PR #794, opened 2016-02-20
ResourceAdapter is broken

klen:
For now, if we have any route which has been added from `router.register_route`, by example any `StaticRoute` then 404 URLs raise 405 (`GET /unknown_url_404 --> 405!`)

It happens because ResourceAdapter **always** returns `allowed_methods`. The PR fixes the problem.

### DIFF
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -223,6 +223,7 @@ class TestUrlDispatcher(unittest.TestCase):
     def test_raise_method_not_found(self):
         handler = self.make_handler()
         self.router.add_route('GET', '/a', handler)
+        self.router.add_static('/static', '/')
         req = self.make_request('GET', '/b')
 
         match_info = self.loop.run_until_complete(self.router.resolve(req))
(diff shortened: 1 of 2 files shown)


tests/test_urldispatch.py
```diff
@@ -715,7 +716,7 @@ def test_resource_adapter_resolve_wildcard(self):
         resource = route.resource
         match_info, allowed = self.loop.run_until_complete(
             resource.resolve('GET', '/path'))
-        self.assertEqual(allowed, {'*'})  # TODO: expand wildcard
+        self.assertEqual(allowed, set())  # TODO: expand wildcard
```
kxepal, line 719: I guess this TODO statement is useful no longer.

klen:
Does nobody care about the error? Actually, for now, AIOHTTP doesn't return 404 as well (for unknown resources). Is it normal?

kxepal:
@klen It could be not a problem. See [webmachine flow](https://raw.githubusercontent.com/basho/webmachine/develop/docs/http-headers-status-v3.png) for example. However, I'm not aware about the underlying idea.

cc @asvetlov

drgarcia1986:
:+1:

asvetlov:
Guys, I'm traveling during last week. Will take a look after returning home.

asvetlov:
I don't that `ResourceAdapter` inherits `Resource`. 

See #826 for proper fix

Issue #797, opened 2016-02-22
Command Line Interface Fails


gledi:
I wanted to test aiohttp and I started by writing a file named `example.py` as shown below:

``` python
from aiohttp import web

async def hi(request):
    name = request.match_info.get('name', 'Anonymous')
    text = 'Hi ' + name
    return web.Response(body=text.encode('utf-8'))

def startapp(args):
    app = web.Application()
    app.router.add_route('GET', '/{name}', hi)
    return app
```

In the docs I saw the section about the [CLI](http://aiohttp.readthedocs.org/en/stable/web.html#command-line-interface-cli) and decide to try that. The example shown in the docs is written as:

``` console
$ python -m aiohttp.web -H localhost -P 8080 package.module.init_func
```

However it wasn't working and I took a look at the code and it turned out that it was using  a `:` to split the module from the callable so I modified my command as follows:

``` console
(aiohttptest)[gledi@testvm aiotest]$ python -d -m aiohttp.web -H localhost -P 8080 example:startapp
usage: aiohttp.web [-h] [-H HOSTNAME] [-P PORT] entry-func
aiohttp.web: error: 'entry-func' not in 'module:function' syntax
```

As `entry-func` actually was in the `module:function` syntax I looked at the code and it turned out that you were passing `sys.argv` to `main`:

``` python
if __name__ == "__main__":
    main(sys.argv)
```

and then later on in `main` after creating a parser you were calling it as such:

``` python
args, extra_args = arg_parser.parse_known_args(argv)
```

However that includes as the first parameter the name of the module (`.... aiohttp/web.py`) and thus it failed to parse it properly. I changed it as follows and it worked for me:

``` python
def main(argv=None):
    if argv is None:
        argv = sys.argv[1:]

...

if __name__ == '__main__':
    sys.exit(main()) # or simply just: main()
```

Also we might need to change the docs so that they use `:` to separate the module from the callable.

asvetlov:
@jashandeep-sohi would you take a look?

jashandeep-sohi:
@asvetlov Yeah, I mixed up `parse_known_args` semantics when called w/ args. Thanks @gledi :bow:

asvetlov:
Fixed by #798

PR #798, opened 2016-02-22
Fix command line arg parsing; issue #797

jashandeep-sohi:
See #797

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -349,7 +349,7 @@ def main(argv):
         type=int,
         default="8080"
     )
-    args, extra_args = arg_parser.parse_known_args(argv)
+    args, extra_args = arg_parser.parse_known_args(argv[1:])
 
     # Import logic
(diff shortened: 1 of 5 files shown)


aiohttp/web.py
```diff
@@ -349,7 +349,7 @@ def main(argv):
         type=int,
         default="8080"
     )
-    args, extra_args = arg_parser.parse_known_args(argv)
+    args, extra_args = arg_parser.parse_known_args(argv[1:])
```
asvetlov: Maybe better to pass `sys.argv[1:]` into `main()` call?

jashandeep-sohi: I don't think it really matters, either is fine. Change it if you like.

asvetlov: Please change. I have a guts feeling that accepting `sys.argv[1:]` is better.
Also I saw the similar approach in past a veral times.

asvetlov:
Thanks!

Issue #792, opened 2016-02-19
ClientSession is leaking cookies across hostnames


panda73111:
Since ClientSession does not link cookies to hostnames, all cookies are sent with every request made using the same session. This leaks cookies across hostnames. Even when using a seperate session for each hostname specific request, the session cookies are sent across HTTP 3xx redirects.

The intuitive fix would be connecting the cookies to their hostname within ClientSession. I'll try to write a patch and test cases if no one has a better idea.

asvetlov:
Yes, the issue describes a very serious bug.
But solution is not so easy.
 I afraid: if aiohttp client should mimic cookie processing by modern browsers it should respect subdomains also and CORS. The first is relative easy but emulating `withCredentials` for CORS is a real nightmare.

gwillem:
Thanks @panda73111 for your PR. I use aiohttp to crawl half the internet and I could not explain why my outbound traffic was 20x inbound, until I ran tcpdump ;-)

asvetlov:
Fixed by Cookie filter #799

PR #799, opened 2016-02-22
Cookie filter

panda73111:
Relevant issue: #792
This branch will add cookie management adhering to [RFC 6265](http://tools.ietf.org/html/rfc6265).
Right now, it adds a simple filter for cookies sent, regarding their "Domain" attribute.

- Cookies passed to `ClientSession` without a "Domain" attribute (as `SimpleCookie()` or `dict`/`tuple`) are still shared across all requests, to keep backwards compatibility.
- Leading dots are ignored (`Domain=.example.com` becomes `Domain=example.com`)
- Cookies with matching domain are sent (`Domain=example.com` is sent to `sub.example.com`, but not the other way around)
- Only shared cookies are sent to addresses that are not a domain, like an IP address

More filters (for the remaining attributes and for storing incoming cookies) will be added as soon as I have the time, until eventually ClientSession behaves like stated in RFC 6265.
CORS is completely new to me and low priority until further research.
This branch can be merged into master at any time, I'll add pull requests as needed and all tests pass.

- [x] The Domain Attribute
- [X] The Secure Attribute
- [X] The Path Attribute
- [X] The Expires Attribute
- [X] The Max-Age Attribute

EDIT 23.02.2016:
- The Domain attribute of received cookies is set to the response hostname if empty, their host-only-flag is set to True
- Cookies from IP addresses are rejected

EDIT 29.02.2016:
- Cookies with secure-flag set are only sent to secure hosts (https://)

EDIT 18.03.2016:
- Cookies with matching path are sent
- If received cookies do not have the path-attribute defined, it is set to the response path

EDIT 20.03.2016:
- Stored cookies are deleted when the time defined in the max-age- or expires-attribute is reached. A timer is set via `loop.call_later()`/`call_at()` when a cookie is added, that removes the cookie from the jar.


aiohttp/client.py
```diff
@@ -353,10 +359,58 @@ def _update_cookies(self, cookies):
...
+                    value["domain"] = domain[1:]
                 dict.__setitem__(self.cookies, name, value)
             else:
                 self.cookies[name] = value
 
+    def _filter_cookies(self, url):
+        """Returns this session's cookies filtered by their attributes"""
+        # TODO: filter by 'expires', 'path', ...
+        netloc = urllib.parse.urlsplit(url).netloc
+        is_ip = IPV4_PATTERN.match(netloc) is not None
```
kxepal: `ipaddress.IPAddress(netloc)` to support both IPv4 and IPv6 in the right way without regexps (which accepts invalid IPv4 addresses).

asvetlov:
Looks interesting. I need reread RFC first for making careful review.
I hope to make it during weekend.

Anyway I have a strong feeling that we need to fix cookies problem in the next aiohttp release.

panda73111:
I added the Expires- and Max-Age-attributes, by scheduling deletion of the cookie via `call_later()`/`call_at()`. But this gets problematic as soon as the user changes these attributes, the cookie would still be removed at the original timestamp. How should this be handled? For "Expires", an absolute date, the cookie could just remain in the cookie jar, and just not be sent if the time of sending the request is after the date defined in the "Expires" attribute. But "Max-Age" only consists of a time span in seconds, of which it is not possible to know when it has been set by the user.

panda73111:
...Or maybe any other comments? This is basically done, as it is.

auvipy:
you could fix the merge conflicts and failing builds

panda73111:
Well, I wanted my code to get approved first, since I don't want to create a merge commit every time master becomes incompatible, until this gets merged or rejected eventually.

And what's up with these coverage comments? Shouldn't these be posted only after all tests pass? The next commit makes it up to 98.26% again only because flake8 is satisfied...
(Also, coveralls doesn't let me view 'changed' lines, because "The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.")

KostyaEsmukov:
Any update on this?
@asvetlov @fafhrd91

aiohttp/helpers.py
```diff
@@ -496,3 +506,257 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+        if cookie_path.endswith("/"):
+            return True
+
+        non_matching = req_path[len(cookie_path):]
+
+        return non_matching.startswith("/")
+
+    @classmethod
+    def _parse_date(cls, date_str):
```
fafhrd91, line 696: can't we reuse some existing function?

fafhrd91:
i am +1 for this change, but we need opinion from @asvetlov before inclusion.

asvetlov:
Squashed and merged as 0016999
@panda73111 thanks you very much for your hard work

PR #803, opened 2016-02-25
Fixed bug in serving static directory

realcr:
When using app.add_static(...) to serve a directory, if the root of the directory is requested, the connection between the server and client is crashed.
This happens because the server tries to read a directory using open with 'rb' and fails.

The fix first makes sure that the requested resource is a file. If it is not a file, it returns a 404 error. I think this is the correct behaviour, feel free to change it if you think it should be something else.

I also include a relevant test case, to avoid regressions with this problem.

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -571,7 +571,7 @@ class StaticRoute(Route):
 
         # Make sure that filepath is a file
         if not filepath.is_file():
-            raise HTTPNotFound() 
+            raise HTTPNotFound()
 
         st = filepath.stat()
 
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -569,6 +569,10 @@ class StaticRoute(Route):
             request.logger.exception(error)
             raise HTTPNotFound() from error
 
+        # Make sure that filepath is a file
+        if not filepath.is_file():
+            raise HTTPNotFound() 
+
         st = filepath.stat()
 
         modsince = request.if_modified_since
--- a/tests/test_web_urldispatcher.py
+++ b/tests/test_web_urldispatcher.py
@@ -27,23 +27,6 @@ def run_timeout(cor, loop, timeout=ASYNC_TEST_TIMEOUT):
 
 
 @pytest.fixture(scope='function')
-def tloop(request):
-    """
-    Obtain a test loop. We want each test case to have its own loop.
-    """
-    # Create a new test loop:
-    tloop = asyncio.new_event_loop()
-    asyncio.set_event_loop(None)
-
-    def teardown():
-        # Close the test loop:
-        tloop.close()
-
-    request.addfinalizer(teardown)
-    return tloop
-
-
-@pytest.fixture(scope='function')
 def tmp_dir_path(request):
     """
     Give a path for a temporary directory
@@ -60,7 +43,7 @@ def tmp_dir_path(request):
     return tmp_dir
 
 
-def test_access_root_of_static_handler(tloop, tmp_dir_path, unused_port):
+def test_access_root_of_static_handler(loop, tmp_dir_path, unused_port):
     """
     Tests the operation of static file server.
     Try to access the root of static file server, and make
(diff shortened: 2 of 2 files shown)


tests/test_web_urldispatcher.py
```diff
@@ -0,0 +1,103 @@
+import pytest
+import tempfile
+import aiohttp
+from aiohttp import web
+import os
+import shutil
+import asyncio
+
+SERVER_HOST = '127.0.0.1'
+SERVER_PORT = 8080
```
asvetlov: Please use `unused_port` fixture here.
See https://github.com/KeepSafe/aiohttp/blob/90f16168fd8510fe930994c4b4170a18cd8f2568/tests/test_run_app.py#L35-L36 for inspiration.

aiohttp/web_urldispatcher.py
```diff
@@ -569,6 +569,10 @@ def handle(self, request):
             request.logger.exception(error)
             raise HTTPNotFound() from error
 
+        # Make sure that filepath is a file
+        if not filepath.is_file():
+            raise HTTPNotFound()
```
asvetlov: flake8 tool blames that you have trailing whitespaces in the line.

asvetlov:
Thanks! I'll merge after you'll fix two mentioned tiny comments.

realcr:
I added the required modifications. Thank you for teaching me about flake8 :)

tests/test_web_urldispatcher.py
```diff
@@ -0,0 +1,104 @@
...
+
+# Timeout in seconds for an asynchronous test:
+ASYNC_TEST_TIMEOUT = 1
+
+
+class ExceptAsyncTestTimeout(Exception):
+    pass
+
+
+def run_timeout(cor, loop, timeout=ASYNC_TEST_TIMEOUT):
```
asvetlov: Do you really need the function? What the purpose of timeout parameter -- looks like you don't use it.

tests/test_web_urldispatcher.py
```diff
@@ -0,0 +1,104 @@
...
+    with open(my_file_path, 'w') as fw:
+        fw.write('hello')
+
+    asyncio.set_event_loop(None)
+    app = web.Application(loop=tloop)
+    # Register global static route:
+    app.router.add_static('/', tmp_dir_path)
+
+    @asyncio.coroutine
+    def inner_cor():
```
asvetlov: I suggest to get rid of inner function.

Declare `test_access_root_of_static_handler` as 

``` @pytest.mark.run_loop
def test_access_root_of_static_handler(...):
```

It allows to use `yield from` statements inside the test function.

tests/test_web_urldispatcher.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def teardown():
+        # Delete the whole directory:
+        shutil.rmtree(tmp_dir)
+
+    request.addfinalizer(teardown)
+    return tmp_dir
+
+
+def test_access_root_of_static_handler(tloop, tmp_dir_path, unused_port):
```
asvetlov: Replace `tloop` with existing `loop` fixture.

tests/test_web_urldispatcher.py
```diff
@@ -0,0 +1,104 @@
...
+        fw.write('hello')
+
+    asyncio.set_event_loop(None)
+    app = web.Application(loop=tloop)
+    # Register global static route:
+    app.router.add_static('/', tmp_dir_path)
+
+    @asyncio.coroutine
+    def inner_cor():
+        handler = app.make_handler()
```
asvetlov: Use `create_app_and_client` fixture https://github.com/KeepSafe/aiohttp/blob/master/tests/conftest.py#L232-L250 for test server starting/stopping.

tests/test_web_urldispatcher.py
realcr: It is used for making sure the test doesn't run too long. Do you have other mechanism to do this in the code base? If not, maybe we can add it in a more global location, to be used with other tests. The ASYNC_TEST_TIMEOUT is the default timeout.

tests/test_web_urldispatcher.py
realcr: Done.

tests/test_web_urldispatcher.py
asvetlov: Got it.
Right now aiohttp test suite has no such feature. I didn't feel the need for it.
Let's add the feature later to hooks that provide `@pytest.mark.run_loop` functionality.

realcr:
I fixed the test case. It now uses mark.run_loop and create_app_and_client, so it is much shorter.

asvetlov:
Thanks!

PR #806, opened 2016-02-28
Add issue and pull request GitHub templates

kxepal:
Following the new GitHub [feature](https://github.com/blog/2111-issue-and-pull-request-templates). I didn't found any good or standard content for these (I was like the one that Google Code had, but those get lost by known reasons), so it's open for discussion and improvements (:

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,28 @@
+<!-- Thank you for your contribution!
+
+     Before submit your Pull Request, make sure you picked
+     the right branch:
+
+     - If you propose a new feature or improvement, select "master"
+       as a target branch;
+     - If this is code bug fix or documentation correction, select
+       the lastest release branch (which looks like "0.xx") -->
+
+## What these changes does?
+
+<!-- Please give a short brief about these changes. -->
+
+## How to test your changes?
+
+<!-- Describe how we can test your changes if they provides
+     any notable behaviour for the end users. -->
+
+## Related issue number
+
+<!-- Is there any issue opened that get fixed by this? -->
+
+## Checklist
+
+- [ ] Code is written and well
+- [ ] Tests for the changes are provided
+- [ ] Documentation reflects the changes
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,23 @@
+## Long story short
+
+<!-- Please describe your problem and why the fix is important. -->
+
+## Expected behaviour
+
+<!-- What the behaviour did you expect? -->
+
+## Actual behaviour
+
+<!-- What's actually happened? -->
+
+## Steps to reproduce
+
+<!-- Please describe steps to reproduce the issue.
+     If you have a script that does that please include it here within
+     markdown code markup -->
+
+## Your environment
+
+<!-- Describe the environment you have that leed to your issue.
+     This includes project version, OS, proxy server and else bits that
+     are related to your case. -->


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+## Related issue number
+
+<!-- Is there any issue opened that get fixed by this? -->
+
+## Checklist
+
+- [ ] Code is written
+- [ ] Tests are added
+- [ ] Documentation reflects your changes
+- [ ] CHANGES.txt is updated
```
asvetlov: I don't want to force the bullet.
Updating CHANGES.txt is not trivial: the file most likely conflicts with other commits.
For now I update CHANGES after every merge manually. Let's keep it up by committers.

In CPython resolving conflicts in CHANGES is the most pain in the ass.
IIRC @brettcannon proposed to collect CHANGES from CPython issue tracker -- it's cool but we don't need it for such small projects as aiohttp with 4 guys with commit rigths for now.

asvetlov:
I'm ok with putting files in the root directory. Otherwise we should relocate CONTRIBUTING also (and change include path for it in the docs.

Maybe adding mentions of well-formed issue and pull request to [CONTRIBUTING](http://aiohttp.readthedocs.org/en/stable/contributing.html) page makes sense?

Also I would like to point base branch for pull request.

After every aiohttp major release we have a list of bugs and regressions which may me solved in bugfix releases (we publish a next major release in two months usually). 
For bugfixes I used to create a branch named like `0.22` on demand but may create a branch for bugfixes on every release (and drop `0.22` after releasing `0.23`).

I suggest adding a guess for making PR basing on bugfix branch if the change doesn't introduce new feature but it's a relative small fix or documentation correction.
Our docs are updated on stable branch changes (last github tag literally), so early doc fixes are important.

Thoughts?

.github/PULL_REQUEST_TEMPLATE.md
brettcannon: So there are two proposals; one is to collect CHANGES entries from the issue tracker, while the other is to have a separate file per CHANGES entry.

kxepal: Fair points. Removed.

kxepal:
> I'm ok with putting files in the root directory. Otherwise we should relocate CONTRIBUTING also (and change include path for it in the docs).

I'm not sure about this. While these two templates are GitHub-related files, the CONTRIBUTING is an important project-related file. Hiding it within some hidden directory wouldn't increase number of his readers while keeping service files (templates are kind of that) will only pollute "global namespace".

So, the lesser evil would be put templates into root directory, but not move CONTRIBUTING anywhere.

> Maybe adding mentions of well-formed issue and pull request to CONTRIBUTING page makes sense?

Idea is good, but who will read them and follow the example? With cross reference from a template that would make a sense. Do you have anything on your mind?

> Also I would like to point base branch for pull request.
> After every aiohttp major release we have a list of bugs and regressions which may me solved in bugfix releases (we publish a next major release in two months usually).
> For bugfixes I used to create a branch named like 0.22 on demand but may create a branch for bugfixes on every release (and drop 0.22 after releasing 0.23).
> I suggest adding a guess for making PR basing on bugfix branch if the change doesn't introduce new feature but it's a relative small fix or documentation correction.
> Our docs are updated on stable branch changes (last github tag literally), so early doc fixes are important.

Suddenly, there is no way to specify default PR branch, but it's possible to put few words about in PR template at the top. How to describe that nicely?

asvetlov:
> So, the lesser evil would be put templates into root directory, but not move CONTRIBUTING anywhere.
> Please do.
> 
> > Maybe adding mentions of well-formed issue and pull request to CONTRIBUTING page makes sense?
> > Idea is good, but who will read them and follow the example? With cross reference from a template that would make a sense. Do you have anything on your mind?
> > Yes, I mean cross references. Really it's not very important though.
> 
> Suddenly, there is no way to specify default PR branch, but it's possible to put few words about in PR template at the top. How to describe that nicely?
> I suggest something like "select master base branch for new features or latest released branch 0.xx for bugfix or documentation correction".
> 0.xx is not very obvious but I cannot pick up better words.

kxepal:
@asvetlov Done.

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,28 @@
...
+     any notable behaviour for the end users. -->
+
+## Related issue number
+
+<!-- Is there any issue opened that get fixed by this? -->
+
+## Checklist
+
+- [ ] Code is written and well
+- [ ] Tests for the changes are provided
```
asvetlov, line 27: Please add a bullet about running flake8 check -- people often forget about it.

kxepal, line 27: I mean this by `Code is written and *well*`. And flake check is a part of `make test` which is recommended in CONTRIBUTING as the way to test your changes. Moreover, TravisCI runs flake checks. So, if you check these two point with flake issues then this only means that you didn't test anything or made it in own way that may be not a part of common workflow.

asvetlov, line 27: Ok

asvetlov:
LGTM. Please merge

kxepal:
Thanks!

popravich:
IMHO these files can be put into `.github` dir not to mess in project's root

PR #814, opened 2016-03-05
UrlDispatcher.add_route with partial coroutine handler

banthonywalker:
fix for problem when partialy applied coroutine is used as handler for UrlDispatcher.add_route

similar issue:
http://bugs.python.org/issue23519

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -93,7 +93,13 @@ class AbstractRoute(abc.ABC):
               issubclass(handler, AbstractView)):
             pass
         else:
-            handler = asyncio.coroutine(handler)
+            @asyncio.coroutine
+            def handler_wrapper(*args, **kwargs):
+                result = handler(*args, **kwargs)
+                if asyncio.iscoroutine(result):
+                    result = yield from result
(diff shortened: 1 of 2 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -93,7 +93,13 @@ def __init__(self, method, handler, *,
               issubclass(handler, AbstractView)):
             pass
         else:
-            handler = asyncio.coroutine(handler)
+            @asyncio.coroutine
+            def handler_wrapper(*args, **kwargs):
+                result = handler(*args, **kwargs)
+                if asyncio.iscoroutine(result):
```
asvetlov, line 99: Please add a check for `asyncio.Future` also.

asvetlov:
Tests needed.

aiohttp/web_urldispatcher.py
ghost, line 99: according to current documentation (for UrlDispatcher.add_route and Resource.add_route) handler should be callable which excludes asyncio.Future

ghost:
test added

asvetlov:
thanks

Issue #728, opened 2016-01-10
Add optional support for aiodns to TCPConnector for async DNS resolving


asvetlov:
See http://stackoverflow.com/questions/34700311/how-to-handle-dns-timeouts-with-aiohttp for reasoning

gwillem:
Thanks for your suggestions! To be solved:
1. The HTTP host header needs to be overwritten (easy)
2.  SSL verification requires the true hostname. I have dug in the code but not found a way yet to cleanly pass the hostname to the SSL context ([relevant code here](https://github.com/python/cpython/blob/6c223d811812f36ac38ec19e2f59d68f2336a0e4/Lib/ssl.py#L254))

asvetlov:
`TCPConnector._resolve` https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/connector.py#L531 performs DNS lookup if `use_dns_cache` for `TCPConnector` http://aiohttp.readthedocs.org/en/stable/client_reference.html#tcpconnector is `True`.
I suggest adding new ctor parameter `use_aiodns` (also `False` by default) to resolve through aiodns if `use_dns_cache` is `True`.

fafhrd91:
-1 for `use_aiodns`, better is to invert dependency.
we need Resolver object, and pass that object as parameter to connector, then we can create default resolver that uses asyncio and aiodns one.

asvetlov:
Yes, `Resolver` sounds better than `use_aiodns`.
Thanks for correction.

asvetlov:
Fixed by #819

PR #747, opened 2016-01-24
Aiodns support

gwillem:
Fixes #728. Todo:

- [x] Mock DNS tests
- [x] Add docs

PR #819, opened 2016-03-07
Aiodns support

lphuberdeau:
## What these changes does?

Allows to use an alternate method for DNS resolution. The default resolution remains to the getaddrinfo system call using thread processors.

Alternate technique is to use aiodns instead, which allows to specify the nameservers to use while resolving. aiodns allows for higher concurrency.

## Related issue number

* Resolves issue #728 
* Follow-up to PR #747

## Checklist

- [x] CI testing works under windows


gwillem:
NB Another benefit of aiodns is that is allows higher concurrency when crawling the Internet. So 100-200 crawlers per core should be possible.

lphuberdeau:
Seems like the issue with appveyor/c-ares is limited to py35-64.

aiohttp/abc.py
```diff
@@ -55,3 +55,15 @@ def __iter__(self):
         @abstractmethod
         def __await__(self):
             return
+
+
+class AbstractResolver(metaclass=ABCMeta):
```
asvetlov: Just inherit from `ABC` like other classes from the module do.

aiohttp/abc.py
```diff
@@ -55,3 +55,15 @@ def __iter__(self):
...
+
+    @asyncio.coroutine
+    @abstractmethod
+    def resolve(self, hostname):
+        """Return IP address for given hostname"""
+
+    @abstractmethod
+    def close(self):
```
asvetlov, line 69: Please make `close` a coroutine -- my guts feel we will need it sometimes.

lphuberdeau:
@asvetlov Requested changes are applied

asvetlov:
Looks good but please fix a merge conflict first.

tests/test_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+    aiodns = None
+
+
+class FakeResult:
+    def __init__(self, host):
+        self.host = host
+
+
+@pytest.fixture
+def loop():
```
jettify: I think `loop` fixture from `conftest.py` could be used here

lphuberdeau:
@asvetlov Branch is now up to date, using the common loop fixture.

asvetlov:
Thanks

PR #834, opened 2016-03-19
Allow to pass None as a timeout value to disable timeout logic

alexkey:
## What these changes does?
Allows to pass `None` as a timeout value to disable timeout logic.

In some cases there is needed to disable timeout logic but keep it in convenient code structure without ugly if-branching like this:

```
if delay:
    with aiohttp.Timeout(delay):
            pass  # ...
else:
        pass  # ...
```

Now it is able to use the same code for both numerical and `None` timeout values:

```
for delay in [None, 0, 0.1, 1]:
    with aiohttp.Timeout(delay):
            pass  # ...

```

This behaviour is similar to `asyncio.wait_for(fut, timeout, *, loop=None)` coroutine, which takes `None` as a timeout value to block until the future completes.

## How to test your changes?

There is a test in `test_timeout.py`: `test_timeout_disable`.

## Related issue number

None.

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -477,7 +477,7 @@ class Timeout:
         if self._task is None:
             raise RuntimeError('Timeout context manager should be used '
                                'inside a task')
-        if self._timeout:
+        if self._timeout is not None:
             self._cancel_handler = self._loop.call_later(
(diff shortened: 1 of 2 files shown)


asvetlov:
Thanks!

Issue #820, opened 2016-03-10
Problems with HEAD responses


lgelo:
RFC7231 forbids sending body content when responding to HEAD requests. Believing that servers are RFC compliant, aiohttp client is not parsing body of HEAD responses. Content of body stays in buffer and when parsing next response from the same session/connection BadStatusLine is raised. Not all servers are compliant, and unfortunately aiohttp server is one of them. Example code is bellow.

I believe the correct behaviour should be according to Postel's law: server won't send body in response to HEAD and will read and discard (maybe with warning) non-compliant responses.

``` python
#!/usr/bin/env python3.5

import asyncio
import aiohttp
from aiohttp import web

async def client(session):
    resp = await session.head('http://127.0.0.1:5000/')
    print("%d: '%s'" % (resp.status, await resp.text()))
    resp.release()

    # so far so good - try again
    resp = await session.get('http://127.0.0.1:5000/')
    print("%d: '%s'" % (resp.status, await resp.text()))
    await resp.release()

async def server(loop):
    app = web.Application()
    handler = app.make_handler()
    srv = await loop.create_server(handler, '127.0.0.1', 5000)
    return srv, handler

loop = asyncio.get_event_loop()
session = aiohttp.ClientSession(loop=loop)

srv, handler = loop.run_until_complete(server(loop))

client = loop.run_until_complete(client(session))
session.close()

loop.stop()
loop.close()
```

asvetlov:
Agree with all your proposals.
Would you make a Pull Request? Or two separate PRs -- one for client and other for server?

lgelo:
Probably just one PR. Proposed fix for the server (web_reqrep.py):

``` python

    def should_send_body(self):
        return (self._req.method.lower() != 'head' and 
               self._status not in [204,304])

    @asyncio.coroutine
    def write_eof(self):
        try:
            body = self._body
            if body is not None and self.should_send_body():
                self.write(body)
        finally:
            self.set_tcp_nodelay(True)
        yield from super().write_eof()
```

I don't know if it is possible to do something  on the client side, because server can set Content-length header to "what it  would have been if the request method had been GET" so there's no way to tell how much data (if any) should be read.

asvetlov:
User still may write

```
resp = web.StreamResponse(status=204)
await resp.prepare(request)
resp.write(b'BODY')
return resp
```

Honestly I not sure if we need to do anything on server side.

lgelo:
[...]

PR #838, opened 2016-03-21
Don't send body in response to HEAD request

lgelo:
## What these changes does?

Skips sending Response body when request method is HEAD or status code is 204 or 304. Satisfies requirements of RFC7230 an RFC7231 that server MUST NOT send message body in such responses.

## How to test your changes?

Test case is included: 
test_head_returns_empty_body (tests/test_web_functional.py)

## Related issue number

Problems with HEAD responses #820 

## Checklist

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -851,11 +851,15 @@ class Response(StreamResponse):
 
         self.body = text.encode(self.charset)
 
+    def should_send_body(self):
+        return (self._req.method.lower() != 'head' and
+                self._status not in [204, 304])
+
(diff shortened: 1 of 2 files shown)


aiohttp/web_reqrep.py
```diff
@@ -851,11 +851,15 @@ def text(self, text):
 
         self.body = text.encode(self.charset)
 
+    def should_send_body(self):
+        return (self._req.method.lower() != 'head' and
```
popravich: No need to convert to lowercase, please use `hdrs.METH_HEAD` instead

fafhrd91:
thanks

PR #847, opened 2016-04-10
Respect CONNECT method to implement a proxy server

drslump:
To be able to implement an HTTP tunneling proxy the payload received on the app logic should be wired to the actual stream and not the `EMPTY_PAYLOAD` one. Of course it would still need to modify the parser in the same way the WebSocket implementation works.

Also added some hardening for the `Content-Length` header being `"0"` or some other weird value.


asvetlov:
Please add a test

drslump:
tests added

PR #856, opened 2016-05-02
'request.logger' > 'request.app.logger' in StaticFileMixin

samuelcolvin:
(I believe this error exists on master too, but i'm using 0.21 as base as per the instructions).

## What these changes does?

prevent 500 exception with extremely long url requests to the static file router.

This is a slightly ugly way of getting the error to occur, presumably there are other causes of a suitable exception, but this how I found the error and seems to demonstrate the point.

The code causing the error was not previously covered.

## How to test your changes?

see test below

## Checklist

- [ ] Code is written and well
- [ ] Tests for the changes are provided

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -566,7 +566,7 @@ class StaticRoute(Route):
             raise HTTPNotFound() from error
         except Exception as error:
             # perm error or other kind!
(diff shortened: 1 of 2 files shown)


samuelcolvin:
@asvetlov any chance of merging this? I see it's still broken in master too.

asvetlov:
Thanks

PR #869, opened 2016-05-08
fixed request.json() method to properly return json object

rkyleg:
previously request.json() failed because request.text() returned a querystring from form data and json.loads expects a proper json string


popravich:
This is invalid behavior! `request.json()` method expects that body is a json data not urlencoded data.
You must use `request.post()` to get 'x-www-form-urlencoded` data.
http://aiohttp.readthedocs.io/en/stable/web.html#file-uploads

PR #871, opened 2016-05-09
If send fails, websocket is not closed

MarcoSulla:
The WebSocketResponse has a _conn_lost attribute that is not augmented if the asyncio's writer does fail to send data. I added a patch, but IMHO asyncio should raise an error instead of logging a simple warning.

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -118,13 +118,21 @@ class WebSocketResponse(StreamResponse):
 
     def exception(self):
         return self._exception
-
+    
+    def check_con(self):
+        if self._conn_lost >= THRESHOLD_CONNLOST_ACCESS:
+            raise RuntimeError('WebSocket connection is closed.')
+    
     def ping(self, message='b'):
         if self._writer is None:
             raise RuntimeError('Call .prepare() first')
         if self._closed:
             raise RuntimeError('websocket connection is closing')
+        
+        conn_lost_start = self._writer.writer._conn_lost
         self._writer.ping(message)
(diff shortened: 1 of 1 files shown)


Marco-Sulla:
Furthermore, I think `ConnectionResetError` should be raised, not `RuntimeError`.

PS: Travis check fails because of W293. I hope you could ignore that warning.

Marco-Sulla:
Excuse me, why you closed it?

asvetlov:
Hmm, github has closed the issue because I've deleted 0.21 branch :(

asvetlov:
After thinking on your patch I cannot accept it, sorry.
`self._writer._conn_lost` is very private api, say uvloop doesn't support it.

Regarding to changing from `RuntimeError` to `ConnectionResetError` -- the last assumes that `errno` code is `ECONNRESET` which is not true for our case.
Using `aiohttp.ClientDisconnectedError` makes more sense.

asvetlov:
What's wrong with current implementation?
You'll get disconnection error on next loop iteration, isn't it?

Marco-Sulla:
No errors, I get warnings from `asyncio` and websocket is never closed.

What about this?

```
    def check_con(self):
        conn_lost_tot = self._conn_lost + self._writer._conn_lost

        if conn_lost_tot >= THRESHOLD_CONNLOST_ACCESS:
            raise RuntimeError('WebSocket connection is closed.')
```

asvetlov:
Sorry for late answer.
Say again, `self._writer._conn_lost` is not public API, we shouldn't rely on it.
E.g. it's not exposed by uvloop IFAIK.

PR #874, opened 2016-05-12
Skip DNS resolving if provided host is already an ip address

lphuberdeau:
When using the AsyncResolver, there is a DNSError when trying to access a url where the host is an ip address.

This patch skips the resolver step if the host is an ip address.

An alternative would have been to move the same patch to the AsyncResolver, but other resolvers would likely have the same issue. Also, there was also a condition formatting the result properly in the TCPConnector.

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -1,5 +1,6 @@
 import asyncio
 import aiohttp
+import ipaddress
 import functools
 import http.cookies
 import ssl
@@ -554,7 +555,7 @@ class TCPConnector(BaseConnector):
 
     @asyncio.coroutine
     def _resolve_host(self, host, port):
-        if not self._use_resolver:
+        if not self._use_resolver or self._host_is_ip(host):
             return [{'hostname': host, 'host': host, 'port': port,
                      'family': self._family, 'proto': 0, 'flags': 0}]
 
@@ -572,6 +573,13 @@ class TCPConnector(BaseConnector):
             res = yield from self._resolver.resolve(
                 host, port, family=self._family)
             return res
+
+    def _host_is_ip(self, host):
+        try:
+            ipaddress.ip_address(host)
+            return True
+        except ValueError:
+            return False
 
     @asyncio.coroutine
     def _create_connection(self, req):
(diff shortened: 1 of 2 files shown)


aiohttp/connector.py
```diff
@@ -573,6 +574,13 @@ def _resolve_host(self, host, port):
                 host, port, family=self._family)
             return res
 
+    def _host_is_ip(self, host):
+        try:
+            ipaddress.ip_address(host)
+            return True
+        except ValueError:
+            return False
```
popravich, line 582: Everybody most of their time use host names not IP addresses,
so most of time this method will raise and except ValueError what is not very efficient.

lphuberdeau, line 582: Anything else to suggest? The ipaddress module does not have a way to verify the format while avoiding the exception. Custom-made validation does not seem like a better route.

kaniini:
For this, we have:

``` py
class IPAwareAsyncResolver(AsyncResolver):
    @asyncio.coroutine
    def resolve(self, host, port=0, family=socket.AF_INET):
        if not host or family not in {socket.AF_UNSPEC, socket.AF_INET, socket.AF_INET6}:
            return []

        if hasattr(socket, 'inet_pton'):
             if family == socket.AF_UNSPEC:
                 afs = [socket.AF_INET, socket.AF_INET6]
             else:
                 afs = [family]

             for af in afs:
                 # Linux's inet_pton doesn't accept an IPv6 zone index after host,
                 # like '::1%lo0', so strip it. If we happen to make an invalid
                 # address look valid, we fail later in sock.connect or sock.bind.
                 try:
                     if af == socket.AF_INET6:
                         socket.inet_pton(af, host.partition('%')[0])
                     else:
                         socket.inet_pton(af, host)
                     return [{'hostname': host, 'host': host, 'port': port, 'family': family, 'proto': 0, 'flags': socket.AI_NUMERICHOST}]
                 except OSError:
                     pass

        # ... if we reach here, this means it's not a valid IP.  try a DNS lookup.
        hosts = yield from super(IPAwareAsyncResolver, self).resolve(host, port, family)
        return hosts
```

asvetlov:
Fixed by #905

PR #879, opened 2016-05-17
Fix the code to use event loop signaling API

1st1:
## What does this change do?

This PR fixes how gunicorn workers set up signals.  In the current version, `GunicornWebWorker` implements an `init_signal` method, but `gunicorn.Worker` doesn't have it!  It has an `init_signals` method.

Everything did work by a pure chance -- `gunicorn.Worker` called its own `init_signals`, which used `signal` module to set up signal handlers (instead of using `loop.add_signal_handler`), which `asyncio` actually doesn't mind (*but it's a bad thing, nevertheless!*)

The existing unit-test is synthetic, it manually calls the wring `init_signal`.

## Related issue number

No related issue, but I've discovered this while investigating https://github.com/MagicStack/uvloop/issues/26.

## Checklist

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -109,14 +109,26 @@ class GunicornWebWorker(base.Worker):
 
         yield from self.close()
 
-    def init_signal(self):
-        # init new signaling
-        self.loop.add_signal_handler(signal.SIGQUIT, self.handle_quit)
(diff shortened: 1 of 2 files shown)


fafhrd91:
thanks

ludovic-gasc:
Hi @1st1, thanks for this pull request, I've also applied this bug fix in API-Hour worker.

Issue #863, opened 2016-05-05
aiohttp support for http/2 protocol


catroot:
Is there any activity to implement http/2 capability to aiohttp or any reason to not implement it?

samuelcolvin:
aiohttp already works fine with nginx-1.10.0 over http2. That should cover much of http2. I guess aiohttp will take some modification to support "Server push", but even nginx [doesn't support that yet](https://www.nginx.com/blog/nginx-1-9-5/).

catroot:
I'm glad, that nginx supports partial http2 features. They takes their own way.
The point is to have some info about native support of http2 for aiohttp.
It will be nice if @asvetlov takes care of them as a developer of http stack for asyncio.
Because aiohttp makes a good pretension to be a favorite python http server for a long time (imho).
Without support for http2 the solution doesn't seems to be completed.
I can't get the point why http2 stays out of focus of developing http stack for asyncio, eventough all browsers already have support for them at their current versions.

ludovic-gasc:
I'm not an aiohttp core dev, this post only represent my personal opinion:

I already spent a lot of my (personal) time only to understand correctly HTTP/1.1 and have enough understanding in aiohttp to put asyncio daemons on production.
I can't imagine the time necessary to implement aiohttp, not only to be an advanced user.

Now, I can deliver and debug daemons quickly with this stack, and for now, it's enough efficient for our needs.
What's the incentive to break everything, relearn at least HTTP/2 debugging, if with HTTP/1.1 + WebSockets already cover the needs ?
Yes, for a company like Google, HTTP/2 is clearly a very good investment, but for a simple human like me who wants feed his family, I'm pretty happy with the features of aiohttp, and it's also the case for my clients ;-)

However, don't misinterpret me: If you want HTTP/2, just code it. As a pull request of aiohttp or a new project, just code it, I'll pretty happy to test.
But, to my understanding, it isn't a small time investment, you need to reimplement a lot of things.
And in case you are interested in to give money for an implementation, certainly people could be interested in to help you.

fafhrd91:
from my production experience i found that frame based protocols are very slow if parser is written in python. so to make aiohttp work well with http/2 we really need c-module for that.

Lukasa:
@fafhrd91 That's not really true.

Frame based protocols have the potential to be _substantially_ faster than line-based protocols written in Python if care is taken with the way the library is designed.

[...]

Issue #320, opened 2015-04-13
HTTP/2 support


ludovic-gasc:
HTTP/2 support isn't our priority for now, at least for me.

Nevertheless, for your information, some people works on HTTP/2 support in CPython: http://bugs.python.org/issue23794

fafhrd91:
`http.client` architecture is quite pure, i used it in very first implementation of tulip.http,
i'm not sure we can use any of `http.client` code in aiohttp.

i have a lot of experience with frame based protocols on top of aiohttp, but thats require a lot of time for implementation. also i'm very concerned about any feature that is not being developed for production use.

ludovic-gasc:
Yes, I'm fully agree with you: as I said in the first line of my issue, it isn't my priority because it isn't production ready yet.

Lukasa:
In the event that you do start caring, you may find [Hyper-h2](http://python-hyper/h2) to be useful for this.

sametmax:
And you got prior art a aio wrapper for inspiration : https://aioh2.readthedocs.org/en/latest/readme.html

PR #882, opened 2016-05-20
[RFC] Discussion PR for HTTP/2 support (DO NOT MERGE).

Lukasa:
## What these changes does?

This is an example of _the most basic_ HTTP/2 support in aiohttp. This is nowhere near close to production ready: instead, it's intended to spur a longer discussion about how best to add HTTP/2 support to aiohttp. This code does the following:
- Allows spinning up a HTTP/2-only aiohttp web server that cannot serve static files.

This code does not support any of the following:
- HTTP/2 flow control
- HTTP/2 stream priority
- HTTP/2 server push
- Serving static files
- 100 Continue
- Correct error pages (404 pages still serve using HTTP/1.1, which breaks things)
- Anything remotely tricky
- Basically, it's not RFC 7540 compliant, or even really all that close

I opened this because I believe that if aiohttp wants to support HTTP/2 it requires a very substantial change, and I did not want to work on that in isolation without discussing it with the aiohttp community. However, it's always tough to discuss such a change without having some concrete code to look at, so this provides that code.
## How to test your changes?

The following test file is a variation on the example from the documentation, and can be tested using the Python hyper library.

This is the server file:

``` python

from aiohttp import web, web2, web_reqrep2


async def handle(request):
    text = "Hello, " + name
    return web_reqrep2.Http2Response(body=text.encode('utf-8'))


app = web.Application(handler_factory=web2.Http2RequestHandlerFactory)
app.router.add_route('GET', '/{name}', handle)

web.run_app(app)
```

To test it, install the Python hyper library (`pip install hyper`) and then run `hyper --h2 GET http://localhost:8080/name`.
## Related issue number
#863, #320
## Discussion

When trying to implement HTTP/2 support in aiohttp, I discovered that aiohttp's architecture and API make it remarkably tricky to plumb HTTP/2 support through in a coherent and sensible way. The biggest problems here are:
1. Lack of encapsulation. A large swathe of the aiohttp code-base knows quite a lot about the wire-format of HTTP/1.1, rather than confining that knowledge to the `ServerHttpProtocol` and `*Parser` classes. For example, `web_urldispatcher` contains a `_defaultExpectHandler` that writes directly to the transport rather than passing that write through the `ServerHttpProtocol` or equivalent object. More generally, `HttpMessage` and friends all write directly to the transport, which means that to make HTTP/2 work we need to propagate quite a lot of state through aiohttp to get it from the `ServerHttp2Protocol` to those objects.
2. Subclassing internally. The aggressive use of subclassing within aiohttp makes it quite difficult to locate all the places that need changing. Ideally, all objects in aiohttp would direct their writes to the transport via the `ServerHttpProtocol`, but because there are so many different objects that have access to the backing transport (and also the reader/writer objects) it is very difficult to enforce that constraint. Additionally, for HTTP/2 they'd all need a correlator (their stream ID), which makes a further substantial change to the codebase. That wouldn't be too bad, except...
3. Subclassing as an API. aiohttp quite widely uses subclassing as an API: for example, `ServerHttpProtocol` and friends all expect to be subclassed. Because of this, there is likely quite a lot of third-party code that subclasses aiohttp's classes and reaches inside of them. That makes it very difficult to do a substantial refactor to the code. If aiohttp was unwilling to break that code, they'd need either to do a big breaking semver-major change (which will annoy a lot of people) or write an entirely parallel stack that is capable of doing HTTP/1.1 and HTTP/2 transparently and then ask people to migrate to that. Either seems like quite a lot of work.

Basically, from where I'm sat I think that adding HTTP/2 support transparently into aiohttp requires a very substantial change to the library, and I'm unwilling to do that without discussing with the community and maintainers first. Is there any interest in having such a substantial change made to aiohttp, or do the maintainers and community believe that HTTP/2 support is not worth it?

### DIFF
--- a/aiohttp/protocol2.py
+++ b/aiohttp/protocol2.py
@@ -1,0 +1,175 @@
+"""Http2 related parsers and protocol."""
+
+from wsgiref.handlers import format_date_time
+
+import aiohttp.hdrs
+import aiohttp.protocol
+
+HttpVersion20 = aiohttp.protocol.HttpVersion(2, 0)
+
+
+class HTTP2Parser:
+    def __init__(self, conn):
+        self._conn = conn
+
+    def __call__(self, out, buf):
+        """
+        Receives data from the buffer, parses it, and then sends any events
+        out.
+
+        This only terminates when the connection is terminated by the remote
+        peer.
+        """
+        while True:
+            # XXX: 65kb is totally arbitrary here: consider tuning.
+            data = yield from buf.readsome(size=65535)
+
+            if not data:
+                out.feed_eof()
+                break
+
+            events = self._conn.receive_data(data)
+            out.feed_data(events, len(data))
+
+
+class Http2Message(aiohttp.protocol.HttpMessage):
+    """
+    A HTTP/2-specific version of the ``HttpMessage`` ABC.
+    """
+    HOP_HEADERS = []
+
+    def __init__(self, conn, transport, stream_id):
+        self._conn = conn
+        self._stream_id = stream_id
+        super().__init__(transport, version=HttpVersion20, close=False)
+
+    def keep_alive(self):
+        return True
+
+    def add_header(self, name, value):
+        # HTTP/2 doesn't do chunked.
+        if name == aiohttp.hdrs.TRANSFER_ENCODING:
+            return
+
+        # Nor does it do Connection.
+        if name == aiohttp.hdrs.CONNECTION:
+            return
+
+        return super().add_header(name, value)
+
+    def send_headers(self, *args, **kwargs):
+        """
+        A complete override of the equivalent method from the ABC.
+        """
+        assert not self.headers_sent, 'headers have been sent already'
+        self.headers_sent = True
+
+        # We always use either the EOF payload writer or the length payload
+        # writer.
+        self.writer = self._write_h2_payload()
+        next(self.writer)
+        self._add_default_headers()
+
+        # Send the headers.
+        headers = [(':status', str(self.status))]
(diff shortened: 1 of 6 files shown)


asvetlov:
@Lukasa let's discuss it during US PyCon in a few days.

Lukasa:
Sounds good to me.

jchampio:
Okay, so as we near the end of the PyCon sprints, I've done far more reading than writing code. I was able to splice @Lukasa's contribution directly into the core in a dev branch (see above commit da7d90367) so that we have a Response object that is capable of both HTTP/1.x and HTTP/2 communication. As expected, the result is an ugly mess. My hope is that the ugly parts will begin to show us where to untangle concerns.

To capture some of the side conversations we had in the sprint:
- It sounds @asvetlov is interested in splitting up the internal classes as necessary, with the general compatibility strategy of forwarding deprecated methods to their new equivalents. I'm sure we'll run into a pain point eventually, but we'll see when we get there.
- As @Lukasa alludes to above, server-side code does not appear to clearly separate the connection/transport from either the protocol being used on that connection or the responses being sent. This is one of the first places I intend to focus next.
- aiohttp supports `sendfile()` for plaintext static file serving. Doing the same for HTTP/2 would require explicit support in hyper-h2; see https://github.com/python-hyper/hyper-h2/issues/236. It is _probably_ less generally useful than its equivalent in HTTP/1.x, since no modern browser currently uses h2c, but it may still be useful for very specific use cases.

fafhrd91:
Closing as no activity for almost a year. In any case client http2 support is more interesting from my prespective

PR #889, opened 2016-05-27
Don't crash logger when transport is None

mpaolini:
When websocket client closes connection I often get this (on the server)

```
[...]

  File "/home/marco/src/aiohttp/aiohttp/helpers.py", line 389, in log
    [message, environ, response, transport, time]))
  File "/home/marco/src/aiohttp/aiohttp/helpers.py", line 376, in _format_line
    return tuple(m(args) for m in self._methods)
  File "/home/marco/src/aiohttp/aiohttp/helpers.py", line 376, in <genexpr>
    return tuple(m(args) for m in self._methods)
  File "/home/marco/src/aiohttp/aiohttp/helpers.py", line 333, in _format_a
    return args[3].get_extra_info('peername')[0]
AttributeError: 'NoneType' object has no attribute 'get_extra_info'
```

Could not cook a test, let me know if you need it

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -330,7 +330,8 @@ class AccessLogger:
 
     @staticmethod
     def _format_a(args):
-        return args[3].get_extra_info('peername')[0]
+        return args[3].get_extra_info('peername')[0] if args[3] is not None \
+            else '-'
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


asvetlov:
I can live without a test for this particular case.

asvetlov:
Thanks

Issue #117, opened 2014-07-15
HTTP GET + compress=True = 501


a-urth:
There are some strange behavior with proxies. Trying to do HTTP GET request, requests package gives correct result, asyncio - 501.

Response headers:
SERVER - squid/3.1.4
MIME-VERSION - 1.0
DATE - Mon, 14 Jul 2014 20:20:15 GMT
CONTENT-TYPE - text/html
CONTENT-LENGTH - 3136
X-SQUID-ERROR - ERR_UNSUP_REQ 0
VIA - 1.0 host (squid/3.1.4)

Proxy - 173.234.226.20:8800

Seems that GET method for such proxy is incorrect, even if request done through http.
Any suggestions?

fafhrd91:
is it still the problem?

danielnelson:
I think it must be still an issue based on comments in #835.

fafhrd91:
@danielnelson what to fix it?

danielnelson:
sure

asvetlov:
Fixed by #891

PR #891, opened 2016-05-31
Don't add Content-Encoding and Transfer-Encoding if no body

danielnelson:
This is a fix for issue #117.  

This fix disables setting the Content-Encoding and Transfer-Encoding headers when a request is created with compress=True but does not have a body.


asvetlov:
Thanks!

PR #905, opened 2016-06-03
Skip resolve on ip

boylea:
## What these changes does?

This is to finish the work done in PR #874 for the PyCon sprints. I have added using a regex to check if a host is an IP address, which is ~2x more efficient according to `timeit`.

## How to test your changes?

I have added tests for the regex, let me know if these should be more extensive. It is more important to make sure that we do not qualify something that needs a DNS lookup as an IP, than the other way around. I was not concerned about host strings that were neither a valid IP or a domain, as this is an error either way.

## Checklist

- [ ] Code is written and well
- [ ] Tests for the changes are provided


asvetlov:
thanks

PR #909, opened 2016-06-03
Include an example of middleware to handle error pages

### LINKED ISSUE #888: Document/example for custom 404 and 405
## Long story short

When using aiohttp to create RESTfull JSON interfaces, one needs to always return JSON, but currently there is no mention on how to do it. #598 refers to this.
## Expected behaviour

I would expect to be able to set app.default_not_found and app.default_not_allowed or something easy to override default responses.
## Actual behaviour

I need to create a middleware (maybe simple).
## Steps to reproduce

Just create an app without routes and request a path.

djmitche:
This serves both as an example of middleware, and as a useful hint as to
how to produce custom error pages (a common request).  Fixes #888.

### DIFF
--- a/docs/web.rst
+++ b/docs/web.rst
@@ -753,6 +753,35 @@ post-processing like handling *CORS* and so on.
    Middlewares accept route exceptions (:exc:`HTTPNotFound` and
    :exc:`HTTPMethodNotAllowed`).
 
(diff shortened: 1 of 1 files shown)


asvetlov:
Thanks

PR #912, opened 2016-06-03
Allow configuration of header size limits

djmitche:
Sorry this took me so long ;)

### DIFF
--- a/aiohttp/server.py
+++ b/aiohttp/server.py
@@ -73,6 +73,12 @@ class ServerHttpProtocol(aiohttp.StreamProtocol):
     :param str access_log_format: access log format string
 
     :param loop: Optional event loop
+
+    :param int max_line_size: Optional maximum header line size
+
+    :param int max_field_size: Optional maximum header field size
+
+    :param int max_headers: Optional maximum header size
     """
     _request_count = 0
     _request_handler = None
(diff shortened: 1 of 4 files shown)


asvetlov:
Please respect pep8: aiohttp/helpers.py:336:80: E501 line too long (92 > 79 characters) ;)

djmitche:
Oops, fixed

Dustin

aiohttp/server.py
```diff
@@ -103,6 +106,14 @@ def __init__(self, *, loop=None,
         self._timeout = timeout  # slow request timeout
         self._loop = loop if loop is not None else asyncio.get_event_loop()
 
+        parser_kwargs = {}
```
asvetlov, line 109: These lines look like python2-ish.
Is there any reason to don't use keyword-only constructor parameters with reasonable default values?

djmitche, line 109: I wanted to avoid giving default values for those parameters away from the
place where the parameters are used.  That is, if I was looking for the
default value of max_headers, I would go looking in the parser class, not
here.  Is there a more py3-ish way to accomplish that?

terrisgit:
What is the right way to set this using an application runner? I googled and read and googled and read.. If not possible via runner then what is the correct modern way?

PR #913, opened 2016-06-04
correct variable name in testing docs

samuelcolvin:
## What do these changes do?

Correct variable name in pytest section of testing docs

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -326,16 +326,16 @@ class TestClient:
 
     :type app: aiohttp.web.Application
 
-    :param protocol: the aiohttp.web application passed to create_test_server
+    :param protocol: http or https
 
-    :type app: aiohttp.web.Application
+    :type protocol: str
 
     TestClient can also be used as a contextmanager, returning
     the instance of itself instantiated.
     """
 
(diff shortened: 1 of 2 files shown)


aiohttp/test_utils.py
```diff
@@ -326,16 +326,16 @@ class TestClient:
...
-    :type app: aiohttp.web.Application
+    :type protocol: str
 
     TestClient can also be used as a contextmanager, returning
     the instance of itself instantiated.
     """
 
     def __init__(self, app, protocol="http"):
-        self._app = app
+        self.app = app
```
asvetlov, line 338: Why did you covert `app` to public name?

samuelcolvin, line 338: Yes, was about to make a comment about this.

Very often in tests it's helpful to get access to the `app` eg. to get the db connection and check data has changed as expected.

samuelcolvin, line 338: Since aiohttp encourages `app` holding all variables this is basically the test client's only way of getting access to serverside data.

asvetlov, line 338: Ok

asvetlov:
thanks

Issue #423, opened 2015-06-25
User application testing facility


saaj:
Does `aiohttp` provide functionality to test application written with its HTTP client? It is in the sense of packages like [httpretty](https://pypi.python.org/pypi/httpretty/), [responses](https://pypi.python.org/pypi/responses) and this [SO question](http://stackoverflow.com/q/9559963/2072035). Basically it may be some mock connector to say _here's the next response your should return on any request_. Or it may be more elaborate with separate queues for HTTP methods, base URLs, et cetera. It's useful for integration and system testing. Workarounds are possible, but require to construct complete server to imitate the target and to patch the application's URLs, which can take some substantial effort.

Even though it's not in my direct interest, the same thing may also apply to HTTP server. Some sort of base test case class that provides means to start user's app, do requests and make assertions about them with ease. Here is [how CherryPy does it](https://cherrypy.readthedocs.org/en/latest/advanced.html#testing-your-application).

I've looked at the test suite and `aiohttp.test_utils` but there seems to be nothing what I'm looking for exactly. If I have not overlooked it and there's no such functionality at the moment, this can be treated as a feature request.

asvetlov:
I believe all mentioned projects are exist because it's hard to write test server in classical blocking way. 
You should execute it in separate thread and synchronize testing code with this server.
However in `asyncio` world co-existing of client and server API's in the same test case is pretty easy. 
We do use it in test suite for `aiohttp`, sure.
Please take a look on our typical test file: https://github.com/KeepSafe/aiohttp/blob/master/tests/test_client_functional_newstyle.py

saaj:
I don't think that internal complexity really matters for an end user. She has certain use cases and there are common approaches to solve them without too much of boilerplate. Your answer, in other words, is DIY, which is actually an answer. Anyway, I want to point to some things.

# Client testing

Specifically it is a white-box testing. A system written with `aiohttp` has endpoints (URLs) to which it performs requests during its normal operation. Unless you patch all the endpoints to _localhost_ or if the design allows application-wide configuration, you can't test it. And ease is of composition isn't relevant here. Fake connector, which you can pass a number of pre-fabricated responses, is.

# Server testing

The threaded server, CherryPy, didn't convince you. Here's my next attempt: http://tornado.readthedocs.org/en/latest/testing.html. You can see the same `unittest` API no mater if the things is blocking IO or not.

asvetlov:
Well, I see your point.

[...]

PR #914, opened 2016-06-04
add pytest_plugins

samuelcolvin:
## What do these changes do?

Add custom plugin to support testing with pytest. Therefore:
* remove the reliance on pytest-asyncio.
* remove all unnecessary boilerplate from tests making getting started with tests easier.

I'm submitting it before I finish it to see if you're willing to merge this idea?

Making testing as easy as possible, for example:

```python
from aiohttp import web
pytest_plugins = 'aiohttp.pytest_plugins'

async def hello(request):
    return web.Response(body=b'Hello, world')

def create_app(loop):
    app.router.add_route('GET', '/', hello)
    return app

async def test_hello(test_client):
    client = await test_client(create_app)
    resp = await client.get('/')
    assert resp.status == 200
    text = await resp.text()
    assert 'Hello, world' in text
```

## Related issue number

#423

## Checklist

### DIFF
--- a/aiohttp/pytest_plugins.py
+++ b/aiohttp/pytest_plugins.py
@@ -3,7 +3,8 @@ import contextlib
 
 import pytest
 
-from .test_utils import TestClient, loop_context, setup_test_loop, teardown_test_loop
+from .test_utils import (TestClient, loop_context, setup_test_loop,
+                         teardown_test_loop)
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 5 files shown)


asvetlov:
Looks good!

samuelcolvin:
ready.

samuelcolvin:
If this is ok it would be great to get it merged before the next release.

asvetlov:
Merged. Thank you very much.

PR #916, opened 2016-06-06
Add optional flag for quoting fields

### LINKED ISSUE #903: How to pass array to form data (multipart/form-data)
I need to call API using `multipart/form-data` with list of strings and list of objects

```
data = aiohttp.FormData()
data.add_field("emails", candidate["email"], content_type="multipart/form-data")
data.add_field("phones[]", [{"value": phone}], content_type="multipart/form-data")
```

If I'm not mistaken FormData is designed to support urlencoded params or bytes for files only. Passing anything other than bytes results in `ClientRequestError: Can not write request body`. Is there a better way that passing everything as files?

What is more the 

```
data.add_field("emails[]", bytes(candidate["email"], "ascii"), content_type="multipart/form-data")
```

results in `emails%5B%5D` and form data doesn't allow bytes as names

```
        elif isinstance(value, (bytes, bytearray, memoryview)):
            if filename is None and content_transfer_encoding is None:
                filename = name

        type_options = multidict.MultiDict({'name': name})
        if filename is not None and not isinstance(filename, str):
            raise TypeError('filename must be an instance of str. '
                            'Got: %s' % filename)
```

lukasz-madon:
## What do these changes do?

Add a flag for quoting fields.

RFC standard requires all fields to be US-ASCII. Additional `quote`
may provide extra security (RFC-1806 p. 2.3), but does not allow
for interoperability with API that requires ASCII characters outside
`quote` set e.g. emails[] becomes emails%5B%5D

There will be no change for current users (the parameter defaults to True). Users that want to skip quoting can pass the flag to constructor.

```
form = helpers.FormData(quote_fields=False)
```
## Related issue number

fixes #903
## Checklist
- [x]  I think the code is well written

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -85,11 +85,12 @@ class FormData:
     """Helper class for multipart/form-data and
     application/x-www-form-urlencoded body generation."""
 
-    def __init__(self, fields=()):
+    def __init__(self, fields=(), quote_fields=True):
         from . import multipart
         self._writer = multipart.MultipartWriter('form-data')
         self._fields = []
         self._is_multipart = False
+        self._quote_fields = quote_fields
 
         if isinstance(fields, dict):
             fields = list(fields.items())
@@ -181,7 +182,9 @@ class FormData:
         for dispparams, headers, value in self._fields:
             part = self._writer.append(value, headers)
             if dispparams:
-                part.set_content_disposition('form-data', **dispparams)
+                part.set_content_disposition(
+                    'form-data', quote_fields=self._quote_fields, **dispparams
+                )
                 # FIXME cgi.FieldStorage doesn't likes body parts with
                 # Content-Length which were sent via chunked transfer encoding
                 part.headers.pop(hdrs.CONTENT_LENGTH, None)
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -851,7 +851,7 @@ class BodyPartWriter(object):
             raise RuntimeError('unknown content transfer encoding: {}'
                                ''.format(encoding))
 
-    def set_content_disposition(self, disptype, **params):
+    def set_content_disposition(self, disptype, quote_fields=True, **params):
         """Sets ``Content-Disposition`` header.
 
         :param str disptype: Disposition type: inline, attachment, form-data.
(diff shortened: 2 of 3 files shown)


asvetlov:
Looks good. 
@lukasz-madon would you add documentation update also?

lukasz-madon:
Actually, the problem is a bit bigger. Aiohttp doesn't implement specification properly see http://greenbytes.de/tech/webdav/rfc2616.html#rfc.section.19.5.1

Field can be a `token` (a limited ascii set) **OR** ascii string inside a quotation mark. Aiohttp only allows tokens. I think this should be a default behaviour, but it may break if users want to update the library.

jchampio:
@lukasz-madon: You'll want to find the latest MIME RFCs to point to instead. Not only is RFC 2616 obsolete, but I don't think it ever applied to multipart payloads to begin with.

https://tools.ietf.org/html/rfc7578#section-4.2, https://tools.ietf.org/html/rfc2183, and https://tools.ietf.org/html/rfc2231 might be good starting points? I am not a MIME expert...

lukasz-madon:
@jchampio not a MIME expert, but looking at RFC and other implementations `filename` should allow any US-ASCII inside `""` https://github.com/pallets/werkzeug/blob/master/tests/test_http.py#L250 https://github.com/pallets/werkzeug/blob/master/werkzeug/http.py#L180

kxepal:
@lukasz-madon This case is invalid [according test suite](http://greenbytes.de/tech/tc2231/#attfnbrokentoken) we used to implement Content-Disposition parser. Though major part of browsers allows that, the spec says the other.

fafhrd91:
@kxepal thoughts?

kxepal:
Hm...I was wrong in previous message about it's wrong - I miss the quotes in test.

@fafhrd91 need a bit time to make sure that's correct, but so far LGFM.

fafhrd91:
@kxepal merge whenever ready

kxepal:
@fafhrd91 ok

fafhrd91:
@kxepal i need you to make a decision

kxepal:
@fafhrd91 Sorry, I forgot about.

fafhrd91:
Thanks!

PR #927, opened 2016-06-16
Lingering close in ServerHttpProtocol 

msornay:
## What do these changes do?

Hello,
 
Quoting [RFC 7230](https://tools.ietf.org/html/rfc7230#page-56) : 

> If a server performs an immediate close of a TCP connection, there is
   a significant risk that the client will not be able to read the last
   HTTP response.  If the server receives additional data from the
   client on a fully closed connection, such as another request that was
   sent by the client before receiving the server's response, the
   server's TCP stack will send a reset packet to the client;
   unfortunately, the reset packet might erase the client's
   unacknowledged input buffers before they can be read and interpreted
   by the client's HTTP parser.

>   To avoid the TCP reset problem, servers typically close a connection
   in stages.  First, the server performs a half-close by closing only
   the write side of the read/write connection.  The server then
   continues to read from the connection until it receives a
   corresponding close by the client, or until the server is reasonably
   certain that its own TCP stack has received the client's
   acknowledgement of the packet(s) containing the server's last
   response.  Finally, the server fully closes the connection.

This is a tentative implementation of a lingering close inspired by [those nginx settings](http://nginx.org/en/docs/http/ngx_http_core_module.html#lingering_close)

The typical use case where the absence of lingering close is worrisome in our application is : 
1. The client starts PUTting a sizeable amount of data ;
2. Server replies with a HTTP 401 Unauthorized after parsing the headers ;
3. The connection  is closed too early for the client to receive / react to the reply. 

It is our experience that some HTTP clients behave badly in front of "early" responses. We believe that having lingerings closes is correct and we are currently investigating how much it will help in practice.

## Checklist

First aiohttp contribution, any input welcome !

### DIFF
--- a/aiohttp/server.py
+++ b/aiohttp/server.py
@@ -58,6 +58,14 @@ class ServerHttpProtocol(aiohttp.StreamProtocol):
 
     :param int timeout: slow request timeout
 
+    :param bool lingering_on: enable lingering close
+
+    :param int lingering_time: maximum time during which the server reads and
+    ignore additionnal data comming from the client when lingering close is on
+
+    :param int lingering_timeout: maximum waiting time for more client data to
+    arrive when lingering close is in effect
+
     :param allowed_methods: (optional) List of allowed request methods.
                             Set to empty list to allow all methods.
     :type allowed_methods: tuple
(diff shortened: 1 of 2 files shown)


asvetlov:
@msornay thank you very much for PR.
I've simplified a little the code in #1050

PR #929, opened 2016-06-16
Fix empty ALLOW Response header for cls based View

singulared:
## What do these changes do?

Fix empty ALLOW Response header for cls based View

We must check existing lower http method names for assembling
ALLOW header and passing into web.HTTPMethodNotAllowed.

yes

## Related issue number

No issue

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -655,7 +655,8 @@ class View(AbstractView):
             return (yield from self.__iter__())
 
     def _raise_allowed_methods(self):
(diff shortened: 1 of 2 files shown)


asvetlov:
Very sorry for late answer but you are asking for ability to write `def get()` instead of `def GET()` as I understand.
Why?

singulared:
Hi! i think lowercase is more pythonic. In other case i did't have valid ALLOW response header if i use `def get()`. But i can use it as request handler.

singulared:
Docs use lowercase methods:
http://aiohttp.readthedocs.io/en/stable/web_reference.html#view

asvetlov:
Ok.
Thanks!

PR #937, opened 2016-06-19
 Show how the caller can get back the HTML text

cclauss:
## What do these changes do?

Modify example to show how the caller can get the results of execution

Shortens the learning cycle for real world uses of aiohttp
## Checklist

- [ x] I think the code is well written
- [ x] Unit tests for the changes exist

Instead of merely printing the HTML _inside_ the coroutine function, demonstrate how the _caller_ can receive the HTML text for further processing.


asvetlov:
Thanks!

PR #941, opened 2016-06-23
Fixed URL matching against quoted characters

IlyaSemenov:
## What do these changes do?

The patch fixes URL matching against ümlauts, spaces and other chars that become URL-quoted.
In particular, this fixes `StaticRoute` not being able to deliver static files with non-ASCII filenames.

It also fixes the incorrect test which expected the successful match of raw `/path/my%2Fpath` against `/{path}/{subpath}`, which contradicted the documentation.

This is backwards compatible, unless users adapted to the old (incorrect) behaviour and, for example, renamed their static files to use %XX URL-quoting notation (which is unlikely), or expected `{}` pattern to match an inner `/` (while the documentation clearly says it won't).

## Related issue number

There is no related issue, I decided to come up with a PR right away.

## Checklist

- [x] Documentation doesn't need to reflect the change

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -174,7 +174,7 @@ class Request(dict, HeadersMixin):
     @reify
     def raw_path(self):
         """ The URL including raw *PATH INFO* without the host or scheme.
-        Warning, the path is unquoted and may contains non valid URL characters
+        Warning, the path is unquoted and may contain non valid URL characters
 
         E.g., ``/my%2Fpath%7Cwith%21some%25strange%24characters``
         """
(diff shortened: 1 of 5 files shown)


IlyaSemenov:
@asvetlov I beg my pardon? This was simply rejected without any feedback? I just installed the latest aiohttp 0.22.2 and the problem with non-ASCII named static files is still there. I believe the patch does the right thing and does it right.

asvetlov:
Github have closed all pending PRs after I had deleted 0.21 branch.

I'll return to reviewing all issues related to URL encoding at working on 0.23

asvetlov:
I believe the issue is solved by yarl now.
If the library still have a bug please make a new issue.

Issue #958, opened 2016-07-05
StaticRoute.sendfile incorrectly uses add_writer/remove_writer


1st1:
See https://github.com/python/asyncio/issues/372 for more details.

It seems that on modern OSes, aiohttp's implementation of sendfile always hits a fast path, i.e. first call to `os.sendfile` succeeds. On Mac OS 10.10, though, it sometimes does not, so the code calls `add_writer` on the underlying transport's socket.

This is a wrong thing to do, as the status of the writer callback (and the active callback itself) is the private state of the transport. It also seems that `StaticRoute.sendfile` may not work at all when `add_writer` is used.

I suggest to duplicate the socket that `get_extra_info` returns before adding/removing readers/writers.

1st1:
Alright, so in the next asyncio release, using add_reader/add_writer on a socket that `transport.get_extra_info('socket')` returns will be an error. I'll also fix uvloop to do that too. This issue has to be fixed asap.

asvetlov:
> Alright, so in the next asyncio release, using add_reader/add_writer on a socket that transport.get_extra_info('socket') returns will be an error.

@1st1 could you point on source code change for the fix? I miss it.

asvetlov:
Aaah, it's not implemented still.

PR #964, opened 2016-07-14
dup a socket for sendfile usage

asvetlov:
Fix for #958

@1st1 please take a look if I've done a fix properly.


1st1:
Looks good.

PR #966, opened 2016-07-15
Add IE support for cookie deletion.

SethMichaelLarson:
## What do these changes do?

Adds support for IE cookie expiration.

Not unless they are deleting cookies, otherwise they will actually delete on IE!

## Related issue number

None.

## Checklist

IE<=11 ( !!! ) doesn't support the MaxAge property for cookies so del_cookie() also needs to set the Expires cookie header to support IE. Uses the UTC epoch for a standard expire time.

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -537,7 +537,7 @@ class StreamResponse(HeadersMixin):
         """
         # TODO: do we need domain/path here?
         self._cookies.pop(name, None)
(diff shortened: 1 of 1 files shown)


sethmlarson:
Obviously need to update the cookie tests along with this pull request, I'll add those as well.

asvetlov:
Looks good. Please update tests.

asvetlov:
Fixed by #994

PR #973, opened 2016-07-19
Added build_aiohttp_client_response pytest fixture

argaen:
## What do these changes do?

Adds a parametrized fixture for the pytest_plugin.py which allows build `ClientResponse` with specified parameters such as method, url, content, headers and status.

Adds extra fixture being able to reduce the number of lines needed for testing external services queried with aiohttp.

## Related issue number

None

## Checklist

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -5,6 +5,7 @@ import pytest
 
 from .test_utils import (TestClient, loop_context, setup_test_loop,
                          teardown_test_loop)
+from .client_reqrep import ClientResponse
 
 
 @contextlib.contextmanager
@@ -65,3 +66,27 @@ def test_client(loop):
 
     if client:
         client.close()
+
+
+@pytest.fixture()
+def build_aiohttp_client_response():
+    """ This is a parametrized fixture for building aiohttp client responses as needed
+
+    An example usage:
+        ```
+        resp = yield from build_aiohttp_client_response(
+            "GET", "http://example.com",
+            b"{'a': 'a', 'b': 'b'},
+            {"CONTENT-TYPE": "application/json"},
+            200)
+        ```
+
+        dict_resp = yield from resp.json()
+    """
+    async def build_response(method, url, content, headers=None, status=200):
+        cr = ClientResponse(method, url)
+        cr._content = content
+        cr.status = status
+        cr.headers = headers or {}
+        return cr
+    return build_response
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -70,9 +70,11 @@ def test_client(loop):
 
 @pytest.fixture()
 def build_aiohttp_client_response():
-    """ This is a parametrized fixture for building aiohttp client responses as needed
+    """
+    This is a parametrized fixture for building aiohttp client responses as
+    needed.
 
-    An example usage:
+    Example usage:
         ```
         resp = yield from build_aiohttp_client_response(
             "GET", "http://example.com",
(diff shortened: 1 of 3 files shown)


asvetlov:
What is an use case for the fixture?

argaen:
I'm using it in a couple of projects for testing requests to external (mocked) services. So, I have the following test:

``` python
async def check_service_raises_exception(self, my_service, build_aiohttp_client_response):

    external_service_response = b'{"a": "a", "b": "b"}'

    with mock.patch("aiohttp.client.ClientSession._request") as patched:
        patched.return_value = build_aiohttp_client_response("GET", '/', external_service_response, status=200)
        with pytest.raises(MyServiceError):
            await my_service.get_items()
```

and then, the `my_service.get_items()` call just uses a `ClientSession.get` call for retrieving the external service items, process and format them and return the values or raise an `Exception` if needed.

As a summary, use case is: Mock external calls.

I don't know if there is already a better approach to do that, but since I didn't find one, I thought I would share it :).

asvetlov:
Sorry, but I don't think the PR is the best way for communicating with external sources.

Long story short it's much easier to create _fake server_ for emulating external resource and run it in a test.

Forcing user to patch private API which can be changed without any announcement/deprecation is **very bad** idea.

Perhaps I should provide an example for _fake server_ in documentation/demo.

argaen:
Hi, sorry for the late response.

I agree with the point that "forcing user to patch private API" is a very bad idea. After checking a bit, I've seen it can also be used while patching the `session.get`, `session.post`, `session.request`, etc... Here is a simple use case:

``` python
import pytest
import aiohttp

from unittest import mock

pytest_plugins = 'aiohttp.pytest_plugin'


async def get_google():
        async with session.request('https://google.com') as resp:
            dict_resp = await resp.json()
            return "I've received {}, {}".format(dict_resp['a'], dict_resp['b'])


@pytest.mark.asyncio
async def test_hello(build_aiohttp_client_response):
    with mock.patch("aiohttp.client.ClientSession.request") as mock_get:

        resp = await build_aiohttp_client_response(
            "GET", "http://example.com",
            b'{"a": "a", "b": "b"}',
            {"CONTENT-TYPE": "application/json"},
            200)

        mock_get.return_value = resp

        ret = await get_google()

        assert ret == "I've received a, b"
```

IMHO I prefer this way of mocking my services rather than having to launch a fake server for each service I want to mock.

asvetlov:
It's up to you but I really found that making fake server with aiohttp is really trivial procedure.

argaen:
If you update the docs with an example I'll check then and decide :). Thanks!

hyzhak:
I love example with fake server from @asvetlov (https://github.com/KeepSafe/aiohttp/blob/master/examples/fake_server.py). But I don't see how it can help to test black-box functions like `get_google()` from @argaen example – without exposing or getting session from outside of function. In other works without converting black box to the white box. 

And this one won't work because we won't able to propagate session inside of `get_google`:

``` python
@pytest.mark.asyncio
async def test_get_google(event_loop):
    fake_google = FakeGoogle(loop=loop)
    info = await fake_google.start()
    resolver = FakeResolver(info, loop=loop)
    connector = aiohttp.TCPConnector(loop=loop, resolver=resolver,
                                     verify_ssl=False)

    async with aiohttp.ClientSession(connector=connector,
                                     loop=loop) as session:
        ret = await get_google()
        assert ret == "I've received a, b"

    await fake_google.stop()

```

argaen:
We've continued using this fixture in our codebase and works well enough. Good thing is that you just need this fixture in order to mock responses for any service you integrate with in your code.

brycedrennan:
@hyzhak @asvetlov @argaen You might find this useful:

https://github.com/CircleUp/aresponses

PR #984, opened 2016-07-23
Fix #976: Add support for websocket send_json and receive_json

### LINKED ISSUE #976: Fix websocket json support
Implement server `resp.send_json()` and client `resp.send_json()`/`resp.receive_json()`

vitaliemaldur:
## What do these changes do?

Add support for websocket server to send json and for websocket client to send and receive json

User can use methods:
`send_json` and `receive_json` both on websocket server and client

## Related issue number
#976 

## Checklist

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -154,6 +154,13 @@ class WebSocketResponse(StreamResponse):
                             type(data))
         self._writer.send(data, binary=True)
 
+    def send_json(self, data, *, dumps=json.dumps):
+        if self._writer is None:
+            raise RuntimeError('Call .prepare() first')
+        if self._closed:
+            raise RuntimeError('websocket connection is closing')
+        self._writer.send(dumps(data), binary=False)
+
     @asyncio.coroutine
     def wait_closed(self):  # pragma: no cover
         warnings.warn(
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -159,7 +159,7 @@ class WebSocketResponse(StreamResponse):
             raise RuntimeError('Call .prepare() first')
         if self._closed:
             raise RuntimeError('websocket connection is closing')
-        self._writer.send(dumps(data), binary=False)
+        self.send_str(dumps(data))
 
     @asyncio.coroutine
     def wait_closed(self):  # pragma: no cover
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -291,12 +291,8 @@ class WebSocketResponse(StreamResponse):
 
     @asyncio.coroutine
     def receive_json(self, *, loads=json.loads):
-        msg = yield from self.receive()
-        if msg.tp != MsgType.text:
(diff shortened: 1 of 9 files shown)


aiohttp/web_ws.py
```diff
@@ -154,6 +154,13 @@ def send_bytes(self, data):
                             type(data))
         self._writer.send(data, binary=True)
 
+    def send_json(self, data, *, dumps=json.dumps):
```
asvetlov, line 157: Please call `.send_str()` instead of working on low level.

aiohttp/websocket_client.py
```diff
@@ -88,6 +89,11 @@ def send_bytes(self, data):
                             type(data))
         self._writer.send(data, binary=True)
 
+    def send_json(self, data, *, dumps=json.dumps):
+        if self._closed:
```
asvetlov: Use `send_str()`

asvetlov:
Would you add tests also?

vitaliemaldur:
Yes I will

vitaliemaldur:
@asvetlov what I can do to avoid

```
AttributeError: 'module' object has no attribute 'JSONDecodeError'
tests\test_web_websocket.py:161: AttributeError
```

I am using `json.decoder.JSONDecodeError` in a `try/except`.

Also, do you have some suggestions for the tests I wrote?

aiohttp/web_ws.py
```diff
@@ -154,6 +154,13 @@ def send_bytes(self, data):
...
+        if self._writer is None:
```
asvetlov: These checks are not required: `.send_str()` performs them anyway.
There is no need to do the check twice.

tests/test_web_websocket.py
```diff
@@ -128,6 +144,25 @@ def receive():
...
+    def test_receive_json_nonjson(self):
+
+        @asyncio.coroutine
+        def go():
+            req = self.make_request('GET', '/')
+            ws = WebSocketResponse()
+            yield from ws.prepare(req)
+
+            @asyncio.coroutine
+            def receive():
```
asvetlov: Use functional test for the case.

vitaliemaldur:
@asvetlov I've noticed that in `web_ws.py` we have `send_str`, `send_bytes`, `send_json` methods and coresponding `receive_str`, `receive_bytes`, `receive_json`. But in `websocket_client.py` we have `send_str`, `send_bytes`, `send_json` methods and only `receive_json`. Is that on purpose? or should I add the rest of methods there?

asvetlov:
I think yes. 
From my perspective we should support the same methods for both client and server for consistency.

vitaliemaldur:
@asvetlov I have added all the methods discussed above, now we have:
- `send_str`
- `send_bytes`
- `send_json`
- `receive_str`
- `receive_bytes`
- `receive_json`

both on `websocket_client.py` and `web_ws.py`. I have also added documentation for them.

asvetlov:
Thanks!

PR #991, opened 2016-07-23
Issue997 default connector limit

rmariano:
## What do these changes do?

Changes the default limit for connections from None, to 20

It should not, unless the user was expecting to always use an unlimited number of connections all the time, in which case, they are now being bounded to 20.

## Related issue number
Issue#997

## Checklist

- [ x ] I think the code is well written
- [ x ] Unit tests for the changes exist
- [ x ] Documentation reflects the changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.swp
 *.bak
 *.egg
 *.egg-info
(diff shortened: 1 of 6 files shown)


asvetlov:
Thank you very much!

PR #996, opened 2016-07-24
Update basic_srv.py

cursorius-cursor:
With a fresh virtual environment (Python 3.5.2) I get an import error
python simplehttpserver.py
  File "simplehttpserver.py", line 9, in <module>
    from aiohttp.multidict import MultiDict
ImportError: No module named 'aiohttp.multidict'

with an alternate import strategie it is working!?


asvetlov:
Yes.
Fixed by 9b06e81

PR #997, opened 2016-07-24
Avoid use mutable CIMultiDict kw param in make_mocked_request

pfreixes:
## What do these changes do?

Avoid use mutable `CIMultiDict` kw param in make_mocked_request

Even though is hard to get an issue, there is still chances to get
unwished behaviors as a side effect because of that.

## Checklist


asvetlov:
Good catch!

Issue #985, opened 2016-07-23
Refactor ProxyConnector


asvetlov:
1. Merge proxying logic into TCPConnector.
2. Support `proxy` and `proxy_auth` parameters in `ClientSession.request` and descendant API.
3. Keep `ProxyConnector` for backward compatibility

asvetlov:
Fixed by #998

PR #998, opened 2016-07-24
Refactored ProxyConnector

mind1master:
## What do these changes do?

1. Proxy request logic is moved from ProxyConnector to TCPConnector
2. Proxy now is set in ClientSession.request() and can be changed for each subsequent request, e.g `session.get("http://httpbin.org/get", proxy="http://localhost:8020")`
3. ProxyConnector stays for backwards compatibility, however now it is just a thin wrapper around TCPConnector.

Old ProxyConnector code still works, however new way is preferred. 
Documentation is updated to reflect that change.

## Related issue number

#985 

## Checklist

- [+] I think the code is well written
- [+] Unit tests for the changes exist
- [+] Documentation reflects the changes

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -104,7 +104,9 @@ class ClientSession:
                 compress=None,
                 chunked=None,
                 expect100=False,
-                read_until_eof=True):
+                read_until_eof=True,
+                proxy=None,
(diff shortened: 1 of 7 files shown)


asvetlov:
Awesome!

PR #1003, opened 2016-07-25
Gunicorn worker ssl

tumb1er:
## What do these changes do?

`gunicorn` provides SSL support for sync workers and for `gaiohttp` worker for aiohttp.
`aiohttp` provides GunicornWebWorker, but it lacks SSL support.
This PR allows GunicornWebWorker to init SSL from gunicorn config.

No regressions expected

## Related issue number

Nope

## Checklist

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -17,10 +17,6 @@ class GunicornWebWorker(base.Worker):
 
     def __init__(self, *args, **kw):  # pragma: no cover
         super().__init__(*args, **kw)
-        if self.cfg.is_ssl:
-            self.ssl_context = self._create_ssl_context(self.cfg)
-        else:
-            self.ssl_context = None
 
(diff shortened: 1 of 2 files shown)


asvetlov:
What is your use case?

From my understanding people starts Reverse Proxy server like Nginx, push Gunicorn behind its and start aiohttp workers under gunicorn.

In this scheme only Nginx serves SSL/TLS connections, all internal communications are insecure and uses raw sockets, isn't it?

tumb1er:
My use case is uploading huge files over ssl (without nginx).
With nginx serving connections, it first dumps full request to disk, then reads it and pass to upstream (aiohttp). Yes, nginx support `client_body_in_file_only` but it is useless in case of multipart requests.

asvetlov:
Ok, makes sense

Issue #922, opened 2016-06-12
Error while closing a websocket


youtux:
## Long story short

I am trying to implement an endpoint that forwards websocket messages between two client. It is necessary that when a client closes the connection, the aiohttp server also close the websocket with the second client.
The problem is my program never get through the `await other_ws.close()` step.
## Expected behaviour

``` python
await other_ws.close()  # should close the other connection
stmts                             # should be able to execute code after awaiting
```
## Actual behaviour

``` python
await other_ws.close()  # never returns
stmts                             # never gets executed
```

Furthermore, I get the following error:

```
Task exception was never retrieved
future: <Task finished coro=<FlowControlDataQueue.read() done, defined at /Users/youtux/.virtualenvs/pebbleremote/lib/python3.5/site-packages/aiohttp/streams.py:589> exception=AssertionError()>
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/Users/youtux/.virtualenvs/pebbleremote/lib/python3.5/site-packages/aiohttp/streams.py", line 591, in read
    result = yield from super().read()
  File "/Users/youtux/.virtualenvs/pebbleremote/lib/python3.5/site-packages/aiohttp/streams.py", line 443, in read
    assert not self._waiter
AssertionError
```
## Steps to reproduce

Run the following file:

``` python
# aiohttp_issue.py
import asyncio
import aiohttp.web

async def echo_handler(request):
    ws = aiohttp.web.WebSocketResponse()

    # if this is the first ws, I should be killed when other clients disconnect
    if 'other_ws' not in request.app:
        request.app['other_ws'] = ws

    await ws.prepare(request)

    async for msg in ws:
        if msg.tp == aiohttp.MsgType.text:
            ws.send_str(msg.data)
        else:
            raise NotImplementedError()

    # if there is another ws, kill it.
    print('client disconnected')

    if 'other_ws' in request.app:
        print('killing other_ws')
        await request.app['other_ws'].close()
        print('this line is not going to be printed twice')
        del request.app['other_ws']

    return ws

app = aiohttp.web.Application()
app.router.add_route('GET', '/echo', echo_handler)

aiohttp.web.run_app(app, host='localhost', port=8080)
```

[...]

PR #1004, opened 2016-07-26
Add terminate method to websocket response

mpaolini:
See #922 

Add a clean way for closing websockets server-side

## Related issue number

#922 

## Checklist

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -79,8 +79,9 @@ class WebSocketResponse(StreamResponse):
         self._writer = writer
         self._protocol = protocol
         self._loop = request.app.loop
-        self._reader_task = weakref.proxy(
-            asyncio.Task.current_task(loop=self._loop))
+        current_task = asyncio.Task.current_task(loop=self._loop)
+        if current_task is not None:
+            self._reader_task = weakref.proxy(current_task)
 
     def start(self, request):
         warnings.warn('use .prepare(request) instead', DeprecationWarning)
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -2,6 +2,7 @@ import sys
 import asyncio
 import json
 import warnings
+import weakref
 
 from . import hdrs, Timeout
 from .errors import HttpProcessingError, ClientDisconnectedError
@@ -37,6 +38,8 @@ class WebSocketResponse(StreamResponse):
         self._timeout = timeout
         self._autoclose = autoclose
         self._autoping = autoping
+        self._reader_task = None
+        self._terminating = False
 
     @asyncio.coroutine
     def prepare(self, request):
@@ -76,6 +79,8 @@ class WebSocketResponse(StreamResponse):
         self._writer = writer
         self._protocol = protocol
         self._loop = request.app.loop
+        self._reader_task = weakref.proxy(
+            asyncio.Task.current_task(loop=self._loop))
 
     def start(self, request):
         warnings.warn('use .prepare(request) instead', DeprecationWarning)
@@ -167,6 +172,17 @@ class WebSocketResponse(StreamResponse):
         yield from self.close()
         self._eof_sent = True
 
+    def _cancel_reader_task(self):
+        self._reader_task.cancel()
+
+    @property
+    def terminating(self):
+        return self._terminating
+
+    def terminate(self):
+        self._terminating = True
+        self._cancel_reader_task()
+
     @asyncio.coroutine
     def close(self, *, code=1000, message=b''):
         if self._writer is None:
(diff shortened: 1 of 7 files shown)


asvetlov:
I'm considering alternative solution for the problem.
1. Not only websockets but regular streamed responses need cancelation mechanics.
2. Internally it's solved by task cancelling, every web hander is executed in separate task context.

Let's introduce 'StreamResponse.task' property.

Accessing to the property before 'await resp.prepare' produces RuntimeError.

After preparing the task is available for cancellation.

User may store response/task outside by strong reference,  there is no need for weak ref unless we control Rask's execution and shutdown it on error/cancelling gracefully.

@mpaolini thoughts?

mpaolini:
Yes, makes sense. I still think a weakref is needed to avoid reference cycles

asvetlov:
Honestly I prefer a strong reference.
We can delete it in task's done callback

mpaolini:
I thought about it some more...

> I'm considering alternative solution for the problem.
> 1. Not only websockets but regular streamed responses need cancelation mechanics.
> 2. Internally it's solved by task cancelling, every web hander is executed in separate task context.
>    
>    Let's introduce 'StreamResponse.task' property.

If we only proxy `.cancel()`, I don't think this is worth having any of these features inside aiohttp: the user can keep references of tasks outside the library instead of keeping references to websocketresponses and he can use `task.cancel()`. Let's just update the docs in this case.

But If the use case is providing a simple way to **cleanly** close a websocket server-side, then both my current implementation (and your proposal @asvetlov ) a bit simplistic.

When `.terminate()` blindly `.cancel()`s the task, it will be a bit hard for the reader implementation to do a clean shutdown: the `CancellationError` might pop up everywhere in the handler call stack, and you basically have to sprinkle `except CancellationError` around your handler to cope with termination in a sensible way.

My alternative proposal is:
- `.terminate()` only cancels **read operations** that by design suspend the handler task and currently can't be easily controlled. It is only available for WebsocketResponse
- a coroutine that is awaiting a read operation on the terminated socket receives a `asyncio.ReadTerminated` exception, that he can catch and provide a clean shutdown
- we dont' expose any other public method or property on the `WebsocketResponse` class apart from `.terminate()`
- we don't add anything to `StreamResponse`: it only can write, so the user can cleanly shut down that already.

mpaolini:
here's the new implementation as proposed in https://github.com/KeepSafe/aiohttp/pull/1004#issuecomment-235306816

with this applied, we can solve the #922 issue this way:

``` python
# aiohttp_issue.py
import asyncio
import aiohttp.web
import aiohttp.websocket

async def echo_handler(request):
    ws = aiohttp.web.WebSocketResponse()
    this_task_id = id(asyncio.Task.current_task(loop=request.app.loop))

    request.app['websockets'].add(ws)
    try:
        await ws.prepare(request)
        async for msg in ws:
            if msg.tp == aiohttp.MsgType.text:
                ws.send_str(msg.data)
            else:
                raise NotImplementedError()
    except aiohttp.websocket.WebSocketTerminate:
        print('{} ws externally terminated'.format(this_task_id))
    else:
        print('{} client disconnected'.format(this_task_id))
        for other_ws in request.app['websockets']:
            if other_ws is not ws:
                print('{} killing other ws'.format(this_task_id))
                other_ws.terminate()
    finally:
        request.app['websockets'].remove(ws)

    print('{} before close'.format(this_task_id))
    await ws.close()
    print('{} after close'.format(this_task_id))
    print('{} this line *is* going to be printed twice'.format(this_task_id))
    return ws


def main():
    loop.set_debug(True)
    app.router.add_route('GET', '/echo', echo_handler)
    app['websockets'] = set()
    aiohttp.web.run_app(app, host='localhost', port=8080)

if __name__ == '__main__':
    main()
```

asvetlov:
Marco, I still think the requirement to have a proper try/finally block around the loop in webscoket handler is a good idea.
It's required anyway.
About `StreamResponse`: typical code looks like:

```
async def stream(request):
    resp = web.StreamResponse()
    await resp.prepare(request)
    try:
        for chunk in read_file_by_chunks():
            resp.write(chunk)
            await resp.drain()
    finally:
        await cleanup()
```

As you see it contains not only writes but kind of reads (`drain()` calls).

Let me sleep on it.
Maybe for more than one night: I'm still traveling over Spain after EuroPython, so my time is limited. I have access to my laptop only in the evening and in the morning.

mpaolini:
hmmm @asvetlov another change of mind here ;) . I think we should scrap this PR and just document the best practices on how to keep track of request handler tasks, how to cancel them, and how to handle `CancelledError`

mpaolini:
#1026  contains an alternative to this PR (that I am closing). I think no specific code should be written in this library to handle this use case, it's a very straightforward thing that can be solved by using vanilla `asyncio`

PR #1005, opened 2016-07-26
Do not filter cookies if unsafe flag provided

bmwant:
## What do these changes do?
For the testing purposes or when user is aware about what he is doing it's handy to be able provide *unsafe* cookie jar (so cookies will be preserved even for domain specified as IP address, e.g. 127.0.0.1)

```
cookie_jar = aiohttp.CookieJar(loop=loop, unsafe=True)
with aiohttp.ClientSession(loop=loop, cookie_jar=cookie_jar) as session:
    resp = await session.get('http://127.0.0.1/')
    print(resp.cookies)
    await resp.release()
```
Now cookies right in place

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -701,7 +701,7 @@ class CookieJar(AbstractCookieJar):
 
             hostname = url_parsed.hostname or ""
 
-            if is_ip_address(hostname):
+            if not self._unsafe and is_ip_address(hostname):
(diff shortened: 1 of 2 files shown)


asvetlov:
Shame on me, I've missed the second IP address check.

asvetlov:
Will be fixed in aiohttp 0.22.3

PR #1006, opened 2016-07-26
Choose an unused port in tests

mpaolini:
When running test locally, I get a failure just because I have a web server running on my machine.

This fixes it

### DIFF
--- a/tests/test_client_functional_oldstyle.py
+++ b/tests/test_client_functional_oldstyle.py
@@ -7,7 +7,6 @@ import os.path
 import json
 import http.cookies
 import asyncio
-import socket
 import unittest
 from unittest import mock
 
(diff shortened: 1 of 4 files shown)


rutsky:
Thanks for the pull request!

There is already [find_unused_port()](https://github.com/elastic-coders/aiohttp/blob/2ef1f449e1fb5d0a1180ef0cc0dcfe5f6ae51746/tests/test_connector.py#L722) method in this test class. I think we should either use it, or (which is better) to remove `find_unused_port` method here and replace it with `aiohttp.test_utils.unused_port` function.

mpaolini:
@rutsky I removed all the dupes (there were a few around)

asvetlov:
Thanks, guys!

PR #1009, opened 2016-07-27
Add the multidict dependency to “Requirements”

flying-sheep:
I also linkified some URLs


flying-sheep:
apparently i failed.

also the template says docs should go to the current branch while the master readme is used!

PR #1020, opened 2016-07-29
Return named_tuple from WebSocketResponse.can_prepare

esaezgil:
## What do these changes do?

Return named tuple from WebSocketResponse.can_prepare
No

## Related issue number

https://github.com/KeepSafe/aiohttp/issues/1016
## Checklist


tests/test_server.py
```diff
@@ -206,6 +206,7 @@ def test_srv_keep_alive(srv):
     assert not srv._keep_alive
 
 
+@pytest.mark.timeout(0)
```
asvetlov: Why did you modify non-related test?

aiohttp/web_ws.py
```diff
@@ -90,16 +91,17 @@ def start(self, request):
         return resp_impl
 
     def can_prepare(self, request):
+        web_socket_ready = namedtuple('web_socket_ready', 'allowed protocol')
```
asvetlov: Don't create namedtuple class on every call.

asvetlov:
1. Implement `__bool__` method.
2. Write tests

tests/test_server.py
esaezgil: ~~This test timed out, (I guess because of the addition of pytest-timeout) and modified it to fix it~~
My bad

asvetlov: As I see tests are passed both locally and on travis-ci.
If timeout is an issue it should be fixed by separate PR anyway

aiohttp/web_ws.py
asvetlov: Rename `allowed` -> `ok`, it sounds better.

aiohttp/web_ws.py
```diff
@@ -37,6 +38,8 @@ def __init__(self, *,
         self._timeout = timeout
         self._autoclose = autoclose
         self._autoping = autoping
+        self._web_socket_ready = namedtuple('web_socket_ready',
```
asvetlov: Sorry, you did't get my point.
Push
`WebSocketReady = namedtuple(`WebSocketReady', ...)` at module namespace.

asvetlov:
Much better but tests and `__bool__` are still required.

tests/test_web_websocket.py
```diff
@@ -178,25 +178,29 @@ def test_write_non_prepared():
 def test_can_prepare_ok(make_request):
     req = make_request('GET', '/', protocols=True)
     ws = WebSocketResponse(protocols=('chat',))
+    assert bool(ws.can_prepare(request=req)) is True
```
asvetlov: Don't change existing tests but create a new one.

You don't need to call `.can_prepare()` -- just create an instance of `WebSocketReady` and test it.
I expect about three or four new tests

tests/test_web_websocket.py
```diff
@@ -178,25 +178,29 @@ def test_write_non_prepared():
...
     assert(True, 'chat') == ws.can_prepare(req)
 
 
 def test_can_prepare_unknown_protocol(make_request):
     req = make_request('GET', '/')
     ws = WebSocketResponse()
+    assert bool(ws.can_prepare(request=req)) is True
```
asvetlov: Don't modify existing test

tests/test_web_websocket.py
```diff
@@ -178,25 +178,29 @@ def test_write_non_prepared():
...
     assert (True, None) == ws.can_prepare(req)
 
 
 def test_can_prepare_invalid_method(make_request):
     req = make_request('POST', '/')
     ws = WebSocketResponse()
+    assert bool(ws.can_prepare(request=req)) is False
```
asvetlov: Don't modify existing test

tests/test_web_websocket.py
```diff
@@ -178,25 +178,29 @@ def test_write_non_prepared():
...
     assert (False, None) == ws.can_prepare(req)
 
 
 def test_can_prepare_without_upgrade(make_request):
     req = make_request('GET', '/',
                        headers=CIMultiDict({}))
     ws = WebSocketResponse()
+    assert bool(ws.can_prepare(request=req)) is False
```
asvetlov: Don't modify existing test

tests/test_web_websocket.py
asvetlov: Did you write unittests before?

aiohttp/web_ws.py
```diff
@@ -17,6 +18,9 @@
 
 THRESHOLD_CONNLOST_ACCESS = 5
 
+WebSocketReady = namedtuple('WebSocketReady', 'ok protocol')
+WebSocketReady.__bool__ = lambda self: self.ok
```
asvetlov: I believe this code is not idiomatic.
Please use inheritance as python doc suggests in https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields

```
>>> class Point(namedtuple('Point', 'x y')):
...     __slots__ = ()
...     @property
...     def hypot(self):
...         return (self.x ** 2 + self.y ** 2) ** 0.5
...     def __str__(self):
...         return 'Point: x=%6.3f  y=%6.3f  hypot=%6.3f' % (self.x, self.y, self.hypot)
```

asvetlov:
Please add a test for `__bool__`.
I believe it is the last missing point.

esaezgil:
Hi Andrew, thanks for reviewing the PR. I updated it after your last comment. Please let me know if you have any other comments. 
Enrique

asvetlov:
Thanks!

PR #1021, opened 2016-07-30
WebSocket internal implementation refactoring

asvetlov:


aiohttp/_ws_impl.py
```diff
@@ -51,12 +45,19 @@
...
 
+
+class MsgType(IntEnum):
+    continuation = 0x0
+    text = 0x1
+    binary = 0x2
+    ping = 0x9
+    pong = 0xa
+    close = 0x8
+    closed = 20
```
kxepal: While else numbers are hex-like, why these are not?

asvetlov: Good catch.
Because technically they are not constants defined by RFC but user values.
I'll start them from 0x100 to never clash with future websocket standard updates.

PR #1022, opened 2016-07-31
Skip IPv6 tests when its not available

kxepal:
This fixes tests for IPv4-only hosts. Also, I found interesting case when Python is build with IPv6 support, but OS actually is not (one case: disabled in kernel). This turns to be undetectable by Python which still thinks IPv6 is ready to go (`socket.has_ipv6 is True`), but it fails meeting the reality.

Obliviously, this is not correct things setup, but we shouldn't bail because of that.


asvetlov:
Looks good, please merge.

PR #1026, opened 2016-07-31
Document easy way to close websockets server-side

mpaolini:
Adds a FAQ entry on how to programmatically close websocket server-side. Alternative to #1004 

See #922


asvetlov:
Thanks.
I still think adding `.task` property to base response class makes sense.

PR #1031, opened 2016-08-01
Some docs fixes

mpaolini:
*  `dumps` is not a term
* .read() ws methods can only be called from request handler tasks
* other minor improvementss

### DIFF
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -851,6 +851,11 @@ WebSocketResponse
       A :ref:`coroutine<coroutine>` that initiates closing
       handshake by sending :const:`~aiohttp.WSMsgType.close` message.
 
+      .. note::
+
+         Can only be called by the request handling task. To programmatically
+         close websocket server side see the :ref:`FAQ section <aiohttp_faq_terminating_websockets>`.
+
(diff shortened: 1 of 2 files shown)


docs/web_reference.rst
```diff
@@ -844,13 +844,18 @@ WebSocketResponse
 
       :raise ValueError: if data is not serializable object
 
-      :raise TypeError: if value returned by :term:`dumps` is not :class:`str`
+      :raise TypeError: if value returned by `dumps` param is not :class:`str`
```
asvetlov: Double backticks should be used here: `dumps`

docs/web_reference.rst
```diff
@@ -932,7 +953,7 @@ json_response
                             dumps=json.dumps)
 
 Return :class:`Response` with predefined ``'application/json'``
-content type and *data* encoded by *dumps* parameter
+content type and *data* encoded by `dumps` parameter
```
asvetlov: double backticks

mpaolini: fixed in next commit

asvetlov:
Thanks

Issue #1027, opened 2016-07-31
Implement StreamResponse.task attribute


asvetlov:
Access to the attr should be `None` if response wasn't prepared or was closed.

See also #922 discussion etc.

mpaolini:
watch out for circular references

asvetlov:
`ServerHttpProtocol` already has `._request_handler` attribute, we need to borrow it.

asvetlov:
Fixed by #1410

PR #1032, opened 2016-08-01
Web resp task

asvetlov:
## What do these changes do?

Add `resp.task` attribute.

It may help with cancelling long-running handlers like WebSockets and Long-Polling.

`resp.task.cancel()` will shutdown the connection

## Related issue number

POC for #1027

## Checklist


asvetlov:
@mpaolini I believe the patch doesn't add a circular reference: resp._task will be assigned to `None` at the end of request processing.

asvetlov:
The master was changed, task should be implemented in another way.
`ServerHttpProtocol` already has `._request_handler` attribute, we need to borrow it.

Issue #921, opened 2016-06-10
IsADirectoryError when accessing folder with a static route


SteffenDE:
When accessing a URL handled by a static route, which is a directory, the server responds with HTTP 200 AND HTTP 500 in the response body instead of just responding by e.g. Forbidden.

```
$ http http://localhost:5000/web/static/
HTTP/1.1 200 OK
CONTENT-LENGTH: 204
CONTENT-TYPE: application/octet-stream
DATE: Fri, 10 Jun 2016 17:27:48 GMT
LAST-MODIFIED: Wed, 08 Jun 2016 14:01:43 GMT
SERVER: Python/3.5 aiohttp/0.21.6

HTTP/1.1 500 Internal Server Error
CONTENT-TYPE: text/html; charset=utf-8
CONTENT-LENGTH: 170
CONNECTION: close
DATE: Fri, 10 Jun 2016 17:27:48 GMT
SERVER: Python/3.5 aiohttp/0.21.6


<html>
  <hea

```

The route is added by calling

``` python
app.router.add_static("/web/static", "web/static")
```

asvetlov:
Please make sure `web/static` both is a directory and web server process has appropriate rights.
As side note better to use absolute path to directory, it helps to avoid weird problems.

SteffenDE:
`web/static` is a directory and the web server has appropriate rights. Accessing files inside of the directory works fine.
I realized that I forgot to include the traceback:

```
======== Running on http://127.0.0.1:8888/ ========
(Press CTRL+C to quit)
Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/aiohttp/server.py", line 266, in start
    yield from self.handle_request(message, payload)
  File "/usr/local/lib/python3.5/site-packages/aiohttp/web.py", line 90, in handle_request
    resp = yield from handler(request)
  File "/usr/local/lib/python3.5/site-packages/aiohttp/web_urldispatcher.py", line 595, in handle
    with filepath.open('rb') as f:
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/pathlib.py", line 1136, in open
    opener=self._opener)
IsADirectoryError: [Errno 21] Is a directory: '/Users/Steffen/Desktop/static'
```

Using an absolute path does not solve the issue.

``` python
import asyncio
from aiohttp import web

app = web.Application()
[...]

PR #1033, opened 2016-08-01
#921 show directory index [WIP]

trimailov:
## What do these changes do?

- By default throws `HTTP 403 - Forbidden` if we access static directory
- If static directory indexing is enabled - lists the static directory contents when we access it

`UrlDispatcher.add_static` got new flag - `show_index`, which will show index of a directory if explicitly set to `True`. By default, accessing directory index is forbidden (HTTP/403, `show_index=False`)

## Related issue number

#921

## Checklist

- [ ] `StaticRoute._directory_as_html()`:
  - [ ] Maybe it should be outside the class? Unit-testing would be easier that way..
  - [ ] I would want a unit-test for this method to explicitly check that HTML is as we want
  - [ ] Method looks cumbersome and has lots of newlines, so the output would be human readable (is this a priority?)
  - Example output (from a test `test_access_root_of_static_handler`):
```
<html>
<head>
<title>Index of /</title>
</head>
<body>
<h1>Index of /</h1>
<ul>
<li><a href="/my_file">my_file</a></li>
</ul>
</body>
</html>
```
- [ ] There is currently missing coverage for [L506](https://github.com/trimailov/aiohttp/blob/show-dir-index/aiohttp/web_urldispatcher.py#L506) and [L531](https://github.com/trimailov/aiohttp/blob/show-dir-index/aiohttp/web_urldispatcher.py#L531-L532)
- [ ] Documentation should reflect the changes, though I have build problems ATM and cannot check docs as HTML

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -497,7 +497,7 @@ class StaticRoute(Route):
         # on opening a dir, load it's contents if allowed
         if filepath.is_dir():
             if self._show_index:
-                ret = Response(text=self._dir_index_html(filepath))
+                ret = Response(text=self._directory_as_html(filepath))
             else:
                 raise HTTPForbidden()
         elif filepath.is_file():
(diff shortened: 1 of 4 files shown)


asvetlov:
Fixed by #1094

Issue #895, opened 2016-06-02
Add shortcut for for request.app.router[path].url(**kw)


jettify:
## Long story short

Currently we need to write rather long line just to get reverse url:

``` python
async def handler(request):
    url = request.app.router['users'].url(**kw)
```

I propose to add shortcut function:

``` python

url_for(request, name, **kw)
```

abusaidm:
I second this.

asvetlov:
PR is welcome :)

asvetlov:
API for routes has changed, `aiohttp-jinja2` supports `url` context function. Subapplications added.
I don't feel we need shortcut like proposed.

PR #1034, opened 2016-08-01
implement Request.named_url and Application.named_url

berislavlopac:
## What do these changes do?

Implement the `named_url` method to `Application` and `Request` branches.

This simplifies constructing the URLs for named routes, as mentioned in #895:

``` python
# old way, still valid
url = request.app.router['users'].url(**kwargs)

# new way
url = request.named_url('users', **kwargs)
url = app.named_url('users', **kwargs)
```

It might not be a huge reduction in pure character count, but is simpler to remember. 
I'll also add the documentation if this is approved.
## Related issue number
#895
## Checklist


asvetlov:
I suggest using just function `named_url(request, name, **kwargs)` instead of request and app methods.

The reason is: router is a pluggable class. User may change it to aiotraversal for example.

Separate function adds less coupled dependency than method.

Maybe I'll change my opinion later, we can add methods then.
But now I have a gut feeling that the function is better.

berislavlopac:
Sure, didn't know that about router. Is there a proper module for such helper functions, or should it simply go to `web` (or maybe `web_reqrep`)?

asvetlov:
I believe `web.py` should be good place.
Later we can extract helpers into `web_helpers.py` if needed.

asvetlov:
Would you update documentation also?
`web.rst` and `web_reference.rst` should be touched.

berislavlopac:
I'm having some problems when running `make doc` to build the documentation; this is the log file:

```
# Sphinx version: 1.4.5
# Python version: 3.5.1 (CPython)
# Docutils version: 0.12 release
# Jinja2 version: 2.8
# Last messages:

# Loaded extensions:
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinx/cmdline.py", line 243, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinx/application.py", line 137, in __init__
    confoverrides or {}, self.tags)
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinx/config.py", line 292, in __init__
    execfile_(filename, config)
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinx/util/pycompat.py", line 130, in execfile_
    exec_(code, _globals)
  File "conf.py", line 63, in <module>
    import sphinxcontrib.spelling  # noqa
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinxcontrib/spelling/__init__.py", line 3, in <module>
    from .builder import SpellingBuilder
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/sphinxcontrib/spelling/builder.py", line 16, in <module>
    from enchant.tokenize import EmailFilter, WikiWordFilter
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/enchant/__init__.py", line 92, in <module>
    from enchant import _enchant as _e
  File "/Users/berislav.lopac/.virtualenvs/aiohttp/lib/python3.5/site-packages/enchant/_enchant.py", line 121, in <module>
    prefix_dir.contents = c_char_p(e_dir)
TypeError: bytes or integer address expected instead of str instance
```

Any ideas?

asvetlov:
Sorry for long silence.
I was working on YARL incorporation and nested application support.
Was not sure if these changes are compatible with the PR.
Hopefully they are.

@berislavlopac would you continue your work on PR?
Documentation should be updated as well (both `web.rst` for pointing best practice and `web_reference.rst`).

Enchant is disabled from docs build now AFAIK.

In aiohttp 1.1 name resolution is changed a little:
1. `.url()` is replaced by `.url_for()` which returns `yarl.URL` instance.
2. `request.app` points to the most nested application in the stack. That's fine for most users: nested app are considered as independent application without any knowledge about top-level ones.
But if the nested app _want_ to know about others it should use `named_url(other_app, 'name', key=value)`.

berislavlopac:
Yes, I definitely want to give another shot to building the docs etc; I've been insanely busy lately.

aiohttp/web.py
```diff
@@ -263,6 +263,24 @@ def __repr__(self):
         return "<Application 0x{:x}>".format(id(self))
 
 
+############################################################
+# Helper functions
+############################################################
+
+def named_url(app_or_request, name, **kwargs):
+    app = getattr(app_or_request, 'app', app_or_request)
```
asvetlov, line 271: `isinstance` check is safer than duck typing.
Parameter which is neither application nor request should raise `TypeError`

berislavlopac: Which is precisely why I wanted to make it a method on those two classes. :smile: 
But no worries, I'm adding the check (finally found the time to work on this).

AraHaan:
@berislavlopac Ok, so can you reaply the changes you did as well as @asvetlov's suggested changes to the latest master branch for inclusion and remake this PR to help make it mergeable with as little commits as possible?

asvetlov:
The patch is obsolete, it doesn't follow new API

PR #1036, opened 2016-08-02
let users of the AioHTTPTestCase class safely define setUp and tearDown

przerull:
## What do these changes do?

Allow developers to define the setUp and tearDown methods on subclasses of AioHTTPTestCase and not have to worry about calling supers or whatever to make sure that the client, loop, and app attributes are setup and torn down correctly.

I prefer to use the unittest framework for testing my personal code and I find that I routinely use setUp and tearDown on my unittest.TestCase subclasses, but I have never had a need to call setUpClass.

Since the unittest framework calls (some subclass of unittest.TestCase).setUpClass automatically, I thought this would be a handy place to decorate AioHTTPTestCase's setUp and tearDown methods.

 

This does not affect the end-user as this is only meant to provide an additional feature for developers using the unittest framework

## Related issue number

There are no open issues that this will resolve.

## Checklist

- [X] _(maybe...)_ Documentation reflects the changes

I wasn't sure if this needed to be explicitly documented. I figure most developers utilizing the unittest framework will assume that setUp and tearDown will behave normally.   However, if anyone feels that this needs extra documentation I'll be happy to write it up.

Thanks for your time considering my pull request.  I look forward to hearing from you.

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -472,14 +472,28 @@ class AioHTTPTestCase(unittest.TestCase):
         """
         pass
 
-    def setUp(self):
-        self.loop = setup_test_loop()
-        self.app = self.get_app(self.loop)
-        self.client = TestClient(self.app)
+    @classmethod
+    def setUpClass(cls):
+        super(AioHTTPTestCase, cls).setUpClass()
 
-    def tearDown(self):
-        del self.client
-        teardown_test_loop(self.loop)
+        def setup_decorator(method):
+            def decorated(self, *args, **kwargs):
+                self.loop = setup_test_loop()
+                self.app = self.get_app(self.loop)
+                self.client = TestClient(self.app)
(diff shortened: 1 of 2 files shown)


kxepal:
I think it's quite complicated and non oblivious to just avoid simple and common `super()` call when you do subclasses. You still would have to do that when you inheritance depth will cross over number 2 or if you will have to override `setUpClass` method to add few more own bits. So what's the profit?

przerull:
yeah I think you have a good point there.

I did something similar to this with django to ensure a mongodb test database it used was set back to empty between each test (similar to how django's test case does with django's internal orm).

In that situation it was probably more appropriate because it might not be obvious to developers that their tests are failing because their database doesn't have a consistent state between tests but in this case if the developer overrides setUp and tearDown and doesn't call a super it'll blow up in their face.

Closing this pull request.

Thanks for the input.

Issue #951, opened 2016-07-04
Need less restrictive HTTP method validation


wolfhechel:
First of all I'm sorry if this topic has been brought up before, I tried to search for a similar issue but couldn't find any.

As it is right now the only methods allowed for routing and requests are the ones defined in [RFC7231](https://tools.ietf.org/html/rfc5789) and [RFC5789](https://tools.ietf.org/html/rfc5789).
In reality there are far more methods used in the open like WebDAV methods (MOVE, COPY, UPDATE, LOCK, UNLOCK etc), GENA methods (SUBSCRIBE, UNSUBCRIBE, NOTIFY).

IANA [lists](http://www.iana.org/assignments/http-methods/http-methods.xhtml) the most common ones defined in RFC's at least, I was surprised to see GENA methods were missing though.

Point being, is there any reason as to why aiohttp is limited in this matter a part from the potential abuse in class-based views?

I know this is an issue already as I've done WebDAV and GENA work with aiohttp in projects, and as it is now I'm monkey-patching the method checks in order to allow route registration there.

wolfhechel:
Also I'm sorry this issue doesn't follow the issue template, but as this is a discussion more than a bug report I hope you can find it in your heart to forgive me.

asvetlov:
Please take a look on #960
Is the proposed solution satisfies your needs?

wolfhechel:
Indeed that solves it for class based views at least, it still doesn't allow regular views as that's limited by the AbstractRoute implementation.

Would it be of any harm to remove lines 84:86 from AbstractRoute, since technically the only definition of a HTTP method is that it only consists of alphanumeric characters.

Actually, according to the HTTP 1.1 RFC a method can be any number of digits, letters or the symbols !#$%&'*+-.^_`|~.

asvetlov:
My main reason for keeping checks was preventing user from silly types. For example I wanted to have a error on writing `app.router.add_route('POSTT', '/', handler)`.

Now thanks to #989 we have shortcut methods for server api, and we already had these for client.

So my objection has gone away, we can relay checks significantly.

@wolfhechel are you volunteered for this task?

wolfhechel:
@asvetlov Absolutely, I still think that methods should be validated according to RFC 1.1 specifications though so I'll implement that instead.

asvetlov:
Fixed by #1037

PR #1037, opened 2016-08-02
Relax HTTP method validation in UrlDispatcher.

wolfhechel:
## What do these changes do?

This pull request changes the HTTP method validation procedure in AbstractRoute.
Previously only "safe HTTP methods" were allowed which made developing unusual
services communicating over HTTP with custom methods (such as WebDAV and UPnP)
impossible without monkey-patching.

The new method AbstractRoute.validate_method() matches the method against a regex compliant with the RFC 2616 specification of a "token".

The only change in behavior is what input raises a ValueError now, this has not been documented anywhere as far as I could tell hence no changes has been documented in the end-user API docs.

## Related issue number

#951 

## Checklist

Previously the validation only compared against "safe HTTP methods",
now we validate against the allowed character-set defined in
RFC 2616 section 5.1.1.

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -30,6 +30,9 @@ __all__ = ('UrlDispatcher', 'UrlMappingMatchInfo',
 PY_35 = sys.version_info >= (3, 5)
 
 
+HTTP_METHOD_RE = re.compile(r"^[0-9A-Za-z!#\$%&'\*\+\-\.\^_`\|~]+$")
+
+
 class AbstractResource(Sized, Iterable):
 
     def __init__(self, *, name=None):
@@ -63,7 +66,6 @@ class AbstractResource(Sized, Iterable):
 
 
 class AbstractRoute(abc.ABC):
-    METHODS = hdrs.METH_ALL | {hdrs.METH_ANY}
 
     def __init__(self, method, handler, *,
                  expect_handler=None,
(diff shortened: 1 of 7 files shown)


wolfhechel:
Okay so I got the object name wrong, the change was in AbstractRoute, not in UrlDispatcher.
Sorry about that!

asvetlov:
Cool!
Would you relax client validation as well?

asvetlov:
Hmm. After rethinking I've decided to remove `validate_method()` and inline regexp check.

The reason is: I don't want to expose yet another _hook_ and don't want to document it.
I doubt we will find many usages for stricter validations: users may override method for relaxing validation rules but never do it for making system stronger.

Fixed by df7b41b

wolfhechel:
It's cool, it's just a habit of mine to separate this kind of logical step in case in the future additional checks not coverable by a single line is required.

Out of curiosity, do you document "protected" methods as well?

Although I realise the naming of the method does not indicate it being protected.

wolfhechel:
Where does the client validation live by the way? As far as I can tell there's no such validation.
Also while digging in the code I noticed, there doesn't seem to be a user-exposed way to perform arbitrary HTTP requests anyway.
The only methods allowed to be sent are the ones exposed through get, post, delete etc. methods from ClientSession.

I think there should be a request() method as well which allows the user to form his/her requests manually. Or did I miss this completely?

PR #1062, opened 2016-08-10
Remove decode parameter from ClientResponse.read

kxepal:
It had been deprecated for two years or so, it's time to move on.


kxepal:
Hm...it's already done and I forgot to fetch the latest changes. Nice!

PR #1072, opened 2016-08-12
Fix test_client fixture to allow multiple clients per test

popravich:
## What do these changes do?

`test_client` fixture now allows multiple client instances:

```python
from project import frontend, api

async def test_full_app(test_client):
    front_client = await test_client(frontend.make_app)
    api_client = await test_client(api.make_app)
    # test complete application
```

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -51,17 +51,17 @@ def loop():
 
 @pytest.yield_fixture
 def test_client(loop):
-    client = None
+    clients = []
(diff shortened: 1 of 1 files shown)


asvetlov:
Not bad but what's you use case?
I'm just curious.

popravich:
Use case is almost as in example above:
project has two servers - public and internal APIs.
I'm testing that touching private API properly propagates data to public API.

asvetlov:
lgtm

PR #1074, opened 2016-08-13
Add multipart coromethod to web request handler

### LINKED ISSUE #1067: will there be huge file upload support ?
when uploading huge file ,there is MemoryError as follows.
## any plan to fixed this?

ERROR:aiohttp.web:Error handling request
  File "D:\soft\prog\python3\python-3.5.1\lib\site-packages\aiohttp\server.py", line 285, in start
    yield from self.handle_request(message, payload)
  File "D:\soft\prog\python3\python-3.5.1\lib\site-packages\aiohttp\web.py", line 90, in handle_request
  File "D:\prog\comp.lang.python\廖2\webfx2.py", line 305, in logger
    return await handler(request)
  File "D:\prog\comp.lang.python\廖2\webfx2.py", line 320, in auth
    return (await handler(request))
  File "D:\prog\comp.lang.python\廖2\webfx2.py", line 338, in response
    r = await handler(request)
  File "D:\soft\prog\python3\python-3.5.1\lib\site-packages\aiohttp\web_urldispatcher.py", line 101, in handler_wrapper
    result = yield from result
  File "D:\prog\comp.lang.python\廖2\webfx2.py", line 211, in **call**
    params = await request.post()
  File "D:\soft\prog\python3\python-3.5.1\lib\site-packages\aiohttp\web_reqrep.py", line 347, in post
    body = yield from self.read()
  File "D:\soft\prog\python3\python-3.5.1\lib\site-packages\aiohttp\web_reqrep.py", line 307, in read
    body.extend(chunk)
MemoryError

kxepal:
This fixes #1067

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -16,7 +16,7 @@ from urllib.parse import parse_qsl, unquote, urlsplit
 
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
 
-from . import hdrs
+from . import hdrs, multipart
 from .helpers import reify, sentinel
 from .protocol import Response as ResponseImpl
 from .protocol import HttpVersion10, HttpVersion11
@@ -321,6 +321,11 @@ class Request(dict, HeadersMixin):
         return loads(body)
 
     @asyncio.coroutine
+    def multipart(self, *, reader=multipart.MultipartReader):
+        """Return async iterator to process BODY as multipart."""
+        return reader(self.headers, self.content)
+
+    @asyncio.coroutine
     def post(self):
         """Return POST parameters."""
         if self._post is not None:
@@ -335,6 +340,10 @@ class Request(dict, HeadersMixin):
                                  'multipart/form-data')):
             self._post = MultiDictProxy(MultiDict())
             return self._post
+
+        if self.content_type.startswith('multipart/'):
+            warnings.warn('To process multipart requests use .multipart'
+                          ' coroutine instead.', DeprecationWarning)
 
         body = yield from self.read()
         content_charset = self.charset or 'utf-8'
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -114,7 +114,7 @@ CHANGES
 
 - Streams documentation added #1150
 
--
+- Add `multipart` coroutine method for web Request object #1067
 
 -
 
--- a/docs/web.rst
+++ b/docs/web.rst
@@ -510,6 +510,7 @@ a container for the file as well as some of its metadata::
 
     async def store_mp3_handler(request):
 
+        # WARNING: don't do that if you plan to receive large files!
         data = await request.post()
 
         mp3 = data['mp3']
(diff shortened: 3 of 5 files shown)


kxepal:
cc @yinzuojie

asvetlov:
We need an usage example in `web.rst`

aiohttp/web_reqrep.py
```diff
@@ -321,6 +321,15 @@ def json(self, *, loads=json.loads, loader=None):
         return loads(body)
 
     @asyncio.coroutine
+    def multipart(self, *, reader=multipart.MultipartReader):
+        """Return async iterator to process BODY as multipart."""
+        if not self.content_type.startswith('multipart/'):
+            warnings.warn('Attempt to process multipart request with'
```
asvetlov: Is it possible to use `.multipart()` for `application/x-www-form-urlencoded` data?

kxepal: `application/x-www-form-urlencoded` - is about `key=value&anotherkey=anothervalue` format. It's not multipart thing. You probably mean `multipart/form-data` which is used to transfer binary data like files. 

From the [spec](https://www.w3.org/TR/html401/interact/forms.html):

> This attribute specifies the content type used to submit the form to the server (when the value of method is "post"). The default value for this attribute is "application/x-www-form-urlencoded". The value "multipart/form-data" should be used in combination with the INPUT element, type="file".

kxepal: For `application/x-www-form-urlencoded` I think we can have just `.form()` which returns `MultiDictProxy(MultiDict())` as like `.post()` and can be called multiple times.

asvetlov: I mean if `Content-Type` is not started with `multipart/` it's maybe not a warning but an exception?

kxepal: No, that's bad idea since content-type validation must be done before any read method call. Raising exception means we have double ctype validation. Also, people may use custom content-type for their requests like `applicaiton/vnd+multipart` or whatever which won't pass this check.

The reason of this warning is to notify that things may go wrong, but since you call that method we think that you know what are you doing. May be it's practically useless though.

asvetlov: Understood

asvetlov:
@kxepal could you add a test for fixing coverage issue.
I really don't want to reduce coverage level but very appreciate to increase it :)

kxepal:
@asvetlov Sorry, yes. And an example. I'll return to the issue in few days.

asvetlov:
thanks. I would like to have this PR in next release.

kxepal:
@asvetlov Done, sorry for delay. It seems windows tests fails for connector due to IPv4 vs IPv6. Is that known issue?

asvetlov:
I've added broken test yesterday, will fix it today ASAP.

kxepal:
Ok, thanks!

asvetlov:
I've fixed tests.
@kxepal could you take a look on coverage for new patch.
I have a dream to reach %100 sometimes, every coverage decrease moves us away from the target.

asvetlov:
Hmm, from first glance all your changes should be covered.

kxepal:
@asvetlov Yes, there shouldn't be a coverage misses. 100% is a really good goal!

asvetlov:
Looks good!
Please merge and update CHANGES.

asvetlov:
But I still don't see a way to process 2GB file without reading the whole content in memory.
multipart has methods for iterating over parts but consumes the part at all.
`reader.read_chunk(size)` doesn't respect `size` parameter :)

kxepal:
@asvetlov Hm...you scary me.

[read_chunk](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L265-L286) does respects that size parameter. And then all depends on what kind of request we process. 

If it provides content length we basically [read](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L288-L301) from `content` which is `StreamReader` of something of that kind. I don't recall any kind of buffering within it implementation, right?

Same story we do for [chunked transfer](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L303-L341), but with a tiny buffer equals a single chunk size since we eventually can read more than we need in single shot, so we need the way to roll back.

Did I miss something?

asvetlov:
Aah, sorry.
You are right. Nevermind.

kxepal:
Great! You just saved travis from test on transferring 2GiB of data ((:

But I found that I'm not able to pass more than 8MiB of data over test client using generator as data emitter. No errors around and warning, just bunch of data get missed.  Throwing all of them with single big blob of data works right. That's all strange.

kxepal:
Finally. I was chasing wrong reason of failure `/)_-`

asvetlov:
Thanks!

PR #1076, opened 2016-08-15
Remove pyenchant and sphinxcontrib-spelling dependencies

traff:
On Mac OSX pychant can't be install by pip, resulting in the following error:
"ImportError: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI."


asvetlov:
Thanks!

PR #1078, opened 2016-08-15
Rename msg.tp -> msg.type

### LINKED ISSUE #1019: Refactor websocket message type
1. `msg.tp` should be `msg.type`. Keep `.tp` alias
2. `.data` and `.extra` are not self-explained names. Let's keep it for backward compatibility but introduce new fields.
   2.1. `.result` in case of normal data frame
   2.2. `.exception` in case of error
   2.3. Special attribute `.closed` for closing message

ganadist:
## What do these changes do?

rename msg.tp field to msg.type

prefer to use msg.type, but user can still use msg.tp

## Related issue number

#1019

## Checklist

- [O] I think the code is well written
- [O] Unit tests for the changes exist
- [O] Documentation reflects the changes

fix #1019 (1)

Signed-off-by: Young Ho Cha <ganadist@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,11 +71,11 @@ This is simple usage example:
         await ws.prepare(request)
 
         async for msg in ws:
-            if msg.tp == web.MsgType.text:
+            if msg.type == web.MsgType.text:
                 ws.send_str("Hello, {}".format(msg.data))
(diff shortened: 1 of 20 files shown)


asvetlov:
Thanks you

Issue #1066, opened 2016-08-10
test_client should accept application object as parameter


mayurmahajan:
## Long story short

This is not an issue, but a feature improvement request. Per this documentation http://aiohttp.readthedocs.io/en/stable/testing.html test_client accepts a create_app function in it's constructor. It will be cool if test_client accepts application object instead.

This feature improvement is not critical. application object is used by any wsgi server to run a service. similarly this test_client should use an application object to create server/client. It's just more convenient to think and use.
## Expected behaviour

instead of

@pytest.fixture
def cli(loop, test_client):
    return loop.run_until_complete(test_client(create_app))

if test_client accepts app then 

@pytest.fixture
def cli(loop, test_client):
    return loop.run_until_complete(test_client(app))

## Actual behaviour

Behavior wise nothing changes, it's just the implementation change.

asvetlov:
Good idea.
Would you provide a Pull Request?
Old functionality should be supported too, we cannot make backward incompatible change without very strong reason.

popravich:
This can be easily achieved without changes to library:

``` python
def test_something(test_client):
    client = await test_client(lambda loop: app)
```

However allowing to pass application object instead of factory can lead to unexpected and
unpredictable errors -- application (or middlewares) may store some state between tests and so
using same input data in different test may lead to unexpected errors or application can be simply
attached to closed loop.
I'm a bit paranoid about it cause I've already seen such bugs.
So I think we can simply add some lines to docs saying how to pass single app instance to `test_client`.

asvetlov:
@popravich but people will either create app in test method or use a fixture for it.
I pretty sure they will never store an app in global var.
Well, they may do it but we all are adult guys.
Passing app instead of factory is more convenient way of thinking.
At least current design was my first spike when I've tried to use the fixture myself.
I've just missed this point on merge/review during PyCon sprints in Portland.

popravich:
> I pretty sure they will never store an app in global var.

Should I remind you several discussions with word "global" (remember aiopg global pool, recent aiohttp "how to get global app", etc)?
Besides in examples above I didn't find `app` to be a fixture )

Anyway, having this feature must require big red warning block in docs saying not to use global app
with test_client.

asvetlov:
I can live with a big red warning but really need the functionality.
Just compare:

```
async def test_1(test_client):
    async def handler(request):
         ...
    def create_app(loop, path, handler):
        app = web.Application(loop=loop)
[...]

PR #1083, opened 2016-08-15
Improvement test_client can accept instance object.

nexusz99:
## What do these changes do?

Improvement test_client can accept instance object as parameter.

No

## Related issue number

#1066 

## Checklist

- [O] I think the code is well written
- [O] Unit tests for the changes exist
- [O] Documentation reflects the changes

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -60,6 +60,10 @@ def test_client(loop):
             app = app_factory(loop, *args, **kwargs)
         else:
             app = app_factory
+
+        assert app.loop is loop, \
+            "Application is attached to other event loop"
+
         client = TestClient(app)
         yield from client.start_server()
         clients.append(client)
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 
 import pytest
+from aiohttp.web import Application
 
 from .test_utils import (TestClient, loop_context, setup_test_loop,
                          teardown_test_loop)
@@ -55,7 +56,10 @@ def test_client(loop):
 
     @asyncio.coroutine
     def _create_from_app_factory(app_factory, *args, **kwargs):
-        app = app_factory(loop, *args, **kwargs)
+        if not isinstance(app_factory, Application):
+            app = app_factory(loop, *args, **kwargs)
+        else:
+            app = app_factory
         client = TestClient(app)
         yield from client.start_server()
         clients.append(client)
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -59,6 +59,8 @@ def test_client(loop):
         if not isinstance(app_factory, Application):
             app = app_factory(loop, *args, **kwargs)
         else:
+            assert not args, "args should be empty"
+            assert not kwargs, "kwargs should be empty"
             app = app_factory
 
         assert app.loop is loop, \
--- a/tests/test_pytest_plugin.py
+++ b/tests/test_pytest_plugin.py
@@ -5,6 +5,8 @@ def test_myplugin(testdir):
     testdir.makepyfile("""\
 import asyncio
 import pytest
+from unittest import mock
+
 from aiohttp import web
 
 pytest_plugins = 'aiohttp.pytest_plugin'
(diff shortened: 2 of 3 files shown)


aiohttp/pytest_plugin.py
```diff
@@ -55,7 +56,10 @@ def test_client(loop):
 
     @asyncio.coroutine
     def _create_from_app_factory(app_factory, *args, **kwargs):
-        app = app_factory(loop, *args, **kwargs)
+        if not isinstance(app_factory, Application):
+            app = app_factory(loop, *args, **kwargs)
+        else:
+            app = app_factory
```
popravich, line 64: There **must** be assertion that `app.loop` is the same as used in test.

tests/test_pytest_plugin.py
```diff
@@ -32,7 +32,9 @@ def test_hello(test_client):
 
 @asyncio.coroutine
 def test_hello_with_loop(test_client, loop):
-    client = yield from test_client(create_app)
+    app = web.Application(loop=loop)
+    app.router.add_get('/', hello)
+    client = yield from test_client(app)
```
popravich, line 39: You shouldn't modify existing tests as `test_client` must support both,
so please add separate tests for this feature.

aiohttp/pytest_plugin.py
ppiyakk2, line 64: It means that there must be assertion like this?

```
assert app.loop is loop
```

tests/test_pytest_plugin.py
ppiyakk2, line 39: OK. I will fix it

aiohttp/pytest_plugin.py
popravich, line 64: Yes!
Plus message like this "Application is attached to other event loop"

ppiyakk2, line 64: Ok. Please check new commit.

popravich:
Looks good,
Can you also add a negative test -- passing to `test_client` application with wrong event loop
and checking assertion is raised

ppiyakk2:
@popravich How to check if assertion is raised? Just use try and except?

popravich:
``` python
with pytest.raises(AssertionError):
    test_client(web.Application(loop=fake_loop)
```

should work

ppiyakk2:
@popravich I added negative test case !

tests/test_pytest_plugin.py
```diff
@@ -122,4 +131,4 @@ def make_app(loop):
 
 """)
     result = testdir.runpytest('-p', 'no:sugar')
-    result.assert_outcomes(passed=8, failed=1)
+    result.assert_outcomes(passed=8, failed=2)
```
popravich: But why is it failing? If it fails it means that assertion is not raised.

nexusz99: Oh. That's my mistake. sorry.

tests/test_pytest_plugin.py
```diff
@@ -60,6 +62,13 @@ def test_hello_fails(test_client):
 
 
 @asyncio.coroutine
+def test_hello_from_app_fails(test_client):
+    with pytest.raises(AssertionError):
+        fake_loop = mock.Mock()
+        test_client(web.Application(loop=fake_loop))
```
popravich: `yield from test_client(...)` -- that's why this test fails

ppiyakk2: oh my god !! thanks !!!

aiohttp/pytest_plugin.py
asvetlov, line 64: Also please make sure that `args` and `kwargs` are both empty.

ppiyakk2, line 64: @asvetlov  When I debug the test case, parameter as a create_app function also args and kwargs are both empty.

aiohttp/pytest_plugin.py
```diff
@@ -55,7 +56,14 @@ def test_client(loop):
...
+
+        assert app.loop is loop, \
+            "Application is attached to other event loop"
```
asvetlov: Please add the following code after the line:
```
assert not args, "args should be empty"
assert not kwargs, "kwargs should be empty"

File: aiohttp/pytest_plugin.py (hunk 1 above)
Comment (asvetlov) [inline aiohttp/pytest_plugin.py:64]: I mean 

```
if isinstance(app_factory, Application):
    assert not args, "arg should be empty"
    assert not kwargs, "kwargs should be empty"
    ....
```

It will prevent cases like: `yield from test_client(app, 1, 2, param=3)`.

asvetlov:
Thanks!

Issue #1002, opened 2016-07-25
Websocket response .close() can hang indefinitely


mpaolini:
## Long story short

By looking at [this loop in the websockertesponse .close() code](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/web_ws.py#L190) it looks like a malicious client can hang the `.close()` method forever by sending regular messages every 10 seconds.
## Expected behaviour

WebSocketResponse `.close()` should timeout by default after some time.
## Actual behaviour

WebSocketResponse `.close()` can wait indefinitely for a close opcode, provided some other data arrives every 10 seconds.
## Steps to reproduce

None yet
## Your environment

linux/python35

asvetlov:
Good catch! I believe client code has the same loop.

It can be solved by either calculating total elapsed time for closing call or counting amount of loops.

@mpaolini are you interesting in providing a patch?

mpaolini:
@asvetlov yep, I can write one maybe later tonight

mpaolini:
won't have time in the near future (next month or so) for this

asvetlov:
Got it

asvetlov:
Fixed by #1084

PR #1084, opened 2016-08-15
Prevent infinite hang on WebSocket response.close() (#1002)

achimnol:
## What do these changes do?

`WebSocketResponse.close()` would no longer block indefinitely even when a malicious client sends bogus messages with an interval shorter than server-side close timeout.

No changes.

## Related issue number

#1002 

## Checklist

* The close() method now checks the time elapsed since the original
   starting time as well as the timeout for each iteration.

* Now close() should wait no more than at most twice of the timeout
   given to WebSocketResponse.

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -187,7 +187,8 @@ class WebSocketResponse(StreamResponse):
             if self._closing:
                 return True
 
-            while True:
+            begin = self._loop.time()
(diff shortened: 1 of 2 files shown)


asvetlov:
Thanks!

mpaolini:
Shouldn't we set self._close_code ? Also we should return true

mpaolini:
And self.exception should be set to a Timeout error just like the Timeout fired a few lines below

PR #1090, opened 2016-08-16
Update README.rst

yoland68:
aiohttp.web_urldispatcher.UrlDispatcher does not have add_get method


popravich:
It does have in master branch

PR #1094, opened 2016-08-18
#921 show directory index (from PR #1033)

tejksat:
## What do these changes do?

- By default throws `HTTP 403 - Forbidden` if we access static directory
- If static directory indexing is enabled - lists the static directory contents when we access it
- If server is not permitted to access a static directory then it responses with `HTTP 403 - Forbidden`

`UrlDispatcher.add_static` got new flag - `show_index`, which will show index of a directory if explicitly set to `True`. By default, accessing directory index is forbidden (HTTP/403, `show_index=False`)

## Related issue number

#921

## Checklist

Points taken from PR #1033:
- [x] `StaticRoute._directory_as_html()`:
 - [x] Maybe it should be outside the class? Unit-testing would be easier that way..
_(This is not necessary as discussed with @asvetlov)_
 - [x] I would want a unit-test for this method to explicitly check that HTML is as we want _(Check added)_
 - [x] Method looks cumbersome and has lots of newlines, so the output would be human readable (is this a priority?) _(It is OK as discussed with @asvetlov)_
 - [x] There is currently missing coverage for L506 and L531 _(Fixed)_
 - [x] Documentation should reflect the changes, though I have build problems ATM and cannot check docs as HTML _(Added)_


asvetlov:
Could you please cover missing three lines?
See https://codecov.io/gh/KeepSafe/aiohttp/compare/b75cb0a021fb9d60307305cba030793c3bf59d02...ee18fe35299655c5e8357c9d7b76cde7da68c3cc

tejksat:
@asvetlov done! Pls check.

asvetlov:
Thanks!

PR #1096, opened 2016-08-19
Support TestClient resource names

aamalev:
## What do these changes do?

Simplifies testing

Renamed first arg methods TestClient path -> path_or_name

## Related issue number

None

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -84,11 +84,19 @@ class TestClient:
         """
         return self._session
 
-    def make_url(self, path):
+    def make_url(self, path_or_name, parts=None):
+        if not path_or_name.startswith('/'):
+            name = path_or_name
+            url_kwargs = {}
+            if parts is not None:
+                url_kwargs['parts'] = parts
+            path = self.app.router[name].url(**url_kwargs)
+        else:
+            path = path_or_name
         return self._root + path
 
     @asyncio.coroutine
-    def request(self, method, path, *args, **kwargs):
+    def request(self, method, path_or_name, *args, **kwargs):
         """Routes a request to the http server.
 
         The interface is identical to asyncio.ClientSession.request,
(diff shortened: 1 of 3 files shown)


asvetlov:
What's you use case? 
Specifying real path instead of route name is more explicit, isn't it?
Also how to process query arguments when route name is specified?

asvetlov:
Hmm. I see.
`await client.get('personal', parts={'name': 'kitty'}, query={'a': 1})`
But in theory params from `ClientSession.request()` may clash with params used by url building.

aamalev:
```
 TypeError: request() got an unexpected keyword argument 'query'
```

"query" is not valid arg. For building url use "name" and "parts". To specify GET-parameters still need to use "params"

asvetlov:
They are not overlapping _right now_ but in future the clash is possible.
Mixing two different _logical_ namespaces in the same _parameters_ namespace is bad idea I believe.

aamalev:
What do you mean by mixing? In my commit no mixing

asvetlov:
I mean bringing up different domains (one for passing params into ClientSession and other for named route resolving) in the same space for `TestClient.get(...)` arguments is bad idea.
It's just my guts feeling.

I still not convinced why the PR makes sense at all.
`client.get('/path/to?a=b')` is very obvious.
```client.get('name', parts={'where': 'to'}, query={'a': 'b'})` looks overcomplicated.

When you write a test you are always know the name resolving, isn't it?

P.S.
I can accept `TestClient.resolve()` method.
Something like this:
`client.get(client.resolve('name', parts={'where': 'to'}, query={'a': 'b'}))`

**Comment by aamalev** [comment]
I agree that the mixing parameters. But does not agree that the option with the client.resolve simplifies testing:

```
url = test_client.resolve('parts', parts={'a': 1}, query={'a': 1})
url = test_client.app.router['parts'].url(parts={'a': 1}, query={'a': 1})
```

test_client.resolve does not shorten, it's useless.

The TestClient must be not just an http client, and should be primarily a tool for testing.
Support the "name" should be primary in comparison with a raw path. This eliminates the need to change tests when you change the schema url. Additionally, the parametrization of the url parameters is not a rare task.

```
@pytest.mark.parametrize(
    "user_id,name", [
        (1, 'joe'),
        (2, 'pit'),
    ])
async def test_dyn_named_resource_(test_client, user_id, name):
    parts = {'user_id': user_id, 'company': 2}
    resp = await test_client.get('participant', parts=parts, params={'a': 2})
    assert resp.status == 200
    result = await resp.json()
    assert result['name'] == name


@pytest.mark.parametrize(
    "path,name", [
        ('/company/2/participant/1?a=2', 'joe'),
        ('/company/2/participant/2?a=2', 'pit'),
    ])
async def test_dyn_named_resource_(test_client, path, name):
    resp = await test_client.get(path)
    assert resp.status == 200
    result = await resp.json()
    assert result['name'] == name
```

For me the obvious first option. Yes, you can write via router.url(...), but the idea of this PR is to make the tool easier

asvetlov:
Sorry.
I respect your strivings but still want to reject the PR.
You may use `url = test_client.app.router['parts'].url(parts={'a': 1}, query={'a': 1})` as workaround anyway.

Issue #1069, opened 2016-08-11
No event loop in the thread 'MainThread' when using test_client fixture


argaen:
## Long story short

When using the test_client pytest fixture and calling other functions that make use of the `get_event_loop` function internally like for example `asyncio.Task.current_task()` it raises the `RuntimeError: There is no current event loop in thread 'MainThread'` exception.

This forces the user to pass the loop explicitly to all calls but we are forcing the user to:
1. Modify the code in order to support testing with it
2. Passing explicitly the loop to all calls that need it which is not always possible because the `request` object is not propagated to all the code.
## Expected behaviour

Normal calls that require a loop should work without needing to pass the loop explicitly.

argaen:
Okay, so pulling out what I wrote here: https://groups.google.com/forum/#!topic/aio-libs/sHE53v8UaG0 which was the same case, the issue #939 and some hours of being lost I've got it working.

In order to have the event loop available to the whole code, here is a test example:

``` python
import pytest

from aiohttp.test_utils import loop_context

from myapp.api import app_factory

@pytest.yield_fixture
def loop():
    with loop_context() as loop:
        yield loop

@pytest.yield_fixture
def event_loop(loop):
    """
    This is needed for correct functioning of the test_client of aiohttp together with
    pytest.mark.asyncio pytest-asyncio decorator. For more info check the following link:
    https://github.com/KeepSafe/aiohttp/issues/939
    """
    loop._close = loop.close
    loop.close = lambda: None
    yield loop
    loop.close = loop._close

@pytest.mark.asyncio
async def test_valid_get_request(test_client, my_mock_data):
    client = await test_client(app_factory)
    resp = await client.get('/', params=my_mock_data)

    assert resp.status == 200
```

Things to note:
- We have to override the `event_loop` pytest-asyncio fixture (yeah, need to install pytest-asyncio). For more info check pytest-dev/pytest-asyncio#29.
- Note that you need `pytest-asyncio` and `pytest-aiohttp` so you have all the fixtures installed.
- With that, calls like `asyncio.Task.get_current()` or `aioes` calls will work without the need of explicitly having to pass the loop over your code.

As an observation, I think we should try to support this in aiohttp transparently without the need of doing this workarounds (i.e. using pytest-asyncio and overriding the event_loop fixture).

Hope this helps to someone else.

asvetlov:
Sorry, but I'm inclining to reject the feature request.
[...]

PR #1097, opened 2016-08-19
Added documentation about testing with the explicit loop

argaen:
## What do these changes do?

Updates on documentation about how to test by passing the loop explicitly and why you should do that.

## Related issue number

#1069


asvetlov:
Thanks!
Feel free for adding very explained texts about aiohttp problems and best usage practices like this one.

argaen:
Sure! :)

PR #1100, opened 2016-08-21
Fix typos in documentation

jwilk:
Found using [mwic](http://jwilk.net/software/mwic) and [anorack](https://github.com/jwilk/anorack).

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -101,11 +101,11 @@ WebSocket utilities
 
    .. attribute:: TEXT
 
-      Text messsage, the value has :class:`str` type.
+      Text message, the value has :class:`str` type.
 
    .. attribute:: BINARY
 
-      Binary messsage, the value has :class:`bytes` type.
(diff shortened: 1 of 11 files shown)


asvetlov:
Cool!
Would you make another PR for adding `make spell` command for spell checking?
We could even integrate spell check into travis-ci run, I believe it makes sense.

jwilk:
`make doc-spelling` already exists, but apparently nobody has run it for a very long time.
I'll try to knock it into shape. :)

f0t0n:
`make doc-spelling` doesn't work for me locally. Should we add `sphinxcontrib-spelling` to requirements-ci.txt?

asvetlov:
Hmm. it's tricky.

I use Ubuntu for local test runs but macosx people had blamed the `enchant` is not  available in their OS.

That's why we have removed `sphinxcontrib-spelling` from requirements.

Issue #1092, opened 2016-08-16
Allow to register application background tasks within an event loop


f0t0n:
## Long story short

 I run an aiohttp application with Gunicorn behind nginx.
The application exposes two websocket connection endpoints for mobile clients: one for RPC and other for Pub/Sub.

When some event is occurred in the system (e,g, there're some news from company for clients) I want to notify all the connected websockets that I store e.g. in my `app['pub_sub_websockets']`. So basically I would iterate them in a loop and send bytes to each one.

Each application instance that Gunicorn runs in a separate process has own collection of connected websockets. But I have to notify all the websockets connected to all workers (or even to all workers on all server nodes). Therefore I would proxy the event through some messaging system like Redis Pub/Sub or ZeroMQ. In my case it's ZeroMQ.

**The problem** is I can't easily setup the "listener" coroutine that will run within the application's event loop to subscribe to ZeroMQ proxy and forward messages to connected websockets.

Therefore I feel like I have to create own ZeroMQ socket zmq.SUB for each of connected websockets inside my websocket request handler and there gather two coroutines - the first will be listening on ws and the second one will be listening on ZeroMQ's socket. So that the amount of ZeroMQ's zmq.SUB sockets will grow linearly depending to the amount of clients connected to the application process via websockets.

Even with this approach I can't tell the application (actually `GunicornWebWorker`) to use [`zmq.asyncio.ZMQEventLoop`](http://pyzmq.readthedocs.io/en/latest/api/zmq.asyncio.html#zmq.asyncio.ZMQEventLoop) that's required to use with `zmq.asyncio`.
## Expected behaviour

**1)** I propose to provide an interface from the `web.Application` like `Application.register_background_task(coro)`  that will allow to register as many tasks as needed during the application instance setup to run along with `GunicornWebWorker._runner` within the [event loop](https://github.com/KeepSafe/aiohttp/blob/v0.22.5/aiohttp/worker.py#L36) the worker creates, say using `asyncio.gather()`.

In this way in my particular case I'd create only one ZeroMQ socket to listen for a topic that will run within the event loop instead of one ZMQ socket per one connected WebSocket.

And if I have to listen more event sources (message queues from different providers, etc.) I'll add one more socket (or other consuming object depending to the event source provider) and not one more thousand sockets for each thousand of currently connected websockets.

[...]

PR #1103, opened 2016-08-22
Define `web.Application.on_startup()` signal handler

f0t0n:
## What do these changes do?

This change introduces new signal `Application.on_startup()`
that allows to register as many `on_startup` signal handlers as needed to run in the event loop along with the application's request handler just after the application's startup.
This is useful to run background tasks such as listening to different event sources (message queues, etc.) to process within an application (e.g. to forward to clients connected via WebSocket).

The change doesn't impact on existing behavior but extends the API and the functionality.

## Related issue number

#1092

## Checklist

---

#### Changes done

- Define `web.Application.on_startup()` signal handler.
- Run `app.on_startup()` along with the request handler within an event
loop in `web.run_app()`.
- Tests for `Application.on_startup` signal
- Tests for multiple tasks to run on startup, including two long running
- Extend tests on `web.run_app()`: ensure that `Application.startup()`
is called within `web.run_app()`.
- Add documentation for `Application.on_startup` signal handler:
    - Describe possible use cases
    - Minor: fix typo

#### Notes

In the `tests/test_run_app.py` increased delay for `loop.call_later` from
`0.01s` to `0.02s` since with the old value loop used to stop
before coroutines gathered for `on_startup` finished, that caused
an exception.

As discussed earlier, the appropriate changes in the `aiohttp.worker.GunicornWebWorker` will be applied in a separate PR.

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -171,6 +171,7 @@ class Application(dict):
         self._on_pre_signal = PreSignal()
         self._on_post_signal = PostSignal()
         self._on_response_prepare = Signal(self)
+        self._on_startup = Signal(self)
         self._on_shutdown = Signal(self)
         self._on_cleanup = Signal(self)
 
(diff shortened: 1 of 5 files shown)


asvetlov:
Thanks you very much!

f0t0n:
Thank you for your help! :)

PR #1105, opened 2016-08-22
Call `Application.startup` in GunicornWebWorker

f0t0n:
## What do these changes do?

- Introduced the same behavior as in `web.run_app()` to call `Application.startup()` along with the request handler.
- Extended test case to check `worker.wsgi.startup` has been called once.

None
## Related issue number

#1092 

## Checklist

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -31,12 +31,11 @@ class GunicornWebWorker(base.Worker):
         super().init_process()
 
     def run(self):
+        self.loop.run_until_complete(self.wsgi.startup())
         self._runner = ensure_future(self._run(), loop=self.loop)
 
         try:
-            self.loop.run_until_complete(asyncio.gather(self._runner,
-                                                        self.wsgi.startup(),
-                                                        loop=self.loop))
+            self.loop.run_until_complete(self._runner)
         finally:
             self.loop.close()
 
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -34,7 +34,9 @@ class GunicornWebWorker(base.Worker):
         self._runner = ensure_future(self._run(), loop=self.loop)
 
         try:
-            self.loop.run_until_complete(self._runner)
+            self.loop.run_until_complete(asyncio.gather(self._runner,
+                                                        self.wsgi.startup(),
+                                                        loop=self.loop))
         finally:
             self.loop.close()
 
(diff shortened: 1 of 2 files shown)


asvetlov:
Please upgrade to master.

appveyor blames:

```
=================================== ERRORS ====================================
____________________ ERROR collecting tests/test_worker.py ____________________
Using @pytest.skip outside a test (e.g. as a test function decorator) is not allowed. Use @pytest.mark.skip or @pytest.mark.skipif instead.
```

But the master has no `pytest.skip` anymore in `test_worker.py`

f0t0n:
It did at https://github.com/KeepSafe/aiohttp/pull/1103 as well. The branch is rebased and is up to date.

aiohttp/worker.py
```diff
@@ -34,7 +34,9 @@ def run(self):
         self._runner = ensure_future(self._run(), loop=self.loop)
 
         try:
-            self.loop.run_until_complete(self._runner)
+            self.loop.run_until_complete(asyncio.gather(self._runner,
```
asvetlov: I suspect parallel execution by `gather` is not desired.
At least it works different that `web.run_app()`.
Better to wait for `self.wsgi.startup()` and than execute `self._runner` for opening server-side socket etc.

f0t0n: I'm not sure I get you right. Do you mean to just schedule `self.wsgi.startup()` with `ensure_future()` and keep the rest as it was done before?

```
ensure_future(self.wsgi.startup(), loop=self.loop)
self._runner = ensure_future(self._run(), loop=self.loop)

try:
    self.loop.run_until_complete(self._runner)
finally:
    self.loop.close()
```

asvetlov:
Every future worth to be awaited :)

```
self.loop.run_until_complete(self.wsgi.startup())
self._runner = ensure_future(self._run(), loop=self.loop)

try:
    self.loop.run_until_complete(self._runner)
finally:
    self.loop.close()
```

Initialize app on first step and run it on the second.
Wait for end of finalization process before starting the app.

f0t0n:
Ah, I already thought to `yield from self.wsgi.startup()` it in `self._run()`, let's keep your version.

aiohttp/worker.py
f0t0n: Fixed as you suggested.

asvetlov:
Thanks!

Issue #1104, opened 2016-08-22
Rewrite doc section about background tasks


asvetlov:
Hmm. After rereading your doc change I see `run_all_long_running_tasks` is not needed and using `gather` is confusing.
Why not just add 

```
app.on_startup.append(listen_to_zeromq)
app.on_startup.append(listen_to_redis)
```

Moreover, listeners are probably tasks not just coroutines.
So you should create a task and leave it running.
Also graceful shutdown is very important.

Most likely working code should look like:

```
import contextlib

async def start_backgrounds(app):
    app['redis_listener'] = app.loop.create_task(listen_to_redis())

async def cleanup_backgrounds(app):
    app['redis_listener'].cancel()
    with contextlib.suppress(asyncio.CancelledError):
        await app['redis_listener']

async def listen_to_redis():
    """Listen to messages from Redis Pub/Sub"""
    pass
```

@f0t0n would you provide a PR for doc fix?

f0t0n:
But if we do in this way, the loop inside the Signal will not continue till it awaits for `listen_to_zeromq` and won't get to `listen_to_redis`, right?

```
app.on_startup.append(listen_to_zeromq)
app.on_startup.append(listen_to_redis)
```

 https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/signals.py#L9

asvetlov:
Yes, that's why I've created new task in `start_backgrounds` from my code snippet.
It's encouraged way.
`gather` will wait for coroutines also BTW.

So from my perspective the doc example will wait forever in application initialization stage, isn't it?

asvetlov:
Related to #1092

f0t0n:
I'll create new PR for docs soon.

asvetlov:
thanks

f0t0n:
Hmm, I've just created an example with the code from my first version and it works well:

https://gist.github.com/f0t0n/f5bd3e41a85e70504ce3d249f5e8d14f

f0t0n:
But your example works much better. Namely it allows to register other quick running tasks after you register the long running.

https://gist.github.com/f0t0n/ca8bdef0ade1ba6358abdf8580ed9594

asvetlov:
Yes, sure.
But please keep doc example as short as possible.
The target is providing a simple code snippet but not describing all allowed opportunities.

asvetlov:
Fixed by #1107

PR #1107, opened 2016-08-22
#1104 Modify docs for background tasks

f0t0n:
## What do these changes do?

Changes:

- Modify docs for background tasks
- Add an example code to examples directory

As discussed the old code example for background tasks creating was working but not optimal.
Removed `asyncio.gather()` from the example and used `app.loop.create_task()` instead.
Also added an example of tasks cancellation on cleanup.

None

## Related issue number

#1104

## Checklist

### DIFF
--- a/examples/background_tasks.py
+++ b/examples/background_tasks.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+"""Example for aiohttp.web.Application.on_startup signal handler
+"""
+
+import asyncio
+from aiohttp import ClientSession
+from aiohttp.web import Application, run_app, Response
+
+
+async def fake_redis_msg(request):
+    return Response(text='fake Redis message...')
+
+
+async def fake_zmq_msg(request):
+    return Response(text='fake ZeroMQ message...')
+
+
+async def get_fake_message(where):
+    async with ClientSession(loop=app.loop) as session:
+        res = await session.get('http://127.0.0.1:8080/{}'.format(where))
+        return await res.text()
+
+
+async def quick_1(app):
+    for x in range(5):
+        print('quck_1')
+
+
+async def quick_2(app):
+    for x in range(5):
+        print('quck_2')
+        await asyncio.sleep(0.15)
+
+
+async def listen_to_redis():
+    try:
+        await asyncio.sleep(0.01)
+        while True:
+            print('Listening to Redis...')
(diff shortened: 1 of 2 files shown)


f0t0n:
@asvetlov Do you think it's short enough or I should remove something (https://github.com/KeepSafe/aiohttp/issues/1104#issuecomment-241540140)?

image

asvetlov:
First.
I think `quickly_notify_monitoring` is not necessary.
If you want to inform users to allowance for doing an instant activity on app startup -- please invite something more obvious than just `sleep`.
I have troubles with a good example personally,

Second.
Your code uses both zmq and redis. For sake of shortness we should choose the single one.
I believe redis is more well known library, let's keep it.

Redis example may be more realistic than, again, just sleeps in a loop.
Create and redis connection using aioredis, subscribe for events and iterate over subscribed messages.

I gives to reader a clue for real use cases.

f0t0n:
@asvetlov Thanks for guidance. I've just pushed new changes and rebased the branch to lastest master.

Currently it looks like this:

image

Also I've rewrote examples/background_tasks.py and ran it with redis to ensure that it works correctly. So the example in docs is an extraction from there and is working fine.

examples/background_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+async def listen_to_redis(app):
+    try:
+        sub = await aioredis.create_redis(('localhost', 6379), loop=app.loop)
+        ch, *_ = await sub.subscribe('news')
+        async for msg in ch.iter(encoding='utf-8'):
+            # Forward message to all connected websockets:
+            for ws in app['websockets']:
+                ws.send_str('{}: {}'.format(ch.name, msg))
+            print("message in {}: {}".format(ch.name, msg))
+    except asyncio.CancelledError:
```
asvetlov, line 34: `finally` is better than `except CancelledError`.
We should close redis subscription unconditionally.

f0t0n: Agreed. Gonna modify now.

f0t0n: Done.

f0t0n:
@asvetlov Should I squash commits or keep for history?

asvetlov:
Doesn't matter -- I'm doing squashing on merge anyway.

asvetlov:
Perfect!

PR #1109, opened 2016-08-23
Issue 1069/document explicit loop for tests

argaen:
Added the lambda example in the first `TestAcceptance` for consistency with the last full tests file example


argaen:
ermmm dunno why it's trying to add a commit already merged again :sweat:

asvetlov:
#1069

PR #1117, opened 2016-08-24
Fix access_log_format in `GunicornWebWorker`

### LINKED ISSUE #705: access log format is invalid when using gunicorn worker
It seems you have to pass in `--access-logformat='%a %l %u %t "%r" %s %b "%{Referrer}i" "%{User-Agent}i"'` to gunicorn for logging to work, they default format from gunicorn is `"%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"` which doesn't work with aiohttp

f0t0n:
## What do these changes do?

- This is a workaround for a problem described in #705.
It's intended to check `GunicornWebWorker.cfg.access_log_format`.
If its value is default Gunicorn's format string, it will be replaced with default aiohttp's log format string to create request handler with.
If it doesn't use default Gunicorn's format string but still sticks to Gunicorn's specification with format options in form of `%(name)s`, the `ValueError` will be thrown with an appropriate message.

- Additionally a note about this added to `Logging` section of the documentation.

Should fix #705 and show correct output in the access log.

## Related issue number

#705
## Checklist

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -2,18 +2,23 @@
 
 import asyncio
 import os
+import re
 import signal
 import ssl
 import sys
 
 import gunicorn.workers.base as base
 
-from aiohttp.helpers import ensure_future
+from gunicorn.config import AccessLogFormat as GunicornAccessLogFormat
+from aiohttp.helpers import AccessLogger, ensure_future
 
 __all__ = ('GunicornWebWorker', 'GunicornUVLoopWebWorker')
 
 
 class GunicornWebWorker(base.Worker):
+
+    DEFAULT_AIOHTTP_LOG_FORMAT = AccessLogger.LOG_FORMAT
+    DEFAULT_GUNICORN_LOG_FORMAT = GunicornAccessLogFormat.default
 
     def __init__(self, *args, **kw):  # pragma: no cover
         super().__init__(*args, **kw)
@@ -48,7 +53,8 @@ class GunicornWebWorker(base.Worker):
             timeout=self.cfg.timeout,
             keep_alive=self.cfg.keepalive,
             access_log=self.log.access_log,
-            access_log_format=self.cfg.access_log_format)
+            access_log_format=self._get_valid_log_format(
+                self.cfg.access_log_format))
 
     @asyncio.coroutine
     def close(self):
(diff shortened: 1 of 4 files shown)


f0t0n:
@asvetlov, does this change reflect the workaround [you suggested](https://github.com/KeepSafe/aiohttp/issues/705#issuecomment-242043986)?

aiohttp/worker.py
```diff
@@ -159,6 +161,11 @@ def _create_ssl_context(cfg):
             ctx.set_ciphers(cfg.ciphers)
         return ctx
 
+    def _fix_log_format(self, source_format, default_format):
+        if re.search(r'%\([^\)]+\)', source_format):
+            return default_format
```
asvetlov: Oh, no.
Silent format replacement should be performed only if source format is exactly _default gunicorn log format_, `%(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"` according to the doc.
For all other gunicorn-styled but not default formats `ValueError` should be raised.

f0t0n: Ha, now I got you. The comment in issue was unclear to me. Going to fix this according to your input. Thanks!

f0t0n: Seems this is done. Also added a note to docs "Logging" section.

image

aiohttp/worker.py
```diff
@@ -158,6 +164,15 @@ def _create_ssl_context(cfg):
             ctx.set_ciphers(cfg.ciphers)
         return ctx
 
+    def _get_valid_log_format(self, source_format):
+        if source_format == self.DEFAULT_GUNICORN_LOG_FORMAT:
+            return self.DEFAULT_AIOHTTP_LOG_FORMAT
+        elif re.search(r'%\([^\)]+\)', source_format):
+            raise ValueError('Gunicorn style using `%(name)s` '
+                             'is not supported for the log formatting.')
```
asvetlov: Please append  http://aiohttp.readthedocs.io/en/stable/logging.html#format-specification link to exception text.
Now exception states that something is wrong but doesn't point on solution.

But it's most likely configuration problem, not programming error.
The exception will be caught by admin or devops, not software developer.

I suggest to provide very detailed error info in these cases.

f0t0n: Agreed. Thanks for the point. Added a link and a text with correct directive.

asvetlov:
Thanks!

Issue #355, opened 2015-05-10
Double slash in path is meaningful (shouldn't be)


mickeyl:
I have a server listening for requests at /infastructure/events. For some reason, it doesn't catch requests sent to infrastructure//events. As far as I know this should work – at least all other web frameworks I tried seem to handle this case.

```
mickey@helios ~ $ http POST localhost:8091/infrastructure/events foo=bar
HTTP/1.1 200 OK
CONNECTION: keep-alive
CONTENT-LENGTH: 27
CONTENT-TYPE: application/json
DATE: Sun, 10 May 2015 11:31:04 GMT
SERVER: Python/3.4 aiohttp/0.15.3

{
    "data": null,
    "result": true
}

mickey@helios ~ $ http POST localhost:8091/infrastructure//events foo=bar
HTTP/1.1 404 Not Found
CONNECTION: keep-alive
CONTENT-LENGTH: 14
CONTENT-TYPE: text/plain; charset=utf-8
DATE: Sun, 10 May 2015 11:31:16 GMT
SERVER: Python/3.4 aiohttp/0.15.3

404: Not Found
```

What's your take on it?

Thanks.

asvetlov:
Normalization procedure from RFC 3986 doesn't allow to remove double-slashes in path, so `http://example.com/path/to` and `http://example.com/path//to` are different paths.
Why do we need join double slashes?

BTW, it can be done in custom middleware if needed.

ludovic-gasc:
@asvetlov maybe via a regex in the route ?

asvetlov:
No. I think catching NotFound error, collapsing slashes and redirect is better.
The same technique should be used for adding trailing slash.

Maybe we need include the middleware in aiohttp batteries, though.

ludovic-gasc:
Ok, I see.
BTW, I already had issues with Flask because it generates automagically a 301 redirect for all endpoints with a slash at the end.
You're right, an optional middleware seems to be the right decision.

kxepal:
`infrastructure//events` is not the same as `infrastructure/events` since `//` should be read as "path segment with zero length name". For most applications that makes no sense, however, you may find it quite often for "collection/element"` REST-like designs (because someone forgot to assert name against empty string).

+1 for optional middleware.

asvetlov:
Is there a volunteer for making a PR?
I can kindly review it :)

asvetlov:
[...]

Issue #682, opened 2015-12-15
modifying request object


ghost:
Hi,

The Request object is currently read only.
I think it would be usefull to be able to modify request headers in a middleware for example.
Being able to add stuff in the request object is nice, but so is having code that don't rely on it (that event works when the middleware is not there)

jashandeep-sohi:
Perhaps you are looking for [`Signals`](http://aiohttp.readthedocs.org/en/latest/web.html#signals)?

jashandeep-sohi:
@asvetlov Ahh, yes -- Been working with responses so much, every "r" word is a "response" in my head now ;-)

asvetlov:
Fixed by #1361

PR #1118, opened 2016-08-24
Added first version of normalize_path_middleware

argaen:
## What do these changes do?

Adds a `normalize_path_middleware` middleware that deals with trailing slashes and double slashes in path.

As a side effect, also `Request.clone()` was implemented.

Users will now be able to do:

``` python
from aiohttp import web
from aiohttp.middlewares import normalize_path_middleware
app = web.Application(middlewares=[normalize_path_middleware])
```
## Related issue number
#355
## Checklist
## Extra checklist
- [ ] Tests for `Request.clone`
- [x] Add args to middleware for append_slash and merge_slashes


argaen:
I'm adding the changes to look for feedback. Right now the test for the append_slash is working. I'll start now with the double slashes one.

asvetlov:
No, `request.clone()` shouldn't accept `RawRequestMessage` -- the later lies on very low abstraction level, much lower than all `aiohttp.web` operations.
The signature should be like this:
`request.clone(*, method=_sentinel, path=_sentinel, ...)`
Not needed to provide changing all request's attrs, let's limit ourself to obvious and most useful ones.
Overrides should be keyword-only.

aiohttp/middlewares.py
```diff
@@ -0,0 +1,39 @@
...
+
+        - Add a trailing slash to the path.
+        - Double slashes are replaced by one.
+    """
+
+    @asyncio.coroutine
+    def middleware(request):
+
+        router = request.app.router
+        match_info = yield from router.resolve(request)
```
popravich: `request` already has `match_info`, no need to resolve it twice

aiohttp/web_reqrep.py
```diff
@@ -381,8 +381,17 @@ def post(self):
         self._post = MultiDictProxy(out)
         return self._post
 
-    def copy(self):
-        raise NotImplementedError
+    def copy(self, *, app=None, message=None, payload=None, transport=None,
+             reader=None, writer=None, secure_proxy_ssl_header=None):
+        return Request(
```
popravich: `Request` is a dict-like object and other middlewares can modify it
so all data from original one must be moved to its clone.

asvetlov: Let's keep `.copy()` as is but create `.clone()` instead.
The methods have different semantics.

setup.cfg
```diff
@@ -1,8 +1,12 @@
+[pep8]
+max-line-length=80
```
asvetlov: Please keep 79: https://www.python.org/dev/peps/pep-0008/#id19

setup.cfg
```diff
@@ -1,8 +1,12 @@
...
+
 [easy_install]
 zip_ok = false
 
 [flake8]
 ignore = N801,N802,N803,E226
+max-line-length=80
```
asvetlov: 79

aiohttp/web_reqrep.py
```diff
@@ -381,6 +381,33 @@ def post(self):
...
+        :param path: str url path to use
+        :param headers: CIMultidictProxy or compatible containing the headers.
+        :param raw_headers: tuple of two element tuples containing the headers
+            as bytearrays.
+        """
+
+        message = RawRequestMessage(
+            method or self.method, path or self.path, self.version,
+            headers or self.headers, raw_headers or self.raw_headers,
+            self.keep_alive, None)
```
asvetlov: Oohh. Our very low level API is really bad.
I believe the `compression` param for this case is always `None` but why we need it here at all?
Nevermind.

asvetlov:
Looks pretty good but please convert `middlewares.normalize_path_middleware` into a factory.

I have a guts feeling: sometimes we'll need to add some parameters to the middleware, e.g.:

`app.middlewares.append(middlewares.normalize_path_middleware(add_trailing=True, remove_trailing=False))`

All params should be keyword-only arguments.

Right now it's not an issue but please respect my guts :)

Also `middlewares.normalize_path_middleware` looks redundant, just `middlewares.normalize_path` should be ok.

aiohttp/web_reqrep.py
argaen: I was wondering the same... Same with the headers and raw_headers. Why do we need to send both replicated? It would be cleaner to just send the `headers` and do the conversion (if ever needed)

argaen:
Yeah! In my current branch I have it as a factory but haven't commited it yet until I have it working with both args (`merge_slashes` and `append_slash`).

Ok with keyword only and Ok with renaming :).

aiohttp/web_reqrep.py
asvetlov: No. `headers` is `CIMultiDict` instance with `str` keys and values.
But `raw_headers` is a list or (header, value) pairs where both are unconverted `bytes`.
Not all possible header and value can be converted using UTF8 codec. Some old clients may send a data using custom encoding and the purpose of `raw_headers` is giving a way to handle this data.

argaen: Ahhh I see... Still feels bit redundant

argaen:
With this last commit functionality is covered. Anyway, in case it is ready to merge in your opinion, I would like to keep it until tomorrow night to see if I can find a more readable way to express it.

A bit off topic but that came while working with this merge request:
- Would it make sense to open a new issue for implementing an alternative `router.resolves`? Current implementation passing the Request object and then when return having to check with `isinstance` IMHO is not clean. It would be better to just send method and path (as you already started in your pull request) and then agree on some syntax to make it easier to check if resolved or not.
- Every time a request is resolved, the `request._match_info` has to be filled automatically. Would it make sense to do that inside the `resolve` method?

argaen:
Finished. We still have a caveat that I don't know how (or if) you want to address it.

Suppose the user has the following route defined: `/resource`.

If the user tries to access to `/resource/`, it will return a 404 no matter if the middleware is active or not. Right now we only deal with cases where we want to **append** trailing slash.

argaen:
I think this would also solve #682

setup.cfg
```diff
@@ -1,8 +1,12 @@
+[pep8]
```
asvetlov, line 1: Why did you add this option? 79 lines is default value.

setup.cfg
```diff
@@ -1,8 +1,12 @@
...
+max-line-length=79
+
 [easy_install]
 zip_ok = false
 
 [flake8]
 ignore = N801,N802,N803,E226
+max-line-length=79
```
asvetlov, line 9: Drop this line.

aiohttp/web_reqrep.py
```diff
@@ -381,6 +381,33 @@ def post(self):
         self._post = MultiDictProxy(out)
         return self._post
 
+    def clone(self, *, method=None, path=None, headers=None, raw_headers=None):
+        """
+        Creates and returns a new instance of Request object. If no parameters
+        are given, an exact copy is returned. If a parameter is not passed, it
+        will reuse the one from the current request object.
+
+        :param method: str http method
```
asvetlov: Don't use autodoc markup

aiohttp/middlewares.py
```diff
@@ -0,0 +1,69 @@
...
+
+        - Add a trailing slash to the path.
+        - Double slashes are replaced by one.
+
+    The middleware returns as soon as it finds a path that resolves
+    correctly. The order if all enable is 1) merge_slashes, 2) append_slash
+    and 3) both merge_slashes and append_slash. If the path resolves with
+    at least one of those conditions, it will redirect to the new path.
+
+    :param append_slash: If True append slash when needed. If a resource is
```
asvetlov: Drop markup

asvetlov: Drop markup

asvetlov: Drop markup

setup.cfg
argaen: People can have configurations at user home level.

asvetlov: Hmm. Ok, make sense.

setup.cfg
asvetlov: Got it. Please leave as is.

asvetlov:
Rename the module to `web_middlewares`,  import it in `web.py` and make accessible in web.py's `__all__`.
The following should work:

```
from aiohttp import web
web.normalize_path_middleware
```

aiohttp/middlewares.py
```diff
@@ -0,0 +1,69 @@
...
+    match_info = yield from request.app.router.resolve(alt_request)
+    alt_request._match_info = match_info
+
+    if not isinstance(match_info.route, SystemRoute):
+        return True, alt_request
+
+    return False, request
+
+
+def normalize_path(
```
asvetlov: Rename to `normalize_path_middleware`. The name is long but it will be used only in one place of program.
Adding `middleware` suffix explicitly describes the function job.

cynecx:
Is the clone function really necessary? As it may impose a performance issue if the server is under a huge amount of load. In my opinion a better approach would be to provide an overload function for `UrlDispatcher::resolve`, so it basically accepts `(method, raw_path)` which can be obtained by calling the appropriate member functions on a `request` object (Or we have to manually iterate over resources and call `resolve` which is again better than actually cloning the whole request object).

Also I am wondering if it's actually better to directly integrate this feature in aiohttp's core (One can enable this feature by a flag), so we can actually drop returning `HTTPMovedPermanently`.

For example the expected behavior for this would be:

A request something like this `Req { method="GET", path="///account//info/" }` would be normalized to `Req { method="GET", path="/account/info/" }`, then we will actually try to find an appropriate handler, this way we are avoiding returning a `HTTPMovedPermanently`.

Again, I think we we should do the normalization as early as possible to elimination possible performance (More like we are being repetitive).

cynecx:
I was working on an temporary approach (More like a hack) where I utilize a middleware which does the normalization then crafts a temporary object where `raw_path` and `method` are stored. Instead of returning a `HTTPMovedPermanently`, the middleware is actually trying to call the appropriate handler. This is the time where I noticed something, this approach might not work with `expect_handler`s because middlewares aren't applied here (https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/web.py#L81), so I guess this won't work without actually changing internal code.

argaen:
@cynecx I have this MR a bit abandoned because Im spending my time with other stuff right now. About what you comment, I remember my first approach trying what you said about not using the `HTTPMovedPermanently` but I found out something (maybe same as you) that made me go with the redirect.

On the other hand, I totally agree that an alternative `resolve` method where only the method and path are sent should exist.

asvetlov:
`.clone()` is implemented by #1361

AraHaan:
Wow looks interresting be nice if this PR could be updated so it can be Merged.

sloria:
This is a useful middleware. And if this gets integrated, I could get rid of my currently-broken implementation in [aiohttp_utils](https://github.com/sloria/aiohttp_utils). 👍

argaen:
I've upgraded the branch to resolve conflicts and apply some comments from @asvetlov. I'm having some problems from 5 test cases that didn't have before:

```python
self = <aiohttp.test_utils.TestServer object at 0x7f1108c474a8>, path = '//resource2//a//b/'

    def make_url(self, path):
        url = URL(path)
>       assert not url.is_absolute()
E       assert not True
E        +  where True = <bound method URL.is_absolute of URL('//resource2//a//b/')>()
E        +    where <bound method URL.is_absolute of URL('//resource2//a//b/')> = URL('//resource2//a//b/').is_absolute
```
I can change the test cases and instead of `//resource2//a//b/` use `///resource2//a//b/` but feels like cheating.


Also, I'm missing the documentation in web_reference.rst. Do you want me to copy paste the docstring there or include it or?

aiohttp/web_middlewares.py
```diff
@@ -0,0 +1,75 @@
...
+
+            if isinstance(request.match_info.route, SystemRoute):
+                paths_to_check = []
+                if merge_slashes:
+                    paths_to_check.append(re.sub('//+', '/', request.path))
+                if append_slash and not request.path.endswith('/'):
+                    paths_to_check.append(request.path + '/')
+                if merge_slashes and append_slash:
+                    paths_to_check.append(
+                        re.sub('//+', '/', request.path + '/'))
```
fafhrd91, line 63: i think `re.sub` should be compiled.

fafhrd91:
Create new section in web_reference "Middlewares", and add you doctoring.

aiohttp/web_middlewares.py
argaen: In terms of optimization is not a big thing: http://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

fafhrd91: ok

fafhrd91:
why `'//resource2//a//b/'` this test fails?

argaen:
Because of the following:

```python
tests/test_web_middleware.py:184: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
aiohttp/client.py:530: in __iter__
    resp = yield from self._coro
aiohttp/test_utils.py:249: in request
    method, self.make_url(path), *args, **kwargs
aiohttp/test_utils.py:237: in make_url
    return self._server.make_url(path)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <aiohttp.test_utils.TestServer object at 0x7fdf41c7a630>, path = '//resource2//a//b/'

    def make_url(self, path):
        url = URL(path)
>       assert not url.is_absolute()
E       assert not True
E        +  where True = <bound method URL.is_absolute of URL('//resource2//a//b/')>()
E        +    where <bound method URL.is_absolute of URL('//resource2//a//b/')> = URL('//resource2//a//b/').is_absolute

aiohttp/test_utils.py:82: AssertionError
```

If the url is absolute its not a valid url to query I don't know why. This is a behavior that was introduced with yarl because I don't remember this happening before.

fafhrd91:
i see. is it possible to skip `make_url` call with some param and construct it manually?

argaen:
I have no idea, the code for the `make_url` is

```
    def make_url(self, path):
        url = URL(path)
        assert not url.is_absolute()
        return self._root.join(url)
```

which doesn't accept extra parameter. I believe it could be a `yarl` bug (or feature). Out of scope of this MR IMO.

fafhrd91:
ok, i will fix that

fafhrd91:
@argaen i added tests that you deleted. some of them actually failing.
https://github.com/KeepSafe/aiohttp/tree/normalize-middleware-fixes

fafhrd91:
@argaen i need you to review failing tests

fafhrd91:
#1583

Issue #1121, opened 2016-08-25
Deprecate debug parameter from app.make_hander()


asvetlov:
Use `app.debug` as a value.
If `app.debug` conflicts with explicitly passes `debug` param -- raise `ValueError`.

The change is backward incompatible a little but it worth to be applied.

f0t0n:
And what about the same story for `loop` parameter?
Currently we pass application's loop to the `RequestHandlerFactory`'s constructor:

```
self._handler_factory(self, self.router, debug=self.debug, loop=self.loop, **kwargs)
```

But user reads in the documentation:

```
    kwargs –
            additional parameters for RequestHandlerFactory
            constructor.

```

So user could decide to pass a loop within `kwargs` that will cause a `TypeError: got multiple values for keyword arg...`.

So my idea is, if we always pass application's loop as a parameter to the `RequestHandlerFactory`'s constructor we should not allow user to do pass own loop as we did with `debug`.

Makes sense?

But the code of the `make_handler` will become ugly with all those checks for different parameters... :/

asvetlov:
`TypeError: got multiple values` is clean and obvious to me.
Let's keep it as is.

asvetlov:
Fixed by #1123

PR #1123, opened 2016-08-25
#1121 Deprecate `debug` parameter of make_handler()

f0t0n:
## What do these changes do?

- Deprecate `debug` parameter of `Application.make_handler`
- Add docs for `debug` parameter in the `Application`'s constructor
- Fix docs display for `Application`'s `loop` parameter
- Add docs for `Application.debug` attribute
- Add a note about the deprecation of the `debug` parameter for
    `Application.make_handler()`

From now `Application.make_handler` will use application's debug mode.
If `debug` parameter is passed to this method and differs with the value of `Application.debug`,
the `ValueError` will be raised.

## Related issue number

#1121
## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -6,6 +6,7 @@ from importlib import import_module
 
 from . import hdrs, web_exceptions, web_reqrep, web_urldispatcher, web_ws
 from .abc import AbstractMatchInfo, AbstractRouter
+from .helpers import _sentinel
 from .log import web_logger
 from .protocol import HttpVersion  # noqa
 from .server import ServerHttpProtocol
@@ -221,8 +222,23 @@ class Application(dict):
         return self._middlewares
 
     def make_handler(self, **kwargs):
-        return self._handler_factory(
-            self, self.router, loop=self.loop, **kwargs)
+        debug = kwargs.pop('debug', _sentinel)
+        if debug is not _sentinel:
+            warnings.warn(
+                "`debug` parameter is deprecated. "
+                "Use Application's debug mode instead", DeprecationWarning)
+            if debug != self.debug:
+                raise ValueError(
+                    "The value of `debug` parameter conflicts with the debug "
+                    "settings of the `Application` instance. The "
+                    "application's debug mode setting should be used instead "
+                    "as a single point to setup a debug mode. For more "
+                    "information please check "
+                    "http://aiohttp.readthedocs.io/en/stable/"
+                    "web_reference.html#aiohttp.web.Application"
+                )
+        return self._handler_factory(self, self.router, debug=self.debug,
+                                     loop=self.loop, **kwargs)
 
     @asyncio.coroutine
     def startup(self):
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -1026,16 +1026,16 @@ Although :class:`Application` is a :obj:`dict`-like object, it can't be
 duplicated like one using :meth:`Application.copy`.
 
 .. class:: Application(*, loop=None, router=None, logger=<default>, \
-                       middlewares=(), **kwargs)
+                       middlewares=(), debug=False, **kwargs)
 
    The class inherits :class:`dict`.
 
    :param loop: :ref:`event loop<asyncio-event-loop>` used
-                for processing HTTP requests.
+    for processing HTTP requests.
 
-                If param is ``None`` :func:`asyncio.get_event_loop`
-                used for getting default event loop, but we strongly
-                recommend to use explicit loops everywhere.
+    If param is ``None`` :func:`asyncio.get_event_loop`
+    used for getting default event loop, but we strongly
+    recommend to use explicit loops everywhere.
 
    :param router: :class:`aiohttp.abc.AbstractRouter` instance, the system
                   creates :class:`UrlDispatcher` by default if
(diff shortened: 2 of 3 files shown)


f0t0n:
@asvetlov Please check if this reflects your vision.

Tests are coming soon.

f0t0n:
Added test coverage. Ready for review.

aiohttp/web.py
```diff
@@ -221,6 +221,22 @@ def middlewares(self):
         return self._middlewares
 
     def make_handler(self, **kwargs):
+        try:
+            debug = kwargs['debug']
```
asvetlov: I prefer `debug = kwargs.pop('debug', _sentinel)`
Please use this form, it's a little shorter and faster than exception based approach.

aiohttp/web.py
```diff
@@ -221,6 +221,22 @@ def middlewares(self):
...
+                raise ValueError(
+                    "The value of `debug` parameter conflicts with the debug "
+                    "settings of the `Application` instance. "
+                    "The application's debug mode setting should be used "
+                    "instead as a single point to setup a debug mode. "
+                    "For more information please check "
+                    "http://aiohttp.readthedocs.io/en/stable/"
+                    "web_reference.html#aiohttp.web.Application"
+                )
+        kwargs['debug'] = self.debug
```
asvetlov: just call `self._handler_factory(self, self.router, debug=debug, loop=self.loop, **kwargs)` instead of adding the flag to `kwargs` back.

aiohttp/web.py
f0t0n: In the issue description you said:

> If app.debug conflicts with explicitly passed debug param -- raise ValueError.

So that if we put default boolean value to `kwargs.pop()` we can't know if the key was in the dictionary or not.
Also we want to compare this value to `self.debug` and raise a `ValueError` if they differ but only if the value was there.
Of course we can use `None` as a default value and then check for `None`.

```
debug = kwargs.pop('debug', None)
if debug is not None and debug != self.debug:
    raise ValueError("...")
return self._handler_factory(self, self.router, debug=self.debug, loop=self.loop, **kwargs)
```

asvetlov: `helpers._sentinel` is a marker which could be used even if `None` is proper value.
Not necessary for boolean parameters though.

f0t0n: Ah, got you. Didn't see that object before. Thanks.

f0t0n: The changes you've suggested are done.

docs/web_reference.rst
```diff
@@ -1128,6 +1135,13 @@ duplicated like one using :meth:`Application.copy`.
       :param kwargs: additional parameters for :class:`RequestHandlerFactory`
                      constructor.
 
+      .. note::
```
asvetlov: Use `.. deprecated:: 1.0` directive.

aiohttp/web.py
```diff
@@ -221,8 +222,18 @@ def middlewares(self):
         return self._middlewares
 
     def make_handler(self, **kwargs):
-        return self._handler_factory(
-            self, self.router, loop=self.loop, **kwargs)
+        debug = kwargs.pop('debug', _sentinel)
```
asvetlov, line 225: Raise `DeprecationWarning` if `debug` is not `_sentinel` like [here](https://github.com/KeepSafe/aiohttp/blob/1a64fa70b819998f165f3e8e1f0395048bb879b6/aiohttp/client.py#L685)

Also explicit test should check if the warning was raised like [here](https://github.com/KeepSafe/aiohttp/blob/471df5c4c9bf54a7843ae97c0d48b8182eda2609/tests/test_web_application.py#L66)

docs/web_reference.rst
f0t0n: I thought to do this but the parameters of `make_handler` are not described and there're just `**kwargs`.
I can add `debug` to the method signature and `debug` parameter description and deprecation hint but I don't think it's good idea to add deprecated parameter to signature now (before the release which deprecates this).

If I just use `.. deprecated:: 1.0` and the text to me it looks ugly and could be confusing.

image

What do you think?

f0t0n: > could be confusing

Namely user may ask "what's deprecated, kwargs or make_handler method? o_O".

aiohttp/web.py
f0t0n: Done.

docs/web_reference.rst
asvetlov: Just add identation to push deprecation directive under parameter spec.

f0t0n: Ok, done.

image

I just doubt if this could confuse user and he could briefly read this as

```
kwargs - additional parameters for RequestHandlerFactory constructor.

    deprecated since 1.0: (blah-blah tl;dr!)
```

and he thinks "ok probably those kwargs are deprecated".

aiohttp/web.py
f0t0n: By the way should `_sentinel` be named as `sentinel` if we expose/use it outside the `helpers` module?

asvetlov: No objections.

docs/web_reference.rst
asvetlov: The best solution is explicitly document all available parameters.
It makes sense not only for getting rid of weird deprecation.
I suspect the most part of users just don't know what parameters are available.

But it's a subject for another PR.
Feel free to create it if you want.

asvetlov:
Thanks!

Issue #944, opened 2016-06-29
Documentation Error for Response parameters


frederikaalund:
Currently, the [documentation](http://aiohttp.readthedocs.io/en/stable/web_reference.html#aiohttp.web.Response) for

```
class aiohttp.web.Response(*, status=200, headers=None, content_type=None, charset=None, body=None, text=None)
```

states that

> content_type (str) – response’s content type. 'text/plain' if text is passed also, 'application/octet-stream' otherwise.

However, if I do not set the `text` argument (but set the `body` argument instead) the `content_type` is not set at all (whereas the documentation says it should be automatically set to `application/octet-stream`).

Python 3.5.1 and aiohttp 0.21.0

avamsi:
@asvetlov What is the intended behavior in this case?
I mean, is this a documentation error or implementation error?

asvetlov:
It's implementation bug.
Would you create  a  Pull Request for fixing it?
Default `content-type` for prepared response should be `application/octet-stream` if the content type was not specified.

avamsi:
Sure thing

asvetlov:
Fixed by #1124

PR #1124, opened 2016-08-25
Fix for #944

avamsi:
## Related issue number
#944 

## Checklist

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -649,7 +649,7 @@ class StreamResponse(HeadersMixin):
         if self._resp_impl is not None:
             if self._req is not request:
                 raise RuntimeError(
-                    'Response has been started with different request.')
+                    "Response has been started with different request.")
             else:
                 return self._resp_impl
         else:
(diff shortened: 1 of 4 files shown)


avamsi:
@asvetlov Just saw the failing tests. Sorry about that.
If you can take a look at the code changes and confirm everything's okay, I'll update affected tests accordingly.

aiohttp/web_reqrep.py
```diff
@@ -774,7 +774,7 @@ def __init__(self, *, body=None, status=200,
         self.set_tcp_cork(True)
 
         if body is not None and text is not None:
-            raise ValueError("body and text are not allowed together.")
+            raise ValueError('body and text are not allowed together')
```
asvetlov: Please use `"` for exception messages

aiohttp/web_reqrep.py
```diff
@@ -794,25 +794,23 @@ def __init__(self, *, body=None, status=200,
                 self._content_dict = {'charset': charset}
                 self.body = text.encode(charset)
             else:
-                self.text = text
                 if content_type or charset:
-                    raise ValueError("Passing both Content-Type header and "
-                                     "content_type or charset params "
-                                     "is forbidden")
+                    raise ValueError('passing both Content-Type header and '
```
asvetlov: `"`

aiohttp/web_reqrep.py
```diff
@@ -794,25 +794,23 @@ def __init__(self, *, body=None, status=200,
...
+                self.text = text
         else:
             if hdrs.CONTENT_TYPE in self.headers:
                 if content_type or charset:
-                    raise ValueError("Passing both Content-Type header and "
-                                     "content_type or charset params "
-                                     "is forbidden")
-            if content_type:
-                self.content_type = content_type
+                    raise ValueError('passing both Content-Type header and '
```
asvetlov: `"`

aiohttp/web_reqrep.py
```diff
@@ -794,25 +794,23 @@ def __init__(self, *, body=None, status=200,
...
+                                     'content_type or charset params '
+                                     'is forbidden')
+            if content_type is None:
```
asvetlov: I think this code should be in `StreamResponse` ctor.
`Content-Type` should be `application/octet-stream` if not explicitly specified.
It's true for all responses.

See https://tools.ietf.org/html/rfc7231#section-3.1.1.5

aiohttp/web_reqrep.py
avamsi: I thought the difference in quotes usage was due to different contributors. I'll change it back

asvetlov:
Please don't forget about:

> I think this code should be in StreamResponse ctor.
> Content-Type should be application/octet-stream if not explicitly specified.
> It's true for all responses.
> 
> See https://tools.ietf.org/html/rfc7231#section-3.1.1.5

asvetlov:
Thanks!

PR #1126, opened 2016-08-26
More convenient `make test`

f0t0n:
## What do these changes do?

- Allow to pass `TESTS` argument to `make test` and `make vtest` targets.
For example:

Following pytest documentation at
http://doc.pytest.org/en/latest/usage.html#specifying-tests-selecting-tests

Run only selected test case: `make test TESTS=test_worker.py::test_run`
Run only selected test module: `make test TESTS=test_worker.py`
Run all tests: `make test`.

- Documented in "Contributing" section.

image

This makes development faster in some cases.
E.g. when you want to run only one test case or one test module quickly to check your changes in the code or in the test itself and don't want to run all tests after each little change it's more convenient and fast to run only the test case/module you're interested in.

## Related issue number

None

## Checklist

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -101,6 +101,20 @@ Please take a look on the produced output.
 
 Any extra texts (print statements and so on) should be removed.
 
+.. tip::
+  You can also run a single test module or a single test case
+  by passing a ``TESTS`` argument as shown below.
+
+  .. code-block:: shell
+
+      $ make test TESTS=test_worker.py
+      $ make test TESTS=test_worker.py::test_run
+
+  .. seealso::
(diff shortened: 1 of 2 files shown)


asvetlov:
Sorry, no.
aiohttp had functionaly like this ages ago but later we've removed it.
Most likely you need to provide more than just test selector, e.g. `-s` and `-p no:timeout` flags.
Even for selecting tests you may use `-k` param.

Moreover, for sake of shortening single test running time you most likely want to skip flake8/isort checks etc.

Conclusion: if you need something more complex than just `make test` does please don't afraid to run `py.test tests` directly.

f0t0n:
@asvetlov Got you. Actually I already use raw `py.test` command to run a single test. Thanks for the explanation.

PR #1127, opened 2016-08-26
Update testing suite

asvetlov:
## What do these changes do?

* Add `TestServer` class and test_server fixture
* Refactor `TestClient`: 
  * `.close` is a coroutine now
  * ctor doesn't create server even if loop is not running
* Move a trick about monkey-patching service into FAQ section

Extracting test server fixture allows to pass server options into `.make_handler()` call.

`TestClient` refactoring is not backward compatible but it affects the only use case where the class is used directly.

Usually people use either `test_client` fixture or `AioHTTPTestCase` base class.

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -453,10 +453,6 @@ def make_mocked_request(method, path, headers=None, *,
                   transport, reader, writer,
                   secure_proxy_ssl_header=secure_proxy_ssl_header)
 
-    assert req.app is app
-    assert req.content is payload
-    assert req.transport is transport
-
     return req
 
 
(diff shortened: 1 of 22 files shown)


aiohttp/test_utils.py
```diff
@@ -34,6 +34,67 @@ def unused_port():
         return s.getsockname()[1]
 
 
+class TestServer:
+    def __init__(self, app, *, protocol="http", host='127.0.0.1'):
+        self.app = app
+        self._loop = app.loop
+        self.port = unused_port()
```
popravich: Grabbing `unused_port` in `__init__` may cause tricky bugs if `TestServer` is used explicitly (not through `test_server` fixture)

asvetlov: Good point

PR #1135, opened 2016-08-28
Change can_start to can_prepare in ws example

greghaynes:
WebSocketResponse.can_start is deprecated and should be replaced by
can_prepare. Lets make our example reflect this.


asvetlov:
Thanks!

PR #1136, opened 2016-08-29
use aiodns.DNSResolver.gethostbyname()

Martiusweb:
aiodns 1.1 has been released this weekend, it adds the support of `DNSResolver.gethostbyname()`. This PR updates `aiohttp.resolver.AsyncResolver` to use it.

The main difference between `DNSResolver.query()` and `gethostbyname()` is that the latter works more like `getaddrinfo()` and reads `/etc/hosts` before it tries to issue a query to a DNS server. It effectively allows to use local hostnames such as `localhost`.

I don't know if it's better to require aiodns 1.1 or if the code should instead check if `gethostbyname()` is available and fallback to `query()` if not.  We can also add a third Resolver class instead of modifying AsyncResolver, but it don't think that's really useful.

In the first case, I will add the requirement notice in the documentation/readme/requirements, otherwise I can rewrite the PR.

### DIFF
--- a/aiohttp/resolver.py
+++ b/aiohttp/resolver.py
@@ -52,8 +52,25 @@ class AsyncResolver(AbstractResolver):
         self._loop = loop
         self._resolver = aiodns.DNSResolver(*args, loop=loop, **kwargs)
 
+        if not hasattr(self._resolver, 'gethostbyname'):
+            # aiodns 1.1 is not available, fallback to DNSResolver.query
+            self.resolve = self.resolve_with_query
+
     @asyncio.coroutine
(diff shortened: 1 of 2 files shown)


asvetlov:
Cool!
The patch should support both versions.
I suggest to check for `_resolver.gethostbyname` existence in the constructor an call appropriate method.

Martiusweb:
Hi,

I updated my patch, rebased master and eventually defeated the CI and coverage checks. Tell me if there is something missing.

asvetlov:
Thanks!

PR #1143, opened 2016-09-06
Make prettier urls if query is empty dict

xen:
## What do these changes do?

Make better urls for named routes when query is empty dict. 

Should not require to update or change existed code.

## Checklist


asvetlov:
Thanks!

Issue #1144, opened 2016-09-07
Incomplete Websockets Documentation


frederikaalund:
The current [websockets documentation](http://aiohttp.readthedocs.io/en/stable/web.html#websockets) only handles two message types: `MsgType.text` and `MsgType.error`. However, a runtime error may occur if `MsgType.closed` is not handled as well. This should be clear from the official documentation.

It took me a long time to figure out why my program returned errors because of this. Now that we are at it: Why not simply catch the `MsgType.closed` in the [`__anext__`](https://github.com/KeepSafe/aiohttp/blob/v0.22.5/aiohttp/web_ws.py#L309) function? It is unintuitive that `autoclose=True` (which is also the default) doesn't deal with `MsgType.closed` when it deals with `MsgType.close`. I realize that the former is your own internal message while the latter is part of the WS standard.

Side note: In the [web_ws example](https://github.com/KeepSafe/aiohttp/blob/v0.22.5/examples/web_ws.py#L33) this is implicitly handled correctly through an else clause.

I'm using Python 3.5.1 and Aiohttp 0.22.5.

asvetlov:
Agree with implicit catching closing message in `__anext__`.
Would you make a Pull Request?

frederikaalund:
> Agree with implicit catching closing message in **anext**.

Great.

> Would you make a Pull Request?

Yeah sure, I can do that.

asvetlov:
Fixed by #1145

PR #1145, opened 2016-09-07
Websockets: Stop iteration when connection closes.

frederikaalund:
## What do these changes do?

Stops `async for msg in ws` iteration when `ws` is closed.

The user no longer needs to handle `WSMsgType.CLOSED` events themselves (this was never documented anyhow).

## Related issue number

See issue #1144.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,7 +45,7 @@ CHANGES
 
 - Ensure TestClient HTTP methods return a context manager #1318
 
--
+- Websockets: Stop `async for` iteration when connection is closed #1144
 
 -
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -53,6 +53,7 @@ Erich Healy
 Eugene Chernyshov
 Eugene Naydenov
 Frederik Gladhorn
+Frederik Peter Aalund
 Gabriel Tremblay
 Gennady Andreyev
 Georges Dubus
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -188,6 +188,6 @@ class ClientWebSocketResponse:
         @asyncio.coroutine
         def __anext__(self):
             msg = yield from self.receive()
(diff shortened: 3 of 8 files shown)


asvetlov:
Please add a test.

asvetlov:
`client_ws.py` should be updated also.

frederikaalund:
> @frederikaalund did you give up the PR?

Sorry for the delay. It's been a busy month. I hope to have time to look at this on Monday.

asvetlov:
@frederikaalund ping

frederikaalund:
@asvetlov Sorry about the delay.
- I've updated `client_ws.py` as well.
- I've added two tests: One for the client interface and one for the server interface.

Due to the nature of the feature (which uses async for), said tests require Python 3.5 and above. Moreover, the tests will _not_ be able to run in Python 3.5 and below since the `async for` is not valid syntax in the previous versions. Of course, this only affects the development pipeline. I do not know of any workarounds to such syntax changes (I tried some `eval`/`exec` tricks but to no avail). I can't find a drop-in replacement for `async for` for Python 3.4 as can be done with `async def` and `async with`. If you know of one, I can update accordingly.

In summary, the test suite now requires Python 3.5 and above.

asvetlov:
Please move your tests into `tests/test_py35` folder.
This directory is ignored when tests are executed by python 3.4

frederikaalund:
> Please move your tests into tests/test_py35 folder.

Ah, so that is how it is done. I completely missed that.

asvetlov:
LGTM.
The last thing: rules for PR was changed. New rules are here: https://github.com/KeepSafe/aiohttp/blob/master/PULL_REQUEST_TEMPLATE.md
Would you update `CHANGES.rst` and `CONTRIBUTORS.txt` as well?

frederikaalund:
Everything should be in order now. Let me know if there is something else.

asvetlov:
Awesome!
Thank you very much!

Issue #1146, opened 2016-09-09
Redirects for HEAD requests should use HEAD


catlee:
## Long story short

I'm trying to do a HEAD request to a service that returns a 303 redirect. aiohttp then does a full GET on the new location.
## Expected behaviour

aiohttp should instead do a HEAD on the new URL.
## Actual behaviour

aiohttp does a GET on the new URL

The code responsible is here: https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client.py#L228
This references this PR from the requests library: https://github.com/kennethreitz/requests/pull/269/files. However, the requests library is preserving the HEAD method for subsequent requests.

asvetlov:
Makes sense.
Would you provide a fix?

PR #1147, opened 2016-09-09
Fix redirects for HEAD requests.

catlee:
## What do these changes do?

Fixes redirects for HEAD requests

HEAD requests that are redirected will result in HEAD requests to the redirected URLs instead of GET requsets.

## Related issue number

#1146 

## Checklist

Fixes: #1146


asvetlov:
Thanks!

PR #1150, opened 2016-09-10
Document streams

### LINKED ISSUE #1106: Explicitly document streams API
http://aiohttp.readthedocs.io/en/stable/api.html looks like a piece of garbage.

Especially when I look on streaming API
http://aiohttp.readthedocs.io/en/stable/api.html#module-aiohttp.streams

It mixes native private methods like `feed_eof()` and `feed_data()` with decided to be used by top-level programmer like `.read_any` etc.
Moreover it doesn't suggest to use `async for` for iterating over incoming data and so on.

The doc should get rid of autodoc usage and switch to explicit documentation writing (as Python itself does).

The far blue sky goal is getting rid of autodoc usage at all for the project.

asvetlov:
Explicitly document streaming API.

Fixes #1106

Also don't inherit `aiohttp.StreamReader` from `asyncio.StreamReader`.

The inheritance makes a mess.

For example `asyncio.StreamReader` has `readuntil()` method but calling it from `aiohttp.StreamReader` leads to crash because internal implementation differs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -107,7 +107,7 @@ CHANGES
 
 -
 
--
+- Default value for StreamReader.read_nowait is -1 from now
 
 -
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -284,7 +284,8 @@ class ClientRequest:
             if hdrs.CONTENT_LENGTH not in self.headers and not self.chunked:
                 self.headers[hdrs.CONTENT_LENGTH] = str(len(self.body))
 
-        elif isinstance(data, (asyncio.StreamReader, streams.DataQueue)):
+        elif isinstance(data, (asyncio.StreamReader, streams.StreamReader,
+                               streams.DataQueue)):
             self.body = data
 
         elif asyncio.iscoroutine(data):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -428,7 +428,8 @@ class ClientRequest:
                             'Bytes object is expected, got: %s.' %
                             type(result))
 
-            elif isinstance(self.body, asyncio.StreamReader):
+            elif isinstance(self.body, (asyncio.StreamReader,
+                                        streams.StreamReader)):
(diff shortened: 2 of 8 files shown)


fafhrd91:
there was reason for inheritance. when i added inheritance asyncio did check `isinstance(..., StreamReader)`. i don't know if it still in place

asvetlov:
There are two checks for `StreamReader` in client API.
Now both are relaxed to accept `asyncio.StreamReader` as well as `aiohttp.StreamReader` explicitly as request's BODY.
I expect no backward compatibility issues.

fafhrd91:
It is opossite, asyncio does not accept anything except asyncio.StreamReader

asvetlov:
Got it.
@fafhrd91 what `asyncio` API call does accept `asyncio.StreamReader` only?
And why do we need these call? 
I've grepped over asyncio source and have found nothing.

fafhrd91:
asyncio still checks for StreamReader instance in `streams.StreamWriter.__init__`

this change will break all installations.

asvetlov:
Yes, is has the check.
But what is your use case?
User very unlikely creates `asyncio.StreamWriter` manually (I mean direct `writer = asyncio.StreamWriter(transport, protocol, reader, loop)`) with passing `aiohttp.StreamReader` instance.
He uses either `asyncio.open_connection()` or `aiohttp` API.

fafhrd91:
Ok, I am fine with change

asvetlov:
@fafhrd91 thanks.
IIRC we've added overloaded `aiohttp.StreamWriter` later for solving another issue but now we really don't need to inherit from `asyncio.StreamReader`.

Honestly I feel the current aiohttp state (parsers-readers-writers-DataQueue) is messy.

Right now I have no proposal for fixing it but want to invest a time into.

I'm considering all mentioned parts as very deep implementation details, e.g. if we'll drop DataQueue  but incorporate processing into `aiohttp.StreamWriter` with keeping existing functionality nothing will fail, isn't it?

P.S.
The proposed change is not a subject for next release.

PR #1164, opened 2016-09-13
Improve speed of native WebSocket frame parsing

SethMichaelLarson:
## What do these changes do?

Improve speed for native parsing WebSocket frames using faster bit-manipulation and constants. Doesn't require any new documentation/unit-tests.

No changes in behavior.

## Related issue number

No issues are closed with this PR.

## Checklist

- [ ] Unit tests for the changes exist (Not needed?)
- [ ] Documentation reflects the changes (Not needed?)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -253,3 +253,5 @@ CHANGES
 - Add manylinux wheel builds
 
 - Dup a socket for sendfile usage #964
+
+- Improve speed of WebSocket frame parsing
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -105,6 +105,7 @@ Sebastian Hüther
 SeongSoo Cho
 Sergey Ninua
 Sergey Skripnick
+Seth Larson
 Simon Kennedy
 Sin-Woo Bang
 Stanislas Plum
--- a/aiohttp/_ws_impl.py
+++ b/aiohttp/_ws_impl.py
@@ -238,7 +238,7 @@ def parse_frame(buf, continuation=False):
     # read header
     first_byte, second_byte = yield from buf.read(2)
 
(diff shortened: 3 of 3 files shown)


asvetlov:
1. I like using hex numbers for bits, masks etc. Why did you replace them with decimals?
2. What is speed improvement? Do you have any benchmark?

sethmlarson:
@asvetlov I will change the decimal back to hex. 
The other changes are just using bit arithmetic instead of integer arithmetic. Simple timeit tests can show that this is substantially faster.

asvetlov:
If the difference is huge -- please provide a benchmark for the function.
I doubt that integer arithmetic is slower that bit based on current CPU.

sethmlarson:
The test that failed in Travis is an asyncio timeout test, not anything related to WebSockets.
You're right that the difference isn't huge, but free cycles are still free cycles.

Tested on an Intel Xeon X5650 6 x 2.67GHz
Before:

```
$ python3.5 -m timeit -s 's = 30' 's // 4'
10000000 loops, best of 3: 0.0644 usec per loop

$ python3.5 -m timeit -s 's = 3289475' 's % 4'
10000000 loops, best of 3: 0.0662 usec per loop
```

After:

```
$ python3.5 -m timeit -s 's = 30' 's >> 2'
10000000 loops, best of 3: 0.0549 usec per loop

$ python3.5 -m timeit -s 's = 3289475' 's & 3'
10000000 loops, best of 3: 0.0586 usec per loop
```

17% speed-up for // vs. >> and 12% speed-up for % vs. &

asvetlov:
I mean the benchmark for the whole `parse_frame`, not single math operation only.
Maybe rewriting with cython will make a value but your change is just invisible.

PR #1169, opened 2016-09-14
Support Content-Transfer-Encoding binary

### LINKED ISSUE #1168: MultipartWriter does not support binary content encoding
## Long story short

There is no support of binary encoding in MultipartWriter
Some services require binary encoding. Seems that easy to implement with just returning unprocessed chunks.
## Expected behaviour

Supported binary encoding
## Actual behaviour

Not implemented encoding
https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L858
## Steps to reproduce

Specify header 'Content-Transfer-Encoding' to 'binary' in MultiPartWriter

kxepal:
This fixes #1168

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -443,7 +443,7 @@ class BodyPartReader(object):
         Supports ``gzip``, ``deflate`` and ``identity`` encodings for
         `Content-Encoding` header.
 
-        Supports ``base64``, ``quoted-printable`` encodings for
+        Supports ``base64``, ``quoted-printable``, ``binary`` encodings for
         `Content-Transfer-Encoding` header.
 
         :param bytearray data: Data to decode.
@@ -477,6 +477,8 @@ class BodyPartReader(object):
             return base64.b64decode(data)
         elif encoding == 'quoted-printable':
             return binascii.a2b_qp(data)
+        elif encoding == 'binary':
+            return data
         else:
             raise RuntimeError('unknown content transfer encoding: {}'
                                ''.format(encoding))
@@ -854,6 +856,8 @@ class BodyPartWriter(object):
         elif encoding == 'quoted-printable':
             for chunk in stream:
                 yield binascii.b2a_qp(chunk)
+        elif encoding == 'binary':
+            yield from stream
         else:
             raise RuntimeError('unknown content transfer encoding: {}'
                                ''.format(encoding))
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -339,6 +339,15 @@ class PartReaderTestCase(TestCase):
         self.assertEqual(b'\xd0\x9f\xd1\x80\xd0\xb8\xd0\xb2\xd0\xb5\xd1\x82,'
                          b' \xd0\xbc\xd0\xb8\xd1\x80!', result)
 
+    def test_read_with_content_transfer_encoding_binary(self):
+        obj = aiohttp.multipart.BodyPartReader(
+            self.boundary, {CONTENT_TRANSFER_ENCODING: 'binary'},
+            Stream(b'\xd0\x9f\xd1\x80\xd0\xb8\xd0\xb2\xd0\xb5\xd1\x82,'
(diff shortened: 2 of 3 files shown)


asvetlov:
LGTM

aiohttp/multipart.py
```diff
@@ -854,6 +856,8 @@ def _apply_content_transfer_encoding(self, stream):
         elif encoding == 'quoted-printable':
             for chunk in stream:
                 yield binascii.b2a_qp(chunk)
+        elif encoding == 'binary':
+            yield from stream
```
AleksandrIakhnev, line 860: Does not work for me (ClientRequestError: Can not write request body'), we need:

```
 for chunk in stream:
      yield chunk
```

kxepal: Hm...ok, but why? This should be roughly the same.

AleksandrIakhnev: Have you tried to make actual POST request? It is exception here https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/client_reqrep.py#L427  (got: class 'generator'  - seems pretty clear that is not the same return

kxepal: Well, basically whole logic here is based around generators. Ok, will try to make a better test then to catch that problem. Thanks for report!

AleksandrIakhnev: Thank you for quick response!

kxepal: @aamalev Could you share the code you use to reproduce that issue? Just to make sure that we're on the same wave here.

AleksandrIakhnev: https://gist.github.com/AleksandrIakhnev/6541a3b3d47a84fc55ffd4790b8bb1e1

kxepal: Great, thanks a lot.

asvetlov:
@kxepal do you want to include the patch into aiohttp 1.0?
Since I've finished #1173 personally I have no show stoppers for next release.

kxepal:
@asvetlov Yes, let's do that. I'll try to get to the code soon to close this one.

aiohttp/multipart.py
kxepal: @AleksandrIakhnev Sorry, but I failed to reproduce your issue: script you provided works for me returning 405 response from Google. I also add [test](https://github.com/KeepSafe/aiohttp/pull/1169/files#diff-a5c957a9c2be1397119d89b7aea0a7efR182) about to make sure that it works right and expected.

Are you sure that nothing is missed there in your example?

AleksandrIakhnev: Sorry, seems that I've tested not on the latest version of aiohttp, after updating it returns 405 response.

asvetlov:
Now the PR is ready, isn't it?

aiohttp/multipart.py
kxepal: Great! (:

kxepal:
@asvetlov Indeed it is.

asvetlov:
Awesome!
Let me make a new release than.
Not all bugs are fixed and still there are improvements worth to be done but we've reached to the next milestone.
aiohttp is ready for 1.0 :)

kxepal:
@asvetlov Great! Do you plan to cut the new release today?

asvetlov:
Yes. We was waiting it for very long, at least couple weeks longer that I expected.

fafhrd91:
@kxepal hey, i am playing with rust async stuff. would you be interested in help me with mime type parsing?

kxepal:
@fafhrd91 I'm not very good with rust, but sure, ping me if you need a help - will try to make it.

fafhrd91:
@kxepal i am not good either, i just started week ago :)
basically, i need mimetype streaming parser, and it could be just general library.
my code is here https://github.com/fafhrd91/simplehttp

kxepal:
@fafhrd91 Ok, will check, thanks!

PR #1174, opened 2016-09-15
Conform to RFC3986 - do not include url fragments in requests

bmuller:
## What do these changes do?

This strips fragments from URLs before making requests.  This is the behavior of all major browsers and is described in [RFC3986](http://www.ietf.org/rfc/rfc3986.txt) Section 3.5.

Nope.

## Related issue number

None.

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -181,6 +181,7 @@ class ClientSession:
                 skip_headers.add(istr(i))
 
         while True:
+            url, _ = urllib.parse.urldefrag(url)
 
             cookies = self._cookie_jar.filter_cookies(url)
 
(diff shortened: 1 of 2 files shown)


bmuller:
As an example of why this is needed, right now `aiohttp` can't be used to download any post pages from tumblr.com because their server goes into an infinite redirect loop when a fragment is included.  Here's an example script:

``` python
import aiohttp
import asyncio

async def fetch(url):
            print(resp.history)

url = "http://best-of-tumblr.tumblr.com/post/150420362886"
asyncio.get_event_loop().run_until_complete(fetch(url))
```

The history that's output is just a list `max_redirects` long showing an infinite redirect loop.  If you use my fork, then there's only one redirect and everything works fine.

bmuller:
Done!  Let me know if you need anything else.

asvetlov:
Nothing more -- your PR is very clean and obvious.
Thanks!

bmuller:
Thanks so much for the quick turnaround!

PR #1195, opened 2016-09-20
fixes #1194

### LINKED ISSUE #1194: Server example on home page
There is a following example on the home page:

```
from aiohttp import web

async def handle(request):
    text = "Hello, " + name
    return web.Response(body=text.encode('utf-8'))

app = web.Application()
app.router.add_get('/{name}', handle)

web.run_app(app)
```

From my perspective this example has following problems:
1. There is no page at `http://localhost:8080` (it's 404), this also means `get('name', "Anonymous")` is useless
2. When I visit `http://locahost:8080/something` it's served with `Content-Type: application/octed-stream`, so it's not user-friendly (page is downloaded rather than shown). Adding a content-type of `plain` or `html` should help.

I have aiohttp 1.0.1 if that matters.

popravich:
Fixes for examples README and docs index page (#1194).

Also there are many examples using `web.Response(body=text.encode('utf-8'))`
and so making content_type fallback to `application/octet-stream`.
I guess all those examples should be fixed aswell. Any thoughts?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,7 @@ This is simple usage example:
     async def handle(request):
         name = request.match_info.get('name', "Anonymous")
         text = "Hello, " + name
-        return web.Response(body=text.encode('utf-8'))
+        return web.Response(text=text)
 
     async def wshandler(request):
         ws = web.WebSocketResponse()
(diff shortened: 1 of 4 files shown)


popravich:
ok

asvetlov:
But examples fixing may be an another PR I think.
Feel free to merge this one if you prefer making two PRs.

popravich:
I want to fix code blocks in docs using `body` where meant to be `text` within this PR

PR #1193, opened 2016-09-19
Ws example fix

fr0stbite:
## What do these changes do?

This minimal pull request fixes depreceated API usage in the `examples/web_ws.py` example.

## Are there changes in behavior for the user?

Users should now be able to run the `examples/web_ws.py` example without any problems.

## Checklist

- [x] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] Add yourself to `CONTRIBUTORS.txt`
- [ ] Add a new entry to `CHANGES.rst`
  * Choose any open position to avoid merge conflicts with other PRs.
  * Add a link to the issue you are fixing (if any) using `#isuue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

PR #1200, opened 2016-09-21
Add missing WSMsgType to web_ws.__all__

popravich:
## What do these changes do?

Add `WSMsgType` to `web_ws.__all__` as it was missing there.

`from aiohttp.web import WSMsgType` will work

## Related issue number

#1193


popravich:
done. changes updated.

PR #1204, opened 2016-09-22
Set documentation page title

bcb:
The documentation page title was not configured in `conf.py`, so it was the default, "aiohttp - aiohttp 1.0.1- documentation". Applying the change it becomes "aiohttp - HTTP client/server for asyncio v1.0.1"

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,8 +3,8 @@
    You can adapt this file completely to your liking, but it should at least
    contain the root `toctree` directive.
 
-aiohttp
-=======
+aiohttp: Asynchronous HTTP Client/Server
+========================================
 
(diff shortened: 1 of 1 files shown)


asvetlov:
If you click on, for example, http://aiohttp.readthedocs.io/en/stable/client.html you'll see that current title's settings make sense.
Maybe what you really need is caption change for `./docs/index.rst`?

bcb:
Yeah I guess ideally you'd only want to change the index page. There must be a way with sphinx - I will look into it.

bcb:
Updated the PR to only the affect the index page. The title is "aiohttp: Asynchronous HTTP Client/Server".

asvetlov:
Thanks!

PR #1215, opened 2016-09-27
Use 1001 Going-Away code in websocket shutdown example

tahajahangir:
The current example (with 999 code) will result in exception: `WebSocketError: Invalid close code: 999`


asvetlov:
Thanks!

PR #1217, opened 2016-09-27
Use yarl in client API

asvetlov:
Switch to yarl.URL in client API


asvetlov:
A coverage miss should be fixed by #1218

PR #1284, opened 2016-10-02
Remove inaccurate reference to HTTP pipelining.

JustinTArthur:
## What do these changes do?
Removes references to HTTP pipelining. While the example shows some kind of pipelining, it is not "HTTP pipelining". 

Nope. Hopefully it will make them less confused if they are looking to do HTTP pipelining.

As far as I know, aiohttp does not support HTTP pipelining. aiohttp requests must be released/exhausted before s connection can be reused (even for the same session).


asvetlov:
Agree

PR #1288, opened 2016-10-05
Yarl in reverse urls

asvetlov:


aiohttp/web_urldispatcher.py
```diff
@@ -296,7 +264,11 @@ def get_info(self):
         return {'path': self._path}
 
     def url(self, *, query=None):
-        return self._append_query(self._path, query)
+        super().url()
```
kxepal, line 278: Warning here will suggest to use `url_for`, but `url_for` doesn't accepts `query` parameter. So deprecation message should be a bit different or it will cause confusion.

aiohttp/web_urldispatcher.py
```diff
@@ -323,130 +295,42 @@ def get_info(self):
...
                 'pattern': self._pattern}
 
-    def url(self, *, parts, query=None):
+    def url_for(self, **parts):
         url = self._formatter.format_map(parts)
-        return self._append_query(url, query)
+        return URL(url)
+
+    def url(self, *, parts, query=None):
+        super().url(**parts)
```
kxepal: Same story here.

aiohttp/web_urldispatcher.py
asvetlov: `url_for` returns `URL`, not `str`.
But functionally it's a replacement for deprecated API.

kxepal: Yes, I see that. I mean that simple replace `url(...)` with `url_for` as message says wouldn't be enough for upgrade.

kxepal: Ok, let's hope for that (:

alexhayes: ~~Where is the documentation for `url_for`?~~ Nevermind, found it

PR #1296, opened 2016-10-09
Build wheels and run tests on OSX as well

zoidbergwill:
Closes https://github.com/KeepSafe/aiohttp/issues/1181

I borrowed some stuff from PyPI and Cryptography's travis stuff. It _might_ be more complicated than we need.

---
This change is [Reviewable](https://reviewable.io/reviews/keepsafe/aiohttp/1296)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,48 @@
 sudo: required
+language: python
 services:
   - docker
 
-language: python
-
-python:
-  - 3.4
-  - 3.5
-#  - 3.5-dev   # for 3.5.2
-#  - nightly
-
-os:
-  - linux
-#  - osx
+matrix:
+    include:
+        - python: 3.4 # these are just to make travis's UI a bit prettier
+          env: TOXENV=py34
+        - python: 3.4
+          env: TOXENV=docs
+          addons:
+           apt:
+            packages:
+            - libenchant-dev
+        - python: 3.5
+          env: TOXENV=py35
+        # - language: generic
+        #   os: osx
+        #   osx_image: xcode8
+        #   env: TOXENV=docs
+        - language: generic
+          os: osx
+          osx_image: xcode6.4
+          env: TOXENV=py34
+        - language: generic
+          os: osx
+          osx_image: xcode6.4
+          env: TOXENV=py35
+        - language: generic
+          os: osx
+          osx_image: xcode7.3
+          env: TOXENV=py34
+        - language: generic
+          os: osx
+          osx_image: xcode7.3
+          env: TOXENV=py35
+        - language: generic
+          os: osx
+          osx_image: xcode8
+          env: TOXENV=py34
+        - language: generic
+          os: osx
+          osx_image: xcode8
+          env: TOXENV=py35
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


asvetlov:
@zoidbergwill very sorry for long silence.
I appreciate your work and will try to merge your PR (probably not in the next release but it is still my goal).

But before publishing MacOSX wheels for aiohttp itself I need to cook dependencies: multidict and yarl.
If you would make wheels for these libraries it would be awesome.
Otherwise I'll try to do it myself sometimes.

zoidyzoidzoid:
Thanks @asvetlov, I'll take a look when I get a chance and fix these conflicts. I'll happily work on it.

asvetlov:
Uggh.
Every my try to enable OSX on travis leads to extremely long build time (typically it is more than a hour).
Perhaps travis is not the best platform for the task :(

fafhrd91:
tests fail for me on OSX 10.12

```
―――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_tcp_connector_resolve_host_use_dns_cache ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――

loop = <_UnixSelectorEventLoop running=False closed=False debug=False>

    @asyncio.coroutine
    def test_tcp_connector_resolve_host_use_dns_cache(loop):
        conn = aiohttp.TCPConnector(loop=loop, use_dns_cache=True)

        res = yield from conn._resolve_host('localhost', 8080)
        assert res
        for rec in res:
            if rec['family'] == socket.AF_INET:
                assert rec['host'] == '127.0.0.1'
                assert rec['hostname'] == 'localhost'
                assert rec['port'] == 8080
            elif rec['family'] == socket.AF_INET6:
>               assert rec['host'] == '::1'
E               assert 'fe80::1%lo0' == '::1'
E                 - fe80::1%lo0
E                 + ::1
```

asvetlov:
@fafhrd91 we *could* solve the particular issue, sure.
But my main point is: adding OSX boxes to travis config increases test run up to 1 - 1.5 hours for *every* build. I don't think if it's appropriate.

AraHaan:
Travis itself is not good for OSX Build might need to find something else that can build for OSX to help speed up build.

AraHaan:
@asvetlov I agree. Unless travis allows you to upload your own OSX VM on github and configure it to use that VM instead of the travis ones to help speed up the travis builds. (which is a good idea and then you could easily manage and update those VM's to what you want as well.

And yes it seems even I was not the only one to suggest it because of the fact that even their version of Linux they have even on trusty has too much of an out of date gcc, g++ and clang to even build C++ 14 code. Which is a shame and get this the people who suggested that of all people was the ones who have C++ code on github and test building on major platforms. But yeah if travis was to do it aiohttp could benefit from it.

AraHaan:
I got the perfect solution for OSX CI that can be fast. But that means using Doozer for it. I already use it on my C++ 14 code as well.

Also this would also require the following file to be added to the repository:

'.doozer.json'

Also I can help on setting it up as well and all that would be needed is to add the webhooks to Doozer to this GitHub repository.

I am not exactly sure however how I can set pypi up on Doozer like it is with Travis. That would need some testing.

Edit:
On the settings on Doozer it seems you can configure environment variables. I am not sure if they can keep the things like the pypi password private. Only way to find out is with some dummy values on one.

img_0112

In aiohttp's case it would probably fit best on the second box for it. Not sure if there is a way to get if it is building an tag or not. And to make environment variables on Doozer it says to do it like so 'ENV=value'. I am however unsure about setting up some sort of code coverage with Doozer.

@asvetlov, @fafhrd91 what you think about getting and using Doozer on the case of OSX?

asvetlov:
We have MacOSX wheels build done thanks to @webknjaz

PR #1297, opened 2016-10-09
Fixed StreamReader._read_nowait

dalazx:
## What do these changes do?

`StreamReader._read_nowait` was returning a single chunk from `self._buffer` even if a much larger chunk was requested and available. This behavior caused performance regressions.

Simple test:
```
import asyncio
import time

import aiohttp


async def measure(chunk_size, timeout):
    start_time = time.time()

    file_size = 0
    chunks_num = 0

        async with session.get('http://localhost:8000/garbage') as response:
            response.content._b_limit = 20 * 1024 * 1024
            while True:
                chunk = await response.content.read(chunk_size)
                if not chunk:
                    break
                file_size += len(chunk)
                chunks_num += 1
                await asyncio.sleep(timeout)

    avg_chunk_size = file_size // chunks_num
    total_time = time.time() - start_time
    print(
        'req chunk size:', chunk_size,
        'avg chunk size:', avg_chunk_size,
        'timeout:', timeout,
        'total time:', total_time)


asyncio.get_event_loop().run_until_complete(measure(10 * 1024 * 1024, 0.1))
asyncio.get_event_loop().run_until_complete(measure(10 * 1024 * 1024, 0.5))
```  

Before:

```
req chunk size: 10485760 avg chunk size: 202427 timeout: 0.1 total time: 53.54 s
req chunk size: 10485760 avg chunk size: 171335 timeout: 0.5 total time: 308.13 s
```

After:
```
req chunk size: 10485760 avg chunk size: 9532509 timeout: 0.1 total time: 1.18 s
req chunk size: 10485760 avg chunk size: 9532509 timeout: 0.5 total time: 5.58 s
```

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,7 +4,8 @@ CHANGES
 1.0.5 (XXXX-XX-XX)
 ------------------
 
--
+- Fix StreamReader._read_nowait to return all available 
+  data up to the requested amount
 
 -
 
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -242,7 +242,7 @@ class StreamReader(AsyncStreamReaderMixin):
                 offset = self._buffer_offset
                 ichar = self._buffer[0].find(b'\n', offset) + 1
                 # Read from current offset to found b'\n' or to the end.
-                data = self._read_nowait(ichar - offset if ichar else -1)
+                data = self._read_nowait_chunk(ichar - offset if ichar else -1)
                 line.append(data)
                 line_size += len(data)
                 if ichar:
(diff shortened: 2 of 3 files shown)


dalazx:
@jettify I guess I found a way to speed up thing further in our particular use case. In the example above there is this line `response.content._b_limit = 20 * 1024 * 1024` that changes the limit in `FlowControlStreamReader` and it makes a difference in the amount of data buffered in the background (64k by default which is too little). The bad thing is that it is not exposed (I have not found the way to publicly set it to whatever I want). Next step is to expose it somehow in a follow-up PR.

asvetlov:
Thanks!

PR #1303, opened 2016-10-11
enhancement to AccessLogger

thehesiod:
populates "extra" dict for clients

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -278,9 +278,9 @@ class AccessLogger:
         'T': 'request_time',
         'Tf': 'request_time_frac',
         'D': 'request_time_micro',
-        'i': 'request_header[%s]',
-        'o': 'response_header[%s]',
-        'e': 'environ[%s]'
+        'i': 'request_header',
+        'o': 'response_header',
+        'e': 'environ'
     }
     LOG_FORMAT = '%a %l %u %t "%r" %s %b "%{Referrer}i" "%{User-Agent}i"'
     FORMAT_RE = re.compile(r'%(\{([A-Za-z0-9\-_]+)\}([ioe])|[atPrsbOD]|Tf?)')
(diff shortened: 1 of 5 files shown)


aiohttp/helpers.py
```diff
@@ -402,8 +408,13 @@ def log(self, message, environ, response, transport, time):
...
         """
         try:
-            self.logger.info(self._log_format % self._format_line(
-                [message, environ, response, transport, time]))
+            fmt_info = self._format_line(
+                [message, environ, response, transport, time])
+
+            extra = {name: value for name, value in fmt_info}
+            self.logger.info(self._log_format %
+                             tuple([value for _, value in fmt_info]),
```
asvetlov: is it `fmt_info.values()`?

thehesiod: not currently as fmt_info is a tuple of tuples to keep items in format order. Would you prefer it be a OrderedDict/namedtuple?

asvetlov: Aaah, got it.
But if you'll return just `extra` `OrderedDict` from `_format_line` the code may look better.

Also  keys for headers/envvars from dicts like `extra={'%{None}i': '-', '%{SPAM}e': 'EGGS', '%{Content-Length}o': 123, '%{User-Agent}i': 'Mock/1.0'}` looks... Well, they looks non-intuitive.
I would expect something like `input_header[User-Agent]` instead of cryptic `%{User-Agent}i`.

Maybe constructing `_log_format` as format accepting params by name in new-styled `.format` form like `"{status} {remote_address} {input_headers[User-Agent]}"` makes sense?

thehesiod: Actually reading the docs for the new formatter it seems like the logging module in python 3.2+ supports the new '{}' style formatting, making this solution a win-win and removes any hesitations I had for this solution, yay!  I'll code something up shortly.

thehesiod:
ok I think this is what we want, extra will contain things like:

``` python
{"response_status": 200, 
"remote_address": "127.0.0.1", 
"request_time_frac": "0.317883", 
"request_header": {
    "User-Agent": "python-requests/2.11.1"
}, 
"first_request_line": "POST /v1/locations HTTP/1.1", 
"response_size": 83}
```

this way users can have a formatter as follows:

```
access_log = aiohttp.log.access_logger
fmtr = logging.Formatter('user_agent: {request_header[User-Agent]}', style='{')
lh2 = logging.StreamHandler()
lh2.setFormatter(fmtr)
access_log.addHandler(lh2)
```

we set the root log handler so if we have the correct format everything "just works" ™️  :)

thehesiod:
I'm not sure if it's doable but it may be worth thinking if it's possible to use a new-style formatter and having the logging system now compose the logging message based on this new `extra` param.

asvetlov:
Awesome!
Would you fix broken tests?

asvetlov:
Thanks!

PR #1304, opened 2016-10-12
Quote special chars in login, password in BasicAuth

kawadia:
## What do these changes do?

Current  implementation of BasicAuth cannot handle a colon in login/username. This change improves it by using urllb.parse.quote/unquote to escape special chars before/after b64 encoding/decoding.

Yes. Eliminates edge case errors when using BasicAuth to encode colon in usernames.

## Related issue number

None.

## Checklist

- [* ] Add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -138,3 +138,4 @@ Yusuke Tsutsumi
 Семён Марьясин
 Pau Freixes
 Alexey Firsov
+Vikas Kawadia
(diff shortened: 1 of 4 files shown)


kawadia:
Checked the RFC again: https://tools.ietf.org/html/rfc1945#section-11.1

Apparently BasicAuth headers should not be quoted. A ":" is not allowed in the username. So the right thing to do is for BasicAuth to raise ValueError if given a username with a ":" in it.

asvetlov:
Agree with closing the PR and adding a check for colon character.

PR #1307, opened 2016-10-12
Raise ValueError if BasicAuth login has a ":"

kawadia:
## What do these changes do?

Improves BasicAuth by raising ValueError if BasicAuth login has a ":".  A colon is not allowed in login/username per RFC 1945#section-11.1.

Yes. Current BasicAuth would silently fail to authenticate if login has a ":". This change would would instead raise a clear error to the user.

## Checklist

- [*] Add yourself to `CONTRIBUTORS.txt`
- [*] Add a new entry to `CHANGES.rst`

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -47,6 +47,10 @@ class BasicAuth(namedtuple('BasicAuth', ['login', 'password', 'encoding'])):
         if password is None:
             raise ValueError('None is not allowed as password value')
 
+        if ':' in login:
+            raise ValueError(
+                'A ":" is not allowed in login (RFC 1945#section-11.1)')
+
         return super().__new__(cls, login, password, encoding)
 
     @classmethod
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -231,3 +231,5 @@ CHANGES
   domains (BACKWARD INCOMPATIBLE) #1125
 
 - Support binary Content-Transfer-Encoding #1169
+
+- Raise ValueError if BasicAuth login has a ":"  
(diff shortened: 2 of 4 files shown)


asvetlov:
Thanks!

aiohttp/helpers.py
```diff
@@ -47,6 +47,10 @@ def __new__(cls, login, password='', encoding='latin1'):
         if password is None:
             raise ValueError('None is not allowed as password value')
 
+        if ':' in login:
+            raise ValueError(
+                'A ":" is not allowed in login (RFC 1945#section-11.1)')
```
kxepal, line 52: Sorry so being late for the party, but this is not very correct RFC reference since: 
1) it about HTTP 1/0 
2) there it disallows non-latin usernames, while we don't.
Here is a better one: https://tools.ietf.org/html/rfc2617#section-2

asvetlov: Thanks.
I'l use `yarl.quote` for both parts separately.

asvetlov: Hmm, after re-reading RFC 7617 I've found:

> The user-id and password MUST NOT contain any control characters (see
>    "CTL" in Appendix B.1 of [RFC5234]).
> 
>   Furthermore, a user-id containing a colon character is invalid, as the first colon in a user-pass string separates user-id and password from one another; text after the first colon is part of the password.
> 
>   User-ids containing colons cannot be encoded in user-pass strings.

Looks like the PR is correct

asvetlov: There is nothing about user-id encoding like _percent-quoting_ etc.

kxepal: That's strange, because without percent-quoting you're not able to use non-ascii names and passwords what is quite awkward today. And since you actually can have them, quoting colon character doesn't breaks the parser while it's quoted.

PR #1315, opened 2016-10-15
File upload fix

demmsnt:
Add missed yield from. As result got MemoryError in big files.
And maybe nessesary specify chunk size. I think in this case we read all data from file.


asvetlov:
Please sync with master

asvetlov:
Fixed by 8a98f552

PR #1317, opened 2016-10-17
Add load and save support for cookie_jar

ahui132:
## What do these changes do?
1. Add load and save support for cookie_jar
2. Add test case for the support above.

No.

## Checklist

- [x ] Unit tests for the changes exist


asvetlov:
Please fix PEP8 blames first.

asvetlov:
The checking bot blames that `cookieJar` is misspelled word.
Changing it to `CookieJar` will fix the problem.
Please do it.

ahuigo:
So sorry for that I didn't response these days. I've re committed my fixes.

asvetlov:
Please add a test for missing case: https://codecov.io/gh/KeepSafe/aiohttp/compare/2013dc4b33abb4c932d5f56f9abb878dd882d99a...a11386475d721f9a2efd8784f464cec376c8c396#61696F687474702F636F6F6B69656A61722E7079-49

I would have 100% coverage someday.
Adding not covered lines definitely goes in wrong direction.

aiohttp/cookiejar.py
```diff
@@ -37,6 +39,17 @@ def __init__(self, *, unsafe=False, loop=None):
         self._next_expiration = ceil(self._loop.time())
         self._expirations = {}
 
+    def load(self, file_path):
+        f = open(file_path, 'rb')
```
asvetlov: Please support both `str` and `pathlib.Path` types.
Just add `file_path = pathlib.Path(file_path)` and use [Path.open](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open) API.

aiohttp/cookiejar.py
```diff
@@ -37,6 +39,17 @@ def __init__(self, *, unsafe=False, loop=None):
...
+        self._cookies = pickle.load(f)
```
asvetlov: The line should be wrapped by `with` statement for sake of managed resource usage.

aiohttp/cookiejar.py
```diff
@@ -37,6 +39,17 @@ def __init__(self, *, unsafe=False, loop=None):
...
+
+    def save(self, file_path):
+        file_dir = os.path.dirname(file_path)
```
asvetlov: The same as above: please support both `str` and `pathlib.Path`.

aiohttp/cookiejar.py
```diff
@@ -37,6 +39,17 @@ def __init__(self, *, unsafe=False, loop=None):
...
+        if file_dir and not os.path.exists(file_dir):
+            os.makedirs(file_dir)
+        f = open(file_path, 'wb')
+        pickle.dump(self._cookies, f)
```
asvetlov: Use `with` context manager.

aiohttp/cookiejar.py
ahuigo: Since the pickle do not support `pathlib.Path`, I choose to use `file_path = str(file_path) if isinstance(file_path, pathlib.Path) else file_path` or `file_path = str(file_path)` instead.

asvetlov:
Sorry, #1326 was finished first.

PR #1318, opened 2016-10-17
Ensure TestClient HTTP methods return a context manager

alexhayes:
## What do these changes do?

A user should be able to run a test as per the [client documentation](http://aiohttp.readthedocs.io/en/stable/client.html);

```python
    async with session.get('https://api.github.com/events') as resp:
        print(resp.status)
        print(await resp.text())
```

However, because the test client doesn't return a context manager the above does not work (unless of course I'm doing something wrong).

This PR ensures the HTTP methods on TestClient return a context manager so that it mimics the same interface as a non-test client

## Are there changes in behaviour for the user?

Yes, the TestClient will behave the same as the non-test client that their code will be using :)

Let's say or instance your code looked as follows;

```python
# client request code
async request_hello(session):
    async with session.get('...') as resp:
        ...
```

Note the use of the context manager: `async with session.get('...') as resp:`

And your test was;

```python
# pytest
from aiohttp import web

async def hello(request):
    return web.Response(text='Hello, world')

async def test_hello(test_client, loop):
    app.router.add_get('/', hello)
    resp = await request_hello(client)
    assert resp.status == 200
    ...
```

On master, this would fail, because TestClient does not return a context manager.

## Related issue number

None.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,10 @@ CHANGES
 
 - Enhancement to AccessLogger (pass *extra* dict) #1303
 
+- Ensure TestClient HTTP methods return a context manager #1318
+
+-
+
 -
 
 -
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -5,6 +5,7 @@ A. Jesse Jiryu Davis
 Alejandro Gómez
 Aleksandr Danshyn
 Aleksey Kutepov
+Alex Hayes
 Alexander Bayandin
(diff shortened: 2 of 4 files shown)


asvetlov:
Please resolve conflict in `CHANGES.rst`.
I'll merge the PR after this.

alexhayes:
Rebased master onto my branch and force pushed.

asvetlov:
Cool! Thanks!

PR #1319, opened 2016-10-18
Show more verbose message on import errors

redixin:
Show more verbose message on import errors when using `python -m aiohttp.web`

User will see more details in case of problems with their code.

For example message:

    aiohttp.web: error: unable to import defiler.server: No module named 'aiomysql'

is more useful then:

    aiohttp.web: error: module 'defiler.server' not found

such messages can be even misleading


asvetlov:
Please fix failed tests

asvetlov:
Thanks!

PR #1335, opened 2016-10-26
Fix invalid invocations of errors.LineTooLong.

filmor:
## What do these changes do?

Fix unhelpful messages on `LineTooLong` exceptions due to invalid or incomplete calls to their constructor.

Only in the message that they see in this case.

## Checklist
  - The format is &lt;Name&gt; &lt;Surname&gt;.
  - Please keep alphabetical order, the file is sorted by names. 
  - Choose any open position to avoid merge conflicts with other PRs.
  - Add a link to the issue you are fixing (if any) using `#issue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

### DIFF
--- a/aiohttp/protocol.py
+++ b/aiohttp/protocol.py
@@ -102,7 +102,9 @@ class HttpParser:
                     header_length += len(line)
                     if header_length > self.max_field_size:
                         raise errors.LineTooLong(
-                            'limit request headers fields size')
+                            'request header field {}'.format(
+                                bname.decode("utf8", "xmlcharrefreplace")),
+                            self.max_field_size)
                     bvalue.append(line)
 
                     # next line
@@ -113,7 +115,9 @@ class HttpParser:
             else:
                 if header_length > self.max_field_size:
                     raise errors.LineTooLong(
-                        'limit request headers fields size')
+                        'request header field {}'.format(
+                            bname.decode("utf8", "xmlcharrefreplace")),
+                        self.max_field_size)
 
             bvalue = bvalue.strip()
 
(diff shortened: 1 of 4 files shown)


aiohttp/protocol.py
```diff
@@ -102,7 +102,8 @@ def parse_headers(self, lines):
                     header_length += len(line)
                     if header_length > self.max_field_size:
                         raise errors.LineTooLong(
-                            'limit request headers fields size')
+                            'request header field %s' % bname.decode("utf8"),
```
popravich: `bname.decode('utf-8')` can fail with own exception

aiohttp/protocol.py
```diff
@@ -113,7 +114,8 @@ def parse_headers(self, lines):
             else:
                 if header_length > self.max_field_size:
                     raise errors.LineTooLong(
-                        'limit request headers fields size')
+                        'request header field %s' % bname.decode("utf8"),
```
popravich: same as above

aiohttp/protocol.py
```diff
@@ -173,7 +175,7 @@ def __call__(self, out, buf):
             raw_data = yield from buf.readuntil(
                 b'\r\n\r\n', self.max_headers)
         except errors.LineLimitExceededParserError as exc:
-            raise errors.LineTooLong(exc.limit) from None
+            raise errors.LineTooLong('request header', exc.limit) from None
```
popravich, line 180: this looks inconvenient: `LineTooLong` above is raised with single argument (formated string) and with two args here

filmor: It's not, and the call here is wrong. Check out the constructor of `LineTooLong` (didn't change that one), it expects a "reason" and a limit that was exceeded. The way this was passed before resulted in a broken error message like `"400 got more than Unknown bytes when reading 32768"` when it should have read `"400 got more than 32768 when reading request header"`.

filmor:
I'll fix the failing tests and add some more.

aiohttp/protocol.py
popravich: true, missed that one.

aiohttp/protocol.py
filmor: Can it though? This is caught before by "Invalid Header". Changed it anyhow :)

asvetlov:
Thanks!

PR #1339, opened 2016-10-27
Fixed JSON response character set detection.

kimvais:
## What do these changes do?

This disables running chardet on responses with `CONTENT_TYPE: application/json` as this is a HUGE performance hit.

RFC 7519 states that JSON *MUST* be encoded in UTF and that
the default encoding, in absence of charset in CONTENT_TYPE header is UTF-8.

This vastly improves performance when doing many small requests in environments
where the server does not specify the character by omitting chardet.

This does not change the behaviour unless a non-conforming server sends a HTTP response using an invalid encoding without specifying the encoding correctly in the header.
## Checklist

- [X] Add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -76,6 +76,7 @@ Julien Duponchelle
 Junjie Tao
 Justas Trimailovas
 Kay Zheng
+Kimmo Parviainen-Jalanko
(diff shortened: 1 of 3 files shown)


asvetlov:
Thanks

PR #1351, opened 2016-10-30
Get rid of aiodns gethostbyname()

amelius0712:
## What do these changes do?

By default aiodns function gethostbyname() (which is actually just a wrapper around c-ares function) resolve host as IPv6 address if socket family set to AF_UNSPEC(default value for TCPConnector). On the other hand ThreadedResolver use asyncio loop function getaddrinfo() which, in accordance with POSIX return both(either IPv4 and IPv6) if socket family set to AF_UNSPEC. So we have unconsistent behavior of ThreadedResolver and AsyncResolver. But c-ares lib doesn't have getaddrinfo() function, so we can't use it in aiohttp resolver. So if I use IPv4 the following simple code does not work for me(if aiodns installed of cause):

``` python
        async with session.get('http://google.com') as resp:
            body = await resp.read()
            print('Done(google.com). status: {}, body len: {}'.format(resp.status, len(body)))
```

To solve this problem in method resolve() of AsyncResolver I query both types if family set to AF_UNSPEC.

No
## Related issue number
# 559
## Checklist
- [ X] I think the code is well written
- [ X] Unit tests for the changes exist
- [ X] Documentation reflects the changes
- [ X] Add yourself to `CONTRIBUTORS.txt`
  - The format is &lt;Name&gt; &lt;Surname&gt;.
  - Please keep alphabetical order, the file is sorted by names. 
- [ X] Add a new entry to `CHANGES.rst`
  - Choose any open position to avoid merge conflicts with other PRs.
  - Add a link to the issue you are fixing (if any) using `#issue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -90,7 +90,7 @@ CHANGES
 
 - Boost performance by adding a custom time service #1350
 
--
+- Get rid of usage aiodns gethostbyname() #559
 
 -
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -105,6 +105,7 @@ Olaf Conradi
 Pankaj Pandey
 Pau Freixes
 Paul Colomiets
+Pavel Sapezhko
 Philipp A.
 Raúl Cumplido
 "Required Field" <requiredfield256@gmail.com>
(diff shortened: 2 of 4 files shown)


psapezhka:
Ok. But what if I want to use aiodns? I think the behavior should be the same as a thread pool. We can set thread pool as default and fix async resolver too. It does not look like they wanted to add getaddrinfo() into c-ares. https://github.com/c-ares/c-ares/issues/70

asvetlov:
Looks like they are looking for volunteer.
If you want to work via aiodns -- connect it explicitly and use on your own.

PR #1366, opened 2016-11-04
fix usage of proxy with https (#845)

### LINKED ISSUE #845: How to use HTTPS proxy with aiohttp?
I noticed the code in ProxyConnector：
`assert proxy.startswith('http://'), ("Only http proxy supported", proxy)`
Is there any way to use HTTPS proxy with aiohttp?
If not, is this feature under consideration?
Any help will be appreciated.

mgwilliams:
Fixes #845, a regression introduced in [e81a47](https://github.com/KeepSafe/aiohttp/commit/e81a4781d80b614da572c2c18635831ad024126b#diff-2fa8ab44317df4fbb7864536a6794d7aL491). I feel this is not the best way to fix this, but it at least points out where the issue is. Perhaps someone more familiar with the code base could rewrite the patch. I'm also not sure how to write a regression test for this, but would be happy to do so if pointed in the right direction.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -104,6 +104,8 @@ class ClientRequest:
         self.update_transfer_encoding()
         self.update_expect_continue(expect100)
 
+        self.path = None
+
     @property
     def host(self):
         return self.url.host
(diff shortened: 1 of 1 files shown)


asvetlov:
Very thank you.
You've found a real source problem.
`.path` hacks should be replaced with better way but I've got your idea.
Will create new PR soon.
Unfortunately we have not yet proper functional tests for proxies so I'll ask you to manually check upcoming PR.

asvetlov:
Fixed

mgwilliams:
@asvetlov Thanks! I'll test ASAP, but it may be a week or so.

PR #1379, opened 2016-11-07
allow starting variables with an underscore

arthurdarcet:
This allows the named resources identifiers to start with an underscore

Mongo ids are useful as a named resource in a route, but they are usually named `_id` and are excluded by the current regex

If this isn't an acceptable change, then the documentation should imo be updated to specify what is an acceptable route identifier (`[a-zA-Z][_a-zA-Z0-9]*` currently)

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -694,12 +694,12 @@ class RoutesView(Sized, Iterable, Container):
 
 class UrlDispatcher(AbstractRouter, collections.abc.Mapping):
 
-    DYN = re.compile(r'\{(?P<var>[a-zA-Z][_a-zA-Z0-9]*)\}')
+    DYN = re.compile(r'\{(?P<var>[_a-zA-Z][_a-zA-Z0-9]*)\}')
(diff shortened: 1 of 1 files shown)


asvetlov:
The change is good to me.
The idea is: route variable names should accept everything that is good for python variables themselves.
Thus I'm ok even with adding non-ascii names on first demand :)

asvetlov:
Thanks

arthurdarcet:
Thanks, and good to know for the non-ascii names. I'll have a go a that too if I ever need it

For reference, if anyone stumble here, python identifiers are:

```
identifier   ::=  xid_start xid_continue*
id_start     ::=  <all characters in general categories Lu, Ll, Lt, Lm, Lo, Nl, the underscore, and characters with the Other_ID_Start property>
id_continue  ::=  <all characters in id_start, plus characters in the categories Mn, Mc, Nd, Pc and others with the Other_ID_Continue property>
xid_start    ::=  <all characters in id_start whose NFKC normalization is in "id_start xid_continue*">
xid_continue ::=  <all characters in id_continue whose NFKC normalization is in "id_continue*">
```

PR #1362, opened 2016-11-04
New server

asvetlov:
Implement low-level web server for humans

PR #1382, opened 2016-11-08
Static range requests

andrewleech:
## What do these changes do?

Add support for html range header in static file handling.

Generally follows  https://tools.ietf.org/html/rfc7233 however does not support multiple ranges in one request.

Usage is by adding the range header similar to {'Range': 'bytes=0-99'} to a request to receive a partial chunk of the file instead of the entire thing.

If the range heading is not explicitly added to a request there should be no change in usage.

## Checklist


aiohttp/file_sender.py
```diff
@@ -152,16 +153,48 @@ def send(self, request, filepath):
...
-        resp = self._response_factory()
+        file_size = st.st_size
+
+        status = 200
+        start = 0
+        end = file_size - 1
+
+        # Handle 206 range response if requested
+        if 'range' in request.headers:
+            from .web_exceptions import HTTPPartialContent, HTTPRequestRangeNotSatisfiable
```
fafhrd91: move it to module level import.

aiohttp/file_sender.py
```diff
@@ -152,16 +153,48 @@ def send(self, request, filepath):
...
 
-        resp = self._response_factory()
+        file_size = st.st_size
+
+        status = 200
+        start = 0
+        end = file_size - 1
+
+        # Handle 206 range response if requested
+        if 'range' in request.headers:
```
fafhrd91: i think it make sense to move range related functionality to helper module or even to separate module. and create `Range` class, it could be useful for external libraries and applications

aiohttp/file_sender.py
andrewleech: I normally would by standard, but was following what had been done a few lines earlier in the same function. I'll fix both.

aiohttp/file_sender.py
andrewleech: I had thought that but wasn't sure where would make sense to put it. 
It could be in a parse_range_header(request) function in helpers, it's not necessarily making things clearer but would certainly would aid in reuse.

andrewleech:
Global range support would be nice sure, but would probably be a significantly larger change. 

That being said most use cases I know of for range is to get part of static resources for pagination or seeking in video files etc. It's less useful on dynamic content where it's harder to know a range to request of something if the content could change in between requests.

asvetlov:
@andrewleech your PR is very long waited addition.
Thank you for it.
I don't think it's ready for merging right now but you are following in right direction.

Couple brief notes:
1. `parse_range_header` should be `web.Request` property like `req.content_lenght` or `req.content_type`. Or even a property of `web.BaseRequest` after #1362 merging. The property result caching is not required, sure.
2. `range_request` should be extracted to module level and presented as Public API. The name is not perfect, maybe `RequestHeader` or `RequestHeaderInfo`

fafhrd91:
1. i'd rename `range_request` to just `Range`
2. move `parse_range_header` to separate module
3. do not import inside function

asvetlov:
Please run `make isort` for fixing travis build.

asvetlov:
The code looks perfect.
The only thing is left -- documentation.
aiohttp doesn't use autodoc markup (at least I'm trying to remove all places where we still do it).
So, please, drop markup from `.http_range` property and add explicit documentation for the property to `docs/web_reference.rst`

asvetlov:
Ok, I'll update docs.
Thank you very much!

asvetlov:
Done by c693473
I've changed `http_range` value from `tuple` to `slice` also.

andrewleech:
Oh, thanks! 
I was going to fix the documentation as suggested but have just been slammed the last week.

That's great about the slice object; I didn't know they existed and my initial search to find one didn't come up with anything. I'll keep that in mind for future!

PR #1385, opened 2016-11-10
ClientSession.request() shouldn't duplicate the param defaults from _request()

mikenerone:
Most of the request methods on `ClientSession` (`get()`, `post()`, etc), have pretty generic "**kwargs" signatures and wrap `self._request()`. `_request()` itself contains sensible defaults, and provides one convenient method to override in a subclass if one would like to change those defaults.

`ClientSession.request()` (no underscore), is an exception, though, and duplicates all of those defaults. As a result, if one wants to make their requests via `request()` (I have my reasons) overriding `_request()` alone is no longer sufficient - `request()` must itself be overridden, as well.

While this is only a minor inconvenience, `request()` should, in my opinion, behave consistently with the other `_request()` wrappers, and have a generic "**kwargs" signature.

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -92,45 +92,9 @@ class ClientSession:
                 context['source_traceback'] = self._source_traceback
             self._loop.call_exception_handler(context)
 
-    def request(self, method, url, *,
-                params=None,
-                data=None,
-                headers=None,
-                skip_auto_headers=None,
-                auth=None,
-                allow_redirects=True,
-                max_redirects=10,
-                encoding='utf-8',
-                version=None,
-                compress=None,
-                chunked=None,
-                expect100=False,
-                read_until_eof=True,
-                proxy=None,
-                proxy_auth=None,
-                timeout=5*60):
+    def request(self, method, url, **kwargs):
         """Perform HTTP request."""
-
-        return _RequestContextManager(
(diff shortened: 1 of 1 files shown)


asvetlov:
OMG. `aiohttp.ClientSession` is not designed for inheritance.
Perhaps forbidding inheritance by aiohttp code is good idea.
What do you need is embedding client session instance into your class like https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/test_utils.py#L173
I could even consider making `_RequestContextManager` a part of public API :)

mikenerone:
I won't be offended if you choose not to accept the PR, of course, but I have no choice but to override certain behaviors (just a couple) that are baked into `ClientSession` that are in conflict with unavoidable requirements, not under my control, of the task I have at hand. Please don't add any inheritance shackles. I'm a consenting adult.

asvetlov:
Ok. You a consenting adult.
I've merged your PR but please don't expect that the API will not be changed without any notification and will respect deprecation period rule.
You've just using "officially unsupported" _gray way_.
It doesn't mean that I'll try to break the API intentionally just for making your life harder but if internal evolution logic will require it -- I'll make a break.

P.S.
Your PR will be available in next 1.1.4 bugfix release.

PR #1391, opened 2016-11-11
Make TestServer.make_url compatible with yarl.URL

imbolc:

Nope

## Related issue number

https://github.com/KeepSafe/aiohttp/issues/1389

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -74,6 +74,8 @@ class TestServer:
                                                           ssl=self._ssl)
 
     def make_url(self, path):
+        if isinstance(path, URL):
+            path = str(path)
         assert path.startswith('/')
         return URL(str(self._root) + path)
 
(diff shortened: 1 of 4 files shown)


aiohttp/test_utils.py
```diff
@@ -78,6 +78,8 @@ def _make_factory(self):
         pass
 
     def make_url(self, path):
+        if isinstance(path, URL):
```
asvetlov: I believe `return self._root.join(URL(path))` is easier solution.

imbolc: I'm not sure what `path.startswith('/')` is checking, but in your solution if we pass `http://foo.com/` as `path` it returns `http://foo.com/` for any `self._root`. Is it ok?

asvetlov: I thinks yes, it's just testing tool.
`.startswith()` does test for relative URL.
The same could be done by `assert not URL(path).is_absolute()`.
Not sure if we need this check but it is harmless.

imbolc:
Done

asvetlov:
Thanks!

asvetlov:
Will be published as part of aiohttp 1.1.4 soon.

PR #1395, opened 2016-11-12
Added third party aiocache and example

argaen:
## Checklist


asvetlov:
I'll love to add aiocache into third-parties list but, sorry, I don't want to include your example.
Let's keep it in aiocache itself.

P.S.
Using global dict for setting smells as anti-pattern for me.
I mean

```
aiocache.settings.set_defaults(
    cache="aiocache.RedisCache",
    namespace="test"
)
```

argaen:
Removed the example :).

> Using global dict for setting smells as anti-pattern for me.

Its not a global dict, they are global variables of the module. I know global variables are an anti pattern but not always. By default aiocache uses Memory backend with default serializer (which does nothing) and default policy (which does nothing).

Now imagine that you want to use always Redis as your backend with a specific serializer at a given endpoint and port. With these global settings feature you just have to change 4 lines, using them explicitly you have to change them everywhere. Example:

```
cached(ttl=10) -> cached(cache=RedisCache, endpoint=foo, port=bar, serializer=blah)
```

Now imagine you have lots of instances all along your code...

If you have a better proposal about how do address this which is both clean and maintainable happy to hear! :)

asvetlov:
I don't know how to do it in framework agnostic way but in case of aiohttp you could push global setting into `Application` object.

argaen:
That would be a partial solution, but in the case of decorators you would still need to pass all the attributes.. =/. 

I prefer to leave it open to the developer :).

Issue #1133, opened 2016-08-27
Add max_size parameter to web.Request reading methods


asvetlov:
It my prevent malicious clients from sending infinite BODY in post requests which may lead to running out of memory.

Guys, what do you think about?
Does it make sense?

asvetlov:
Should raise 413 on size overloading

f0t0n:
I think you'll never get to request body size check when you run behind nginx in production deployment since nginx has own setting [`client_max_body_size`](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) (default 1MB) and returns 413 if the size of request's body exceeds that value.

Do we expect users to run aiohttp applications without nginx in production?

f0t0n:
So my point is RAM management is more up to DevOps engineer that is responsible for the server infrastructure, so that he should configure correct limits for nginx, optimal database connection pool size and so on.

f0t0n:
Say attacker could try to overload RAM with huge request body. And we check the max size. Ok. Then he starts to DoS the application, should we now check his request rate? I think both cases are out of application's scope and should be delegated one level upper, namely to servers setup by DevOps team.

asvetlov:
Checking for max length by nxing is satisfactory solution but writing complex configuration with different values for different locations may be painful.
So why not support it in aiohttp too?
In other case most likely DevOps will relax the size for all application even if only couple endpoints should accept huge file uploads.

argaen:
I think it's useful to have this feature supported by aiohttp as long as:
1. We set a reasonable default size
2. We provide a way of disabling it (like setting `max_size=0` for example?)

@f0t0n with the example you say about request rate, there are libraries like djangorestframework that provide this functionality. Is the users choice if they want to use this feature or use something else in the server layer.

PR #1414, opened 2016-11-20
[web_request] add handling of client_max_size in request body

pawelmhm:
## What do these changes do?

Adding size limits to request body so that malicious client can't cause trouble by trying to upload request with extremely big body. 

* default limit set to value same as in nginx 1mb;
* if client tries to upload body that exceeds this limit it will get 413 HTTP Error;
* limit can be updated by passing client_max_size argument to Application object

Yes, too big uploads will raise HTTP 413 error.

## Related issue number

#1133 

## Checklist


pawelmhm:
I stumbled on two issues here:

1) there are `aiohttp.web_exceptions` and `aiohttp.errors`. Difference between the two is slightly unclear. They seem to contain similar code (e.g. 400, or 405 exception definittions) It would be nice to add some docstrings explaining their purposes. It seems like aiohttp.errors are mostly internal errors raised by library, and web_exceptions are intended to be raised from client code, is that correct? I added 413 to errors, is that ok? I cant add it to web_exceptions because of circular imports.

2) web_server.RequestHandler.handle_request will fail with UnboundLocalVariable error in finally block (line 69)  on every exception that is not HTTPException. It looks like a bug, am I right? `self._handler()` call in `try` part can raise multiple things that are not HTTPExceptions? It seems this was introduced recently, I did some change in code last weekend and behavior was different. Anyway I 'fixed' that, but without test, because I'm not exactly sure about this. Let me know if we should consider this bug or feature.

I guess code-cow failing because of missing test for 2.

aiohttp/web_reqrep.py
```diff
@@ -70,6 +74,11 @@ def __init__(self, message, payload, transport, reader, writer,
         self._state = {}
         self._cache = {}
         self._task = task
+        if client_max_size:
+            max_size = client_max_size
+        else:
+            max_size = self._CLIENT_MAX_SIZE
```
asvetlov: `max_size` could be `None` if user operates with low-level API (without `web.Application`, routers, signals, middlewares and other related stuffs).

aiohttp/web_reqrep.py
```diff
@@ -363,6 +372,8 @@ def read(self):
             while True:
                 chunk = yield from self._payload.readany()
                 body.extend(chunk)
+                if len(body) >= self._client_max_size:
```
asvetlov: `_client_max_size` could be `None`

aiohttp/web_reqrep.py
```diff
@@ -363,6 +372,8 @@ def read(self):
...
+                    raise HttpRequestEntityTooLarge()
```
asvetlov: For this particular case `from .web_exceptions import HTTPRequestEntityTooLarge` is allowed evil.
Yes, I know the import from function has ugly adore.
But I'm ok with this for the case.

aiohttp/web_server.py
```diff
@@ -53,17 +54,21 @@ def handle_request(self, message, payload):
 
         request = self._request_factory(message, payload, self)
         self._request = request
+        resp = None
```
asvetlov: Yes, it's a bug. Thank you for finding.

aiohttp/web_server.py
```diff
@@ -53,17 +54,21 @@ def handle_request(self, message, payload):
...
 
         try:
             try:
                 resp = yield from self._handler(request)
             except HTTPException as exc:
                 resp = exc
+            except HttpProcessingError as exc:
```
asvetlov: All exceptions not derived from `HttpProcessingError` should raise `500 Internal Server Errror` (now it's done by `server.py` code).

aiohttp/web_server.py
```diff
@@ -53,17 +54,21 @@ def handle_request(self, message, payload):
...
+                raise exc
 
             resp_msg = yield from resp.prepare(request)
             yield from resp.write_eof()
         finally:
-            resp._task = None
+            if hasattr(resp, "_task"):
```
asvetlov: If you'll use `web_exceptions` this code is not needed.

fafhrd91:
@pawelmhm are you still interested in working on this PR?

AraHaan:
Uh @pawelmhm you might want to backup some of your changes and go with the latest commit on the master branch and work from there and remake this PR after that so that way it can fix the merge conflicts easier. Also it would be a good idea to also apply the requested changes that @asvetlov said.

pawelmhm:
hey @AraHaan @fafhrd91 sorry for delay I was bit busy, yeah I'll do rebase against master and apply suggestions from @asvetlov I'll try to do this over the weekend

fafhrd91:
Cool, thanks

aiohttp/web_reqrep.py
```diff
@@ -372,6 +375,11 @@ def read(self):
             while True:
                 chunk = yield from self._payload.readany()
                 body.extend(chunk)
+                if self._client_max_size \
+                        and len(body) >= self._client_max_size:
+                    # local import to avoid circular imports
+                    from aiohttp import web_exceptions
```
fafhrd91, line 381: Try to import `aiohttp.web`. All web exceptions are available i that namedspace

fafhrd91:
thanks

aiohttp/web_reqrep.py
pawelmhm: ah I see you merged that, I can check if import of aiohttp.web works, should I open new PR with that + doc updates, adding myself to contributors?

fafhrd91: I fixed import and added you to contributors already. thanks

Issue #1108, opened 2016-08-23
Deprecate aiohttp.wsgi


asvetlov:
Currently the module is not 100% compatible with WSGI standard, it works bad with big payload but introduces `'async.reader'` and `'async.writer'` fields in WSGI environment dictionary.

It's not documented also.

Should we support it at all?

I suggest adding explicit deprecation and removing `wsgi` mentions from documentation.
After 1.5 years of deprecation period we may drop the module at all.

If somebody want to support the module -- please do it outside of aiohttp project.

fafhrd91:
gunicorn.gaiohttp depends on wsgi

asvetlov:
Uuups.
Sorry, I've forgot this fact.

fafhrd91:
we can replace gaiohttp with newer implementation that support both aiohttp.wsgi and aiohttp.web 
and later deprecate aiohttp.wsgi support

asvetlov:
Agree with the plan

xpinguin:
Hi! I'd like to put my ten kopeks in, regarding aiohttp.wsgi. Turned out useless for me, as my goal was to seamlessly attach inherently synchronous wsgi app (WsgiDav in my case) to an asyncio application, where aiohttp is used only for HTTP-based protocols.
The major drawback is that aiohttp.wsgi supports only wsgi/async - custom wsgi extension, out of pep-3333 scope. This renders conforming (i.e. non-async) wsgi apps out of the game.

The solution was to develop the class which automatically performs 2way coroutine<->function conversion, thus allowing, for instance, to wrap async StreamReader into non-async one and pass it to an app. Well, no magic in here, drawbacks are obvious:
- function->corofunc <=> loop.run_in_executor(function(...)), eg. default ThreadPool
- corofunc->function <=> run_coroutine_threadsafe(corofunc(...))

... so at least we could easily have a stall in our wsgi app, although never in the main event loop, so we could adjust the pool size, swap pools, etc. - all hypothetical, had no occasion to try that out, as our real loads are laughable :)

For those, interested in actual code, here is the gist of the class (NB. it is from non-public application in alpha-stage so the clumsy code, I've also removed some aux. stuff, eg. our approach to routing, in order to be clean, just don't expect the code below to work as-is; moreover, WSGIHttpServerProtocol has poorly fitted my preference for abstractions used, so class is more-or-less application-like (conceptually, not literaly the web.Application)):
[...]

PR #1418, opened 2016-11-22
add wsgi support to GunicornWebWorker

fafhrd91:
this is prerequisite for #1108, after aiohttp public release we can replace AiohttpWorker from gunicorn and then we can start wsgi deprecation process

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -1,6 +1,7 @@
 """Async gunicorn worker for aiohttp.web"""
 
 import asyncio
+import logging
 import os
 import re
 import signal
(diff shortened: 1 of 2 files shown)


fafhrd91:
@asvetlov

AraHaan:
oh nice, would be nice if the checks on those 2 webhooks passed. :octocat:

asvetlov:
Looks good.
Would you create a functional wsgi test?
Frankly speaking I don't trust mock-only testing.

fafhrd91:
i don't really care about wsgi. i just want to start deprecation process.

asvetlov:
Ok

PR #1431, opened 2016-11-27
Fixed stream reading method of BodyPartReader

### LINKED ISSUE #1428: Method "read_chunk" of "BodyPartReader" returns zero bytes before eof
## Long story short

I've implemented a multipart file upload handler inspired on code from the [docs](http://aiohttp.readthedocs.io/en/stable/web.html#file-uploads). My code is truncating part's data. I believe the problem is in the method `_read_chunk_from_stream`, which is used by `read_chunk` of `BodyPartReader`. That method is returning a zero-length `bytearray` before the part's EOF. This is the pseudo-code.

    part = await reader.next()
    arr = bytearray()

    while True:
        chunk = await part.read_chunk()  # 8192 bytes by default.

        if not chunk:
            break

        arr.extend(chunk)

## Expected behaviour

The loop ends when all the part's data has been read.

## Actual behaviour

The loop ends before the part's data is exhausted, i.e., chunk becomes a zero-length `bytearray` prematurely.

## Steps to reproduce

The code is part of a large web application so it's hard for me to give reproducible steps. But replacing the break condition to `if not part._at_eof` made the problem go away.

    part = await reader.next()
    arr = bytearray()

    while True:
        chunk = await part.read_chunk()  # 8192 bytes by default.

        if not part._at_eof:  # This fixed the problem.
            break

        arr.extend(chunk)

## Your environment

Aiohttp 1.1.5
Python 3.5.1 from PSF
macOS Sierra 10.12.1

viotti:
The method `BodyPartHeader._read_chunk_from_stream` from `aiohttp.multipart` module was returning an zero length Byte object before reaching EOF while reading a part in a multipart request.

Added missing virtualenv activation instructions in `CONTRIBUTING.rst`.

Fixed a typo in the docs.

Fixes #1428.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -39,6 +39,7 @@ If you like to use *virtualenv* please run:
 
    $ cd aiohttp
    $ virtualenv --python=`which python3` venv
+   $ . venv/bin/activate
 
 For standard python *venv*:
 
@@ -46,6 +47,7 @@ For standard python *venv*:
 
    $ cd aiohttp
    $ python3 -m venv venv
+   $ . venv/bin/activate
 
 For *virtualenvwrapper* (my choice):
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -108,6 +108,7 @@ Pankaj Pandey
 Pau Freixes
 Paul Colomiets
 Philipp A.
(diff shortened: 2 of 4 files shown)


kxepal:
+1

asvetlov:
Thanks!

PR #1442, opened 2016-12-01
Update tutorial.rst

gyermolenko:
The goal of this update is to show user gradual development progress, with working project at each iteration of it.
Here are some items I've updated:
 - bump aiohttp version
 - move views section way up. We need it for first "hello aiohttp". And we don't need configs or db at this point.
- remove obsolete self argument from index view and one from the first route.


asvetlov:
Cool! Thanks!

PR #1453, opened 2016-12-06
preserve cookie escaping for old servers

thehesiod:
this preserves the coded value so quotes won't get inserted and break cookie processing from certain servers like: http://hydro1.sci.gsfc.nasa.gov/data/NLDAS/NLDAS_FORA0125_H.002/2010/360/NLDAS_FORA0125_H.A20101226.0000.002.grb  which uses: http://disc.sci.gsfc.nasa.gov/alerts/access-to-ges-disc-data-will-require-all-users-to-be-registered-with-the-earthdata-login-system

This updates aiohttp to mimic chrome's behavior in terms of preserving the cooking encoding.


thehesiod:
w/o this fix you hit max_redirects trying to re-auth since it doesn't recognize the auth cookie it set

thehesiod:
btw, for changes like these do we bump up versioning?  What's the versioning strategy?

asvetlov:
1. What the change exactly does?
2. Where are tests for proving new behavior?
3. Couple existing tests a broken, I suspect they are broken in unexpected way, e.g. domain info was dropped.

thehesiod:
There are two values for a cookie, a coded_value and the non-coded value.  Python always encodes the coded_value according to cookie version 1 rules (double quoting the value if the value contains certain characters). The issue is when you've received a version=0 value from a server w/o dbl quoting, it would then dbl quote it and send it back causing the server to not recognize its own cookie. Actually the test is failing because we're now preserving the domain whereas it was not before.  I'll check to see what is correct.  Btw if anything we now preserve more information than less.

Basically we should send cookies back in the format they were received.

thehesiod:
ok updated the unittests to expect the domain post filter, and added one to verify that double quoted values are preserved. I believe the tests were previously incorrect as if you do an "output" pre filter it now matches post-filter.

thehesiod:
it's kinda confusing given there's http.cookies.SimpleCookie, not to be mistaken by http.cookiejar.Cookie which is expected by http.cookiejar.CookieJar.set_cookie.  And then aiohttp has its own CookieJar.

tests/test_cookiejar.py
```diff
@@ -240,20 +240,33 @@ def test_preserving_ip_domain_cookies(loop):
...
 
 
 def test_ignore_domain_ending_with_dot(loop):
     jar = CookieJar(loop=loop, unsafe=True)
     jar.update_cookies(SimpleCookie("cookie=val; Domain=example.com.;"),
                        URL("http://www.example.com"))
     cookies_sent = jar.filter_cookies(URL("http://www.example.com/"))
-    assert cookies_sent.output(header='Cookie:') == "Cookie: cookie=val"
+    assert cookies_sent.output(header='Cookie:') \
+           == "Cookie: cookie=val; Domain=www.example.com; Path=/"
```
popravich: This is wrong as this will appear as 3 cookies: `cookie`, `Domain` and `Path`.
Please see rfc for `Cookie` header http://httpwg.org/specs/rfc6265.html#sane-cookie
Its the `Set-Cookie` header that can accept domain/expire/path option.

thehesiod: ah, sorry I was figuring that would get filtered out later, I've updated the code to filter these out upfront

asvetlov:
Thank you

thehesiod:
thanks!  sorry about the misstep :)

Issue #1443, opened 2016-12-01
Use trafaret_config in polls demo


asvetlov:
I like https://github.com/tailhook/trafaret_config and want to use it in demo: https://github.com/KeepSafe/aiohttp/tree/master/demos/polls

The task is trivial.
Any volunteer?

asvetlov:
Fixed by #1462

PR #1456, opened 2016-12-06
Update db preparation script

gyermolenko:
This PR does the following:

- Create `init_db.py` script to create tables via sqlalchemy according to existing models and not recreate them with raw sql (replacement for `create_tables.sql` and, consequently, for `install.sh`).

- fix db config initialization
- fix title in `results.html`
    
If/when it merges I could update `Database schema` part of tutorial documentation

### DIFF
--- a/demos/polls/aiohttpdemo_polls/db.py
+++ b/demos/polls/aiohttpdemo_polls/db.py
@@ -36,7 +36,7 @@ class RecordNotFound(Exception):
 
 
 async def init_pg(app):
-    conf = app['config']
+    conf = app['config']['postgres']
     engine = await aiopg.sa.create_engine(
         database=conf['database'],
         user=conf['user'],
--- a/demos/polls/aiohttpdemo_polls/templates/results.html
+++ b/demos/polls/aiohttpdemo_polls/templates/results.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 
-{% set title = question.question %}
+{% set title = question.question_text %}
 
 {% block content %}
 <ul>
--- a/demos/polls/init_db.py
+++ b/demos/polls/init_db.py
@@ -1,0 +1,50 @@
+import subprocess
+
+import sqlalchemy as sa
+import yaml
+
+from aiohttpdemo_polls.db import question, choice
+
+
+config_path = 'config/polls.yaml'
+with open(config_path) as f:
+    data = yaml.load(f)
+
+CONF = data['postgres']
+
+DSN = '{drivername}://{username}:{password}@{host}:{port}/{database}'.format(
+    drivername='postgresql',
+    username=CONF['user'],
+    password=CONF['password'],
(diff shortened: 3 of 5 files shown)


asvetlov:
@gyermolenko would you take a look on #1443 first?

I really want to have it implemented before working on this PR.

Next topic is: @jettify used simple SQL scripts instead of `alembic` tool for sake of simplicity.
But if we want to move ahead a proper demonstration of alembic usage is better maybe?

gyermolenko:
Ok, no problem )
I will take a look at trafaret (and try to apply it for polls). Never heard of it before. 
Alembic is also a good idea, I will think about it.

asvetlov:
Please split tasks 
`trafaret_config` first and `alembic` later.
For saving your time please discuss the need for alembic first.

gyermolenko:
Split tasks - sure.
As for "discuss the need for alembic" - where/with who?

jettify:
In `aiohttp_admin` demos  I end up with fake data generating script + docker-compose file (https://github.com/aio-libs/aiohttp_admin/blob/master/demos/polls/aiohttpdemo_polls/generate_data.py). Having only docker it is easy to get demo running. 
```
make docker_start_pg
make fake_data
make run
```
and no more fighting with postgres user/permission configuration.

jettify:
+1 for `alembic` it will be very useful.

popravich:
-1 for alembic — it has its own tutorials.
polls demo db is quite simple and can be initialized from python (or shell) script.

jettify:
depends what is our goal, we want to show simple project as a demo of aiohttp related features? or we want full feature project to show users how to organize code in the project?

gyermolenko:
For the sake of comparison with other tutorials: 
- Django uses native migrations
- Flask ([Flaskr app](http://flask.pocoo.org/docs/0.11/tutorial/dbinit/)) doesn't use migrations (and no models.py or alike)
- [Flask mega tutorial](https://blog.miguelgrinberg.com/post/the-flask-mega-tutorial-part-iv-database) uses SQLAlchemy-migrate, although this part was written in 2012.

popravich:
> depends what is our goal, we want to show simple project (**A**) as a demo of aiohttp related features? or we want full feature (**B**) project to show users how to organize code in the project?

IMHO:
I'm for **A** — less code to learn in example — faster to start. And it's more kinda *aiohttp demo* not *how to code demo*...
**B** can be an option for separate set of tutorials

AraHaan:
So, is this still being worked on or not? Also is this even needed anymore? @fafhrd91 @asvetlov mind letting me know if this is really needed anymore now that a lot of things has changed since then?

asvetlov:
Sorry, but I'm closing the PR.
Subprocess calls for DB tables creation is bad practice I believe.
Better to leave the demo as is -- people will learn about db setup etc. anyway

PR #1457, opened 2016-12-07
Introduce trafaret config validator

gyermolenko:
## Related issue number
#1443

### DIFF
--- a/demos/polls/aiohttpdemo_polls/db.py
+++ b/demos/polls/aiohttpdemo_polls/db.py
@@ -36,7 +36,7 @@ class RecordNotFound(Exception):
 
 
 async def init_pg(app):
-    conf = app['config']
+    conf = app['config']['postgres']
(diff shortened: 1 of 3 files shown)


asvetlov:
Thanks.
Next step is converting from `read_and_validate()` to `commandline.standard_argparse_options(ap, default_config='config.yaml')` :)

gyermolenko:
Let me clarify:
1. you wanted to validate config onload (that part was in this PR)
2. and you also want to validate some custom config from command line, w/o starting server

Is it correct? I just want to understand the use case.
Also..  with this merge new dependency is introduced. Did I miss some requirements.txt for that?

asvetlov:
I've fixed requirements by 004ae3f
No, I want keep trafaret from the PR but instead of explicit `read_and_validate` call use configured argument parser approach.

PR #1480, opened 2016-12-13
Deprecate changing the status code of a response after it has been prepared

arthurdarcet:
Changing the status of a Response after it has been prepared doesn't do anything and should raise an exception imo.
Since breaking all apps that do this might not be an option, I've added a warning when this happens and I would suggest changing the warning to an exception in the next major release (not sure what the policy on breaking changes is, so up to you!)

(this PR is missing updates in CHANGES/…, I can add those if this looks good to you)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,9 @@ CHANGES
 - Avoid a race when application might start accepting incoming requests
   but startup signals are not processed yet e98e8c6
 
+- Raise a `RuntimeError` when trying to change the status of the HTTP response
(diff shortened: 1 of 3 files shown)


asvetlov:
I prefer considering the PR as bug fix and raise `RuntimeError` without deprecation period.
I doubt if there are many improper uses of `.set_status` in existing code base.

arthurdarcet:
@asvetlov works for me, I have updated the PR to raise a RuntimeError (+ added an entry in CHANGES.rst and added a test)

PR #1484, opened 2016-12-16
use UTF-8 as the default encoding for multipart text parts

arthurdarcet:
After some tests on Chrome, it seems in multipart/form-data bodies, regular field=value parts are encoded using utf-8, not latin1

https://www.w3.org/TR/html5/forms.html#multipart/form-data-encoding-algorithm
https://dvcs.w3.org/hg/xhr/raw-file/tip/Overview.html#dom-xmlhttprequest-send

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -394,7 +394,9 @@ class BodyPartReader(object):
         :rtype: str
         """
         data = yield from self.read(decode=True)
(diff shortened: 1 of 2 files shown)


asvetlov:
Sounds reasonable

kxepal:
It's good to see more UTF-8 defaults around (: Thanks!

PR #1496, opened 2016-12-19
add notice about sup-application URL reversing from main app

toidi:
Subj.

Ref: https://github.com/KeepSafe/aiohttp/issues/1489


asvetlov:
Thanks

PR #1500, opened 2016-12-23
Update Connection header to match RFC 6455 case

### LINKED ISSUE #1495: Webscockets client timeout when connecting to Kodi
## Long story short

There is an incompatibility causing a timeout in aiohttp when connecting to the Kodi websockets interface. Kodi is working with the `websockets` python library. This is very possibly some poor behavior on Kodi's part.

Background: https://github.com/home-assistant/home-assistant/pull/4951#discussion_r93008906

## Expected behaviour

Websocket connection is opened successfully.

## Actual behaviour

aiohttp is hanging here: https://github.com/KeepSafe/aiohttp/blob/v1.2.0/aiohttp/client.py#L302 In the logs, `HTTP/1.1 400 Bad Request` is shown.

## Steps to reproduce

```python
import aiohttp
import asyncio

@asyncio.coroutine
def main(loop):
    client = aiohttp.ClientSession(loop=loop)
    ws = yield from client.ws_connect('ws://192.168.1.193:9090/jsonrpc')
    while True:
        msg = yield from ws.receive()
            print(msg)
        elif msg.type == aiohttp.WSMsgType.CLOSED:
            break
        elif msg.type == aiohttp.WSMsgType.ERROR:
            break

```

## Your environment
Python 3.4.3
aiohttp 1.2.0
Kodi 16.1

armills:
## What do these changes do?

This change modifies the case of the 'Connection' header value to match the case used in [RFC 6455](https://tools.ietf.org/html/rfc6455#section-4.2.1). Although section 4.2.1 paragraph 4 instructs the server to treat the value as case-insensitive, if aiohttp needs to pick a case, we should use the one defined in the spec.

No

## Related issue number

Fixes #1495 

## Checklist

### DIFF
--- a/aiohttp/protocol.py
+++ b/aiohttp/protocol.py
@@ -684,7 +684,7 @@ class HttpMessage(ABC):
         # set the connection header
         connection = None
         if self.upgrade:
-            connection = 'upgrade'
+            connection = 'Upgrade'
         elif not self.closing if self.keepalive is None else self.keepalive:
             if self.version == HttpVersion10:
                 connection = 'keep-alive'
--- a/tests/test_protocol.py
+++ b/tests/test_protocol.py
@@ -216,7 +216,7 @@ def test_default_headers_connection_upgrade(transport):
     msg.upgrade = True
     msg._add_default_headers()
 
-    assert msg.headers['Connection'] == 'upgrade'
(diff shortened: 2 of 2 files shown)


AraHaan:
hmm this seems good I think.

asvetlov:
@armills thanks

PR #1501, opened 2016-12-24
Add recreate method to aiohttp.ClientSession().

AraHaan:
This allows a person to create a ``ClientSession`` that was closed without having to reinitialize a class.

There is a problem this fixes. Classes are not destructed after initialized even after the variable that the initialization goes it is overridden by another instance causing a Resource Leak until the Python interpreter is closed. This fixes that issue by explicitly adding an recreate method to eliminate the need to reinitialize the class.

## What do these changes do?

Adds a way to create a ClientSession without having to reinitialize the class causign a Resource and a Memory leak if not careful.

Yes, instead of overwriting a closed Client Session Instance with a new instance they should use the new ``recreate`` method I made in this PR instead to reduce Memory and Resource leaks.

## Related issue number

#1503 

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,10 @@ CHANGES
 
 - Fix polls demo run application #1487
 
--
+- Add recreate method to aiohttp.ClientSession(). #1501
+  This addresses an Issue I spotted with ClientSession in a comment to #1468 found here https://github.com/KeepSafe/aiohttp/pull/1468#issuecomment-269052523.
+
+- 
 
 1.2.0 (2016-12-17)
 ------------------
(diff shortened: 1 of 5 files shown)


AraHaan:
Working on docs and unit tests and all that to get all those checks done.

AraHaan:
I am thinking that the failing tests needs some fixing. If only all tests had a logger that shows what all fails and their stack trace (if there is a traceback generated) to actually know what tests fails, is skipped, and passes by name.

asvetlov:
aiohttp will never have hacks like this.
ClientSessions's life circle is: Init -> Use -> Close

AraHaan:
But this does help those who do things just like discord.py does.

PR #1502, opened 2016-12-24
add numprocs in supervisord

o3o3o:
add `numprocs` in supervisord

## What do these changes do?

use  `numprocs` filed in supervisord.conf for supporting run the mutilple app.


asvetlov:
Thanks!

PR #1505, opened 2016-12-24
Changed close method to delete the instance itself.

AraHaan:
This is to help those who forget to delete the instance of ClientSession
after closing it.

## What do these changes do?

delete the instance in close() so that way people do not forget to delete it after closing their Client Session

No user code should get broken with this. This should help the end user so they don't have to worry about instances not getting deleted on their end.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ CHANGES
 
 -
 
+- close() now deletes the instance for those who forget to delete it in their code. #1505
+
 1.2.0 (2016-12-17)
 ------------------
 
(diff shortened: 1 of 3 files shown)


AraHaan:
Ok, @asvetlov this is better than the previous PR as this deletes the instance for the people who forget to (or do not know to delete it after closing) and this actually helps make it optional for people to delete it when doing it this way way. It also catches any issues when the client session is lazy ``__init__``'d in a class when that session is closed  and the class instance variable is not deleted before reinitializing the client session class. Also it eliminates the hacks and does what you want as well as cleans up the instance.

asvetlov:
The PR doesn't make any sense.
@AraHaan please read how reference counting works first.
`del` doesn't call destructor immediately but removes a variable name from namespace and decrements refcount.

AraHaan:
hmm

PR #1510, opened 2016-12-25
remove Application dependency from UrlDispatcher

fafhrd91:
I think UrlDispatcher should not depend on Application object. 
Here is backward compatible change to remove this dependency.

@asvetlov

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -18,7 +18,7 @@ from .web_exceptions import *  # noqa
 from .web_reqrep import *  # noqa
 from .web_server import Server
 from .web_urldispatcher import *  # noqa
-from .web_urldispatcher import PrefixedSubAppResource, _wrap_add_subbapp
+from .web_urldispatcher import PrefixedSubAppResource
 from .web_ws import *  # noqa
 
 __all__ = (web_reqrep.__all__ +
@@ -40,6 +40,7 @@ class Application(MutableMapping):
             router = web_urldispatcher.UrlDispatcher()
         assert isinstance(router, AbstractRouter), router
 
+        # backward compatibility until full deprecation
         router.add_subapp = _wrap_add_subbapp(self)
 
         if debug is ...:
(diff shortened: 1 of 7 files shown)


fafhrd91:
we need to update docs and add deprecation warning to `router.add_subapp`

i think `UrlDispatcher` should be part of public api

aiohttp/web.py
```diff
@@ -125,6 +126,23 @@ def handler(app):
         reg_handler('on_shutdown')
         reg_handler('on_cleanup')
 
+    def add_subapp(self, prefix, subapp):
+        if self.frozen:
+            raise RuntimeError(
+                "Cannod add sub application to frozen application")
+        if subapp.frozen:
+            raise RuntimeError("Cannod add frozen application")
```
popravich: "cannod" typo

aiohttp/web_urldispatcher.py
```diff
@@ -891,20 +883,25 @@ def add_delete(self, *args, **kwargs):
...
+        super().freeze()
+        for resource in self._resources:
+            resource.freeze()
+
+
+def _wrap_add_subbapp(app):
+
+    def add_subapp(prefix, subapp):
         if subapp.frozen:
             raise RuntimeError("Cannod add frozen application")
```
popravich: ...a-ha, copied from here — "cannod"

fafhrd91: good catch

fafhrd91:
@asvetlov objections?

also what about `UrlDispatcher` as part of public api?

asvetlov:
Hmm. I dislike `router.add_subapp = _wrap_add_subbapp(self)` -- looks like a hack.

Also I doubt if it will work with all possible router implementations.
Adding `add_subapp` to `Application` doesn't look reasonable -- we are adding a way to register sub applications only but not regular routes.

I suggest keeping the code as is.

fafhrd91:
`router.add_subapp` just for backward compatibility until next release

PR #1507, opened 2016-12-25
Fixed ClientSession initialization warning

AraHaan:
This fixes an issue with ClientSession throwing a warnings when feeding
an event loop into it when calling it from a normal function. This
should help silence annoying warnings for libraries that "lazy"
initializes ClientSession and properly closes the session and handles
it. Now those libraries would have no need for a useless bug reports.

## What do these changes do?

Fixes the warning when calling a ClientSession instance with an event loop that is passed to the class.

## Are there changes in behavior for the user?

No, just adds more control over the ClientSession instance.

## Related issue number

https://github.com/KeepSafe/aiohttp/pull/1468#issuecomment-269112177

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new entry to `CHANGES.rst`
  * Choose any open position to avoid merge conflicts with other PRs.
  * Add a link to the issue you are fixing (if any) using `#issue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

PR #1518, opened 2016-12-28
Fixed ClientSession initialization warning

AraHaan:
This fixes an issue with ClientSession throwing a warning when feeding
an event loop into it when calling it from a normal function. This
should help silence annoying warnings for libraries that "lazy"
initializes ClientSession and properly closes the session and handles
it. Now those libraries would have no need for a useless bug reports.

## What do these changes do?

Fixes the warning when calling a ClientSession instance with an event loop that is passed to the class.

No, just adds more control over the ClientSession instance.

## Related issue number

https://github.com/KeepSafe/aiohttp/pull/1468#issuecomment-269112177

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -41,29 +41,42 @@ class ClientSession:
                  version=aiohttp.HttpVersion11,
                  cookie_jar=None):
 
-        if connector is None:
-            connector = aiohttp.TCPConnector(loop=loop)
-            loop = connector._loop  # never None
-        else:
-            if loop is None:
+        self._set_explicit_loop = False
+        try:
+            if connector is None:
+                connector = aiohttp.TCPConnector(loop=loop)
                 loop = connector._loop  # never None
-            elif connector._loop is not loop:
-                raise ValueError("loop argument must agree with connector")
+                self._set_explicit_loop = True
+            else:
+                if loop is None:
+                    loop = connector._loop  # never None
+                    self._set_explicit_loop = True
+                elif connector._loop is not loop:
+                    raise ValueError(
+                        "loop argument must agree with connector")
 
-        self._loop = loop
-        if loop.get_debug():
-            self._source_traceback = traceback.extract_stack(sys._getframe(1))
+            self._loop = loop
+            if loop.get_debug():
+                self._source_traceback = traceback.extract_stack(
+                    sys._getframe(1))
+        except RuntimeError:
+            # We know for sure now that there was no event loop specified.
+            # TODO: throw a warning now here to but the warning must not use
+            # the exception handler from event loops as there is None.
+            self._set_explicit_loop = False  # for now RIP.
 
-        if not loop.is_running():
-            warnings.warn("Creating a client session outside of coroutine is "
-                          "a very dangerous idea", ResourceWarning,
-                          stacklevel=2)
-            context = {'client_session': self,
-                       'message': 'Creating a client session outside '
-                       'of coroutine'}
-            if self._source_traceback is not None:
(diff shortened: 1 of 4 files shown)


AraHaan:
I had to remake #1507 to fix a Merge Conflict for just 1 commit. Sorry about that.

aiohttp/client.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, *, connector=None, loop=None, cookies=None,
         if loop.get_debug():
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
-        if not loop.is_running():
+        if not loop.is_running() and loop is not None:
```
asvetlov: Check for `None` shoud be first: it's fastpath.

AraHaan: ok.

AraHaan:
Ok, doing another squash commit of this that contains the change you suggested by having the ``None`` check 1st.

AraHaan:
Well now there is an issue with the tests. They both now Fail because of the fact of no event loop running in the main loop because ``get_event_loop()`` throws an RuntimeError. The problem with that is that I dont want to create a new event loop within it if loop is None so, it now prevents people from creating a ClientSession outside of a coroutine without speechifying an event loop in it.

aiohttp/client.py
AraHaan: I fixed an issue in the None check I wrote ``is not none`` instead of ``is None``.

AraHaan:
Well with these changes. There is no way to successfully create a client session without an event loop as no matter what ``asyncio.get_event_loop()`` will fail on the main thread. Sadly even the test is now broken still.

AraHaan:
hmm @asvetlov mind helping with fixing up the test? I have no idea why it seems to always fail no matter what I do so that way it returns the coverage and covers in a way where it does warn with this patch, The issue is that the warning comes after an RuntimeError is thrown from ``get_event_loop`` and so the warning just never shows making the tests fail. If the warning was moved up to be the first thing before getting the event loop maybe it would help with the tests as then I could just make the tests catch that ``RuntimeError``.

aiohttp/client.py
AraHaan: uh, so I tested it and sadly the final for loop line must be ``if loop is not None:`` because otherwise if I check if the loop is running with the current ``None`` check it would raise an ``AttributeError``.

AraHaan:
Ok, now I think it is set. Sadly the only issue is that when ``loop`` is ``None`` then the call to ``loop.call_exception_handler`` fails on it sadly currently. So this means that the class would need to make it's own exception handler for this case to work correctly.

AraHaan:
ok, am trying my best however to handle this perfectly. However I think you might need to check and see if asyncio throws an ``RuntimeError``, intercept it and pass from it. and then if ``loop`` is ``None`` then raise an ``RuntimeError`` instead otherwise the exception handler for the loop fails because the loop is None.

AraHaan:
Ok, I had to sadly turn that warning into an ``RuntimeError`` but the real problem is that I had to intercept an asyncio ``RuntimeError just so that way the test case that this is supposed to hit finally actually works. This should now not warn when an event loop is not running more like when ``ClientSession`` has no event loop and asyncio fails to give it an event loop as well. So in this case when they try to create an ClientSession outside of a coroutine and do not pass in ``loop=`` where loop is an event loop.

AraHaan:
Ok, So Sadly I might need to reopen this PR so this is only 1 commit. And so it would be also up to date with this repo.

AraHaan:
Remade as #1572.

PR #1520, opened 2016-12-29
Simplify waiter in streams.py

AraHaan:
This should reduce some uneeded branches and lines of code that are
partials because the fact that there is no case where waiter is canceled
before the if statement in the tests. This should also simplify it a bit
as well.

## What do these changes do?

Simplifies the waiter and helps reduce the amount of partial lines on the coverage as well as reduces some branches on it.

There should not really be any noticed changes in behavior for the user.

## Related issue number

None.

## Checklist

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -118,14 +118,26 @@ class StreamReader(AsyncStreamReaderMixin):
     def exception(self):
         return self._exception
 
-    def set_exception(self, exc):
-        self._exception = exc
-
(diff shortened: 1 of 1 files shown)


AraHaan:
There should now be at least 4 partials instead of 3 in ``streams.py`` Until the Tests or something actually cancels the waiter so that way the other part of the cancellation check can be considered fulfilled on the Coverage. Other than that this should help somewhat.

PR #1521, opened 2016-12-29
Fixed misses in helpers

AraHaan:
This commit fixes all 4 of the misses in helpers.py.

## What do these changes do?

Fix the Misses in helpers.py

No changes in behavior at all.

## Related issue number

None

## Checklist

### DIFF
--- a/aiohttp/__init__.py
+++ b/aiohttp/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '1.2.1a'
+__version__ = '1.2.1a0'
 
 # Deprecated, keep it here for a while for backward compatibility.
 import multidict  # noqa
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -21,10 +21,13 @@ from multidict import MultiDict, MultiDictProxy
 
 from . import hdrs
 
-try:
-    from asyncio import ensure_future
-except ImportError:
-    ensure_future = asyncio.async
+# Because this has asyncio.async as a 3rd arg for getattr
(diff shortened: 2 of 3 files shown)


AraHaan:
@fafhrd91 What you think of this PR? Also I noticed there was a missing 0 in the version so I fixed that as well in this PR, btw you could also look in #1518 as well to. It also contains my entry in ``CONTRIBUTORS.txt``

fafhrd91:
i don't think it worth to create new module just for import checks.
also it seems nothing uses `check_loop`

AraHaan:
So, Check loop can be safely removed. I see.

AraHaan:
Ok, then check_loop has now been removed completely since it is not getting used anyway so yeah.

AraHaan:
hmm if only asyncio had that get_future instead. hmmm

fafhrd91:
I leave this PR to @asvetlov
He is active maintainer

AraHaan:
Ok, found a way to help get what we need. Using ``getattr`` for if ``asyncio.ensure_future`` does not exist it returns ``asyncio.async`` and use 1 line.

ss 2016-12-30 at 04 02 04

AraHaan:
Looks like I just had to monkey patch 1 of the tests as sometimes it would fail at ``web.run_app`` hmm but is more often then not rare. So, when it happens should make it instead of the ``pass`` that I am doing right now to make it ``xfail``

asvetlov:
I've dropped `check_loop` and fixed version number. 39b8568
`ensure_future` is covered by running on different python versions.

PR #1523, opened 2017-01-03
separate read + connect + request timeouts

thehesiod:
This is an attempt at fixing the timeout issues raised in https://github.com/KeepSafe/aiohttp/issues/1524 + https://github.com/KeepSafe/aiohttp/issues/1180


thehesiod:
I updated the implementation to allow for unlimited timeouts by passing`request(timeout=None)`

adamrothman:
@jettify @asvetlov what do you guys think about this? Would make it a lot easier to use `aiohttp` in more places.

thehesiod:
adding @fafhrd91

fafhrd91:
looks good. i am not sure about proper api though. lets add read_timeout but lets wait with documentation.

i did some work on TimeService but for server, we may introduce time_service for client as well, so it'd be possible to use third-party time service.

thehesiod:
I think it was a mistake making the default `_request` timeout 5min, it should have been None and the details handled internally, it makes backwards compatible fixes difficult as we can't differential between the user setting it to 5minutes, or them wanting the default behavior.

thehesiod:
cool, let me know what you want me do.

fafhrd91:
`request timeout 5min` where is it? if we have 5min timeout, it should painless to replace it with `None` i doubt anyone notice.

fafhrd91:
i'll merge this change. if you'd like you can look into time_service integration

fafhrd91:
thanks

thehesiod:
https://github.com/KeepSafe/aiohttp/pull/1523/files#diff-7dd84b5ef8d5eea2de1dfc5329411dfcL127 let me know if you want me change it to None and update the logic

fafhrd91:
ok, lets replace it with None

PR #1530, opened 2017-01-06
fix #1525: add missing sub MultipartWriter headers

### LINKED ISSUE #1525: writing nested multipart messages does not work correctly
## Long story short

Write the following multipart message:
```python
from aiohttp.multipart import MultipartWriter
with MultipartWriter('mixed') as root:
  with MultipartWriter('mixed') as subwriter1:
    subwriter1.append('first message')
  root.append(subwriter1)

  with MultipartWriter('mixed') as subwriter2:
    subwriter2.append('second message')
  root.append(subwriter2)

print(b''.join(root.serialize()).decode('ascii'))
```

## Expected behaviour

```
--0d20b463316a462d839e93130150f5ef
Content-Type: multipart/mixed; boundary="05af9536a8374279a8bccc2ea099c82b"

--05af9536a8374279a8bccc2ea099c82b
Content-Length: 13

first message
--05af9536a8374279a8bccc2ea099c82b--

--0d20b463316a462d839e93130150f5ef
Content-Type: multipart/mixed; boundary="1d4fdf1ddf8e46019da125c22c247649"

--1d4fdf1ddf8e46019da125c22c247649
Content-Length: 14

second message
--1d4fdf1ddf8e46019da125c22c247649--

--0d20b463316a462d839e93130150f5ef--

```

## Actual behaviour

```
--d3b608e9626a4f3cb2329fa01ebb1fce

--9310b9861e56425daf7ef0773fd2ae54
Content-Length: 13

first message
--9310b9861e56425daf7ef0773fd2ae54--

--d3b608e9626a4f3cb2329fa01ebb1fce

--6fa9c511ee2d4a36bec615b15da6f1b2
Content-Length: 14

second message
--6fa9c511ee2d4a36bec615b15da6f1b2--

--d3b608e9626a4f3cb2329fa01ebb1fce--

```

## Steps to reproduce

Writer a nested multipart message as shown in the documentation.

**"Fix"**: `append(subwriter, headers=subwriter.headers)`

_I'll try to commit a patch, but just reporting the issue right now_

terencehonles:
When appending MultipartWriter objects to another MultipartWriter use
the subwriter's headers in the constructed BodyPartWriter

fixes issue #1525 

## What do these changes do?

Fixes issue #1525 by allowing the wrapping BodyPartWriter to use the containing MultipartWriter's headers. This fixes the issue with the missing headers and also allows additional header changes until `.serialize()` is called.

Documentation is now correct, and if the user previously explicitly passed the subwriter's headers when appending the object the headers will already be in the subwriter's headers object so it is effectively a no-op replacing existing values with their same values.

## Related issue number

#1525 

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,7 @@ CHANGES
 
 - Fix polls demo run application #1487
 
--
+- Fix sub-Multipart messages missing their headers on serialization #1525
 
 1.2.0 (2016-12-17)
 ------------------
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -133,6 +133,7 @@ Steven Seguin
 Sviatoslav Bulbakha
 Taha Jahangir
 Taras Voinarovskyi
+Terence Honles
 Thomas Grainger
 Tolga Tezel
 Vaibhav Sagar
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -670,7 +670,11 @@ class BodyPartWriter(object):
     """Multipart writer for single body part."""
 
     def __init__(self, obj, headers=None, *, chunk_size=8192):
-        if headers is None:
+        if isinstance(obj, MultipartWriter):
+            if headers is not None:
+                obj.headers.update(headers)
+            headers = obj.headers
+        elif headers is None:
             headers = CIMultiDict()
         elif not isinstance(headers, CIMultiDict):
             headers = CIMultiDict(headers)
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -838,6 +838,12 @@ class BodyPartWriterTestCase(unittest.TestCase):
         multipart = aiohttp.multipart.MultipartWriter(boundary=':')
         multipart.append('foo-bar-baz')
         multipart.append_json({'test': 'passed'})
+        multipart.append_form({'test': 'passed'})
+        multipart.append_form([('one', 1), ('two', 2)])
+        sub_multipart = aiohttp.multipart.MultipartWriter(boundary='::')
+        sub_multipart.append('nested content')
+        sub_multipart.headers['X-CUSTOM'] = 'test'
+        multipart.append(sub_multipart)
         self.assertEqual(
             [b'--:\r\n',
              b'Content-Type: text/plain; charset=utf-8\r\n'
@@ -845,10 +851,36 @@ class BodyPartWriterTestCase(unittest.TestCase):
              b'\r\n\r\n',
              b'foo-bar-baz',
              b'\r\n',
+
              b'--:\r\n',
              b'Content-Type: application/json',
              b'\r\n\r\n',
              b'{"test": "passed"}',
+             b'\r\n',
+
+             b'--:\r\n',
+             b'Content-Type: application/x-www-form-urlencoded',
+             b'\r\n\r\n',
+             b'test=passed',
+             b'\r\n',
+
+             b'--:\r\n',
+             b'Content-Type: application/x-www-form-urlencoded',
+             b'\r\n\r\n',
+             b'one=1&two=2',
+             b'\r\n',
+
+             b'--:\r\n',
+             b'Content-Type: multipart/mixed; boundary="::"\r\nX-Custom: test',
+             b'\r\n\r\n',
+             b'--::\r\n',
+             b'Content-Type: text/plain; charset=utf-8\r\n'
+             b'Content-Length: 14',
+             b'\r\n\r\n',
+             b'nested content',
+             b'\r\n',
+             b'--::--\r\n',
+             b'',
              b'\r\n',
              b'--:--\r\n',
              b''],
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -1014,6 +1014,13 @@ class BodyPartWriterTestCase(unittest.TestCase):
         self.part.set_content_disposition('related', filename='foo.html')
         self.assertEqual('foo.html', self.part.filename)
 
+    def test_wrap_multipart(self):
+        writer = aiohttp.multipart.MultipartWriter(boundary=':')
+        part = aiohttp.multipart.BodyPartWriter(writer)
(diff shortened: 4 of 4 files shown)


terencehonles:
@kxepal not sure if you're the multipart owner, but this is related to the other multipart issue I raised. Ty

kxepal:
@terencehonles eventually (:
But why not to let MultipartWriter have others MultipartWriters as a part? Having twisted mix of MultipartWriters that includes BodyPartWriters which holds else MultpartsWriters and so on and so deeper looks and feels quite strange. Assuming Multpart message is a tree, MultipartWriters are nodes while BodyPartWriters are the leaves.

terencehonles:
@kxepal but that's not what is currently written and written that way would actually make it more complicated in code. If multipart messages were nodes in the tree then they would have to know if they are the root node or not. If they were _not_ the root node they would serialize their headers and body as the `BodyPartWriter.serialize()` does. If they were at the root they would omit that information and let whatever was containing it write the headers as needed.

By keeping all serialization code in the `BodyPartWriter` it simplifies how the encoding and serialization is done (and it doesn't change how the code is currently organized). Which leaves `MultipartWriter` as a sort of tree builder (via `.append(...)`) and a root to call `.serialize()` on.

If you think it is awkward that the user can find out that a `BodyPartWriter` is being used, then we _could_ special case the append to allow both `BodyPartWriter` or `MultipartWriter` in the `parts` list and when serializing wrap the `MultipartWriter` in a `BodyPartWriter` and serialize that, but the downside is we are doing more work during serialization which means there are more chances for exceptions to be raised.

The nice thing about how the code is currently written is `MultipartWriter.serialize()` only cares that it emits boundaries between all its parts. The part handles how it needs to be serialized and it does so. If the part is _also_ a MultipartWriter it allows it to handle the **body** serialization, but it still handles everything else. The way it is written is _really_ seems like how it should be written.

kxepal:
@terencehonles 
Good points. I'm indeed missed few details there. Ok, let's have it so. No lightweight alternatives comes to mind anyway.

Please, add few more tests about this change to let us remember about this use case.

terencehonles:
@kxepal sounds good :+1:

terencehonles:
@kxepal I'm not sure what needs to be done to satisfy those checks that failed

kxepal:
@terencehonles 
It's about patch code coverage. Check [codecov report](https://codecov.io/gh/KeepSafe/aiohttp/compare/63d1ad544b6489a538e53988f05dfdf5f45fceed...59f4ce7c930ce4cd1dbede9fa4aef71cde7ed6d1).
The yellow line is a partial covered branch. This means that there are two possible cases for this line: when condition is True and you fall into the `if` case and when it's False and you don't enter there. It seems that the False-case wasn't satisfied. Probably, worth to add a case when you append a MultipartWriter and not pass custom headers.

terencehonles:
@kxepal how do you even know what is covering it or not? This is _not_ actually the case you are talking about because `MultipartWriter.append(...)` **always** passes an object for `headers` and it is never `None`. If there was something covering it before it was part of another test case (which goes back to my original question)

kxepal:
@terencehonles 

>  how do you even know what is covering it or not? 

The codecov report says that with colors (;

> This is not actually the case you are talking about because MultipartWriter.append(...) always passes an object for headers and it is never None.

I'm about `headers` argument in `BodyPartWriter.__init__` - you have a condition there. We're talking about [the same](https://github.com/KeepSafe/aiohttp/pull/1530/files#diff-235eeadd1960735e1191879aa500b7b7R674) codeline, right?

terencehonles:
@kxepal you're answering a slightly different question. I asked _what is covering a line_ you're answering _if a line is covered_.

We're talking about the same code path, but the only way to exercise that code path is to create a `BodyPartWriter` with a multipart object during construction. None of the tests currently explicitly check construction of the `BodyPartWriter` but rely on the one created in the `setUp` method. (This is why I was asking the first question). I added a test case which covers instantiation, but I'd be willing to bet our code coverage report is lying to us.

kxepal:
@terencehonles 
Sorry, I didn't understand your question (actually, I can't even find it) .

But anyway, yes, it seems the only way to cover those branch is to test BodyPartWriter constructor explicitly - that's not a problem, I think (: There were no else such tests because setup test suite and else tests did that job well. Not very unit approach, but quite dry. So that one should fill the blanks.

Coverage report here doesn't lies, just tells which branches were not covered by test run.

terencehonles:
> how do you even know what is covering it [a line] ~~or not~~?
Was the question.

Unless you know _what_ is covering a line and what data it was using there's a possibility it is not being meaningfully covered. The colors only show that _some_ line in the test suite exercised the code not that it completely exercised it with the correct data (that's why I was wondering if there was a way to backtrack). I don't have enough time right now to look at the tests to figure out if this is currently a problem, but it's great that we even have as many tests as we do.

kxepal:
@terencehonles 
Well, if you put there those condition, you had some reasons for it. But every condition causes branches and that's what coverage ask you to do: cover them all e.g. make sure each to be executed somehow. It's not and never about the data or any meaning. You can remove those condition and code coverage will be full, but not the data, so there be a bug. But that's about another kind of coverage and it need different tools to use. 

The current failure seems to be not your fault, but some flaky test in webserver about serving static content. I triggered it retry.

terencehonles:
@kxepal it seems like we're not understanding each other, but it's not really something that needs to be continued on this PR.

My point was we need to make sure we write tests to test the code not just to make the coverage tool show green. The tool helps us see what we might have missed, but I'm arguing it might be more generous than it should be when saying a line is "covered". Anyways, I'm fine leaving the discussion wherever it is :)

kxepal:
@terencehonles 
No, I finally get your point, but it's still about different. Coverage tool checks only code coverage, so we need at least such set of tests that covers all the code statements (not even lines). But this doesn't means that this set of tests is enough to say that we tested our code right and complete. There could be more tests about edge cases. Suddenly, there is no tools to show how better you catch all such cases, only to help you reach them.

Anyway, for current PR it all looks fine. +1

@asvetlov, any notes before merge?

fafhrd91:
lgtm

PR #1532, opened 2017-01-07
fix #1526: multipart messages should consume their epilogue

### LINKED ISSUE #1526: reading nested multipart messages does not work correctly
## Long story short

Multipart reader breaks after reading a sub multipart end boundary and starting the next part

## Expected behaviour

Nested multipart reader reads a message created with the multipart writer correctly

## Actual behaviour

```python
ValueError: Invalid boundary b'', expected b'--b0b69248b3a345cf8256a8dd25f07874'
```

## Steps to reproduce

Receive the multipart response from #1525 

Server:
```python
from aiohttp.multipart import MultipartWriter
from aiohttp.web import Response

def handle(request):
  with MultipartWriter('mixed') as root:
    with MultipartWriter('mixed') as subwriter1:
      subwriter1.append('first message')
    root.append(subwriter1, headers=subwriter1.headers)

    with MultipartWriter('mixed') as subwriter2:
      subwriter2.append('second message')
    root.append(subwriter2, headers=subwriter2.headers)
  return Response(body=b''.join(root.serialize()), headers=root.headers)

# ... create web app which responds with the handler above ...
```

Client:
```python
import aiohttp
import asyncio
from aiohttp.multipart import BodyPartReader, MultipartReader

@asyncio.coroutine
def read_multipart(reader):
  while True:
    part = yield from reader.next()
    if part is None: break
    if isinstance(part, BodyPartReader):
      body = yield from part.read(decode=True)
      print('body part: %r' % body)
    else:
      print('nested part')
      yield from read_multipart(part)

@asyncio.coroutine
def request(url):
  response = yield from aiohttp.get(url)
  yield from read_multipart(MultipartReader.from_response(response))

# ... drive event loop and call request(handler_url) ...
```

## Issue

Lines [multipart.py:767](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L767) and [multipart.py:969](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L969) have line endings which result in an empty line after the end boundary of a multipart message (valid). However the reader terminates after the end boundary and the parent reader now expects the next boundary, but what is found is the blank line from [multipart.py:767](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L767)

## Possible fix

```diff
diff --git a/aiohttp/multipart.py b/aiohttp/multipart.py
index af7f19b1..82ad2306 100644
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -639,6 +639,7 @@ class MultipartReader(object):
             pass
         elif chunk == self._boundary + b'--':
             self._at_eof = True
+            yield from self._readline()
         else:
             raise ValueError('Invalid boundary %r, expected %r'
                              % (chunk, self._boundary))
```

terencehonles:
Currently a multipart message epilogue is left unread which when nested
in another multipart message causes a ValueError to be raised because
the parent reader expects to find its multipart boundary not the
sub-message's epilogue.

fixes issue #1526 

## What do these changes do?

Allows reading multipart sub-messages that were created with `aiohttp.multipart.MultipartWriter`

ValueError for an incorrect boundary is not thrown anymore.

## Related issue number

#1526 

## Checklist

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -639,6 +639,7 @@ class MultipartReader(object):
             pass
         elif chunk == self._boundary + b'--':
             self._at_eof = True
+            yield from self._readline()
         else:
             raise ValueError('Invalid boundary %r, expected %r'
                              % (chunk, self._boundary))
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -594,6 +594,7 @@ class MultipartReaderTestCase(TestCase):
                    b'\r\n'
                    b'passed\r\n'
                    b'----:----\r\n'
+                   b'\r\n'
                    b'--:--'))
         yield from reader.release()
         self.assertTrue(reader.at_eof())
--- a/tests/test_py35/test_multipart_35.py
+++ b/tests/test_py35/test_multipart_35.py
@@ -46,6 +46,7 @@ async def test_async_for_reader(loop):
             b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03K\xcc\xcc\xcf())'
             b'\xe0\x02\x00\xd6\x90\xe2O\x08\x00\x00\x00',
             b'--::--',
+            b'',
             b'--:--',
             b''])))
     idata = iter(data)


fafhrd91:
@kxepal ^^^

kxepal:
@fafhrd91 I'm watching you, don't worry (: But thanks anyway!

kxepal:
From [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.7.2):

> Unlike in RFC 2046, the epilogue of any multipart message MUST be empty; HTTP applications MUST NOT transmit the epilogue (even if the original multipart contains an epilogue). These restrictions exist in order to preserve the self-delimiting nature of a multipart message- body, wherein the "end" of the message-body is indicated by the ending multipart boundary. 


So, technically, we need to decide what to do with a bad HTTP applications: throw and error or silently ignore their violations. There is also preambula case uncovered, but both preambula and epilogue originally [recommended to be ignored](https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html):

> NOTE: These "preamble" and "epilogue" areas are not used because of the lack of proper typing of these parts and the lack of clear semantics for handling these areas at gateways, particularly X.400 gateways. 


But assuming that issue was triggered by aiohttp client, I think we need to let it not do it in the first place (:

~Rereading the multipart spec, it says nothing about `\r\n` chars on closing~

~So imho, correct fix is to strip `\r\n` chars from [our closing-delimiter yield](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L969). Not sure why it was put there, much likely my copy-paste from regular delimiter yield.~ 

~And then we can decide what to do with epilogues: swallow or crash. Hopefully we have a room for choice because of this bug (:~

That was wrong assumption. From where is epilogue comes from? /o?

kxepal:
Ok, got it. The epilogue eventually happens after [part serialization](https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/multipart.py#L767). If part is another MultipartWriter, then it end with like `--boundary--\r\n\r\n` and we crash. Sounds like fix all boundaries yield to how spec defines them (e.g. `\r\n` in head, not in tail) is a good idea in any way:

```
diff --git a/aiohttp/multipart.py b/aiohttp/multipart.py
index af7f19b..4e87975 100644
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -963,9 +963,9 @@ class MultipartWriter(object):
             return
 
         for part in self.parts:
-            yield b'--' + self.boundary + b'\r\n'
+            yield b'\r\n--' + self.boundary + b'\r\n'
             yield from part.serialize()
         else:
-            yield b'--' + self.boundary + b'--\r\n'
+            yield b'\r\n--' + self.boundary + b'--'
 
         yield b''
```

We can't remove trailing `\r\n` from regular delimiter there we yield first part of encapsulation:
```
encapsulation := delimiter CRLF body-part 
delimiter := CRLF "--" boundary   ; taken from  Content-Type 
field. 
```

But now we can remove one extra newline from a body part serialization:
```
diff --git a/aiohttp/multipart.py b/aiohttp/multipart.py
index af7f19b..1155044 100644
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -764,7 +764,6 @@ class BodyPartWriter(object):
             )
         yield b'\r\n\r\n'
         yield from self._maybe_encode_stream(self._serialize_obj())
-        yield b'\r\n'
 
     def _serialize_obj(self):
         obj = self.obj
```

This looks right. This will also break all the tests which expects that body part serialization ends with newline, but it's about to replace those assertions with StopIteration expectation.

Need to test that idea anyway, but does that looks good for you, @terencehonles ?

terencehonles:
@kxepal that's what I was wondering about since it is slightly ambiguous. I _believe_ it means that the epilogue should be blank (which is what is currently being done) *not* that it should be non existent. When looking around on the web it did seem like most examples included a blank epilogue rather than placing the next boundary immediately following the nested multipart boundary end.

I do think the code may need to be lenient (not what it happening), but I wanted to start a conversation over this because I _do_ depend on this code (both the reader and writer)

terencehonles:
@kxepal I refrained from changing the writer because I was only seeing examples with blank lines for the epilogue not omitted epilogues. Do you have examples (other than the test case) which show otherwise?

kxepal:
@terencehonles 
The biggest problem of multipart examples is that you cannot be sure where newlines were added for formatting and where they are mandatory (and where they are CRLF, not just CR).

No, I don't have any good examples right-by-the-hand, need to search over and check the others implementations - how they do handle such cases.

Personally, I just feel bad about having additional blind readline because there you may eventually read not what you expect to read. Say, what if there will be no epilogue at all and the "blank" means nothing, not a "blank line" as you expect? 

So I would prefer to fix the writter at the first place as there we won't loose or break the data that we do not own. And only if this doesn't helps, try to fix the reader with extra caution. So far it's possible to do without any notable breaks.

Will return to this issue a bit later after few more investigations.

terencehonles:
@kxepal well I'm not advocating keeping it blind. I wanted to start the discussion, but I think we _should_ probably be handling the case when there is an epilogue. So by _just_ fixing the writer we are sweeping the problem under the rug. I'm fine with doing both, but I wasn't sure if the epilogue really should be omitted rather than blank (I'm still not convinced)

kxepal:
@terencehonles 
Yea, I got your point. And you're absolute right. There are actually two fixes: let writer not emit an epilogue (no any good reason to have it) and let reader handle epilogue right (by ignoring as everybody does) (and may be preambula as well with the same effect if we're going to fix these holes). Each can solve your problem, but both of them is good to have. 

I started from a writer side since it's simpler to fix. But this certainly doesn't makes epilogue problem go away, there just need a little bit more dirty work to do.

fafhrd91:
@terencehonles @kxepal what is status of this ticket?

terencehonles:
@fafhrd91 I don't have enough previous multipart experience to know what should be allowed and how strict the code should be. I'm not sure if @kxepal had any time to look into the issue any further as we were spending time on #1530, but now that it is closed I'd like to spend more time on this PR.

Not that it really matters, but I've kept my work unblocked by using this branch for my personal development, but I will need a proper solution by the time I push my work to production.

fafhrd91:
as i understood #1530 was ready and it is merged now.

@kxepal i need your assistance

kxepal:
That rogue readline really annoys me, but I don't have a time to propose anything better right now and it seems that it really solves the issue and it seems like it doesn't breaks anything. So let's merge it and leave room open for further improvements. If it'll break something, we would have a way to look into.

+1

fafhrd91:
@terencehonles could you rebase your branch against master and then re-create PR against master

terencehonles:
@fafhrd91 will do.

@kxepal I'll look at the code path this week and see if I can propose something nicer (I'm also not very much a fan of the blind readline with the limited documentation I've read, but I needed to propose something)

fafhrd91:
@terencehonles i merged pr, if you want to make more improvements let's move it to separate pr.

thanks!

terencehonles:
@fafhrd91 Ok, sorry I got a flu for the past couple days and didn't have a chance to update this.

@kxepal I looked and the code and there are two things we could do. I decided to just implement something and you can check out #1581

PR #1535, opened 2017-01-09
do not use readline when reading the content of a part in the multipart reader

arthurdarcet:
multipart uploads to an aiohttp server with a "large" file containing no new line fails with a `ValueError('Line is too long')`:

```python
import aiohttp
import aiohttp.web
import asyncio


async def view(request):
	async for part in reader:
		v = await part.read(decode=True)
		print(part, len(v))
	return aiohttp.web.Response(text="OK")

async def server(loop):
	app.router.add_post('/', view)
	handler = app.make_handler()
	await loop.create_server(handler, '0.0.0.0', 1111)

async def main(loop):
	srv = await server(loop)
	async with aiohttp.post('http://0.0.0.0:1111', data={'val':b'0' * (2 ** 16 - 100)}) as resp:
		print('--->', resp.status)
	print('--')
	async with aiohttp.post('http://0.0.0.0:1111', data={'val':b'0' * (2 ** 16 + 1)}) as resp:
		print('--->', resp.status)

```
```
<aiohttp.multipart.BodyPartReader object at 0x109c10f28> 65436
---> 200
--
Error handling request
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/web_server.py", line 61, in handle_request
    resp = yield from self._handler(request)
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/web.py", line 268, in _handle
  File "poc.py", line 9, in view
    v = await part.read(decode=True)
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/multipart.py", line 257, in read
    data.extend((yield from self.readline()))
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/multipart.py", line 351, in readline
    line = yield from self._content.readline()
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/streams.py", line 517, in wrapper
    result = yield from func(self, *args, **kw)
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/streams.py", line 587, in readline
    return (yield from super().readline())
  File "/Users/arthur/Documents/reaaad/lib/aiohttp/aiohttp/streams.py", line 251, in readline
    raise ValueError('Line is too long')
ValueError: Line is too long
---> 500
```

This is because the `StreamReader` class only accepts lines up to 64KB.
I've updated the `MultipartReader` class to avoid `readline` when it doesn't really make sense, and to use `read_chunk` instead.

As far as I can see, it seems to work, but that might not be the right approach to solve this at all…

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -252,12 +252,8 @@ class BodyPartReader(object):
         if self._at_eof:
             return b''
         data = bytearray()
-        if self._length is None:
-            while not self._at_eof:
-                data.extend((yield from self.readline()))
-        else:
-            while not self._at_eof:
-                data.extend((yield from self.read_chunk(self.chunk_size)))
+        while not self._at_eof:
+            data.extend((yield from self.read_chunk(self.chunk_size)))
         if decode:
             return self.decode(data)
         return data
@@ -377,12 +373,8 @@ class BodyPartReader(object):
         """
         if self._at_eof:
             return
-        if self._length is None:
-            while not self._at_eof:
-                yield from self.readline()
-        else:
-            while not self._at_eof:
-                yield from self.read_chunk(self.chunk_size)
+        while not self._at_eof:
+            yield from self.read_chunk(self.chunk_size)
 
     @asyncio.coroutine
     def text(self, *, encoding=None):
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -255,14 +255,13 @@ class PartReaderTestCase(TestCase):
         self.assertEqual(b'.' * size, result)
         self.assertTrue(obj.at_eof())
 
-    def test_read_does_reads_boundary(self):
+    def test_read_does_not_read_boundary(self):
         stream = Stream(b'Hello, world!\r\n--:')
         obj = aiohttp.multipart.BodyPartReader(
             self.boundary, {}, stream)
         result = yield from obj.read()
         self.assertEqual(b'Hello, world!', result)
-        self.assertEqual(b'', (yield from stream.read()))
-        self.assertEqual([b'--:'], list(obj._unread))
+        self.assertEqual(b'--:', (yield from stream.read()))
 
     def test_multiread(self):
         obj = aiohttp.multipart.BodyPartReader(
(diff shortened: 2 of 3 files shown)


fafhrd91:
@kxepal 
you are very popular :)

kxepal:
@arthurdarcet 
Good catch!

@fafhrd91 
Ahaha! Indeed. As ours multipart feature (:

aiohttp/multipart.py
```diff
@@ -252,12 +252,8 @@ def read(self, *, decode=False):
         if self._at_eof:
             return b''
         data = bytearray()
-        if self._length is None:
```
kxepal, line 255: Nice!

tests/test_multipart.py
```diff
@@ -501,6 +499,17 @@ def test_filename(self):
             None)
         self.assertEqual('foo.html', part.filename)
 
+    def test_reading_long_part(self):
+        # should stay larger than aiohttp.streams.DEFAULT_LIMIT
+        size = 2 ** 18
```
kxepal: Can this value be picked from aiohttp.streams.DEFAULT_LIMIT to be in sync?

tests/test_multipart.py
```diff
@@ -255,14 +256,13 @@ def test_read_chunk_properly_counts_read_bytes(self):
         self.assertEqual(b'.' * size, result)
         self.assertTrue(obj.at_eof())
 
-    def test_read_does_reads_boundary(self):
+    def test_read_does_not_read_boundary(self):
```
kxepal, line 260: Well, actually test reads the boundary (: In previous case it just was placed into unread buffer.

tests/test_multipart.py
arthurdarcet: sure, i've updated the commit

tests/test_multipart.py
arthurdarcet: I'm not entirely sure, but i think it reads the boundary then put in back in the stream (with `unread_data`) ; so it's as if it wasn't read

fafhrd91:
what is status of this ticket?

fafhrd91:
@kxepal ?

kxepal:
@fafhrd91 Sorry, a bit busy times. I have all the tickets you ping me stared in by inbox, will check them in few hours.

fafhrd91:
thanks!

kxepal:
+1

PR #1542, opened 2017-01-16
Accept an errors parameter for text decoding ClientResponses

bmuller:
## What do these changes do?

Users can now give a parameter specifying what should happen to decoding errors when asking for the `text()` from a `ClientResponse`.  This is passed directly into the [bytes() decode method](https://docs.python.org/3/library/stdtypes.html#bytearray.decode).

The internet is full of websites that have incorrectly specified their character encoding or that have one or two characters on a page that are not encoded correctly.  Right now, the `aiohttp` library just throws an exception when this happens - for a more resilient solution, users should be able to optionally provide an alternative error handling strategy for incorrectly encoded characters.

Nope - just a new optional parameter.

## Related issue number

Nope.

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,8 +12,7 @@ CHANGES
 
 - Fix typo in FAQ section "How to programmatically close websocket server-side?".
 
-- Allow users to specify what should happen to decoding errors when calling a responses `text()` method
-
+- Allow users to specify what should happen to decoding errors when calling a responses `text()` method #1542
  
 1.2.0 (2016-12-17)
 ------------------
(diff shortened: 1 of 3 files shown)


fafhrd91:
Could you add test for this change?

bmuller:
@fafhrd91 sure - what would you like to be tested though?  This is a pretty inconsequential change that just passes another optional parameter down to Python std lib call.

fafhrd91:
test that `errors` param get passed to decode function, also could you use `errors` keyword argument for decode function.

bmuller:
OK - let me know if there's anything else you need @fafhrd91

fafhrd91:
great! thanks

Issue #1211, opened 2016-09-25
RuntimeError: Cannot pause_reading() when closing, triggered via FlowControlStreamReader.feed_data


zmedico:
## Long story short

When trying to read the response to a get request from an https endpoint, I get a `RuntimeError: Cannot pause_reading() when closing` exception.
## Expected behaviour

I expect it to read the response successfully.
## Actual behaviour

```
  File "/usr/lib64/python3.5/site-packages/aiohttp/client_reqrep.py", line 739, in read
    self._content = yield from self.content.read()
  File "/usr/lib64/python3.5/site-packages/aiohttp/streams.py", line 507, in wrapper
    result = yield from func(self, *args, **kw)
  File "/usr/lib64/python3.5/site-packages/aiohttp/streams.py", line 562, in read
    return (yield from super().read(n))
  File "/usr/lib64/python3.5/site-packages/aiohttp/streams.py", line 265, in read
    raise self._exception
  File "/usr/lib64/python3.5/site-packages/aiohttp/parsers.py", line 190, in set_parser
    next(p)
  File "/usr/lib64/python3.5/site-packages/aiohttp/protocol.py", line 313, in __call__
    yield from self.parse_length_payload(out, buf, length)
  File "/usr/lib64/python3.5/site-packages/aiohttp/protocol.py", line 359, in parse_length_payload
    out.feed_data(chunk, len(chunk))
  File "/usr/lib64/python3.5/site-packages/aiohttp/streams.py", line 554, in feed_data
    self._stream.transport.pause_reading()
  File "/usr/lib64/python3.5/asyncio/sslproto.py", line 336, in pause_reading
    self._ssl_protocol._transport.pause_reading()
  File "/usr/lib64/python3.5/asyncio/selector_events.py", line 647, in pause_reading
    raise RuntimeError('Cannot pause_reading() when closing')
RuntimeError: Cannot pause_reading() when closing
```
## Steps to reproduce

The relevant code is something like this, though I've only been able to reproduce it in an application which creates many sessions simultaneously:

```
async def ssl_fetch(url):
    with aiohttp.ClientSession(
        connector=aiohttp.TCPConnector(
        ssl_context=ssl._create_unverified_context())) as session:

        resp = await session.get(url)
        content = await resp.read()
        await resp.release()
        return content
```
## Your environment

[...]

PR #1551, opened 2017-01-23
Do not pause transport during set_parser stage

fafhrd91:
Added `allow_pause` to flow control related streams, so we can prevent transport pausing during `set_parser` stage. this should solve #1211 

@asvetlov @kxepal i need your review and thoughts

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,11 +10,14 @@ CHANGES
 
 - Fix sub-Multipart messages missing their headers on serialization #1525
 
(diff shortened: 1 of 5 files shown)


fafhrd91:
i think `pause_reading` should be no-op during transport closing.

asvetlov:
I think `allow_pause` should be a private attribute (`_allow_pause`).
Otherwise we need to document it.

fafhrd91:
good point, will change

PR #1552, opened 2017-01-23
Having a `:` or `@` in a route does not work

arthurdarcet:
The URL escaping done for the request urls and in the router when adding resources should be consistent.
Currently, `:` and `@` are considered "safe" by yarl and not escaped in `request.rel_url.raw_path` ; but the router does escape them and so the created resource never match:

```python
import aiohttp
import aiohttp.web
import asyncio


async def server(loop):
	app.router.add_get('/a-b', lambda _: aiohttp.web.Response(text="OK"))
	app.router.add_get('/a:b', lambda _: aiohttp.web.Response(text="nope"))
	app.router.add_get('/a%3Ab', lambda _: aiohttp.web.Response(text="still nope"))
	handler = app.make_handler()
	await loop.create_server(handler, '0.0.0.0', 8080)

async def main(loop):
	srv = await server(loop)
	async with aiohttp.get('http://0.0.0.0:8080/a-b') as resp:
		print('--->', resp.status)
	async with aiohttp.get('http://0.0.0.0:8080/a:b') as resp:
		print('--->', resp.status)
	async with aiohttp.get('http://0.0.0.0:8080/a%3Ab') as resp:
		print('--->', resp.status)

```
Output:
```
% python a.py
---> 200
---> 404
---> 404
```

I have updated the router to avoid using `yarl.quote` and use `yarl.URL(…).raw_path` instead, ensuring the escaping is consistent in both cases. (and I added a test that fails without this patch)

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -11,7 +11,10 @@ from collections.abc import Container, Iterable, Sized
 from pathlib import Path
 from types import MappingProxyType
 
-from yarl import URL, quote, unquote
+# do not use yarl.quote directly,
+# use `URL(path).raw_path` instead of `quote(path)`
+# Escaping of the URLs need to be consitent with the escaping done by yarl
+from yarl import URL, unquote
 
 from . import hdrs
 from .abc import AbstractMatchInfo, AbstractRouter, AbstractView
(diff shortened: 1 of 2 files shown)


fafhrd91:
thanks!

PR #1556, opened 2017-01-25
fix #1555 AioHTTPTestCase method name mistype - get_applicaion

### LINKED ISSUE #1555: AioHTTPTestCase method name mistype - get_applicaion instead of get_application
## Long story short

missed letter in word looks confusing and misleading

## Expected behaviour

method to be called get_application

## Actual behaviour

it is called get_applicaion

## Steps to reproduce

check commit by Andrew: https://github.com/KeepSafe/aiohttp/commit/8bfabd7ab25f85def9124e8a87d5681b14234afb

## Your environment
just mistype

jazz-jack:
## What do these changes do?

fix mistype in method name

yes, it changes AioHTTPTestCase public method name

## Related issue number

#1555 

## Checklist
it is just a typo fix

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -382,7 +382,8 @@ class AioHTTPTestCase(unittest.TestCase):
 
     def setUp(self):
         self.loop = setup_test_loop()
-        self.app = self.loop.run_until_complete(self.get_application(self.loop))
+        self.app = self.loop.run_until_complete(
+            self.get_application(self.loop))
         self.client = TestClient(self.app)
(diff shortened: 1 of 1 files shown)


fafhrd91:
Good catch! Could you add old name and add deprecation warning

jazz-jack:
i doubt anybody used it, the issue is quite obvious
1-month old bug, not released...
i'd prefer to keep sources cleaner and just leave it

fafhrd91:
good

fafhrd91:
thanks!

PR #1563, opened 2017-01-29
Remove obsolete remark about `BaseConnector.limit` defaulting to `None`

ExplodingCabbage:
This hasn't been true since `limit` was given a default of 20 in the `BaseConnector.__init__`.


fafhrd91:
thanks!

PR #1578, opened 2017-02-02
Ensure same connection is reused during client keep-alive tests

JustinTArthur:
Forces reuse of the same connection in client keep-alive tests by setting pool limit to 1.

I'm working on pipelining options for connectors and this helps provide a clean baseline. Added myself to contributors.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -83,6 +83,7 @@ Joongi Kim
 Julien Duponchelle
 Junjie Tao
 Justas Trimailovas
+Justin Turner Arthur
 Kay Zheng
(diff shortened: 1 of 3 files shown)


fafhrd91:
there is no context for this change, and intention is not clear. if you want to work on something just create new set of tests

JustinTArthur:
@fafhrd91, the tests as they are now are intended to test keep-alive on a connection. However, the way they are written, there is no guarantee that the same connection is being used for subsequent requests.

Keep-alive functionality might regress and the tests could continue to pass even if the subsequent request(s) work only because a second connection was established. This is the context of my submission.

fafhrd91:
thanks for explanation!

PR #1589, opened 2017-02-06
Indicate in the docs that response history contains the earliest request first, rather than the most recent

ExplodingCabbage:
Either was plausible, and I'd wrongly assumed the latter. Seems worth noting this explicitly in the docs.


fafhrd91:
Thanks!

PR #1608, opened 2017-02-09
run_app and CLI can listen on Unix domain sockets

JustinTArthur:
## What do these changes do?
Makes it easy to serve HTTP or HTTPS requests over Unix domain sockets using pre-existing deployment features. Accomplishes this by:
* Adding a `path=` keyword argument to the `run_app` utility function that can be used to supply a file system path for a Unix domain socket.
* Adding an optional `-U` or `--path` command line argument to the CLI web server that can be used to supply a file system path for a Unix domain socket.

Users serving HTTP or HTTPS over TCP/IP should see no change in behavior.

## Needs Review Before Merge
* Is the term _path_ a good one to use in our APIs? It mimics the `path` parameter passed to asyncio's implementation and is small. Is it intuitive enough?
* CPython <3.6 fails to remove a stale unix socket path from the file system before attempting to bind again, causing an error if a stale path exists (see python/asyncio#425). Do we want to work around this for our 3.4 and 3.5 users and delete the stale path before calling `create_unix_server`?
* I had to choose a URI or URL format to use for portraying the HTTP-over-Unix-domain-socket resource. In this version the text for humans that gets spit out by `run_app` uses [the same format as nginx's proxy_pass directive](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass), e.g. `http://unix:/tmp/backend.sock:`. Some other possible locator or identifier formats we could use:
   * `http+unix://%2Ftmp%2Fbackend.sock` as used by [requests-unix](https://github.com/msabramo/requests-unixsocket).
  * Place a dummy host as used by [curl](https://curl.haxx.se/)'s `curl` tool: `http://localhost/`
    * It does this partly because its Unix domain socket is supplied separately.
  * `unix:/tmp/backend` as used in nginx's `upstream` `server` directive.
    * It won't portray whether the socket is using SSL or not.

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -1,4 +1,5 @@
 import asyncio
+import socket
 import sys
 import warnings
 from argparse import ArgumentParser
@@ -269,12 +270,12 @@ class Application(MutableMapping):
         return "<Application 0x{:x}>".format(id(self))
 
 
-def run_app(app, *, host='0.0.0.0', port=None,
+def run_app(app, *, host='0.0.0.0', port=None, path=None,
             shutdown_timeout=60.0, ssl_context=None,
             print=print, backlog=128, access_log_format=None,
             access_log=access_logger):
     """Run an app locally"""
-    if port is None:
+    if port is None and path is None:
         if not ssl_context:
             port = 8080
         else:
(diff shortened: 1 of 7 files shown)


aiohttp/web.py
```diff
@@ -289,16 +290,30 @@ def run_app(app, *, host='0.0.0.0', port=None,
...
+        srv_creation = loop.create_unix_server(
+            handler, path, ssl=ssl_context, backlog=backlog
+        )
+        # nginx-style URLs for Unix socket paths
+        if isinstance(path, str):
+            urls = ('{}://unix:{}:'.format(scheme, path),)
+        else:
+            urls = ['{}://unix:{}:'.format(scheme, p) for p in path]
+
+    srv = loop.run_until_complete(srv_creation)
```
fafhrd91: i think you should be able to run both unix and tcp servers at the same time

docs/deployment.rst
```diff
@@ -96,23 +96,24 @@ Next we need to configure *aiohttp upstream group*:
 
    http {
      upstream aiohttp {
-       fail_timeout=0 means we always retry an upstream even if it failed
+       # fail_timeout=0 means we always retry an upstream even if it failed
```
fafhrd91, line 99: may be there should be two sections one for tcp server and for unix socket server

aiohttp/web.py
JustinTArthur: Doesn't sound bad. I will have to consider the impact on the parameters (for example, when assuming a host of '0.0.0.0')

docs/deployment.rst
JustinTArthur: That would be easy enough to do. I can commit a version like this and we'll evaluate if it's too verbose.

fafhrd91: sounds good

aiohttp/web.py
fafhrd91: les start tcp server only if host is explicitly provide

JustinTArthur:
I have now…
* Added stale socket file cleanup on start for the older event loop implementations that don't do this.
* Added support for passing a bytes-like object for the host(s) or path(s).
* Added tests to validate Linux abstract path support.
  * The URI that gets printed may not be human-usable in this current version.
* Made it so you can supply as many hosts or paths as you want at the same time per @fafhrd91's recommendation..
  * If none are supplied or only a port is supplied, the default behavior of serving on 0.0.0.0 occurs just as has for a long time.
* Documented code snippets for both TCP sockets and Unix domain sockets in the deployment doc per @fafhrd91's recommendation.
* Got code coverage for the web module back up to par.

@fafhrd91 how does it look?

fafhrd91:
Perfect! Thanks!

PR #1616, opened 2017-02-10
http pipelining

fafhrd91:
@asvetlov @kxepal

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -225,8 +225,9 @@ class ClientSession:
                         resp.close()
                         conn.close()
                         raise
-                except (aiohttp.HttpProcessingError,
-                        aiohttp.ServerDisconnectedError) as exc:
+                except aiohttp.ServerDisconnectedError:
+                    raise
+                except aiohttp.HttpProcessingError as exc:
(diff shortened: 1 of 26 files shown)


kxepal:
Wow! That's great! Will be happy to test this.

fafhrd91:
It doesn't give any increase in plain requests. It may give some increase if you do extra io in handlers. This shows that slow part is not in protocol write track, it is probably whole aiohttp.web stack is slightly slow

fafhrd91:
so on "ok" responses pipelining is ~5% slower than non-pipelining mode.
i think i can speed it up by refactoring request parsing.

fafhrd91:
merged to master

fafhrd91:
I did some more optimizations, now it starting look better

Env: uvloop, -OO, no logs

1.2 ~7.4k rps
1.4a ~9.8k rps
1.4a pipelining ~11.7k rps

kxepal:
Meanwhile it seems that HTTP/1 Pipelining is going to be deprecated: https://bugzilla.mozilla.org/show_bug.cgi?id=1340655

PR #1621, opened 2017-02-11
remove gc.collect() from teardown_test_loop

samuelcolvin:
Was there a good reason to call `gc.collect()` while tearing down a loop in tests?

In the particular code I'm working on atm. removing that line roughly halves the time taken to run tests.

If you really think it's necessary perhaps use `gc.collect(1)` which is less aggressive and seems to still yield the 2x speed increase.

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -3,7 +3,6 @@
 import asyncio
 import contextlib
 import functools
-import gc
 import socket
 import sys
 import unittest
@@ -457,7 +456,6 @@ def teardown_test_loop(loop):
         loop.call_soon(loop.stop)
         loop.run_forever()
         loop.close()
-    gc.collect()
     asyncio.set_event_loop(None)
 
 


samuelcolvin:
The failure is spurious.

From what I can tell from travis; time taken for the main test suit has has gone from ~1min to ~25secs with this change.

fafhrd91:
it is easy to forget about un-completed future or open socket. gc.collect() helps with __del__ method.
you can not just remove it, but maybe we should introduce some special mode that would increase tests speed, and `full` mode for CI

samuelcolvin:
I'm not convinced that `gc.collect()` helps with pending futures, from my experience `gc.collect()` is both very slow and doesn't help much with pending tasks/futures.

It would be good if there was some way to check for pending futures at the end of a test and throw an error if any exist, but that's a different issue.

**edit:** would `asyncio.Task.all_tasks()` allow the teardown method to check for incomplete futures?

jettify:
I think `gc.collect()` more about making sure that any `__del__` logic triggered, for instance unclosed sockets. This is reliable way to tests things like https://github.com/KeepSafe/aiohttp/blob/b521674037d17ecc7849040ae12075a3884ec87b/aiohttp/client.py#L112-L122 also helps surface warnings after specific test.

fafhrd91:
I am fine with introducing `light` mode

fafhrd91:
added `AIOHTTP_TESTS_LIGHT` environ var

fafhrd91:
in 6c097c96a7b60f26f846936ec993fddf4da1d7d3

PR #1622, opened 2017-02-12
Cull some completely pointless 'sentinel' logic

ExplodingCabbage:
Here `TCPConnector.__init__` takes a `sentinel` object as its default
argument to `use_dns_cache`, but then sets `self._use_dns_cache` to True
if the sentinel is given (and to the given value otherwise). That
logic is identical to scrapping the sentinel-specific logic and
just setting `self._use_dns_cache` to the given value, which is simpler.

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -477,7 +477,7 @@ class TCPConnector(BaseConnector):
     """
 
     def __init__(self, *, verify_ssl=True, fingerprint=None,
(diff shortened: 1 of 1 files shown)


ExplodingCabbage:
Note also that the sentinel doesn't get explicitly passed in as a value of `use_dns_cache` by any code within `aiohttp` that instantiates a `TCPConnector`.

fafhrd91:
thanks

Issue #1632, opened 2017-02-15
Explicitly require yarl versions <= 0.9.3


Gallaecio:
## Long story short

The example code at http://aiohttp.readthedocs.io/en/stable/client_reference.html fails with yarl > 0.9.3 due to https://github.com/aio-libs/yarl/issues/55

## Expected behaviour

The aiohttp package, that installed using pip, should prevent the installation of any yarl version with this issue, so far 0.9.4 and 0.9.5.1. Otherwise, 'pip install aiohttp' installs a broken aiohttp.

fafhrd91:
Fixed with yarl 0.9.6

PR #1634, opened 2017-02-15
Fix URL instance comparsion

insideable:
## What do these changes do?

yarl 0.9.5 introduces BaseURL class, so isinstance(foo, URL) doesn't work anymore.

No

## Related issue number

#1632 #1631 #1635


insideable:
Build failed only on Linux/Python 3.4.2, and I have no idea why this happens.
https://travis-ci.org/KeepSafe/aiohttp/builds/201841920

kxepal:
@insideable 
That was well known flaky static test. Rerun fixed that.

fafhrd91:
This should be fixed with yarl0.9.6

PR #1643, opened 2017-02-16
Fix multiple ws_connect with headers dict

armills:
## What do these changes do?

Currently in my client, I'm generating a headers dictionary that is used for all websockets connections. What I found is that when ws_connect is called, the websocket key is added to this dict, which is then passed back on the next call to ws_connect. Since the default headers don't overwrite the headers dict, the stale secret key is sent to the server, and the connection fails.

This change forces the secret key to overwrite the existing headers, since it is (and should) be regenerated for each connection. We might also want to consider copying the incoming dict before making changes, but that's another issue.

No

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 CHANGES
 =======
 
+- Fix multiple calls to client ws_connect when using a shared header dict #1643
+
 1.3.1 (2017-02-09)
 ------------------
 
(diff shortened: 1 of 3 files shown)


fafhrd91:
thanks!

fafhrd91:
thats fine. i'll fix

emlove:
#1644

Issue #1651, opened 2017-02-17
request.files in addition to request.data


ilkarman:
It seems the only way to add files to a request with aiohttp is by referencing them in the data parameter. However this becomes problematic when you have an API that expects to see something in 'files', not 'data.

For example, this works (using requests):
```
files = {'imagefile': open(fname, 'rb')}
requests.post(url, files=files)
```
However this doesnt:
```
files = aiohttp.FormData()
files.add_field('imagefile', open(fname, 'rb'))
response = yield from s.request('post', url, data=files)
```

AraHaan:
So, what you need is an ``files=`` parameter for ``request``?

AraHaan:
I created an PR that adds that parameter however I have not implemented the usage of it fully yet until I find out more on how requests works on it.

AraHaan:
I have void the github repo for the requests package I am looking into how they implemented their ``files=`` on it to hopefully add it into aiohttp.

AraHaan:
Ok, @iksteen I am working on it to where this would work for you (or at least something of the similar):

```py
files = {'imagefile': open(fname, 'rb')}
response = yield from s.request('post', url, files=files)
```

fafhrd91:
i do not think `files=` is good idea. 

maybe something like ```aiohttp.request(data=FormData(File('file.name'), File(fileObject, filename=''), Stream(streamReader)))``` etc

AraHaan:
@ilkarman doe's @fafhrd91 comment help fix this issue?

asvetlov:
IIRC we had `files` param but it was too ambiguous.
Let's think about API changing very carefully.
Personally I feel a need for sleeping on it before next discussion iteration.

fafhrd91:
I agree, we need to think about this api carefully. I envision some `UploadItem` interface that we can use as `data` value or as argument to `FormData` which in turn can implement `UploadItem` interface. it has to support sizable/unsizable stream types. also we should be able to use it in `aiohttp.web` as well.

fafhrd91:
better name is `PayloadItem`

fafhrd91:
@asvetlov let's move conversation to #1172

PR #1653, opened 2017-02-17
[Fix] #1651 - request.files in addition to request.data

AraHaan:
This is a starter commit that adds ``files=`` to ``ClientSession()._request`` to allow it to be used in ``ClientSession().request``. This should hopefully make things that works in the requests package also work in aiohttp. I just need to look into the actual way to make it really work for when an API expects to see something in 'files', but not in 'data'.

## What do these changes do?

They allow someone to tell request to use files and not data when an aip does not check for files in 'data' but does in 'files'.

No changes other than they can tell request files instead of data to make requests work on uploading files to servers.

## Related issue number

#1651 

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -134,6 +134,9 @@ class ClientSession:
     def _request(self, method, url, *,
                  params=None,
                  data=None,
+                 # TODO: make this actually work for files.
+                 # see issue  for more info. #1651
+                 files=None,
                  headers=None,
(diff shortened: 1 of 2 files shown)


AraHaan:
@fafhrd91 mind helping me get this to work based on the requests package's ``files=`` param. Currently they way they have it in their implementation is so confusing.

fafhrd91:
i am generally against `files` param. we already have some much parameters

PR #1662, opened 2017-02-19
Fix static location in index when prefix is used

lovebug356:
When the prefix is used (e.g. not /), the links generated in the
index page are wrong. In this patch the relative path is correctly
calculated by the pathlib module and a absolute url is used as link
in the index page.

The original bug can be demonstrated with following example:

```
import pathlib

from aiohttp import web

app = web.Application()
app.router.add_static('/static', pathlib.Path(__file__).parent,
show_index=True)

web.run_app(app)
```


fafhrd91:
@asvetlov could you review this PR

AraHaan:
Well looks good other than the posix part because it would probably jack with running aiohttp on windows?

Like I am not sure how it would interfere with how the ``\``'s are on Windows however it *should* support both ``/`` and ``\``.

AraHaan:
Mind rebasing this PR, or reopen it?

fafhrd91:
could you open PR against https://github.com/aio-libs/aiohttp

AraHaan:
Seems that windows also accepts posix paths to an extent as well.

asvetlov:
Thanks!

PR #1664, opened 2017-02-20
Payload refactoring

fafhrd91:
Payload refactoring


fafhrd91:
some goals

1. unify how we handle various types of payloads
2. unify client/server implementation
3. it needs to be backward compatible with existing clients

note: this is only about how we send payload

fafhrd91:
added `streamer` decorator for coroutine support as data provider

for usage check `test_client_functional.py::test_POST_STREAM_DATA`

fafhrd91:
@kxepal is this `content-disposition` right? `inline; filename="test.txt"; filename*=utf-8''test.txt`

fafhrd91:
@kxepal I need you to review MultipartWriter related changes specifically

aiohttp/client_reqrep.py
```diff
@@ -219,86 +217,54 @@ def update_auth(self, auth):
...
+        if isinstance(body, FormData):
+            body = body(self.encoding)
 
-        else:
-            if not isinstance(data, helpers.FormData):
-                data = helpers.FormData(data)
+        try:
+            body = payload.PAYLOAD_REGISTRY.get(body)
+        except payload.LookupError:
+            body = FormData(body)(self.encoding)
```
kxepal, line 244: Why not to send data as is?

aiohttp/multipart.py
```diff
@@ -938,53 +682,191 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
     def append_json(self, obj, headers=None):
         """Helper to append JSON part."""
-        if not headers:
+        if headers is None:
             headers = CIMultiDict()
-        headers[CONTENT_TYPE] = 'application/json'
-        return self.append(obj, headers)
+
+        *_, params = parse_mimetype(headers.get(CONTENT_TYPE))
+        charset = params.get('charset', 'utf-8')
```
kxepal, line 750: I think this is a bit redundant and may confuse. Because you can throw here any content-type, but in the end it will be `application/json` without any charset and else params. 

Better have explicit charset function argument or drop this idea. We're in 2017 anyway and Python`s json escapes all non ascii chars, so it doesn't really matter which encoding will be applied.

aiohttp/payload.py
```diff
@@ -0,0 +1,267 @@
...
+    _headers = None
+    _content_type = 'application/octet-stream'
+
+    def __init__(self, value, *, headers=None,
+                 content_type=sentinel, filename=None, encoding=None):
+        self._value = value
+        self._encoding = encoding
+        self._filename = filename
+        if headers is not None:
+            self._headers = CIMultiDict(headers)
```
kxepal, line 62: Shouldn't here be a multidict proxy? Because, technically, I can update `Payload.headers` content type and it will cause a conflict with the content type property.

kxepal:
I'll take a second look at the evening if it wouldn't be too late (: 
Massive changes, but I like the way it goes.

aiohttp/client_reqrep.py
fafhrd91: body could be dict or list or whatever. I am fine with changing FormData api

aiohttp/multipart.py
fafhrd91: I don't care much about this, we can remove it

kxepal: +1

aiohttp/client_reqrep.py
kxepal: Yes, I was though to just throw an "encode me" error in this case.

fafhrd91: first i try to convert `body` into Payload, then if it fails I try to use FormData
when you do `FormData(body)(self.encoding)` it returns bytes for url encoded or 
MultipartWriter (which is Payload) or fails

aiohttp/client_reqrep.py
```diff
@@ -219,86 +217,54 @@ def update_auth(self, auth):
...
-                mime = 'application/octet-stream' if mime is None else mime
-                self.headers[hdrs.CONTENT_TYPE] = mime
-
-        elif isinstance(data, MultipartWriter):
-            self.body = data.serialize()
-            self.headers.update(data.headers)
-            self.chunked = True
+        # FormData
+        if isinstance(body, FormData):
+            body = body(self.encoding)
```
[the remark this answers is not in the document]
fafhrd91: ?  I dont understand

aiohttp/abc.py
```diff
@@ -129,3 +129,20 @@ def update_cookies(self, cookies, response_url=None):
...
+    def write(self, chunk):
+        """Write chunk into stream"""
+
+    @asyncio.coroutine  # pragma: no branch
+    @abstractmethod
+    def write_eof(self, chunk=b''):
+        """Write last chunk"""
+
+    @asyncio.coroutine  # pragma: no branch
+    @asyncio.coroutine
```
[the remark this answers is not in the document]
fafhrd91: fixed

aiohttp/abc.py
```diff
@@ -129,3 +129,20 @@ def update_cookies(self, cookies, response_url=None):
...
+
+
+class AbstractPayloadWriter(ABC):
+
+    @abstractmethod
+    def write(self, chunk):
+        """Write chunk into stream"""
+
+    @asyncio.coroutine  # pragma: no branch
+    @abstractmethod
```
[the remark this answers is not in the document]
fafhrd91: good

aiohttp/payload.py
fafhrd91: yes, maybe. this api is not settled yet.

fafhrd91:
accidentally removed branch, now GitHub does not allow reopen PR.

@kxepal let's continue here, in any case I am mostly done with refactoring.

aiohttp/client_reqrep.py
kxepal: Sorry, I should be more clear. I mean that just try to send data as is without implicit transformations and fail if that's not possible (dict, list, custom object that didn't implement required interface). Implicit transformation may play a bad joke.

aiohttp/client_reqrep.py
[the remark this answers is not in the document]
kxepal: I mean that what if body could be an arbitrary callable object? FormData uses that interface to be processed. We can reuse that way for everything else. Though not sure if this would be helpful.

aiohttp/client_reqrep.py
fafhrd91: I want to be explicit, it may be not clear if it fails on actual write

aiohttp/client_reqrep.py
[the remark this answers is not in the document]
fafhrd91: We can add this later, right now I do not see value in this. Developer should be able to prepare spendable data before calling client. Case with custom data generation from coroutine already covered with `streamer` decorator

aiohttp/client_reqrep.py
kxepal: Ok.

aiohttp/client_reqrep.py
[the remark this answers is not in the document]
kxepal: Fair enough. Thanks!

fafhrd91:
@asvetlov comments?

Issue #1618, opened 2017-02-10
allow HEAD requests to default routes


samuelcolvin:
Just checked and both flask and django by default allow `HEAD` requests (but not `POST`, `PUT` etc.) to the default routed endpoints.

The closest equivalent in aiohttp is `add_get` but understandably that doesn't allow `HEAD` requests and returns 405.

aiohttp should add a method to add routes which accept the "noop" or safe http methods eg. `GET`, `HEAD` and `OPTIONS`.

To be consistent with other servers and act in the spirit of http, this method should be the default. AFAIK returning 405 to `HEAD` requests to `GET` endpoints is wrong.

I'm trying to think about the best name for such a method which is both clear and not verbose, it's not obvious. [Wikipedia](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol) calls these "safe methods" but `add_safe` could be a confusing; making people think aiohttp is giving some kind of protection which it isn't.

Perhaps the simplest options would be just  **`add`**. 

So code would be

```python
app.router.add('/', handler)
app.router.add_post('/post', post_handler)
app.router.add_put('/put', put_handler)
```

What exactly it should do with `OPTIONS` requests is debatable. The fact it should gracefully accept `HEAD` requests is a no-brainer for me.

samuelcolvin:
bump, any news one this? I think it's important and requires a fix.

fafhrd91:
i am not entirely sure what you are proposing?
Default methods or way how to add default methods?

samuelcolvin:
I'm proposing a new method `add` which allows HEAD requests as well as GET requests. 

Then change in documentation to encourage people to use that method instead of `add_get` with a comment saying "If you really want views which do not allow HEAD requests (which is contrary to the spirit of http) you can use the `add_get` method."

samuelcolvin:
Perhaps easiest if I do a pull request then you can comment on that.

kxepal:
Shouldn't allow GET also allow HEAD?

samuelcolvin:
yes it should, but I do understand why it's confusing if `add_get()` allowed `HEAD` so we need another method but `add_get_or_head()` would be annoying so I think just `add()` would work.

I'm easy, could just modify `add_get()`.

fafhrd91:
i think i'm missing something. you can do `add_route('HEAD', ...)`

samuelcolvin:
```python
app.router.add_get('/', index, name='index')
[...]

PR #1668, opened 2017-02-21
add allow_head to add_get

samuelcolvin:
## What do these changes do?

Add the `allow_head` keyword argument for `add_get` to encourage aiohttp server applications to correctly implement http and allow `HEAD` requests to endpoints generally used for `GET` requests. 

Also adds `app.router.set_defaults(allow_head=True)` to switch behaviour of all endpoints for the app.

## Related issue number

#1618

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -715,6 +715,7 @@ class UrlDispatcher(AbstractRouter, collections.abc.Mapping):
         super().__init__()
         self._resources = []
         self._named_resources = {}
+        self._default_allow_head = False
 
     @asyncio.coroutine
     def resolve(self, request):
@@ -754,6 +755,9 @@ class UrlDispatcher(AbstractRouter, collections.abc.Mapping):
 
     def named_resources(self):
         return MappingProxyType(self._named_resources)
+
+    def set_defaults(self, *, allow_head):
+        self._default_allow_head = allow_head
 
     def register_resource(self, resource):
         assert isinstance(resource, AbstractResource), \
@@ -855,11 +859,16 @@ class UrlDispatcher(AbstractRouter, collections.abc.Mapping):
         """
         return self.add_route(hdrs.METH_HEAD, *args, **kwargs)
 
-    def add_get(self, *args, **kwargs):
+    def add_get(self, *args, allow_head=None, name=None, **kwargs):
         """
-        Shortcut for add_route with method GET
+        Shortcut for add_route with method GET, if allow_head is true another
+        route is added allowing head requests to the same endpoint
         """
-        return self.add_route(hdrs.METH_GET, *args, **kwargs)
+        if allow_head or (allow_head is None and self._default_allow_head):
+            # the head route can't have "name" set or it would conflict with
+            # the GET route below
+            self.add_route(hdrs.METH_HEAD, *args, **kwargs)
+        return self.add_route(hdrs.METH_GET, *args, name=name, **kwargs)
 
     def add_post(self, *args, **kwargs):
         """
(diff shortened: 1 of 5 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -755,6 +756,9 @@ def routes(self):
     def named_resources(self):
         return MappingProxyType(self._named_resources)
 
+    def set_defaults(self, *, allow_head):
```
asvetlov: I doubt if we need default for the flag.
It might lead to many confusing cases, e.g. if third party library has something like `register_routes(router, prefix)` method its behavior varies depending on outer flag.

samuelcolvin: Humm, ok I'll remove it if you prefer. But the current setup is wrong according to the RFC and without the option to set the default you have to explicitly set `allow_head=True` on every route you add:

[RFC 7231](https://tools.ietf.org/html/rfc7231#page-25):

> The server SHOULD send the same
   header fields in response to a HEAD request as it would have sent if
   the request had been a GET, except that the payload header fields
   (Section 3.3) MAY be omitted.

For me, the simplest, clearest and most correct option would be to have `allow_head=True` by default.

samuelcolvin:
~~This is failing due to an error in docs I've fixed in #1669~~ fixed

aiohttp/web_urldispatcher.py
fafhrd91: problem is, we currently do not control when handler writes body, so in most cases
developer needs to support `head` explicitly

samuelcolvin: The developer can choose to support head explicitly but nothing breaks if they don't.

Allowing HEAD to endpoints which act like they got a GET request and still write body doesn't break anything.

Both django and flask allow users to still (by default) write body even for HEAD requests. I'm not saying they're always right, but their solution is more correct than aiohttp currently.

Here's an example:

**app.py:**
```python
import asyncio
from aiohttp import web

async def handle(request):
    print('preparing slow response...')
    await asyncio.sleep(2, loop=request.app.loop)
    return web.Response(body=b'this is the body\n')

app = web.Application()
app.router.add_get('/', handle, allow_head=True)

web.run_app(app, host='localhost', port=8000)
```

**shell:**
```
~ ➤  curl -v http://localhost:8000
* Rebuilt URL to: http://localhost:8000/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET / HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 17
< Content-Type: application/octet-stream
< Date: Wed, 22 Feb 2017 10:08:03 GMT
< Server: Python/3.5 aiohttp/2.0.0a0
< 
this is the body
* Connection #0 to host localhost left intact
~ ➤  curl -I http://localhost:8000
HTTP/1.1 200 OK
Content-Length: 17
Date: Wed, 22 Feb 2017 10:08:10 GMT
Server: Python/3.5 aiohttp/2.0.0a0

~ ➤  
```

Nothing went wrong here. 

The app could choose to notice the HEAD request and not spend a lot of time/resources writing body, but I don't see any problem if those not concerned by this (fairly niche) optimisation don't bother with it.

**In summary:** your current approach explicitly contravenes the HTTP RFC. The issue you're worried about is an optimisation which will apply in a very small percentage of requests and where app authors already have all the tools they need to achieve the optimization.

fafhrd91: Ok. Let's remove url dispatch part and change default allow_head to True

samuelcolvin: Great. I've updated the PR.

fafhrd91:
very good

PR #1688, opened 2017-02-27
Add 'stacklevel=2' to deprecation message.

bitdancer:
This will cause the message to reference the line where the user code calls the function, rather than the line where the deprecation message is in the aiohttp source.  This will make it easier for the user to fix their code.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -116,3 +116,6 @@ CHANGES
 
 - Deprecated: `Application.finish()` and `Application.register_on_finish()`
   will be removed in 1.4 #1602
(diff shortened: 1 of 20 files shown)


bitdancer:
Hmm.  I followed the instructions in the text box to select the 1.3 branch since this is a bug fix, but I guess it pulled in the changes relative to master because I used the 'edit' function on the master branch :(  I don't have time to make a checkout and rebase this right now.

fafhrd91:
could you create PR against https://github.com/aio-libs/aiohttp

PR #1698, opened 2017-03-07
preserve bname case in raw_headers

thehesiod:
fixes https://github.com/KeepSafe/aiohttp/issues/1570


thehesiod:
I'll work on the unittests later

fafhrd91:
fixed in https://github.com/aio-libs/aiohttp
@thehesiod could you close this PR

thehesiod:
based at looking at the code seems good, thanks

PR #1663, opened 2017-02-20
Fix web Stream Response started flag and object representation.

singulared:
## What do these changes do?

This patch will fix Exception in __repr__ method for StreamResponse class.

## Are there changes in behavior for the user?

No

## Related issue number

No

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] Add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new entry to `CHANGES.rst`
  * Choose any open position to avoid merge conflicts with other PRs.
  * Add a link to the issue you are fixing (if any) using `#issue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

PR #1700, opened 2017-03-07
_resp_impl going back to None is insane

arthurdarcet:
Introduced by #1663 :

 - After a `write_eof`, `Response.prepared` and `Response.started` are False again (?!)
 - Because of https://github.com/KeepSafe/aiohttp/blob/master/aiohttp/web_server.py#L90 the request is prepared twice
 - and the headers are sent again in the pipe after the complete body has been sent (!)

This behaviour is very problematic, but quite hidden because most clients use the `Content-Length` or `Transfer-Encoding` headers to stop reading what the server is sending once they have received all the data they expect.
Even a `curl` shows nothing anormal. But an actual telnet is pretty obvious:
```
% telnet 0.0.0.0 1111
Trying 0.0.0.0...
Connected to 0.0.0.0.
Escape character is '^]'.
GET / HTTP/1.1
 
HTTP/1.1 200 OK
Server: …
Date: Tue, 07 Mar 2017 16:49:48 GMT
Transfer-Encoding: chunked

10
My actual body
0

HTTP/1.1 200 OK
Server: …
Date: Tue, 07 Mar 2017 16:49:48 GMT
Transfer-Encoding: chunked
```

Also, nginx when doing a `proxy_pass`, does not actually respect the `Transfer-Encoding` header, read the additional data sent by aiohttp, and send them back to the client.

I tried adding a test catching this, but this seems pretty hard because the test_client drops the connections as soon as he has read enough data.

### DIFF
--- a/aiohttp/web_reqrep.py
+++ b/aiohttp/web_reqrep.py
@@ -926,7 +926,6 @@ class StreamResponse(HeadersMixin):
         self._eof_sent = True
         self._body_length = self._resp_impl.body_length
         self._req = None
-        self._resp_impl = None
 
     def __repr__(self):
         if self._eof_sent:
--- a/tests/test_web_response.py
+++ b/tests/test_web_response.py
@@ -390,8 +390,8 @@ def test_repr_after_eof():
     resp.write(b'data')
     writer.drain.return_value = ()
     yield from resp.write_eof()
-    assert not resp.started
-    assert not resp.prepared
+    assert resp.started
+    assert resp.prepared
     resp_repr = repr(resp)
     assert resp_repr == '<StreamResponse OK eof>'
 


fafhrd91:
@arthurdarcet could you hold a bit with PR's. we are now deciding how we want to do development of aiohttp. at the moment I moved all my code to separate repo, it has all fixes.

arthurdarcet:
@fafhrd91 
We are also managing our fixes separately, so we don't need you to merge this.
But do you realise how broken the current release is? it makes the whole lib unusable…

fafhrd91:
I understand that, but that is beyond my control. I don have write access to repo anymore and I don't need one.

philippb:
@fafhrd91 not sure where we lost your access to the repo. Never intentioned on our end. Might have been Laura in admin. I just gave you all permissions back. Sorry about that.

fafhrd91:
@philippb as I told you I am not planing to participate in KeepSafe's aiohttp repo

fafhrd91:
@arthurdarcet could you close this PR

PR #1712, opened 2017-03-12
run_app: allow for passing preexisting sockets

malor:
## What do these changes do?

Make `run_app()` an additional argument `sock`, so that it's possible to pass one or more preexisting socket objects to accept new connections on. This can be useful for implementing socket-based activated applications, e.g. with systemd [1]. In this case `bind()` is done by the parent process and the app just need to use the given socket (or multiple), and does not create a new one.

[1] http://0pointer.de/blog/projects/socket-activation.html

Existing workflows are not changed. This just extends the interface of `run_app()` to allow for a new use case -  passing preexisting socket objects.

## Related issue number

N/A

## Checklist

- [X] Add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,10 @@ CHANGES
 - `run_app` and the Command Line Interface now support serving over Unix domain sockets for
   faster inter-process communication.
 
+- `run_app` now supports passing a preexisting socket object. This can be useful
+  e.g. for socket-based activated applications, when binding of a socket is
+  done by the parent process.
+
 - Revert timeout handling in client request
 
 - Fix StreamResponse representation after eof
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -122,6 +122,7 @@ Rafael Viotti
 Raúl Cumplido
 Required Field
 Robert Lu
+Roman Podoliaka
 Samuel Colvin
 Sebastian Hanula
 Sebastian Hüther
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -301,7 +301,7 @@ def _wrap_add_subbapp(app):
     return add_subapp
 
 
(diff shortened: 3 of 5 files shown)


fafhrd91:
Should we accep file descriptor id as well?

Could you open PR against https://github.com/aio-libs/aiohttp

ikalnytskyi:
@fafhrd91 so this repo is not official anymore, and everyone should use the one from aio-libs from now on?

ikalnytskyi:
@fafhrd91 wrote:
> Should we accep file descriptor id as well?

asyncio doesn't support file descriptors, and is expecting sockets. I think it's ok to keep file descriptor to socket conversion up to user code, as it's not difficult and keeps aiohttp code a bit simpler. :)

malor:
@fafhrd91 

> Should we accept file descriptor id as well?

Technically we could use https://docs.python.org/3/library/socket.html#socket.fromfd to convert a given FD into a proper socket object (that we can pass to asyncio `create_server()`), but then we'd need to make sure we pass the very same `family`/`type` params, that were used for creation of this socket (as stdlib docs specifically say that such checks are not performed, but inconsistency *may* lead to errors of subsequent operations on the FD). In systemd case there is a shared library that implements helper functions - https://www.freedesktop.org/software/systemd/man/sd_is_socket.html - it's my understanding that we'd need to implement similar logic in aiohttp.

I guess, I'd rather we keep this burden on the users, at least for now. FWIW, they create these sockets either directly or by the means of systemd/inetd/etc + `socket.fromfd()` - in both cases they have knowledge about correct params for creation of socket objects.

fafhrd91:
@ikalnytskyi yes.

@malor ok, makes sense.

ikalnytskyi:
@fafhrd91 why it wasn't moved then? it would preserve stars and forks numbers.

malor:
Merged to aio-libs/aiohttp

PR #1735, opened 2017-03-21
Allow route registration with decorator

fafhrd91:
I think it is useful, especially combined with sub-applications

### DIFF
--- a/tests/test_web_application.py
+++ b/tests/test_web_application.py
@@ -3,8 +3,8 @@ from unittest import mock
 
 import pytest
 
-from aiohttp import helpers, log, web
-from aiohttp.abc import AbstractRouter
+from aiohttp import hdrs, helpers, log, web
+from aiohttp.abc import AbstractRouter, abstractmethod
 
 
 def test_app_ctor(loop):
@@ -109,6 +109,36 @@ def test_non_default_router():
     router = mock.Mock(spec=AbstractRouter)
     app = web.Application(router=router)
     assert router is app.router
+
+
+class MockAbstractRouter(AbstractRouter):
+
+    @abstractmethod
+    def add_resource(self, oath, name=None):
+        pass
+
+
+def test_add_route(loop):
+    router = mock.Mock(spec=MockAbstractRouter)
+    app = web.Application(router=router)
+
+    @app.route('/index.html')
+    def get(req):
+        pass
+
+    router.add_resource.assert_called_with(
+        '/index.html', name=None)
+    router.add_resource.return_value.add_route(
+        hdrs.METH_GET, get, expect_handler=None)
+
+    @app.route('/index.html', hdrs.METH_POST, name='test')
+    def post(req):
+        pass
+
+    router.add_resource.assert_called_with(
+        '/index.html', name='test')
+    router.add_resource.return_value.add_route(
+        hdrs.METH_POST, post, expect_handler=None)
 
 
 def test_logging():
--- a/tests/test_web_application.py
+++ b/tests/test_web_application.py
@@ -141,6 +141,15 @@ def test_add_route(loop):
         hdrs.METH_POST, post, expect_handler=None)
 
 
+def test_add_route_different_module(loop):
+    from test_web_application_helpers import app_test
+
+    with pytest.raises(TypeError):
+        @app_test.route('/404')
+        def get(req):
+            pass
+
+
 def test_logging():
     logger = mock.Mock()
     app = web.Application()
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import os
 import socket
 import stat
@@ -58,6 +59,7 @@ class Application(MutableMapping):
         self._secure_proxy_ssl_header = None
         self._loop = loop
         self._handler_args = handler_args
+        self._module = None
         self.logger = logger
 
         self._middlewares = FrozenList(middlewares)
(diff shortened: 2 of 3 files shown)


fafhrd91:
@asvetlov @kxepal @popravich @socketpair thoughts?

tests/test_web_application.py
```diff
@@ -91,6 +91,36 @@ def test_non_default_router():
...
+    @abstractmethod
+    def add_resource(self, oath, name=None):
+        pass
+
+
+def test_add_route(loop):
+    router = mock.Mock(spec=MockAbstractRouter)
+    app = web.Application(router=router)
+
+    @app.route('/index.html')
```
socketpair, line 125: in my own impplementation of such decorator, I add reference to decorated method and all decorator's kwargs and args  to some global array. And also I have a function that adds this saved information into application's routes.

In short (pseudocode):
```python
global_array = []
def handler(*args, **kwargs):
    def decorator(function):
        global_array.append((function, args, kwargs))
        return function
    return decorator

@handler(....)
async def myhandler(request):
    ....

class MyApplication(web.Application):
    def add_saved_routes(self):
        for (method, args, kwargs):
            self.add_route(....)

def main():
    app = MyApplication()
    app.add_saved_routes()
    ...
```

This allows me to use decorators BEFORE creating application (during imports). This is especially important since I have many handlers which are placed in many files.

So, could you provide complete example showing usage of your decorators for the case when there are many files with handlers ?

kxepal:
I think this is a bad idea and goes against everything we have now. Routes as decorators forces you to use module-level scope to define application, loop and the rest of bits, returning you to the ~stone age~ flask-like design. While it's obliviously simple for starters, I don't think it's right and is good in mid/long term of application life.

Yesterday I meet one interesting issue with that kind of design using aiohttp: I investigated a project, where application and uvloop where defined at the module level. This module imported another one just before uvloop was set as the default, where Consul client was created as well at the module level. And since there was no loop argument passed for him, that client picked current default asyncio loop that was available - certainly not the uvloop one, since it was not yet even created. You may guess what happened next, when half of you application works with the one loop and some parts with the else one. And obliviously, author didn't even expected surprises of that kind.

I believe decorator routes will only make more situations like this.

For current design, I tend to think that Pyramid-style will fits more without much differences. For people who need flask-like style design, there is sanic already.

jettify:
agree with @kxepal, there is even FAQ entry about such routing: http://aiohttp.readthedocs.io/en/stable/faq.html#are-there-any-plans-for-app-route-decorator-like-in-flask
Could this be implemented in separate package?

popravich:
Agree with @kxepal completely.
Such decorators should be used with extreme caution:
first you start with global app, loop, etc, etc; then your tests add more `if`s to your app logic;
then import order begins to matter; next imports induce circular dependencies, and so on and so on.
This feature works until whole project fits single module.

f0t0n:
Agreed with @kxepal.
I'd add that even in my Flask applications I never use such decorator-based routing but keep everything in the dict routing table such as the one below for the simple authentication service. Then just loading it with some simple functions to add routes to the application. The same I always did for the aiohttp-based projects.
So the routing table could be either nested to include sub-applications or separate tables for each sub-application, but anyway it's always explicit, declarative and configurable in one place. So it's easier to manage for any scale of projects.


```py
from .public import session, signature, user

api_resources = {
    user: {
        'create': ('/users', ['POST'], {}),
        'verify': ('/users/verify/<uuid:user_id>/<string:verification_key>',
                   ['PATCH'],
                   {}),
    },
    session: {
        'create': ('/users/sessions', ['POST'], {}),
        'delete': ('/users/sessions', ['DELETE'], {}),
    },
    signature: {
        'verify': ('/signatures/verify', ['HEAD'], {}),
    },
}
```

fafhrd91:
sorry for delay.

I see two problems we discuss here.
@socketpair and @kxepal describe problem with configuration context. aiohttp does not have one and pyramid has it explicitly. and problem is worse with asyncio applications because of global event loop.
(that is the reason why I deprecated `loop` parameter in Application ctor). maybe we should introduce
configuration context, but that would be aiohttp 3.0 version.

I totally agree on configuration as a side effect of import problem, but I find `Application.route()` decorator super convenient. it allows to export module functionality in consolidated way (keeps context).
what if we limit route() decorator to one module, if you use it in one module you can not use it in another.
as @popravich mentioned router() decorator works when app fits into one module.

app configuration func could look like:
```python

from . import users, content, ...

def build_app():
    app = web.Application()

    app.add_subapp('/users', users.USERS_APP)
    app.add_subapp('/content', content.CONTENT_APP)
    ...
    return app

```

fafhrd91:
added one module usage limitation

kxepal:
> added one module usage limitation

I still think this is a bad move. Yes, this feature will allow us to have nice hello-world examples. Yes, this feature will make simple to make single modules web applications as like it was for single page web apps. But. If application will grow and user will be forced to split it into modules he/she will have to rewrite whole the routing system, application initialization and drop global context usage. And all this may not be quite oblivious. So there will be two chairs problem: or stay within  a single module forever or cry for help.

Even for small applications, decorator-way may not be convenient. I'm working now on service which have only two routes. But, these routes are belongs to different API versions and I don't want to mix them or their logic in a single module because of coupling problem. So they are defined in two different modules, each of those defines how each API version should act.

There will be also a problem to use different routers with such kind of design. For instance, [aiohttp_traversal](https://github.com/zzzsochi/aiohttp_traversal). Decorator routes cannot be used with this one by design reasons.

kxepal:
I think we have here a lot of words against this feature. Is there any good reasons to have it?

fafhrd91:
Convenience, that is it, nothing else. I think we think too big, big applications, huge load, big teams. But I have a felling that in reality python get pushed back to area of glue language and data analytics. And most users just need to build something fast with minimal effort. Of course we can say f... them, learn how to write code, but again that is my feeling, those people are our primary audience. And It seems client part of aiohttp is more popular than server, which proves my point of view.

fafhrd91:
It is not about replacing normal registration with decorators, it is just adding ergonomics for small apps

asvetlov:
Let's create a small third party library for route decorators, put it into aio-libs and reference from aiohttp documentation.

fafhrd91:
I don't like idea of separate repo. this feature is for new developers with prior flask and similar experience.
so it would be useless if you need to look for it. actually I think this feature could change how apps get build, one app per module, good isolation.

PR #1738, opened 2017-03-21
add ids to loop params in pytest plugin

samuelcolvin:
this should make the two loop params clearer, particularly when users suddenly find their tests "running twice". When using pytest with `-v` the tests should now use these ides instead of `[loop-o]` and `[loop-1]`.

I'm open to better names than "uvloop" and "pyloop", but I couldn't immediately think of better options.

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -83,22 +83,28 @@ def pytest_configure(config):
         without_uvloop = True
 
     LOOP_FACTORIES.clear()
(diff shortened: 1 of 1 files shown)


fafhrd91:
"default" or "asyncio"?

fafhrd91:
I think we should use "uvloop", everybody know it now.
I think "pyloop" is fine.

fafhrd91:
I'll merge this PR, we always can rename change id later

PR #1744, opened 2017-03-23
check for runtime warnings in tests

samuelcolvin:
## What do these changes do?

Checks for `coroutine 'X' was never awaited` runtime warnings and raises an exception, thus preventing tests which silently don't await coroutines (via pytest's muting of stdout/stderr).

No


samuelcolvin:
Not just useful for aiohttp itself, but also for other projects with use it's pytest plugin.

fafhrd91:
thanks

Issue #1746, opened 2017-03-24
Need possibility to modify client.Session default headers in constructed session


penguinolog:
## Long story short

`client.Session._default_headers` is "protected" field, while sometimes required to modify it in runtime (set specific 'User-Agent' and etc) without restart.

## Expected behaviour

Getter property `client.Session.default_headers` (it's mutable -> setter is not needed) is suitable solution for changing headers between requests.

## Actual behaviour

`client.Session._default_headers` is used.

fafhrd91:
I do not think this is good idea. What is your specific use case?

penguinolog:
Currently not mine case: zeep library sets HTTP-Client header on received session, constructed outside.

fafhrd91:
But that's bad behavior. Let's say library set default header, but then owner of session modifies it, so no one would be able to rely what's in default headers.

-1

penguinolog:
Accepted concern.

PR #1747, opened 2017-03-24
Added `default_headers` property to `ClientSession`.

penguinolog:
## What do these changes do?

Added `default_headers` property to `ClientSession`.

Session default_headers is public available.

## Related issue number

#1746


penguinolog:
Abandon

PR #1755, opened 2017-03-27
Remove copy of body headers into request headers

JohanLorenzo:
Thank you for this library! Mozilla uses aiohttp to upload artifacts on Amazon S3.

We recently migrated to 2.0.0, which gave this particular side-effect: https://github.com/mozilla-releng/scriptworker/issues/98. We noticed 'Content-Disposition' is now added to PUT request that contain a data payload (like a text/plain document). This has the unwanted effect to confuse Firefox when downloading artifacts from S3. Before then, artifacts could be directly displayed in browser, but now they are considered as blobs to be downloaded.

I found headers are now extended by the ones detected from the body. in https://github.com/aio-libs/aiohttp/commit/d69838a5b0ea8ce8e6ac5e39cb797ac81f943028. I'm not sure what was the context at the time.

## What do these changes do?

With this patch, body headers are not included in the actual headers of a request. Based on https://github.com/aio-libs/aiohttp/commit/d69838a5b0ea8ce8e6ac5e39cb797ac81f943028, this was the behavior before this commit.

What do you think @fafhrd91 ?

That fixes a regression from 1.x.

## Related issue number

* https://github.com/mozilla-releng/scriptworker/issues/98

## Checklist

(Side note: I just deleted some lines 😃 )

It seems this particular area wasn't tested. I haven't changed the tests. 

I'm not sure if changes are needed in the docs.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,8 @@ Changes
 
 - Dropped "%O" in access logger #1673
 
+- Headers from body are not in Request headers anymore
+
 
 2.0.3 (2017-03-24)
 ------------------
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -82,6 +82,7 @@ Jeroen van der Heijden
 Jesus Cea
 Jinkyu Yi
 Joel Watts
+Johan Lorenzo
 Joongi Kim
 Julien Duponchelle
 Junjie Tao
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -275,12 +275,6 @@ class ClientRequest:
                 hdrs.CONTENT_TYPE not in skip_auto_headers):
             self.headers[hdrs.CONTENT_TYPE] = body.content_type
 
-        # copy payload headers
-        if body.headers:
-            for (key, value) in body.headers.items():
-                if key not in self.headers:
-                    self.headers[key] = value
-
     def update_expect_continue(self, expect=False):
         if expect:
             self.headers[hdrs.EXPECT] = '100-continue'


fafhrd91:
I do not think this is good solution. I think default expectation is to have content-disposition header if you send file. What if we change attachment to inline type? That should solve your problem

JohanLorenzo:
Thank you for your rapid answer 😃 

I'm not sure to understand the necessity of `Content-Disposition`, though. Could you help me out, here?

From what I read on the [Amazon S3 docs](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html#RESTObjectPUT-requests-request-headers) and the [W3C specs](https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.5.1), this is meant for clients to present the data differently (like changing a file name). Because this is not part of the HTTP specs, this is a nice to have, in my opinion.

Hence, forcing this header as a default on POST/PUT requests doesn't seem natural to me. Moreover, this might also represent a security risk. For example, one may not want to share the actual file name stored on a local disk. I also don't see a way filter out this header (with `skip_auto_header` for instance) before the request is sent.

fafhrd91:
aiohttp adds this header only if you send file object. I'll replace attachment with inline for 2.0.4 release
and think about removing header in 2.1.0. does it sound ok for you?

JohanLorenzo:
Sounds good to me. Let's go with this plan!

fafhrd91:
@kxepal what do you think about `inline` by default?

@JohanLorenzo another option is to drop content-disposition for non-multipart requests

JohanLorenzo:
Dropping it for non-multiparts makes sense to me.

kxepal:
+1 to drop. This header is not what to add for the requests. It's the server response header.

fafhrd91:
ok, then I'll leave attachment as default, and will drop content-disposition from stream

fafhrd91:
fixed in 2.0 and master. will release new version later today

fafhrd91:
2.0.4 is released

JohanLorenzo:
Thank you :)

PR #1772, opened 2017-03-30
Update client_reqrep.py to fix exception when proxied

ATXadam:
Ensure _connection.protocol is actually an attribute before polling on _response_eof to avoid exception thrown.


fafhrd91:
could you add test

fafhrd91:
at least post exception you got

ATXadam:
Just saw these emails, exception was that _connection did not have attribute protocol, hence the hasattr sanity check.

PR #1775, opened 2017-03-31
use io.BytesIO instead of joining bytes to avoid copying everything

arthurdarcet:
The most important change here is the update in `web_request.Request.read`: reading the request body entirely then return `b''.join(chunks)` is very inefficient because it has to copy the complete body just before returning.
This means that currently, when a 200MB request is sent to an aiohttp server, the server would need up to 400MB of RAM to handle it (it would be less in real life because the client does not transmit the entire body at once. But it's still a lot).
With `io.BytesIO`, the actual underlying buffer is returned when calling `BytesIO.getvalue`, this way only 200MB of RAM is needed to handle the same request.

I have also updated other parts of the library (mainly the HTTP payload writer) to use BytesIO instead of joining lists of bytes.

### DIFF
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -2,6 +2,7 @@
 
 import asyncio
 import collections
+import io
 import socket
 import zlib
 from urllib.parse import SplitResult
@@ -140,7 +141,7 @@ class PayloadWriter(AbstractPayloadWriter):
         self.output_size = 0
 
         self._eof = False
-        self._buffer = []
+        self._buffer = io.BytesIO()
         self._compress = None
         self._drain_waiter = None
 
@@ -150,14 +151,17 @@ class PayloadWriter(AbstractPayloadWriter):
         elif acquire:
             self._stream.acquire(self)
 
+    def _flush_buffer(self, keep_buffer=False):
+        chunk = self._buffer.getvalue()
+        if chunk:
+            self._transport.write(chunk)
+            if not keep_buffer:
+                self._buffer = io.BytesIO()
+
     def set_transport(self, transport):
         self._transport = transport
 
-        chunk = b''.join(self._buffer)
-        if chunk:
-            transport.write(chunk)
-            self._buffer.clear()
-
+        self._flush_buffer()
         if self._drain_waiter is not None:
             waiter, self._drain_waiter = self._drain_waiter, None
             if not waiter.done():
@@ -190,7 +194,7 @@ class PayloadWriter(AbstractPayloadWriter):
             size = len(chunk)
             self.buffer_size += size
             self.output_size += size
-            self._buffer.append(chunk)
+            self._buffer.write(chunk)
 
     def _write(self, chunk):
         size = len(chunk)
@@ -198,14 +202,13 @@ class PayloadWriter(AbstractPayloadWriter):
         self.output_size += size
 
         if self._transport is not None:
-            if self._buffer:
-                self._buffer.append(chunk)
-                self._transport.write(b''.join(self._buffer))
-                self._buffer.clear()
+            if self._buffer.tell():
+                self._buffer.write(chunk)
+                self._flush_buffer()
             else:
                 self._transport.write(chunk)
         else:
-            self._buffer.append(chunk)
+            self._buffer.write(chunk)
 
     def write(self, chunk, *, drain=True, LIMIT=64*1024):
         """Writes chunk of data to a stream.
@@ -248,11 +251,7 @@ class PayloadWriter(AbstractPayloadWriter):
         headers = status_line + ''.join(
             [k + SEP + v + END for k, v in headers.items()])
         headers = headers.encode('utf-8') + b'\r\n'
-
-        size = len(headers)
-        self.buffer_size += size
-        self.output_size += size
-        self._buffer.append(headers)
+        self.buffer_data(headers)
 
     @asyncio.coroutine
     def write_eof(self, chunk=b''):
@@ -287,10 +286,7 @@ class PayloadWriter(AbstractPayloadWriter):
     @asyncio.coroutine
     def drain(self, last=False):
         if self._transport is not None:
-            if self._buffer:
-                self._transport.write(b''.join(self._buffer))
-                if not last:
-                    self._buffer.clear()
+            self._flush_buffer(keep_buffer=last)
             yield from self._stream.drain()
         else:
             # wait for transport
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -1,5 +1,6 @@
 import asyncio
 import collections
+import io
 import traceback
 
 from . import helpers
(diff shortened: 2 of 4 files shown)


fafhrd91:
I do not think this is right solution. you should use streaming interfaces, aiohttp supports flow controls on both side.

when sending a lot of data you should call drain(), or just use "yield from resp.write()"
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_writer.py#L118

on reading side you should "request.content.read()" instead of "request.read()", "request.content" handles back pressure automatically.

arthurdarcet:
the streams.py file is doing exactly the same in read: "append `bytes` objects to a list ; then return b''.join(parts)". So using `request.content.read()` instead of `request.read()` does not change anything

This approach means a simple handler like this:
```python
async def handler(request):
    await request.content.read()
    return aiohttp.web.Response(text='ok')
```
requires 440MB of RAM on my machine to handle a single 200MB request, versus 217MB with this patch

fafhrd91:
`StreamReader` has several methods you can use for partial reading.

```python
async def handler(request):
    stream = aiohttp.web.StreamResponse()
    await stream.prepare(request)

    while True:
         chunk = await request.content.readany()
         stream.write(chunk)

    return stream
```

arthurdarcet:
and this still won't solve anything: if the client's data comes in fast enough, the first chunk will be several hundreds megabytes and that's that much RAM going to waste because readany needs 2B of RAM for every byte of body it reads (and since Python does not release RAM on Unix, you really just discarded that RAM).

Of course there are workaround, you just have to define a maximum chunk size, then use `read(CHUNK_SIZE)` and you will trade a very tiny slowdown for the insurance that aiohttp won't waste any more than CHUNK_SIZE of RAM.

But I don't understand why we are even discussing this: this patch ensures aiohttp waste 0 amount of RAM, with no change at all to the API and doesn't even add complexity to the code (i think it even makes things simpler). Could you explain why you think this patch is the wrong approach?

fafhrd91:
you should test first and do benchmarks, we choose to use "".join() because it is fast.
also please test you server tcp stack, I've never saw several megabytes reads from socket.recv()
usually tcp frame mtu is around 1500 bytes

fafhrd91:
also if you stop reading from socket, kernel stops accepting data over tcp, and client waits until server continues. aiohttp supports flow control for tcp transports

arthurdarcet:
The test server I mentioned earlier, which used 400MB of RAM to handle a 200MB-request, was a complete aiohttp server and the request were sent using curl from the outside, through the network stack of my machine (albeit through the lo interface)


I understand the need for speed at all cost, but I still think some memory is copied unnecessarily here, with performance hits on memory mostly but probably also on time.
This `BytesIO` version is indeed slower ; I think the main issue is the double "append to list, return b''.join(parts)" done by read: first done here https://github.com/aio-libs/aiohttp/blob/master/aiohttp/streams.py#L291 but this merges parts that come from here https://github.com/aio-libs/aiohttp/blob/master/aiohttp/streams.py#L372 and the body is then three time in memory (original blobs, _read_no_wait version, and the result of read)
I'll update the PR to fix only this.

```
python -m timeit -s "buffer = [b'0' * 1500] * $N" "res = []; [res.append(p) for p in buffer]; b''.join(res)"
python -m timeit -s "buffer = [b'0' * 1500] * $N; import io" "buf = io.BytesIO(); [buf.write(p) for p in buffer]; buf.getvalue()"
```

```
- N=1
  join: 0.58 usec per loop
  BytesIO: 0.954 usec per loop

- N=10
  join: 2.25 usec per loop
  BytesIO: 4.31 usec per loop

- N=100
  join: 29.1 usec per loop
  BytesIO: 98.8 usec per loop

- N=1000
  join: 220 usec per loop
  BytesIO: 934 usec per loop
```

fafhrd91:
I still do not understand what are you trying to fix

fafhrd91:
also you can check repo history, we used bytearray, then it was proved that simple join is better

arthurdarcet:
Imagine you want to process the complete body posted to a server [1].
You can of course build the handler using `readany`, and buffering the content by hand until you've built the complete bytestring – probably using BytesIO or some kind of buffer – but that's overly complicated and completely unneeded: I want the complete body sent to the handler, i expect i can get it with `content.read()`. Handling buffering and reconstructing the body from its chunks is the job of the http server and that's exactly what `content.read()` is supposed to do.

Except calling `content.read()` need to duplicate the complete body three time in memory, which is completely inefficient and definitely not needed.

[1] In case you don't believe such use case exist: imagine a server receiving a zipfile as the body of the request, and then doing something with it.

fafhrd91:
but loading whole payload into memory is bad solution in first place. anyone can kill such service in matter of seconds. check sanic, anyone can kill it very easy, because it does exactly this.

in any case aiohttp does not prevent you from doing this.

here is your helper function:

```python
async def load(request):
   body = io.Bytes()

    while True:
         chunk = await request.content.readany()
         if chunk:
             body.write(chunk)
        else:
             break

    return body

async def handler(request):
    body = await load(request)
    ...

```

arthurdarcet:
I know perfectly well that loading the whole body is not a good thing. that does not mean it can be avoided. Again, same example: loading a zipfile require to load everything in memory.

And again, as i said earlier, i know it can be worked around by manually merging the chunks. But that's unnecessarily complicated. Why do you think read() has to be inefficient?!

fafhrd91:
it is trade-off, we trade memory usage for speed.
especially because your use case is wrong in first place, and most users does not have problem with loading whole blob into memory. current implementation does not prevent you from implementing more efficient usage of memory, while with io.Bytes everyone would pay price for nothing.

fafhrd91:
did you check io.Bytes implementation? it may reallocate memory.
there is no magic in memory management, if you need extend allocated memory chunk, you just need reallocate and copy old memory to new.

arthurdarcet:
What i'm trying to achieve here is making sure the memory is only copied once. And there seem to be a lot of `b''.join(…)` going around to should copy the bytes.

I checked the BytesIO implementation of course, using BytesIO means the copy appends during the "write", so that when we want to access the value, everything is already contiguous.

What I did not know, and is very important here, is that `b''.join(…)` is very optimised for some very specific calls:
```python
>>> b = b'abc'; b''.join([b]) is b
True
```
So calling `b''.join([b])` is really just unwinding the input list without copying anything

This is why the bytes don't get copied around too much with the current implementation, without this, everything would get duplicated several times over. A comment mentioning that in streams.py would help.


I'm closing this, I agree the current implementation is optimal:
 - The requests come in chunks, so we have them in memory but not in contiguous parts
 - Storing the request contiguously would require moving the chunks as they come, which is not an option to avoid unnecessarily slowing down everything
 - Since the request is not stored contiguously, getting the body as a single bytestring requires copying things around (this could be done with a bit less memory by releasing the original chunks as they are moved in the result, but this would need to be done in C to have any hope of matching the perfs of `b''.join`, so I won't try it)

fafhrd91:
if you really want very low-level access to stream, help me with async-tokio project

PR #1794, opened 2017-04-06
Fixing the documentation of AioHTTPTestCase get_application at testing.rst

diogommartins:
On versions 1.x  `AioHttpTestCase.get_application` had to be overwritten with a `loop` parameter. It doesn`t anymore and the documentation doesn't reflect that change


fafhrd91:
thanks!

Issue #1805, opened 2017-04-10
Unable to iterate over chunks as sent by server.


ahane:
## Long story short

I am talking to a HTTP server that uses `transfer-encoding: chunked` to stream records, whith each chunk representing one record. 
aiohttp currently offers no API for iterating over the chunks exactly as sent by the server.

## Expected behaviour

I would expect the response.content.iter_chunked(n) method to default to the chunking already defined by the server if n=None or a special flag is passed.

See the [similar method in the Requests library](http://docs.python-requests.org/en/master/api/#requests.Response.iter_content) that shows the desired behaviour.

## Actual behaviour

n=None returns an empty bytestring, while n=-1 doesn't return at all.

There is no method to recover the chunking of the server.

fafhrd91:
you can use `readany`, it returns data available in buffer. but right now it is not possible.
would you like to provide PR?

ahane:
yes ;) readany does join all chunks in the buffer into one bytestring though
https://github.com/aio-libs/aiohttp/pull/1806

fafhrd91:
fixed by you :)

PR #1806, opened 2017-04-10
Add method for iterating of chunks sent by server

ahane:
## What do these changes do?

This patch adds the `iter_chunks` and the `readchunk` method to the `StreamReader`.
`readchunk` will read a single element from the buffer, which represents one chunk as received from the server.
This enables the `iter_chunks` method which allows for asynchronous iteration over the chunks in the size as received fromt the server.

See [similar behaviour in the Requests library](http://docs.python-requests.org/en/master/api/#requests.Response.iter_content)

The existing method `iter_chunked` is not modified.
Instead I added the `iter_chunks` method to avoid breaking something by changing the behaviour of the `iter_chunked` signature.

I believe that my choice of naming is a bit unfortunate as it could lead to confusion between the two similarly named methods. On the other hand the method does iterate over the chunks of the response so the name is on point.

## Improvements
Ideally there would be only one method with a default value of `n=None` which would implement the desired behaviour. Any thoughts on this?

## Related issue number

#1805

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,8 @@ Changes
 
 - Dropped "%O" in access logger #1673
 
+- Added `iter_chunks` to response.content object. #1805
+
 
 2.0.6 (2017-04-04)
 ------------------
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -3,6 +3,7 @@ Contributors
 
 A. Jesse Jiryu Davis
 Adam Mills
+Alec Hanefeld
 Alejandro Gómez
 Aleksandr Danshyn
 Aleksey Kutepov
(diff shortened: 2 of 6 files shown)


fafhrd91:
lets rename "readone" to "readchunk"
otherwise lgtm.

@asvetlov @kxepal any questions?

kxepal:
Good, helpful, method. Thanks for contribution!

> Ideally there would be only one method with a default value of n=None which would implement the desired behaviour. Any thoughts on this?

I think it's a good idea, but for another PR. In the end, we should stay with a single `iter_chunk(s|ed)` method, so one of the have to go.

ahane:
Done, thanks aswell!

fafhrd91:
thanks!

Issue #1180, opened 2016-09-16
Client request timeout should be optional


adamrothman:
## Long story short

As described in #877, attempting to use `aiohttp.Timeout` in a Tornado-based app results in a `RuntimeError`. In that issue, @bdarnell offers a solution that uses Tornado's own timeout utility. Through version 0.22.5, `aiohttp`'s client request plumbing does not use `aiohttp.Timeout` internally; users were expected to apply it externally when desired.

The 1.0.0 release of `aiohttp` includes dc79e14b69bae57db3bfb09ae4640c4c80dc0d1c, which forces all requests to use a `Timeout`. This seems a bit heavy-handed as it makes it impossible to use `aiohttp`'s client with Tornado, even with Ben's suggested workaround. If the `timeout` kwarg to `aiohttp.ClientSession::request` is `None`, it should work the way it used to.
## Steps to reproduce

The following test server can be used to consistently reproduce a `RuntimeError` as in #877 (trace follows):

``` python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
from asyncio import get_event_loop

from aiohttp import ClientSession
from tornado.httpserver import HTTPServer
from tornado.platform.asyncio import AsyncIOMainLoop
from tornado.web import Application
from tornado.web import RequestHandler


class Handler(RequestHandler):

    async def get(self):
        async with ClientSession() as client:
            async with client.get('https://www.google.com') as response:
                self.write(await response.text())


if __name__ == '__main__':
    AsyncIOMainLoop().install()

    app = Application([
        (r'/', Handler),
    ])
    server = HTTPServer(app)
    server.listen(9999)

    get_event_loop().run_forever()

```

And the resulting error:

```
Uncaught exception GET / (127.0.0.1)
HTTPServerRequest(protocol='http', host='localhost:1337', method='GET', uri='/', version='HTTP/1.1', remote_ip='127.0.0.1', headers={'Host': 'localhost:1337', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive', 'User-Agent': 'HTTPie/0.9.6', 'Accept': '*/*'})
Traceback (most recent call last):
  File "/Users/adam/.pyenv/versions/3.5.2/lib/python3.5/site-packages/tornado/web.py", line 1469, in _execute
    result = yield result
  File "/Users/adam/.pyenv/versions/3.5.2/lib/python3.5/site-packages/tornado/gen.py", line 1015, in run
    value = future.result()
[...]

PR #1817, opened 2017-04-15
Avoid creating TimerContext when there is no timeout

jcugat:
This will allow to use aiohttp with ioloops that don't implement
asyncio.Task.current_task, like Tornado's couroutine runner.

More info: https://github.com/aio-libs/aiohttp/issues/1180

## What do these changes do?

This will allow to use aiohttp with ioloops that don't implement asyncio.Task.current_task, like Tornado's couroutine runner.

This will restore previous functionality that was available in v1.3.3, which was possible to run with Tornado when setting `timeout=None`. This was lost in this refactor: 38360f9a05aa09ac2bf9a7077edbc2fb0e2520aa

## Related issue number

More information is available in the original issue: https://github.com/aio-libs/aiohttp/issues/1180

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,8 @@ Changes
 
 - Added `iter_chunks` to response.content object. #1805
 
+- Avoid creating TimerContext when there is no timeout to allow compatibility with Tornado. #1817 #1180
+
 
 2.0.7 (2017-04-12)
 ------------------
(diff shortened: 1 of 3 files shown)


asvetlov:
I believe Tornado should fix its issue with asyncio compatibility instead of trying to invent workarounds with tornado in every async library.

jcugat:
Yes, I agree than the best outcome would be to implement `asyncio.Task.current_task` in Tornado, but since this behaviour was previously possible (in aiohttp v1.3.3) but it's broken now, I'd treat this as an actual bug.

fafhrd91:
thanks

fafhrd91:
I am implementing alternative event loop at the moment, and `asyncio.Task.current_task` api is so bad and ugly.

PR #1819, opened 2017-04-16
Add new ttl option to expire DNS entries after N seconds

pfreixes:
## What do these changes do?

Add a new option to expire DNS cached entries at some point, this will avoid having invalid DNS entries for those environments where translations changes in some time policy.

By default it behaves as backward compatibility, entries are cached forever. Using the new `ttl_dns_cache` parameter the user can specify how many seconds a DNS entry will be keep into the table before to be refreshed. 

## Checklist

### DIFF
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -689,7 +689,7 @@ TCPConnector
 
 .. class:: TCPConnector(*, verify_ssl=True, fingerprint=None,\
                         use_dns_cache=True, \
-                        ttl_dns_cache=None, \
+                        ttl_dns_cache=10, \
                         family=0, ssl_context=None, conn_timeout=None, \
                         keepalive_timeout=30, limit=None, \
                         force_close=False, loop=None, local_addr=None, \
@@ -732,7 +732,7 @@ TCPConnector
          The default is changed to ``True``
 
    :param int ttl_dns_cache: expire after some seconds the DNS entries, ``None``
-      by default therefore cached forever.
+      means cached forever. By default 10 seconds.
 
       By default DNS entries are cached forever, in some environments the IP
       addresses related to a specific HOST can change after a specific time. Use 
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -689,6 +689,7 @@ TCPConnector
 
 .. class:: TCPConnector(*, verify_ssl=True, fingerprint=None,\
                         use_dns_cache=True, \
+                        ttl_dns_cache=None, \
                         family=0, ssl_context=None, conn_timeout=None, \
                         keepalive_timeout=30, limit=None, \
                         force_close=False, loop=None, local_addr=None, \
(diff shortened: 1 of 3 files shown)


ludovic-gasc:
Hi @pfreixes,

Normally, you have a ttl at server side for each DNS record.
Are you sure it's kept in cache forever or the ttl is respected ?

Thanks for your answer.

pfreixes:
Hi @GMLudo,

Unfortunately, retrieve the TTL as a related metadata of a DNS query can end up being a mess, the default implementation of `getaddrinfo` relays on the C implementation that implements for POSIX systems the most common RFC, including as a response a limited structure [1]. I would say that using `aiodns` and avoiding the naive `gethostbyname` implementation and using the raw access to the `cares` via the `query` function it should be possible.

IMHO the current commit avoids all of this complexity just adding a new user parameter, with that the user has the chance to avoid having wrong entries in his DNS cache and also taking advantage of the cache yet. In fact, I would prefer to put some value such as 60 seconds per default ... but it could create too much controversial because of breaking change.

There is a discussion also open in the Tornado repo about that [2]

Yes, If I'm not wrong a `Session` caches the entries forever. Use the same session for a long time can end up having DNS entries that are not valid, perhaps if the services are behind an AWS ELB [3] 

[1] http://man7.org/linux/man-pages/man3/getaddrinfo.3.html
[2] https://github.com/tornadoweb/tornado/pull/626
[3] https://aws.amazon.com/articles/1636185810492479

fafhrd91:
I like your changes. could you just use loop.time() instead of monotonic.

fafhrd91:
some more notes:

1. let's use some default value, something like 10
2. also could you add doc for this parameter to client_reference.rst
3. add yourself to contributors list

ludovic-gasc:
Hi @pfreixes,

Thanks for your explanations, it's now clear to me, and thanks for your pull request BTW ;-)

Have a nice week.

pfreixes:
@fafhrd91 done. The documentation was already there and I've contributed before and I was already in at CONTRIBUTORS.rst file

FYI This week I'm gonna work in give support for multiple hosts address per hostname, it might mean some changes in TCPConnector and the DNS resolvers. The question is how backward compatibility has to be the `cached_hosts` property [1]? I didn't want put the timestamp and the address as a unique value to avoid break the data type exposed by this property, but the changes behind the commit related to give support for multiple support might need to change the string type to something like a set. What do you think ? 

[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L590

pfreixes:
I've renamed a bit the unit tests about the DNS stuff, just let me know if you are happier with the new approx.

docs/client_reference.rst
```diff
@@ -730,6 +731,16 @@ TCPConnector
...
 
+   :param int ttl_dns_cache: expire after some seconds the DNS entries, ``None``
+      means cached forever. By default 10 seconds.
+
+      By default DNS entries are cached forever, in some environments the IP
+      addresses related to a specific HOST can change after a specific time. Use 
+      this option to keep the DNS cache updated refreshing each entry after N
+      seconds.
+
+      .. versionadded:: 2.0.8
```
fafhrd91, line 742: change it to 2.1

fafhrd91:
thanks!

Issue #1791, opened 2017-04-06
aiohttp VS requests proxy from environment variables difference


MRigal:
## Long story short

I know the problem was already shortly discussed in the issues #529 and #102 but I would like to present it again as I believe the discussion was not satisfying for people who would assume aiohttp would work as similar as possible to the requests library

## Expected behaviour

I expect aiohttp to work as similar as possible to requests. This is not the case for proxies from environment variables

## Actual behaviour

proxy info from environment variables is not used.

## Steps to reproduce

The following code will work differently on a machine with proxy and without proxy. The aiohttp code will just raise `TimeoutError` permanently...

```
import aiohttp
import asyncio
import requests

async def main():
    params = {'language': 'DE', 'address': 'Yorckstr. 1, Berlin'}
    url = 'https://maps.googleapis.com/maps/api/geocode/json'
    timeout = 5.0
    # The first synchronous request works the same with or without proxy
    with requests.Session() as session:
        location = session.get(url=url, params=params, timeout=timeout)
        print(location)
    # The asynchronous aiohttp request works properly without proxy 
    # but times out on a machine requiring a proxy and having it in an environment variable
    async with aiohttp.ClientSession() as session:
        location = await session.get(url=url, params=params, timeout=timeout)
        print(location)

if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(main())

```

## Your environment

As well OSX as Linux

## Suggestions

1) Similarly to requests (https://github.com/kennethreitz/requests/blob/master/requests/compat.py#L57), we could import the `getproxies()` (https://docs.python.org/3/library/urllib.request.html#urllib.request.getproxies) function from urllib.requests. Although people expressed concerns about it, it would be similar behaviour
2) If not, it should be stated explicitely in the docs that the behaviour towards environment variables is different from requests!
3) We could create a `session_from_env()` method. But I would then also tend to update the `_request()` method to be able to take `proxy_from_env=False` as kwargs, in order to give enough flexibility to the user.

I'm willing to work on these ones, just tell me where the road goes!

fafhrd91:
I don't have strong opinion on this. I am fine with add new class method to ClientSession, something like `from_env`. It just needs to be documented.

fafhrd91:
thanks again, fixed by #1825

MRigal:
[...]

Issue #529, opened 2015-09-26
Add support for http_proxy environment variable


wrobell:
A lot of software is checking for "http_proxy" env variable and if it exists, then it uses the proxy specified in the variable.

Would it be possible to add appropriate code in `aiohttp.request` or should each software using aiohttp do it on its own?

asvetlov:
I think handling `http_proxy` environment variable is out of aiohttp duties.
`requests` library doesn't handle it also for example.

aeroevan:
IIRC, requests does handle http{,s}_proxy environment variables:
http://docs.python-requests.org/en/latest/user/advanced/#proxies

asvetlov:
@aeroevan that sounds strange because I cannot find `http_proxy` string neither in request nor in urllib3 source code.

wrobell:
Also, `urllib.request` does

```
In addition, if proxy settings are detected (for example, when a *_proxy environment variable like http_proxy is set), ProxyHandler is default installed and makes sure the requests are handled through the proxy.
```

https://docs.python.org/3/library/urllib.request.html#module-urllib.request

asvetlov:
@wrobell urllib is not the best example.
Anyway, I've made decision: proxy environment variables will not be processed by aiohttp

aeroevan:
@asvetlov I think they build the environment variable lookup programmatically based on protocol to get HTTPS_PROXY, FTP_PROXY, etc.

qsz13:
@asvetlov  Sorry, but I still don't get it. 
Actually, requests use urllib.request.getproxies() to check the env:
https://github.com/kennethreitz/requests/blob/master/requests/compat.py
So, why not add support for the http_proxy in env?

popravich:
Hi @qsz13, using proxy settings fetched from env will work in a one single case -- when you use all defaults (I mean `with ClientSession() as sess: sess.get/post/etc`).
Implicitly changing default connector to other one is bad idea.
Consider following example:
You have a project which fetches web pages parses them and store results in elasticsearch.
Lets say that crawler part depends on `http_proxy` setting and its ok, but elasticsearch
uses http protocol as well, so elasticsearch client will end up trying to connect to your
local elasticsearch server using the same proxy as fetcher (one from http_proxy env var).
Such cases are really hard to debug. Obvious decision in this case to use another env var `no_proxy`
with elasticsearch server address and so you end up "programming" your environment instead of
correctly programming your system.
[...]

Issue #102, opened 2014-07-07
Use the system proxy settings when detected


m13253:
There is a `urllib.request.getproxies()` function provided by Python standard library.

I hope aiohttp can detect system proxy settings and enable proxy automatically if proxy is set.

joshfriend:
As a long time user of `requests`, this confused me for a while before i realized that `requests` does proxy configuration automatically, and `aiohttp` does not.

PR #1825, opened 2017-04-19
Proxy from environment variables

MRigal:
## What do these changes do?

Allows optional reading proxy from environment variables. Addresses #1791 #529 and #102

They can use proxy from environment variables like in requests, even if it is disabled by default here

## Related issue number

Resolving open issue #1791

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -292,7 +292,10 @@ class ClientRequest:
         if expect:
             self._continue = helpers.create_future(self.loop)
 
-    def update_proxy(self, proxy, proxy_auth):
+    def update_proxy(self, proxy, proxy_auth, proxy_from_env):
+        if proxy_from_env and not proxy:
+            proxy_url = getproxies().get(self.original_url.scheme)
+            proxy = URL(proxy_url) if proxy_url else None
         if proxy and not proxy.scheme == 'http':
             raise ValueError("Only http proxies are supported")
         if proxy_auth and not isinstance(proxy_auth, helpers.BasicAuth):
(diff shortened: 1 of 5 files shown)


aiohttp/client_reqrep.py
```diff
@@ -96,7 +98,7 @@ def __init__(self, method, url, *,
         self.update_cookies(cookies)
         self.update_content_encoding(data)
         self.update_auth(auth)
-        self.update_proxy(proxy, proxy_auth)
+        self.update_proxy(proxy, proxy_auth, proxy_from_env)
```
fafhrd91, line 101: let's use False instead of None

MRigal: ok. I'll wait for more comments and push the changes

fafhrd91:
comments anyone?

@asvetlov @kxepal ^^

aiohttp/client_reqrep.py
MRigal: I've added the change and squashed a bit the commit history

kxepal:
LGFM, though I'm pretty sure people will get annoyed by `proxy_from_env` flag they forced to set on most of time. Quite common behavior is just fallback to environment variables if they are available and no explicit value was provided. But I can imagine cases when you want to turn off such behavior, however.

I think that flag should be described quite notable since it's potential gotcha.

MRigal:
@kxepal this is also fully my opinion and why I reported it as an issue in #1791 !!! But as pointed in the various discussions, @fafhrd91 and @asvetlov are strictly against falling back to environment variables...

Unfortunately for now, aiohttp doesn't even allow this possibility, so this is a first step.
Enabling it by default, is, IMHO, desired, but should happen only in a 3.0 release, as this will break the behaviour of some installs.

This WAS indeed a real gotcha towards `requests` and in this PR I make the actual behaviour quite clear in the documentation.

kxepal:
@MRigal Sorry, I didn't follow those discussions. But indeed, that's a good step forward.

fafhrd91:
good work!

Issue #1797, opened 2017-04-07
aiohttp 2.0.x regression on timeouts (proxy combined with throttling through Semaphore?)


MRigal:
## Long story short

Exactly the same code runs on aiohttp 1.3.x and fails in 2.0.5

## Expected behaviour

I use async.gather() to create 1000 futures occurence of an API call, where timeout is set to 5 secs.
I use a semaphore to limit to 100 the amount of parallel running requests.
When running on 2.x it works if I do not use a proxy, but it fails if I define a proxy.
The whole process should take around 20s and does so on aiohttp 1.3.5

## Actual behaviour

After 5 seconds, the requests are failing due to TimeoutError. I fear the proxy connections of the futures are started (aenter) all at once, not when the semaphore actually releases them and thus a TimeoutError is probably raised before the requests were actually sent out.

## Steps to reproduce

I've tried to describe it above. My code is actually pretty complex, but as I said a simple downgrade of aiohttp fixes the problem.
Maybe the description is enough to fix the bug. Else, if required, I could try to write a proper test, but this is not so tricky as it requires finding real-life API calls to be made to trigger the timeouts... Maybe the description is enough...

## Your environment

Ubuntu 16.04, python 3.6.0 or 3.5.1, pyenv virtualenv and proxy in local network

fafhrd91:
it is hard to say what is the problem. timeouts and connection limits are changed significantly in 2.0

with 2.0 you don't really need separate semaphore, you can use "limit" parameter for TcpConnection,
it limits number of concurrent requests.

fafhrd91:
as example for this kind of test you can use test_client_functionional.py 
you can start local http server and implement exact logic you need

MRigal:
Thanks @fafhrd91 for pointing to the right test, I might try to write something if I can find the time for it...

Just as a note: I am using the semaphore also because behind these 1000 queries (which are even more on production), I have the same amount of read to the DB (!) And I don't want to build up thousands of connections. The advantage of the semaphore is that it should limit all async functions underneath. This is why I think that aiohttp should handle it properly in the version 2.x also! :)

fafhrd91:
I don't see any reason why it should not. Default limit is 100, try to set it to 0



Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

MRigal:
You might have misunderstood me. I have no doubt "limit" is working for TcpConnection.
[...]

PR #1831, opened 2017-04-21
WIP 1797 timeout regression

MRigal:
## What do these changes do?

Trying first to write a solid failing test case to illustrate #1797 
Not stable for now, but still maybe worth looking at the results

Nope

## Related issue number

Not good to merge for now, aiming at #1797 

## Checklist

### DIFF
--- a/tests/test_py35/test_client.py
+++ b/tests/test_py35/test_client.py
@@ -118,7 +118,8 @@ async def test_aiohttp_request(loop, test_server):
     (100, None),
     (200, None)
 ])
-async def test_no_timeout_using_semaphore(loop, test_server, semaphore_value, conn_limit):
+async def test_no_timeout_using_semaphore(loop, test_server,
+                                          semaphore_value, conn_limit):
 
     async def handler(request):
         resp = web.Response()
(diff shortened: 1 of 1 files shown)


fafhrd91:
did you read my comment on actual issue?

could you try to shield actual session.get() call.

MRigal:
I'm actually not very sure what you mean with it... Would you like to submit a PR to my branch or just write down code-wise what you are meaning?

MRigal:
OK, I am not able to write a proper test case simulating the full functionality of a proxy, but I could put in evidence on my server that I had to go through a proxy to be able to reproduce the problem. Therefore I updated the issue. Still, downgrading to 1.3.x simply fixes the problem

MRigal:
The report of the travis 3.6 is pretty complete (https://travis-ci.org/aio-libs/aiohttp/jobs/225670002):
```
       6 failed
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[pyloop-10-None]
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[pyloop-3-1000]
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[pyloop-10-1000]
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[uvloop-10-None]
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[uvloop-3-1000]
         - tests/test_py35/test_client.py:104 test_timeout_using_conn_limit[uvloop-10-1000]
```
While applying the same tests on the 1.3 branch will only cause the `*loop-*-1000` to fail... Which shows a behaviour difference (regression?) when not touching the `limit` parameter on the 2.x branch

fafhrd91:
here is difference how aiohttp handles timeout.

1.3 - timeout get applied only to connection + send request, after that there is no timeout on operation
2.0 - timeout get applied to all operations related to request, connection, send request, get response.

in your test you create all connections at the same time, so aiohttp tries to connect and start timeout at the same time. but remote service can not serve you requests fast enough, so at some point you start to see timeout errors. I start to see timeouts around 450 requests, so aiohttp able to complete about 450 request within first 4 seconds. and that is the reason why you can not reproduce this with local server, it just able to serve all your requests within 4 seconds.

also `asyncio.gather` does not protect you from exceptions, (shield() does)
lets say you try to gather 100 requests (coroutines), but if one of those raises exception (TimeoutError in your case) it cancels `asyncio.gather()` operation. for crawler case you can not really just send request and wait response, you have to handle exceptions from requests, bad responses, etc and return result.
generally, you have to "shield" request if you want to use it with `gather()`

fafhrd91:
@MRigal ping

fafhrd91:
@MRigal ?

tests/test_py35/test_client.py
```diff
@@ -99,3 +99,42 @@
...
+    conn_kwargs = dict(loop=loop)
+    if conn_limit is not None:
+        conn_kwargs['limit'] = conn_limit
+    connector = aiohttp.TCPConnector(**conn_kwargs)
+    async with aiohttp.ClientSession(connector=connector,
+                                     loop=loop) as session:
+        for i in range(connector._limit * rows_multiplier):
+            rows.append(asyncio.ensure_future(
+                read_resp(session, server), loop=loop))
+        await asyncio.gather(*rows)
```
AraHaan, line 140: Use asyncio.shield() here. This is what @fafhrd91 also suggested above.

PR #1834, opened 2017-04-22
add secure_proxy_ssl_header to Application init

### LINKED ISSUE #1762: secure_proxy_ssl_header with gunicorn
## Long story short

I think it's not possible to set `secure_proxy_ssl_header` when deploying with gunicorn.

## Expected behaviour

gunicorn config or environment variable should allow me to set `secure_proxy_ssl_header='X-Forwarded-Proto'` so `request.scheme` is correct.

## Actual behaviour

No apparent way to set it.

## Steps to reproduce

Can give more detail if required.

## Your environment

Heroku, but should be the same with any other.

samuelcolvin:
fix #1762

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -42,10 +42,15 @@ __all__ = (web_protocol.__all__ +
 
 
 class Application(MutableMapping):
-
-    def __init__(self, *, logger=web_logger, router=None, middlewares=(),
-                 handler_args=None, client_max_size=1024**2,
-                 loop=None, debug=...):
+    def __init__(self, *,
+                 logger=web_logger,
+                 router=None,
+                 middlewares=(),
+                 handler_args=None,
+                 client_max_size=1024**2,
+                 secure_proxy_ssl_header=None,
+                 loop=None,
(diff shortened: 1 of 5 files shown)


fafhrd91:
whats about 'X-Forwarded-Proto' header? should we check for it as well?

samuelcolvin:
you mean on `.secure`? I thought `.scheme` already did that?

fafhrd91:
oh, np. I missed code in _schema

fafhrd91:
could you fix spelling in web_reference, 

https://travis-ci.org/aio-libs/aiohttp/jobs/224700291

fafhrd91:
btw do we need special treatment for "wss:" ?

samuelcolvin:
I don't think so, I assume `schema` will be `https` for `wss` (which is arguably wrong) but `secure` would then be correct.

From the docs:

> The scheme is 'https' if transport for request handling is SSL or secure_proxy_ssl_header is matching.

fafhrd91:
ok, good.

LGTM

PR #1836, opened 2017-04-23
Choose addr cached based on a round robin strategy

pfreixes:
At each attempt to open a new connection, the addrs related to that
specific host will be retrieved using a round robin strategy. As a
result all hosts resolved by the DNS query will be used.

## What do these changes do?

Enable the use of all addresses resolved for a specific host name. The current implementation only falls to the next host if and only if the first one fails. Taking into account that DNS entries returned do not have weights all addresses must be eligibles.

No

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -6,6 +6,8 @@ import traceback
 import warnings
 from collections import defaultdict
 from hashlib import md5, sha1, sha256
+from itertools import cycle, islice
+from time import monotonic
 from types import MappingProxyType
 
 from . import hdrs, helpers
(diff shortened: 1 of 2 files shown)


fafhrd91:
lgtm. 

@kxepal any comments?

fafhrd91:
@asvetlov comments?

kxepal:
Looks good since it's totaly private thing, but I wonder if we need to have something like this rather than suggest to install some local dns cache service instead, like dnsmasq, systemd or whatever. Windows has it own OS local DSN cache. Double caching problem may be not good.

kxepal:
I can tell you a story about Hadoop namenode which uses sort of local dns cache as well and how this turns things into hell when you need to migrate some nodes to another zone.

asvetlov:
Looks good.
People would want to use the option on early development stages.
What bother me is missing new ctor parameters in http://aiohttp.readthedocs.io/en/stable/client_reference.html#tcpconnector

@pfreixes would you fix it?
It could be done by another PR

fafhrd91:
@pfreixes thanks!

pfreixes:
Hi, first thanks for the reviewing to all of  you @fafhrd91 @asvetlov and @kxepal 

Regarding the constructor param in the `TCPConnector` remember that this is a replacement of the old cache system that didn't have a way to use all `addrs` during the happy path execution, therefore IMHO the `use_dns_cache` is, in fact, the related param to enable or disable this feature. Or were you thinking in another case?

@kxepal The idea behind this it's put some continuation in this other work [1], having an expiration of the  DNS table with an enough low value and having the chance to use all `addr` resolved by a host you will be in the best position to use services that have in front LBs that use the DNS as a way scale up or scale down. Take as an example the case of the ELBs orchestrated by AWS that you use in your daily work, they behave as it has been explained.

There are examples that show you how the software [2] spent some time to react to the change in the expiration latency, IMHO the app layer has to be ready to live in an environment where the resolutions can change quickly.

[1] https://github.com/aio-libs/aiohttp/pull/1819
[2] http://discourse.haproxy.org/t/trouble-with-dns-resolvers-sticking-to-single-ip-address/146/4

kxepal:
@pfreixes 
Suddenly (or fortunately?) , I don't and never used AWS things, but I feel that there is a real problem that this could solve for some cases. Thank you for the explanation! This became more clear for me.

PR #1849, opened 2017-04-27
Add client side certificates verification example to docs

omerbenamram:
I've spent some time figuring this out on my own (and have written an answer to the only other guy on SO that asked himself the same question http://stackoverflow.com/questions/41701791/aiohttp-and-client-side-ssl-certificates/43664652#43664652)

Requests documentation is pretty straightforward about this http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates

I think this will be a useful example for future users of this feature (without any high-level API changes like requests implemented).

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -505,6 +505,17 @@ pass it into the connector::
   session = aiohttp.ClientSession(connector=conn)
   r = await session.get('https://example.com')
 
+If you need to verify **client-side** certificates, you can do the same thing as the previous example,
+but add another call to ``load_cret_chain`` with the key pair::
+
+  sslcontext = ssl.create_default_context(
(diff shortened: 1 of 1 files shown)


fafhrd91:
great! thanks!

MerlijnWajer:
Hi, sorry for not creating a new issue, but:

* documentation states `load_cret_chain` (cret, not cert)
* the code example uses 'pub.pem' and 'priv.pem', suggesting the two arguments are keys. This is not true, as per Python documentation, the first argument is the certificate, and the (optional) second argument is the privkey.

asvetlov:
Also aiohttp master accepts `ssl_context` parameter in methods like `client.get()` etc.
I think better to use this approach in docs: creation a custom `TCPConnector` is overhead in this particular case.

PR #1869, opened 2017-05-07
Add a parameter to optionally not close loop on exit

EliRibble:
The issue here is that some clients, like me, may have other cleanup
they need to do to gracefully shut down other coroutines outside of the
aiohttp server request handler coroutines. Because run_app closes the
loop those coroutines that are running the same loop but are outside of
aiohttp's control will immediately emit ugly stack traces because they
are attempting to continue operating on a closed loop.

By adding a parameter that controls closing the loop clients like me can
simply disable closing the loop and do our own additional cleanup before
closing the loop ourselves

## What do these changes do?

Allow users to disable closing the event loop on KeyboardInterrupt

No, the default behavior remains the same, it just gives users the option, if they want it, of changing the behavior

## Related issue number

No

## Checklist

- [] Unit tests for the changes exist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -313,8 +313,9 @@ class Application(MutableMapping):
 def run_app(app, *, host=None, port=None, path=None, sock=None,
             shutdown_timeout=60.0, ssl_context=None,
             print=print, backlog=128, access_log_format=None,
-            access_log=access_logger, loop=None, close_loop=True):
+            access_log=access_logger, loop=None):
     """Run an app locally"""
+    user_supplied_loop = loop is not None
     if loop is None:
         loop = asyncio.get_event_loop()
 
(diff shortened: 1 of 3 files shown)


fafhrd91:
I am fine with this change. 

@kxepal @asvetlov any comments?

aiohttp/web.py
```diff
@@ -421,7 +421,8 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
         loop.run_until_complete(app.shutdown())
         loop.run_until_complete(handler.shutdown(shutdown_timeout))
         loop.run_until_complete(app.cleanup())
-    loop.close()
+    if close_loop:
```
kxepal: I think there should be a check if loop was explicitly passed externally instead of any kind of a control flag.

Why? Because, if I pass the loop explicitly from outside, I control it from the outside and I have to take care about it from the outside, where I should close it whenever it's appropriate. If it's implicit loop created within the `run_app` function, then it's responsible to close it when it no longer need in it as it only owns it and nobody else.

kxepal:
LGFM except one design bit. May be it worth to pay attention on it if it suites the case.

aiohttp/web.py
fafhrd91: I think this is very good idea. instead of extra parameter, let's just close owned loop. if it come from outside just do nothing.

EliRibble:
Alright, let me see if I can hack that together

fafhrd91:
thanks!

PR #1872, opened 2017-05-08
Fix note block wording and position

kxrd:
Fix wording of the note block instructing users to use the streaming response if lots of data is going to be read; move it down a notch for better positioning and making it less confusion-prone.


fafhrd91:
thanks

fafhrd91:
could you rebase against master

Issue #1134, opened 2016-08-27
Support X-Forwarded-* and Forwarded implicitly, deprecate secure_proxy_ssl_header.


asvetlov:
It's not obvious and the current doc is misleading: it proposes `secure_proxy_ssl_header='X-Forwarded-Proto'` but should do `secure_proxy_ssl_header=('X-Forwarded-Proto', 'https')`.

Looks like @popravich made a mistake on documenting it by d4954ef4
1. `X-Forwarded-For`, `X-Forwarded-By`, and  `X-Forwarded-Proto` should be supported. See the spec at http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html
2. `Forwarded` header from RFC 7239 https://tools.ietf.org/html/rfc7239 should be supported as well.

The issue doesn't require a single Pull Request, most likely it should be several PRs, each for other aspect for sake of easy reviewing etc.

Until we'll remove deprecated `secure_proxy_ssl_header` support it should be processed first before all other rules.

evertlammerts:
What's the status on this? I'm going to use:

- [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)
- [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto), [X-Forwarded-Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host) and [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)
- [Host](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)

... to try to determine how & where my service can be reached during runtime. I can do this in my own code but I can also shoot in a PR. How about this approach:

- add the above headers to [aiohttp.hdrs](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/hdrs.py)
- use them in the above order in `Request._scheme`, `Request.host` and `Request.url` in an attempt to determine the correct vales
- explain resolution order in the docs

Would that be useful enough to be included?

fafhrd91:
@evertlammerts PR would be very good!

evertlammerts:
How about something like this? There are no tests in there yet and I'm not sure how deprecation works in aiohttp. Any pointers?

I'll revisit soon, just about to leave on a trip.

[...]

PR #1881, opened 2017-05-11
Use Forwarded, X-Forwarded-Scheme and X-Forwarded-Host for better scheme and host resolution

evertlammerts:
## What do these changes do?

They enhance the resolution mechanism for scheme and host in `aiohttp.web_request.Baserequest`.

The scheme is resolved based on:

* the currently used transport
* the `secure_proxy_ssl_header` argument of `BaseRequest.__init__`
* the `Forwarded` header
* the `X-Forwarded-Proto` header

in that order.

The host is resolved based on:

* the `Forwarded` header
* the `X-Forwarded-Host` header
* the `Host` header

in that order.

This PR does not include changes in the contract.

## Related issue number

#1134 Support X-Forwarded-* and Forwarded implicitly, deprecate secure_proxy_ssl_header.

## Checklist


CHANGES.rst
```diff
@@ -41,6 +41,8 @@ Changes
 
 - Do not unquote `+` in match_info values #1816
 
+- Use Forwarded, X-Forwarded-Scheme and X-Forwarded-Host for better scheme and host resolution. #1134
```
AraHaan: PEP8 here. Please split this to 2 lines or more of up to 79 characters per line. For more information on PEP8 see the www.python.org page on PEP8.

asvetlov:
Implementation looks good on fist glaze.
Let me reread RFC before merging.
I'll add required deprecation records after merging if @evertlammerts has troubles with it.

evertlammerts:
does c2848ed cover the deprecation of `secure_proxy_ssl_header`?

evertlammerts:
Ping @asvetlov

asvetlov:
Sorry, I'm on US PyCon right now and quite busy.

asvetlov:
https://tools.ietf.org/html/rfc7239 says that `Forwarded` header could be present several times but the very first occurrence points to host which initiates the request.
@evertlammerts could you add an explicit test for it?

aiohttp/web_request.py
```diff
@@ -151,15 +151,30 @@ def secure(self):
         return self.url.scheme == 'https'
 
     @reify
+    def _forwarded(self):
+        forwarded = self._message.headers.get(hdrs.FORWARDED)
+        if forwarded is not None:
+            parsed = re.findall(
+                r'^by=([^;]*); +for=([^;]*); +host=([^;]*); +proto=(https?)$',
```
asvetlov: I suspect the parser should be more complicated.
```
The header field value can be
   defined in ABNF syntax as:

       Forwarded   = 1#forwarded-element

       forwarded-element =
           [ forwarded-pair ] *( ";" [ forwarded-pair ] )

       forwarded-pair = token "=" value
       value          = token / quoted-string

       token = <Defined in [RFC7230], Section 3.2.6>
       quoted-string = <Defined in [RFC7230], Section 3.2.6>
```
It means the `forwarded-pair` elements could be in random order: `by=a; host=b.com` and `host=b.com, by=a` are both valid combinations.
Also value could be optionally quoted, aiohttp shoud unquote it automatically. 
pair-name is case-insensitive: all 'Host', 'host', and 'HOst` are valid names:
```
       Forwarded: for="_gazonk"
       Forwarded: For="[2001:db8:cafe::17]:4711"
       Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43
       Forwarded: for=192.0.2.43, for=198.51.100.17

   Note that as ":" and "[]" are not valid characters in "token", IPv6
   addresses are written as "quoted-string".
```

asvetlov: Please note: the lase example contains two `for` pairs, it's not supported by your regexp now.

evertlammerts:
:blush: I should have been way more thorough. Sorry about this, I will go
fix. Thanks for looking into it!

evertlammerts:
Alright, this actually adds a public `forwarded` property to `BaseRequest`. It makes an effort to parse Forwarded headers as specified by the RFC:

- It adds all parameters (by, for, host, proto) in the order it finds them; starting at the topmost / first added 'Forwarded' header, at the leftmost / first-added parameter.
- It checks that the value has valid syntax in general as specified in section 4: either a 'token' or a 'quoted-string'.
- It un-escapes found escape sequences.
- It does NOT validate 'by' and 'for' contents as specified in section 6.
- It does NOT validate 'host' contents (Host ABNF).
- It does NOT validate 'proto' contents for valid URI scheme names.

Don't you think the place for parsing the Forwarded header is rather a middleware? I'm in doubt myself now.

Otherwise, what's a good place for the regex constants? Would be nice to compile the regex only once, but module level doesn't do much for esthetics.

aiohttp/web_request.py
```diff
@@ -21,6 +22,35 @@
...
+_QUOTED_PAIR = r'\\[\t {tchar}]'.format(tchar=_TCHAR)
+
+_QUOTED_STRING = r'"(?:{quoted_pair}|{qdtext})*"'.format(
+    qdtext=_QDTEXT, quoted_pair=_QUOTED_PAIR)
+
+_FORWARDED_PARAMS = (
+    r'[bB][yY]|[fF][oO][rR]|[hH][oO][sS][tT]|[pP][rR][oO][tT][oO]')
+
+_FORWARDED_PAIR = (
+    r'^ *({forwarded_params})=({token}|{quoted_string}) *$'.format(
```
asvetlov: IIRC spaces around `=` are allowed.

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+                    forwarded_pairs = tuple(
+                        _FORWARDED_PAIR_RE.findall(pair)
+                        for pair in forwarded_elm.split(';'))
+                    for forwarded_pair in forwarded_pairs:
+                        if len(forwarded_pair) != 1:
+                            # non-compliant syntax, ignore
+                            continue
+                        param = forwarded_pair[0][0].lower()
+                        value = forwarded_pair[0][1]
+                        if len(value) and value[0] == '"':
```
asvetlov: Add a check for `value[-1] == '"'`

asvetlov: If empty value allowed? I mean `len(value) == 0`.
Also `if value:` gives the same result as `if len(value)` but works a bit faster.

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+        - It un-escapes found escape sequences.
+        - It does NOT validate 'by' and 'for' contents as specified in section
+          6.
+        - It does NOT validate 'host' contents (Host ABNF).
+        - It does NOT validate 'proto' contents for valid URI scheme names.
+
+        Returns a dict(by=tuple(...), for=tuple(...), host=tuple(...),
+        proto=tuple(...), ) 
+        """
+        params = {'by': [], 'for': [], 'host': [], 'proto': []}
```
asvetlov: I believe the property's type should be `MultyDict` instead of dict of lists.
It preserves all information about proxies in right order but original host still could be checked as `req.forwarded.get('host')`

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+                            continue
+                        param = forwarded_pair[0][0].lower()
+                        value = forwarded_pair[0][1]
+                        if len(value) and value[0] == '"':
+                            # quoted string: replace quotes and escape
+                            # sequences
+                            value = _QUOTED_PAIR_REPLACE_RE.sub(
+                                r'\1', value[1:-1])
+                        params[param].append(value)
+        return params
```
asvetlov: return `MultiDictProxy(params)` -- it's an immutable view.

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+        - It un-escapes found escape sequences.
+        - It does NOT validate 'by' and 'for' contents as specified in section
+          6.
+        - It does NOT validate 'host' contents (Host ABNF).
+        - It does NOT validate 'proto' contents for valid URI scheme names.
+
+        Returns a dict(by=tuple(...), for=tuple(...), host=tuple(...),
+        proto=tuple(...), )
+        """
+        params = {'by': [], 'for': [], 'host': [], 'proto': []}
```
asvetlov: Use `MultiDict` here instead of dict of pairs. It preserves all information about proxies.

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+        if hdrs.FORWARDED in self._message.headers:
+            for forwarded_elm in self._message.headers.getall(hdrs.FORWARDED):
```
asvetlov: Drop previous line, `in` is not free but `getall(hdrs.FORWARDED, ())` does the check using single call to data structure.

aiohttp/web_request.py
evertlammerts: I'm not sure but as far as I can see that's not the case. Section 4 only says `forwarded-pair = token "=" value` and section 7 recalls `Note that an HTTP list allows white spaces to occur between the identifiers, and the list may be split over multiple header fields.`

aiohttp/web_request.py
```diff
@@ -151,15 +181,60 @@ def secure(self):
...
+        - It does NOT validate 'proto' contents for valid URI scheme names.
+
+        Returns a dict(by=tuple(...), for=tuple(...), host=tuple(...),
+        proto=tuple(...), ) 
+        """
+        params = {'by': [], 'for': [], 'host': [], 'proto': []}
+        if hdrs.FORWARDED in self._message.headers:
+            for forwarded_elm in self._message.headers.getall(hdrs.FORWARDED):
+                if len(forwarded_elm):
+                    forwarded_pairs = tuple(
```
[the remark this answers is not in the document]
evertlammerts: This is supported by the regexes and I've included tests for it

aiohttp/web_request.py
evertlammerts: There's no need to check for a closing quote: the regex validates `quoted-string` so any value starting with `"` must also have a closing `"`.

A `token` may be empty (but note again that I don't check the syntax in section 6)

evertlammerts:
I forgot Python < 3.5 doesn't like an [arbitrary number of unpackings](https://www.python.org/dev/peps/pep-0448/) (`*iterable`). Just to be certain, is Python >= 3.4.2 still the requirement?

aiohttp/web_request.py
asvetlov: IMHO it's rare case but still allowed

asvetlov:
Yes, we sill support Python 3.4.2+
Will drop it soon but not right now.
It will be trivial but huge change -- replace all `yield from` with `await` :)

evertlammerts:
I seem to have missed a couple of things:

- field values are separated by `comma`: my implementation rather assumes that each header has a single field value in which each parameter may occur multiple times (i.e. I allow `host=abc; host=def;` which is explicitly not allowed and should be `host=abc, host=def`)
- MultiDict maintains order between values, so I can just use `md.add()`

Regarding OWS around `=`, I can't find examples where that is allowed. Actually, I think it isn't allowed around `;` either, or at least the ABNF doesn't include OWS or RWS. (Another bug in my regexes.) The only place where it seems to be allowed is the field-value separator (`,`).

I looked for examples of whitespace around `=` in other places as well. I can only think of the `Content-*` (`charset=...`) and `Accept`(`q=...`) headers as ones that also accept key-value pairs and RFC 7231 explicitly doesn't allow whitespace. Why do you think it is allowed? Maybe something to do with clients not being compliant?

I'll do my best to add some more commits later this week. When are you planning the next release, @asvetlov?

asvetlov:
Ok, I'm convinced regarding to spaces.
But commas and proper multidict usage (`md.add(k, v)`) should be fixed.
There is no scheduled date for the release.
I love to have this issue in next version and could wait for you for a while if @fafhrd91 agrees with me.

fafhrd91:
@asvetlov it is your call. I am busy with different project in any case

evertlammerts:
d1ba86b fixes the last issues, as far as I can see.

- It is now possible to provide multiple field-values (comma separated list) in a single Forwarded header. See the tests / code in d1ba86b for the supported semantics.
- I've fixed the unpacking issue. The patch should be compatible with Python 3.4 now.
- I've fixed an issue with the quoted-pairs regex, for quoted-strings (ie `host="something \"escaped\""` now works correctly)

As far as the return type goes, I've changed that to a tuple containing MappingProxies. The semantics of the Forwarded header are so that every proxy may add a Forwarding field-value; `by`, `for`, `host` and `proto` in any combination, but _not_ multiple times per field-value. When a proxy adds a new Forwarding field-value, it either appends it to an already existing Forwarding header (after a comma) or it adds a new Forwarding header at the bottom. I think it's useful to maintain the information about the order in which proxies added something. So now the result will look something like:

```
req.forward = (
    MappingProxy({'by': ..., 'for': ...}),  # added by proxy 1
    MappingProxy({'host': '...', 'for': ..., 'proto': ...}),  # added by proxy 2
    # ... etc
)
```

If a field-value added by a proxy cannot be parsed, the code will add a `MappingProxy(dict())` so that users can still see that something is there, and that it was added by a proxy at that point in the chain.

Let me know what you think, @asvetlov.

asvetlov:
Awesome work @evertlammerts 
Very thank you!

adamcharnock:
For anyone else chasing their tail trying to figure out where this feature has gone, it has been removed and moved into middleware. See here:

https://github.com/aio-libs/aiohttp/issues/2171#issuecomment-346689565

Issue #1883, opened 2017-05-11
ProcessPoolExecutor in background cause freeze after keep-alive timeout


voronin-opensoft:
## Long story short

If we run heavy task in `ProcessPoolExecutor` in background, `aiohttp` don't handle request after some timeout. It happens after ~75 seconds. It's equal `keepalive_timeout` value at https://github.com/aio-libs/aiohttp/blob/2.0.7/aiohttp/web_protocol.py#L82
If we change `keepalive_timeout` value we get freeze after specific time.

## Expected behaviour

Handle all requests.

## Actual behaviour

Freeze on request after keep-alive timeout.

## Steps to reproduce

__We set keep-alive timeout to 1 second for faster bug reproducing.__

1. Run server
```python
import asyncio
import logging
from multiprocessing import Manager
from concurrent.futures import ProcessPoolExecutor

from aiohttp.web import Application, Response, run_app


logging.basicConfig(level=logging.DEBUG)


async def index(request):
    body = '''<!DOCTYPE html>
    <html>
    <body>
    <script>
        function request(path) {
            var xhr = new XMLHttpRequest();
            
            xhr.open('GET', path, false);
            xhr.onload = function (e) {
                if (xhr.readyState == 4 && xhr.status == 200) {
                    console.log(xhr.responseText);
                }
            }
            xhr.send(null);
        }
        
        function start() {
            return request('/start');
        }
        
        function stop() {
            return request('/stop');
        }
    </script>
    <button style="width: 200px; height: 100px;" onclick="start()"> Start </button>
    <button style="width: 200px; height: 100px;" onclick="stop()"> Stop </button>
    </body>
    </html>
    '''
    return Response(body=body, content_type='document')


async def start(request):
    await request.app['worker'].queue.put('data')
    return Response(body='Start')


async def stop(request):
    request.app['worker'].stop_event.set()
    return Response(body='Stop')


def hard_work(stop_event):
    while True:
        if stop_event.is_set():
            break


class Worker:

    async def start(self, loop):
        self.queue = asyncio.Queue(loop=loop)

        while True:
            data = await self.queue.get()

            executor = ProcessPoolExecutor()
            manager = Manager()
            stop_event = manager.Event()
            self.stop_event = stop_event

            hard_future = loop.run_in_executor(executor, hard_work, stop_event)

            try:
                await asyncio.wait_for(hard_future, timeout=999999, loop=loop)
            except asyncio.TimeoutError:
                print('Time out')
[...]

PR #1884, opened 2017-05-12
Write to transport \r\n before closing after keepalive timeout #1883

fafhrd91:


fafhrd91:
@kxepal @asvetlov comments?

aiohttp/web_protocol.py
```diff
@@ -268,6 +268,7 @@ def eof_received(self):
         pass
 
     def data_received(self, data):
+        print("data:", data)
```
kxepal: print

aiohttp/web_protocol.py
```diff
@@ -336,13 +337,15 @@ def close(self):
             if not waiter.done():
                 waiter.cancel()
 
-    def force_close(self):
+    def force_close(self, keepalive=False):
```
kxepal: I think the argment name a bit misleading. `send_last_heartbeat` - might be more precisely about what will happen. The `keepalive` fits more to class property which can be enabled or disabled.

fafhrd91: good catch, thanks

PR #1886, opened 2017-05-13
Fix examples/docs for websocket send commands being a coroutine

balloob:
## What do these changes do?

Version 1.3 made websocket send commands awaitable: b3c80ee3f7d5d8f0b8bc27afe52e4d46621eaf99. However that change did not update tests/examples. This PR finishes that work.

It went unnoticed because when a send method is called, the bytes are actually written to the socket and a coroutine is returned for the drain command.

Noticed it because setting `PYTHONASYNCIODEBUG` to 1 started printing errors about unawaited coroutines.

No

## Related issue number

No

## Checklist

### DIFF
--- a/demos/chat/aiohttpdemo_chat/views.py
+++ b/demos/chat/aiohttpdemo_chat/views.py
@@ -19,11 +19,11 @@ async def index(request):
     name = (random.choice(string.ascii_uppercase) +
             ''.join(random.sample(string.ascii_lowercase*10, 10)))
     log.info('%s joined.', name)
-    resp.send_str(json.dumps({'action': 'connect',
-                              'name': name}))
+    await resp.send_str(json.dumps({'action': 'connect',
+                                    'name': name}))
     for ws in request.app['sockets'].values():
-        ws.send_str(json.dumps({'action': 'join',
-                                'name': name}))
+        await ws.send_str(json.dumps({'action': 'join',
+                                      'name': name}))
     request.app['sockets'][name] = resp
 
     while True:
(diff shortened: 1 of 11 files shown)


asvetlov:
Thanks!

balloob:
When I started yielding from sending messages I quickly ran into [Python bug 29930](http://bugs.python.org/issue29930): 2 drains in parallel will raise an `AssertionError` error.

Do you think that we should add this to the aiohttp docs?

AraHaan:
Yeah, this is exactly why I use the latest commit of 3.6.1+ by building python from source every day for 32 and 64 bit windows. Also all the more reason to use 'async def' and 'await' as well. I also can't wait for the day when python 3.4.x reaches no further patches by EOL so that way aiohttp can drop python 3.4 and go full 'async def' and 'await'.

asvetlov:
I think the main showstopper for dropping Python 3.4 support is absence of 3.5 in Debian stable.

AraHaan:
Yeah, if I was part of Debian development I would change that to latest commit of 3.6.1+.

After all I do not think 3.6.1+ from github would not break most things for 3.4. But it would allow people to use 'async def' and 'await' on the fly without needing for them to manually compile 3.5 or 3.6 to begin with.

thehesiod:
@asvetlov stretch shipped, IIRC it's 3.5

asvetlov:
Yes, I know.
We should give people several months for migrating on Debian Stretch.
Maybe Autumn is good time for shipping aiohttp 3.0 with dropped Python 3.4 support.

@balloob do you know plans for Raspberry PI?

balloob:
Raspberry PI is expecting a Debian Stretch based release over the summer. Last time it was ~5 months after the Debian stable release.

([source](https://www.raspberrypi.org/blog/a-raspbian-desktop-update-with-some-new-programming-tools/) - last paragraph "One final thing")

PR #1900, opened 2017-05-17
aiohttp/http_parser.py: pure-python fix for big chunk decoding (#1899)

### LINKED ISSUE #1899: Big chunk decoding makes content.iter_any() iterate over partial content
## Long story short

When receiving a very big chunk (response has `Transfer-Encoding: chunked.` header), result of `aiohttp.ClientSession().request(...).content.iter_any()` yields partial chunks.

## Expected behaviour

I would expect chunks integrity to be respected, and `iter_any()` to yield reasonably-decodable json content.

## Actual behaviour

`iter_any()` yields partial chunks, which is Bad(tm).

## Steps to reproduce

I think this basic code below is enough if you find a server that answers big chunks. We internally managed to reproduce the bug by querying a docker server with URL `GET /v1.24/containers/json?all=1 HTTP/1.1.`. Server would reply a very big 350 000 bytes long chunk, and instead of yielding that chunk iter_any() just served us an about-30k-bytes-long beginning of chunk.

```python
response = await aiohttp.ClientSession().request(...)
async for chunk in response.content.iter_any():
    json.loads(chunk)
```

If ValueError is raised, you lose the game.

## Your environment
Ubuntu Xenial (python 3.5.2) with aiohttp 2.0.7 (bug is reproducible on aiohttp master branch).

(update: updated code sample)

Anvil:
## What do these changes do?

It should be a pure-python fix #1899 
The .pyx implementation is not done. Not sure I'll be able to do that.

response.content.iter_any() will not yield partial chunk

## Related issue number

#1899 

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,9 @@ Changes
 
 - Make test server more reliable #1896
 
+- Pure-python-only fix for partial chunk decoding #1899
+
+
 2.0.7 (2017-04-12)
 ------------------
 
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -535,14 +535,20 @@ class HttpPayloadParser:
                     required = self._chunk_size
                     chunk_len = len(chunk)
 
-                    if required >= chunk_len:
-                        self._chunk_size = required - chunk_len
-                        if self._chunk_size == 0:
-                            self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
-
+                    if required == chunk_len:
+                        # We have read just enough data, feed decoded
+                        # chunk, wait for chunk eof and return.
+                        self._chunk_size = 0
+                        self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
                         self.payload.feed_data(chunk, chunk_len)
                         return False, None
+                    if required > chunk_len:
+                        # We have not read enough data, wait for more data.
+                        self._chunk_tail = chunk
+                        return False, None
                     else:
+                        # We have read more data than required, feed
+                        # decoded chunk and continue
                         self._chunk_size = 0
                         self.payload.feed_data(chunk[:required], required)
                         chunk = chunk[required:]
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -535,24 +535,23 @@ class HttpPayloadParser:
                     required = self._chunk_size
                     chunk_len = len(chunk)
 
+                    if required > chunk_len:
+                        # We have not read enough data, wait for more.
+                        self._chunk_tail = chunk
+                        return False, None
+
+                    self._chunk_size = 0
+                    self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
                     if required == chunk_len:
                         # We have read just enough data, feed decoded
                         # chunk, wait for chunk eof and return.
-                        self._chunk_size = 0
-                        self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
-                        self.payload.feed_data(chunk, chunk_len)
-                        return False, None
-                    if required > chunk_len:
-                        # We have not read enough data, wait for more data.
-                        self._chunk_tail = chunk
+                        self.payload.feed_data(chunk, required)
                         return False, None
                     else:
                         # We have read more data than required, feed
                         # decoded chunk and continue
-                        self._chunk_size = 0
                         self.payload.feed_data(chunk[:required], required)
                         chunk = chunk[required:]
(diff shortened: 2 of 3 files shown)


Anvil:
I can rebase on 2.0 if required.

fafhrd91:
could you add test?

Anvil:
I'm trying...

Anvil:
OK, this is the best I could come up with. I know it does the job, I dont know if it's minimalist enough.

Anvil:
The failure in the tests in the C parser failing, not the python one.

fafhrd91:
I don't think this is good design choice. This will load whole chunk into memory before processing, so it would be very easy to flood server memory.

@asvetlov @kxepal comments?

kxepal:
I agree with concern about memory footprint. And I guess this feature should be done for both parsers in anyway in order to preserve similar behaviour.

As about the problem, I think here is the server who should be fixed in order to not send too big chunks. Otherwise, there could be some magic constant which user explicitly will change in order to receive big chunks. But certainly, that's not good idea to have support of those by default for everyone without any way to disable it.

asvetlov:
I don't think this is the bug at all.
Chunked encoding is just transport protocol only, semantically it's a sequence of bytes without boundaries. Just like a TCP socket is semantically a stream of bytes while the content is transferred by packets.

From my perspective partial chunks are totally correct.
Thoughts?

asvetlov:
Moreover proxy *could* change transfer encoding, it even could compress/decompress the body.
End user's code should not rely on it.

fafhrd91:
I agree, developer should not rely on transfer-encoding

asvetlov:
@Anvil do you have a strong reason for supporting this behavior.
I'm inclining to closing the issue as 'no a bug'

Anvil:
(Ok for reference and reminder, chunk encoding is
\<hex size\>\r\n
\<content of given size\>\r\n
[repeat size + content scheme]
0\r\n
\r\n
EOT
)

I think that you cannot "fix" servers and tell them "dont send big chunk". No offense, but the choice is yours to make.

Now about the bug itself. Because it _is_ a bug. This is not about big chunks, this is about partial chunks.

Thing is, aiohttp client/parser detects content is chunk-encoded (through HTTP headers) and starts interpreting the data by catching it and by _stripping the size information from the communication_. In case two chunks are received at once, aiohttp knows how many bytes to expect, nicely performs the split and the `content.iter_any()` successively yields the first and then the second chunk.

Seeing that, we thought that it was aiohttp job to handle all cases and to return us chunks. Not raw data. Sounded like a very nice feature.

Turned out that when a chunk is partially received, iter_any yields partial content. Annnd I think this defeats the purpose of the generator.
When iter_any yields garbage instead of a chunk, not only we become blind because we dont know the size of the chunk: aiohttp keeps that information for itself but if that can happen once, how can i trust any value yielded by iter_any to be consistent ?
Since, we dont have access to the data size information any more, the only approach left would be "make an educated guess, dude"

This is not exactly a reliable behavior.

IMHO, this patch brings consistency back. At the cost a potentially big memory footprint, yes. My opinion is that if someone wants not to have the whole chunk in memory, (s)he should disable chunk parsing in aiohttp and should implement his/her own way to handle chunks.

Also, if I should not rely in transfer-encoding, then, with all due respect, neither should you, nor the aiohttp client.

(If my opinion matters on the subject, yes, both parsers should implement this.)

Regards.

(Edit: fixed `<>` escaping for markdown)

fafhrd91:
I wouldn't mind adding this feature if it would be possible to limit to one use case, this would affect everyone. Probably in your case it is ok to load everything into memory, but I am sure for most users that would be very big surprise. Also discovering that anyone can kill your service without much effort would be not very pleasant experience with framework

hubo1016:
What about add a max-read-size parameter as the ones in read() functions? We may set it to something like 1MB by default, if the chunk size exceeds the limit, partial result is returned. And if the user set it to None, it means that he is willing to accept the risk. That's the same as open("xxx","r").read().

hubo1016:
And also I would like to remind that for usages that yields multiple JSON objects through chunked-encoded data, it is not necessary to depend on the chunk boundary to split it. JSON *objects* (data surrounded by {}) can be splitted by a state machine e.g. counting the braces in a smart way, this is used in JSON-RPC protocol.
Actually many implementations are just returning JSON objects on different lines, and using readline() is enough. Usually we would consider chunked encoding as an implementation detail of HTTP protocol, but not an attribute of content; if the HTTP is transferred through a proxy, the proxy may break the chunked boundary - because RFC does not say NO.

asvetlov:
I could live with chunk splitting by default but supporting opt-in flag for brave guys who have unlimited memory and don't care about attacks.
Personally I don't feel a strong motivation for providing a fix but will do review/merge if somebody will create a PR (both Python and Cython parts should be fixed at once).

PR #1906, opened 2017-05-19
adding exception & warning logging middleware

samuelcolvin:
## What do these changes do?

Add middleware to log exceptions and warnings.

This is part code, part proposition - aiohttp should make it as easy as possible for people to start getting proper error reports. This allows that.

I'm submitting it just like this to see if you'll consider merging it. If the answer is "yes" I'll add tests and update docs.

It provides middleware which reports exceptions and optionally warnings, it should work with sentry "out of the box" (you'll just need to setup sentry as a log handler) or any other setup which uses python's logging module.

Usage should be as simple as:

```python
# when initialising the app
middleware = (
    ErrorLoggingMiddleware(),
    example_middleware,
    ...
)
app = web.Application(middlewares=middleware)

# when setting up logging

logging.config.dictConfig({
    'version': 1,
    'disable_existing_loggers': False,
    'formatters': {
        ...
    },
    'handlers': {
        'sentry': {
            'level': 'WARNING',
            'class': 'raven.handlers.logging.SentryHandler',
            'dsn': os.getenv('RAVEN_DSN', None),
            'release': os.getenv('COMMIT', None),
            'name': os.getenv('SERVER_NAME', '-')
        },
        ...
    },
    'loggers': {
        'aiohttp.server': {
            'handlers': ['sentry'],
        },
        ...
    },
})
```

(by the way, the plural of middleware is middleware - no "s" on the end).

I've been using middleware very similar to this in production for some time, see [here](https://github.com/tutorcruncher/socket-server/blob/master/tcsocket/app/middleware.py#L50).

If they want - use of `ErrorLoggingMiddleware` is optional.

## Checklist

### DIFF
--- a/aiohttp/web_middlewares.py
+++ b/aiohttp/web_middlewares.py
@@ -1,7 +1,8 @@
 import asyncio
 import re
+import logging
 
-from aiohttp.web_exceptions import HTTPMovedPermanently
+from aiohttp.web_exceptions import HTTPException, HTTPInternalServerError, HTTPMovedPermanently
 from aiohttp.web_urldispatcher import SystemRoute
 
 __all__ = (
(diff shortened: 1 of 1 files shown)


samuelcolvin:
oops, I'll obviously make it 3.4 compatible if you agree with the idea.

fafhrd91:
Lgtm from me, but I am not a middleware user.

@asvetlov ? 
@kxepal ?

aiohttp/web_middlewares.py
```diff
@@ -73,3 +74,63 @@ def middleware(request):
...
+                )
+                if http_exception:
+                    raise http_exception
+                else:
+                    r = await handler(request)
+            except HTTPException as e:
+                if self.should_log_warnings and e.status >= 400:
+                    await self.log_warning(request, e)
+                raise
+            except BaseException as e:
```
kxepal, line 129: I don't think you should eventually catch here KeyboardInteruptException and deny me a chance to stop a server.

samuelcolvin: yes, should probably be `Exception`, not sure why I never had a problem with this.

aiohttp/web_middlewares.py
```diff
@@ -73,3 +74,63 @@ def middleware(request):
...
+                await self.log_exception(e, request)
+                raise HTTPInternalServerError()
```
kxepal, line 131: I guess this exception users might want to customize in some way. Say, I have a service with JSON API which should always return JSON responses. In this case it will break the protocol badly.

kxepal:
While I'm understand the reason and the need to have something like this one, I'm skeptical about to have any of such in aiohttp itself. 

As for sentry, there is nice `aiohttp_sentry` which catches exception, not the logs (actually, I didn't get an idea to send logs to sentry. ELK stack looks better candidate for that). And we may not want to get blocked on attempt to log errors.

I'm -0 here. What would the users say about?

samuelcolvin:
Happy to fix the two comments, just want to see if a complete PR would be accepted before doing all that work.

For me `aiohttp_sentry` isn't acceptable, it doesn't provide sufficient information and uses the much uglier "raven_client" approach. I wrote my own middleware precisely because I didn't want to use aiohttp_sentry.

Using logging is much cleaner because:
* you can capture errors from other services (eg. arq) with the raven config only defined once.
* can unit test error reporting by just changing the log handlers with no ugly mocking.
* As you said, you could send the data anywhere you like - to ELK, to papertrail, to sentry, to email via `SMTPHandler`, to foobar inc. application monitoring which doesn't even exist yet .... One implementation, all the integrations.

I understand (and completely agree with) keeping aiohttp lean, but this seems like a feature which:
1. a high proportion of people will want to use
2. it is important that people have the highest degree of confidence in it working - if your error reporting breaks "you're up **** creek with no paddle".

Up to you, I could release it as a seperate package if you'd prefer.

fafhrd91:
+-0 

@asvetlov should make decision

kxepal:
@samuelcolvin 
Don't get me wrong, I'm not against the feature. I'm just skeptical. Logging and errors processing is too intimate process and everyone does it in own fashion.

> For me aiohttp_sentry isn't acceptable, it doesn't provide sufficient information and uses the much uglier "raven_client" approach. I wrote my own middleware precisely because I didn't want to use aiohttp_sentry.

Heh, same story actually, because lack of customization and too-simple-to-not-implement-it-by-myself (:

asvetlov:
Guys, please don't rush.
I feel your needs but skeptical about solution.
Please let me share my objections a bit later (US PyCon is really a time eater).

fafhrd91:
do we have decision for this PR?

asvetlov:
Sorry, I'm in travel, have very limited free time.
Will return to normal work in a week and half.

samuelcolvin:
Closing in favour of https://github.com/getsentry/raven-python/issues/1124, for me that would be a better way of achieving the same goal.

PR #1910, opened 2017-05-22
Fix #1828: make enable_compression work on HTTP/1.0

### LINKED ISSUE #1828: enable_compression causes exception on HTTP/1.0
## Long story short

call set_compression() on a response object and return it to aiohttp server causes a RuntimeError ("Using chunked encoding is forbidden for HTTP/1.0")

## Expected behaviour

There should be at least two ways to use compression on HTTP/1.0:

1. Fully compress the content, and return the length of the compressed data in the Content-Length header;
2. Do not return Content-Length header, return the compressed stream and set Connection: close. Close the connection after transmit.

## Actual behaviour

An runtime error is raised, and the connection is closed immediately.

In fact, it seems that by default, even a HTTP/1.1 request is responded with a HTTP/1.0 response, so it always raises this exception.

## Steps to reproduce

Call enable_compression on any response object before returning.

## Comment

I think it is absolutely not correct to enable chunked encoding on compression. Content-Encoding is not a part of transfer, it is a part of the content. It should never affect the transfer encoding type.

hubo1016:
## What do these changes do?

This pr fix #1828 i.e. generate a correctly compressed response for HTTP/1.0 requests.

Call enable_compression on **Response** class with **body** set to some bytes now generates a response with content-length instead of chunk-encoded response.

## Related issue number

#1828 

## Checklist

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -298,7 +298,7 @@ class StreamResponse(HeadersMixin):
 
     def _do_start_compression(self, coding):
         if coding != ContentCoding.identity:
-            self._headers[hdrs.CONTENT_ENCODING] = coding.value
+            self.headers[hdrs.CONTENT_ENCODING] = coding.value
             self._payload_writer.enable_compression(coding.value)
             # Compressed payload may have different content length,
             # remove the header
@@ -347,13 +347,13 @@ class StreamResponse(HeadersMixin):
         version = request.version
         writer = self._payload_writer = request._writer
 
-        if self._compression:
-            self._start_compression(request)
-
         headers = self._headers
         for cookie in self._cookies.values():
             value = cookie.output(header='')[1:]
             headers.add(SET_COOKIE, value)
+
+        if self._compression:
+            self._start_compression(request)
 
         if self._chunked:
             if version != HttpVersion11:
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -628,7 +628,8 @@ class Response(StreamResponse):
             self._compressed_body = compressobj.compress(self._body) +\
                 compressobj.flush()
             self._headers[hdrs.CONTENT_ENCODING] = coding.value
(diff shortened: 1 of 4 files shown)


asvetlov:
Looks like you've committed an syntax error :)

hubo1016:
@asvetlov fixed

asvetlov:
Codecov reports that your change is not fully covered by tests.
Albeit it's sometimes acceptable we want to keep our test coverage level.
Please run `make cov-dev` and fix missing cases by adding new tests.

aiohttp/web_response.py
```diff
@@ -355,11 +359,14 @@ def _start(self, request,
...
-            if writer.length is None and version >= HttpVersion11:
-                writer.enable_chunking()
-                headers[TRANSFER_ENCODING] = 'chunked'
-                if CONTENT_LENGTH in headers:
-                    del headers[CONTENT_LENGTH]
+            if writer.length is None:
+                if version >= HttpVersion11:
+                    writer.enable_chunking()
+                    headers[TRANSFER_ENCODING] = 'chunked'
+                    if CONTENT_LENGTH in headers:
```
hubo1016, line 373: This branch cannot be covered in current version because all the **content_length** implementations do not return None when there is a Content-Length header. But I think it is necessary in case there are other implementations in the future. We should not rely on the current implementations.

hubo1016:
@asvetlov Seems that codecov report is not updating, now it is reported as
aiohttp/web_response.py          383      4    186      3    99%

Among the missing statements and branches, only one statement and one branch is related to this pr. The reason is updated in the above comment.

hubo1016:
@asvetlov any response?

asvetlov:
Please create a PR against master.
Also your change fails one test on master branch.

asvetlov:
The last tiny change: move a record in `CHANGES.txt` to 2.3 version

asvetlov:
Ok, I'll update CHANGES in merging.

PR #1915, opened 2017-05-24
Concurrent drains

asvetlov:
asyncio doesn't support concurrent `.drain()` calls (https://github.com/aio-libs/aiohttp/pull/1886#issuecomment-301355495)
This PR proposes a workaround for the issue.

### DIFF
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -4,6 +4,7 @@ import asyncio
 import collections
 import socket
 import zlib
+from collections import deque
 from urllib.parse import SplitResult
 
 import yarl
@@ -39,6 +40,8 @@ class StreamWriter:
         self._waiters = []
         self.available = True
         self.transport = transport
+        self._is_drain = False
+        self._drain_waiters = deque()
 
     def acquire(self, writer):
         if self.available:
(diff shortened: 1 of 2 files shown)


aiohttp/http_writer.py
```diff
@@ -125,7 +128,18 @@ def drain(self):
           yield from w.drain()
         """
         if self._protocol.transport is not None:
-            yield from self._protocol._drain_helper()
+            if self._is_drain:
+                fut = asyncio.Future(loop=self._loop)
```
popravich, line 132: `self._loop.create_future()` I guess

aiohttp/http_writer.py
```diff
@@ -125,7 +128,18 @@ def drain(self):
...
+                self._drain_waiters.append(fut)
+                yield from fut
+            self._is_drain = True
+            try:
+                yield from self._protocol._drain_helper()
+            finally:
+                self._is_drain = False
+                if self._drain_waiters:
+                    fut = self._drain_waiters.popleft()
+                    fut.set_result(None)
```
popravich, line 142: The same "deadlock" issue (a,b) can happen here: if future on line 134 is cancelled then other task that
sets result for next waiter will do nothing and all other futures will be in `_drain_waiters` forever.

a: bugs.python.org/issue27585
b: https://github.com/python/asyncio/pull/393 and propbably https://github.com/python/cpython/pull/1031

popravich: I think cancellation tests should be added as well.

aiohttp/http_writer.py
fafhrd91, line 132: can't we use same future for all callers?

asvetlov:
The code is outdated

Issue #1932, opened 2017-05-30
web.run_app(): Terminate properly with signal TERM and INT


cecton:
## Long story short

This is more or less related to #63. My issue is that `web.run_app()` gracefully exit when SIGINT is sent to the process but not when SIGTERM is sent.

This is because `web.run_app()` catches SIGINT (with `KeyboardInterrupt` exception) but does not override SIGTERM handling (there is no exception raised for that, one can only use `signal.signal`). Most of the time it is SIGTERM that is used to gracefully shutdown a process but that will actually kill the application instead of letting the requests finish properly and response. `kill`, Docker and system.d use SIGTERM by default.

## Expected behaviour

I suppose the default expected behavior of `web.run_app()` is to allow the requests to be finished before terminating the process. That's what most people would think it does. The feeling is also harder since it does it already for SIGINT.

## Actual behaviour

Only SIGINT triggers a graceful exit. SIGTERM does not let the running requests finish.

## Steps to reproduce

1. run this script
```
import asyncio
from aiohttp import web

async def handle(request):
    await asyncio.sleep(60)
    text = "Hello"
    return web.Response(text=text)

app = web.Application()
app.router.add_get('/', handle)

web.run_app(app)
```
2. do a GET request on /
3. send SIGTERM to the process
4. you will notice the process ends before the 60 seconds and your request has been canceled

## Your environment

Python 3.6, Arch Linux

hubo1016:
Maybe raising a customized exception in a signal handler (instead of SystemExit) is enough to solve this?

asvetlov:
Ok, `run_app` should accept a flag for enabling SIGTERM handling (switched on by default).
Implementation could raise a custom exception as @hubo1016 suggested.
Raising an exception derived from SystemExit is totally fine.

@cecton would you create a Pull Request?

cecton:
Yeay 😄 thx @asvetlov

cecton:
Solved by #1946

PR #1946, opened 2017-06-02
Add SIGTERM handling in web.run_app

cecton:
## What do these changes do?

Makes `web.run_app()` handle SIGTERM by default to gracefully exit the application.

The application run with `web.run_app()` will override any existing SIGTERM handler in the current process.

## Related issue number

#1932 

## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -1,5 +1,6 @@
 import asyncio
 import os
+import signal
 import socket
 import stat
 import sys
@@ -318,10 +319,18 @@ class Application(MutableMapping):
         return "<Application 0x{:x}>".format(id(self))
 
 
+class GracefulExit(SystemExit):
+    code = 1
+
+
+def raise_graceful_exit():
+    raise GracefulExit()
+
+
 def run_app(app, *, host=None, port=None, path=None, sock=None,
             shutdown_timeout=60.0, ssl_context=None,
             print=print, backlog=128, access_log_format=None,
-            access_log=access_logger, loop=None):
+            access_log=access_logger, handle_signals=True, loop=None):
     """Run an app locally"""
     user_supplied_loop = loop is not None
     if loop is None:
@@ -414,12 +423,19 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
         asyncio.gather(*server_creations, loop=loop)
     )
 
-    print("======== Running on {} ========\n"
-          "(Press CTRL+C to quit)".format(', '.join(uris)))
+    if handle_signals:
+        try:
+            loop.add_signal_handler(signal.SIGINT, raise_graceful_exit)
+            loop.add_signal_handler(signal.SIGTERM, raise_graceful_exit)
+        except NotImplementedError:  # pragma: no cover
+            # add_signal_handler is not implemented on Windows
+            pass
 
     try:
+        print("======== Running on {} ========\n"
+              "(Press CTRL+C to quit)".format(', '.join(uris)))
         loop.run_forever()
-    except KeyboardInterrupt:  # pragma: no cover
+    except (GracefulExit, KeyboardInterrupt):  # pragma: no cover
         pass
     finally:
         server_closures = []
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,7 @@ Changes
 - Retain method attributes (e.g. :code:`__doc__`) when registering synchronous
   handlers for resources. #1953
 
--
+- Added signal TERM handling in `run_app` to gracefully exit #1932
 
 - Fix websocket issues caused by frame fragmentation. #1962
 
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -37,6 +37,7 @@ Brett Cannon
 Brian C. Lane
 Brian Muller
 Carl George
+Cecile Tonglet
 Chien-Wei Huang
 Chih-Yuan Chen
 Chris AtLee
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -2085,10 +2085,11 @@ Utilities
 
 
 .. function:: run_app(app, *, host=None, port=None, path=None, \
-                      loop=None, shutdown_timeout=60.0, \
+                      sock=None, shutdown_timeout=60.0, \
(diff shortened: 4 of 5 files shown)


cecton:
The testing is a bit complicated so I only tested using the script given in the original issue.

aiohttp/web.py
```diff
@@ -414,12 +423,15 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
         asyncio.gather(*server_creations, loop=loop)
     )
 
+    if handle_signals:
+        signal.signal(signal.SIGTERM, sigterm_handler)
```
asvetlov: https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals should be used for signal registration.
Also signal subscription should be skipped on Windows

aiohttp/web.py
```diff
@@ -318,10 +319,18 @@ def __repr__(self):
...
+
+def sigterm_handler(signum, frame):
+    raise TerminateSignal()
+
+
 def run_app(app, *, host=None, port=None, path=None, sock=None,
             shutdown_timeout=60.0, ssl_context=None,
             print=print, backlog=128, access_log_format=None,
-            access_log=access_logger, loop=None):
+            access_log=access_logger, handle_signals=True, loop=None):
```
asvetlov, line 333: maybe `handle_sigterm` is better name?

cecton: Yes actually I was thinking that you may want actually that run_app handles all the signals or none. For the user it's not really about handling SIGTERM in particular, it's more about handling the signals (INT and TERM) or not. It's unfortunate somehow that Python handles SIGINT for you by default.

I just realize now that maybe we should also handle SIGINT using https://docs.python.org/3/library/asyncio-eventloop.html#unix-signals but I'm not sure at all about it, what do you think?

cecton:
@asvetlov Do you mind if I move the print() inside the try...except? https://github.com/aio-libs/aiohttp/pull/1946/commits/87fa801fa85f5d07cec6b2d5f7673d8fae6e8c3a

Otherwise I can't possibly know when to terminate.

aiohttp/web.py
asvetlov: SIGINT is handled out of the box, why should we care about it?

asvetlov:
>  Do you mind if I move the print() inside the try...except?
I totally ok with it

asvetlov:
Could you use asyncio api for subprocesses in test instead of `subprocess.Popen`?
It's not critical but why not if it is exist.

cecton:
@asvetlov Yes that's what I wanted in the beginning but the test method is not asynchronous. There is no point and it would make thing unreadable.

cecton:
@asvetlov I mean:

The test function is synchronous so I suppose the tests are run synchronously. The thing I'm testing spawns a process, wait for it and assert. There is only one thing done at a time.

If I want to make it asynchronous it means I will have to move the test to a coroutine and the test itself will do a `run_until_complete` on that coroutine.

There is no point of doing so because the event loop will be busy with only one task. And the perk of using `asyncio.create_subprocess` is also not so good but it is actually using subprocess behind the scene (it's just a wrapper).

So unless you know how to make the tests run asynchronously in parallel like mocha.parallel, there is no real advantage to this.

```python
async def real_test():
    proc = await asyncio.create_subprocess....
    async for line in proc.stdout:
        ....
    proc.terminate()
    retcode = await proc.wait()
    assert ...

def test_sigterm(loop):
    loop.run_until_complete(real_test())
```

cecton:
@asvetlov I just checked the code of asyncio and I believe we should also manage the SIGINT the same way.

The reason is simply that if not managed, Python has this default behavior to raise KeyboardInterrupt. But apparently they made `add_signal_handler` in asyncio because using `signal.signal` wasn't enough.

This is from asyncio:

```python
    def add_signal_handler(self, sig, callback, *args):
(...)
        handle = events.Handle(callback, args, self)
        self._signal_handlers[sig] = handle

        try:
            # Register a dummy signal handler to ask Python to write the signal
            # number in the wakup file descriptor. _process_self_data() will
            # read signal numbers from this file descriptor to handle signals.
            signal.signal(sig, _sighandler_noop)

            # Set SA_RESTART to limit EINTR occurrences.
            signal.siginterrupt(sig, False)
(...)
```

This is the doc for siginterrupt:
> Change system call restart behaviour: if flag is False, system calls will be restarted when interrupted by signal signalnum, otherwise system calls will be interrupted. Returns nothing. Availability: Unix (see the man page siginterrupt(3) for further information).
>
> **Note that installing a signal handler with signal() will reset the restart behaviour to interruptible by implicitly calling siginterrupt() with a true flag value for the given signal.**
From https://docs.python.org/3/library/signal.html#signal.siginterrupt

The other reason that makes me think we should handle it is that they are managing the signals with the event loop itself.

I'm not sure what are the consequences because KeyboardInterrupt and the `signal.signal` call I just added work fine in a simple example but I have a strong feeling about it. I propose to keep the variable `handle_signal` and actually handle INT and TERM with `add_signal_handler`.

aiohttp/web.py
cecton: I actually answered later in the PR comments. It's not really "out of the box" for the event loop, it's out of the box for default synchronous usage of Python.

cecton:
@asvetlov I'm going to give up the signal testing on Windows if you don't mind...

Apparently the console ask to confirm when you send a signal CTRL_BREAK_EVENT or CTRL_C_EVENT:

```
tests\test_resolver.py ssssss....s.sss..
tests\test_run_app.py ...................sssss.s.^CTerminate batch job (Y/N)? 
```

By the way, you may have notice that your Windows CI (appveyor) is a bit stuck? This is because each build wait 1 hour before being killed. Since there are 6 builds every time, it's waiting 6 hours. I didn't realize fast enough and I'm afraid I'm blocking your CI for 7 hours now and there are still 5 hours to go.

Hum.... sorry?

asvetlov:
Oooh. If Windows hangs with user prompt -- please don't run test on it. I'm fine with your proposal.
Don't worry about appveyor blocking -- I believe it was not intentional :)

cecton:
@asvetlov ok finally done :) Are you sure about the subprocess? Did I convince you?

asvetlov:
Thanks

Issue #1954, opened 2017-06-05
2.x: ClientSession read_timeout should not be cumulative


thehesiod:
read_timeout is usually implemented as time from receiving each chunk of data.  This is to prevent servers which are not data back w/in a certain amount of time.

2.x changed the behavior for this to be cumulative, which means that short downloads will work ok, but large downloads will fail even for servers that are still responding correctly.  I have never seen read_timeout be a cumulative timeout.  This change is similar to https://github.com/aio-libs/aiohttp/issues/1907 in that it's a change in behavior that no-one will expect.  

Further there's no real reason to apply a cumulative read_timeout as this will be handled by the `_request` `timeout` parameter.

In a minimum this needs to be changed to have a flag to disable this behavior and make it work like before as to do the previous behavior clients will have to take over the read timeout handling.

fafhrd91:
I think `read_timeout` in general is useless or unneeded. async_timeout is perfectly fine solution

thehesiod:
actually it's really important, there's a distinction between a request taking 'N' seconds, and a server responding w/n N seconds.

AWS for instance can immediately give you back a response stream, but then hang mid-way through reading the stream.

We have one of our services which downloads files from 10k to 800MB.  We can't say we want to request to take N seconds because it depends how big the file is.  What we can say is to fail the request if the server doesn't send any data w/in N seconds.

fafhrd91:
you get back response object when server return http message, so technically it is "time to until" you can start reading response payload. that is reason why I think it is useless.

thehesiod:
http://docs.python-requests.org/en/master/user/advanced/#timeouts

fafhrd91:
so what method can you use to timeout requests in sync world?

thehesiod:
note here too: http://docs.python-requests.org/en/master/user/quickstart/#timeouts  This is how all native libraries I've ever used works too.

you mean an overarching max timeout? You'd have to manually iterate over the chunks and keep your own max timeout..given you know each chunk will be a max of `read_timeout`

fafhrd91:
requests as an example does not make sense. with requests you just don't have other options.

you can implement whatever method you want for aiohttp. it just not possible to satisfy all use cases.
I dont care about `read_timeout`. you can provide PR and doc changes if everyone agrees with reasoning.

@asvetlov @kxepal

thehesiod:
more examples:
http://botocore.readthedocs.io/en/latest/reference/config.html
[...]

PR #1966, opened 2017-06-07
traditional read_timeout implementation

thehesiod:
First hack at implementing the traditional implementation of read_timeout support.  It layers upon the current implementation.

I also added a note that the current impl of read_timeout is not quite correct as it really should be implemented as a pool of time that is pulled from when waiting for data from the server, instead of a fixed timeout from when the request is started as it doesn't take into account how the client uses the API.

attempts to fix https://github.com/aio-libs/aiohttp/issues/1954


thehesiod:
ok, this is my first implementation, @fafhrd91  let me know what you think.  It is kinda hacky...if I were to do it nicer I would completely re-do how timeouts are handle to make it a pool of time that is drawn from when we're waiting for data from the server, but that's a much bigger change.  Also a nicer impl would probably pass the read_timeout across the http parser and to the stream class.

fafhrd91:
Complexity for nothing. I don't like idea in first place. If everyone agrees then we need to redesign feature from the ground.

@asvetlov @kxepal

thehesiod:
I still don't understand how you don't see this a a critical feature.  The primary use case is you want to guard against a server that has hung (think of FTP use case).  The current impl requires the client to calculate how long the request will take based on your current network conditions, size of file, and load. Furthermore, in the case of a large file, you'll probably sent the timeout to something like 5 minutes, this means it will take you 5 minutes to detect that a server has hung instead of the 2s it takes to receive a chunk and detect the hang. This also means you would have to have different times for different sized files.  Further like I've mentioned before several times network read_timeouts are traditionally handled the way I've mentioned, if you wanted a total timeout it should not have been called read timeout, instead it should have been called request timeout or something else.

fafhrd91:
what is wrong with with code?
```python

with async_timeout.timeout(1):
    session.get()
```

thehesiod:
that doesn't affect reading to body of the respose, how would you propose detecting a body read timing out?

fafhrd91:
```python

with Timeout(1):
    resp = yield from session.get()

with Timeout(2):
   content = yield from resp.read()

with Timeout(3):
   chunk = yield from resp.content.read_any()
```

thehesiod:
ya, vs: `async with aiohttp.ClientSession(conn_timeout=1, read_timeout=(1, None)) as session` with my impl, much nicer no?  This works especially well with streaming responses: https://gist.github.com/thehesiod/f562d7b998eda92d4806a144f16bcf2f

If we go your route there's no point for read_timeout or conn_timeout either.  So either it should all be removed, or this enhancement added IMHO.

thehesiod:
libraries should make clients write less code, not more code :)

fafhrd91:
I do not want to argue anymore. -1 for this solution on top of existing one

PR #1977, opened 2017-06-14
Update args in unittest_run_loop()

juliatem:
Update args in unittest_run_loop()

## What do these changes do?
Small fix for unittest_run_loop

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -413,7 +413,7 @@ def unittest_run_loop(func, *args, **kwargs):
     """
 
     @functools.wraps(func, *args, **kwargs)
-    def new_func(self):
+    def new_func(self, *args, **kwargs):
         return self.loop.run_until_complete(func(self, *args, **kwargs))
 
     return new_func
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -413,8 +413,9 @@ def unittest_run_loop(func, *args, **kwargs):
     """
 
     @functools.wraps(func, *args, **kwargs)
-    def new_func(self, *args, **kwargs):
-        return self.loop.run_until_complete(func(self, *args, **kwargs))
+    def new_func(self, *inner_args, **inner_kwargs):
+        return self.loop.run_until_complete(
+            func(self, *inner_args, **inner_kwargs))
 
     return new_func
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,7 +15,8 @@ Changes
 
 - Raise RuntimeError is you try to set the Content Length and enable
   chunked encoding at the same time #1941
--
+
+- Small update for `unittest_run_loop`
 
 -
 


juliatem:
The decorator async_data_provider(fn) can be used in cases when some unit tests have to check similar logic on the different sets of input and output values.

aiohttp/test_utils.py
```diff
@@ -573,3 +573,17 @@ def mock_coro(*args, **kwargs):
...
+
+def async_data_provider(fn_data_provider):
+    """Data provider decorator, allows another callable to provide the data for the async test"""
+    def test_decorator(fn):
+        async def repl(self, *args):
+            for args in fn_data_provider():
+                try:
+                    await fn(self, *args)
+                except AssertionError:
+                    print("Assertion error caught with data set ", args)
```
kxepal: I think better throw own AssertionError with that message preserving original stacktrace. Like `raise AssertionError("Assertion error caught with data set: %s " % args) from exc`

kxepal: Also, if some else error happens, wouldn't it be helpful to get for which data it happened?

kxepal:
On behalf of provocateur: why not just to use [pytest](https://docs.pytest.org/en/latest/fixture.html) instead?

asvetlov:
I'm with @kxepal 
Why do we need this functionality in aiohttp at all?

juliatem:
Do you mean `@pytest.fixture`?
I see, it can provide similar functionality. So I just will commit chenges in str 416, ok? It's realy needed. I lost it in previous time.

aiohttp/test_utils.py
```diff
@@ -413,7 +413,7 @@ def unittest_run_loop(func, *args, **kwargs):
     """
 
     @functools.wraps(func, *args, **kwargs)
-    def new_func(self):
+    def new_func(self, *args, **kwargs):
```
kxepal: I fear now there is a collision and it's a breaking change: previously args and kwargs were taken from unittest_run_loop arguments, now they get overridden by new_func arguments.

kxepal:
> It's realy needed. I lost it in previous time.

No worries. Could you describe your case? What do you want to achieve and why pytest is not an option?

kxepal:
As workaround you could try this (didn't run):
```
class MyTests(AioHTTPTestCase):
    @unittest_run_loop
    async def test_ok(self):
        for input_value, expected_result in get_data_stubs():
            with self.subTest(input_value=input_value, expected_result=expected_result):
                result = await my_function(input_value)
                self.assertEqual(expected_result)
```
A bit more nicer and doens't requires aiohttp changes.

aiohttp/test_utils.py
juliatem: When you use other decorators, for example `@patch` with `@unittest_run_loop` then:
1. `unittest_run_loop()` transform your test-function to `new_func(self)`
2. `@patch` (or other decorator) will applied for `new_func()` which can't take argument from the patch

kxepal: Ah, got it. Tricky.

juliatem: So, what about this small  fix?

asvetlov: Could you rename these params to `inner_args`/`inner_kwargs` or `new_args`/`new_kwargs` maybe?
Otherwise people will always stuck with this code.

asvetlov:
Thanks

PR #1983, opened 2017-06-16
[WIP] Cythonize frozenlist and signals

asvetlov:
Right now it's quite slow.
Adding C Extension might make it faster

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,10 @@
 .noseids
 .tox
 .vimrc
-aiohttp/_multidict.c
-aiohttp/_multidict.html
+aiohttp/_frozenlist.c
+aiohttp/_frozenlist.html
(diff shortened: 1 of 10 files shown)


kxepal:
Quite stupid question, may be, but why not to just have a plain list when state could be mutable and cast it to tuple when you need to make it immutable? 

Idea of data structure that eventually could change own behavior depending on some method call is quite strange for me.

kxepal:
For example, frozenset. It's just immutable set, not a set which you need to freeze to make immutable.

asvetlov:
Well, having a new object for immutable state is viable idea.
But I think the ship has sailed: we already use `FrozenList` and derive `Signal` from it.
Also we use *frozen* concept for application and router already. `FrozenList` is not very different from frozen application.

PR #1988, opened 2017-06-19
Fix #1985: Keep a reference to ClientSession in response object

### LINKED ISSUE #1985: Keep a reference to ClientSession in response object
https://github.com/aio-libs/aiohttp/issues/1981#issuecomment-309293373

asvetlov:
Fix for #1985

@fafhrd91 is the change good for you?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,8 @@ Changes
 
 - Fix BadStatusLine caused by extra `CRLF` after `POST` data #1792
 
+- Keep a reference to ClientSession in response object #1985
+
 
 2.1.0 (2017-05-26)
 ------------------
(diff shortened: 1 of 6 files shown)


fafhrd91:
you should set session to None in `_cleanup_writer`, just in case

asvetlov:
Done.
Honestly I think it's not necessary: request and response have an unidirection reference to session but the session doesn't keep references to req/rep. Thus we have no circle and regular decref should release everything very well.

CHANGES.rst
```diff
@@ -42,6 +42,14 @@ Changes
 
 - Keep a reference to ClientSession in response object #1985
 
+-
+
+- Deprecate undocumented app.on_loop_available signal #1978
```
kxepal: These changes looks accidental here.

PR #1989, opened 2017-06-19
Fix #1978: Deprecate on_loop_available

### LINKED ISSUE #1978: Deprecate app.on_loop_available signal
From my understanding on `app.on_startup` the loop is already present.
Why do we need additional signal?
`app.on_loop_available` is not documented BTW
@fafhrd91 ?

asvetlov:
Fix for #1978

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -36,6 +36,14 @@ Changes
 
 - Fix BadStatusLine caused by extra `CRLF` after `POST` data #1792
 
+-
+
+- Deprecate undocumented app.on_loop_available signal #1978
+
+-
+
+-
+
 
 2.1.0 (2017-05-26)
 ------------------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -36,6 +36,14 @@ Changes
 
 - Fix BadStatusLine caused by extra `CRLF` after `POST` data #1792
 
+-
+
+- Deprecate undocumented app.on_loop_available signal #1978
+
+-
+
+-
```
kxepal, line 45: Reserved for future needs? (:

asvetlov: It reduces the probability of merge conflict.
pytest project used to follow this practice but now they switched to new approach: a folder for change records.
We need the same maybe.

kxepal: Interesting. Thanks.

PR #2000, opened 2017-06-21
Fix tests

### LINKED ISSUE #1994: Switch to new multidict version
Current `multidict` master has changed behavior: keys for `CIMultiDict` are always `str` (before they was either `str` or `istr`) and they are stored as-is without title casing. But key comparison is performed on title-cased basis.

Also `CIMultiDict` lookup is about 30% faster.

It is backward incompatible change, some aiohttp tests are failing. But all functionality is working on just released aiohttp 2.2.0 (it's crucial for smooth transition).
 
I need release new multidict first and fix failed aiohttp tests.

asvetlov:
Fix #1994

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -161,7 +161,7 @@ class ClientRequest:
 
     def update_auto_headers(self, skip_auto_headers):
         self.skip_auto_headers = CIMultiDict(
-            (hdr, None) for hdr in skip_auto_headers)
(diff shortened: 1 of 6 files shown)


fafhrd91:
Nice PR number :)

asvetlov:
A lot of work since 2013.
Unbelievable, the project will celebrate 4 years this Autumn.

fafhrd91:
time flies

PR #2004, opened 2017-06-22
Route definitions

asvetlov:
Alternative to #2003

@fafhrd91 your [proposal](https://github.com/aio-libs/aiohttp/pull/2003#issuecomment-310492680 ) was very seductive.

I've implemented it :)


asvetlov:
The usage:
```
    routes = web.RouteDef()

    @routes.get('/path')
    @asyncio.coroutine
    def handler(request):
        pass

    app.router.add_routes(routes)
```

fafhrd91:
cool!

what if `Routes` behave more like sub app, then you can manage set of routes as one entity and apply rules to all routes at the same time

asvetlov:
I don't follow.
What rules are you talking about?

fafhrd91:
for example, url prefix, or some kind of permissions, etc. or even use different route style. 
maybe it is too far and too broad

asvetlov:
Right now we have no permissions, let's don't discuss it here.
Regarding to url prefix -- it's doable but have own drawback.
Let's consider the following:
```
    routes = web.RoutesDef()

    @routes.get('/path')
    @asyncio.coroutine
    def handler(request):
        pass

    app.router.add_routes(routes, prefix='/base')
```
Final url will be `'/base/path'`. The behavior is pretty simple and obvious.
But my guts feel this is not very good idea.
With `RoutesDef` the route table is scratched over many places in many files.
The most practical way to find proper web handler is grepping the source code by URL.
Prefixes make the process much harder: route definitions may have the same path but different prefixes, it should very common for REST based solutions.
Thus better to have longer paths in decorators for easy searching. Otherwise user should look on two places: decorator and `RoutesDef` registration.

P.S.
I prefer `RoutesDef` to `Routes`: we already have `Route` concept and it's used for **registered** route. But `RoutesDef` is for **unregistered** route definition, let's don't mix them.

Now I think `RouteInfo` namedtuple should be renamed to `RouteDef` (the name will almost newer be used in application code).
More important `router.add_routes()` should renamed `.add_routes_def()`. It's 4 chars longer but supports more native semantics.

fafhrd91:
I think of `RoutesDef` as common context for all routes. same as sub application. without `common context` it becomes just list of routes and what is the point then introduce new concept

asvetlov:
Sub-application has a context: state aka `app['name']`, own signals and middlewares. Why do we need another stateful concept?
`RoutesDef` in opposite is just a state-less list of route definitions but with convenient methods like `routes.get(...)` for usage as web-handler decorators.

fafhrd91:
Too much ceremony for convenience. -1
Just use `@web....` decorator

asvetlov:
What the decorator should do?
It's supposed to be a plain function in `aiohttp.web` namespace, isn't it?

fafhrd91:
Yes, 
it create some hidden module level variable, put route data in it. Then routes method, something like `load_mod(module object)` reads variable and register routes

jairhenrique:
I like de [apistar](https://github.com/tomchristie/apistar) approach, using `Router` and `Include` classes.

asvetlov:
@jairhenrique  it's done by PR:
```
    app.router.add_routes([
        web.get('/path', handler1),
        web.post('/path', handler2),
    ])
```
apistar.Include counterpart is not implemented yet but it could be done in several lines of code.
Supporting class based views is trivial also.
Looks like table based route definition is not a subject for discussion -- it's good enough and maybe requires a couple extra functions.
Decorator-based approach is much more contradictory question. Maybe I need splitting the PR in two for easy discussing/accepting.

asvetlov:
> it create some hidden module level variable, put route data in it. Then routes method, something like `load_mod(module object)` reads variable and register routes

This is a movement in opposite direction than your [original](https://github.com/aio-libs/aiohttp/pull/2003#issuecomment-310492680) propose.

1. Adding a hidden module variable increases a level of *hidden magic* a bit. I cannot name the solution as *very explicit*.
2.  Another problem is: how to determine a modude which should be used by decorator? 

At first glance it could be `sys.modules[handler.__module__]` and this works for 99% use cases but silently fails if the handler is defined in one module but decorated in other like `handler = web.get('/path')(handler)`. It's very unusual and code looks weird but and don't want to add this kind of *weird error-by-design* at all, it's very not pythonic. Looks like design bomb.

Fetching data by analyzing `sys._getframe()` frames is even worse by very many reasons: user written helpers, pypy support, cython support and I dont know what else potential errors.

P.S. 
RoutesDef instance might be shared between modules if user found it convenient to him -- why not.
Like not all Flask's blueprints and limited to single module only -- it might be a package also with routes scattered by package modules.

P.P.S.
`app.router.load_mod(module object)` doesn't look pretty for single file projects. `__main__` module has no a reference to itself, the only way to make a call is `app.router.load_mod(sys.modules['__name__'])`. Not very elegant solution.

aiohttp/web_urldispatcher.py
```diff
@@ -28,12 +29,25 @@
...
            'AbstractRoute', 'ResourceRoute',
-           'StaticResource', 'View')
+           'StaticResource', 'View', 'RoutesDef',
+           'head', 'get', 'post', 'patch', 'put', 'delete', 'route')
 
 HTTP_METHOD_RE = re.compile(r"^[0-9A-Za-z!#\$%&'\*\+\-\.\^_`\|~]+$")
 PATH_SEP = re.escape('/')
 
 
+class RouteDef(namedtuple('_RouteDef', 'method, path, handler, kwargs')):
```
jairhenrique: Why not only `Route`? 
What's the meaning of `Def`?

aiohttp/web_urldispatcher.py
```diff
@@ -894,3 +908,85 @@ def freeze(self):
...
+
+        Parameter should be a sequence of RouteDef objects.
+        """
+        # TODO: add_table maybe?
+        for route in routes:
+            route.register(self)
+
+
+def route(method, path, handler, **kwargs):
+    return RouteDef(method, path, handler, kwargs)
```
jairhenrique: I think the code gets more readable when we do not use abbreviations. Can I open a `PR` by renaming` hdrs` to `headers`?

aiohttp/web_urldispatcher.py
```diff
@@ -894,3 +908,85 @@ def freeze(self):
...
+
+
+def head(path, handler, **kwargs):
+    return route(hdrs.METH_HEAD, path, handler, **kwargs)
```
jairhenrique, line 936: I think the code gets more readable when we do not use abbreviations. Can I open a `PR` by renaming` hdrs` to `headers`?

aiohttp/web_urldispatcher.py
asvetlov: Because we already use `Route` name for routes added by `app.router.add_route()` call.
`RouteDef` is a scratch for non-added-yet route definition

aiohttp/web_urldispatcher.py
asvetlov: It would be good in general but for this particular case it doesn't make sense I think.
`hdrs` is not a part of our public API anyway.

aiohttp/web_urldispatcher.py
```diff
@@ -894,3 +908,85 @@ def freeze(self):
...
+        """Append routes to route table.
+
+        Parameter should be a sequence of RouteDef objects.
+        """
+        # TODO: add_table maybe?
+        for route in routes:
+            route.register(self)
+
+
+def route(method, path, handler, **kwargs):
```
Arfey, line 931: i think after refactoring this function is not required

aiohttp/web_urldispatcher.py
```diff
@@ -894,3 +908,85 @@ def freeze(self):
...
+def delete(path, handler, **kwargs):
+    return route(hdrs.METH_DELETE, path, handler, **kwargs)
+
+
+class RoutesDef(Sequence):
+    """Route definition table"""
+    def __init__(self):
+        self._items = []
+
+    # TODO: add __repr__
```
Arfey:
```
import operator

def __repr__(self):
    items = sorted(
        self._items,
        key=operator.attrgetter('path', 'method')
    )
    return "<RoutesDef\n\t{}\n>".format(
        "\n\t".join(map(str, items))
    )

____
result

<RoutesDef
	< RouteDef GET /article -> 'get_article', {}>
	< RouteDef GET /user -> 'get_user', {}>
	< RouteDef POST /user -> 'post_user', {}>
>
```

asvetlov: Thank you for suggestion for route definitions should be sorted by insertion order.
It is crucial for router implementation.

aiohttp/web_urldispatcher.py
```diff
@@ -28,12 +29,25 @@
...
-           'StaticResource', 'View')
+           'StaticResource', 'View', 'RoutesDef',
+           'head', 'get', 'post', 'patch', 'put', 'delete', 'route')
 
 HTTP_METHOD_RE = re.compile(r"^[0-9A-Za-z!#\$%&'\*\+\-\.\^_`\|~]+$")
 PATH_SEP = re.escape('/')
 
 
+class RouteDef(namedtuple('_RouteDef', 'method, path, handler, kwargs')):
+    # TODO: add __repr__
```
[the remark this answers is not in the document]
asvetlov: As I said `Route` name  is captured by already existing class hierarchy.
It's present *instantiated route*.
In opposite *route definition* is for list of routes to be added to router instance.

I still think the distinction is matter.

asvetlov:
@fafhrd91 and others.

I love to merge the PR after polishing.

Now we have *directive* way for adding new routes: `app.router.add_get(...)`
The PR proposes `router.add_routes()` for  adding routes in Django style: https://github.com/aio-libs/aiohttp/pull/2004/files#diff-d24e0645e091e8104b32633c7a3cab93
Looks like nobody objects.

Also it adds decoration based style like flask and bottle: https://github.com/aio-libs/aiohttp/pull/2004/files#diff-1fe46683db6b1c2e6b977a6b3f5da19f
The only difference the PR introduces `RoutesDef` class (maybe `RouteTableDef` is better name) for collecting routes by decorator appliance and adding them to router explicitly.

I feel people need it too. 
I did my best to keep aiohttp spirit: everything is explicit, no magic even if users should write more code.
No side effects and potential break downs.

If @fafhrd91 will say *strict no* for PR I'll split it into separate ones for table and decorators. 
But please don't do it, I very want to have both features in next aiohttp release.

fafhrd91:
lgtm

asvetlov:
Thank you @fafhrd91

aiohttp/web_urldispatcher.py
asvetlov: It is: user might want to provide custom HTTP method in very rare cases

asvetlov:
BTW the PR is pretty close to https://github.com/IlyaSemenov/aiohttp_route_decorator

asvetlov:
Docs added, code is 100% covered.
The PR is done.
Will merge it after tomorrow if nobody object.

aiohttp/web_urldispatcher.py
```diff
@@ -38,7 +38,13 @@
...
 
 class RouteDef(namedtuple('_RouteDef', 'method, path, handler, kwargs')):
-    # TODO: add __repr__
+    def __repr__(self):
+        info = []
+        for name, value in sorted(self.kwargs.items()):
+            info += ", {}={}".format(name, value)
+        return ("<RouteDef {method} {path} -> {handler.__name__!r}"
+                "{info}>".format(method=self.method, path=self.path,
+                                 handler=self.handler, info=''.join(info)))
```
Arfey: i think will be better if info will be string 
then you can delete ''.join(info)

asvetlov: This style is used by asyncio in many places, I like to keep it in aiohttp too

PR #2008, opened 2017-06-24
[WIP] Introducing scheduler for background tasks

asvetlov:
#For aiohttp people need long-running background tasks very often.
The solution is `loop.create_task(coro)` usage but this very simple approach has own drawbacks:
1. No report about failed task except printout to console
2. No way to gracefully shutdown all active tasks on application exit
3. No limit for amount of concurrently executed tasks

The solution is adding `app.scheduler` attribute for starting, waiting and closing these long tasks aka jobs.

To start a job call `await app.scheduler.run(coro())`. At app finish all active jobs will be stopped.
`.scheduler.run()` returns a Job instance.

User could explicitly await for job finishing:
```
job = await app.scheduler.run(coro())
ret = await job.wait(timeout=1.0)
```
In case of explicit waiting the error is raised. Failed non-awaited jobs a reported by `app.scheduler.call_exception_handler()`. 
`.call_exception_handler()` is very similar to `loop.call_exception_handler()` and calls it by default but user might setup own handler by `app.scheduler.set_exception_handler()`.

All active jobs are finished on `app.shutdown()`.

By default amount of active jobs are limited to 100 (user can change this parameter).
If there are 100 active jobs and 101th is scheduled `JobRunner` will wait for one active job finishing.

There is no documentation yet, I'll make it soon.


asvetlov:
@jettify you might have an interest in this PR.

aiohttp/web.py
```diff
@@ -133,6 +135,8 @@ def _set_loop(self, loop):
         if self._debug is ...:
             self._debug = loop.get_debug()
 
+        self._jobs = JobRunner(loop=loop)
```
jettify: May be stupid idea, but I think we may want to pass JobRunner as parameter. I suspect with time, we will see interesting custom implementation of runner, for instance `ExecutorJobRunner`, that uses `Executor` to do sync work

jettify:
I think it is very good  idea. Quick question, want happens if coroutine is cancelled?
```
job = await app.jobs.run(coro())
ret = await job.wait(timeout=1.0)
```
Cancel on first `await`,  job never submitted?
What about second? User should use `asyncio.shield` for protection from cancel?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+import traceback
+from collections import deque
+from collections.abc import Container
+
+import async_timeout
+
+from .helpers import ensure_future
+
+
+__all__ = ['JobRunner']
```
kxepal: `__all__` should be a tuple. And may be add Job here?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+class Job:
+    _source_traceback = None
+    _closed = False
+    _explicit_wait = False
+    _task = None
+
+    def __init__(self, coro, manager, loop):
+        self._loop = loop
+        self._coro = coro
+        self._runner = manager
```
kxepal: I think these names should be in sync.

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+        self._runner = None  # drop backref
+        self._closed = True
+
+
+class JobRunner(Container):
+    def __init__(self, *, loop):
+        self._loop = loop
+        self._jobs = set()
+        self._timeout = 0.1
+        self._concurrency = 100
```
kxepal: Concurrency and timeout need to be configured from user side.

kxepal: Ah, I see property setters, but imho it's better to set them on `__init__` rather than encourage change them somewhere after.

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+    def concurrency(self, concurrency):
+        self._concurrency = concurrency
+
+    @property
+    def active_count(self):
+        return self._active
+
+    @property
+    def close_timeout(self):
+        return self._timeout
```
kxepal: I was quite hard to get what `_timeout` is for until I saw public interface to it. May be rename to keep them in sync?

aiohttp/web.py
kxepal: Naming bikeshedding alert. 

Jobs is a JobRunner, not a list of jobs which you might think following you intuition. I always mess with `app.on_startup` / `app.on_cleanup` callbacks because these are not a callable accumulators, but list-like objects.

May be use `scheduler` word and have Scheduler class instead of JobRunner? Yeap, a bit longer, but quite clear what it about. (Personally, I just don't like Job and any Runners).

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+class JobRunner(Container):
+    def __init__(self, *, loop):
+        self._loop = loop
+        self._jobs = set()
+        self._timeout = 0.1
+        self._concurrency = 100
+        self._exception_handler = None
+        self._failed_tasks = asyncio.Queue(loop=loop)
+        self._failed_waiter = ensure_future(self._wait_failed(), loop=loop)
+        self._pending = deque()
```
kxepal: I think pending list should be exposed as well. This class is basically queue of tasks which may overflow and you would like to catch that moment until it will be too late.

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+        handler = self._exception_handler
+        if handler is None:
+            handler = self._loop.call_exception_handler(context)
+        else:
+            handler(self, context)
+
+    def get_exception_handler(self):
+        return self._exception_handler
+
+    def set_exception_handler(self, handler):
```
kxepal: Why exception_handler is not a property?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+        self._active = 0
+        self._closed = False
+
+    @asyncio.coroutine
+    def run(self, coro):
+        job = Job(coro, self, self._loop)
+        self._jobs.add(job)
+        if self._active < self._concurrency:
+            self._active += 1
+            job._start()
```
kxepal: Why job has no public api to start it?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+    def close(self):
+        if self._closed:
+            return
+        jobs = self._jobs
+        if jobs:
+            yield from asyncio.gather(*[job.close() for job in jobs],
+                                      loop=self._loop, return_exceptions=True)
+        self._failed_tasks.put_nowait(None)
+        yield from self._failed_waiter
+        self._closed = True
```
kxepal: What happens with pending tasks on close? Should be raised some warning about them? Shouldn't those dequeue be cleared to drop the references?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+        self._failed_tasks = asyncio.Queue(loop=loop)
+        self._failed_waiter = ensure_future(self._wait_failed(), loop=loop)
+        self._pending = deque()
+        self._active = 0
+        self._closed = False
+
+    @asyncio.coroutine
+    def run(self, coro):
+        job = Job(coro, self, self._loop)
+        self._jobs.add(job)
```
kxepal: The _jobs is a set, so we don't have duplicates there, but why I still can start the same task twice?

aiohttp/web_jobs.py
```diff
@@ -0,0 +1,223 @@
...
+
+import async_timeout
+
+from .helpers import ensure_future
+
+
+__all__ = ['JobRunner']
+
+
+class Job:
```
kxepal: I know Job name is sweet and short, but it introduces a new entity on the board. How about to stay with Task name? Internally, you still uses `_task`, not something else.

aiohttp/web.py
asvetlov: Maybe later if people will request it.
Easier to add a new `.run_sync()` method than inviting to reimplement the whole logic.
`.run_sync(func, *args, **kwargs)` should have a different signature than `.run(coro)`, it's not just a replacement.
Cancellation of synchronous jobs is also problematic challenge.

asvetlov:
> Quick question, want happens if coroutine is cancelled?

@jettify are you asking about support for short task that should be executed atomically and maybe even with 100% guarantee (always scheduled and newer cancelled)?
We could support atomic  by `.run(coro(), atomic=True)` (internaly implementation could use `shield()`) but 100% guarantee could give only message queue like RabbitMQ etc.

aiohttp/web_jobs.py
asvetlov: Technically better to remove `__all__` at all -- both `Job` and `JobRunner` are not exposed to `aiohttp.web` namespace.

aiohttp/web_jobs.py
asvetlov: `Job` API is different from `Task`. That's why I invited new name.

aiohttp/web_jobs.py
asvetlov: Sure.

aiohttp/web_jobs.py
asvetlov: In current implementation user should never instantiate `JobRunner` but use `.app.jobs` property.

aiohttp/web_jobs.py
asvetlov: Not sure if we need to expose the list -- people should do nothing with pending tasks except getting a count of them. But it's available with current API: `pending_count = len(app.jobs) - app.jobs.active_count`

aiohttp/web_jobs.py
asvetlov: It's a set of `Job` instances where every job is unique. If you start a task twice -- you give two different jobs.

aiohttp/web_jobs.py
asvetlov: It should be started by runner not by user.
Only `JobRunner` decides a time of job starting basing on concurrency limit and number of active jobs.

aiohttp/web_jobs.py
asvetlov: Backref to runner is removed on `.job.close()`.
And yes, let's clear both `.pending` and `.jobs`

aiohttp/web_jobs.py
asvetlov: Agree.

aiohttp/web_jobs.py
asvetlov: To be mimic to exception handler API in loop.

asvetlov:
Well, using `scheduler` instead of `jobs` makes sense.

aiohttp/web_jobs.py
kxepal: May be expose that math via another property? It is not quite oblivious unless you familiar with implementation. Which eventually coudl be changed.

aiohttp/web_jobs.py
kxepal: Ok. Though it looks a bit inconsistent here. On other hand, with propertry `call_exception_handler` will do the same.

aiohttp/web_jobs.py
kxepal: By references I mean that JobRunner keeps references to jobs which keeps references to tasks. But clear all the things will fix that.

asvetlov: Pending tasks are closed without starting its coroutines.
I not sure if we need a warning here.
Because stopping web server most likely raises this warning always.

kxepal:
@asvetlov Did I get you right that you want to keep the jobs thing completely private/internal for now?

asvetlov:
I don't understand what do you mean by *private/internal*.

kxepal:
> I don't understand what do you mean by private/internal.

That user can't use own Scheduler, own jobs classes, interact with in any other way except via `app`.

asvetlov:
Yes. It makes experimenting with scheduler implementation easier to us.
`Scheduler` is complex enough concept, we have no stable ABC and contract for it. 
In this situation writing custom scheduler compatible with all others is very hard.
On other hand user might write own scheduler-like class, instantiate it in `app.on_startup()` and finish in `on_cleanup()`.

aiohttp/web_jobs.py
asvetlov: ok

aiohttp/web_scheduler.py
```diff
@@ -0,0 +1,226 @@
...
+    def concurrency(self, concurrency):
+        self._concurrency = concurrency
+
+    @property
+    def active_count(self):
+        return len(self._jobs) - len(self._pending)
+
+    @property
+    def pending_count(self):
+        return len(self._pending)
```
kxepal, line 206: Just a question: how about to return active/pending list of tasks instead of their count? Then user may decide what to do with them: count them all, close them all, filter out to find out something.

asvetlov: I think better to add `Job.active` property. User might iterate over all jobs now.
BTW what name is better: `active` or `running`?

asvetlov:
I've changed my mind about pending jobs close.
Instead of ignoring these jobs we should start them and cancel immediately.
I think new behavior is more consistent, at least job's coroutine may be informed about cancellation instead of just silently skipped.
It makes wrapping a coroutine into `shield` reliable also.

aiohttp/web_scheduler.py
asvetlov: `job.active` and `job.pending` added

kxepal: I think `active` is good. Running means that these tasks are actually running, but with asyncio they don't - only one task is actually running at the single moment of time.

asvetlov: Ok

samuelcolvin:
Is it really a good idea to encourage running background tasks in the same event loop and process as a web app? I think not.

Whether it be multiprocessing, separate container, separate VM, worker dynos etc. etc. most common deployment setups allow (and encourage) developers to implement any task which is divorced from the event loop in a separate, isolated environment.

Most people (I think) assume that background tasks shouldn't compete directly for resources with the front-end server.

Worker environments often have:
* different resource requirements
* different dependencies
* very different shutdown behavior - you wouldn't want to cancel a slow background task because you just want to scale down the number of web-server processes.
* different connections and connection pools

In other words, they're a completely different subsystem. It's the first tiny step towards a microservice architecture.

If I didn't know better, I could imagine starting a new project, deciding to use aiohttp and being encouraged to implement all background tasks using aiohttp `Scheduler` and getting into big trouble. What if a scheduled task is CPU bound, is it ok that the webserver will bog down? What happens if a scheduled task blocks, is it ok that the webserver will cease to process requests.

If this idea gets merged: can I suggest that there's a big, hard to ignore warning explaining that it's not a substitute for a proper worker setup and should only be used for short tasks that can never block or use significant CPU resources.

asvetlov:
The scheduler is not a replacement for distributed systems.
And yes, crash of process/server may lead to data loss.
But there are many applications for running something in the same process after handling HTTP request. People do it already but I pretty sure their code is weak in terms of error handling, cancellation and graceful  shutdown. I'm trying to solve it.

Let's assume we need collecting some statistics. Usually we could allow to miss something in case of crash. Also we most likely want to respond quickly, pushing statistics in separate task sounds reasonable. 

On other hand starting new servers just for getting stats from frontend and putting it into a storage is very big overhead most likely (maybe not -- it depends). Anyway we need background tasks for pushing stats to some place.

jettify:
Other very common use cases is sending letters and publishing notification via REST. Deploying celery with rabbitmq just for this reason is overkill.

samuelcolvin:
Ok, I guess I see that there are some applications for this. 

But I still think it would be good to have a big fat warning about what this is good for and what it's not good for in the docs?

Perhaps with a short description of how to properly distribute critical/heavy tasks

asvetlov:
Agree

samuelcolvin:
(please forgive/ignore the self promotion)

In terms of a solution for distributing tasks, I'm using [arq](https://github.com/samuelcolvin/arq) in production and it works very well. All it needs is redis which I'm generally using anyway for the cache so there are no extra requirements except running another container for the worker.

asvetlov:
Yes, i'l aware about `arq`. This is only one of many available options. Regardless we need jobs anyway.

fafhrd91:
I wonder why is this needs to be part of framework and not standalone package? Seems like departure from web framework to process management

samuelcolvin:
True even Django "I will tell you how to do everything" doesn't define how
background tasks should be executed.

asvetlov:
Because Django's view almost never starts new thread. 
But people calls `loop.create_task()` from web-handlers very often *right now* without support from aiohttp.
And they handle these tasks wrong in 95% cases as I mentioned in very first message.
That's why we need provide a reliable tool to help them write error-safe code.

fafhrd91:
I still don't understand why it needs to be part of aiohttp.

asvetlov:
To encourage people use best techniques.
But if you reluctant for inclusion so much -- I could extract it into third-party with adding a mention to aiohttp docs, sure.

fafhrd91:
I think it should be external package. aiohttp already has big api surface and adding new api that is not related to http is very good.

asvetlov:
Ok. Let's keep a PR open until I'll extract it into separate library.

kxepal:
I agree about to have this feature as aiohttp addon. It's quite complicated topic in long term and eventually it will envolve by it own. There could be quite a lot of features that aims only background task problem, not aiohttp itself. Continuous jobs vs short term ones. How to keep jobs alive. How to backpressure their restart. How to integrate with popular libs that solve the same goal, like celery. Etc.

kxepal:
Also, as stand alone, this project could achieve more than as a part of aiohttp. Background jobs is not a topic of aiohttp, but any asyncio-based application.

asvetlov:
Code moved into https://github.com/aio-libs/aiojobs
First official release will be published pretty soon.

Issue #2058, opened 2017-07-06
Can not run a subprocess in a test: "the child watcher does not have a loop attached"


cecton:
## Long story short

The loop created for the testing purpose doesn't allow you to spawn subprocesses using `asyncio.create_subprocess_*`.

## Expected behaviour

You can spawn subprocesses without problem like you would do with a default loop.

## Actual behaviour

When you try to spawn a subprocess, you automatically get:

```
self = <asyncio.unix_events.SafeChildWatcher object at 0x7fe80eb15e10>, pid = 9399
callback = <bound method _UnixSelectorEventLoop._child_watcher_callback of <_UnixSelectorEventLoop running=False closed=False debug=False>>
args = (<_UnixSubprocessTransport pid=9399 running>,)

    def add_child_handler(self, pid, callback, *args):
        if self._loop is None:
            raise RuntimeError(
>               "Cannot add child handler, "
                "the child watcher does not have a loop attached")
E           RuntimeError: Cannot add child handler, the child watcher does not have a loop attached

/usr/lib64/python3.6/asyncio/unix_events.py:858: RuntimeError
```

## Steps to reproduce

```python
import asyncio


# tokio event loop does not allow to override attributes
def skip_if_no_dict(loop):
    if not hasattr(loop, '__dict__'):
        pytest.skip("can not override loop attributes")


def stopper(loop):
    def f(*args):
        loop.call_later(0.001, loop.stop)
    return f


async def test_subprocess_co(loop):
    proc = await asyncio.create_subprocess_shell("exit 0", loop=loop)
    await proc.wait()
    assert proc.retcode == 0


def test_subprocess(loop, mocker):
    skip_if_no_dict(loop)
    loop.run_until_complete(test_subprocess_co(loop))
```

## Your environment

Arch Linux, Python 3.6

cecton:
@asvetlov is it possible that it is because the current loop is set to None when running the test?

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L442

I have no idea why we put that to None but setting it to loop fixes the issue. Should I make a PR?

kblin:
The PR fixes an issue I was seeing trying to use a pytest-aiohttp event loop to test a function using `asyncio.subprocess`. Thanks!

cecton:
@kblin yw \o/

[...]

PR #2059, opened 2017-07-06
Fix test event loop

cecton:
## What do these changes do?

Set the current event loop properly during the test.

Some functions that use the default event loop will now work properly: asyncio.create_subprocess_exec and asyncio.create_subprocess_shell

## Related issue number

#2058 

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -441,6 +441,11 @@ def setup_test_loop(loop_factory=asyncio.new_event_loop):
     """
     loop = loop_factory()
     asyncio.set_event_loop(None)
+    if sys.platform != "win32":
+        policy = asyncio.get_event_loop_policy()
+        watcher = asyncio.SafeChildWatcher()
+        watcher.attach_loop(loop)
+        policy.set_child_watcher(watcher)
     return loop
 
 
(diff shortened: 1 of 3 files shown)


asvetlov:
No. We intentionally disabled default event loop to check for passing the loop explicitly everywhere in aiohttp.
This behavior should be kept at least while we support Python < 3.5.3.

But your problem has another solution: just call `loop.get_child_watcher()` after a new loop creation.

cecton:
@asvetlov ah thanks! Makes totally sense.

Then would it be a good idea to add this call in the setup_test_loop? Because apparently I'm not the only one to have failed to understand.

cecton:
Actually that won't work. `get_child_watcher` is a method of a policy, not a loop...

asvetlov:
Ooops, I see.
But should be a way for attaching a new loop without making it default.

cecton:
I found a way ^_^

cecton:
Actually I'm not entirely sure that you want to enforce the developers to give the event loop explicitly. According to my memory of the doc of asyncio, this is highly "recommended". But enforcing it will only break all the applications that rely on that behavior. For once, I'm the one who suggest to be flexible on that.

cecton:
I close this PR and re-open it for master because it will now be a feature and not a fix.

Issue #2063, opened 2017-07-06
ClientSession.__aexit__ synchronous session close


litwisha:
## Long story short

When using `ClientSession` as an asynchronous context manager, `self.close` is not awaited and it throws `DeprecationWarning`.

## Expected behaviour

Returns `self.close()` in `ClientSession.__aexit__`

## Actual behaviour

Calls `self.close()` in `ClientSession.__aexit__`

## Steps to reproduce

```
import asyncio
from aiohttp import ClientSession

loop = asyncio.get_event_loop()


async def coro():
    async with ClientSession(loop=loop):
        pass


loop.run_until_complete(coro())
```

## Your environment
```
OS X 10.12.3
python 3.6.0
aiohttp==2.2.3
```

asvetlov:
Fixed by #2064

asvetlov:
Thanks!

PR #2064, opened 2017-07-06
ClientSession.__aexit__ returns ClientSession.close coroutine

litwisha:
## What do these changes do?

Correct session closing for asynchronous context manager.

## Related issue number

#2063

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -582,8 +582,9 @@ class ClientSession:
         def __aenter__(self):
             return self
 
+        @asyncio.coroutine
         def __aexit__(self, exc_type, exc_val, exc_tb):
-            return self.close()
+            yield from self.close()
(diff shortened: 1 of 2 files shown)


aiohttp/client.py
```diff
@@ -582,9 +582,8 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         def __aenter__(self):
             return self
 
-        @asyncio.coroutine
         def __aexit__(self, exc_type, exc_val, exc_tb):
-            self.close()
+            return self.close()
```
asvetlov: Why `return`? I think it should be `yield from`

litwisha: I have done `timeit` for those two cases, and your variant is faster. Fixed it.

Issue #2032, opened 2017-06-28
Making tests with AioHTTPTestCase


cecton:
## Long story short

I'm trying to make my test to run on the same port instead of a random port. The reason is that I'm making integration tests and I need it to receive calls from another service that I can't really configure the port easily.

My idea is that I need to make a class FixedPortTestServer that will inherit from TestServer and override the `start_server` method to force a specific port.

Then I need to pass this instance to the TestClient initialization. The AioHTTPTestCase use the result of `get_application` and give it to TestClient init parameters. This makes me think that I can make `get_application` return the FixedPortTestServer instance instead of the application instance:

```python
    async def get_application(self):
        return FixedPortTestServer(my_app)
```

Which is wrong... Now I have a properly initialized TestClient using my TestServer but `self.app` in the AioHTTPTestCase give me the TestServer instead of my application which doesn't make any sense.

## Expected behaviour

I expect to get the application from `self.app` within a test method of AioHTTPTestCase and I want to be able to override the TestServer instance with whatever BaseTestServer instance I want without needing to override a private method.

## Actual behaviour

Currently, my only way to override the TestServer instance in the initialization of the TestClient is to override the private method `_get_client` in the AioHTTPTestCase.

## Steps to reproduce

```python
import asyncio
import socket
from aiohttp.helpers import sentinel
from aiohttp.test_utils import AioHTTPTestCase, TestServer, unittest_run_loop
from yarl import URL

from mymodule import myapp


class FixedPortTestServer(TestServer):
    @asyncio.coroutine
    def start_server(self, loop=None, **kwargs):
        if self.server:
            return
        self._loop = loop
        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self._socket.bind((self.host, 8080)) # force port
        self.port = self._socket.getsockname()[1]
        self._ssl = kwargs.pop('ssl', None)
        if self.scheme is sentinel:
            if self._ssl:
                scheme = 'https'
            else:
                scheme = 'http'
            self.scheme = 'http'
        self._root = URL('{}://{}:{}'.format(self.scheme,
                                             self.host,
                                             self.port))

        handler = yield from self._make_factory(**kwargs)
        self.server = yield from self._loop.create_server(
            handler, ssl=self._ssl, sock=self._socket)


class MyTestCase(AioHTTPTestCase):
    async def get_application(self):
        return FixedPortTestServer(myapp)

    @unittest_run_loop
    async def test_initialization(self):
[...]

PR #2074, opened 2017-07-10
Added get_client and get_server methods to AioHTTPTestCase

cecton:
This change is needed if we want a clear distinction in the
unittest.TestCase between "app" (the application) and "server" the
actual test server. This will also allow a developer to override these
methods to use their own TestServer and TestClient.

## What do these changes do?

1. It transform the method `_get_client()` of AioHTTPTestCase to a public method (inviting the developer to override it)
2. It adds a method `get_server()` to AioHTTPTestCase that will, by default, instantiate the default TestServer
3. It replaces the flexible first parameter of TestClient which was `app_or_server` to `server` only
4. It removes the now unrelevant parameter host an port of TestClient

1. (breaking change) You can no longer return a BaseTestServer instance from `get_application`. `get_application` must only return an Application instance.
2. If you want a different TestServer you must override `get_server`.
2. You can now override `get_client` method because it is now a public method (you always could but... now it's more obvious that you actually should).
3. (breaking change) `_get_client` has change of name.

## Related issue number

#2032 

## Checklist

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -1,4 +1,5 @@
 import asyncio
+import collections
 import contextlib
 import tempfile
 import warnings
@@ -9,8 +10,8 @@ from py import path
 from aiohttp.web import Application
 
 from .test_utils import unused_port as _unused_port
-from .test_utils import (RawTestServer, TestClient, TestServer, loop_context,
-                         setup_test_loop, teardown_test_loop)
+from .test_utils import (BaseTestServer, RawTestServer, TestClient, TestServer,
+                         loop_context, setup_test_loop, teardown_test_loop)
 
 
 try:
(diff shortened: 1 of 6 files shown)


asvetlov:
I'm ok with adding `get_client()` and `get_server()` public methods to `AioHTTPTestCase`.
But prohibiting an application passing to `TestClient` is too invasive change (you could notice how many tests are touched by this PR).

cecton:
I think you're right....

My point is mostly for the unittest's TestCase where it makes much more sense and also makes the code simple (if you compare the TestClient class before and after, it's clearly just easier). But I completely ignored the *test_* methods.

I'm giving it another try.

asvetlov:
You proposal for `TestClient` ctor change makes sense but ship has sailed more than year ago.
Historically the client supported `Application` only, after that I added `TestServer` also.
I could live with accepting both app and server. Not very elegant code but handy for usage.

Hmm. We could change `test_client` fixture and get rid of application in `TestClient` itself.
It should be relative smooth transition.

cecton:
Yes that's what I just realized. I need to make the tests in test_test_utils.py files more elegant but it's getting closer.

cecton:
Actually no I can't make it look better because the purpose of the `test_test_utils.py` is to test `test_utils.py`. Since I changed the API, there are changes in those files. Makes sense.

I just made the passing of argument loop more explicit.

@asvetlov do you think this PR looks much better now?

aiohttp/pytest_plugin.py
```diff
@@ -227,19 +228,22 @@ def test_client(loop):
...
+                not isinstance(__param, (Application, BaseTestServer)):
+            __param = __param(loop, *args, **kwargs)
+            kwargs = {}
+        else:
             assert not args, "args should be empty"
+
+        if isinstance(__param, Application):
+            server = TestServer(__param, loop=loop, **server_kwargs)
+            client = TestClient(server, loop=loop, **kwargs)
+        elif isinstance(__param, BaseTestServer):
```
cecton, line 243: I dropped here the duplicated test on TestServer and RawTestServer. The argument to TestClient must now be a BaseTestServer instance which includes both. Maybe someone would want to make a TestServer not based on TestServer but on BaseTestServer. That also reflects better the first argument of TestClient.

aiohttp/pytest_plugin.py
```diff
@@ -227,19 +228,22 @@ def test_client(loop):
...
-        elif isinstance(__param, TestServer):
-            assert not args, "args should be empty"
-            client = TestClient(__param, loop=loop, **kwargs)
-        elif isinstance(__param, RawTestServer):
+    def go(__param, *args, server_kwargs={}, **kwargs):
+
+        if isinstance(__param, collections.Callable) and \
+                not isinstance(__param, (Application, BaseTestServer)):
+            __param = __param(loop, *args, **kwargs)
+            kwargs = {}
```
cecton, line 236: This reproduces the exact behavior than before but a little more strict:
1. __param must be a callable, in the previous code it was the "else" case
2. kwargs is reset just after because we don't want them to be passed to the client or server (identical than before)
3. by extension, the result of the callable must be an instance of Application or BaseTestServer or it would complain explicitly about it

cecton:
I just added setUpAsync and tearDownAsync. They are very useful in development because you don't need to write synchronous methods for the setUp and tearDown anymore, you can simply override those. (If you want I have made tests cases for those in particular, it instantiate the test case in a test and check that both functions executed.)

That was my last change for this PR, thanks for your time.

cecton:
@asvetlov this is ready for your review.

asvetlov:
Thanks!

PR #2088, opened 2017-07-13
Fix bug in CookieJar when a cookie would be expired by mistake

### LINKED ISSUE #2084: Subsequent removal and addition of a cookie may leave the cookie removed
## Long story short

I encountered a flaky test failure when testing login-logout-login and discovered that a login right after a logout may not work, in other words, the authentication cookie may not be set.

This boiled down to the following error in `CookieJar`: for event loops with poor time resolution (in my case `uvloop`), subsequent removal and addition of a cookie may not add the cookie.

Here is a test case:
```python
import time

from aiohttp.cookiejar import CookieJar
from uvloop import new_event_loop


def test_uvloop_cookies():
    loop = new_event_loop()

    for _ in range(1000):
        jar = CookieJar(unsafe=True, loop=loop)

        # Remove`test` cookie.
        jar.update_cookies(SimpleCookie('test=""; expires=Thu, 01 Jan 1970 00:00:00 GMT; Max-Age=0; Path=/'))
        # Set `test` cookie to `bar`.
        jar.update_cookies(SimpleCookie('test="bar"'))

        # Assert that there a cookie.
        assert list(jar)
        time.sleep(0.1)
        assert list(jar)
```

Although it took me quite a bit of time to discover what's going on, it seems there is an extremely simple fix: replace strict comparison (`<`) with non-strict one (`<=`) [here](https://github.com/aio-libs/aiohttp/blob/eee65a4eed846394082c5e931fc75f9c1e7f30cd/aiohttp/cookiejar.py#L77).
I'll provide a PR soon if maintainers agree.

## Expected behaviour

Test passes

## Actual behaviour

Test fails

## Steps to reproduce

Run the test

## Your environment

uvloop=0.8.0, aiohttp=2.1.0

roganov:
## What do these changes do?
Fixes #2084
Unfortunately I couldn't come up with a test case other than to record uvloop's timestamps and replay them. Obviously it highly depends on the internals on CookieJar and the test may become irrelevant it code changes.

Hopefully not.

## Related issue number
#2084 

## Checklist

### DIFF
--- a/tests/test_cookiejar.py
+++ b/tests/test_cookiejar.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import itertools
 import os
 import tempfile
 import unittest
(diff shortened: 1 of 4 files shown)


tests/test_cookiejar.py
```diff
@@ -577,3 +578,21 @@ def test_invalid_values(self):
...
+            timestamps, itertools.cycle([timestamps[-1]]))
+
+        jar = CookieJar(unsafe=True, loop=loop)
+        # Remove`foo` cookie.
+        jar.update_cookies(SimpleCookie('foo=""; Max-Age=0'))
+        # Set `foo` cookie to `bar`.
+        jar.update_cookies(SimpleCookie('foo="bar"'))
+
+        # Assert that there a cookie.
+        assert jar
```
asvetlov: Please use more obvious check.
`assert len(jar) == 1` for example

roganov: Done.

asvetlov:
Thanks!

Issue #2092, opened 2017-07-14
Signals Startup & Cleanup still not good


cecton:
## Long story short

In my application:
 *  In attribute
     *  Client sessions to connect to APIs
 *  On startup:
     *  Background tasks
     *  Foreground asynchronous code
 *  On cleanup:
     *  Cancel background tasks
     *  Close sessions

My goal is to properly clean up and exit the application if one of the startup hook did not succeeded. But I just can't figure out a way at the moment to properly cleanup my application on exit. 

## Expected behaviour

1. ~~Failure during a startup hook should prevent the application to startup~~ (cannot reproduce)
2. Cleanup should be sent even if the startup signal failed
3. We should have an helper for background task because it's way too complicated to handle properly
4. Any error during a startup hook should be displayed properly

## Actual behaviour

1.  The cleanup signal is sent only if all the startup signals succeeded. I believe this is problematic at least in my case because I need to close the client sessions in a kind of "finally" statement.

The following workaround will work for failures during startup signal, but this call to `app.cleanup` will be duplicated with the one in web.run_app if the application did not fail during startup but is cancelled by a manual KeyboardInterrupt:

```python
loop = asyncio.get_event_loop()
try:
    web.run_app(app, loop=loop)
except (SystemExit, KeyboardInterrupt):
    loop.run_until_complete(app.cleanup())
finally:
    loop.close()
```

2.  Now I remember: any code executed in a startup hook that raises an exception does not see its exception being retrieved (`Task exception was never retrieved`).

```python
def safe_startup(func):
    async def wrapper(app):
        try:
            await func(app)
        except:
            logger.exception("Startup failed")
            raise
    return wrapper
```

3.  An exception during a cleanup hook prevents the rest of the cleanup hooks to be executed. But that's probably fine.

```python
async def stop_my_bg_task(app):
    if 'my_bg_task' in app:
        app['my_bg_task'].cancel()
        try:
            await app['my_by_task']
        except:  # NOTE: awaiting on a failed bg task retrieves also the exception
            pass  # NOTE: hide that exception
```

## Steps to reproduce

(Examples given above)

## Your environment

ArchLinux, Python 3.6, aiohttp 2.2.3

asvetlov:
We don't support atomic execution for startup/cleanup pair for pairs like `start_my_bg_task`/`stop_my_bg_task`.
But we could run cleanup stage unconditionally and use trick like:
```
async def cleanup(app):
    bg_task = app.get('bg_task')
[...]

PR #2096, opened 2017-07-14
Improve startup cleanup signals

cecton:
## What do these changes do?

1. cleanup signal is sent even if startup signal failed
2. ~helpers to create and manage background tasks~

1. The cleanup hooks will be called more often at the end of the process. You now try to connect to a database in a startup hook and get your client session cleaned automatically in the cleanup even if the database connection failed.
2. ~You no longer need to use the dict methods of Application to create tasks to manage background processes, you can use the helper `add_task` and provide a coroutine. The coroutine will be called with the app argument. When the application finishes, the tasks will be cancelled and awaited properly (except if a task raise SystemExit or any not Exception based exception).~

## Related issue number

#2092 

## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -420,9 +420,10 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
         loop = asyncio.get_event_loop()
 
     app._set_loop(loop)
-    loop.run_until_complete(app.startup())
 
     try:
+        loop.run_until_complete(app.startup())
+
         make_handler_kwargs = dict()
         if access_log_format is not None:
             make_handler_kwargs['access_log_format'] = access_log_format
--- a/changes/2092.feature
+++ b/changes/2092.feature
@@ -1,0 +1,1 @@
+Cleanup signal is sent even if an error occur during startup signal
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -1299,6 +1299,14 @@ duplicated like one using :meth:`Application.copy`.
           async def on_cleanup(app):
               pass
 
+      .. versionchanged:: 2.3.0
+
+         The cleanup signal is sent even if an error occurs during the
+         processing of the startup signal. An error during startup will
+         probably leave your application partially initialized. If you are
+         using the :class:`Application` dictionary, you must check first that
+         the key exists before cleaning up anything.
+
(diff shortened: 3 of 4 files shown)


cecton:
@asvetlov  I tested on 3.4.6, 3.5.3, 3.6.0 and 3.6.1 and it works. Unfortunately asyncio is a bit different in 3.4.3 and 3.4.2. Do you really want to support those specific versions?

cecton:
(In 3.4.4 it works already)

cecton:
@asvetlov last one but biggest one. You can maybe cherry pick ca5fb94 separately if the rest doesn't meet your expectations.

asvetlov:
Hmm. I definitely don't want `app.add_task()` and others.
People was reluctant for incorporating `aiojobs`, your proposal overlaps with the library a little bit.
Could you make a PR for ca5fb94 only?
I can cherry-pick it but prefer well-formed pull requests :)

cecton:
@asvetlov done! I will check aiojobs later to see if I can replace some code in my projects.

Besides I think it's still confusing, the documentation says that you can spawn background processes in that fashion: http://aiohttp.readthedocs.io/en/stable/web.html#background-tasks

It mostly works but it is not perfect, there is a little bit more to do to get a background job working, logging and terminating properly.

If people find that aiohttp should not handle background jobs at all, then it would make more sense to remove this example of the documentation.

If background jobs are okay, then we should either give an example that works perfectly or implement the helpers like I did (my helpers are still 37 lines of code in Application (still smaller than aiojobs).

asvetlov:
Docs should be updated, sure.
Let's left this task to other PR.

Regarding to this one I think we need mention the behavior in docs too both in Usage and Reference.
After PR applying user should always check if her *key* is in app dict.

It is a breaking change.
On other hand we definitely don't want to have partially initialized application.
Also it correlates with #2121 a little

Mybe we need a `ServerStartUp` object with methods for saving all activities: startup signals and created handlers.
The object could be used for proper shutdown procedure.

Now it is a wild idea, let me sleep on it.

cecton:
@asvetlov I haven't give much thought about this PR yet but I suppose we can conclude it by adding in the documentation that the startup/cleanup behavior is changed. I've just added it to the doc, I think we are good.

asvetlov:
Better solution proposed

PR #2097, opened 2017-07-14
Add Eyepea + ALLOcloud links and better links rendering

GMLudo:
Hi,

As discussed with @asvetlov yesterday, this links of Eyepea and ALLOcloud, where we use a lot aiohttp with success ;-)
At the same time, I have converted all links with a title.

You might use alphabetical ordering for this list ?

Thanks for the merge.


asvetlov:
Thanks.
I don't think we need alphabetical order now, while number of links is relative low.

Issue #1992, opened 2017-06-19
[enhancement] Add ability to disable automatic decompression of body


thehesiod:
In our library we'd like the ability to disable the automatic decompression of the request body as we need to compare a hash in the headers to the original body.

Right now this doesn't seem to be possible as the HTTP parser creates a DeflateBuffer which will automatically decompress the body.  If guidance is given how this should be achieved I can create a PR and test case.

fafhrd91:
let's split this into two parts

1. add extra parameter to HttpParser, so it can create HttpPayloadParser with compression=None. This should with both python parser and c-extension parser. this should be simple
2. modify client/web api. this is complicated part, because decompression could start before handler get called.

@asvetlov thoughts?

asvetlov:
I believe @thehesiod talks about client API.
It could be done by simple flag in `session.get()` and family.

thehesiod:
I just noticed that if I try a aiobotocore s3 client `get_object` of a key that a `Content-Encoding` metadata set to `gzip`, it's automatically decompressing the body as well, not-good...this means it returns the wrong datatype for an expected operation due to the automatic decompression.  I'll link another bug report to this.

thehesiod:
this PR only handles `ClientResponse`, other functionality can be added in separate PRs

thehesiod:
ok, created PR which resolves this

PR #2110, opened 2017-07-17
option to disable automatic client response body decompression

thehesiod:
enhancement for https://github.com/aio-libs/aiohttp/issues/1992


thehesiod:
this PR only handles ClientResponse, other functionality can be added in separate PRs

asvetlov:
Good note!
Do you object parameter name, isn't it?

fafhrd91:
right now aiohttp applies two type of transformation, `transfer-encoding` and `content-encoding`.
I'd disable both with one parameter.

asvetlov:
Makes sense.

asvetlov:
@cecton could you take a look on failed tests?
Sure they could accept `loop` fixture explicitly but I worry about backward degradation in our test tools.
Looks like the PR was passed without errors unless I've merged your changes into master.

tests/test_test_utils.py
```diff
@@ -58,7 +69,39 @@ def test_get_route():
...
-                assert "Hello, world" in text
+                assert _hello_world_str == text
+
+            loop.run_until_complete(test_get_route())
+
+
+def test_auto_gzip_decompress():
+    with loop_context() as loop:
+        app = _create_example_app()
+        with _TestClient(app, loop=loop) as client:
```
cecton: @asvetlov  Here is the problem. If you really want to use TestClient directly, you need a  TestServer instance. It's best to use the `test_client` fixture if you can:

```python
def test_auto_gzip_decompress(test_client):
    app = _create_example_app() 
    with test_client(app) as client:
```

Otherwise:

```python
     with loop_context() as loop:
         app = _create_example_app()
         with _TestClient(_TestServer(app, loop=loop), loop=loop) as client:
```

asvetlov: I see no significant degradation.
@thehesiod please update tests to use `test_client` fixture explicitly

thehesiod: I've updated the tests to run like the tests surrounding it, I can't use `test_client` fixture because I use `_create_example_app`

thehesiod:
ya I think botocore may need transfer-encoding disabled as well.  I'll try working on that today.

tests/test_test_utils.py
cecton: @thehesiod I don't understand the problem. As far as I know you can give the app created by `_create_example_app` as argument to `test_client`. The fixture gives you a function, not a client instance.

tests/test_test_utils.py
```diff
@@ -58,7 +69,39 @@ def test_get_route():
...
+        app = _create_example_app()
+        with _TestClient(_TestServer(app, loop=loop), auto_decompress=False) as client:
+
+            @asyncio.coroutine
+            def test_get_route():
+                nonlocal client
+                resp = yield from client.request("GET", "/gzip_hello")
+                assert resp.status == 200
+                data = yield from resp.read()
+                assert data == _hello_world_gz
```
cecton, line 105: This test coroutine is duplicated at a lot of different places. Maybe you can make a helper outside the test and provide the client instance in argument. I think it would be more readable. What do you think?

```python
async def _assertBody(client, url, body):
    resp = await ...
    assert resp.status == 200
    data = await ...
    assert data == body
```

tests/test_test_utils.py
```diff
@@ -14,11 +15,20 @@
                                 teardown_test_loop, unittest_run_loop)
 
 
-def _create_example_app():
+_hello_world_str = "Hello, world"
```
cecton, line 18: At this rate you can even generate a random string. Like:

```python
uuid.uuid4().hex
```

I would actually even put the str, bytes and gzip inside the app instance:

```python
app = web.Application()
app["body_str"] = uuid.uuid4().hex
app["body_bytes"] = app["body_str"].encode('utf-8')
app["body_gz"] = gzip.compress(app["body_bytes"])
```

tests/test_test_utils.py
thehesiod: I've been avoiding too much refactoring because @asvetlov has frowned upon that in the past in PRs that implement new features. He's wanted these types of changes in multiple PRs. If he gives the go-ahead I can do these suggested changes.

thehesiod:
@fafhrd91 @asvetlov are there any aiohttp examples of client/server gzip transfer encoding?  I don't see any tests or examples in the web.

tests/test_test_utils.py
thehesiod: I must have been looking at an old version of that function, in either case all the functions should be updated, so if @asvetlov says ok I can change them all to use this fixture

fafhrd91:
what do you mean by example?  implementation is shared between client and server, 
http_parser.DeflateBuffer and http_writer.PayloadWriter for reading and for writing

tests/test_test_utils.py
cecton: The fixture `test_client` won't work... I just realized that this fixture is overridden in the test file itself.

asvetlov:
Thanks!

PR #2111, opened 2017-07-18
Throttle simultaneous DNS requests #1924

pfreixes:
## What do these changes do?

Allows to the connector throttle simultaneous DNS requests for a specific
domain when the DNS cache is enabled, having this enabled by default.

This feature get rid of the dogpile side effect when there is a miss in the cache, having
just one request doing the DNS resolution - at hostname granularity, requests that continue arriving while the ongoing DNS resolution is not finished will wait till it finishes.

When the DNS cache is not used. All requests will end up making a DNS query.

No

## Related issue number

No

## Checklist


pfreixes:
Trying to get a green pipeline, just a copy of the https://github.com/aio-libs/aiohttp/pull/1924. Now issues with the output of the coverage having a really weird error

```
INTERNALERROR>   File "/home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages/_pytest/main.py", line 105, in wrap_session
INTERNALERROR>  ........
```

Again 

> it works on my computer

Any recommendation to fix that issue with the pipeline? @asvetlov @kxepal @fafhrd91

fafhrd91:
do not worry about "Python: nightly". 
your build actually fail on "migration.rst:152:api’s:"

https://travis-ci.org/aio-libs/aiohttp/jobs/255018892#L491

aiohttp/locks.py
```diff
@@ -0,0 +1,77 @@
+import asyncio
+import collections
+
+from .helpers import create_future
+
+
+class Event:
```
kxepal: Isn't it possible to inherint from asyncio.Event and override set method instead? I'm not sure we want to maintain whole own Even lock.

pfreixes:
Thanks for the tip @fafhrd91 fixed.

Aside of that, after some research I decided to get rid of the full Event class, the same behavior can be reached with the new and simple wrapper called `ErrorfulOneShotEvent` (thanks to Nathaniel for that approx [1])

**Disclaimer**, there is still a red-flag raised by @asvetlov regarding the Futures related to these tasks pending to be awakened, the whole thread can be followed here [2]. BTW my opinion about that is still the same, the proposed implementation is not nocive at all or at least is guided by the same principles than the current one. 

[1] https://mail.python.org/pipermail/python-ideas/2017-July/046486.html
[2] https://github.com/aio-libs/aiohttp/pull/1924

aiohttp/locks.py
pfreixes: No longer used that pattern

pfreixes:
Any chances to keep the discussion that we had @asvetlov ? any chances to gather the opinion from other maintainers such as @fafhrd91 ?

fafhrd91:
you should to cleanup all event in connector's close method. just set `CancelledError`.

otherwise I think PR is fine.

aiohttp/connector.py
```diff
@@ -597,10 +598,18 @@ def __init__(self, *, verify_ssl=True, fingerprint=None,
 
         self._use_dns_cache = use_dns_cache
         self._cached_hosts = _DNSCacheTable(ttl=ttl_dns_cache)
+        self._throttle_dns_events = {}
         self._ssl_context = ssl_context
         self._family = family
         self._local_addr = local_addr
 
+    def close(self):
```
pfreixes, line 606: @fafhrd91 and @asvetlov now the throttled requests are canceled if the connector is closed explicitly. Not really sure if I'm happy with the implementation, to make it possible I had to save all waiters and chain them with the waiters related to the `Event`, take a look alternatives are welcomed.

fafhrd91:
this is for 2.3 release. otherwise lgtm. 

@asvetlov ?

pfreixes:
@asvetlov looks you have the last word for this PR. BTW adapted to last CHANGES protocol and one small fix about the explicit loop.

kxepal:
Please do rebase feature branch against master instead of merging master into feature branch. Twisted git log history is quite hard to follow.

asvetlov:
@kxepal it doesn't needed, we usually do quash PRs anyway

kxepal:
@asvetlov ah, cool then!

aiohttp/locks.py
```diff
@@ -0,0 +1,42 @@
+import asyncio
+import collections
+
+from .helpers import ensure_future
+
+
+class ErrorfulOneShotEvent:
```
asvetlov: The name is too long.
Also it starts with `Error`, usually this schema is used for nameing exception classes.

pfreixes: Changed to `EventResultOrError`, still long but at least more meaningful. Let me know if it suits better for you.

asvetlov:
Ok, thanks. The name is not public, we could change it later anyway.

Issue #2123, opened 2017-07-23
Suggest: add Access Logging option to show `X-Forwarded-For`


harobed:
Suggest: add Access Logging option to show `X-Forwarded-For`

Exemple `%f` option return `X-Forwarded-For` value or Remote IP-address if `X-Forwarded-For` don't exist.

asvetlov:
I suggest another change: pass `request` instead of `message` into `AccessLogger.log` (since aiohttp 2.0 `web_protocol.RequestHandler` always create a request instance.
In turn `web.BaseRequest` has `req.host` property, the property respects both `Forwarded` and `X-FORWARDED-HOST` headers.

asvetlov:
@harobed do you want to make a PR for implementing my proposal?

harobed:
Yes, maybe this evening.

PR #2124, opened 2017-07-23
Add X-Forward-For option to Access Log #2123

harobed:
Add X-Forward-For option to Access Log #2123


asvetlov:
Should be implemented as suggested in #2123

Issue #2109, opened 2017-07-17
do we need http pipelining?


fafhrd91:
@asvetlov @kxepal thoughts?

asvetlov:
Not sure.
On one hand why not if we could support it easy.
On other hand if pipelining requires extra steps from users they never do it.

fafhrd91:
removing pipelining would simplify request handling flow. (some speedup is expected too) 
pipelining requires read/write ordering. I doubt pipelining is in use, especially for python servers.

fafhrd91:
on other hand pipelining help in synthetic benchmarks

kxepal:
I'm not sure about as well. If HTTP/1 pipelining support cause maintenance issues, slows down general HTTP usage and gives only profit in synthetic benchmark, I think there is nothing much to think about. 

It's pretty cool, but not very popular feature for HTTP/1, but it's a core one of HTTP/2. May be move it HTTP/2 implementation if any will happens?

Btw, do we know any aiohttp usage with pipelining in the wild?

fafhrd91:
Pipelining gives nice 25% boost in synthetic tests :)

socketpair:
https://medium.freecodecamp.org/million-requests-per-second-with-python-95c137af319

fafhrd91:
how much code does it actually run? add business logic written in python and some logging, and this thing would show same speed as any other python framework.

fafhrd91:
I added pipelining because of benchmarks :)

samuelcolvin:
I think pipelining is only really used in tests. The mistake it tests/benchmarks allowing it/being too simple. 

For me, remove if it makes code simpler or provides any other speedups.

pfreixes:
I have a proposal for that.

Right now Aiohttp implements HTTP pipelining but this feature is not fully enabled - correct me if Im wrong @fafhrd91. The default concurrence per connection [1] does not allow Aiohttp to process more than one request at a time. 

Having this scenario and taking into account the complexity that this feature adds in terms of code and the low usage in production systems, almost in microservices architectures,  I would propose to reduce the current HTTP pipelining to a more naive implementation that **does not support concurrence in the same connection**.

Why ? 

1) Reduce the codebase and as a result the performance footprint.
2) Give still support for HTTP pipelining for the none happy path and meeting the HTTP 1.1 requirements.

How ?

[...]

PR #2126, opened 2017-07-24
Slow request body copy

arthurdarcet:
When sending a request with a large body (e.g.`session.post('…', data=b'0' * 2_000_000_000)`,) `PayloadWriter._write` is taking very long (5+ seconds for a 2G body).

This is especially problematic because this locks the event loop and nothing else gets executed (so if you are running this in a request handler, the whole server becomes unresponsive).

As far as I understand this, there is no need to actually copy the data before passing it to the underlying asyncio transport: its `write` method handles fragmented chunks and we spare a big `''.join(buffer)`.

This leads to a x2 speed up for the actual write, what ever the actual body size is (I tested with bodies from 2KB to 2GB)

## What do these changes do?

When the asyncio transport is open, leave the buffering to the OS and write data directly to it.
I have also factorised some code, to keep the `_buffer` / `_transport` variables in a clear state (exactly one of them is None at all time)

The headers are not buffered until the first byte of the body is ready if the transport is already there. I'm not sure why this was done, I can't find anything related to that in the git history. @fafhrd91, since this is your code, any idea why sending the headers was buffered even when the transport is ready?

Also, see the changed test: some user might get a ClientPayloadError instead of a ServerDisconnectedError since the headers are now sent even if write_eof raises. This is mostly a side effect of the tests: in real life I would expect users already receiving a bit of both `ClientPayloadError` and `ServerDisconnectedError` for this use case (depending on when exactly the connection was cut off).

### DIFF
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -154,8 +154,7 @@ class PayloadWriter(AbstractPayloadWriter):
 
         if self._drain_waiter is not None:
             waiter, self._drain_waiter = self._drain_waiter, None
-            if not waiter.done():
-                waiter.set_result(None)
+            waiter.set_result(None)
 
     async def get_transport(self):
         if self._transport is None:
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -148,8 +148,8 @@ class PayloadWriter(AbstractPayloadWriter):
         self._transport = transport
 
         if self._buffer is not None:
-            for b in self._buffer:
-                transport.write(b)
+            for chunk in self._buffer:
+                transport.write(chunk)
             self._buffer = None
 
         if self._drain_waiter is not None:
@@ -160,10 +160,9 @@ class PayloadWriter(AbstractPayloadWriter):
     async def get_transport(self):
         if self._transport is None:
             if self._drain_waiter is None:
-                self._drain_waiter = create_future(self.loop)
+                self._drain_waiter = self.loop.create_future()
             await self._drain_waiter
 
-        assert self._transport is not None
         return self._transport
 
     @property
@@ -195,7 +194,6 @@ class PayloadWriter(AbstractPayloadWriter):
 
         # see set_transport: exactly one of _buffer or _transport is None
         if self._transport is not None:
-            assert self._buffer is None
             self._transport.write(chunk)
         else:
             self._buffer.append(chunk)
@@ -272,7 +270,7 @@ class PayloadWriter(AbstractPayloadWriter):
         self._transport = None
         self._stream.release()
 
-    async def drain(self, last=False):
+    async def drain(self):
         if self._transport is not None:
             await self._stream.drain()
         else:
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -124,7 +124,8 @@ class PayloadWriter(AbstractPayloadWriter):
 
     def __init__(self, stream, loop, acquire=True):
         self._stream = stream
-        self._transport, self._buffer = None, []
+        self._transport = None
+        self._buffer = []
 
         self.loop = loop
         self.length = None
@@ -137,22 +138,33 @@ class PayloadWriter(AbstractPayloadWriter):
         self._drain_waiter = None
 
         if self._stream.available:
-            self._transport, self._buffer = self._stream.transport, None
+            self._transport = self._stream.transport
+            self._buffer = None
             self._stream.available = False
         elif acquire:
             self._stream.acquire(self)
 
     def set_transport(self, transport):
-        assert self._transport is None and self._buffer is not None
+        self._transport = transport
 
-        for b in self._buffer:
-            transport.write(b)
-        self._transport, self._buffer = transport, None
+        if self._buffer is not None:
+            for b in self._buffer:
+                transport.write(b)
+            self._buffer = None
 
         if self._drain_waiter is not None:
             waiter, self._drain_waiter = self._drain_waiter, None
             if not waiter.done():
                 waiter.set_result(None)
+
+    async def get_transport(self):
+        if self._transport is None:
+            if self._drain_waiter is None:
+                self._drain_waiter = create_future(self.loop)
+            await self._drain_waiter
+
+        assert self._transport is not None
+        return self._transport
 
     @property
     def tcp_nodelay(self):
@@ -265,7 +277,4 @@ class PayloadWriter(AbstractPayloadWriter):
             await self._stream.drain()
         else:
             # wait for transport
-            if self._drain_waiter is None:
-                self._drain_waiter = self.loop.create_future()
-
-            await self._drain_waiter
+            await self.get_transport()
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -124,7 +124,7 @@ class PayloadWriter(AbstractPayloadWriter):
 
     def __init__(self, stream, loop, acquire=True):
         self._stream = stream
-        self._transport = None
+        self._transport, self._buffer = None, []
 
         self.loop = loop
         self.length = None
@@ -133,23 +133,21 @@ class PayloadWriter(AbstractPayloadWriter):
         self.output_size = 0
 
         self._eof = False
-        self._buffer = []
         self._compress = None
         self._drain_waiter = None
 
         if self._stream.available:
-            self._transport = self._stream.transport
+            self._transport, self._buffer = self._stream.transport, None
             self._stream.available = False
         elif acquire:
             self._stream.acquire(self)
 
     def set_transport(self, transport):
-        self._transport = transport
+        assert self._transport is None and self._buffer is not None
 
-        chunk = b''.join(self._buffer)
-        if chunk:
-            transport.write(chunk)
-            self._buffer.clear()
+        for b in self._buffer:
+            transport.write(b)
+        self._transport, self._buffer = transport, None
 
         if self._drain_waiter is not None:
             waiter, self._drain_waiter = self._drain_waiter, None
@@ -178,25 +176,15 @@ class PayloadWriter(AbstractPayloadWriter):
                      if encoding == 'gzip' else -zlib.MAX_WBITS)
         self._compress = zlib.compressobj(wbits=zlib_mode)
 
-    def buffer_data(self, chunk):
-        if chunk:
-            size = len(chunk)
-            self.buffer_size += size
-            self.output_size += size
-            self._buffer.append(chunk)
-
     def _write(self, chunk):
         size = len(chunk)
         self.buffer_size += size
         self.output_size += size
 
+        # see set_transport: exactly one of _buffer or _transport is None
         if self._transport is not None:
-            if self._buffer:
-                self._buffer.append(chunk)
-                self._transport.write(b''.join(self._buffer))
-                self._buffer.clear()
-            else:
-                self._transport.write(chunk)
+            assert self._buffer is None
+            self._transport.write(chunk)
         else:
             self._buffer.append(chunk)
 
@@ -241,11 +229,7 @@ class PayloadWriter(AbstractPayloadWriter):
         headers = status_line + ''.join(
             [k + SEP + v + END for k, v in headers.items()])
         headers = headers.encode('utf-8') + b'\r\n'
-
-        size = len(headers)
-        self.buffer_size += size
-        self.output_size += size
-        self._buffer.append(headers)
+        self._write(headers)
 
     async def write_eof(self, chunk=b''):
         if self._eof:
@@ -268,9 +252,9 @@ class PayloadWriter(AbstractPayloadWriter):
                     chunk = b'0\r\n\r\n'
 
         if chunk:
-            self.buffer_data(chunk)
+            self._write(chunk)
 
-        await self.drain(True)
+        await self.drain()
 
         self._eof = True
         self._transport = None
@@ -278,10 +262,6 @@ class PayloadWriter(AbstractPayloadWriter):
 
     async def drain(self, last=False):
         if self._transport is not None:
-            if self._buffer:
-                self._transport.write(b''.join(self._buffer))
-                if not last:
-                    self._buffer.clear()
             await self._stream.drain()
         else:
             # wait for transport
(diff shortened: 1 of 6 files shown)


arthurdarcet:
Also, kind of related:
I have not seen any section of the documentation relating to large request/response bodies, both from the server and client point of view.
There are some gotchas when dealing with this in production, to avoid locking the event loop and the server, and it would be nice to gather a list of best practice on how to handle heavy requests.
This section could also mention that doing CPU-intensive jobs on the main thread/process will lock the GIL, the event loop, and the server which might be unexpected for someone coming from multi-threaded frameworks like Flask/…

arthurdarcet:
And lastly, using `session.post('…', data=io.BytesIO(…))` makes everything work correctly, but if I'm not mistaken, this isn't documented anywhere (i had to read the code to see that chunking was happening for BytesIO and not for raw bytes)

I think one of three things should be done:
 - This should be documented somewhere, probably in this same "big requests" section of the docs I mentioned earlier
 - The `payload.BytesPayload` class should do some chunking automatically in my opinion, but I can understand that it makes sense to reserve that for BytesIO payloads.
 - The `BytesPayload` class could log a warning when a payload above ~1MB is passed directly (something like `2 * streams.DEFAULT_LIMIT`?)

Let me know which solution(s) sound good, and I can open PR for them

aiohttp/http_writer.py
```diff
@@ -129,7 +129,7 @@ class PayloadWriter(AbstractPayloadWriter):
 
     def __init__(self, stream, loop, acquire=True):
         self._stream = stream
-        self._transport = None
+        self._transport, self._buffer = None, []
```
fafhrd91: use separate lines

aiohttp/http_writer.py
```diff
@@ -138,29 +138,38 @@ def __init__(self, stream, loop, acquire=True):
         self.output_size = 0
 
         self._eof = False
-        self._buffer = []
         self._compress = None
         self._drain_waiter = None
 
         if self._stream.available:
-            self._transport = self._stream.transport
+            self._transport, self._buffer = self._stream.transport, None
```
fafhrd91: separate lines

tests/test_web_server.py
```diff
@@ -90,8 +90,9 @@ def handler(request):
     server = yield from raw_test_server(handler, logger=logger)
     cli = yield from test_client(server)
 
-    with pytest.raises(client.ServerDisconnectedError):
-        yield from cli.get('/path/to')
+    resp = yield from cli.get('/path/to')
+    with pytest.raises(client.ClientPayloadError):
```
fafhrd91: why exception is different?

fafhrd91:
regarding payloads. I agree. you are welcome to make PR!

arthurdarcet:
@fafhrd91 you agree with using chunking for `BytesPayload`, or with logging a warning for large `bytes` payloads?

fafhrd91:
all tree. but I am not sure if chunking in BytesPayload makes sense.

arthurdarcet:
i would say, either we handle the chunking in `BytesPayload` and it does not really make sense to log a warning (nothing bad would be happening with big payloads); or we do not do any chunking there (like now), and we log something to warn the user that sending big payloads as bytes directly is a bad idea

For the documentation, I … don't really have the time to write it… but I'm planning on writing a blog article related to our experience with all this. I think it will be easy to extract some guidelines from there and i will open a doc PR when i'm ready. Do you want me to open another pending issue to track this until then?

aiohttp/http_writer.py
arthurdarcet: done

aiohttp/http_writer.py
arthurdarcet: done

fafhrd91:
we have already doc related issue for payload system.

lets add log to BytesPayload.

tests/test_web_server.py
arthurdarcet: until now, the headers were buffered and not sent to the transport until the first byte of the body was written.

I changed that to send the headers immediately if the transport is ready when `write_headers` is called; it impacts this test because if the server cancels the `write_eof`, the headers are still already sent, so the client does not see the ServerDisconnectedError and instead I have to try to read the body to get a `ClientPayloadError` (the server did not send a complete body)

see the last paragraph of the issue

fafhrd91: oh, I see. reason is performance for small responses. for small response it usually one syscall, with your change it is two. so I'd like to measure performance on small responses.

fafhrd91: other note. buffering is needed because of http pipelining. but if we decide on #2109 then it may go away.

fafhrd91:
@asvetlov any comments?

aiohttp/web_fileresponse.py
```diff
@@ -20,17 +20,16 @@
...
 
-    def set_transport(self, transport):
-        self._transport = transport
-
-        if self._drain_waiter is not None:
-            waiter, self._drain_maiter = self._drain_maiter, None
-            if not waiter.done():
-                waiter.set_result(None)
+    def __init__(self, *args, **kwargs):
+        self.__buffer = []
```
asvetlov: It's mostly taste question but why don't invite new name for this buffer instead of mangling?
We never use mangling in aiohttp.

aiohttp/http_writer.py
```diff
@@ -130,6 +130,7 @@ class PayloadWriter(AbstractPayloadWriter):
     def __init__(self, stream, loop, acquire=True):
         self._stream = stream
         self._transport = None
+        self._buffer = []
```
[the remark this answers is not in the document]
arthurdarcet: if i remember correctly, aiohttp switched from bytearrays to list for performance reasons too, so this probably need to be benchmarked…

aiohttp/http_writer.py
```diff
@@ -138,29 +139,39 @@ def __init__(self, stream, loop, acquire=True):
...
 
     def set_transport(self, transport):
         self._transport = transport
 
-        chunk = b''.join(self._buffer)
-        if chunk:
-            transport.write(chunk)
-            self._buffer.clear()
+        if self._buffer is not None:
+            for b in self._buffer:
```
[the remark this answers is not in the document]
arthurdarcet: if i'm understanding this correctly, the buffer is usually just one item, the headers, so sending them in a for loop or joining won't change much.

The tradeoff however is when the buffer is exactly a small chunk (the headers) and one very large chunk (the body). This will copy the body to create a slightly larger chunk and will be much slower than calling write twice.

I have no idea of how the internals of asyncio works, so I'll of course defer to you here ; but do we have benchmarks for this?

aiohttp/http_writer.py
```diff
@@ -138,29 +139,39 @@ def __init__(self, stream, loop, acquire=True):
...
                 waiter.set_result(None)
 
+    @asyncio.coroutine
+    def get_transport(self):
+        if self._transport is None:
+            if self._drain_waiter is None:
+                self._drain_waiter = create_future(self.loop)
+            yield from self._drain_waiter
+
+        assert self._transport is not None
```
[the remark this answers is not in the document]
arthurdarcet: nothing that i can think of, i was using the asserts as safeguards/way to comment the code. I'll remove them

aiohttp/web_fileresponse.py
arthurdarcet: yes, you're perfectly right, i'll update the PR with a new name instead of mangling

aiohttp/http_writer.py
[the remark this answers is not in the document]
asvetlov: We just follow asyncio streams design, it switched from bytearrays to lists and back to bytearrays :)

aiohttp/http_writer.py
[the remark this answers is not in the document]
asvetlov: Well, nevermind.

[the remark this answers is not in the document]
arthurdarcet: i'll try to write some benchmark of how multiple syscalls behave. I'm pretty sure the transport is almost always set before the body is written, so it's not really buffered here.

The main optimisation in all this is sending the buffered data as soon as the transport is set. Right now, the buffered data waits for the next chunk, and then is joined with this new chunk and everything is sent together. That behaves very poorly with a large body because it means copying everything to avoid a second syscall. For very small buffers it probably makes way more sense to bundle everything and do only one syscall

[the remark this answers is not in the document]
fafhrd91: Should we check size of chunk and do join for small chunk?

tests/test_web_server.py
arthurdarcet: See #2179 for some benchmarks. I'm not really sure how to benchmark the syscalls, but impact seems negligible

aiohttp/http_writer.py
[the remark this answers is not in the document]
arthurdarcet: I tried to benchmark bytearray vs join in #2179, but the results were inconclusive at best (or lean in favor of `b''.join`). I'll try to dig up why asyncio changed again to bytearray

aiohttp/http_writer.py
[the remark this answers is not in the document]
arthurdarcet: checking the size of the chunk would mean buffering the headers until the first chunk is written - as is done currently. I'm not a big fan of that because it makes the code really more complicated (you might end up having data buffered *and* a transport already set). Plus I don't think the syscall matters that much

arthurdarcet:
@asvetlov @fafhrd91 Could you have a look at this? Are there still issue blocking this PR?

fafhrd91:
I can not comment on this, I didn't work on aiohttp for long time.  this is question for other contributors.

@asvetlov @kxepal 

btw can we ping contributors somehow?

asvetlov:
I'll try to find a time tomorrow maybe

@aio-libs/aiohttp-committers should work for pinging all aiohttp committers in the list

aiohttp/http_writer.py
```diff
@@ -273,25 +266,19 @@ def write_headers(self, status_line, headers, SEP=': ', END='\r\n'):
...
-    def drain(self, last=False):
+    async def drain(self):
         if self._transport is not None:
-            if self._buffer:
-                self._transport.write(b''.join(self._buffer))
-                if not last:
-                    self._buffer.clear()
-            yield from self._stream.drain()
+            print(self._stream)
+            print(self._stream.drain)
```
kxepal: print

aiohttp/http_writer.py
```diff
@@ -138,29 +139,36 @@ def __init__(self, stream, loop, acquire=True):
         self.output_size = 0
 
         self._eof = False
-        self._buffer = []
         self._compress = None
         self._drain_waiter = None
 
         if self._stream.available:
             self._transport = self._stream.transport
+            self._buffer = None
```
kxepal, line 142: I think it's not good idea to jungle variable type. How empty buffer `[]` is different from `None`?

arthurdarcet: I'm using `[]` and `None` a bit differently here: once the buffer has been emptied into the new transport, it should never be used again. So I would prefer setting it to None and crash if something very wrong happens

aiohttp/http_writer.py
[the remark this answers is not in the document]
arthurdarcet: Benchmarks were added in a separate pull request: https://github.com/aio-libs/aiohttp/pull/2179

aiohttp/http_writer.py
arthurdarcet: oops, thanks, fixed

aiohttp/http_writer.py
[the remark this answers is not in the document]
kxepal: Awesome! Thanks!

tests/test_client_functional_oldstyle.py
```diff
@@ -710,7 +710,7 @@ def connection_lost(self, exc):
             await r.read()
             self.assertEqual(1, len(connector._conns))
 
-            with self.assertRaises(aiohttp.ServerDisconnectedError):
+            with self.assertRaises(aiohttp.ClientConnectionError):
```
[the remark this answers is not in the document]
arthurdarcet: the headers are now sent even if write_eof raises. This is mostly a side effect of the tests: in real life I would expect users already receiving a bit of both ClientPayloadError and ServerDisconnectedError for this use case (depending on when exactly the connection was cut off).

aiohttp/http_writer.py
kxepal: Hm...I see. Well, then such crash shold be controlled, imho. Otherwise it would be hard to distinguish transport re-use case from else error.

tests/test_client_functional_oldstyle.py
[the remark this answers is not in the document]
kxepal: Ok, thanks for explanation!

aiohttp/http_writer.py
arthurdarcet: It's purely internal to PayloadWriter, so everything is controlled where it should be.

This None means that at any given time, exactly one of `(self._transport, self._buffer)` is None. It makes it easier to check the code / prove that it's correct.

arthurdarcet:
I've rebase this against master, again. Could we reach a decision here?

@asvetlov @kxepal

asvetlov:
Looks good.
Could you add `CHANGES/2126.feature` changelog note.

arthurdarcet:
@asvetlov sure, thanks, just pushed the commit. let me know if the message isn't informative enough

asvetlov:
Let's wait for a while for feedback from other devs.

asvetlov:
Looks like nobody objects, let's go ahead.

@arthurdarcet I've dropped `tests/test_client_functional_oldstyle.py` and moved all remaining tests into `tests/test_client_functional.py`.
Could you solve resolved conflicts?
Also 100% test coverage would be awesome.

arthurdarcet:
@asvetlov 

 - rebase is done

 - I pushed two more tests, that cover two branches in `PayloadWriter`. The only things left untested in this class are the tcp cork related functions (i have no idea what that is, so i'm not sure i should try to write the tests...) and an edge case where the PayloadWriter receives two large concurrent writes before the transport is set (i could test this with a manual calls to `PayloadWriter.get_transport`, but that's not really "public-api").

 - I removed a "if" that wasn't covered with https://github.com/aio-libs/aiohttp/pull/2126/commits/78eb7b44f252cf84b4e1a81567df6cd91402225f . I'm pretty confident it was impossible for the if condition to become false, but please review this commit


(Also, sorry I push over you commit fixing my 2126.feature file and I haven't found a way to cherry pick your commit back, so I rewrote your changes in my original commit)

asvetlov:
@arthurdarcet thanks.
For your information: rebase is not required, just merge is enough. The PR is squashed on merging anyway.

PR #2127, opened 2017-07-24
log a warning when sending requests with a large body of bytes

arthurdarcet:
As discussed in #2126 , log a warning when sending a request with a large body payload passed as bytes directly: aiohttp does not do any chunking for this use case, and it can freeze the event loop.

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -3,6 +3,7 @@ import io
 import json
 import mimetypes
 import os
+import warnings
 from abc import ABC, abstractmethod
 
 from multidict import CIMultiDict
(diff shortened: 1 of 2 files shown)


aiohttp/payload.py
```diff
@@ -150,6 +151,11 @@ def __init__(self, value, *args, **kwargs):
 
         self._size = len(value)
 
+        if self._size > DEFAULT_LIMIT:
```
fafhrd91: I think `DEFAULT_LIMIT` is too small. lets define new const, something close to 1mb

arthurdarcet: ok, do you want it in `payload.py` or in `stream.py` next to the other constants?

fafhrd91: it should be payload specific

fafhrd91:
thanks

PR #2128, opened 2017-07-24
Multipart base64 encoding as  RFC 2045

### LINKED ISSUE #2087: Multipart base64 encoding produces long line
## Long story short

When using the aiohttp.multipart module, and its MultipartWriter class, we can create multipart documents whose payload can be transfert-encoded in base64.

We use this module to retrieve some http content via aiohttp client, and _stream_ a multipart document containing the Responses provided by the client to another processus via stdout (so we're not buffering the contents into memory, but act like a proxy).

The MultipartWriter generates base64 subdocuments without newlines. In version 0.2.1 of aiohttp, it was with multiple lines.

I know the base64 format does not define line length, but :
- there are very few reasons to use the multipart module with base64 encoding in http transaction (since http can handle binary contents)
- there are very good reasons to use multipart + base 64 in emails since emails are 7 bits only
- emails format is described in RFC5322
- RFC5322 states that line length must be more than 998 characters (and should not be more than 78 characters) (https://tools.ietf.org/html/rfc5322.html#section-2.1.1)
- RFC2045, which define MIME (englobing the multipart format), states that "The encoded output stream must be represented in lines of no more than 76 characters each. " (https://tools.ietf.org/html/rfc2045.html#section-6.8)

I think the MultipartWriter should allow us to generate base64 sub contents with newlines (potentially with an option).

## Expected behaviour

When producing a multipart document with base64 tranfert encoding, the final document should not have lines longer than 78 characters (more or less).

## Actual behaviour

When producing a multipart document with base64 tranfert encoding, the final document has the entire base64 subdocument on one line.

## Steps to reproduce

```
import sys

import asyncio
import aiohttp.multipart as mp

class Writer:
	@asyncio.coroutine
	def write(self, chunk):
		sys.stdout.buffer.write(chunk)


a = mp.MultipartWriter()
a.append("i" * 200, {"Content-transfer-encoding": "base64"})

loop.run_until_complete(a.write(Writer()))
```

## Fix

In aiohttp.multipart.MultipartPayloadWriter.write() method, the call to base64.b64encode could be replaced by base64.encodebytes which takes care of the newlines.

alex-eri:
## What do these changes do?
Fixes #2087 
Trivial, but need discussion.

https://docs.python.org/3/library/base64.html#base64.encodebytes

No affected for user, but can broke some side apps:

Maybe unexpected '\n' because http exprcts '\r\n' .

Lose performance 
```
>>> timeit.timeit('base64.b64encode(data)','import base64;data= b"i"*1000')
3.0879702609963715
>>> timeit.timeit('base64.encodebytes(data)','import base64;data= b"i"*1000')
15.915579186999821

```

## Related issue number

#2087 

## Checklist

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -915,7 +915,7 @@ class MultipartPayloadWriter:
                 enc_chunk, self._encoding_buffer = (
                     buffer[:div * 3], buffer[div * 3:])
                 if enc_chunk:
-                    enc_chunk = base64.b64encode(enc_chunk)
+                    enc_chunk = base64.encodebytes(enc_chunk)
                     yield from self._writer.write(enc_chunk)
         elif self._encoding == 'quoted-printable':
             yield from self._writer.write(binascii.b2a_qp(chunk))
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -876,7 +876,6 @@ class MultipartPayloadWriter:
     def enable_encoding(self, encoding):
         if encoding == 'base64':
             self._encoding = encoding
-            self._encoding_buffer = bytearray()
         elif encoding == 'quoted-printable':
             self._encoding = 'quoted-printable'
 
@@ -893,11 +892,6 @@ class MultipartPayloadWriter:
                 self._compress = None
                 yield from self.write(chunk)
 
-        if self._encoding == 'base64':
-            if self._encoding_buffer:
-                yield from self._writer.write(base64.b64encode(
-                    self._encoding_buffer))
-
     @asyncio.coroutine
     def write(self, chunk):
         if self._compress is not None:
@@ -907,16 +901,14 @@ class MultipartPayloadWriter:
                     return
 
         if self._encoding == 'base64':
-            self._encoding_buffer.extend(chunk)
+            if chunk:
+                enc_chunk = base64.encodebytes(chunk)
+                # eol convertion needed because python API returns simple \n
+                # and RFC 2045 needs \r\n. It's safe because base64 doesn't
+                # produce \n inside a payload
+                enc_chunk = enc_chunk.replace(b"\n", b"\r\n")
+                yield from self._writer.write(enc_chunk)
 
-            if self._encoding_buffer:
-                buffer = self._encoding_buffer
-                div, mod = divmod(len(buffer), 3)
-                enc_chunk, self._encoding_buffer = (
-                    buffer[:div * 3], buffer[div * 3:])
-                if enc_chunk:
-                    enc_chunk = base64.encodebytes(enc_chunk)
-                    yield from self._writer.write(enc_chunk)
         elif self._encoding == 'quoted-printable':
             yield from self._writer.write(binascii.b2a_qp(chunk))
         else:
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -914,6 +914,19 @@ def test_writer_with_content_transfer_encoding_base64(buf, stream, writer):
 
 
 @asyncio.coroutine
+def test_writer_with_content_te_base64_no_long_lines(buf, stream, writer):
+    writer.append('Time to write long lines!' * 10,
+                  {CONTENT_TRANSFER_ENCODING: 'base64'}
+    )
+    yield from writer.write(stream)
+    headers, message = bytes(buf).split(b'\r\n\r\n', 1)
+
+    line0 = message.split(b'\r\n')
+    assert len(line0) < 80
+
+
+
+@asyncio.coroutine
 def test_writer_content_transfer_encoding_quote_printable(buf, stream, writer):
     writer.append('Привет, мир!',
                   {CONTENT_TRANSFER_ENCODING: 'quoted-printable'})
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -923,6 +923,7 @@ def test_writer_with_content_te_base64_no_long_lines(buf, stream, writer):
     line0 = message.split(b'\r\n')
     assert len(line0) < 80
 
+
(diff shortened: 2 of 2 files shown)


fafhrd91:
@kxepal thoughts?

alex-eri:
@manuco one line for fix. You can make regression tests

alex-eri:
I suggest to NOT merge it besause of 

```
>>> timeit.timeit('base64.b64encode(data)','import base64;data= b"i"*1000')
3.0879702609963715
>>> timeit.timeit('base64.encodebytes(data)','import base64;data= b"i"*1000')
15.915579186999821

```

fafhrd91:
does speed matter for this code?

alex-eri:
This kind of Multipart is often big amount of data (mjpeg video stream for example).
It difficult to cache on proxy. 
So speed preferred then readability.

updated topic

manuco:
Base64 is not designed to be used on pure http streams : http can handle binary streams. If speed matters, on big data like mpeg streams, base64 should't be considered as a good transport encoding... 

IMHO, the speed problem is not a major one.

_edit: typo_

manuco:
@alex-eri: what's the best workflow to include the unit test, and some others fixes for this problem on the code ?
- Should I fork + PR your repo, or should I fork + PR the main repo ? Or should I just push on your branch once you allow me to do that ?

arthurdarcet:
I agree with @manuco that speed is not a concern here: the base64 encoding is not the default one when sending a multipart request, and you have to explicitly enable it. Sending huge files encoded in base64 is a waste of resources anyway.

The 6x overhead for `encodebytes` over `b64encode` is insane though, opening a bug report against Python for this might be a good idea

alex-eri:
@manuco i dont git pro =)  Dont know the best way. You can create new PR. I close this

https://github.com/alex-eri/aiohttp/invitations

manuco:
I've pushed my proposal on your repo on the branch patch-5. If everyone accept it, you can create a Pull Request for merging into the main repo, and close #2128 and #2087.

alex-eri:
u are shure for? 
+                enc_chunk = enc_chunk.replace(b"\n", b"\r\n")
imaybe is there better way to do this?

what would happen if concatenate two or more chunks (one with tail < 76 chars)? 
one string with longer or less length.

manually splitting  it maybe

manuco:
Ha! I see the problem:
- with multiple write(), there will be output lines that will not be at the maximum size, or even contains a padding in the middle of the string. I missed that, and now I understand the work on the "multiple of 3" split stuff.

I'll have to write some test case testing this case, and the code of it. For today, I don't have any more time for this. I'll do this (I hope) before the end of the week.

Is it allowed to rewrite the branch history ?

manuco:
For the replace() stuff, it only touches the encoded part of the stream. So wathever the input content is, it can't be converted with "\n" in the base64 format. So we can safely transform it as long as we don't used any of the base64 alphabet.

But since this first implementation has the issue seen earlier, I'll try a smarter approach.

asvetlov:
> Is it allowed to rewrite the branch history ?

Do what you want, it's your branch :)
Even `git push --force` is allowed.

asvetlov:
Why test is failed?

manuco:
The test shows that when one stream reader reads multiple chunks, the produced base64 document is not valid. It means that my previous code doesn't work well. It was my first step about producing a better solution for base64 :

1. create a test that shows the problem
1. fix the problem

The next step will be done tomorrow or the next week (i'm busy with many other things to do).

aiohttp/multipart.py
```diff
@@ -907,16 +901,14 @@ def write(self, chunk):
...
-            if self._encoding_buffer:
-                buffer = self._encoding_buffer
-                div, mod = divmod(len(buffer), 3)
-                enc_chunk, self._encoding_buffer = (
-                    buffer[:div * 3], buffer[div * 3:])
-                if enc_chunk:
-                    enc_chunk = base64.b64encode(enc_chunk)
-                    yield from self._writer.write(enc_chunk)
+            if chunk:
+                enc_chunk = base64.encodebytes(chunk)
```
socketpair, line 905: Are you serious?

Base64-encoding adds equal-signs at the end for alignment. So two single-byte chunks like `A` will write `QQ==QQ==` instead of `QUE=`.

socketpair (changes requested):

aiohttp/multipart.py
manuco: > Are you serious?

What's your point?

This is a work in progress. As you can see in my [previous comment](https://github.com/aio-libs/aiohttp/pull/2128#issuecomment-321599180) I'm already aware of this problem.

The previous code that you show indeed fixes the problem of the multiple read chunks, but doesn't adress the long line problem that I reported initially. I will try to solve these two constraints soon (I hope).

socketpair: Usually incomplete code (I'm not talking about tests/linting/docs) should not reach PR, since it will just uselessly bother reviewers.

Your changes are welcome, but please show us complete working code even if you are unusure about the idea. When the idea is approved, you can add tests/docs and so on.

kxepal: @socketpair 
I wouldn't be so strict here. When you open PR, your remote branch becomes liked with PR. Any changes there automagically happens here. Do you suggest to not do any push after that until you done requested changes? That's not cool.

Much better when contributor pings you once requested changes becomes satisfied. He can continue work on in his branch while you don't pay attention to WIP PRs.

asvetlov:
What is the status of PR?

manuco:
Still on my todolist (but not on the top) but I miss time to complete it for now. I'll try to prioritize it for the next weeks.

asvetlov:
Good

Issue #2114, opened 2017-07-19
aiohttp.client does not support HTTP redirect 308 (RFC 7538)


himikof:
Currently, `aiohttp.client` does not support recent HTTP redirect code 308 Permanent Redirect, defined in [RFC 7538](https://tools.ietf.org/html/rfc7538). It is simply a permanent version of already handled 307 redirect, or a legacy-free version of 301 redirect.

I would also argue that `normalize_path_middleware` default redirect class should be changed to `HTTPPermanentRedirect`, as it is currently very surprising that a user agent will likely retry a POST request as a GET one after receiving a 301 redirect from path normalization, but that is probably a breaking API change.
Still, supporting 308 in `aiohttp.client` would allow `normalize_path_middleware` to be used with POST requests in asyncio <-> asyncio communication by specifying redirect 308 manually.

fafhrd91:
Would you provide PR for client?

asvetlov:
Fixed by #2134

PR #2134, opened 2017-07-27
Support HTTP 308 Permanent redirect in client class.

penguinolog:
## What do these changes do?

Implements HTTP 308 code support in client

## Related issue number
Implements: #2114


asvetlov:
Thanks

Issue #1987, opened 2017-06-19
Add support for retrieving randomly selected port numbers


abingham:
If you pass `0` as the port to `aiohttp.web.run_app()`, aiohttp will select any available port. However, it is not currently possible to determine *which* port was selected. It would be useful in some cases to be able to determine which port is being used.

A motivating example for this is [the traad Python refactoring server](https://github.com/abingham/traad/tree/aiohttp). This listens for HTTP requests to perform various kinds of refactoring operations. It is typically used via a text editor like Emacs. Users of traad don't normally care what port it's running on; they may not even know that they're running an HTTP server at all. They just expect Emacs to "do the right thing". As such, it doesn't make sense to ask them to specify a port for the server. 

The emacs client for traad starts the server, telling it to choose any available port. Once the server starts, it prints its actual port to stdout and the emacs client parses that. This is possible in the current bottle-based implementation of traad, but the experimental aiohttp version is a impractical because we can't determine the port that it selects.

For my purposes, it would be great if I could register an `on_startup` listener that could then fetch the port from the app. But right now there's no way to get the port.

FWIW, there was [some discussion of this on stackoverflow](https://stackoverflow.com/questions/44610441/how-to-determine-which-port-aiohttp-selects-when-given-port-0).

alexmnazarenko:
You can always do something like this:

```py
  sock = socket.socket()
  sock.bind(("127.0.0.1", 0))
  # Aiohttp will call 'listen' inside.
  # But it must be called before we actually use the port,
  # any attempts to connect before the 'listen' call will
  # be rejected.
  sock.listen(128)
  _, port = sock.getsockname()
  print("Port: %d" % (port,))

  aiohttp.web.run_app(app, sock=sock)
```

I use similar code in unittests. Note that while 'listen' is optional and kinda ugly, there is no way to reliably 'wait' for listen call by aiohttp (as on_startup is called before setting up asyncio servers). As about this weird looking '128' param - took it from aiohttp defaults.

Hope that helps )

asvetlov:
The problem is `Application` might be used for listening on several tcp and unix sockets.
Right not we don't register them into application but PR is welcome.

abingham:
FWIW, [this SO post](https://stackoverflow.com/questions/44610441/how-to-determine-which-port-aiohttp-selects-when-given-port-0) has a solution that works well for my situation.

abingham:
> Right not we don't register them into application but PR is welcome.

[...]

PR #2135, opened 2017-07-27
Added "pre_serve" signal and registered asyncio servers with app.

abingham:
## What do these changes do?

This introduces two related changes. The first is that it adds a "pre-serve" hook which is run right before `run_forever()` is called. Second, it attaches all of the "servers" to the app under the key `asyncio_servers`. The proximal need for this is so that users can determine which ports are being used by aiohttp when they specify port=0.

This is an entirely backwards-compatible change. Existing users won't need to change anything, though of course they can start using the new hook and server information if they want.

## Related issue number

#1987 

## Checklist

I have not done most of the following checklist. I'd like to get some indication of whether these changes will be accepted at all before I go through the effort of updating tests, documentation, etc. (which I will happily do if I think this PR is going anywhere). I'm particularly concerned that the names I'm using (e.g. pre-serve, asyncio_servers) are not appropriate, so I'd like some feedback on those. 


asvetlov:
1. Why do you need `pre_serve` signal?
2. Info about started servers should be stored outside of app in separate class maybe. But definitely not in app's dict: it is for users and third party libraries.

abingham:
> Why do you need pre_serve signal?

Maybe I don't, but as far as I can tell no other signal is run between a) when the "servers" are created and b) when `run_forever` is started.

> Info about started servers should be stored outside of app in separate class maybe. But definitely not in app's dict: it is for users and third party libraries.

That's fine with me. How do I communicate the server information to users? I'm completely open to any suggestions you have, which is why I started the discussion in #1987 in the first place. This PR is actually an attempt on my part to get any sort of feedback from the powers-that-be in the project to get a real problem solved.

My concern is not about what signals exist, where server details are stored, or any implementation detail like that. I simply need to be able to find out which port(s) my server is listening on.

abingham:
I might also point out your comment from #1987:

> The problem is Application might be used for listening on several tcp and unix sockets. Right not we don't register them into application but PR is welcome.

I took this as an indication that we *should* "register them into application", hence the nature of this PR.

asvetlov:
Got it. I've missed the context.

aiohttp/web.py
```diff
@@ -436,6 +451,7 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
         servers = loop.run_until_complete(
             asyncio.gather(*server_creations, loop=loop)
         )
+        app['asyncio_servers'] = servers
```
cecton: You could also simply implement a public attribute like "server" in the instance and make a setter that will update it for you. Just like `set_loop`.

abingham:
Yep, that would work just fine. Though do you mean that we would need a
setter function to modify the attribute? I'm not sure that's warranted. And
if it is necessary, a property would probably be more "pythonic".

aiohttp/web.py
pfreixes: IMHO if the idea is just give a way to gather that information, I will go for a simple `getter`. Indeed I would avoid the `setter` taking into account that modifies the content will do nothing.

aiohttp/web.py
```diff
@@ -259,6 +266,14 @@ def startup(self):
         yield from self.on_startup.send(self)
 
     @asyncio.coroutine
+    def pre_serve(self):
+        """Causes on_pre_serve signal
+
+        Should be called just before run_forever().
+        """
+        yield from self.on_pre_serve.send(self)
```
pfreixes: Im missing all of the proper test coverage of this new signal

abingham: Your point is good. But as I mentioned in the description of this PR, I have not done all of the normal "due diligence" yet. And I won't do it until I can get some assurance that the basic design is acceptable to the maintainers.

pfreixes: As a member and reviewer, I can see the point of publishing the servers as a new attribute of the app just as a simple property (only getter). Please cover it with the proper unit tests.

abingham:
> Looks like that we need access to the app.servers object attribute to set the servers. 

Why? The property as written now exposes a `list` which is mutable. Short of replacing the object with another object, adding a setter to the property doesn't let you do anything semantically new, and I'd argue that forcing `servers` to be a `list` (or *some* fixed type) is probably a good thing.

If I'm missing some point, though, it would be good for me to know.

> this is not gonna work,

What do you mean by "this"? It's a bit vague.

> with tests will be easier check if the code works or not.

Believe me, I'll get to those when the time is right. Nagging won't speed that up. As I've said from literally the beginning of this PR, I don't want to spend time on tests until I'm sure that the design passes muster with the project. That's the exercise we're involved in right now.

abingham:
FWIW, I've added a few tests around `on_pre_serve` and `Application.servers`. These aren't very thorough tests, but they seem to be at about the same level as the others in the suite.

abingham:
Due to maintainer inactivity (disinterest?), this PR has bit rotted to the point where I would essentially need to re-write it. Would it be worth my time to do so? Or would that work simply be ignored as well?

asvetlov:
I'm very sorry.
Implementation of `run_app` was rewritten from scratch.
Let's continue discussion in #1987 issue.

PR #2136, opened 2017-07-27
Raise ContentTypeError by json() when headers does not …

pfreixes:
## What do these changes do?

Basically, avoid having a nocive fall back as the following one

```python
try:
    resp = await req.json()
except ClieintResponseError:
    logging.warning("Invalid content-type")
    resp = loads(await req.read())
```

The `ClieintResponseError` can be raised by many reasons, and the case presented the fallback can be misleading having as a  result twice `ClieintResponseError` because the source of the problem wasn't related to the content type.

With an ad-hoc exception, the client can code a more specific fallback case taking into account the differents source of problems with a cleaner interface, for example:

```python
try:
    resp = await req.json()
except ContentTypeError:
    logging.warning("Invalid content-type by servcie ....")
    resp = loads(await req.read())
except ClieintResponseError:
    logging.exception("something went wrong")
    raise
```

Thanks to the current `json()` implementation once we have had a `ContentTypeError` exception is not possible to have a `ClieintResponseError`. Therefore, the developer can handle the errors as the previous example.

For backward compatibility the `ContentTypeError` inherits from `ClieintResponseError`.

No

## Checklist

### DIFF
--- a/aiohttp/client_exceptions.py
+++ b/aiohttp/client_exceptions.py
@@ -40,7 +40,7 @@ class ClientResponseError(ClientError):
 
 
 class ContentTypeError(ClientResponseError):
-    """ ContentType found is not valid """
+    """ContentType found is not valid."""
 
 
 class WSServerHandshakeError(ClientResponseError):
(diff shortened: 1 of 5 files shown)


asvetlov:
The intention is pretty clean but `ClientResponseContentTypeError` is 30 chars long!
Please invite shorter name.

pfreixes:
Exception name changed to `ContentTypeError`

asvetlov:
Thanks!

PR #2137, opened 2017-07-27
Make dist folder be created under right privileges

webknjaz:
@asvetlov it's the same as for multidict.

Ref: aio-libs/multidict#117


webknjaz:
I've decided to make script run atomic by moving dist dir creation there.

PR #2142, opened 2017-07-28
Allow overriding ClientSession class in TestClient

cecton:
## What do these changes do?

It adds a keyword argument to the TestClient class to override the default ClientSession used.

None.

## Related issue number

#2032 #2074 

## Checklist

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -374,7 +374,7 @@ class AioHTTPTestCase(unittest.TestCase):
         Use .get_application() coroutine instead
 
         """
-        pass  # pragma: no cover
+        raise RuntimeError("Did you forget to define get_application()?")
 
     def setUp(self):
         self.loop = setup_test_loop()
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -186,7 +186,8 @@ class TestClient:
 
     """
 
-    def __init__(self, server, *, cookie_jar=None, loop=None, **kwargs):
+    def __init__(self, server, *, cookie_jar=None, loop=None,
+                 session_class=ClientSession, **kwargs):
         if not isinstance(server, BaseTestServer):
             raise TypeError("server must be web.Application TestServer "
                             "instance, found type: %r" % type(server))
@@ -194,7 +195,7 @@ class TestClient:
         self._loop = loop
         if cookie_jar is None:
             cookie_jar = aiohttp.CookieJar(unsafe=True, loop=loop)
-        self._session = ClientSession(loop=loop,
+        self._session = session_class(loop=loop,
                                       cookie_jar=cookie_jar,
                                       **kwargs)
         self._closed = False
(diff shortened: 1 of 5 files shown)


asvetlov:
Sorry, I still don't understand why do you need to use custom client session.
I believe inheritance from `ClientSession` is discouraged (while not prohibited)

cecton:
Really? I thought it was common usage. That's how docker-py works https://github.com/docker/docker-py/blob/master/docker/api/client.py#L44 I did the same for aiosparql: https://github.com/aio-libs/aiosparql/blob/master/aiosparql/client.py#L81

(In case of aiosparql I just did it because then I can use the `async with` automatically, otherwise I have to implement all the methods for it.)

asvetlov:
@cecton looks like I've lost control onto aiohttp classes usage (again and again).
My initial thought about `ClientSession` is the session should be *sealed*, if anybody need to extend it -- please make a wrapper with overriding `session.get`, `session.post` and others.
Embedding is better than inheritance etc.

But I see people inherit from session widely.
Well, I could live with it.

The PR is good but please add a check for missed `RuntimeError` coverage.
After that feel free to merge.

cecton:
@asvetlov I can hardly test at that level because it's testing the test. But I will think of it... if you have any idea... :)

cecton:
The example 2 here could help http://www.programcreek.com/python/example/7163/unittest.expectedFailure

Will do that

asvetlov:
Well, `#pragma: no cover` will work too.

cecton:
@asvetlov there is something fishy with the loop fixture that I will check another time. If I don't use the fixture loop, I can see this in the logs (by using `-s` of py.test).

```
============================================== 1 passed in 0.45 seconds ===============================================
Exception ignored in: <bound method BaseEventLoop.__del__ of <_UnixSelectorEventLoop running=False closed=True debug=Fa
lse>>
  File "/usr/lib/python3.5/asyncio/base_events.py", line 431, in __del__
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 58, in close
  File "/usr/lib/python3.5/asyncio/unix_events.py", line 139, in remove_signal_handler
  File "/usr/lib/python3.5/signal.py", line 47, in signal
TypeError: signal handler must be signal.SIG_IGN, signal.SIG_DFL, or a callable object
py35 runtests: commands[1] | mv .coverage .coverage.py35
_______________________________________________________ summary _______________________________________________________
  py35: commands succeeded
  congratulations :)
```

Not blocking though... just unclean.

cecton:
@asvetlov I have the feeling the tests are run 3 times on travis... it takes ages compared to appveyor... something wrong too

cecton:
@asvetlov If you think one should not inherit from ClientSession, then maybe it's best to not encourage this behavior at all. Maybe it is best that we keep only the second commit which is simply handy and drop the first one. Anyone who really needs that will find their way to hack around to get it done anyway.

asvetlov:
Yes, the second commit only looks much better

cecton:
Closed in favor of #2151

PR #2144, opened 2017-07-28
Fix FileResponse sending empty chunked body on 304

### LINKED ISSUE #2143: FileResponse sends empty chunked body when returning 304
## Long story short

`FileResponse` (and by extension, `add_static`) may return `Transfer-Encoding: chunked` and the empty chunked body when returning a 304 Not Modified, in violation of [RFC 2616 §10.3.5](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5), which states: "The 304 response MUST NOT contain a message-body, and thus is always terminated by the first empty line after the header fields."

## Expected behaviour

A 304 response with no body.

## Actual behaviour

A 304 response with `Transfer-Encoding: chunked` and a 1-byte body (`0`).

## Steps to reproduce
1. Run this with an empty `index.html` in the same directory:
```
from aiohttp import web
from pathlib import Path

app = web.Application()
app.router.add_static('/', Path(__file__).parent)
web.run_app(app, port=8080)
```

2. Send an HTTP/1.1 request with an `If-Modified-Since` that triggers a 304 Not Modified, e.g.
```
# nc localhost 8080
GET /index.html HTTP/1.1
If-Modified-Since: Fri, 28 Jul 2017 18:42:00 GMT

```

The response will be something like:

```
HTTP/1.1 304 Not Modified
Transfer-Encoding: chunked
Date: Fri, 28 Jul 2017 18:42:34 GMT
Server: Python/3.6 aiohttp/2.2.3

0

```

## Your environment

This is using aiohttp 2.2.3, the newest version at this time.

es1024:
## What do these changes do?

Fixes a bug where FileResponse/add_static will include a non-empty body and use Transfer-Encoding: chunked for 304 responses.

FileResponse/add_static now won't give malformed 304 responses with bodies.

## Related issue number

This fixes #2143.

## Checklist

- [n/a] Documentation reflects the changes
- [X] Add a new news fragment into the `changes` folder


tests/test_web_sendfile_functional.py
```diff
@@ -165,7 +165,9 @@ def handler(request):
     resp.close()
 
     resp = yield from client.get('/', headers={'If-Modified-Since': lastmod})
+    body = yield from resp.read()
     assert 304 == resp.status
+    assert b'' == body
```
cecton: Hello @es1024 ,

According to what I read on the RFC:
 *  http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3
 *  http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.5

The Content-Length header should not even be there. Can you add an assert to enforce it?

aiohttp/web_response.py
```diff
@@ -362,7 +362,8 @@ def _start(self, request,
                 del headers[CONTENT_LENGTH]
         elif self._length_check:
             writer.length = self.content_length
-            if writer.length is None and version >= HttpVersion11:
+            if (writer.length is None and version >= HttpVersion11 and
```
asvetlov: I think fix should be applied to `web_fileresponse.py` instead of base `StreamResponse` class.

asvetlov:
pretty sure `BIN  aiohttp/.web_response.py.swo` was not intentionally added.

aiohttp/web_fileresponse.py
```diff
@@ -179,6 +179,7 @@ def prepare(self, request):
         modsince = request.if_modified_since
         if modsince is not None and st.st_mtime <= modsince.timestamp():
             self.set_status(HTTPNotModified.status_code)
+            self._length_check = False
```
asvetlov, line 182: This line looks hacky but I could live with it.
Looks like `StreamResponse` is too coupled with `Response` class, let's left it for other PR

tests/test_web_sendfile_functional.py
cecton: Thx!

This habit of so many dev to always use `.get()` 😕 This is so much more readable (and slightly more correct):

```python
assert "Content-Length" not in resp.headers
```

(Don't bother though, not important)

cecton:
@asvetlov can I do the merge? (I will drop the bin don't worry)

asvetlov:
@cecton please do

cecton:
Fixed by cc614ca3ecbcb3d49854a2a7411cf64eb724923d

PR #2147, opened 2017-07-30
Patch aiohttp classes module

asvetlov:
The PR fixes `__module__` attribute for aiohttp classes.

You know people don't import our classes properly: the use `from aiohttp.client.ClientSession` instead of `aiohttp.Session` etc. I saw it very many times.
It prevents us for deep refactoring (extracting source code into new file is a potential source of backward compatibility problems for our users).

The proposal pins module names to expected by us: `aiohttp` for client usage and `aiohttp.web` for server. 

I raised the issue many times proposing mangling modules by underscore etc.
Patching the module looks like robust and it's an easy change.
I'll check tomorrow how it works for PyCharm IDE.

### DIFF
--- a/aiohttp/__init__.py
+++ b/aiohttp/__init__.py
@@ -4,6 +4,7 @@ __version__ = '2.3.0a0'
 
 from . import hdrs  # noqa
 from .client import *  # noqa
+from .client_exceptions import * # noqa
 from .formdata import *  # noqa
 from .helpers import *  # noqa
 from .http import (HttpVersion, HttpVersion10, HttpVersion11,  # noqa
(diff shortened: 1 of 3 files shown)


kxepal:
What should I see with this PR? PyCharm still happy to suggest me do `from aiohttp.client import ClientSession` as like as `from aiohttp import ClientSession`.

If I just do `import aiohttp`, than in `aiohttp.` suggestion wouldn't be `client` module, but `client_exceptions` and `client_reqrep`. The `ClientSession` is available as well and has  `__module__ == "aiohttp"` .

asvetlov:
No.
```
In [1]: import aiohttp
In [2]: aiohttp.ClientSession
Out[2]: aiohttp.client.ClientSession
```

After the patch applied:
```
In [1]: import aiohttp

In [2]: aiohttp.ClientSession
Out[2]: aiohttp.ClientSession
```

kxepal:
That's the only change, yes.

asvetlov:
Well, I cannot fix PyCharm (but I'll discuss the issue with PyCharm team, sure).
The change is useless and might make more confusion than solve problems.
I still hate imports like `from aiohttp.client import ClientSession`.
Maybe it is my personal pain.

@kxepal do you suggest to close the PR without applying?

kxepal:
@asvetlov I think I do. I don't see how it solve our problem, sorry. 

May be we could try play with `sys.meta_path` in order to force Python import system forget about some modules? Wonder if this even possible, but still.

Or may be we could propose some PEP for, say, `__export__` feature which defines what stuff could be imported from current module/package by caller?

asvetlov:
Respecting `__all__` by hinters should be enough.
The main problem I see is `aiohttp.Client` is not found in `aiohttp.__init__.py` as well as `aiohttp.web.Request` is actually stored in `aiohttp.web_request.py`.

But if user goes to source code reading he could understand instantly where actual code is placed.

PR #2149, opened 2017-07-31
use osx postgres user

thijstriemstra:
This PR makes the polls demo sql installation script pick the default user for pg on osx. Otherwise it'll fail with:

```
$ bash sql/install.sh 
sudo: unknown user: postgres
sudo: unknown user: postgres
sudo: unknown user: postgres
sudo: unknown user: postgres
sudo: unknown user: postgres
```


asvetlov:
thanks

Issue #1949, opened 2017-06-02
Doc clarification -  StreamReader.iter_any() vs iter_chunks()


pkch:
It might help to clarify in the documentation the difference between [`iter_any`](http://aiohttp.readthedocs.io/en/stable/streams.html#aiohttp.StreamReader.iter_any) and [`iter_chunks`](http://aiohttp.readthedocs.io/en/stable/streams.html#aiohttp.StreamReader.iter_chunks). Currently it says

> Iterates over data chunks in order of intaking them into the stream

> Iterates over data chunks as received from the server

But it's not clear what the difference is between the chunks received from the server and the chunks that were taken into the stream.

My (naive) reading suggests the (almost certainly incorrect) interpretation that `iter_chunks` guarantees to preserve the chunks produced by the server under chunked transfer encoding, and yield them one by one. OTOH, since the `aiohttp` implementation is free to decide how to accept the chunks from the server into the stream (it can combine them or even split them), `iter_any` provides no guarantees about each yielded chunk is (other than that, if they are put together, they will of course combine to whatever the server sent).

BTW, this came up when I was reading the discussion about the chunked transfer encoding and `iter_any`, and specifically [this clarification](https://github.com/aio-libs/aiohttp/pull/1900#issuecomment-302840811):

> [...] Chunked encoding is just transport protocol only, semantically it's a sequence of bytes without boundaries. Just like a TCP socket is semantically a stream of bytes while the content is transferred by packets. From my perspective partial chunks are totally correct.

asvetlov:
@pkch suggestions?

pkch:
I tried to read the code, but for the life of me I can't see what the difference between these two functions is. As far as I can tell, they do exactly the same thing (except for a possible wording in the warning message).

asvetlov:
Fixed by #2150

PR #2150, opened 2017-07-31
Feature/bugfix: make the HTTP client able to return HTTP chunks when chunked transfer encoding is used

jlacoline:
## What do these changes do?

Those changes make the `StreamReader.readchunk` method return HTTP chunks when `Transfer-Encoding: chunked` is used.
The `._buffer` attribute of `StreamReader` now contains the actual HTTP chunks instead of the TCP chunks received by the server (only if chunk encoding is used, otherwise it is the same as before)
 
Previous behaviour: In all cases, `StreamReader.readchunk` returns the oldest chunk, the chunks beeing as received by the server (i.e. TCP chunks)

New behaviour: If chunked encoding is used, `StreamReader.readchunk` returns the oldest HTTP chunks. Otherwise it will return the oldest TCP chunk.

## Why?

The current implementation of the HTTP client does not take into account how the payload is split when chunked encoding is used. The actual chunks that are stored and returned to the user correspond to TCP chunks, which are virtually meaningless in HTTP communication.
This leads to problems if someone needs to retreive the HTTP chunks as sent by the server: for example issue #1899  (note that I work in the same team as @Anvil and experience the same problems about Docker response parsing)
When reading the documentation, it seems that the `iter_chunks` generator was intended to yield HTTP chunks and not TCP chunks (related issue: #1949). 

As @asvetlov argued [here](https://github.com/aio-libs/aiohttp/pull/1900#issuecomment-302840811), the way the payload is chunked is a transport problem and should be transparent to the user.
However some HTTP servers consider the HTTP chunks they send as atomic blocks of information that have a meaning by themselves (at least Docker does, but I suppose they are not the only ones to have made that choice for data streaming). More precisely, Docker streamed responses are chunks of json objects and the json decoding is done once for each chunk, so in this scenario preserving chunk integrity is mandatory.

From my understanding, allowing to retreive HTTP chunks is a quite common feature for HTTP clients (docker clients use the requests package in the Python version and the native client in the Go version)

## How?

The C parser fires the signals `on_chunk_header` and `on_chunk_complete` at the right moments but no handler was defined to act on those signals, so I added two handlers in `_http_parser.pyx`. Those two handlers basically tell the `StreamReader` payload when to start and stop storing chunks.
I modified the Python parser to reflect this behaviour.

## Checklist


asvetlov:
I still don't like the feature.
What servers except Docker relies on chunks?
What Docker API relies on it?
https://docs.docker.com/engine/api/v1.30/#operation/ContainerAttach sends exact size of every message explicitly.

fafhrd91:
This was already discussed. This not going to happen, not in this way.

I see two possible solution:

1. Each chunk is separate stream.
2. Parser should emit end of chunk. It might be callback or special value.

Solution should not affect users who does not care about transfer encoding.

jlacoline:
@asvetlov The "attach" endpoint triggers a HTTP Upgrade so this not about chunked encoding. 
`/images/create` is an endpoint that shows the behaviour I mentionned (docker pull command). The progression of the image download is sent in several json chunks that look like that: 
(capture of my TCP packets during streaming)
```
T ::1:2376 -> ::1:52905 [AP]
b9.
{"status":"Downloading","progressDetail":{"current":32264,"total":1990402},"progress":"[\u003e                                                  ]  32.26kB/1.99MB","id":"88286f41530e"}.
.


T ::1:2376 -> ::1:52905 [AP]
ba.
{"status":"Downloading","progressDetail":{"current":130279,"total":1990402},"progress":"[===\u003e                                               ]  130.3kB/1.99MB","id":"88286f41530e"}.
.


T ::1:2376 -> ::1:52905 [AP]
ba.
{"status":"Downloading","progressDetail":{"current":228583,"total":1990402},"progress":"[=====\u003e                                             ]  228.6kB/1.99MB","id":"88286f41530e"}.
.
```



@fafhrd91 The solution I propose is not the same as the one Anvil implemented earlier, the current version of the C parser uses your second solution.
As for user impact, I am not sure there is a problem with my implementation, basically `._chunk_buffer` will be filled temporary instead of `._buffer` in the `StreamReader` class. I could add a size limit to avoid problems if this is what you suggest

fafhrd91:
With your implementation I could write very simple script that would flood server memory, and application would not be able to react.

fafhrd91:
As for impact, everyone suddenly would need to care about transfer encoding

jlacoline:
If the script you suggest consists in sending huge chunk sizes, the problem seems to be exactly the same as users sending huge Content-Length sizes isn't it? I guess it should be handled the same way?
Actually I do not know what is done in the current code to prevent this but I suppose I could apply the same logic

fafhrd91:
with content-length application has full control on how to handle huge payloads. it can close connection, it can pause connection, etc. with this patch, application can not make any decision. I think, this is biggest problem.

asvetlov:
I like suggested approach for emitting special tokens on end of chunk if opt-in parameter is passed.

fafhrd91:
Or we can just break `readchunk`

asvetlov:
I don't care about backward compatibility of `readchunk` -- now it is almost useless and not widely used I pretty sure.

fafhrd91:
ok, good.

@jlacoline would like to rewrite your PR?

asvetlov:
@fafhrd91 to be clean: the proposal is rewriting `readchunk` to emit either bytes or END_CHUNK token on getting the chunk ending. All other methods should skip END_CHUNK and concatenate bytes from the stream silently.
Do I understand you properly?

fafhrd91:
yes, that is right. maybe it should return tuple: (data, eof_chunk)
otherwise users would need to use: isinstance(data, END_OF_CHUNK)

fafhrd91:
btw do we need parameter at all if we do not care about backward compatibility?
lets just change method.

asvetlov:
I think no, the method behavior change is ok to me.

fafhrd91:
ok, good. then +1 for new proposal

asvetlov:
I not sure about returning tuple.
`data is END_OF_CHUNK` check is pretty fast, there is no need for `isinstance` check.

fafhrd91:
changing type of returning data doesn't seem very good, but that is matter of taste
here is case with json:

```python

while 1:
   data, end_of_chunk = yield from payload.readchunk()
   buffer += data
   if end_of_chunk:
        decode_json_and_do_somthing(buffer)
```

as oposite to:

```python

while 1:
   data = yield from payload.readchunk()
   if data is END_OF_CHUNK:
        decode_json_and_do_somthing(buffer)
   else:
        buffer += data
```

seems documentation should be simpler with tuple as we would need to describe each field in tuple only.
but that is up to you.

jlacoline:
I am OK with this new logic, seems to be a good compromise for usability and security.
About `iter_chunks()`, since it is using  `readchunk()`, then the generator will yield tuples of (data, end_of_chunk) and it will be up to the user to concatenate the different parts to get full chunks (maybe add a helper fonction later that does it with a size limit?).
I will update the PR when I get some time (before the end of the week I think).
As for the choice between a tuple or a special value, it seems pretty equivalent so I let you tell me which one you prefer.

asvetlov:
@After careful thinking I agree that `(bytes, bool)` tuple is better: it breaks existing code fast, users will upgrade without long pain period.
And yes, let's left up to user partial chunks concatenation.
It is low level API anyway.
Moreover helper with limit doesn't help: user still should add a code for concatenation if chunk size exceeds the limit.

jlacoline:
I updated the code to make `readchunk()` behave like specified, ready for your feedback.
However I do not understand what caused the CI job to fail, the test that failed does not seem to be linked to my changes.

jlacoline:
ping @asvetlov @fafhrd91: are you ok with the changes?

asvetlov:
Will review in a few hours.

aiohttp/streams.py
```diff
@@ -40,6 +40,17 @@ def __anext__(self):
                 raise StopAsyncIteration  # NOQA
             return rv
 
+    class ChunkTupleAsyncStreamIterator(AsyncStreamIterator):
+        @asyncio.coroutine
+        def __anext__(self):
+            try:
```
asvetlov: Please add tests for covering these lines

aiohttp/streams.py
```diff
@@ -58,20 +69,21 @@ def iter_chunked(self, n):
...
-            """Returns an asynchronous iterator that yields chunks of the
-            size as received by the server.
+            """Returns an asynchronous iterator that yields chunks of data
+            as they are received by the server. The yielded objects are tuples
+            of (bytes, bool) as returned by the StreamReader.readchunk method.
 
             Python-3.5 available for Python 3.5+ only
             """
-            return AsyncStreamIterator(self.readchunk)
+            return ChunkTupleAsyncStreamIterator(self.readchunk)
```
asvetlov, line 83: Add a test for the call.

aiohttp/streams.py
```diff
@@ -320,16 +347,34 @@ def readany(self):
...
 
-        if self._buffer:
-            return self._read_nowait_chunk(-1)
+        if not self._buffer:
+            # end of file
+            return (b"", False)
+        elif self._http_chunk_splits is not None:
+            if self._http_chunk_splits:
+                pos = self._http_chunk_splits[0]
+                if pos > self._cursor:
```
asvetlov: Add a test for case if `not pos > self._cursor`.

aiohttp/streams.py
```diff
@@ -320,16 +347,34 @@ def readany(self):
...
+        encoding is used, end_of_http_chunk is a boolean indicating if the end
+        of the data corresponds to the end of a HTTP chunk , otherwise it is
+        always False.
+        """
         if self._exception is not None:
             raise self._exception
 
         if not self._buffer and not self._eof:
+            # end of http chunk without available data
+            if self._http_chunk_splits and \
```
asvetlov: Replace backslash with brackets.

aiohttp/http_parser.py
```diff
@@ -547,18 +548,17 @@ def feed_data(self, chunk, SEP=b'\r\n', CHUNK_EXT=b';'):
                     required = self._chunk_size
                     chunk_len = len(chunk)
 
-                    if required >= chunk_len:
+                    if required > chunk_len:
```
asvetlov, line 551: Please explain why you replaced `>=` with `>`.

jlacoline: If `required == chunk_len`, then `self._chunk_size == 0` and the "else" part does what is needed. 
So lines 552 and 553 looked like duplicated code, and deduplicating this allowed me to put `self.payload.end_http_chunk_receiving()` at only one place in the code.

But the code will continue after the 'else' as opposed to the previous version where (False, None) was returned immediately. So the behaviour is not exactly the same even if the same result will be eventually returned.

So imo it looks better but if you want me to revert to the previous version tell me.

Also I'm going to remove the useless blank line(554)

aiohttp/streams.py
jlacoline: done

aiohttp/streams.py
jlacoline: done

aiohttp/streams.py
jlacoline: done

aiohttp/streams.py
jlacoline: FTR this condition was here because I was afraid of what would happen if some other calls than `readchunk` were made before. 
So I did a small modification to the condition here and added tests for those situations, covering cases where `stream.unread_data` and `stream.read` are used before `stream.readchunk`.

jlacoline:
Thanks for the review. I updated my branch according to your feedback.
I will modify `docs/client_reference.rst` when I get some time, probably next week.

asvetlov:
Good. Please drop a message when you will be ready for next review.

jlacoline:
@asvetlov ready for next review

jlacoline:
@asvetlov pinging you again in case you didn't see the last message

aiohttp/http_parser.py
asvetlov: ok, fine with the change

Issue #2157, opened 2017-08-02
StaticResource url versioning


sky-code:
Add feature for generating urls to static files with version hash to have more control over caching in browser.

## Expected behaviour
When you resolve url from `StaticResource` you can use `append_version` flag to add a hash of file (version) into url, which will change if file content changed and than for browser it's will be new url (which not exists in cache and will be reloaded)

## Example
```html
<script src="{{ url('static', filename='build/js/bundle.js', append_version=True) }}"></script>
```
with `append_version` flag `StaticResource` will append `v` param into url query string  with hash of the file
```
/static/build/js/bundle.js?v=aDc7UcFqFqXFUOYI8crlpxDAW-i41AGSNdDP8eQebSM%3D
```

asvetlov:
Not sure if I want to have the feature in aiohttp.
It looks like solving very specific problem in core while it could be done in third party easily.

Do other libraries/frameworks have the feature? Do nginx support file hashes?

sky-code:
Yes, another web frameworks already have same feature
https://www.davepaquette.com/archive/2015/05/06/link-and-script-tag-helpers-in-mvc6.aspx > Cache Busting

https://www.google.com.ua/search?q=asp.net+core+mvc+append-version

this feature implementation very similar to asp.net core mvc implementation.
I don't see any reason to reject this feature, it's fully optional (you need to use this flag to have any impact) and fully backward compatible with exists code

asvetlov:
Ok

asvetlov:
Done by #2158

PR #2158, opened 2017-08-02
add append_version feature into StaticResource url resolver

sky-code:
Fully backward compatible feature (has no impact on exists code)

## Related issue number
Please goto #2157 for description and discussion

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -427,9 +427,10 @@ class StaticResource(PrefixResource):
     VERSION_KEY = 'v'
 
     def __init__(self, prefix, directory, *, name=None,
-                 expect_handler=None, chunk_size=256*1024,
+                 expect_handler=None, chunk_size=256 * 1024,
                  response_factory=StreamResponse,
-                 show_index=False, follow_symlinks=False):
+                 show_index=False, follow_symlinks=False,
+                 append_version=False):
         super().__init__(prefix, name=name)
         try:
             directory = Path(directory)
@@ -446,6 +447,7 @@ class StaticResource(PrefixResource):
         self._chunk_size = chunk_size
         self._follow_symlinks = follow_symlinks
         self._expect_handler = expect_handler
+        self._append_version = append_version
 
         self._routes = {'GET': ResourceRoute('GET', self._handle, self,
                                              expect_handler=expect_handler),
@@ -453,15 +455,17 @@ class StaticResource(PrefixResource):
                         'HEAD': ResourceRoute('HEAD', self._handle, self,
                                               expect_handler=expect_handler)}
 
-    def url(self, *, filename, append_version=False, query=None):
+    def url(self, *, filename, append_version=None, query=None):
         # is there with_query need to be used?
         # with_query remove previous query options
         url = self.url_for(filename=filename, append_version=append_version)
         if query is not None:
-            return str(url.with_query(query))
+            return str(url.update_query(query))
         return str(url)
 
-    def url_for(self, *, filename, append_version=False):
+    def url_for(self, *, filename, append_version=None):
+        if append_version is None:
+            append_version = self._append_version
         if isinstance(filename, Path):
             filename = str(filename)
         while filename.startswith('/'):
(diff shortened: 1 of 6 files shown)


asvetlov:
Please mention new parameter and behavior in docs

sky-code:
I have added documentation for `url_for` but not for `url` because it's deprecated

asvetlov:
The PR only generates new version tag by `url_for()` call.

Should it return redirection response on version mismatch?

Should we move `append_version` parameter to constructor? 

It might simplify the usage: just add static files handler with parameter set (`app.router.add_static('/prefix', path_to_static, append_version=True)`), no need to pass `append_version=True` to every `url_for()` call.

sky-code:
Should it return redirection response on version mismatch?
Not necessarily, no at this time. If someone asks to make it and provide good explanation of advantages then it will be added

Should we move append_version parameter to constructor?
maybe, set default value in constructor than override it in each call if necessary

aiohttp/web_urldispatcher.py
```diff
@@ -450,17 +453,51 @@ def __init__(self, prefix, directory, *, name=None,
...
                                               expect_handler=expect_handler)}
 
-    def url(self, *, filename, query=None):
-        return str(self.url_for(filename=filename).with_query(query))
+    def url(self, *, filename, append_version=False, query=None):
+        # is there with_query need to be used?
+        # with_query remove previous query options
+        url = self.url_for(filename=filename, append_version=append_version)
+        if query is not None:
+            return str(url.with_query(query))
```
asvetlov: Should be `update_query`

aiohttp/web_urldispatcher.py
```diff
@@ -450,17 +453,51 @@ def __init__(self, prefix, directory, *, name=None,
...
+                # TODO cache file content
+                # with file watcher for cache invalidation
+                with open(str(filepath), mode='rb') as f:
+                    file_bytes = f.read()
+                h = self.get_file_hash(file_bytes)
+                url = url.with_query({self.VERSION_KEY: h})
+                return url
+        return url
+
+    def get_file_hash(self, byte_array):
```
asvetlov: Add `_` prefix to method name: it is not a part of public API

aiohttp/web_urldispatcher.py
```diff
@@ -450,17 +453,51 @@ def __init__(self, prefix, directory, *, name=None,
...
+                    filename = filename[1:]
+                filepath = self._directory.joinpath(filename).resolve()
+                if not self._follow_symlinks:
+                    filepath.relative_to(self._directory)
+            except (ValueError, FileNotFoundError) as error:
+                # relatively safe
+                return url
+            except Exception as error:
+                # perm error or other kind!
+                # TODO log error here?
```
asvetlov: Not sure if we should catch the exception.
url is useless: next GET will raise an error.
Moreover next `open` call might fail too.

aiohttp/web_urldispatcher.py
```diff
@@ -450,17 +453,51 @@ def __init__(self, prefix, directory, *, name=None,
...
-        return URL(url)
+        url = URL(url)
+        if append_version is True:
+            try:
+                if filename.startswith('/'):
+                    filename = filename[1:]
+                filepath = self._directory.joinpath(filename).resolve()
+                if not self._follow_symlinks:
+                    filepath.relative_to(self._directory)
+            except (ValueError, FileNotFoundError) as error:
```
asvetlov: What case could raise `ValueError`?

sky-code: I don't know, this is copied code from `_handle` method, anyway i remove it

aiohttp/web_urldispatcher.py
sky-code: removed

aiohttp/web_urldispatcher.py
sky-code: done

aiohttp/web_urldispatcher.py
sky-code: done

aiohttp/web_urldispatcher.py
```diff
@@ -443,24 +447,57 @@ def __init__(self, prefix, directory, *, name=None,
...
-    def url(self, *, filename, query=None):
-        return str(self.url_for(filename=filename).with_query(query))
-
-    def url_for(self, *, filename):
+    def url(self, *, filename, append_version=None, query=None):
+        # is there with_query need to be used?
+        # with_query remove previous query options
+        url = self.url_for(filename=filename, append_version=append_version)
+        if query is not None:
+            return str(url.update_query(query))
```
asvetlov: Please add a test for this case

aiohttp/web_urldispatcher.py
```diff
@@ -443,24 +447,57 @@ def __init__(self, prefix, directory, *, name=None,
...
             filename = str(filename)
         while filename.startswith('/'):
             filename = filename[1:]
         filename = '/' + filename
         url = self._prefix + URL(filename).raw_path
-        return URL(url)
+        url = URL(url)
+        if append_version is True:
+            try:
+                if filename.startswith('/'):
```
asvetlov, line 476: Please add a test if `filename` doesn't start with `/`

aiohttp/web_urldispatcher.py
```diff
@@ -443,24 +447,57 @@ def __init__(self, prefix, directory, *, name=None,
...
+                    filename = filename[1:]
+                filepath = self._directory.joinpath(filename).resolve()
+                if not self._follow_symlinks:
```
asvetlov, line 479: Test for `_follow_symlink` needed

aiohttp/web_urldispatcher.py
sky-code: Done

aiohttp/web_urldispatcher.py
sky-code: Done

aiohttp/web_urldispatcher.py
sky-code: Done

asvetlov:
Thanks!

Issue #2170, opened 2017-08-05
Support custom Forwarded header parameters


vfaronov:
## Long story short

[`aiohttp.web.BaseRequest.forwarded`](http://aiohttp.readthedocs.io/en/stable/web_reference.html#aiohttp.web.BaseRequest.forwarded) does not parse the `Forwarded` header in accordance with [RFC 7239](https://tools.ietf.org/html/rfc7239). In particular, it does not support custom parameters.

The whole point of deploying `Forwarded` instead of the legacy `X-Forwarded-*` zoo is extensibility. For example, custom parameters could be very useful for identifying proxies reliably (as opposed to the current practice of “take the *n*th proxy from the end”).

## Expected behaviour

Given a request with headers:

```
Forwarded: for=203.40.91.97;by=10.1.2.3;secret=JxED1spY6t;
Forwarded: for=10.1.2.3;by=10.4.5.6;;some="other, stuff"
```

the `forwarded` attribute should have the value:

```
(
    mappingproxy({
        'for': '203.40.91.97',
        'by': '10.1.2.3',
        'secret': 'JxED1spY6t',
    }),
    mappingproxy({
        'for': '10.1.2.3',
        'by': '10.4.5.6',
        'some': 'other, stuff',
    }),
)
```

## Actual behaviour

The `forwarded` attribute has the value `({}, {}, {})`.

Note how a comma inside a quoted-string confuses aiohttp, which could even lead to security problems if the application is programmed to look at the *n*th value from the end or the beginning.

## Steps to reproduce

Run this server program:

```
from aiohttp import web
async def handle(request):
    return web.Response(text=repr(request.forwarded))
app = web.Application()
app.router.add_get('/', handle)
web.run_app(app)
```

and send requests to it with curl:

```
$ curl -v localhost:8080/ \
>   -H 'Forwarded: for=203.40.91.97;by=10.1.2.3;secret=JxED1spY6t;' \
>   -H 'Forwarded: for=10.1.2.3;by=10.4.5.6;;some="other, stuff"'
```

## Your environment

aiohttp Git master, Python 3.5, Linux

vfaronov:
I’m experimenting with a different parser:
https://gist.github.com/vfaronov/890ec14572d4ccacfa54616e39a43783
It is more correct than aiohttp’s current parser, while being about as fast (in my unscientific tests) and simple. Let me know if you like this approach and want me to make a PR.

fafhrd91:
+1 I think this makes sense

PR #2173, opened 2017-08-06
Fix parsing the Forwarded header

vfaronov:
## What do these changes do?

Parse the `Forwarded` header more carefully, while staying about as fast, simple, and robust in the face of potential injection attacks. (Speed was measured with IPython’s `%timeit` on my laptop on a few typical and pathological header values.)

Most valid `Forwarded` headers that used to be parsed incorrectly by aiohttp are now parsed correctly. In particular:

- commas and semicolons are allowed inside quoted-strings;
- empty forwarded-pairs (as in ``for=_1;;by=_2``) are allowed;
- non-standard parameters are allowed (although this alone could be easily done in the previous parser).

Valid headers containing obs-text are still not parsed correctly, as this was an intentional decision in the previous parser (see comments) that I did not change.

Also, the previous parser used to bail out of (invalid) forwarded-elements containing duplicate parameter names. No rationale was given in the code, and I don’t think this is important, so the new parser doesn't enforce this.

## Related issue number

#2170

## Checklist

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -27,7 +27,7 @@ FileField = collections.namedtuple(
 _TCHAR = string.digits + string.ascii_letters + r"!#$%&'*+.^_`|~-"
 # '-' at the end to prevent interpretation as range in a char class
 
-_TOKEN = r'[{tchar}]*'.format(tchar=_TCHAR)
+_TOKEN = r'[{tchar}]+'.format(tchar=_TCHAR)
 
 _QDTEXT = r'[{}]'.format(
     r''.join(chr(c) for c in (0x09, 0x20, 0x21) + tuple(range(0x23, 0x7F))))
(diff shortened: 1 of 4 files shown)


aiohttp/web_request.py
```diff
@@ -208,37 +204,45 @@ def forwarded(self):
...
-                            # duplicate param in field-value
-                            break
-                        if value and value[0] == '"':
-                            # quoted string: replace quotes and escape
-                            # sequences
-                            value = _QUOTED_PAIR_REPLACE_RE.sub(
-                                r'\1', value[1:-1])
-                        fvparams[param.lower()] = value
+        elems = []
+        for field_value in self._message.headers.getall(hdrs.FORWARDED, ()):
```
socketpair, line 208: @fafhrd91 @asvetlov 
It is easily possible to build FSM which will parse this structure using `regex` module. Unfortunatelly built-in regexps are not able to capture all repeating matches.

Pros:
* Less code
* Less error-prone
* Faster
* Ability to parse other structures using higly-efficient FSMs

Cons:
* Dependency on regex module
* More difficult to understand (unsure, maybe more simplier)
* Non-pythonic way.

(Anyway, maybe merge as shown here, but re-write later, I can do that)

asvetlov: @socketpair please merge the PR first than feel free to create a new one for regexps based approach.

socketpair: @asvetlov how can I sure that other members agree with my decision to merge ? (sorry, I'm asking BEFORE making something awful)

vfaronov: @socketpair I’m not sure an FSM can handle my `test_single_forwarded_header_injection1` case, which necessarily involves a form of backtracking. And I think this case is important to handle, at least until deployment experience shows that all reverse proxies take precautions against that.

asvetlov: @socketpair the trick borrowed from CPython works: declare intention to merge (LGTM or github pull request approval), wait a day and do merge tomorrow if nobody objects.

asvetlov:
Thanks

Issue #2174, opened 2017-08-07
Insecure way to handle json requests


fduxiao:
## Long story short

[BaseRequest.json](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_request.py#L499) coroutine returns request body directly, whose `Content-Type` should have been checked. 

## Expected behaviour

`Content-Type` should be ensured to be `application/json` so that one can make sure it is sent by ajax, since there are always csrf attacks while ajax can not be forged cross site.

## Actual behaviour

`Content-Type` is not checked. 

## Steps to reproduce

I setup the server using following code. 
``` python
#!/usr/bin/env python3

from aiohttp import web
import asyncio


app = web.Application()


async def index(request):
    if request.method == 'GET':
        return web.Response(text='Hello Aiohttp!')
    elif request.method == 'POST':
        j = await request.json()
        return web.json_response(j)


def setup_routes(app):
    app.router.add_route('*', '/', index)


if __name__ == '__main__':
    setup_routes(app)
    web.run_app(app, host='127.0.0.1', port=8080)
```

And I request the server with following command. 

```shell
# both of the following will work
curl -v -d '{"abc":2}' -H "Content-Type: application/json" http://127.0.0.1:8080
curl -v -d '{"abc":2}' http://127.0.0.1:8080  # This should fail.
```

Actually [flask](http://flask.pocoo.org/) handles json requests only with `Content-Type: application/json`

## Environment
```shell
> uname -a
Darwin bogon 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64 x86_64

> python3 --version  # installed with homebrew
Python 3.6.2

> python3 -c 'import aiohttp; print(aiohttp.__version__)'
2.0.7

> brew --version 
Homebrew 1.3.0
Homebrew/homebrew-core (git revision 51fde5; last commit 2017-08-04)

```

socketpair:
Please use latest aiohttp. See #1723

fduxiao:
That's not what I need. I'm not trying to get the request from a server but to handle the request sent to the server. Now the way to handle will cause CSRF attacks. An error should be raised or directly return 405 instead of getting the json result. Please read my code carefully.

asvetlov:
For preventing CSRF you need much more than just content type check.

fduxiao:
[...]

PR #2177, opened 2017-08-08
Check the content type before returning the json value.

fduxiao:
## What do these changes do?

Check the content type before returning the json value.

When getting json requests one could directly `await request.json()`.
If one'd like to ignore the `Content-Type`, he/she should specify `if_check` manually i.e. `await request.json(if_check=False)`. 

## Related issue number

referencing #2174 

## Checklist

  * The format is Xiao Tan.
  * name it `2174.bugfix` for example (588.bug)


aiohttp/web_request.py
```diff
@@ -510,8 +510,19 @@ def text(self):
...
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, wrong_format=None, if_check=True):
+        """
+        Return BODY as JSON.
+        If the `Content-Type` is not `application/json`,
+        then the request is rejected.
+        You can also handle it manually by passing the
+        `wrong_format` parameter.
+        """
+        if if_check and self.content_type != 'application/json':
```
kxepal: How would I use, say, [jsonapi](http://jsonapi.org/) if it will never match this hardcoded content-type?

aiohttp/web_request.py
```diff
@@ -510,8 +510,19 @@ def text(self):
         return bytes_body.decode(encoding)
 
     @asyncio.coroutine
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, wrong_format=None, if_check=True):
```
kxepal: I think signature is quite overloaded. I suggest just to allow to pass what exact content-type should be considered as valid. No match - throw an exception, nothing more. For BC, default value could be None until some release when we change it to 'application/json'. Thoughts?

aiohttp/web_request.py
fduxiao: Reasonable. I will add it.

aiohttp/web_request.py
```diff
@@ -510,19 +510,26 @@ def text(self):
...
-        You can also handle it manually by passing the
-        `wrong_format` parameter.
+        You can also handle it manually by passing the `err_handler` parameter.
         """
-        if if_check and self.content_type != 'application/json':
-            if wrong_format is None:
+        wrong_format_flag = False
+        if isinstance(allowed_types, str):
+            wrong_format_flag = allowed_types != self.content_type
+        elif isinstance(allowed_types, list):
```
kxepal: What if allowed_types be a set or tuple or frosenset or whatever container?

aiohttp/web_request.py
```diff
@@ -510,19 +510,26 @@ def text(self):
...
+        """ Return BODY as JSON. If the `Content-Type` is not in allowed_types,
         then the request is rejected.
-        You can also handle it manually by passing the
-        `wrong_format` parameter.
+        You can also handle it manually by passing the `err_handler` parameter.
         """
-        if if_check and self.content_type != 'application/json':
-            if wrong_format is None:
+        wrong_format_flag = False
+        if isinstance(allowed_types, str):
```
kxepal: What if I receive `application/json` in legacy form when it will have a charset option defined explicitly like `application/json;charset=utf-8`?

aiohttp/web_request.py
```diff
@@ -510,19 +510,26 @@ def text(self):
...
+        elif allowed_types is None:  # do not check the content_type
+            wrong_format_flag = False
+
+        # if fail to pass the check
+        if wrong_format_flag:
+            if err_handler is None:
                 raise HTTPNotAcceptable()
-            elif callable(wrong_format):
-                yield from wrong_format()
+            elif callable(err_handler):
```
kxepal: You expect callable, but below err_handler is a coroutine actually.

aiohttp/web_request.py
```diff
@@ -510,19 +510,26 @@ def text(self):
...
+                yield from err_handler()
```
kxepal: Why to pass here err_handler after all? You can wrap `req.json()` call with try/except and do the rest on user code side.

aiohttp/web_request.py
```diff
@@ -510,19 +510,26 @@ def text(self):
...
-    def json(self, *, loads=json.loads, wrong_format=None, if_check=True):
-        """
-        Return BODY as JSON.
-        If the `Content-Type` is not `application/json`,
+    def json(self, *, loads=json.loads, allowed_types=None, err_handler=None):
+        """ Return BODY as JSON. If the `Content-Type` is not in allowed_types,
         then the request is rejected.
-        You can also handle it manually by passing the
-        `wrong_format` parameter.
+        You can also handle it manually by passing the `err_handler` parameter.
```
kxepal: I think there is need to note what will happens if content-type won't match.

aiohttp/web_request.py
fduxiao: I meant to have reduced the use of try/except so one could get rid of the fact that if you want to handle the error yourself, you need the try/except block i.e. one doesn't need to raise the error, catch it and handle it since catch an error is [costly](https://docs.python.org/3/faq/design.html#how-fast-are-exceptions).

aiohttp/web_request.py
fduxiao: So will iterable make sense? I mean how about `hasattr(alowed_types, '__iter__')`?

aiohttp/web_request.py
fduxiao: I can only use module `inspect` to decide whether an object is a coroutine. Should I decide whether it is a coroutine or not or even something else(says a string) and make calling correspondingly. I think more details should be given.

aiohttp/web_request.py
kxepal: May be do like it was don't in `ClientResponse.json()` ? It's quite abstract from what kind of type was passed and the behaviour would be consistent between client and server.

aiohttp/web_request.py
kxepal: I would suggest to not pass `exc_handler` at all. No argument - no problems (:

aiohttp/web_request.py
kxepal: No need to worry about try/except usage. If you do it, you explicitly sign about that you're aware about that things here could go wrong. Some custom argument that does the same makes it not so oblivious. 

As about speed, you shouldn't worry about catch performance. I bet, your app business logic will be million times slower than that.

aiohttp/web_request.py
fduxiao: How about just checking the [type, subtype and suffix](http://aiohttp.readthedocs.io/en/stable/api.html#aiohttp.helpers.parse_mimetype)? And should json be [decoded according to the charset](https://github.com/pallets/flask/blob/0.12-maintenance/flask/wrappers.py#L152)?

asvetlov: I suggest just supporting a container (`list`, `tuple`, whatever) only and raising explicit TypeError on single string.
By default it should be `None` for backward compatibility. That's sad but we cannot change it.
Application might have a parameter for enabling checking json types in the application by default though.

kxepal: > How about just checking the type, subtype and suffix? 

That's correct, but will make things too much complicated.

> And should json be decoded according to the charset?

In modern RFC's it's mandatory should be UTF-8. However, there are lot of legacy around which forces charset via params. I don't thing we should care about non-UTF8 JSON, but if client asks about it explicitly, there is no reason to deny him, imho. 

> I suggest just supporting a container (list, tuple, whatever) only and raising explicit TypeError on single string.

It's good idea, but it conflicts with how client's `json()` acts. I don't think we should have different behaviour of a similar things - that would annoy people all around.

fduxiao: So I think we can use either `Container` or `str`, and `None` is for wildcard (or should I use a `'*'` as in `add_route`, since `allowed_types=None` looks weird for wildcard); otherwise a `TypeError` is raised

asvetlov: @kxepal 

Client's json has `content_type` parameter which is forced to always check content type.
Server API is not compatible already (it doesn't do check).
What is your suggestion? Rename `allowed_types` to content type and accept `str` only?

I'm -0 for *string or list* parameters in API but could live with them.

aiohttp/web_request.py
```diff
@@ -510,8 +510,26 @@ def text(self):
...
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, allowed_types=None, err_handler=None):
+        """ Return BODY as JSON. If the `Content-Type` is not in allowed_types,
+        then the request is rejected.
+        You can also handle it manually by passing the `err_handler` parameter.
+        """
+        wrong_format_flag = False
+        if isinstance(allowed_types, str):
+            wrong_format_flag = allowed_types != self.content_type
```
socketpair: The `self.content_type` may be `application/json; charset=UTF-8`. Am I right ? So, use `startswith` or handle that in a different way.

aiohttp/web_request.py
```diff
@@ -510,8 +510,26 @@ def text(self):
...
+        elif isinstance(allowed_types, list):
```
socketpair: what about tuple and set ?

aiohttp/web_request.py
```diff
@@ -510,8 +510,26 @@ def text(self):
         return bytes_body.decode(encoding)
 
     @asyncio.coroutine
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, allowed_types=None, err_handler=None):
```
socketpair: I vote for `allowed_types='application/json'`

kxepal: That would be breaking change. See previous discussions.

aiohttp/web_request.py
```diff
@@ -510,8 +510,26 @@ def text(self):
...
+            wrong_format_flag = self.content_type not in allowed_types
+        elif allowed_types is None:  # do not check the content_type
+            wrong_format_flag = False
+
+        # if fail to pass the check
+        if wrong_format_flag:
+            if err_handler is None:
+                raise HTTPNotAcceptable()
+            elif callable(err_handler):
+                yield from err_handler()
```
socketpair: 1. Please do not check if it is callable. Instead, check return value of call. if it is awaitable -- await it. (see signal handlers in aiohttp)
2. In current code, with `err_handler=42` -- error will be just ignored, which is a bug.
3. errors must not be handled as callbacks or so. There are exceptions for that.

aiohttp/web_request.py
fduxiao: Yes. So how do you think of these codes:
```python
content_type = self.content_type
(mime_type, subtype, suffix, parameters) = parse_mimetype(content_type)
content_type = mime_type + '/' + subtype
if suffix != '':
    content_type += '+' + suffix

charset = parameters.get('charset', None)
```

aiohttp/web_request.py
fduxiao: Maybe all Containers should be supported, says
```python
if isinstance(allowed_types, str):
    wrong_format_flag = allowed_types != self.content_type
elif isinstance(allowed_types, collections.Container):
    wrong_format_flag = self.content_type not in allowed_types
elif allowed_types is None:  # do not check the content_type
    wrong_format_flag = False
else:  # for other type
    raise TypeError("Unsupported container of `Content-Type`")
```

aiohttp/web_request.py
fduxiao: Shall we directly check if the mime type starts with `'application/'` and ends with `'+json'` i.e. [flask's way to deal with it](https://github.com/pallets/flask/blob/0.12-maintenance/flask/wrappers.py#L119)?  Hence no need to consider about how should allowed_type behavior

aiohttp/web_request.py
fduxiao: As we discussed above, there's no need to give callbacks. Instead, only the 406 error is raised.

aiohttp/web_request.py
socketpair: seems OK, I want to review this part in context of whole path.

aiohttp/web_request.py
socketpair: `-1` for any containers, though. Even `list`.

aiohttp/web_request.py
socketpair: 1. So, if backward compatibility is a very important in this question (I don't agree with that) than maybe force checking by default, but define some compatibility switch which turns old behavior. I mean some swith of form `aiohttp.DEFAULT_JSON_CT=None`. Like `logging.raiseExceptions=False`.
2. I'm `-1` for having API to pass list/tuple/container, since I can't imagine real case.

kxepal: Good point on containers. Server is pretty specific about content type it could handle.

kxepal: I don't like global switches. You have argument and it's default. `None` disables the check. For awhile it will be default to preserve BC. After some release it becomes `application/json` in defaults and may be we could prohibit `None` usage, but that's another story.

aiohttp/web_request.py
kxepal: If we stay with string only argument as @socketpair suggested, that's would be friendly default behaviour with almost no cost.

aiohttp/web_request.py
fduxiao: Could I suggest a function(mapping a string to boolean) be passed as an argument to filter the `Content-Type` and `str`, `None` are also accepted and nothing more is supported?

kxepal: I would recommend to avoid callback style. This is the road to make aiohttp acts weird depending on what functions user passed around. There is pretty simple rules to find out if MIME type is json-like. In the rest cases user could override that behavior by explicit string value.

fduxiao: So that means the user won't want to behavior weirdly? Then I just need to check whether the `type` is `str` or `tuple` or `list` or `set` and compare them directly, or that it is something that means json-like or wildcard. Also I need to derive the charset from `Content-type` and pass it to loads.

kxepal: Why he should? `.json()` provides stable interface and guarantee about what it does. User may customize default expected content-type via string argument. Why there is need to do more?

If user need to do more complex content type check, it should make it at middleware level, not in `.json()` call.

aiohttp/web_request.py
```diff
@@ -510,10 +511,42 @@ def text(self):
...
+        content_type = self.content_type
+        (mime_type, subtype, suffix, parameters) = parse_mimetype(content_type)
+        content_type = mime_type + '/' + subtype
+        if suffix != '':
+            content_type += '+' + suffix
+
+        charset = parameters.get('charset', None)
+
+        if isinstance(allowed_types, str):
+            if allowed_types == 'json':
```
socketpair: Useless. Users must supply full contenttype like xxxx/yyyyy

aiohttp/web_request.py
```diff
@@ -510,10 +511,42 @@ def text(self):
...
+                if content_type != 'application/json':
+                    if not content_type.startswith('application') or \
+                       not content_type.endswith('+json'):
+                        raise HTTPNotAcceptable()
+            elif allowed_types != content_type:
+                raise HTTPNotAcceptable()
+        elif isinstance(allowed_types, set):
```
socketpair: -2 for containers. Why they are intended for ? Which user-story (case) do you suggest ?

aiohttp/web_request.py
```diff
@@ -510,10 +511,42 @@ def text(self):
...
+            if content_type not in allowed_types:
+                raise HTTPNotAcceptable()
+        elif allowed_types is None:  # do not check the content_type
+            pass
+        else:  # for other type
+            raise TypeError("Unsupported type for allowed `Content-Type`")
+
         body = yield from self.text()
-        return loads(body)
+        return loads(body, encoding=charset)
```
socketpair: The other arguments have the same meaning as in load(), except encoding which is ignored and deprecated.

socketpair: Charset should go as additional argument of the function. And should be passed to .text() if defined.

aiohttp/web_request.py
fduxiao: This means 'json-like'. View my code after that. I will then check  whether the `Content-Type` is 'application/json' or starts with 'application' and ends with '+json'. If it is a string and is not 'json', I will compare it with `Content-Type` directly. This may be convenient i.e. one won't need to always specify the argument for different json format

aiohttp/web_request.py
fduxiao: Previously, we discussed what if one sends 'application/json; charset=utf-8' and 'application/json' is hard-coded and now I strip the charset. My concern is that will there be something else the frontend will send fail to pass the examination, says the handler need to handle both `application/json` and `application/javascript` or some uncontrollable i.e. for compatibility of an old frontend framework.

aiohttp/web_request.py
fduxiao: What if the frontend sends a request specifying charset? And I think a json object should be parsed with loads, with [encoding supported by python](https://docs.python.org/3/library/json.html#json.loads) itself. Should `.text()` decode something like `"\\u4e16\\u754c\\u4f60\\u597d"`?

aiohttp/web_request.py
socketpair: We should keep code of aiohttp simple and clear. Simple thing should be done in  a simple way, and difficult ones should be possible.

If an specific application needs to handle such very obscure situation, it may control content-type by it's own before decoding of json. And also, I think, situation you have provided is just an imagination, not the real one. If no, please describe deeper.

aiohttp/web_request.py
socketpair: PLEASE look at the aiohttp sources before asking and before changing the code. `.text()` will analyze conten-type header (regarding charset) before converting bytes to text.

fduxiao: That's not what I mean, says `b'"\\u4e16\\u754c\\u4f60\\u597d"'.decode('utf-8')` will never decode anything but `json.loads('"\\u4e16\\u754c\\u4f60\\u597d"')` will give `'世界你好'`. Thus the argument you give to `.text()` will never be used when you decode json.

aiohttp/web_request.py
```diff
@@ -510,8 +511,24 @@ def text(self):
         return bytes_body.decode(encoding)
 
     @asyncio.coroutine
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, allowed_type=None):
+        """ Return BODY as JSON. If the `Content-Type` is not in allowed_types,
+        then the request is rejected.
+        """
+        content_type = self.content_type
```
socketpair: it's not required to parse `self.content_type` if  `allowed_type` is `None`

aiohttp/web_request.py
```diff
@@ -510,8 +511,24 @@ def text(self):
...
+        (mime_type, subtype, suffix, parameters) = parse_mimetype(content_type)
+        content_type = mime_type + '/' + subtype
+        if suffix != '':
+            content_type += '+' + suffix
+
+        if isinstance(allowed_type, str):
```
socketpair: Much simplier:

```
if allowed_type is not None:
    parse content type
    if parsed_ct != allowed_type:
        raise ...
```

without any `else`

aiohttp/web_request.py
```diff
@@ -510,8 +511,24 @@ def text(self):
         return bytes_body.decode(encoding)
 
     @asyncio.coroutine
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, allowed_type=None):
+        """ Return BODY as JSON. If the `Content-Type` is not in allowed_types,
```
socketpair: `is not in` is wrong now.

aiohttp/web_request.py
socketpair: JSON-decoding is two-step process. First, it needs to convert from bytes to characters (symbols). Thi is done by UTF-8 (typically) decoder in `.text()`. Second-step -- is to decode slash-escapes according to JSON-format. this is done using `loads()`. So, encoding you are talking about is for first step only. As you can see, letter `u` after every slash signifies that every item is UNICODE symbol (as opposed to some byte value). And so encoding is not applicable.

aiohttp/web_request.py
```diff
@@ -510,8 +511,26 @@ def text(self):
...
 
     @asyncio.coroutine
-    def json(self, *, loads=json.loads):
-        """Return BODY as JSON."""
+    def json(self, *, loads=json.loads, allowed_type=None):
+        """ Return BODY as JSON. If the `Content-Type` does not equal
+        allowed_type, then the request is rejected.
+        """
+
+        if allowed_type is None:  # do not check the content_type
```
socketpair, line 523: I vote for replacing this chain with only one condition like:

```python
if allowed_type is not None:
   ...
```

aiohttp/web_request.py
```diff
@@ -510,8 +511,26 @@ def text(self):
...
+        elif isinstance(allowed_type, str):
+            # parse the content_type
+            content_type = self.content_type
+            (mime_type, subtype, suffix, _) = parse_mimetype(content_type)
+            content_type = mime_type + '/' + subtype
+            if suffix != '':
+                content_type += '+' + suffix
+
+            if allowed_type != content_type:
+                raise HTTPNotAcceptable()
```
socketpair, line 534: Also, I'm not sure if this concrete exception should be raised. Also please provide an message in exception.

@asvetlov @fafhrd91

aiohttp/web_request.py
fduxiao: Yes. I did a search and found it so I removed these codes.

aiohttp/web_request.py
fduxiao: So, what do you guys think now? Should the message be passed as an argument? I think the error should be raised directly since that's exactly what I want for a JSON API in most cases and I think we've reached the agreement that for special requirements one should write a middleware or catch the exception as we discussed before.

samuelcolvin: What would be wrong with just 

```python
def json(self, *, loads=json.loads, allowed_type='application/json'):
    if allowed_type and self.content_type != allow_type:
        raise HTTPNotAcceptable(text='invalid content-type: "{}" not "{}"'.format(self.content_type, allowed_type))
    ...
```

?

If users really want more sophisticated behaviour than this they can add their own logic, there's nothing very complicated here that couldn't be replicated in a custom coroutine.

samuelcolvin:
also is `406` really the correct response?

from wikipedia:

> The requested resource is capable of generating only content not acceptable according to the Accept headers sent in the request.

That doesn't sound quite right. `400` is very generic but is technically more correct I think.

asvetlov:
Fixed by another PR #3889

PR #2181, opened 2017-08-08
2176.removal of TimeService 

greg-barnett:
removed TimeService, which had bugs causing it to return very incorrect dates.

It has been mentioned that TimeService was written as a performance optimization, but I have not been able to recreate any conditions in which the very small performance improvement would be noticeable.


greg-barnett:
wrong branch.

PR #2180, opened 2017-08-08
Simplify time caching in DATE: header formatting (#2176)

socketpair:
## Related issue number

Issue #2176

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new news fragment into the `changes` folder
  * name it `<issue_id>.<type>` for example (588.bug)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #2182, opened 2017-08-08
issue 2176 Time service removal

greg-barnett:
removed TimeService, which had bugs causing it to return very incorrect dates.

It has been mentioned that TimeService was written as a performance optimization, but I have not been able to recreate any conditions in which the very small performance improvement would be noticeable.


fafhrd91:
-1 we should cache date value

greg-barnett:
@fafhrd91 If my benchmarking is correct, caching the formatted date string saves approximately
0.00000142 seconds per request (will vary on hardware of course).  That is a small enough amount of time that I don't think it will ever be noticed.

tests/test_helpers.py
```diff
@@ -379,47 +379,8 @@ def test_is_ip_address_invalid_type():
...
-    def test_recalc_time(self, time_service, mocker):
-        time = time_service._loop.time()
-        time_service._time = time
-        time_service._strtime = 'asd'
-        time_service._count = 1000000
-        time_service._on_cb()
-        assert time_service._strtime is None
-        assert time_service._time > time
-        assert time_service._count == 0
+def time_service():
```
socketpair: What's this ?

fafhrd91:
oh, so that means all my assumptions about performance was wrong. all my performance related work is probably wrong as well, nice :)

@asvetlov comments?

tests/test_helpers.py
greg-barnett: I missed some dead code in tests, have removed it.

greg-barnett:
@fafhrd91 Do you have any notes on the benchmarking/profiling you did?  If you do, I can try to replicate your results.  There might be some case where time.time() or time.gmtime() is slow, in which case TimeService/caching would make sense.

fafhrd91:
I don't have any notes, that was relatively long time ago. but @asvetlov came to similar conclusion, he added TimeService.

greg-barnett:
If it matters, the strftime format I'm using is more performant than the one used in TimeService.  Don't have the numbers handy right now, but it ran about 30-40% faster.  After I tried that, I noticed that the same format was used elsewhere, so I made it into a constant.  hdrs.RFC822_FORMAT

fafhrd91:
I am fine with removing cache. just waiting for @asvetlov to approve.

asvetlov:
IIRC adding time service reduced response time by at about 10% for cases like 
```
async def handler(request):
    return web.Response('OK')
```

Your number is 0.00000142 sec per request for time service disabled.
Now aiohttp server shows about 16_000 RPS for handling trivial requests.
 `0.00000142 * 16000 == 0.02272`, about 2% of response time is spent on date string formatting.

It's much less than 10% but still significant degradation.
Honestly I don't recall what numbers was at time of adding time service.

But your proposal to use `RFC822_FORMAT = '%a, %d %b %Y %H:%M:%S GMT'` as time format is wrong. 
Formatting used by aiohttp  is borrowed from `wsgiref` module IIRC.
It explicitly uses English names for weekdays and months but  `%a` is **Locale’s** abbreviated weekday name.
On servers with non-English locale it produces incorrect string.

Maybe we need resurrection of #2180 but I still don't like global vars, we could rewrite `TimeService` class to use the approach -- yet another syscall during response processing is cheap but string formatting is expensive.

fafhrd91:
I think global var is fine, it is private. Removing time service reduces amount of doc we need

asvetlov:
We don't mention `TimeService` in docs, it's totally private API.
Why not just drop updating on `call_later` and use `time.time`?

Anyway, if @fafhrd91 @socketpair and other committers prefer globals in this particular case -- I can live with it.
#2180 changes private API only, we could change it again without breaking our users in next releases.

fafhrd91:
it is part of WebRequest api, technically we should document it.

asvetlov:
@fafhrd91 really? Why?
I have strong feeling either `TimeService` nor `get_formatted_datetime` from #2180 should be documented.
We could add underscore prefix to names if needed but users should never import these names explicitly -- we don't provide them in `__all__` for `aiohttp` and `aiohttp.web`.

socketpair:
@greg-barnett Could you benchmark old and new code in same environment ?
0.00000142 got from one computer, and 16_000 RPS got from another.

fafhrd91:
@asvetlov time_service is public attribute. 

I'd prefer to replace class with one function if we touch it in any case. Amount of functionality is too small for whole class. But I am fine with rather decision

asvetlov:
@fafhrd91 I see your point now.
After your split `web_reqrep.py` into `web_request.py` and `web_response.py` the history is lost, I'm too lazy for digging deeper.
Maybe I've added the property but maybe you.
Most likely it's my change.

Nevermind, the property should be deleted. Fortunately it's not documented.

I'm fine with rather decision too.

@socketpair please reopen #2180 
Wait for @greg-barnett activity for a week (we don't schedule next release yet, a week is good enough  for getting feedback).
If he explicitly argue against the proposal -- let's discuss.
If not -- please merge your PR.

socketpair:
It's better to fix current PR (add caching) than edit mine. Well, I will mix them up tomorrow.

greg-barnett:
I'm probably not going to have time to work on this for at least a few days.

If the solution is going to involve global variables, I'd also prefer that someone else do the commit.  I don't want someone to see a global, do a a git blame, and then go away with a bad opinion of me :)

@asvetlov I missed that it was using Locale's names.  So, in addition to TimeService, the existing code for headers where I replaced format string with the constant will also need to be fixed.

I suspect that the proper way to do this is something like the following (completely untested)

    class TimeService:
        def __init__(self):
            if locale.get_locale(LC_TIME) in LOCALES_WITH_ENGLISH_MONTH_AND_DAY_NAMES:
                self._format_method = self._rfc822_fast
            else:
                self._format_method = self._rfc822_safe

            self._formatted_now = None
            self._time = None

        def rfc822_format(timestamp):
            return self._format_method(timestamp)

        def rfc822_now():
             now = time.time()
            if now != self._time:
                self._time = now
                self._formatted_now = self._format_method(self._time)
            return self._formatted_now

        def _rfc822_fast(timestamp):
             return strtime(RFC822_FORMAT, timestamp)

        def _rfc822_safe(timestamp):
             # use the slightly slower formatting code from existing TimeService

asvetlov:
@greg-barnett I like your sketch.
Let's modify time service behavior but not add global vars.
I hate them, @fafhrd91 agree with any decision, @socketpair has no strong objection at least.
Perhaps we need a new PR.
@greg-barnett @socketpair I hope you'll find a time for working on it soon.
I love to get the problem fixed but personally prefer to work on other issues.

socketpair:
I have nechmarked `.strftime()` approach and one, used in original code (with tuple lookup). And guess what ? `.strftime()` is SLOWER (!)

```
In [29]: %time xxx()
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 28.1 µs
```

```
%time time.strftime('%a, %d %b %Y %H:%M:%S GMT')
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 38.6 µs
```

```
def xxx():
   ...:     year, month, day, hh, mm, ss, wd, y, z = time.gmtime(time.time())
   ...:     return "%s, %02d %3s %4d %02d:%02d:%02d GMT" % (_weekdayname[wd], da
   ...: y, _monthname[month], year, hh, mm, ss)
```

socketpair:
when I remove argument of `gmtime()` it becomes even more faster:

```
In [39]: %time xxx()
CPU times: user 0 ns, sys: 0 ns, total: 0 ns
Wall time: 20.7 µs
```

socketpair:
I updated my PR with code borrowed from @greg-barnett. This PR probably could be closed.

greg-barnett:
@socketpair your PR doesn't cover the additional issue we've discovered: the locale dependent format being used for headers was already being used for last-modified headers in aiohttp/webresponse.py

This is what I believe to be the requirements to fix this issue:

- Properly format all date/time header values to match rfc822.
- Always use the current time for the Date header.
- Either improve performance, or have only a trivial degradation.

These are not requirements, but they should probably be followed anyway:

- not introduce any global variables
- the code should be simpler if at all possible

My current plan, which I hope to have completed within the next 2 days.

Write benchmarking code that tests:
- getting an rfc822 formatted time for the Date header
- getting an rfc822 formatted time for the Last-Modified header, passing in a timestamp
- how many requests per second for a GET request that returns Date header
- how many requests per second for a GET request that returns both Date and Last-Modified headers

Run the benchmarking code twice, once in en_US, once in a locale where "%a" breaks rfc822 format.

I think my solution is going to be similar to the TimeService I posted in my last comment, but make the class a singleton (with a different name, probably RFC822DateFormatter), and have one public method in helpers.py that calls it.  I'm thinking a singleton is less ugly than private global variables.

greg-barnett:
I created a very simple server, and ran apache bench against it.

All variations tried came in around 2,000 requests per second, with a fairly large amount of variation.

Code that attempted to pick a formatting strategy based on locale did not increase performance noticeably.

My latest version uses class variables, which are probably just as bad as global variables.  @socketpair's code, if expanded to fix Last-Modified headers, is probably the way to go.

If my current approach is the way you want to go, I'll add some unit tests.

asvetlov:
I believe it's fixed by #2180

asvetlov:
@socketpair ?

socketpair:
Yes, TimeService is removed, so bug is closed. But I did not check if caching of datetime may be removed completely. Maybe it saves nano-seconds. I don't know.

Issue #1128, opened 2016-08-26
Move ssl-related params from TCPConnector to request args


asvetlov:
## Long story short

Now ssl configuration is tightly coupled with `TCPConnector`.

It is not convenient (user should create a connector and pass it into session) and disallows to use the same `ClientSession` for working with different ssl contexts etc.

I think we should move `verify_ssl`, `fingerprint` and `ssl_context` into `ClientSession.request` and `ClientSession.ws_connect` like we have done for `proxy` and `proxy_auth` (see #998).
Current TCPConnector's parameters should be kept as default values which may be overridden by request without any warning/error.

magicgoose:
I think they should not be moved to single requests but to **ClientSession** instead. Multiple requests to the same host+port can reuse connection (or also not reuse, it's statically unknown in general) — https://github.com/aio-libs/aiohttp/blob/3da3a0142e3f8a50c990824ea6d4669643a9fa9e/aiohttp/connector.py#L340 and TLS params are inherently property of the connection, not a HTTP request, so if the user gives different TLS parameters to these requests, that would be very misleading and even non-stable behavior (because it's not known in general if the connection will be reused for a specific request).

@asvetlov do you agree?

asvetlov:
I think if `requests` accepts `verify` and `cert` for every HTTP request call -- why `aiohttp` cannot do the same?

magicgoose:
It can, of course, but I'm not sure if there are real advantages of doing this. But on second thought I see that it's probably fine for user, it would be just a little harder to implement. So, okay…

(at first I thought that there would be also things like which versions of TLS and which ciphers to negotiate with the server, and then there's no meaningful way of changing these constraints for next request with the same connection in general way, other than closing the connection and doing the handshake again, but if this level of control is not going to be allowed, then setting just `verify` and `cert` per request can be defined easily)

asvetlov:
Fixed by #2184

PR #2184, opened 2017-08-09
Add SSL related params to ClientSession.request

cecton:
## What do these changes do?

...

...

## Related issue number

#1128 

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -2,9 +2,11 @@ import asyncio
 import collections
 import io
 import json
+import ssl
 import sys
 import traceback
 import warnings
+from hashlib import md5, sha1, sha256
 from http.cookies import CookieError, Morsel
 from urllib.request import getproxies
 
@@ -32,6 +34,16 @@ __all__ = ('ClientRequest', 'ClientResponse', 'RequestInfo')
 
 RequestInfo = collections.namedtuple(
     'RequestInfo', ('url', 'method', 'headers'))
+
+
+HASHFUNC_BY_DIGESTLEN = {
+    16: md5,
+    20: sha1,
+    32: sha256,
+}
+
+
+_SSL_OP_NO_COMPRESSION = getattr(ssl, "OP_NO_COMPRESSION", 0)
 
 
 class ClientRequest:
@@ -66,7 +78,13 @@ class ClientRequest:
                  chunked=None, expect100=False,
                  loop=None, response_class=None,
                  proxy=None, proxy_auth=None, proxy_from_env=False,
-                 timer=None, session=None, auto_decompress=True):
+                 timer=None, session=None, auto_decompress=True,
+                 verify_ssl=None, fingerprint=None, ssl_context=None):
+
+        if verify_ssl is False and ssl_context is not None:
+            raise ValueError(
+                "Either disable ssl certificate validation by "
+                "verify_ssl=False or specify ssl_context, not both.")
 
         if loop is None:
             loop = asyncio.get_event_loop()
@@ -89,6 +107,8 @@ class ClientRequest:
         self.response_class = response_class or ClientResponse
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
+        self._verify_ssl = verify_ssl
+        self._ssl_context = ssl_context
 
         if loop.get_debug():
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
@@ -101,6 +121,7 @@ class ClientRequest:
         self.update_content_encoding(data)
         self.update_auth(auth)
         self.update_proxy(proxy, proxy_auth, proxy_from_env)
+        self.update_fingerprint(fingerprint)
 
         self.update_body_from_data(data)
         self.update_transfer_encoding()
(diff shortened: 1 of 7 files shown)


cecton:
Having issues with too much mocking.... if someone has any advice?

asvetlov:
I suggest the following:

1. Make a self signed certificate and push it into `tests` (or just reuse `tests/sample.crt` if it's good for you).
2. Write functional tests using real app started with self signed certs.

Feel free to ask if my recommendation is not clean.

P.S.
Very thank you for picking up the issue and working on implementation.

asvetlov:
P.P.S.
Side note: as committer you are free to drop github template for new issues and PRs.
These templates are created for helping newbies to file PR properly.
You know our rules already and might use free form for issue/PR description.
For example PR checklist could be skipped if you are keeping the list in your memory.
The only requirement: if PR is related to existing issue please mention it in description text -- it simplifies navigation between issue and PR.

asvetlov:
Obviously it's up to you: use free form or file all template fields if template is more comfortable to you.
Personally I prefer free form for both.

asvetlov:
Regarding to proxy tests: take a look on `tests/proxy_functional.py` design.

cecton:
@asvetlov PR template is very fine for me :) it helped me a lot to not forget anything (documentation and change logs especially) but thx

I will have to change a few (all?) tests that mock the entire ClientRequest class. This thing is settings properties in the ClientRequest that I don't want.

asvetlov:
Are you talking about proxy tests which should be affected by the PR, not fingerprint related tests, right?

In this case please don't hesitate to update all mocks if method signatures changed. I see no way to do it better.

cecton:
Yes at this point the code I did should have kept the original behavior intact and the tests should pass unchanged. But 3 of the 4 failures are due to the fact that the mocking is actually giving a fingerprint to the ClientRequest objects.

asvetlov:
Please update failed mocks, it's totally correct.

asvetlov:
@cecton would you continue work on the PR?
I have a feeling that we need a new aiohttp release, will dedicate a full time week to lift up pending issues.

cecton:
@asvetlov it shouldn't have take that much time but I got stuck with the mocking thing and then I had a lot of changes recently (change of project basically).

I was going to ask for help with the tests actually.

cecton:
@asvetlov okay ready for review

asvetlov:
Looks good. Feel free to merge, while I very appreciate full test coverage.

cecton:
@asvetlov Thanks I will improve that! I thought I had all the cases but I hadn't check the diff on codecov.

asvetlov:
Just a hint: install codecov browser extension: https://docs.codecov.io/v4.3.6/docs/browser-extension
It allows to see coverage just in github 'Files Changes' tab.
If you'll update contributors instruction to mention the plugin -- it would be awesome!

jettify:
that is nice trick!

aiohttp/client_reqrep.py
```diff
@@ -307,6 +328,35 @@ def update_proxy(self, proxy, proxy_auth, proxy_from_env):
...
         self.proxy_auth = proxy_auth
 
+    def update_fingerprint(self, fingerprint):
+        if fingerprint:
+            digestlen = len(fingerprint)
+            hashfunc = HASHFUNC_BY_DIGESTLEN.get(digestlen)
+            if not hashfunc:
+                raise ValueError('fingerprint has invalid length')
+            elif hashfunc is md5 or hashfunc is sha1:
+                warnings.simplefilter('always')
```
kxepal: Wouldn't this cause suddenly global side effect?

cecton: I have no idea. I copy-pasted that from there a610cf12209326fba5e6660c1644cbff55b7c07d

cecton: I think you're right. After this the "warning filter specifications" is altered. https://docs.python.org/3.5/library/warnings.html#warnings.simplefilter @asvetlov can you confirm you want to keep this behavior?

asvetlov:
@jettify 
> that is nice trick!

What are you talking about?

aiohttp/client_reqrep.py
asvetlov: No.
@kxepal good catch.
Please don't alter `warnings` filter but duplicate the message to `aiohttp.log.client_logger`.
From my experience people are much more sensible to log messages than deprecation warnings.
Log level should be [`warning'](https://docs.python.org/3/library/logging.html#logging.Logger.warning) sure.

jettify:
@asvetlov codecov browser extension, very usefull

asvetlov:
@jettify thanks

aiohttp/client_reqrep.py
cecton: I suppose I change the one in connector too? ^_^

cecton:
@asvetlov I just saw that the original ticket mention a "move" of those parameters. She I implement a deprecation for the using the parameter in the connector?

asvetlov:
I not sure.
Obviously there is a need for per-request ssl configuration but I don't know is deprecation makes sense.
Maybe yes.
Guys what do you think?

cecton:
@asvetlov the fingerprint is something you will get per server basis so per request makes sense. On the other hand the connector has the argument fingerprint like it would connect always to the same server. To me it doesn't make any sense. Unless maybe for a UNIX socket?

`verify_ssl` and `ssl_context` could make sense because you may want to disable the entire SSL checking for a complete ClientSession.

asvetlov:
Make sense

asvetlov:
Thank you!

PR #2187, opened 2017-08-10
use trafaret.IP instead

thijstriemstra:
Trafaret provides an `IP` type.


thijstriemstra:
And without this change the demo breaks with Trafaret >= 0.11

asvetlov:
Thanks!
I very appreciate PRs for aiohttp demos.
Unfortunately demos are not test covered by their nature. Test suite need Postgres and manual steps for adding pg user, etc. and etc.
Thus the baby lives without parent control and it is outdated very often.

asvetlov:
If you want to polish http://aiohttp.readthedocs.io/en/stable/tutorial.html -- you are welcome

Issue #2167, opened 2017-08-04
Enable to remove "Content-Length: 0" for GET Request


ga0:
## Long story short

When I send GET request to some http server with the  "Content-Length: 0" header, will cause a aiohttp.ServerDisconnectedError. If I remove this header, everything goes well.

I don't know why these server hate empty content-length header so much.

## Expected behaviour

Some args to remove Content-Length if it is 0 and method is GET.

## Steps to reproduce

You can test by using the comands below.

```
curl -v -H "Content-Length: 0" http://115.238.34.168

curl -v http://115.238.34.168 
```

## Your environment
aiohttp: 2.0.7

OS & Python: 
* macOS with Python 3.6.1
* CentOS7 with Python 3.4.5

socketpair:
https://stackoverflow.com/questions/978061/http-get-with-request-body?answertab=active#tab-top

Content length mean presensce of a body, but with zero length. I prefer to close as "not a bug". No one http-client will send such requests. And sending body with GET-request difinitely signify a bug in client. RFC on HTTP is not strict enough, so current implementation of this case still leaves aiohttp RFC-compliant. Handling such reqests will not bring any benefits, just code bloat. 

Why you want such functionality ?

Also please update aiohttp and re-test.

ga0:
* Why you want such functionality ?

Because I found send GET request with  "Content-Length: 0" header to  some kind of servers will cause a bad request error.

* Also please update aiohttp and re-test.

I have tried aiohttp-2.2.5, still a "Content-Length: 0" in the request's header.

socketpair:
Since it's not an security error, does not fix any real problem, does not add new functionality, I'm -3 for  this issue.

fafhrd91:
is it possible to disable `content-length` with `skip_auto_headers`?

asvetlov:
Unfortunately no.

fafhrd91:
I think we should support this use case

asvetlov:
Agree.
I see two options:
1. Respect `skip_auto_headers` in `Content-Length` and `Content-Encoding`
2. Implement #2019 and add a low level control for all HTTP request transformation details.

I prefer second bullet but it requires much more work.
Skipping mandatory fields like `Content-Length` in our prime API could make a mess (not 100% sure but I have bad guts feeling about this approach).

socketpair:
1. `skip_auto_headers` controls RESPONSE headers generation.
2. Topic started say us about `content-length` in REQUEST.
[...]

PR #2191, opened 2017-08-12
Do not send `Content-length: 0` in GET/HEAD/TRACE/OPTIONS requests (#2167)

socketpair:
## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -51,11 +51,14 @@ ConnectionKey = namedtuple('ConnectionKey', ['host', 'port', 'ssl'])
 
 
 class ClientRequest:
-
-    GET_METHODS = {hdrs.METH_GET, hdrs.METH_HEAD, hdrs.METH_OPTIONS}
+    GET_METHODS = {
+        hdrs.METH_GET,
+        hdrs.METH_HEAD,
+        hdrs.METH_OPTIONS,
+        hdrs.METH_TRACE,
+    }
     POST_METHODS = {hdrs.METH_PATCH, hdrs.METH_POST, hdrs.METH_PUT}
-    ALL_METHODS = GET_METHODS.union(POST_METHODS).union(
-        {hdrs.METH_DELETE, hdrs.METH_TRACE})
+    ALL_METHODS = GET_METHODS.union(POST_METHODS).union({hdrs.METH_DELETE})
 
     DEFAULT_HEADERS = {
         hdrs.ACCEPT: '*/*',
@@ -129,7 +132,8 @@ class ClientRequest:
         self.update_fingerprint(fingerprint)
 
         self.update_body_from_data(data)
-        self.update_transfer_encoding()
+        if data or self.method not in self.GET_METHODS:
+            self.update_transfer_encoding()
         self.update_expect_continue(expect100)
 
     @property
--- a/tests/test_client_request.py
+++ b/tests/test_client_request.py
@@ -264,7 +264,7 @@ def test_skip_default_useragent_header(make_request):
 
 
 def test_headers(make_request):
-    req = make_request('get', 'http://python.org/',
+    req = make_request('post', 'http://python.org/',
                        headers={'Content-Type': 'text/plain'})
 
     assert 'CONTENT-TYPE' in req.headers
@@ -273,7 +273,7 @@ def test_headers(make_request):
 
 
 def test_headers_list(make_request):
-    req = make_request('get', 'http://python.org/',
+    req = make_request('post', 'http://python.org/',
                        headers=[('Content-Type', 'text/plain')])
     assert 'CONTENT-TYPE' in req.headers
     assert req.headers['CONTENT-TYPE'] == 'text/plain'
@@ -529,16 +529,16 @@ def test_connection_header(loop, conn):
 def test_no_content_length(loop, conn):
     req = ClientRequest('get', URL('http://python.org'), loop=loop)
     resp = req.send(conn)
-    assert '0' == req.headers.get('CONTENT-LENGTH')
+    assert req.headers.get('CONTENT-LENGTH') is None
     yield from req.close()
     resp.close()
 
 
 @asyncio.coroutine
-def test_no_content_length2(loop, conn):
+def test_no_content_length_head(loop, conn):
     req = ClientRequest('head', URL('http://python.org'), loop=loop)
     resp = req.send(conn)
-    assert '0' == req.headers.get('CONTENT-LENGTH')
+    assert req.headers.get('CONTENT-LENGTH') is None
     yield from req.close()
     resp.close()
 
@@ -586,7 +586,7 @@ def test_content_type_skip_auto_header_form(loop, conn):
 
 
 def test_content_type_auto_header_content_length_no_skip(loop, conn):
-    req = ClientRequest('get', URL('http://python.org'),
+    req = ClientRequest('post', URL('http://python.org'),
                         data=io.BytesIO(b'hey'),
                         skip_auto_headers={'Content-Length'},
                         loop=loop)
@@ -645,6 +645,7 @@ def test_pass_falsy_data_file(loop, tmpdir):
     yield from req.close()
 
 
+# Elasticsearch API requires to send request body with GET-requests
 @asyncio.coroutine
 def test_get_with_data(loop):
     for meth in ClientRequest.GET_METHODS:
@@ -673,7 +674,7 @@ def test_bytes_data(loop, conn):
 
 @asyncio.coroutine
 def test_content_encoding(loop, conn):
-    req = ClientRequest('get', URL('http://python.org/'), data='foo',
+    req = ClientRequest('post', URL('http://python.org/'), data='foo',
                         compress='deflate', loop=loop)
     with mock.patch('aiohttp.client_reqrep.PayloadWriter') as m_writer:
         resp = req.send(conn)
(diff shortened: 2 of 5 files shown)


tests/test_client_request.py
```diff
@@ -626,11 +626,11 @@ def test_post_data(loop, conn):
 @asyncio.coroutine
 def test_pass_falsy_data(loop):
     with mock.patch(
-            'aiohttp.client_reqrep.ClientRequest.update_body_from_data'):
+            'aiohttp.client_reqrep.ClientRequest.update_body_from_data') as upd:
         req = ClientRequest(
             'post', URL('http://python.org/'),
             data={}, loop=loop)
-        req.update_body_from_data.assert_called_once_with({})
```
socketpair: thanks to Pycharm.

tests/test_client_request.py
```diff
@@ -652,12 +652,10 @@ def test_pass_falsy_data_file(loop, tmpdir):
 @asyncio.coroutine
 def test_get_with_data(loop):
     for meth in ClientRequest.GET_METHODS:
-        req = ClientRequest(
-            meth, URL('http://python.org/'), data={'life': '42'},
-            loop=loop)
-        assert '/' == req.url.path
-        assert b'life=42' == req.body._value
```
socketpair: Completely wrong test.

aiohttp/client_reqrep.py
```diff
@@ -82,6 +88,10 @@ def __init__(self, method, url, *,
         self.url = url.with_fragment(None)
         self.original_url = url
         self.method = method.upper()
+
+        if self.method in self.GET_METHODS and any((data, compress, chunked)):
+            raise ValueError('Method %s should not contain request-data.' % self.method)
```
kxepal: This will eventually break requests to elastic search. Technically, you can send GET with payload and that's correct from the point of spec.

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html

socketpair: OK, I didn't know that ElasticSearch predates RFC, where it's said that HTTP-server SHOULD ignore payload of GET-requests. I'm very upset about everything that.

kxepal: Yeap, that was a big surprise for me as well, but that's how things are. You can google the subject, iirc, on some w3c mailing list that case was discussed and considered as fine. Also, SHOULD is a strong recommendation, but it's not a mandatory behaviour as what MUST does.

socketpair: I removed the checks and fix tests accordingly.

aiohttp/client_reqrep.py
```diff
@@ -35,11 +35,14 @@
 
 
 class ClientRequest:
-
-    GET_METHODS = {hdrs.METH_GET, hdrs.METH_HEAD, hdrs.METH_OPTIONS}
+    GET_METHODS = {
+        hdrs.METH_GET,
+        hdrs.METH_HEAD,
+        hdrs.METH_OPTIONS,
+        hdrs.METH_TRACE,
```
kxepal, line 58: Should DELETE method be here are well?

socketpair: Nothing is said regarding DELETE in RFC. And also:

1. a message-body is explicitly forbidden in responses to HEAD requests (section 9, and 9.4 specifically)
2. an entity-body is explicitly forbidden in TRACE requests only, all other request types are unrestricted (section 9, and 9.8 specifically)

Should I add raise ValueError in constructor in these requests ?

aiohttp/client_reqrep.py
```diff
@@ -103,7 +106,8 @@ def __init__(self, method, url, *,
         self.update_proxy(proxy, proxy_auth, proxy_from_env)
 
         self.update_body_from_data(data)
-        self.update_transfer_encoding()
+        if data or self.method not in self.GET_METHODS:
```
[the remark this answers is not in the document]
socketpair: Sure. It is required to send `Content-Length: 0` in these requests.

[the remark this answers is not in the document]
kxepal: Hm, ok. Asking because you can send POST without content-length set and that's fine as well. curl does so.

aiohttp/client_reqrep.py
kxepal: > Should I add raise ValueError in constructor in these requests ?

Let's open another issue for that. If we can help users to not produce invalid requests that would be good.

socketpair: OK: #2192

asvetlov:
Should we don't send zero `Content-Length` by **default** or add and option for skipping the header autogeneration explicitly?
I feel the second option is more correct, proper web server should respect either `Content-Length` or use chunked encoding, otherwise HTTP 1.1 protocol is broken, isn't it?

fafhrd91:
I think we should set and allow to skip.

asvetlov:
@fafhrd91 please make your last message clean: do you mean we should set `Content-Length: 0` by default and skip it if `skip_auto_headers` in `ClientSession` contains `Content-Length`?

fafhrd91:
@asvetlov your interpretation is right. I think our current behavior is right we just need a way to override defaults

socketpair:
No! totally disagree with you.
1. GET requests SHOULD NOT contain body. HEAD/TRACE MUST NOT contain body.
2. Body presence is detected either by having `content-length` or `transfer-ecoding`.
3. Yes, server should respect any of these headers, but presence of body in GET very often considered as a bug, since RFC say us SHOULD NOT. For example, libhtp (yes with one t) used in Suricata IDS/IPS AFAIK consider such requests as wrong. AFAIK
4. All HTTP-clients I know do not set these headers in GET-requests. Why we should ?

fafhrd91:
I agree in general, but there is reality. Dropping `content-length` would be huge semantic change, can we reliably test it? `content-length` exists from the begging of the client code, I'd play it safe.

socketpair:
First of all, this change will actually affect only GET-requests. People should never rely on the `content-length` in such requests (if data was not passed). So, I consider we are not breaking anything real. But instead, we are fixing ancient bug.

fafhrd91:
Ok. Your call.

@asvetlov @kxepal ?

kxepal:
I'm with @socketpair, he has a point. `Content-Length` signs a presense of message body. No message body - no need to send a header. This, I think, has a good analogue: `({}).get('key')` vs `({'key': None}).get('key')` - both produces the same result, but in first case there is no `key` while in second there is, but it has `None` value. Missing value != empty one.

Here is a story about `Content-Length: 0` with GET request: http://redmine.lighttpd.net/issues/1717
It turns into two edge case: servers have to understand `Content-Length: 0` as no content at all, while clients shouldn't send such HEAD/GET requests.

asvetlov:
At least ElasticSearch uses GET with BODY for multi-get requests.
I think it is gray zone but we should support it somehow (we do call elastic by aiohttp at work for example).

kxepal:
Can we ask aioes maintainers to run their tests with this pathc and no `Content-Length: 0`? I'm pretty sure everything will be fine.

asvetlov:
aioes is not maintained anymore but there are async transports for official elasticsearch-py.

Do you suggest not send `Content-Length` if body is not present but keep it only if the body is sent?

kxepal:
Yes, I do. There indeed could be semantical difference since no body means `resp.content is None` while zero sized `resp.content == b''`, but I don't think there is a practical difference between. When you send GET/POST requests via curl with no payload, it doesn't sends `Content-Length` header as well:

```
$ curl -v -XPOST http://localhost:5984
* Rebuilt URL to: http://localhost:5984/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 5984 (#0)
> POST / HTTP/1.1
> Host: localhost:5984
> User-Agent: curl/7.50.1
> Accept: */*
```

vs

```
$ curl -v -XPOST http://localhost:5984 --data ''
Note: Unnecessary use of -X or --request, POST is already inferred.
* Rebuilt URL to: http://localhost:5984/
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 5984 (#0)
> POST / HTTP/1.1
> Host: localhost:5984
> User-Agent: curl/7.50.1
> Accept: */*
> Content-Length: 0
> Content-Type: application/x-www-form-urlencoded
```

socketpair:
https://stackoverflow.com/questions/7323958/are-put-and-post-requests-required-expected-to-have-a-request-body


image

socketpair:
So, current fix is right :) and CURL has a bug (regarding POST without an entity).

asvetlov:
Well, looks good but I like to have explicit test for GET with BODY -- we should support this mode.

socketpair:
@asvetlov there is one. An original test:

https://github.com/socketpair/aiohttp/blob/7601e892343bdbc2b9b12c0cdc0d8fe729a461ae/tests/test_client_request.py#L654

Even more, I added the comment `# Elasticsearch API requires to send request body with GET-requests`

asvetlov:
This test doesn't perform actual call.
I believe we need a test in `test_client_functional.py` with explicit check that GET BODY was passed through the wire correctly.

socketpair:
Myabe just add `assert  int(req.headers['CONTENT_LENGTH']) > 0` ?

asvetlov:
Please no, only with functional test I could sleep safe and quiet :)

socketpair:
ohhh. I forgot about this PR. Will fix tests in nearest time.

asvetlov:
Ping

kxepal:
@socketpair @asvetlov 
Anything else blocks  to ship it?

socketpair:
Only specific tests are missing (@asvetlov asks me to write them). I'm in Dubai now (Gitex), so I can't work on it. Beside this, everything is ready and coulud be merged.

socketpair:
I just have rebased, no code change.

asvetlov:
Ok

Issue #1342, opened 2016-10-27
Route for subdomains


alikzao:
## Describe

Sometimes there are situations when it is necessary for the current project to create a subdomain just as another rout.
And hang on to this subdomain a handler that is already there.
## Expected behaviour

For example it might look something like this:

```
app.router.add_route('GET',  '/',  main_page)
app.router.add_route('GET',  '/',  handler, subdomain="domain")

def handler():
      print('text')
```

asvetlov:
After #1301 it's possible to register subapplications by prefix.

I suggest the same technique but for resolving by subdomain.

cynecx:
I am currently using the approach of adding a custom middleware which specifically filters out unknown hostnames, the only issue I am currently having is that it quite hard to support real micro-services with different domains, but this isn't an issue anymore especially after @asvetlov has landed sub-applications :).

asvetlov:
I mean we could add a method `router.add_subdomain('sub.domain.com', subapp)` like recently added `router.add_subapp('/prefix', subapp)`.

cynecx:
@asvetlov Do you also consider supporting wildcards in the domain name as well a falback so unknown domains are handled too?

asvetlov:
Why not?

cynecx:
@asvetlov Well, by looking at your example I thought you would like to keep it very simple, so in that case I assumed it would come without wildcard and fallback domain support. But yeah, it's good to know that you will include that too :).

ghost:
Is it implemented?

fafhrd91:
no, it is not

ghost:
Very sad :(

fafhrd91:
Would you provide PR?

ghost:
Nah, I'm not familiar with aiohttp

alikzao:
@TiberiumPY I did not wait for decisions, so switched to another framework

ghost:
@alikzao what framework?

fafhrd91:
You have plenty of options. Pyramid, flask.

espositofulvio:
If this has not been done yet, I'd like to have add_domain instead of add_subdomain. It would be a nice feature which could easily enable multi-site support for aiohttp applications. Further, having remove_subapp() and remove_domain() would complete the feature allowing to dinamically add and remove features to a website and in my opinion could be a nice thing to have (paid features come to mind but also a multi-site cms where user can actually load the features they want could be handled this way easily).

alex-eri:
[...]

PR #2194, opened 2017-08-12
Implement Application.add_subdomain (#1342)

amirouche:
## What do these changes do?

Implement basic subdomain support 

Yes, one can add a subapp for a given domain as follow:

```python
    subdomain = web.Application()
    subdomain.router.add_get('/to', handler)

    # setup root app
    app = web.Application()
    app.add_subdomain('example.aiohttp.test', subdomain)
```

## Related issue number

#1342 

## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -27,7 +27,7 @@ from .web_request import *  # noqa
 from .web_response import *  # noqa
 from .web_server import Server
 from .web_urldispatcher import *  # noqa
-from .web_urldispatcher import PrefixedSubAppResource
+from .web_urldispatcher import PrefixedSubAppResource, SubDomainResource
 from .web_ws import *  # noqa
 
 
@@ -190,6 +190,23 @@ class Application(MutableMapping):
             raise ValueError("Prefix cannot be empty")
 
         resource = PrefixedSubAppResource(prefix, subapp)
+        self.router.register_resource(resource)
+        self._reg_subapp_signals(subapp)
+        self._subapps.append(subapp)
+        if self._loop is not None:
+            subapp._set_loop(self._loop)
+        return resource
+
+    def add_subdomain(self, subdomain, subapp):
+        if self.frozen:
+            raise RuntimeError(
+                "Cannot add sub domain application to frozen application")
+        if subapp.frozen:
+            raise RuntimeError("Cannot add frozen application")
+        if subdomain == '':
+            raise ValueError("Subdomain cannot be empty")
+
+        resource = SubDomainResource(subdomain, subapp)
         self.router.register_resource(resource)
         self._reg_subapp_signals(subapp)
         self._subapps.append(subapp)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -672,6 +672,54 @@ class PrefixedSubAppResource(PrefixResource):
             prefix=self._prefix, app=self._app)
 
 
+class SubDomainResource(AbstractResource):
+
+    def __init__(self, subdomain, app):
+        super().__init__()
+        self._subdomain = subdomain
+        self._app = app
+
+    def add_prefix(self, prefix):
+        super().add_prefix(prefix)
+        for resource in self._app.router.resources():
+            resource.add_prefix(prefix)
+
+    def url_for(self, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


amirouche:
It's not exactly what is explained in #1342 but I think it's what was intended.

Mind the fact, that I don't think this can be merged given the checklist is not complete and that IMO the user experience is not good enough. With the current code, the domain is hardcoded which, I think, will be a pain during dev because the dev has to override DNS resolution in her `/etc/hosts` making it impossible to access the production website via the normal URL. What would be best, is somehow to add somewhere the information about the domain to the root app. So that, it can be configured in production and dev.

WDYT?

aiohttp/web_urldispatcher.py
```diff
@@ -672,6 +672,54 @@ def __repr__(self):
...
+        self._subdomain = subdomain
+        self._app = app
+
+    def add_prefix(self, prefix):
+        super().add_prefix(prefix)
+        for resource in self._app.router.resources():
+            resource.add_prefix(prefix)
+
+    def url_for(self, *args, **kwargs):
+        raise RuntimeError(".url_for() is not supported "
```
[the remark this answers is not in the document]
amirouche: Ok.

aiohttp/web.py
```diff
@@ -197,6 +197,23 @@ def add_subapp(self, prefix, subapp):
...
         return resource
 
+    def add_subdomain(self, subdomain, subapp):
+        if self.frozen:
+            raise RuntimeError(
+                "Cannot add sub domain application to frozen application")
+        if subapp.frozen:
+            raise RuntimeError("Cannot add frozen application")
+        if subdomain == '':
+            raise ValueError("Subdomain cannot be empty")
```
amirouche: Ok.

aiohttp/web_urldispatcher.py
```diff
@@ -672,6 +672,54 @@ def __repr__(self):
...
+        raise RuntimeError(".url() is not supported "
+                           "by subdomain sub-application root")
+
+    def get_info(self):
+        return {'app': self._app,
+                'subdomain': self._subdomain}
+
+    @asyncio.coroutine
+    def resolve(self, request):
+        if not request.headers['Host'] == self._subdomain:
```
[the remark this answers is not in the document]
amirouche: It will fail. Should I do something specific here?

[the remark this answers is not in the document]
samuelcolvin: Surely just `.get()` would work?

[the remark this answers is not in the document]
amirouche: Ok.

aiohttp/web.py
```diff
@@ -197,6 +197,23 @@ def add_subapp(self, prefix, subapp):
             subapp._set_loop(self._loop)
         return resource
 
+    def add_subdomain(self, subdomain, subapp):
```
espositofulvio, line 200: Can we rename this to add_domain(self, domain, subapp)? The way you did it actually enables multi-site apps on different domains, not only subdomains I think, so why don't just change the name?

aiohttp/web_urldispatcher.py
```diff
@@ -672,6 +672,54 @@ def __repr__(self):
             prefix=self._prefix, app=self._app)
 
 
+class SubDomainResource(AbstractResource):
+
+    def __init__(self, subdomain, app):
+        super().__init__()
+        self._subdomain = subdomain
```
espositofulvio, line 679: If we go for changing subdomain to domain, I guess it might be useful to use Yarl to parse the subdomain here so we get for free the port number and can actually dispatch on the actual host (domain + port number)

aiohttp/web_urldispatcher.py
espositofulvio, line 702: Again, pending the decision to have domain in place of subdomain, I guess using Yarl to parse the Host header here is also safer.

aiohttp/web.py
amirouche: I agree.

aiohttp/web_urldispatcher.py
amirouche: Do you mean that we should patch yarl to parse the domain wiht subdomain?

amirouche:
I think https://github.com/aio-libs/aiohttp/pull/2809 is better and more general

Issue #754, opened 2016-01-29
How to close websockets coroutines ?


Insoleet:
Hello,

I'm running the following code which hangs : 

``` python
import quamash
import asyncio
import aiohttp
import sys


class Testor:
    def __init__(self):
        self.loop = asyncio.get_event_loop()
        self.myWS = None

    async def connect(self):
        self.myWS = aiohttp.ws_connect('ws://metab.ucoin.io:9201/ws/block')
        async with self.myWS as ws:

            async for msg in ws:
                if msg.tp == aiohttp.MsgType.text:
                    print(msg.data)
                elif msg.tp == aiohttp.MsgType.closed:
                    print("closed")
                    break
                elif msg.tp == aiohttp.MsgType.error:
                    print("error")
                    break

    def disconnect(self):
        self.myWS.close()

    def run(self):
        self.loop.call_later(2, self.disconnect)
        self.loop.run_until_complete(self.connect())

test = Testor()
test.run()
```

What am I doing wrong and how am I supposed to close websockets sessions ?

I also tried the following, which resulted in the same behaviour : 

``` python

    def disconnect(self):
        asyncio.wait_for(self.myWS.close(), timeout=2)
```

Thanks a lot for your great library.

jettify:
Do you have any error? Because your code looks good, since you are using `async with`, connection is closed implicitly on moving out from context manager block.

Insoleet:
It hanks, no error available. If I kill the program, I get : 

```
Traceback (most recent call last):
  File "/home/inso/code/quamash/bug.py", line 37, in <module>
    test.run()
  File "/home/inso/code/quamash/bug.py", line 34, in run
    self.loop.run_until_complete(self.connect())
  File "/home/inso/.pyenv/versions/3.5.0/lib/python3.5/asyncio/base_events.py", line 330, in run_until_complete
    self.run_forever()
  File "/home/inso/.pyenv/versions/3.5.0/lib/python3.5/asyncio/base_events.py", line 301, in run_forever
    self._run_once()
  File "/home/inso/.pyenv/versions/3.5.0/lib/python3.5/asyncio/base_events.py", line 1162, in _run_once
    event_list = self._selector.select(timeout)
  File "/home/inso/.pyenv/versions/3.5.0/lib/python3.5/selectors.py", line 432, in select
    fd_event_list = self._epoll.poll(timeout, max_ev)
KeyboardInterrupt
Unclosed response
[...]

PR #2208, opened 2017-08-16
Fix ws close faq

mpaolini:
Update FAQ entry to leverage #754


docs/faq.rst
```diff
@@ -162,49 +162,43 @@ For example we have an application with two endpoints:
...
-tasks for a user in the :class:`aiohttp.web.Application` instance and
-``cancel()`` them in ``/logout_user`` handler::
+One simple solution is keeping a shared registry of websocket responses
+for a user in the :class:`aiohttp.web.Application` instance and
+call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in ``/logout_user`` handler::
 
     async def echo_handler(request):
 
-        ws = web.WebSocketResponse()
+        ws = aiohttp.web.WebSocketResponse()
```
asvetlov: Please keep `web.WebSocketResponse` name.
Canonical way for using aiohttp web is `from aiohttp import web`.

docs/faq.rst
```diff
@@ -162,49 +162,43 @@ For example we have an application with two endpoints:
...
         user_id = authenticate_user(request)
 
-        for task in request.app['websockets'][user_id]:
-            task.cancel()
+        ws_closers = [ws.close() for ws in request.app['websockets'][user_id] if not ws.closed]
 
-        # return response
-        ...
+        # Watch out, this will keep us from returing the response until all are closed
+        ws_closers and await asyncio.wait(ws_closers)
```
asvetlov: `aynsio.wait` doesn't raise exceptions, please use `asyncio.gather()`.

docs/faq.rst
```diff
@@ -162,49 +162,43 @@ For example we have an application with two endpoints:
...
+
+        return aiohttp.web.Response(text='OK')
```
asvetlov: Name

docs/faq.rst
```diff
@@ -162,49 +162,43 @@ For example we have an application with two endpoints:
...
 
     def main():
         loop = asyncio.get_event_loop()
         app = aiohttp.web.Application(loop=loop)
         app.router.add_route('GET', '/echo', echo_handler)
         app.router.add_route('POST', '/logout', logout_handler)
-        app['websockets'] = defaultdict(set)
+        app['websockets'] = defaultdict(deque)
```
asvetlov: Where `deque` comes from and why `set` is not enough?

mpaolini: WebSocketResponse is not hashable, that's why I used `collections.deque`

docs/faq.rst
```diff
@@ -162,50 +162,44 @@ For example we have an application with two endpoints:
...
+call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in ``/logout_user`` handler::
 
     async def echo_handler(request):
 
         ws = web.WebSocketResponse()
         user_id = authenticate_user(request)
         await ws.prepare(request)
-        request.app['websockets'][user_id].add(asyncio.Task.current_task())
-
+        request.app['websockets'][user_id][id(ws)] = ws
```
asvetlov: `id(ws)` is not reliable: after garbage collection there is a chance to create new `web.WebSocketResponse()` at the same memory address.

asvetlov:
Thanks!

Issue #434, opened 2015-07-09
Implementing auth handlers with `aiohttp.client`


alexdutton:
Is there a preferred way to hook in auth handlers to the HTTP client? If not, can we specify one, maybe with code changes to `ClientSession` to support it?

asvetlov:
I don't follow. What do you want to do with auth?

alexdutton:
Implement e.g. OAuth, Negotiate, Digest auth on the client side, instead of just supporting Basic.

More generally, have some way of hooking in error handlers for 4xx responses that can perform additional requests before returning a response. Could also be used to transparently back off on 420 and 429 responses.

alexdutton:
My thinking was putting something near the bottom of the `while True` loop in `ClientSession.request` to allow something to be hooked in to handle 4xx (or any other) status by manipulating the request and performing it again. Nowhere near sure on the details yet though.

alunduil:
Custom authentication mechanisms in requests would be awesome.  Right now it looks like I have to subclass the ClientRequest and overload update_auth to perform my custom authentication logic.  

If we simply didn't have the expectation that an auth was a BasicAuth but something that responded to an encode method (or something similar) that would be awesome.  I'd also really like it if the request itself could be available for the encoding process for mechanisms that want to have a signature of the request itself.

mattpepin:
Any progress on this one? Digest auth is often needed to interface with IoT devices/IP cameras.

fafhrd91:
PR would be awesome. Personally, I don't have time for this

fidiego:
Would this be a good place to mention I'd like to add support for AWS4Auth?

dadocsis:
I have a use case for ntlm auth

dadocsis:
Would it make sense to add event hooks similar to the requests package? 
http://docs.python-requests.org/en/master/user/advanced/#event-hooks
This could allow devs to register custom authentication methods. 

IE:

```python
class MyAuth:
      async def do_custom_auth(self, response):
            if r.status_code == 401:                  
                  # do custom

      def __call__(self, request):
            request.register_hook('response', do_custom_auth)
```

fidiego:
That would be very cool.

sky-code:
maybe just add `before_send` hook as last point where request can be changed so by using one hook we can inject auth headers before send request ?

asvetlov:
Public client request object is not exist (yet).

RobertoPrevato:
[...]

PR #2213, opened 2017-08-20
Added a digest authentication helper

### LINKED ISSUE #4939: support for digest authentication
I used aiohttp for async http requests,but It dose not support digest authentication.I think it is a basic feature for http authentication, I found some one has done some effort to implement the goal, as  follows https://github.com/aio-libs/aiohttp/pull/2213. can support this features officially?

feus4177:
## What do these changes do?

Added a digest authentication helper.

None

## Related issue number

Resolves #4939

## Checklist

I didn't go through the checklist. I just wanted to see if this is something you are interested in incorporating into the aiohttp package. To be able to use this DigestAuth helper like the aiohttp BasicAuth helper would take some restructuring so I decided to keep it as isolated as possible right now. Let me know what you think and I can make any changes as necessary.


asvetlov:
Sorry, I don't follow how to use the helper.
Full test coverage is required also.

feus4177:
Sorry, I meant to add this snippet as well.
```python

import aiohttp
import asyncio

client = aiohttp.ClientSession()
auth = aiohttp.auth.DigestAuth('usr', 'psswd', client)

async def fetch(url, **kwargs):
    return await auth.request('GET', url, **kwargs)


response = loop.run_until_complete(fetch('http://httpbin.org/digest-auth/auth/usr/psswd/MD5/never'))
```

feus4177:
Obviously, I can add tests, documentation, etc. Before I go through all that though, I just wanted to see if there is any interest in it.

asvetlov:
Well, idea is good but it should be proved by test suite.

oleksandr-kuzmenko:
This is a good idea!
But I think it would be better to add digest authentication support to standard client, something like:
```
                                     |
                                     v
async with aiohttp.ClientSession(digest_auth=...) as session:
    async with session.get('https://api.github.com/events') as resp:
        print(resp.status)
```

kxepal:
@alxpy 
Yeap, that's looks simpler for users. But it will overload ClientSession signature too much. Because digest and basic auths are not the only around. Having sort-of decorators / middlewares instead gives you more control: you can implement own auth method like this without touching aiohttp sources. And aiohttp developers wouldn't have to maintain such a jack of all trades object.

oleksandr-kuzmenko:
@kxepal yes, that makes sense, I agree

feus4177:
That was the kind of discussion I was looking for. It sounds like you guys would prefer to keep it separated like it currently is. I'll go ahead and add the tests and documentation and then update the pull request. Thanks.

aiohttp/helpers.py
```diff
@@ -216,6 +217,183 @@ def encode(self):
...
+        self.username = username
+        self.password = password
+        self.last_nonce = ''
+        self.nonce_count = 0
+        self.challenge = None
+        self.num_401 = 0
+        self.args = {}
+        self.session = session
+
+    @asyncio.coroutine
```
redixin: Nobody uses this syntax here anymore.

aiohttp/helpers.py
```diff
@@ -216,6 +217,183 @@ def encode(self):
...
+                method.upper(), url
+            )
+
+        response = yield from self.session.request(
+            method, url, headers=headers, **kwargs
+        )
+
+        # If the response is not in the 400 range, do not try digest
+        # authentication.
+        if not 400 <= response.status < 500:
```
redixin: Why not just always send auth header if user has decided to use basic auth?

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+        """
+        Takes the given response and tries digest-auth, if needed.
+        :rtype: ClientResponse
+        """
+        auth_header = response.headers.get('www-authenticate', '')
+
+        if 'digest' in auth_header.lower():
+            pattern = re.compile(r'digest ', flags=re.IGNORECASE)
+            self.challenge = parse_key_value_list(
+                pattern.sub('', auth_header, count=1)
```
kxepal: Sounds like you want to strip first word from string if it's match `digest` in lower case. No reason for regexp here.

docs/client_reference.rst
```diff
@@ -1395,6 +1395,118 @@ BasicAuth
...
+      (:class:`ClientResponse` or derived from).
+
+      :param str method: HTTP method
+
+      :param url: Requested URL, :class:`str` or :class:`~yarl.URL`
+
+      :param dict params: Parameters to be sent in the query
+                          string of the new request (optional)
+
+      :param data: Dictionary, bytes, or file-like object to
```
kxepal: `param dict|bytes|file data` btw.

docs/client_reference.rst
```diff
@@ -1395,6 +1395,118 @@ BasicAuth
...
+      :param bool read_until_eof: Read response until EOF if response
+                                  does not have Content-Length header.
+                                  ``True`` by default (optional).
+
+      :param loop: :ref:`event loop<asyncio-event-loop>`
+                   used for processing HTTP requests.
+                   If param is ``None``, :func:`asyncio.get_event_loop`
+                   is used for getting default event loop.
+
+         .. deprecated:: 2.0
```
kxepal: indention

docs/client_reference.rst
```diff
@@ -1395,6 +1395,118 @@ BasicAuth
...
+
+      :return ClientResponse: a :class:`client response <ClientResponse>` object.
```
kxepal: rtype?

feus4177:
So I updated the pull request to following the contributing guidelines (test, docs, etc.). Since it has been a while I also went ahead and merged the latest from aio-libs/aiohttp:master into the pull request. It looks like you guys switched over to the async/await syntax so I did that as well. However, half of all the tests are failing now. It looks like it's that way for all the tests though, so I'm assuming it isn't an issue. Let me know if there is anything else you want me to do.

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
+    def H(x):
+        if isinstance(x, str):
+            x = x.encode()
+        return hashlib.md5(x).hexdigest()
+
+    @asyncio.coroutine
+    def handler(request):
+        realm = 'test@example.com'
+        if 'Authorization' in request.headers:
+            pattern = re.compile(r'digest ', flags=re.IGNORECASE)
```
kxepal: You seems to repeat yourself here in tests.

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
```
kxepal: triple times.

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
+    def H(x):
+        if isinstance(x, str):
+            x = x.encode()
+        return hashlib.sha1(x).hexdigest()
+
+    @asyncio.coroutine
+    def handler(request):
+        realm = 'test@example.com'
+        if 'Authorization' in request.headers:
+            pattern = re.compile(r'digest ', flags=re.IGNORECASE)
```
kxepal: ok, multiple ones.

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
+    auth = helpers.DigestAuth(username, password, client)
+
+    resp = yield from auth.request('GET', '/?test=true')
+    assert 200 == resp.status
+
+
+@asyncio.coroutine
+def test_digest_auth_MD5_sess(loop, test_client):
+    opaque = binascii.hexlify(os.urandom(16))
+    username = 'usr'
```
kxepal: wh usr nd psswd whl username nd password wrwhr ls?

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
+    def H(x):
+        if isinstance(x, str):
+            x = x.encode()
+        return hashlib.md5(x).hexdigest()
+
+    @asyncio.coroutine
+    def handler(request):
+        realm = 'test@example.com'
+        if 'Authorization' in request.headers:
+            pattern = re.compile(r'digest ', flags=re.IGNORECASE)
```
kxepal: again

tests/test_helpers.py
```diff
@@ -494,6 +499,515 @@ def test_set_content_disposition_bad_param():
...
```
kxepal: ok, tests are the same as the regular code - please, show it some love in the name of maintenance.

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+                method.upper(), url
+            )
+
+        response = yield from self.session.request(
+            method, url, headers=headers, **kwargs
+        )
+
+        # If the response is not in the 400 range, do not try digest
+        # authentication.
+        if not 400 <= response.status < 500:
```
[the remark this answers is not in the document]
feus4177: It wouldn't try to do auth in the in the case of a 503. It will only try to perform digest authentication if the response status is from 400 up to but not including 500. Perhaps this would be clearer?
```python
        # Only try performing digest authentication if the response status is
        # from 400 to 500.
        if 400 <= response.status < 500:
            return await self._handle_401(response)
        
        return response
```
As for why we don't just check for 401, you can see the discussion here: https://github.com/requests/requests/issues/3772.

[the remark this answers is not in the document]
feus4177: You'll notice the behavior should be equivalent but perhaps this is more readable.

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+
+        # lambdas assume digest modules are imported at the top level
+        if algorithm == 'MD5' or algorithm == 'MD5-SESS':
+            hash_fn = hashlib.md5
+        elif algorithm == 'SHA':
+            hash_fn = hashlib.sha1
+        else:
+            return ''
+
+        def H(x):
```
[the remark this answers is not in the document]
feus4177: A lot of the nomenclature is pulled directly from the [RFC](https://tools.ietf.org/html/rfc2617) (e.g. H, KD, HA1, HA2, cnonce, ncvalue, etc.)

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+            return hash_fn(x.encode()).hexdigest()
+
+        def KD(s, d):
+            return H('%s:%s' % (s, d))
+
+        parsed = urlparse(url)
```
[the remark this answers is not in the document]
feus4177: I was unfamiliar with yarl; I will switch it over.

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+        A1 = '%s:%s:%s' % (self.username, realm, self.password)
+        A2 = '%s:%s' % (method, path)
+
+        HA1 = H(A1)
+        HA2 = H(A2)
+
+        if nonce == self.last_nonce:
+            self.nonce_count += 1
+        else:
+            self.nonce_count = 1
```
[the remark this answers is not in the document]
feus4177: If you re-use the instance, the `nonce_count` count can grow without bound. The nonce count is important in preventing replay attacks. There probably aren't too many cases where the nonce will need to reset but perhaps in the event of some failed messages.

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+
+        if not qop:
+            respdig = KD(HA1, '%s:%s' % (nonce, HA2))
+        elif qop == 'auth' or 'auth' in qop.split(','):
+            noncebit = '%s:%s:%s:%s:%s' % (
+                nonce, ncvalue, cnonce, 'auth', HA2
+            )
+            respdig = KD(HA1, noncebit)
+        else:
+            raise client_exceptions.ClientError(
```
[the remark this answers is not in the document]
feus4177: There would have to be some duplication of logic around `qop` and I didn't think that the performance was that big of an issue. I can move it though.

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+                'Unsupported qop value: %s' % qop
+            )
+
+        base = ', '.join([
+            'username="%s"' % self.username,
+            'realm="%s"' % realm,
+            'nonce="%s"' % nonce,
+            'uri="%s"' % path,
```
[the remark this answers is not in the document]
feus4177: So I believe the header is supposed to follow the rules outlined in [RFC 2616 section 2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html). I couldn't figure out how that specification intended for quotes to be escaped. I did a little more searching and this [SO post](https://stackoverflow.com/questions/7886782/what-is-the-exact-syntax-and-semantics-of-a-quoted-string-in-the-http1-1-rfc2616) was the only other thing I found and it wasn't any help.

I have been using [httpbin](https://httpbin.org/digest-auth/auth/user/password/MD5/never) to do some testing. I opened up the developer tools in chrome and tried going to the httpbin with a username and password that contained a space and a double quote. Looking at the request headers, Chrome seems to perform a URL encoding (i.e. using %20 and %22). httpbin didn't seem to handle it correctly. I also tried doing the same thing with the python requests library and it didn't work either:
```python
import requests
from requests.auth import HTTPDigestAuth
auth = HTTPDigestAuth('us er', 'pass"word')
url = 'http://httpbin.org/digest-auth/auth/us er/pass"word/MD5/never'
requests.get(url, auth=auth)
```

My guess is that the URL encoding is probably the best way to go. Also, I added `parse_key_value_list` in helpers.py to parse the key-value pairs that are coming in the response header but it probably isn't very complete either. Do you guys already have a tool to parse these or should I try to improve my implementation?

aiohttp/helpers.py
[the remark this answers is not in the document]
feus4177: I should clarify, if you re-use the instance of `DigestAuth` for performing multiple requests.

tests/test_helpers.py
feus4177: I was trying to demonstrate that the helper isn't just hardcoded to work with one username and password. Although, this really doesn't challenge it at all. Once you decide what you want to do about the string escaping I can change this to include something more difficult like spaces and quotes.

asvetlov:
Tests still are red :(

feus4177:
When I merged your master in I made the mistake of leaving the test_dummy_cookie_jar test in when I was fixing the merge conflicts. Removing it appears to have fixed the tests.

asvetlov (changes requested):

asvetlov:
Sorry, wrote comments two days ago but forgot to press "Submit" button.

asvetlov:
The code is written good and well tested, but let's discuss the usage design.
Now the API is:
```
auth = aiohttp.DigestAuth('user', 'password', client)
resp = await auth.request('get', '/')
```
The API is not very handy:
1. `async with auth.request('get', url) as resp:` is not supported
2. `await auth.get(url)` is not supported
3 `async with client.get(url, auth=aiohttp.DigestAuth(...)` is not supported

We could fix 1. and 2. easy but it doesn't make a value.
What is really desirable is accepting `DigestAuth` as `auth` parameter -- like `requests` does [1]
Later we can add support for OAuth and other authorization protocols.
The approach requires inviting an abstract interface for auth classes and supporting it by both `BasicAuth` and `DigestAuth`.

@feus4177 what do you think about?

[1]: http://docs.python-requests.org/en/master/user/authentication/#digest-authentication

feus4177:
The difficult part is that for digest auth there is potentially a second request that needs to be launched, and the whole point of the helper is so that you don't have to worry about handling that additional request. If you want to be able to do use the `client.get(url, auth=aiohttp.DigestAuth...)` syntax, the easiest way to go about it would be to wrap the `aiohttp.ClientSession._request` method. This way, the digest auth can make multiple calls to the underlying `_request` method.

This is kind of what I had been asking about earlier though. @alxpy had mentioned doing `aiohttp.ClientSession(digest_auth=...)` but I thought it sounded like you guys didn't like that. I think for the users `client.get(url, auth=aiohttp.DigestAuth...)` is the best option but I think it will be a little messier to implement. At the end of the day though, I'm fine with however you guys want to do it.

dadocsis:
I am wrapping up pr to implement custom auth handlers similar to the requests package. I am also adding a response event hook where the auth handler could be registered to handle a 401. So in this case the usage could be:

`
auth = aiohttp.DigestAuth('user', 'password', client)
session = aiohttp.ClientSession(auth=auth)
resp = await auth.request('get', '/')
`

Does this make sense?

dadocsis:
See issue 434
https://github.com/aio-libs/aiohttp/issues/434

feus4177:
In your case, wouldn't it be `resp = await session.request('get', '/')`?

I referenced the requests library pretty heavily when I was creating this feature and they use event hooks so I think it would fit in well. In my mind we would want event hooks directly before and after the request, and then some mechanism to retry the request. The before hook would help in the case of digest auth because then we could implement the auth-int quality of protection which you need the request body entity to perform.

dadocsis:
Correct. 
Right now I am just adding a response hook. Looking at the Digest Auth class I do not see where it needs the "before" hook. The response hook would pass the original request to the handler (I mistyped in my original comment about passing in a response) It makes more sense to pass in the request because from the request you will have access to the response and client session. The "_handle_401" could be registered as the handler and should fit in nicely.

feus4177:
The before hook would be needed to add support for the auth-int quality of protection. Right now, if they server specifies auth-int for qop, we raise an error. But if we were to add support for it, you would need the before hook.

aiohttp/helpers.py
```diff
@@ -103,6 +104,177 @@ def encode(self):
...
+        self.challenge = previous.get('challenge')
+        self.args = {}
+        self.session = session
+
+    async def request(self, method, url, *, headers=None, **kwargs):
+        if headers is None:
+            headers = {}
+
+        # Save the args so we can re-run the request
+        self.args = {
```
WGH-, line 151: Why do you store last request as a class attribute? Won't it cause everything to break when `DigestAuth` instance is used by multiple coroutines simultaneously?

feus4177: Yeah, that's a good point. We can just store it as a local variable and then pass it to `_handle_401`.

WGH-: I think this applies to `self.challenge` as well.

aiohttp/helpers.py
```diff
@@ -103,6 +104,177 @@ def encode(self):
...
+                method.upper(), url
+            )
+
+        response = await self.session.request(
+            method, url, headers=headers, **kwargs
+        )
+
+        # Only try performing digest authentication if the response status is
+        # from 400 to 500.
+        if 400 <= response.status < 500:
```
WGH-, line 169: In one of my tests, when username-password pair was incorrect, this led to infinite recursion. I think the fix is to call `_handle_401` only if `self.challenge` is `None`.

RouquinBlanc:
Hi guys,

What do you think the status of this pull request is and do you have a rough idea if it is going to be ready any time soon-ish? It seems quite close from an external eye...

At the moment AFAIK there is no pure asyncio framework supporting digest authentication, you have to either go back to `tornado` or to give up on asyncio and look at `requests`...

Thanks!

DavHau:
> At the moment AFAIK there is no pure asyncio framework supporting digest authentication, you have to either go back to `tornado` or to give up on asyncio and look at `requests`...

@RouquinBlanc 
This is not true. There is:
Pulsar: https://github.com/quantmind/pulsar (file upload in combination with DigestAuth is broken at the time of writing)
or
yieldfromrequests: https://github.com/rdbhost/yieldfromrequests (not updated since 4 years, but works for me. Only usable with old generator style async syntax. write yourself a wrapper ;) )

I'm also really looking forward to have DigestAuth available in aiohttp since it is my favorite async http lib.

sonicblind:
@asvetlov 
I found a small bug related to Digest Auth Header in helpers.py.

The Digest header can include spaces in realm name, for example:
`Digest realm="iMC RESTful Web Services", qop="auth", nonce="MTU0....=="`

My suggestion as per below:
```
def parse_key_value_list(header):
    return {
        key: value for key, value in
#       [parse_pair(header_pair) for header_pair in header.split(' ')]
        [parse_pair(header_pair) for header_pair in header.split(', ')]
    }
```

But this is probably also not bulletproof as the realm could include also comma.
So maybe `header.split('", ')` would work even better.

And if the above make sense, then the "remove trailing comma" section in parse_pair(pair) should also be updated accordingly:

```
def parse_pair(pair):
    key, value = pair.split('=', 1)

    # If it has a trailing comma, remove it.
#   if value[-1] == ',':
#       value = value[:-1]

    # If it is quoted, then remove them.
    if value[0] == value[-1] == '"':
        value = value[1:-1]
    
    return key, value
```

asvetlov:
@sonicblind thanks for the comment.
The PR doesn't go to merge but I'd like to keep it open as a reference for eventual new implementation built from scratch.

bdraco:
Thanks for working on this.   @feus4177 @asvetlov Is there a shot that support for digest auth is forthcoming any time soon?

kiknaio:
Still nothing new. I think it would be amazing if you complete this feature request.

docs/client_reference.rst
```diff
@@ -1397,6 +1397,118 @@ BasicAuth
       :return: encoded authentication data, :class:`str`.
 
 
+DigestAuth
+^^^^^^^^^^
+
+.. class:: DigestAuth(login, password', session)
+
+   HTTP digest authentication helper. Unlike :class:`DigestAuth`, this helper
```
svisser, line 1405: Unlike `BasicAuth` ?

aiohttp/helpers.py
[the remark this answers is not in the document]
rroller: parse_key_value_list should at the very least, split on the comma and not on the space... and then trim the results.

aiohttp/helpers.py
```diff
@@ -103,6 +104,177 @@ def encode(self):
...
+    if value[0] == value[-1] == '"':
+        value = value[1:-1]
+
+    return key, value
+
+
+def parse_key_value_list(header):
+    return {
+        key: value for key, value in
+        [parse_pair(header_pair) for header_pair in header.split(' ')]
```
ascott18, line 124: Splitting on space does not work because values can be quoted strings that contain spaces.

E.g. `'realm="Login to AMC072A731527963121B", qop="auth", nonce="839506863", opaque="f5a15b51387c97add984d4d948213e563fe3eede"'`

skewty:
I take it this died and never got merged.  Any way to bring this back to life?  What was holding up the merge?

skewty:
I still don't see `aiohttp.helpers.DigestAuth` so it would seem this never got merged. Can we get this merged?

aiohttp/helpers.py
```diff
@@ -182,6 +184,194 @@ def encode(self):
...
+
+    @asyncio.coroutine
+    def _handle_401(self, response):
+        """
+        Takes the given response and tries digest-auth, if needed.
+        :rtype: ClientResponse
+        """
+        auth_header = response.headers.get('www-authenticate', '')
+
+        if 'digest' in auth_header.lower():
```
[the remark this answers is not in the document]
skewty: perhaps `'digest '` would be a better check (with a space at the end)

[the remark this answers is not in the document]
skewty: and do want want to make sure it starts with digest?  what if somebody's login name is 'digest'?

flashnuke:
This would be of tremendous value... any chance of merging?

webknjaz:
Hey @feus4177, are you planning to resume this effort? It looks like a lot of reviews accumulated over the years. Plus, there's merge conflicts that would have to be addressed.

bdraco:
It would be great to get this one implemented as its one of the key drivers for Home Assistant devs choosing alternate (usually slower) libraries because the device they are integrating needs digest auth.

feus4177:
I have no intention of continuing this work. I tried to get feedback on the overall approach of the work before I invested a bunch of time and effort into finalizing the details: https://github.com/aio-libs/aiohttp/pull/2213#issuecomment-323602760. After some small feedback I went through and cleaned everything up, made sure all the tests were passing and everything. Then I was hit with this: https://github.com/aio-libs/aiohttp/pull/2213#issuecomment-346964207. I didn't enjoy jumping through all the hoops just to then revisit the initial feedback I was trying to solicit. I have no intention of going through all that work again knowing that it will all be in vain if the owners decide that they still don't like the architecture of the solution.

bdraco:
> I have no intention of continuing this work

The current authentication interface design needs to accommodate additional future methods, and a redesign is required. Unfortunately, this conclusion was realized late in the review process, and I'm sure you wanted to do something other than sign up to do that redesign. I can understand how that is frustrating.

Thank you for your contributions thus far and for following up on this PR.

feus4177:
@bdraco, I appreciate your understanding. I also think a redesign is completely reasonable. I just think it's up to the maintainers at this point.

Dreamsorcerer:
> The current authentication interface design needs to accommodate additional future methods, and a redesign is required.

I think there's a discussion elsewhere, but I noted that `auth` in other libraries often allows arbitrary functions to run and modify the request. I suggested we could just call this a middleware, and make it work very similarly to the server-side middlewares. Then people can use it for various auth methods or whatever else they might want to do.

bdraco:
I started thinking about this in https://github.com/aio-libs/aiohttp/pull/9732

PR #2215, opened 2017-08-21
Without this clear, we will get exception in unittest: RuntimeError: …

patryk4815:
Without this clear, we will get exception in unittest: RuntimeError: web.Application instance initialized with different loop

## Checklist


asvetlov:
Please provide a test for reproducing the error.
I suspect it's related to cases where the loop is not explicitly passed to app but like to have an explicit test case for it.

patryk4815:
I had mistake in my unittest, i used global variable to store "app" instance, this was wrong.

PR #2221, opened 2017-08-24
Add deferredssl

machina-widmo:
Wraps ssl in a deferred wrapper.
aiohttp does not require SSL to function. The codepaths involved with
SSL will only be hit upon SSL usage. Using a wrapper allows installs
that don't use SSL to function (without erroring on import) and still
allows full function on SSL enabled systems.

## Checklist

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -93,6 +93,7 @@ Jeroen van der Heijden
 Jesus Cea
 Jinkyu Yi
 Joel Watts
+Jon Nabozny
 Joongi Kim
 Josep Cugat
 Julia Tsemusheva
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -3,6 +3,12 @@ import functools
 import sys
 import traceback
 import warnings
+
+try:
+    import ssl
+except ImportError:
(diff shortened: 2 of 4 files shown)


asvetlov:
What is your use case?
I pretty sure all Python interpreters are shipped with ssl support nowdays.

ghost:
My use case is a bit contrived. I was trying to use this with buildozer / python-for-android, which by default don't include ssl (but can).

Just figured that in case SSL isn't available (although, like you pointed out, it almost always is), seems better to only fail when it's required.

ghost:
The only reason I opted for a module was to be a bit forward looking (and DRY).
But, whatever fits best with the project makes most sense.

asvetlov:
@machina-widmo would you rewrite a PR for using approach proposed by me?
I really don't like `deferredsll.py`.

ghost:
I went back and fixed up only the actual "deployable" runtime code.
Anywhere else, I don't think it's necessary as they are either examples, or tests.

May be worth doing conditional imports for the tests so at least the tests could run and fail on usage.

asvetlov:
Merged to master from command line.

Thanks for contribution.

Issue #2223, opened 2017-08-24
Pytest: Allow coroutine fixtures


k4nar:
[pytest-asyncio](https://github.com/pytest-dev/pytest-asyncio) provides a very nice feature, which is to allow coroutine fixtures. For example it makes possible to create a fixture making asynchronous DB requests.

Unfortunately pytest-asyncio and pytest-aiohttp seems to no longer be compatible. So far this is the only feature that I'm missing, so it would be awesome if it were added to pytest-aiohttp.

Here is the implementation in pytest-asyncio:
https://github.com/pytest-dev/pytest-asyncio/blob/92b34e5e9a14c45062d6ce2e2686ed8b11c7c03e/pytest_asyncio/plugin.py#L52-L134

I've tried to adapt it to pytest-aiohttp but unsuccessfully so far.

samuelcolvin:
It's very simple right now to create fixtures which execute asynchronous code, for example:

```python
@pytest.yield_fixture
def redis(loop):
    async def _redis():
        redis = await create_redis(('localhost', 6379))
        await redis.flushdb()
        return redis

    redis = loop.run_until_complete(_redis())

    yield redis

    async def _close():
        redis.close()
        await redis.wait_closed()
    loop.run_until_complete(_close())
```

samuelcolvin:
Personally I think the way aiohttp executes async tests is good and I think @asvetlov built it this way specifically because he didn't agree with how pytest-asyncio was setup.

k4nar:
Yes, I know that it's possible to execute asynchronous code in a fixture as in your example, however I find that it's a bit too much boilerplate.

Actually I've managed to adapt pytest-asyncio's hack to pytest-aiohttp, I'll open a PR shortly.

samuelcolvin:
so what happens with fixtures which are session scoped? Is the same loop running thoughout all tests?

From my understanding, accepting this kind of fixture in aiohttp without hacks would require a big change.

k4nar:
It only works for function fixtures as it implies the usage of the loop fixture. So far what I have is not intrusive, it's just a pytest hook (mostly like the one in pytest-asyncio). I'm currently writing tests for it.

k4nar:
I think the PR is ready (#2226). The tests pass and I'm using it with success in my project.

asvetlov:
Fixed by #2226

PR #2226, opened 2017-08-25
Pytest async fixtures

k4nar:
## What do these changes do?

With this PR, pytest fixtures can be coroutines. For example, in order to create a client, you can use:

```python
@pytest.fixture
async def cli(test_client):
    return await test_client(setup_app)
```

This is not a killer feature as you could already do the same thing using `loop.run_until_complete`, but I find it more idiomatic.

The main gotcha is that in order for an async fixture to work, it must use the `loop` fixture, or be called in a test using it. The previous example works because `test_client` uses the `loop` fixture.
I don't think this is a big issue because I can't think of any case where one would want an asynchronous fixture without using the loop.

None.

## Related issue number

#2223

## Checklist

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -188,6 +188,7 @@ Will McGugan
 Willem de Groot
 Wilson Ong
 Yannick Koechlin
+Yannick Péroux
 Yegor Roganov
 Young-Ho Cha
 Yuriy Shatrov
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -243,7 +243,7 @@ def isasyncgenfunction(obj):
         return inspect.isasyncgenfunction(obj)
     return False
 
-  
+
 def parse_mimetype(mimetype):
     """Parses a MIME type into its components.
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -6,6 +6,7 @@ import binascii
 import cgi
 import datetime
 import functools
+import inspect
 import os
 import re
 import sys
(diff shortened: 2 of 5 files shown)


asvetlov:
Your PR is awesome!
I really love it.

Would you add add couple words to http://aiohttp.readthedocs.io/en/stable/testing.html#pytest for introducing the feature?

k4nar:
Yes I'll had some documentation soon. I wanted to have some feedback first, but now we're good :) .

mikenerone:
Just a note based on a glance at this (so apologies if I'm missing some important difference), but this seems like reinventing [pytest-asyncio](https://pypi.python.org/pypi/pytest-asyncio), which provides a more robust feature set. aiohttp already has deps on some other pytest plugins, and this seems like a natural one to add to that set.

asvetlov:
Thanks!

samuelcolvin:
Despite my initial scepticism, this looks like it should be really useful.

Thanks a @k4nar, you'll save me writing `loop.run_until_complete` a few hundred times a year.

mikenerone:
@asvetlov @k4nar I guess my comment went under the radar, as there was no response. I'll mention it just this one more time in case it was just missed: I haven't had to write `loop.run_until_complete()` for quite some time, as I've been using asynchronous pytest fixtures via [pytest-asyncio](https://pypi.python.org/pypi/pytest-asyncio), which provides the same functionality as these changes and more. This seems like a concern that's separate from aiohttp, IMO.

asvetlov:
pytest-asyncio is not fully compatible with pytest-aiohttp (while last pytest-asyncio commits has smoothed some corners).

The crucial difference is disabling global event loop by default in pytest-aiohttp.
Another difference is `event_loop` vs `loop` fixture name. From my perspective `loop` is more native: it's used widely by asyncio itself, no `event_loop` variable in asyncio code.

After `asyncio.get_event_loop()` fix in Python 3.5.3+ default event loop issue becomes negligible but it's still very important until aiohttp will drop older Python versions support (I hope we'll do it it after next release).

PR #2229, opened 2017-08-27
Speed up Signals when there are no receivers.

pfreixes:
## What do these changes do?

Reduce the footprint of the `send()` method by a 10% when there are
no receivers connected.

Before
```bash
$ python ../test_aiohttp_signals.py
Signals per second: 2296186.5549501474
```

After
```bash
$ python ../test_aiohttp_signals.py
Signals per second: 2759791.7854124783
```

Script used [1]

[1] https://gist.github.com/pfreixes/baea4ecc7b5380d534d60a8c1e8da32f 

No

## Checklist

### DIFF
--- a/aiohttp/_frozenlist.pyx
+++ b/aiohttp/_frozenlist.pyx
@@ -4,9 +4,11 @@ from collections.abc import MutableSequence
 cdef class FrozenList:
 
     cdef readonly bint frozen
+    cdef readonly bint _receivers 
     cdef list _items
 
     def __init__(self, items=None):
+        self._receivers = False
         self.frozen = False
         if items is not None:
             items = list(items)
@@ -17,6 +19,9 @@ cdef class FrozenList:
     cdef object _check_frozen(self):
         if self.frozen:
             raise RuntimeError("Cannot modify frozen list.")
+
+    cdef object _fast_len(self):
+        return len(self._items)
 
     def freeze(self):
         self.frozen = True
@@ -33,7 +38,7 @@ cdef class FrozenList:
         del self._items[index]
 
     def __len__(self):
-        return self._items.__len__()
+        return self._fast_len()
 
     def __iter__(self):
         return self._items.__iter__()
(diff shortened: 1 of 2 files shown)


kxepal:
Nice! Since this is a FrozenList, his length is constant. What if you compute length once at `__init__` and return that static value from `__len__` ?

Update: I always forgot that it's not frozen like frozenset. Ok, what if do the same - compute length once on freeze while keep it dynamic while it's not frozen?

pfreixes:
From what I've seen the freeze is done explicitly by the developer at some moment, would be needed calculate it at the moment of the `freeze()` call. But, seeing the `Signal` implementation seems that this characteristic is not really used to determine if a Signal can be used or not, looks like that the freeze only avoids modify the items saved from that moment. Why ? no fucking idea :(

aiohttp/_frozenlist.pyx
```diff
@@ -18,6 +18,9 @@ cdef class FrozenList:
         if self.frozen:
             raise RuntimeError("Cannot modify frozen list.")
 
+    cdef object _fast_len(self):
```
asvetlov: The function is used only once, please inline it -- the speed should be the same.

pfreixes: done

asvetlov:
You've raised two separate questions:
1. Why raising non-frozen signals are allowed?
2. Does length calculation on `.freeze()` makes sense?

1. It's implementation detail. I just forgot to add check for `.frozen` in `Signal.send()`.
I see no harm in behavior changing, we could start from raising a warning.
2. Technically we could add `cdef int _len` attribute but I pretty sure `PyList_Size()` is pretty fast too. Extra field adds unnecessary complexity.

pfreixes:
Yes, the cost of `PyList_Size()` vs doing explicitly with an attribute is negligible. And makes the code more complicated.

asvetlov:
Thanks!

Issue #2225, opened 2017-08-25
Middleware factory is run for every single request


butla:
## Long story short
Maybe that's just my unwaranted assumption, but I thought that a middleware factory should run once for every handler, creating the wrapped handlers. Turns out, it's run for every request. This seems to be terribly inefficient.

Is this an expected behaviour, something that I'm seeing for some strange reasons, or a regression in the library?

## Expected behaviour
Run each middleware factory for each handler at one point, at the start of application's lifecycle. Let the new handlers handle requests from that point without being constantly recreated.
   
## Actual behaviour
Middleware factory is run for every request to a handler.

## Steps to reproduce
What I'm creating is a middleware that will parse JWT tokens. I need a mechanism to exculde this parsing on some handlers, since some will be accesses without authorization. I didn't use auihttp_security, since it imposes structure that I don't need (I don't have sessions, for example). I may open-source this code later.

The important bits of my code (with some alterations):

```python
# authorization.py

from functools import wraps

from aiohttp import web
import jwt

NON_SECURED_HANDLERS_KEY = "set of endpoints that don't expect a token"
USER_ID_KEY = 'user ID given by token middleware'

async def token_reading_middleware(app, handler):
    """Reads the token from the authorization header and determines the user ID based on the
    scope inside it.

    Raises 401 if there's something wrong with the header or it's missing.
    
    This check can be disabled by adding the function to the set of non secured handlers
    (on the application object, under `NON_SECURED_HANDLERS_KEY`).
    """
    # assume that APP_CONFIG is imported
    app_secret = app[APP_CONFIG].app_secret

    @wraps(handler)
    async def middleware_handler(request):
        """Adds the user ID parsed from the authorization header to request or raises 401 error.
        """
        # let's say that this function is implemented somewhere and returns 401 is something's wrong with the header
        user_id = _get_user_id_from_auth_header(request, app_secret)
        request[USER_ID_KEY] = user_id
        return await handler(request)

    # this is to tell us that the factory code has been run
    print('XXX', flush=True)
    
    # let's say that this function determines if the handler is esxccepted from security or not
    if _is_except_from_security(handler, app):
        return handler
    return middleware_handler

# app.py
from aiohttp import web

async def _health_check(_):
    return web.HTTPNoContent()


loop =  asyncio.get_event_loop()
app = web.Application(
    loop=loop,
    middlewares=[authorization.token_reading_middleware])
app.router.add_route('GET','/', _health_check)

web.run_app(
    app,
[...]

PR #2230, opened 2017-08-29
Document that middleware factories are run for every request.

butla:
I guess the title says it all. Related to #2225

### DIFF
--- a/docs/web.rst
+++ b/docs/web.rst
@@ -904,6 +904,11 @@ the **next** *middleware factory*. The last *middleware factory* always receives
 the :ref:`request handler <aiohttp-web-handler>` selected by the router itself
 (by :meth:`UrlDispatcher.resolve`).
 
+.. note::
+
+   Both the outer *middleware_factory* coroutine and the inner
+   *middleware_handler* coroutine are called for every request handled.
+
 *Middleware factories* should return a new handler that has the same signature
 as a :ref:`request handler <aiohttp-web-handler>`. That is, it should accept a
 single :class:`Request` instance and return a :class:`Response`, or raise an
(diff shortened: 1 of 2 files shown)


butla:
Wait a sec, I should have based that on the released tag... Gotta change the branch.

docs/web.rst
```diff
@@ -904,6 +904,10 @@ the **next** *middleware factory*. The last *middleware factory* always receives
 the :ref:`request handler <aiohttp-web-handler>` selected by the router itself
 (by :meth:`UrlDispatcher.resolve`).
 
+.. note::
+
+   The chain of factories is run for every handled request.
```
samuelcolvin: I think someone could still read this and think it's just reiterating that each middleware is called for all views.

Perhaps we could be more explicit, something like 

"Both the outer `middleware_factory` coroutine and the inner `middleware_handler` coroutine are called for every request handled."

butla:
@samuelcolvin Good idea, changed according to your suggestion.

samuelcolvin:
done. Thanks @butla.

@asvetlov depending on the outcome of #2225 we might need to add this to master too, do we have somewhere to record that so we don't forget?

asvetlov:
I've merged 2.2 into master.

asvetlov:
@samuelcolvin in aiohttp we do merging release branches like `2.2` into master after every PR like this.
Well, the rule was never pronounced but I always did it.
aiohttp committers, please do it too.
After switching changenotes to towncrier it is not painful job, we have no merge conflicts in CHANGES.rst anymore.

samuelcolvin:
humm, I'm not seeing this commit or the change on master?

asvetlov:
Missed to make `git push` before sending the message.

PR #2252, opened 2017-09-07
simplifying middleware

samuelcolvin:
## What do these changes do?

This simplifies middleware so middlewares are simple coroutines rather than coroutine factories returning coroutines.

Simply:

```python
async def my_middleware(request, handler):
    return await handler(request)
```

Old style middleware:

```python
async def middleware_factory(app, handler):
    async def middleware_handler(request):
        return await handler(request)
    return middleware_handler
```

Continues to work with a depreciation warning.

They can simplify their middleware if they wish

## Rationale

1. The new style is cleaner, less code, quicker to read, quicker to write.
2. Old style leaves a big bad bear trap for developers to fall into: assuming the outer coroutine is only called once and using it for setup code. 

## Related issue number

#2225

## Checklist


samuelcolvin:
@asvetlov 

> Then we'll open a kind of voting for collecting feedback from other devs.

How do you propose to vote? Using github reactions?

I've delayed updating docs until we agree this is required. Also if we are changing to this, perhaps we should give some kind of indication how long old-style middleware will continue to work.

asvetlov:
Hmm, in your proposal all middlewares are either new-styled or old ones.
It's impossible to combine new-style middleware from some third-party library with old-fashioned one.
Therefore it looks like very hard backward compatibility breaker.

samuelcolvin:
ok, I'll extend it to check every middleware.

samuelcolvin:
ok, now old and new middleware can be mixed.

aiohttp/web.py
```diff
@@ -272,6 +274,20 @@ def _make_request(self, message, payload, protocol, writer, task,
             self._loop,
             client_max_size=self._client_max_size)
 
+    def _prepare_middleware(self):
+        for m in reversed(self._middlewares):
+            sig = inspect.signature(m)
```
asvetlov: I don't like signature inspecting. The check is too weak.
In my mind we should mark a new style middleware very explicitly.
Let's wrap it by `@newstylemiddleware` decorator.
The decorator will add `__middleware_version__` attr to decorated object with value `1`.

@samuelcolvin what do you think?

samuelcolvin: Seems ugly, but I agree the signature approach is brittle.

I'll add it unless anyone can think of a third, less cleaner approach?

samuelcolvin: instead of the decorator we could use a type annotation for new style middleware? So if you want to use new style middleware you have to use 

```python
async def my_middleware(request, app) -> Request:
    ...
```

After all, this is what type annotations are for.

samuelcolvin: We could then add a depeciation warning in future if middleware doesn't include a type annotation declaring its return type.

asvetlov: Yep, type annotations is even better than decorator!

asvetlov: Sure, your snippet is incorrect but I've gotten your idea

asvetlov: Hmm, after careful thinking I have objections for type annotations usage.
It's intended for static analysis, processing types information in runtime doesn't sound like good idea.
Moreover supporting https://www.python.org/dev/peps/pep-0563/ is complicated, extracting types information from `.pyi` files and other sources opens a can of worms.

For sake of performance we need checking for middleware version on application freeze stage and storing a flag somewhere anyway.

kxepal: How about to not change signature so dramatically? If function takes two arguments - ok, that's old style. If three - new one. Application as a standalone parameter is very handy in middlewares in anyway.

samuelcolvin: The intentions for type hinting are deliberately vague, I would say the zen was "use it for whatever you want". For example I use it a lot in [pydantic](https://github.com/samuelcolvin/pydantic) and it works well.

O don't think there's anything complicated with extracting information etc, just:

```python
 class Foo:
     pass
 
 async def foo(a, b) -> Foo:
     pass
 
 foo.__annotations__
 {'return': __main__.Foo}
```

@kxepal extra argument out would be another solution, but I think it's kind of unclear and prone to confuse people.

asvetlov: Check for return value type only is too weak: both new and old style middlewares return a response:
```def old_style(app: Application, handler: Awaitable) -> StreamResponse```
and
```def new_style(request: Request, handler: Awaitable) -> StreamResponse```
check for arguments count is also not strict:
```def new_style(self, request: Request, handler: Awaitable) -> StreamResponse```

**Comment by samuelcolvin** [comment]
I think, that's wrong. Old middleware returns a coroutine. Or am I missing
something?

File: aiohttp/web.py (hunk 1 above)
Comment (kxepal) [reply aiohttp/web.py]: @asvetlov the self is not a problem if you inspect class instance method rather than class method itself.

Comment (kxepal) [reply aiohttp/web.py]: Ok, how about different way: currently middleware definition is basically hand-written decorator. What if we introduce own middleware decorator as @asvetlov proposed early, but no need to do any private attributes. It's ours decorator which demands some protocol and expects that decorated function will accept two specific arguments. Like what `signal` module does.

Comment (asvetlov) [reply aiohttp/web.py]: Yep, it's true.
Another important point is supporting default values:
```def new_style(app, request, handler, param=None)```.
We cannot protect it, users will make very wild signatures.

Comment (samuelcolvin) [reply aiohttp/web.py]: (I replied via email and my message came in the wrong place, so I'ved deleted it and am replying here).

@asvetlov I think you have the signature confused.

Old style signature:
```python
def old_style(app: Application, handler: Awaitable) -> Awaitable
    ...
```

New style signature:
```python
def new_style(app: Application, handler: Awaitable) -> StreamResponse
    ...
```

They're completely different. The return type alone is enough to distinguish the two styles of middleware.

(update, corrected signature code)

kxepal:
```
>>> from aiohttp import StreamResponse
ImportError: cannot import name 'StreamResponse'
```

From where I should import it? Why I should import it if I don't use it? I think annotations shouldn't be made mandatory. Python itself doesn't requires them to be mandatory to achieve any profits, so we cannot demand them from users.

samuelcolvin: why not?

It's there as a language feature, we can use it if we want.

We can implement the decorator solution too in case some people can't (eg. 3.4) or don't want to use type annotations.

asvetlov: @samuelcolvin yes, new and old styles are different by return type, now I see it.

The language feature's usage is static type checking only. Technically you could use it for different purposes but this is discouraged.
Strings as type annotations are supported now by `typing` and will become mandatory in next Pythons. Stub files make situation even more complex.
Let's use decorator approach.
@kxepal could you explain your idea?

samuelcolvin: Genuine question: "but this is discouraged", can you point me to where that's stated? I wasn't aware of it.

I really don't see that stub files have have relevance here, AFAIK they're particular to mypy. I was just suggesting using `function.__annotations__`, same as another application might use `function.__name__`. Stub files and strings as type annotations have no bearing on this.

However, I see a real point to not using type annotations: developers assume they're either advisory or for static type analysis; therefore it doesn't make much sense to use them for business logic in one place in the code.

I'll implement the decorator solution now.

asvetlov: @samuelcolvin I don't remember exact blogpost or article but Guido told many times about his vision for annotations: previously it was not utilized by python stdlib etc., people used to apply them on they own.
But now annotations are for type checking only. Other approaches are still supported to not break existing code (and will be supported at least for decade) though.

kxepal: @asvetlov 

```
def old_style_middleware(app, handler):
  async def middleware_handler(request):
    ...
  return middleware_handler

@middleware
def new_style_middleware(request, handler):
  ...

```

where `middleware` is  something like
```
def middleware(func):
  def old_style_middleware(app, handler):
    async def middleware_handler(request):
        return await func(request, handler)
    return middleware_handler
  return old_style_middleware
```

Or whatever it should be to create new style arguments. But, since aiohttp implements `middleware` decorator, it may dictate what it should decorate and what to pass for decorated function.

Thought, I still not quite follow what's wrong with current approach. Yes, it's a bit verbose, but quite simple and explicit.

aiohttp/web.py
```diff
@@ -272,6 +273,18 @@ def _make_request(self, message, payload, protocol, writer, task,
             self._loop,
             client_max_size=self._client_max_size)
 
+    def _prepare_middleware(self):
+        for m in reversed(self._middlewares):
+            if getattr(m, '__middleware_version__', None) == 1:
+                # first argument is request or r and this is
```
asvetlov, line 279: The comment is irrelevant

aiohttp/web_middlewares.py
```diff
@@ -23,6 +24,11 @@ def _check_request_resolves(request, path):
     return False, request
 
 
+def new_middleware(f):
```
asvetlov: Hmm.
I don't like the name. We have `new_middleware` and have no `old_middleware`.
Mybe just `@middleware` is enough?

samuelcolvin: I agree that I don't like `new_middleware`, I'll switch to `middleware`.

kxepal: +1

docs/web.rst
```diff
@@ -997,42 +997,31 @@ the keyword-only ``middlewares`` parameter when creating an
    app = web.Application(middlewares=[middleware_factory_1,
                                       middleware_factory_2])
 
-A *middleware factory* is simply a coroutine that implements the logic of a
-*middleware*. For example, here's a trivial *middleware factory*::
+A *middleware* is just a coroutine that can modify either the request or
+response. For example, here's a simple *middleware* which appends
+``' wink'`` to the response::
```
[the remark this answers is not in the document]
samuelcolvin: done.

asvetlov:
Perfect!
Feel free to merge after making spell checker happy.
Hint: use *does not* instead of *doesn't*

samuelcolvin:
Passing except for codecov. I can't merge.

Can I suggest configuring codecov so it's not this brittle?

asvetlov:
That's because line https://github.com/aio-libs/aiohttp/pull/2252/files#diff-724d27c86f7c189969e6112c6976feacR62 is only partially covered but I could live with it.

@samuelcolvin if you'll cover the line -- I very appreciate it.
If not -- just say "don't want", I'll merge the PR as is.

Hmm, maybe I'm wrong -- feel free to make a PR with codecov config improvement.

samuelcolvin:
I don't want to here, I thought about it but it's not really related to this PR (the line was there before, with just the indent changed).

asvetlov:
Accepted

asvetlov:
Thanks for such valuable improvement, @samuelcolvin

Issue #2248, opened 2017-09-06
Test only asyncio loop by default.


asvetlov:
aiohttp uses parametrized `loop` fixture which runs all test suite several times: one for asyncio itself and another ones for uvloop and tokio if they are installed.
It's good for aiohttp testing itself but bad for `pytest-aiohttp` users.
For example test suites for my jobs take about 15 mins, we don't like to double the time (we use uvloop, sure).
We could solve the problem by pushing custom `loop` fixture into our `conftest.py` but it is not obvious and robust solution.

I suggest moving `pytest_configure` from `pytest_plugin.py` into `tests/conftest.py` and using only asyncio loop for tests by default.
As an option we could add support for command line `--aiohttp-loops` option and `aiohttp-loops` in `pytest.ini` for helping people to run their tests on specific asyncio loop implementations.

argaen:
If you go for this, I would strongly +1 for adding a parameter where you can specify with which loops you want to run and pick only one by default (at least for development).

However for travis builds, I would keep passing tests for all the loops (dunno if that was included already in your intentions)

asvetlov:
@argaen  yep, it is my intention.
BTW, would you work share my duty and on PR? 
The code change is not very hard.

argaen:
Sure can do

argaen:
After checking, I see that the `--loop` option is already supported with the pytest_plugin. The behavior is the following:

1. If nothing is specified, it picks installed loops.
2. If --loop specified, only picks the specified one (you can specify multiple one separated by comma).

With this I have couple of proposals:

1. Leave current behavior

With this, I think its just a matter of updating the `Makefile` needed commands to just use one loop. From what I've seen: `test`, `vtest`, `cov-dev`. Not so sure about `cov` which ends up using tox.

Then I would leave `cov-ci-no-ext`, `cov-ci-aio-debug`, `cov-ci-run` which are the ones used by travis as they are

2. Change default behavior so by default it always picks one loop (no matter if uvloop or tokio are installed) and if you want to run tests with multiple loops, you have to use the `--loop` option. Ofc this requires updating the commands used by travis to use all loops

asvetlov:
I'm talking not about aiohttp test suite (running all available loops is perfectly fine) but changing `pytest-aiohttp` default behavior. Now it shares `loop` fixture with internal `./tests`.

samuelcolvin:
[...]

PR #2253, opened 2017-09-07
use pyloop as default for pytest --loop

samuelcolvin:
## What do these changes do?

Change the default loop for pytest tests to be the python asyncio loop.

Yes, the default loop is now `pyloop` rather than all, all available loops can be used via `--loop all`

## Related issue number

#2248

## Checklist

- [ ] ~~Documentation reflects the changes~~ no docs for this anyway

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -31,7 +31,7 @@ def pytest_addoption(parser):
         help='run tests faster by disabling extra checks')
     parser.addoption(
         '--loop', action='store', default='pyloop',
-        help='run tests with specific loop: pyloop, uvloop, tokio')
+        help='run tests with specific loop: pyloop, uvloop, tokio or all')
     parser.addoption(
         '--enable-loop-debug', action='store_true', default=False,
         help='enable event loop debug mode')
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -30,7 +30,7 @@ def pytest_addoption(parser):
         '--fast', action='store_true', default=False,
         help='run tests faster by disabling extra checks')
     parser.addoption(
-        '--loop', action='append', default='pyloop',
+        '--loop', action='store', default='pyloop',
         help='run tests with specific loop: pyloop, uvloop, tokio')
     parser.addoption(
         '--enable-loop-debug', action='store_true', default=False,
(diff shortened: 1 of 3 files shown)


aiohttp/pytest_plugin.py
```diff
@@ -30,7 +30,7 @@ def pytest_addoption(parser):
         '--fast', action='store_true', default=False,
         help='run tests faster by disabling extra checks')
     parser.addoption(
-        '--loop', action='append', default=[],
+        '--loop', action='store', default='pyloop',
         help='run tests with specific loop: pyloop, uvloop, tokio')
```
argaen: Can you add the option `all` in `help`?

aiohttp/pytest_plugin.py
```diff
@@ -119,29 +119,19 @@ def pytest_configure(config):
...
+        loops = 'pyloop,uvloop?,tokio?'
+
+    for name in loops.split(','):
+        required = not name.endswith('?')
+        name = name.strip(' ?')
+        if name in factories:
+            LOOP_FACTORIES.append(factories[name])
+            LOOP_FACTORY_IDS.append(name)
+        elif required:
+            raise ValueError(
```
samuelcolvin, line 132: I tried adding a test to add coverage for this case:

```python
def test_wrong_loop(testdir, capsys):
    testdir.makepyfile("""\
import asyncio
pytest_plugins = 'aiohttp.pytest_plugin'

@asyncio.coroutine
def test_foobar():
    assert 1 == 1
""")
    result = testdir.runpytest('-p', 'no:sugar', '--loop', 'foobar')
    print(result.outlines)
```

But I get back an error saying the `--loop` argument is not recognised, I get the same error if I use a real loop eg. `pyloop` so I think this is a limitation of the pytest self test fixtures.

samuelcolvin: it wasn't covered before.

asvetlov: Ok, let's leave these lines untested. I could live with it.

samuelcolvin:
This is only failing because python nightly failed to build and codecov.

Surely `nightly` show be allowed to fail?

asvetlov:
Yes, nightly build is unstable for months, please just ignore it.

asvetlov:
Thanks!

PR #2260, opened 2017-09-11
run_app: make print=None disable printing

seirl:
run_app currently takes a callable compatible with `print()`. If you want to disable output completely (which happens all the time when you're writing test suites or similar), you have to call run_app with `print=lambda *_, **__: None`, which is not very pretty and confusing.

This behavior is annoying, so this change makes print=None a special case that disables output completely.

This change is backwards-compatible.

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -447,8 +447,9 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
                 pass
 
         try:
-            print("======== Running on {} ========\n"
-                  "(Press CTRL+C to quit)".format(', '.join(uris)))
+            if print:
+                print("======== Running on {} ========\n"
+                      "(Press CTRL+C to quit)".format(', '.join(uris)))
             loop.run_forever()
         except (GracefulExit, KeyboardInterrupt):  # pragma: no cover
             pass
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -29,6 +29,7 @@ Andrew Leech
 Andrew Lytvyn
 Andrew Svetlov
 Andrii Soldatenko
+Antoine Pietri
 Anton Kasyanov
 Arthur Darcet
 Ben Bader
--- a/changes/2260.feature
+++ b/changes/2260.feature
@@ -1,0 +1,1 @@
+run_app: Make print=None disable printing
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -2400,7 +2400,8 @@ Utilities
                        ``None`` for HTTP connection.
 
    :param print: a callable compatible with :func:`print`. May be used
-                 to override STDOUT output or suppress it.
+                 to override STDOUT output or suppress it. Passing `None`
+                 disables output.
 
    :param int backlog: the number of unaccepted connections that the
                        system will allow before refusing new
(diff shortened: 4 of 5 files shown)


samuelcolvin:
+1: for this, but I've always thought the `print` argument was ugly. 

Really this should be logged, but since python (very unfortunately) python doesn't have a `NOTICE` log level I'm not sure of the proper solution.

Perhaps `aiohttp` should have a log which is setup to write to stdout out at INFO level by default?

(I'm guess you had to create the PR before you could add a file to `changes`?)

seirl:
2017-09-11 12:39 GMT+02:00 Samuel Colvin <notifications@github.com>:
>
> +1: for this, but I've always thought the print argument was ugly.
>
> Really this should be logged, but since python (very unfortunately) python doesn't have a NOTICE log level I'm not sure of the proper solution.
>
> Perhaps aiohttp should have a log which is setup to write to stdout out at INFO level by default?

I agree with all of that, and probably a verbose=True argument instead
of print=. That said, I did not feel confident to add logging since
aiohttp doesn't use much logging at all, and since this would have
been a backwards incompatible change I'd rather have this quick and
easy solution in first before we find a "proper" way to replace it.

> (I'm guess you had to create the PR before you could add a file to changes?)

Yeah, sorry, it was in my staged changes but I went to lunch and forgot.
It's there now.

seirl:
The lambda assignment made QA explode so I'm doing a regular `def` instead...

asvetlov:
I'm ok with PR but disagree with logging usage for printing text like `Press Ctrl+C to quit`.

aiohttp/web.py
```diff
@@ -422,6 +422,10 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
     app._set_loop(loop)
     loop.run_until_complete(app.startup())
 
+    if print is None:
```
samuelcolvin: much easier just do `if print` below before calling it.

seirl: I guess, if that's the only place you ever plan to use that print argument?

seirl: There, I updated it.

samuelcolvin: looks like you having a linting error :smile:

seirl: Woops, failed my block selection. There you go, sorry :-P

seirl:
... wtf?

samuelcolvin:
codecov occasionally goes mad and does strange things. Is the branch bases off current master?

seirl:
Sure, parent commit is c7e835656abeabac81fe3b8cddd70343df0c4eba .

samuelcolvin:
Would probably be good to add a test for `print=None`, that might prompt codecov to correct itself.

seirl:
Okay, rather than adding a new test I updated one that already contained `print=lambda *args: None`.

aiohttp/web.py
```diff
@@ -447,8 +447,9 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
                 pass
 
         try:
-            print("======== Running on {} ========\n"
-                  "(Press CTRL+C to quit)".format(', '.join(uris)))
+            if print:
```
zopieux, line 450: `if callable(print):` maybe? The discoverability of this argument is poor, intuition makes me want to use `print=True` which will fail miserably.

samuelcolvin: better to "fail miserably" than fail silently. In neither case would `print=True` do what you expected.

Much better to use python typing hinting for public methods like this once 3.4 support is dropped.

asvetlov:
Thanks!

Issue #673, opened 2015-12-10
Compression Extensions for WebSocket


kxepal:
See [RFC-7692](http://tools.ietf.org/html/rfc7692).

asvetlov:
See also #309

jzwinck:
`permessage-deflate` is supported in Chrome (since 32), Firefox (since 37), Tornado, Autobahn, and wspy. It would be a great addition to aiohttp.

asvetlov:
@jzwinck totally agree with you.
Would you work on Pull Request?

fanthos:
aaugustin/websockets has it implemented.
That project licensed under BSD-3.
The extension source code is at:
https://github.com/aaugustin/websockets/blob/master/websockets/extensions/permessage_deflate.py

asvetlov:
Thanks, @fanthos

PR #2273, opened 2017-09-16
Add initial support for PerMessage Deflate

fanthos:
## What do these changes do?

Add initial support for WebSocket server side deflate compress.
Compress for WS client is not written yet.

No

## Related issue number

#673

## Checklist


fafhrd91:
this is good! but could you add unit and functional tests

fanthos:
The file upload tests in `test_client_functional.py` keep failing with HTTP 500 on my device.
I am using Windows 10 1703 with Python 3.6.2.

aiohttp/client.py
```diff
@@ -380,7 +384,8 @@ def _ws_connect(self, url, *,
                     origin=None,
                     headers=None,
                     proxy=None,
-                    proxy_auth=None):
+                    proxy_auth=None,
+                    compress=0):
```
asvetlov, line 405: Let's use `compress=15` as initial value and don't accept `True`

aiohttp/client.py
```diff
@@ -26,6 +26,8 @@
 from .helpers import (PY_35, CeilTimeout, TimeoutHandle, _BaseCoroMixin,
                       deprecated_noop, sentinel)
 from .http import WS_KEY, WebSocketReader, WebSocketWriter
+from .http_websocket import extensions_gen as ws_ext_gen
```
asvetlov: Why do you rename imports?
Just use `ws_ext_gen` and `ws_ext_parse` in `http_websocket.py`.

aiohttp/http_websocket.py
```diff
@@ -146,6 +147,52 @@ def _websocket_mask_python(mask, data):
...
+            break
+
+    return compress, server_notakeover, client_notakeover
+
+
+def extensions_gen(compress=0, server_notakeover=False,
+                   client_notakeover=False):
+    # compress wbit 8 does not support in zlib
+    if compress < 9 or compress > 15:
+        return False
```
asvetlov: Raise `ValueError` here.

aiohttp/http_websocket.py
```diff
@@ -146,6 +147,52 @@ def _websocket_mask_python(mask, data):
...
+        return 0, False, False
+
+    extensions = [s.strip() for s in extstr.split(',')]
+    compress = 0
+    server_notakeover = False
+    client_notakeover = False
+    for ext in extensions:
+        if ext.startswith('permessage-deflate'):
+            compress = 15
+            for param in [s.strip() for s in ext.split(';')][1:]:
```
asvetlov: I believe regex parse is more robust.

aiohttp/client.py
```diff
@@ -420,6 +424,11 @@ def _ws_connect(self, url, *,
             headers[hdrs.SEC_WEBSOCKET_PROTOCOL] = ','.join(protocols)
         if origin is not None:
             headers[hdrs.ORIGIN] = origin
+        if compress:
+            if compress is True:
```
asvetlov: The check is not needed: let's assume `compress` is always integer.

aiohttp/client.py
```diff
@@ -478,12 +485,35 @@ def _ws_connect(self, url, *,
...
 
+            # websocket compress
+            notakeover = False
+            if compress:
+                if hdrs.SEC_WEBSOCKET_EXTENSIONS in resp.headers:
+                    try:
+                        compress, notakeover = ws_ext_parse(
+                            resp.headers[hdrs.SEC_WEBSOCKET_EXTENSIONS]
+                        )
+                        if compress == 0:
```
asvetlov: What the check does?

aiohttp/client.py
```diff
@@ -478,12 +485,35 @@ def _ws_connect(self, url, *,
                         protocol = proto
                         break
 
+            # websocket compress
+            notakeover = False
+            if compress:
+                if hdrs.SEC_WEBSOCKET_EXTENSIONS in resp.headers:
```
asvetlov: Use `resp.headers.get(hdrs.SEC_WEBSOCKET_EXTENSIONS)`, do getting the header only once.

aiohttp/http_websocket.py
```diff
@@ -106,6 +108,13 @@ def __init__(self, code, message):
         super().__init__(message)
 
 
+class WSHandshakeError(Exception):
+    """WebSocket protocol handshake error."""
+
+    def __init__(self, message):
```
asvetlov: Constructor from parent class is ok, isn't it?
Drop `__init__` function.

docs/client_reference.rst
```diff
@@ -524,6 +524,8 @@ The client session supports the context manager protocol for self closing.
       :param aiohttp.BasicAuth proxy_auth: an object that represents proxy HTTP
                                            Basic Authorization (optional)
 
+      :param bool compress: Enable Per-Message Compress Extension support
```
asvetlov: Update method signature too.

asvetlov:
Last question: should we enable websockets by default?
It could break existing programs. But we can fix bugs quickly I hope.
As an option for smooth transition we could use `0` compression by default in next release and switch to `15` in after next.
Thoughts?

fanthos:
I think we can disable compression for client side by default, maybe need more test with other server software to make sure they are compatible.
The server code is tested with Firefox and Chrome and works as excepted.

asvetlov:
@fanthos sounds good

asvetlov:
`WebSocketResponse(..., compress=True)` should be documented

asvetlov:
Thanks!

Issue #2193, opened 2017-08-12
KeyError: <aiohttp.connector._TransportPlaceholder>


rndusr:
I'm not sure if this is a bug in aiohttp, but if it's a bug in my code, this error doesn't tell my anything:

```
  File "/home/ich/code/stig/stig/client/aiotransmission/rpc.py", line 328, in __post
    response = await self.__session.post(repr(self.__url), data=data, headers=self.__headers)
  File "/home/ich/code/aiohttp/aiohttp/helpers.py", line 102, in __await__
    ret = yield from self._coro
  File "/home/ich/code/aiohttp/aiohttp/client.py", line 233, in _request
    conn = yield from self._connector.connect(req)
  File "/home/ich/code/aiohttp/aiohttp/connector.py", line 386, in connect
    self._acquired.remove(placeholder)
KeyError: <aiohttp.connector._TransportPlaceholder object at 0x7f270bc78c88>
```

So, `placeholder` is added to `self._acquired` (line 379) and then removed again (line 386), but inbetween it has already been removed, hence the `KeyError`.

The culprit is `self._create_connection(req)` (line 382), and eventually `self._loop.create_connection()` (line 732), and then I'm losing track.

Anyway, a simple `if placeholder in self._acquired:` inserted in line 386 seems to fix the issue for me. Or it could be something wrong with my code, but then I still have no idea what it is I'm doing wrong. I'd appreciate any pointers what could cause this.

I can reproduce, but unfortunately not with simple, sharable code.

rndusr:
After some more investigating, I think the issue is that I somehow `close()` my `ClientSession` before `_create_connection()` is finished.  `close()` calls `self._acquired.clear()` (line 328) and that explains the KeyError when `placeholder` is to be removed from `self._acquired` after `_create_connection()` finished.

I've fixed my code, but it seems to me that this is still an issue in aiohttp. I should either get a more helpful exception or `placeholder` should be removed with the `discard()` method, which does nothing if `placeholder` isn't in `self._acquired`.

fafhrd91:
fixed by 941a218

fafhrd91:
e404760

PR #2282, opened 2017-09-20
Handle session close during connection, KeyError: <aiohttp.connector.…

fafhrd91:
…_TransportPlaceholder> #2193


asvetlov:
@fafhrd91 could you make a test for fixing partial coverage blame?
If not -- I could live with it.
Your code is safe, the change is obvious.
I'm talking just about raising coverage level a bit.
If we'll merge the PR as is -- nobody will write a test for the case, sure.

Issue #2284, opened 2017-09-21
Server request's state is not cloned.


asvetlov:
`req.clone()` call doesn't clone request state, changes made by `req['key'] = val` are not propagated to clone.

The solution is calling `req._state.copy()` in `clone`

Issue #2171, opened 2017-08-05
Insecure algorithm for determining remote/scheme/host


vfaronov:
## Long story short

`aiohttp.web.BaseRequest` has attributes [`scheme`](http://aiohttp.readthedocs.io/en/latest/web_reference.html#aiohttp.web.BaseRequest.scheme), [`host`](http://aiohttp.readthedocs.io/en/latest/web_reference.html#aiohttp.web.BaseRequest.host) and, since 2.3, [`remote`](http://aiohttp.readthedocs.io/en/latest/web_reference.html#aiohttp.web.BaseRequest.remote). Their values are determined by looking at request headers that can be manipulated by a remote user. Under most configurations, the user can set them to **any desired values**. This is especially bad for the `remote` attribute, because the user’s IP address is often relied upon for access controls.

(I’m reporting this in the open rather than privately, because `remote` is not yet released, while `scheme` and `host` do not seem like much of a security problem.)

## Expected behaviour

This involves the headers [`Forwarded`](https://tools.ietf.org/html/rfc7239) and `X-Forwarded-For`. The **last element** (comma-separated) in these headers can be trusted **as long as a trusted proxy is configured to append it**.

The same applies to the headers `X-Forwarded-Host` and `X-Forwarded-Proto`, with the exception that they are usually single values (to be replaced by the proxy), not comma-separated lists.

## Actual behaviour

aiohttp takes the **first element** from `Forwarded`/`X-Forwarded-For`, **without knowing which of these headers (if any) are controlled by a trusted proxy**.

In most deployments where aiohttp sits behind nginx, the `Forwarded` header is not controlled by the proxy. Nobody is aware that it needs to be controlled. It is not mentioned in the [example nginx configuration](http://aiohttp.readthedocs.io/en/latest/deployment.html#nginx-configuration) from aiohttp docs. In such deployments, an external user can trivially force `scheme`, `host` and `remote` to any desired values by sending a header like:

```
Forwarded: for=10.0.0.1;host=example.net;proto=https
```

In fact, it’s impossible to configure current versions of nginx to *correctly* append a `Forwarded` header (without writing some C or possibly Lua code).

But even if aiohttp was sitting behind a proxy that correctly controlled all of the involved headers, nothing would change, because the proxy would *append* a comma-separated element to the remote user’s `Forwarded` header, while aiohttp is looking at the *first* element — which is still controlled by the user.

## Steps to reproduce

Run this server program:

    from aiohttp import web
[...]

PR #2287, opened 2017-09-23
Clone web request's state

asvetlov:
Fix for #2284

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -62,7 +62,10 @@ class BaseRequest(collections.MutableMapping, HeadersMixin):
 
     def __init__(self, message, payload, protocol, writer, time_service, task,
                  loop,
-                 *, secure_proxy_ssl_header=None, client_max_size=1024**2):
+                 *, secure_proxy_ssl_header=None, client_max_size=1024**2,
+                 state=None):
+        if state is None:
+            state = {}
         self._message = message
         self._protocol = protocol
         self._transport = protocol.transport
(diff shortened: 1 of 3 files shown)


aiohttp/web_request.py
```diff
@@ -120,7 +123,8 @@ def clone(self, *, method=sentinel, rel_url=sentinel,
             self._time_service,
             self._task,
             self._loop,
-            secure_proxy_ssl_header=self._secure_proxy_ssl_header)
+            secure_proxy_ssl_header=self._secure_proxy_ssl_header,
+            state=self._state.copy())
```
kxepal, line 127: I hope there are no nested dicts/list in state, right?

asvetlov:
It's impossible to perform deep state copy for an obvious reasons: values are not copyable very often, they could be an opened socket for example.

Thus we have 3 options:
1. Cleanup a state on copy.
2. Do swallow copy.
3. Share state between all copies.

I believe the option 2 is  better than others.

kxepal:
@asvetlov What's a case for `req.clone()`? May be we can make it immutable instead to do not any copies?

asvetlov:
`.clone()` is intended to be used in middlewares:
```
@asyncio.coroutine
def middleware_factory(app, next_handler):
    @asyncio.coroutine
    def middleware(request):
        request2 = request.clone(rel_url="/new_url", method="GET")
        response = yield from next_handler(request2)
    return middleware
```

I'd like to extend `clone()` params for fixing #2171 too.

kxepal:
Hm...I see. Then it seems swallow copy is the only option.

Issue #2294, opened 2017-09-28
Implement aiohttp.client_exceptions.CertificateError and aiohttp.client_exceptions.SSLError


hellysmile:
## Long story short

My idea is implement 2 new exceptions `aiohttp.client_exceptions.CertificateError` and `aiohttp.client_exceptions.SSLError` which has bases `ssl.CertificateError` and `ssl.SSLError` as well `aiohttp.client_exceptions.ClientOSError`. When users catch `aiohttp.client_exceptions.ClientOSError` there is no way to understand is it ssl error, dns error or networking error or whatever else

There is a list of all exceptions and their bases in ssl module

```
<class 'ssl.CertificateError'> (<class 'ValueError'>,)
<class 'ssl.SSLEOFError'> (<class 'ssl.SSLError'>,)
<class 'ssl.SSLError'> (<class 'OSError'>,)
<class 'ssl.SSLSyscallError'> (<class 'ssl.SSLError'>,)
<class 'ssl.SSLWantReadError'> (<class 'ssl.SSLError'>,)
<class 'ssl.SSLWantWriteError'> (<class 'ssl.SSLError'>,)
<class 'ssl.SSLZeroReturnError'> (<class 'ssl.SSLError'>,)
```

## Expected behaviour

`
aiohttp.client_exceptions.CertificateError: Cannot connect to host expired.badssl.com:443 ssl:True [[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748)]
` 

## Actual behaviour

will raise 

`
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host expired.badssl.com:443 ssl:True [[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:748)]
`

```python
import aiohttp
import asyncio

async def main():
    async with aiohttp.ClientSession() as session:
        async with session.get('https://expired.badssl.com/') as response:
            return await response.text()

loop = asyncio.get_event_loop()
loop.run_until_complete(main())

```

## Your environment

Python 3.6.2
Mac OS X 10.12.6
latest aiohttp master branch

asvetlov:
Fixed by #2297

PR #2297, opened 2017-09-28
ssl.SSLError explicit support

hellysmile:
## What do these changes do?

Explicit `ssl.SSLError` support and `ssl.CertificateError`

Right now `Connector` wraps any `OSError` to `ClientConnectorError `
https://github.com/aio-libs/aiohttp/blob/0b807a849564f5fefec5630b4a478789a9de1d22/aiohttp/connector.py#L813
Later `connect` wraps it one more time https://github.com/aio-libs/aiohttp/blob/0b807a849564f5fefec5630b4a478789a9de1d22/aiohttp/connector.py#L393 at this point real `OSError` is lost and 2 times `raise from`

Not sure is there any reason to catch  `OSError` in `connect`

If explicitly re raise `ClientConnectorError` only `tests.test_connector.test_connect_connection_error` fails,  but it seems mocked wrong, cuz there is no way to raise `OSError` from `_create_connection` it is already re raised to `ClientConnectorError` in `_create_direct_connection`

For first, can we remove catching `OSError` from `connect`? It makes no sense as for me

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2294

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -3,7 +3,7 @@ import functools
 import sys
 import traceback
 import warnings
-from collections import defaultdict, namedtuple
+from collections import defaultdict
 from hashlib import md5, sha1, sha256
 from itertools import cycle, islice
 from time import monotonic
@@ -138,9 +138,6 @@ class _TransportPlaceholder:
 
     def close(self):
         pass
-
-
-ConnectionKey = namedtuple('ConnectionKey', ['host', 'port', 'ssl'])
 
 
 class BaseConnector(object):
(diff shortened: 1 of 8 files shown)


hellysmile:
I am not sure hot to add tests when `ssl` module is not available

aiohttp/connector.py
```diff
@@ -390,6 +394,11 @@ def connect(self, req):
                 if self._closed:
                     proto.close()
                     raise ClientConnectionError("Connector is closed.")
+            except ClientConnectorSSLError as exc:
```
hellysmile: What about just `raise` here?

hellysmile:
What about rename `aiohttp.ClientConnectorSSLError` just to `aiohttp.ClientSSLError` ?

fafhrd91:
will look later today

Issue #2270, opened 2017-09-14
aiohttp client support for Brotli compression


kakarukeys:
## Long story short

Currently unable to read brotli-compressed response. Is the support on the roadmap?

## Expected behaviour

Able to read.

## Actual behaviour

awaiting `response.text` will raise
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 5: invalid continuation byte

## Steps to reproduce

```python
import asyncio, aiohttp
run = asyncio.get_event_loop().run_until_complete

async def f():
    async with aiohttp.ClientSession(headers={'Accept-Encoding': 'br'}) as session:
        async with session.get("https://www.amazon.com") as resp:
            html = await resp.text()

run(f())
```

## Your environment

Python 3.6.2
aiohttp==2.2.5
chardet==3.0.4
on Linux Mint 13

asvetlov:
Never heard about Brotli before :)
But PR is welcome!

fafhrd91:
I think brotli is from dropbox? should be easy to add.

yes, PR is always welcome

asvetlov:
Fixed by #2312

PR #2312, opened 2017-10-10
Implement Brotli support

alxpy:
## What do these changes do?
 Implement [Brotli](https://github.com/google/brotli) support

No.

## Related issue number
https://github.com/aio-libs/aiohttp/issues/2270

## Checklist

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -16,6 +16,7 @@ from .http_writer import HttpVersion, HttpVersion10
 from .log import internal_logger
 from .streams import EMPTY_PAYLOAD, FlowControlStreamReader
 
+
 try:
     import brotli
     HAS_BROTLI = True
@@ -619,7 +620,7 @@ class DeflateBuffer:
             self.decompressor = brotli.Decompressor()
         else:
             zlib_mode = (16 + zlib.MAX_WBITS
-                        if encoding == 'gzip' else -zlib.MAX_WBITS)
+                         if encoding == 'gzip' else -zlib.MAX_WBITS)
             self.decompressor = zlib.decompressobj(wbits=zlib_mode)
 
     def set_exception(self, exc):
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -16,6 +16,12 @@ from .http_writer import HttpVersion, HttpVersion10
 from .log import internal_logger
 from .streams import EMPTY_PAYLOAD, FlowControlStreamReader
 
+try:
+    import brotli
+    HAS_BROTLI = True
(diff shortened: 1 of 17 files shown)


asvetlov:
Good start but tests are required.

oleksandr-kuzmenko:
I hope this is not a big problem
`Coverage   97.23%   97.18%   -0.05% `

asvetlov:
I pretty happy with the value.
Please add `# pragma: no cover` for lines which are executed if no `brotlipy` installed.

hellysmile:
Hello there, recently I was playing with exactly brotli and were using official python package.

Is there any difference between https://pypi.python.org/pypi/Brotli and https://pypi.python.org/pypi/brotlipy

Can we support both of them?

oleksandr-kuzmenko:
@hellysmile hello
I can't find docs for google `brotli` py-package =(

`Brotlipy` is a collection of bindings to the Brotli C library by Google, has docs and:
> Brotlipy has a very similar interface to the standard library’s zlib module

I think no:
> Can we support both of them?

oleksandr-kuzmenko:
`Coverage   97.23%   97.23%   +<.01%`
FYI @asvetlov

oleksandr-kuzmenko:
In the near future I will try to add brotli support to server side

asvetlov:
I mean now we do support brotli decompression only, compressing requests/responses would be very useful too.

oleksandr-kuzmenko:
I got what you meant

kxepal:
awesome!

hellysmile:
I've tried to play around brotli\brotlipy, at least test is green on both of them.

asvetlov:
Looks like we need mention both libraries in doc

oleksandr-kuzmenko:
ok, i'll update docs

aiohttp/http_parser.py
```diff
@@ -605,23 +612,29 @@ def __init__(self, out, encoding):
         self.encoding = encoding
         self._started_decoding = False
 
-        zlib_mode = (16 + zlib.MAX_WBITS
-                     if encoding == 'gzip' else -zlib.MAX_WBITS)
-
-        self.zlib = zlib.decompressobj(wbits=zlib_mode)
+        if encoding == 'br':
+            if not HAS_BROTLI:  # pragma: no cover
```
webknjaz, line 616: Why this has `no cover`? Good coverage matters

asvetlov:
We don't use extras in aiohttp (but I like the idea).
Hmm, we should mention `brotly`/`brotlipy` as optional dependency in `index.rst` and `README.rst` along with `cchardet` and `aiodns`.

Issue #2315, opened 2017-10-11
aiohttp.wep.Application.make_handler should support pluggable AccessLogger


hellysmile:
## Long story short

There is now way to use custom `aiohttp.helpers.AccessLogger` for `aiohttp.wep.Application.make_handler`

## Expected behaviour

`aiohttp.wep.Application.make_handler` support pluggable `aiohttp.helpers.AccessLogger`

## Actual behaviour

`aiohttp.web.Application.make_handler` do not support pluggable `aiohttp.helpers.AccessLogger`

asvetlov:
Fixed by #2316

PR #2316, opened 2017-10-12
aiohttp.wep.Application.make_handler access_log_class support

hellysmile:
## What do these changes do?

Adds support pluggable `aiohttp.helpers.AccessLogger` for `aiohttp.wep.Application.make_handler`

No breaking changes.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2315

## Checklist

### DIFF
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -152,7 +152,21 @@ class AbstractAccessLogger(ABC):
 
     def __init__(self, logger, log_format):
         self.logger = logger
+        self.log_format = log_format
 
     @abstractmethod
+    def _log(self, request, response, time):
+        """
+
+        :param request: Request object.
+        :param response: Response object.
+        :param float time: Time taken to serve the request.
+        """
+
     def log(self, request, response, time):
         """Emit log to logger"""
+
+        try:
+            self._log(request, response, time)
+        except Exception:
+            self.logger.exception("Error in logging")
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -146,3 +146,13 @@ class AbstractPayloadWriter(ABC):
     @abstractmethod
     def drain(self):
         """Flush the write buffer."""
+
+
+class AbstractAccessLogger(ABC):
+
+    def __init__(self, logger, log_format):
+        self.logger = logger
+
+    @abstractmethod
+    def log(self, request, response, time):
+        """Emit log to logger"""
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -155,18 +155,5 @@ class AbstractAccessLogger(ABC):
         self.log_format = log_format
 
     @abstractmethod
-    def _log(self, request, response, time):
-        """
-
-        :param request: Request object.
-        :param response: Response object.
-        :param float time: Time taken to serve the request.
-        """
-
     def log(self, request, response, time):
         """Emit log to logger"""
-
-        try:
-            self._log(request, response, time)
-        except Exception:
-            self.logger.exception("Error in logging")
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -237,8 +237,10 @@ class Application(MutableMapping):
                      **kwargs):
 
         if not issubclass(access_log_class, AbstractAccessLogger):
-            raise TypeError('access_log_class must be subclass of '
-                            'aiohttp.abc.AbstractAccessLogger')
+            raise TypeError(
+                'access_log_class must be subclass of '
+                'aiohttp.abc.AbstractAccessLogger, got {}'.format(
+                    access_log_class))
 
         self._set_loop(loop)
         self.freeze()
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -17,6 +17,7 @@ from . import (hdrs, web_exceptions, web_fileresponse, web_middlewares,
                web_urldispatcher, web_ws)
 from .abc import AbstractAccessLogger, AbstractMatchInfo, AbstractRouter
 from .frozenlist import FrozenList
+from .helpers import AccessLogger
 from .http import HttpVersion  # noqa
 from .log import access_logger, web_logger
 from .signals import FuncSignal, PostSignal, PreSignal, Signal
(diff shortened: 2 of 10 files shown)


hellysmile:
Here is example of `AccessLogger` which produces messages exactly same like current `aiohttp.helpers.AccessLogger` with default `log_format`

```python
class AccessLogger(AbstractAccessLogger):

        msg = '{request.remote} - - {t} "{r}" {response.status} {response.body_length} "{referrer}" "{user_agent}"'

        r = '%s %s HTTP/%s.%s' % tuple((request.method,
                                        request.path_qs) + request.version)

        t = datetime.datetime.utcnow().strftime('[%d/%b/%Y:%H:%M:%S +0000]')

        self.logger.info(msg.format(
            t=t,
            r=r,
            referrer=request.headers.get('Referrer', '-'),
            user_agent=request.headers.get('User-Agent', '-'),
            request=request,
            response=response,
            time=time,
        ))
```

I've played a bit with this `AccessLogger`, and server can be speeded up to 5% in comparison to the default `aiohttp.helpers.AccessLogger`, but it is not big deal.

Most useful example is something like:

```python

class AccessLogger(AbstractAccessLogger):

        self.logger.info(f'{request.remote} "{request.method}" -> {response.status}')
```

If proposal at least makes some sense, I'll do full test coverage/docs/etc

asvetlov:
Please make the feature well documented

aiohttp/abc.py
```diff
@@ -146,3 +146,13 @@ def write_eof(self, chunk=b''):
     @abstractmethod
     def drain(self):
         """Flush the write buffer."""
+
+
+class AbstractAccessLogger(ABC):
+
+    def __init__(self, logger, log_format):
+        self.logger = logger
```
kxepal, line 154: `log_format` seems to be lost.

aiohttp/abc.py
```diff
@@ -146,3 +146,13 @@ def write_eof(self, chunk=b''):
...
+
+    @abstractmethod
+    def log(self, request, response, time):
```
kxepal: Need to document which objects and types that method accepts.

aiohttp/abc.py
hellysmile: Yes, but it is exactly how current `AccessLogger` works, as well default `log_format` is `AccessLogger.LOG_FORMAT` not sure is it needs to be stored as `self.log_format`. My idea was that, `logger` itself is only one required attribute, `log_format` is actually optional

aiohttp/web_protocol.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, manager, *, loop=None,
                  tcp_keepalive=True,
                  slow_request_timeout=None,
                  logger=server_logger,
+                 access_log_factory=helpers.AccessLogger,
```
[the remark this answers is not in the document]
hellysmile: What about `access_log_cls` or `access_log_class` ?

aiohttp/abc.py
hellysmile: Sure, it will be documented

hellysmile:
@kxepal What do You think to make `self._log` `@abstractmethod` and `self.log` will be:

```python

    try:
        self._log(request, response, time)
    except Exception as e:
        self.logger.exception("Error in logging", exc_info=e)
```

aiohttp/abc.py
kxepal: So, if I'm making own `AccessLogger` I should keep in mind, that super requires two arguments, one of which it will set to attributes and one that it will ignore? A bit not friendly behavior. The `AccessLogger` may have overloaded `__init__` to accept `log_format`, but if the `AbstractAccessLogger` doesn't uses it in anyway there is no point to keep it there.

kxepal:
@hellysmile 
Good idea if we don't want to let logger crash the request. Though, it may hide errors for awhile. Also, no need to pass `exc_info` for `logger.exception` - it already catches it by default.

aiohttp/web_protocol.py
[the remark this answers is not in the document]
kxepal: `access_log_class` is good one.

hellysmile:
There is one more question, it will be nice to place somewhere

```python
assert issubclass(access_log_class, AbstractAccessLogger), access_log_class
```

but where to place is, actually `access_logger` will be created lazy on first request

asvetlov:
@hellysmile push the check into `app.make_handler()`.

aiohttp/web.py
```diff
@@ -231,6 +231,14 @@ def middlewares(self):
         return self._middlewares
 
     def make_handler(self, *, loop=None, **kwargs):
+        if 'access_log_class' in kwargs:
```
asvetlov: Just add `access_log_class=helpers.AccessLogger` into method signature.

aiohttp/web.py
```diff
@@ -231,6 +231,14 @@ def middlewares(self):
...
+            access_log_class = kwargs['access_log_class']
+
+            assert issubclass(access_log_class, AbstractAccessLogger), (
```
asvetlov: Should be `TypeError`, not `AssertionError`.

aiohttp/abc.py
```diff
@@ -146,3 +146,27 @@ def write_eof(self, chunk=b''):
...
+        :param request: Request object.
+        :param response: Response object.
+        :param float time: Time taken to serve the request.
+        """
+
+    def log(self, request, response, time):
+        """Emit log to logger"""
+
+        try:
+            self._log(request, response, time)
```
asvetlov: Let's drop `_log()` and make `log()` abstract method.
On one hand formatting logic always should be wrapped by catching exception obviously.
On other -- let's not make any decision about implementation details.
Implementor might want to do other things than just registering caught exception.
Let's keep ABC very abstract, my guts feel that there are many contradictory might occur if we will bake a code in abstract class.

aiohttp/abc.py
```diff
@@ -146,3 +146,27 @@ def write_eof(self, chunk=b''):
...
+
+    def __init__(self, logger, log_format):
+        self.logger = logger
+        self.log_format = log_format
+
+    @abstractmethod
+    def _log(self, request, response, time):
+        """
+
+        :param request: Request object.
```
asvetlov: Please don't use sphinx formatting in docstrings -- it is a text for readers.
I'd like to rip it out everywhere and use only hand written documentation.

aiohttp/web.py
```diff
@@ -230,7 +231,15 @@ def router(self):
...
 
-    def make_handler(self, *, loop=None, **kwargs):
+    def make_handler(self, *,
+                     loop=None,
+                     access_log_class=AccessLogger,
+                     **kwargs):
+
+        if not issubclass(access_log_class, AbstractAccessLogger):
+            raise TypeError('access_log_class must be subclass of '
+                            'aiohttp.abc.AbstractAccessLogger')
```
kxepal: Would be nice to add `got {}` to error message to simplify understanding what's wrong with the logger.

PR #2319, opened 2017-10-12
 Fix web.FileResponse overriding user supplied Content-Type

### LINKED ISSUE #2317: web.FileResponse overrides content_type
## Long story short

When returning a `web.FileResponse` from a route, any changes to the `content_type` member is not taken into account when the file is served. Instead, the file extension is always used to determine the content type of the response.

## Expected behaviour

`Content-Type` header of the response be that of `web.FileResponse.content_type` when such has been set.

## Actual behaviour

`Content-Type` header is guessed from the file extension of the served file.

## Steps to reproduce

Create a route that responds something along the lines of

```python
    response = web.FileResponse('/path/to/some/image.png')
    response.content_type = 'application/octet-stream'
    return response
```

then `curl -i` it and notice the `Content-Type` is `image/png`.

## Your environment

```
[~]$ uname -mrs
Linux 4.11.9-1-ARCH x86_64
[~]$ python --version
Python 3.6.2
[~]$ pip show aiohttp | grep Version
Version: 2.2.5
```

thelink2012:
## What do these changes do?

If the user modifies either the `web.FileResponse.content_type` property, or the `Content-Type` header, the choice is now respected instead of guessing the content type from the file extension.

No notable changes should happen.

## Related issue number

Fixes #2317 

## Checklist

- [ ] Documentation reflects the changes **(no need I guess)**
- [X] Add a new news fragment into the `changes` folder

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -49,6 +49,7 @@ Daniel García
 Daniel Nelson
 Danny Song
 David Michael Brown
+Denilson Amorim
 Denis Matiychuk
 Dima Veselov
 Dimitar Dimitrov
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -182,9 +182,13 @@ class FileResponse(StreamResponse):
             self._length_check = False
             return (yield from super().prepare(request))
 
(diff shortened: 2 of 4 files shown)


asvetlov:
Thanks!

PR #2322, opened 2017-10-13
Cleanup

asvetlov:
Minor code and documentation cleanups

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -41,8 +41,7 @@ else:
     from .backport_cookies import SimpleCookie  # noqa
 
 
-__all__ = ('BasicAuth', 'create_future', 'parse_mimetype',
-           'Timeout', 'ensure_future', 'noop')
+__all__ = ('BasicAuth', 'Timeout')
 
 
 sentinel = object()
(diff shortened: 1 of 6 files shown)


aiohttp/helpers.py
```diff
@@ -514,11 +514,11 @@ def _format_line(self, request, response, time):
     def log(self, request, response, time):
         """Log access.
 
-        :param message: Request object. May be None.
-        :param environ: Environment dict. May be None.
-        :param response: Response object.
-        :param transport: Tansport object. May be None
-        :param float time: Time taken to serve the request.
+        message: Request object. May be None.
```
hellysmile: these parameters is not relevant actually

kxepal:
I'm not sure about global sphinx markup rid. Even if not for autodocs, it's very useful for one more thing: type hints. PyCharm, for instance, can use that information if no annotations provided. And when you code, it's easy to note what kind of objects are required.

I would recommend postpone that change and made it with migration to type annotations to not loose anything.

asvetlov:
@kxepal markup is not in actual state very often and present for very small subset of API.
We need review it anyway.
I don't think that removed sphinx markup makes any sense.

kxepal:
@asvetlov ok then.

PR #2325, opened 2017-10-15
Fix #2311: Accept partial status line

### LINKED ISSUE #2311: Default Aiohttp request parser `HttpRequestParserC` fails, parsing URLs of allowed size.
## Long story short

Default Aiohttp request parser `HttpRequestParserC` fails, parsing URLs of allowed size (up to 8190 bytes allowed).
>  https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_parser.py#L658

## Expected behaviour

URL parser should not raise unexpected errors, working with URLs of allowed size and length.

## Actual behaviour

```
url = 'http://<host>:<port>/api?ids=ac886e02-bda7-41db-b3a2-257cba5824cf&ids=...&ids=85a8b805-bf92-45ac-994b-9ceb1bfa981a'
# 100 key/values in get parameters
# Totally 4 Kb, which is half of allowed 8.19 Kb for URL size
response = await client.get(url)

ERROR aiohttp.server
  File "/usr/local/lib/python3.6/dist-packages/aiohttp/web_protocol.py", line 278, in data_received
    messages, upgraded, tail = self._request_parser.feed_data(data)
  File "aiohttp/_http_parser.pyx", line 274, in aiohttp._http_parser.HttpParser.feed_data (aiohttp/_http_parser.c:4364)
  File "aiohttp/_http_parser.pyx", line 334, in aiohttp._http_parser.cb_on_url (aiohttp/_http_parser.c:5381)
  File "aiohttp/_http_parser.pyx", line 546, in aiohttp._http_parser._parse_url (aiohttp/_http_parser.c:8903)
aiohttp.http_exceptions.InvalidURLError: 400, message='invalid url b'=9a93-80409935daa8&ids=...&ids=85a8b805-bf92-45ac-994b-9ceb1bfa981a'
```
Invalid URL is actually a last truncated part of original.

## Steps to reproduce
Leave environment variable `AIOHTTP_NO_EXTENSIONS` unchanged, `False` by default.
That makes aiohttp app use default `HttpRequestParser = HttpRequestParserC`.
Run your application
```
import asyncio
import sys

import uvloop

from aiohttp import ClientSession, web


async def api_handler(request: web.Request) -> web.Response:
    get_params = request.rel_url.query
    # ...
    return web.json_response()


def main(host: str, port: int):
    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())

    app.router.add_route('GET', '/api', api_handler)

    client = ClientSession(loop=loop)
    handler = app.make_handler(loop=loop)

    try:
        loop.run_until_complete(app.startup())
        loop.run_until_complete(loop.create_server(handler, host, port))
        loop.run_forever()
    except KeyboardInterrupt:
        ...
    finally:
        loop.run_until_complete(app.cleanup())
        loop.run_until_complete(app.shutdown())
        client.close()
        loop.close()


if __name__ == '__main__':
    sys.exit(main('localhost', 8000))
```

Try to access your API endpoint from remote (*important - not from same localhost*)
response = await client.get('http://<host_ip>:8000/api?ids=uuid1&ids=...&ids=uuid100'
Raised error, described above.

## Your environment

Python 3.6.3
Aiohttp 2.2.5
Ubuntu 14.04 LTS

### LINKED ISSUE #3211: Forcibly closed connection to web.Application results in infinite loop of Exceptions that cannot be caught
## Long story short

When the TCP connection to my websocket server is forcibly closed (outside Python's control, such as with TCPView), there is an infinite loop of Exceptions that my user-level code cannot catch.

## Expected behaviour

I would expect to get a connection closed message or exception as with other kinds of service interruptions.

## Actual behaviour

I get an infinite loop of exceptions that cannot be caught.
```python-traceback
Exception in callback BaseSelectorEventLoop._read_from_self()
handle: <Handle BaseSelectorEventLoop._read_from_self()>
  File "C:\Python37-32\lib\asyncio\events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "C:\Python37-32\lib\asyncio\selector_events.py", line 125, in _read_from_self
    data = self._ssock.recv(4096)
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

## Steps to reproduce

The code below is a smallest-representation of the problem.  Run the websocket server on localhost, and connect to it with a client such as the tool at http://www.websocket.org/echo.html?location=ws://localhost:9999 .  Then forcibly close the TCP connection using a tool like TCPView to simulate an OS-level event.

```python
#!/usr/bin/env python3
import asyncio
import webbrowser

from aiohttp import web


async def setup():
    app = web.Application()
    app.router.add_get("/", ws_handler)
    runner = web.AppRunner(app)
    await runner.setup()
    site = web.TCPSite(runner, "localhost", port=9999)
    await site.start()


async def ws_handler(request: web.BaseRequest):
    ws = web.WebSocketResponse()
    try:
        await ws.prepare(request)
        await ws.send_str("Welcome to this demo.")
        await ws.send_str("Trying forcing closed this connection with something like TCPView:")
        await ws.send_str("https://docs.microsoft.com/en-us/sysinternals/downloads/tcpview")
        while not ws.closed:
            ws_msg = await ws.receive()
            print("Received:", ws_msg)
    except Exception as ex:
        print("I cannot catch the Exception when connection is forcibly closed.", flush=True)
        print("Exception", ex, flush=True)
        ex.with_traceback()
    finally:
        print("I am never run when connection is forcibly closed.", flush=True)

    return ws


webbrowser.open("http://www.websocket.org/echo.html?location=ws://localhost:9999")
loop.create_task(setup())
loop.run_forever()
```

## Your environment

- Windows 10
- aiohttp v3.3.2
- Python 3.7 32bit

asvetlov:
Fix for #2311

### DIFF
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -48,6 +48,7 @@ cdef class HttpParser:
 
         bint    _started
         object  _url
+        bytearray   _buf
         str     _path
         str     _reason
         list    _headers
@@ -92,6 +93,7 @@ cdef class HttpParser:
         self._loop = loop
         self._timer = timer
 
+        self._buf = bytearray()
         self._payload = None
         self._payload_error = 0
         self._payload_exception = payload_exception
@@ -223,6 +225,9 @@ cdef class HttpParser:
     cdef _on_chunk_complete(self):
         self._payload.end_http_chunk_receiving()
 
+    cdef object _on_status_complete(self):
+        pass
+
     ### Public API ###
 
     def http_version(self):
@@ -311,6 +316,22 @@ cdef class HttpRequestParserC(HttpParser):
                     max_line_size, max_headers, max_field_size,
                     payload_exception, response_with_body)
 
+    cdef object _on_status_complete(self):
+         cdef Py_buffer py_buf
+         if not self._buf:
+             return
+         self._path = self._buf.decode('utf-8', 'surrogateescape')
+         if self._cparser.method == 5:  # CONNECT
+             self._url = URL(self._path)
+         else:
+             PyObject_GetBuffer(self._buf, &py_buf, PyBUF_SIMPLE)
+             try:
+                 self._url = _parse_url(<char*>py_buf.buf,
+                                        py_buf.len)
+             finally:
+                 PyBuffer_Release(&py_buf)
+         self._buf.clear()
+
 
 cdef class HttpResponseParserC(HttpParser):
 
@@ -323,6 +344,11 @@ cdef class HttpResponseParserC(HttpParser):
                    max_line_size, max_headers, max_field_size,
                    payload_exception, response_with_body, auto_decompress)
 
+    cdef object _on_status_complete(self):
+        if self._buf:
+            self._reason = self._buf.decode('utf-8', 'surrogateescape')
+            self._buf.clear()
+
 
 cdef int cb_on_message_begin(cparser.http_parser* parser) except -1:
     cdef HttpParser pyparser = <HttpParser>parser.data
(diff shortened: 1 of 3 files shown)


asvetlov:
Fixed incorrect issue number in commit message

Issue #1699, opened 2017-03-07
Client basic HTTP auth domain mismatch


krya:
## Long story short

I was making calls with aiohttp client to one twilio API server (which requires basic HTTP authentication) and it was working fine untill I reached an endpoint that does a redirect to amazon S3.
But it appears that aiohttp passes basic authentication headers after redirection to a location defined in last response header even if domain doesnt match with the one that did redirection which causes an issue in my case because amazon requires that there should be only one auth method (in my cause token was in the url for redirection) 
I've checked the requests lib and it works fine in this exact case.

## Expected behaviour

once lib follows redirection and domain doesnt match with previous it should NOT send basic auth headers to this domain

## Actual behaviour

end response from S3 ends with an authentication error since there is an auth token in the URL AND a basic auth headers

## Steps to reproduce

I'm not sure on available to everyone server that does a redirection to S3

## Your environment

ubuntu 14/16
aiohttp==1.3.3 (same with git master)

fafhrd91:
makes sense for me.

@kxepal @asvetlov ?

asvetlov:
I think it is security issue and should be fixed

PR #2328, opened 2017-10-16
Auth on redirect

asvetlov:
Fix for #1699

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -331,6 +331,10 @@ class ClientSession:
                         elif not scheme:
                             r_url = url.join(r_url)
 
+                        if url.origin() != r_url.origin():
+                            auth = None
+                            headers.pop(hdrs.AUTHORIZATION)
+
                         url = r_url
                         params = None
                         resp.release()
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -333,7 +333,7 @@ class ClientSession:
 
                         if url.origin() != r_url.origin():
                             auth = None
-                            headers.pop(hdrs.AUTHORIZATION)
+                            headers.pop(hdrs.AUTHORIZATION, None)
 
                         url = r_url
                         params = None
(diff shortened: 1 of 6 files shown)


asvetlov:
Committers, I need a review of the issue.
Sure, I can merge it without approval (I'm an admin btw) but I like other eyes on it

asvetlov:
Waiting for tomorrow

aiohttp/client.py
```diff
@@ -231,6 +222,24 @@ def _request(self, method, url, *,
...
+                    if auth is None:
+                        auth = auth_from_url
+                    if auth is None:
+                        auth = self._default_auth
+                    # It would be confusing if we support explicit
+                    # Authorization header with auth argument
+                    if (headers is not None and
+                            auth is not None and
+                            hdrs.AUTHORIZATION in headers):
+                        raise ValueError("Cannot combine AUTHORIZATION header "
```
kxepal, line 239: Can we have tests which triggers these ValueError's? I'm hardly imagine the case for them.

aiohttp/client.py
```diff
@@ -322,6 +331,10 @@ def _request(self, method, url, *,
                         elif not scheme:
                             r_url = url.join(r_url)
 
+                        if url.origin() != r_url.origin():
```
kxepal, line 334: Side note: it seems like relative `Location` [is legit](https://tools.ietf.org/html/rfc7231#section-7.1.2) now.

aiohttp/client.py
asvetlov: Yes, the PR has a test for the case

aiohttp/client.py
asvetlov: Relative URL is joined to original two lines above.
`r_url` is absolute at this line.

aiohttp/client.py
kxepal: Hm...seems  it's slipped away from me. Ok.

asvetlov:
@kxepal thank you for careful review.
Relative urls was processed by aiohttp as I've commented.

aiohttp/client.py
kxepal: I'm interpreted it as schemaless url rather then relative one. Fine then.

PR #2329, opened 2017-10-16
Fix #2189: Wakeup pending waiters

### LINKED ISSUE #2189: _RequestContextManager will not close connection if ClientConnectorError has happened
## Long story short

When using _RequestContextManager and try to get a host which can not be connected to, connection will not be released to the pool.

## Expected behaviour

Raises an Exception and connection releases.

## Actual behaviour

Raises an Exception.

## Steps to reproduce

```
import asyncio

import aiohttp


async def get_image(session, url):
        if response.status != 200:
            print('Something went wrong!')
            return None

        return await response.read()


async def worker():
    conn = aiohttp.TCPConnector(limit=1)
    async with aiohttp.ClientSession(connector=conn) as session:
        await asyncio.gather(
            asyncio.ensure_future(get_image(session, 'http://unavailable:9090/test.jpg')),
            asyncio.ensure_future(get_image(session, 'http://unavailable:9090/test.jpg'))
        )


if __name__ == '__main__':

    loop.run_until_complete(worker())
    loop.close()
```

## Your environment

macOS 10.12.6
Python 3.6.1
aiohttp 2.2.5

asvetlov:
Fix for #2189 

@fafhrd91 please review.
Sure, I'll add tests if you agree in general.

BTW logic with waiters is too complicated.
Better to rewrite it with `asyncio.Condition`.

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -391,12 +391,8 @@ class BaseConnector(object):
                     proto.close()
                     raise ClientConnectionError("Connector is closed.")
             except:
-                # signal to waiter
-                waiters = self._waiters[key]
-                if waiters:
-                    waiter = waiters.pop(0)
-                    waiter.set_result(None)
-                    raise
+                self._release_waiter()
+                raise
             finally:
                 if not self._closed:
                     self._acquired.remove(placeholder)
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -390,6 +390,13 @@ class BaseConnector(object):
                 if self._closed:
                     proto.close()
                     raise ClientConnectionError("Connector is closed.")
+            except:
+                # signal to waiter
+                waiters = self._waiters[key]
+                if waiters:
+                    waiter = waiters.pop(0)
+                    waiter.set_result(None)
+                    raise
             finally:
                 if not self._closed:
                     self._acquired.remove(placeholder)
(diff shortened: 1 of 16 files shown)


aiohttp/connector.py
```diff
@@ -390,6 +390,13 @@ def connect(self, req):
                 if self._closed:
                     proto.close()
                     raise ClientConnectionError("Connector is closed.")
+            except:
+                # signal to waiter
+                waiters = self._waiters[key]
+                if waiters:
+                    waiter = waiters.pop(0)
```
fafhrd91: I don't think this is required, this could cause IndexError on line 379

fafhrd91: maybe you should just call `_release_waiter`?

fafhrd91:
I don't think you can use `Condition`, semaphore maybe. but that is so heavy weight solution

aiohttp/connector.py
asvetlov: Yep, figured it out after writing tests.
Please wait for update -- polishing test suite

asvetlov:
Well, let's keep complicated synchronization logic for sake of speed.
Really I don't want to add such intrusive change just before a release.

asvetlov:
@fafhrd91 unfortunately `self._release_waiter()` leads to hung.
I should go.
Could you pick up the PR?
Otherwise I'll continue working on it tomorrow.

asvetlov:
Nevermind, new implementation passes all tests

Issue #2330, opened 2017-10-16
Propagating match_info in make_mocked_request


argaen:
When using `make_mocked_request` I find myself doing many times the following:

```
request = make_mocked_request(...)
request._match_info = UrlMappingMatchInfo({'whatever': 'you want', mock.Mock())
await my_handler(request)
```

Would it make sense that users could pass `match_info` into the `make_mocked_request` call so this is done inside? Right now its being set as empty https://github.com/aio-libs/aiohttp/blob/729eaff8d7e553658c1b7d8ce2cea99b29bc69b8/aiohttp/test_utils.py#L567 so it shouldn't be a big change

asvetlov:
+1

asvetlov:
Fixed by #2331

PR #2331, opened 2017-10-16
Propagate match_info in make_mocked_request

argaen:
https://github.com/aio-libs/aiohttp/issues/2330

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -565,7 +565,8 @@ def make_mocked_request(method, path, headers=None, *,
                   protocol, payload_writer, task, loop,
                   client_max_size=client_max_size)
 
-    match_info = UrlMappingMatchInfo({} if match_info is sentinel else match_info, mock.Mock())
+    match_info = UrlMappingMatchInfo(
+        {} if match_info is sentinel else match_info, mock.Mock())
     match_info.add_app(app)
     req._match_info = match_info
 
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -494,6 +494,7 @@ def _create_transport(sslcontext=None):
 
 
 def make_mocked_request(method, path, headers=None, *,
+                        match_info=sentinel,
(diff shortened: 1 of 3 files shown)


aiohttp/test_utils.py
```diff
@@ -564,7 +565,7 @@ def timeout(*args, **kw):
                   protocol, payload_writer, task, loop,
                   client_max_size=client_max_size)
 
-    match_info = UrlMappingMatchInfo({}, mock.Mock())
+    match_info = UrlMappingMatchInfo({} if match_info is sentinel else match_info, mock.Mock())
```
asvetlov: Isn't the line too long?

argaen: didn't give any problems to me, waiting for the automated checks to pass

argaen:
Added the doc about the new attribute but the docs of that method are with sphinx format. I'll check how do display that info in that format. Have to leave for 1/2 hours now.

asvetlov:
NP, I've polished doc.

FYI: http://aiohttp.readthedocs.io/en/stable/contributing.html#documentation

aiohttp/test_utils.py
asvetlov: You didn't run `make test`, travis blames exactly the line

argaen: Yup, didn't run because was failing on my machine for some reason, that's why I was letting travis do the job.

In case you had this before, that's what I'm getting: `ValueError: option names {'--fast'} already added`

argaen: uninstalling `pytest-aiohttp` worked for me

asvetlov:
You shouldn't install `pytest-aiohttp` for developing aiohttp itself because it depends on `aiohttp`.
The test suite setups `pytest-aiohttp` fixtures in `conftest.py` without plugin installation, that's how things work.

argaen:
Makes sense. I think I had it installed because of reusing venvs between aiohttp and pytest-aiohttp

PR #2332, opened 2017-10-17
http parser non strict mode

popravich:
## What do these changes do?

Build http-parser in non-strict mode.

By default http-parser was built in strict mode.
This was causing requests having non-ascii URLs in request line to be
rejected with `InvalidURLErrors`.
Although it is considered to be invalid HTTP request most servers handle such requests.
So it is decided to **build http-parser in non-strict mode by default**.

(more discussion [here on gitter](https://gitter.im/aio-libs/Lobby?at=59d34704614889d4756b42c0))

This PR allows to control http-parse mode by forwarding `HTTP_PARSER_STRICT` environment
variable to extension compiler.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,9 +34,6 @@ check_changes:
 	@./tools/check_changes.py
 
 .develop: .install-deps $(shell find aiohttp -type f) .flake check_changes
-	# rebuild http-parser extention depending on environ flags
-	@HTTP_PARSER_STRICT=$(HTTP_PARSER_STRICT) python setup.py build_ext \
-		--inplace --force
 	@pip install -e .
 	@touch .develop
 
@@ -58,10 +55,6 @@ cov-dev: .develop
 cov-ci-no-ext: .develop
 	@echo "Run without extensions"
 	@AIOHTTP_NO_EXTENSIONS=1 py.test --cov=aiohttp tests
-cov-ci-http-parser-strict: HTTP_PARSER_STRICT = 1
-cov-ci-http-parser-strict: .develop
-	@echo "Run with http-parser built with strict mode"
-	@py.test --cov=aiohttp --cov-append tests
 cov-ci-aio-debug: .develop
 	@echo "Run in debug mode"
 	@PYTHONASYNCIODEBUG=1 py.test --cov=aiohttp --cov-append tests
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,9 @@ check_changes:
 	@./tools/check_changes.py
 
 .develop: .install-deps $(shell find aiohttp -type f) .flake check_changes
+	# rebuild http-parser extention depending on environ flags
+	@HTTP_PARSER_STRICT=$(HTTP_PARSER_STRICT) python setup.py build_ext \
+		--inplace --force
 	@pip install -e .
 	@touch .develop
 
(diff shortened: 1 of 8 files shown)


asvetlov:
Do you want to hop on the bus at last second?
Ok, I'll wait for a while. The change looks pretty important.

popravich:
Well, I just don't know where in docs should I put a note on this or a news in changes is enough?..

asvetlov:
Hmm, strict mode requires compilation from sources, isn't it?
We will not ship wheels in strict mode, right?

popravich:
True

popravich:
Dropped unneeded control flags.

Makefile
```diff
@@ -50,7 +50,7 @@ cov-dev: .develop
 	@echo "Run without extensions"
 	@AIOHTTP_NO_EXTENSIONS=1 py.test --cov=aiohttp tests
 	@py.test --cov=aiohttp --cov-report=term --cov-report=html --cov-append tests
-        @echo "open file://`pwd`/coverage/index.html"
+	@echo "open file://`pwd`/coverage/index.html"
```
asvetlov, line 53: Good catch!
I was wondering why I don't see the line in console output :)

changes/2332.feature
```diff
@@ -0,0 +1 @@
+Build http-parser extension in non-strict mode by default.
```
kxepal, line 1: I'm not sure we want non-strict mode by default.

kxepal: Also, does this switch implies that we have to ship two wheels with strict and non strict parsers?

kxepal:
What's the behavior of pure Python parser btw?

changes/2332.feature
asvetlov: Why not?
We are strict on sending HTTP data but relaxed on receiving the data from foreign servers.
Basically we do the same in very many places.

asvetlov:
IMHO it works in relaxed mode.
@popravich ?

changes/2332.feature
kxepal: According the sources, this check does a bit more than just allow non-ascii characters in URL. If I read correctly, users may send "GET / OLOLO/1.1" requests without worry about, servers may return malformed status codes, etc. I don't think we want to support this.

popravich:
pure-python parser works in relaxed mode

changes/2332.feature
popravich: `GET / OLOLO/1.1` does not work — have just checked this.

popravich:
Update the test to use all parsers

changes/2332.feature
kxepal: Hm, then I read it wrong way. What exact it controls then? Can't found any documentation about.

popravich: I was looking at tests: https://github.com/nodejs/http-parser/blob/05525c5fde1fc562481f6ae08fa7056185325daf/test.c#L766-L811

asvetlov: After brief overview I've found the following behavior changes for non-strict mode.

It allows 
1. spaces, tabs and linefeeds in headers
2. utf-8 in URLs
3. underscore in hostname
4. `\9` (ht) and `\12` (np) symbols. 

Also looks like it has relaxed checks for `\r\n` (`\r` and `\n` are allowed).
"GET / OLOLO/1.1" if forbidden but "GET / HLOL/1.1" is allowed.

kxepal: @asvetlov Thanks! 

Are we ok with 1, 3 and 4?

asvetlov: I think yes. It doesn't raise security issues IMHO

asvetlov:
@fafhrd91 what is your opinion? Is it ok to ship C parser in relaxed mode?

fafhrd91:
I think it is fine

asvetlov:
Thanks @popravich

PR #2334, opened 2017-10-17
Sync in-docker wheels creation w/ best practices

webknjaz:


tools/build-wheels.sh
```diff
@@ -26,28 +31,35 @@ done
...
+
+echo
+echo
 echo "Install packages and test"
 echo "dist directory:"
 ls /io/dist
 
 for PYTHON in ${PYTHON_VERSIONS}; do
+    # clear python cache
+    find /io -type d -name __pycache__ -delete
```
asvetlov: Does `-delete` remove directories?
I thought it works for files only

webknjaz: Oops, you're right. I'll fix this

PR #2337, opened 2017-10-17
Fix deploy

asvetlov:
Fix tests running on built wheels, don't upload non-aiohttp wheels

### DIFF
--- a/tools/build-wheels.sh
+++ b/tools/build-wheels.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
+
 PYTHON_VERSIONS="cp34-cp34m cp35-cp35m cp36-cp36m"
 
 # Avoid creation of __pycache__/*.py[c|o]
@@ -42,6 +44,7 @@ for PYTHON in ${PYTHON_VERSIONS}; do
     echo
     echo -n "Test $PYTHON: "
     /opt/python/${PYTHON}/bin/python -c "import platform;print(platform.platform())"
+    /opt/python/${PYTHON}/bin/pip install -r /io/requirements/ci-wheel.txt
     /opt/python/${PYTHON}/bin/pip install "$package_name" --no-index -f file:///io/dist
     /opt/python/${PYTHON}/bin/py.test /io/tests
 done
--- a/tools/build-wheels.sh
+++ b/tools/build-wheels.sh
@@ -47,4 +47,7 @@ for PYTHON in ${PYTHON_VERSIONS}; do
     /opt/python/${PYTHON}/bin/pip install -r /io/requirements/ci-wheel.txt
     /opt/python/${PYTHON}/bin/pip install "$package_name" --no-index -f file:///io/dist
     /opt/python/${PYTHON}/bin/py.test /io/tests
+
+    # clear python cache
+    find /io -name __pycache__ | xargs rm -rf
 done
--- a/tools/run_docker.sh
+++ b/tools/run_docker.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
+
 package_name="$1"
 if [ -z "$package_name" ]
 then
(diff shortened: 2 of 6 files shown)


tools/run_docker.sh
```diff
@@ -17,7 +19,7 @@ do
 done
 
 echo Creating dist folder with privileges of host-machine user
-mkdir dist  # This is required to be created with host-machine user privileges
+mkdir -p dist  # This is required to be created with host-machine user privileges
```
webknjaz, line 22: This is not needed. `--path` argument only takes effect if you create subpaths, like `dist/subfolder` and `dist` doesn't exist. This won't change anything for a single atom.

tools/run_docker.sh
```diff
@@ -33,3 +35,5 @@ do
 
     dock_ext_args=""  # Reset docker args, just in case
 done
+
+find dist -type f -not -name "*aiohttp*" | xargs rm -f
```
webknjaz, line 39: I've covered this with https://github.com/aio-libs/aiohttp/pull/2334/files#diff-efcf4fcbe5802c1378b182a49aaf85e8R40 (tested with other repo, but should also work)

tools/build-wheels.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
```
webknjaz, line 2: https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail

webknjaz: `set -euxo pipefail`

webknjaz: Even better:
```bash
if [ -n "$DEBUG" ]
then
  set -x
fi

set -euo pipefail
```

tools/run_docker.sh
asvetlov: I've added it for running `./tools/run_docker.sh "aiohttp"` locally.
On travis the change makes no harm but I've been tired to run `rm -rf dist` before `run_docker.sh` every time.

tools/run_docker.sh
asvetlov: Maybe you mean another link -- but let's keep the line as is for sake of quick merge.
You can polish the line later.
I'm striving to release aiohttp 2.3 ASAP

tools/build-wheels.sh
asvetlov: Do you mean I should change it to `set -euxo pipefail`?
Sounds good but please left it for next PRs.
I hope you'll add the proposal into your work on travis polishing.
Right now I want to fix travis deployment for getting 2.3 release ASAP.
Every change requires a waiting for travis for completion.

tools/run_docker.sh
kxepal: find has `-delete` option, no need to pipe to xargs.

webknjaz: yes, but the match expression is a bit more complete :)

tools/build-wheels.sh
webknjaz: sure

tools/run_docker.sh
webknjaz: Yea, I see. But I'm talking about the `-p` argument :)

tools/run_docker.sh
webknjaz: `find /io/dist -maxdepth 1 -type f ! -path '/io/dist/'"$package_name"'-*-manylinux1_${arch}.whl' -exec rm -fv {} \;` @asvetlov I'm referring to this line

tools/run_docker.sh
asvetlov: I'm also talking about `-p`.
With it `mkdir` doesn't create `dist` folder if it already exists.

PR #2345, opened 2017-10-19
create MimeType object

jairhenrique:


asvetlov:
Makes sense.
The change looks safe
Moreover MIME types parsing is not exposed as our public API at all.

asvetlov:
thanks

Issue #2343, opened 2017-10-18
aiohttp 3.0 release


asvetlov:
I'd like to name the next release 3.0
3.0 means the major.

Let's drop Python 3.4 and use `async/await` everywhere.
Another important question is what 3.5 release to cut off?

Honestly I want to support `3.5.3+` only: the release fixes well known ugly bug for `asyncio.get_event_loop()`.
Debian stable has shipped with Python 3.5.3, not sure about RHEL.
All other distributions with faster release cycle support 3.5.3 too at least or event 3.6.

The transition should not be done at once. 
1. Let's pin minimal required Python version first in `setup.py`. 
2. Translate test suite to use `async/await`.
3. Modify aiohttp itself to use new syntax.

Every bullet except number one could be done in a long series of PRs, part by part.

BTW at the end I hope to get minor performance increase :)

samuelcolvin:
`async/await` everywhere. :champagne: :fireworks:

kxepal:
I think at 3.0 time it would better to drop even 3.5 support.

fafhrd91:
+1

We should drop all depredations. Convert write method for response and send method for websockets to real coroutines

fafhrd91:
@kxepal what we can simplify by dropping 3.5?

samuelcolvin:
> @kxepal what we can simplify by dropping 3.5?

* async generators are way simpler with just 3.6
* f-strings can make formatted strings much cleaner
* less confusion about which patch of 3.5 is support and which not. Simply ">=3.6".

The show stopper I guess is Debian, I don't know when it will upgrade to 3.6.

asvetlov:
> We should drop all deprecations. Convert write method for response and send method for websockets to real coroutines

+100. But let's be careful about dropping everything. Let's look on every deprecation one by one. 
Sure, we have a deprecation warning for old-styled middlewares for example.
Removing the support for these middlewares just in two months after introduction is too rude.

samuelcolvin:
on old middleware: I agree with have to keep it for longer.

kxepal:
@samuelcolvin 
I wonder if we should count one of Linux distributive as stopper for us. So far, we don't target to any specific Linux distro support AFAIK.

asvetlov:
@samuelcolvin @kxepal I would love to drop 3.5 at all but one of our major downstreams is https://home-assistant.io/
@balloob asked me to keep 3.4 support until RaspberryPi will have 3.5 in batteries by default.
[...]

PR #2350, opened 2017-10-19
moving to async/await syntax

eteamin:
## What do these changes do?

Changed some files to use async/await syntax.

## Related issue number

issue #2343

### DIFF
--- a/tests/test_classbasedview.py
+++ b/tests/test_classbasedview.py
@@ -13,50 +13,44 @@ def test_ctor():
     assert view.request is request
 
 
-@asyncio.coroutine
-def test_render_ok():
+async def test_render_ok():
     resp = web.Response(text='OK')
 
     class MyView(View):
-        @asyncio.coroutine
-        def get(self):
+        async def get(self):
             return resp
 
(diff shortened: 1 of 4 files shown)


asvetlov:
I suggest not converting everything at once but create a PR per file. Tests should go first, sure.
Reviewing single file is much easier than the whole project.

I'll tune CI bots to disable 3.4 this evening

samuelcolvin:
I understand why you say this, but doing a build for each file will take a loooong time and will also impact build times for all other aio-lib projects since travis throttles builds per organisation.

I spent about two hours yesterday waiting for aiohttp builds to finish to aiohttp-devtools builds would start.

webknjaz:
@samuelcolvin normal push builds are relatively fast: they don't run OS X at all.

asvetlov:
Well, a PR for a 2-3 small files is an option -- but reviewing 10000 long diff is a nightmare.
Pretty sure everybody understands it very well.

samuelcolvin:
yes I agree, makes sense.

PR #2352, opened 2017-10-19
Changed some files to async/await syntax

eteamin:
issue #2350 
@asvetlov I will make the next PRs limited to 2-3 smalls files.


alxpy:
I think, first you should to remove this https://github.com/aio-libs/aiohttp/blob/master/setup.py#L95

oleksandr-kuzmenko:
I think, first you should to remove this https://github.com/aio-libs/aiohttp/blob/master/setup.py#L95 and edit this https://github.com/aio-libs/aiohttp/blob/master/setup.py#L113

asvetlov:
Left `setup.py` changes to me.

eteamin:
@asvetlov Is this mergable?

eteamin:
CI seems to be good :)

asvetlov:
thanks!

PR #2353, opened 2017-10-19
coverage -> htmlcov

asvetlov:
Rename to follow well known coverage defaults


setup.cfg
```diff
@@ -28,4 +28,4 @@ source = aiohttp, tests
 omit = site-packages
 
 [coverage:html]
-directory = coverage
+directory = htmlcov
```
samuelcolvin: surely this is now unnecessary?

asvetlov: Actually not

asvetlov:
Notes fixed.

PR #2360, opened 2017-10-20
Cleanup ABC

asvetlov:
Use `async def`.


jettify:
Changes looks good, but do not know if you want to add period to other doc string in file.

PR #2361, opened 2017-10-20
test_client_request.py and test_client_response.py now use async/await syntax

eteamin:
issue #2343


eteamin:
by the way, shouldn't https://github.com/aio-libs/aiohttp/blob/master/tests/test_client_request.py#L965 be a coroutine?

asvetlov:
Good catch!
Yes, it should.
The code was working because the old styled coroutines are just generators technically -- but if you touch the file please fix it.

Also would you add your name to CONTRIBUTORS.txt? I'd like to see you in the list.

eteamin:
I fixed it. Also added my name to CONTRIBUTORS.txt. thanks for your consideration.

PR #2364, opened 2017-10-20
test_client_ws.py and test_client_ws_functional.py now use async/await syntax

eteamin:
issue #2343


eteamin:
There is something about [this](https://github.com/eteamin/aiohttp/blob/e254e3c13b60c0bd037b223735974a47b299c1bb/tests/test_connector.py#L381). If I change the inner `coro`  to `async coro()`, `test_tcp_connector_use_dns_cache_disabled()` will fail due to `Runtime Error` for awaiting a coroutine twice [here](https://github.com/eteamin/aiohttp/blob/e254e3c13b60c0bd037b223735974a47b299c1bb/tests/test_connector.py#L429). I found some info [here](https://bugs.python.org/issue25887) but I couldn't find a workaround to solve it.

asvetlov:
Tests are passed correctly.
Thanks.

asvetlov:
@eteamin JFYI I've fixed by `test_tcp_connector_use_dns_cache_disabled()` 146a660a

PR #2373, opened 2017-10-22
test_multipart.py now uses async/await syntax

eteamin:
issue #2343


eteamin:
I did!

asvetlov:
There are many other `yield from` expressions left in `test_multipart.py`.
The file is slightly not accurate, not all coroutines are marked by `@asyncio.coroutine` decorator -- but all functions containing `yield from` should be converted into `async def` with replacement `yield from` to `await`.

asvetlov:
Yep, too many commits but I could live with it for now.
For next PR sync your clone with upstream.
Or just make a new clone.

Issue #2121, opened 2017-07-21
Split async server running functions


txomon:
## Long story short

The requirement to have `run_app` as only execution of the server is not enough for some applications. It would be nice if this run_app functionality could be divided in two functions, `start()` and `cleanup()`, as many other libraries do. After all aiohttp is not the core of my app.

The idea is I want to be in charge of the main loop and control how and when it runs.

## Expected behaviour

Not to modify existing applications when including aiohttp.

## Actual behaviour

`aiohttp` get's to be the centre of the application by running `run_forever()` inside the only provided function to run the web server part.

## Steps to reproduce

1. Write `asyncio` application
2. Try to include web interface using aiohttp
3. Modify application to be run with aiohttp as the centre of it, synchronously.

arthurdarcet:
aiohttp does not need to be the center of your app. Have a look at what the `run_app` *helper* is doing: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web.py#L412

You can copy some of that logic in your code, and trigger your `loop.run_forever` whenever you want. But of course, you should also handle some "process-related" issues on your own, such as handling the SIGTERM/INT signals to allow aiohttp to shutdown gracefully.

asvetlov:
Anyway keeping separate startup and shutdown functions in public api makes sense.
Honestly I cannot remember all needed steps without looking into source code

alexmnazarenko:
In my project, I've implemented AsyncServer class for the very same reasons.

Something like this:

```python
class AsyncServer:
    def __init__(...run_app params..)
    async def start()
    async def shutdown()
    async def wait()
    async def __aenter__()
    async def __aexit__()
```

Works fine )

Something like this will be nice to have right in the library.

gjcarneiro:
I have to add my +1 to fixing this.  IMHO, the best thing would be to have a `run()` coroutine method on `aiohttp.web.Application()`, so that my main app looks like (simplified):

```python

async def my_suff():
   ...

def main():
    loop = asyncio.get_event_loop()
    app = web.Application()
    ....
    loop.run_until_complete(
        asyncio.gather(
            app.run("0.0.0.0", 80),
            run_my_stuff(),
        )
    )
```
Now wouldn't that be lovely? ;-)

asvetlov:
No, it should be a separate class.
Think about running the same app on different hosts/ports and stopping the app.

gjcarneiro:
[...]

PR #2376, opened 2017-10-23
added run_async_app, fixes #2375

### LINKED ISSUE #2375: add a way to run an app asynchron
## Long story short

It would be nice to have an async / coroutine that will take an app and run it. This is very usefull if you have an long running that wants to expose a rest interface.

## Steps to reproduce

go to the web.py file and look for run function that runs an app asynchron

hwmrocker:
## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -425,12 +425,12 @@ def _make_server_creators(handler, *, loop, ssl_context,
             uris.append(str(base_url.with_host(host).with_port(port)))
     return server_creations, uris
 
-@asyncio.coroutine
-def run_async_app(app, *, host=None, port=None, path=None, sock=None,
-            shutdown_timeout=60.0, ssl_context=None,
-            print=print, backlog=128, access_log_format=None,
-access_log=access_logger, handle_signals=True, loop=None):
-    yield from app.startup()
+
+async def run_async_app(app, *, host=None, port=None, path=None, sock=None,
+                  shutdown_timeout=60.0, ssl_context=None,
+                  print=print, backlog=128, access_log_format=None,
+                  access_log=access_logger, handle_signals=True, loop=None):
+    await app.startup()
     try:
         make_handler_kwargs = dict()
         # if access_log_format is not None:
@@ -443,7 +443,7 @@ access_log=access_logger, handle_signals=True, loop=None):
             loop=loop, ssl_context=ssl_context,
             host=host, port=port, path=path, sock=sock,
             backlog=backlog)
-        servers = yield from asyncio.gather(*server_creations, loop=loop)
+        servers = await asyncio.gather(*server_creations, loop=loop)
 
         if handle_signals:
             try:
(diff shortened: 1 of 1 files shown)


aiohttp/web.py
```diff
@@ -425,6 +425,50 @@ def _make_server_creators(handler, *, loop, ssl_context,
             uris.append(str(base_url.with_host(host).with_port(port)))
     return server_creations, uris
 
+@asyncio.coroutine
+def run_async_app(app, *, host=None, port=None, path=None, sock=None,
+            shutdown_timeout=60.0, ssl_context=None,
+            print=print, backlog=128, access_log_format=None,
+access_log=access_logger, handle_signals=True, loop=None):
+    yield from app.startup()
```
oleksandr-kuzmenko: now python3.4 is dropped and you should to use async / await syntax

asvetlov:
How would you use `run_async_app()`?
Please provide a small snippet. No need PR modification, a comment would be enough.

hwmrocker:
here is an example how i would run it:

```loop = asyncio.get_event_loop()
tasks = [
    asyncio.ensure_future(web.run_async_app(app, loop=loop)),
    asyncio.ensure_future(other_async_function()),
]
loop.run_until_complete(asyncio.gather(*tasks))
```

I will write some tests in the next days

aiohttp/web.py
```diff
@@ -426,6 +426,51 @@ def _make_server_creators(handler, *, loop, ssl_context,
     return server_creations, uris
 
 
+async def run_async_app(app, *, host=None, port=None, path=None, sock=None,
```
jettify, line 429: This code very similar what we have in `run_app`, should we somehow reuse this logic in `run_app`, example is not correct but still:
```python
def run_app(...):
    loop.run_until_complet(run_async_app(...))
```

samuelcolvin:
@hwmrocker can you give some more information about what "other_async_function" is and what it's doing? This is a code example, not really a use case.

My guess is that you would have a better time starting `other_async_function` as a task via `on_startup`.

jettify:
I use similar approach for testing, run server in background while hitting it with client.

https://github.com/aio-libs/aiozipkin/blob/a1a239d6f5a42fce35ecc9810c09eb4ac1d89780/examples/aiohttp_example.py#L45-L66

jettify:
Other use case, I have 5 services I want to run all of them in one event loop locally to simplify testing, 
https://github.com/aio-libs/aiozipkin/blob/a1a239d6f5a42fce35ecc9810c09eb4ac1d89780/demos/microservices/runner.py#L1-L33

Otherwise I have to run each service in separate thread or copy paste `run_app` in order make startup/shutdown logic work.

asvetlov:
Well, I understand the motivation pretty clean but asked about not only spawning a new task but it cancellation also.
https://github.com/aio-libs/aiohttp/pull/2376/files#diff-8e4ccf3aaf43100064462a4df10952e5R460 looks pretty incorrect: `KeyboardInterrupt` will never appear.
I pretty sure making a class for handling all gory details as proposed by #2121 is much better solution.

@hwmrocker and others: would you change your vision by providing a new PR?
I remember @cecton was interesting on the feature too.

The issue is in my priority list anyway but I would be very happy if somebody make it done.

cecton:
@asvetlov the motivation is unclear to me so I asked on #2375 for details.

To me it seems that the web service is the long running task of your process and your process should end with it. Unless you have a scenario where you want to have your process continue running without the web service, I think you can do everything with aiojobs (or without) and startup/clean hooks.

PR #2401, opened 2017-10-25
WIP: Add a better example for file upload

barrachri:
## What do these changes do?

Improve docs.

I think most of the time when you upload a file you also have related fields.

This PR add a general case to the docs., where you have a text field and file field.

No.

## Checklist

### DIFF
--- a/docs/web.rst
+++ b/docs/web.rst
@@ -743,15 +743,20 @@ should use :meth:`Request.multipart` which returns a :ref:`multipart reader
 
         # /!\ Don't forget to validate your inputs /!\
 
-        mp3 = await reader.next()
+        # reader.next() will `yield` the fields of your form
 
-        filename = mp3.filename
+        field = await reader.next()
+        assert field.name == 'name'
+        name = await field.read(decode=True)
 
+        field = await reader.next()
+        assert field.name == 'mp3'
+        filename = field.filename
         # You cannot rely on Content-Length if transfer is chunked.
         size = 0
         with open(os.path.join('/spool/yarrr-media/mp3/', filename), 'wb') as f:
             while True:
-                chunk = await mp3.read_chunk()  # 8192 bytes by default.
+                chunk = await field.read_chunk()  # 8192 bytes by default.
                 if not chunk:
                     break
                 size += len(chunk)
@@ -789,10 +794,10 @@ with the peer::
         print('websocket connection closed')
 
         return ws
-    
-The handler should be registered as HTTP GET processor::    
-    
-    app.router.add_get('/ws', websocket_handler)    
+
+The handler should be registered as HTTP GET processor::
+
+    app.router.add_get('/ws', websocket_handler)
 
 .. _aiohttp-web-websocket-read-same-task:
 
--- a/CHANGES/2401.doc
+++ b/CHANGES/2401.doc
@@ -1,0 +1,1 @@
+Small improvement docs: better example for file uploads.


asvetlov:
Hmm. The code is definitely robust but the snippet goes to very low details maybe.
@kxepal please review.

barrachri:
Thanks @asvetlov.

Do you think is too verbose or more comments/explanations outside of the code would improve this?

fafhrd91:
I’d add note in file operations, in general case file operations block

docs/web.rst
```diff
@@ -737,28 +742,49 @@ that :meth:`Request.post` reads the whole payload in memory, resulting in possib
...
 <aiohttp-multipart>`::
 
-    async def store_mp3_handler(request):
-
-        reader = await request.multipart()
-
-        # /!\ Don't forget to validate your inputs /!\
+    async def store_multipart_handler(request):
+        """Handle a multipart form."""
+        if request.method == 'POST':
```
kxepal: I think this is a bad pattern to recommend. You should have this condition on router level.

docs/web.rst
```diff
@@ -737,28 +742,49 @@ that :meth:`Request.post` reads the whole payload in memory, resulting in possib
...
+                            content_type.startswith('text/'):
+                        charset = field.get_charset(default='utf-8')
+                        value = value.decode(charset)
+                    data[field.name] = value
+                    size += len(value)
+
+            return web.Response(text='Thanks {} for uploading {}'
+                                .format(data['name'], data['file']))
+
+        return web.Response(body=page, content_type="text/html")
```
kxepal: page is not defined.

kxepal:
@fafhrd91 
Good point. However, that's not the only edge case for this example.

barrachri:
So:

* Add a note that file operations generally block.

@kxepal a small note in the official docs completed with a link to a good example in `/examples`  would be better?

asvetlov:
I think we need the docs change but it should look like:
```python
    async def store_mp3_handler(request):

        # /!\ Don't forget to validate your inputs /!\

        field = await reader.next()
        assert field.name == 'name'
        name = await field.read(decode=True)

        field = await reader.next()
        assert field.name == 'mp3'
        filename = field.filename
        # You cannot rely on Content-Length if transfer is chunked.
        size = 0
        with open(os.path.join('/spool/yarrr-media/mp3/', filename), 'wb') as f:
            while True:
                chunk = await field.read_chunk()  # 8192 bytes by default.
                if not chunk:
                    break
                size += len(chunk)
                f.write(chunk)

        return web.Response(text='{} sized of {} successfully stored'
                                 ''.format(filename, size))
```

The key is using `field.name` and `field.read()`.
Maybe we need process `field.get_charset()` too.

@kxepal does it make sense?

kxepal:
@barrachri 
Yes.

@asvetlov 
That's nice. I believe you would like to use `name`  instead of `filename` on file open line. Otherwise it's unused and may confuse.

asvetlov:
I have no strong opinion.
My intention is showing all useful field methods and attributes

barrachri:
The current example and @asvetlov's are very clear in showing how you can chunk files.

The problem is that usually in other frameworks you don't loop or iterate on an object to get your params and the same is also true for aiohttp, with `Request.query` and `data = await request.post()`.

My problem was that you jump from this example

```python
    mp3 = data['mp3']

    # .filename contains the name of the file in string format.
    filename = mp3.filename

    # .file contains the actual file data that needs to be stored somewhere.
    mp3_file = data['mp3'].file

    content = mp3_file.read()
```

to this example

```python
async def store_mp3_handler(request):

    # /!\ Don't forget to validate your inputs /!\

    mp3 = await reader.next()

    filename = mp3.filename

    # You cannot rely on Content-Length if transfer is chunked.
    size = 0
    with open(os.path.join('/spool/yarrr-media/mp3/', filename), 'wb') as f:
        while True:
            chunk = await mp3.read_chunk()  # 8192 bytes by default.
            if not chunk:
                break
            size += len(chunk)
            f.write(chunk)

    return web.Response(text='{} sized of {} successfully stored'
                             ''.format(filename, size))
```
and you completely miss what's going on with `next` because there isn't a context anymore.
The short paragraph before the example talk about the multipart reader and if you follow the [link](http://aiohttp.readthedocs.io/en/stable/multipart.html#aiohttp-multipart) everything is discussed apart from `Reading Multipart Request`.

After I checked the implementation of `Request.post` everything was clear but I agree is too verbose to just paste the code.

I think that a simple comment will make the example more clear.
Something like this:

```python
    async def store_mp3_handler(request):

        # /!\ Don't forget to validate your inputs /!\

        # reader.next() will `yield` the fields of your form

        field = await reader.next()
        assert field.name == 'name'
        name = await field.read(decode=True)

        field = await reader.next()
        assert field.name == 'mp3'
        filename = field.filename
        # You cannot rely on Content-Length if transfer is chunked.
        size = 0
        with open(os.path.join('/spool/yarrr-media/mp3/', filename), 'wb') as f:
            while True:
                chunk = await field.read_chunk()  # 8192 bytes by default.
                if not chunk:
                    break
                size += len(chunk)
                f.write(chunk)

        return web.Response(text='{} sized of {} successfully stored'
                                 ''.format(filename, size))
```

asvetlov:
I think now is better. Please update the RP.
@kxepal ?

barrachri:
Shall I add also an example in `examples/` or just updating the PR?

kxepal:
Let's update `examples/` in another PR. It's not a blocker for docs improvement.

asvetlov:
Thanks!

PR #2403, opened 2017-10-26
convert tests files to async/await syntax

douglasfarinelli:
issue https://github.com/aio-libs/aiohttp/issues/2343


asvetlov:
thanks!

PR #2404, opened 2017-10-26
all tests now use async/await syntax

eteamin:
issue #2343 

BTW, can `__call__` be async? [here](https://github.com/aio-libs/aiohttp/blob/master/tests/test_web_middleware.py#L286)


asvetlov:
Yes, it can.
I've added these test lines :)

PR #2407, opened 2017-10-26
changed web_ws.py and worker.py to async/await syntax

eteamin:
issue #2343


asvetlov:
The issue is blocked by `payload_writer.drain()` call.

eteamin:
[this](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_ws.py#L94) remained the old style. If I change it to async/await style, `await payload_writer.drain()` raises `TypeError: object tuple can't be used in 'await' expression`

asvetlov:
thanks

PR #2411, opened 2017-10-26
Fix a middleware example in docs

ahui132:
## What do these changes do?
Fix a middleware example in `docs/web.rst`.

No

## Related issue number
No

## Checklist

### DIFF
--- a/docs/web.rst
+++ b/docs/web.rst
@@ -1043,7 +1043,7 @@ The following code demonstrates middlewares execution order::
        return response
 
    @web.middleware
-   async def middleware2(app, handler):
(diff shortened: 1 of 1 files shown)


samuelcolvin:
@asvetlov how strict are we on adding to `CHANGES` even for small changes like this?

asvetlov:
I think we don't need CHANGES update for such trivial PRs.
The behavior is not changed, no new features added.

samuelcolvin:
Makes sense. I'll merge when passed.

samuelcolvin:
Thanks!

PR #2413, opened 2017-10-26
Fix #2412: add faq section about sharing db connection

### LINKED ISSUE #2412: How to access db connection stored in app from subapp
## Long story short

I'm trying to find a way to access the database connection details stored in app (via dict-like interface) from the subapp.

## Expected behaviour

There must be some access method like subapp.get_root_app() so that I can then use that to get the database connections stored via subapp.get_root_app()['db']

## Actual behaviour

I can't find anything in the documentation that shows a way to do this. So I'm using the following approach:

```
app = web.Application()
app['db'] = db

# sub application setup
subapp = web.Application()
subapp['parent'] = app

# add subapp to main app
app.add_subapp('/sub/', subapp)
```
Then in the subapp's request handler, I'm doing request.app['parent']['db'] - it works when I know that the subapp is just one level below main/parent app. But what if there are nested subapps?

## Steps to reproduce

As shown above.

## Your environment

CentOS / aiohttp 2.3 / Nginx

asvetlov:
between main app and subapp

Fix for #2412

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -368,3 +368,16 @@ use :class:`aiohttp.DummyCookieJar`. If you need separate cookies
 for different http calls but process them in logical chains use single
 :class:`aiohttp.TCPConnector` with separate
 client session and ``own_connector=False``.
+
+
+How to access db connection stored in app from subapplication?
+--------------------------------------------------------------
+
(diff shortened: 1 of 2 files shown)


asvetlov:
Sure, I can merge the PR without approval -- but will wait for green card until tomorrow.

PR #2416, opened 2017-10-27
fix a TypeError where MultiDict keys should be either str or subclass…

rschalch:
fix a TypeError where MultiDict keys should be either str or subclass str

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -559,14 +559,14 @@ class BaseRequest(collections.MutableMapping, HeadersMixin):
 
                     ff = FileField(field.name, field.filename,
                                    tmp, content_type, field.headers)
-                    out.add(field.name, ff)
+                    out.add(str(field.name), ff)
                 else:
                     value = yield from field.read(decode=True)
                     if content_type is None or \
                             content_type.startswith('text/'):
                         charset = field.get_charset(default='utf-8')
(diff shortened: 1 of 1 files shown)


asvetlov:
But `field.name` is always a string, isn't it?

rschalch:
@asvetlov I was able to capture this output from a default pytest call:
```
  File "/home/rbs/Projects/aiohttp/aiohttp/web_protocol.py", line 416, in start
  File "/home/rbs/Projects/aiohttp/aiohttp/web.py", line 323, in _handle
  File "/home/rbs/Projects/aiohttp/tests/test_client_functional.py", line 1439, in handler
    data = await request.post()
  File "/home/rbs/Projects/aiohttp/aiohttp/web_request.py", line 569, in post
    out.add(field.name, value)
  File "multidict/_multidict.pyx", line 385, in multidict._multidict.MultiDict.add
  File "multidict/_multidict.pyx", line 346, in multidict._multidict.MultiDict._add
  File "multidict/_multidict.pyx", line 270, in multidict._multidict._str
TypeError: MultiDict keys should be either str or subclasses of str
```

asvetlov:
What is your web form and it values?

rschalch:
Sorry, that's a sign that if different values are passed to form, the validation was not supposed to be where I put them, I will close this PR.

PR #2418, opened 2017-10-27
web.py and test_utils.py now use async/await syntax

eteamin:
issue #2343


asvetlov:
The build has failed.
I suggest not touching problematic places -- I'll convert hard code myself.
Hopefully it relates to 20 places at max -- all others are trivial.
Let's do them first,

eteamin:
Sure. I shouldn't have dived so deep!

PR #2427, opened 2017-10-28
Globally exclude *.so and *.pyd files from dist

### LINKED ISSUE #2414: Build fails on Alpine when building in not a x86 architecture
## Long story short
When trying to build aiohttp version 2.3.1 in Alpine (ppc64le and s390x architectures) the build fails at Alpine stripping part because it tries to strip "_http_parser.cpython-36m-x86_64-linux-gnu.so" shared object that is shipped in source code and fail in any architecture different from x86.

I guess the "_http_parser.cpython-36m-x86_64-linux-gnu.so" should not be shipped in the source code

## Steps to reproduce
- Clone Alpine aports project
- build py3-aiohttp project (community/py3-aiohttp) with command `abuild -r`

## Your environment
Alpine Linux 3.6.2 (ppc64le or s390x)

webknjaz:
Fixes #2414.

P.S. I still feel the need to properly configure `setuptools-git` to avoid need to track files manually.


asvetlov:
The tricky thing about `setuptools-git` integration is inclusion of generated `*.c` files for Cython extensions.

webknjaz:
@asvetlov those could be easily included via `setup.py`

Issue #2313, opened 2017-10-11
ClientSession signals to track events related to each request.


pfreixes:
The current implementation of `ClientSession` does not provide a clean and straightforward way to track events related to each request.

Right now in our organization, we are tracking the time spent and the status code overriding the `_request` private method and implementing a customized `ClientSession`. The solution that we have implemented is far from being the best.

To address this situation, `ClientSession` could implement a new set of methods that will allow the user to instrumentalize different metrics related to each request.  The following is a list of new methods that will be used for this purpose:

* `on_pre_request` Executed before the request is sent.
* `on_post_request` Executed after the request is sent.
* `on_queued` Executed when the request is queued.
* `on_unqueued` Executed when the request is unqueued.
* `on_create_conn` Executed when a new connection is created.

We are planning on have support for three new metrics: number of TCP connections created, number of operations queued because the limit has been reached, time spent in the queue
The previous methods will allow us to track all of these metrics.

Any comments appreciate.

asvetlov:
Not methods but signals I guess.
Also we need to track request/response headers and bodies.

pfreixes:
Yes sorry, I meant signals.

Can you evolve that please: 

> track request/response headers and bodies.

pfreixes:
Current `Signal` implementation is coupled to the `App` interface [1]. I was wondering how changing the name of the class to `AppSignal` and implementing a more naive `Signal` class without the app dependency - and probably without the pre and post signals - might break the API.

First thoughts?


[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/signals.py#L20

jettify:
I was going to ask something similar for my tracing project https://github.com/aio-libs/aiozipkin other tools like DataDog, NewRelic, etc will benefit too. 

Just FYI popular java http client `okhttp` has two types of hooks: interceptors (can modify request, add retries and caching) and events (does not change anything).  Signals looks like events:
<img width="850" alt="events" src="https://user-images.githubusercontent.com/223335/31439505-0f2d195a-ae95-11e7-8695-27e13270299c.png">
https://github.com/square/okhttp/wiki/Events

So should we also add similar hooks? like on_dsn_start/stop, on_headers_send?

asvetlov:
There are at least three signal sets:
1. Request: *send headers* -> [ *send body part* ] -> *finish*.
2. Response: *receive headers* -> [ *receive body part* ] -> *finish*.
[...]

PR #2429, opened 2017-10-29
Client signals #2313

pfreixes:
## What do these changes do?

Implementation of the client signals exposed by the `ClientSession`
class, to get a list of the all signals implementation please visit
the documentation.

Yes

## Related issue number

#2313 

## Checklist


pfreixes:
List of signals implemented here: https://github.com/aio-libs/aiohttp/pull/2429/files#diff-7dd84b5ef8d5eea2de1dfc5329411dfcR695

jettify:
@asvetlov lets summarize what should be done to move this PR forward

1) move code _on_headers_received and _on_content_received to higher level
2) as for body stuff, is there any point where we can signal that full body consumed? I believe this approach will cover 90% of use cases. Other 10% we can address in next PRs

anything else?

pfreixes:
I had my doubts about the parsers and how to implement the signals triggered by sending or receiving some pieces of the HTTP protocol. The current implementation is just an intention or a POC of how other signals that have more granularity such as `on_content_chunk_sent` could be implemented, and I have my own rationale for that ( Obviously I can have it in a wrong way)

But I would prefer to get rid of this variable and try to put all of us on the same page. My proposal here is get rid of the `on_headers_sent`, `on_headers_received`, `on_content_sent`, `on_content_received` signals to be implemented later, and if we consider that they are a must implement them together with the `chunk` ones. It is gonna give us the freedom of focus our efforts right now in a set of signals that already give a user value.

What do you think  @asvetlov and @jettify ?

asvetlov:
Sounds good

pfreixes:
@asvetlov I think that your requested changes are no longer valid and the branch is ready to be either merged or moving on with other reviewing stuff.

aiohttp/client.py
```diff
@@ -218,6 +227,18 @@ def _request(self, method, url, *,
...
 
         url = URL(url)
+
+        if trace_context is None:
+            trace_context = SimpleNamespace()
+
+        yield from self.on_request_start.send(
+            trace_context,
+            method,
+            url.host,
```
asvetlov: The whole URL object maybe? Query part might be interested for tracer.

aiohttp/client.py
```diff
@@ -291,6 +315,9 @@ def _request(self, method, url, *,
                     # redirects
                     if resp.status in (
                             301, 302, 303, 307, 308) and allow_redirects:
+
+                        self.on_request_redirect.send(trace_context, resp)
```
asvetlov: What is intended usage for the signal?
How to figure out what initial request was redirected?
The same is true for all other signals.
Maybe we should always pass URL, method and headers?

aiohttp/client.py
```diff
@@ -654,6 +683,60 @@ def loop(self):
...
+
+    @property
+    def on_request_exception(self):
+        return self._on_request_exception
+
+    # connector signals
+
+    @property
+    def on_request_queued_start(self):
+        return self._connector.on_queued_start
```
asvetlov: I think every session should have own signals.
Connector might be shared between sessions by `connector_owner=False` or `session.detach()`.
Shared subscriptions make a mess: nobody know when recipient is dead.

aiohttp/connector.py
```diff
@@ -347,8 +354,12 @@ def closed(self):
         """
         return self._closed
 
-    async def connect(self, req):
+    async def connect(self, req, trace_context=None):
```
asvetlov: Let's pin `trace_context` to mandatory params set.
I'm ok with breaking backward compatibility in connectors API.

aiohttp/connector.py
```diff
@@ -347,8 +354,12 @@ def closed(self):
...
         """Get from pool or create new connection."""
+
+        if trace_context is None:
+            trace_context = SimpleNamespace()
```
asvetlov: We should not create an empty context here -- just pass a parameter as is.

aiohttp/connector.py
asvetlov: The same for other places with `trace_context=None` in signatures.

aiohttp/connector.py
```diff
@@ -722,15 +782,21 @@ def clear_dns_cache(self, host=None, port=None):
...
+    async def _create_connection(self, req, trace_context=None):
         """Create connection.
 
         Has same keyword arguments as BaseEventLoop.create_connection.
         """
         if req.proxy:
-            _, proto = await self._create_proxy_connection(req)
+            _, proto = await self._create_proxy_connection(
+                req,
+                trace_context=None
```
asvetlov: `trace_context=trace_context` I pretty sure.

aiohttp/connector.py
```diff
@@ -722,15 +782,21 @@ def clear_dns_cache(self, host=None, port=None):
...
+            )
         else:
-            _, proto = await self._create_direct_connection(req)
+            _, proto = await self._create_direct_connection(
+                req,
+                trace_context=None
```
asvetlov: `trace_context=trace_context`

aiohttp/client.py
```diff
@@ -218,6 +227,18 @@ def _request(self, method, url, *,
         handle = tm.start()
 
         url = URL(url)
+
+        if trace_context is None:
+            trace_context = SimpleNamespace()
```
asvetlov: I think the API is wrong here.
User will never send a `trace_context` into `async with session.get()` explicitly.
It's another level of abstraction.
What the user will do is setting up session properly on initialization stage by substribing on signals and (optionally) providing a *trace context factory* for creating a new container for user data.
I even doubt if we need a factory parameter, at least at current stage.

asvetlov:
New comments.
Sorry, I'm on conference these days, have very limited time for careful review.
There are other issues for discussion not covered by my notes yet.

aiohttp/client.py
pfreixes: The rationale behind this implementation is the following one:

Give the proper freedom to the developer to have a grain control of his requests calls building trace context for each request.

Perhaps

```python
async def on_request_start(trace_context, host, port, headers):
    trace_context['start'] = loop.time()

async def on_request_end(trace_context, resp):
    await send_metrics(
        time=loop.time() - trace_context['start']
        query=trace_context['query']
    )


sesion = ClientSession()
session.add_on_requests_start(on_request_start)
session.add_on_requests_end(on_request_end)

resp = session.get("http://localhost?query=foo", trace={'query':'foo'})
resp = session.get("http://localhost?query=foo", trace={'query':'bar'})
```

This example shows how the same ClientSession is used to make different queries that might have divergent traces. 

In case the user is keen on share information between all requests that belong to the same `ClientSession` might use a closure pattern, perhaps:

```python
def on_request_end(query):
    async def on_request_end(trace_context, resp):
        await send_metrics(
            time=loop.time() - trace_context['start']
            query=query
        )
    return on_request_end

sesion = ClientSession()
session.add_on_requests_start(on_request_start)
session.add_on_requests_end(on_request_end(query='foo'))
```

I can see that your point about forcing the user to populate each request call can be less kindy, but from my experience have the way to pass a context that has information about the current execution is a must. Also, take into account that having this granularity allows the user to implement Session or Requests context.

Another solution would pass for implement two different contexts, one for the session and another one for the request. But IMHO overcomplicates right now the implementation.

Thoughts ?

aiohttp/client.py
pfreixes: Yes, I will do that

aiohttp/client.py
pfreixes: The idea was to give a way to trace when a redirect happens. Completly agree that the signal parameters are not enough consistency to give enough information to the user, lets add the `method`, `URL`, `headers`.

What do you mean with :: ```The same is true for all other signals.```

aiohttp/client.py
pfreixes: Oks I can see right now the idea of the `connector_owner` parameter, my fault I havent check it and it head me to a wrong implementation. If at last the connector is shared between sessions its absolutely necessary don't mess the signals.

In another way worries me a bit the `connector_owner` implementation, leaving to the user's hand the power of making it `True` or `False`, meaning that in case the user gives an alternative connector but forget to pass the `connector_owner` set as `False`, when the session is gonna be closed this automatically will close the connections of the connector that might be shared by another `Session`. 

Should the `connector_owner` param handled internally by the `Session` object` ?

aiohttp/connector.py
pfreixes: Oks.

aiohttp/connector.py
pfreixes: Okis.

aiohttp/client.py
pfreixes: I will do the same with the `on_request_end`, and `on_request_exception` allowing the user to know the source URL in case the request came from a redirect.

pfreixes:
Something that I don't like about the current proposal after applying two comments regarding your concerns::

1) Making the `trace_context` an arg instead of a `kwarg` breaks many function signatures such as `connect`, `direct_connect` and so on. And also IMHO makes the API of at least the `Connector` and its derivates dirtier. The same will happen with other classes that are used by the `ClientSession` and have to use signals to trace a specific event related to a specific request.
2) Removing the signals from the `Connector` and putting them together into the `ClientSession` forces us to add another param, the session, that will even make dirtier these functions. The same will happen with other classes.

To fix the second issue, trying to minimize the number of parameters needed I will be keen on creating an object that stores all of the dependencies needed to send a signal, perhaps::

```python
class Trace:
    def __init__(self, session, trace_context):
        self._session = session
        self._trace_context = trace_context

    def on_connect_start(self, host):
        self._session.on_connect_start.send(self._trace_context, host)

def connect(self, host, trace):
    trace.on_connect_start.send(host)
    trace.on_connect_start.send(host)
```

The first issue, that was other the concerns that you expressed @asvetlov, related to make the `trace` argument mandatory, I will be still more inclined to make them optional having, as a result, the following snippet::

```python
def connect(self, host, trace=None):
    if trace:
        trace.on_connect_start.send(host)
    if trace:
        trace.on_connect_start.send(host)
```

The `Trace` function won't be a first citizen class that could be created per each request in the `ClientSession._request` function:

```pyhton
def _request(self, ..)
    trace = Trace(self, trace_context)
    self.connector.connect(trace=trace)
```

I'm still wondering if it's the best implementation have the signals attached to `ClientSession`, and if its worth it to move them to another object called `TraceRequest`, passing this object  to the `ClientSession` at constructor time.  perhaps::

```python
trace_request = TraceRequest()
trace_request.on_request_start.append(my_function)
session = ClientSession(..., trace_request=trace_request)
```

When the developer wants to trace a specific client session gives it as a parameter, meanwhile the default behavior doesn't trace the requests. It will allow us to use the `freeze` method of the signals, once the `TraceRequest` instance is given to `ClientSession` all signals belonging to that specific `TraceRequest` instance are frozen.

Worth mentioning that having the signals to trace a request decoupled from the session, it forces us to pass the session instance as the first parameter for all of the signals.

Thoughts ?

pfreixes:
I couldn't help it and I wrote a POC of the last proposal changes without testing here [1], two things that I dont like it

1) The repeated pattern `if trace: trace.send(...)`, but its just a matter of cosmetic issue Indeed having this pattern, the performance impact when there is not tracing enabled should be the minimum.
2) The `trace.send('signal')` function, it breaks a bit the normal usage of signals.
3) Linked with the previous one, if in the future we have support for function signals we must force to refactorize the `Trace` object.

[1] https://github.com/pfreixes/aiohttp/commit/4caddd9a1ed22c941167d806dff397806d8375de

aiohttp/signals.py
```diff
@@ -24,6 +24,25 @@ class Signal(BaseSignal):
...
     """
 
+    __slots__ = ()
+
+    @asyncio.coroutine
+    def send(self, *args, **kwargs):
+        """
+        Sends data to all registered receivers.
+        """
+        yield from self._send(*args, **kwargs)
```
kxepal: async/await

docs/client.rst
```diff
@@ -747,6 +747,36 @@ If no redirects occurred or ``allow_redirects`` is set to ``False``,
...
+
+The execution flow of a specific request can be followed attaching listeners functions
+to the signals provided by the :class:`ClientSession` instance, for example the following
+snippet shows how the start and the end of the request flow can be followed::
+
+    async def on_request_start(trace_context, method, host, port, headers):
+        print("Starting request")
+
+    async def on_request_end(trace_context, resp):
+        print("Ending request")
```
kxepal: Wording bit: is `on_request_end` happens just before request is ended or when it's actually ended. `-ing` suffix confusing here.

aiohttp/client.py
```diff
@@ -354,15 +386,29 @@ def _request(self, method, url, *,
...
 
-        except Exception:
+        except Exception as e:
             # cleanup timer
             tm.close()
             if handle:
                 handle.cancel()
                 handle = None
 
+            yield from self.on_request_exception.send(
```
kxepal: await

aiohttp/client.py
```diff
@@ -354,15 +386,29 @@ def _request(self, method, url, *,
                     handle.cancel()
 
             resp._history = tuple(history)
+            yield from self.on_request_end.send(
```
kxepal: await

aiohttp/client.py
```diff
@@ -266,7 +286,10 @@ def _request(self, method, url, *,
                     # connection timeout
                     try:
                         with CeilTimeout(self._conn_timeout, loop=self._loop):
-                            conn = yield from self._connector.connect(req)
+                            conn = yield from self._connector.connect(
```
kxepal: await

aiohttp/client.py
```diff
@@ -218,6 +227,17 @@ def _request(self, method, url, *,
         handle = tm.start()
 
         url = URL(url)
+
+        if trace_context is None:
+            trace_context = SimpleNamespace()
+
+        yield from self.on_request_start.send(
```
kxepal: await

tests/test_client_session.py
```diff
@@ -474,3 +476,121 @@ def test_client_session_implicit_loop_warn():
...
+
+    yield from session.get('http://example.com')
+    assert MyClientRequest.headers['foo'] == 'bar'
+
+
+@asyncio.coroutine
+def test_request_tracing_proxies_connector_signals(loop):
+    connector = TCPConnector(loop=loop)
+    session = aiohttp.ClientSession(connector=connector, loop=loop)
+    assert id(session.on_request_queued_start) == id(connector.on_queued_start)
```
kxepal: Why not to do simpler `is` check?

tests/test_client_session.py
```diff
@@ -474,3 +476,121 @@ def test_client_session_implicit_loop_warn():
 
     asyncio.set_event_loop(None)
     loop.close()
+
+
+@asyncio.coroutine
```
kxepal: I guess you know what (:

pfreixes:
@asvetlov and @kxepal the way to implement the client signals have been rewritten at least the way are exposed to the user and how the user has to enable them into the ClientSession object, its based on the rationale exposed in this comment [1].

A good point of start is the documentation [2].

Disclaimer: I didn't take care of the awaitish work that has to be done in the `ClientSession` object taking into account that it might be done by another PR that will take care of it plus all of the tests.

[1] https://github.com/aio-libs/aiohttp/pull/2429#issuecomment-342659190
[2] https://github.com/aio-libs/aiohttp/pull/2429/files#diff-e5af816206d783c3bd169522ceecf99bR750

aiohttp/client.py
```diff
@@ -218,6 +226,22 @@ def _request(self, method, url, *,
...
+        if self._trace_config:
+            if trace_context is None:
+                trace_context = self._trace_config.trace_context()
+            trace = Trace(self._trace_config, self, trace_context)
+        else:
+            trace = None
+
+        if trace:
+            yield from trace.send(
+                'on_request_start',
```
kxepal: Can we avoid strings here and have `trace.on_request_start` and friends which actually does `trace.send`, but with proper event name (`'on_request_start'` in this case)?

pfreixes: The `Trace` acts as a proxy to call the signal with the proper arguments, aside of the arguments provided by the same signal such as `metod`, `url` and `headers` the `session` and the `trace_context` are sent implicitly by the `Trace.send` method. 

I came up with another implementation, see the following snippet, but I didn't like it::

```python
class Trace:
    ....

    async def on_request_start(self, *args, **kwargs):
        return await self._trace_config.on_request_start.send(
            self._session,
            self._trace_context,
            *args,
            **kwargs
        )
```

IMHO this implementation raises another issue, the class exposes a method `on_request_start` as it would be a typical signal, meanwhile is just a method that wrappers the real signal.

I know that `Trace` is not a first level class and we can do whatever we want but I have preferred to leave something consistency than having something almost equal but not consistency.

Anyway, I'm open to suggestions.

kxepal: I didn't get the issue with the snippet, could you elaborate a bit more, please?

pfreixes: The usual way that the signals are exposed in aiohttp follows the following pattern::

```python
class Foo:
    def __init__(self):
        self._on_some_event = Signal(self)

    @property
    def on_some_event(self):
        return self._on_some_event

f = Foo()
f.on_some_event.send(arg1, arg2)
```

The interface of the signal is fully exposed, its responsibility of the caller to call it using the `send` method and the arguments to trigger all listeners for a specific event.

The previous snippet shows how the signal is not fully exposed and the `send` method is hidden to allow the `Trace` object add the two first params. 

IMHO the snippet implementation is kinda inconsistent if you compare it with the traditional way of exposing signals. 

To avoid that misinterpretation I preferred to create an ad-hoc method called `send` for the `Trace` object.

asvetlov: What's about `Trace.send_request_start()` method instead of `on_request_start`?

kxepal: Good idea, Andrew!

pfreixes:
@kxepal and @asvetlov I've changed the way of signals are handled internally by the `Trace` and the `TraceConfig` object, just making them explicitly and allowing to me create the proper ad-hoc method suggested `send_name_of_signal`.

The code might be a bit verbose but at least the change leave it more readable.

pfreixes:
I'm thinking about changing the `trace_config` param for the `ClientSession` to an array, it will allow the client to install different `TraceConfig` instances, perhaps:

```pyhthon
from aws_xray_aiohttp import ClientConfigXRaySignals
from .mymodule import ClientConfigOpentTSDB

session = ClientSession(trace_configs=[ClientConfigXRaySignals(), ClientConfigOpentTSDB()]
```

asvetlov:
> The code might be a bit verbose but at least the change leave it more readable.

Readability beats code succinctness, at least for libraries.
Also it simplifies debugging.

asvetlov:
I don't like a list of traceconfig objects but support your idea in general.
We could provide a meta config class with `TraceConfig` interface: 
```
from aws_xray_aiohttp import ClientConfigXRaySignals
from .mymodule import ClientConfigOpentTSDB

session = ClientSession(trace_configs=TraceConfigComposite(ClientConfigXRaySignals(), ClientConfigOpentTSDB()))
```

I believe the proposal keeps internals simple but allows to specify multiple tracers easy.

pfreixes:
The good thing about the array is that it follows the same idea of the `middlewares`, and the implementation its pretty straightforward having only to modify the ad-hoc `Trace.send_<signa>()` functions iterating for all of the `TraceoConfig` instances.

@asvetlov ^^

asvetlov:
Do you want to support both single value and a list?

pfreixes:
IMHO doesnt matter, for the sake of API clearness we can force to give always a list, if not this is something that can be handled internally by the `ClientSession` init:

```python
self._trace_configs = trace_configs if isinstance(trace_configs, list) else [trace_configs]
```

asvetlov:
Ok

pfreixes:
@asvetlov and @kxepal more changes, now the `ClientSession` accepts a list of `TraceConfig` objects, allowing the user to install different trace config objects. To make so, the `trace_context` has become unique per `TraceConfig`. Therefore, to allow the user progress custom values from the `request()` method has been created a new parameter, in favor of the old one `trace_context`, called `trace_request_context`.

I think that the documentation reflects pretty well these changes [1], please read it and review it. Sorry for the many changes, but I'm an almost absolutely believe that these changes are necessary for the user.

[1] https://github.com/aio-libs/aiohttp/pull/2429/files#diff-e5af816206d783c3bd169522ceecf99bR750

asvetlov:
`docs/client.rst` was split up to `client_quickstart` and `client_advanced`. Most likely you need the second one or, even better, create a new `client_tracing.rst` for *Tracing Usage*. I suspect it is a complex thing, the document content will grow.
`docs/toc.rst` gas gone, add `tracing_reference` to `client.rst` `toctree`.

asvetlov:
Thanks for hard work @pfreixes 
Is it possible to rename `trace_request_context` -> `trace_context`?
The name is very long, I believe *request* application is obvious in the context of using the feature.

The PR is really very huge, I'm inclining to merge it after fixing very obvious changes and continue the work on client tracing in next PRs. Including a discussion for every supported signal parameters.

docs/client_reference.rst
```diff
@@ -294,6 +294,12 @@ The client session supports the context manager protocol for self closing.
 
          .. versionadded:: 2.3
 
+      :param trace_context: Object used to give as a param for the all signals
+        triggered by the ongoing request. Default uses the object returned 
+        by the :class:`TraceConfig.trace_context()` method.
```
asvetlov: :meth:`TraceConfig.trace_context` maybe. Parenthesis is not required by sphinx

docs/tracing_reference.rst
```diff
@@ -0,0 +1,156 @@
...
+   .. method:: trace_context()
+
+      Return a new trace context.
+
+   .. attribute:: on_request_start
+
+      Property that gives access to the signals that will be executed when a
+      request starts, based on the :class:`~signals.Signal` implementation.
+
+      The coroutines listening will receive as a param the `session`,
```
asvetlov: Spinx prefer double backticks, e.g. ``` `session` should be replaced by ``session`` ```.
On other hand CPython documentation uses `*session*` for parameters bu looks like in aiohttp docs we have not pronounced agreement to use double backticks.

tests/test_client_request.py
```diff
@@ -500,7 +500,8 @@ def test_gen_netloc_no_port(make_request):
         '012345678901234567890'
 
 
-async def test_connection_header(loop, conn):
+@asyncio.coroutine
+def test_connection_header(loop, conn):
```
asvetlov: Why removing `async def`?

pfreixes:
Just to put all of us in the same page, sure the large MR and many changes do not help. The trace_context naming is reserved for  the  TraceConfig objects that use this param to pass along the different signals a place to store internal information related to the same request and same TraceConfig

Meanwhile the trace_request_context is the param that the user can pass at the begining of the request that can be used at any signal.

My proposal for the sake of clearness will be:

- trace_context to trace_config_ctx
- trace_request_context to trace_request_ctx

Or other alternatives?

asvetlov:
`trace_config_ctx` and `trace_request_ctx` are perfect!

tests/test_client_request.py
pfreixes: My fault, fixed.

docs/tracing_reference.rst
pfreixes: Fixed, using `` as you proposed.

docs/client_reference.rst
pfreixes: no longer required, this documentation has been deprecated.

pfreixes:
Documentation accommodated to the new architecture, basically added the tracing as a new section of the `client_advanced` chapter and added a new chapter as `tracing_reference`.

More or less the PR is done to be Merged or move on with the reviewing.

Disclaimer: once the code is merged I will open a discussion about if there is any chance to make a port to the 2.X series for this feature to have later a new 2.4 release. The goal is to receive feedback to the community and improve the API for the 3.X version and also important make it usable by the developers right now.

kxepal:
🎉

asvetlov:
@pfreixes the PR has merged.
Thanks for hard work again.

Personally I doubt if we need 2.4 release:
1. We have no strong schedule for 3.0 but I hope to release it after New Year, definitely in January.
Not much time is left.
2. IMHO the proposal is not ready yet -- we need a careful review of tracing signals and, most important, their signatures.
3. Current master has dramatic changes: `async/await` and documentation refactoring. Keeping backports is nightmare in this situation.

asvetlov:
I support @kxepal emoji. Sorry, I'm not emoji jedi but share the delight of merging the proposal into master branch!

pfreixes:
Oks January is not so far away. But my intentions are starting to work with the support of AWS XRay for Aiohttp and also refactor one of the middlewares that instrumentalizes the calls made through ClientSession, using for both the new tracing system.

I hope that this is gonna give us a bit of feedback.

asvetlov:
Check a concept on real usage is always very useful.

PR #2435, opened 2017-10-29
drop support to python <3.5.3

jairhenrique:


aiohttp/helpers.py
```diff
@@ -219,14 +215,8 @@ def proxies_from_env():
...
-if PY_352:
-    def create_future(loop):
-        return loop.create_future()
-else:
-    def create_future(loop):  # pragma: no cover
-        """Compatibility wrapper for the loop.create_future() call introduced in
-        3.5.2."""
-        return asyncio.Future(loop=loop)
+def create_future(loop):
+    return loop.create_future()
```
hellysmile: You can remove this function at all and just use ‘loop.create_future’ everywhere

jairhenrique: :+1:

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 language: python
 
 python:
-- 3.5.2
+- 3.5.3
```
asvetlov: 3.5.2 was here to emphasize the difference between 3.5.2 and 3.5.3
Let's just drop redundant travis job. I could help to reduce travis build time.

jairhenrique: @asvetlov removed.

Issue #2424, opened 2017-10-27
Attempts to connect to multiple ips from dns response is broken


hellysmile:
## Long story short

If dns server return multiple ip records for same host, `aiohttp` should try to connect for all them one by one until connection is established

## Expected behaviour

`aiohttp` tries to connect to all ips one by one

## Actual behaviour

If first connection fails, non of the rest ips is used

It were broken at https://github.com/aio-libs/aiohttp/pull/2297

There will be fix after https://github.com/aio-libs/aiohttp/pull/2410 will be merged

asvetlov:
Fixed by #2447

PR #2447, opened 2017-10-30
Fix connection for multiple dns hosts

hellysmile:
## What do these changes do?

Fix for `aiohttp.connector.TCPConnector._create_direct_connection` tries to use all of dns hosts from dns response to connect one by one

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2424

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -807,36 +807,44 @@ class TCPConnector(BaseConnector):
             # it is problem of resolving proxy ip itself
             raise ClientConnectorError(req.connection_key, exc) from exc
 
+        last_exc = None
+
         for hinfo in hosts:
             host = hinfo['host']
             port = hinfo['port']
-            transp, proto = yield from self._wrap_create_connection(
-                self._factory, host, port,
-                ssl=sslcontext, family=hinfo['family'],
-                proto=hinfo['proto'], flags=hinfo['flags'],
-                server_hostname=hinfo['hostname'] if sslcontext else None,
-                local_addr=self._local_addr,
-                req=req, client_error=client_error)
+
+            try:
+                transp, proto = yield from self._wrap_create_connection(
+                    self._factory, host, port,
+                    ssl=sslcontext, family=hinfo['family'],
+                    proto=hinfo['proto'], flags=hinfo['flags'],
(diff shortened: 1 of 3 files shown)


pfreixes:
Yeps, Maybe Im reading in the wrong way but looks like you are trying to implement a kind of failover or circuit breaker pattern inside the logic of the `_create_direct_connection ` function.

IMHO this is a bad idea and it makes the API not predictable. What happen if we wanna log in our app each time that there is an error connecting to a specific IP address ? What to do with the errors at this level has to be done by upper layers, these layers that are in the domain of the developer.

hellysmile:
asyncio itself has this functionality. Aiohttp had it prior 2.3.0, where it was removed by mistake. I am trying to fix my own mistakes. I was talking with @asvetlov offline regarding logging of failed connections and decision was do not implement it, cuz asyncio do not log it

hellysmile:
Another solution instead of logging can be raise each new exception from previous one

pfreixes:
@hellysmile can you give me a link which shows how this is implemented by "asyncio" and another one that shows how it was implemented before by aiohttp?

hellysmile:
`loop.create_connection` https://github.com/python/cpython/blob/master/Lib/asyncio/base_events.py#L752

`aiohttp` 2.2
https://github.com/aio-libs/aiohttp/blob/2.2/aiohttp/connector.py#L733

pfreixes:
You are right, thanks for the info!

hellysmile:
@pfreixes can I ask You for a review?

aiohttp/connector.py
```diff
@@ -834,9 +842,15 @@ def _create_direct_connection(self, req,
...
                     if not self._cleanup_closed_disabled:
                         self._cleanup_closed_transports.append(transp)
-                    raise ServerFingerprintMismatch(
+                    last_exc = ServerFingerprintMismatch(
                         expected, got, host, port)
+                    continue
+
             return transp, proto
+        else:
+            assert last_exc is not None
```
asvetlov: Just drop the line.
Check doesn't point on potential problem, next line will fail enyway (raise None is forbidden in Python 3).

PR #2452, opened 2017-10-31
payload.py now uses async/await syntax

eteamin:
issue #2343 
however `IOBasePayload` and `BytesPayload` classes still use the old style async syntax for their `write` coroutines. Because changing their `write` coroutine to new syntax raises `TypeError: object tuple can't be used in 'await' expression` that I was unable to fix.


asvetlov:
Thanks!

Issue #2444, opened 2017-10-30
all except BaseException should be except Exception


hellysmile:
## Long story short

https://github.com/aio-libs/aiohttp/pull/2383 fixed flake8 blind exception, but later for example https://github.com/aio-libs/aiohttp/pull/2441 used `except BaseException`

It should be fixed separately in 2.3 and master branches

## Expected behaviour

everywhere should be `except Exception`

## Actual behaviour

Some parts of code uses `except Exception`, some `except BaseException`

fafhrd91:
For 2441 BaseException makes sense, especially because it just re-raise exception

socketpair:
@asvetlov it is actual again. We need a linter.

See https://github.com/aio-libs/aiohttp/blob/7a324fd46ff7dc9bb0bb1bc5afb326e04cf7cef0/aiohttp/client_reqrep.py#L974 for example. Exception in `close()` prevent, for example, exitting from the application.

asvetlov:
No. 
The code catches an exception, closes the response and reraises the exception again.
It should work pretty fine with `CancelledError`, isn't it?

Exception in `close()` unwinds the stack anyway.

socketpair:
Yes, unwinds, but exception in `.close()` will REPLACE, say, SystemExit exception, does not it ?

asvetlov:
Chicken-egg problem.
I have a feeling that a response should be closed in case of exception.
The question is: how should look like an exception in finalizer.
@socketpair do you have a suggestion?

socketpair:
1. We should track only `except Exception`  (and CancelledError also, if applicable).
2. AFAIK, calling `close()` - just an GC optimisation. In good code, GC should kill such items.
3. Another option - try...except...warning+ignore wrapper over `close()`.

There are plenty of exceptions, which are not errors actually. It's wrong design. Like Stopiteration,systemexit,Cancellederror....We should close something only by ERROR. If someone raises such exceptions it should understand what will happen.

That's my opinion.

PR #2453, opened 2017-10-31
BaseException -> Exception

kvirt:
## What do these changes do?

Fix swallow `BaseException` (`KeyboardInterrupt`)

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2444

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -724,7 +724,7 @@ class _SessionRequestContextManager(_RequestContextManager):
     def __iter__(self):
         try:
             return (yield from self._coro)
-        except BaseException:
+        except Exception:
             yield from self._session.close()
             raise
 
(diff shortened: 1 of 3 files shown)


fafhrd91:
This change will leak resources

asvetlov:
@fafhrd91 do you suggest catching `BaseException`?

fafhrd91:
For sure If we need to release resourse. We still can re-raise

arthurdarcet:
For the two except in `client.py`, I agree with @fafhrd91, they should definitely stay `BaseException`. They are immediately re-raised so this PR would change any behaviour except that the transport wouldn't be properly closed when a BaseException is raised. This is exactly why #2441 was needed.

For the other two excepts, I think the `set_exception` call should be replaced with a `raise` if the exception is a BaseException (and not an Exception). But I'm having a hard time imagining use cases where the current code would lead to issues.
I guess there might be some edge cases where a BaseException would be swallowed into a future, and never really raised, so raising it immediately is safer.

asvetlov:
Fixed in master

Issue #1670, opened 2017-02-22
ClientResponse ergonomics


fafhrd91:
I think we should add following:

1. `content-disposition` header fast access
2. `.multipart()` coro
3. something similar to `.post()`  from WebRequest

asvetlov:
Something is done already, other things look lack an interest.

PR #2455, opened 2017-10-31
Content-Disposition fast access in ClientResponse

redixin:
Add content_disposition and content_disposition_dict properties

Partially implements #1670

## What do these changes do?

Improves ClientResponse ergonomics

All changes are documented.

## Related issue number

#1670

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -9,16 +9,17 @@ import warnings
 from collections import namedtuple
 from hashlib import md5, sha1, sha256
 from http.cookies import CookieError, Morsel, SimpleCookie
+from types import MappingProxyType
 
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
 from yarl import URL
 
-from . import hdrs, helpers, http, payload
+from . import hdrs, helpers, http, multipart, payload
 from .client_exceptions import (ClientConnectionError, ClientOSError,
                                 ClientResponseError, ContentTypeError,
                                 InvalidURL)
 from .formdata import FormData
-from .helpers import HeadersMixin, TimerNoop, noop
+from .helpers import HeadersMixin, TimerNoop, noop, reify
 from .http import SERVER_SOFTWARE, HttpVersion10, HttpVersion11, PayloadWriter
 from .log import client_logger
 from .streams import FlowControlStreamReader
@@ -31,6 +32,10 @@ except ImportError:  # pragma: no cover
 
 
 __all__ = ('ClientRequest', 'ClientResponse', 'RequestInfo')
+
+
+ContentDisposition = collections.namedtuple(
+    'ContentDisposition', ('type', 'parameters', 'filename'))
 
 
 RequestInfo = collections.namedtuple(
@@ -527,6 +532,7 @@ class ClientResponse(HeadersMixin):
         self._request_info = request_info
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
+        self._cache = {}  # reqired for @reify method decorator
 
     @property
     def url(self):
@@ -549,6 +555,16 @@ class ClientResponse(HeadersMixin):
     @property
     def request_info(self):
         return self._request_info
+
+    @reify
+    def content_disposition(self):
+        raw = self._headers.get(hdrs.CONTENT_DISPOSITION)
+        if raw is None:
+            return None
+        disposition_type, params = multipart.parse_content_disposition(raw)
+        params = MappingProxyType(params)
+        filename = multipart.content_disposition_filename(params)
+        return ContentDisposition(disposition_type, params, filename)
 
     def _post_init(self, loop, session):
         self._loop = loop
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1109,12 +1109,19 @@ Response object
 
    .. attribute:: charset
 
-   Read-only property that specifies the *encoding* for the request's BODY.
+      Read-only property that specifies the *encoding* for the request's BODY.
 
       The value is parsed from the *Content-Type* HTTP header.
 
       Returns :class:`str` like ``'utf-8'`` or ``None`` if no *Content-Type*
       header present in HTTP headers or it has no charset information.
+
+   .. attribute:: content_disposition
+
+      Read-only property that specified the *Content-Disposition* HTTP header.
+
+      Instance of :class:`ContentDisposition` or ``None`` if no *Content-Disposition*
+      header present in HTTP headers.
 
    .. attribute:: history
 
@@ -1561,6 +1568,21 @@ All exceptions are available as members of *aiohttp* module.
 
       Invalid URL, :class:`yarl.URL` instance.
 
+.. class:: ContentDisposition
+
+    Represent Content-Disposition header
+
+    .. attribute:: value
+
+    A :class:`str` instance. Value of Content-Disposition header itself, e.g. ``attachment``.
+
+    .. attribute:: filename
+
+    A :class:`str` instance. Content filename extracted from parameters. May be ``None``.
+
+    .. attribute:: parameters
+
+    Read-only mapping contains all parameters.
 
 Response errors
 ^^^^^^^^^^^^^^^
--- a/tests/test_client_response.py
+++ b/tests/test_client_response.py
@@ -458,6 +458,42 @@ def test_charset_no_charset():
     assert response.charset is None
 
 
+def test_content_disposition_full():
+    response = ClientResponse('get', URL('http://def-cl-resp.org'))
+    response.headers = {'Content-Disposition':
+                        'attachment; filename="archive.tar.gz"; foo=bar'}
+
+    assert 'attachment' == response.content_disposition.type
+    assert 'bar' == response.content_disposition.parameters["foo"]
+    assert 'archive.tar.gz' == response.content_disposition.filename
+    with pytest.raises(TypeError):
+        response.content_disposition.parameters["foo"] = "baz"
+
+
+def test_content_disposition_no_parameters():
+    response = ClientResponse('get', URL('http://def-cl-resp.org'))
+    response.headers = {'Content-Disposition': 'attachment'}
+
+    assert 'attachment' == response.content_disposition.type
+    assert response.content_disposition.filename is None
+    assert {} == response.content_disposition.parameters
+
+
+def test_content_disposition_no_header():
+    response = ClientResponse('get', URL('http://def-cl-resp.org'))
+    response.headers = {}
+
+    assert response.content_disposition is None
+
+
+def test_content_disposition_cache():
+    response = ClientResponse('get', URL('http://def-cl-resp.org'))
+    response.headers = {'Content-Disposition': 'attachment'}
+    cd = response.content_disposition
+    ClientResponse.headers = {'Content-Disposition': 'spam'}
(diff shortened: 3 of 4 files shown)


redixin:
@fafhrd91 should I use issue number or pull request number for CHANGES? I'm asking because this PR implements only first item in issue #1670

fafhrd91:
1. Aiohttp has content disposition parsing function. I think cgi module doesn’t handle all cases.
2. Regarding changes, it is question for asvetlov

fafhrd91:
I like idea of `ContentDisposition` class

asvetlov:
PR number is ok for the case -- it doesn't cover the whole issue.

fafhrd91:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/multipart.py#L35

aiohttp/helpers.py
```diff
@@ -737,3 +755,10 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
 
         if content_length:
             return int(content_length)
+
+    @property
+    def content_disposition(self, *, header=hdrs.CONTENT_DISPOSITION):
```
asvetlov: Why the property is in `HeadersMixin`?
Should `web.Request` has the property?
I thought it is only for `ClientResponse`

aiohttp/helpers.py
```diff
@@ -714,6 +720,18 @@ def _parse_content_type(self, raw):
...
 
+    def _parse_content_disposition(self, _CONTENT_DISPOSITION):
+        from . import multipart
+        raw = self._headers.get(_CONTENT_DISPOSITION)
+        if raw is None:
+            self._content_disposition = None
+        else:
+            value, params = multipart.parse_content_disposition(raw)
+            filename = multipart.content_disposition_filename(params)
+            self._content_disposition = ContentDisposition(value, params,
```
asvetlov: Convert `params` into `types.MappingProxyType`. The value should be immutable

aiohttp/helpers.py
```diff
@@ -737,3 +755,10 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
 
         if content_length:
             return int(content_length)
+
+    @property
+    def content_disposition(self, *, header=hdrs.CONTENT_DISPOSITION):
+        """The value of Content-Disposition HTTP header."""
+        if self._content_disposition == False:
```
kxepal: `is False`

aiohttp/helpers.py
```diff
@@ -703,6 +708,7 @@ class HeadersMixin:
 
     _content_type = None
     _content_dict = None
+    _content_disposition = False
```
kxepal: Quite strange default. Why `False`?

aiohttp/helpers.py
```diff
@@ -714,6 +720,18 @@ def _parse_content_type(self, raw):
         else:
             self._content_type, self._content_dict = cgi.parse_header(raw)
 
+    def _parse_content_disposition(self, _CONTENT_DISPOSITION):
+        from . import multipart
```
kxepal: Why hidden import?

aiohttp/helpers.py
kxepal: Indeed it's response only header.

aiohttp/helpers.py
redixin: Will do in next patchset.

aiohttp/helpers.py
redixin: False if header are not parsed/cached yet
None if there was an attempt to parse, but no header found
ContentDisposition if header was found and parsed

aiohttp/helpers.py
redixin: It was here due to circular import

aiohttp/client_reqrep.py
```diff
@@ -500,6 +505,7 @@ class ClientResponse(HeadersMixin):
     raw_headers = None  # Response raw headers, a sequence of pairs
 
     _connection = None  # current connection
+    _content_disposition = False  # content disposition cache
```
kxepal: I still think that's not good choice. `None` means no value. We tried to parse it, nothing happened. Yes, this will trigger headers lookup for each property access - but that's not a problem.

asvetlov: `helpers.sentinel` is an option

kxepal: Good alternative, but `None` is fine as well. However, even better would be to use reify here and avoid internal attributes. Headers are immutable in anyway.

asvetlov: +1 for `reify`

aiohttp/client_reqrep.py
```diff
@@ -550,6 +556,16 @@ def _headers(self):
...
 
+    @reify
+    def content_disposition(self):
+        raw = self._headers.get(hdrs.CONTENT_DISPOSITION)
+        if raw is None:
+            return None
+        value, params = multipart.parse_content_disposition(raw)
+        params = MappingProxyType(params)
+        filename = multipart.content_disposition_filename(params)
+        return ContentDisposition(value, params, filename)
```
asvetlov: Convert params into immutable structure

asvetlov:
`flake8` tool still disagrees with your code style :)

aiohttp/client_reqrep.py
redixin: What do you mean? `params` is already converted to MappingProxyType line 565

aiohttp/client_reqrep.py
```diff
@@ -527,6 +532,7 @@ def __init__(self, method, url, *,
         self._request_info = request_info
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
+        self._cache = {}
```
kxepal: This looks like redundant.

tests/test_client_response.py
```diff
@@ -458,6 +458,32 @@ def test_charset_no_charset():
...
 
 
+def test_content_disposition_full():
+    response = ClientResponse('get', URL('http://def-cl-resp.org'))
+    response.headers = {'Content-Disposition':
+                        'attachment; filename="archive.tar.gz"; foo=bar'}
+
+    assert 'attachment' == response.content_disposition.value
+    assert 'bar' == response.content_disposition.parameters["foo"]
+    assert 'archive.tar.gz' == response.content_disposition.filename
```
kxepal: Worth to add two more cases:
1. `content_disposition.parameters` is immutable
2. second access to `content_disposition` returns the same result if `ClientResponse.headers` were modified in-between (I'm surprised to see it's not read-only property).

docs/client_reference.rst
```diff
@@ -1561,6 +1568,21 @@ All exceptions are available as members of *aiohttp* module.
 
       Invalid URL, :class:`yarl.URL` instance.
 
+.. class:: ContentDisposition
+
+    Represent Content-Disposition header
+
+    .. attribute:: value
+
+    Value of Content-Disposition header itself, e.g. ``attachment``.
```
kxepal: ``"attachment"`` - you want to notice that the value is a string typed.

docs/client_reference.rst
```diff
@@ -1561,6 +1568,21 @@ All exceptions are available as members of *aiohttp* module.
...
+
+    .. attribute:: filename
+
+    Content filename extracted from parameters. May be ``None``.
+
+    .. attribute:: parameters
+
+    A :class:`dict` instance contains all parameters.
```
kxepal: Actually, it's not a dict, but `MapptingProxy`. Difference in mutability.

aiohttp/client_reqrep.py
```diff
@@ -33,6 +34,10 @@
 __all__ = ('ClientRequest', 'ClientResponse', 'RequestInfo')
 
 
+ContentDisposition = collections.namedtuple(
+    'ContentDisposition', ('value', 'parameters', 'filename'))
```
kxepal: I think `value` is not the best choice. The [RFC](https://tools.ietf.org/html/rfc6266#section-4.1) defines it as a `type` which suites more to what the value it holds.

aiohttp/client_reqrep.py
```diff
@@ -550,6 +556,16 @@ def _headers(self):
     def request_info(self):
         return self._request_info
 
+    @reify
+    def content_disposition(self):
+        raw = self._headers.get(hdrs.CONTENT_DISPOSITION)
+        if raw is None:
+            return None
+        value, params = multipart.parse_content_disposition(raw)
```
kxepal: It seems [only two](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) values here are valid for HTTP headers case: `inline` and `attachment`. May be worth to raise a warning about invalid type or turn the `value` in Enum?

aiohttp/client_reqrep.py
redixin: Looks like no. All classes with @reify methods have this and AttributeError is raising without this line.

aiohttp/client_reqrep.py
redixin: Agree

aiohttp/client_reqrep.py
asvetlov: Sorry.
Nevermind

docs/client_reference.rst
asvetlov: We could just replace it with `*mapping*` for sake of simplicity.
I pretty sure not many python developers know what `MappingProxy` is :)

aiohttp/client_reqrep.py
kxepal: Implicit magic. Worth to add notice about.

aiohttp/client_reqrep.py
redixin: According to https://tools.ietf.org/html/rfc6266#section-4.1 any token is valid

aiohttp/client_reqrep.py
asvetlov: Yes, it's required.

aiohttp/client_reqrep.py
asvetlov: I like `Enum` but how the change affects `multidict` and existing users.
I suspect `parse_content_disposition` is internal API but not 100% sure.

docs/client_reference.rst
redixin: Will do

docs/client_reference.rst
redixin: Oops

aiohttp/client_reqrep.py
kxepal: @redixin Yes, but RFC defines more general Content-Disposition definition. Mozilla explicitly splits it into HTTP headers and Multpart headers. The second one utilize `token`  rule completely, but here we are in HTTP response land.

docs/client_reference.rst
kxepal: Good idea.

tests/test_client_response.py
redixin: Will do

aiohttp/client_reqrep.py
asvetlov: I mean the attr is required, not a notice :)
But the notice makes no harm

kxepal: Oh, ok. Last time I saw `reify` in acttion, no special attrs were required.

asvetlov:
Great work!
Thank you!

PR #2465, opened 2017-11-05
WIP Server class to start applications

cecton:
## What do these changes do?

Attempt to answer to the issue raised on #2121 and #2375 by creating a `Server` class capable of registering multiple applications and providing a `start()` and a `stop()` asynchronous methods to allow the developers to open and close the server sockets to the application.

Added a Server class. `web.run_app` is preserved.

## Related issue number

#2121 #2375

## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -352,7 +352,7 @@ class Server:
         self.backlog = backlog
         self.access_log_format = access_log_format
         self.access_log = access_log
-        self.base_url = URL('{}://localhost'.format(scheme)).with_port(port)
+        self.base_url = URL.build(scheme=scheme, host='localhost', port=port)
 
         if path is None:
             self.paths = ()
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -473,14 +473,15 @@ class Server:
             yield from app.cleanup()
 
 
-def run_app(app, *, print=print, handle_signals=True, loop=None, **kwargs):
+def run_app(*apps, print=print, handle_signals=True, loop=None, **kwargs):
     """Run an app locally"""
     user_supplied_loop = loop is not None
     if loop is None:
         loop = asyncio.get_event_loop()
 
     server = Server(loop=loop, **kwargs)
-    server.register(app)
+    for app in apps:
+        server.register(app)
 
     uris = loop.run_until_complete(server.start())
 
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -27,7 +27,7 @@ from .web_middlewares import *  # noqa
 from .web_protocol import *  # noqa
 from .web_request import *  # noqa
 from .web_response import *  # noqa
-from .web_server import Server
+from .web_server import Server as WebServer
 from .web_urldispatcher import *  # noqa
 from .web_urldispatcher import PrefixedSubAppResource
 from .web_ws import *  # noqa
@@ -240,9 +240,9 @@ class Application(MutableMapping):
             for k, v in self._handler_args.items():
                 kwargs[k] = v
 
-        return Server(self._handle, request_factory=self._make_request,
-                      access_log_class=access_log_class,
-                      loop=self.loop, **kwargs)
+        return WebServer(self._handle, request_factory=self._make_request,
+                         access_log_class=access_log_class,
+                         loop=self.loop, **kwargs)
 
     async def startup(self):
         """Causes on_startup signal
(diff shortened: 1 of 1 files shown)


cecton:
@asvetlov is it more or less the idea you had in mind?

aiohttp/web.py
```diff
@@ -27,7 +27,7 @@
 from .web_protocol import *  # noqa
 from .web_request import *  # noqa
 from .web_response import *  # noqa
-from .web_server import Server
+from .web_server import Server as WebServer
```
kxepal, line 30: Maybe rename web_server.Server into web_server.WebServer to avoid collisions?

aiohttp/web.py
```diff
@@ -332,140 +332,176 @@ def raise_graceful_exit():
...
+        self._handler = app.make_handler(loop=loop, access_log=access_log,
+                                         **make_handler_kwargs)
+
+        self.shutdown_timeout = shutdown_timeout
+        self.scheme = scheme = 'https' if ssl_context else 'http'
+        self.ssl_context = ssl_context
+        self.backlog = backlog
+        self.access_log_format = access_log_format
+        self.access_log = access_log
+        self.base_url = URL('{}://localhost'.format(scheme)).with_port(port)
```
kxepal: `URL.build` was pretty nice.

aiohttp/web.py
```diff
@@ -332,140 +332,176 @@ def raise_graceful_exit():
...
+                                                   sock.getsockname()))
+            else:
+                host, port = sock.getsockname()
+                uris.append(str(self.base_url.with_host(host).with_port(port)))
+
+        self.uris = uris
+        return asyncio.gather(*server_creations, loop=self.loop)
+
+    @asyncio.coroutine
+    def start(self):
```
kxepal, line 449: How could I start specific app in runtime after starting all the apps?

aiohttp/web.py
```diff
@@ -332,140 +332,176 @@ def raise_graceful_exit():
...
+        for app in self._apps:
+            app._set_loop(self.loop)
+            yield from app.startup()
+
+        self._servers = yield from self._create_servers(self._handler)
+
+        return self.uris
+
+    @asyncio.coroutine
+    def stop(self):
```
kxepal, line 459: How could I stop specific app in runtime?

aiohttp/web.py
```diff
@@ -332,140 +332,176 @@ def raise_graceful_exit():
...
-        # Multiple hosts bound to same server is available in most loop
-        # implementations, but only send multiple if we have multiple.
-        host_binding = hosts[0] if len(hosts) == 1 else hosts
-        server_creations.append(
-            loop.create_server(
-                handler, host_binding, port, ssl=ssl_context, backlog=backlog
+        self._apps = []
+        self._servers = None
+
+    def register(self, app, *, prefix="/"):
```
kxepal, line 389: How could I unregister specific app in runtime?

aiohttp/web.py
cecton: This will break retro compatibility but we can maybe

aiohttp/web.py
cecton: Oops sorry I missed this one 😀 I based my branch on an old version by mistake then I rebased too late when I realized my mistake.

aiohttp/web.py
cecton: You can't at the moment. It's not meant to be mutable. I can try to implement it but this is not my priority. The code to create a server and register apps is pretty small and I suppose it's best to just re-instantiate a new one if you need to change something.

aiohttp/web.py
cecton: I'm not sure what you mean. This is not meant to start selectively the applications. This is meant to group all the target applications and run them together on one connection. I mostly inspired on hapi's Glue for the concept.

cecton:
I will give thorough examples and documentation if this is what we want for aiohttp to handle multiple applications (and running the applications without being the main process). Right now all I can say is that you could potentially do this:

Imagine you want to run side by side a grpc server and an application. Now you can by creating dynamically a `Server` object and registering your application. You just start it and it will work while your main program can do something else. You can stop anytime if you don't want to expose it all the time.

It also helps to gather `Application` objects and run them together. ~But then maybe it would be a good idea to allow `web.run_app` to take multiple positional arguments for apps and register all the apps to the server.~ [edit: done]

I suppose there are a lot of use cases for this but to be honest I have never encountered one yet professionally since I mostly did microservices.

Maybe @alexmnazarenko could give his input on this feature. It's just a sample at this point, we can add `wait()` or other things quite easily I suppose.

aiohttp/web.py
cecton: Actually I just saw this in a test traceback:

```
>       srv = web.Server(handler)
E       TypeError: __init__() takes 1 positional argument but 2 were given
```

Which means the name Server is exported in `aiohttp.web` so I need to find another name than `Server` to avoid breaking compatibility.

asvetlov:
Looks as good concept.
I have a couple ideas for discussion.
We have too many server classes :) 
We need another name. 
What is the best name for `Application` instance serving specific TCP or UNIX port? `Site` maybe? I'm thinking about documentation.

`Server` from `./aiottp.web_server` is a protocol factory for making a HTTP protocol for incoming connections. It could be used alone for making proxies etc.

`Application` is the most used kind of `Server`. It has a state, route table, nested apps etc.
The application creates a `Server` internally with `handler` hook to process routes etc.

Running `Application` could serve several *sites* on different ports or unix paths.

P.S.
From my taste `web._make_server_creators` function is a mess: it tries to solve too many use cases.
What's about making a `Site` base class with `TCPSite`, `UNIXSite` and `SocketSite` derivatives for processing `host/port`, `path` and `sock` parameters of `run_app` correspondingly? These classes shares almost everything except socket creation procedure.

aiohttp/web.py
kxepal: Yea, better to avoid this name collision somehow.

aiohttp/web.py
kxepal: Re-instantiate a new server means to stop all the applications and cause a downtime. Not a good behaviour for adding a new app.

aiohttp/web.py
kxepal: I mean if Server object manages multiple applications, then why not you can't use it to register/start/stop arbitrary application is runtime? Quite oblivious task for it.

cecton:
> Looks as good concept.

It tried to read in your mind 😀 

> We have too many server classes :) 

👍 

> We need another name. 
> What's about making a `Site` base class with `TCPSite`, `UNIXSite` and `SocketSite` derivatives for processing `host/port`, `path` and `sock` parameters of `run_app` correspondingly? These classes shares almost everything except socket creation procedure.

That sounds like a great idea!

> Running `Application` could serve several *sites* on different ports or unix paths.

Wait I'm lost. So you would rename Application to Site and this new Server to Application? But then it's TCPApplication, UnixApplication and SocketApplication isn't it? Also you can do a major release if you rename everything like that.

aiohttp/web.py
cecton: I haven't make it with the idea of changing the applications during runtime at all but it sounds like an interesting use case.

aiohttp/web.py
cecton: (Same comment: it wasn't meant to avoid downtime, it was meant to allow running the apps without having to make aiohttp the center of the process)

asvetlov:
> Wait I'm lost. So you would rename Application to Site and this new Server to Application? But then it's TCPApplication, UnixApplication and SocketApplication isn't it? Also you can do a major release if you rename everything like that.

No. `Application` should not be touched.
`Site` use `Application` for serving the app on specific host/port.
Another site could do the same for other port (think about HTTP/HTTPS).

asvetlov:
Superseded by #2530

asvetlov:
@cecton I believe #2530 solves the problem better.
Big thank you for working on prototypes -- it allowed me to look on the problem from different points and find the final solution.

cecton:
You're welcome :) I'm glad it helped!

PR #2471, opened 2017-11-06
Update FAQ

asvetlov:
Reflect the fact that aiohttp supports route decorators now.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -7,7 +7,8 @@ FAQ
 Are there any plans for @app.route decorator like in Flask?
 -----------------------------------------------------------
 
-We have it already: :ref:`aiohttp-web-alternative-routes-definition`.
+We have it already (*aiohttp>=2.3* required):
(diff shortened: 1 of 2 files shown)


socketpair:
Please add since which version it works.

asvetlov:
@socketpair done

socketpair:
Is it possible to use this decorator on top of View (http://aiohttp.readthedocs.io/en/stable/web_reference.html#view) ?

I think, it should be mentioned in both, FAQ and main doc (http://aiohttp.readthedocs.io/en/stable/web_reference.html#routetabledef). (I'm going to use it, I still did not test)

asvetlov:
Views are not supported (yet), PR is welcome!

socketpair:
So, it will be nice to mention, that Views are not supported yet.

asvetlov:
I hope somebody will find a time for #2472 before 3.0
Let me merge as is.

PR #2475, opened 2017-11-07
Await in websockets

asvetlov:
Make `send_*()` and `ping()`/`pong()` true coroutines.

Get rid of `yield from`


asvetlov:
Need a doc update before merging

Issue #2419, opened 2017-10-27
Deprecate non-coroutine signal handlers


asvetlov:
Now we have 2 checks: `iscoroutine()` and `isfuture()`.
Dropping them may speedup signal handling significantly.

asvetlov:
Moreover we can do check in `freeze()` call and wrap plain functions into coroutines as we do it for web handlers in `AbstractRoute`.

asvetlov:
Fixed by #2480

PR #2480, opened 2017-11-08
Signals refacoting

asvetlov:
* Convert more API to async/await syntax
* Simplify and speedup signals implementation
* Drop support for non-coroutine signal handlers

Fix for #2419

### DIFF
--- a/aiohttp/signals.py
+++ b/aiohttp/signals.py
@@ -1,95 +1,32 @@
-import asyncio
-from itertools import count
-
 from aiohttp.frozenlist import FrozenList
 
 
-class BaseSignal(FrozenList):
-
-    __slots__ = ()
-
-    async def _send(self, *args, **kwargs):
-        for receiver in self:
-            res = receiver(*args, **kwargs)
-            if asyncio.iscoroutine(res) or asyncio.isfuture(res):
-                await res
-
-
-class Signal(BaseSignal):
+class Signal(FrozenList):
     """Coroutine-based signal implementation.
 
     To connect a callback to a signal, use any list method.
 
-    Signals are fired using the :meth:`send` coroutine, which takes named
+    Signals are fired using the send() coroutine, which takes named
     arguments.
     """
 
-    __slots__ = ('_app', '_name', '_pre', '_post')
+    __slots__ = ('_owner',)
 
-    def __init__(self, app):
+    def __init__(self, owner):
         super().__init__()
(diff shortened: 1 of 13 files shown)


asvetlov:
@pfreixes the RP breaks your work on client tracing but I pretty sure the change is not big and is done for good.

asvetlov:
Sure.
These signals was never documented and I pretty sure nobody uses it.

aiohttp/signals.py
```diff
@@ -1,95 +1,32 @@
...
-        self._counter = count(1)
-
-    def ordinal(self):
-        return next(self._counter)
-
-
-class PostSignal(DebugSignal):
-
-    __slots__ = ()
+            await receiver(*args, **kwargs)
```
kxepal, line 32: How about to add more friendly exception if plain func is going to be registered as handler? I guess now such code will fail with cryptic `TypeError: object NoneType can't be used in 'await' expression` or similar.

asvetlov: Do you suggest catching `TypeError` and raise a new `TypeError` with more informative exception message?

asvetlov: As an option we could check ll callbacks on freezing stage.
It prevents registering a regular function which returns a future but it's fine I guess.

kxepal: Yes for TypeError and yes, better check this early. On append call if possible. The stacktrace will help to find that bad signal usage with easy.

asvetlov: FrozenList can be modified by too many ways, `.append` is not the only one.
I think `.freeze()` is a good compromise.

kxepal: It's a pity. Ok then if error message would contains function name which is not async.

asvetlov: Well, but I see you point.
Maybe we need specialized `CheckedFrozenList` class with registered callback? It could improve UX slightly.
Please let me merge the PR as is and make a new issue.
I don't want to block @pfreixes with his client tracing PR.

kxepal: Ok. Let's do that.

pfreixes:
LGTM. Just take into account that Func signals might be needed later, but let's implement it once we know for sure that are gonna needed.

asvetlov:
1. API simplicity
2. A check for coroutine is quite expensive

PR #2483, opened 2017-11-08
Polish docs and ABC, convert everything except multipart and client to async/await syntax

asvetlov:
* Make AbstractPayloadWriter.write() an async function
* Drop response.drain() from docs

The drain is still supported but removed from documentation: users should call `await resp.write()`.
`write()` calls writer's `drain()` behind the scene.

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -393,7 +393,7 @@ class StreamResponse(HeadersMixin):
 
         return writer
 
-    def write(self, data):
+    async def write(self, data):
         assert isinstance(data, (bytes, bytearray, memoryview)), \
             "data argument must be byte-ish (%r)" % type(data)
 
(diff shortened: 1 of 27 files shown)


aiohttp/web_response.py
```diff
@@ -408,6 +408,9 @@ def write(self, data):
         assert not self._eof_sent, "EOF has already been sent"
         assert self._payload_writer is not None, \
             "Response has not been started"
+        warnings.warn("drain method is deprecated, use await resp.write()",
```
kxepal, line 411: `write` method should be changed to async. Otherwise this suggestion will lead to errors.

asvetlov: Please elaborate

kxepal: `StreamResponse.write` is not `async` now. What it returns is a mystery. According the interface `StreamResponse.write` breaks it. So it would be fine to expect strange things including TypeErrors on trying to await non-coroutine value.

asvetlov: `write` is async now.

asvetlov:
@kxepal please review again.
I believe the PR is done and ready for merging.

asvetlov:
@kxepal thanks!

asvetlov:
@kxepal say again I very appreciate your reviews.
Please keep it going.

kxepal:
@asvetlov you're welcome! Glad to help how I can (:

PR #2484, opened 2017-11-08
Add drain() method to ClientWebSocketResponse.

iksteen:
## What do these changes do?

These changes add a drain() method to the ClientWebSocketResponse class, analogous to the drain() method of the WebSocketResponse class.

There should be no changes in existing applications, only newly added functionality.

## Related issue number

I am not aware of a related issue number.

## Checklist

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -126,6 +126,9 @@ class ClientWebSocketResponse:
     async def send_json(self, data, *, dumps=json.dumps):
         await self.send_str(dumps(data))
 
+    async def drain(self):
+        await self._writer.drain()
+
     async def close(self, *, code=1000, message=b''):
         # we need to break `receive()` cycle first,
         # `close()` may be called from different task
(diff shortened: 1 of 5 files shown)


aiohttp/client_ws.py
```diff
@@ -126,6 +126,9 @@ def exception(self):
     async def send_json(self, data, *, dumps=json.dumps):
         await self.send_str(dumps(data))
 
+    async def drain(self):
```
webknjaz, line 129: I'd prefer to see a docstring here as well.

asvetlov:
I think the PR is invalid.
1. `WebSocketResponse` **has no** `drain()` method.
2. All senders are coroutines now, just use `await ws.send_str()` etc -- they do draining under the hood.

iksteen:
That's what I get for not paying attention to changes in master. In hindsight I agree that this is not necessary anymore. It was originally developed for 2.x.

For completeness though: WebSocketResponse has a drain() method via StreamWriter.

PR #2500, opened 2017-11-10
Web protocol cleanup

asvetlov:
* Drop deprecated param
* Drop kwargs

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -3,7 +3,6 @@ import asyncio.streams
 import http.server
 import socket
 import traceback
-import warnings
 from collections import deque
 from contextlib import suppress
 from html import escape as html_escape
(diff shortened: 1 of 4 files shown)


aiohttp/web_protocol.py
```diff
@@ -94,11 +92,7 @@ def __init__(self, manager, *, loop=None,
                  max_headers=32768,
                  max_field_size=8190,
                  lingering_time=10.0,
-                 max_concurrent_handlers=1,
```
kxepal, line 97: Is that an accidental removal?

asvetlov: `max_concurrent_handlers` is not removed, see below.
But `**kwargs` is dropped

kxepal: Bah! GitHub UI tricked me.

Issue #2339, opened 2017-10-18
Check markup and spelling on every PR


asvetlov:
It could be done by running `towncrier --yes` on travis *without* committing changed files.

Otherwise I should fix spelling just before release.

webknjaz:
It could be added to build stage with linters as a separate job. Doing it right before deploy would fail late (which is much more time-consuming) in contrary to fail fast approach.

asvetlov:
Agree

PR #2503, opened 2017-11-10
Check markup and spelling on every PR

asvetlov:
Fix for #2339

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -143,6 +143,7 @@ jobs:
     env:
     - docs
     script:
+    - towncrier --yes
     - make doc-spelling
 
   - <<: *_lint_base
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -143,6 +143,7 @@ jobs:
     env:
     - docs
     script:
+    - towncrier --yes
```
webknjaz, line 146: Do we have it installed at this point?

webknjaz: Nevermind, just checked: it's okay.

CHANGES/2396.doc
```diff
@@ -1 +1 @@
-Improve tutorial docs, using literalinclude to link to the actual files.
+Improve tutorial docs, using `literalinclude` to link to the actual files.
```
webknjaz, line 1: This change is unrelated.

asvetlov: It's a fix for already committed CHANGES.
`literalinclude` is a spelling error without backticks.

Issue #2415, opened 2017-10-26
Deprecate returning web.HTTPException from handlers and middlewares


asvetlov:
Now an endpoint might either return `web.HTTPException` (`HTTPForbidden` for example) or raise it.

It makes a mess, very often a middleware should check for type of response and process it as exception if the response is an instance of `HTTPException`.

Deprecation could be added just before this line: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_protocol.py#L428

Next step in transition plan is adding `HTTPException.make_response()` method.

At the end we could drop inheritance from `web.Response` at all.

@tailhook told me about the problem a year ago.
`aiohttp.web` was designed after Pyramid framework. 
I still think Pyramid is much better designed than Django and Flask but the library is not such popular because alternatives acquired much better promotion.
Anyway, Pyramid inherits HTTP exceptions from both `Response` and `Exception` classes, we did the same.
Now, after years of working on aiohttp and derived projects, it's clean for me: the choice was definitely wrong.

I was not ready for such breaking change at the moment but now I have a strong feeling that the change makes sense and very important.

Sure, we should do transition smoothly: a warning and `make_response()` method could be added in next aiohttp 3.0 release but `web.Response` removal should wait for a year and half at least.

Champions for first two steps are welcome!

fafhrd91:
I agree with motivation, but I think 1 year is too long. 3.0 should deprecate httpexception return. And turn to a hard error in half a year or 3.2

Replacing return with raise is not big change

fafhrd91:
Btw transition to 2.0 was not that hard, but amount of changes was relatively big.

asvetlov:
@fafhrd91 maybe you right. But maybe a year will be good enough. Or 1.5 years.
Honestly I don't care right now, we need to prepare for transition in aiohttp 3.0 and only after this date decide how long we should support deprecated logic.

fafhrd91:
I would even drop instead of deprecate for 3.0. migration to 3.0 would need attention to software in any case.

kxepal:
I'm +1 to drop as well.

webknjaz:
+1 to drop inheritance from Response

asvetlov:
Fixed by #2515

PR #2515, opened 2017-11-13
Add deprecation warning when HTTPException is returned instead of raised

panagiks:
## What do these changes do?

Add DeprecationWarning when an HTTPException is returned instead of being raised.

If user's handler (view function) returns an HTTPException, a DeprecationWarning will be displayed.

## Related issue number

Discussion in #2415

## Checklist

### DIFF
--- a/CHANGES/2415.removal
+++ b/CHANGES/2415.removal
@@ -1,0 +1,1 @@
+Add DeprecationWarning for returning HTTPException
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -140,6 +140,7 @@ Nicolas Braem
 Nikolay Novik
 Olaf Conradi
 Pahaz Blinov
+Panagiotis Kolokotronis
 Pankaj Pandey
 Pau Freixes
 Paul Colomiets
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -3,6 +3,7 @@ import asyncio.streams
 import http.server
 import socket
 import traceback
+import warnings
 from collections import deque
 from contextlib import suppress
 from html import escape as html_escape
@@ -426,6 +427,14 @@ class RequestHandler(asyncio.streams.FlowControlMixin, asyncio.Protocol):
                 # log access
                 if self.access_log:
                     self.log_access(request, resp, loop.time() - now)
+
+                # Deprication warning (See #2415)
+                if isinstance(resp, HTTPException):
+                    warnings.warn(
+                        "returning HTTPException object is deprecated (#2415) "
+                        "and will be removed, "
+                        "please raise the exception instead",
+                        DeprecationWarning)
 
                 # check payload
                 if not payload.is_eof():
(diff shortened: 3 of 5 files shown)


asvetlov:
Sorry, travis checks is broken by me.
Will fix it soon.

panagiks:
[INVALID QUESTION] I will add the build_response() method along with test in a while. Do we need reference to HTTPException instance in Response? (should I instantiate the object as done in [HTTPException's constructor](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_exceptions.py#L79) or by `Response(...)`)

panagiks:
I'm not really a fan of the implementation in my last commit (as it includes some code repetition) but there were a few issues with my initial approach:
* If build_response() creates a new Response there is the issue that, at this point, self.body and self.text are already both populated and this fails [this check](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_response.py#L448). Similarly content_type is present along with a Content Type header that fails [this check](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_response.py#L463).
* Removing the inheritance of Response (so the above values don't get populated during HTTPException initialization) is, as you noted, too invasive, as a few aiohttp components along the way depend on it's behaviour as both Response and Exception.

If there are any suggestions for a better approach I'd be more than happy to implement them :)

panagiks:
I also thought about returning `self` in `build_response()` as self is currently a Response object. But (besides the fact that I'm not sure it would be a good practice since the target is to eventually get rid of the Response dependency) it wouldn't force middlewares to handle the non-duality of the response (i.e. functionalities that depended on response's Exception nature would still work when calling build_response(), although that feature is deprecated through the build_response() interface).

asvetlov:
Making a new `_as_response` in constructor looks... ugly, sorry.

Let's go further.
In my mind `build_response()` should replace returning the `HTTPException` derived class.
#2521 shows my fault in changing everything at once.

But you can get the [main idea](https://github.com/aio-libs/aiohttp/pull/2521/files#diff-95ccfe4e2abcc03ee3ee36a74e6864b7L140) : `headers`, `reason`, `body`, `text` and `content_type` parameters should be deprecated.

Every class like `_HTTPMove` (a base for every `3xx` redirection response) should override `build_response()` method.

In the future (aiohttp 4.0) we need to support the only our own parameters like `location`, `link`, `allowed_methods` etc.

Sorry for long message but sure,  I've got my point.

panagiks:
Wouldn't this change deprecate the `return web.HTTPException()` interface outright?

From #2415 I got the sense that deprecation should come in waves. i.e.:
* in one release add a deprecation warning and a new interface that you will encourage people to use (i.e. `build_response()`)
* in another release fully deprecate the `return web.HTTPException()` interface

I have no stance/opinion/position on whether the deprecation should come in waves, just the sense that I got from the issue mentioned above.

In any case, I'll look into modeling the changes needed here based on the example you provided.

Sorry for the back-and-forth this pr required and thank you for the guideline :)

asvetlov:
I suggest merging this PR with deprecation warning only (with proper documentation update) and making a new PR for `build_response()`.

asvetlov:
thanks!

PR #2517, opened 2017-11-14
MSWord doc was provided rather than a simple txt file

pfreixes:
## What do these changes do?

Get rid of a MSWord file that was provided as a way to document the changes, rather than give a simple text file.

No


pfreixes:
@asvetlov master is kind of broken because of this.

PR #2519, opened 2017-11-15
Move documentation to docs.aiohttp.org

asvetlov:
http://docs.aiohttp.org/en/stable/ already exists BTW

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ Async http client/server framework
    :alt: Latest PyPI package version
 
 .. image:: https://readthedocs.org/projects/aiohttp/badge/?version=latest
-   :target: http://aiohttp.readthedocs.io/en/latest/
+   :target: http://docs.aiohttp.org/
    :alt: Latest Read The Docs
 
 .. image:: https://badges.gitter.im/Join%20Chat.svg
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -226,7 +226,7 @@ class GunicornWebWorker(base.Worker):
                 "Gunicorn's style options in form of `%(name)s` are not "
                 "supported for the log formatting. Please use aiohttp's "
                 "format specification to configure access log formatting: "
-                "http://aiohttp.readthedocs.io/en/stable/logging.html"
+                "http://docs.aiohttp.org/en/stable/logging.html"
                 "#format-specification"
             )
         else:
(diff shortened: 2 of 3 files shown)


samuelcolvin:
Looks good.

What about google webmaster tools and `<link rel="canonical" href="...">` to encourage google to show the right domain?

Or can you do a 301 from http://aiohttp.readthedocs.io? Might be bad for SEO to have "duplicated" content at `docs.aiohttp.org` and `aiohttp.readthedocs.io`?

asvetlov:
Installed google webmaster, waiting for crawler.

popravich:
https?

kxepal:
https://docs.aiohttp.org
```
docs.aiohttp.org uses an invalid security certificate. The certificate is only valid for the following names: *.readthedocs.io, readthedocs.io Error code: SSL_ERROR_BAD_CERT_DOMAIN
```

I don't think that would be simple.

asvetlov:
Unfortunately it doesn't work: rtfd/readthedocs.org#2652

samuelcolvin:
Surely just use cloudflare, it's free and should fix in it with a few seconds work.

(I mean just use their HTTPS proxy in front of readthedocs)

samuelcolvin:
Also on http://docs.aiohttp.org/en/stable/:

```
<link rel="canonical" href="http://aiohttp.readthedocs.io/en/stable/" />
```

Is this because this PR hasn't been released yet?

asvetlov:
@samuelcolvin makes sense, I'll setup it.
Try http://docs.aiohttp.org/en/latest/ for master.
stable is for last tagged release.

asvetlov:
@samuelcolvin cloudflare works like a charm, very thank you: https://docs.aiohttp.org/en/stable/
Just for information: I use cloudflare for both `aiohttp.org` and `aio-libs.org`.

`aio-libs.org` is just a CNAME for `aio-libs.github.io`,  https://github.com/aio-libs/aio-libs.github.io repo.

`aiohttp.org` is hosted on http://netlify.com (thanks again @samuelcolvin) and proxied by cloudflare.
The repo is https://github.com/aio-libs/aiohttp-site
https://docs.aiohttp.org/ is proxied by cloudflare as well.

Looks like Free Plan for both used tools doesn't allow me to add you guys as collaborators -- but I believe we can live with it.

If you need access for git repos for sites -- please ask me.

Future plans are in #2520

samuelcolvin:
Cloudflare's proxy/CDN does not play well with netlify's, and sites are much faster without the cloudflare proxy. I found out the hard way that the best option is to disable cloudflare's proxy. You can still get free automatic ssl, you just have to prompt netlify to generate the ssl cert after disabling cloudflare's proxy.

Doesn't matter much, but I'm pretty sure netlify allows multiple contributors on all tiers.

PR #2523, opened 2017-11-16
Close unclosed client sessions properly

hynek:
## What do these changes do?

`ClientSession.__del__` uses a bare `self.close()` which causes a warning.  This fixes that warning.  I think this should be fine like that?

One warning less.

## Related issue number

n/a

## Checklist


asvetlov:
It's a complex question.
`__del__` is called on session destruction.
It could be even interpreter finalization stage if the session has global scope (we discourage this approach but still support it).

If the loop is closed before `ClientSession.__del__` -- nothing to do.

On other hand the session could be dropped in coroutine context:
```
async def f():
    ...
    del session  # or just implicit decref on frame destruction
```   
In this case the loop is running, `loop.run_until_complete()` is prohibited.

hynek:
Yeah, I suspected it’s not clear cut.

I’d suggest to either drop the close altogether or wrap it in a `with suppress(Exception):` for an best-effort try.  As it’s now it’s kind useless, isn’t it?

asvetlov:
I think dropping is just fine.
Let's keep a warning report part but don't try to close the session: it is not very reliable.

asvetlov:
Thanks!
Backported to 2.3, aiohttp 2.3.3 released

PR #2526, opened 2017-11-17
Rephrase middleware intro

hynek:
The old intro still explained middleware using the deprecated factories.

I also took the liberty to remove a “just” – it’s rather poor style and sounds patronizing.


asvetlov:
Thanks!

PR #2530, opened 2017-11-17
Site and AppRunner

asvetlov:
I'm trying to cleanup infrastructure for web application runners.

The PR introduces a `site` and `AppRunner` concepts.

Fixes #2121 and #2375 
Supersedes #2465

The work is still not finished yet but I believe my intention is clear.

PR #2531, opened 2017-11-17
Drop user supplied loop

asvetlov:
For sake of #2530

The mode is exists for not closing the loop for sake of later reusage, I believe we can invite more robust way by #2530

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -376,15 +376,6 @@ def _make_server_creators(handler, *, loop, ssl_context,
         )
         uris.append('{}://unix:{}:'.format(scheme, path))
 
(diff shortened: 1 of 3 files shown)


kxepal:
So now we always use asyncio global loop and if I want/need to run an app in different one, I should run it in different thread?

asvetlov:
Running two loops in parallel in the same thread is impossible, isn't it?
The program should run one loop per thread (except unittest case), most likely the single loop in main thread per process.

donkahuna:
If I want to change the loop, for example use uvloop ? How can I do this ?

asvetlov:
https://docs.python.org/3/library/asyncio-eventloops.html#asyncio.set_event_loop_policy

Issue #2454, opened 2017-10-31
Middlewares and nested applications


olexandr-klymenko:
## Long story short

There is no clear notion about how middlewares should work in nested applications.
By common sense if middleware is mentioned in sub application it shouldn't be invoked for root application. But as the matter of fact it doesn't work the I expected. Every middleware is applied for every endpoint which is a bit wierd.

## Expected behaviour

middlewares used for declare sub application shouldn't be invoked when root application enpoint is requested

## Actual behaviour

every middleware is applied for every ebdpoint

## Steps to reproduce

1. Create root application
2. Create route in root application
3. Create sub app with middlewares
4. Attach sub app to root application
5. Request root application endpoint

## Your environment

python 3.6.3, aiohttp 2.3.1

asvetlov:
Middlewares are nested too.
If endpoints is handled by subapp -- middlewares from main app will be applied first, than subapp middlewares.

olexandr-klymenko:
I mean this is true the other way around as well. A middleware from sub app
affects on root application

asvetlov:
Hmm. If it's true -- this is a regression and behavior should be fixed.

asvetlov:
Steps to reproduce:

A sub application is not working and it is not very convenient.

```python


from aiohttp import web

async def index(request):
    print('*'*8, request.path, '*'*8)
    print('Index!')
    return web.Response(body=b'Index', content_type='text/plain')


async def login(request):
    print('*' * 8, request.path, '*' * 8)
    print('Login!')
    return web.Response(body=b'Login', content_type='text/plain')

async def cabinet(request):
    print('*' * 8, request.path, '*' * 8)
    print('Cabinet!')
    return web.Response(body=b'Cabinet', content_type='text/plain')


@web.middleware
async def auth_middleware(request, handler):
    print('m' * 8, request.path, 'm' * 8)
    print('Auth middleware!')
    return await handler(request)


root_app = web.Application()
admin_app = web.Application(
    middlewares=[auth_middleware],
)
admin_login_app = web.Application()

root_app.router.add_get('/', index, name='index')
admin_app.router.add_get('/cabinet/', cabinet, name='admin:cabinet')
admin_login_app.router.add_route('*', '/', login, name='login')

admin_app.add_subapp('/login', admin_login_app)
root_app.add_subapp('/admin/', admin_app)


web.run_app(root_app)

```

```bash
$ python server.py 
======== Running on http://0.0.0.0:8080 ========
[...]

PR #2532, opened 2017-11-18
App handle split

NewVadim:
## What do these changes do?

Split Application._handler() by functions.
Fix of list middleware in assert response is instance StreamResponse.

The user can override the behavior a application handler in your own applications.
For example override only middleware.

## Checklist


asvetlov:
I see no strong motivation.
The API is *internal*, user should never override it by custom code -- neither all together nor partially.

NewVadim:
I override all `_handler` to add their own processing middlewares.
Also the logical division of functions will not be superfluous.

Skip middleware example:
```python
async def wrap_middlewares(self, match_info):
    handler = match_info.handler
    skip_middlewares = getattr(handler, '__skip_middlewares__', set())

    for app in match_info.apps:
        for m, new_style in app._middlewares:
            full_path = '{}.{}'.format(m.__module__, m.__name__)
            if '__all__' in skip_middlewares:
                break
            elif full_path in skip_middlewares:
                continue

            if new_style:
                handler = partial(m, handler=handler)
            else:
                handler = await m(app, handler)

    return handler
```
P.S.: I don't insist.

asvetlov:
In my mind skipping not needed middlewares should be done in the following way:
1. Create main app
2. Push every lightweight handler not required by complex middleware stack into one sub application.
3. Put others into another app with corresponding middlewares.

Thus in theory lightweight handlers should be avoided by middlewares from bullet 3.
Maybe it is not true: there is #2454 issue.
I pretty sure I had tests for the case but maybe the behavior is broken.
I didn't test the issue yet -- if we have a broken functionality the situation should be fixed. Could you check?

P.S.
From my perspective the `Application` class is *final*/*sealed* -- no class should derive from it.
If users still inherit from `Application` -- it is a sign of every the concept misunderstanding or lack in aiohttp.web functionality. Let's fix it!

P.P.S.
*Freezing* classes from inheritance is a very good idea, it keeps a place for further modifications of the library. Otherwise we'll get always blocked by some potential backward incompatibilities.
The library is used in the wild, making every implementation detail a part of public API simply stops every improvement: maybe it breaks some code. That's why public API exists.

NewVadim:
A sub application is not working and it is not very convenient.

```python


from aiohttp import web

async def index(request):
    print('*'*8, request.path, '*'*8)
    print('Index!')
    return web.Response(body=b'Index', content_type='text/plain')


async def login(request):
    print('*' * 8, request.path, '*' * 8)
    print('Login!')
    return web.Response(body=b'Login', content_type='text/plain')

async def cabinet(request):
    print('*' * 8, request.path, '*' * 8)
    print('Cabinet!')
    return web.Response(body=b'Cabinet', content_type='text/plain')


@web.middleware
async def auth_middleware(request, handler):
    print('m' * 8, request.path, 'm' * 8)
    print('Auth middleware!')
    return await handler(request)


root_app = web.Application()
admin_app = web.Application(
    middlewares=[auth_middleware],
)
admin_login_app = web.Application()

root_app.router.add_get('/', index, name='index')
admin_app.router.add_get('/cabinet/', cabinet, name='admin:cabinet')
admin_login_app.router.add_route('*', '/', login, name='login')

admin_app.add_subapp('/login', admin_login_app)
root_app.add_subapp('/admin/', admin_app)


web.run_app(root_app)

```

```bash
$ python server.py 
(Press CTRL+C to quit)
******** / ********
Index!
mmmmmmmm /admin/login/ mmmmmmmm
Auth middleware!
******** /admin/login/ ********
Login!
mmmmmmmm /admin/cabinet/ mmmmmmmm
Auth middleware!
******** /admin/cabinet/ ********
Cabinet!
```

NewVadim:
So, i close pull request.

PR #2535, opened 2017-11-19
Drop md5 and sha1 fingerprints

### LINKED ISSUE #2267: Drop md5 and sha1 fingerprints
Should be done in 3.0

asvetlov:
Fix #2267


asvetlov:
@kxepal thanks

Issue #2310, opened 2017-10-10
WebSocketResponse's timeout does not apply to send_bytes/send_json/send_str calls.


frederikaalund:
## Long story short

This is more of a question. Is this behaviour by design?

We have to timeout-related parameters on WebSocketResponse:

* `timeout`
* `receive_timeout`

The `receive_timeout` works for all receive operations. Consequently, I assumed that the `timeout` parameter worked for all send operations. Apparently, I was mistaken. The documentation is not so clear about this.

From the look of the implementation, `timeout` is actually only used in the `close()` call.

## Expected/Actual behaviour

I expected that the `timeout` parameter applied to the `send_bytes`/`send_json`/`send_str` calls. It does not.

## Discussion

Currently, I simply wrap my `send_x` calls in `with Timeout(...)`. It's not that much effort but it would be nice to see this directly implemented in aiohttp.

I can see two options:

* Let `timeout` parameter apply to the `send_x` calls
* Add a `send_timeout` parameter to the WebSocketResponse that applied a timeout to the `send_x` calls

This also relates to #2309. In my actual implementation, the `send_x` calls are submitted to the event loop in a fire-and-forget fashion. Consequently, if the connection is abruptly cut, said `send_x` calls are left lingering and never cancelled. By using a timeout, said calls will time out and not fill up in memory.

## Your environment

Aiohttp 2.2.3 and Python 3.6.1.

asvetlov:
Hmm. Perhaps we need `receive_timeout`, `close_timeout` and `send_timeout`.
At deprecation period `timeout` should be an alias for `close_timeout`.
After deprecation expiring and `timeout` removal (or just at time of aiohttp 4.0 release) we could resurrect `timeout` as default mutable exclusive value for all three params.
@frederikaalund if your would provide a Pull Request -- you are welcome!
I love to see it in aiohttp 3.0 -- we need a time for deprecation period.
Otherwise `timeout` cannot be reused with proper meaning for a long.

frederikaalund:
I'm on it.

Issue #2309, opened 2017-10-10
WebSocketResponse does not throw/close when connection is abruptly cut


frederikaalund:
## Long story short

This is a continuation of #961 (with some ties to #2283). Apparently, the issue is still valid (I don't know if it was ever really fixed).

I have a single websocket handler. It does two things on connection:

* Starts a sender task that sends `N` bytes over the websocket every second
* Receives (and discards) any incoming websocket messages

I connect to this handler through a simple javascript client running in Google Chrome. Everything works as expected: I receive `N` bytes every second, indefinitely.

However, when the connection is abruptly cut (by disconnecting the network cable), the websocket sender task takes 15 minutes to stop (even though the heartbeat timeout is only 5 seconds).

## Expected behaviour

I would expect the `send_bytes` operation to throw *or* the `async for msg in ws` loop to terminate/throw.

## Actual behaviour

The `send_bytes` keep buffering up writes (eventually causing the process to run out of memory and crash). Furthermore, the `async for msg in ws` loop takes 15 minutes to terminate/throw.

## Steps to reproduce

The websocket server:

	#!/usr/bin/env python3.6
	import aiohttp
	import asyncio
	import logging
	from aiohttp import web

	logger = logging.getLogger()
	logger.setLevel(logging.INFO)

	async def sender(ws):
		try:
			logging.info('sender begin')
			while True:
				# Sender never stops on disconnect for N = 10000
				N = 100000
				# Sender stops on disconnect for N = 10
				#N = 10
				logging.info(f'sending {N} bytes')
				try:
					ws.send_bytes(bytes(N))

					# Explicit drain will stall the sender on disconnect
					# but not throw (despite of the heartbeat)
					#await ws.drain()

					# The only fix is to query transport.is_closing()
					# This stops the transport as soon as the heartbeat times out
					#if ws._req.transport.is_closing():
					#	logging.info(f'sender transport is closing')
					#	break
				except Exception as error:
					logging.info(f'sender error:', exc_info=error)
				await asyncio.sleep(1.0)
		# Catch all sender Exceptions
		except Exception as error:
			logging.info('sender error:', type(error))
		finally:
			logging.info('sender end')

	async def handler(request):
		# Note the hearbeat is set to 5 seconds
		ws = web.WebSocketResponse(heartbeat=5.0)
		await ws.prepare(request)

		# Launch send task
		sender_task = asyncio.ensure_future(sender(ws))

		# Pass any message
		logging.info('receiver begin')
[...]

PR #2537, opened 2017-11-19
Add send_timeout, close_timeout, and send_x timeout parameter

frederikaalund:
## What do these changes do?

See the title and the discussion in #2310.

No. The old `timeout` parameter will take precedence over the new `close_timeout` parameter. The default `close_timeout` of 10.0 seconds is still in place.

## Related issue number

#2310

## Checklist


frederikaalund:
I tried writing a functional test for the `send_x` calls. However, I failed to do so. Seemingly, the `send_x` calls return immediately despite my best efforts to stall them. I know from experience that the `send_x` can stall in practice (#2309) but I'm unable to simulate these conditions in a test setting.

Here is my attempt (that doesn't work):

```python
async def test_send_bytes_timeout(loop, test_client):
    closed = loop.create_future()
    raised = False

    async def handler(request):
        ws = web.WebSocketResponse(send_timeout=0.1)
        await ws.prepare(request)

        try:
            # The send_bytes call returns immediately.
            # I also tried with 1 GB of data to defeat buffering. That didn't work.
            await ws.send_bytes(b'0')
            # Calling drain doesn't help either.
            await ws.drain()
        except asyncio.TimeoutError:
            nonlocal raised
            raised = True

        closed.set_result(1)
        await ws.close()
        return ws

    app = web.Application()
    app.router.add_get('/', handler)
    async with client.ws_connect('/') as ws:
        await closed
    assert raised
```

aiohttp/web_ws.py
```diff
@@ -47,8 +47,9 @@ def __init__(self, *,
         self._loop = None
         self._waiting = None
         self._exception = None
-        self._timeout = timeout
         self._receive_timeout = receive_timeout
+        self._send_timeout = send_timeout
+        self._close_timeout = timeout or close_timeout
```
asvetlov, line 56: Please deprecate `timeout`.
Something like
```
if timeout is not None:
    warnings.warn("timeout parameter is deprecated, use close_timeout instead", DeprecationWarning)
    close_timeout = timeout
```

docs/web_reference.rst
```diff
@@ -846,7 +846,8 @@ Response
 WebSocketResponse
 ^^^^^^^^^^^^^^^^^
 
-.. class:: WebSocketResponse(*, timeout=10.0, receive_timeout=None, \
+.. class:: WebSocketResponse(*, timeout=None, receive_timeout=None, \
+                             send_timeout=None, close_timeout=10.0, \
```
asvetlov, line 860: Please describe added params below.

asvetlov (changes requested):
Would you write tests for expired timeouts?

Should we use a timeout for `ping` and `pong`?
Pretty sure yes.

Also please add `send_timeout` etc to `client_ws.py` -- there is no reason to have no such functionality in client web socket as well.

frederikaalund:
Sorry about the delay. I'll try to find time to implement the changes this week. Thanks for the detailed review.

asvetlov:
No problem!

PR #2539, opened 2017-11-19
Added socks proxy support

nibrag:
## What do these changes do?

Added Socks4/Socks5 proxies support

## Checklist


aiohttp/connector.py
```diff
@@ -978,6 +1001,60 @@ def _get_fingerprint_and_hashfunc(self, req):
...
+                    proto=hinfo['proto'], flags=hinfo['flags'],
+                    local_addr=self._local_addr, req=req,
+                    client_error=ClientProxyConnectionError,
+                    server_hostname=req.host if sslcontext else None)
+            except ClientConnectorError as exc:
+                last_exc = exc
+                continue
+
+            try:
+                self._check_fingerprint(req, transp, req.host, req.port)
```
asvetlov, line 1051: The connection is proxies by SOCKS server.
Certeficate, if any, is the SOCKS certificate (I doubt if SOCKS specifies the mode at all).
Why do fingerprint checking?
Fingerprints exist for checking a real server, not proxy.

asvetlov (changes requested):

asvetlov (changes requested):
*Client reference* documentation part should be updated: a new exception added, new meaning for `proxy` and `proxy_auth` params.

asvetlov (changes requested):
Please add functional tests for SOCKS -- like `test_proxy_functional.py`.
Mocked tests are not reliable, we had broken functionality for http proxies many times before added functional tests.

aiohttp/connector.py
nibrag: To connect with socks server `aiosocks` not use the certificate.
First, communication with socks server, then start tls with destination server.

docs/client_reference.rst
```diff
@@ -515,6 +528,9 @@ The client session supports the context manager protocol for self closing.
 
          .. versionadded:: 2.3
 
+      :param bool remote_socks_resolve: Resolve destination host on the remote socks server.
```
asvetlov: Sorry, I still not follow what is the difference between different `remote_socks_resolve` values.
`True` for host DNS resolution on socks server. `False` for DNS resolution on client?
When user should provide non-default parameter?

aiohttp/connector.py
asvetlov: Fingerprint is checked for transport's certificate.
Transport is a connection to SOCKS server.
Thus the check is performed for SOCKS server cert, not proxied host.
Please correct me if I'm wrong.

asvetlov:
@nibrag ping

nibrag:
@asvetlov I'll do it on the weekend.

asvetlov:
thanks!

docs/client_reference.rst
nibrag: > Fingerprint is checked for transport's certificate.

Look at this test:
https://github.com/nibrag/aiohttp/blob/90bf6d9c20570e3e05c92ab02a69ee3c0d8818d7/tests/test_proxy_functional.py#L672-L692

nibrag: > True for host DNS resolution on socks server. False for DNS resolution on client?

Yes. Docs updated.

asvetlov:
Please describe why `socks_remote_resolve` is part of public API?

nibrag:
> Please describe why socks_remote_resolve is part of public API?

I would like to manage this setting for each proxy server (or host) individually.

asvetlov:
Why do you need to manage it at all?
Why resolving on proxy server is not enough?
HTTP proxies for example have no such setting.

nibrag:
Some host names are blocked on proxy server (filter by host).

meejah:
> Why resolving on proxy server is not enough?

This won't work with Tor, for example (where you should always resolve on the remote end) and AFAIK you MUST be able to send unresolved hosts through SOCKS5 (and not resolve on the host at all).

aiohttp/connector.py
Hasimir: Pretty much.  Essentially there's only two possibilities; either:

 1. The TLS/SSL check will be performed on a cert for the proxy server, assuming there is one; or
 
 2. The target server will report what the TLS credentials should be, but the presence of the proxy will trigger a MitM alert of some kind.
 
The latter is most commonly seen with improperly configured transparent proxies.

NotoriousRebel:
any updates on this? @nibrag

kamikaze:
any news?

bitnom:
I'm sure I don't have to explain how important socks support is.

rigens:
> any news?

> I'm sure I don't have to explain how important socks support is.

There is quite stable 3rd party SOCKS implementation for aiohttp: [aiohttp-socks](https://github.com/romis2012/aiohttp-socks)

PR #2541, opened 2017-11-20
Drop await aiohttp.request() syntax for sake of context manager

### LINKED ISSUE #2540: Drop await aiohttp.request(...)
Closing request/session is too error-prone in the case, especially if an exception was generated.

The only robust solution is prohibiting `await request()` syntax but supporting `async with request() as resp:` only.

After implementing proposed restriction we'll get very clean context and behavior.

asvetlov:
Fixes #2540

### DIFF
--- a/CHANGES/2267.removal
+++ b/CHANGES/2267.removal
@@ -1,0 +1,1 @@
+Drop `md5` and `sha1` finger-prints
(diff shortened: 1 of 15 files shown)


asvetlov:
PR is done.
Will merge it tomorrow if nobody objects.

PR #2543, opened 2017-11-20
Forbid synchronous context managers

### LINKED ISSUE #2540: Drop await aiohttp.request(...)

asvetlov:
for ClientSession and test client/server.

Fixes #2540

### DIFF
--- a/CHANGES/2362.feature
+++ b/CHANGES/2362.feature
@@ -1,0 +1,2 @@
+Forbid synchronous context managers for `ClientSession` and test
+server/client.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -711,7 +711,7 @@ class ClientSession:
         self._connector = None
 
     def __enter__(self):
-        raise RuntimeError("Use async with instead")
+        raise TypeError("Use async with instead")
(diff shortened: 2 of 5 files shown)


pfreixes:
LGTM

asvetlov:
Implementation note: I use `raise RuntimeError` instead of just removing `__enter__` methods for better exception text.
Figured out that `TypeError` is even better exception type.

asvetlov:
@pfreixes please next time approve by github UI, not onlt by `LGTM` comment.
Sure, I have administrator privilege for skipping the check but very appreciate the following formal procedure.

pfreixes:
Oki doki, one question can I click merge button but taking into account that must pass the CI pipeline to finally do it? Something like "Merge if CI passes"

asvetlov:
IMHO Merge button becomes enabled only if CI passed.
There is no "delayed merge".

PR #2545, opened 2017-11-22
Don't quote boundary parameter for multipart unless necessary

### LINKED ISSUE #2544: Some servers don't understand quoted boundaries in multipart documents
## Long story short

After a prolonged debugging session I discovered that the usage of quotes in the boundary of a multipart document caused my request to be rejected by the server since it didn't support this.
Preventing the quotation marks from being added in the code makes this work.

The server in question is using nginx, but I don't know which version yet (asked their support).

## Specification

https://tools.ietf.org/html/rfc7231#section-3.1.1.5
https://tools.ietf.org/html/rfc7231#appendix-D
https://tools.ietf.org/html/rfc7230#section-3.2.6

    Content-Type = media-type

    media-type = type "/" subtype *( OWS ";" OWS parameter )
    parameter = token "=" ( token / quoted-string )

    token          = 1*tchar
    quoted-string  = DQUOTE *( qdtext / quoted-pair ) DQUOTE
    qdtext         = HTAB / SP /%x21 / %x23-5B / %x5D-7E / obs-text
    obs-text       = %x80-FF
    quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text )
    tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                       / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                       / DIGIT / ALPHA
                       ; any VCHAR, except delimiters

## Expected behaviour

Specify the multipart boundary without quotes, unless they are required.

## Actual behaviour

Multipart boundary is always quoted.

## Steps to reproduce

Since I don't know which nginx version is affected, the only reliable way to reproduce right now is by filing a request at [share-online's upload API](http://www.share-online.biz/uploadapi).

A sample script is provided below (fill in your data):

<details>

```py
import asyncio
import pathlib

import aiohttp

SESSION_URL = "http://www.share-online.biz/upv3_session.php"
USER = ""
PASS = ""
path = pathlib.Path(".gitignore")


async def async_main():
        data = {'username': USER, 'password': PASS}
        async with session.post(SESSION_URL, data=data) as resp:
            text = await resp.text()
        session_name, session_endpoint = text.split(";")
        session_endpoint = f"http://{session_endpoint}"

        file_size = path.stat().st_size
        with path.open('rb') as fp:
            data = aiohttp.FormData()
            data.add_field('username', USER)
            data.add_field('password', PASS)
            data.add_field('upload_session', session_name)
            data.add_field('chunk_no', "1")
            data.add_field('chunk_number', "1")
            data.add_field('filesize', str(file_size))
            data.add_field('fn', fp, filename=path.name, content_type="application/octet-steam")
            data.add_field('finalize', "1")

            print(f"Starting upload of {path.name}; size: {file_size}")
            async with session.post(session_endpoint, data=data) as resp:
                # print(f"response headers: {resp.headers}")
                resp.raise_for_status()
                text = await resp.text()
                print(f"reply: {text}")


def main():
    task = asyncio.ensure_future(async_main())
    try:
        loop.run_until_complete(task)
    finally:
        loop.close()


if __name__ == '__main__':
    main()
```

</details>

With this script, the upload will fail with `*** EXCEPTION session creation/reuse failed - 11-21-2017, 3:13 pm ***`.

Removing the quotes at https://github.com/aio-libs/aiohttp/blob/439c732f5bf29a83779dce6da4a55e7efd433cf8/aiohttp/multipart.py#L640 makes the request work.

## Your environment

Arch Linux, Python 3.6, aiohttp 2.3.3 (client)

FichteFoll:
## What do these changes do?

Only quote multipart boundary when necessary.
~~Also accept byte-strings, although only in a limited form due to an
architectural restriction.~~
Also properly quote boundaries when they contain `"` or `\` characters.

Boundaries provided by the user, which contain `"` or `\` characters, are properly escaped and invalid characters such as `\n` cause an exception.

## Related issue number

Fixes #2544.

## Checklist

---

Is it okay for me to add myself to `CONTRIBUTORS.txt` with my pseudonym? I saw nobody else doing it, which is why I didn't for now.


aiohttp/helpers.py
```diff
@@ -733,3 +733,35 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
...
+    # qdtext         = HTAB / SP / %x21 / %x23-5B / %x5D-7E / obs-text
+    # obs-text       = %x80-FF
+    # quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text )
+    # tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
+    #                  / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+    #                  / DIGIT / ALPHA
+    #                  ; any VCHAR, except delimiters
+    # VCHAR           = %x21-7E
+    valid_tchar_regex = br"^[!#$%&'*+\-.^_`|~\w]+$"
+    if re.match(valid_tchar_regex, value):
```
asvetlov: Please use compiled regular exception

aiohttp/helpers.py
```diff
@@ -733,3 +733,35 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
...
+        return value
+
+    invalid_qdtext_char_regex = br"[\x00-\x08\x0A-\x1F\x7F]"
+    if re.match(invalid_qdtext_char_regex, value):
```
asvetlov: Compile regex

aiohttp/helpers.py
FichteFoll: Could you explain the rationale for this?

asvetlov: Compiled regexs are faster, isn't it?

aiohttp/helpers.py
```diff
@@ -654,3 +654,35 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
...
+        raise ValueError("parameter value contains invalid characters")
+
+    quoted_value_content = value.replace(b'\\', b'\\\\')  # %x5C
+    quoted_value_content = value.replace(b'"', b'\\"')  # %x22
+
+    return b'"' + quoted_value_content + b'"'
```
kxepal: Why bytes here in and out?

aiohttp/multipart.py
```diff
@@ -633,11 +633,21 @@ class MultipartWriter(Payload):
...
-        ctype = 'multipart/{}; boundary="{}"'.format(subtype, boundary)
+            if isinstance(boundary, bytes):
+                boundary.decode('ascii')
+                self._boundary = boundary
+            else:
+                self._boundary = boundary.encode('ascii')
+        except (UnicodeEncodeError, UnicodeDecodeError):
+            raise ValueError('boundary should contain ASCII only chars') \
+                from None
+        boundary_value = format_parameter_value(self._boundary).decode('ascii')
```
kxepal: Because it looks strange. We take string boundary, turns it into bytes, and then turn it back to stirng to apply formatting.  This wouldn't be a bottleneck for sure, but looks like useless work.

aiohttp/multipart.py
```diff
@@ -633,11 +633,21 @@ class MultipartWriter(Payload):
...
+        # The underlying Payload API demands a str (utf-8), not bytes,
+        # so we need to ensure we don't lose anything during conversion.
+        # As a result, require the boundary to be ASCII only.
+        # In both situations.
         try:
-            self._boundary = boundary.encode('us-ascii')
-        except UnicodeEncodeError:
-            raise ValueError('boundary should contains ASCII only chars')
-        ctype = 'multipart/{}; boundary="{}"'.format(subtype, boundary)
+            if isinstance(boundary, bytes):
```
kxepal: What is the case for binary boundary input?

kxepal: What is the case for binary boundary input?

aiohttp/multipart.py
```diff
@@ -633,11 +633,21 @@ class MultipartWriter(Payload):
...
+                boundary.decode('ascii')
+                self._boundary = boundary
+            else:
```
kxepal: If this `if-else` still exists, would be good to move default boundary generation here into another branch. No point to do isinstance check if boundary passed as None since we know that it will not be a bytes.

aiohttp/multipart.py
```diff
@@ -633,11 +633,21 @@ class MultipartWriter(Payload):
...
+                self._boundary = boundary.encode('ascii')
+        except (UnicodeEncodeError, UnicodeDecodeError):
```
kxepal: I would recommend to minimize exception catch scope.

aiohttp/helpers.py
FichteFoll: The HTTP spec operates on bytes, which is why I designed this helper function to follow that. The only reason why we need to convert back to a string is because the underlying Payload API requires a string for header values. Or at least I think so. If it doesn't, I'll be glad to remove the conversion back to a unicode string. 

 I could of course special-case the function for just our situation here, but it should go into multipart.py then.

aiohttp/helpers.py
FichteFoll: Yes, but only if they are cached properly so they aren't recompiled every time. Python does some internal caching for regular expressions already, so unless we need a lot of regular expressions, the internal cache should cover most use cases.

In fact, if we don't cache the regex and compile it every time we need it, we actually degrade performance compared to having Python (or the re module) handle caching automatically.

aiohttp/helpers.py
FichteFoll: Also, we need to use bytes in the actual multipart body, so we should try to convert exactly that onto a potentially quoted parameter value.

aiohttp/multipart.py
FichteFoll: It would complicate (or at least worsen code beauty) of the next part when we try to catch encoding errors.

aiohttp/multipart.py
FichteFoll: It is true to how it is actually used and defined in the spec. In this particular situations, users don't get more value out of it because we have to convert the boundary back to a unicode string for the Payload header. Otherwise characters above 0x80 could be used inside the boundary. 

Ideally, all headers would work on byte strings (only or additionally), but I didn't consider this to be in the scope of this PR.

aiohttp/helpers.py
asvetlov: I propose to cache regexps at module level, like [here](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L452
`re` module has 512 slots for LRU cache of compiled regexps.
Thus the program may degrade performance if other code uses non-complied exceptions for solving own tasks. If the total regexps number exceeds 512 -- `re` will recompile them constantly.

aiohttp/multipart.py
FichteFoll: *bytes above 0x80 could be used. HTTP is an ascii-readable byte-based protocol.

aiohttp/helpers.py
FichteFoll: This function will only be run once for every multipart document sent, so I believe the performance impact will be minimal at ~~best~~worst, but we can do it anyway if that's your standard.

asvetlov: Thanks.
1. It is the aiohttp standard.
2. The function runs once per multipart document but for high-load site it could clash with other handlers and their functionality.

FichteFoll:
Regarding the header stuff, I think I can implement byte string headers and automatic conversion using `ascii` (and not `utf-8`!) reasonably by converting unicode strings in Payload's `__init__`, if necessary, and then changing `MultipartWriter.append_payload` to expect the headers to be byte strings already. I haven't looked at where else I would need to implement this especially regarding the main request/response's and would rather not persue that within this PR, but Payload seems to only be used by multipart related code (and its headers property is only accessed within MultipartWriter).

Regarding the motivation of that suggested change:
To my knowledge, the entire HEAD part (also in multipart) of HTTP is byte-based and only characters in the ASCII range are used by indicators (i.e. header names, values in the IANA encoding registry, etc.) while other bytes >=0x80 may be used in some locations, for example in the boundary parameter here. A such, I believe this internal representation should be reached as early as possible and only public API should accept unicode strings in the ASCII range because those translate to a single byte encoding transparently and to maintain backwards compatiblity, obviously. It's also more intuitive for most users because byte strings *seem* like an abstract concept to many, I think.

Also, what about my `CONTIBUTORS.txt` question?

fafhrd91:
Aiohttp treats header value as an utf-8. I think a lot of servers does the same nowadays.

asvetlov:
Yes, historically aiohttp uses utf-8 for headers with a backdoor for getting bytes for communication with malformed client/server, maybe we need a hooks for sending custom bytes too.

Anyway for me there is no reason for switching to byte-based headers, especially for boundary -- the value perfectly fits in ascii, isn't it?

FichteFoll:
No, technically bytes 0x80 to 0xff are allowed as well. When quoted, that is. It's the obs-text (or char?) rule in the abnf I included in the issue and in the comment for the helper function. (I'm on mobile, otherwise I'd quote it myself.)

Again, if you don't feel good about the weird conversion between bytes and unicode, I can remove it and it won't be backwards incompatible. It wouldn't allow any bytes higher than 0x7f, but the current implementation doesn't as well and it's perfectly backwards compatible. It just bothers me in a way that the specification isn't respected entirely.

asvetlov:
@FichteFoll don't get me wrong.
The PR exists for fixing current 100% correct implementation for sake of better compatibility with software which is in turn not 100% follows the specification.
I have a guts feeling that allowing non-ascii boundary will break other software (but pretty good from RFC). Let's restrict boundary to ascii on multipart message generation -- this way is:
1. Backward compatible with current aiohttp implementation
2. Cannot break other software

We can return to discussion if somebody will complain aiohttp in not supporting non-ascii boundaries.

FichteFoll:
I agree with this sentiment. Will adjust the PR tomorrow.

Do you want me to squash? Otherwise I'd just commit on top of this so we can keep the other commit in case this ever comes up. It might end up being useful, maybe.

asvetlov:
Just merge master and add required changes.
PR is squashed before merging, PR's history doesn't matter.

aiohttp/helpers.py
```diff
@@ -654,3 +654,35 @@ def content_length(self, *, _CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
...
+    #                  / DIGIT / ALPHA
+    #                  ; any VCHAR, except delimiters
+    # VCHAR           = %x21-7E
+    valid_tchar_regex = br"^[!#$%&'*+\-.^_`|~\w]+$"
+    if re.match(valid_tchar_regex, value):
+        return value
+
+    invalid_qdtext_char_regex = br"[\x00-\x08\x0A-\x1F\x7F]"
+    if re.match(invalid_qdtext_char_regex, value):
+        raise ValueError("parameter value contains invalid characters")
```
asvetlov: The line is not covered by tests

tests/test_multipart.py
```diff
@@ -945,9 +945,21 @@ def test_default_subtype(self):
...
+    def test_quoted_boundary(self):
+        writer = aiohttp.multipart.MultipartWriter(boundary=R'\"')
+        self.assertEqual({CONTENT_TYPE: R'multipart/mixed; boundary="\\\""'},
+                         writer.headers)
+
     def test_bad_boundary(self):
         with self.assertRaises(ValueError):
             aiohttp.multipart.MultipartWriter(boundary='тест')
+        with self.assertRaises(ValueError):
+            aiohttp.multipart.MultipartWriter(boundary='\n')
```
kxepal: How about `boundary=test\n`? (;

aiohttp/multipart.py
```diff
@@ -657,6 +664,41 @@ def __iter__(self):
...
+        # tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
+        #                  / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+        #                  / DIGIT / ALPHA
+        #                  ; any VCHAR, except delimiters
+        # VCHAR           = %x21-7E
+        value = self.boundary
+        if re.match(self._valid_tchar_regex, value):
+            return value.decode('ascii')  # cannot fail
+
+        if re.match(self._invalid_qdtext_char_regex, value):
```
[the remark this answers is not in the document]
FichteFoll: Good call. Fixed that and another issue with the previous regex not considering that `$` also matches before a newline.

kxepal:
@FichteFoll, Thanks! @asvetlov?

docs/multipart_reference.rst
```diff
@@ -161,8 +161,12 @@ Multipart reference
 
    Multipart body writer.
 
+   ``boundary`` may be an ASCII-only string.
+
    .. attribute:: boundary
 
+      The byte-string representation of the boundary.
```
asvetlov, line 168: The type bothers me.
We pass optional `str` value into constructor but the property is byteish.
Maybe the method should return `self._boundary.decode('ascii')`?
It is a breaking change but we could do it in aiohttp 3.0.
Amount of affected user code is very low IMHO.

FichteFoll: This behavior is unchanged. I just documented it.

https://github.com/aio-libs/aiohttp/pull/2545/files#diff-235eeadd1960735e1191879aa500b7b7L637

asvetlov: Yes, I see.
But maybe now is the time for behavior changing?
Current one is confusing a little.

FichteFoll: Well, imo everything should be saved as byte-strings anyway (as I explained before), so having the boundary as a byte-string doesn't seem far off. It's only used that way, too, so either we'd add an additional property that converts the boundary from a unicode to a byte string, or we encode it at every point we need it (not a fan).

I'd say we leave this for now. I would rather spend time to convert all header strings to byte-strings rather than unwrapping the boundary here.

kxepal: I fear it was bytes all the time. Let's change that for 3.0! This could be done with separate PR with breaking change notice.

asvetlov: Ok, let's merge.
Technically single Pull Request can have several records in `CHANGES/` but I'm ok with separate PR.

asvetlov: @FichteFoll I have a feeling that all headers should be either strings or bytes.
All means really **all**, including `request.headers` etc.
This change in too intrusive, the ship has sailed many years ago.
Also I think 99% aiohttp users are happy with string headers but byte header values will confuse them.

asvetlov:
@FichteFoll thanks!

docs/multipart_reference.rst
FichteFoll: I agree that this should ideally be standardized on a single header type. For compatibility, all unicode strings *could* be converted to byte-strings using ASCII encoding, so that the outgoing API doesn't change much except that it allows byte-strings for all headers too.

However, this goes way beyond the scope of this PR and would be a large architectural consideration that I'm sure you're going to have a much better time deciding on. It might be something worth considering for the next major release, i.e. 3.0, as long as someone puts in the work.

PR #2550, opened 2017-11-23
Root app middleware receives wrong app context

popravich:
## Long story short

When handling request to sub-application root application's middlewares see wrong
`request.app` instance — sub-app instead of the one they attached to.

This PR demonstrates it (adds failing test) but has no fix yet.

### DIFF
--- a/CHANGES/2550.bugfix
+++ b/CHANGES/2550.bugfix
@@ -1,0 +1,1 @@
+Make `request.app` point to proper application instance when using nested applications (with middlewares).
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -22,6 +22,7 @@ from .signals import Signal
 from .web_exceptions import *  # noqa
 from .web_fileresponse import *  # noqa
 from .web_middlewares import *  # noqa
+from .web_middlewares import _fix_request_current_app
 from .web_protocol import *  # noqa
 from .web_request import *  # noqa
 from .web_response import *  # noqa
@@ -267,6 +268,8 @@ class Application(MutableMapping):
                               'see #2252'.format(m),
                               DeprecationWarning, stacklevel=2)
                 yield m, False
+        if self._middlewares:
+            yield _fix_request_current_app(self), True
 
     async def _handle(self, request):
         match_info = await self._router.resolve(request)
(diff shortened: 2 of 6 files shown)


asvetlov:
Is it related to #2454 ?

popravich:
not sure

popravich:
`request.app` property always returns the last app in `match_info.apps` list.
We need to fix this.

popravich:
So, the fix is as following:
* make `request.app` normal property, not `reify`'ed attribute;
* make `request.app` return `match_info.current_app`
* add `match_info.current_app` property which, by default, returns last app (same as `request.app` did).
* add a context manager to match_info that updates `current_app`;
* update application's `_prepare_middleware` method to append one extra middleware
  `_fix_request_current_app` that switches current application in match info and, as a result, in request.

Whew... profit!

Guys, your thoughts?

asvetlov:
Adding a new middleware is not ideal maybe but I can live with it -- an extra middleware is better than broken functionality.
We can invite a better fix later anyway.

SoumenDass:
This change has broken existing functionality e.g. earlier within a view in a sub-application, I could always do request.app and get the sub-application instance. Now it always returns the main application instance which basically defeats the purpose of sub-application. E.g. if I have a routes defined in the sub-application, earlier they used to work fine via request.app.router['<route_name>']. Now since request.app always returns the main app, it is not aware of the subapplication route names and errors out:

```
    rpt_url = request.app.router['getreport'].url_for(rpt_id=rpt_id)
  File "/home/sdass/apivenv/lib/python3.6/site-packages/aiohttp/web_urldispatcher.py", line 851, in __getitem__
    return self._named_resources[name]
KeyError: 'getreport'
```

popravich:
Hi, @SoumenDass ,
does your applications (main and sub) have any middlewares setup?

SoumenDass:
All my middlewares are in the main app

popravich:
ok, thanks, I will check your issue

SoumenDass:
Thanks - for now, I've rolled back my deployment to 2.3.3 where this issue does not happen.

popravich:
I've just updated the test and made a fix.
Indeed, if sub-app has no middleware it was accessing main app instance.

Issue #2555, opened 2017-11-24
Simplify stream classes


asvetlov:
- [x] drop `ChunksQueue` -- not used by aiohttp code
- [x] drop `FlowControlChunksQueue` -- not used by aiohttp code
- [x] Merge `FlowControlStreamReader` into `StreamReader`, drop `FlowControlStreamReader`. The reader without flow control is not used by aiohttp, `FlowControlStreamReader` is too long name. Getting rid of `await super()` calls may boost performance a little.
- [x] Add hysteresis to `StreamReader` to prevent flipping between paused and resumed states too often.

kxepal:
FlowControlChuckQueue is useful when server emits item-per-chunk - that's quite handy to process. CouchDB does that for instance. Let me check if this approach is still actual.

asvetlov:
If CouchDB needs `FlowControlChuckQueue` -- please copy the implementation into the project.
Supporting not used classes is a burden, you know.

kxepal:
Not a problem, I know (: Probably, that even wouldn't be need after switching to ijson.

kxepal:
Just wanted to describe the case when that flow was useful. Some else users eventually may use it as well.

asvetlov:
Fixed by @2256

PR #2556, opened 2017-11-24
Simplify streams

asvetlov:
Fix for #2555


aiohttp/streams.py
```diff
@@ -112,8 +114,10 @@ def __repr__(self):
             info.append('%d bytes' % self._size)
         if self._eof:
             info.append('eof')
-        if self._limit != DEFAULT_LIMIT:
-            info.append('l=%d' % self._limit)
+        if self._low_water != DEFAULT_LIMIT:
+            info.append('low=%d' % self._low_water)
+        if self._high_water != DEFAULT_LIMIT * 2:
```
kxepal: May be always show low/high watter in repr? Otherwise it's not oblivious: it has no limits or just some hidden default ones?

aiohttp/streams.py
```diff
@@ -265,7 +274,7 @@ def end_http_chunk_receiving(self):
                 if ichar:
                     not_enough = False
 
-                if line_size > self._limit:
+                if line_size > self._high_water:
```
kxepal, line 275: Previously, max line size was `2 ** 16` by default, but not it's `2 ** 17`. Is that intentional?

asvetlov: Yes, it is intentional.
Line size should be less than high watermark is my mind -- was larger than low watermark.

aiohttp/streams.py
asvetlov: Done

aiohttp/streams.py
kxepal: Ok.

asvetlov:
@kxepal all your notes are fixed.

PR #2559, opened 2017-11-25
Websocket handshake cleanup

asvetlov:
`do_handshake` is very private API, let's move the implementation into `web.WebSocketResponse` -- the only user of the function.

### DIFF
--- a/aiohttp/http.py
+++ b/aiohttp/http.py
@@ -9,7 +9,8 @@ from .http_parser import (HttpParser, HttpRequestParser, HttpResponseParser,
                           RawRequestMessage, RawResponseMessage)
 from .http_websocket import (WS_CLOSED_MESSAGE, WS_CLOSING_MESSAGE, WS_KEY,
                              WebSocketError, WebSocketReader, WebSocketWriter,
-                             WSCloseCode, WSMessage, WSMsgType, do_handshake)
+                             WSCloseCode, WSMessage, WSMsgType, ws_ext_gen,
+                             ws_ext_parse)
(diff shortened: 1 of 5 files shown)


aiohttp/http.py
```diff
@@ -27,7 +28,7 @@
 
     # .http_websocket
     'WS_CLOSED_MESSAGE', 'WS_CLOSING_MESSAGE', 'WS_KEY',
-    'WebSocketReader', 'WebSocketWriter', 'do_handshake',
+    'WebSocketReader', 'WebSocketWriter', 'ws_ext_gen', 'ws_ext_parse',
```
kxepal, line 31: Why expose `ws_ext_gen` and `ws_ext_parse` then?

asvetlov: It's a part of internal API, why not.

My main intention for `do_handshake` changes is not preventing exposing new functions but not converting `http` exceptions into `web` ones.
PR raises proper exceptions directly. I believe it simplifies understanding a little, plus stack traces are shorter.

kxepal: Aha, I see. Thanks!

kxepal:
Oh, is that [reference](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_ws.py#L98) is about just dropped `do_handshake()`?

asvetlov:
The code is for do nothing on second `prepare` call.
Not related to `do_handshake()` details.

kxepal:
Ok, good then.

PR #2560, opened 2017-11-25
Client signals get rid of trace_request_ctx as signal kw

pfreixes:
## What do these changes do?

Try to reduce the verbosity of the signal's handler, before this PR all signals handlers had to expect a `trace_request_ctx` kw param. The proposal of this param was and is give a way to the developer for passing information that is only available at request time.

This PR moves wire the `trace_request_ctx` kw param that can be passed at each request as a new an unique param when the `trace_config_ctx` is built. Therefore, by default this param will be available under the `trace_config_ctx.trace_request_ctx` attribute.

For example:

```python
async def on_request_end(session, trace_config_ctx, resp):
    assert trace_config_ctx.trace_request_ctx == 'foo'

trace_config = TraceConfig()
trace_config.on_request.end.append(on_request_end)
sesion = ClientSession(trace_configs=[trace_config])
await session.get("http://python.org", trace_request_ctx='foo')
```
 

Yes

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -226,9 +226,11 @@ class ClientSession:
 
         traces = [
             Trace(
-                trace_config,
                 self,
-                trace_request_ctx=trace_request_ctx)
+                trace_config,
+                trace_config.trace_config_ctx(
+                    trace_request_ctx=trace_request_ctx)
+            )
             for trace_config in self._trace_configs
         ]
 
--- a/aiohttp/tracing.py
+++ b/aiohttp/tracing.py
@@ -10,7 +10,7 @@ class TraceConfig:
     """First-class used to trace requests launched via ClientSession
     objects."""
 
-    def __init__(self, trace_config_ctx_class=SimpleNamespace):
+    def __init__(self, trace_config_ctx_factory=SimpleNamespace):
         self._on_request_start = Signal(self)
         self._on_request_end = Signal(self)
         self._on_request_exception = Signal(self)
@@ -25,11 +25,11 @@ class TraceConfig:
         self._on_dns_cache_hit = Signal(self)
         self._on_dns_cache_miss = Signal(self)
 
-        self._trace_config_ctx_class = trace_config_ctx_class
+        self._trace_config_ctx_factory = trace_config_ctx_factory
 
     def trace_config_ctx(self, trace_request_ctx=None):
         """ Return a new trace_config_ctx instance """
-        return self._trace_config_ctx_class(
+        return self._trace_config_ctx_factory(
             trace_request_ctx=trace_request_ctx)
 
     def freeze(self):
--- a/aiohttp/tracing.py
+++ b/aiohttp/tracing.py
@@ -27,9 +27,10 @@ class TraceConfig:
 
         self._trace_config_ctx_class = trace_config_ctx_class
 
-    def trace_config_ctx(self):
+    def trace_config_ctx(self, trace_request_ctx=None):
         """ Return a new trace_config_ctx instance """
-        return self._trace_config_ctx_class()
+        return self._trace_config_ctx_class(
+            trace_request_ctx=trace_request_ctx)
 
     def freeze(self):
         self._on_request_start.freeze()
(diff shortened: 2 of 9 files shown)


aiohttp/client.py
```diff
@@ -226,9 +226,11 @@ def request(self, method, url, **kwargs):
 
         traces = [
             Trace(
-                trace_config,
                 self,
-                trace_request_ctx=trace_request_ctx)
+                trace_config,
+                trace_config.trace_config_ctx(
```
kxepal, line 231: Why trace config context is separated from trace_config object?

pfreixes: There is a `trace_config_ctx` object instantiated per request, each handler receives as a param this instance to save information during the signals triggered by the same request.

The TraceConfig instance is registered as a param for the `ClientSession`, therefore the relationship between the config and the context is 1:N where N means requests.

kxepal: What if we put trace config inside context once context is created? Seems like they are bounded, so better not pass them around separated. 

Is there any cases when we could change context / context in the middle of request processing?

pfreixes: IMHO I would prefer to keep the `trace_config` as an explicit argument of the signal handler. Both `trace_config` and `trace_config_ctx` have different purposes. Meanwhile, the first one is the class that allows you configure the signals that would be triggered, the second one is a pure temporary storage used to make the trace of a specific request stateful.

Change in the meaning of creating a new instance? No, I don't think so. Take into account that the pattern that we are allowing having this parameter is the following one

```python
ascync def start(session, trace_config, trace_config_ctx, *args, **kwargs):
    tace_config_ctx.start = session.loop.time()

ascync def end(session, trace_config, trace_config_ctx. *args, **kwargs):
    print("Total time {}".format(session.loop.time() - trace_config_ctx.start)
```

pfreixes: thoughts ?

kxepal: Well, for me they looks like fragments of a single thing: `trace_config`.  Assume we have `TraceConfig` object. It has some attributes, methods, properties...well, everything about to configure the tracer as the name stands. At the same time it may wraps a block of code into the context, where some different rules shifts the original state, but only for that block.

I didn't dive into design deep enough to judje, but it just looks like that we can have a single `trace_config` object with `ctx` property or have standard context manager protocol with a bit help of `__call__` act in the similar way. But again, just a thoughts. And all because naming tells us that these two things are heavily coupled - probably too much to think that they could be the single entity.

kxepal: Anyway, this discussion is irrelevant to PR. It should be started on initial tracer PR, but there were too much changes to stop on every bit.

pfreixes:
can we unblock this? /cc @asvetlov

asvetlov:
Will review tomorrow.
Sorry, was busy by bugfixing.

jettify:
@pfreixes @kxepal I am working on aiozipkin integration with aiohttp singnals, can you please take a look? https://github.com/aio-libs/aiozipkin/pull/42 I am using `trace_request_ctx`, will update my code once this PR merged.

pfreixes:
@jettify if you can provide feedback about usability and what you would like to have to improve the way tracing is handled will help us to improve the current design.

I'm still a bit reluctant about how the `trace_request_ctx` is handled, despite this PR reduces the verbosity of the signal handler is still necessary to pollute the caller, having something like `session.get(...., trace_request_ctx=value)`.

But this is IMHO a cosmetic discussion and it's up to the user who writes his own trace config used to trace his own requests to use the `trace_request_ctx` to share information that is only available in request time with the trace config handlers.

For third libraries, the intention is avoiding the use of this parameter. The user must be able to install, configure and forget it. The usage of the `trace_request_ctx` will create friction and will make more complex the maintainability of the code. Also worth mentioning that the use of this kw for more than one `TraceConfig` can be a nightmare.

So, how to pass information that is only available at request time for third-party libraries without polluting the `trace_request_ctx`? either wait for the context variables that will come with Python 3.7 or use an early implementation such as aiotask-context [1]

In any case, as I said I'm still a bit reluctant.  

[1] https://github.com/Skyscanner/aiotask-context

asvetlov:
Looks good.
Minor note: is `trace` prefix required for classes in `tracing.py`?
E.g. `TraceConfig.trace_config_ctx()` could be renamed to `TraceConfig.config_ctx()`, `trace_request_ctx` is just `request_ctx`, `self._trace_config_ctx_class` is `self._config_ctx_class` or event better `self._config_ctx_factory` -- technically it could be any callable with `request_ctx` argument.

Names outside of `tracing.py` like `trace_config` in client.py` are perfect.

pfreixes:
Renamed `trace_config_ctx_class` to `trace_config_ctx_factory` for consistency.

If nobody objects once it passes the CI steps if I have permissions I will merge it as it is, FYI @jettify

asvetlov:
@pfreixes sorry, I've pushed "Merge" before you.

jettify:
@pfreixes thanks! I decided to go with explicit metadata propagation, as next step will add task local stuff.

PR #2567, opened 2017-11-28
Fix connector

asvetlov:
Prevent hang on connection acquiring after connection failure

### DIFF
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -343,6 +343,28 @@ def test_release_close(loop):
     assert proto.close.called
 
 
+def test__drop_acquire_per_host1(loop):
+    conn = aiohttp.BaseConnector(loop=loop)
+    conn._drop_acquired_per_host(123, 456)
+    assert len(conn._acquired_per_host) == 0
+
+
+def test__drop_acquire_per_host2(loop):
+    conn = aiohttp.BaseConnector(loop=loop)
+    conn._acquired_per_host[123].add(456)
+    conn._drop_acquired_per_host(123, 456)
+    assert len(conn._acquired_per_host) == 0
+
+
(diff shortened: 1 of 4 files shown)


tests/test_connector.py
```diff
@@ -343,6 +343,28 @@ def test_release_close(loop):
     assert proto.close.called
 
 
+def test__drop_acquire_per_host1(loop):
```
webknjaz, line 346: @asvetlov is this `double__underscore` notation intentional?

asvetlov: Yes. It is a test for `_drop_acquire_per_host` method.

pfreixes:
Why the clients that are waiting for an available connection hang in case of the previous one had an error? I cant see this bad behavior in the current code. Could you please point me to the right path?

Also, I don't get the point of the  `test_error_in_performing_request` test, how is it testing that situation?

asvetlov:
self._waiters was not cleaned up, https://github.com/aio-libs/aiohttp/blob/2.3/aiohttp/connector.py#L357 said that `available` is not positive.
`connect()` method created a waiter but no other connection was exist, nobody set resumed the waiter future.

That's why `test_error_in_performing_request` failed.

I need review the logic for `available`, looks like `len(self._waiters)` is a wrong check.

asvetlov:
Dropped a check for `len(self._waiters)`, `self._acquired` is enough.

pfreixes:
Oks, got it. LGTM can be merged once all checks passes

pfreixes:
IMHO this kind of bug is enough important to have a bugfix release for the 2.X series

asvetlov:
@pfreixes will publish 2.3.4 just after merging

PR #2575, opened 2017-11-30
PEP fixes

thehesiod:
- don't use private attributes
- mark where abstract methods need to be implemented
- don't use shadow var name from global scope
- use snake case


thehesiod:
@asvetlov looks like there's a test framework issue:
`TypeError: unhashable type: 'HTTPMethodNotAllowed'`

asvetlov:
Yes, it is the regrssion: #2579

asvetlov:
Sorry, accidental closing

asvetlov:
Something is still wrong.
@thehesiod could you take a look?

asvetlov:
Thanks!

thehesiod:
TY!

PR #2577, opened 2017-11-30
Correct request.app context (for handlers not just middlewares)

popravich:
This is an update to #2550 
`_fix_request_current_app` middleware was added if application had any middlewares.
Now it is yielded unconditionally to set proper context for handlers as well.

### DIFF
--- a/CHANGES/2577.bugfix
+++ b/CHANGES/2577.bugfix
@@ -1,0 +1,1 @@
+Correct `request.app` context (for handlers not just middlewares).
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -268,8 +268,7 @@ class Application(MutableMapping):
                               'see #2252'.format(m),
                               DeprecationWarning, stacklevel=2)
                 yield m, False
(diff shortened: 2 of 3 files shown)


popravich:
@asvetlov , I'm not sure about target branch — should it be master or some bugfix branch?

asvetlov:
I'll cherry-pick to 2.3 myself

asvetlov:
thanks!

pfreixes:
Just FYI this fix impacted with a loss btw 5%-10% of the performance, we should try to make it in some way that does not reuse the middlewares code to setup also the current app when there are no middlewares installed.

amirouche:
Does it make sens to optimize for the case where there is no middleware?

asvetlov:
The check should analyze not only middlewares but nested applications also.
Pull Request is welcome!

popravich:
The performance loss might be introduced by making `Request.app` a real property and not cached one.
So I guess optimizing for the case when there are no middlewares might not bring those percents back.

Issue #2580, opened 2017-12-01
Where is the `proxies` parameter of ClientSession? And how could I add proxy for whole session?


fy0:
## Long story short

Document:

http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientSession

```python
class aiohttp.ClientSession(*, connector=None, loop=None, ..., proxies=None)
```

Real world:

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L53-L62

Well, what happened?

## Expected behaviour

-

## Actual behaviour

-

## Steps to reproduce

-

## Your environment

aiohttp 2.3.5

fy0:
working on fix

asvetlov:
Please do

RatanShreshtha:
Please merge the PR for creating sessions with the proxy.

PR #2582, opened 2017-12-01
Added proxy parameters for ClientSession

fy0:
## What do these changes do?

Added proxy parameters for ClientSession

New parameters of ClientSession\():
* proxy
* proxy_auth
* proxy_headers

```python
async with ClientSession(proxy=proxy) as session:
    async with session.get('http://test.com') as resp:
        print(await resp.text())
```

## Related issue number

#2580 

## Checklist

### DIFF
--- a/CHANGES/2580.feature
+++ b/CHANGES/2580.feature
@@ -1,0 +1,1 @@
+Added proxy parameters for ClientSession.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -74,6 +74,7 @@ Erich Healy
 Eugene Chernyshov
 Eugene Naydenov
 Evert Lammerts
+Fan Yu
 Frederik Gladhorn
 Frederik Peter Aalund
 Gabriel Tremblay
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -165,13 +165,13 @@ class ClientSession:
                        chunked=None,
                        expect100=False,
                        read_until_eof=True,
-                       proxy=NotImplemented,
-                       proxy_auth=NotImplemented,
+                       proxy=sentinel,
+                       proxy_auth=sentinel,
                        timeout=sentinel,
                        verify_ssl=None,
                        fingerprint=None,
                        ssl_context=None,
-                       proxy_headers=NotImplemented,
+                       proxy_headers=sentinel,
                        trace_request_ctx=None):
 
         # NOTE: timeout clamps existing connect and read timeouts.  We cannot
@@ -203,13 +203,13 @@ class ClientSession:
 
         # Merge with session proxy
         proxy = self._proxy_info['proxy'] \
-            if proxy == NotImplemented else proxy
+            if proxy is sentinel else proxy
 
         proxy_auth = self._proxy_info['auth'] \
-            if proxy_auth == NotImplemented else proxy_auth
+            if proxy_auth is sentinel else proxy_auth
 
         proxy_headers = self._proxy_info['headers'] \
-            if proxy_headers == NotImplemented else proxy_headers
+            if proxy_headers is sentinel else proxy_headers
 
         # Merge with default headers and transform to CIMultiDict
         headers = self._prepare_headers(headers)
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -59,7 +59,8 @@ class ClientSession:
                  cookie_jar=None, connector_owner=True, raise_for_status=False,
                  read_timeout=sentinel, conn_timeout=None,
                  auto_decompress=True, trust_env=False,
-                 trace_configs=None):
+                 trace_configs=None,
+                 proxy=None, proxy_auth=None, proxy_headers=None):
 
         implicit_loop = False
         if loop is None:
@@ -110,6 +111,11 @@ class ClientSession:
         self._raise_for_status = raise_for_status
         self._auto_decompress = auto_decompress
         self._trust_env = trust_env
+        self._proxy_info = {
+            'proxy': proxy,
+            'auth': proxy_auth,
+            'headers': proxy_headers,
+        }
 
         # Convert to list of tuples
         if headers:
@@ -159,13 +165,13 @@ class ClientSession:
                        chunked=None,
                        expect100=False,
                        read_until_eof=True,
-                       proxy=None,
-                       proxy_auth=None,
+                       proxy=NotImplemented,
+                       proxy_auth=NotImplemented,
                        timeout=sentinel,
                        verify_ssl=None,
                        fingerprint=None,
                        ssl_context=None,
-                       proxy_headers=None,
+                       proxy_headers=NotImplemented,
                        trace_request_ctx=None):
 
         # NOTE: timeout clamps existing connect and read timeouts.  We cannot
@@ -194,6 +200,16 @@ class ClientSession:
         redirects = 0
         history = []
         version = self._version
+
+        # Merge with session proxy
+        proxy = self._proxy_info['proxy'] \
+            if proxy == NotImplemented else proxy
+
+        proxy_auth = self._proxy_info['auth'] \
+            if proxy_auth == NotImplemented else proxy_auth
+
+        proxy_headers = self._proxy_info['headers'] \
+            if proxy_headers == NotImplemented else proxy_headers
 
         # Merge with default headers and transform to CIMultiDict
         headers = self._prepare_headers(headers)
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -47,7 +47,8 @@ The client session supports the context manager protocol for self closing.
                          conn_timeout=None, \
                          raise_for_status=False, \
                          connector_owner=True, \
-                         auto_decompress=True, proxies=None)
+                         auto_decompress=True, \
+                         proxy=None, proxy_auth=None, proxy_headers=None)
 
    The class for creating client sessions and making requests.
 
--- a/tests/test_proxy_session.py
+++ b/tests/test_proxy_session.py
@@ -1,0 +1,126 @@
+import asyncio
+import os
+from unittest import mock
+
+import pytest
+from yarl import URL
+
+import aiohttp
+from aiohttp import web
+
+
+@pytest.fixture
+def proxy_test_server(raw_test_server, loop, monkeypatch):
+    """Handle all proxy requests and imitate remote server response."""
+
+    _patch_ssl_transport(monkeypatch)
+
+    default_response = dict(
+        status=200,
+        headers=None,
+        body=None)
+
+    proxy_mock = mock.Mock()
+
+    async def proxy_handler(request):
+        proxy_mock.request = request
+        proxy_mock.requests_list.append(request)
+
+        response = default_response.copy()
+        if isinstance(proxy_mock.return_value, dict):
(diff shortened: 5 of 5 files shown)


asvetlov:
Any test?

fy0:
@asvetlov OK, the code is very simple, but there is something different.
I use `NotImplemented` as a placeholder instead of `None`, because I thought sometimes we need send request **without proxy** when a proxy applied to the whole session.

```
async def func():
    async with ClientSession(conn_timeout=5, read_timeout=10, proxy=proxy) as session:
        async with session.get('http://test.com') as resp:
            print(await resp.text())

        async with session.get('http://test.com/2', proxy=None) as resp:
            print(await resp.text())

        async with session.get('http://test.com/3') as resp:
            print(await resp.text())
```

But actually it's not working, even value of `proxy` is None in function, it still access proxy three times.

Something went wrong, so I tested another example:

```
async def func2():
    async with ClientSession(conn_timeout=5, read_timeout=10) as session:
        async with session.get('http://test.com/1') as resp:
            print(await resp.text())
            #print(resp.status)

        async with session.get('http://test.com/2', proxy=proxy) as resp:
            print(await resp.text())
            #print(resp.status)

        async with session.get('http://test.com/2', proxy=proxy) as resp:
            print(await resp.text())
            #print(resp.status)
```

This time the proxy was not used, so I guess first request of session decided proxy of whole session and can not be changed?

Emmmmmm...

BTW, if you don't like the `NotImplemented` as placeholder, I'll change it to None.

I will added tests soon.

asvetlov:
`helpers.sentinel` is even better.

Design thoughts.
`requests` accepts a `dict` for `proxies`:
```
  #: Dictionary mapping protocol or protocol and host to the URL of the proxy
        #: (e.g. {'http': 'foo.bar:3128', 'http://host.name': 'foo.bar:4012'}) to
        #: be used on each :class:`Request <Request>`.
```
Maybe we need multiple proxies support too?

fy0:
@asvetlov 
I think keep it simple is better.
For complex situation, we can provide something like `ProxySwitchHandler`.

```python
class ProxySwitchHandler:
    def choose_proxy(self, url):
        # ...
        return 'foo.bar:3128'
```

fy0:
@asvetlov I'm not sure `proxy_headers` works or not. Nothing happened to `proxy.headers`  even if `sess.get(url, proxy_headers={'foo': 'bar'})` be called.

```python
async def test_session_proxy_http_headers(proxy_test_server, loop):
    url = 'http://aiohttp.io/path'
    proxy = await proxy_test_server()

    conn = aiohttp.TCPConnector(loop=loop)
    sess = aiohttp.ClientSession(connector=conn, loop=loop, proxy=proxy.url)
    await sess.get(url)
    assert 'ProxyFlag' not in proxy.request.headers

    conn = aiohttp.TCPConnector(loop=loop)
    sess = aiohttp.ClientSession(connector=conn, loop=loop, proxy=proxy.url, proxy_headers={'ProxyFlag': '1'})
    await sess.get(url)
    assert 'ProxyFlag' in proxy.request.headers
```

asvetlov:
Very often the session uses two different proxies: one for HTTP and other for HTTPS.
Maybe support a dict makes sense (we can merge proxy auth into proxy url to reduce parameters count).
Not sure about support for global proxy headers.

@aio-libs/aiohttp-committers please share your opinions.

fy0:
Very often? I don't think so. Use dict means no code completion and limited ability, so I prefer to a class.

Or just support change proxy during the session, then add another api to set `ClientSession._proxy_info`, it's enough in most situations.

asvetlov:
The mine idea is: specify proxies and do **any** requests over the session.

> Or just support change proxy during the session, then add another api to set ClientSession._proxy_info, it's enough in most situations.

Evolving your idea I can say that you don't need to pass proxy info into constructor at all: just provide corresponding proxy parameters for every HTTP request.

fy0:
> Evolving your idea I can say that you don't need to pass proxy info into constructor at all: just provide corresponding proxy parameters for every HTTP request.

Actually I wrote some code like that:
```python
def new_session(connector=None, cookie_jar=None, cookies=None, headers=None):
    # set cookies, headers, timeout, etc.
    cookie_jar = cookie_jar if cookie_jar else CookieJar(unsafe=True)
    return ClientSession(conn_timeout=5, read_timeout=10, connector=connector, cookie_jar=cookie_jar, cookies=cookies, headers=headers)

# `new_session` used by hundreds of files
async def fetch():
    async with new_session(headers=get_headers()) as session:
        async with session.post('http://xxxx', data=data) as resp:
            text = await resp.text()
```

Obviously modify global settings is much better than rewrite hundreds of local settings.
In this case I don't need change proxy, but the PR is for all users so I want to make it more flexible. 

I thought keep simple and flexible is very important: We just write general code for general case, it can easily works. If we need more, after add options or do something, it still works for unusual cases.

My suggestion:
```python
sess = ClientSession(proxy='http/https/socks://...')
sess.set_proxy('http/https/socks://...')

proxy: [str, ProxyProvider]


class ProxyProvider:  # Provider/Feeder/Switcher ...
    def pick(self, sess, url, method):
        return { headers: {}, url: {}, auth: {} }

class OSEnvProxyProvider(ProxyProvider):
    def __init__(self):
        self.http = read_from_env('http_proxy')
        self.https = read_from_env('https_proxy')

    def pick(...):
        ret = url_parse ...
        if ret.schema == 'http': return self.http
        if ret.schema == 'https': return self.https

class SimpleProxyProvider(ProxyProvider):
    def __init__(url, auth, headers):
        self.default = {...}

    def pick(...):
        return self.default
```

fy0:
already 10 days

asvetlov:
My main objection is that session constructor is complicated already, with PR it becomes even more complicated.
We need opinions from other @aio-libs/aiohttp-committers 
Strong -1 for `session.set_proxy()`: we have no methods for changing session state, I don't want to start adding them.

kxepal:
I agree with @asvetlov . Personally, I see proxy configuration as a sort of middleware for sessions. Could it be implemented in such fashion? Something like:
```
session = ClientSession(...)
session = ProxySession(session, auth=..., headers=...)
...
```
A bit more code to write, but proxy thing becomes completely decoupled from session one. Thoughts?

cecton:
I liked the first idea:

```python
async with ClientSession(proxy=proxy) as session:
    async with session.get('http://test.com') as resp:
        print(await resp.text())
```

Simple to use, optional to use... Sounds cool. But then @asvetlov made a good point about having multiple proxies.

About the alternatives:

```python
session.set_proxy(....)
```

Yes I agree with @asvetlov that this will introduce a state to the application. Imagine you run 2 HTTP calls using 2 different proxies in parallel (asynchronously), you don't know what's going to happen.

```python
session = ClientSession(...)
session = ProxySession(session, auth=..., headers=...)
```
It looks awful to use, sorry @kxepal 😁 

Now why not having a proxy object that actually has all the information to connect using any scheme and the headers to use, etc...? So you merge `proxy`, `proxy_auth` and `proxy_headers` to a single object?

You can still write this:

```python
async with ClientSession(proxy=proxy) as session:
    async with session.get('http://test.com') as resp:
        print(await resp.text())
```

But you need to create first your proxy object:

```python
proxy = Proxy(http="some_url", http_headers={...}, http_auth=("user", "pass"),
              https="other_url", ...)

# or

proxy = Proxy(all="some_url", all_headers={...}, all_auth=("user", "pass"))
```

fafhrd91:
I like @kxepal idea. but in slightly different way

```python

proxy = Proxy(...)

async def main():
     with proxy.with(ClientSession()).get(url='') as resp:
         ...

```

lysyloren:
I think @cecton idea about `Proxy` class, which encapsulates all proxy parameters and alghorithms is a good idea.

Actual implementation of proxy support is not so flexible, some issues are:
-  it's hard to fully control headers sent to proxy even with `proxy_headers` setting, e.g. CONNECT:443 send headers: `Accept: */*` and `Accept-Encoding: gzip, deflate`, which are inappropriate in this case. These headers cannot be skipped, because lack of  `proxy_skip_auto_headers`,
- some proxy types may need even more `proxy_` settings, e.g. see discussion  #2630,
- implementation CONNECT:80 proxy type by end user requires inheritance from `TCPConnector`, copying whole `_create_proxy_connection` method and change one or two parameters in its body (and then end user must track all changes in this method in future `aiohttp` releases),
- implementation CONNECT:80 proxy type by `aiohttp` requires passing around more `proxy_` parameters.

With separate `Proxy` class it will be easier address these issues. `Proxy` class can have whole bunch of  `proxy_` parameters without complicating `ClientSession()` or `session.request()` API. Also, one can easily inherit `Proxy` object and modify its behavior for his own needs.


Using `Proxy` class in simple case can be as easy as `URL` from `yarl`. `proxy` parameter can accept `str` or `Proxy` instance:
```
proxy = Proxy(url)
proxy = Proxy(url, http_headers={...}, http_auth=("user", "pass"), ...)
session.get('http://test.com', proxy=proxy)

# or in the simplest case:
session.get('http://test.com', proxy='http://my.proxy.com')
```

`proxy` parameter can be supported by `ClientSession.__init__` and by `ClientSession._request`. It is possible then to override proxy for specific request with other proxy, e.g:
```
async with ClientSession(proxy=proxy1) as session:
    async with session.get(url1) as resp: # proxy1 is used here
        ...
    async with session.get(url2, proxy=proxy2) as resp:
        ...
    async with session.get(url3, proxy=proxy3) as resp:
        ...
```

`Proxy` instance will by passed down to `ClientRequest` instance and then proper methods will be called from `Connector` instance (e.g. `req._proxy.create_connection()`).

I can imagine even `ProxySet` class which holds set of `Proxy` instances. End user can then inherit `ProxySet` and implement his own strategy for using proxies (e.g. different proxy for different protocols or urls, or rotating proxy strategy, etc.). `ProxySet` instance can be provided in `proxy` parameter also for API simplicity.

In this design more advanced proxy use can be easily implemented like e.g. `ProxyChain`.

RatanShreshtha:
It seems Travis has forgotten about this PR, I can send a new after resolving the merge conflict.

webknjaz:
@RatanShreshtha Travis CI didn't forget about it. It's just that it has `continuous-integration/travis-ci/pr` status reported from the old integration, but `Travis CI - Pull Request` is for the new one, which we've enabled in the end of the spring. There were now updates to the PR since then so the new integration haven't received any events wrt to this PR and therefore didn't create any builds for it.
We can close/reopen it to trigger the CI build but as you said it should be updated anyway.

fy0:
@RatanShreshtha  

It's really a long time. Since the pr was sent, I was busied with works for some days. It's about two months later when I was free, and I nearly forgot it.

I thought it's must be someone solved the problem in these two months, so my PR was already finished. One year past, there still no proxy api for ClientSession, really?

I must say origin PR is too simple, api about proxies must be redesigned.

RatanShreshtha:
@fy0 are you working on new PR ?

fy0:
@RatanShreshtha No, not now. I suggest that you could use a temporary local patch if you are in a hurry. Even if someone works on it today, merged can not be so quickly.

RatanShreshtha:
@fy0 Should I send a new PR with your changes after resolving the merge conflicts?

fy0:
@RatanShreshtha OK, of course.

asvetlov:
Overtaken #3405

Issue #2581, opened 2017-12-01
Support .netrc by trust_env


asvetlov:
`.netrc` has information about logins and password.
Applying it to client request when `trust_env=True` would be awesome.

Volunteers are welcome!

asvetlov:
Fixed by #2584

PR #2584, opened 2017-12-03
Support .netrc by trust_env

linw1995:
## What do these changes do?

Support .netrc by trust_env

None

## Related issue number

#2581 

## Checklist


aiohttp/helpers.py
```diff
@@ -115,6 +116,25 @@ def strip_auth_from_url(url):
 
 
 def proxies_from_env():
+    netrc_obj = None
+    netrc_path = os.environ.get('netrc')
+    if netrc_path is None:
+        home_dir = os.path.expanduser('~')
```
asvetlov: Could you use `pathlib` instead of `os.path`?
E.g. https://docs.python.org/3/library/pathlib.html#pathlib.Path.home for getting user home folder.

aiohttp/helpers.py
```diff
@@ -115,6 +116,25 @@ def strip_auth_from_url(url):
 
 
 def proxies_from_env():
+    netrc_obj = None
+    netrc_path = os.environ.get('netrc')
```
asvetlov: Environment names are case sensitive, according to https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html the name should be upper-cased: `NETRC`.

asvetlov:
`.netrc` could be used not only for proxy credentials but for any resource, `requests` supports it.
What's about reading the file in `ClientRequest.update_auth()` https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L281 if no explicit auth info is provided?

tests/test_proxy_functional.py
```diff
@@ -511,6 +512,28 @@ def _make_ssl_transport_dummy(self, rawsock, protocol, sslcontext,
     assert proxy.request.headers['Proxy-Authorization'] == auth.encode()
 
 
+async def test_proxy_from_env_http_with_auth_from_netrc(
+        proxy_test_server, get_request, tmpdir, mocker):
+    url = 'http://aiohttp.io/path'
+    proxy = await proxy_test_server()
+    auth = aiohttp.BasicAuth('user', 'pass')
+    netrc_file = tempfile.mktemp()
```
asvetlov: Why you use `tempfile` module? Is `tmpdir` fixture functionality not enough?

linw1995:
I'm so sorry, take so long to changes code. not familiar to pathlib, I should take more time to think then code.

asvetlov:
No problem

linw1995:
if I use `netrc_from_env` into `ClientRequest.update_auth()` https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L281, the `ClientRequest` needs to add a param `trust_env`. How about add it into `ClientSession` if no explict auth or AUTHORIZATION in headers? https://github.com/aio-libs/aiohttp/blob/dbb8d9a63f6ffba0ae7e4592158045e54c26433a/aiohttp/client.py#L252-L265

asvetlov:
Sorry, had no time to look on changes.
Will review tomorrow or on Tuesday.

aiohttp/helpers.py
```diff
@@ -111,12 +112,40 @@ def strip_auth_from_url(url):
...
+            home_dir = Path.home()
+            if os.name == 'nt':  # pragma: no cover
+                netrc_path = home_dir.joinpath('_netrc')
+            else:
+                netrc_path = home_dir.joinpath('.netrc')
+
+        if netrc_path and netrc_path.is_file():
+            try:
+                netrc_obj = netrc.netrc(str(netrc_path))
+            except (netrc.NetrcParseError, IOError) as e:
```
asvetlov: Rename `IOError` to `OSError` -- they are aliases but `OSError` is the preferable name.

aiohttp/helpers.py
```diff
@@ -125,6 +154,12 @@ def proxies_from_env():
             client_logger.warning(
                 "HTTPS proxies %s are not supported, ignoring", proxy)
             continue
+        if netrc_obj and auth is None:
+            auth_from_netrc = netrc_obj.authenticators(proxy.host)
+            if auth_from_netrc is not None:
+                *logins, password = auth_from_netrc
+                auth = BasicAuth(logins[0] if logins[0] else logins[1],
```
asvetlov: `logins[-1]` should be enough.
Please add a comment when `logins[0]` can be empty (file is started with tab I guess).

linw1995: https://github.com/requests/requests/blob/24092b11d74af0a766d9cc616622f38adb0044b9/requests/utils.py#L203-L207
`Requests` code shows me how to handle it, and i saw the `netrc` source code.
`netrc.authenticators` method return a (`user`, `account`, `password`) tuple.
I guess `user` and `account` both can be username. 

```python
        if netrc_obj and auth is None:
            auth_from_netrc = netrc_obj.authenticators(proxy.host)
            if auth_from_netrc is not None:
                # auth_from_netrc is a (`user`, `account`, `password`) tuple,
                # `user` and `account` both can be username,
                # if `user` is None, use `account`
                *logins, password = auth_from_netrc
                auth = BasicAuth(logins[0] if logins[0] else logins[-1],
                                 password)
```

linw1995:
[Create a .netrc or _netrc file](https://www.labkey.org/Documentation/wiki-page.view?name=netrc) says the netrc file should be set to at least Read (400), or Read/Write (600), unreadable to everyone else. 
Is it necessity to verify that?

asvetlov:
No, `.netrc` permission rules are for administration and security, no reason to raise an exception.
If user's host is not secure -- it is not aiohttp problem.

aiohttp/helpers.py
asvetlov: Ok

asvetlov:
Thanks!

PR #2585, opened 2017-12-03
Don't create a new resource with allow_head=True

asvetlov:
Backward incompatible but I believe almost nobody is affected.

### DIFF
--- a/CHANGES/2585.removal
+++ b/CHANGES/2585.removal
@@ -1,0 +1,1 @@
+Do not create a new resource on `router.add_get(..., allow_head=True)`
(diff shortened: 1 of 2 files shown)


CHANGES/2585.removal
```diff
@@ -0,0 +1 @@
+Do not create a new resource on `router.add_get(..., allow_head=True)`
```
kxepal, line 1: I guess it's improvement, not removal.

asvetlov: It's both optimization and breaking change: in aiohttp 2.x the call `router.add_get('/', handler, name='name')` adds two named resources: `name` for GET and `name-head` for `HEAD`.
But both resources are identical.

After PR merging I want add generic optimization for sequential `add_route()` calls if created resource is the same.
```
router.add_get('/', ...)
router.add_post('/', ...)
```
should create the single resource if *path* and *name* are the same.

Issue #2277, opened 2017-09-18
Missing response content if header "Upgrade: h2,h2c" and "Connection: Upgrade" are available


segal90:
## Long story short

Content of a POST response is None or b'' (based on how do you want access the response content) if "Upgrade: h2,h2c" and "Connection: Upgrade" headers are in the HTTP response.
After disable Apache's mod_http2 everything works like a charm.

## Expected behaviour

Ignore these headers or handle correctly the HTTP2 upgrade request but not ignore the response body.

## Actual behaviour

Response body is ignored

## Steps to reproduce

1) configure an Apache with mod_http2
2) run this code snippet:
```python
auth = aiohttp.BasicAuth('user1', 'VerySecretPassword')
data = {'just': 'some', 'dummy': 'data'}
async with aiohttp.ClientSession(auth=auth, conn_timeout=10) as session:
    async with session.post(<SERVER_ADDRESS>, data=data) as response:
        print(await response.json()) 
```

## Your environment

OS: Linux (Fedora)
aiohttp version: 2.2.5
uvloop is set as asyncio event loop

asvetlov:
aiohttp doesn't support HTTP/2 yet.

Please describe what response do you request for the case.

segal90:
Lack of HTTP/2 is not the problem. Problem is that the HTTP client is not ignore the request for HTTP2 from the webserver.
I just want access the HTTP response body which is now missing.

mutaboricha:
```Python
async def f():
    stream = None
    async with ClientSession(connector=TCPConnector(verify_ssl=False)) as session:
        async with session.get(url[1]) as resp:
            print(resp.raw_headers)
            if resp.status == 200:
                stream = io.BytesIO()
                async for chunk in resp.content.iter_chunked(1024):
                    stream.write(chunk)

    if stream:
        mime = magic.from_buffer(stream.getvalue(), mime=True)
        print(mime)
        print(stream.getvalue())


loop = asyncio.get_event_loop()
loop.run_until_complete(f())
loop.close()
```
Response headers:
(b'Date', b'Mon, 25 Sep 2017 14:52:15 GMT'), 
(b'Server', b'Apache/2.4.27 (cPanel) OpenSSL/1.0.2k mod_bwlimited/1.4'), 
**_(b'Upgrade', b'h2,h2c'), (b'Connection', b'Upgrade')_**, 
(b'Last-Modified', b'Fri, 26 Jul 2013 05:35:15 GMT'), 
[...]

PR #2588, opened 2017-12-04
Limit protocol upgrades to supported protocols

lphuberdeau:
## What do these changes do?

Fix issue #2277 by limiting connection upgrades to supported protocols. Currently only websocket as h2 and h2c are not supported.

There are no unit tests on this change as all my attempts caused the test to be skipped. Could use some help.

Responses will be obtained when the server advertises connection upgrades to http2 without the client requesting it.

## Related issue number

#2277 

## Checklist

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -327,6 +327,12 @@ class HttpParser:
             elif v == 'upgrade':
                 upgrade = True
 
+        # restrict connection upgrade to supported protocols
+        upgrade_to = headers.get(hdrs.UPGRADE)
+        if upgrade_to:
+            v = conn.lower()
+            upgrade = upgrade and v in {'websocket'}
+
         # encoding
         enc = headers.get(hdrs.CONTENT_ENCODING)
         if enc:


asvetlov:
No, aiohttp supports upgrade to custom protocol IMHO.
@fafhrd91 please correct if I'm wrong.
@lphuberdeau what is your use case?

fafhrd91:
I see one main reason is http/2, if server can not upgrade to http/2 it should ignore upgrade header and serve request as a http/1 request.
But for this use case custom logic should be implemented

fafhrd91:
Though I have plans to add http/2 support to aiohttp

lphuberdeau:
@asvetlov I have the same issue as reported. aiohttp provides blank responses on some sites and it's because of this apparently. #2277 

It seems to be related to mod_h2 in apache where the server always responds with Upgrade: h2 even if the client did not request it. Browsers and all other tools we've tested give the response just fine. aiohttp hangs for a bit and errors out based on stream not being readable.

Commenting out this line fixes the issue (although it's not a real fix)
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_proto.py#L173

This condition is also part of the problem, although I'm not sure I understand it. 
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_parser.py#L166

I'm fully aware the proposed fix is missing out on a lot of context ;) Glad to have this conversation.

lphuberdeau:
@fafhrd91 The issue here is really that the server includes the Upgrade: h2 header in the response even if the client does not request it. We actually found several cases of this, all of which were under Apache so far.

Chunked response also, which kind of points towards that http_parser.py#L166 line.

HTTP2 support would be great!

fafhrd91:
“upgrade: h2” is standard http/2 upgrade process for clear connection upgrade. Spec states that server should respond with http/1 if it does not understand http/2. So this is actually aiohttp bug

lphuberdeau:
The patch could easily be reversed to use a black list on connection upgrade to h2 instead and still resolve the issue if the whitelist is too restrictive. It could possibly also be resolved higher in the stack.

Let me know how I can help.

asvetlov:
Fixed on master.
Please feel free to create a new issue if problem will back again.

lphuberdeau:
Thanks for taking care of this. For educational purposes, could you identify which commit fixes the issue?

It would also help track which release the fix will be in.

asvetlov:
Sorry, I've missed concrete commit hash number.
Just this script https://gist.github.com/Findus23/3aaedbac0dd2520a9b65d0c9a6431b22 works on master but fails on aiohttp 2.3

The master will be published as aiohttp 3.0

yelite:
@lphuberdeau I think it's 79655ebce1d61f6b425c67a7f8cea3e27eee9f0a. You can also check the commit in upstream, https://github.com/nodejs/http-parser/commit/05525c5fde1fc562481f6ae08fa7056185325daf

Issue #2206, opened 2017-08-15
json request for certain media types fail


iherman:
## Long story short

The `resp.json()` call reports a bug if the return type is something like `application/XXX+json` (as opposed to `application/json`). However, `application/XXX+json` is a perfectly fine media type, "extending" the _semantics_ (not the syntax) of the core json syntax, ie, should be parsed as bona fide json...

## Your environment

OS X 10.11.6, Python 3.6.2

jscholes:
You can pass a custom content_type parameter to the ClientResponse.json method, although I agree that this isn't perhaps the most sensible API.  I just switched some code from using Requests to using this library and while Requests could decode the JSON without a problem, aiohttp reported an error which will trip some people up.

asvetlov:
Related to #2174

asvetlov:
Extending the list of accepted json content types is fair easy change.
Please propose a Pull Request. `application/XXX+json` is not acceptable but I pretty sure you could grab a list of wide spread json types from `requests` library.

asvetlov:
use `content_type=None` for disabling the check

asvetlov:
What to do?
On one hand we want to prevent processing non-JSON data in `.json()` method.
On other we don't want to be too strict.

Maybe you can pass a custom content type when fetching a data from your server?
Or create a custom json reading function on top of `resp.read()`?

iherman:
@asvetlov,

The fundamental fact is that data coming with `XXX+json` **is** JSON. From the point of view of aiohttp that all that matters. A properly registered `XXX+json` media type means that the content _is_ JSON but it may have some additional syntactic or structural constraint _within the framework of JSON_. This may mean imposing a particular JSON Schema, or some more complicated specification on top of what JSON means, or may be no extra syntactic restriction whatsoever. 

To take a specific example. `application/ld+json` is a bona fide media type defined through the structural syntax; the content is defined via the [JSON-LD Recommendation](https://www.w3.org/TR/json-ld/). That specification adds some additional semantics to individual terms, requires some specific structure to the JSON data, but the essential point is that a JSON-LD content is a perfectly fine JSON content. (This is probably one of the most complex examples. Others may be a few paragraphs only.)

[...]

PR #2594, opened 2017-12-06
handle custom content types

wuan:
## What do these changes do?

When using a custom content type like `application/vnd.aiohttp-test+json` the client now ignores content type parameters (e. g. an encoding setting) when checking for matching types.

Nothing expected here. Clients which include the parameters in `content_type` (as a workaround) should still work.

## Related issue number

#2206 is somehow related

## Checklist

### DIFF
--- a/CHANGES/2594.bugfix
+++ b/CHANGES/2594.bugfix
@@ -1,0 +1,1 @@
+Fixes content type check in client by ignoring content type parameters like encoding etc.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -25,6 +25,7 @@ Alexey Popravka
 Alexey Stepanov
 Amin Etesamian
 Amy Boyle
+Andreas Würl
 Andrei Ursulenko
 Andrej Antonov
 Andrew Leech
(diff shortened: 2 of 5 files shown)


asvetlov:
Sorry, I don't follow.
What problem do you want to solve?
What is `Content-Type` header for your server?

wuan:
We are reading data via HTTP and get a response which has a `Content-Type` header in the form `application/vnd.foo+json; encoding=utf-8`.

To read JSON data from the response, the full content type has to be specified:
```
    data = response.json(content_type='application/vnd.foo+json; encoding=utf-8')
```
As `json()` already handles the encoding option according to the servers `Content-Type` header internally, it feels wrong to specify (and even fix) the encoding through the content type string. If the server leaves out that information some day it will even break the client.

This PR will allow us to use
```
    data = response.json(content_type='application/vnd.foo+json')
```
and still get a correct response with any of these `Content-Type` headers: `application/vnd.foo+json; encoding=utf-8`, `application/vnd.foo+json; encoding=cp1251`, `application/vnd.foo+json`

asvetlov:
In `requests` `.json()` implementation is pretty simple: it doesn't accept custom content type and encoding.

I have a strong temptation to drop these params from aiohttp client too.

wuan:
I would be fine with dropping the content type verification.

asvetlov:
Let's close the PR as `wontfix`. 3bbe4d76 stresses a way for content type disabling

Issue #2586, opened 2017-12-04
Reuse last resource from url dispatch table


asvetlov:
If name and path is the same -- add a route to existing resource instead of creating a new one.

E.g.
```
router.add_get('/', ...)
router.add_post('/', ...)
```
should create the single resource if path and name are the same.

It reduces resources table size and boosts url lookup if handlers for the same url is grouped together.

The change is easy, Pull Request is welcome.

sheb:
I'm working on it

sheb:
Hmm it's not allowed to add a route with existing path and name as it's already handled by 

> aiohttp.web_urldispatcher.UrlDispatcher().register_resource(..)

See below
```
app.router.add_get("/", lambda x: x, name="foo")
app.router.add_post("/", lambda x: x, name="foo")
>>> ValueError: Duplicate 'foo', already handled by <PlainResource 'foo'  /
```

Should I match only the path ?

asvetlov:
The feature should check both `path` and `name` for last added resource.
If they are matched to added one -- return last item.
All functionality could be done in `.add_resource()` method.

sheb:
Ok got it, pull request soon

asvetlov:
Fixed by #2603

PR #2603, opened 2017-12-12
Avoid to create unnecessary resources (#2586)

sheb:
## What do these changes do?

Do not create a new resource when adding a route with the same
name and path of an existing resource

Nothing expected

## Related issue number

#2586 

## Checklist

- [x] Documentation reflects the changes - No documentation needed here

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -439,7 +439,7 @@ class PrefixResource(AbstractResource):
         self._prefix = prefix + self._prefix
 
     def raw_match(self, prefix):
-        return self._prefix == prefix
+        return False
 
     # TODO: impl missing abstract methods
 
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -92,6 +92,10 @@ class AbstractResource(Sized, Iterable):
     def freeze(self):
         pass
 
+    @abc.abstractmethod
+    def raw_match(self, path):
+        """Perform a raw match against path"""
+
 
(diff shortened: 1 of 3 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -830,6 +840,10 @@ def register_resource(self, resource):
     def add_resource(self, path, *, name=None):
         if path and not path.startswith('/'):
             raise ValueError("path should be started with / or be empty")
+        # avoid to create unnecessary resources
+        resource = self._named_resources.get(name)
```
asvetlov: The only last resource in table should be checked -- otherwise the PR changes route order implicitly.

aiohttp/web_urldispatcher.py
```diff
@@ -299,6 +299,10 @@ def register_route(self, route):
         else:
             return None, allowed_methods
 
+    @abc.abstractmethod
+    def raw_match(self, path):
```
asvetlov: Move the method to `AbstractResource` please

aiohttp/web_urldispatcher.py
```diff
@@ -428,6 +438,9 @@ def add_prefix(self, prefix):
         assert len(prefix) > 1
         self._prefix = prefix + self._prefix
 
+    def raw_match(self, prefix):
+        return self._prefix == prefix
```
asvetlov: Reusing `StaticResource` and subapp resource is dangerous.
Let's always return False.

asvetlov:
Thanks!

PR #2605, opened 2017-12-12
Add response hook

dadocsis:
None

## Related issue number
#434

## Checklist

- [x ] Unit tests for the changes exist
- [ x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
- [ x] Add a new news fragment into the `CHANGES` folder


asvetlov:
I dislike the proposal by 2 reasons:
1. Named hooks are too subtle. I understand, `requests` uses this approach but if we need a hook for calling a code just after retrieving response header -- let's add just it without inviting very generic and potentially error-prone approach.
2. The PR calls auth helper passing `ClientRequest` instance. `ClientRequest` is not designed to be a part of public API, it is not stable and changed very often. I don't like the class design but can live with it unless we don't expose the class.

hellysmile:
It looks like solution for implementing https://github.com/wikibusiness/aioelasticsearch/issues/48 , requests has https://pypi.python.org/pypi/requests-aws4auth , right now it is very tricky to implement something like requests-aws4auth for aiohttp

dadocsis:
@asvetlov 
Just want to clarify
#1. So just create a single static hook?
#2. I see what you are saying about it not being apart of the api. Unfortunately non trivial auth schemes  will need the original request object. Do you have an actionable suggestion for me?

asvetlov:
Please give me a pause.
I'll try to cook something.

I understand the need and share the intention, but maybe I can invent an alternative implementation.

Cellebyte:
Will this be merged? I need this to implement some custom auth handlers for OpenID token-exchange client Authorization and Authentication in my async client implementation.

asvetlov:
I understand your needs but the functionality should be implemented in a slightly different way than proposed by this PR.

Cellebyte:
How do I implement it now? Should I use decorators?

asvetlov:
Sorry about that.
I really want to improve things, it is on my todo list.
The solution touches several aspects, it is not 20-lines change.
I hope to introduce something in a month or two.

Issue #2620, opened 2017-12-24
DNSCache BUG


socketpair:
(please help me to choose good subject)

It's not easy to explain why this bug happens. I will try:

Here is current code (cutted):

```python
class _DNSCacheTable:
    def add(self, host, addrs):
        self._addrs[host] = addrs
        self._addrs_rr[host] = cycle(addrs)
    def next_addrs(self, host):
        # Return an iterator that will get at maximum as many addrs
        # there are for the specific host starting from the last
        # not itereated addr.
        return islice(self._addrs_rr[host], len(self._addrs[host]))
```  

another place (very simplified):

```python
    @asyncio.coroutine
    def _create_direct_connection(....):
        hosts = dns_cache.next_addrs()
        for hinfo in hosts:
            host = hinfo['host']
            transp, proto = yield from self._wrap_create_connection(   # YIELD FROM IS CRUSHIAL
```
Suppose now, that `_create_direct_connection` is run in parallel. Since `dns_cache.next_addrs()` returns `islice` which points to `cycle` which points to internal state, it's state IS REUSED  in parallel instances of  `_create_direct_connection`. This lead to wrong results returned by `islice`.

And now, let's try to reconstruct situation in test program:

```python
from itertools import cycle, islice
import random
import asyncio

addrs = ['a','b','c','d']
rr_addrs = cycle(addrs)

async def cccc(ident):
    await asyncio.sleep(random.random())
    hosts = islice(rr_addrs, len(addrs))
    for i in hosts:
        print(ident, i)
        await asyncio.sleep(random.random())

async def amain():
    await asyncio.wait([cccc(i) for i in range(5)])

loop = asyncio.get_event_loop()
loop.run_until_complete(amain())
```

and run it:

```
./qwe.py | sort
0 a
0 a
0 b
0 c
1 a
1 b
1 d
1 d
2 a
2 b
2 d
2 d
3 b
3 b
3 c
3 d
4 a
4 c
4 c
4 c
```
As you can see, it's not what author expect! No one instance received a,b,c and d!

How did I come to this bug? I was just doing parallel requests to Dropbox. The resolver gives me IPv6 address as well as IPv4. But my system does not have IPv6 and so `connect()` reports EUNREACH for IPv6 addresses. In my case sometimes `islice` gave only IPv6 addresses! and whole request failed.

How to cure?
simpliest solution is to return `list(islice(....))`

Except for my problem, there is two more:
1. uneven distribution
[...]

PR #2621, opened 2017-12-24
Fix DNSCache race-condition

socketpair:
## Related issue number

#2620

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -578,10 +578,11 @@ class _DNSCacheTable:
         self._timestamps.clear()
 
     def next_addrs(self, host):
-        # Return an iterator that will get at maximum as many addrs
-        # there are for the specific host starting from the last
-        # not itereated addr.
-        return islice(self._addrs_rr[host], len(self._addrs[host]))
+        loop = self._addrs_rr[host]
+        addrs = list(islice(loop, len(self._addrs[host])))
+        # Consume one more element to shift internal state of `cycle`
+        next(loop)
+        return addrs
 
     def expired(self, host):
         if self._ttl is None:
--- a/CHANGES/2620.bugfix
+++ b/CHANGES/2620.bugfix
@@ -1,0 +1,1 @@
+Fixed race-condition for iterating addresses from the DNSCache.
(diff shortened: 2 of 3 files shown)


aiohttp/connector.py
```diff
@@ -560,28 +559,24 @@ def addrs(self):
...
         self._timestamps.clear()
 
     def next_addrs(self, host):
-        # Return an iterator that will get at maximum as many addrs
-        # there are for the specific host starting from the last
-        # not itereated addr.
-        return islice(self._addrs_rr[host], len(self._addrs[host]))
+        addrs = self._addrs[host].copy()
+        shuffle(addrs)
+        return addrs
```
pfreixes: Why not keep the original round robin but making sure that we get rid of the corner case ? such as

```
>>> addrs = cycle([1, 2, 3, 4])
>>> list(islice(addrs, 4))
[1, 2, 3, 4]
>>> next(addrs)
1
>>> list(islice(addrs, 4))
[2, 3, 4, 1]
```
Following your idea of using an ad-hoc list per call to `next_addrs` it's just a matter of make a next after built the list to make sure that we always use a shifted list per call.

socketpair: Because, anyway, cycle will start from actually random element. Why not to randomize explicitly ? Why we should stick on round-robin ?

```
In [7]: timeit list(islice(xxx, 4))
The slowest run took 15.44 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 447 ns per loop

In [8]: timeit shuffle(qwe)
The slowest run took 5.42 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 2.82 µs per loop
```

Well, it seems you are right....

socketpair: But no! if everyone will consume 4 items from slice, everytime everyone will receive exactly the same items, since every call to `nex_addrs` will rotate whole array.

socketpair: So, I fixed.

pfreixes:
So, ready to be merged? flake8 is failing for some reason. Up to yo to merge it once it is fixed.

asvetlov:
Let me review soon.
P.S. Travis is red still :(

asvetlov:
Looks good

PR #2622, opened 2017-12-25
Fix location of DummyCookieJar in docs

lysyloren:
Trivial doc fix. DummyCookieJar can't be imported from aiohttp.helpers.


asvetlov:
Thanks!

PR #2623, opened 2017-12-25
Simplify HTTP pipelining implementation

pfreixes:
## What do these changes do?

HTTP pipelining is still supported but implemented
as a simple FIFO queue with just 1 Asyncio task attending
it. It might allow optimizing infrastructures based on microservices
that do not use HTTP pipelining.

No

## Related issue number

#2109


pfreixes:
@fafhrd91 my idea was working on getting rid of some of the layers in a separated PR once it gets merged, just for simplicity and make easier the reviewing process. But, if you agree I can try to do it in the same PR.

/cc @asvetlov

fafhrd91:
it is up to you, I'd remove StreamWriter at the same time

asvetlov:
I prefer two steps: the PR is huge already.

asvetlov:
Thanks, @pfreixes

PR #2625, opened 2017-12-26
avoid code duplicity

jairhenrique:


asvetlov:
I doubt if the PR should be applied:
1. It saves not too much code.
2. I'd like to deprecate and eventually remove `fingerprint` parameter from TCPConnector.

asvetlov:
Superseded by #2628

PR #2629, opened 2017-12-28
Skip middleware code when there are no user middlewares installed

pfreixes:
## What do these changes do?

Fixes performance issue introduced by #2577, boosting from 8K req/sec
to almost 9K req/sec.

If there are no middlewares installed by the user the attribute
`request._match_info.current_app` already looks at to the right app,
this is by design.

No

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -110,15 +110,18 @@ class Application(MutableMapping):
             return
 
         self._frozen = True
-        self._middlewares = tuple(self._prepare_middleware())
+        self._middlewares.freeze()
         self._router.freeze()
         self._on_response_prepare.freeze()
         self._on_startup.freeze()
         self._on_shutdown.freeze()
         self._on_cleanup.freeze()
+        self._run_middlewares = len(self.middlewares) > 0
+        self._middlewares_handlers = tuple(self._prepare_middleware())
 
         for subapp in self._subapps:
             subapp.freeze()
+            self._run_middlewares |= len(subapp._middlewares) > 0
 
     @property
     def debug(self):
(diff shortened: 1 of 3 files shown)


pfreixes:
Test numbers

```
$ python3 benchmark.py --reference=origin/master --wrk-path=/home/ec2-user/wrk/wrk /home/ec2-user/aiohttp
[origin/master] Benchmark req/sec 8206

$ python3 benchmark.py --reference=origin/fix_degradation_performance_middleware --wrk-path=/home/ec2-user/wrk/wrk /home/ec2-user/aiohttp-pfreixes
[origin/fix_degradation_performance_middleware] Benchmark req/sec 9049
```

I've used this benchmark [1]

[1] https://github.com/pfreixes/aiohttp_regression_performance_test

aiohttp/web_app.py
```diff
@@ -110,15 +110,18 @@ def freeze(self):
...
         self._on_response_prepare.freeze()
         self._on_startup.freeze()
         self._on_shutdown.freeze()
         self._on_cleanup.freeze()
+        self._run_middlewares = len(self.middlewares) > 0
+        self._middlewares_handlers = tuple(self._prepare_middleware())
 
         for subapp in self._subapps:
             subapp.freeze()
+            self._run_middlewares |= len(subapp._middlewares) > 0
```
asvetlov: The line is cryptic a little.
Please replace with `self._run_middlewares = self._run_middlewares or subapp._run_middlewares`
The change respects middlewares from deep nested subapps: app -> subapp1 -> subapp2.

pfreixes: Fair enough, done. Added also a bit of internal documentation about this flag and the root cause of its usage.

Issue #2635, opened 2017-12-31
Unclosed client session Error in aiohttp.request


soulomoon:
## Long story short

using aiohttp.request from the example in [document](https://docs.aiohttp.org/en/stable/client_reference.html#basic-api) emit (Unclosed client session) error

## Steps to reproduce

```python
import aiohttp
import asyncio

async def fetch():
    async with aiohttp.request('GET', 'http://python.org/') as resp:
        assert resp.status == 200
        print(await resp.text())

if __name__=="__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(fetch())
```

## Your environment
python 3.6.3
Darwin
aiohttp==2.3.7

asvetlov:
This is well-known behavior, fixed in aiohttp 3.0.
Cannot be backported -- the fix is not 100% backward compatible.

P.S.
Why do you recreate a session every time? If you need slowdown your program -- `sleep()` should be enough.

soulomoon:
@asvetlov Thank you for the reminder , now I use ClientSession in my code now, but 
Probably should state that in the document too? It would be really confusing for those who are trying out examples from the document.

asvetlov:
Is note from https://docs.aiohttp.org/en/stable/client.html#make-a-request not enough?
If it is -- please feel free to make Pull Request with documentation update.

PR #2636, opened 2017-12-31
Add note about `aiohttp.request` usage problem

soulomoon:
## What do these changes do?

Add note about `aiohttp.request`'s example which would cause `Unclosed client session`

## Related issue number

[#2635](https://github.com/aio-libs/aiohttp/issues/2635)


asvetlov:
Documentation is a not good place for reflecting every aiohttp bug, especially already fixed in master

PR #2640, opened 2018-01-03
Raise cancelled on closed

### LINKED ISSUE #2499: Raise an exception if trying to write into closed response
See https://stackoverflow.com/questions/47175297/handling-premature-client-disconnection-in-aiohttp

The problem is: `await drain()` does nothing because the buffer is not overflown, the buffer is not overflown because sent data are ignored and not accumulated in internal buffer.
`transport.is_closing()` should be used for check.

It should be fixed in payload writer, affects both client and server.

maqquettex:
## What do these changes do?

Add raising of asyncio.CancelledError to PayloadWriter write() method to avoid writing to closing transport.

According to https://github.com/aio-libs/aiohttp/issues/2499#issuecomment-352364059 is was possible to catch this situation earlier by manually checking transport status of request. Thus, change should not broke any previously written code.

## Related issue number

Resolves #2499 

## Checklist

- [*] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/CHANGES/2499.bugfix
+++ b/CHANGES/2499.bugfix
@@ -1,1 +1,1 @@
-Fix writing to closed transport by raising asyncio.CancelledError
+Fix writing to closed transport by raising `asyncio.CancelledError`
--- a/CHANGES/2499.bugfix
+++ b/CHANGES/2499.bugfix
@@ -1,0 +1,1 @@
+Fix writing to closed transport by raising asyncio.CancelledError
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -1,5 +1,6 @@
 """Http related parsers and protocol."""
 
+import asyncio
 import collections
 import socket
 import zlib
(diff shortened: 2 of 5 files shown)


pfreixes:
LGTM changes approved, thanks!

pfreixes:
but please first I would ask you to synchronize your branch with the master one.

azhpushkin:
@pfreixes Done! (i guess merge is enough for further rebase&squash? )

pfreixes:
thanks!

asvetlov:
Hold on!
Is CancelledError raised by client API?
CancelledError without actual cancellation on `resp.write()` is confusing also.
I suggested raising `RuntimeError` or maybe even better a custom exception type derived from RuntimeError.

azhpushkin:
@asvetlov 
Agree, make sence. But why RuntimeError and not ConnectionError, for example? It could be a good match.

pfreixes:
Taking into account that write_eof calls the drain and it can have as a side effect a cancellation error if the client disconnects, don't we have to try to handle both cases in such a way that the developer can catch easily them using just one exception?

pfreixes:
Yeps forgot my last comments, drain as other waiters are wake up with the proper exception once the connection gets closed.  Should we use a `ConnectionError` [1] exception or one of its children?

[1] https://docs.python.org/3/library/exceptions.html#ConnectionError

azhpushkin:
asyncio's FlowControlMixin uses ConnectionResetError, so we could use it too, but I have a concern whether transport could be closed only on the other than aiohttp side (otherwise usual ConnectionError looks as a better match to me).

asvetlov:
No way right now.
But you didn't cancel a task -- something canceled it from outside :)

PR #2646, opened 2018-01-09
Minor fix of exception type

maqquettex:
## What do these changes do?

Minor fix discussed in PR #2640 

## Related issue number

Issue #2499

### DIFF
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -1,6 +1,5 @@
 """Http related parsers and protocol."""
 
-import asyncio
 import collections
 import socket
 import zlib
@@ -129,7 +128,7 @@ class PayloadWriter(AbstractPayloadWriter):
         self.output_size += size
 
         if self._transport.is_closing():
-            raise asyncio.CancelledError('Cannot write to closing transport')
+            raise ConnectionError('Cannot write to closing transport')
         self._transport.write(chunk)
 
     def write(self, chunk, *, drain=True, LIMIT=64*1024):
--- a/tests/test_http_writer.py
+++ b/tests/test_http_writer.py
@@ -1,5 +1,4 @@
 """Tests for aiohttp/http_writer.py"""
-import asyncio
 import zlib
 from unittest import mock
 
@@ -157,5 +156,5 @@ def test_write_to_closing_transport(stream, loop):
     msg.write(b'Before closing')
     stream.transport.is_closing.return_value = True
 
-    with pytest.raises(asyncio.CancelledError):
+    with pytest.raises(ConnectionError):
         msg.write(b'After closing')


kxepal:
Hm...will this exception happens if client interrupts response read? For instance, when browser stops awaiting server response.

azhpushkin:
@kxepal  I did testing by the following script:
```python
from aiohttp import web

async def view(request: web.Request) -> web.StreamResponse:
    response = web.StreamResponse(headers={'content-type': 'text/plain'})
    await response.prepare(request)

    for i in range(100000):
        await response.write(b'test response')
    return response

app = web.Application()
app.router.add_get('/', view)
web.run_app(app, host='0.0.0.0', port=8080)
```

After `curl -XGET http://0.0.0.0:8080` and `Ctrl-C` there is expected error in the console
``` python
Error handling request
  File "/home/mqtx/projects/aiohttp/aiohttp/web_protocol.py", line 410, in start
  File "/home/mqtx/projects/aiohttp/aiohttp/web.py", line 326, in _handle
  File "bug.py", line 9, in view
    await response.write(b'test response')
  File "/home/mqtx/projects/aiohttp/aiohttp/web_response.py", line 407, in write
    return self._payload_writer.write(data)
  File "/home/mqtx/projects/aiohttp/aiohttp/http_writer.py", line 228, in write
    self._write(chunk)
  File "/home/mqtx/projects/aiohttp/aiohttp/http_writer.py", line 198, in _write
    raise ConnectionError('Cannot write to closing transport')
ConnectionError: Cannot write to closing transport
```

Also, wrapping `await response.write` in `try: ... except ConnectionError: ...` works.

Worth mentioning, that for some reason this is not working with master branch. On master for me creating and requesting any view fails (curl says `Connection reset by peer`), and after aiohttp server is closed following traceback appears:
```
future: <Task finished coro=<RequestHandler.start() done, defined at /home/mqtx/projects/aiohttp/aiohttp/web_protocol.py:343> exception=AttributeError("'RequestHandler' object has no attribute '_force_close'",)>
  File "/home/mqtx/projects/aiohttp/aiohttp/web_protocol.py", line 357, in start
    while not self._force_close:
AttributeError: 'RequestHandler' object has no attribute '_force_close'
```
but this looks like error related to another issue.
So I did a testing on 2.3 branch with cherry-picked commits and over there everything is fine

asvetlov:
The proper exception type is not obvious.
Say, in client API we are trying to hide almost all standard exception by derived from base `aiohttp.ClientError`. 
It makes a code messy sometimes.
Maybe aiohttp 3.0 is a great opportunity for deep exceptions hierarchy refactoring?

azhpushkin:
For me this looks like very low-level error that is expected to be handled by manual checking of transport status, much like ConnectionResetError in [asyncio.streams.py, FlowControlMixin._drain_helper](https://github.com/python/cpython/blob/ce5b0e9db1b9698e6ffc43ae41cf3a22ca5a6ba6/Lib/asyncio/streams.py#L200-L202). Generally speaking I believe this raise should even be somewhere in the asyncio lib.

Correct me if I'm wrong, this is not part of the client API but both server and client, is it? Did not get too deep into aiohttp's exceptions so not sure whether I can determine best way to go.

asvetlov:
@maqquettex sorry, I've forgot about the PR.
Rebased as #2989 
Will merge in master when green

Issue #2649, opened 2018-01-09
Wrong default `access_log_format` in `web.run_app`


azhpushkin:
## Long story short
web.run_app does not provide correct default `access_log_format`, because of this `None` is being passed into `AccessLogger.compile_format` and causes an error. Because of this application cannot proceed requests correctly.

https://github.com/aio-libs/aiohttp/blob/19c138c864adb37823c4723901a9b84bd2197085/aiohttp/web.py#L39-L43

## Expected behaviour
Having `helpers.AccessLogger.LOG_FORMAT` as default value
## Actual behaviour
`None` as default value for access log format
## Steps to reproduce
```python
from aiohttp import web

async def index(request):
    return web.Response()

app = web.Application()
app.router.add_get('/', index)
web.run_app(app, host='0.0.0.0', port=8080)
```
Querying `0.0.0.0:8080` returns nothing, stopping aiohttp produces following traceback before finish of the loop:
```python
Task exception was never retrieved
future: <Task finished coro=<RequestHandler.start() done, defined at /.../aiohttp/aiohttp/web_protocol.py:343> exception=AttributeError("'RequestHandler' object has no attribute '_force_close'",)>
Traceback (most recent call last):
  File "/.../aiohttp/aiohttp/web_protocol.py", line 357, in start
    while not self._force_close:
AttributeError: 'RequestHandler' object has no attribute '_force_close'
```

azhpushkin:
Closed in #2650

PR #2650, opened 2018-01-10
Fix default `access_log_format` to correct value

maqquettex:
## What do these changes do?

Fixing wrong default value of access log format

Should not affect anything

## Related issue number

#2649 

## Checklist

### DIFF
--- a/CHANGES/2649.bugfix
+++ b/CHANGES/2649.bugfix
@@ -1,0 +1,1 @@
+Fix default value of `access_log_format` argument in `web.run_app`
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -40,8 +40,7 @@ def run_app(app, *, host=None, port=None, path=None, sock=None,
             shutdown_timeout=60.0, ssl_context=None,
             print=print, backlog=128, access_log_class=helpers.AccessLogger,
             access_log_format=helpers.AccessLogger.LOG_FORMAT,
-            access_log=access_logger,
-            handle_signals=True):
+            access_log=access_logger, handle_signals=True):
     """Run an app locally"""
     loop = asyncio.get_event_loop()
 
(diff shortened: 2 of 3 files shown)


fafhrd91:
aiohttp uses `None` for specific reason. enabled server logging significantly reduce performance even if logging it self is not enabled.

azhpushkin:
Problem of this PR is more about having bad parameters, I tried to describe everything in details in the issue. 
Access logger is enabled while default argument of class is set (`access_logger` is not False). I could  set it to None in this PR also, however this is another story so I would like to leave it the way it is .

fafhrd91:
drop performance by 20% is not good change, especially for default parameters.

azhpushkin:
@fafhrd91 Sorry, maybe I`m not following something but with current default params for me views are not working at all. Access logger is being initialized for each request in the following part of code:
https://github.com/aio-libs/aiohttp/blob/19c138c864adb37823c4723901a9b84bd2197085/aiohttp/web_protocol.py#L135-L139

As access logger over here is True `access_log_class(...)` is called. Afterwards, default [AccessLogger.__init__](https://github.com/aio-libs/aiohttp/blob/19c138c864adb37823c4723901a9b84bd2197085/aiohttp/helpers.py#L264) is called, than compile_format() is called and it fails in the following line:
https://github.com/aio-libs/aiohttp/blob/19c138c864adb37823c4723901a9b84bd2197085/aiohttp/helpers.py#L352

This is happening because log_format over here is None. 
Also there is no exception or traceback being produced before loop is closed. Can you please try Steps to reproduce to see whether you getting this problem too?

fafhrd91:
you are right, sorry. somehow I thought that access log is disabled by default.

asvetlov:
Access log was enabled by default in aiohttp 2.x
The issue is a result of my refactoring for `AppRunner`, I forgot to set proper default value.

asvetlov:
Thanks

PR #2651, opened 2018-01-10
Get rid of legacy class StreamWriter #2109

pfreixes:
## What do these changes do?

Legacy StreamWriter class as a pure proxy of the transport and the protocol is
no longer needed. All of the functionalities that were behind this class
has been moved to the `PayloadWriter` class.

Some changes that have to be considered that impacted by this change

* TCP Operations have been isolated in a module rather than move them
into the `PayloadWriter` class. All of these operations have been simplified and state less.

* `WebSocketWriter` class had a dependency with the StreamWriter, to get rid of
that dependency the constructor has been modified to take the protocol
and the transport.

A next step changing the name of the `PayLoadWriter` class for the `StreamWriter` - it reborn!- might be considered, making it consistent with the reader side.

No

## Related issue number

#2109

## Checklist

- [] Documentation reflects the changes


fafhrd91:
looks good to me

aiohttp/http_writer.py
```diff
@@ -110,11 +30,15 @@ def __init__(self, stream, loop):
...
-        self._transport = self._stream.transport
 
-    async def get_transport(self):
+    @property
+    def transport(self):
         return self._transport
 
+    @property
+    def protocol(self):
+        return self._protocol
```
asvetlov, line 40: Please add a test for the property

pfreixes: Done.

asvetlov:
Another question: why do you return a value from `tcp_cork`/`tcp_nodelay` but never use it in aiohttp code?
Let's drop this functionality.
Tests can check a real TCP option by `getsockopt` call instead of checking the return value.

pfreixes:
@asvetlov removed the value returned by the TCP operations. My fault Ive ended up changing the code to have an easy way to test the functions, once these have become stateless there is some edge cases where I cant see how to assert properly. For example the case of the socket `None`. 
Any idea how to improve this [1]?

[1] https://github.com/aio-libs/aiohttp/pull/2651/commits/87928bea98956da46663efcd394cb5e33e245b8d#diff-ef69e535e777102409aee677f44b4137R75

asvetlov:
You can monkey-patch socket object to make sure that its methods are not called but I think current code is good enough.

asvetlov:
Thanks!

PR #2654, opened 2018-01-11
Rename PayloadWriter to StreamWriter

pfreixes:
## What do these changes do?

Rename the `PayloadWriter` class to `StreamWriter`, having consistency with the reader side which has a class called `StreamReader`.

No

## Related issue number

#2109

## Checklist


asvetlov:
Thanks!

Issue #2656, opened 2018-01-11
Adding routes to subapps is broken after attaching subapp to the parent app


webknjaz:
## Long story short

If one adds routes to subapp before `app.add_subapp` accessing them works well, otherwise it results in 404.

## Expected behaviour

It's naturally expected that routing will work well regardless of the order of function calls.

## Actual behaviour

Given `sub_app_cbv.py`:
```python
import sys
import aiohttp.web

API_URL_PREFIX = r'/api/v1'

class PingView(aiohttp.web.View):
    async def get(self):
        return aiohttp.web.Response(text='pong')

def just_app_with_routes_def(app, api_app):
    app.router.add_route('GET', API_URL_PREFIX + r'/ping/', PingView)

def sub_app_after_routes_def(app, api_app):
    api_app.router.add_route('GET', r'/ping/', PingView)
    app.add_subapp(API_URL_PREFIX, api_app)

def sub_app_before_routes_def(app, api_app):
    app.add_subapp(API_URL_PREFIX, api_app)
    api_app.router.add_route('GET', r'/ping/', PingView)

def run_app(coro_name):
    print(f'Going to run {coro_name}')
    setup_app = globals()[coro_name]
    app = aiohttp.web.Application()
    api_app = aiohttp.web.Application()
    setup_app(app, api_app)
    aiohttp.web.run_app(app)

def main():
    if len(sys.argv) < 2:
        raise RuntimeError('Supply one cli argument plz [just_app_with_routes_def|sub_app_after_routes_def|sub_app_before_routes_def]')
    coro_name = sys.argv[1]
    run_app(coro_name)

__name__ == '__main__' and main()
```

##### Demo 1

```shell
$ python sub_app_cbv.py just_app_with_routes_def
Going to run just_app_with_routes_def
======== Running on http://0.0.0.0:8080 ========
(Press CTRL+C to quit)
```

```shell
$ http :8080/api/v1/ping/
HTTP/1.1 200 OK
Content-Length: 4
Content-Type: text/plain; charset=utf-8
Date: Thu, 11 Jan 2018 23:17:52 GMT
Server: Python/3.6 aiohttp/2.3.7
pong
```

##### Demo 2

```shell
$ python sub_app_cbv.py sub_app_after_routes_def
Going to run sub_app_after_routes_def
======== Running on http://0.0.0.0:8080 ========
(Press CTRL+C to quit)
```

```shell
$ http :8080/api/v1/ping/
[...]

PR #2657, opened 2018-01-12
Freeze subapp on adding to parent

asvetlov:
Fix for #2656

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -163,6 +163,7 @@ class Application(MutableMapping):
         self.router.register_resource(resource)
         self._reg_subapp_signals(subapp)
         self._subapps.append(subapp)
+        subapp.freeze()
         if self._loop is not None:
             subapp._set_loop(self._loop)
         return resource
--- a/CHANGES/2656.bugfix
+++ b/CHANGES/2656.bugfix
@@ -1,0 +1,1 @@
+Freeze sub-application on adding to parent app
--- a/tests/test_web_app.py
+++ b/tests/test_web_app.py
@@ -236,3 +236,11 @@ def test_app_run_middlewares():
     root.add_subapp('/sub', sub)
     root.freeze()
     assert root._run_middlewares is True
+
+
+def test_subapp_frozen_after_adding():
+    app = web.Application()
+    subapp = web.Application()
+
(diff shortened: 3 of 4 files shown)


aiohttp/web_app.py
```diff
@@ -163,6 +163,7 @@ def add_subapp(self, prefix, subapp):
         self.router.register_resource(resource)
         self._reg_subapp_signals(subapp)
         self._subapps.append(subapp)
+        subapp.freeze()
```
pfreixes, line 166: Even though is not part of that PR I'm wondering if makes sense does not allow add sub apps that have been frozen before. There is an explicit check for that few lines above.

asvetlov: What is your use case?
Sharing sub-app instance between different parents / prefixes can open a can of worms.
The restriction was exist from very beginning IIRC

pfreixes: Just trying to understand the why of this check. Looks like an implicit way to check that the sub app has been already added as a child of one app (indeed could be the same).

asvetlov: Yes, subapp should be used only once. If you need reuse -- please create another subapp instance.

pfreixes:
No `CHANGES`?

webknjaz:
If the app is one-shot instance, let's store its prefix there.

asvetlov:
Forgot to push already committed changes.

asvetlov:
> If the app is one-shot instance, let's store its prefix there.

Sorry, I don't follow.
Application prefix is stored in `PrefixedSubAppResource` (the resource is required anyway).

webknjaz:
So how do I access `app.prefix`?

asvetlov:
What is your use case? There is no such attribute as `handler.path`, why do you need `app.prefix`?

pfreixes:
Feel free to merge when you consider that the conversation is resolved.

webknjaz:
@asvetlov I've got swagger package behaving strange with base url. I have a hope to try improving it by analyzing subapp prefixes.

asvetlov:
Do you mean aiohttp-swagger?
If the library has a bug -- it should be fixed.
I doubt if exposing subapp prefix in public API is the best way to do it.

Please open a new issue if needed

Issue #2662, opened 2018-01-15
ImportError: cannot import name 'unquote'


ilmaruk:
## Long story short

As of yarl 1.0.0 (available since one hour ago), yarl.unquote is no longer available, thus causing an import error at web_urldispatcher.py.

## Expected behaviour

Been able to run: `from aiohttp import web`

## Actual behaviour

`from aiohttp import web` raises the following error:
```Traceback (most recent call last):
  File "web.py", line 1, in <module>
    from aiohttp import web
  File "/home/ruggero/Development/mqtt/.venv/lib/python3.6/site-packages/aiohttp/web.py", line 15, in <module>
    from . import (hdrs, web_exceptions, web_fileresponse, web_middlewares,
  File "/home/ruggero/Development/mqtt/.venv/lib/python3.6/site-packages/aiohttp/web_middlewares.py", line 5, in <module>
    from aiohttp.web_urldispatcher import SystemRoute
  File "/home/ruggero/Development/mqtt/.venv/lib/python3.6/site-packages/aiohttp/web_urldispatcher.py", line 21, in <module>
    from yarl import URL, unquote
ImportError: cannot import name 'unquote'
```

## Steps to reproduce

Simply: `from aiohttp import web`

## Your environment

aiohttp 2.3.7 (both)
Ubuntu Linux 16.04
Python 3.6.3

## Workaround

Downgrade to yarl 0.18.0

gcarq:
Same problem since yarl 1.0.0, downgrading to `yarl==0.18.0` helps as workaround.

```
13:36:06   File "/var/jenkins/workspace/sometest-01-test/test/__init__.py", line 19, in <module>
13:36:06     from aiohttp import web
13:36:06   File "/var/jenkins/shiningpanda/jobs/bd59e0bb/virtualenvs/d41d8cd9/lib/python3.6/site-packages/aiohttp/web.py", line 15, in <module>
13:36:06     from . import (hdrs, web_exceptions, web_fileresponse, web_middlewares,
13:36:06   File "/var/jenkins/shiningpanda/jobs/bd59e0bb/virtualenvs/d41d8cd9/lib/python3.6/site-packages/aiohttp/web_middlewares.py", line 5, in <module>
13:36:06     from aiohttp.web_urldispatcher import SystemRoute
[...]

PR #2665, opened 2018-01-15
Freeze YARL version to avoid issue #2662

pfreixes:
## What do these changes do?

Temporary fix for #2662

No

## Related issue number

#2662


samuelcolvin:
this duplicates #2664

asvetlov:
Fixed by #2667

PR #2690, opened 2018-01-25
Use attrs library for data classes

asvetlov:


aiohttp/client_reqrep.py
```diff
@@ -39,12 +39,18 @@
 __all__ = ('ClientRequest', 'ClientResponse', 'RequestInfo', 'Fingerprint')
 
 
-ContentDisposition = collections.namedtuple(
-    'ContentDisposition', ('type', 'parameters', 'filename'))
+@attr.s(frozen=True)
```
pfreixes, line 42: I have my doubts about the usage of `frozen=True` taking into account that immutability depends on the type of fields that you have in the data class. In the worst case scenario having a mutable attribute, the `hash` will fail and you won't prevent modify these attributes.

asvetlov: Sorry, I don't follow your point.

`ConetentDisposition` was immutable named tuple, I've just converted the class into immutable `attrs` representation. All `type`, `parameters` and `filename` attributes are immutable too.

aiohttp/helpers.py
```diff
@@ -187,7 +192,12 @@ def isasyncgenfunction(obj):
     return False
 
 
-MimeType = namedtuple('MimeType', 'type subtype suffix parameters')
+@attr.s(frozen=True)
+class MimeType:
+    type = attr.ib(type=str)
+    subtype = attr.ib(type=str)
+    suffix = attr.ib(type=str)
+    parameters = attr.ib(type=MultiDict)
```
pfreixes, line 200: `MultiDict` is mutable `frozen=True` is IMHO inconsistent in that case.

aiohttp/client_reqrep.py
pfreixes:
```>>> @attr.s(frozen=True)
... class Foo:
...     a = attr.ib(type=MappingProxyType)
...
>>> d = {}
>>> m = MappingProxyType(d)
>>> f = Foo(m)
>>> hash(f)
  File "<attrs generated hash a57df858a086a8e0acf97d938d787d39b6f33d51>", line 4, in __hash__
TypeError: unhashable type: 'mappingproxy'```

File: aiohttp/web_request.py [hunk 3]
```diff
@@ -24,8 +26,15 @@
 
 __all__ = ('BaseRequest', 'FileField', 'Request')
 
-FileField = collections.namedtuple(
-    'Field', 'name filename file content_type headers')
+
+@attr.s(frozen=True, slots=True)
```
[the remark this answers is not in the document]
asvetlov: Actually not . I forgot adding new param and was experimenting with different usage.

kxepal:
Btw, what do you think about dataclasses? Would we migrate to them eventually or will stay with attrs?

aiohttp/web_request.py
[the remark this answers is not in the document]
asvetlov: Let's not use slots in attrs, they are not bring new value but slightly slower.

asvetlov:
I want keeping `frozen` classes at least by default. It's safer.

webknjaz:
By the way, `attrs` has a built-in `@dataclass` decorator

asvetlov:
@webknjaz I know about dataclass support embedded into attrs, but we cannot use it until dropping Python 3.5 support.

kxepal:
@asvetlov 
I mean...dataclasses are currently provided as standalone package, aren't they? We could just use them as optional dependency for <3.7.

asvetlov:
https://pypi.python.org/pypi/dataclasses requires Python 3.6, we are limited to 3.5.3

kxepal:
Yeah. Ok, I got your point. Thanks!

webknjaz:
What if there was a backport?

asvetlov:
@webknjaz don't follow

pfreixes:
@asvetlov I don't see the point of usage by default `frozen` because its safer. The old `namedtuples` didn't give you that behavior. Start using the `frozen=True` means that you would give this feature for just for those params that are immutable. So, safer means that some attributes will be protected and others no? What I'm missing?

asvetlov:
`frozen` prevents inplace modification:
```
In [1]: import attr

In [2]: @attr.s
   ...: class A:
   ...:     a = attr.ib()
   ...:     

In [3]: a = A(1)

In [4]: a.a
Out[4]: 1

In [5]: a.a = 2

In [6]: a.a
Out[6]: 2

In [7]: @attr.s(frozen=True)
   ...: class B:
   ...:     b = attr.ib()
   ...:     

In [8]: b = B(1)

In [9]: b.b
Out[9]: 1

In [10]: b.b = 2
---------------------------------------------------------------------------
FrozenInstanceError                       Traceback (most recent call last)
<ipython-input-10-dc6c31b33bcf> in <module>()
----> 1 b.b = 2

~/.virtualenvs/aiohttp/lib/python3.6/site-packages/attr/_make.py in _frozen_setattrs(self, name, value)
    349     Attached to frozen classes as __setattr__.
    350     """
--> 351     raise FrozenInstanceError()
    352 
    353 

FrozenInstanceError: 
```

webknjaz:
@asvetlov I mean, shouldn't it be possible to backport dataclasses implementation to support 3.5?

asvetlov:
@webknjaz no, [PEP 526](https://www.python.org/dev/peps/pep-0526/) was implemented in Python 3.6 only.
Dataclasses require this feature.

pfreixes:
this is clear, Im just saying that the usage of  frozen can become error
prone, this feature do not prevent manipulate mutable data, perhaps
foo.headers.pop('header name').

IMHO this flag can give you a false sensation of protection.

webknjaz:
Fair enough

asvetlov:
@pfreixes I understand that `frozen` is not recursive property and all consequence pitfalls.
But frozen classes are better than non-frozen, at least in namedtuple replacement context.

pfreixes:
Well, at least the whole replacement is safer for all variables, and must be the developer who has the last call to expose a variable mutable or not mutanle. A triki situatuion.

Anyway, lets move forward. LGTM

asvetlov:
`slots=True` should be mandatory also -- without the flag attrs doesn't prevent setting a custom attribute, which is not desired behavior usually.

PR #2697, opened 2018-01-30
Runner refactoring

asvetlov:


aiohttp/web_runner.py
```diff
@@ -218,3 +220,51 @@ def _unreg_site(self, site):
...
         self._sites.remove(site)
+
+
+class ServerRunner(BaseRunner):
+    """Low-level web server runner"""
+
+    __slots__ = ('_web_server',)
+
+    def __init__(self, web_server, *, handle_signals=False, **kwargs):
+        super().__init__(handle_signals=handle_signals, **kwargs)
```
kxepal, line 231: Is there any need to duplicate `False` value even if base class has the same default?

aiohttp/test_utils.py
```diff
@@ -42,27 +36,26 @@ def __init__(self, *, scheme=sentinel, loop=None,
...
-            self._socket.bind((self.host, self.port))
-        else:
-            self._socket.bind((self.host, 0))
-            self.port = self._socket.getsockname()[1]
         self._ssl = kwargs.pop('ssl', None)
+        self.runner = await self._make_runner(**kwargs)
+        await self.runner.setup()
+        if not self.port:
+            self.port = unused_port()
+        site = TCPSite(self.runner, host=self.host, port=self.port,
```
kxepal, line 56: Previously, if port was None, it turns into 0 e.g. random. Now it would set as some default `8443 if self._ssl_context else 8080`. Is that ok?

kxepal: Also, what about port/address reuse? Sounds like they are pretty useful in tests where server starts and shuts down quite frequently.

kxepal: Ah, nvm about zero port - missed `unused_port()` call right in front of me /:

asvetlov: asyncio reuses address on UNIX, should be enough.

aiohttp/web_runner.py
asvetlov: It is a tradeoff.
I believe duplication is better in terms of code readability and documentation.

kxepal: ok

aiohttp/test_utils.py
kxepal: Good to know!

Issue #2686, opened 2018-01-24
Change client tracing signatures


asvetlov:
Now tracing signal signatures look like 
```
async def on_request_start(session, trace_config_ctx, method, host, port, headers): ...
async def on_send_connection_queued_start(session, trace_config_ctx): ...
```

The problem is the approach is has no space for future extensions.
Most likely we will want to add a connection info to `on_send_connection_queued_start` but it is impossible to do without breaking all existing signal handlers: sending additional `connection` parameter cannot be accepted.

Sure, we can enforce users to write handlers with mandatory `*args` and `**kwargs` like
```
async def on_send_connection_queued_start(session, trace_config_ctx, *args, **kwargs): ...
```
but I expect that users will constantly forgot these extra arguments.

As a solution we can change signal signature for all signals to `async def on_sig(session, ctx, data): ...` where `ctx` is `trace_config_ctx` and `data` is a structure with all signal parameters: url, host, port, headers, connection etc.
It allows to add new param to data in backward compatible manner.

Next question is the `data` type.
I see several options:
1. Plain dict with keys like `host` and `connection`. It works but looks too old fashion, everybody prefer attributes to string keys.
2. [SimpleNamespace](https://docs.python.org/3/library/types.html?highlight=simplenamespace#types.SimpleNamespace). `data.url` looks better than `data['url']`
3. Custom data classes. Honestly it is even better than `SimpleNamespace` at least from documentation perspective: we will need to document every used data structure explicitly.
4. [attrs](https://pypi.python.org/pypi/attrs) classes. The library brings extra dependency (I don't care, aiohttp has several deps already) but provides neat helpers. Using data classes from Python 3.7 is even better but we should support Python 3.5/3.6 for long time.

`Trace` class already has explicit methods for sending every signal, constructing a data class on call is easy and straightforward.

@pfreixes and others, what do you think about?

pfreixes:
Definitly I wil go for something explicit that incorporates documentation. What about thr idea of custom data classes but based on simple namedtuples?

asvetlov:
I think deriving from SimpleNamespace makes no sense.

asvetlov:
Do you have objections against attrs library?

pfreixes:
I didnt know about attrts, looks great. But just for simplicity and consistence - current aiohttp code usage - and taking into account that the use case is just a need of having a explicit way to declare our data params.

Do you believe that right now the features that this library gives are gonna make a difference from the begining?

[...]

PR #2699, opened 2018-01-30
Use custom classes to pass client signals parameters #2686

pfreixes:
## What do these changes do?

This will allow Aiohttp development add new parameters in the future without
break the signals signature.

Yes

## Related issue number

#2686

## Checklist

### DIFF
--- a/docs/tracing_reference.rst
+++ b/docs/tracing_reference.rst
@@ -31,8 +31,8 @@ the request flow.
       Property that gives access to the signals that will be executed when a
       request starts, based on the :class:`aiohttp.signals.Signal` implementation.
 
-      The coroutines listening will receive as a param the ``session``,
-      ``trace_config_ctx`` and :class:`aiohttp.TraceRequestStartParams` params.
+      The signal handler signature is ``async def on_request_start(session, context, params): ...``
+      where ``params`` is :class:`aiohttp.TraceRequestStartParams` instance
 
       .. versionadded:: 3.0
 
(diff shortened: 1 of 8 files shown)


docs/tracing_reference.rst
```diff
@@ -29,10 +29,10 @@ the request flow.
    .. attribute:: on_request_start
 
       Property that gives access to the signals that will be executed when a
-      request starts, based on the :class:`~signals.Signal` implementation.
+      request starts, based on the :class:`aiohttp.signals.Signal` implementation.
 
       The coroutines listening will receive as a param the ``session``,
-      ``trace_config_ctx``, ``method``, ``url`` and ``headers``.
+      ``trace_config_ctx`` and :class:`aiohttp.TraceRequestStartParams` params.
```
asvetlov: Minor question.
Maybe ```The signal handler signature is ``async def on_request_start(session, context, params): ...`` where ``params`` is :class:`aiohttp.TraceRequestStartParams` instance``` for better readability?

The same for all other tracing signals.

asvetlov:
Looks good in overall.

docs/tracing_reference.rst
pfreixes: Done!

jettify:
I reworked aiozipkin (https://github.com/aio-libs/aiozipkin/pull/95) to use this API, I think new one is much better and easier to use! 👍

asvetlov:
cool!

Issue #2397, opened 2017-10-25
Design good distribution extras support


webknjaz:
Follow-up https://github.com/aio-libs/aiohttp/pull/2312#issuecomment-339335603: this issue is intended to encourage discussion about adding extras support.

There's several optional dependencies mentioned in docs, and they instruct users to do `pip install {{dep_name}}`.

Common practice is to add feature-flags allowing to depend on certain additional ability of the framework w/o having to install those dependencies manually.

For example, aiohttp could have `speedups` extra and then user's could've just `pip install aiohttp[speedups]` instead of `pip install aiodns aiohttp cchardet`.

Also, it is possible to use environment markers as extra names, in this case certain dependencies get installed automatically if the environment matches specific expression, which might be useful for windows-only crutches etc.

Please consider treating extras as a **public API**, because they are exposed to end-users.

/cc: @asvetlov

asvetlov:
I like the proposal, please go ahead

asvetlov:
Can we close it? 
#3412 is merged.

webknjaz:
Let's keep it for discussion and close upon final edits. It's a public API, we'll end up maintaining it as accepted, remember?

asvetlov:
Agree

webknjaz:
`charset-guessing-speedup`, `async-dns` and `brotli-compression`?

cc @WouldYouKindly

asvetlov:
Perhaps my type speed is not enough for `pip install aiohttp[charset-guessing-speedup,async-dns,brotli-compression]`.

I'm starting to wonder if we need separate entry points for every feature. One point for installing everything is better maybe? If a user wants something specific -- he can learn from `extra_requires` what to install.

WouldYouKindly:
@webknjaz @asvetlov why not just `speedups`, `async-dns`, and `brotli`? 

`charset-guessing-speedup` / `speedups`
Those who want speed either don't really care where it comes from, or are willing to look into `setup.py` and see exact libraries — so I see no need for such a descriptive name. If later you add more speedups, extras API would not need to change.

`brotli` — the word is almost always used for a compression algorithm, so we may omit that part.

kxepal:
I think it would be better to limit support of extras by two cases: all extras on and all extras off. Otherwise we'll be burried by exponential growing combinations of those which need to support. 

Is there a case when you need just some specific ones?

webknjaz:
[...]

PR #2703, opened 2018-02-02
Added check install idna-ssl or not on Python3.7

hellysmile:
## What do these changes do?

idna_ssl is not installed on Python3.7 as dependency 

otherwise on Python3.7 will be

```bash
idna-ssl requires Python '<3.7.0' but the running Python is 3.7.0
```


asvetlov:
Yep, maybe your proposal is more elegant -- but I don't care too much

webknjaz:
it's just a note for future if I ever get to doing #2397

PR #2709, opened 2018-02-05
Fix typo from `pyenv` to `pyvenv`

delirious-lettuce:
## What do these changes do?

Fix a typo, [`pyenv`](https://github.com/pyenv/pyenv) is different from [`pyvenv`](https://docs.python.org/3/library/venv.html?highlight=pyvenv)

## Related issue number

N/A, trivial update

### DIFF
--- a/docs/deployment.rst
+++ b/docs/deployment.rst
@@ -214,7 +214,7 @@ Create a directory for your application::
   >> mkdir myapp
   >> cd myapp
 
-`Ubuntu` has a bug in pyenv, so to create virtualenv you need to do some
+`Ubuntu` has a bug in pyvenv, so to create virtualenv you need to do some
 extra manipulation::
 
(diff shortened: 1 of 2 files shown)


webknjaz:
@delirious-lettuce plz add it to dictionary, so that sphinx's spellcheck won't complain

delirious-lettuce:
@webknjaz ,

My mistake, I'll add it right away. Just to double check, you mean add it to `aiohttp/docs/spelling_wordlist.txt` ?

webknjaz:
Yes please

asvetlov:
Is the info about Ubuntu bug in pyvenv still relevant?

asvetlov:
Fixed in master.
Thanks for pointing on the problem

delirious-lettuce:
@asvetlov ,

No problem, thanks for your time!

Issue #2578, opened 2017-11-30
Review pytest fixture names


asvetlov:
Perhaps we need to add `aiohttp_` prefix to all fixtures to prevent clashing with other libraries.
Existing names should be supported for sake of backward compatibility at least up to aiohttp 4.0.

Thoughts?

fafhrd91:
Those just tests and I think should be used by third party packages. It was a not logical change, upgrade is just find and replace. I’d add prefixes for 3.0

asvetlov:
The fixtures are used widespread, by `aio-libs/aiohttp-*` also.
I want to not break all these libs at moment of aiohttp 3.0 release.

asvetlov:
Fixed by #2714

PR #2714, opened 2018-02-09
Fixture renaming

asvetlov:
Fix for #2578
Rename everything (keeping backward compatibility though)


tests/test_pytest_plugin.py
```diff
@@ -116,29 +116,29 @@ def test_noncoro():
...
     with pytest.raises(RuntimeError):
-        await test_client(make_app)
+        await aiohttp_client(make_app)
 
 
-async def test_custom_port_test_client(test_client, unused_port):
-    port = unused_port()
-    client = await test_client(create_app, server_kwargs={'port': port})
+async def test_custom_port_aiohttp_client(aiohttp_client, aiohttp_unused_port):
+    port = aiohttp_unused_port()
```
webknjaz, line 129: @asvetlov 
Why use it at all if it's possible to instruct app to bind 0 and read the allocated port when needed? That fixture unbinds from port, meaning it might be intercepted by smth else as a result of race condition.
I'd go for `SO_REUSEPORT` or socket activation in order to make it more reliable.

asvetlov: The approach is widely used by asyncio tests, I just borrowed the idea.

port 0 may work but the change is definitely out of the PR scope.
Feel free to create a new PR if needed.

webknjaz: Yea, I understand. It's just that I've been concerned about it during some time now. And decided to shout out about possible flaky behavior in test flow. I would at least not expose this as a public API to end-users, encouraging them to use port=0

asvetlov: As first step we should get rid of the function in aiohttp test suite itself.
Maybe in next releases?

PR #2735, opened 2018-02-13
add a ssl_context option to fingerprint - to allow the developer to override the automatically created ssl_context

stephan48:
## What do these changes do?

I faced the problem that with the aiohttp 3.0 ssl api changes i could no longer use fingerprint verification and client cert auth. As you could only specify one or the other. After a short talk with the developer we came to the conclusion that a way to solve the problem would be to add a ssl_context kwargs to the Fingerprint. This PR does exactly that.

Documentation is still missing.

As with the previous API aiohttp <=2 you can now use both methods again.

## Related issue number

Nope

## Checklist

- will work on this tomorrow - its already late -


aiohttp/connector.py
```diff
@@ -771,12 +778,18 @@ def _get_ssl_context(self, req):
             sslcontext = req.ssl
             if isinstance(sslcontext, ssl.SSLContext):
                 return sslcontext
+            if not (not isinstance(sslcontext, Fingerprint) or not (sslcontext.ssl_context is not None) or not isinstance(
```
webknjaz: There's too many negations, which is hard to understand instantly and easy to misunderstand. I think, this can (and should) be simplified.

aiohttp/connector.py
```diff
@@ -771,12 +778,18 @@ def _get_ssl_context(self, req):
...
+                    sslcontext.ssl_context, ssl.SSLContext)):
+                return sslcontext.ssl_context
             if sslcontext is not None:
                 # not verified or fingerprinted
                 return self._make_ssl_context(False)
             sslcontext = self._ssl
             if isinstance(sslcontext, ssl.SSLContext):
                 return sslcontext
+            if not (not isinstance(sslcontext, Fingerprint) or not (sslcontext.ssl_context is not None) or not isinstance(
```
webknjaz: This is a duplicate check. Consider code reuse, so that it won't result in diverse copy-paste in future.

aiohttp/connector.py
webknjaz: It seems to me that you want this check:
```python
isinstance(sslcontext, Fingerprint) and isinstance(sslcontext.ssl_context, ssl.SSLContext)
```

stephan48:
Hi, thank you for your comments I used some function in PyStorm to break the line - this apparently lead to it rewriting the expression without me checking. I reworked it - i would like to keep the "is not None" check as a precondition. I fixed the negation mess throu.

But I see that this check is not done in the other branches - is isinstance "None"-safe?

webknjaz:
@stephan48 this is why I use Vim :)

isinstance is "None-safe": `None` is instance of the `NoneType`

stephan48:
yea :) for work stuff i tend to use windows & a IDE. for everything else vim on a ssh connection.

thanks for that info.

how about this:

```python
            if ssl is None:  # pragma: no cover
                raise RuntimeError('SSL is not supported.')

            def check_ssl(ssl_ctx):
                if isinstance(ssl_ctx, ssl.SSLContext):
                    return ssl_ctx
                if isinstance(ssl_ctx, Fingerprint) and isinstance(ssl_ctx.ssl_context, ssl.SSLContext):
                    return ssl_ctx.ssl_context
                if ssl_ctx is not None:
                    # not verified or fingerprinted
                    return self._make_ssl_context(False)

            if req.ssl is not None:
                return check_ssl(req.ssl)

            if self._ssl is not None:
                return check_ssl(self._ssl)

            return self._make_ssl_context(True)
```

webknjaz:
Your last if/return in `check_ssl()` is always true (I mean, `if` not needed there at all). But if looks unreliable to me in current form.

webknjaz:
Oh, and please update code in branch, cause it's fairly complicated to comment on such snippet + it's ripped out of context, which makes it hard to imagine how it would play with other parts of code around it. TIA.

asvetlov:
ping

stephan48:
Hi,

sadly I got overwhelmed by private and work stuff and i totally forgot to check back in here :(

Would the changes/the version in the last commit(pre merge) be adequate?

-- stephan

stephan48:
Do I assume correctly, that the code coverage complains about the new check_ssl function being undocumented?

aiohttp/connector.py
```diff
@@ -781,18 +789,23 @@ def _get_ssl_context(self, req):
...
-                return self._make_ssl_context(False)
-            sslcontext = self._ssl
-            if isinstance(sslcontext, ssl.SSLContext):
-                return sslcontext
-            if sslcontext is not None:
+
+            def check_ssl(ssl_ctx):
+                if isinstance(ssl_ctx, ssl.SSLContext):
+                    return ssl_ctx
+                if isinstance(ssl_ctx, Fingerprint) and \
```
asvetlov: Please use brackets:
```
    if (isinstance(ssl_ctx, Fingerprint) and
            isinstance(ssl_ctx.ssl_context, ssl.SSLContext)):
        return ssl_ctx.ssl_context
```

aiohttp/connector.py
```diff
@@ -781,18 +789,23 @@ def _get_ssl_context(self, req):
...
+                   isinstance(ssl_ctx.ssl_context, ssl.SSLContext):
+                    return ssl_ctx.ssl_context
```
asvetlov: The line is not covered by tests

aiohttp/connector.py
```diff
@@ -781,18 +789,23 @@ def _get_ssl_context(self, req):
...
-                return sslcontext
-            if sslcontext is not None:
-                # not verified or fingerprinted
-                return self._make_ssl_context(False)
-            sslcontext = self._ssl
-            if isinstance(sslcontext, ssl.SSLContext):
-                return sslcontext
-            if sslcontext is not None:
+
+            def check_ssl(ssl_ctx):
```
asvetlov: The function has no closures, please rename it into `_check_ssl` and move to the module's top-level namespace

aiohttp/connector.py
```diff
@@ -764,13 +764,23 @@ def _make_ssl_context(verified):
             sslcontext.set_default_verify_paths()
             return sslcontext
 
+    def _check_ssl(self, ssl_ctx):
+        if isinstance(ssl_ctx, ssl.SSLContext):
+            return ssl_ctx
+        if (isinstance(ssl_ctx, Fingerprint) and
+                isinstance(ssl_ctx.ssl_context, ssl.SSLContext)):
+            return ssl_ctx.ssl_context
```
asvetlov, line 865: Please add a test for covering the line execution.

aiohttp/connector.py
```diff
@@ -781,18 +791,13 @@ def _get_ssl_context(self, req):
...
-                # not verified or fingerprinted
-                return self._make_ssl_context(False)
-            sslcontext = self._ssl
-            if isinstance(sslcontext, ssl.SSLContext):
-                return sslcontext
-            if sslcontext is not None:
-                # not verified or fingerprinted
-                return self._make_ssl_context(False)
+
+            if req.ssl is not None:
```
webknjaz, line 888: smth like this would look nicer that duplicate if-clauses:
```python
_ssl = req.ssl or self._ssl
if _ssl is not None:
    return self._check_ssl(_ssl)
```

asvetlov:
I like the idea but sorry, I should close the PR.
It has no tests and documentation updates.
Please feel free to open a new PR that has all the required fixes.

PR #2743, opened 2018-02-16
Fix bug with None value in server response headers

### LINKED ISSUE #2742: [server] Unhandled exception when None in header values
## Long story short

When a dictionary is passed to the `headers` kwarg (in any response, either returned or raised) and a value in this dictionary is `None`, an unhandled exception is raised that results in the client getting no response at all.

## Expected behavior

The expected behavior would be to handle it in the same way an empty string header is handled (i.e. discard it) or to catch the exception so that a 500 error can be returned.

## Actual behavior

No response arrives at the client.

## Steps to reproduce

```py
import asyncio

from aiohttp import web


def view_func(request):
    raise web.HTTPUnauthorized(headers={'a':''})


def view_raise_fail(request):
    raise web.HTTPUnauthorized(headers={'a': None})


def view_return_fail(request):
    return web.Response(status=200, headers={'a': None})


def get_app():
    app = web.Application()
    app.router.add_route('GET', '/', view_func)
    app.router.add_route('GET', '/raise/fail/', view_raise_fail)
    app.router.add_route('GET', '/return/fail/', view_return_fail)
    return app


if __name__ == '__main__':
    app = get_app()
    web.run_app(app, port=8080, host='127.0.0.1')

```

```sh
wget http://127.0.0.1:8080/
--2018-02-16 08:47:40--  http://127.0.0.1:8080/
Connecting to 127.0.0.1:8080... connected.
HTTP request sent, awaiting response... 401 Unauthorized
```

```sh
wget http://127.0.0.1:8080/raise/fail/
--2018-02-16 08:48:31--  http://127.0.0.1:8080/raise/fail/
Connecting to 127.0.0.1:8080... connected.
HTTP request sent, awaiting response... No data received.
Retrying.
```

```sh
wget http://127.0.0.1:8080/return/fail/
--2018-02-16 08:48:43--  http://127.0.0.1:8080/return/fail/
Connecting to 127.0.0.1:8080... connected.
HTTP request sent, awaiting response... No data received.
Retrying.
```

The Traceback:

```py
Unhandled exception
  File "<PATH>/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 396, in start
    await resp.prepare(request)
  File "<PATH>/lib/python3.6/site-packages/aiohttp/web_response.py", line 300, in prepare
    return self._start(request)
  File "<PATH>/lib/python3.6/site-packages/aiohttp/web_response.py", line 605, in _start
    return super()._start(request)
  File "<PATH>/lib/python3.6/site-packages/aiohttp/web_response.py", line 367, in _start
    writer.write_headers(status_line, headers)
  File "<PATH>/lib/python3.6/site-packages/aiohttp/http_writer.py", line 98, in write_headers
    [k + SEP + v + END for k, v in headers.items()])
  File "<PATH>/lib/python3.6/site-packages/aiohttp/http_writer.py", line 98, in <listcomp>
    [k + SEP + v + END for k, v in headers.items()])
TypeError: must be str, not NoneType
```

## Your environment

OS version : Linux deb9 4.9.0-3-amd64 #1 SMP Debian 4.9.30-2+deb9u5 (2017-09-19) x86_64 GNU/Linux

Python version : 3.6.4

aiohttp server version: 3.0.1

panagiks:
## What do these changes do?

Fixes #2742 . Handle None in server Response Headers.

No. Unless the user had encountered this bug.

## Related issue number

#2742 

## Checklist


aiohttp/http_writer.py
```diff
@@ -95,7 +95,7 @@ def write_headers(self, status_line, headers, SEP=': ', END='\r\n'):
         """Write request/response status and headers."""
         # status + headers
         headers = status_line + ''.join(
-            [k + SEP + v + END for k, v in headers.items()])
+            [k + SEP + v + END for k, v in headers.items() if v is not None])
```
webknjaz: "eating" some headers is misleading for the end-user. This must fail loudly. Better would be raising exception with a sane problem description.

panagiks: It's already done implicitly when `''` is passed as a header value. This was my reasoning (as mentioned in the related issue). 

I'm open to changing it to raise an error by try caching it but I think that:

a) The same should be done for `''` values, which is trickier
b) It's a bit tricky to catch the `TypeError` and raise, since `TypeError` can be many things.

But considering the above, feel free to tell me what you prefer and I'll change it :)

asvetlov:
Sending empty header value is pretty correct but passing anything except str is a error

panagiks:
With the new change the following will be logged to the console when a None value is passed to a response's header:
``Error handling request
TypeError: Header value cannot be None! ``

In order to have the actual trace in the error log the following would have to be executed and try/excepted:
```py 
await resp.prepare(request)
```
But:
a) Catching TypeError at such a broad scope (I think) would not be the best practice
b) I would require recalling the exact same command after running `handle_error`. (code repetition)
c) The stack that is returned only includes aiohttp code so (again I think) it wouldn't really help a user of the framework (i.e. developer) understand what he/she did wrong.

I tried to make it as close to the existing error handling as possible. If you would like a different message or any other change please let me know :)

asvetlov:
I think you've applied the check to wrong place.
`response.prepare()` could be called from handler in case of streaming response or websocket. Handling it there where you do is too late for such response types.

panagiks:
Ok I moved the check in StreamResponse constructor. I wanted to avoid it as it doesn't cover the case where someone edits the `._headers' after the creation of `StreamResponse` object, but considering that it's an implementation detail (and there is no method, that I could find, for the user to set the headers after the object's creation) it's probably ok.

aiohttp/web_response.py
```diff
@@ -54,6 +54,8 @@ def __init__(self, *, status=200, reason=None, headers=None):
 
         if headers is not None:
             self._headers = CIMultiDict(headers)
+            if None in headers.values():
+                raise RuntimeError("Header value can't be None!")
```
webknjaz: m/b `ValueError`?

panagiks: Sure :) I'll change it now.

aiohttp/web_response.py
```diff
@@ -54,6 +54,8 @@ def __init__(self, *, status=200, reason=None, headers=None):
 
         if headers is not None:
             self._headers = CIMultiDict(headers)
+            if None in headers.values():
```
webknjaz, line 57: This must be slow.
@asvetlov wouldn't it be more efficient to put the check in multidict around `__setitem__()`?

asvetlov:
I very sorry, but after careful thinking I'm inclining to adding a function called `build_headers` or something like this to multidict itself.
The implementation can be optionally Cythonized.
@panagiks would you continue your work but in multidict library?

panagiks:
Sure :) I'll have a look at the library and open a PR to multidict with a proof of concept implementation (probably around the coming weekend) and we can continue the discussion there.

I'll go ahead and close this PR.

PR #2747, opened 2018-02-17
App cleanup context

asvetlov:
A fix for #2092


asvetlov:
The feature is done (except documentation).
The only questionable thing is the new signal name.
Now it is `app.on_cleanup_ctx`.
Suggestions?

asvetlov:
@aio-libs/aiohttp-committers please take a look.
If nobody will invite a better name than `app.on_cleanup_ctx` I'll merge the PR after adding a documentation for the feature.

kxepal:
I don't see any `app.on_cleanup_ctx` around in this PR, but `app.cleanup_ctx` fits nice.

asvetlov:
Sorry, it is definitely `app.cleanup_ctx` property.

kxepal:
It has a good name since it refrect exactly what it reference to - CleanupContext. It's hard to figure something better for me. Why do you question this name? Do you have any different ideas about?

asvetlov:
No other ideas actually. 
But I was not sure, maybe somebody can suggest something better?

code-cro:
I suggest full name cleanup_context to be more clear for the reader.

asvetlov:
Thanks for sharing opinion. Ctx is pretty common abbreviation.
Python uses such shortcuts everywhere in stdlib.
cleanup_context is really long name

Issue #2689, opened 2018-01-25
Sharing db connection to nested nested app


fduxiao:
## Long story short

I've read  your answer in #2413 (raised in #2412 ).
The solution (`subapp['db'] = mainapp['db']`) does not work when it comes to nested subapps.
For example, if I have to build an app A, which have some subapp B and I need to pass the db connection from A to B, which can be done in that behavior. But, when I need to nest app A as a subapp to C, another app, and how should I pass db from C to A and to B. Clearly, the sharing between A and B exists before that between C and A, when A does not know the existence of db (which is even set inside startup) and C should also know nothing about the inner structure of A. A more complicated mechanism should be provided. 

## Expected behaviour

A more complicated solution. 

## Actual behaviour

Too simple to solve the problem. 

## Steps to reproduce

```python
from aiohttp import web

app_b = web.Application()

app_a = web.Application()
app_b['db'] = app_a['db']  # error
app_a.add_subapp('/b', app_b)

app_c = web.Application()
app_c['db'] = 'db'  # actually inside a startup event
app_c.add_subapp('/a', app_a)  # how should I pass app_c['db']

web.run_app(app_c)

```
## My solution
I'm trying to solve the problem by making the db become a `promise` (or a monad, which makes more sense about the side effect of the main app on the sub app ?). 
```python
from aiohttp import web


class MyApplication(web.Application):

    get_db = None

    def set_db(self, db):
        self.get_db = lambda: db

    def add_subapp(self, prefix, subapp):
        # The db should be a `promise` (like a monad).
        # When the `db` property is called,
        # the promise is then calculated.
        # So, when the db is set, a promise is set,
        # and when it is passed to the subapp, the promise is passed

        async def set_db(app):
            app.get_db = lambda: self.get_db()
        subapp.on_startup.append(set_db)
        #  subapp.get_db = lambda: self.get_db()
        return super().add_subapp(prefix, subapp)

```
While flask provides a `g` so I can share the db between different apps (with in the same request), will you consider adding a same mechanism (or a monadic implementation if your consider about the implicit `side effect` introduced by `g`), since not only the db is shared(more data should be shared?).

asvetlov:
We can organize a kind of chained maps, like https://docs.python.org/3/library/collections.html#collections.ChainMap
The chained mapping should be an explicit request's attribute most likely.

It breaks subapp isolation *by definition* but maybe we can live with it.
Need more thoughts and investigations.

fduxiao:
[...]

Issue #2314, opened 2017-10-11
access subapp routes via <app-name>:<route-name>


samuelcolvin:
From aio-libs/aiohttp-jinja2#163

> I think sub app routes should be available via `subapp:view-name`, which would require a `route_name` argument to `add_subapp`.

This would roughly mirror django routing and would make stuff like the `url()` jinja function work with subapps.

It would also make all `url_for` calls refering to subapps more succinct.

I think @asvetlov has some thoughts on implementation.

asvetlov:
Hmm. I not sure if `subapp:view-name` syntax should be a part of aiohttp or just `aiohttp_jinja2`.
Opinions?

samuelcolvin:
For me this should definitely be part of aiohttp. There are numerous situation other than templates where it can be useful to declare routes via a simple string.

Just to give a few random examples:
* an API framework might have a `URL(route_name)` field which converts a `pk` int to the url for that item
* a menu might be declared via a list of tuples `[('account:details', 'My Account'), ('main:dashbaord', 'Dashboard')]`

and lots of others, these don't work if there's no simple immutable reference for a route.

ansrivas:
I also think this could be a part of `subapp()` implementation. Because if the context is not known,  `aiohttp_jinja2.setup()` will be required to be setup for all the sub apps, like this:
```   
    app.add_subapp(r'/api/v1/todo', todos_app)
    app['todos_app'] = todos_app

    aiohttp_jinja2.setup(todos_app,
                         loader=jinja2.PackageLoader('app', 'templates'))

    app.router.add_static('/static/',
                          path=str(PROJECT_ROOT / 'static'),
                          name='static')```

asvetlov:
Unfortunately url dispatcher doesn't know about applications stack.
If you want to have such functionality in aiohttp you need adding `request.url_for()` method.

Paladiamors:
Quick query on this topic, is there a time line on when having namespaced subapps for the url function might be integrated? Just saw that development has been stopped as the maintainers have been busy for the last while.

I've started using aiohttp as I've moved from django as this tech seems really interesting.

asvetlov:
The issue is not in my personal todo-list (yet)

Paladiamors:
Understood, thanks for letting me know!

orumaxon:
Hi
Please tell me the status of this thread.
I have a similar problem generating urls for sub-applications, so I had to create my own processing as "url_subapp".

[...]

PR #2756, opened 2018-02-24
Implement dynamic prefix for subapps

achimnol:
## What do these changes do?

This allows use of variables (regular expressions) in the subapp prefixes.

Existing codes are not affected.

## Related issue number

I am making this PR after discussion with @asvetlov in gitter.

## Checklist

- [x] I'm already in `CONTRIBUTORS.txt`

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -700,6 +700,64 @@ class PrefixedSubAppResource(PrefixResource):
             prefix=self._prefix, app=self._app)
 
 
+class DynamicSubAppResource(DynamicResource):
+
+    def __init__(self, prefix, app):
+        super().__init__(prefix)
+        self._app = app
+        self._prefix = prefix
+
+    def add_prefix(self, prefix):
+        super().add_prefix(prefix)
+        for resource in self._app.router.resources():
+            resource.add_prefix(prefix)
+
+    def url_for(self, *args, **kwargs):
+        raise RuntimeError(".url_for() is not supported "
+                           "by sub-application root")
+
+    def get_info(self):
+        return {'app': self._app,
+                'prefix': self._prefix}
+
+    def _match(self, path):
+        # match from the beginning of path
+        match = self._pattern.match(path)
+        if match is None:
+            return None, -1
+        else:
+            mdict = {key: URL.build(path=value, encoded=True).path
+                     for key, value in match.groupdict().items()}
+            return mdict, match.end()
+
+    async def resolve(self, request):
+        mdict, mend = self._match(request.url.raw_path)
+        if mdict is None:
+            return None, set()
+        subrequest = request.clone(
+            rel_url=request.url.with_path(request.url.raw_path[mend:]))
+        match_info = await self._app.router.resolve(subrequest)
+        for k, v in mdict.items():
+            if k not in match_info:
+                match_info[k] = v
+        match_info.add_app(self._app)
+        if isinstance(match_info.http_exception, HTTPMethodNotAllowed):
+            methods = match_info.http_exception.allowed_methods
+        else:
+            methods = set()
+        return match_info, methods
+
+    def __len__(self):
+        return len(self._app.router.routes())
+
+    def __iter__(self):
+        return iter(self._app.router.routes())
+
+    def __repr__(self):
+        return "<DynamicSubAppResource {prefix} -> {app!r}>".format(
+            prefix=self._prefix, app=self._app)
+
+
 class ResourceRoute(AbstractRoute):
     """A route with resource"""
 
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -690,11 +690,6 @@ class PrefixedSubAppResource(PrefixResource):
         for subresource in app.router.resources():
             subresource.add_prefix(self)
 
-    def add_prefix(self, resource):
-        super().add_prefix(resource)
-        for subresource in self._app.router.resources():
-            subresource.add_prefix(self)
-
     def get_info(self):
         return {'app': self._app,
                 'prefix': self._prefix}
@@ -732,18 +727,12 @@ class DynamicSubAppResource(DynamicResource):
         assert len(prefix) > 1
         super().__init__(prefix, name=name)
         self._app = app
-        self._prefix = prefix
         for subresource in app.router.resources():
-            subresource.add_prefix(self)
-
-    def add_prefix(self, resource):
-        super().add_prefix(resource)
-        for subresource in self._app.router.resources():
             subresource.add_prefix(self)
 
     def get_info(self):
         info = {'app': self._app,
-                'prefix': self._prefix}
+                'prefix': self._formatter}
         if self._name:
             info['name'] = self._name
         return info
@@ -791,9 +780,9 @@ class DynamicSubAppResource(DynamicResource):
     def __repr__(self):
         if self._name:
             return "<DynamicSubAppResource {name} {prefix} -> {app!r}>".format(
-                name=self._name, prefix=self._prefix, app=self._app)
+                name=self._name, prefix=self._formatter, app=self._app)
         return "<DynamicSubAppResource {prefix} -> {app!r}>".format(
-            prefix=self._prefix, app=self._app)
+            prefix=self._formatter, app=self._app)
 
 
 class ResourceRoute(AbstractRoute):
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -106,7 +106,7 @@ class AbstractResource(Sized, Iterable):
         Return (UrlMappingMatchInfo, allowed_methods) pair."""
 
     @abc.abstractmethod
-    def add_prefix(self, prefix):
+    def add_prefix(self, resource):
         """Add a prefix to processed URLs.
 
         Required for subapplications support.
@@ -365,16 +365,18 @@ class PlainResource(Resource):
         super().__init__(name=name)
         assert not path or path.startswith('/')
         self._path = path
+        self._parent = None
 
     def freeze(self):
         if not self._path:
             self._path = '/'
 
-    def add_prefix(self, prefix):
-        assert prefix.startswith('/')
-        assert not prefix.endswith('/')
-        assert len(prefix) > 1
-        self._path = prefix + self._path
+    def add_prefix(self, resource):
+        self._parent = resource
+        # assert prefix.startswith('/')
+        # assert not prefix.endswith('/')
+        # assert len(prefix) > 1
+        # self._path = prefix + self._path
 
     def _match(self, path):
         # string comparison is about 10 times faster than regexp matching
@@ -389,8 +391,9 @@ class PlainResource(Resource):
     def get_info(self):
         return {'path': self._path}
 
-    def url_for(self):
-        return URL.build(path=self._path, encoded=True)
+    def url_for(self, **kwargs):
+        prefix = self._parent.url_for(**kwargs).path if self._parent else ''
+        return URL.build(path=prefix + self._path, encoded=True)
 
     def __repr__(self):
         name = "'" + self.name + "' " if self.name is not None else ""
@@ -438,13 +441,15 @@ class DynamicResource(Resource):
         assert formatter.startswith('/')
         self._pattern = compiled
         self._formatter = formatter
+        self._parent = None
 
-    def add_prefix(self, prefix):
-        assert prefix.startswith('/')
-        assert not prefix.endswith('/')
-        assert len(prefix) > 1
-        self._pattern = re.compile(re.escape(prefix)+self._pattern.pattern)
-        self._formatter = prefix + self._formatter
+    def add_prefix(self, resource):
+        self._parent = resource
+        # assert prefix.startswith('/')
+        # assert not prefix.endswith('/')
+        # assert len(prefix) > 1
+        # self._pattern = re.compile(re.escape(prefix)+self._pattern.pattern)
+        # self._formatter = prefix + self._formatter
 
     def _match(self, path):
         match = self._pattern.fullmatch(path)
@@ -462,9 +467,10 @@ class DynamicResource(Resource):
                 'pattern': self._pattern}
 
     def url_for(self, **parts):
+        prefix = self._parent.url_for(**parts).path if self._parent else ''
         url = self._formatter.format_map({k: URL.build(path=v).raw_path
                                           for k, v in parts.items()})
-        return URL.build(path=url)
+        return URL.build(path=prefix + url)
 
     def __repr__(self):
         name = "'" + self.name + "' " if self.name is not None else ""
@@ -478,16 +484,23 @@ class PrefixResource(AbstractResource):
         assert not prefix or prefix.startswith('/'), prefix
         assert prefix in ('', '/') or not prefix.endswith('/'), prefix
         super().__init__(name=name)
-        self._prefix = URL.build(path=prefix).raw_path
+        self._parent = None
+        self._prefix = prefix
+        # self._path_prefix = URL.build(path=prefix).raw_path
 
-    def add_prefix(self, prefix):
-        assert prefix.startswith('/')
-        assert not prefix.endswith('/')
-        assert len(prefix) > 1
-        self._prefix = prefix + self._prefix
+    def add_prefix(self, resource):
+        self._parent = resource
+        # assert prefix.startswith('/')
+        # assert not prefix.endswith('/')
+        # assert len(prefix) > 1
+        # self._parent = prefix + self._parent
 
     def raw_match(self, prefix):
         return False
+
+    def url_for(self, **parts):
+        prefix = self._parent.url_for(**parts).path if self._parent else ''
+        return URL.build(path=prefix + self._prefix, encoded=True)
 
     # TODO: impl missing abstract methods
 
@@ -523,7 +536,7 @@ class StaticResource(PrefixResource):
                         'HEAD': ResourceRoute('HEAD', self._handle, self,
                                               expect_handler=expect_handler)}
 
-    def url_for(self, *, filename, append_version=None):
+    def url_for(self, *, filename, append_version=None, **kwargs):
         if append_version is None:
             append_version = self._append_version
         if isinstance(filename, Path):
@@ -533,7 +546,8 @@ class StaticResource(PrefixResource):
         filename = '/' + filename
 
         # filename is not encoded
-        url = URL.build(path=self._prefix + filename)
+        prefix = self._parent.url_for(**kwargs).path if self._parent else ''
+        url = URL.build(path=prefix + self._prefix + filename)
 
         if append_version is True:
             try:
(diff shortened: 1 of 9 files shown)


achimnol:
My implementation approach:
 * Instead of calling `add_prefix()` to all resources in the subapp at the constructor of `DynamicSubAppResource` class, I strip the matched prefix from the request to resolve the subapp's routes, cloning the request with a new URL.
    - This would incur some performance overheads, but I tried to implement it "correctly" first.
    - Existing codes that use plain subapp prefixes won't be affected.
    - User-defined web handlers still receive the original request object, so I belive that user codes should not be affected in any way.
 * Overlapped variable names are overrided by subapps. (e.g., a subapp with a `/{name}.txt` resource with a dynamic prefix `/{name}` will get "def" for the `"name"` key of `match_info` if resolved with "/abc/def.txt")

achimnol:
To allow access to all overriden variables, we could use `ChainMap` / `MultiDict` instead of `dict` for building `UrlMappingMatchInfo` class, but I'm not sure if just enumerating multi-values in the nesting order (like its `_apps` property) would be sufficient or not.
@asvetlov How about your ideas?

achimnol:
Okay, I implemented using ChainMap. :wink:

tests/test_urldispatch.py
```diff
@@ -1003,7 +1005,7 @@ def test_url_for_in_resource_route(router):
 
 def test_subapp_get_info(app, loop):
     subapp = web.Application()
-    resource = subapp.add_subapp('/pre', subapp)
```
asvetlov, line 1006: Good catch!

asvetlov:
@achimnol thanks for PR.
It covers 99% but I see the only one problem: what if `{name}` is present in both dynamic rules: once in subapp prefix itself and other one in routed sub application?

I see the solution:
1. Add `.parent` property to `UrlMappingMatchInfo`. For nested apps it will point on subapp prefix route.
2. Don't derive `UrlMappingMatchInfo` from dict but implement `collections.abc.Mapping` interface and `__getitem__` method. The method can look into `.parent` if the key is absent. 

Right now I don't care about performance, clean API is the first target. Later we can add Cythonized version of `UrlMappingMatchInfo` if needed.

What do you think about?

achimnol:
1. My PR handles such cases using `ChainMap` (see the test case `test_nested_dynamic_subapp_resolution`) -- `.maps` can serve the purpose of `.parent` you suggest, like the `_apps` internal property. I think `.parent` may not be very useful because from the subapp's perspective it is anyway "unknown" how many times it should repeat traversal recursively to get the "right" parent. I thought it is better to expose all maps as-is, like `ChainMap` already does. However, if you prefer `.parent`, I can rewrite it along with 2.

2. The original code was deriving `UrlMappingMatchInfo` from `dict` so I followed that style. If you'd like to change this, I can rewrite it.

achimnol:
To change `UrlMappingMatchInfo` to be a self-nested data structure (e.g., to use `match_info.parent.parent`), I think we need to rewrite its `apps` property handling as well because it also treats the list of nested apps as its own property rather than having a single reference to the current app -- so that the user code becomes `match_info.current_app` / `match_info.parent.current_app`.

Also I still don't get the benefit of having multiple per-app-level `match_info` objects for a single resolution operation with duplicate references to routes for each of them since sub-app resources are not treated as final destinations.

I think it is sufficient to have nested maps via ChainMap just like `apps` property. 😕

asvetlov:
I've convinced. Let's use chained map approach.
But please wait for my careful review.

aiohttp/web_urldispatcher.py
```diff
@@ -172,12 +172,13 @@ def url_for(self, *args, **kwargs):
         return await self._expect_handler(request)
 
 
-class UrlMappingMatchInfo(dict, AbstractMatchInfo):
+class UrlMappingMatchInfo(AbstractMatchInfo, Mapping):
```
asvetlov, line 200: I really like the fact that `UrlMappingMatchInfo` is immutable now.

aiohttp/web_urldispatcher.py
```diff
@@ -172,12 +172,13 @@ def url_for(self, *args, **kwargs):
...
 
     def __init__(self, match_dict, route):
-        super().__init__(match_dict)
+        super().__init__()
```
asvetlov: `super()` call is not needed, Python machinery will work fine without it.

aiohttp/web_urldispatcher.py
```diff
@@ -230,8 +231,24 @@ def set_current_app(self, app):
     def freeze(self):
         self._frozen = True
 
+    @property
+    def variable_maps(self):
+        return tuple(reversed(self._variables.maps))
+
+    def add_variables(self, match_dict):
+        self._variables.maps.append(match_dict)
```
asvetlov: I suspect you should insert a new dict *in front* of maps -- like we do it for `add_app()`.
Moreover maybe merging `add_app()` and `add_variables()` functionality makes sense.
What do you think about `_add_app(app, match_dict)` method as replacement for `add_app`.
Adding underscore emphasizes that the method is private. `add_app()` was never a part of documented public API, I think the change is safe in backward compatibility perspective.

aiohttp/web_urldispatcher.py
```diff
@@ -230,8 +231,24 @@ def set_current_app(self, app):
...
+
+    def __iter__(self):
+        return self._variables.__iter__()
+
+    def __len__(self):
+        return self._variables.__len__()
+
     def __repr__(self):
-        return "<MatchInfo {}: {}>".format(super().__repr__(), self._route)
+        return "<MatchInfo {}: {}>".format(dict(self).__repr__(), self._route)
```
asvetlov: `repr(self._variables)` instead of `repr(dict(self))` is better I believe. There is no reason to hide chained variables.

aiohttp/web_urldispatcher.py
```diff
@@ -230,8 +231,24 @@ def set_current_app(self, app):
     def freeze(self):
         self._frozen = True
 
+    @property
+    def variable_maps(self):
```
asvetlov: Maybe let's use just `maps` name?
`maps` is shorted, in documentation we could refer to `ChainMap.maps`. `maps` is mentally more familiar to people if they know about `ChainMap` already.

docs/web_reference.rst
```diff
@@ -2181,8 +2189,8 @@ In general the result may be any object derived from
 
 .. class:: UrlMappingMatchInfo
 
-   Inherited from :class:`dict` and :class:`AbstractMatchInfo`. Dict
-   items are filled by matching info and is :term:`resource`\-specific.
+   Inherited from :class:`collections.ChainMap` and :class:`AbstractMatchInfo`.
```
asvetlov: The doc is not relevant anymore, `Mapping` should replace `ChainMap`.
Also `.variable_maps` or `.maps` attribute should be documented.

asvetlov (commented):
Sorry for delay and thank you for patience.

I have some comments.
Let's discuss my proposals.

aiohttp/web_urldispatcher.py
achimnol: This was to make existing test cases that compares the `repr` result with a constant string to pass. Then I will change those test cases to match with `ChainMap`'s `repr` result.

aiohttp/web_urldispatcher.py
achimnol: `ChainMap` resolves from the beginning of its `maps`. The point here is that `add_variables()` is called *after* resolving the subapp's routes, which means that subapp's match variables are added to the chain map first. Since we want to give priority to the innermost subapp's variables in the merged view of this chain map, this `append` is natural.

I think merging this method into `add_app(app, match_dict)` is a good idea. :+1:

aiohttp/web_urldispatcher.py
achimnol: This is also related to the fact that resolution order of the chain map and its build order are reversed. The use of `reversed()` here is to be consistent with the order of the return value of `apps` property.
Also I wanted to make it read-only since `ChainMap`'s `maps` property is mutable -- so I've used a different name because they are not the same. Still, I'd like to change the name to `maps` if you still want it. :) If so, I will update the docs as well.

aiohttp/web_urldispatcher.py
achimnol: Hm.... merging into a single method causes a lot of problem:
* When instantiating `UrlMappingMatchInfo` we already have `match_dict` (so the chain map begins with a single layer) but `_app` is None and `_apps` is an empty tuple. This makes the number of nested maps and apps inconsistent.
* Changing `UrlMappingMatchInfo`'s constructor to accept an explicit `app` argument requires `Route` and `Resource` classes to keep track of their owner `app` because `resolve()` method should be able to indicate the exact `app` instance to create the match info object. `request.app` is a self null reference there.
* `add_app()` is called at many other places such as `test_utils.py` and `web_app.py` where `match_dict` is out-of-scope.

The root cause of this mess is that `app` and `match_dict` are added to `UrlMappingMatchInfo` at different places and timings. :confused:

aiohttp/web_urldispatcher.py
```diff
@@ -227,11 +227,29 @@ def set_current_app(self, app):
         finally:
             self._current_app = prev
 
+    def add_map(self, match_dict):
+        if self._frozen:
+            raise RuntimeError("Cannot change maps stack after .freeze() call")
+        self._variables.maps.append(match_dict)
```
asvetlov: Let's consider the following:
```
async def handler(request):
    return web.Response(text=request.math_info['name'])

subapp = web.Application()
subapp.add_routes([web.get('/{name}', handler)])

app = web.Application()
app.add_subapp('/{name}', subapp)
```
What `name` should be returned by the handler?

aiohttp/web_urldispatcher.py
```diff
@@ -674,6 +693,57 @@ def __repr__(self):
...
 
 
+class DynamicSubAppResource(DynamicResource):
+
+    def __init__(self, prefix, app):
+        super().__init__(prefix)
+        self._app = app
+        self._prefix = prefix
+
+    def url_for(self, *args, **kwargs):
```
asvetlov: I see more complex problem.
`.url_for()` for all resources from sub-application are broken for dynamic prefixed resource.

Sorry, but the problem is blocker for merging.

aiohttp/web_urldispatcher.py
achimnol: In this case, `request.match_info['name']` in the handler will return the matched value of the innermost `{name}` definition: the route one.

aiohttp/web_urldispatcher.py
achimnol: Yes, that's what I mentioned in the gitter channel.
I think rewriting `url_for()` to have a recursive architecture so that route resources of different types in different subapp levels could be chained requires a lot of work, so I'd like to suggest it as a separate PR.
Maybe we could mark this PR as "provisional" if merged before it.

achimnol: I'm currently very busy due to my business work -- once I get some time after a few weeks, I'll give a look on the extension plan for `url_for()`.

aiohttp/web_urldispatcher.py
asvetlov: Perfect!

aiohttp/web_urldispatcher.py
asvetlov: The problem is that without `url_for` fix not only `url_for()` for subapp itself doesn't work but all `url_for()` calls for subapp resources are broken.
That's why I think everything should be done in single PR.

achimnol:
Implemented support for `url_for()`.
There is an issue to decide on: unlike the chained `match_info`, the `**kwargs` passed to `url_for()` does not distinguish which key is for which subapp prefix.

* Let it just do what it does now. Existing code should work as-is.
* Add a mandatory property `name` to `PrefixedSubAppResource` and `DynamicSubAppResource`.
  Let users prefix specific keys with those names, e.g., "myprefix.version" and "version" will treated differently. In this case, the chained `match_dict` may be also replaced with a flat dict with prefixed keys. But this requires a little change to existing codes that use subapps.

@asvetlov What do you think?

achimnol:
Ah, since dynamic prefix is not a released API yet, I think it is safe to make the proposed name-and-predixed-keys scheme *optional*.
If the user wants to avoid key collision then they can use this optional name prefixes, but when not, they can just use as-is. No breaking for existing codes. Yay.

aiohttp/web_urldispatcher.py
```diff
@@ -637,27 +669,26 @@ class PrefixedSubAppResource(PrefixResource):
...
+        for subresource in self._app.router.resources():
+            subresource.add_prefix(resource)
 
     def get_info(self):
         return {'app': self._app,
                 'prefix': self._prefix}
 
     async def resolve(self, request):
-        if not request.url.raw_path.startswith(self._prefix):
+        if not request.url.path.startswith(self._prefix):
```
asvetlov: Why not `raw_path`?

achimnol: What if the prefix includes arbitrary unicode characters?

achimnol:
Now if we have a nested dynamic subapp chain:
```python
root_app = web.Application()
outer_app = web.Application()
inner_app = web.Application()
resource = inner_app.router.add_resource('/{var}')
route = resource.add_route('GET', ...)
outer_app.add_subapp('/{var}', inner_app, name='b')
root_app.add_subapp('/{var}', outer_app, name='a')
```

Then `match_dict` after resolving `"/abc/def/ghi"` has the content:
```python
{
  'a.var': 'abc',
  'b.var': 'def',
  'var': 'ghi',
}
```

`route.url_for(**{'a.var': '123', 'b.var': '456', 'var': '789'})` gives:
```python
'/123/456/789'
```

When `inner_app` and `outer_app` has beend added *without* names,
the match result contains only the innermost variable:
```python
{
  'var': 'ghi'
}
```
and `route.url_for(var='xxx')` gives:
```python
'/xxx/xxx/xxx'
```

No more ChainMap!

achimnol:
Rebased against latest master.

aiohttp/web_urldispatcher.py
asvetlov: Mixing PCT encoded and decoded data can make a mess easy.
I believe we should follow the strategy:
1. Convert all user provided *path*, *prefix* etc into *normalized* (PCT-encoded actually) form by using `yarl`. Already normalized paths passed *as-is* by yarl.
2. Normalize given HTTP path (we do it).
3. Match normalized path to already pre-normalized patterns.
4. Return `request.match_info['name']` in **PCT-decoded** form because users most likely want getting `이름` instead of `%EC%9D%B4%EB%A6%84`

If aiohttp code violates the rule somewhere -- it should be fixed.
Do you agree?

achimnol:
I think it's ready to get reviewed. :)

aiohttp/web_urldispatcher.py
achimnol: Okay, I agree to keep the rule consistent. Let me inspect and fix up this PR.

achimnol: I think then `PlainResource.__init__()` also need this, like `PrefixResource`:
```python
self._path = URL.build(path=path).raw_path
```

asvetlov:
Let me meditate for a while. 
``"name."`` looks... strange. How to manage a situation if `name` argument for `app.add_subapp` was not passed? 
Another thing worries me is: suppose I have a common-usable application like admin or debug toolbar.
I'm adding it with variable prefix and the subapp becomes broken: it has a knowledge about its required parents for generating urls. 
The worse thing is that breakage appears on subapp usage (url generations for filing HTTP templates), not on subapp adding.

If you have ideas how to make it steady -- please share your vision. 
We a free to invite a new API or even restrict subapp type to `VariablePrefixAwareSubapp` inherited from `web.Application` but the game rules should be discussed and fixed in stone before the PR merging.

I understand your striving to make subapp prefix flexible but my guts feel that we should be very careful with chosen design. Otherwise fitting a opened can of worms will be painful.

I suggest excluding the PR from aiohttp 3.1 but working hard to publish it as part of aiohttp 3.2
Your idea for chaining `match_info` is brilliant.
There is a very visible request for providing chained info for nested applications: 
- #2689 for accessing parent configuration like `request.app['db']`
- #2314 for building URLs from sibling sub-apps
We should make a uniform approach for all of these.
 
P.S.
I'll go to Ho Chi Minh tomorrow evening for a couple weeks. Not sure if I will have enough time to work on aiohttp these dates.
The city is much closer to Seoul than Kiev but unfortunately still so far for meeting with you offline :(

achimnol:
I see. You want more isolation between apps -- and I agree with that.
I thought having explicit names was better than nested scopes because it is clear to read.

I'm fine to go back to the first ChainMap design for `match_info`, keeping my "recursive" implementation.
The problem is `url_for()`.

To allow use `url_for()` in the innermost ("leaf") app *without* knowing whatever variables are included in its outer subapp prefixes, then I think we need a strong assumption that `url_for()` is only used inside request handlers where `match_info` is available in the current task-local context.
Once `url_for()` has access to the nested `match_info`'s ChainMap then it can fill the prefix variables using the matched values of the current request's path.

Still we have two issues:

1. Python asyncio currently lacks task-local context support which automatically propagates to subtasks. To workaround this, it may be necessary to change the `url_for()` API to accept the `request` object (since it serves the role of handler-local context now) or its `match_info` attribute. But this would break use of `url_for()` in template rendering. #2314's `request.url_for()` may be an option here.
2. `url_for()` may be used to refer URLs with prefixes other than the current request's path has. In this case users need to provide the prefix variables anyway, like my "name." design (of course the concrete method may be different).

@asvetlov What do you think?

achimnol:
More thoughts:
~Maybe could we ingore the 2nd case if we stick with "isolated" apps?
If each app never calls `url_for()` against routes in other apps, then filling the prefix variables with the current request's path should be ok. But in the reality, like in one of my Django projects, cross-reference to URLs in other apps in one app's template rendering often happens because a single rendered view of webpage usually involves multiple apps (e.g., dashboard, user authentication, posts, comments, etc.). Nonetheless it is also possible to argue that such cases must be implemented as a single monolithic app — there is no reason to match the semantic of aiohttp's app with Django's app.~
See below!

achimnol:
More more thoughts:
When composing multiple apps, cross-reference to URLs in other apps usually happen in a specific direction only: from outer apps to inner apps, not to parent/sibling apps.

From the perspective of inner apps, it's safe to assume that the aiohttp's web framework will take care of filling outer prefix variables with the handler-local match_info context.

From the perspective of outer apps, they need to explicitly give the value of prefix variables when calling `url_for()` of routes of inner apps. In this specific case, we need a way to distinguish the duplicate variable names in the prefix names.

asvetlov:
Maybe name prefixes are not bad but I need a time to experiment with.
Regarding to mandatory params from outer apps etc:
1. We can invite a new API for it, something like `request.build_url()` (the name is a subject for changing)
2. Context local variables (implicit context) are not required as we always have *explicit context* as `request`.
3. Inner/outer applications can be useful terminology.

Dreamsorcerer:
@achimnol Do we want to revive this PR? I suspect it'll need some adjustments since we overhauled the routing algorithm sometime ago.

PR #2759, opened 2018-02-26
Fix MultipartWriter.append* no longer returning part/payload.

terencehonles:
## What do these changes do?

- Fixes commit: caa6bdb2484bf821b52a65322efa98a889b593de

  MultipartWriter.append methods used to return the part appended to the writer so one could set the content_disposition, etc. This patch restores that functionality so the code matches the documentation in multipart.rst

- This patch also makes append_json use the JsonPayload object instead of duplicating functionality.

Fix regression

## Checklist

### DIFF
--- a/CHANGES/2759.bugfix
+++ b/CHANGES/2759.bugfix
@@ -1,0 +1,1 @@
+Fix MultipartWriter.append* no longer returning part/payload.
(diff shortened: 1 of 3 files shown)


asvetlov:
Thanks!

Issue #2698, opened 2018-01-30
Change flow control for writers


asvetlov:
Since `writer.write()` is a coroutine -- no need for internal writer buffer at all.
On socket's internal send buffer overflow the writer should be paused without trying to fill application level transport buffer -- it makes no sense but wastes extra memory and CPU cycles.

fafhrd91:
you should run some benchmarks before 3.0 release

fafhrd91:
btw raw results for round 15 techempower benchmarks are ready https://tfb-status.techempower.com/

fafhrd91:
aiohttp consistently outperform sanic :)

despite fact that sanic is super fast

asvetlov:
@fafhrd91 yes, sure.
As side effect of enforcing write to bufferless coroutine I expect getting rid of reported aiohttp problems like #2669

asvetlov:
@fafhrd91 is there a page with human readable results?
My brain does not parse JSON easy.

fafhrd91:
write buffer can help with performance.

what I do in actix is write everything to buffer and flush this buffer only once per event loop iteration. so for example if you have multiple request in processing queue, you can write all response once before pushing data to a socket. only with this optimization actix can process 1m requests a second

fafhrd91:
human readable results are not ready yet. you should open json file in Firefox, it shows json nicely, at least nightly Firefox.

fafhrd91:
@asvetlov here is script

```python
import json

#data = open("results.2018-01-27-17-41-53-405.json", 'r').read()
#data = open("results.2018-01-27-17-41-53-405.json", 'r').read()
data = open("results.2018-01-18-09-55-11-468.json", 'r').read()
data = json.loads(data)
raw_data = data['rawData']

res = {}
for key in ['fortune', 'plaintext', 'db', 'update', 'json', 'query']:
    results = []
    d = raw_data[key]
    for name, val in d.items():
        #if len(val) < 5:
        #    continue
        info = val[0]
        s, e = info['startTime'], info['endTime']
        if 'totalRequests' not in info:
            continue

        reqs = info['totalRequests']
        persec = reqs / (e-s)
        results.append((persec, name, reqs))

    results.sort()
    res[key] = [(name, r1, r2) for (r1, name, r2) in list(reversed(sorted(results)))]

print(json.dumps(res))
```

pfreixes:
[...]

PR #2762, opened 2018-02-26
Async write headers

asvetlov:
Make `writer.write_headers()` a coroutine

### DIFF
--- a/CHANGES/2762.feature
+++ b/CHANGES/2762.feature
@@ -1,0 +1,1 @@
+Make ``writer.write_headers`` a coroutine.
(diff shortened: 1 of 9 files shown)


kxepal:
LGFM, but could you remind me, please, why all this need? What's the goal?

webknjaz:
I guess it's related to https://github.com/aio-libs/aiohttp/pull/2743#issuecomment-368309217, right?

asvetlov:
No, it is for #2698

asvetlov:
After switching all writes to coroutines we can drop write buffers.

kxepal:
Cool! Thanks.

Issue #2755, opened 2018-02-23
Client request tracing. Allow to get the request body in the trace.


kowalski:
## Long story short

I would like to get the `request_body` in the tracing events. Obviously it cannot be included in `on_request_start` because body can come from generator. 

I would be happy to contribute another event, say `on_body_sent` which would fire after the body is written after the body is finished.

pfreixes:
Please go ahead!

kowalski:
Ok, great! Thanks for encouragement. 

Before I get into coding I would like to discuss an idea around this. 
The barest possible idea is to simply save the request body bytes into a variable and pass it to a `Signal` after everything is written. Unfortunately I can see how in certain use cases this could have catastrophic performance and memory footprint. 
So instead I would like to propose, that I will move registering for grabbing a request/response body into a method on:

```
class TraceConfigContext(SimpleNamespace):

    def register_content_handlers(self, request_stream, response_stream):
          ....
```

The idea would be that this method is called from `on_request_start` callback and the results are gathered in `on_request_end`. 

Also this means, that if user code chooses not to read the response body it will not be available for tracing. I mean getting 400 status when the code looks like:
```
async with session.get(url) as response:
    if response.status == 400:
        raise SomeError
    # here the normal processing follows, call response.json() and do something with in
```
What do you think ?

pfreixes:
Umm ... once the strings are allocated in memory I would say that internally the only thing that is given as a parameter is just a String object that points to that region. So, passing the request or the response as a parameter won't affect the memory, and only if the signal handle mutates the string a new copy will be made.

am I missing something?

kowalski:
Sorry for not being clear.
I meant a scenario when request body is not passed as string, but as for example `aiohttp.payload.BufferedReaderPayload` pointing to a file which will not fit in memory.

asvetlov:
We should call signal as many times as we call `transport.write`. In this case everything should fit in memory.

pfreixes:
We might thing on rename the signals as `on_request_chunk_sent` and `on_response_chunk_received`, if chunk is used this function will be called many times if not will be called once.

asvetlov:
Fixed by #2767

PR #2767, opened 2018-02-26
Add tracking signals for getting request/response bodies.

kowalski:
This is PR for issue: https://github.com/aio-libs/aiohttp/issues/2755

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -443,7 +443,7 @@ class ClientRequest:
                     self.body = (self.body,)
 
                 for chunk in self.body:
-                    writer.write(chunk)
+                    await writer.write(chunk)
 
             await writer.write_eof()
         except OSError as exc:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -22,6 +22,7 @@ from .formdata import FormData
 from .helpers import PY_36, HeadersMixin, TimerNoop, noop, reify, set_result
 from .http import SERVER_SOFTWARE, HttpVersion10, HttpVersion11, StreamWriter
 from .log import client_logger
+from .signals import Signal
 from .streams import StreamReader
 
 
@@ -168,7 +169,8 @@ class ClientRequest:
                  proxy=None, proxy_auth=None,
                  timer=None, session=None, auto_decompress=True,
                  ssl=None,
-                 proxy_headers=None):
+                 proxy_headers=None,
+                 traces=[]):
 
         if loop is None:
             loop = asyncio.get_event_loop()
@@ -209,6 +211,7 @@ class ClientRequest:
         if data or self.method not in self.GET_METHODS:
             self.update_transfer_encoding()
         self.update_expect_continue(expect100)
+        self.traces = traces
 
     def is_ssl(self):
         return self.url.scheme in ('https', 'wss')
@@ -475,7 +478,13 @@ class ClientRequest:
             if self.url.raw_query_string:
                 path += '?' + self.url.raw_query_string
 
-        writer = StreamWriter(conn.protocol, conn.transport, self.loop)
+        async def on_chunk_sent(chunk):
+            for trace in self.traces:
+                await trace.send_request_chunk_sent(chunk)
+
+        writer = StreamWriter(
+            conn.protocol, conn.transport, self.loop)
+        writer.on_chunk_sent.append(on_chunk_sent)
 
         if self.compress:
             writer.enable_compression(self.compress)
@@ -509,11 +518,16 @@ class ClientRequest:
 
         self._writer = self.loop.create_task(self.write_bytes(writer, conn))
 
+        async def on_chunk_received(chunk):
+            for trace in self.traces:
+                await trace.send_response_chunk_received(chunk)
+
         self.response = self.response_class(
             self.method, self.original_url,
             writer=self._writer, continue100=self._continue, timer=self._timer,
             request_info=self.request_info,
             auto_decompress=self._auto_decompress)
+        self.response.on_chunk_received.append(on_chunk_received)
 
         self.response._post_init(self.loop, self._session)
         return self.response
@@ -572,6 +586,9 @@ class ClientResponse(HeadersMixin):
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
+
+        # avoid circular reference so that __del__ works
+        self.on_chunk_received = Signal(owner=None)
 
     @property
     def url(self):
@@ -796,6 +813,8 @@ class ClientResponse(HeadersMixin):
         if self._content is None:
             try:
                 self._content = await self.content.read()
+                self.on_chunk_received.freeze()
+                await self.on_chunk_received.send(self._content)
             except BaseException:
                 self.close()
                 raise
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -477,13 +477,9 @@ class ClientRequest:
             if self.url.raw_query_string:
                 path += '?' + self.url.raw_query_string
 
-        async def on_chunk_sent(chunk):
-            for trace in self.traces:
-                await trace.send_request_chunk_sent(chunk)
-
         writer = StreamWriter(
             conn.protocol, conn.transport, self.loop,
-            on_chunk_sent=on_chunk_sent
+            on_chunk_sent=self._on_chunk_request_sent
         )
 
         if self.compress:
@@ -540,6 +536,10 @@ class ClientRequest:
             if not self.loop.is_closed():
                 self._writer.cancel()
             self._writer = None
+
+    async def _on_chunk_request_sent(self, chunk):
+        for trace in self.traces:
+            await trace.send_request_chunk_sent(chunk)
 
 
 class ClientResponse(HeadersMixin):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -210,7 +210,9 @@ class ClientRequest:
         if data or self.method not in self.GET_METHODS:
             self.update_transfer_encoding()
         self.update_expect_continue(expect100)
-        self._traces = traces or []
+        if traces is None:
+            traces = []
+        self._traces = traces
 
     def is_ssl(self):
         return self.url.scheme in ('https', 'wss')
@@ -583,7 +585,9 @@ class ClientResponse(HeadersMixin):
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
-        self._traces = traces or []
+        if traces is None:
+            traces = []
+        self._traces = traces
 
     @property
     def url(self):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -169,7 +169,7 @@ class ClientRequest:
                  timer=None, session=None, auto_decompress=True,
                  ssl=None,
                  proxy_headers=None,
-                 traces=[]):
+                 traces=None):
 
         if loop is None:
             loop = asyncio.get_event_loop()
@@ -210,7 +210,7 @@ class ClientRequest:
         if data or self.method not in self.GET_METHODS:
             self.update_transfer_encoding()
         self.update_expect_continue(expect100)
-        self.traces = traces
+        self._traces = traces or []
 
     def is_ssl(self):
         return self.url.scheme in ('https', 'wss')
@@ -519,7 +519,7 @@ class ClientRequest:
             writer=self._writer, continue100=self._continue, timer=self._timer,
             request_info=self.request_info,
             auto_decompress=self._auto_decompress,
-            traces=self.traces,
+            traces=self._traces,
         )
         self.response._post_init(self.loop, self._session)
         return self.response
@@ -538,7 +538,7 @@ class ClientRequest:
             self._writer = None
 
     async def _on_chunk_request_sent(self, chunk):
-        for trace in self.traces:
+        for trace in self._traces:
             await trace.send_request_chunk_sent(chunk)
 
 
@@ -566,7 +566,7 @@ class ClientResponse(HeadersMixin):
     def __init__(self, method, url, *,
                  writer=None, continue100=None, timer=None,
                  request_info=None, auto_decompress=True,
-                 traces=[]):
+                 traces=None):
         assert isinstance(url, URL)
 
         self.method = method
@@ -583,7 +583,7 @@ class ClientResponse(HeadersMixin):
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
-        self._traces = traces
+        self._traces = traces or []
 
     @property
     def url(self):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -22,7 +22,6 @@ from .formdata import FormData
 from .helpers import PY_36, HeadersMixin, TimerNoop, noop, reify, set_result
 from .http import SERVER_SOFTWARE, HttpVersion10, HttpVersion11, StreamWriter
 from .log import client_logger
-from .signals import Signal
 from .streams import StreamReader
 
 
@@ -483,8 +482,9 @@ class ClientRequest:
                 await trace.send_request_chunk_sent(chunk)
 
         writer = StreamWriter(
-            conn.protocol, conn.transport, self.loop)
-        writer.on_chunk_sent.append(on_chunk_sent)
+            conn.protocol, conn.transport, self.loop,
+            on_chunk_sent=on_chunk_sent
+        )
 
         if self.compress:
             writer.enable_compression(self.compress)
@@ -518,17 +518,13 @@ class ClientRequest:
 
         self._writer = self.loop.create_task(self.write_bytes(writer, conn))
 
-        async def on_chunk_received(chunk):
-            for trace in self.traces:
-                await trace.send_response_chunk_received(chunk)
-
         self.response = self.response_class(
             self.method, self.original_url,
             writer=self._writer, continue100=self._continue, timer=self._timer,
             request_info=self.request_info,
-            auto_decompress=self._auto_decompress)
-        self.response.on_chunk_received.append(on_chunk_received)
-
+            auto_decompress=self._auto_decompress,
+            traces=self.traces,
+        )
         self.response._post_init(self.loop, self._session)
         return self.response
 
@@ -569,7 +565,8 @@ class ClientResponse(HeadersMixin):
 
     def __init__(self, method, url, *,
                  writer=None, continue100=None, timer=None,
-                 request_info=None, auto_decompress=True):
+                 request_info=None, auto_decompress=True,
+                 traces=[]):
         assert isinstance(url, URL)
 
         self.method = method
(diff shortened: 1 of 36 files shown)


aiohttp/http_writer.py
```diff
@@ -56,13 +59,18 @@ def _write(self, chunk):
...
 
-    def write(self, chunk, *, drain=True, LIMIT=64*1024):
+    def write(self, chunk, *, drain=True, LIMIT=64 * 1024):
         """Writes chunk of data to a stream.
 
         write_eof() indicates end of stream.
         writer can't be used after write_eof() method being called.
         write() return drain future.
         """
+        self.on_chunk_sent.freeze()
```
asvetlov: I pretty sure there is better place for signal freezing than first write attempt.

aiohttp/http_writer.py
```diff
@@ -56,13 +59,18 @@ def _write(self, chunk):
             raise asyncio.CancelledError('Cannot write to closing transport')
         self._transport.write(chunk)
 
-    def write(self, chunk, *, drain=True, LIMIT=64*1024):
+    def write(self, chunk, *, drain=True, LIMIT=64 * 1024):
```
asvetlov: Don't touch if you don't change the logic.
If you want update code style -- do it in a separate PR.
For this particular line I pretty happy with status quo.

aiohttp/http_writer.py
```diff
@@ -56,13 +59,18 @@ def _write(self, chunk):
...
         """Writes chunk of data to a stream.
 
         write_eof() indicates end of stream.
         writer can't be used after write_eof() method being called.
         write() return drain future.
         """
+        self.on_chunk_sent.freeze()
+        self.loop.create_task(
```
asvetlov: Spawning a new task is not what we need.
Technically `.write` is a coroutine, we need convert it to genuine async function.
Maybe I'll do it very soon in separate PR.

aiohttp/client_reqrep.py
```diff
@@ -573,6 +587,9 @@ def __init__(self, method, url, *,
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
 
+        # avoid circular reference so that __del__ works
+        self.on_chunk_received = Signal(owner=None)
```
asvetlov: Not sure if we need a signal here.
It looks like signal concept overusage: you are creating `on_chunk_received` signal with the only usage by aiohttp internals itself. It is not very effective, signal subscription is literally a waste of time.
I suggest passing `traces` directly to response object.

aiohttp/http_writer.py
```diff
@@ -31,6 +32,8 @@ def __init__(self, protocol, transport, loop):
         self._compress = None
         self._drain_waiter = None
 
+        self.on_chunk_sent = Signal(self)
```
asvetlov: Again pass traces instead of creating a new internal-only signal.

asvetlov:
Documentation should be updated as well

aiohttp/http_writer.py
kowalski: That's wasn't me, it's my emacs ;) I will revert this change

aiohttp/http_writer.py
kowalski: Well, please consider that `StreamWriter` is used both by the http client and server. If I pass the `traces` in `__init__()` I will have to make something like:
```
def write( ):
    for trace in self.traces:
        await trace.sent_request_chunk_sent(...)
    ...
```
Now if you consider that code in the context of usage by http server, it clearly stops making sense....
This is why I introduced a signal and names it `on_chunk_sent`. Without specifying if it's http request it's sending or http response (as a server).

aiohttp/client_reqrep.py
kowalski: Here the approach with passing `traces` will actually work quite well. I've introduced the signal to make it symetrical to the situation in request. I will update this to pass `traces` and remove the signal

aiohttp/http_writer.py
kowalski: Are you sure we don't want this in another task ? Whatever handler does asynchronously will slow down writing the response. 

Also `.write` isn't actually coroutine, because it's not called like:
```
                for chunk in self.body:
                    writer.write(chunk)
```
(without `await`) The couroutine returned is never triggered.
I guess I would have to track all the places there `StreamWriter.write` is used and add `await` in there. That's not undoable, but it's rather big refactor.

aiohttp/http_writer.py
kowalski: Actually I was quite surprised with the whole concept of `freeze()`. It's very tricky! Do you know why it's necessary at all ? 
The only reason I can think of is that the author wanted to prevent modifications to list of handlers while processing `send()`. 

Still, it's possible that we will decide to remove this `Signal` altogether, but if we don't, what do you think is the correct place to call freeze() ? 
- I can't do this in `__init__` because I want to add handlers later
- I can't delegate this responsibility to the user of the class, because this introduces a braking change where a code which works now like:
  ```
  writer = StreamWriter(...)
  await writer.write(...)
  ```
  will start raising `RuntimeError`

aiohttp/http_writer.py
asvetlov: Let's avoid signals for internal-only usage.
Here we have two options: pass a callback for sending signals or derive a `ClientStreamWriter` class. Personally I prefer the second approach.

aiohttp/http_writer.py
asvetlov: Yes, I'm sure. Spawning a task without waiting for result has an ugly smell.
More important spawning a task may reorder received signals, e.g. user may get *request chunk* after *request finished*.

`.write` should be async method. I'm working on this refactoring (required for #2698 anyway).
I suggest waiting for my job done.

aiohttp/http_writer.py
kowalski: I will remove the `Signal` in favor of accepting an optional callable to be used as a callback. 
I prefer not to create a child class for this - too many layers of inheritance hurts code readability.

aiohttp/http_writer.py
kowalski: Got it, I will wait until the refactoring is merged.

aiohttp/http_writer.py
kowalski: Fixed

kowalski: Also I fixed `setup.cfg` so that this line is not altered by anyone else who uses `autopep8`

aiohttp/http_writer.py
asvetlov: Ok

aiohttp/http_writer.py
asvetlov: Cool

asvetlov:
@kowalski #2774 is done, `write()` is async now

kowalski:
@asvetlov all fixes done, including adding documentation. 
I'm not squashing commits so that you can see what changes. I can squash changes after review  to keep history clean.

aiohttp/client_reqrep.py
```diff
@@ -475,7 +477,14 @@ def keep_alive(self):
             if self.url.raw_query_string:
                 path += '?' + self.url.raw_query_string
 
-        writer = StreamWriter(conn.protocol, conn.transport, self.loop)
+        async def on_chunk_sent(chunk):
```
asvetlov: Convert the function into ClientRequest's method. No need to create a nested function on every HTTP request.

aiohttp/http_writer.py
```diff
@@ -62,6 +70,9 @@ def _write(self, chunk):
         writer can't be used after write_eof() method being called.
         write() return drain future.
         """
+        if self._on_chunk_sent:
```
asvetlov: Please do check `self._on_chunk_sent is not None`. It is more idiomatic.

aiohttp/tracing.py
```diff
@@ -9,13 +9,14 @@
...
 
 __all__ = (
-    'TraceConfig', 'TraceRequestStartParams', 'TraceRequestEndParams',
-    'TraceRequestExceptionParams', 'TraceConnectionQueuedStartParams',
-    'TraceConnectionQueuedEndParams', 'TraceConnectionCreateStartParams',
-    'TraceConnectionCreateEndParams', 'TraceConnectionReuseconnParams',
-    'TraceDnsResolveHostStartParams', 'TraceDnsResolveHostEndParams',
-    'TraceDnsCacheHitParams', 'TraceDnsCacheMissParams',
-    'TraceRequestRedirectParams'
+    'TraceConfig', 'TraceRequestStartParams',
```
asvetlov: Do you really need to touch all these lines?
Appending two classes to the end of sequence should be enough.

docs/tracing_reference.rst
```diff
@@ -147,6 +147,20 @@ TraceConfig
 
       ``params`` is :class:`aiohttp.TraceRequestStartParams` instance.
 
+   .. attribute:: on_request_chunk_sent
```
asvetlov, line 160: Please also upgrade flow diagrams at the beginning of file.

asvetlov:
Not squashing is fine, I'll do it on merging anyway. Well, github button will do actually.

aiohttp/client_reqrep.py
```diff
@@ -168,7 +168,8 @@ def __init__(self, method, url, *,
                  proxy=None, proxy_auth=None,
                  timer=None, session=None, auto_decompress=True,
                  ssl=None,
-                 proxy_headers=None):
+                 proxy_headers=None,
+                 traces=[]):
```
kxepal: Let's not use mutables as defaults. These may cause some nasty bugs we could easily avoid.

aiohttp/client_reqrep.py
```diff
@@ -572,6 +583,7 @@ def __init__(self, method, url, *,
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
+        self._traces = traces
```
kxepal: Why in ClientRequest traces is public attribute while in ClientResponse it's private?

aiohttp/client_reqrep.py
kowalski: done

aiohttp/client_reqrep.py
```diff
@@ -555,7 +565,8 @@ class ClientResponse(HeadersMixin):
 
     def __init__(self, method, url, *,
                  writer=None, continue100=None, timer=None,
-                 request_info=None, auto_decompress=True):
+                 request_info=None, auto_decompress=True,
+                 traces=[]):
```
kxepal: Same defaults story.

aiohttp/http_writer.py
kowalski: changed

aiohttp/tracing.py
kowalski: Ok, reverted and added new classes at the end

docs/tracing_reference.rst
kowalski: done. Here I added new signals to description. Please let me know if you meant something more elaborate.
zrzut ekranu z 2018-02-27 15-04-09

aiohttp/client_reqrep.py
asvetlov: I can answer: `ClientResponse` is a public class visible by user.
The attr should be private.
About request I don't care.

kxepal: Ok, fine for me.

docs/tracing_reference.rst
```diff
@@ -34,8 +34,8 @@ Overview
      exception[shape=flowchart.terminator, description="on_request_exception"];
 
      acquire_connection[description="Connection acquiring"];
-     got_response;
-     send_request;
+     got_response[description="on_response_chunk_received"];
```
asvetlov: Please rename `got_response` and `send_request` nodes to be close to actual signal names. Maybe that long names could be shortened on diagram -- be creative.

Also I like to see arrows what explicitly points that chunk_sent and chunk_received events can come more than once.

aiohttp/client_reqrep.py
kowalski: true, changed to `None`

aiohttp/client_reqrep.py
kowalski: Agreed. I switched now to `_traces` in both classes

aiohttp/client_reqrep.py
kowalski: also changed to `None`

asvetlov:
Wait, send_request and got_response in my mind is mostly for headers. We need different nodes for body chunks.

asvetlov:
Please add `..versionadded:: 3.1` for new signals and data classes

setup.cfg
```diff
@@ -1,5 +1,6 @@
 [pep8]
 max-line-length=79
+ignore=E225,E226
```
kxepal: I'm in doubt that we should ignore these. Any reasons why to disable them globally?

kowalski: Well, I'm not in place to judge this. I did it to be able to save a file which has `64*1000` literal without spaces around `*`. Otherwise autopep just fixes it for me.

kowalski:
@asvetlov 
If I split nodes for headers and chunks this is what I get. What do you think ? 
zrzut ekranu z 2018-02-27 15-25-33

setup.cfg
kxepal: That autopep suggestion looks correct. You may also write `64000` instead without any loss in readability.

kowalski: I'm with you.  @asvetlov specifically asked me to keep it as was.

asvetlov: For me the strange thing is successful passing flake8 checks without the `ignore` setting.
I'm totally fine with 64000 or even better 0x10000. Pretty sure it should be 64KiB instead of 64kB

asvetlov:
The diagram looks better but I don't see arrows from sent/recevied nodes to exception.
Maybe resulting picture will be unreadable, in this case I suggest extracting headers/body signals into separate diagram -- like I did for connection establishment and DNS lookup.
I don't know -- please do the best but if you'll fail I'll polish the diagram after merging.

Also the diagram shows that we need `on_request_headers_sent` and `on_response_headers_received` signals maybe :) -- but in separate PR obviously.

docs/tracing_reference.rst
```diff
@@ -147,6 +155,23 @@ TraceConfig
...
+      ``params`` is :class:`aiohttp.TraceRequestChunkSentParams` instance.
+
+   .. attribute:: on_response_chunk_received
+
+
+      Property that gives access to the signals that will be executed
+      when a chunk of response body is received.
+
+      ``params`` is :class:`aiohttp.TraceResponseChunkReceivedParams` instance.
+
```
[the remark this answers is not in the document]
kowalski: done

kowalski:
@asvetlov done

aiohttp/client_reqrep.py
```diff
@@ -475,7 +477,10 @@ def keep_alive(self):
             if self.url.raw_query_string:
                 path += '?' + self.url.raw_query_string
 
-        writer = StreamWriter(conn.protocol, conn.transport, self.loop)
+        writer = StreamWriter(
+            conn.protocol, conn.transport, self.loop,
+            on_chunk_sent=self._on_chunk_request_sent
```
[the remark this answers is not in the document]
kowalski: `StreamWriter` is used in (at least) two scenarios:
- here, for writing request body
- in RequestHandler for writing response to incoming request

In the later case, having `traces` an calling `on_request_chunk_sent` seems wrong, because the class is used for writting something completely different. So `on_chunk_sent` is introduced to have something generic.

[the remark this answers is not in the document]
pfreixes: Oks good point!

aiohttp/client_reqrep.py
```diff
@@ -209,6 +210,7 @@ def __init__(self, method, url, *,
         if data or self.method not in self.GET_METHODS:
             self.update_transfer_encoding()
         self.update_expect_continue(expect100)
+        self._traces = traces or []
```
asvetlov: In other aiohttp code we usually use another pattern:
```
if traces is None:
    traces = []
self._traces = traces
```
Please change the one-liner to explicit check.
It is a very subtle optimization for if `traces` is an empty list already you code calculates `False` for `bool(traces)` check and creates an empty list again.

aiohttp/client_reqrep.py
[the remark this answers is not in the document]
asvetlov: +1

aiohttp/client_reqrep.py
```diff
@@ -572,6 +583,7 @@ def __init__(self, method, url, *,
         self._timer = timer if timer is not None else TimerNoop()
         self._auto_decompress = auto_decompress
         self._cache = {}  # reqired for @reify method decorator
+        self._traces = traces or []
```
asvetlov: The same: replace `or` with explicit check for `is None` please

aiohttp/http_writer.py
```diff
@@ -16,7 +17,12 @@
 
 class StreamWriter(AbstractStreamWriter):
 
-    def __init__(self, protocol, transport, loop):
+    def __init__(self, protocol, transport, loop, on_chunk_sent=None):
+        assert (
```
asvetlov: The assert is in the hot-pass route.
I understand your striving for type checking but all `StreamWriter` usage is under the library control, we always should pass the proper values for `on_chunk_sent` parameter.
Please remove the assert statement.

docs/tracing_reference.rst
```diff
@@ -34,16 +34,26 @@ Overview
      exception[shape=flowchart.terminator, description="on_request_exception"];
 
      acquire_connection[description="Connection acquiring"];
-     got_response;
-     send_request;
+     headers_received;
```
asvetlov, line 37: The diagram is cryptic a little  but I can live with it.

setup.cfg
asvetlov: Please drop the change. Feel free to replace the limit with `0x10000` if needed -- I'm +-0 for it.

asvetlov:
@kxepal @pfreixes please review again

kxepal:
I agree with already requested changes, so nothing more could add.

kxepal:
@asvetlov 
Ok, approving to not be a blocker. It seems you're on pulse of further changes here.

asvetlov:
@kxepal thanks

pfreixes:
Just missing the proper tests in `test_http_write.py` and `test_client_response.py` that must check the changes at the function level, for the default path - `traces` and `on_chunk_sent` as None - and 
 the none default  path. We alredy have it for `test_connection.py` right now.

I know that globally new handlers are tested via `test_client_session.py`, but its more an integration test, local tests allow us to check al branches explicitly.

asvetlov:
@pfreixes agree.
@kowalski please add required tests

setup.cfg
kowalski: done

aiohttp/client_reqrep.py
kowalski: done

aiohttp/client_reqrep.py
kowalski: done

aiohttp/http_writer.py
kowalski: done

aiohttp/http_writer.py
```diff
@@ -62,6 +70,9 @@ def _write(self, chunk):
         writer can't be used after write_eof() method being called.
         write() return drain future.
         """
+        if self._on_chunk_sent is not None:
+            await self._on_chunk_sent(chunk)
+
```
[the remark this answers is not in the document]
kowalski: done

aiohttp/client_reqrep.py
```diff
@@ -796,6 +808,8 @@ def _notify_content(self):
         if self._content is None:
             try:
                 self._content = await self.content.read()
+                for trace in self._traces:
+                    await trace.send_response_chunk_received(self._content)
```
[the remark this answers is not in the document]
kowalski: done

kowalski:
all done

pfreixes:
Thanks !

kowalski:
@asvetlov done (added `./CHANGES/2767.feature`)

asvetlov:
Great job, thanks to everyone.
Will merge when tests pass.

Issue #1749, opened 2017-03-24
Type hinting in code


penguinolog:
## Long story short

While working on big project, typing check helps to exclude shot in the leg scenario.

## Expected behaviour

Type annotations on public methods.

## Actual behaviour

No any type annotations, including the most docstrings.

## Note

**Need a decision: typing in global requirements or use version-specific requirements (PEP496/PEP508)**

penguinolog:
@fafhrd91 Please comment decision, if possible.

fafhrd91:
Personally, I do not use type annotations. But I am fine with type annotations on public methods

penguinolog:
I'm about requirements on `typing` library. We can require it on every python version and can specify 3.4 only due to it presence in 3.5+ out of box.

fafhrd91:
That's fine

iceboy233:
I personally dislike type annotations in performance sensitive code in python. Any footprint bloat would slow things down, especially in interpreted languages.

fafhrd91:
I thought there should be no runtime penalty?

iceboy233:
the annotations are accessible at runtime through `__annotations__`, so it will increase the memory footprint, allocator load (for the hash table for `__annotations__`), incref for the type objects, parsing time, etc.

fafhrd91:
Ah, but compared to overall speed of python that doesn't really matter. I guess it is possible to come up with method how to strip annotations data at startup

iceboy233:
The annotations are not necessarily for type hints and are runtime visible. It's just the IDEs using them for type hints. I have used annotation for other things in my programs, such as param sanitizing.

asvetlov:
Annotations are perfect concept but before applying it to aiohttp we need reliable type checker tool.
The only existing tool for now is `mypy` and it's not production ready yet.

At least I had big problems when tried to check annotations for smaller aio-libs projects.
Let's postpone the issue (without closing).

thehesiod:
extra datapoint: all the code I write has type hints as it greatly helps finding issues in pycharm while developing

kxepal:
I also had negative experience with type hints as annotations. Tools are not ready indeed and you have to choose between two evils: do additional imports for getting types and accidentally end with circular ones or double number of files to maintain. So far good old sphinx docstrings are more handy and PyCharm can use it for type hinting. You still have to write docs anyway.

[...]

PR #2769, opened 2018-02-26
Add type hinting

mosquito:
## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -9,6 +9,7 @@ import sys
 import traceback
 import warnings
 from collections.abc import Coroutine
+from typing import Coroutine
 
 from multidict import CIMultiDict, MultiDict, MultiDictProxy, istr
 from yarl import URL
@@ -42,6 +43,8 @@ __all__ = (client_exceptions.__all__ +  # noqa
 
 # 5 Minute default read and connect timeout
 DEFAULT_TIMEOUT = 5 * 60
+
+_ResponseType = Coroutine[ClientResponse]
 
 
 class ClientSession:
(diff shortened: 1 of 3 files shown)


mosquito:
For me it's a little bit painful when I don't see `web.Application.router`s methods,  or don't see client response attributes and methods.

This changes already tested by me on my PyCharm (with `pip install -e ../aiohttp`).

asvetlov:
No. Tools are not ready yet, adding type hints without checks is pointless and can make a harm easy.
See #1749

asvetlov:
Say, your PR declares a return type for `client.request` as a coroutine with `ClientResponse` return type but actually it is a much more complex object with async context manager interface.

mosquito:
@asvetlov this "complex object" inherited from coroutine. 

Note about tooling: PyCharm and SublimeText (with anaconda) supports this. This fixes needs only for "developer experience".

It's sad that I have to write every time `# type: ClientResponse`:

```python

...
        async with self.client_session(headers=headers) as session:
            async with session.get(url) as response:  # type: ClientResponse
                result = await response.json()
...
```

asvetlov:
1. Type for async context manager with `__await__` method should be different from mypy/types perspective.
2. You are talking about "developer experience", I am about type correctness checking. If we will add types we should add type checker to our test suite. Unchecked type declaration is ad bad as code not covered by tests from my perspective.

So we need a robust tool. Mypy doesn't support https://www.python.org/dev/peps/pep-0561/ yet unfortunately, thus it cannot be used *right now*. I hope mypy dev team will fix it. Personally I have no time for working on mypy unfortunately.

mosquito:
Should we make `.pyi` file with type declaration?

asvetlov:
We should have working checker with PEP 0561 implemented first.
After that we can start a discussion about typing strategy.
I suspect we will have a mix of stub files for C extensions at least and embedded annotations.

Issue #2773, opened 2018-02-27
"InvalidStateError: invalid state" when data is chunked


rq4w7z:
I am use simple aiohttp (3.0.2) server for Telegram bot:

```
API_TOKEN = 'XXXXxxXXXXxxxxxx-XXXXXXX_xxxxXXXXXx'
bot = telebot.TeleBot(API_TOKEN)
app = web.Application()
async def handle(request):
    if request.match_info.get('token') == bot.token:
        request_body_dict = await request.json()
        update = telebot.types.Update.de_json(request_body_dict)
        bot.process_new_updates([update])
        return web.Response()
    else:
        return web.Response(status=403)

app.router.add_post('/{token}/', handle)
bot.remove_webhook()
bot.set_webhook(url=WEBHOOK_URL_BASE+WEBHOOK_URL_PATH,
                    certificate=open(WEBHOOK_SSL_CERT, 'r'))
context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
context.load_cert_chain(WEBHOOK_SSL_CERT, WEBHOOK_SSL_PRIV)
web.run_app(
        app,
        host=WEBHOOK_LISTEN,
        port=WEBHOOK_PORT,
        ssl_context=context)
```

When message send from Telegram server I receive this error:

> ERROR:asyncio:Exception in callback _SelectorSocketTransport._read_ready()
> handle: <Handle _SelectorSocketTransport._read_ready()>
> Traceback (most recent call last):
>   File "/usr/lib64/python3.6/asyncio/events.py", line 127, in _run
>     self._callback(*self._args)
>   File "/usr/lib64/python3.6/asyncio/selector_events.py", line 731, in _read_ready
>     self._protocol.data_received(data)
>   File "/usr/lib64/python3.6/asyncio/sslproto.py", line 516, in data_received
>     self._app_protocol.data_received(chunk)
>   File "/home/telegram/bwb_env/lib64/python3.6/site-packages/aiohttp/web_protocol.py", line 252, in data_received
>     self._waiter.set_result(None)
> asyncio.base_futures.InvalidStateError: invalid state

I get the response completely, but further after a number of responses there are big delays and denial of service

When I use curl for make test request I did not receive this error

[...]

PR #2775, opened 2018-02-27
Don't raise InvalidStateError on two consequenced  data_received() calls

asvetlov:
Fix for #2773


pfreixes:
Thanks for the fixing!

asvetlov:
Welcome!
Every major aiohttp release has such bugs.
Your changes was been made for good reason.
Unfortunately our test suite did not catch the problem.
Now everything is fine I hope.

PR #2792, opened 2018-03-03
added _reuse_address and _reuse_port to web_runner.TCPSite.__slots__

jancespivo:
Hi,
`_reuse_address` and `_reuse_port` are not included in `TCPSite.__slots__`

```python
import aiohttp.web
aiohttp.web.run_app(app)
```
Invokes:
`AttributeError: 'TCPSite' object has no attribute '_reuse_address'`

Best regards

### DIFF
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -57,7 +57,7 @@ class BaseSite(ABC):
 
 
 class TCPSite(BaseSite):
-    __slots__ = ('_host', '_port')
+    __slots__ = ('_host', '_port', '_reuse_address', '_reuse_port')
 
     def __init__(self, runner, host=None, port=None, *,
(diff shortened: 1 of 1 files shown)


asvetlov:
Hmm. This is a bug definitely but I wonder why tests are passed without slots.
Your code snipped didn't raise an error too.

drpoggi:
I believe this error only happens on 3.7 where now `ABC` defines a `__slots__`

asvetlov:
You are right. This explains everything.

jancespivo:
Yes, sorry. It was actually python 3.8.0a0 I forgot change the `venv` :man_facepalming:
I am a little suprised the using of `__slots__`  is useless for class inherited from `ABC` in python < 3.7

drpoggi:
It's not useless, it just requires subclasses to be explicit.

PR #2809, opened 2018-03-06
The mechanism of the rules for the matching of subapps

aamalev:
I propose to implement subdomain as part of a general approach, when subapp can be matched according to a certain rule

    subapp = web.Application()
    app.add_subapp(Domain('example.com'), subapp)

No

## Related issue number
https://github.com/aio-libs/aiohttp/issues/1342
https://github.com/aio-libs/aiohttp/pull/2194

## Checklist


asvetlov:
I think `app.add_domain()` is more readable.

Regarding to router changes: they are either incomplete or sense-less. Say, I see no usage of `DefaultRule` and many others.

aamalev:
@asvetlov add_domain is more readable but highly specialized.

DefaultRule was born because of the lack of add an application without a prefix.
AbstractRuleMatching allow users to create rules themselves.

UseCases:
* Merging routers of two or more subapps without prefixes
* Serving different domains through a Domain rule
* Serving different masks of domains and subdomains through user rules
* Service of authorized and public requests in various subapps
* Disabling of subdomains by billing rules

asvetlov:
It still looks too complex, `add_doumain()` is better at least for documenting and teaching.
For example in Python classes usually are created by `class A(Base):` statement while technically `type(...)` call should be enough, and `type()` is potentially more flexible.

amirouche:
Should this be closed?

amirouche:
I am asking because I would like to continue to work on #1342

aamalev:
@asvetlov please review

aamalev:
@asvetlov Final realization adds only one public interface Application.add_domain(str, subapp) which encapsulates the mechanism

aiohttp/abc.py
```diff
@@ -145,3 +145,18 @@ def __init__(self, logger, log_format):
     @abstractmethod
     def log(self, request, response, time):
         """Emit log to logger."""
+
+
+class AbstractRuleMatching(ABC):
```
asvetlov: I doubt if we need this abstract class: it is used for `Domain` class only.

tests/test_urldispatch.py
```diff
@@ -1054,14 +1150,14 @@ def test_frozen_router_subapp(app, loop):
     subapp = web.Application()
     subapp.freeze()
     with pytest.raises(RuntimeError):
-        app.add_subapp('/', subapp)
+        app.add_subapp('/pre', subapp)
```
asvetlov, line 1148: Why the test has changed?

tests/test_urldispatch.py
```diff
@@ -1054,14 +1150,14 @@ def test_frozen_router_subapp(app, loop):
...
 
 
 def test_frozen_app_on_subapp(app, loop):
     app.freeze()
     subapp = web.Application()
     with pytest.raises(RuntimeError):
-        app.add_subapp('/', subapp)
+        app.add_subapp('/pre', subapp)
```
asvetlov, line 1155: Is the change required? Why?

tests/test_urldispatch.py
aamalev: Empty prefix is ​​not supported, and now the prefix checks first then subapp

tests/test_urldispatch.py
aamalev: to check subapp rather than the prefix

    raise ValueError("Prefix cannot be empty")

asvetlov: ok

tests/test_urldispatch.py
asvetlov: good

aiohttp/abc.py
asvetlov: Still the question

aamalev: There are two and a half ways to decide his fate:

1 Throw away

2 To refine the subsystem of rules

2+1/2. Combine MatchedSubAppResource and PrefixedSubAppResource as they differ practically only by the method resolve

    class PrefixedSubAppResource(PrefixResource):
        async def resolve(self, request):
            if not request.url.raw_path.startswith(self._prefix):
                return None, set()

    class MatchedSubAppResource(PrefixedSubAppResource):
        async def resolve(self, request):
            if not await self._rule.match(request):
                return None, set()


and implement the rule

    class Prefix(AbstractRuleMatching):
        def __init__(self, prefix):
            self._prefix = prefix
    
        @property
        def prefix(self):
            return self._prefix
        
        def match(self, request):
            return request.url.raw_path.startswith(self._prefix)

asvetlov: I'd like to not introduce a new **public** ABC class.
Moreover, in aiohttp 4.0 I want to drop a pluggable router. By this, we'll have the only supported router implementation -- the current one. In this light adding a new ABC to API part that will never be extended outside doesn't make any sense.

If you have a feeling that ABC is valuable -- feel free to keep it but move into `web_urldispatcher.py`.

aamalev:
@asvetlov I moved AbstractRuleMatching to web_urldispatcher
but mypy shot in an unexpected place
```
aiohttp/helpers.py:222: error: Too few arguments for "BasicAuth"
aiohttp/http_websocket.py:547: error: The return type of "__init__" must be None
```

and local mypy-0.641
```
aiohttp/helpers.py:221: error: Argument 1 to "BasicAuth" has incompatible type "Optional[str]"; expected "str"
aiohttp/helpers.py:221: error: Argument 2 to "BasicAuth" has incompatible type "Optional[str]"; expected "str"
```

asvetlov:
Please merge master, mypy is updated and `helpers.py` is fixed

asvetlov:
Thanks!

docs/web_reference.rst
```diff
@@ -1387,6 +1387,21 @@ duplicated like one using :meth:`Application.copy`.
 
       :returns: a :class:`PrefixedSubAppResource` instance.
 
+   .. method:: add_domain(domain, subapp)
```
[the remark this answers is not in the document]
asvetlov: The pull request is welcome!

PR #2812, opened 2018-03-07
partially fixed #2805 (multiple http byte ranges are not yet supported)

### LINKED ISSUE #2805: BaseRequest.http_range returns the wrong slice when "start" is missing
## Long story short

When streaming range based requests from file, the method [http_range](
https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.http_range) comes quite handy, unfortunately negative indexes don't work as expected.

The offending line 442 is in [web_request.py](https://github.com/aio-libs/aiohttp/blame/20928e1d6ba4887b9137eacd8ad7dda2dbbb09e6/aiohttp/web_request.py#L442) it should be "start = -end" instead of "end = -end"

Other valid ranges are described in https://tools.ietf.org/html/rfc7233#section-2.1 
I don't know if it's allowed to just return a 416 on the more exotic ones though:)

## Expected behaviour
[Request Header]
Range: bytes=-5678

[BaseRequest.http_range]
slice(start=-5678, stop=None, step=1)

## Actual behaviour
[Request Header]
Range: bytes=-5678

[BaseRequest.http_range]
slice(start=None, stop=-5678, step=1)

## Steps to reproduce

run an aiohttp server and request a resource using the range header.
(do I have to post the whole setup, code included?)

## Your environment

Running aiohttp 1.2.0 on debian 9 (stretch) ... I know, I know :)

Fahrradkette:
Now the slice returned by Request.http_range() is also correct when there is no "start" value specified i.e. when the requester only wants the tail of the range.

If they rely on rng.start and rng.stop this change will break their app.

## Related issue number

Partial fix for #2805, if we want to create a set or list of ranges from those exotic headers, we probably need another method name (http_multi_ranges) or something

## Checklist

### DIFF
--- a/CHANGES/2805.bugfix
+++ b/CHANGES/2805.bugfix
@@ -1,1 +1,2 @@
-Property http_range of BaseRequest now returns a python-like slice when requesting the tail of the range. It's now indicated by a negative value in rng.start rather then in rng.stop
+Property `BaseRequest.http_range` now returns a python-like slice when requesting the tail of the range. 
+It's now indicated by a negative value in `range.start` rather then in `range.stop`
--- a/CHANGES/2805.bugfix
+++ b/CHANGES/2805.bugfix
@@ -1,0 +1,1 @@
+Property http_range of BaseRequest now returns a python-like slice when requesting the tail of the range. It's now indicated by a negative value in rng.start rather then in rng.stop
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -193,9 +193,9 @@ class FileResponse(StreamResponse):
         # If a range request has been made, convert start, end slice notation
         # into file pointer offset and count
         if start is not None or end is not None:
-            if start is None and end < 0:  # return tail of file
-                start = file_size + end
-                count = -end
+            if start < 0 and end is None:  # return tail of file
+                start = file_size + start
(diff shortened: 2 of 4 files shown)


ikanobori:
What happens to `end` here, aren't you going to get a `start` of say -500 combined with an `end` of 500 and get to a `slice(-500, 500, 1)`?

Maybe `end` needs to be reset :)

Fahrradkette:
This logic should make from a "-500" the slice(-500, None, 1) if I understood it correctly.

Edit: I think this PR needs more work on the receiving end of http_range, i.e. in [web_fileresponse.py](https://github.com/aio-libs/aiohttp/blob/b348e0f47c8ad16f898fd0dffa00c814b259131e/aiohttp/web_fileresponse.py#L186-L213)

I guess we're also breaking code which relies on the current slice notation

Fahrradkette:
slice(None, 42, None) != slice(0, 42, 1) even though they're referencing the same data.

I'm not sure on how to test this throughly, should we rather set _payload then passing it as a request payload? If it's OK to set the payload, we could use an int-list and therefore we would have unique values in the examined range.

I also don't know which tests are required to check if it fails correctly, should I just add one which checks some chars outside ascii 48-57 to check the regex?

PR #2815, opened 2018-03-08
doc fix: Response does not take positional parameters.

JulienPalard:
Fixed doc about Response not taking positional parameters but used as `Response("some text")`.


asvetlov:
Thanks

PR #2824, opened 2018-03-12
Add static helpers

### LINKED ISSUE #2795: Add web.static and web.RouteTableDef.static
As aliases for UrlDispatcher.add_static.

The idea is adding static file resources in declarative way: `app.add_routes([web.static(...)])` and `table.static(...)`.

Part of #2766

asvetlov:
Fixes #2795

### DIFF
--- a/tests/test_route_def.py
+++ b/tests/test_route_def.py
@@ -272,6 +272,17 @@ def test_repr_route_def_with_extra_info():
     assert repr(rd) == "<RouteDef GET /path -> 'handler', extra='info'>"
 
 
+def test_repr_static_def():
+    routes = web.RouteTableDef()
+
+    routes.static('/prefix', '/path', name='name')
+    async def handler(request):
+        pass
+
+    rd = routes[0]
+    assert repr(rd) == "<StaticDef /prefix -> /path, name='name'>"
+
+
 def test_repr_route_table_def():
     routes = web.RouteTableDef()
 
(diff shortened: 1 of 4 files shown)


tests/test_route_def.py
```diff
@@ -256,6 +272,17 @@ def test_repr_route_def_with_extra_info():
     assert repr(rd) == "<RouteDef GET /path -> 'handler', extra='info'>"
 
 
+def test_repr_static_def():
+    routes = web.RouteTableDef()
+
+    routes.static('/prefix', '/path', name='name')
```
kxepal, line 278: `@`

asvetlov: No, this method is intentionally not a decorator: static has no user-supported handler, only `prefix`, `path` and optional parameters. I'll document it properly.

The reason why `url.static()` and `routes.static` should exist is it is a part of configuration which very often is constant (safe to be done at module level). No need to add `app.add_statiic()` shortcut.

`app.router.add_static()` is redundant (but still exist and do all work internally).
`app.add_subapp` is a different beast: it requires `subapp` instance, that's why I don't want to add shourcuts for `add_subapp`.

kxepal: Aha, I see. Now it looks fine. On previous version there was a function right after and that looks like missed @ from decorator.

asvetlov:
Thanks @kxepal

PR #2828, opened 2018-03-13
Close a connection if an unexpected exception occurs while sending a request

### LINKED ISSUE #2827: Exception handling doesn't close connection
## Long story short

With certain malformed urls, the connection is not properly closed.   This causes an "Unclosed connection".

## Expected behaviour

Connection cleaned up when there is an exception writing headers.

## Actual behaviour

Connection not properly cleaned up.

## Steps to reproduce
Here are two examples, one mocked and one using a live site.  Both should produce an "Unclosed connection" error.

requires pytest and [aresponses](https://github.com/circleup/aresponses)
```python
@pytest.mark.asyncio
async def test_invalid_location(aresponses):
    # real example: 'http://www.skinhealthcanada.com/index.cfm?pagepath=Brands&id=66172'
    url = 'http://badwebsite.circleup.com/'
    bad_url = 'http://badwebsite.circleup.com/in\udcaedex.html'

    aresponses.add('badwebsite.circleup.com', aresponses.ANY, 'GET', aresponses.Response(status=302, headers={'Location': bad_url}))
    aresponses.add('badwebsite.circleup.com', aresponses.ANY, 'GET', aresponses.Response())

        try:
            response = await session.get(bad_url)
        except Exception as e:
            print(repr(e))

        response = await session.get(bad_url)


@pytest.mark.asyncio
async def test_invalid_location_live():
    url = 'http://www.skinhealthcanada.com/index.cfm?pagepath=Brands&id=66172'
    bad_url = 'http://www.skinhealthcanada.com/index.cfm?pagepath=Brands/Environ\udcae&id=66220'
        try:
            response = await session.get(bad_url)
        except Exception as e:
            print(repr(e))

        response = await session.get(bad_url)

```

## Your environment

client

Python 3.6.4
Ubuntu 17.10
virtualenv
aiohttp==3.0.6

## possible solution
change aiohttp/client.py lines 328-343 to
```python
                    try:
                        try:
                            resp = req.send(conn)
                        except BaseException:
                            conn.close()
                            raise
                        try:
                            await resp.start(conn, read_until_eof)
                        except BaseException:
                            resp.close()
                            conn.close()
                            raise
                    except ClientError:
                        raise
                    except OSError as exc:
                        raise ClientOSError(*exc.args) from exc
```

brycedrennan:
## What do these changes do?

Close a connection if an unexpected exception occurs while sending a request

No.

## Related issue number

Resolves #2827

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -326,7 +326,11 @@ class ClientSession:
                     tcp_nodelay(conn.transport, True)
                     tcp_cork(conn.transport, False)
                     try:
-                        resp = await req.send(conn)
+                        try:
+                            resp = await req.send(conn)
+                        except BaseException:
+                            conn.close()
+                            raise
                         try:
                             await resp.start(conn, read_until_eof)
                         except BaseException:
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -328,13 +328,12 @@ class ClientSession:
                     try:
                         try:
                             resp = await req.send(conn)
+                            try:
+                                await resp.start(conn, read_until_eof)
+                            except BaseException:
+                                resp.close()
+                                raise
                         except BaseException:
-                            conn.close()
-                            raise
-                        try:
-                            await resp.start(conn, read_until_eof)
-                        except BaseException:
-                            resp.close()
                             conn.close()
                             raise
                     except ClientError:
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -371,12 +371,49 @@ async def test_reraise_os_error(create_session):
         # return self.transport, self.protocol
         return mock.Mock()
     session._connector._create_connection = create_connection
+    session._connector._release = mock.Mock()
 
     with pytest.raises(aiohttp.ClientOSError) as ctx:
         await session.request('get', 'http://example.com')
     e = ctx.value
     assert e.errno == err.errno
     assert e.strerror == err.strerror
+
+
+async def test_close_conn_on_error(create_session):
+    class UnexpectedException(Exception):
+        pass
+
+    err = UnexpectedException("permission error")
+    req = mock.Mock()
+    req_factory = mock.Mock(return_value=req)
+    req.send = mock.Mock(side_effect=err)
+    session = create_session(request_class=req_factory)
+
+    connections = []
+    original_connect = session._connector.connect
+
+    async def connect(req, traces=None):
+        conn = await original_connect(req, traces=traces)
+        connections.append(conn)
+        return conn
+
+    async def create_connection(req, traces=None):
+        # return self.transport, self.protocol
+        conn = mock.Mock()
+        return conn
+
+    session._connector.connect = connect
+    session._connector._create_connection = create_connection
+    session._connector._release = mock.Mock()
+
+    with pytest.raises(UnexpectedException):
+        await session.request('get', 'http://example.com')
+
+    # normally called during garbage collection.  triggers an exception
+    # if the connection wasn't already closed
+    for c in connections:
+        c.__del__()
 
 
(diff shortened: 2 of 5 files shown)


pfreixes:
@asvetlov and @brycedrennan  some comments on this MR and the issue related

- Using the context manager way of sending a request rather than use an `await` this issue shouldn't appear, am I right ?

- And specifically of that MR, Why `close` and not `release`? the context manager uses as a teardown the `release` method.

asvetlov:
@pfreixes 
1. Yes, from my understanding context manager should close connection on exception
2. `release` returns a connection back to pool. If exception has occurred the connection is in undefined state: it can have not-read-yet data etc. Dropping the connection entirely by `close()` is safer.

brycedrennan:
@pfreixes 
1. It doesn't appear to help to use the context manager.  I modified the test to use the context manager and it still has trouble without the fix.

2.  I don't know the difference. I used `close` because thats what was used to handle exceptions during `.start`

tests/test_client_session.py
```diff
@@ -379,6 +380,42 @@ def test_context_manager(connector, loop):
     assert e.strerror == err.strerror
 
 
+async def test_close_conn_on_error(create_session):
+    class UnexpectedException(Exception):
```
kxepal: Your try/except block catches BaseExceptions, but here you test just for Exception. It's easy to change the code and break the behavior, but not the tests.

aiohttp/client.py
```diff
@@ -326,7 +326,11 @@ def request(self, method, url, **kwargs):
                     tcp_nodelay(conn.transport, True)
                     tcp_cork(conn.transport, False)
                     try:
-                        resp = await req.send(conn)
+                        try:
+                            resp = await req.send(conn)
+                        except BaseException:
+                            conn.close()
+                            raise
```
kxepal: It seems we're don't the same thing twice. How about to regroup try/except a bit?
```
try:
  resp = await req.send(conn)
  try:
    await resp.start(conn, read_unit_eof)
  except:
    resp.close()
    raise
except:
  conn.close()
  raise
```

kxepal:
And sign question about context managers. It's strange that they doesn't helps.

tests/test_client_session.py
brycedrennan: Done.

aiohttp/client.py
brycedrennan: Done.

kxepal:
@brycedrennan Thanks!

pfreixes:
Checked, sorry for the noise but finally I was able to connect all of the dots, I would like to propose a silly change in this MR.

The context manager is not able to close the connection because there is no response yet and won't be able to make it. My bet is that with your example @brycedrennan related with the context manager you are having another exception raised within the original exception.

Basically, the change that I propose is adding a check to make sure that the resp attribute is present [1].

Thoughts @asvetlov and @brycedrennan ?

[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L790

brycedrennan:
@pfreixes  `__aexit__` is never called because the exception is thrown during `__aenter__`.   I just put some breakpoints in the debugger, and confirmed by adding a try/except into the `__aenter__`.

Also the test is catching the specific exception that was thrown.

pfreixes:
@brycedrennan my fault I was totally confused, the exception happens in the `__aenter__` so as you said the `__aexit__` is not called.

Forget my comment. The response and the connection teardown in case of exception within Client.__request can't rely on the context manager.

asvetlov:
Everything looks good, will merge tomorrow if nobody objects.
The PR should be backported to 3.0 branch and new 3.0.x release should be published (I'll do it myself).
@brycedrennan thanks for your work!

PR #2838, opened 2018-03-15
Import the web package when aiohttp is imported

pfreixes:
## What do these changes do?

Import `aiohttp.web` package and its explicit dependencies when the user makes the typical `import aiohttp`.

I'm wondering if this was on purpose ...

### DIFF
--- a/aiohttp/__init__.py
+++ b/aiohttp/__init__.py
@@ -16,6 +16,7 @@ from .resolver import *  # noqa
 from .signals import *  # noqa
 from .streams import *  # noqa
 from .tracing import *  # noqa
+from .web import *  # noqa
 
 try:
     from .worker import GunicornWebWorker, GunicornUVLoopWebWorker  # noqa


drpoggi:
Couldn't this significantly slow down import time (sorry I haven't actually measured it yet)? `web` adds a lot of extra imports, this would hurt CLI programs which likely don't use any of that.

EDIT:
I measured and it's not as big a difference as I expected:

- Without `from .web import *` ([gist](https://gist.github.com/drpoggi/7888a56065391b9fe6a460d6d7887d86))
- With `from .web import *` ([gist](https://gist.github.com/drpoggi/55d3f28be1ce3552c5c8611ceb0842c5))

asvetlov:
Separating `aiohttp.web` namespace is intentional.

pfreixes:
Just for curiosity which is the rationale @asvetlov?

asvetlov:
To don't clash client response with server one etc.

pfreixes:
Thanks!

Issue #2844, opened 2018-03-17
FileResponse failed to prepare proper headers for Range requests


spcharc:
## Long story short
Please describe your problem and why the fix is important.

Users may pause their downloads of large static resources.
Many download managers including those browser-builtin ones check for Content-Range headers. Missing this header would cause downloads to fail.
I checked the source code and found out static resources are handled by FileResponse in web_fileresponse.py. But {0}.format(__title__)

## Expected behaviour
What is the behaviour you expect?

Assume you have a BIG file big_file
big_file.size() == 10000 (9.77kb)
A user downloaded 500 bytes (0-499) and connection lost.
And then the user resumes download with request: "Range:bytes=500-"
Expected header: "Content-Range:bytes 500-9999/10000"   <- this one is missing

## Actual behaviour
What's actually happening?

The downloads just failed immediately after clicking resume button.
Actually aiohttp handles Range requests very well. HTTP 206 and Content-Length are both correctly returned to users.

## Steps to reproduce
Please describe steps to reproduce the issue.
     If you have a script that does that please include it here within
     markdown code markup

Check the web_fileresponse.py source code, or try a download like this
```
    from aiohttp import web
    import asyncio
    loop = asyncio.get_event_loop()
    app = web.Application(loop = loop)
    app.add_static('/', folder_containing_large_file, show_index = True)
    web.run_app(app, host = '0.0.0.0', port = 4321)
```
goto http://your_addr:4321/
click the big_file

## Your environment

Describe the environment you have that lead to your issue.
     This includes aiohttp version, OS, proxy server and other bits that
     are related to your case. 
     
     IMPORTANT: aiohttp is both server framework and client library.
     For getting rid of confusing please put 'server', 'client' or 'both'
     word here.

Server
Python/3.6.4
aiohttp/3.0.9

## Temporary workaround
in file web_fileresponse.py:
    `line 228 == "if count:"`
in line 227 I added:
    `self.headers[hdrs.ACCEPT_RANGES] = 'bytes'`
(Accept-Ranges header, which is to indicate that the server supports Range requests, is also missing. But I  found out that usually web clients just ignore it. They still send Range requests even though this is missing. Maybe an Accept-Ranges:none header is needed to prevent Range requests?)
in line 229 I added:
`if start is not None: self.headers[hdrs.CONTENT_RANGE] = f'bytes {start}-{start+count-1}/{file_size}'`
(After this line is added, download managers work like a charm)
[...]

PR #2847, opened 2018-03-18
fileresponse content-range header

spcharc:
Feel free to close this request if it is not properly written. I'm still learning how to use github
## What do these changes do?
1. Now aiohttp provides Content-Range header for Range requests
2. Accept-Ranges:bytes for files
3. Out-of-range start-range requests now cause HTTP 416 Request Range Not Satisfiable according to rfc7233
    for instance, `file.size() == 200`
    but client gives this header `request.header['Range'] == 'bytes=99999-'`
4. Out-of-range tail-range requests would not break FileResponse now
    for instance, `file.size() == 200`
    but client gives this header `request.header['Range'] == 'bytes=-99999'`
5. Two reify(s): (in web_request.py)
    ```
    Request.if_unmodified_since
    Request.if_range
    ```
    Why not one function to handle them all (including Request.if_modified_since)?
    It seems that reify class uses `function.__name__` as the key to store results ...
    Their code is just the same with if_modified_since actually
    Maybe I should create a _private_func (of course not a reify decorated one) and use that. Copy and paste the same piece of code is not interesting.
6. Some more test cases. They send requests with If-Range or If-Unmodified-Since now
    Besides, some existing test functions get updated. They check Content-Range header now.

New headers returned
Out-of-range start-range requests result in HTTP 416
these are behavior changes, nothing more

## Related issue number
#2844 However it won't be fully solved, as multipart/byteranges is still not supported.
multipart/byteranges support involves:
1. changing the behavior of request.http_range function.
It should be able to handle multiple ranges and return a tuple of slices
This will break existing code.
2. It seems that MultipartWriter in multipart.py need to be updated to handle multipart/byteranges

## Checklist

- [?] I think the code is well written

I think this question is for you. Everbody considers him/her-self good code writers
I added a few comments though. I think the comments are well written

- [Y] Unit tests for the changes exist

Old test cases modified + new ones

- [Y] Documentation reflects the changes

Request.if_range and Request.if_unmodified_since are now reflected in Server->reference page

- [Y] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

I didn't know about this until everything is ready for a pull request. Went back and updated it

- [Y] Add a new news fragment into the `CHANGES` folder

OK. I went back and added 2844.feature
file content:"Provide Content-Range header for Range requests"
actually it's an improvement for an existing feature, not a new one

## Question
In web_fileresponse.py, line 170
self._length_check = False
What is this for? I failed to figure out what it is.
_length_check is in StreamResponse(web_response.py). I commented out this line and everything still worked. That's strange
line 170 is returning an http error.
Returning other http errors in FileResponse do not involve setting _length_check to False.

Edit: Question solved. Actually I had made a `grep -r _length_check` search and found it in web_response.py before I made this PR. But I failed to understand that users could also set the Content-Length header at that time.


spcharc:
Sorry about the commits
Now the coverage would be a little better

asvetlov:
@spcharc thank you for contribution.
I'll review in my first free time window.

spcharc:
Thank you!
Feel free to tell me if there's anything wrong. I'll try fixing it ASAP.

spcharc:
Now Content-Range is also provided when HTTP 416 is returned.
`file.size() == 200`
client request:`headers['Range'] = 'bytes=500-'`
server response:`headers['Content-Range'] == 'bytes */200'` <- tell client the actual size/allowed range
It seems that this is necessary as indicated in rfc7233
However I investegated several sites around the web, some belongs to large companies, and found out they usually do not provide Content-Range with HTTP 416. (Neither do many of them support multipart/byteranges. Some return 416 some even return 403 in this case. 403 Forbidden! what a simple solution. I like it.)

No test case change for this commit (577c81a).

spcharc:
No more changes of this PR beyond this point. ETag not supported yet (seems unnecessary)
Tested multiple kinds of stream video players / browsers / download managers, but none of them sends multi-range requests. multipart/byteranges support seems unnecessary too

asvetlov:
> Why not one function to handle them all (including Request.if_modified_since)?
It seems that reify class uses function.__name__ as the key to store results ...
Their code is just the same with if_modified_since actually
Maybe I should create a _private_func (of course not a reify decorated one) and use that. Copy and paste the same piece of code is not interesting.

It's up to you. I'm ok with `_private_func` for the case (we already have an exception for `Content-Type` handling.

asvetlov:
I'm ok with not supporting multi-range requests.

aiohttp/web_fileresponse.py
```diff
@@ -166,7 +167,12 @@ def __init__(self, path, chunk_size=256*1024, *args, **kwargs):
         modsince = request.if_modified_since
         if modsince is not None and st.st_mtime <= modsince.timestamp():
             self.set_status(HTTPNotModified.status_code)
-            self._length_check = False
+            # self._length_check = False     <- why this line is needed?
```
asvetlov: See https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_response.py#L340
Looks like is is important for chunked encoding, I don't recall details honestly.
Maybe @fafhrd91 can shed a light?

docs/web_reference.rst
```diff
@@ -329,6 +329,24 @@ and :ref:`aiohttp-web-signals` handlers.
...
 
+   .. attribute:: if_unmodified_since
+
+      Read-only property that returns the date specified in the
+      *If-Unmodified-Since* header.
+
+      Returns :class:`datetime.datetime` or ``None`` if
+      *If-Unmodified-Since* header is absent or is not a valid
+      HTTP date.
+
```
asvetlov, line 340: Here should be `..versionadded:: 3.1` tag.

docs/web_reference.rst
```diff
@@ -329,6 +329,24 @@ and :ref:`aiohttp-web-signals` handlers.
...
+   .. attribute:: if_range
+
+      Read-only property that returns the date specified in the
+      *If-Range* header.
+
+      Returns :class:`datetime.datetime` or ``None`` if
+      *If-Range* header is absent or is not a valid
+      HTTP date.
+
```
asvetlov, line 351: `.. versionadded::`

spcharc:
Thanks for review.
I'll adding version 3.1 tag right now
for content _length_check, if headers['Content-Length'] is set, the flag deletes it. because returning 304 should not contain any body
but content-length header is set after that line. (line 269 in web_fileresponse.py)

spcharc:
I suddenly realized that content-length could be set by a user!
If you do not use `add_static('/static_path/', static_folder)` but directly return `web.FileResponse(path, headers={'Content-Length':'100'})`, this would cause problem.
You are right. I'll modify it

asvetlov:
@spcharc thank you, very high quality work!

PR #2855, opened 2018-03-20
Update README.rst, add requirement: attrs

thomaszdxsn:
README file is difference with docs, with requirements section. 

README file missing `attrs` package, but it is required

## What do these changes do?

fix `README.rst`.

nothing.

## Related issue number

nothing.

## Checklist

nothing, just change README


asvetlov:
Thank you

Issue #2841, opened 2018-03-16
Client DNS start/end events not fired


MykolaPolovyi:
## Long story short

Unable to track dns timing via TraceConfig using client with TcpConnector, since `None` intead of `traces` arg passed to direct and proxy connections methods

```
    async def _create_connection(self, req, traces=None):
        """Create connection.

        Has same keyword arguments as BaseEventLoop.create_connection.
        """
        if req.proxy:
            _, proto = await self._create_proxy_connection(
                req,
                traces=None
            )
        else:
            _, proto = await self._create_direct_connection(
                req,
                traces=None
            )

        return proto
```
## Expected behaviour

working DNS tracing with TCP

## Actual behaviour

dns events missed

## Your environment
aiohttp: 3.0.9
OS X

asvetlov:
That's sad but the fix looks easy.
Would you provide a patch?

MykolaPolovyi:
yes

pfreixes:
shame of me, the integration test that wires all of the tracing does not cover this signal [1] ... if its not a big deal would be nice if we can add these signals to that test. That would help in the future to avoid the same mistake.

[1] https://github.com/aio-libs/aiohttp/blob/master/tests/test_web_functional.py#L1641

MykolaPolovyi:
Thanks for pointing the test. I actually wrote a different one and seems like in a wrong place. Will do there.

pfreixes:
Well this just checks the happy path an upper level, if you have others at function level please commit them as well!

PR #2858, opened 2018-03-20
Do fire DNS tracing events

asvetlov:
Fix for #2841

### DIFF
--- a/tests/test_web_functional.py
+++ b/tests/test_web_functional.py
@@ -2,6 +2,7 @@ import asyncio
 import io
 import json
 import pathlib
+import socket
 import zlib
 from unittest import mock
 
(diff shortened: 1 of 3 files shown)


tests/test_web_functional.py
```diff
@@ -1663,20 +1668,50 @@ def make_handler(appname):
...
+                return await self._resolver.resolve(host, port, family)
 
-    client = await aiohttp_client(app, trace_configs=[trace_config])
+    resolver = FakeResolver({'example.com': server.port})
+    connector = aiohttp.TCPConnector(resolver=resolver)
+    client = aiohttp.ClientSession(connector=connector,
+                                   trace_configs=[trace_config])
 
-    await client.get('/redirector', data="foo")
+    await client.get('http://example.com/redirector', data="foo")
```
pfreixes, line 1705: a `f"localhost:{server.port}/redirector"` does not fire a DNS event? If it does we wouldn't need it the `FakeResolver`

asvetlov: Resolver doesn't perform DNS lookup for IP addresses, only for domain names.
`localhost` should be sufficient but unfortunately there are too many boxes with declared but actually disabled IPv6 support.

pfreixes: sorry I don't get the following sentence.

> there are too many boxes with declared but actually disabled IPv6 support.

pfreixes: I dont want to block the PR just for a test that in any case can be at any moment refactorized.

asvetlov: Ok, I'll try to describe. Some broken environments (Travis-CI included IIRC) resolves `localhost` to both `127.0.0.1` and `::1` but IPv6 stack is disabled.
By this communication over IPv6 `::1` fails.
That's why `aiohttp.test_utils` and aiohttp test suite use IPv4 `127.0.0.1` explicitly.
The situation could be improved I hope but working on it is definitely out if the issue scope.

Issue #2834, opened 2018-03-15
Update TechEmpower benchmark


fafhrd91:
Benchmark still uses 2.1 version

https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Python/aiohttp

fafhrd91:
They run it regularly, so it is possible to see difference in human readable form

https://tfb-status.techempower.com

asvetlov:
Looks like the issue is addressed to me.
If nobody will pick it up I'll find a time but not sure when.
Will be pretty busy next two weeks.

fafhrd91:
doesn't really matter, someone need to update requirements. maybe ill be able to do this next week.

fafhrd91:
TechEmpower/FrameworkBenchmarks#3421

fafhrd91:
merged

fafhrd91:
3.0 is consistently slower than 2.x

pfreixes:
Do you have two URLs to compare them in that site? I'm feeling a bit stupid browsing through this site, I didn't manage to get that kind of comparison.

asvetlov:
I suspect it is a result of my change: d0e6e72947a70dea3b0b25be5fe3ad4875f6f0c6

The intention was forbidding a wild assignment a custom attrs to app and request.
I didn't want converting classes to `__slot__` based but pass through a deprecation period.
We can do support `__setattr__` only if the code is executed under `PYTHONASYNCIODEBUG=1` if it is very slow.

I should measure performance first.
Looks like a blocker for aiohttp 3.1 release

fafhrd91:
Results are available on this site https://tfb-status.techempower.com
You need to check commit id

fafhrd91:
here is link to previous results https://tfb-status.techempower.com/results/1dfecb5c-16ff-4622-aa14-d02987749f52

PR #2860, opened 2018-03-20
Prevent wild assignment in DEBUG mode only

asvetlov:
For sake of performance

Related to #2834


spcharc:
I like it.
but would this check impact performance?
of course disabling it would be better, I just wonder the extent

asvetlov:
Yes, technically it is a Python function call on every internal attribute assignment.
Very cheap if the check is implemented by `CPython C`/`Cython` but quite expensive if `__setattr__` is a pure Python function.

PR #2864, opened 2018-03-20
Fix dangling dash character in version number in docs.

manisenkov:
## What do these changes do?

Just fix annoying dangling dash character in the index page of docs:
screen shot 2018-03-20 at 21 05 36

No


kxepal:
Thanks! But does tag always contains `-` to be delimited from the version?

manisenkov:
Actually no, current version in `master` written as `3.1.0a0`
https://github.com/aio-libs/aiohttp/blob/4f8ea944bb049bc9af02a17e91716154db88afea/aiohttp/__init__.py#L1
Should tag always be separated with `-`?

kxepal:
`a0` isn't a tag in that mean, but that version fullfils PEP-440 requirements. I'm not sure what the tag value could be there, but if it's `a0`, `dev0` or sort of that thing, than patch is good.

@asvetlov what do you think?

asvetlov:
I like the change.
I've added a dash looking on pre-release docs. Released version looks confusing, agree.

Regarding to tags.
@kxepal ironically you are the author of this naming: https://github.com/aio-libs/aiohttp/blame/master/docs/conf.py#L26-L31
I don't care too much, let's just drop a dash and that's it.

asvetlov:
@manisenkov thanks

PR #2869, opened 2018-03-22
Set write buffer limit to zero on client and server

asvetlov:
#2698

### DIFF
--- a/CHANGES/2698.feature
+++ b/CHANGES/2698.feature
@@ -1,0 +1,1 @@
+Set write buffer limit to zero for both client and server.
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -58,6 +58,7 @@ class ResponseHandler(DataQueue, asyncio.streams.FlowControlMixin):
 
     def connection_made(self, transport):
         self.transport = transport
+        transport.set_write_buffer_limits(0)
 
     def connection_lost(self, exc):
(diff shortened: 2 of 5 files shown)


fafhrd91:
You should run some benchmarks

asvetlov:
Sure will do

pfreixes:
@asvetlov just some general considerations, I would like to put myself int he same page as you because sure I'm missing something.

The `StreamWriter.drain` method uses the Asyncio method `_drain_helper` which checks if the stream is paused, in case that is paused will wait till the stream is resumed. Having a new limit of 0 for the high water mark - that is what this MR is doing - shouldn't we expect that the stream is never paused? Making useless the call to the drain? 

At the same time, the call to the `StreamWriter.drain` made by the `StreamWriter.write` it's at least a bit confusing IMHO - sure I'm missing something again. It calls explicitly the `drain` if the size of the buffer sent till that specific moment is greater than `0x10000`, my questions here are:

1) Why 16 bytes?
2) Why call the drain when we have sent 16 bytes?
3) Why we are not just calling the `Asyncio._drain_helper` setting the high watermark to the desired value?

asvetlov:
Good questions. The initial idea was not using *aiohttp/asyncio buffering* but only OS kernel write buffer for a socket.
On benchmarking I see ~7% slowdown, need to investigate more.

Issue #2802, opened 2018-03-06
Support async iterators for data uploading on client


asvetlov:
Generators look like much more native solution over `@aiohttp.streamer` decorator.
Python 3.6 supports it out-of-the-box, on Python 3.5 users could install `async-generator` library.
`@streamer` should be deprecated.

asvetlov:
aiobotocore is broken, I think publishing the feature as part of bugfix release is possible, nothing should be broken again.

asvetlov:
See #2557 discussion for `StreamReader` dropping motivation

asvetlov:
Async iterators are done by #2873

PR #2873, opened 2018-03-24
Async iterators payload

asvetlov:
Implementation of #2802

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -81,6 +81,8 @@ class PayloadRegistry:
             self._normal.append((factory, type))
         elif order is Order.try_last:
             self._last.append((factory, type))
+        else:
+            raise ValueError("Unsupported order {!r}".format(order))
 
 
 class Payload(ABC):
@@ -316,7 +318,7 @@ class AsyncIterablePayload(Payload):
         try:
             # iter is not None check prevents rare cases
             # when the case iterable is used twice
-            while self._iter is not None:
+            while True:
                 chunk = await self._iter.__anext__()
                 await writer.write(chunk)
         except StopAsyncIteration:
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -316,8 +316,8 @@ class AsyncIterablePayload(Payload):
             while True:
                 chunck = await self._iter.__anext__()
                 await writer.write(chunck)
-        except AsyncStopIteration:
-            pass
+        except StopAsyncIteration:
+            self._iter = None
 
 
 PAYLOAD_REGISTRY = PayloadRegistry()
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -30,9 +30,9 @@ class LookupError(Exception):
 
 
 class Order(enum.Enum):
-    normal = enum.auto()
-    try_first = enum.auto()
-    try_last = enum.auto()
+    normal = 'normal'
+    try_first = 'try_first'
+    try_last = 'try_last'
 
 
 def get_payload(data, *args, **kwargs):
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -1,9 +1,12 @@
+import enum
 import io
 import json
 import mimetypes
 import os
 import warnings
 from abc import ABC, abstractmethod
+from collections.abc import AsyncIterable
+from itertools import chain
 
 from multidict import CIMultiDict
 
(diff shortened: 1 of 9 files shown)


asvetlov:
Working on docs

asvetlov:
Ready for review

aiohttp/payload.py
```diff
@@ -278,6 +296,30 @@ def __init__(self, value,
             content_type=content_type, encoding=encoding, *args, **kwargs)
 
 
+class AsyncIterablePayload(Payload):
+
+    def __init__(self, value, *args, **kwargs):
+        assert isinstance(value, AsyncIterable), \
```
kxepal: Shouldn't here be a regular `if not isinstace: raise TypeError` check?

aiohttp/payload.py
```diff
@@ -50,19 +61,26 @@ class PayloadRegistry:
     """
 
     def __init__(self):
-        self._registry = []
+        self._first = []
+        self._normal = []
+        self._last = []
 
-    def get(self, data, *args, **kwargs):
+    def get(self, data, _CHAIN=chain, *args, **kwargs):
```
kxepal: Shouln't `_CHAIN`  be keyword-only argument? Otherwise it will be overrided by a second argument which should get into `args`.

aiohttp/payload.py
```diff
@@ -50,19 +61,26 @@ class PayloadRegistry:
...
-        for factory, type in self._registry:
+        for factory, type in _CHAIN(self._first, self._normal, self._last):
             if isinstance(data, type):
                 return factory(data, *args, **kwargs)
 
         raise LookupError()
 
-    def register(self, factory, type):
-        self._registry.append((factory, type))
+    def register(self, factory, type, order=Order.normal):
```
kxepal: Everywhere else `order` is keyword-only argument. Shouldn't it be there the same for consistency?

aiohttp/payload.py
asvetlov: It should probably but other classes in the module use `assert` statements.

aiohttp/payload.py
asvetlov: good catch

aiohttp/payload.py
```diff
@@ -278,6 +296,30 @@ def __init__(self, value,
...
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/octet-stream'
+
+        super().__init__(value, *args, **kwargs)
+
+        self._iter = value.__aiter__()
+
+    async def write(self, writer):
+        try:
+            while True:
```
kxepal: Here need a check that `self._iter` is not None for the case when the same instance of this class will be used twice.

aiohttp/payload.py
asvetlov: fixed

aiohttp/payload.py
```diff
@@ -278,6 +296,30 @@ def __init__(self, value,
...
+                chunck = await self._iter.__anext__()
```
kxepal: s/chunck/chunk/

aiohttp/payload.py
kxepal: Ok.

aiohttp/payload.py
asvetlov: Done

aiohttp/payload.py
asvetlov: fixed

asvetlov:
Fixed everything except assert statements.

Working on asserts.

Thanks for review

asvetlov:
I've reverted the change for:

> Here needs a check that self._iter is not None for the case when the same instance of this class will be used twice.

Don't want to support such trick case. If an instance of the class will be used twice -- this is not intended behavior, better to fail in the case.
Failing fast is even better but I not sure if we really need to add extra checks. Anyway, passing the same generator instance twice into `itertools` API leads to unpredictable behavior too without any protection. Why should aiohttp be stricter?

kxepal:
@asvetlov 
No, I didn't ask for support. I was about to throw RuntimeError or sort of explicitly about that case instead of having weird `AttributeError: 'NoneType' object has no attribute '__anext__'`.

asvetlov:
Done

asvetlov:
async generators have an internal check for the case but I want to support any async iterable object (which could have no such strict checking).

PR #2877, opened 2018-03-25
Generate the most compact JSON representation in various places.

socketpair:
## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -7,6 +7,7 @@ import cgi
 import datetime
 import functools
 import inspect
+import json
 import netrc
 import os
 import re
@@ -741,3 +742,7 @@ def set_result(fut, result):
 def set_exception(fut, exc):
     if not fut.done():
         fut.set_exception(exc)
+
+def json_dumps(obj):
+    # Return the most compact serialisation.
+    return json.dumps(obj, ensure_ascii=False, allow_nan=False, separators=(',', ':'))
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -737,8 +737,7 @@ async def test_writer_serialize_io_chunk(buf, stream, writer):
 async def test_writer_serialize_json(buf, stream, writer):
     writer.append_json({'привет': 'мир'})
     await writer.write(stream)
-    assert (b'{"\\u043f\\u0440\\u0438\\u0432\\u0435\\u0442":'
-            b' "\\u043c\\u0438\\u0440"}' in buf)
+    assert '{"привет":"мир"}'.encode('utf-8') in buf
 
 
 async def test_writer_serialize_form(buf, stream, writer):
@@ -778,8 +777,8 @@ async def test_writer_write(buf, stream, writer):
 
          b'--:\r\n'
          b'Content-Type: application/json\r\n'
-         b'Content-Length: 18\r\n\r\n'
-         b'{"test": "passed"}'
+         b'Content-Length: 17\r\n\r\n'
+         b'{"test":"passed"}'
          b'\r\n'
 
          b'--:\r\n'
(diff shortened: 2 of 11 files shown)


tests/test_multipart.py
```diff
@@ -737,8 +737,7 @@ class CustomReader(aiohttp.MultipartReader):
 async def test_writer_serialize_json(buf, stream, writer):
     writer.append_json({'привет': 'мир'})
     await writer.write(stream)
-    assert (b'{"\\u043f\\u0440\\u0438\\u0432\\u0435\\u0442":'
-            b' "\\u043c\\u0438\\u0440"}' in buf)
+    assert '{"привет":"мир"}'.encode('utf-8') in buf
```
socketpair, line 740: Don't know how to remake tests without dependency on JSON dumps settings.

aiohttp/helpers.py
```diff
@@ -741,3 +742,7 @@ def set_result(fut, result):
 def set_exception(fut, exc):
     if not fut.done():
         fut.set_exception(exc)
+
+def json_dumps(obj):
+    # Return the most compact serialisation.
+    return json.dumps(obj, ensure_ascii=False, allow_nan=False, separators=(',', ':'))
```
socketpair, line 748:
```
asd='превед я русский текст, как же фигово китайцам с юникодом. да уж, вот же шь у них проблемы то!!'
timeit json.dumps(asd, ensure_ascii=True)  # before patch
100000 loops, best of 3: 3.06 µs per loop

timeit json.dumps(asd, ensure_ascii=False)   # after patch
100000 loops, best of 3: 4.8 µs per loop
```

BUT

```
len(json.dumps(asd, ensure_ascii=True))  # before patch
457

len(json.dumps(asd, ensure_ascii=False))  # after patch
97
```

Interesting, for English, situation is the same (about speed):
```
zxc='some english text that. It should be exactly the same by speed of encoding to JSON. Will verify that.'
timeit json.dumps(zxc, ensure_ascii=True)
The slowest run took 6.92 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 2.31 µs per loop

timeit json.dumps(zxc, ensure_ascii=False)
100000 loops, best of 3: 4.99 µs per loop

```

socketpair: So, this is VOLUME vs CPU trade-off. Don't know what to do next -- maybe some way to configure it globally ?

asvetlov:
I think it's ok to apply `ensure_ascii`: JSON is utf-8 encoded data.
But `allow_nan` and custom separators modify output too much. Please use a custom serializer in your project to change these arguments.

asvetlov (changes requested):
See comment above

kxepal:
Disallowing NaN is actually good since NaNs aren't in JSON specification - that's rudiment from the dark days. But `ensure_ascii=False` may lead to breaking compatibility case with some json libraries which don't expect utf-8 characters and strictly follows json.org specification.

UPD: that BC is ethemeral since we could say that we are  following [RFC-7159](https://tools.ietf.org/html/rfc7159) - that's a solution too.

samuelcolvin:
Surely no JSON library will be broken by a change to separators. How can this be a problem?

hellysmile:
Hello there! What about small idea, to make it configurable per application? It can be really useful! I often see rapidjson/ujson/etc in projects and it makes annoying always pass dumps=...

kxepal:
@hellysmile 
Hi! Eventually, it's already configurable. Just create your own JsonEncoder instance with proper setup and pass it around via dump argument.

asvetlov:
I suggest closing the issue.
If somebody wants to get a compact JSON dump he can use a custom JSON serializer.

socketpair:
@asvetlov

socketpair:
`json_response.__kwdefaults__['dumps'] = partial(json.dumps, ensure_ascii=False)` for those who find this issue.

PR #2883, opened 2018-03-27
Fix MsgType, replace it with WSMsgType

thomaszdxsn:
I found a error in README, the server side websocket usage have a error, that is `web.MsgType.text`. because according [Migration to 2.x](https://github.com/aio-libs/aiohttp/blob/d67332f7332875df395e93e469fd29e01fcf0a62/docs/migration_to_2xx.rst), it is change to `aiohttp.WSMsgType`. looks like a longtime error.


CHANGES/2784.bugfix
```diff
@@ -0,0 +1 @@
+Fix websocket client return EofStream.
```
asvetlov: Is this file here by accident?

thomaszdxsn: It is previous PR, maybe i should take a new branch

asvetlov:
Thanks

PR #2885, opened 2018-03-27
Use weakref.WeakSet for tracking websocket objects in the doc sample

achimnol:
Update the example code for "Graceful Shutdown" documentation to use `weakref.WeakSet` instead of plain list. Strong references here may leak the memory in certain situations.


asvetlov:
Thanks

PR #2886, opened 2018-03-27
setup editorconfig

jairhenrique:
Inspired on #2882 

http://EditorConfig.org


asvetlov:
I like it, thank you very much

PR #2889, opened 2018-03-28
fix client_ref doc.

thomaszdxsn:
add `headers` argument and it's description.

and fix some other doc error

ps.

I think `session.ws_connect()`'s argument `heartbeat`'s doc is not clear, not description about default value(`None`)'s behavior


asvetlov:
Thanks

Issue #2853, opened 2018-03-19
Publish release lifecycle, support duration and security reports


samuelcolvin:
I'm about to start a new project (an open source event/ticketing platform) I obviously want to use aiohttp but I have a few questions/requests about long term support and security.

It's particularly relevant in this project as I hope the platform will run for the client for years without significant maintenance from me or any sysadmin.

Could aiohttp publish a life-cycle for major releases: eg. how long they will be fully supported for, how long critical security issues will be fixed for?

Also, as per #2751 it would be good if aiohttp had an official procedure for reporting security issues and stated how we will deal with them.

I guess in summary I'm asking for a simplified version of https://docs.djangoproject.com/en/2.0/internals/security/. Is this possible?

asvetlov:
Are you asking about a email to send found security problems?
Or LTS burden etc?

samuelcolvin:
Ideally we would:
* add to the docs a way to report security problems, perhaps security@aiohttp.org which goes to you and perhaps other trusted aiohttp maintainers.
* publish how long major versions will be supported. Eg. 1 year, so v3 will be supported until February 2019.
* have an "announce" mailing list for release notices and warnings only.

asvetlov:
Sorry for a pause.
I'll keep it on my radar.
1. `security` email looks very easy technically.
2. A commitment for major version support is more questionable. 
  a. Should we keep the support of 3.x after 4.0 release?
  b. For what time period?
  c. How bug fixes from master should go to latest supported 3.x
  d. I have a vision how CPython works with these question but ideas from aiohttp community and especially from aiohttp committers are very valuable and important.
3. We have https://groups.google.com/forum/#!forum/aio-libs but even I has stopped to publish announcements because the procedure is too boring. Should we automate it? How? Ideas?

samuelcolvin:
> 1

great

> 2

I think 2 months of support (security bugs only) for an old major version after a new major version is released is sufficient, maybe others need longer?

Personally I think after a major version is released and stable nothing new should be released for that major version except critical operational bugs and security bugs. I should be able to merge and deploy a new minor version once tests pass with no more thinking. New features in minor releases just cause confusion. I guess this is how many projects treat patch releases.

> 3

[...]

PR #2896, opened 2018-04-01
Add backporting instructions

asvetlov:
A work on #2853

Thanks to @Mariatta quick feedback we have `cherry_picker` tool setup and working.
Now cherry-picking should be performed manually by running `cherry_picker <commit> 3.1` command but later we can setup a robot like `miss-islington.


CONTRIBUTING.rst
```diff
@@ -18,216 +18,16 @@ Workflow is pretty straightforward:
```
webknjaz, line 18: @asvetlov instead of keeping 2 `contributing.rst` files in repo, it is possible to have just one in `docs/` folder: I've noticed that github picks it up as well now.

webknjaz: https://help.github.com/articles/setting-guidelines-for-repository-contributors/

asvetlov: I'd like to have sphinx markup in docs version and keep a simple file on top-level for sake of visibility for newcomers.

asvetlov: But anyway good to know. Looks like `.rst` is not supported, only Markdown and text.

webknjaz: I think I saw RST being supported in some of my projects (JFYI)

Issue #2865, opened 2018-03-21
Improve client timeouts


danielrobbins:
## Long story short

Currently, aiohttp imposes a default 5 minute timeout on client HTTP requests, after which the connection will be aborted. While deemed a useful feature by some, it does mean that without special workarounds, aiohttp by default violates RFC 2616 section 9.3 ("GET") which is likely to be unexpected behavior for developers who expect an RFC-compliant HTTP implementation.

In summary: the most straightforward interpretation of RFC 2616 section 9.3 ("GET") is that under normal conditions, and without any explicit attempts to cancel the connection, the client HTTP request is supposed to run until the full entity identified by the Request-URI has been transmitted to the client. By default, this may not occur, and in fact under many circumstances, will not occur.

## Expected behaviour

An aiohttp client should, by default, hold the connection open indefinitely, as long as the underlying TCP connection is active, and allow the full entity identified by the Request-URI to be transmitted as per the RFC.

aiohttp should provide the ability for a client to abort a connection at any point, but this should not be the default behavior, since it is not the default behavior of HTTP, and aiohttp is supposed to implement HTTP.

## Rationale

RFC 2616 states that "The GET method means retrieve whatever information (in the form of an
   entity) is identified by the Request-URI." It does not specify that there is a time limit imposed upon this exchange of information. The entity is supposed to be returned to the client in its entirety. And in turn, the client is supposed to retrieve the entity, in its entirety. This is supposed to occur regardless of the amount of time required to send the data over the network. In other words, assuming a stable network, this is the intended behavior of HTTP.

There are only two exceptions to this behavior mentioned to this in the RFC: the conditional GET and the partial GET (Range: header). Unless these features are used, the intended behavior of the client is to retrieve the remote resource in its **entirety**, and for the server to send the resource in its **entirety**.

Again, the HTTP protocol does **not impose any time restriction** on the completion of an HTTP request. In fact, it is perfectly acceptable to an HTTP request to last more than 5 minutes -- **even hours** or days. This behavior is actually very common when transferring large amounts of data over a single HTTP request over a relatively slow connection.

[...]

PR #2897, opened 2018-04-01
Drop explicit async_timeout usage from docs

asvetlov:
As discussed in #2865


samuelcolvin:
Looks good to me, simple but helpful change for newbies.

PR #2910, opened 2018-04-04
fix resolve cancellation

thehesiod:
In my situation I have groups of 3 similar requests, where I take the fastest one out of the three and cancel the other two.

Lets say we have two groups of 3: A and B

B.1 succeeds
B.2 blocks on resolve (https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L702)

A.1 waits on resolve of B.2 (https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L689)

if we cancel  B.2, it ends up cancelling A.1 because the exception is broadcast to all waiters: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/locks.py#L31

so if a resolve is cancelled, only the task blocked on the resolve should be cancelled, not all the items waiting on the resolve.  Only if it's a DNS error should the error be broadcast.


aiohttp/connector.py
```diff
@@ -813,10 +813,13 @@ def _get_fingerprint(self, req):
...
         try:
-            hosts = await self._resolve_host(
+            # Cancelling this lookup should not cancel the underlying lookup
+            #  or else the cancel event will get broadcast to all the waiters
+            #  across all connections.
+            hosts = await  asyncio.shield(self._resolve_host(
                 req.url.raw_host,
                 req.port,
-                traces=traces)
+                traces=traces))
```
asvetlov: Please pass `loop=self._loop` into `shield()`.
Consider it as tiny speed optimization.

aiohttp/connector.py
```diff
@@ -813,10 +813,13 @@ def _get_fingerprint(self, req):
         fingerprint = self._get_fingerprint(req)
 
         try:
-            hosts = await self._resolve_host(
+            # Cancelling this lookup should not cancel the underlying lookup
+            #  or else the cancel event will get broadcast to all the waiters
+            #  across all connections.
+            hosts = await  asyncio.shield(self._resolve_host(
```
asvetlov: Is there two spaces between `await` and `asyncio.shield()`?

asvetlov:
Thank you for patch, looks almost good on the first glaze.
I need to review it more carefully, I hope to find a time for it in the evening.

aiohttp/connector.py
thehesiod: fixed

aiohttp/connector.py
thehesiod: fixed

aiohttp/connector.py
```diff
@@ -813,10 +813,13 @@ def _get_fingerprint(self, req):
         fingerprint = self._get_fingerprint(req)
 
         try:
-            hosts = await self._resolve_host(
+            # Cancelling this lookup should not cancel the underlying lookup
+            #  or else the cancel event will get broadcast to all the waiters
+            #  across all connections.
+            hosts = await asyncio.shield(self._resolve_host(
```
pfreixes, line 816: So this `shield` [1] can be removed, it was put in place just for the situation that you are commenting but taking a look at your MR - and the shield implementation - I was wrong.  

[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L702

thehesiod:
Ok was wondering about that, will remove

aiohttp/connector.py
thehesiod: Ok was wondering about that, removed

asvetlov:
The change looks good and reasonable but please add CHANGES file.
I'll merge and publish bugfix release than.

thehesiod:
@asvetlov added

changes/2910.bugfix
```diff
@@ -0,0 +1 @@
+fix cancellation broadcast during DNS resolve
```
asvetlov: I'm sorry but the folder name must be upper-cased: `CHANGES/2910.bugfix`.
The idea is having change-note on the top of changed files list.

thehesiod: I think this folder was renamed in master, so I ran into this: https://coderwall.com/p/5w3jca/git-case-sensitivity-on-os-x   fixed

asvetlov:
It was uppercased last in 2017, I don't remember the exact date.
The folder has `.gitignore` file, it always exists even if there are no feature files in `CHANGES`.

asvetlov:
LGTM.
@pfreixes please review.

Issue #2909, opened 2018-04-03
_on_chunk_sent at write_eof


vasilenkoalexey:
## Long story short

Looks like _on_chunk_sent not called when last chunk sent at write_eof

## Expected behaviour

_on_chunk_sent called

## Actual behaviour

_on_chunk_sent not called

## Steps to reproduce

at http_writer.py
```
async def write(self, chunk, *, drain=True, LIMIT=0x10000):
...
    if self._on_chunk_sent is not None:
        await self._on_chunk_sent(chunk)
...
    if chunk:
        if self.chunked:
            chunk_len = ('%x\r\n' % len(chunk)).encode('ascii')
            chunk = chunk_len + chunk + b'\r\n'

        self._write(chunk)
```
but
```
async def write_eof(self, chunk=b''):
...
     if chunk:
         self._write(chunk)
```
looks like last chunk has been sent and there is not a single call _on_chunk_sent 

## Your environment

Python 3.6
aiohttp 3.1.1
Windows\Linux

pfreixes:
Yes, looks like a bug.

do you feel strong enough to make a MR with the bugfix and the proper unit test? should be easy.

vasilenkoalexey:
OMG, fixed faster than I was able to answer. Cool, thank you.
I'm not sure how it matters, but at `write` function `self._on_chunk_sent(chunk)` called before `chunk = self._compress.compress(chunk)`, and as i see at fixed `write_eof` function `self._on_chunk_sent(chunk)` called after `chunk = self._compress.compress(chunk)`. Probably the logic should be the same in both functions.

vasilenkoalexey:
Can be closed?

asvetlov:
Yes, sure.
Thanks for the appointment.

PR #2915, opened 2018-04-05
Backport 6a6ab7d and c5d399 to 3.1

pfreixes:
## What do these changes do?

Backports the fix related to the #2909 issue for the 3.1 branch.

Note that due that fix was finally made using two commits the backport has been by hand but trying to keep the nomenclature imposed by cherry_picker /cc @Mariatta


pfreixes:
Releases are automatic?

PR #2925, opened 2018-04-10
Add URL with fragment to ClientResponse

hdk5:
## What do these changes do?

ClientResponse and RequestInfo now have real_url property, which is request url without fragment part being stripped

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -191,8 +191,8 @@ class ClientRequest:
             url2 = url.with_query(params)
             q.extend(url2.query)
             url = url.with_query(q)
-        self.url = url.with_fragment(None)
         self.original_url = url
+        self.url = url.with_fragment(None)
         self.method = method.upper()
         self.chunked = chunked
         self.compress = compress
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -60,6 +60,7 @@ class RequestInfo:
     def real_url_default(self):
         return self.url
 
+
 class Fingerprint:
     HASHFUNC_BY_DIGESTLEN = {
         16: md5,
@@ -248,7 +249,8 @@ class ClientRequest:
 
     @property
     def request_info(self):
-        return RequestInfo(self.url, self.method, self.headers, self.original_url)
+        return RequestInfo(self.url, self.method,
+                           self.headers, self.original_url)
 
     def update_host(self, url):
         """Update destination host, port and connection type (ssl)."""
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -54,7 +54,11 @@ class RequestInfo:
     url = attr.ib(type=URL)
     method = attr.ib(type=str)
     headers = attr.ib(type=CIMultiDictProxy)
+    real_url = attr.ib(type=URL)
 
+    @real_url.default
+    def real_url_default(self):
+        return self.url
 
 class Fingerprint:
     HASHFUNC_BY_DIGESTLEN = {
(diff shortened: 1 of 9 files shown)


hdk5:
I had a trouble retrieving URL fragment from response URL. I am aware that request should be sent without URL fragment, but I don't think that URL from response should be modified.

asvetlov:
HTTP response has no URL (unless it is not 3xx redirection).
`ClientResponse.url` is an URL of request.
I think that keeping a fragment part in response:
1. Makes inconsistency mess.
2. Breaks existing API

hdk5:
> unless it is not 3xx redirection

This is what i meant. I need to know full URL I am getting redirected to because this is how I get application token. 

Here is a little example: https://dry-meadow-80563.herokuapp.com/tokenask

It redirects to `/tokengive#token=hereisyourtoken`, and now it is not possible to get `#token=hereisyourtoken` part from response object.

aiohttp/client_reqrep.py
```diff
@@ -191,8 +191,8 @@ def __init__(self, method, url, *,
             url2 = url.with_query(params)
             q.extend(url2.query)
             url = url.with_query(q)
-        self.url = url.with_fragment(None)
         self.original_url = url
```
hdk5, line 199: Also, this code before my commit looks strange to me:

```python
        self.url = url.with_fragment(None)
        self.original_url = url
```

Both `url` and `original_url` are not modified after (at least by the object itself), so they are always the same. 

And then `original_url` is passed as `url` for ClientResponse. 

So, wasn't it actually intended to keep URL fragment in response object?

asvetlov:
Let's consider `ClientRequest` as *private* API. I don't care too much about its attributes.
The library needs a public client request object but it is another story.

The *public* API is `ClientResponse` and `resp.history` for looking on redirection URLs.
The history is a list of `RequestInfo` data structures.
I propose adding a new attribute called `real_url` (`unstripped_url` or whatever) to save URL with a fragment.

By this, we will keep the behavior of existing code and provide a way for getting original unmodified URLs.

Opinions?

hdk5:
> I propose adding a new attribute called real_url

This might be a solution.

But can you please explain my previous comment on code and [this commit](https://github.com/aio-libs/aiohttp/commit/c34f4e9a2c0ff97db0ad2e45062e3ab6a34fd40c)?

asvetlov:
Looks like @fafhrd91 wanted to implement exactly your initial intention, later we broke the behavior again.
I see two reasons for it:
1. Logical complexity: there is no obvious answer to the question: what is response's URL. I have a feeling that we need two properties.
2. The mentioned commit has no tests/documentation update, that's why we did break it insensibly.

hdk5:
Should I just put real url in `real url` attribute then, like you suggested, and update docs?

asvetlov:
I think we need both `ClientResponse.real_url` and `RequestInfo.real_url`.

Documentation and tests are mandatory.

docs/client_reference.rst
```diff
@@ -1056,6 +1056,10 @@ Response object
 
       URL of request (:class:`~yarl.URL`).
 
+   .. attribute:: real_url
+
+      Unmodified URL of request (:class:`~yarl.URL`).
+
```
asvetlov, line 1065: Please add `.. versionadded:: 3.2` tag

docs/client_reference.rst
```diff
@@ -1450,6 +1454,10 @@ RequestInfo
 
       HTTP headers for request, :class:`multidict.CIMultiDict` instance.
 
+   .. attribute:: real_url
+
+      Requested *url* with URL fragment unstripped, :class:`yarl.URL` instance.
+
```
asvetlov, line 1465: `.. versionadded:: 3.2`

asvetlov:
Thanks!

Issue #2929, opened 2018-04-11
TypeError: server must be web.Application TestServer instance, found type: <class 'aiohttp.web.Application'>


argaen:
## Long story short

Can't create a `TestClient` instance with a `web.Application` instance. According to the docs (https://docs.aiohttp.org/en/stable/testing.html#pytest) and exception message it should be possible but in code aiohttp just checks if its an instance of `BaseTestServer`: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L161

So either the docs or the implementation are wrong. Tried with `aiohttp==3.1.2` (also reproduced with older versions).

## Steps to reproduce

```
Python 3.6.0+ (3.6:1555e7776321, Jun 27 2017, 18:36:56)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from aiohttp import web
ap>>> app = web.Application()
>>> from aiohttp.test_utils import TestClient
>>> TestClient(app)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/manuelmiranda/.pyenv/versions/bb36/lib/python3.6/site-packages/aiohttp/test_utils.py", line 193, in __init__
    "instance, found type: %r" % type(server))
TypeError: server must be web.Application TestServer instance, found type: <class 'aiohttp.web.Application'>
>>>
```

argaen:
I guess the usage has to be `TestClient(TestServer(app))`?

asvetlov:
Thanks for the report.
Documentation should be fixed, `web.Application` should be dropped from the error message.

asvetlov:
Yes, `TestClient(TestServer(app))`

argaen:
You can assign the issue to me. Will provide MR to update docs and exception message today or tomorrow

asvetlov:
Thank you

argaen:
@asvetlov I was double checking `TestClient` docs (https://docs.aiohttp.org/en/stable/testing.html#test-client). They say the following:
```
app_or_server –
BaseTestServer instance for making client requests to it.

If the parameter is aiohttp.web.Application the tool creates TestServer implicitly for serving the application.
```
[...]

PR #2933, opened 2018-04-12
Update docs on how to start test client with app object

argaen:
#2929

PS: I haven't added change entry because its simple doc update. Let me know if you still want it.


argaen:
I've added a change to actual test to check the exception message.

asvetlov:
Thanks

Issue #1821, opened 2017-04-17
Degrading performance over time...


vgoklani:
I wrote a quick AsyncScraper class below:

	import logging, datetime, time
	import aiohttp
	import asyncio
	import uvloop

	# asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())

	logger = logging.getLogger(__name__)
	logger.setLevel(logging.DEBUG)
	logging.basicConfig(format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
	logger.addHandler(logging.StreamHandler())

	class AsyncScraper(object):
		headers = {"User-Agent" : 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/34.0.1847.131 Safari/537.36'}
		def __init__(self, max_connections=1000, timeout=10):
			self.max_connections = max_connections
			self.timeout = timeout
			
		async def get_response(self, url, session):
			with aiohttp.Timeout(timeout=self.timeout):
				async with session.get(url, allow_redirects=True, headers=AsyncScraper.headers, timeout=self.timeout) as response:
					try:
						content = await response.text()
						return {'error': "", 'status': response.status, 'url':url, 'content': content, 'timestamp': str(datetime.datetime.utcnow())}
					except Exception as err:
						return {'error': err, 'status': "", 'url':url, 'content': "", 'timestamp': str(datetime.datetime.utcnow())}
					finally:
						response.close()

		def get_all(self, urls):
			loop = asyncio.get_event_loop()
			with aiohttp.ClientSession(loop=loop, connector=aiohttp.TCPConnector(keepalive_timeout=10, limit=self.max_connections, verify_ssl=False)) as session:
				tasks = asyncio.gather(*[self.get_response(url, session) for url in urls], return_exceptions=True)
				results = loop.run_until_complete(tasks)
				return results

	def chunks(l, n):
		for i in range(0, len(l), n):
			yield l[i:i + n]

	def process_urls(urls, chunk_size=1000):
		scraper = AsyncScraper()

		results = []
		t0 = time.time()
		for i, urls_chunk in enumerate(chunks(sorted(set(urls)), chunk_size)):
			t1 = time.time()
			result = scraper.get_all(urls_chunk)
			success_size = len( [_ for _ in result if ((isinstance(_, Exception) is False) and (_['status']==200)) ] )
			results.extend(result)
[...]

PR #2937, opened 2018-04-13
speed up connector limiting

thehesiod:
see conversation in https://github.com/aio-libs/aiohttp/issues/1821

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -181,6 +181,10 @@ class BaseConnector:
         self._acquired_per_host = defaultdict(set)
         self._keepalive_timeout = keepalive_timeout
         self._force_close = force_close
+
+        # {host_key: FIFO list of waiters}
+        # NOTE: this is not a true FIFO because the true order is lost amongst
+        #       the dictionary keys
         self._waiters = defaultdict(list)
 
         self._loop = loop
@@ -392,11 +396,18 @@ class BaseConnector:
 
             try:
                 await fut
-            finally:
-                # remove a waiter even if it was cancelled
-                waiters.remove(fut)
-                if not waiters:
-                    del self._waiters[key]
+            except BaseException:
+                # remove a waiter even if it was cancelled, normally it's
+                #  removed when it's notified
+                try:
+                    waiters.remove(fut)
+
+                    if not waiters:
+                        del self._waiters[key]
+                except ValueError:  # fut may no longer be in list
+                    pass
+
+                raise
 
             if traces:
                 for trace in traces:
@@ -423,10 +434,8 @@ class BaseConnector:
             except BaseException:
                 # signal to waiter
                 if key in self._waiters:
-                    for waiter in self._waiters[key]:
-                        if not waiter.done():
-                            waiter.set_result(None)
-                            break
+                    waiters = self._waiters[key]
+                    self._release_key_waiter(key, waiters)
                 raise
             finally:
                 if not self._closed:
@@ -470,6 +479,22 @@ class BaseConnector:
         del self._conns[key]
         return None
 
+    def _release_key_waiter(self, key, waiters):
+        if not waiters:
+            return False
+
+        try:
+            waiter = waiters.pop(0)
+            waiter.set_result(None)
+
+            if not waiters:
+                del self._waiters[key]
+
+        except:
+            raise
+
+        return True
+
     def _release_waiter(self):
         # always release only one waiter
 
@@ -477,18 +502,14 @@ class BaseConnector:
             # if we have limit and we have available
             if self._limit - len(self._acquired) > 0:
                 for key, waiters in self._waiters.items():
-                    if waiters:
-                        if not waiters[0].done():
-                            waiters[0].set_result(None)
+                    if self._release_key_waiter(key, waiters):
                         break
 
         elif self._limit_per_host:
             # if we have dont have limit but have limit per host
             # then release first available
             for key, waiters in self._waiters.items():
-                if waiters:
-                    if not waiters[0].done():
-                        waiters[0].set_result(None)
+                if self._release_key_waiter(key, waiters):
                     break
 
     def _release_acquired(self, key, proto):


aiohttp/connector.py
```diff
@@ -392,11 +396,18 @@ def closed(self):
 
             try:
                 await fut
-            finally:
-                # remove a waiter even if it was cancelled
-                waiters.remove(fut)
```
thehesiod, line 397: this is slow because it's trying to find a future in a list for each request that's waiting on a connector

aiohttp/connector.py
```diff
@@ -470,25 +479,33 @@ def _get(self, key):
...
+
     def _release_waiter(self):
         # always release only one waiter
 
         if self._limit:
             # if we have limit and we have available
             if self._limit - len(self._acquired) > 0:
                 for key, waiters in self._waiters.items():
-                    if waiters:
-                        if not waiters[0].done():
```
thehesiod, line 481: with the new model we can guarantee that there are only active waiters in the list so we can greatly simplify this to always popping and notifying the first item in the list

aiohttp/connector.py
```diff
@@ -470,25 +479,33 @@ def _get(self, key):
         del self._conns[key]
         return None
 
+    def _release_key_waiter(self, key, waiters):
+        if not waiters:
+            return False
+
+        waiter = waiters.pop(0)
+        waiter.set_result(None)
```
pfreixes: You can not guarantee that the element is a none canceled future. Even you are removing from the list here [1].

Once the `Future` is set with an exception the callback [2] that will trigger this Exception will be scheduled by the loop, so before the callback that cancels the Task is really executed, the task that was holding a connection and wants to release it might `pop` a canceled future. 

[1] https://github.com/aio-libs/aiohttp/pull/2937/files#diff-7f25afde79f309e2f8722c26cf1f10adR399
[2] https://github.com/python/cpython/blob/master/Lib/asyncio/futures.py#L254

pfreixes: I will mimic this code [1] https://github.com/python/cpython/blob/master/Lib/asyncio/locks.py#L450

aiohttp/connector.py
```diff
@@ -392,11 +396,18 @@ def closed(self):
 
             try:
                 await fut
-            finally:
-                # remove a waiter even if it was cancelled
-                waiters.remove(fut)
-                if not waiters:
-                    del self._waiters[key]
+            except BaseException:
+                # remove a waiter even if it was cancelled, normally it's
```
pfreixes: if the future has been canceled, we do need to wake up another waiter. Take a look at the semaphore implementation [1]

[1] https://github.com/python/cpython/blob/master/Lib/asyncio/locks.py#L478

aiohttp/connector.py
```diff
@@ -181,6 +181,10 @@ def __init__(self, *, keepalive_timeout=sentinel,
         self._acquired_per_host = defaultdict(set)
         self._keepalive_timeout = keepalive_timeout
         self._force_close = force_close
+
+        # {host_key: FIFO list of waiters}
+        # NOTE: this is not a true FIFO because the true order is lost amongst
+        #       the dictionary keys
```
pfreixes: I would suggest using the `deque` [1] data strucutre that is the one used by al of the implementations of `asyncio.locks`. More likely because lists have the following constraint:

> lists incur O(n) memory movement costs for pop(0)


[1] https://github.com/python/cpython/blob/master/Lib/asyncio/locks.py#L437

aiohttp/connector.py
thehesiod: I don't see how this can happen because the async stack ends on the waiter for `Future` and there are no done callbacks associated with it, see this example: https://ideone.com/izIfqw

Think of how this would work, currently no-one sets an exception on this future besides cancelling the waiter at [1], we can verify this by seeing the future is only available from `self._waiters` and no one sets an exception on it.

so, if you cancel task at [1], you have two options:

1) you don't wait on the cancelled task:
  1.1) you can release it (ok, removed from list), and then wait on cancelled (item already gone from list)
  1.2) you can not release it (ok), then wait on cancelled (item will be removed from list)
2) you wait on the cancelled task, in which case the baseexcept runs and removes item from list.  Nothing else can happen because we're just waiting on the future and no-one else is waiting on said future or has callbacks against said future.

aiohttp/connector.py
thehesiod: this method created the future, why would we need to wake up another waiter?  That doesn't make sense as it would imply yet another connector is available.  This is 1-1, one waiter was added, one removed.  Also note that code is if the future was not cancelled, in this scenario it can only be cancelled

aiohttp/connector.py
thehesiod: let me know if I'm missing something in my logic, I know it can get complex

pfreixes: The task will end up once the reactor schedules again the task, the result that is given back to the task depends on the future taht yielded the task before. The cancel is only a signal that wakes up the task with an CanceledError exception.

So the task might be still there just waiting for its turn in the reactor/loop. While its happenning you have chances to have the situation that I told tou.

Please review the asyncio lock code, all of the different implementations take this into consideration.

aiohttp/connector.py
pfreixes: My mistake, the wake up will be done automariaclly by the exit of the context manager in whatever scenario. So forget about this

aiohttp/connector.py
thehesiod: I think what's not clear perhaps, is that when you cancel the task the task is not marked cancelled until the task can process the event, in that case the waiter will be removed. The release cannot happen in this scenario because the removal code is at the bottom of the event chain, it could happen as the cancel bubbles up. Can you give a concrete example of your scenario, I still don't think it's possible.

pfreixes: Better an image than thousand words

```python
import asyncio

waiters = []

async def task(waiter):
    try:
        await waiter
    finally:
        print("task finalitzed")
        if waiter in waiters:
            waiters.remove(waiter)
        else:
            print("waiter not present")

def wakeup_next_waiter():
    if not waiters:
        return

    waiter = waiters.pop(0)
    try:
        waiter.set_result(None)
    except Exception as e:
        print(f"Exception calling set_result {e!r}")
        raise

async def main(loop):
    waiter = loop.create_future()
    waiters.append(waiter)

    # create the task that will wait till either the waiter
    # is finised or the task is cancelled.
    t = asyncio.ensure_future(task(waiter))

    # make a loop iteration to allow the task reach the
    #     await waiter
    await asyncio.sleep(0)

    # put in the loop a callback to wake up the waiter.
    loop.call_soon(wakeup_next_waiter)

    # cancel the task, this will mark the task as cancelled
    # but will be pending a loop iteration to wake up the
    # task, having as a result a CanceledError exception.
    # This implicitly will schedule the Task._wakeup function
    # to be executed in the next loop iteration.
    t.cancel()

    try:
        await t
        pass

```

The output shows how `waiter.set_result(None)` statment raises an exception, the task never had the chance to execute the tear down code that must remove its waiter dependency.

So, the solution is the same applied in many other similiar situations in the asyncio lock implementation, check always [1] if the waiter has been finised or not.

[1] https://github.com/python/cpython/blob/master/Lib/asyncio/locks.py#L450

thehesiod: thanks @pfreixes for that example, so it seems it depends on the ordering of the tasks called during cancellation. Here's a question, can you have an outstanding release for a single connector?  It seems like you could end up releasing two connectors for a single cancel no if the except clause executes before `wakeup_next_waiter`

pfreixes: Not sure about the question, but once the race condition is considered and mitigated the code should work as is expected.

aiohttp/connector.py
```diff
@@ -392,11 +396,18 @@ def closed(self):
 
             try:
                 await fut
-            finally:
-                # remove a waiter even if it was cancelled
-                waiters.remove(fut)
-                if not waiters:
-                    del self._waiters[key]
+            except BaseException:
```
asvetlov: Sorry, I don't understand why the deletion is moved from `finally` to `except`.
Why shouldn't we remove the waiter if **no** exceptions was raised?

aiohttp/connector.py
thehesiod: @pfreixes my "picture" :)

```python
import asyncio

waiters = []


async def task(waiter):
    try:
        await waiter
    except:
        print("task finalized")
        if waiter in waiters:
            print("waiter removed")
            try:
                waiters.remove(waiter)
            except ValueError:
                print("Waiter already removed")
        else:
            print("waiter not present")
        raise


def wakeup_next_waiter():
    if not waiters:
        return

    waiter = waiters.pop(0)
    try:
        if not waiter.done():
            waiter.set_result(None)
    except Exception as e:
        print(f"Exception calling set_result {e!r}")
        raise


async def main(loop):
    # We add two waiters
    waiter = loop.create_future()
    waiters.append(waiter)

    waiter = loop.create_future()
    waiters.append(waiter)

    # create the task that will wait till either the waiter
    # is finished or the task is cancelled.
    t = asyncio.ensure_future(task(waiter))

    # make a loop iteration to allow the task reach the
    #     await waiter
    await asyncio.sleep(0)

    # put in the loop a callback to wake up the waiter.
    loop.call_later(0.1, wakeup_next_waiter)

    # cancel the task, this will mark the task as cancelled
    # but will be pending a loop iteration to wake up the
    # task, having as a result a CanceledError exception.
    # This implicitly will schedule the Task._wakeup function
    # to be executed in the next loop iteration.
    t.cancel()

    try:
        await t
        pass

    # wait for the release to run
    await asyncio.sleep(1)

    # now we have zero waiters even though only one was cancelled
    print(len(waiters))

```

note in this example I add two waiters, cancel one, but at the end none are left because I ensure that the `wakeup_next_waiter`  happens after `task`

aiohttp/connector.py
thehesiod: @asvetlov this is for two reasons: performance, and that if no exception is thrown the removal happened by the release method.

aiohttp/connector.py
pfreixes: I don't see the issue in the code - be carefull with extra ifs that are not needed in thte `task` function. If you are claiming that both waiters are removed this IMO is perfectly fine.

The first waiter that is created is removed through the happy path of the `wakeup_next_waiter`, obviously being removed by `waiter = waiters.pop(0)`, no task is wakened up because you didn't start any task with that waiter. The second waiter is canceled and the task related will remove the waiter by the `waiters.remove(waiter)`. 

Just to put all of us in the same page, the release of a used connection is done automatically by the context manager of a request calling the `resp.release()` [1]. So every time that a code path goes out of the context will release the connection and will try to wake up pending requests.

[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L786

pfreixes:
@thehesiod will be nice if we can get the CI in green, right now tests related to stuff that you are improving are failing.

If the numbers that you claim [1] improve the Aiohttp in such a measure we would need to release ASAP a new version.

So, let's try to focus on resolving the MR comments and move ahead this MR.

[1] https://github.com/aio-libs/aiohttp/issues/1821#issuecomment-381313397

thehesiod:
failures don't seem like my fault:
`AttributeError: module 'asyncio.coroutines' has no attribute 'debug_wrapper'`

pfreixes:
@thehesiod seems that you are still having a linter issue 

```
12.70s$ flake8 aiohttp examples tests demos
tests/test_connector.py:1752:5: F841 local variable 'i' is assigned to but never used
```

aiohttp/connector.py
```diff
@@ -470,25 +479,34 @@ def _get(self, key):
...
+            return False
+
+        waiter = waiters.popleft()
+        if not waiter.done():
+            waiter.set_result(None)
+
+        if not waiters:
+            del self._waiters[key]
+
+        return True
```
cecton, line 491: If you don't do anything with the result, don't implement a result at all, it's like dead code. This is just a small remark

aiohttp/connector.py
```diff
@@ -181,7 +181,11 @@ def __init__(self, *, keepalive_timeout=sentinel,
         self._acquired_per_host = defaultdict(set)
         self._keepalive_timeout = keepalive_timeout
         self._force_close = force_close
-        self._waiters = defaultdict(list)
+
+        # {host_key: FIFO list of waiters}
+        # NOTE: this is not a true FIFO because the true order is lost amongst
+        #       the dictionary keys
```
pfreixes: I guess that this can be removed, the FIFO is for each dictionary key. Different keys mean different tuple values of `(hosts, port)` which FIFO does not make sense.

pfreixes:
@asvetlov this MR is quite critical and once merged will ship a good performance improvement when the client suffers a lot of concurrent connections and starts to apply backpressure, the figures speak by them self

Before
```
limit_builtin:
  Responses: Counter({200: 9999})
  Took:      49.72031831741333 seconds
limit_semaphore:
  Responses: Counter({200: 9999})
  Took:      22.222461223602295 seconds
```

After
```
limit_builtin:
  Responses: Counter({200: 9999})
  Took:      19.374465942382812 seconds
limit_semaphore:
  Responses: Counter({200: 9999})
  Took:      23.060685396194458 seconds
```

Let's keep pushing/discussing the changes that are asked to allow Aiohttp merge at some point this improvement. Also as a critical part of the code, I would like to have a second approval from @asvetlov , mine will be there once the two pending issues that I've commented are solved.

aiohttp/connector.py
thehesiod: each deque is indeed a FIFO as the first one in will be the first one out (amongst items in that list), however across keys it's not a FIFO because it currently iterates across keys (which theoretically can be in any order) when choosing from which FIFO to release.

aiohttp/connector.py
cecton: Ah I see sry

pfreixes:
@thehesiod thanks for the hard work that has been done, let's try to close the opened discussions. 

I don't want to block the PR for just one source code comment, but the exception masked its IMO something that needs to be addressed.

aiohttp/connector.py
```diff
@@ -392,11 +396,18 @@ def closed(self):
...
-            finally:
-                # remove a waiter even if it was cancelled
-                waiters.remove(fut)
-                if not waiters:
-                    del self._waiters[key]
+            except BaseException:
+                # remove a waiter even if it was cancelled, normally it's
+                #  removed when it's notified
+                try:
+                    waiters.remove(fut)
```
cecton, line 401: I just checked and there is no `remove` method in defaultdict.

aiohttp/connector.py
```diff
@@ -392,11 +396,18 @@ def closed(self):
...
+
+                    if not waiters:
+                        del self._waiters[key]
+                except ValueError:  # fut may no longer be in list
+                    pass
```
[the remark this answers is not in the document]
cecton: I agree with @pfreixes , try to scope what you are catching as much as possible.

This is much better imo:

```python
try:
    ...

    if not waiters:
        try:
            del self._waiters[key]
        except ValueError:
            pass

    ...

except:
    ...
```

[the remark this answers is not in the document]
thehesiod: I don't see what you're saying here, `waiters.remove(fut)` throws the `ValueError`, `del self._waiters[key]` could throw a `KeyError`, not another `ValueError`. Not going to change this unless it's really needed.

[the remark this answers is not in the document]
pfreixes: Read my first comment, unless you make it explicit with a `raise from e`, the second try/except masks the original exception in case of ValueError.

thehesiod:
enlightening PR for all of us it seems, mostly me I guess :)  thanks guys

aiohttp/connector.py
```diff
@@ -470,25 +477,34 @@ def _get(self, key):
         del self._conns[key]
         return None
 
+    def _release_key_waiter(self, key, waiters):
+        if not waiters:
+            return False
+
+        waiter = waiters.popleft()
```
pfreixes, line 484: Umm, you have to persist till reaches a waiter not done or you run out of waiters, see the implementation of CPython [1].

[1] https://github.com/python/cpython/blob/master/Lib/asyncio/locks.py#L450

thehesiod: are you saying the old implementation was wrong: https://github.com/aio-libs/aiohttp/pull/2937/files#diff-7f25afde79f309e2f8722c26cf1f10adL481 ?  I don't believe this is the case.  There are two ways a waiter can be removed:

1. An exception happened while waiting (in exception handler)
    1. a release was dispatched for said waiter (someone will see a release)
2. through this method

What you describe would create a double release for 1.i.  This is in fact the scenario you before alluded to

pfreixes: Yes, the issue was already there, indeed I can see the following issues with the code that we have in master:

- The iteration till reach a none canceled waiter has to be done through all of the items of a list, right now is only done on top of the head of each list.
- Each time that we try to release a waiter we have to calculate if the limit and the number of concurrent connections allows us to make it. This is done only in one when the release_waiter is called explicitly but not when we had an exception trying to make the connection.
- The limit per host, TBH, i would say that is not well calculated.

So we have to fix them, but true that they were already there and would be nice if we decouple both things.

thehesiod: ya I have a feeling this is the tip of the iceberg :)  I have a strong suspicious there's a leak in aiohttp or something aiohttp uses as right now we're leaking ~40MB / week in prod

pfreixes:
LGTM, I cant merge it because of the coverage issue. Can you help us @asvetlov?

In any case, let's hold on a new release I would like to work on the issues that I've commented here [1]

[1] https://github.com/aio-libs/aiohttp/pull/2937#discussion_r184523761

asvetlov:
Sure, I can merge it but what prevents from adding new tests for ensuring the full coverage?
@pfreixes if you want to do it in separate PR -- I'm fine with it.

pfreixes:
I will provide a new MR with the fixes and more coverage.

PR #2943, opened 2018-04-17
Raise TooManyRedirects error when client gets redirected too many times

### LINKED ISSUE #2631: There should be an exception in a case of "too many redirects"
## Long story short

When loading `http://relax-tip.de/` it redirects to `http://deepskin.de/`, but `http://deepskin.de/` redirects back to `http://relax-tip.de/`.

aiohttp has built-in limit `max_redirects=10`, but when this limit is reached users get last redirect response and no way to figure-out that is cycle redirect, as well aiohttp follows redirects automatically and it is no expected to get `304` response.

## Expected behaviour

I have an idea, when redirects limit is reached raise `TooManyRedirects` (requests do it)

## Actual behaviour

10 redirect response returned as successful response
 
## Steps to reproduce

```python
import aiohttp
import asyncio


async def main():
        async with session.get('http://relax-tip.de/') as response:
            import ipdb; ipdb.set_trace()
            response


```

## Your environment

client

If such idea makes sense, I'll do pull request to implement this feature

roganov:
Closes #2631

## What do these changes do?
Client now raises TooManyRedirects when gets redirected too many times instead of returning last response.

It is a breaking change.

## Related issue number
#2631 

## Checklist

### DIFF
--- a/CHANGES/2631.feature
+++ b/CHANGES/2631.feature
@@ -1,0 +1,2 @@
+Raise ``TooManyRedirects`` exception when client gets redirected too many times
+instead of returning last response.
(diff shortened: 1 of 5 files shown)


asvetlov:
Thanks

PR #2949, opened 2018-04-22
Request config

asvetlov:
## What do these changes do?

Add `request.config` and `request.config_dict` properties.
Both props do lookup into parent app if the name is not found in sub-application.
`.config` is a namespace (`req.config.key`) but `.config_dict` is a mapping (`req.config_dict['key']`).

Both structures are immutable, to setup config value use `app['key'] = val` on application initialization stage.

The change is backward compatible but two new attributes are added to request instance.
## Related issue number

#2689 
## Checklist


aiohttp/helpers.py
```diff
@@ -741,3 +742,63 @@ def set_result(fut, result):
...
+    def __len__(self):
+        # reuses stored hash values if possible
+        return len(set().union(*self._maps))
+
+    def __iter__(self):
+        d = {}
+        for mapping in reversed(self._maps):
+            # reuses stored hash values if possible
+            d.update(mapping)
+        return iter(d)
```
kxepal, line 777: It feels like you can have there the same `iter(set().union(*self._maps))` instead of another dict construction.

aiohttp/helpers.py
```diff
@@ -741,3 +742,63 @@ def set_result(fut, result):
 def set_exception(fut, exc):
     if not fut.done():
         fut.set_exception(exc)
+
+
+class ChainedProps(Mapping):
```
kxepal: I think the ChainedMaps would be better name here since it's a mapping. Props suffix leaves the feeling that it's something about attr-based access. While ChainedMaps causes collision with stdlib ChainMap, still, they have a bit different naming what should cause suspicions. However, if we make this class for internal-only usage, than own ChainMap is fine.

aiohttp/helpers.py
```diff
@@ -741,3 +742,63 @@ def set_result(fut, result):
...
+
+    def __contains__(self, key):
+        return any(key in m for m in self._maps)
+
+    def __bool__(self):
+        return any(self._maps)
+
+    def __repr__(self):
+        content = ", ".join(map(repr, self._maps))
+        return 'ChainedProps({})'.format(content)
```
kxepal: Here we should reuse class name in repr to avoid any confusions with subclasses. I know, there wouldn't any subclasses of those, but still this little bit could save few debugging hours for some hacker.

aiohttp/web_request.py
```diff
@@ -664,6 +665,18 @@ def app(self):
...
+    def config_dict(self):
+        lst = self._match_info.apps
+        app = self.app
+        idx = lst.index(app)
+        sublist = list(reversed(lst[:idx + 1]))
+        return ChainedProps(sublist)
+
+    @property
+    def config(self):
+        return Namespace(self.config_dict)
```
kxepal: I'm in a very doubt if we need not a mapping class here. This will cause two problems:
1. Inconsistent access to config values. While names are valid Python attributes it's all fine, but how to access to a value with `mycomponent.cache`? Here, user have to fallback to `getattr(request.config, 'mycomponent.cache')`.
2. Linters. Linters will complain about accessing to unknown class attributes. This will be very annoying. 

I would prefer to have one way to work with the config that works and right. So far, `config_dict` is more powerful and useful interface to it.

aiohttp/helpers.py
asvetlov: Naming is the hardest thing!
Maybe we can invite a name without *chain* word?

aiohttp/helpers.py
asvetlov: I don't want to allow subclassing. Hmm, should do it explicitly.

aiohttp/web_request.py
asvetlov: I heard complains many times that IDE doesn't autocomplete string keys but does it for custom attributes.

aiohttp/helpers.py
kxepal: Hm...How about to just a Mapping? With the trick that all the maps will get merged right at the moment of it creation?

kxepal: I mean, if we do merge all the maps, than this class has no reason to exists. The merge result could be returned as MappingProxy - well known immutable mapping.

aiohttp/web_request.py
kxepal: Yes, they don't do keys autocomplete unless you explicitly define those keys in the same scope. However, their ability to autocomplete custom attributes is limited and based on type knowledge and similar usages. 

However, for pylint you will have to put Namespace class into  `ignored-classes` since it will think that it's your fault since you're referencing to unknown class instance attribute.

aiohttp/helpers.py
asvetlov: Let's name it `ChainMapProxy` and keep existing implementation -- I don't like early merging.

aiohttp/web_request.py
asvetlov: @socketpair I recall you wanted the feature.
Please explain your opinion

aiohttp/helpers.py
asvetlov: Let's keep it for sake of shared implementation with `ChainMap`

asvetlov: Iteration is not the most common call, `get`/`__getitem__` doesn't create a dict.
Moreover your proposal constructs a set, not sure what is better.

aiohttp/helpers.py
asvetlov: Done

aiohttp/web_request.py
asvetlov: @Deepwalker ?

asvetlov:
Silence.
@kxepal I suggest dropping `request.config` but keeping `request.config_dict`.
Let's reserve `.config` for future usage (I still doubt about the best API).

asvetlov:
Done

PR #2964, opened 2018-04-28
Fixes some bugs in the limit connection feature

pfreixes:
There are two places which both try to release the next waiter. When the ongoing connection is being canceled calling explicitly to wake up a waiter and when the connection is being released, either because was called explicitly or implicitly due to the close of a response.

In both situations, we call the same method `_release_waiter` that will iterate per all of the
queues till found one that has a waiter that is not already canceled.

This iteration will also check that the limit and the limit per host is not overcome. The logic to check how many available connections/free slots there are has been placed behind the new `_available_connections` function.

So far this PR fixes:

- Take into account the limit and limit per host at any time that we try
to wake up a waiter. No connection available, no wake-up.
- Don't give up till reach a waiter that is not finished.
- Iterate always through all of the different hosts.

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -415,10 +415,10 @@ class BaseConnector:
                 except ValueError:  # fut may no longer be in list
                     pass
 
+                raise e
+            finally:
                 if not waiters:
                     del self._waiters[key]
-
-                raise e
 
             if traces:
                 for trace in traces:
@@ -508,19 +508,11 @@ class BaseConnector:
                 continue
 
             waiters = self._waiters[key]
-
             while waiters:
                 waiter = waiters.popleft()
                 if not waiter.done():
                     waiter.set_result(None)
-
-                    if not waiters:
-                        del self._waiters[key]
-
                     return
-
-            if not waiters:
-                del self._waiters[key]
 
     def _release_acquired(self, key, proto):
         if self._closed:
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import random
 import sys
 import traceback
 import warnings
@@ -494,8 +495,15 @@ class BaseConnector:
         Iterates over all waiters till found one that is not finsihed and
         belongs to a host that has available connections.
         """
+        if not self._waiters:
+            return
 
-        for key in list(self._waiters.keys()):
+        # Having the dict keys ordered this avoids to iterate
+        # at the same order at each call.
+        queues = list(self._waiters.keys())
+        random.shuffle(queues)
+
+        for key in queues:
             if self._available_connections(key) < 1:
                 continue
 
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -414,11 +414,8 @@ class BaseConnector:
                 except ValueError:  # fut may no longer be in list
                     pass
 
-                try:
-                    if not self._waiters[key]:
-                        del self._waiters[key]
-                except KeyError:
-                    pass
+                if key in self._waiters and not self._waiters[key]:
+                    del self._waiters[key]
 
                 raise e
 
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -418,7 +418,11 @@ class BaseConnector:
                 raise e
             finally:
                 if not waiters:
-                    del self._waiters[key]
+                    try:
+                        del self._waiters[key]
+                    except KeyError:
+                        # the key was evicted before.
+                        pass
 
             if traces:
                 for trace in traces:
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -415,7 +415,7 @@ class BaseConnector:
                 except ValueError:  # fut may no longer be in list
                     pass
 
-                if key in self._waiters and not self._waiters[key]:
+                if not waiters:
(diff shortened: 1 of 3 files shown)


pfreixes:
Having the concern of #1821, no degradation perceived with this PR

```bash
imit_semaphore:
  Responses: Counter({200: 9999})
  Took:      20.797451972961426 seconds
limit_builtin:
  Responses: Counter({200: 9999})
  Took:      20.224246501922607 seconds
```

pfreixes:
Also worth it if @fafhrd91 can take a look.

asvetlov:
I'm at a conference, sorry.
Will try to find a time for review tomorrow.

thehesiod:
will try to carve out some time too later today, was working on a bunch of aiobotocore PRs that just landed, am hoping to switch to that and this fix in prod for the perf improvement.

thehesiod:
btw I was thinking, should a "FIFO semaphore" abstraction be refactored from this?  I think it would make the main code clearer.

asvetlov:
@thehesiod What "FIFO semaphore" abstraction is?

aiohttp/connector.py
```diff
@@ -394,18 +407,18 @@ def closed(self):
...
+            except BaseException as e:
                 # remove a waiter even if it was cancelled, normally it's
                 #  removed when it's notified
                 try:
                     waiters.remove(fut)
                 except ValueError:  # fut may no longer be in list
                     pass
 
-                if not waiters:
+                if key in self._waiters and not self._waiters[key]:
```
asvetlov: What is the case when `waiters is not self._waiters[key]`?

aiohttp/connector.py
```diff
@@ -477,35 +490,37 @@ def _get(self, key):
...
+        if not self._waiters:
+            return
 
-        waiter = waiters.popleft()
-        if not waiter.done():
-            waiter.set_result(None)
+        # Having the dict keys ordered this avoids to iterate
+        # at the same order at each call.
+        queues = list(self._waiters.keys())
+        random.shuffle(queues)
```
asvetlov: Why do we need shuffling?
The operation is not free.

aiohttp/connector.py
```diff
@@ -477,35 +490,37 @@ def _get(self, key):
...
+            found = False
+            waiters = self._waiters[key]
 
-    def _release_waiter(self):
-        # always release only one waiter
+            while waiters and not found:
+                waiter = waiters.popleft()
+                if not waiter.done():
+                    waiter.set_result(None)
+                    found = True
```
asvetlov: Isn't `break` enough? It saves a check.

aiohttp/connector.py
pfreixes: Yes, that hurts me. But without the shuffle the iteration of the queues will be done always in the same order,  having chances of starvation for certain queues.

We could try to get rid of that operation, but it will imply some important changes.

pfreixes: The idea of @thehesiod of having a "true" FIFO would get rid of this operation, but my gut feeling says that this FIFO queue will have enough footprint - has to take care of the limit per host - to make it slower than just a call `shuffle`.

PD: a shuffle operation takes around 1 microsecond for a list of 10 elements.

asvetlov: `waiters` is a `deque`.
`deque.rotate()` is cheap. Can we use it?

thehesiod:
@asvetlov basically what this code is trying to do is be a semaphore on the connectors, with the caveat that the waiters are released in order.  So why I said a FIFO Semaphore :)

asvetlov:
@thehesiod got you

aiohttp/connector.py
```diff
@@ -477,35 +490,37 @@ def _get(self, key):
...
-            # then release first available
-            for key, waiters in self._waiters.items():
-                if self._release_key_waiter(key, waiters):
-                    break
+                    if not waiters:
+                        del self._waiters[key]
+
+                    return
+
+            if not waiters:
```
asvetlov: The condition is always true.

pfreixes: Yes it doesn't make sense right now let me get rid of this part of the code and I will check the tests.

aiohttp/connector.py
```diff
@@ -394,19 +407,19 @@ def closed(self):
...
+            except BaseException as e:
                 # remove a waiter even if it was cancelled, normally it's
                 #  removed when it's notified
                 try:
                     waiters.remove(fut)
                 except ValueError:  # fut may no longer be in list
                     pass
 
+                raise e
+            finally:
```
pfreixes, line 419: @asvetlov can you check this new change? My concern with the previous implementation was because having a `del self._waiters[key]` in two places was a tick bomb and hard to maintain.

asvetlov: Looks good

pfreixes:
@asvetlov any clue about how can I increase the coverage?

asvetlov:
I think how everything is perfect.

asvetlov:
thank you, @pfreixes

pfreixes:
Thanks to you, any chance to get a new bugfix release?

thehesiod:
sorry I didn't have a chance to help review, been in a killer time sink project

asvetlov:
@thehesiod no problem.
If you see something -- we can fix it in a separate PR.

Issue #2458, opened 2017-11-01
TCPConnector when family == 0, should sort dns responses to try ipv4 first


hellysmile:
## Long story short

Most widely used scenario is resolving localhost to ipv6, https://github.com/aio-libs/aiohttp/issues/2402 , but it can produce similar issues. aiohttp silently tries all responses from dns response and fails on ipv6 then connection established on ipv4. 

My proposal to sort dns responses by family and use ipv4 first(if family is ipv6 there wouldn't be any ipv4 ip's at all)

## Expected behaviour

ipv4 tried before ipv6

## Actual behaviour

arbitrary order of dns responses tried one by one

webknjaz:
I've got an additional note:

I've faced an issue when Travis's box has disabled IPv6 (detected via `sysctl -a | grep ipv6`), but `localhost` gets resolved to `::1` because of dual entries in `/etc/hosts`.

I think, `aiohttp` could additionally probe availability of INET families and filter out unsupported ones.

asvetlov:
Probing in DNS resolver are too expensive.
On other hand we are trying to connect on all resolved families already.

webknjaz:
I'm talking more about try/except around interfaces domain resolves to, not DNS resolver (we've already got all IPs it resolves into). Maybe some list of available addresses interfaces could give us an idea about IPv6 support, for example.

asvetlov:
How is it different from https://github.com/aio-libs/aiohttp/blob/bff50e6f58307bfdb830d7c1658baa6eb1b3e69a/aiohttp/connector.py#L803-L817 ?

asvetlov:
Happy Eyeball is needed. Python 3.8 is a reasonable limitation I believe.

bdraco:
Happy eyeballs was implemented and it used by default. There should be no need to worry about sorting addresses manually anymore as they happen in the order defined in RFC 8305

PR #2978, opened 2018-05-04
Improve DNS resolution

asvetlov:
In client API if socket family is 0 resolve DNS for both IPv4 and
IPv6, return found IPv4 addresses first.

#2458


asvetlov:
Would somebody review?
I'm going to merge the PR tomorrow in the evening.

aiohttp/resolver.py
```diff
@@ -67,7 +68,8 @@ def __init__(self, loop=None, *args, **kwargs):
     async def resolve(self, host, port=0, family=socket.AF_INET):
         if family == 0:
             hosts = await self._resolve(host, port, socket.AF_INET)
-            hosts.extend(await self._resolve(host, port, socket.AF_INET6))
+            with suppress(OSError):
```
pfreixes: why this only for the `AsyncResolver`?

aiohttp/resolver.py
```diff
@@ -26,6 +27,15 @@ def __init__(self, loop=None):
         self._loop = loop
 
     async def resolve(self, host, port=0, family=socket.AF_INET):
+        if family == 0:
```
pfreixes: What about to use an auto descriptive constant such as `ALL_FAMILIES`? This would help the user.

aiohttp/resolver.py
```diff
@@ -26,6 +27,15 @@ def __init__(self, loop=None):
...
+            hosts = await self._resolve(host, port, family=socket.AF_INET)
+            hosts.extend(await self._resolve(
+                host, port, family=socket.AF_INET6))
```
pfreixes: I have the feeling that both asynchronous calls can be run concurrently ....

aiohttp/resolver.py
```diff
@@ -56,30 +65,41 @@ def __init__(self, loop=None, *args, **kwargs):
...
-        hosts = []
-        for address in resp.addresses:
-            hosts.append(
-                {'hostname': host,
-                 'host': address, 'port': port,
-                 'family': family, 'proto': 0,
-                 'flags': socket.AI_NUMERICHOST})
+        if family == 0:
+            hosts = await self._resolve(host, port, socket.AF_INET)
+            with suppress(OSError):
```
pfreixes: the same as before, we can run them concurrently. Also why in this specific case the `suppress`?

aiohttp/resolver.py
asvetlov: Done. `socket.AF_UNSPEC` has `0` value.

aiohttp/resolver.py
pfreixes: Just found that today [1], regarding the part that this code is trying to achieve the recommendation by the RFC is launch both asynchronously but always first the IPV6 to give it more priority.

In any case, most of the algorithm presented by the RFC is not followed by Aiohttp. At some point we might change the implementation to make it more aligned, I like some of the ideas.

[1] https://tools.ietf.org/html/rfc8305#page-4

asvetlov: Thanks for the link.
I should find a time for a second attempt.

aiohttp/resolver.py
```diff
@@ -26,6 +27,15 @@ def __init__(self, loop=None):
         self._loop = loop
 
     async def resolve(self, host, port=0, family=socket.AF_INET):
+        if family == socket.AF_UNSPEC:
+            hosts = await self._resolve(host, port, family=socket.AF_INET)
```
socketpair, line 31: Use `asyncio.gather` to resolve in parallel. here and below

aiohttp/resolver.py
```diff
@@ -26,6 +27,15 @@ def __init__(self, loop=None):
         self._loop = loop
 
     async def resolve(self, host, port=0, family=socket.AF_INET):
```
socketpair, line 29: 1. Since it is private API, I suggest changing default value for family from AF_INET to AF_UNSPEC.
2. I would prefer IPv6 to be the first if AF_UNSPEC requested.

aiohttp/resolver.py
```diff
@@ -26,6 +27,15 @@ def __init__(self, loop=None):
...
+        if family == socket.AF_UNSPEC:
+            hosts = await self._resolve(host, port, family=socket.AF_INET)
+            hosts.extend(await self._resolve(
+                host, port, family=socket.AF_INET6))
```
webknjaz, line 33: It looks like `AF_UNSPEC` would already return all IPv6 and IPv4 addresses: http://man7.org/linux/man-pages/man3/getaddrinfo.3.html

So what is the point of separate calls for those?

socketpair: Yes, you are right. Seems this hack should be applied only in those resolve methods who don't support resolving both families using one API request.

webknjaz: Oh.. I see: stdlib call is synchronous, but running in the executor https://github.com/python/cpython/blob/bec2372b7e1da5dfdbadaf242aa8e994b164cace/Lib/asyncio/base_events.py#L771

asvetlov: Seems like cares doesn't support AF_UNSPEC

webknjaz: @asvetlov what will happen if the server is bound to `socket.AF_UNIX`, but the client code will call this function with `socket.AF_UNSPEC`, for example (or smth else mismatching)?

aiohttp/resolver.py
webknjaz: +1 for `AF_UNSPEC`

requirements/ci-wheel.txt
```diff
@@ -21,7 +22,6 @@ twine==1.11.0
 yarl==1.1.1
 
 # Using PEP 508 env markers to control dependency on runtimes:
-aiodns==1.1.1; platform_system!="Windows"  # required c-ares will not build on windows
```
webknjaz, line 24: Why are you decommissioning it?

aiohttp/resolver.py
webknjaz: @asvetlov "cares doesn't support AF_UNSPEC "?

I'm not sure I've got what you mean here. Could you please add more detail?

requirements/ci-wheel.txt
asvetlov: aiodns **does** work on Windows (at least appveyor tests are passed)

aiohttp/resolver.py
asvetlov: The resolver is not used for unix sockets.
async resolver doesn't support AF_UNSPEC, it requires two explicit calls for DNS resolving -- one for IPv4 and another one for IPv6.
At least I've tried to do it in the single call but failed

requirements/ci-wheel.txt
webknjaz: But why remove the whole of it, not just conditional?

aiohttp/resolver.py
webknjaz: @asvetlov according to https://github.com/python/cpython/blob/bec2372b7e1da5dfdbadaf242aa8e994b164cace/Lib/asyncio/base_events.py#L771 it just bypasses all args to underlying synchronous `socket.getaddrinfo`, which supports `UNSPEC` https://github.com/python/cpython/blob/master/Modules/socketmodule.c#L6189 and bypasses that into underlying syscall.

The only thing I can think of if a small possibility of behaving differently in non-UNIX envs, but anyway I don't understand how could it fail for you. Can you please post a traceback here?

NotoriousRebel:
@asvetlov Any updates on this?

asvetlov:
Happy Eyeballs is the right solution for the problem, not this fix.

PR #2983, opened 2018-05-07
Forbid reading response BODY after release

asvetlov:
Now the following code produces unspecified behavior:
```
      pass
await resp.read()
```
The same for reading from `resp.content`.

If the response has a data into the internal buffer -- the data is returned. Otherwise, an exception is raised.
Reading from the released response is a logical error, the PR raises an exception unconditionally in the case.

### DIFF
--- a/CHANGES/2983.feature
+++ b/CHANGES/2983.feature
@@ -1,1 +1,1 @@
-Disable reading response BODY after release
+Forbid reading response BODY after release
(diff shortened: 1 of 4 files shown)


asvetlov:
The fix may break incorrect existing code but I believe that better to fail early than keeping existing inconsistent behavior.

kxepal:
I would you stronger word like Forbid instead of Disable. But :+1: in any case.

asvetlov:
Done

PR #2985, opened 2018-05-08
Don't reuse a connection with the same URL but different proxy/TLS settings

### LINKED ISSUE #2981: proxy behavior is inconsistent
## Long story short

request proxy behavior is inconsistent within one session:

```Python
import asyncio

import aiohttp


class ProxyCrawler:
    def __init__(self):
        self.session = aiohttp.ClientSession(loop=loop)

    async def req(self):
        async with self.session.get('http://httpbin.org/ip') as resp:
            text = await resp.text()
            print('First request:')
            print(text)

        async with self.session.get('http://httpbin.org/ip', proxy='http://127.0.0.1:1087') as resp:
            text = await resp.text()
            print('Second request:')
            print(text)

        await self.session.close()


if __name__ == '__main__':
    c = ProxyCrawler()
    loop.run_until_complete(c.req())

```

## Expected behaviour

The first request goes directly, and the second request goes through the specified proxy.

## Actual behaviour

Both requests go directly without proxy.

( But if you set the first request go through a proxy, the second request will go through the specified proxy correctly. )

## Steps to reproduce

Run the code above ( replace the proxy with yours ).

## Your environment

aiohttp: 3.2.0
Mac OS X

asvetlov:
Fixes #2981

### DIFF
--- a/CHANGES/2981.bugfix
+++ b/CHANGES/2981.bugfix
@@ -1,0 +1,1 @@
+Don't reuse a connection with the same URL but different proxy/TLS settings
--- a/aiohttp/client_exceptions.py
+++ b/aiohttp/client_exceptions.py
@@ -241,7 +241,7 @@ class ClientConnectorCertificateError(*certificate_errors_bases):
 
     @property
     def ssl(self):
-        return self._conn_key.ssl
+        return self._conn_key.is_ssl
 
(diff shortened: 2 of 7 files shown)


asvetlov:
Ready for review

PR #2986, opened 2018-05-08
Implement base protocol class

asvetlov:
To avoid a dependency from internal `asyncio.streams.FlowControlMixin`

### DIFF
--- a/aiohttp/base_protocol.py
+++ b/aiohttp/base_protocol.py
@@ -1,0 +1,64 @@
+import asyncio
+
+from .log import internal_logger
+
+
+class BaseProtocol(asyncio.Protocol):
+    def __init__(self, loop=None):
+        if loop is None:
+            self._loop = asyncio.get_event_loop()
+        else:
+            self._loop = loop
+        self._paused = False
+        self._drain_waiter = None
+        self._connection_lost = False
+        self.transport = None
+
+    def pause_writing(self):
+        assert not self._paused
+        self._paused = True
+        if self._loop.get_debug():
+            internal_logger.debug("%r pauses writing", self)
+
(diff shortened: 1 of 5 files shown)


kxepal:
> To avoid a dependency from internal asyncio.streams.FlowControlMixin

Why to avoid this? Do you expect that internals will change in some breaking way soon?

asvetlov:
The class is not a part of public API. Technically it should be prefixed by underscore but we've missed it.

The second reason is type hinting: we have a public `.transport` property and expose our protocols in our public API (not sure if we should do it but the change can be done in 4.0 only).
Marking a property as "returning `asyncio.Protocol`" crashes mypy.

kxepal:
Got it. Thanks!

aiohttp/base_protocol.py
```diff
@@ -0,0 +1,64 @@
...
+class BaseProtocol(asyncio.Protocol):
+    def __init__(self, loop=None):
+        if loop is None:
+            self._loop = asyncio.get_event_loop()
+        else:
+            self._loop = loop
+        self._paused = False
+        self._drain_waiter = None
+        self._connection_lost = False
+        self.transport = None
```
kxepal, line 15: Do we want transport as public attribute or instead public read-only property? I think transport is only valid to set via `connection_made` method, not somehow adhoc'y.

asvetlov: I'm trying don't touch transport attr, many aiohttp codes use it internally.
Tests modify `.transport` wildly as well.
Better to deprecate `.protocol` at all in our public API.

kxepal: :+1:

PR #2989, opened 2018-05-10
Raise ConnectionReserError instead of CancelledError

asvetlov:
on trying to write to a closed stream.


CHANGES/2499.feature
```diff
@@ -0,0 +1,2 @@
+Raise ``ConnectionReserError`` instead of ``CancelledError`` on trying
```
kxepal: `ConnectionResetError`

asvetlov: Fixed, thank you

PR #2990, opened 2018-05-10
Travis CI migration and packaging improvements

webknjaz:
## What do these changes do?
* port Travis CI config changes from multidict
* CI config polishing
* CI links update
* port dist setup changes
* other packaging and CI related stuff
* all changes are put into atomic commits making it possible to rebase-merge that as is

‒

## Related issue number
‒

## Checklist


setup.py
```diff
@@ -120,16 +118,28 @@ def run(self):
...
-    url='https://github.com/aio-libs/aiohttp/',
+    url=repo_url,
+    project_urls={
+        'Chat: Gitter': 'https://gitter.im/aio-libs/Lobby',
+        'CI: AppVeyor': 'https://ci.appveyor.com/project/{}'.format(appveyor_slug),
+        'CI: Circle': 'https://circleci.com/gh/{}'.format(repo_slug),
+        'CI: Shippable': 'https://app.shippable.com/github/{}'.format(repo_slug),
+        'CI: Travis': 'https://travis-ci.com/{}'.format(repo_slug),
+        'Coverage: codecov': 'https://codecov.io/github/{}'.format(repo_slug),
+        'Docs: RTD': 'https://{}.readthedocs.io'.format(name),
```
asvetlov: It is `http://docs.aiohttp.org`

setup.py
```diff
@@ -83,21 +82,19 @@ def read(f):
...
-        errno = subprocess.call([sys.executable, '-m', 'pytest', 'tests'])
-        raise SystemExit(errno)
-
-
-tests_require = install_requires + ['pytest', 'gunicorn',
-                                    'pytest-timeout', 'async-generator']
+tests_require = ['pytest', 'gunicorn',
+                 'pytest-timeout', 'async-generator']
 
+name = 'aiohttp'
```
asvetlov: I don't like `name` substitution.
A plain constant is more obvious than string-formatted, isn't it?

webknjaz: I don't want to create code diversity, which often leads to forgetting to update some of places. Also, I use this approach in lots other distributions including multidict, which proved to be useful.

setup.py
```diff
@@ -83,21 +82,19 @@ def read(f):
...
+appveyor_slug = 'asvetlov/{}'.format(name)  # FIXME: move under aio-libs/* slug
```
webknjaz: @asvetlov how about finally fixing AppVeyor URL?

setup.py
webknjaz: *https

setup.py
asvetlov: I've approved it for multidict but now I'm inclining to revert bare strings back.
It is written once but read many times.
Every time when I see the line I should remember what *name* is.
You've convinced me that we need overcomplicated travis config. Perhaps it is true, I'm not Travis expert.
But for `setup.py` I pretty sure that the simplest approach is the best.

setup.py
asvetlov: I don't remember how to setup it.
Could you do it yourself?

webknjaz: This requires owner privileges AFAIR, so I can't. It's a bit tricky, but the bottom line is that you create a team there mapped to the team in GitHub and give it some access there. Also you need to somehow create an org link there.

webknjaz: Here you go: https://www.appveyor.com/docs/team-setup/#setting-up-appveyor-account-for-github-organization
Create a separate account called `aio-libs` and then add your personal account as its administrator. You can also create team mapped to some team in GitHub, allowing its members to cancel/rerun builds.

setup.py
webknjaz: I'd agree if we were using declarative `setup.cfg`, but this is code here.

webknjaz: By the way I have an 80-line shim snippet for supporting newer setuptools, which would allow us to move a vast of meta/options to `setup.cfg`

asvetlov: Say again I prefer constants to string substitutions in `setup.py`

webknjaz: ok, as you wish. but how about moving that to `setup.cfg`?

asvetlov: If you want to migrate -- I'm ok probably but want to see how it looks like first.
Could you point on another project with `setup.cfg` integration?

webknjaz: AFAIR I only tried this in commercial ones, but I also tested out a shim, which makes it possible to support this in old versions of setuptools as well.

webknjaz: Here's how it would've looked in ansible if we accepted the patch: https://github.com/ansible/ansible/pull/38413/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R1

webknjaz:
@asvetlov please remove this deprecated integration `continuous-integration/travis-ci/push Pending — The Travis CI build is in progress Required Details` and make new `Travis CI - Pull Request In progress — Build Started` required instead.

webknjaz:
Sure, thanks :)

webknjaz:
I'm still seeing it @ https://github.com/aio-libs/aiohttp/settings/installations.
Will remove it myself.

webknjaz:
@asvetlov can we accept it now?

.appveyor.yml
```diff
@@ -14,10 +14,10 @@ install:
 build: false
 
 test_script:
-  - "tools/build.cmd %PYTHON%\\python.exe setup.py test"
+  - "tools/build.cmd %PYTHON%\\python.exe -m pytest tests"
```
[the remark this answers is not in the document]
asvetlov: Ok, I see

.appveyor.yml
```diff
@@ -14,10 +14,10 @@ install:
...
 
 after_test:
-  - "tools/build.cmd %PYTHON%\\python.exe setup.py sdist bdist_wheel"
+  - "tools/build.cmd %PYTHON%\\python.exe -m setup sdist bdist_wheel"
```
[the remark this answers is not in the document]
asvetlov: I know I can but it looks confusing. At least for me.

.travis.yml
```diff
@@ -1,3 +1,5 @@
+version: "= 0"
```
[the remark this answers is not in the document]
asvetlov: Nice! Good to know.

setup.py
kxepal: +1 to move all to setup.cfg

.travis.yml
```diff
@@ -136,20 +136,25 @@ jobs:
   include:
   - <<: *_doc_base
     env:
-    - docs
+    - docs <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- <- =
```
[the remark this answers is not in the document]
asvetlov: FIxed

setup.py
webknjaz: I know how to do it, it's just not going to be in the same PR.

webknjaz: Also, there's certain complications related to this change. So let's not rush.

.appveyor.yml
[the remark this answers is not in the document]
webknjaz: It might look confusing for someone who is not used to it. It works though, I'm using it all the time now.

PR #2993, opened 2018-05-11
Add path_pattern property to routes and resources

### LINKED ISSUE #2968: Add canonical name for the route
## Long story short

For third party libraries sometimes it is useful to know canonical name of the route from request, for example https://github.com/aio-libs/aiozipkin/issues/115. 

Something  like `/foo/bar/{name:\d+}` could be exposed with property:  `request.match_info.route.canonical_name`

pposca:
The path_pattern is the raw path used to add a new route. For example,
/foo/bar/{name:\d+}.

 - Add path_pattern property to AbstractRoute and AbstractResource
 - Add path pattern implementation to PlainResource, DynamicResource,
 PrefixResource, StaticResource, ResourceRoute and SystemRoute.
 - Add tests

Closes #2968

## What do these changes do?

Add path_pattern property to routes and resources.
The path_pattern is the raw path used to add a new route. For example, /bar/{name:\d+}.

The public API of Route and Resource has now a new property path_name to expose directly the raw path used to add a new route. Significant values have been selected for each implementation of the abstract classes.

## Related issue number

2968

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -45,6 +45,15 @@ class AbstractResource(Sized, Iterable):
     def name(self):
         return self._name
 
+    @property
+    @abc.abstractmethod
+    def canonical(self):
+        """Exposes the resource's canonical path.
+
+        For example '/foo/bar/{name}'
+
+        """
+
     @abc.abstractmethod  # pragma: no branch
     def url_for(self, **kwargs):
         """Construct url for resource with additional params."""
@@ -300,6 +309,10 @@ class PlainResource(Resource):
         assert not path or path.startswith('/')
         self._path = path
 
+    @property
+    def canonical(self):
+        return self._path
+
     def freeze(self):
         if not self._path:
             self._path = '/'
@@ -373,6 +386,10 @@ class DynamicResource(Resource):
         self._pattern = compiled
         self._formatter = formatter
 
+    @property
+    def canonical(self):
+        return self._formatter
+
     def add_prefix(self, prefix):
         assert prefix.startswith('/')
         assert not prefix.endswith('/')
@@ -413,6 +430,10 @@ class PrefixResource(AbstractResource):
         assert prefix in ('', '/') or not prefix.endswith('/'), prefix
         super().__init__(name=name)
         self._prefix = URL.build(path=prefix).raw_path
+
+    @property
+    def canonical(self):
+        return self._prefix
 
     def add_prefix(self, prefix):
         assert prefix.startswith('/')
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -1823,6 +1823,13 @@ Resource classes hierarchy::
 
       Read-only *name* of resource or ``None``.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. For example
+      ``/path/to`` or ``/path/{to}``
+
+      .. versionadded:: 3.3
+
    .. comethod:: resolve(request)
 
       Resolve resource by finding appropriate :term:`web-handler` for
@@ -1886,6 +1893,12 @@ Resource classes hierarchy::
    The class corresponds to resources with plain-text matching,
    ``'/path/to'`` for example.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the path
+      used to create the PlainResource. For example ``/path/to``
+
+      .. versionadded:: 3.3
 
    .. method:: url_for()
 
@@ -1900,6 +1913,13 @@ Resource classes hierarchy::
    :ref:`variable <aiohttp-web-variable-handler>` matching,
    e.g. ``'/path/{to}/{param}'`` etc.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the
+      formatter obtained from the path used to create the DynamicResource.
+      For example, from a path ``/get/{num:^\d+}``, it returns ``/get/{num}``
+
+      .. versionadded:: 3.3
 
    .. method:: url_for(**params)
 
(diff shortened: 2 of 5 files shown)


asvetlov:
Thank for the PR.
We have 2 options: use regex pattern or formatter string.
The first is stricter, the second looks nicer: `/foo/bar/{name}`.

I have no personal opinion.
@aio-libs/aiohttp-committers what do you think?

kxepal:
I'm for formatted string, or, actually, placeholders.

Regexps are more powerful, but with their power they move part of data validation on router side. Say, `/user/:id` vs `/user/{id:\d+}` - placeholder knowns nothing about what ids are valid and delegates their check to handler side. Regular expression tries to ensure that they are numeric ones. This may sounds sweet, but you still would do `int(id)` in handler and if that id is not valid you still would like to throw some exception back.

Also, I would like to note, that `/foo/bar/{baz}` syntax not quite common. The `/foo/bar/:baz` would be a bit more wide used.

pposca:
As far as I've understood, regex patterns only exist in DynamicResource. If other resources return a string and DynamicResource returns a regex, it looks a bit like a broken API, inconsistent. It seems more logic to me that all resources return a string.

On the other hand, I think regex is more useful than formatted string, as `/user/{id:\d+}` provides more information than `/user/{id}`, and this is why I chose to return exactly the same string it is used when calling add_routes., whether it is a normal string (PlainResource) or a regex pattern string (DynamicResource)

PD: Sorry if I've misunderstood something. It is the first time I dive into aiohttp code.

pposca:
@asvetlov , do you prefer a 2nd commit with the update or that I amend the first commit?

aiohttp/web_urldispatcher.py
```diff
@@ -131,6 +140,15 @@ def handler(self):
     def name(self):
         """Optional route's name, always equals to resource's name."""
 
+    @property
+    @abc.abstractmethod
+    def canonical(self):
```
asvetlov: Should we have `route.canonical`?
`route.resource.canonical` should be enough.
`route.name` exists for historical reasons, routes were added before the resource invention.

pposca: The issue was asking for a route property, that's why I added it. 

I will change it and I will update the docs (I completely forgot them...)

aiohttp/web_urldispatcher.py
```diff
@@ -457,6 +478,10 @@ def __init__(self, prefix, directory, *, name=None,
                         'HEAD': ResourceRoute('HEAD', self._handle, self,
                                               expect_handler=expect_handler)}
 
+    @property
+    def canonical(self):
+        return self._prefix + str(self._directory)
```
asvetlov: Looks like there is an error here.
Canonical URL should contain the prefix only, the directory is a *local* path.
It never returns in URLs but `url_for()` constructs a return value from the prefix and a *relative* to the directory path.

pposca: Ok, I will remove it from StaticResource, as canonical is already the prefix in PrefixResource and StaticResource will inherit the property.

docs/web_reference.rst
```diff
@@ -1823,6 +1823,11 @@ Resource classes hierarchy::
 
       Read-only *name* of resource or ``None``.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. For example
+      ``/path/to`` or ``/path/{to}``
+
```
asvetlov, line 1830: `.. versionadded:: 3.3`

docs/web_reference.rst
```diff
@@ -1886,6 +1891,10 @@ Resource classes hierarchy::
    The class corresponds to resources with plain-text matching,
    ``'/path/to'`` for example.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the path
+      used to create the PlainResource. For example ``/path/to``
```
asvetlov, line 1899: `versionadded`

docs/web_reference.rst
```diff
@@ -1900,6 +1909,11 @@ Resource classes hierarchy::
    :ref:`variable <aiohttp-web-variable-handler>` matching,
    e.g. ``'/path/{to}/{param}'`` etc.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the
+      formatter obtained from the path used to create the DynamicResource.
+      For example, from a path ``/get/{num:^\d+}``, it returns ``/get/{num}``
```
asvetlov, line 1920: `versionadded`

docs/web_reference.rst
```diff
@@ -1918,6 +1932,11 @@ Resource classes hierarchy::
    The class corresponds to resources for :ref:`static file serving
    <aiohttp-web-static-file-handling>`.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the prefix
+      used to create the StaticResource. For example ``/prefix``
+
```
asvetlov, line 1945: `versionadded`

docs/web_reference.rst
```diff
@@ -1945,6 +1964,12 @@ Resource classes hierarchy::
    A resource for serving nested applications. The class instance is
    returned by :class:`~aiohttp.web.Application.add_subapp` call.
 
+   .. attribute:: canonical
+
+      Read-only *canonical path* associate with the resource. Returns the
+      prefix used to create the PrefixedSubAppResource.
+      For example ``/prefix``
+
```
asvetlov, line 1980: `versionadded`

asvetlov:
Well done!
Thank you for patience.

pposca:
My pleasure. Really happy to help.

jettify:
Thanks @pposca @asvetlov ! I have already added this property to aiozipkin integration (https://github.com/aio-libs/aiozipkin/pull/138)

PR #2995, opened 2018-05-13
Cythonize reify

asvetlov:
Another 5% boost on simple aiohttp web server

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,8 @@
 .vimrc
 aiohttp/_frozenlist.c
 aiohttp/_frozenlist.html
+aiohttp/_helpers.c
+aiohttp/_helpers.html
 aiohttp/_websocket.c
 aiohttp/_websocket.html
 aiohttp/_http_parser.c
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ script:
 - make cov-ci-run
 
 after_success:
-- codecov
+ # - codecov
+ - bash <(curl -s https://codecov.io/bash)
 
 _helpers:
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -23,6 +23,8 @@
 .vimrc
 aiohttp/_frozenlist.c
 aiohttp/_frozenlist.html
+aiohttp/_helpers.c
+aiohttp/_helpers.html
```
webknjaz, line 27: @asvetlov it looks like you forgot to update MANIFEST.in.
I wish we could just use a setuptools plugin, which would rely on git to do this automatically.

.travis.yml
```diff
@@ -19,7 +19,8 @@ script:
 - make cov-ci-run
 
 after_success:
-- codecov
+ # - codecov
+ - bash <(curl -s https://codecov.io/bash)
```
webknjaz, line 23: why do this? we install it in CI pinned to certain version:
https://github.com/aio-libs/aiohttp/blob/master/requirements/ci-wheel.txt#L25

.gitignore
asvetlov: `*.c` files should be included but `*.html` should not.
Not very big problem though: on travis html files are never generated.

.travis.yml
asvetlov: It was my try to fix coverage reports.
Now upload on codecov is failed (with both codecov plugin and bash script).
Any ideas?

.gitignore
asvetlov: Fixed by ffa4532e

PR #3021, opened 2018-05-22
Deprecate custom router

asvetlov:
A part of #2766

### DIFF
--- a/CHANGES/3021.removal
+++ b/CHANGES/3021.removal
@@ -1,0 +1,1 @@
+Deprecate custom application's router.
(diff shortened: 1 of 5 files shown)


vangheem:
Why is router deprecated? What should we do to replace this?

vangheem:
Nevermind, I've read https://github.com/aio-libs/aiohttp/issues/2766

Just FYI, https://github.com/plone/guillotina is using a traversal router right now. We'll move to using default `UrlDispatcher` with a wildcard resource match.

PR #3037, opened 2018-05-30
Implement runner.addresses property

### LINKED ISSUE #3036: Access to socket information when using AppRunner
## Long story short

Sometimes I like to listen on port 0 to allow for any and need to access the actual port after binding.

## Expected behaviour

A public way to access socket information like the port number.

## Actual behaviour

I have to reach into the private `TCPSite._server`.

asvetlov:
Fix #3036

### DIFF
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -148,11 +148,17 @@ class BaseRunner(ABC):
         self._handle_signals = handle_signals
         self._kwargs = kwargs
         self._server = None
-        self._sites = set()
+        self._sites = []
 
     @property
     def server(self):
         return self._server
+
+    @property
+    def addresses(self):
+        return [sock.getsockname()
+                for site in self._sites
+                for sock in site._server.sockets]
 
     @property
     def sites(self):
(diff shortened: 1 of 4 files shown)


asvetlov:
`IPEndpoint(ipaddress, int)` or `UNIXEndpoint(str)`?

asvetlov:
Hmm. IPv6 has additional `flow_info` and `scope_id`.
I have not enough information how people use it but suspect keeping everything is essential.
My motivation is: let's support the same naming structures as well-known `socket` module do.
Later we can invent a better API by adding `.endpoints` property :)

aiohttp/web_runner.py
```diff
@@ -148,12 +148,18 @@ def __init__(self, *, handle_signals=False, **kwargs):
         self._handle_signals = handle_signals
         self._kwargs = kwargs
         self._server = None
-        self._sites = set()
+        self._sites = []
```
webknjaz, line 151: @asvetlov what is the motivation behind this?

asvetlov: To keep a stable order of returned addresses.
Registering a new site in a runner is not a bottleneck, we can live with O(n) for it.

PR #3040, opened 2018-05-31
3040 Typo

da4089:
Fix for an apparent typo.


asvetlov:
Thanks!
Will merge when green.

PR #3041, opened 2018-05-31
Remove heartbeat on closing connection on keepalive timeout

### LINKED ISSUE #2782: Unsolicited response received on idle HTTP channel starting with "\r\n"; err=<nil>
## Long story short

By default, aiohttp closes keep-alive connections after 75 seconds of inactivity. Just before closing it, server sends `\r\n` data to client in order to notify it about closing. However, not all the clients are happy with that. One of those is standard golang `net/http` which logs errors in this case not about closed connection, but about bad data received in place when there should be any of such.

Referenced commit: https://github.com/aio-libs/aiohttp/commit/bf49f8e75fb263113d6a448a532293107d2e03bc

Question is that behaviour correct or common, or based on some RFC? 
What recommendations could we give to clients which complains about to relax their communication with aiohttp servers?

## Expected behaviour

```
2018/03/01 13:52:12 Hello, Anonymous
```

## Actual behaviour

```
2018/03/01 13:52:12 Hello, Anonymous
2018/03/01 13:53:27 Unsolicited response received on idle HTTP channel starting with "\r\n"; err=<nil>
```

## Steps to reproduce

1. Run any aiohttp example. Let's take `examples/web_srv.py`
2. Save the following as `main.go`, for example:
```
package main

import (
    "fmt"
    "io/ioutil"
    "log"
    "net/http"
    "net/url"
    "time"
)

func main() {
    Url, err := url.Parse("http://localhost:8080/hello")
    if err != nil {
        log.Panic(err)
        return
    }

    client := http.Client{
        Timeout: time.Duration(time.Millisecond * 10000),
    }

    resp, err := client.Get(Url.String())
    if err != nil {
        log.Panic(err)
        return
    }

    body, err := ioutil.ReadAll(resp.Body)
    resp.Body.Close()

    log.Print(string(body))
    time.Sleep(100 * time.Second)
}
```
3. Run it as `go run main.go`
4. Wait for awhile and see the error from subject.

## Your environment

1. Python 3.6 / aiohttp 3.0.5
2. go 1.9.1

asvetlov:
Revert  bf49f8e75fb263113d6a448a532293107d2e03bc
The used hack violates HTTP protocol.

Fixes #2782

### DIFF
--- a/CHANGES/3041.feature
+++ b/CHANGES/3041.feature
@@ -1,0 +1,2 @@
+Remove heartbeat on closing connection on keepalive timeout.
+The used hack violates HTTP protocol.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -286,14 +286,12 @@ class RequestHandler(BaseProtocol):
         if self._waiter:
             self._waiter.cancel()
 
-    def force_close(self, send_last_heartbeat=False):
+    def force_close(self):
         """Force close connection"""
         self._force_close = True
         if self._waiter:
             self._waiter.cancel()
         if self.transport is not None:
-            if send_last_heartbeat:
-                self.transport.write(b"\r\n")
             self.transport.close()
             self.transport = None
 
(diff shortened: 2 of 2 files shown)


kxepal:
Thanks!

PR #3051, opened 2018-06-04
Do not use translated path

william-gr:
## What do these changes do?

Allow redirect to work when using url encoded paths and normalize path middleware, e.g.

/foo/bar%2Fbar will redirect to /foo/bar%2Fbar/ instead of /foo/bar/bar/

Not as far as I am aware, since its a bug fix.

### DIFF
--- a/aiohttp/web_middlewares.py
+++ b/aiohttp/web_middlewares.py
@@ -73,7 +73,7 @@ def normalize_path_middleware(
                 resolves, request = await _check_request_resolves(
                     request, path)
                 if resolves:
-                    raise redirect_class(request.path + query)
+                    raise redirect_class(path + query)
 
         return await handler(request)
 
--- a/aiohttp/web_middlewares.py
+++ b/aiohttp/web_middlewares.py
@@ -73,7 +73,7 @@ def normalize_path_middleware(
                 resolves, request = await _check_request_resolves(
                     request, path)
                 if resolves:
-                    raise redirect_class(path + query)
+                    raise redirect_class(request.raw_path)
 
         return await handler(request)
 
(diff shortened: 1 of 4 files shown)


webknjaz:
Please fill out the PR template, so that it would be easier to understand why and what you are fixing.

asvetlov:
Sorry, I don't follow what is failed and what is expected behavior.

Unit test and changelog entry are required.

webknjaz:
https://raw.githubusercontent.com/aio-libs/aiohttp/master/.github/PULL_REQUEST_TEMPLATE.md

william-gr:
@asvetlov @webknjaz PR updated and summary changed to follow template.

Let me know if something is missing.

Thanks

william-gr:
Ping. Just want to make sure there isn't anything missing here. I will be travelling the next few days and want to leave it ready for a review.

aiohttp/web_middlewares.py
```diff
@@ -73,7 +73,7 @@ def normalize_path_middleware(
                 resolves, request = await _check_request_resolves(
                     request, path)
                 if resolves:
-                    raise redirect_class(request.path + query)
```
asvetlov, line 76: I wonder if `request.path` can be replaced with `request.raw_path`

william-gr: Thats the same as `path`, no? request.raw_path is being used above to populate paths_to_check.

asvetlov: No. It is the percent encoded *canonical* path representation.

william-gr: And thats exactly the issue I am solving here, and its passing all the tests so I dont get it.

william-gr:
@asvetlov call updated. Thanks for the helpful explanation as of why.

asvetlov:
thanks

PR #3054, opened 2018-06-05
`Line too long` error when proxying a request

arthurdarcet:
When sending a request with a body set to a `StreamReader` (ie when your aiohttp server proxy a request to another server), the body could contain very long lines.

Such line go over the high water mark, and trigger an exception.
The `payload` module should not write the `StreamReader` by (async)-iterating over the lines, but should instead iterate over whatever is available. This patch implements that.

Here is a small script reproducing the error:
```python
import aiohttp
import asyncio


async def handle(request):
	return aiohttp.web.Response(body=b'0' * 2 ** 18)

async def main(loop):
	app.add_routes([aiohttp.web.get('/', handle)])
	handler = app.make_handler()
	await loop.create_server(handler, '0.0.0.0', 1111)

		async with session.get('http://0.0.0.0:1111') as resp:
			# this is just a stream of 2**18 0-bytes
			stream = resp.content

			async with session.post('http://httpbin.org/anything', data=stream):
				pass


```

### DIFF
--- a/CHANGES/3054.bugfix
+++ b/CHANGES/3054.bugfix
@@ -1,0 +1,1 @@
+When using a server-request body as the `data=` argument of a client request, iterate over the content with `readany` instead of `readline` to avoid `Line too long` errors.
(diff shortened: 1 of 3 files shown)


asvetlov:
Thanks!

arthurdarcet:
Thank you for the release!

asvetlov:
Welcome!

PR #3058, opened 2018-06-06
Do not use async_timeout 3.0.0 on the 2.3 branch

romuald:
## What do these changes do?

This change "pins" the async_timeout dependency to 2.x

Because async_timeout >= 3.0 dropped python3.4, not pinning this will break on python 3.4

I do not know about aiohttp policies for the 2.3 branch, so this may not be the correct way to do this 

Also this will since this will pin to 2.x even on python 3.5+, I may add a version check if you wish

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ with codecs.open(os.path.join(os.path.abspath(os.path.dirname(
 
 
 install_requires = ['chardet', 'multidict>=4.0.0',
-                    'async_timeout>=1.2.0', 'yarl>=1.0.0']
+                    'async_timeout>=1.2.0,<3.0.0', 'yarl>=1.0.0']
 
 
 if sys.version_info < (3, 7):


and800:
> not pinning this will break on python 3.4

I believe the packaging toolchain must handle this. `async_timeout==3.0` has `python_requires='>=3.5.3'` in its metadata, so Pip won't install it for old Python.

On the other side, if I use python3.7 and aiohttp2.*, why can't I use new async_timeout?

asvetlov:
If you use old aiohttp -- why do you need the new `async-timeout`?
What prevents you to upgrade aiohttp if you use a new Python?

asvetlov:
But I agree with the note about toolchain: aiohttp 2.3 has loose `asyncio-timeout` restriction.
What `pip install` does? Why it installs `asyncio-timeout` 3.0 if the version doesn't match requirements?
What `pip` version do you use?

and800:
> What prevents you to upgrade aiohttp if you use a new Python?

laziness for overcoming v3 breaking changes other than python drop :)  

> why do you need the new async-timeout?

well I don't but another dependency may. I'm just pointing that the tighter requirements are, the closer dependency hell is. And the benefits of tightening requirements aren't obvious in this case

romuald:
@asvetlov I'm using pip 10.0.1 in this case to reproduce the issue

@and800 I'ved tested a `python setup.py install` from the 2.3 branch with a python 3.4 and it installed the 3.0.0 version of async-timeout

(The problem I actually had was with the pip from Debian's jessie dh_virtualenv)

and800:
@romuald if you want to install aiohttp from source, try `pip install .`

and800:
`setup.py install` command uses `easy_install` package downloader, not `pip`. So this issue is caused by that aiohttp doesn't support `easy_install` (and I think it would be hard to add that support)

@romuald what is your reason to install aiohttp from source? Pip allows you to [install any specific version](https://pip.pypa.io/en/stable/reference/pip_install/#requirement-specifiers) of the package.

asvetlov:
Hmm. I've merged but see no reason to make a new release on PyPI.
There are two reasons: 
1. autodeployment for 2.3 branch is outdated and broken. Definitely it can be fixed but requires additional time.
2. It is not an issue of aiohttp itself: 2.3 works with any async-timeout version installable on Python 3.4. `pip` has tricks and pitfalls for version dependencies but I believe installation from PyPI just worked fine.
3. Anyway versions can be pinned by requirements.txt etc.

PR #3070, opened 2018-06-11
Disable keep-alive when working with proxy

asvetlov:
Many HTTP proxies has buggy keepalive support.  

Let's not reuse connection but close it after processing every response.

### DIFF
--- a/CHANGES/3070.bugfix
+++ b/CHANGES/3070.bugfix
@@ -1,0 +1,2 @@
+Many HTTP proxies has buggy keepalive support.
+Let's not reuse connection but close it after processing every response.
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -45,6 +45,9 @@ class ResponseHandler(BaseProtocol, DataQueue):
                 self._payload_parser is not None or
                 len(self) or self._tail)
 
+    def force_close(self):
+        self._should_close = True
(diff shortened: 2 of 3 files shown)


socketpair:
I disagree about that. This is performance issue. Could you provide exact cases where that problem reproduced ?

asvetlov:
The feature was disabled from the very beginning.
I've enabled it in version 3.3 but we have constant timeout issues on our production servers.

Would be nice to have opt-in flag but it requires another issue/PR (maybe an attribute in `ProxyInfo` dataclass).

socketpair:
Maybe bug in your production servers ? does not it ?

asvetlov:
No. 
It is definitely a bug of our proxies (`force_close` for connector removes all timeout issues).
I recall the same problems 4 years ago (different project and proxy provider).
I suspect that buggy proxies are pretty common.

asvetlov:
I had a bad experience in my previous job. 
We used paid proxy fleets from different companies to mask our crawler agents.
Sorry, cannot provide more info.

webknjaz:
Why not have a switch for it, for users who want to risk it?

socketpair:
I vote for the switch, with default which ENABLES keep-alive.

asvetlov:
PR is welcome!

PR #3073, opened 2018-06-12
configure raise_for_status per request feature

sky-code:
## What do these changes do?

Add an ability to specify `raise_for_status` parameter per request

This changes has full backward compatibility

## Checklist

### DIFF
--- a/CHANGES/3073.feature
+++ b/CHANGES/3073.feature
@@ -1,0 +1,1 @@
+Add ``raise_for_status`` request parameter
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -231,6 +231,7 @@ class ClientSession:
                        compress=None,
                        chunked=None,
                        expect100=False,
+                       raise_for_status=None,
                        read_until_eof=True,
                        proxy=None,
                        proxy_auth=None,
@@ -465,7 +466,9 @@ class ClientSession:
                     break
 
             # check response status
-            if self._raise_for_status:
+            if raise_for_status is None:
+                raise_for_status = self._raise_for_status
+            if raise_for_status:
                 resp.raise_for_status()
 
             # register connection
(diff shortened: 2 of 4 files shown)


asvetlov:
The change has a price: additional API parameter that should be documented and remembered.
Client API already has too long list of available parameters, we should be careful with providing a new one.

On another hand the change doesn't bring a new functionality or saves lines of code. The following snippets are equal:
```
async with client.get(url, raise_for_status=True) as resp:
    await resp.text()
...
async with client.get(url) as resp:
    resp.raise_for_status()
    await resp.text()
```

I need a serious reason to approve the Pull Request.

sky-code:
Main reason for this new parameter is to override value provided to `ClientSession`

For example, you have `ApiClient` class with configured `ClientSession` object property, and you use this client session in `ApiClient` methods, for 10 methods it's ok to have `raise_for_status=True` but for another 3 methods you need `raise_for_status=False`, with current api this is impossible.
If you set `raise_for_status=True` - you set it for **all request** without ability to disable it for any request, so this feature not usable if you need it for **most of cases** but not for **all cases**.

I want to enable `raise_for_status` by default and override it when I need to check response code beyond 200, so this is inverted pattern from checking `resp.raise_for_status()` manually every time to setting `raise_for_status=False` when I expect non 200 response.

```
async with client.get(url) as resp:
    assert resp.status == 404 # if raise_for_status=True you won't be able to get here
    # so you end up with not use raise_for_status in ClientSession

async with client.get(url, raise_for_status=False) as resp:
    assert resp.status == 404 # now now you can reach here
```

asvetlov:
Well, this is another story.
Documentation should be updated, your use case should be mentioned.

sky-code:
I have added documentation

asvetlov:
Thanks

PR #3078, opened 2018-06-13
Add FAQ section Why is creating a ClientSession outside of an event loop dangerous?

### LINKED ISSUE #3072: Why is creating a ClientSession outside of an event loop "dangerous"?
If you create a `ClientSession` outside of a running event loop, you get:

    Creating a client session outside of coroutine is a very dangerous idea

Which comes from [here](https://github.com/aio-libs/aiohttp/blob/3dfc12379afc1657516e297fdd483c3b8a0db031/aiohttp/client.py#L122-L123)

**First and foremost, I would really like to know *why* this is dangerous.**

Ideally, I'd like to have the actual warning inform me, so that I would not need to dig through the code to determine it. But looking at the code, I'm still not sure: why is this dangerous?

To explain my use case, hopefully: at program initialization we create a few global `ClientSession`s, and pass those into various instances of things that require them. I don't really see anything inherent in `ClientSession` that should prevent it from initializing itself synchronously, and this pattern (creating needed connection pools at a high level context in the program, and passing them, typically via constructor, to lower level contexts) is essentially how I think anyone would expect a connection-pool like object to operate. (Now, I'm writing server-side code, so these pools have a lifetime of effectively forever; we shut them down when the server shuts down. I can of course see a more client-side code wanting a very brief `with ClientSession()`, but that isn't my use-case.)

asvetlov:
Fix #3072

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -332,3 +332,68 @@ operations::
 A :class:`aiohttp.web.Response` object must be returned. This is
 required by aiohttp web contracts, even though the response
 already been sent.
+
+
+Why is creating a ClientSession outside of an event loop dangerous?
+-------------------------------------------------------------------
+
+Short answer is: life-cycle of all asyncio objects should be shorter
+than life-chicle of event loop.
+
+Full explanation is longer.  All asyncio object should be correctly
+finished/disconnected/closed before event loop shutdown.  Otherwise
+user can get unexpected behavior. In the best case it is a warning
+about unclosed resource, in the worst case the program just hangs,
+awaiting for coroutine is never resumed etc.
+
+Consider the following code from ``mod.py``::
+
+    import aiohttp
+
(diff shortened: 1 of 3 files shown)


docs/faq.rst
```diff
@@ -332,3 +332,68 @@ operations::
 A :class:`aiohttp.web.Response` object must be returned. This is
 required by aiohttp web contracts, even though the response
 already been sent.
+
+
+Why is creating a ClientSession outside of an event loop dangerous?
+-------------------------------------------------------------------
+
+Short answer is: life-cycle of all asyncio objects should be shorter
+than life-chicle of event loop.
```
socketpair: `chicle` ?

https://en.wikipedia.org/wiki/Chicle

asvetlov: Fixed. Thanks

PR #3081, opened 2018-06-13
Postpone web task creation to connection establishment

### LINKED ISSUE #3052: SSL with closed connections
## Long story short

Receiving strange errors when bots are connecting to my SSL (HTTPS) server without sending any data and disconnecting afterwards.
SSL protocol handler inside aiohttp shows unexpected errors. This never happens to plain HTTP connections.

## Expected behaviour

No messages received (as for HTTP protocol).

## Actual behaviour

Multiple errors:
task: <Task pending coro=<RequestHandler.start() done, defined at C:\Program Files\Python36\lib\site-packages\aiohttp\web_protocol.py:342> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x000001E0FBF59D98>()]>>

## Steps to reproduce

Run simple example:

```
import asyncio  
import aiohttp.web
import ssl

async def fetch_page():
    while True:
        reader, writer = await asyncio.open_connection('127.0.0.1', 443)
        writer.close()
    
async def request_handler(request: aiohttp.web.BaseRequest):
    await request.read()
    return aiohttp.web.Response(status=200, content_type='text/html')

async def start_server():
    ssl_context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
    ssl_context.check_hostname = False
    ssl_context.load_cert_chain('any.crt', 'any.key')

    server = aiohttp.web.Server(request_handler)
    await loop.create_server(server, '127.0.0.1', 443, ssl=ssl_context) # removing ssl=ssl_context disables SSL and fixes the bug
    await fetch_page()

if __name__ == '__main__':  
    loop.run_until_complete(asyncio.gather(asyncio.Task(start_server())))
```

## Your environment

Checked both Windows 10 and CentOS 7, aiohttp 3.0.4 and 3.3.0

asvetlov:
Fix #3052

### DIFF
--- a/CHANGES/3052.bugfix
+++ b/CHANGES/3052.bugfix
@@ -1,0 +1,1 @@
+Postpone web task creation to connection establishment.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -109,7 +109,7 @@ class RequestHandler(BaseProtocol):
 
         self._waiter = None
         self._error_handler = None
-        self._task_handler = self._loop.create_task(self.start())
+        self._task_handler = None
 
         self._upgrade = False
         self._payload_parser = None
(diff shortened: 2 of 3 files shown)


pfreixes:
Umm, but moving the task creation when the connection made is triggered is only delaying a loop iteration the task creation. migth be the SSL triggering this issue just because of the time needed to make the SSL handshake? So, maybe better just consider the case and just cancel the task ?

Or am I missing something ?

asvetlov:
From my understanding, a protocol instance is created immediately on socket accepting but `connection_made` is called after SSL handshake. If handshake is not successful the protocol is destroyed by GC *without* `connection_lost` call.
Another solution could be dropping the task in `__del__` but I'd like to avoid this.

PR #3093, opened 2018-06-19
raise 413 from request.post() not ValueError

### LINKED ISSUE #3087: request.post() should raise HTTPRequestEntityTooLarge same as request.read()
`request.read()`

https://github.com/aio-libs/aiohttp/blob/0bc7d52ccdcc8014293130bb5be0a7075cf1c8bc/aiohttp/web_request.py#L526-L528

`request.post()`

https://github.com/aio-libs/aiohttp/blob/0bc7d52ccdcc8014293130bb5be0a7075cf1c8bc/aiohttp/web_request.py#L583-L585

Seems inconsistent to me that the two methods behave differently. `Request.post()` should raise `HTTPRequestEntityTooLarge` too. 

Or am I missing something?

samuelcolvin:
## What do these changes do?

Raise 413 "Payload Too Large" rather than raising ValueError in request.post()

post will raise `HTTPRequestEntityTooLarge` rather than ValueError

## Related issue number

fix #3087

## Checklist

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -587,8 +587,7 @@ class BaseRequest(collections.MutableMapping, HeadersMixin):
                         tmp.write(chunk)
                         size += len(chunk)
                         if 0 < max_size < size:
-                            raise ValueError(
-                                'Maximum request body size exceeded')
+                            raise HTTPRequestEntityTooLarge
                         chunk = await field.read_chunk(size=2**16)
                     tmp.seek(0)
 
@@ -605,8 +604,7 @@ class BaseRequest(collections.MutableMapping, HeadersMixin):
                     out.add(field.name, value)
                     size += len(value)
                     if 0 < max_size < size:
-                        raise ValueError(
-                            'Maximum request body size exceeded')
+                        raise HTTPRequestEntityTooLarge
 
                 field = await multipart.next()
         else:
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -529,9 +529,13 @@ class BaseRequest(collections.MutableMapping, HeadersMixin):
             while True:
                 chunk = await self._payload.readany()
                 body.extend(chunk)
-                if self._client_max_size \
-                        and len(body) >= self._client_max_size:
-                    raise HTTPRequestEntityTooLarge
+                if self._client_max_size:
+                    body_size = len(body)
+                    if body_size >= self._client_max_size:
+                        raise HTTPRequestEntityTooLarge(
+                            max_size=self._client_max_size,
+                            actual_size=body_size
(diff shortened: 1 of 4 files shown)


samuelcolvin:
by the way, when I run `make test`, I'm getting

```
tests/test_helpers.py:319: in <module>
    class TestCReify(ReifyMixin):
tests/test_helpers.py:320: in TestCReify
    reify = helpers.reify_c
E   AttributeError: module 'aiohttp.helpers' has no attribute 'reify_c'
```

This seems to be new, I wasn't getting it before.

asvetlov:
Perhaps you didn't compile cython extensions.
`make test` should do it.
Also please don't forget `git submodule init` / `git submodule update`

samuelcolvin:
`git submodule init` / `git submodule update` was what I needed. Thank you.

aiohttp/web_request.py
```diff
@@ -605,8 +604,7 @@ def body_exists(self) -> bool:
                     out.add(field.name, value)
                     size += len(value)
                     if 0 < max_size < size:
-                        raise ValueError(
-                            'Maximum request body size exceeded')
+                        raise HTTPRequestEntityTooLarge
```
kxepal: May be add some message like what size was expected and which one is actual? That's good for debug.

asvetlov: Makes sense.
We do it for too large headers already.

samuelcolvin: Ok, so something like `Maximum request body size exceeded: 1234 > 123`?

samuelcolvin: Or perhaps clearer `Maximum request body size 123 exceeded, actual body size 1234`?

asvetlov: I like the second

kxepal: :+1:  the second. A bit verbose, but that wouldn't hurt.

aiohttp/web_request.py
```diff
@@ -605,8 +609,8 @@ def body_exists(self) -> bool:
                     out.add(field.name, value)
                     size += len(value)
                     if 0 < max_size < size:
-                        raise ValueError(
-                            'Maximum request body size exceeded')
+                        raise HTTPRequestEntityTooLarge(
+                            max_size=max_size, actual_size=size)
```
kxepal: That style is a bit different from above: 
```
+                if self._client_max_size:
+                    body_size = len(body)
+                    if body_size >= self._client_max_size:
+                        raise HTTPRequestEntityTooLarge(
+                            max_size=self._client_max_size,
+                            actual_size=body_size
+                        )
```
Let's be consistent.

asvetlov:
thanks!

PR #3095, opened 2018-06-19
Micro-optimization

asvetlov:


aiohttp/web_protocol.py
```diff
@@ -75,10 +75,17 @@ class RequestHandler(BaseProtocol):
...
     KEEPALIVE_RESCHEDULE_DELAY = 1
 
+    __slots__ = ('_request_count', '_keep_alive', '_manager',
+                 '_request_handler', '_request_factory', '_tcp_keepalive',
+                 '_keepalive_time', '_keepalive_handle', '_keepalive_timeout',
+                 '_lingering_time', '_messages', '_message_tail',
+                 '_waiter', '_error_handler', '_task_handler',
+                 '_upgrade', '_payload_parser', '_request_parser',
+                 '_reading_paused', 'logger', 'debug', 'access_log',
+                 'access_logger', '_close', '_force_close')
```
kxepal, line 87: May be have it one-item-per-line and sorted? Would simplify management and new items addition if that happens.
```
    __slots__ = (
        '_close',
        '_error_handler',
        '_force_close',
        '_keep_alive',
        '_keepalive_handle',
        '_keepalive_time',
        '_keepalive_timeout',
        '_lingering_time',
        '_manager',
        '_message_tail',
        '_messages',
        '_payload_parser',
        '_reading_paused',
        '_request_count',
        '_request_factory',
        '_request_handler',
        '_request_parser',
        '_task_handler',
        '_tcp_keepalive',
        '_upgrade',
        '_waiter',
        'access_log',
        'access_logger',
        'debug',
        'logger',
    )
```

asvetlov: Sorting is boring and weak: very easy to make a mistake in sort order.

I'm thinking about ultimately applying *black* but not ready for it yet.

kxepal: _black_ is good, but you wouldn't like what it will do with string quotes - it prefers double ultimately.

asvetlov: Maybe we can live with it.

samuelcolvin: What is black?

asvetlov: https://github.com/ambv/black

asvetlov:
Not so much on macro benchmarks (4-5%), but HTTP parser is 2 times faster now.

kxepal:
Cool!

asvetlov:
You know, boosting existing pre-optimized code is a hard job.
I see several points for speed up. No one is fascinating but the cummulative effect could be very visible.

aiohttp/web_protocol.py
kxepal: Yeah, that's a new feature. I agree that black isn't perfect, but eventually, there are no much good alternatives. Good news: it's quite easy to fork it and fix stuff to make you like it. In the end, we could have some aio-libs style guide which is backed by ours formatter. Still that's better than nothing.

PR #3098, opened 2018-06-21
Make HTTPException truthy

### LINKED ISSUE #3096: HTTPException should retain cause information
## Long story short

I would like to use a pattern like:
```python
try:
    do_something()
except SomeException as e:
    raise HTTPInternalServerError(reason='My message') from e
```
with cause information retained. Does it make sense to implement `HTTPException.__bool__()` to always `True`?

## Expected behaviour

Two tracebacks separated by `The above exception was the direct cause of the following exception:`

## Actual behaviour

The cause information is lost, because `traceback.TracebackException` evaluates the [truthiness of the exception value](https://github.com/python/cpython/blob/de86073a761cd3539aaca6f886a1f55effc0d9da/Lib/traceback.py#L464) to determine whether cause lines are produced. For `HTTPException`, [truthiness is resolved](https://github.com/aio-libs/aiohttp/blob/7cf5785f5383ee67756d0f8d86ea3516b823b940/aiohttp/web_response.py#L426) via `StreamResponse.__len__()` which returns 0.

jcjf:
## What do these changes do?

Fixes #3096 

More informative tracebacks when using `raise from` syntax.

## Related issue number

#3096

## Checklist

### DIFF
--- a/tests/test_web_exceptions.py
+++ b/tests/test_web_exceptions.py
@@ -1,5 +1,6 @@
 import collections
 import re
+from traceback import format_exception
 from unittest import mock
 
 import pytest
@@ -172,3 +173,14 @@ def test_link_header_451(buf, request):
 
     assert 'http://warning.or.kr/' == resp.link
     assert '<http://warning.or.kr/>; rel="blocked-by"' == resp.headers['Link']
+
+
+def test_HTTPException_retains_cause():
+    with pytest.raises(web.HTTPException) as ei:
+        try:
+            raise Exception('CustomException')
+        except Exception as exc:
+            raise web.HTTPException() from exc
+    tb = ''.join(format_exception(ei.type, ei.value, ei.tb))
+    assert 'CustomException' in tb
+    assert 'direct cause' in tb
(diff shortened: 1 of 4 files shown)


tests/test_web_exceptions.py
```diff
@@ -172,3 +173,14 @@ def test_link_header_451(buf, request):
...
     assert '<http://warning.or.kr/>; rel="blocked-by"' == resp.headers['Link']
+
+
+def test_HTTPException_retains_cause():
+    with pytest.raises(web.HTTPException) as ei:
+        try:
+            raise Exception('CustomException')
+        except Exception as exc:
+            raise web.HTTPException() from exc
+    tb = ''.join(format_exception(ei.type, ei.value, ei.tb))
```
webknjaz, line 184: is `ei` available outside of `with` block?

asvetlov: Yes it is.
We use it widely in the test suite:
```
    with pytest.raises(aiohttp.ClientResponseError) as info:
        await response.start(connection)

    assert info.value.request_info is request_info
```

asvetlov:
@jcjf Thank you

PR #3107, opened 2018-06-25
Fix StreamResponse equality

### LINKED ISSUE #3100: StreamResponse instances are all equal
## Long story short

Since #2494 , `StreamResponse` inherits `collections.MutableMapping.__eq__`, which makes them basically all equal. 
The implementation in `Mapping` looks like `return dict(self.items()) == dict(other.items())`.

This is especially the case for `WebSocketResponse` : when following https://aiohttp.readthedocs.io/en/stable/faq.html#how-do-i-programmatically-close-a-websocket-server-side, if you use a `list` instead of a `set`, it's impossible to `remove` the correct websocket.

## Expected behaviour

```python
>>> from aiohttp.web_ws import WebSocketResponse
>>> r1 = WebSocketResponse()
>>> r2 = WebSocketResponse()
>>> r1 == r2
False
>>> id(r1) == id(r2)
False
>>> r1 is r2
False
>>> hash(r1) == hash(r2)
False
```

As a rule, `a == b` implies `hash(a) == hash(b)`. But it's now broken.

## Actual behaviour

Since v3.0:
```python
>>> r1 == r2
True
>>> id(r1) == id(r2)
False
>>> r1 is r2
False
>>> hash(r1) == hash(r2)
False
```

## Steps to reproduce

Described above

## Your environment

* `aiohttp >= 3.0`

NotSqrt:
## What do these changes do?

StreamResponse as subclass of MutableMapping introduced equality defined as `return dict(self.items()) == dict(other.items())`.
Now, it is back to identity of addresses.

Back to the behaviour of pre-3.0 aiohttp version.

## Related issue number

Fixes #3100 

## Checklist

- [x] Documentation reflects the changes (no specific mention of  `__eq__` in the docs)

### DIFF
--- a/CHANGES/3100.bugfix
+++ b/CHANGES/3100.bugfix
@@ -1,0 +1,1 @@
+Fix `StreamResponse` equality, now that they are `MutableMapping` objects.
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -432,6 +432,9 @@ class StreamResponse(collections.MutableMapping, HeadersMixin):
     def __hash__(self):
         return hash(id(self))
 
+    def __eq__(self, other):
+        return self is other
+
 
 class Response(StreamResponse):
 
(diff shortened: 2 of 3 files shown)


asvetlov:
Thanks!

Issue #3102, opened 2018-06-23
Fix (make robust) test_timer_context_cancelled under Python 3.7


webknjaz:
## Long story short

With emerging of 3.7, we need to adjust test suite to pass under this env

## Expected behaviour

test passes

## Actual behaviour

One of coroutine cancellation tests fails:

```python-traceback
$ make cov-ci-no-ext
...
――――――――――――――――――――――――― test_timer_context_cancelled ―――――――――――――――――――――――――
    def test_timer_context_cancelled():
        with mock.patch('aiohttp.helpers.asyncio') as m_asyncio:
            m_asyncio.TimeoutError = asyncio.TimeoutError
            loop = mock.Mock()
            ctx = helpers.TimerContext(loop)
            ctx.timeout()
    
            with pytest.raises(asyncio.TimeoutError):
                with ctx:
                    pass
    
>           assert m_asyncio.Task.current_task.return_value.cancel.called
E           AssertionError: assert False
E            +  where False = <MagicMock name='asyncio.Task.current_task().cancel' id='23073207868104'>.called
E            +    where <MagicMock name='asyncio.Task.current_task().cancel' id='23073207868104'> = <MagicMock name='asyncio.Task.current_task()' id='23073207876128'>.cancel
E            +      where <MagicMock name='asyncio.Task.current_task()' id='23073207876128'> = <MagicMock name='asyncio.Task.current_task' id='23073207863616'>.return_value
E            +        where <MagicMock name='asyncio.Task.current_task' id='23073207863616'> = <MagicMock name='asyncio.Task' id='23073207799992'>.current_task
E            +          where <MagicMock name='asyncio.Task' id='23073207799992'> = <MagicMock name='asyncio' id='23073210485560'>.Task
tests/test_helpers.py:430: AssertionError
```

## Steps to reproduce

See https://travis-ci.com/aio-libs/aiohttp/jobs/130964628#L793-L814

## Your environment

Travis CI

asvetlov:
Thanks for catch up. 
The test is outdated, need to review and fix it.

PR #3113, opened 2018-06-29
Fix 3102: add check for to py37 in test_timer_context_cancelled

gyermolenko:
bugfix for #3102

Has implications for all tests marked with `@pytest.mark.skipif(not PY_36, reason="Python 3.6+ required")`

But current PY_36 check is rather wrong because for python 3.7
`PY_36 = sys.version_info >= (3, 6)` gives us `PY_36` as `True`

**upd:**
added fix for `test_tcp_connector_certificate_error`, [msg updated in py37](https://docs.python.org/3/library/ssl.html#ssl.CertificateError)

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -34,8 +34,8 @@ from .log import client_logger
 
 __all__ = ('BasicAuth', 'ChainMapProxy')
 
-PY_36 = sys.version_info >= (3, 6)
-PY_37 = sys.version_info >= (3, 7)
+PY_36 = (3, 6) <= sys.version_info < (3, 7)
+PY_37 = (3, 7) <= sys.version_info < (3, 8)
 
 if not PY_37:
     import idna_ssl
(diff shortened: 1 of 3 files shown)


aiohttp/helpers.py
```diff
@@ -34,8 +34,8 @@
 
 __all__ = ('BasicAuth', 'ChainMapProxy')
 
-PY_36 = sys.version_info >= (3, 6)
-PY_37 = sys.version_info >= (3, 7)
+PY_36 = (3, 6) <= sys.version_info < (3, 7)
```
asvetlov, line 37: It is not obvious but aiohttp uses the following convention: `PY_36` means that at least `Python 3.6` is present.
It may be 3.6.3 or even 3.7
The convention makes sense for a feature support check. If the feature appears in 3.6 it is supported by all future versions.

asvetlov:
Fixed by #3128
@gyermolenko thanks for contribution

PR #3114, opened 2018-06-30
Fix install_requires following PEP-508

item4:
## What do these changes do?

pipenv can not support code-based deps list.
this PR change install requires list following PEP-508

Please see https://github.com/pypa/pipenv/issues/2473 for details.

no, there is only change to installation

## Related issue number

https://github.com/pypa/pipenv/issues/2473

## Checklist

Please help me to check below. I do not check it myself

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ from distutils.command.build_ext import build_ext
 from distutils.errors import (CCompilerError, DistutilsExecError,
                               DistutilsPlatformError)
 
+import setuptools
 from setuptools import Extension, setup
 
 
@@ -71,38 +72,6 @@ try:
 except IndexError:
     raise RuntimeError('Unable to determine version.')
 
-
-def get_environment_marker_support_level():
-    """
-    Copied from https://github.com/pytest-dev/pytest/blob/master/setup.py
-
-    Tests how well setuptools supports PEP-426 environment marker.
-    The first known release to support it is 0.7 (and the earliest on PyPI seems to be 0.7.2
-    so we're using that), see: https://setuptools.readthedocs.io/en/latest/history.html#id350
-    The support is later enhanced to allow direct conditional inclusions inside install_requires,
-    which is now recommended by setuptools. It first appeared in 36.2.0, went broken with 36.2.1, and
-    again worked since 36.2.2, so we're using that. See:
-    https://setuptools.readthedocs.io/en/latest/history.html#v36-2-2
-    https://github.com/pypa/setuptools/issues/1099
-    References:
-    * https://wheel.readthedocs.io/en/latest/index.html#defining-conditional-dependencies
-    * https://www.python.org/dev/peps/pep-0426/#environment-markers
-    * https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
-    """
-    try:
-        version = pkg_resources.parse_version(setuptools.__version__)
-        if version >= pkg_resources.parse_version("36.2.2"):
-            return 2
-        if version >= pkg_resources.parse_version("0.7.2"):
-            return 1
-    except Exception as exc:
-        sys.stderr.write("Could not test setuptool's version: %s\n" % exc)
-
-    # as of testing on 2018-05-26 fedora was on version 37* and debian was on version 33+
-    # we should consider erroring on those
-    return 0
-
-
 install_requires = [
     'attrs>=17.3.0',
     'chardet>=2.0,<4.0',
@@ -112,11 +81,15 @@ install_requires = [
 ]
 extras_require = {}
 
-environment_marker_support_level = get_environment_marker_support_level()
-if environment_marker_support_level >= 2:
+setuptools_version_tuple = setuptools.__version__.split('.')
+setuptools_version = tuple(
+    int(setuptools_version_tuple[0]),
+    int(setuptools_version_tuple[1]),
+    int(setuptools_version_tuple[2]),
+)
+
+if setuptools_version >= (36, 2, 2):
     install_requires.append('idna-ssl>=1.0;python_version<"3.7"')
-elif environment_marker_support_level == 1:
-    extras_require[':python_version<"3.7"'] = ["idna-ssl>=1.0"]
 else:
     if sys.version_info < (3, 7):
         install_requires.append("idna-ssl>=1.0")
--- a/setup.py
+++ b/setup.py
@@ -79,7 +79,6 @@ install_requires = [
     'async_timeout>=3.0,<4.0',
     'yarl>=1.0,<2.0',
 ]
-extras_require = {}
 
 setuptools_version_tuple = setuptools.__version__.split('.')
 setuptools_version = (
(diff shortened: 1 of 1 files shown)


uranusjr:
Another related issue: #2703. The approach used here is considered the best way currently available to specify environment-specific requirements. Both the Python-based (currently in use) and `extra_requires` (mentioned in #2703) are no longer recommended, as they may cause problems in dependency resolution in package managers, and could provide incorrect metadata reporting on PyPI.

setup.py
```diff
@@ -72,13 +72,14 @@ def build_extension(self, ext):
...
-
-if sys.version_info < (3, 7):
-    install_requires.append('idna-ssl>=1.0')
+install_requires = [
+    'attrs>=17.3.0',
+    'chardet>=2.0,<4.0',
+    'multidict>=4.0,<5.0',
+    'async_timeout>=3.0,<4.0',
+    'yarl>=1.0,<2.0',
+    'idna-ssl>=1.0;python_version<"3.7"',
```
webknjaz: AFAIR pip (setuptools?) doesn't support this kind of notation in `install_requires` (at least in certain range of versions).
I'd say it's better to go for extras-based syntax:
```python
extra_requires = {
    ':python_version<"3.7"': [
        'idna-ssl>=1.0',
    ],
}
```

webknjaz:
@uranusjr while it might be not recommended we target a wider range of envs and extras-based notation is better in this case. I'll need to check feature-support first.

setup.py
Victor-Savu: Is there any way to test this? There have been quite a few improvements in this area in pip v10.x, and support for PEP 508 came up more than once.

uranusjr:
@webknjaz Unfortunately `extra_requires` also has issues during detection. I generally do not recommend sticking to the old syntax, since the current recommendation has been stable for more than a year, and newer versions of Setuptools is available to everyone to upgrade.

If Setuptools compatibility is an issue, Pytest’s approach may be more appropriate:
https://github.com/pytest-dev/pytest/blob/master/setup.py#L26

I would also like to mention that I have not seen any real-life reports on Setuptools support on `installs_require` markers being an issue.

uranusjr:
@Victor-Savu The warning only applies for `extra`; _environment_ markers do work in pip 9.

----

Update: There is a table in the middle of the PEP document listing available markers. All of them except `extra` work in relatively up-to-date Setuptools and pip.

https://www.python.org/dev/peps/pep-0508/

setup.py
kxepal: @webknjaz it does today. And at least bundled pip for 3.5 support that. For older I think we shouldn't care. 

In anyway, rule of thumb is: before install anything in new virtualenv, upgrade pip and setuptools first.

kxepal:
I'm +1 this patch. While I share @webknjaz concerns about old setuptools/pip, today that shouldn't be a problem and, in general, I think we shouldn't let them slow down us, because users may easily upgrade them at any moment for free. 

-1 for extras - that's not right solution, because:
- It's easy to forget about that extra (how would you know about it without reading the sources?)
- You will break all the packages that depends on aiohttp - they don't have any extras in own install_requires list, so they wouldn't get that package
- All downstreams would have to maintain own if-condition: if 3.5, then use aiohttp's extra, if not - don't. Not cool.

item4:
I changed code. please review again.

setup.py
```diff
@@ -72,13 +72,54 @@ def build_extension(self, ext):
...
+    again worked since 36.2.2, so we're using that. See:
+    https://setuptools.readthedocs.io/en/latest/history.html#v36-2-2
+    https://github.com/pypa/setuptools/issues/1099
+    References:
+    * https://wheel.readthedocs.io/en/latest/index.html#defining-conditional-dependencies
+    * https://www.python.org/dev/peps/pep-0426/#environment-markers
+    * https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
+    """
+    try:
+        version = pkg_resources.parse_version(setuptools.__version__)
```
kxepal: You forgot the import

setup.py
```diff
@@ -72,13 +72,54 @@ def build_extension(self, ext):
...
+        if version >= pkg_resources.parse_version("36.2.2"):
+            return 2
+        if version >= pkg_resources.parse_version("0.7.2"):
```
kxepal: Should we care about such old versions? I think we shouldn't.

setup.py
```diff
@@ -72,13 +72,54 @@ def build_extension(self, ext):
...
-install_requires = ['attrs>=17.3.0', 'chardet>=2.0,<4.0',
-                    'multidict>=4.0,<5.0',
-                    'async_timeout>=3.0,<4.0',
-                    'yarl>=1.0,<2.0']
-
-if sys.version_info < (3, 7):
-    install_requires.append('idna-ssl>=1.0')
+def get_environment_marker_support_level():
+    """
+    Copied from https://github.com/pytest-dev/pytest/blob/master/setup.py
```
kxepal: If you copy code from third party project, you need to add this reference to NOTICE file as ASF license stands for.

setup.py
item4: I will fix it

setup.py
item4: OK

setup.py
item4: I will drop this copied code

setup.py
```diff
@@ -71,14 +72,27 @@ def build_extension(self, ext):
...
+)
 
-install_requires = ['attrs>=17.3.0', 'chardet>=2.0,<4.0',
-                    'multidict>=4.0,<5.0',
-                    'async_timeout>=3.0,<4.0',
-                    'yarl>=1.0,<2.0']
-
-if sys.version_info < (3, 7):
-    install_requires.append('idna-ssl>=1.0')
+if setuptools_version >= (36, 2, 2):
```
kxepal: environment markers are supported since 20.5 version, but really get worked since 21.0.  No need to require such high setuptools version.

setup.py
```diff
@@ -71,14 +72,27 @@ def build_extension(self, ext):
...
     raise RuntimeError('Unable to determine version.')
 
+install_requires = [
+    'attrs>=17.3.0',
+    'chardet>=2.0,<4.0',
+    'multidict>=4.0,<5.0',
+    'async_timeout>=3.0,<4.0',
+    'yarl>=1.0,<2.0',
+]
+extras_require = {}
```
kxepal: extras seems to be unused.

setup.py
kxepal: Also, version 21.0 released two years ago. Once again, do we need any checks against setuptools version at all? Unlikely any sane user will install latest aiohttp having legacy package management toolset.

item4:
@kxepal I think checking setupools version is useless and I want to reset --hard to first commit. but someone commented setuptools versioni issue. I want to hear opinion of member of aiohttp

kxepal:
@webknjaz 
Are you ok to not check setuptools version?

asvetlov:
Dropping setuptools version check is ok for me: env markers support is pretty old.

PR #3124, opened 2018-07-05
Do not hang on `206 Partial Content` response with `Content-Encoding: gzip`

### LINKED ISSUE #3123: Client hangs on 206 Partial Content response with Content-Encoding: gzip
## Long story short

When server responds with 206 to range request - `response.read()` hangs.
Also `requests` library handles the same scenarios successfully.

## Expected behaviour

Partial content is a valid content when responding to a range request. The client should not hang and just decode as much as possible.

## Actual behaviour

The `response.read()` hangs indefinitely if Content-Encoding: gzip. Does not hang when Content-Encoding is deflate or brotli.

## Steps to reproduce

```python
import asyncio

import aiohttp
import aiohttp.web
import requests

async def client():
    async with aiohttp.ClientSession(headers={"Range": "bytes=0-1000"}) as session:
        # async with session.get("https://www.airbnb.com") as response:
        async with session.get("https://www.tripadvisor.com") as response:
            await response.read()
            print("Never reached")

print(requests.get("https://www.tripadvisor.com", headers={"Range": "bytes=0-1000"}).text)
print(requests.get("https://www.airbnb.com", headers={"Range": "bytes=0-1000"}).text)

asyncio.get_event_loop().run_until_complete(client())
```

## Your environment

aiohttp==3.3.2
requests==2.19.1
macOS 10.13.5
no proxy
aiohttp client

pauleikis:
## What do these changes do?

Do not raise `ContentEncodingError('deflate')` for `gzip` encoding when the response is partial. This can be a valid response in case of 206 Partial Content. The bug was not reproducible with deflate or brotli encoding

User behaviour is not affected

## Related issue number

This fixes #3123

## Checklist

- [x ] I think the code is well written
- [x ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
- [ x] Add a new news fragment into the `CHANGES` folder


pauleikis:
Actually I was trying to reproduce it aiohttp server, but did not manage to send the required response. Luckily I found some in the wild (can be seen in #3123)  Not sure how to write the test. If I understand correctly, aiohttp server does not support 206 for non-file responses.

asvetlov:
You can gzip the body in web handler and return a regular response with the partial body and `Content-Encoding` header set.

pauleikis:
OK, will try to cook something up

pauleikis:
Added tests. Please review again.

PR #3152, opened 2018-07-23
Improve lingering logic and documentation

Harmon758:
### What do these changes do?

Improve logic for lingering and documentation of [`Application.make_handler`](http://docs.aiohttp.org/en/latest/web_reference.html#aiohttp.web.Application.make_handler) parameters

### Are there changes in behavior for the user?

No

### Related issue number

N/A

## Checklist


aiohttp/web_protocol.py
```diff
@@ -431,8 +431,7 @@ def resume_reading(self):
                         with suppress(
                                 asyncio.TimeoutError, asyncio.CancelledError):
                             while not payload.is_eof() and now < end_t:
-                                timeout = min(end_t - now, lingering_time)
-                                with CeilTimeout(timeout, loop=loop):
+                                with CeilTimeout(end_t - now, loop=loop):
```
asvetlov, line 424: You are changing a logic here. Why?

Harmon758: This just optimizes the logic, since `end_t - now` is always less than or equal to `lingering_time`.
As `end_t` is set to `now + lingering_time`, `end_t - now` is equivalent to `lingering_time` in the first iteration of the while loop. Afterwards, `end_t` and `lingering_time` don't change and `now` is set to `loop.time()` in each iteration of the loop, so `end_t - now` becomes increasingly smaller and is never larger than `lingering_time` at any point, so it just sets `timeout` to `end_t - now` each time.

Issue #3162, opened 2018-07-29
Build/install broken without cython


apatrushev:
## Long story short

Latest master cannot be installed without cython.
```
clang: error: no such file or directory: 'aiohttp/_websocket.c'
```

## Expected behaviour

```
pip install git+http://github.com/aio-libs/aiohttp
```
Should install aiohttp even without cython. It worked at least in previous versions.

## Actual behaviour

Build failed:
```
    running build_ext
    building 'aiohttp._websocket' extension
    creating build/temp.macosx-10.13-x86_64-3.7
    creating build/temp.macosx-10.13-x86_64-3.7/aiohttp
    gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -I/Users/tosha/.pythonz/pythons/CPython-3.7.0/Frameworks/Python.framework/Versions/3.7/include/python3.7m -c aiohttp/_websocket.c -o build/temp.macosx-10.13-x86_64-3.7/aiohttp/_websocket.o
    clang: error: no such file or directory: 'aiohttp/_websocket.c'
    clang: error: no input files
    error: command 'gcc' failed with exit status 1
```

## Steps to reproduce

Run
```
pip install git+http://github.com/aio-libs/aiohttp
```
in fresh empty venv.

## Your environment

macOS 10.13.6/Python 3.7/empty vevn
I am sure that the problem is OS/Python version independent.

webknjaz:
Clone with recursive submodules and install from there

webknjaz:
Ref #3138

apatrushev:
Sorry, my fault. May be it will make a sense to write the link to CONTRIBUTING.txt (the only place where I found `recursive` word) somewhere in issue template?

apatrushev:
Yeah, and before it was possible to install `aiohttp` with command I mentioned above:
```
pip install git+http://github.com/aio-libs/aiohttp
```
I think that vendoring with submodules is a little step back. Just an opinion.

webknjaz:
GitHub shows the link to the CONTRIBUTING in the side bar when you create an issue or PR

asvetlov:
Make sense.
Contributing is awailable from aiohttp docs but if you want a new link -- please add it.

asvetlov:
Sorry, nevermind

webknjaz:
[...]

PR #3164, opened 2018-07-29
Fix build without cython (#3162)

apatrushev:
## What do these changes do?

Fixes build/install without cython

Nope

## Related issue number

#3162

## Checklist

- [-] Unit tests for the changes exist
- [-] Documentation reflects the changes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ class ve_build_ext(build_ext):
     def build_extension(self, ext):
         try:
             build_ext.build_extension(self, ext)
-        except (DistutilsExecError,
+        except (CCompilerError, DistutilsExecError,
                 DistutilsPlatformError, ValueError):
             raise BuildFailed()
 
--- a/CHANGES/3162.bugfix
+++ b/CHANGES/3162.bugfix
@@ -1,0 +1,1 @@
+Build without cython fixed
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -33,6 +33,7 @@ Andrew Svetlov
 Andrii Soldatenko
 Antoine Pietri
 Anton Kasyanov
+Anton Patrushev
 Anton Zhdan-Pushkin
 Arthur Darcet
 Ben Bader


setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
     def build_extension(self, ext):
         try:
             build_ext.build_extension(self, ext)
-        except (DistutilsExecError,
+        except (CCompilerError, DistutilsExecError,
```
webknjaz, line 60: This is a workaround. We need a real fix, finding out why the file is missing or having a clear explanation of why this is happening.

webknjaz: Okay, it might make sense. Let's wait for @asvetlov.
P.S. I still want to see this tested in CI.

apatrushev: Nope. We already discussed the problem with @asvetlov in #3162 and I probably caught his idea and pain about development environment. Though the build is broken anyway: it is impossible to build/install from sources without gcc. So I plan to improve setup.py.

webknjaz: Feel free to try, but I'll require adding env combos to Travis CI as mandatory anyway.

webknjaz: And while you're on it you might want to extend docs to mention how things work.

apatrushev: ```I'll require adding env combos to Travis CI as mandatory anyway.```

@asvetlov mentioned that this is the bad idea, because it will increase even more the testing time.

webknjaz: Not really, just add it and then I'll make sure that it doesn't hit performance. I'll basically do to it what I did to OS X and it'll be a fine compromise: running those only on cron as opposed to having them in each build. Andrew trusts me to handle CI.

P.S. If we recognize that we need it on each commit we could always move it to another CI service.

setup.py
```diff
@@ -57,7 +57,7 @@ def run(self):
...
                 DistutilsPlatformError, ValueError):
             raise BuildFailed()
```
webknjaz, line 62: It's also useful to add `from original_exception` part.

setup.py
webknjaz: Remark: while working on changing `setup.py` it might be worth adding some env var for toggling strict mode for dev.

It could be smth like:
```python
import os
...
class ve_build_ext(build_ext):
    ...
    def build_extension(self, ext):
        valid_build_exceptions = {CCompilerError, DistutilsExecError, DistutilsPlatformError, ValueError}
        suppressed_build_exceptions = set()

        if os.getenv('AIOHTTP_DEV_MODE'):
            suppressed_build_exceptions = {CCompilerError, DistutilsExecError, DistutilsPlatformError}

        try:
            build_ext.build_extension(self, ext)
        except tuple(valid_build_exceptions - suppressed_build_exceptions) as build_exc:
            raise BuildFailed() from build_exc
```

cc: @asvetlov

apatrushev: Yeah, that's what I planned to implement.

asvetlov:
Superseded by #3189

PR #3173, opened 2018-08-03
Add option to remove slash to `normalize_path_middleware`

kryptn:
## What do these changes do?

This change adds new functionality to the `normalize_path_middleware`. I wanted the option to remove trailing slashes and it made sense to just add the functionality to the existing middleware.

No breaking changes to existing usage. The `normalize_path_middleware` factory only accepts keyword arguments, and because there's now an assertion that only one of `append_slash` and `remove_slash` can be enabled this change shouldn't be breaking.

## Related issue number

None that I could find.

## Checklist

### DIFF
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -2762,27 +2762,37 @@ Normalize path middleware
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. function:: normalize_path_middleware(*, \
-                                        append_slash=True, merge_slashes=True)
+                                        append_slash=True, \
+                                        remove_slash=False, \
+                                        merge_slashes=True, \
+                                        redirect_class=HTTPMovedPermanently)
 
   Middleware factory which produces a middleware that normalizes
   the path of a request. By normalizing it means:
 
-      - Add a trailing slash to the path.
+      - Add or remove a trailing slash to the path.
       - Double slashes are replaced by one.
 
   The middleware returns as soon as it finds a path that resolves
-  correctly. The order if all enabled is:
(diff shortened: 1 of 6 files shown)


kryptn:
Thanks! Can do, but I'm not sure which version to add in there. should it be the expected release or the current version i made the change to?

kryptn:
Went ahead and assumed the next minor version and updated.

Thanks again!

docs/web_reference.rst
```diff
@@ -2762,27 +2762,41 @@ Normalize path middleware
...
 .. function:: normalize_path_middleware(*, \
-                                        append_slash=True, merge_slashes=True)
+                                        append_slash=True, \
+                                        remove_slash=False, \
+                                        merge_slashes=True, \
+                                        redirect_class=HTTPMovedPermanently)
 
-  Middleware factory which produces a middleware that normalizes
-  the path of a request. By normalizing it means:
+   Middleware factory which produces a middleware that normalizes
```
asvetlov, line 2770: In sphinx the normal indentation is 3 spaces.
We use this default everywhere in aiohttp docs.
Why did you change it?

kryptn: The indentation in that section was two, it's now three. I noticed the difference when adding the `.. versionadded::` tag.

asvetlov: Got it. Sorry, spaces are hard to count in github review tool

kryptn: No problem, I even had to open it back up in my editor to verify because it did look like four in github. 

Thanks!

Issue #3134, opened 2018-07-12
make_mocked_request's default app should be a MagicMock


butla:
## Long story short
App is, among other things, a container for objects with an application-wide use. The default right now is to populate the `app` in the `Request` returned by `make_mocked_request` with `unittest.mock.Mock`, which doesn't support indexing, thus crashing when values in the `app` are accessed, unless MagicMock (or a dictionary) is explicitly put as `app` in the the call to `make_mocked_request`. I think the default should be to create the app as a MagicMock.

I couldn't test if that would brake any tests, because running tox was just spewing missing .h file errors at me and wouldn't run anything.

## Expected behaviour

```python
>>> from aiohttp.test_utils import make_mocked_request
>>> r = make_mocked_request('GET', '/bla')
>>> r.app['some_value']
<MagicMock name='mock.__getitem__()' id='139715764977336'>
```

## Actual behaviour

```python
>>> from aiohttp.test_utils import make_mocked_request
>>> r = make_mocked_request('GET', '/bla')
>>> r.app['some_value']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'Mock' object does not support item assignment
'Mock' object does not support item assignment
```

## Your environment
Aiohttp server: 3.3.2
OS: Kubuntu 17.10 x64

asvetlov:
Well, make sense.
Would you create a pull request for the proposed fix?
A simple unit test is required.

butla:
@asvetlov I'd need to know how to set up the tests, because trying to run tox on master failed miserably. Or should I base the PR on a tagged release commit?

webknjaz:
@butla, `tox` is not yet supported (basically I'm trying to convince Andrew to use it as an underlying layer here, in a smaller project https://github.com/aio-libs/multidict/pull/263), so you'll have to relay on `make` (see `Makefile`), also, you can find inspiration in `.travis.yml`.
Would this help?

asvetlov:
@butla https://docs.aiohttp.org/en/stable/contributing.html has some info about contributions into the project. A fork from the current master should work perfectly fine.

@webknjaz sorry for the delay about tox decision. Will do it soon.

butla:
[...]

PR #3174, opened 2018-08-05
make_mocked_request: subscriptable default app

butla:
## What do these changes do?

App in the default request created with make_mocked_request won't crash when accessed with `[]`.

Nope.

## Related issue number

#3134

### DIFF
--- a/CHANGES/3174.feature
+++ b/CHANGES/3174.feature
@@ -1,0 +1,2 @@
+The default ``app`` in the request returned by ``test_utils.make_mocked_request``
+can now have objects assigned to it and retrieved using the ``[]`` operator.
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -448,7 +448,17 @@ def teardown_test_loop(loop, fast=False):
 
 
 def _create_app_mock():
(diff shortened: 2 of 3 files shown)


asvetlov:
The PR doesn't crash, it's true.
I wonder is it helpful?
```
req.app['key'] = 'val'
assert req.app['key'] == 'val'
```
fails.
Adding a mock for `app.__getitem__` works but it looks complicated.

What do you think?

butla:
Well, it'd be useful for some of the tests I have, where something is accessed from the app, but whatever gets accessed doesn't get used in a meaningful way for the logic.

To make it more useful I could implement `__setitem__` and `__getitem__` inside `_create_app_mock` more or less like this:

```python
def set_dict(app, key, value):
    app.__app_dict[key] = value

def get_dict(app, key):
     return app.__app_dict[key]

app = MagicMock()
app.__app_dict = {}
app.__getitem__ = get_dict
app.__setitem__ = set_dict
```

asvetlov:
yes, it is more robust

butla:
@asvetlov I've updated the code.

butla:
@asvetlov Added.

PR #3189, opened 2018-08-12
Require cython when building from git clone only

### LINKED ISSUE #3162: Build/install broken without cython

asvetlov:
Fixes #3162

Supersedes #3164


setup.py
```diff
@@ -12,13 +12,27 @@
...
+if (here / '.git').exists() and not USE_CYTHON:
+    print("Install cython when building from git clone")
+    print("pip install cython")
+    sys.exit(1)
+
+
+if (here / '.git').exists() and not (here / 'vendor/http-parser/README.md'):
+    print("Install submodules when building from git clone")
+    print("git submodule init")
+    print("git submodule update")
```
kxepal: It's simpler to do `git submodule update --init`.

kxepal: Also, how about to do that automagically if condition is met? Since we know which commands are need to run. Suddenly, it wouldn't be useful for Cython - setup_requires can't handle wheels /:

asvetlov: It is for dev module only.
I prefer don't hide git commands from user here.
There is a rare possibility to have no `git` command in `PATH`, better to give a hint only

kxepal: Ok, that's fine for me.

setup.py
```diff
@@ -12,13 +12,26 @@
...
 
 try:
     from Cython.Build import cythonize
     USE_CYTHON = True
 except ImportError:
     USE_CYTHON = False
 
+if (here / '.git').exists() and not USE_CYTHON:
+    print("Install cython when building from git clone")
+    print("pip install cython")
```
webknjaz: Add this to `pyproject.toml` and it'll be autoinstalled for pip 10+ users.

setup.py
```diff
@@ -12,13 +12,26 @@
...
+here = pathlib.Path(__file__).parent
 
 try:
     from Cython.Build import cythonize
     USE_CYTHON = True
 except ImportError:
     USE_CYTHON = False
 
+if (here / '.git').exists() and not USE_CYTHON:
+    print("Install cython when building from git clone")
```
webknjaz: Please use stderr

setup.py
```diff
@@ -12,13 +12,26 @@
...
+    print("pip install cython")
+    sys.exit(1)
+
+
+if (here / '.git').exists() and not (here / 'vendor/http-parser/README.md'):
+    print("Install submodules when building from git clone")
+    print("git submodule update --init")
```
webknjaz: Try calling this from here it'd be smooth for users then.

.travis.yml
```diff
@@ -152,12 +152,16 @@ jobs:
   - <<: *_doc_base
     env:
     - TARGET=towncrier
+    install:
+    - pip install -r requirements/towncrier.txt
```
webknjaz: You're also overriding `- *upgrade_python_toolset` here.

Issue #3201, opened 2018-08-18
compression + encoding should probably happen in executor


thehesiod:
I just implemented a watchdog utility for our aiohttp servers which tracks when the main thread hangs for more than 5 seconds, and it caught the below callstack.  This response returned a large json object with compression enabled.  Based on the callstack I think probably compression should be done in a background thread otherwise the web-server can get blocked for several seconds and cause the health checks to fail (note in AWS the default is 5s so if any operation takes more than 5s your server may get recycled).

```

  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/fbn/fbn.com/api/weather/services/weatherapi/service.py", line 1430 in <module>
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/fbn/fbn.com/api/weather/services/weatherapi/service.py", line 1426 in start_app
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/site-packages/fbn_async_server/entrypoint.py", line 237 in run_server
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/asyncio/base_events.py", line 422 in run_forever
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/asyncio/base_events.py", line 1434 in _run_once
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/asyncio/events.py", line 145 in _run
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/site-packages/fbn_async_server/entrypoint.py", line 44 in start
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 398 in start
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/site-packages/aiohttp/web_response.py", line 300 in prepare
  | Aug 18 00:08:25.467 |   |   | WeatherAPIInternalService | File "/usr/local/lib/python3.6/site-packages/aiohttp/web_response.py", line 605 in _start
[...]

PR #3205, opened 2018-08-20
body thread helpers

thehesiod:
work on resolving https://github.com/aio-libs/aiohttp/issues/3201


thehesiod:
I also fixed some subclass methods which didn't match the parent class parameter declarations.

aiohttp/web_response.py
```diff
@@ -33,6 +34,10 @@ class ContentCoding(enum.Enum):
 # HTTP Response classes
 ############################################################
 
+# Length in bytes of body which will trigger sync methods to process in a
+# thread pool to avoid blocking the main thread
+_BODY_LENGTH_THREAD_CUTOFF = 1024
```
kxepal: Could we have this configurable?

aiohttp/web_response.py
```diff
@@ -635,3 +654,25 @@ def json_response(data=sentinel, *, text=None, body=None, status=200,
             text = dumps(data)
     return Response(text=text, body=body, status=status, reason=reason,
                     headers=headers, content_type=content_type)
+
+
+async def async_json_response(data=sentinel, *, text=None, body=None,
```
kxepal: I think we shouldn't have this function. While it may not be simple to improve `json_response`, it's much more better to explain users to do dump outside of it call and pass the result as `text` argument to original `json_response`.

aiohttp/web_response.py
```diff
@@ -597,26 +604,38 @@ def content_length(self, value):
...
         if self._body_payload or self._chunked:
             return super()._do_start_compression(coding)
         if coding != ContentCoding.identity:
             # Instead of using _payload_writer.enable_compression,
             # compress the whole body
             zlib_mode = (16 + zlib.MAX_WBITS
                          if coding.value == 'gzip' else -zlib.MAX_WBITS)
+            if len(self._body) > _BODY_LENGTH_THREAD_CUTOFF:
+                loop = asyncio.get_event_loop()
+                await loop.run_in_executor(None,
```
kxepal: What if default executor will be process one? That may hurt instead of making things better.

aiohttp/web_response.py
```diff
@@ -597,26 +604,38 @@ def content_length(self, value):
...
+                                           self._compress_body(zlib_mode))
```
kxepal: Here seems be a typo since compression will happens outside of executor.

aiohttp/web_response.py
thehesiod: that's already a problem in aiohttp's use of `getaddrinfo`

aiohttp/web_response.py
thehesiod: where would you suggest this be settable?

aiohttp/web_response.py
thehesiod: I think we should because it plays into the `_BODY_LENGTH_THREAD_CUTOFF` value, further I think `json_response` should then be deprecated

aiohttp/web_response.py
thehesiod: doh!  ty!  still working through tests =)

aiohttp/web_response.py
kxepal: Perfect place would be Application, but real one could be Response. If you want custom tricky things you can construct Response object with proper settings. In the end, you only know which views returns big responses which could get any benefits from compression/json.dump in executor.

aiohttp/web_response.py
kxepal: That's all because global private constant `_BODY_LENGTH_THREAD_CUTOFF`. Once it will be configurable, composition or creating custom object instead will be the answer. Say, `JsonResponse` one which will do dump properly in executor and old `json_response` will be just a handy helper for compability reasons. Users even wouldn't need to change their code a lot to gain benefits from defaults.

thehesiod: ya but remember the original reason for this helper method, it's to avoid people from having to re-implement things the same way, I think if we can provide something that helps reduce people typing the same code over and over it's a win.   With the above parameterization change I think it makes even more sense, let me know what you think.  Sorry going to keep pushing for a bit more because I know for me it reduces a bunch of our services from having to do this independently.

aiohttp/web_response.py
thehesiod: made param

aiohttp/web_response.py
```diff
@@ -597,26 +603,37 @@ def content_length(self, value):
...
                          if coding.value == 'gzip' else -zlib.MAX_WBITS)
+            if len(self._body) > self._zlib_thread_size:
+                await self._loop.run_in_executor(
+                    None, self._compress_body, zlib_mode)
+            else:
+                self._compress_body(zlib_mode)
+
             compressobj = zlib.compressobj(wbits=zlib_mode)
-            self._compressed_body = compressobj.compress(self._body) +\
+            self._compressed_body = compressobj.compress(self._body) + \
```
thehesiod: forgot to remove this, will be removed in next submit

aiohttp/web_response.py
```diff
@@ -440,7 +441,8 @@ class Response(StreamResponse):
 
     def __init__(self, *, body=None, status=200,
                  reason=None, text=None, headers=None, content_type=None,
-                 charset=None):
+                 charset=None, loop=None,
+                 zlib_thread_size=None):
         if body is not None and text is not None:
```
thehesiod, line 486: I'm not adding a loop param because this was not required before, and if we were to do `self._loop = loop or asyncio.get_event_loop()` it may fail due to there not being an active loop.

thehesiod:
hrm, codecov stuck

thehesiod:
failed due to: `The job exceeded the maximum log length, and has been terminated.`

another error: `aiohttp.client_exceptions.ClientConnectorSSLError: Cannot connect to host 127.0.0.1:36583 ssl:None [[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:871)]`  doesn't seem like I caused these

aiohttp/web_response.py
```diff
@@ -647,3 +661,27 @@ def json_response(data=sentinel, *, text=None, body=None, status=200,
...
+    if data is not sentinel:
+        if text or body:
+            raise ValueError(
+                "only one of data, text, or body should be specified"
+            )
+        else:
+            if asyncio.iscoroutinefunction(dumps):
+                text = await dumps(data)
+            elif executor_body_size is not None and \
+                    len(data) > executor_body_size:
```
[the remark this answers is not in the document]
webknjaz: M/b try (recursive) sizeof?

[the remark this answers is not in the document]
kxepal: `sizeof` will require you somehow convert objects memory usage into string length, which wouldn't be trivial. And all this complexity just to prevent users explicitly define behaviour on their side with two lines of code?

asvetlov:
Please merge with master.
I've reduced output log size (I hope) by disabling `pytest-sugar` plugin.
The plugin is pretty nice, I'm still using it locally but it writes to many terminal control symbols to make the output colorized and interactive.

aiohttp/web_response.py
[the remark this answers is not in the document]
webknjaz: Yeah, I agree with users needing to just opt-in by themselves.
As for converting to string length, it's not needed, just bytes it occupies in memory would be enough :)

webknjaz:
@asvetlov sugar is not the issue. The issue is that you run a lot of tests, which I'm trying to address for a long time now. You should be running test suite in just one mode per job, it'd help dramatically. That's one of the reasons of me pushing towards using tox or smth similar to unify testing layer, maybe parallelise it + possibly distribute across CIs.

webknjaz:
@thehesiod this PR needs conflict resolution (rebase or merge)

aiohttp/web_response.py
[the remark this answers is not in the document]
thehesiod: hah!  wow that was a huge oversight, thanks...however thinking about this, I still need something like a sizeof, I don't know how large the object is without doing the dump ;) I'll have to think about this for a bit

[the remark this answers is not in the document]
thehesiod: removing this from this PR, thanks guys and sorry for the distraction...however it remains a problem to be solved.  Given a dict, to programmatically determine if it should be dumped in a thread or not.  I suppose you just need to decide if you always want to run it in a thread or not.

aiohttp/web_response.py
thehesiod: removed per discussion below

aiohttp/web_response.py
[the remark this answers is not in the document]
webknjaz: No worries :)
One of the solutions might be recursively accumulating the size of stuff in the nested object with interruption of the process once reached some threshold as a flag for compression being needed as an optimization.

thehesiod:
@webknjaz the merge from master brought in those changes, not sure how to revert?

thehesiod:
ok I think I fixed it, followed: https://stackoverflow.com/a/23668025.  Not sure how that happened as I never touched the submodules

aiohttp/web_response.py
```diff
@@ -609,27 +613,36 @@ def content_length(self, value):
...
             # Instead of using _payload_writer.enable_compression,
             # compress the whole body
             zlib_mode = (16 + zlib.MAX_WBITS
                          if coding.value == 'gzip' else -zlib.MAX_WBITS)
-            compressobj = zlib.compressobj(wbits=zlib_mode)
-            self._compressed_body = compressobj.compress(self._body) +\
-                compressobj.flush()
+            if self._zlib_thread_size is not None and \
+                    len(self._body) > self._zlib_thread_size:
+                await asyncio.get_event_loop().run_in_executor(
```
webknjaz, line 680: Wouldn't it be better to have a thread pool in place?

kxepal: I think it would. If someone will set default executor as mutiprocessing one he will get very surprised by aiohttp behaviour. However, inplace thread pool will cause DoS situation - you need to limit overall threads by some sane number.

aiohttp/web_response.py
asvetlov: Fine with it

aiohttp/web_response.py
asvetlov: Multiprocessing pool as default is deprecated by asyncio: it doesn't work well with API like `loop.getaddrinfo()` anyway.

thehesiod: Perhaps ability to pass in thread pool? This can get complicated later if you want to support multiple pools

asvetlov: Make sense. Feel free to update the PR.

asvetlov:
Sorry, I broke unit tests trying to resolve merge conflicts.
@thehesiod would you take a look?

docs/web_reference.rst
```diff
@@ -830,6 +831,10 @@ Response
    :param str charset: response's charset. ``'utf-8'`` if *text* is
                        passed also, ``None`` otherwise.
 
+   :param int zlib_executor_size: length in bytes which will trigger zlib compression
+                            of body to happen in an executor
+
```
asvetlov, line 836: Please insert `.. versionadded:: 3.5` directive here.

docs/web_reference.rst
```diff
@@ -830,6 +831,10 @@ Response
...
+   :param int zlib_executor: executor to use for zlib compression
```
asvetlov, line 839: `.. versionadded::` again

docs/web_reference.rst
asvetlov: aiohttp `3.5.0` was not released yet, only `3.5.0 alpha 0`.
So please use just `3.5`, the last zero should be omitted.

docs/web_reference.rst
```diff
@@ -833,8 +833,10 @@ Response
 
    :param int zlib_executor_size: length in bytes which will trigger zlib compression
                             of body to happen in an executor
+        .. versionadded:: 3.5.1
 
    :param int zlib_executor: executor to use for zlib compression
+        .. versionadded:: 3.5.1
```
asvetlov: `3.5` as above.

thehesiod:
ty guys for the help!

Issue #3208, opened 2018-08-21
TCP not closing properly after POST


Swamp-Ig:
## Long story short

I'm connecting to an IOT device (airconditioner) with a REST interface.

Posting commands using >=3.0 doesn't work. The post goes through and the status OK is returned, but there's some issue with the TCP connection. 

## Steps to reproduce

Within the [wireshark packet capture](https://github.com/aio-libs/aiohttp/files/2306190/AIO-Issue.zip) there's two different TCP transactions with exactly the same aiohttp code running (also within the zip file). The short version of the code without imports etc looks like this:

```python
device_ip = '10.0.0.107'

async def send_command_async(command: str, data: Any):
    body = {command : data}
    url = f"http://{device_ip}/{command}"
    async with ClientSession() as session:
        async with session.post(url, json=body) as response:
            response.raise_for_status()

loop: AbstractEventLoop = asyncio.get_event_loop()
loop.run_until_complete(loop.create_task(send_command_async('SystemMODE', 'heat')))
```

If you get aiohttp<3.0 you get the first packet capture, up until packet 10. Everything is happy.

With the latest version and any version above 3.0, all goes well until the ACK gets sent for the OK packet, then a whole spume of out of order and dup ack packets goes back and forth.

I can't actually see any difference in what gets sent or anything. It's possible it's the device that's the issue, I've emailed the guys that produce the device so they are on this issue too, but the low level TCP stuff is a bit beyond me.

The odd thing is that http GET works just fine.

## Environment

* Python 3.6
* Windows 10
* Latest aiohttp (but did a regression test back until <3.0 version)
* The device is an [iZone air conditioner](http://izone.com.au/)

webknjaz:
I assume that the issue _might_ be within HTTP parser in C from nodejs that we use.
You may try downgrading git submodule in `vendor/http-parser` folder to some version in past. Let's say some commit a year or two ago. This should make it clear whether my guess is related.

Swamp-Ig:
I just tried python 3.7... didn't help. Will do as you suggest

*Edit:*
Err....  I would do as you suggest, but I'll have to figure out how to do it. I did get as far as checking out aiohttp I've never really used python until a couple of weeks ago.

webknjaz:
Oh, just make sure that you do `git clone` with `--recurse-submodules`, then go to the folder I mentioned and `git checkout` to some old commit.
[...]

PR #3209, opened 2018-08-23
Bump vendored http-parser to v 77310ee commit

webknjaz:


asvetlov:
@webknjaz I doubt if we should bump to not-released-yet versions of `http-parser`.
According to their log the last is 2.8.1: https://github.com/nodejs/http-parser/releases/tag/v2.8.1

webknjaz:
Fair enough. But it has some fixes, which may be related to #3208

asvetlov:
Which commit from http-parser unreleased master fixes #3208 ?

webknjaz:
It seems to be https://github.com/nodejs/http-parser/commit/a7c2e8626bf6b5ddc0c891a88cb77ff9b979f006 from history, but needs verification.

asvetlov:
Well. Let's keep the PR but avoid bumps to unreleased versions in the future.

webknjaz:
Got it

Swamp-Ig:
Can confirm that doing this fixed #3208

asvetlov:
Ok, thank you

PR #3235, opened 2018-09-02
Make METHRE RFC-7230 compliant

kxepal:
Definition of method is:
```
    method = token
    tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
            "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
    token = 1*tchar
```

So he had two issues:
1. Not all the characters were allowed.
2. Actually, we did allowed too much characters since `$-_` parsed as:
"all characters in range between code 36 ($) and code 95 (_)" instead of
"characters with codes 36, 45 and 95". So we did match methods like
`[GET]` which are malformed according the spec.

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -29,7 +29,15 @@ __all__ = (
     'RawRequestMessage', 'RawResponseMessage')
 
 ASCIISET = set(string.printable)
-METHRE = re.compile('[A-Z0-9$-_.]+')
+
+# See https://tools.ietf.org/html/rfc7230#section-3.1.1
+# and https://tools.ietf.org/html/rfc7230#appendix-B
+#
+#     method = token
+#     tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
+#             "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
+#     token = 1*tchar
+METHRE = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
 VERSRE = re.compile(r'HTTP/(\d+).(\d+)')
 HDRRE = re.compile(rb'[\x00-\x1F\x7F()<>@,;:\[\]={} \t\\\\\"]')
 
--- a/tests/test_http_parser.py
+++ b/tests/test_http_parser.py
@@ -540,7 +540,7 @@ def test_http_request_parser_two_slashes(parser):
 
 def test_http_request_parser_bad_method(parser):
     with pytest.raises(http_exceptions.BadStatusLine):
-        parser.feed_data(b'!12%()+=~$ /get HTTP/1.1\r\n\r\n')
+        parser.feed_data(b'=":<G>(e),[T];?" /get HTTP/1.1\r\n\r\n')
 
 
 def test_http_request_parser_bad_version(parser):
--- a/tests/test_web_protocol.py
+++ b/tests/test_web_protocol.py
@@ -239,7 +239,7 @@ async def test_simple(srv, loop, buf):
 
 async def test_bad_method(srv, loop, buf):
     srv.data_received(
-        b'!@#$ / HTTP/1.0\r\n'
+        b':BAD; / HTTP/1.0\r\n'
         b'Host: example.com\r\n\r\n')
 
     await asyncio.sleep(0, loop=loop)
--- a/CHANGES/3235.bugfix
+++ b/CHANGES/3235.bugfix
@@ -1,0 +1,1 @@
+Make method match regexp RFC-7230 compliant


kxepal:
After review of https://github.com/aio-libs/aiohttp/pull/3233 I found this inconsistentcy between our regex and the spec.  If fix is ok, will finish the rest bits.

kxepal:
Oh, I didn't run full test suite, just test_http_parser. Now should be fixed.

tests/test_web_protocol.py
```diff
@@ -239,7 +239,7 @@ def test_srv_keep_alive_disable(srv):
 
 async def test_bad_method(srv, loop, buf):
     srv.data_received(
-        b'!@#$ / HTTP/1.0\r\n'
+        b':BAD; / HTTP/1.0\r\n'
```
webknjaz, line 242: Is there a test case for having multiple spaces in Request-Line? (>2)
What about test with non-ascii?

kxepal: If we want to have such tests, we would need [hypothesis](https://github.com/HypothesisWorks/hypothesis/tree/master/hypothesis-python) to generate such malformed requests. That's a bit bigger story than fixing a single regexp, but certainly will improve tests quality a lot.

asvetlov: If somebody wants to write hypothesis tests -- you are welcome (requires another PR)

webknjaz: Sure, just wanted to point out what I personally faced with when redesigning another HTTP parser :)

aiohttp/http_parser.py
```diff
@@ -29,7 +29,7 @@
     'RawRequestMessage', 'RawResponseMessage')
 
 ASCIISET = set(string.printable)
-METHRE = re.compile('[A-Z0-9$-_.]+')
+METHRE = re.compile("[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
```
webknjaz: I think It might make sense to put a comment with that rule somewhere nearby:
```
#     method = token
#     tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
#             "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
#     token = 1*tchar
```

webknjaz: Yeah, I know about blame. But it's visually more convenient to see it next to the code. Like it's done in yarl, for example.

aiohttp/http_parser.py
```diff
@@ -29,7 +29,15 @@
...
-METHRE = re.compile('[A-Z0-9$-_.]+')
+
+# See https://tools.ietf.org/html/rfc7230#section-3.1.1
+# and https://tools.ietf.org/html/rfc7230#appendix-B
+#
+#     method = token
+#     tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." /
+#             "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
+#     token = 1*tchar
+METHRE = re.compile("[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
```
webknjaz: I'd use a raw-string literal `r''`

webknjaz: Also, `+` and `*` might also need escaping and `|` needs escaping (otherwise it divides groups in `[]`)

webknjaz: Oh, and `.` might stand for any single character.

webknjaz: Probably they have restricted meanings within `[]`, but IMHO better safe than sorry.

tests/test_http_parser.py
```diff
@@ -540,7 +540,7 @@ def test_http_request_parser_two_slashes(parser):
 
 def test_http_request_parser_bad_method(parser):
     with pytest.raises(http_exceptions.BadStatusLine):
-        parser.feed_data(b'!12%()+=~$ /get HTTP/1.1\r\n\r\n')
+        parser.feed_data(b'=":<G>(e),[T];?" /get HTTP/1.1\r\n\r\n')
```
webknjaz, line 543: Add `|`, `+`, `-` and `.`

aiohttp/http_parser.py
kxepal: All true, but not for `[]` sets.
```
>>> import re
>>> re.match('[.]', 't')
>>> re.match('[.]', 't') is None
True
```

tests/test_http_parser.py
kxepal: Why? These characters are valid.

aiohttp/http_parser.py
webknjaz: digits can be replaced with `\d` (to match style below)

tests/test_http_parser.py
asvetlov: `with pytest.raises` at least.
Personally I dont insist on pytest parametrization

asvetlov:
@kxepal please merge when ready

PR #3239, opened 2018-09-04
Enhanced basic auth decode

dalazx:
## What do these changes do?

These changes prevent passing illegal chars in the base64 payload.
It was possible to use `Authorization: Basic ???` to get `BasicAuth(login='', password='')` without exceptions.
Also, the related RFC https://www.ietf.org/rfc/rfc2617.txt allows the username and password to be blank, but the colon must be present.
```
      credentials = "Basic" basic-credentials
      basic-credentials = base64-user-pass
      base64-user-pass  = <base64 [4] encoding of user-pass, except not limited to 76 char/line>
      user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT
```
and
```
curl -vv -u '' example.com
Authorization: Basic Og==  # which is just ":"
```

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -95,20 +95,29 @@ class BasicAuth(namedtuple('BasicAuth', ['login', 'password', 'encoding'])):
     @classmethod
     def decode(cls, auth_header: str, encoding: str='latin1') -> 'BasicAuth':
         """Create a BasicAuth object from an Authorization HTTP header."""
-        split = auth_header.strip().split(' ')
-        if len(split) == 2:
-            if split[0].strip().lower() != 'basic':
-                raise ValueError('Unknown authorization method %s' % split[0])
-            to_decode = split[1]
-        else:
+        try:
+            auth_type, encoded_credentials = auth_header.split(' ', 1)
+        except ValueError:
             raise ValueError('Could not parse authorization header.')
 
+        if auth_type.lower() != 'basic':
+            raise ValueError('Unknown authorization method %s' % auth_type)
+
         try:
-            username, _, password = base64.b64decode(
-                to_decode.encode('ascii')
-            ).decode(encoding).partition(':')
+            decoded = base64.b64decode(
+                encoded_credentials.encode('ascii'), validate=True
+            ).decode(encoding)
         except binascii.Error:
             raise ValueError('Invalid base64 encoding.')
+
+        try:
+            # RFC 2617 HTTP Authentication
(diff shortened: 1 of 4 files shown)


tests/test_helpers.py
```diff
@@ -83,8 +84,10 @@ def test_basic_auth4():
     assert auth.encode() == 'Basic bmtpbTpwd2Q='
 
 
-def test_basic_auth_decode():
-    auth = helpers.BasicAuth.decode('Basic bmtpbTpwd2Q=')
+@pytest.mark.parametrize('header', (
+    'Basic bmtpbTpwd2Q=', 'basic bmtpbTpwd2Q='))
```
webknjaz: Nitpick: if placed one under the other, sequence items are better readable.

aiohttp/helpers.py
```diff
@@ -95,21 +95,30 @@ def __new__(cls, login: str,
...
+        if auth_type.lower() != 'basic':
+            raise ValueError('Unknown authorization method %s' % auth_type)
+
         try:
-            username, _, password = base64.b64decode(
-                to_decode.encode('ascii')
-            ).decode(encoding).partition(':')
+            decoded = base64.b64decode(
+                encoded_credentials.encode('ascii'), validate=True
+            ).decode(encoding)
```
webknjaz, line 109: Not sure whether it's applicable to current PR, but JFYI some browsers (Firefox) don't encode some of UTF-8 bytes correctly, assuming `ISO-8859-1` for unicode input and loosely encodes that input, which cuts some bytes in two-byte encoded chars (try entering `€öäü` in browser and see what you receive in server), which results in error during encoding since it cannot understand byte sequence when reaches those characters. Ref https://github.com/cherrypy/cherrypy/issues/1680

asvetlov: Thanks for the reminder.
We discussed it 2 or 3 years ago and decided to do nothing until users report.
There is no blame yet :)

setup.cfg
```diff
@@ -34,6 +34,8 @@ branch = True
 source = aiohttp, tests
 omit = site-packages
 
+[mypy]
+incremental = false
```
asvetlov, line 38: Related mypy issue: python/mypy#5534

aiohttp/helpers.py
dalazx: thanks @webknjaz. noted, but it seems to be out of scope of this PR though.

asvetlov:
Thanks!

PR #3267, opened 2018-09-13
Quiet warnings about netrc

nicktimko:
## What do these changes do?

If there isn't a .netrc file specified by an environment variable, it can be confusing to see warnings about it. If NETRC isn't set, don't warn. Only warn if one of

* can't resolve HOME
* can load, but can't parse file
* can't find file
* file appears to exist at the default location but is unreadable for some reason.

When using the client from a command line application, the user will no longer see `could't find .netrc file` if `~/.netrc` is missing and/or the envvar `NETRC` isn't set.

## Related issue number

None.

## Checklist

- [x] I think the code is well written: Tried to reduce how much code the try/except was catching, added some early exits to reduce number of conditionals
- [ ] Unit tests for the changes exist: *n/a: UI change?*
- [ ] Documentation reflects the changes: *n/a, UI change?*
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`: *n/a trivial change*
- [ ] Add a new news fragment into the `CHANGES` folder: *trivial change?*

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -9,6 +9,7 @@ import functools
 import inspect
 import netrc
 import os
+import platform
 import re
 import sys
 import time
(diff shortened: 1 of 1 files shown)


aiohttp/helpers.py
```diff
@@ -146,30 +146,38 @@ def strip_auth_from_url(url: URL) -> Tuple[URL, Optional[BasicAuth]]:
...
+    else:
+        try:
             home_dir = Path.home()
-            if os.name == 'nt':  # pragma: no cover
-                netrc_path = home_dir.joinpath('_netrc')
-            else:
-                netrc_path = home_dir.joinpath('.netrc')
+        except RuntimeError as e:  # pragma: no cover
+            # if pathlib can't resolve home, it may raise a RuntimeError
+            client_logger.warning("Could not find .netrc file: %s", e)
```
webknjaz: This message seems to be misleading

aiohttp/helpers.py
```diff
@@ -146,30 +146,38 @@ def strip_auth_from_url(url: URL) -> Tuple[URL, Optional[BasicAuth]]:
...
-            except (netrc.NetrcParseError, OSError) as e:
-                client_logger.warning(".netrc file parses fail: %s", e)
-
-        if netrc_obj is None:
-            client_logger.warning("could't find .netrc file")
-    except RuntimeError as e:  # pragma: no cover
-        """ handle error raised by pathlib """
-        client_logger.warning("could't find .netrc file: %s", e)
-    return netrc_obj
+        if os.name == 'nt':  # pragma: no cover
```
webknjaz: Use `platform.system()` instead

webknjaz: `netrc_path = home_dir / ('_netrc' if platform.system() == 'Windows' else '.netrc')`

agronholm:
I also couldn't understand why these warnings were printed in the first place.

nicktimko:
I dunno, if I just submit a diff where I delete all of them, that fine?

asvetlov:
Thanks!

agronholm:
@nicktimko Why is it even necessary to print a warning if the home directory cannot be resolved? A typical Docker container will not have a resolvable home directory unless one is explicitly created in the Dockerfile.

asvetlov:
If your code is not supposed to work with `trust_env=True` -- why do you want to pass the flag at all?

nicktimko:
@agronholm This was a bit monkey-see, monkey-do. I was trying to preserve the existing behavior that I didn't understand, as I haven't used a `.netrc` file didn't fully grok it. I was going on the assumption that *the existing code was correct* (#2584) for most circumstances, and I don't think I changed much in the functionality aside from the narrow case that I cut out (nothing set in the env, the file doesn't exist.)

agronholm:
> If your code is not supposed to work with trust_env=True -- why do you want to pass the flag at all?

What are you talking about? I use `trust_env=True` so it could pick up my proxy settings from the environment variables.

asvetlov:
Ok, I see nothing bad if we will suppress a warning if the home directory doesn't exist..

@agronholm would you prepare a pull request?

agronholm:
Sure, I'll try to get that done this week.

PR #3282, opened 2018-09-22
Make Request.post() also work with content-type application/json

marconifeo:
## What do these changes do?

Make Request.post() also work with content-type application/json. So the user doesn't have to check content-type to use either Request.post() or Request.json(). Added test that fails without the change, and that passes with the change.

When a user calls Request.post() for a request with json data in the body and content-type application/json, he/she will get the data as expected (instead of empty Multidict and having to investigate and discover that in this case Request.json() should be used instead of Request.post())

## Checklist

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -135,6 +135,7 @@ Makc Belousow
 Manuel Miranda
 Marat Sharafutdinov
 Marco Paolini
+Marconi Feo
 Mariano Anaya
 Martin Melka
 Martin Richard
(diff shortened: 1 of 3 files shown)


marconifeo:
I added a condition and added to the test that accesses that method. I first wrote the test and saw it fail. Then I wrote the code and the test passed. How can it have less coverage? Can someone please point me in the right direction to fix this? Thank you.

asvetlov:
Thanks for PR.
Sorry, but I decline it.

Let me describe the decision.

On request processing you already have a context: is it REST API call, HTML form or something special.
`req.post()` is for HTML forms, `req.json()` is for API.

'application/x-www-form-urlencoded' and 'multipart/form-data' are for handling HTTP forms (sent by different content-type, sure).

`application/json` is a totally different beast, sorry.
If the web-handler is already made for web forms processing a JSON is not the thing to accept.
The same for REST handler.

What is your use case for the PR?
Maybe I've missed something?

marconifeo:
I understand. Thank you for the explanation.

PR #3294, opened 2018-09-26
Add typings for Payload

kornicameister:
## What do these changes do?

Adds typing for the `aiohttp/payload.py`

No, these are just hints for the type checkers.

## Related issue number

#1749 

## Checklist

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -16,7 +16,7 @@ from .abc import AbstractStreamWriter
 from .helpers import (PY_36, content_disposition_header, guess_filename,
                       parse_mimetype, sentinel)
 from .streams import DEFAULT_LIMIT, StreamReader
-from .typedefs import JSONEncoder
+from .typedefs import JSONEncoder, _CIMultiDict
 
 
 __all__ = ('PAYLOAD_REGISTRY', 'get_payload', 'payload_type', 'Payload',
@@ -102,14 +102,14 @@ class PayloadRegistry:
 class Payload(ABC):
 
     _size = None  # type: Optional[float]
-    _headers = None  # type: Optional[CIMultiDict[str]]
+    _headers = None  # type: Optional[_CIMultiDict]
     _content_type = 'application/octet-stream'  # type: Optional[str]
 
     def __init__(self,
                  value: Any,
                  headers: Optional[
                      Union[
-                         CIMultiDict[str],
+                         _CIMultiDict,
                          Dict[str, str],
                          Iterable[Tuple[str, str]]
                      ]
@@ -142,7 +142,7 @@ class Payload(ABC):
         return self._filename
 
     @property
-    def headers(self) -> Optional[CIMultiDict[str]]:
+    def headers(self) -> Optional[_CIMultiDict]:
         """Custom item headers"""
         return self._headers
 
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -369,7 +369,7 @@ else:
 
 class AsyncIterablePayload(Payload):
 
-    _iter: Optional[_AsyncIterator] = None
+    _iter = None  # type: Optional[_AsyncIterator]
 
     def __init__(self,
                  value: _AsyncIterable,
@@ -393,7 +393,7 @@ class AsyncIterablePayload(Payload):
                 # iter is not None check prevents rare cases
                 # when the case iterable is used twice
                 while True:
-                    chunk: bytes = await self._iter.__anext__()
+                    chunk = await self._iter.__anext__()
                     await writer.write(chunk)
             except StopAsyncIteration:
                 self._iter = None
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -44,13 +44,14 @@ def get_payload(data: Any, *args: Any, **kwargs: Any) -> 'Payload':
 
 def register_payload(factory: Type['Payload'],
                      type: Any,
-                     *, order: Order=Order.normal) -> None:
+                     *,
+                     order: Order=Order.normal) -> None:
     PAYLOAD_REGISTRY.register(factory, type, order=order)
 
 
 class payload_type:
 
-    def __init__(self, type: Any, order: Order=Order.normal) -> None:
+    def __init__(self, type: Any, *, order: Order=Order.normal) -> None:
         self.type = type
         self.order = order
 
@@ -115,7 +116,8 @@ class Payload(ABC):
                  ] = None,
                  content_type: Optional[str]=sentinel,
                  filename: Optional[str]=None,
-                 encoding: Optional[str]=None) -> None:
+                 encoding: Optional[str]=None,
+                 **kwargs: Any) -> None:
         self._value = value
         self._encoding = encoding
         self._filename = filename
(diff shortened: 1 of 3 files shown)


aiohttp/payload.py
```diff
@@ -107,27 +126,27 @@ def __init__(self, value, *, headers=None, content_type=sentinel,
         self._content_type = content_type
 
     @property
-    def size(self):
+    def size(self) -> Optional[float]:
```
kornicameister, line 135: All those `Optional` seem a little awkward, but honestly speaking that's how I read the logic. Might as well be that I got that wrong.

asvetlov:
Thanks for the PR!
Will find a time for review tomorrow.

asvetlov:
> aiohttp/payload.py:191: error: Too many arguments for "__init__" of "Payload"

Since `Payload` accepts the only `value` positional argument and the value is already passed in `super().__init__()` the `*args` should be dropped from `BytesPayload` constructor.
Perhaps the same for other similar errors.

aiohttp/payload.py
```diff
@@ -87,12 +100,18 @@ def register(self, factory, type, *, order=Order.normal):
...
 class Payload(ABC):
 
-    _size = None
-    _headers = None
-    _content_type = 'application/octet-stream'
-
-    def __init__(self, value, *, headers=None, content_type=sentinel,
-                 filename=None, encoding=None, **kwargs):
+    _size = None  # type: Optional[float]
+    _headers = None  # type: Optional[LooseHeaders]
```
asvetlov: `_headers` has the concrete type `Optional[CIMultiDict[str]]`.

`LooseHeaders` is used for passing any mapping-like type.
aiohttp client API converts `LooseHeaders` into `CIMultiDict` internally.

I'm not 100% sure but a payload should accept `CIMultiDict` objects only.

aiohttp/payload.py
```diff
@@ -60,12 +65,16 @@ class PayloadRegistry:
...
+    def __init__(self) -> None:
+        self._first = []  # type: List[Tuple[Type[Payload], Any]]
+        self._normal = []  # type: List[Tuple[Type[Payload], Any]]
+        self._last = []  # type: List[Tuple[Type[Payload], Any]]
 
-    def get(self, data, *args, _CHAIN=chain, **kwargs):
+    def get(self,
+            data: Any,
+            *args: Any,
+            _CHAIN: Callable[..., Tuple[Type['Payload'], Any]]=chain,
```
asvetlov: Use `_CHAIN: Any` here, it's ok to shut up the type checker for the argument.

aiohttp/payload.py
```diff
@@ -29,27 +32,29 @@ class LookupError(Exception):
     pass
 
 
-class Order(enum.Enum):
+class Order(str, enum.Enum):
```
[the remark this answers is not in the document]
kornicameister: Well, all the entries in the enum are derived from string. It's like extra precaution to prevent adding other members that would be derived from different data type. FYI: https://docs.python.org/3.7/library/enum.html#others.

aiohttp/payload.py
kornicameister: @asvetlov Given [this](http://werkzeug.pocoo.org/docs/0.14/datastructures/) documentation, I think that it won't hurt to write typing for `__init__` to reflect all the possible types `CIMultiDict` can accept. I will push the next commit with that and you can check it out later on.

kornicameister:
Ok, so mypy no longer complains, but I am getting `TypeError: __init__() got an unexpected keyword argument 'disposition'` all over the test execution :/. Wonder what I've changed to cause that. I'd assume that adding types shouldn't have nothing to do with changing the actual functionality thus failing tests.

----

NVM, found that. Removed one extra `**kwargs` elsewhere.

aiohttp/typedefs.py
```diff
@@ -23,8 +24,13 @@
     _MultiDictProxy = MultiDictProxy
 
 Byteish = Union[bytes, bytearray, memoryview]
-JSONEncoder = Callable[[Any], str]
-JSONDecoder = Callable[[str], Any]
+JSONObj = Dict[str, Any]
+JSON = Union[
```
asvetlov: Technically JSON can contain a single integer, ``None``  or string.
Moreover, it can be **any** python object if there is python -> JSON converter passed by `json.dums(obj, defaul=my_converter)`.
Let's use `Any` for json encoders/decoders.

aiohttp/payload.py
```diff
@@ -197,20 +232,33 @@ def __init__(self, value, *args,
             if content_type is None:
                 content_type = 'text/plain; charset=%s' % encoding
 
-        super().__init__(
-            value.encode(encoding),
-            encoding=encoding, content_type=content_type, *args, **kwargs)
+        if encoding is None:
```
kornicameister: Ok, this bit I completely do not get. It's literally impossible for encoding to be None, based on the logic. But the typing says otherwise and we can't exactly redefine the value with `mypy`. We can't also do the 

```
_encoding: str
_content_type: str
```

letting the assignment to come later, because tests on `3.5` complains. We could've do the `typing.cast` but not really sure if that's correct. Sounds a bit hacky to me.

@asvetlov any suggestions ?

kornicameister: Alternatively, we can keep that `raise ValueError` and add some `noqa` there, but honestly that sounds bad as well to me.

kornicameister: Oh, and for the sake of discussion, this is the alternative:

```
class StringPayload(BytesPayload):

    def __init__(self,
                 value: str,
                 *args: Any,
                 encoding: Optional[str]=None,
                 content_type: Optional[str]=None,
                 **kwargs: Any) -> None:
        _encoding, _content_type = self._parse_args(encoding, content_type)
        super().__init__(
            value.encode(_encoding),
            encoding=_encoding,
            content_type=_content_type,
            *args,
            **kwargs)

    @staticmethod
    def _parse_args(
        encoding: Optional[str]=None,
        content_type: Optional[str]=None,
    ) -> Tuple[str, str]:
        if encoding and content_type:
            return encoding, content_type
        elif encoding is None and content_type is None:
            return 'utf-8', 'text/plain; charset=utf-8'
        elif encoding is None and content_type:
            mimetype = parse_mimetype(content_type)
            return (
                mimetype.parameters.get('charset', 'utf-8'),
                content_type,
            )
        else:
            return (
                cast(str, encoding),
                'text/plain; charset=%s' % cast(str, encoding),
            )
```

asvetlov: Revert back the change, drop `if encoding is None` check.
Use the following trick:
```
        if encoding is None:
            if content_type is None:
                 real_encoding = 'utf-8'
                 content_type = 'text/plain; charset=utf-8'
             else:
                 mimetype = parse_mimetype(content_type)
                 readl_encoding = mimetype.parameters.get('charset', 'utf-8')
         else:
            if content_type is None:
                content_type = 'text/plain; charset=%s' % encoding
            real_encoding = encoding

 super().__init__(
                value.encode(encoding),
                encoding=real_encoding,
                content_type=content_type,
                *args,
                **kwargs)
```

`real_encoding` variable has strict string type, not `Optional[str]`.

I believe mypy understands it pretty well.

aiohttp/payload.py
```diff
@@ -290,18 +342,39 @@ def size(self):
...
+                 dumps: JSONEncoder=json.dumps,
+                 *args: Any,
+                 **kwargs: Any) -> None:
 
         super().__init__(
             dumps(value).encode(encoding),
             content_type=content_type, encoding=encoding, *args, **kwargs)
 
 
+if TYPE_CHECKING:
```
asvetlov: Please add `# pragma: no cover` comment to the line for hinting coverage tool

asvetlov:
Please use another approach.
`CIMultiDict` is a generic class, strict mypy mode forbids unspecialized generics.
`CIMultiDict[str]` should be used instead.
But it produces a runtime exception, sure.
The easiest backward compatible way to solve it is forward declarations.
Just wrap the problematic annotation in quotes: `'CIMultiDict[str]'` and `Union[`CIMultiDict[str]`, Dict[str, str]]`. The later most likely can be collapsed into `Mapping[str, str]` though.

kornicameister:
Doesnt it mean that _CIMultiDict should be removed and replaced with forward references ? I mean at all...not as a sugestiom for me. Given your comment  I got the feeling I used sth I should not have 😎

asvetlov:
Thanks!
I've missed underscore in `_CIMultiDict` on review.
I discovered forward-declaration based solution after adding `_CIMultiDict` to `typedefs.py`.
Now I have no strong preference, both ways are acceptable.
Let's add type hints everywhere first. Maybe later we'll polish it, I don't care right now.

kornicameister:
Ok, working currently on `web_response`, but that is a huge beast. I am still trying to figure out all the bits and pieces.

asvetlov:
`tracing.py` should be a relatively easy thing.

aiohttp/payload.py
```diff
@@ -87,12 +101,23 @@ def register(self, factory, type, *, order=Order.normal):
 
 class Payload(ABC):
 
-    _size = None
-    _headers = None
-    _content_type = 'application/octet-stream'
-
-    def __init__(self, value, *, headers=None, content_type=sentinel,
-                 filename=None, encoding=None, **kwargs):
+    _size = None  # type: Optional[float]
```
kxepal, line 104: A bit late for the party, but why size is typed as float? This value is used for Content-Length header which defined number of bytes of payload size - it's hard to pass over HTTP some fraction of a byte. What was the case for float type?

kxepal: https://github.com/aio-libs/aiohttp/pull/3484

PR #3303, opened 2018-09-29
Fix #3290 missing parameter 'raise_for_status' for aiohttp.request()

### LINKED ISSUE #3290: Missing raise_for_status parameter to aiohttp.request
## Missing raise_for_status parameter to aiohttp.request
Well the title stands for itself.
Documentation on [aiohttp.request](https://docs.aiohttp.org/en/stable/client_reference.html#basic-api) stands, that this method have argument `raise_for_status`, which was added in version 3.4. 
But looking into [current code](https://github.com/aio-libs/aiohttp/blob/044996a59ad574c101bb4fe5c47a78f6dc0067cc/aiohttp/client.py#L892) of file `aiohttp.client.py` shows that this method do not have such argument.

## Expected behaviour

Expected that this code will work like a charm:
```python
import aiohttp
async with aiohttp.request('GET', 'www.example.com', raise_for_status=True, ) as response:
    pass
```

## Actual behaviour

Get type error:
```python
TypeError: request() got an unexpected keyword argument 'raise_for_status'
```

## Steps to reproduce
```python
import aiohttp
import asyncio

async def fetch():
    async with aiohttp.request('GET', 'http://python.org/', raise_for_status=True) as resp:
        assert resp.status == 200
        print(resp.status)

if __name__ == '__main__':
    loop.run_until_complete(fetch())
```

## Your environment
aiohttp==3.4.4
Client

artslob:
## What do these changes do?

Fix missing parameter 'raise_for_status' for aiohttp.request()

No notable changes

## Related issue number

#3290 

## Backport

I think there is need for backport to 3.4, because [documentaion](https://docs.aiohttp.org/en/stable/client_reference.html#basic-api) says request() have this parameter from 3.4.

## Checklist


asvetlov:
While the documentation mentions `raise_for_status` the code is absent.
Adding a new parameter is a new functionality, it should land on aiohttp 3.5 only.

asvetlov:
Thanks!

Issue #3307, opened 2018-10-01
Server access log format `%b` is the size of the whole response, not excluding headers


cdunklau:
## Long story short

Access log format `%b` is the size of the whole response, not excluding headers. The documentation says it should exclude the headers... "%b - Size of response in bytes, excluding HTTP headers".

https://github.com/aio-libs/aiohttp/blob/708d1f07bb7878c09b47df620cd6b38a9bec470b/docs/logging.rst#format-specification

While this looks like an implementation bug, I don't really understand the rationale behind the idea of excluding the headers from the length calculation... ISTM that the full response size would be a more useful metric. I'd recommend leaving the behavior as-is, and just updating the docs to correctly reflect the behavior.

## Expected behaviour

Returning `web.Response(body=None, status=200)` to result in 0 for the `%b` replacement value, and returning `web.Response(body=b'foo', status=200)` to result in 3.

## Actual behaviour

Results are 149 and 152, respectively. These are the lengths of the entire response, including status line and headers.

## Steps to reproduce

Run this:

```python
import sys
import logging
import asyncio

from aiohttp import web


def nobody(request):
    return web.Response(body=None, status=200)


def somebody(request):
    return web.Response(body=b'foo', status=200)


def init_app(*, loop):
    app = web.Application()
    app.router.add_route('GET', '/nobody', nobody)
    app.router.add_route('GET', '/somebody', somebody)
    return app


def main():
    logging.basicConfig(stream=sys.stderr, level=logging.DEBUG)

    loop = asyncio.get_event_loop()
    app = init_app(loop=loop)
    web.run_app(
        app,
        host="127.0.0.1",
        port=8080,
        access_log_format='%r response_length=%b',
    )

if __name__ == '__main__':
    main()
```

Hit the two endpoints:

```
curl -vv http://localhost:8080/nobody
curl -vv http://localhost:8080/somebody
```

Notice the log lines have the incorrect values for response length, if the documentation is to be believed:

```
INFO:aiohttp.access:GET /nobody HTTP/1.1 response_length=149
INFO:aiohttp.access:GET /somebody HTTP/1.1 response_length=152
```

## Your environment

Mac, Python 3.6.3. Installed versions:

```
$ venv/bin/pip freeze
aiohttp==3.4.4
async-timeout==3.0.0
attrs==18.2.0
chardet==3.0.4
idna==2.7
idna-ssl==1.1.0
multidict==4.4.2
yarl==1.2.6
```

[...]

Issue #1673, opened 2017-02-22
Drop "%O" in access logger


fafhrd91:
implementation for "%O" is wrong. `output_size` was never designed to use as response size.

@asvetlov @kxepal

fafhrd91:
I will change "%O" to use same value as "%b", but we need to decide what to do with it in general

kxepal:
I'm fine with either decision here since I didn't get the idea of these magic tokens at all (:

PR #3312, opened 2018-10-02
Make log format %b doc reflect behavior and docstring

cdunklau:
## What do these changes do?

Update logging doc to reflect behavior and `aiohttp.helpers.AccessLogger` docs, fixing #3307.

I dug into the tests to make sure the full-response-size behavior is tested, but I'm confused about what this test is doing with `response.body_length`, or rather, what that attribute actually means:

https://github.com/aio-libs/aiohttp/blob/4c4401ad06c931929031d919180ee1f1e0c06835/tests/test_helpers.py#L185-L216

ISTM that `body_length` would be the length of the response content, excluding status line, headers, and terminating blank line, but that doesn't reflect how it's used... it looks like that's the length of the entire response, not just the content. @asvetlov could you suggest how to make sure this test exercises that behavior, or should I work on a separate one?

## Checklist

### DIFF
--- a/CHANGES/3307.doc
+++ b/CHANGES/3307.doc
@@ -1,0 +1,2 @@
+Make server access log format placeholder %b documentation reflect
+behavior and docstring.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -45,6 +45,7 @@ Brian Muller
 Bryce Drennan
 Carl George
 Cecile Tonglet
+Colin Dunklau
 Chien-Wei Huang
 Chih-Yuan Chen
 Chris AtLee
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -71,7 +71,7 @@ request and response:
 +--------------+---------------------------------------------------------+
 | ``%s``       | Response status code                                    |
 +--------------+---------------------------------------------------------+
-| ``%b``       | Size of response in bytes, excluding HTTP headers       |
+| ``%b``       | Size of response in bytes, including HTTP headers       |
 +--------------+---------------------------------------------------------+
 | ``%T``       | The time taken to serve the request, in seconds         |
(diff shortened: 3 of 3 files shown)


asvetlov:
Sorry, was pretty busy today.
You've raised a complex question actually.

The source of misunderstanding is:
`%b` uses `resp.body_length` but `.body_length` is calculated from `writer.output_size`.
Ideally we can support both *full message size* and *body length*.

Logger codes were designed after Apache mnemonics: https://httpd.apache.org/docs/2.4/mod/mod_log_config.html#formats
Don't ask me why, but let's be consistent.

`%b` doesn't count http headers size.
To do this in aiohttp we can save `writer.output_size` after `StreamResponse.prepare()` for storing the size of headers. `resp.body_length` can substract the value from `writer.output_size`.

Do you like the idea?

cdunklau:
It looks like "both full message size and body length" were indeed supported... `%O` did exist at one point, and was dropped in 0e99cfb0333f39cc824ef6a8f92cc9b4ad1ef590. At the same time, the docstring was changed to say `%b  Size of response in bytes, including HTTP headers`, as it is now. It looks like that's the change that brought the docstring and the logging.rst doc out of sync. Unfortunately, the discussion in #1673 is not all that enlightening. Perhaps @fafhrd91 and @kxepal can provide their input on how this should proceed?

I seem to have stumbled on something more complex than just the doc bug. My feeling is that we should for now just merge this doc change to re-sync, and address the underlying "what the behavior should be" question elsewhere.

kxepal:
Sorry, I can't help much here since the only things I do with access logger is turn it off or keep it in default state.

But if docs says one, but code works in other way, it's good to fix the docs at the first place and then propose changes to the code if they are needed.

asvetlov:
Merged!

@cdunklau thanks!

PR #3317, opened 2018-10-03
Add typings for web_response

kornicameister:
Add typings for `aiohttp/web_response.py`

For now it is not clear.

## Related issue number

#1749 

## Checklist


aiohttp/helpers.py
```diff
@@ -484,13 +484,13 @@ def _format_r(request: 'BaseRequest',
     @staticmethod
     def _format_s(request: 'BaseRequest',
                   response: 'StreamResponse',
-                  time: float) -> str:
+                  time: float) -> int:
```
kornicameister: Correct if I am wrong, but doing typings in `web_response.py` reveled that it should be `int` most likely.

aiohttp/payload.py
```diff
@@ -101,7 +101,7 @@ def register(self,
 
 class Payload(ABC):
 
-    _size = None  # type: Optional[float]
+    _size = None  # type: Optional[int]
```
kornicameister, line 104: Yeah, I am pretty sure that making this `float` in the beginning was an err on my side.

aiohttp/web_response.py
```diff
@@ -268,53 +304,63 @@ def last_modified(self, value):
...
+    def _generate_content_type_header(
+        self,
+        CONTENT_TYPE: str=hdrs.CONTENT_TYPE,
+    ) -> None:
+        params = '; '.join(
+            ("%s=%s" % (k, v) for k, v in self._content_dict.items()),
+        )
         if params:
-            ctype = self._content_type + '; ' + params
+            ctype = self.content_type + '; ' + params
```
kornicameister, line 320: Not really sure why variable is `Optional[str]` and attr getter from super returns `str`. Can't tell yet which one is correct.

aiohttp/web_response.py
```diff
@@ -409,35 +455,38 @@ def _start_compression(self, request):
...
+    def __delitem__(self, key: str) -> None:
         del self._state[key]
 
-    def __len__(self):
+    def __len__(self) -> int:
         return len(self._state)
 
-    def __iter__(self):
-        return iter(self._state)
+    def __iter__(self) -> Iterator[Tuple[str, str]]:
```
kornicameister: I will try to look into this once again  but basically this one says that `aiohttp/web_response.py:483: error: Return type of "__iter__" incompatible with supertype "Iterable"
`. An `Iterable` is last or the one before last in the chain of inheritance starting from `MutableMapping`. I can't really say why this one is bad if typings for `MutableMapping` says exactly `MutableMapping[str, str]`.

aiohttp/web_response.py
```diff
@@ -506,27 +555,24 @@ def __init__(self, *,
...
-    def body(self, body,
-             CONTENT_TYPE=hdrs.CONTENT_TYPE,
-             CONTENT_LENGTH=hdrs.CONTENT_LENGTH):
+    def body(self,
+             body: Any,
+             CONTENT_TYPE: str=hdrs.CONTENT_TYPE,
+             CONTENT_LENGTH: str=hdrs.CONTENT_LENGTH) -> None:
         if body is None:
             self._body = None
-            self._body_payload = False
```
kornicameister, line 518: So, it turns out that `_body` can be either `bytes` or `Payload`. Seemed more correct to rely on `isinstance` then keeping extra property in the class.

aiohttp/web_response.py
```diff
@@ -548,13 +594,17 @@ def body(self, body,
...
 
     @property
-    def text(self):
+    def text(self) -> Optional[str]:
         if self._body is None:
             return None
-        return self._body.decode(self.charset or 'utf-8')
+        else:
+            if isinstance(self._body, payload.Payload):
+                raise ValueError('Cannot extract text from payload')
```
kornicameister, line 607: Most likely this is wrong :/

aiohttp/helpers.py
asvetlov: You are right.
As an option, we can return `str` but should apply `str(response.status)`.
Let's change annotation as you did already.

aiohttp/payload.py
asvetlov: Agree, `3.5` bytes will never happen.

aiohttp/web_response.py
```diff
@@ -35,60 +41,71 @@ class ContentCoding(enum.Enum):
...
+        self._keep_alive = None  # type: Optional[bool]
         self._chunked = False
         self._compression = False
-        self._compression_force = None
+        self._compression_force = None  # type: Optional[bool]
         self._cookies = SimpleCookie()
 
-        self._req = None
-        self._payload_writer = None
+        self._req = None  # type: Optional[Request]
```
asvetlov: `Optional[BaseRequest]` here

aiohttp/web_response.py
```diff
@@ -35,60 +41,71 @@ class ContentCoding(enum.Enum):
...
         self.set_status(status, reason)
 
     @property
-    def prepared(self):
+    def prepared(self) -> bool:
         return self._payload_writer is not None
 
     @property
-    def task(self):
+    def task(self) -> asyncio.Task[_T]:
```
asvetlov: `asyncio.Task[None]`. https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_protocol.py#L336 returns nothing.

aiohttp/web_response.py
```diff
@@ -101,22 +118,22 @@ def set_status(self, status, reason=None, _RESPONSES=RESPONSES):
...
         self._keep_alive = False
 
     @property
-    def body_length(self):
+    def body_length(self) -> int:
         return self._body_length
 
     @property
-    def output_length(self):
+    def output_length(self) -> None:
```
asvetlov: I guess the length is `int`

aiohttp/web_response.py
```diff
@@ -101,22 +118,22 @@ def set_status(self, status, reason=None, _RESPONSES=RESPONSES):
...
         warnings.warn('output_length is deprecated', DeprecationWarning)
-        return self._payload_writer.buffer_size
+        return self._payload_writer.buffer_size  # type: ignore
 
-    def enable_chunked_encoding(self, chunk_size=None):
+    def enable_chunked_encoding(self, chunk_size: Any=None) -> None:
```
asvetlov: `chunk_size: Optional[int]`. 
The parameter will be removed in aiohttp 4.0 but now let's be consistent.

kornicameister: I figure that since it is not used in anyway and setting it to anything else other than `None` will anyway result in an error, we actually don't have to care about that. But if you prefer to have it as `Optional[int]`, let it be so.

kornicameister:
@asvetlov sorry for longer inactive. Didn't exactly had time in the evenings to take a look here. 

The current status is that I have two errors for which I can't find the solution:
```
aiohttp/web_response.py:105: error: Cannot determine type of 'RESPONSES'
aiohttp/web_response.py:356: error: Cannot determine type of 'SERVER_SOFTWARE'
```

Surprisingly that happens only if I use `make mypy`. Running mypy directly does not report that. I think I saw the fix for it elsewhere I just can't find it, so any help appreciated.

asvetlov:
I've fixed mypy errors but looks like the logic is broken somehow -- 10 tests are failed.

asvetlov:
Done by 22a58e25

kornicameister:
@asvetlov I was aware that logic will be broken somehow. I couldn't actually get around mypy here and wanted to handle tests after that's done. But if you managed to resolve that, I am more than happy ;-)

kornicameister:
BTW: @asvetlov really nice product. Maybe you recall me mentioning performance comparision (or maybe I've mentioned that over in [connexion](https://github.com/zalando/connexion)), but comparing `aiohttp` with ordinary `flask` based application, well we, roughly speaking, improved performance about 30%-50%. So really great product :+1:

kornicameister:
Heh, it was real application. Although for the sake of POC I've compared just two endpoints (those destined to work under heaviest load). But application was normally working inside the cluster (i.e. it was normally deployed) and tests were ran also in there. So we've been actually sending bunch of requests, sometimes even spawning the service. And what was nearly a heart attack for older app, new one was still catching up.

asvetlov:
Sweet!

PR #3320, opened 2018-10-04
Add example for SSL client auth.

danielgtaylor:
## What do these changes do?

For my use-case I needed to set up the SSL context to do client auth. It will not work unless I pass in the right purpose when creating the context. I searched around a long time and couldn't find anything about it, so I'm adding it to the docs to help others (and my own forgetful future self).

No

## Related issue number

None

## Checklist

### DIFF
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -424,7 +424,15 @@ same thing as the previous example, but add another call to
                              '/path/to/client/private/device.key')
   r = await session.get('https://example.com', ssl=sslcontext)
 
-There is explicit errors when ssl verification fails
+If you need to use client-side SSL certificates for authentication
+you can create the :class:`ssl.SSLContext` with that purpose::
+
+  sslcontext = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+  sslcontext.load_cert_chain('/path/to/client/public/device.pem',
+                             '/path/to/client/private/device.key')
+  r = await session.get('https://example.com', ssl=sslcontext)
+
+There are explicit errors when ssl verification fails
 
 :class:`aiohttp.ClientConnectorSSLError`::
 
--- a/CHANGES/3320.doc
+++ b/CHANGES/3320.doc
@@ -1,0 +1,1 @@
+Add SSL client auth example to docs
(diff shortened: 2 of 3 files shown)


docs/client_advanced.rst
```diff
@@ -424,7 +424,15 @@ same thing as the previous example, but add another call to
                              '/path/to/client/private/device.key')
   r = await session.get('https://example.com', ssl=sslcontext)
 
-There is explicit errors when ssl verification fails
+If you need to use client-side SSL certificates for authentication
+you can create the :class:`ssl.SSLContext` with that purpose::
+
+  sslcontext = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
```
asvetlov, line 430: Should it be `ssl.Purpose.SERVER_AUTH`?
https://docs.python.org/3/library/ssl.html#ssl.Purpose.SERVER_AUTH

danielgtaylor: Here's a good article that describes my use case:

https://medium.com/@sevcsik/authentication-using-https-client-certificates-3c9d270e8326

TL;DR: The server authenticates the client during the SSL handshake by using the client's cert. This is different from the "normal" flow where the client checks the server's cert. In fact, in a development environment we have required client certs that are used to talk to servers with self-signed certs so the *only* check is that the client sends a valid cert that the server is preconfigured to accept.

asvetlov: Got you. The problem is: you are trying to describe a working and useful but not common use case.
I afraid added 10 lines produce more confusion than help.
A much more comprehensive text is needed to describe a motivation for the mode, like the mentioned article.

But I have a feeling that aiohttp doc is not a good place for such articles, the library *uses* `ssl`, describes very trivial cases but doesn't dig into SSL too deep -- the subject is out of `aiohttp` scope (like basic Python programming and many other things).

danielgtaylor:
Closing this based on feedback that this use case is too specific and uncommon to warrant going into the docs. Thanks for considering the change!

asvetlov:
Welcome and thanks for the PR!
Even not merged pull requests are valuable, I've learned a new certificates game :)

webknjaz:
I actually think some info would be useful. While it may be uncommon for generic apps, I believe it's very useful in corporate envs, for example.

asvetlov:
A FAQ section?

webknjaz:
Yep, at least there. Let's raise people's awareness :)

PR #3323, opened 2018-10-05
Clarify merge strategy

asvetlov:
Add a hint to avoid unnecessary `git rebase` when working on a PR.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -29,6 +29,15 @@ Workflow is pretty straightforward:
   7. Optionally make backport Pull Request(s) for landing a bug fix
      into released aiohttp versions.
 
+.. note::
+
+   The project uses *Squash-and-Merge* strategy for *GitHub Merge* button.
+
+   Basically it means that there is **no need to rebase*** a Pull Request against
+   *master* branch. Just ``git merge`` *master* into your working copy (a fork) if
+   needed. The Pull Request is automatically squashed into the single commit on the PR
+   accepting.
+
 Preconditions for running aiohttp test suite
 --------------------------------------------
 
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -33,10 +33,10 @@ Workflow is pretty straightforward:
 
    The project uses *Squash-and-Merge* strategy for *GitHub Merge* button.
 
-   Basically it means that there is **no need to rebase*** a Pull Request against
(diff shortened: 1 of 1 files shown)


docs/contributing.rst
```diff
@@ -29,6 +29,15 @@ Workflow is pretty straightforward:
   7. Optionally make backport Pull Request(s) for landing a bug fix
      into released aiohttp versions.
 
+.. note::
+
+   The project uses *Squash-and-Merge* strategy for *GitHub Merge* button.
+
+   Basically it means that there is **no need to rebase*** a Pull Request against
```
webknjaz: `***` -> `**`

webknjaz: Also, I'd rephrase to `you don't have to`. Because it's a person's choice. I personally like to be doing this a lot.

docs/contributing.rst
```diff
@@ -29,6 +29,15 @@ Workflow is pretty straightforward:
...
+   *master* branch. Just ``git merge`` *master* into your working copy (a fork) if
+   needed. The Pull Request is automatically squashed into the single commit on the PR
```
webknjaz: ... once the PR is accepted ...

kxepal:
Well, rebase and master merge may produce quite different results, especially for changes against similar things.  I would prefer to suggest rebase always on top of master to avoid any confusions. However, still final diff would be good to review for some stranges after any operation.

asvetlov:
1. We always squash PRs on merging.
2. Rebase on top of master is harder than `git merge master`: very often it requires resolving more conflicts.
3. After squashing all PR commits disappear, from master's point of view merge and rebase produces the same output.
4. Given all this, I think that rebasing in *our* workflow is pointless: more work for the same result.
5. The proposal targets on making life easier for newbies. Git jedies can do what they want, the jedi doesn't need such instructions at all :)

webknjaz:
Take a look at this nice workflow: https://www.gentoo.org/glep/glep-0066.html

asvetlov:
Sorry, I don't follow.
Do you propose to change the workflow? If yes -- please open a new issue.
This PR documents the current status quo, that's it.

webknjaz:
It's more like sharing a learning resource. I think we could borrow some git configuration advices and instructions from there.

asvetlov:
Ok.
I've merged the PR.
If you want to discuss Gentoo workflow -- please open a new issue.

PR #3341, opened 2018-10-12
Wrap parse_mimetype in an lru_cache

orf:
## What do these changes do?

`parse_mimetype` seems like an ideal candidate for an `lru_cache`. The return result is immutable, it's non-trivial, called fairly often and has a single, immutable input.

Nope.

## Related issue number

None.

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -25,7 +25,7 @@ from urllib.request import getproxies
 
 import async_timeout
 import attr
-from multidict import MultiDict
+from multidict import MultiDict, MultiDictProxy
 from yarl import URL
 
 from . import hdrs
@@ -234,7 +234,7 @@ class MimeType:
     type = attr.ib(type=str)
     subtype = attr.ib(type=str)
     suffix = attr.ib(type=str)
-    parameters = attr.ib(type=MultiDict)  # type: MultiDict[str]
+    parameters = attr.ib(type=MultiDictProxy)  # type: MultiDictProxy[str]
 
 
 @functools.lru_cache(maxsize=56)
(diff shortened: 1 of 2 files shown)


aiohttp/helpers.py
```diff
@@ -237,6 +237,7 @@ class MimeType:
     parameters = attr.ib(type=MultiDict)  # type: MultiDict[str]
 
 
+@functools.lru_cache(maxsize=56)
```
kxepal, line 240: Nice idea. But why 56?

orf: No particular reason. It's potentially controlled by an attacker, so it cannot be too high. Could make it 100?

kxepal: There 75 mime types in nginx default mime.types file. Quite much of them are not common / popular. I'm pretty sure that 10 will be more than enough.

asvetlov:
I'm curious what is the performance boost?

orf:
Master:
```
> %timeit parse_mimetype('text/html; charset=utf-8')
> 5.06 µs ± 47.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```
This patch:
```
> %timeit parse_mimetype('text/html; charset=utf-8')
> 95.8 ns ± 0.771 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

orf:
Ahh, one small thing: the `parameters` object needs to be immutable. I've changed it to use `MultiDictProxy`.

Not sure if this is a breaking change, but the speedup is pretty nice if we can accept this.

orf:
Ahh, one small thing: the `parameters` object needs to be immutable. I've changed it to use `MultiDictProxy`.

Not sure if this is a breaking change, but the speedup is pretty nice if we can accept this.

asvetlov:
thanks

PR #3343, opened 2018-10-15
Add typings for Tracing

vananabu:
## Related issue number
[#1749](https://github.com/aio-libs/aiohttp/issues/1749)
[](url)

## Checklist

### DIFF
--- a/aiohttp/tracing.py
+++ b/aiohttp/tracing.py
@@ -2,15 +2,15 @@ from types import SimpleNamespace
 from typing import TYPE_CHECKING, Awaitable, Callable, Type
 
 import attr
-from multidict import CIMultiDict
+from multidict import CIMultiDict  # noqa
 from yarl import URL
 
 from .client_reqrep import ClientResponse
 from .signals import Signal
 
 
-if TYPE_CHECKING:
-    from .client import ClientSession
+if TYPE_CHECKING:  # pragma: no cover
+    from .client import ClientSession  # noqa
 
     _Signal = Signal[Callable[['TraceConfig'], Awaitable[None]]]
 else:
--- a/aiohttp/tracing.py
+++ b/aiohttp/tracing.py
@@ -11,9 +11,8 @@ from .signals import Signal
 
 if TYPE_CHECKING:
     from .client import ClientSession
-    from .web_app import Application
 
-    _Signal = Signal[Callable[[Application], Awaitable[None]]]
+    _Signal = Signal[Callable[['TraceConfig'], Awaitable[None]]]
 else:
     _Signal = Signal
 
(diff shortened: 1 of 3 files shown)


aiohttp/tracing.py
```diff
@@ -24,31 +30,37 @@ class TraceConfig:
...
-        self._on_dns_cache_miss = Signal(self)
-
-        self._trace_config_ctx_factory = trace_config_ctx_factory
-
-    def trace_config_ctx(self, trace_request_ctx=None):
+    def __init__(
+        self,
+        trace_config_ctx_factory: Type[SimpleNamespace]=SimpleNamespace
+    ) -> None:
+        self._on_request_start = Signal(self)  # type: Signal[Callable[['Application'], Awaitable[None]]] # noqa
```
kxepal: Maybe define `Signal[Callable[['Application'], Awaitable[None]]]` only once at top level and reuse it everywhere? You wouldn't need typing comments and qa disable in this case.

dbohomiakov: Refactored

aiohttp/tracing.py
```diff
@@ -8,6 +9,15 @@
 from .signals import Signal
 
 
+if TYPE_CHECKING:
+    from .client import ClientSession
+    from .web_app import Application
+
+    _Signal = Signal[Callable[[Application], Awaitable[None]]]
```
asvetlov: Please replace `Application` with `TraceConfig` to fix the signature

dbohomiakov: done

aiohttp/tracing.py
```diff
@@ -24,31 +33,37 @@ class TraceConfig:
...
-        self._on_dns_cache_miss = Signal(self)
-
-        self._trace_config_ctx_factory = trace_config_ctx_factory
-
-    def trace_config_ctx(self, trace_request_ctx=None):
+    def __init__(
+        self,
+        trace_config_ctx_factory: Type[SimpleNamespace]=SimpleNamespace
+    ) -> None:
+        self._on_request_start = Signal(self)  # type: _Signal
```
kxepal, line 40: Hm. Do we have any conventions about when to use type comment and when - annotation? Why not to use annotations everywhere? IIRC, those comments were made just for Python 2.x  typing support.

dbohomiakov: I didn't find any documented convention regarding style for annotations so it looks like the question to @asvetlov.

asvetlov: 1. This comment is for Python 3.5; `a: type = val` is supported by Python 3.6 only.
2. We are in the middle of the transition to a fully annotated code. After the work is done mypy strict mode will be enabled.

The line is correct.

kxepal: I almost forgot what 3.5 is. Thanks!

asvetlov:
Thanks!

PR #3349, opened 2018-10-18
Bump aiohttp-theme

alxpy:
## What do these changes do?
Fix docs theme
Ref: https://github.com/aio-libs/aiohttp-theme/pull/2

No


asvetlov:
Please note: docs site will be not updated automatically, there is a need for a new aiohttp release.
P.S.
Doc scheme (as well as other deps) is updated by pyup bot but I see nothing wrong in pinning it manually if you've created a PR for it.

oleksandr-kuzmenko:
ok, thx

Issue #3364, opened 2018-10-26
HTTP response not released with raise_for_status=True


Grendel7:
## Long story short

When using ClientSession with `raise_for_status=True`, HTTP responses are not released. This means the related TCP connection will remain in `CLOSE_WAIT` state indefinitely, when the called server returns an error status code and a large body. Eventually, these waiting connections will hit the connection limit and cause timeout errors on ALL open connections.

## Expected behaviour

When using Client Session with `raise_for_status=True`, TCP connections are released properly and they are either closed or reused (depending on keep alive configuration).

## Actual behaviour

When using Client Session with `raise_for_status=True`, the TCP connections 

## Steps to reproduce

client.py
```
from aiohttp import ClientSession, ClientResponseError
import asyncio


async def send_request():
    client = ClientSession(raise_for_status=True)

    try:
        async with client.get('http://hansadema.com/404.php') as response:
            print('Received good status %d' % response.status)
    except ClientResponseError as e:
        print('Received bad status %d' % e.status)


loop = asyncio.get_event_loop()
asyncio.ensure_future(send_request())
loop.run_forever()
```

Dockerfile
```
FROM python:3.6-slim

COPY . /usr/src/app
RUN pip install aiohttp

CMD python /usr/src/app/client.py
```

Commands:
```
$ docker build -t aiohttp-test .
$ docker run --rm --name aiohttp-test aiohttp-test
# Linux command to check open connections
$ sudo nsenter -t (docker inspect -f '{{.State.Pid}}' "aiohttp-test") -n ss --numeric --tcp
```

## Your environment

aiohttp 3.4.4
Python 3.6.6
Debian 9.5

asvetlov:
Fixed by #3365

PR #3365, opened 2018-10-26
Release HTTP response before raising status exception

HansAdema:
## What do these changes do?

A bugfix for issue #3364 - it releases the HTTP response before the ClientResponse exception is thrown.

If there is some way to get the HTTP response from a ClientResponseException (which, AFAIK, isn't possible), the response would be closed already.

Beyond that, no behavioral changes except for that connections are closed properly.

## Related issue number

#3364

## Checklist

- [ ] Unit tests for the changes exist: _I could not reproduce this issue with aiohttp's server_.
- [X] Documentation reflects the changes: _I could not find relevant docs._

### DIFF
--- a/CHANGES/3364.bugfix
+++ b/CHANGES/3364.bugfix
@@ -1,0 +1,1 @@
+Release HTTP response before raising status exception
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -93,6 +93,7 @@ Gregory Haynes
 Gus Goulart
 Gustavo Carneiro
 Günther Jena
+Hans Adema
 Harmon Y.
 Hu Bo
 Hugo Herter
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -907,6 +907,7 @@ class ClientResponse(HeadersMixin):
     def raise_for_status(self) -> None:
         if 400 <= self.status:
             assert self.reason  # always not None for started response
+            self.release()
             raise ClientResponseError(
                 self.request_info,
                 self.history,
--- a/tests/test_client_response.py
+++ b/tests/test_client_response.py
@@ -665,6 +665,7 @@ def test_raise_for_status_4xx() -> None:
         response.raise_for_status()
     assert str(cm.value.status) == '409'
     assert str(cm.value.message) == "CONFLICT"
+    assert response.closed
 
 
 def test_resp_host() -> None:


webknjaz:
> I could not reproduce this issue with aiohttp's server.

We need regression tests. And if you cannot reproduce an issue within aiohttp, is it a real bug then? I'd like more details on this before proceeding. Maybe _togather_ we could come up with a solution...

Grendel7:
Thank you @webknjaz for checking in.

I can confirm this issue is real. However, the only times I have been able to confirm this issue with a remote host and a large response body.

Come to think, it might not actually be a limitation of aiohttp's server, but a problem with local networking. I haven't tried to run an aiohttp server on a remote system yet to see if I could reproduce the issue.

It might be that local networking is the issue. And by issue, I mean I'm not sure whether either
- the issue does not occur on a loopback interface; or
- the command I used to check it doesn't work on loopback interfaces.

**Question:** Does Python or one of the lower level components of aiohttp expose information about TCP connection statuses? If so, that would make testing the connection states a lot easier.

This was a script I used to test a big server response:

server.py
```python


async def body(request):
    client = ClientSession()
    async with client.get('https://en.wikipedia.org/wiki/Main_Page#/media/File:Lewis_Carroll_-_Henry_Holiday_-_Hunting_of_the_Snark_-_Plate_9.jpg') as response:
        body = await response.read()
    await client.close()
    return web.Response(body=str(body)+str(body)+str(body), status=404)

app = web.Application()
app.add_routes([web.get('/', body)])

web.run_app(app)
```

asvetlov:
Better fix is calling `self.close()` from `resp.raise_for_status()`.
Unit-test for `resp.closed` looks easy.
@HansAdema would you improve the PR?

Grendel7:
No problem, I've updated the PR with a check on `response.closed` in the tests.

asvetlov:
Thanks @HansAdema !

haizaar:
Hi guys,

Sorry for necroing, but after quite some time we've upgraded to recent version of aiohttp and lots of our error handling code stopped working due to the change introduced in this PR. Here example of the code we use for error handling:

```python
    async with aiohttp.ClientSession() as s:
        res = await s.get("https://google.com/not-such-thing")
        try:
            res.raise_for_status()
        except aiohttp.ClientResponseError as err:
            print("Failed. The error body:")
            print(await res.text())
```

Awaiting for `res.text()` now raises `ClientConnectionError: Connection closed`.

If I understand right, the reason for the fix was to prevent connection leaking if exception was raised, right?

Anyhow, for us it is quite a big breaking change. What would be your recommendation? Switch to manual status code parsing? Namely:
```python
    async with aiohttp.ClientSession() as s:
        res = await s.get("https://google.com/not-such-thing")
        if 400 <= res.status:
            print("Failed. The error body:")
            print(await res.text())
```

Thank you.

/CC @worroc @motinani

elupus:
Ran into this as well. In general it's annoying that the ClientResponseError doesn't include body. But now the simple workaround doesn't exist. You have to implement your own raise_for_status method.

asvetlov:
I'm sorry, guys, for the design mistake made years ago.

If you need a response body the manual status code parsing is maybe the best option available (currently `raise_for_status()` contains 12 code lines).

Providing a response body as an attribute of `ClientResponseError` is not an option because the response has potentially unlimited size; it leads to a high memory footprint (not a leak technically but the very not desirable behavior).

elupus:
I don't understand the need for the close in method thou. Either there is a context manager or you need be prepared for manual closing in some finally clause right. I suppose it would be an easy case to miss though.

Work around isn't too hard though.

asvetlov:
Generally speaking, you don't know at the exception catch point if the response was closed or not. 
It generates an additional point of confusing.

haizaar:
I'll second asvelov here. Though this change caught us off guard and we had to fix thing on the spot, eventually I think it brings clarity wrt resource management.

PR #3366, opened 2018-10-26
add facebook´s doh-proxy to built_with list

miraculli:
## What do these changes do?

it adds the doh-proxy from Facebook to the of 'built_with' aiohttp list
doh-proxy is a proof of concept for an DNS-over-HTTP proxy.


asvetlov:
Thanks for the PR.
I've broke unit tests for Python 3.7 yesterday but think this change can bypass the test suite.

PR #3372, opened 2018-10-30
Encourage creation of aiohttp objects from coroutines

### LINKED ISSUE #3331: Encourage creation of aiohttp public objects inside a coroutine
I would enforce it. For example, constructing `CleintSession` in a module global namespace is a bad smell and leads to hard-to-debug errors very often.

The plan is:
1. Add `helpers.get_running_loop()` function with the following implementation:
```
def get_running_loop(loop=None):
    if loop is None:
    if not loop.is_running():
        warnings.warn("The object should be created from async function",
                                 DeprecationWarning,
                                 stacklevel=3)
```
2. Replace `asyncio.get_event_loop()` with `get_running_loop()` in all aiohttp objects.
3. Replace a deprecation warning with `RuntimeError()` in aiohttp 4.0

asvetlov:
Fixes #3331

### DIFF
--- a/CHANGES/3331.removal
+++ b/CHANGES/3331.removal
@@ -1,0 +1,1 @@
+Encourage creation of aiohttp public objects inside a coroutine
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,8 @@
 all: test
 
 .install-deps: $(shell find requirements -type f)
-	@pip install -r requirements/cython.txt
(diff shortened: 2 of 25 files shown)


kxepal:
This wouldn't affect creation of web.Application object, right?

asvetlov:
In `web.Application` `loop` parameter is deprecated already.

kxepal:
Ah, right. Thanks!

PR #3373, opened 2018-10-30
update `BaseRunner` and `AppRunner` to honor an event loop other than the current loop

thatmattbone:
## What do these changes do?

These changes make it possible to explicitly pass an event loop into an `AppRunner`. Previously the current event loop returned by `asyncio.get_event_loop()` was always used.

By default there are no changes for the user. A user must opt-in to this new behavior.

## Related issue number

I couldn't find any. Happy to create a new issue if that is useful.

## Checklist

### DIFF
--- a/CHANGES/3373.feature
+++ b/CHANGES/3373.feature
@@ -1,0 +1,1 @@
+Enable users to an explict event loop in `AppRunner`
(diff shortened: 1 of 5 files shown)


asvetlov:
Please describe your use case. When and why do you need non-current event loop?

thatmattbone:
@asvetlov The more I think about it, it's not required at all and explicitly passing around the loop seems frowned upon by the asyncio devs.

 I was working with some code that passed it around explicitly and it just broke my brain. I will change THAT code instead!

PR #3404, opened 2018-11-22
Header cleanup

das7pad:
## What do these changes do?

Fix import of cpython headers.

Drop invalid header start `PROXY_`.

No, this is a low level optimization.

## Related issue number

The PR that introduced the header optimization: #3095

## Checklist


tools/gen.py
```diff
@@ -16,11 +15,17 @@ def factory():
 
 
 TERMINAL = object()
+HIDDEN_FROM_FIND_HEADER = (
+    'Proxy-Authorization',
```
asvetlov: Please elaborate why do you want to skip the header

asvetlov: Ah, thanks.
I don't recall details but IMHO there is no reason for `Proxy-Authorization` header skipping.

Please revert skipping logic in `gen.py` and regenerate `_find_headers.c`

das7pad: done

PR #3407, opened 2018-11-23
Update client_quickstart.rst

nyurik:
Total timeout is shown as disabled in the docs, but in reality it seems like it is set to 5 minutes. http://docs.aiohttp.org/en/stable/client_reference.html#clienttimeout

### DIFF
--- a/docs/client_quickstart.rst
+++ b/docs/client_quickstart.rst
@@ -422,10 +422,11 @@ Supported :class:`ClientTimeout` fields are:
 
       The whole operation time including connection
       establishment, request sending and response reading.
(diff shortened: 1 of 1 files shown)


asvetlov:
I don't agree.
1. 5 mins for total timeout is mentioned just 2 paragraphs before your change and repeated in the next paragraph again.
2. If you want to mention default values for `total` but not for other fields. I don't think we need it at all, a link to the reference chapter should be enough.

asvetlov:
Closing for sake of aio-libs/aiohttp@20d36cb6
Feel free to send a docs improvement PR but based on my last comments and changes.

PR #3412, opened 2018-11-29
Allow installing aiodns and cchardet as extras

WouldYouKindly:
Users will be able to install `aiodns` and `cchardet` using `pip install aiohttp[aiodns,cchardet]`

Related to https://github.com/aio-libs/aiohttp/issues/2397

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,6 +109,11 @@ tests_require = [
     'pytest-xdist',
 ]
 
+extras_require = {
+    'cchardet': ['cchardet'],
+    'aiodns': ['aiodns'],
+}
+
 
 args = dict(
     name='aiohttp',
@@ -153,6 +158,7 @@ args = dict(
     install_requires=install_requires,
     tests_require=tests_require,
     setup_requires=pytest_runner,
+    extras_require=extras_require,
     include_package_data=True,
     ext_modules=extensions,
     cmdclass=dict(build_ext=ve_build_ext),
--- a/README.rst
+++ b/README.rst
@@ -158,7 +158,9 @@ Requirements
 - yarl_
 
 Optionally you may install the cChardet_ and aiodns_ libraries (highly
-recommended for sake of speed).
+recommended for sake of speed)::
+
+    $ pip install aiohttp[cchardet,aiodns]
 
 .. _chardet: https://pypi.python.org/pypi/chardet
 .. _aiodns: https://pypi.python.org/pypi/aiodns


asvetlov:
Thanks!

setup.py
```diff
@@ -109,6 +109,11 @@ def read(f):
     'pytest-xdist',
 ]
 
+extras_require = {
+    'cchardet': ['cchardet'],
+    'aiodns': ['aiodns'],
```
webknjaz, line 114: I'd go for smth like `dns-speedup` and `charset-speedup` to make extras represent feature flags, not end-dependencies. This also allows to abstract extra names and make those deps easily replaceable.

asvetlov: Sorry, looks like I've pressed 'merge' too fast.

The chance of libraries replacement is very low.
I doubt if we'll do it at all.
But I like `speedup` requirement name.
What if we add `'speedup`: ['aiodns', 'cchardet']` line to requirements for quick installation of all available accellerators?

P.S.
I've missed 'brotli', it is another optional dependency.

asvetlov: For example, user wants to use aiodns but cchardet installation faild on his system.
Or aiodns doesn't work well for some reason, e.g. it doesn't respect all OS configuration files.

webknjaz: Well, there's `speedups` in plural I saw used in other distributions. We could use that for "everything".

As for "it's not likely to change", while it's likely true I'd like to point out that there will probably be other dependencies and I'd like to see extra names being consistent as it's a public API, so changing names would be difficult.

I don't want to end up with a bunch of extras named using different random strategies. And I think extras should name features, not implementations.

Bottom line: we might want to have separate extra deps + one "wholesale" extra.

asvetlov: I like it.
The only thing we need to think about is naming.
`cchardet` is definitely a speedup.
`aiodns` is not. I *can* boost the execution time but the main point is using nonblocking async DNS API.
`brotli` is a feature, basically a support for another comression method (implemented partially now).

What are the best names for these features? Suggestions?

asvetlov: Taking into account the conversation I've reverted the PR.
Let's make an agreement about exposed names first and only after it implement them.
The issue priority is low but I love to add changes only once without future renaming.

webknjaz: `charset-guessing-speedup`, `async-dns` and `brotli-compression`?

PR #3417, opened 2018-12-01
Deprecate bare connector close

asvetlov:
Use `async with connector:` and `await connector.close()` instead

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -82,6 +82,10 @@ def noop(*args, **kwargs):  # type: ignore
     return  # type: ignore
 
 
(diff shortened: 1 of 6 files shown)


aiohttp/helpers.py
```diff
@@ -82,6 +82,10 @@ def noop(*args, **kwargs):  # type: ignore
     return  # type: ignore
 
 
+async def noop2(*args: Any, **kwargs: Any) -> None:
```
webknjaz, line 85: Let's rename it to represent the intent:
```suggestion
async def async_noop(*args: Any, **kwargs: Any) -> None:
```

PR #3421, opened 2018-12-02
WIP: Reformat with black

asvetlov:
1. Add `make fmt` command for source code reformatting
2. Check `black` style in `make flake`


kxepal:
I don't like it as well (quotes and the way how it exands / breaks exanded things in parentheses), but having consistent code style is better than anything mixed.

I don't think we have good git history here to heavily relay on blame. For instance, because of squash + merge strategy you wouldn't have a single limited changeset with detailed explanation why, just a big changeset for the whole related PR. At the sametime GitHub provides you to see changes beyong the blamed ones with a single click. `git log --diff` is still a thing as well.

kxepal:
@asvetlov you also would need to tell flake8, isort and editorconfig that the new line length limit is 88.

samuelcolvin:
I did this on pydantic recently. I agree black isn't perfect but it's another thing not to think about.

I would recommend using the "-S" setting to avoid changing all quotes. I prefer single quotes but also it reduces the number of changes by about 80%.

asvetlov:
@samuelcolvin thanks for sharing your experience.

I prefer don't touch black config. Reformatting changes very many lines anyway regardless `-S` is used or not.

asvetlov:
https://github.com/ambv/black/pull/618 is blocker, now black formatted code is not compatible with Python 3.5

webknjaz:
> I don't think we have good git history here to heavily relay on blame

Which doesn't mean that we should make it worse.

> We have two options

Not true, there's also devs with commit access who actually know how to use Git. It's the third option, which allows newbies to keep using squash, but doesn't break advanced workflows.

> I would recommend using the "-S" setting to avoid changing all quotes. I prefer single quotes but also it reduces the number of changes by about 80%.

:+1: I also prefer single quotes style

webknjaz:
:robot: https://github.com/Mariatta/black_out

asvetlov:
I don't see a reason why a single PR should land with multiple commits.
My opinion is: If someone has unrelated changes inside a PR -- he should create multiple isolated PRs instead.

webknjaz:
Because this way I can make separate chunks of code revertible without touching other chunks. While all of them would serve one global PR purpose, for example. If you want to know more, read https://git-scm.com/book/en/v2. Git is way more powerful.

asvetlov:
> Because this way I can make separate chunks of code revertible without touching other chunks. 

Chunks isolation is not checked by CI, I doubt if they can be reverted without implications, at least in a generic case. Commits it PR *tends* to be tightly coupled. Isolation requires an extra care.

But my the main point is: the squashing model is used by CPython with very success.
I don't recall any blame or request for more complex workflow.
aiohttp is a much simpler project than CPython.
Squashing in aiohttp works fine for me, I don't want to replace a simple but working thing with something more complex.

Maybe other projects require complex workflows but it is not our case.

webknjaz:
It's not about projects, it's about making Git human-friendly

asvetlov:
Closing, better to recreate a PR that fix merge conflicts.

PR #3434, opened 2018-12-06
Add 'speedups' extra dependencies group

### LINKED ISSUE #2397: Design good distribution extras support

webknjaz:
Resolves #2397

## What do these changes do?

Add extra deps so that it'd possible to install them along with the main project as `pip install aiohttp[speedups]`

Enable users to use `pip install aiohttp[speedups]`

## Related issue number

#2397

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -151,6 +151,13 @@ args = dict(
     packages=['aiohttp'],
     python_requires='>=3.5.3',
     install_requires=install_requires,
+    extras_require={
+        'speedups': [
+            'aiodns',
+            'brotlipy',
+            'cchardet',
+        ],
+    },
     tests_require=tests_require,
     setup_requires=pytest_runner,
     include_package_data=True,


kxepal:
s/speedups/faster/ - that's nicer and could be sold with easy. Also, why there is no uvloop?

webknjaz:
I don't want `faster`, because it's a name of group of deps, not an action. It's the same as when you call a function with action/verb but you call a var/class as an object.

webknjaz:
@asvetlov so what about `uvloop`?

asvetlov:
Just library installation doesn't make sense, uvloop policy should be initialized by `asyncio.set_event_loop_policy()`.

I don't see a place for the call in aiohttp itself.

webknjaz:
So I think the idea is to check whether it's present and do it on init.

asvetlov:
What do you mean by *on init*?
I don't see any 100% reliable scenario.
A user always can do weird things, e.g. calling asyncio API **before** `aiohttp` import.
If `aiohttp` will change the default loop policy on import (which is a side effect BTW) previously opened connection will just hang etc.

PR #3437, opened 2018-12-08
Pin idna version

asvetlov:
To maximum supported by requests (a dependency for twine tool)


requirements/ci.txt
```diff
@@ -1,6 +1,7 @@
 setuptools-git==1.2
 mypy==0.641; implementation_name=="cpython"
 mypy-extensions==0.4.1; implementation_name=="cpython"
+idna==2.7
```
webknjaz: Maybe `<2.8`? Just did that for Molecule project.

asvetlov: Let's try and see how things go

PR #3442, opened 2018-12-08
Create a task per request handling.

### LINKED ISSUE #3406: ContextVar support?
## Long story short
I am trying to check ContextVar integration with aiohttp.  https://docs.python.org/3/library/contextvars.html#asyncio-support
ContextVar seems not deals well with aiohttp server... Am I doing something wrong?

## Expected behaviour

ContextVar should be different for each request.

## Actual behaviour

ContextVar has the same value between requests, but  seems to be cleared on failure.

### server.py
```python
from contextvars import ContextVar
from aiohttp import web

ctx = ContextVar("test")

async def handle(request):
    try:
        ctx.get()
        assert False, "Shouldn't ever be called. ContextVar has value"
    except LookupError:
        pass

    ctx.set("Test")

    return web.Response(text="Hello, " + ctx.get())

app = web.Application()
app.add_routes([web.get('/', handle)])

web.run_app(app, port=8000)
```

### Other process client.py
Client is working fine, just for example
```python
import aiohttp
import asyncio

async def main():
        while True:
            async with session.get("http://127.0.0.1:8000/") as response:
                html = await response.text()

            print(html)

asyncio.run(main())
```

### Response
Server is constantly being failed on each even request.

```text
Hello, Test
<html><head><title>500 Internal Server Error</title></head><body><h1>500 Internal Server Error</h1>Server got itself in trouble</body></html>
Hello, Test
<html><head><title>500 Internal Server Error</title></head><body><h1>500 Internal Server Error</h1>Server got itself in trouble</body></html>
Hello, Test
...
```

### Extra example

I tried almost the same for `sanic`, and seems that it works fine.
```python
from contextvars import ContextVar

from sanic import Sanic
from sanic.response import text

app = Sanic()

ctx = ContextVar("test")


@app.route('/')
async def test(request):
    try:
        ctx.get()
        assert False, "Shouldn't ever be called. ContextVar has value"
    except LookupError:
        pass

    ctx.set("Test")

    return text('hello, ' + ctx.get())
app.run(host='0.0.0.0', port=8000)
```

### Response
```text
hello, Test
hello, Test
hello, Test
hello, Test
hello, Test
hello, Test
hello, Test
```
## Your environment
macOs Mojave
Python 3.7.0 (default, Jun 29 2018, 20:13:13) \n[Clang 9.1.0 (clang-902.0.39.2)]
aiohttp 3.4.4, also tried 3.5.0a1 from master branch

asvetlov:
Fixes #3406

A separate task for every request doesn't produce any performance degradation on benchmarks.

Tests update is needed to wait for not only a task started by connection_made() but also a nested task.

### DIFF
--- a/tests/test_web_protocol.py
+++ b/tests/test_web_protocol.py
@@ -237,7 +237,7 @@ async def test_simple(srv, buf) -> None:
     srv.data_received(
         b'GET / HTTP/1.1\r\n\r\n')
 
-    await asyncio.sleep(0.01)
+    await asyncio.sleep(0.05)
     assert buf.startswith(b'HTTP/1.1 200 OK\r\n')
 
 
@@ -450,7 +450,7 @@ async def test_lingering_disabled(make_srv,
         b'Content-Length: 50\r\n\r\n')
     await asyncio.sleep(0)
     assert not transport.close.called
-    await asyncio.sleep(0.01)
+    await asyncio.sleep(0.05)
     transport.close.assert_called_with()
 
 
@@ -465,7 +465,7 @@ async def test_lingering_timeout(
     srv.connection_made(transport)
     request_handler.side_effect = handle_request
 
-    await asyncio.sleep(0.01)
+    await asyncio.sleep(0.05)
     assert not transport.close.called
 
     srv.data_received(
@@ -475,7 +475,7 @@ async def test_lingering_timeout(
     await asyncio.sleep(0)
     assert not transport.close.called
 
-    await asyncio.sleep(0.01)
+    await asyncio.sleep(0.05)
     transport.close.assert_called_with()
 
 
@@ -490,7 +490,7 @@ async def test_handle_payload_access_error(
         b'some data'
     )
     # start request_handler task
-    await asyncio.sleep(0.01)
+    await asyncio.sleep(0.05)
 
     with pytest.raises(web.PayloadAccessError):
         await request_handler.call_args[0][0].content.read()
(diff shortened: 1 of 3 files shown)


tests/test_web_protocol.py
```diff
@@ -237,7 +237,7 @@ def test_srv_keep_alive_disable(srv) -> None:
     srv.data_received(
         b'GET / HTTP/1.1\r\n\r\n')
 
-    await asyncio.sleep(0)
+    await asyncio.sleep(0.01)
```
kxepal: Does it worth to have some helper function which does `await asyncio.sleep` with proper time? To not change a lot of constants everywhere all the time?

asvetlov: I think all 5-6 places in the single test file are fixed.
Don't want to add a helper, at least now.

asvetlov:
@samuelcolvin what docs part do you suggest?
A chapter in `web_quickstart.rst` maybe?

samuelcolvin:
I guess.

Not sure if really matters, just if someone searches "aiohttp ContextVar" they should get a result in docs.

asvetlov:
I've started to write a doc but found that there are questions about contextvars initialization:
* Where is safe to set up a context?
* How it works with the context modification code in `Application.on_startup()`?
* The PR creates a context copy for every request processing, sure. But what is the original context source?

Let me merge the PR as is and solve the questions in separate one (along with documentation update). Maybe `run_app()` code should be modified as well, I need to check.

PR #3446, opened 2018-12-11
Contextvars support

asvetlov:
@samuelcolvin maybe you have a suggestion for documentation polishing?


docs/web_advanced.rst
```diff
@@ -393,6 +393,82 @@ and safe to use as the key.
...
+
+  E.g. all context modifications made on application startup a visible on teardown.
+
+* On every request handling *aiohttp* creates a context copy. :term:`web-handler` has
+  all variables installed on initialization stage. But the context modification made by
+  a handler or middleware is invisible to another HTTP request handling call.
+
+An example of context vars usage::
+
+    from contextvars import ContextVar
```
webknjaz, line 426: Please mark it as a python block

asvetlov: Why?
Python is a default language for the project, the doc uses this convention everywhere

docs/web_advanced.rst
```diff
@@ -393,6 +393,82 @@ and safe to use as the key.
...
+
+Starting from Python 3.7 asyncio has :mod:`Context Variables <contextvars>` as a
+context-local storage (a generalization of thread-local concept that works with asyncio
+tasks also).
+
+
+*aiohttp* server supports it in the following way:
+
+* A server inherits a current task's context used by the server
+  creation. :func:`aiohttp.web.run_app()` runs a task for handling all underlying dirty
```
samuelcolvin: > runs a task for handling all underlying dirty job

 Doesn't make sense to me. I'm not entirely sure what you're trying to say, but perhaps something like

```
A server inherits the current task's context used when creating it. 
:func:`aiohttp.web.run_app()` runs a task for handling all underlying jobs running 
the app, but alternatively :ref:`aiohttp-web-app-runners` can be used.
```

PR #3460, opened 2018-12-22
Serve brotli static files

2-5:
## What do these changes do?

Enable serving of statically compressed Brotli files (.br) if they are present, just like for .gz

If the user has statically compressed .br files present when trying to server a file using FileResponse, they will now be served instead of ignored. The .br files also take priority over .gz files.

## Checklist


asvetlov:
Tests for a new functionality are required.

2-5:
I tried to write tests, but failed to.

For behavioral tests, I looked into `test_web_sendfile.py` but failed to understand how they work. Those tests somehow test if the `.png` or `.png.gz` file is returned, depending on the `Accept-Encoding` headers, yet the `.gz` extension doesn't appear nowhere.

I also tried adding functional tests in `test_web_sendfile_functional.py` similar to the `gzip` ones, but failed again. When I try serving a `.br` file, and request it back with the proper headers (since the default one is `gzip, deflate`):

```python
resp = await client.get('/', headers={'Accept-Encoding':'gzip, deflate, br'})
```
it fails with this error: 
```python
aiohttp.client_exceptions.ClientPayloadError: 400, message='Can not decode content-encoding: br'
```
even if I have the `brotli` package installed.

I also failed writing a test for `app.router.add_static` like the ones at the end of the file.

The manual tests that I did do: use Chrome as a client, `aiohttp.web` with `add_static` as server, and confirm that when a `.br` file is present it's served instead of the `.gz` or the regular one (confirmed by looking at `Content-Type`, `Content-Length` and `Content-Encoding`)

CONTRIBUTORS.txt
```diff
@@ -1,5 +1,6 @@
 - Contributors -
 ----------------
+2-5
```
webknjaz: I think we prefer actual name.

aiohttp/web_fileresponse.py
```diff
@@ -222,12 +235,16 @@ def __init__(self, path: Union[str, pathlib.Path],
             return await super().prepare(request)
 
         if hdrs.CONTENT_TYPE not in self.headers:
-            ct, encoding = mimetypes.guess_type(str(filepath))
+            # Guess Content-Type and encoding of original path,
```
webknjaz, line 238: I think you can register new mime types there.

aiohttp/web_fileresponse.py
```diff
@@ -198,13 +198,26 @@ def __init__(self, path: Union[str, pathlib.Path],
     ) -> Optional[AbstractStreamWriter]:
         filepath = self._path
 
+        encoding = None
+        accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, '')
+
+        brotli = False
+        if 'br' in accept_encoding:
```
webknjaz, line 205: You also need some check for whether brotli dist is installed in the env at all.

CONTRIBUTORS.txt
2-5: Removed, I don't actually want attribution

aiohttp/web_fileresponse.py
2-5: Not really. We are just serving the raw compressed bytes as they are found on disk. The client requesting them (with indication of support for brotli in the headers) does the decoding. This client might not even be a Python script, like Chrome.

aiohttp/web_fileresponse.py
2-5: I thought about that, but there is no consensus about the filename/encoding name (see https://bugs.python.org/issue32021). Also, registering mimetypes in request handlers which might be seldom/never called seemed wrong, with a good posibility of creating hard to track down bugs for the user of the library.

asvetlov:
Closing the PR.
1. It fails on CI
2. I never saw `file.br` archive, while `file.gzip` is pretty common. There is no reason for supporting the former.
3. The PR author didn't provide tests for added functionality.

2-5:
> 2. I never saw `file.br` archive, while `file.gzip` is pretty common. There is no reason for supporting the former.

That is because Brotli files are either statically created by an asset pipeline (webpack/...) or dynamically compressed when serving (nginx/...). You will probably never see a freee one in the wild.

asvetlov:
The lack of unittests is still the main objection.
If you want to finish the PR -- you are welcome

PR #3462, opened 2018-12-24
Web exceptions refactoring

asvetlov:
Don't inherit `web.HTTPException` from `Response`


aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
-                          headers=headers, reason=reason,
-                          body=body, text=text, content_type=content_type)
-        Exception.__init__(self, self.reason)
-        if self.body is None and not self.empty_body:
-            self.text = "{}: {}".format(self.status, self.reason)
+        if reason is None:
+            try:
+                reason = RESPONSES[self.status_code][0]
+            except Exception:
+                reason = ''
```
kxepal: It feels like not a good reason to swallow the exception. However, I don't feel the intentions for this. Are they good?

aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
+
+        if text is None:
+            if not self.empty_body:
+                text = "{}: {}".format(self.status_code, reason)
+        else:
+            if self.empty_body:
+                warnings.warn(
```
kxepal, line 100: For which cases this warning is corrent?

aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
-        Response.__init__(self, status=self.status_code,
-                          headers=headers, reason=reason,
-                          body=body, text=text, content_type=content_type)
-        Exception.__init__(self, self.reason)
-        if self.body is None and not self.empty_body:
-            self.text = "{}: {}".format(self.status, self.reason)
+        if reason is None:
+            try:
+                reason = RESPONSES[self.status_code][0]
+            except Exception:
```
webknjaz: KeyError + IndexError?

aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
+                warnings.warn("content_type without text is deprecated",
+                              DeprecationWarning,
+                              stacklevel=2)
+            real_headers[hdrs.CONTENT_TYPE] = content_type
+        elif hdrs.CONTENT_TYPE not in real_headers and text:
+            real_headers[hdrs.CONTENT_TYPE] = 'text/plain'
+
+        super().__init__(reason)
+
+        self._reason = reason
```
kxepal: This assignment looks strange. Shouldn't it be handled in super-classes?

aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
-                DeprecationWarning)
-        Response.__init__(self, status=self.status_code,
-                          headers=headers, reason=reason,
-                          body=body, text=text, content_type=content_type)
-        Exception.__init__(self, self.reason)
-        if self.body is None and not self.empty_body:
-            self.text = "{}: {}".format(self.status, self.reason)
+        if reason is None:
+            try:
+                reason = RESPONSES[self.status_code][0]
```
webknjaz: Maybe add method for this with lru cache?

aiohttp/web_exceptions.py
```diff
@@ -83,23 +88,71 @@ class HTTPException(Response, Exception):
...
+            except Exception:
+                reason = ''
+
+        if text is None:
+            if not self.empty_body:
+                text = "{}: {}".format(self.status_code, reason)
+        else:
+            if self.empty_body:
+                warnings.warn(
+                    ("text argument is deprecated for HTTP status {} ,"
```
webknjaz: No need to have braces around text here

kxepal: Don't we have yet black to handle such styling issues?

aiohttp/web_exceptions.py
```diff
@@ -241,15 +297,22 @@ def __init__(self,
...
                  text: Optional[str]=None,
                  content_type: Optional[str]=None) -> None:
         allow = ','.join(sorted(allowed_methods))
         super().__init__(headers=headers, reason=reason,
-                         body=body, text=text, content_type=content_type)
+                         text=text, content_type=content_type)
         self.headers['Allow'] = allow
-        self.allowed_methods = set(allowed_methods)  # type: Set[str]
-        self.method = method.upper()
+        self._allowed = set(allowed_methods)  # type: Set[str]
```
webknjaz, line 297: Should `allowed_methods` be uppercased as well?

aiohttp/web_exceptions.py
webknjaz: Dunno, but it's ugly like this now:(

aiohttp/web_exceptions.py
webknjaz: Isn't it faster in case of exception handling branch invocation?

aiohttp/web_exceptions.py
asvetlov: Black reformatting fails on python 3.5.
Will apply it when ready

aiohttp/web_exceptions.py
asvetlov: Good question.
HTTP verbs are case sensitive by RFC.
Maybe even `method.upper()` is redundant.
aiohttp dispatcher never raises the lowercased method.

aiohttp/web_exceptions.py
asvetlov: Are you talking about `lru_cache` decorator from `functools`? This decorator is not for free.
Anyway, I'll switch to direct `HTTPStatus` usage.

aiohttp/web_exceptions.py
```diff
@@ -410,3 +468,19 @@ class HTTPNotExtended(HTTPServerError):
 
 class HTTPNetworkAuthenticationRequired(HTTPServerError):
     status_code = 511
+
+
+def _initialize_default_reason() -> None:
+    for obj in globals().values():
```
webknjaz, line 473: How about using
```suggestion
    for obj in HTTPException.__subclasses__():
```
which gets immediate children of the class.

asvetlov: Interesting, but we need a recursive iteration to collect all classes

webknjaz: Like here? https://stackoverflow.com/a/3862957/595220
Such helper looks quite clean. If we were on Python 3.6 I'd suggest `__init_subclass__()` for such tracking instead...

aiohttp/web_exceptions.py
```diff
@@ -410,3 +468,19 @@ class HTTPNotExtended(HTTPServerError):
...
+        if isinstance(obj, type) and issubclass(obj, HTTPException):
```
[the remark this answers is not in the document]
asvetlov: The first argument of `issubclass` should be a type:
```
>>> issubclass(1, int)
TypeError: issubclass() arg 1 must be a class
```
`globals()` have many values, not all of them are classes

asvetlov:
@webknjaz thanks for the review.
Now I'm working on the documentation update.
Will address notes a little later.

aiohttp/web_exceptions.py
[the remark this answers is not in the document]
webknjaz: Oh... How about `__subclasses__()` recursive traversal?

aiohttp/web_exceptions.py
asvetlov: Let's keep the code as is

asvetlov:
Everything is done from my perspective

PR #3467, opened 2018-12-26
Fix aiohtto.client.ClientSession type annotations

JZQT:
See change.


webknjaz:
mypy didn't catch this?

asvetlov:
That's strange, agree.
We have `strict_optional` flag in config: https://github.com/aio-libs/aiohttp/blob/master/setup.cfg#L34

PR #3470, opened 2018-12-28
Docs: re-style command line options

shvechikov:
Command line options look weird now on the official docs site:

image

Should look better after the fix:

image

https://docs.aiohttp.org/en/stable/deployment.html


asvetlov:
Agree, monospace font looks better.

asvetlov:
Thanks!

PR #3472, opened 2018-12-30
Update towncrier template

asvetlov:
Get rid of custom `tools/fix_changelog.py` postprocess procedure


pyproject.toml
```diff
@@ -1,5 +1,11 @@
+[build-system]
+requires = ["setuptools", "wheel"]
```
webknjaz: @asvetlov you were against it earlier, suggesting that sdist fallback should work, have you changed your mind?

asvetlov: I've copied the lines from `attrs` project honestly.
`sdist` fallback should still work.
We do `from setuptools import setup` already in `setup.py`.
[PEP  518](https://www.python.org/dev/peps/pep-0518/) recommends this lines.
Did I miss something?

webknjaz: Back when I proposed it you told me that fallback to pure python version won't work if the user does not have gcc. OTOH it was probably about cython presence there.

webknjaz: Anyway, I'd set s minimum setuptools version at least anyway.

asvetlov: I don't remember that discussion, sorry. 
Maybe it was about `setup.py`, not `pyproject.toml`?

What setuptools version should be used?

asvetlov: I've decided to drop the section: I want to backport the PR into 3.5.
Even a tiny chance of breaking build system doesn't worth it.

webknjaz: It wouldn't break the build. It's just about pip making a virtualenv and installing build deps there to use them in the next step. Without cython it wouldn't affect the build.

asvetlov:
I recall the idea to add `cython` to build-system requirements list but the change is completely unrelated.
The idea is adding explicit lines for parameters already used implicitly.
But anyway it is the subject for another patch (if any).

PR #3475, opened 2019-01-02
Preserve MultipartWriter parts headers on write

### LINKED ISSUE #3035: part.set_content_disposition is ignored
## Long story short

Using part.set_content_disposition for some multipart file stuff, couldn't get it to work.
So I checked the part header, which is as I expected and then made a fake writer to check what the payload being sent actually was. The content disposition in the payload doesn't at all match what I have set in the part.

## Expected behaviour

The content disposition in the payload is as set in the part.

## Actual behaviour

It appears to generate a fresh content disposition.

## Steps to reproduce

Demo
```py
class Writer:

    def __init__(self):
        self.buffer = ""

    async def write(self, stuff):
        self.buffer += stuff.decode('utf-8','ignore')

    def done(self):
        print(self.buffer[:300])

# Generate multipart
with aiohttp.MultipartWriter('form-data') as mp:
    part = mp.append(open(path, 'rb'))
    part.set_content_disposition(
        'form-data',
        quote_fields=False,
        name='files[]',
        filename=os.path.split(path)[1]
    )
    # Fix aiohttp being retarded
    part._headers['Content-Disposition'] = part._headers['Content-Disposition'].split(" filename*=utf-8", 1)[0]
    print("\n", part._headers['Content-Disposition'], "\n")
    w = Writer()
    await mp.write(w)
    w.done()
```

## Your environment

Aiohttp 3.2.1
Python 3.6
macOS 10.13.4

# The issue (as far as I can tell)

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/multipart.py#L754-L755
It appears there that the payload headers are saved separately and so no matter what you do to the returned part after it's appended, the headers will never actually be updated in the multipart.

# Hotfix (until this issue is resolved)

Create your payload yourself, set headers etc. before you append to the multipart.
eg.
```py
# Hotfix
from multidict import CIMultiDict
payload = aiohttp.payload.get_payload(open(path, 'rb'), headers=CIMultiDict())
payload.set_content_disposition(
    'form-data',
    quote_fields=False,
    name='files[]',
    filename=os.path.split(path)[1]
)
payload._headers['Content-Disposition'] = payload._headers['Content-Disposition'].split(" filename*=utf-8", 1)[0]

# Add part to multipart
part = mp.append(payload)
```

kxepal:
This fixes #3035

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -167,7 +167,8 @@ class Payload(ABC):
     @property
     def _binary_headers(self) -> bytes:
         if self.headers is None:
-            return b''
+            # FIXME: This case actually is unreachable.
+            return b''  # pragma: no cover
         return ''.join(
             [k + ': ' + v + '\r\n' for k, v in self.headers.items()]
         ).encode('utf-8') + b'\r\n'
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -165,6 +165,14 @@ class Payload(ABC):
         return self._headers
 
     @property
+    def _binary_headers(self) -> bytes:
+        if self.headers is None:
+            return b''
+        return ''.join(
+            [k + ': ' + v + '\r\n' for k, v in self.headers.items()]
+        ).encode('utf-8') + b'\r\n'
+
+    @property
     def encoding(self) -> Optional[str]:
         """Payload encoding"""
         return self._encoding
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -1008,9 +1008,6 @@ class TestMultipartWriter:
         part = writer._parts[0][0]
         assert part.headers[CONTENT_TYPE] == 'test/passed'
 
-    async def test_write(self, writer, stream) -> None:
-        await writer.write(stream)
-
     def test_with(self) -> None:
         with aiohttp.MultipartWriter(boundary=':') as writer:
             writer.append('foo')
@@ -1032,6 +1029,25 @@ class TestMultipartWriter:
         with pytest.raises(TypeError):
             with aiohttp.MultipartWriter(boundary=':') as writer:
                 writer.append(None)
+
+    async def test_write_preserves_content_disposition(
+        self, buf, stream
+    ) -> None:
+        with aiohttp.MultipartWriter(boundary=':') as writer:
+            part = writer.append(b'foo', headers={CONTENT_TYPE: 'test/passed'})
+            part.set_content_disposition('form-data', filename='bug')
+        await writer.write(stream)
+
+        headers, message = bytes(buf).split(b'\r\n\r\n', 1)
+
+        assert headers == (
+            b'--:\r\n'
+            b'Content-Type: test/passed\r\n'
+            b'Content-Length: 3\r\n'
+            b'Content-Disposition:'
+            b' form-data; filename="bug"; filename*=utf-8\'\'bug'
+        )
+        assert message == b'foo\r\n--:--\r\n'
 
 
 async def test_async_for_reader() -> None:
--- a/CHANGES/3035.bugfix
+++ b/CHANGES/3035.bugfix
@@ -1,0 +1,1 @@
+Preserve MultipartWriter parts headers on write.
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -677,7 +677,7 @@ class MultipartReader:
             self._last_part = None
 
 
-_Part = Tuple[Payload, 'MultiMapping[str]', str, str]
+_Part = Tuple[Payload, str, str]
 
 
 class MultipartWriter(Payload):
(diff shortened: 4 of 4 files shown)


aiohttp/payload.py
```diff
@@ -146,6 +146,14 @@ def headers(self) -> Optional[_CIMultiDict]:
         """Custom item headers"""
         return self._headers
 
+    @property
+    def _binary_headers(self) -> Optional[bytes]:
```
kxepal: I'm really not sure about this property, that's why it's private, but that serialization happens in two different places. Any better ideas how to handle it?

tests/test_multipart.py
```diff
@@ -1005,9 +1005,6 @@ def test_append_multipart(self, writer) -> None:
         part = writer._parts[0][0]
         assert part.headers[CONTENT_TYPE] == 'test/passed'
 
-    async def test_write(self, writer, stream) -> None:
-        await writer.write(stream)
```
kxepal, line 1012: Removed that test because it tests nothing.

aiohttp/payload.py
asvetlov: No idea, I don't recall this part of code

asvetlov:
@kxepal please fix `make mypy` run

kxepal:
No worries. Actually, that failure was quite right one: I didn't to fix the `_Part` annonation which continued thinks that detached headers are passed around.

I'm actually wonder why coverage failed so dramatically. Should I worry about and fix that? Probably, it's because removed useless test. May be I should return it with few meaninful assertions...

kxepal:
Ah, too outdated local master branch probably. Rebased now.

asvetlov:
Coverage report shows a couple missed lines: https://codecov.io/gh/aio-libs/aiohttp/pull/3475/diff
If you'll fix it -- would be nice!

kxepal:
@asvetlov 
Well, it's possible to fix coverage here with some, but that fix will have no links with the real life. Proper one will require quite of refactoring which will touch too many bits and unrelated with bugfix itself. For now I added this commit: https://github.com/aio-libs/aiohttp/pull/3475/commits/7c3623c551cab21aecb0425d18961648c5964c6a to make CI happy and will handle this issue in another PR.  Are you ok with it?

asvetlov:
@kxepal sounds perfect!

elsampsa:
Just tried the latest main branch.  Unfortunately, the bug still persists.

### Expected behaviour
I want to modify the Content-Disposition field in the headers to this:
```
Content-Disposition: attachments; filename="5c2ca3d448cf7d16922acfdc.jpg"
```
### Actual behaviour
Whatever I try, it always reads:
```
Content-Disposition: attachment; filename="test.jpg"; filename*=utf-8''test.jpg
```
### Example code
Please try this simple code snippet:
```
# https://docs.aiohttp.org/en/stable/multipart.html

# nix
import sys
import io

# async
import asyncio
import aiohttp

class debugWriter:
    
    def __init__(self):
        self.buffer=""
        
    async def write(self, stuff):
        self.buffer += stuff.decode("utf-8","ignore")
        
    def done(self):
        print(self.buffer[:200])
 

async def bug1():
    """This doesn't work because:
    
    - payload.py : Payload.set_content_disposition : gets called and it overwrites header with some default values
    
    """
    
    mpwriter = aiohttp.MultipartWriter('form-data')
    f=open("test.jpg","wb"); f.write("123".encode("utf-8")); f.close() # create a dummy file
    
    f=open("test.jpg","rb")
    subwriter = mpwriter.append(f, {
        aiohttp.hdrs.CONTENT_DISPOSITION : 'attachments; filename="5c2ca3d448cf7d16922acfdc.jpg"', # has no effect
        aiohttp.hdrs.CONTENT_TYPE : "image/whatever"
            })

    w = debugWriter()
    await mpwriter.write(w)
    w.done()
    f.close()
    

async def bug2():
    """This doesn't work because:
    
    - multipart.py : MultiparWriter.append : headers get constructed once and only when append is called.
    - .. calling set_content_disposition after that has no effect
    
    """
    mpwriter = aiohttp.MultipartWriter('form-data')
    f=open("test.jpg","wb"); f.write("123".encode("utf-8")); f.close() # create a dummy file
    
    f=open("test.jpg","rb")
    subwriter = mpwriter.append(f, {
        aiohttp.hdrs.CONTENT_TYPE : "image/whatever"
            })
    
    mpwriter.set_content_disposition('attachments', filename='5c2ca3d448cf7d16922acfdc.jpg') # has no effect
    
    w = debugWriter()
    await mpwriter.write(w)
    w.done()
    f.close()
    
    
if __name__ == '__main__':
    asyncio.get_event_loop().run_until_complete(bug1())
    asyncio.get_event_loop().run_until_complete(bug2())
```

### Quickfix
Modify **Payload.set_content_disposition**
```
def set_content_disposition(self,
                                disptype: str,
                                quote_fields: bool=True,
                                **params: Any) -> None:
        """Sets ``Content-Disposition`` header."""
        if self._headers is None:
            self._headers = CIMultiDict()

        if (hdrs.CONTENT_DISPOSITION in self._headers):
            print("set_content_disposition : CONTENT_DISPOSITION found : ", self._headers[hdrs.CONTENT_DISPOSITION])

        else:
            print("set_content_disposition : overwriting ")
            self._headers[hdrs.CONTENT_DISPOSITION] = content_disposition_header(
                disptype, quote_fields=quote_fields, **params)
        print("set_content_disposition : now ", self._headers[hdrs.CONTENT_DISPOSITION])
```
.. this way the case "bug1" in the code snippet above starts working

asvetlov:
@kxepal ?

kxepal:
@elsampsa 
Thanks! That's actually another story, but with the similar consequences. Let me see how to fix this once and for all time.

elsampsa:
Thanks!  I'll try it tomorrow then ..

elsampsa:
Tested this (version 3.5.4 from pypi).  Works as expected.

Thanks for the great library and keep up the good work!

asvetlov:
Welcome!

Issue #3313, opened 2018-10-03
Work with files asynchronously


asvetlov:
There is a very old tech debt in aiohttp: the library uses blocking synchronous calls to work with files.

It should be fixed by replacement such direct calls by `loop.run_in_executor()`.

Affected parts:
- [x] `web_fileresponse.py`
- [x] `payload.py`

Splitting the task into two pull requests makes sense in term of easier review.

The issue is pretty easy and straightforward, it waits for a champion.

socketpair:
Do not forget to `open()` files also in a thread.

I think, internal API should be the following:

```python
import aiofiles

file = await aiofiles.open()
result = await file.read()
result = await file.write()
await file.chmod()
await file.fsync()
...
```

UPD: https://github.com/Tinche/aiofiles  buth this solution is overingeneered. I would write my own with implementing each method explicitly.

asvetlov:
I don't want to use aiofiles.
`run_in_executor` is just enough for the library needs.
aiohttp calls `open()` in `file_response.py` only, client API accepts already opened regular IO objects.
Wrapping them into aiofiles looks more complex than explicit thread pool usage.

chirgjn:
👋 would love to pick this up, if not fixed already

asvetlov:
Please go ahead!

kxepal:
This could raise a holywar, but I'll risk to ask you: why not aiofiles?

asvetlov:
Did you read messages above?
I can send a ball back: why aiofiles despite the fact that we need only run in executor a couple IO calls?

kxepal:
Yes, I did, but some abstraction may be useful to not care about that bit if it will eventually have changed. If that's the only reason than it satisfied my curiousity (:

asvetlov:
Changes will not touch the public API.
Choosing an abstraction layer for async files support is the subject for future discussion, let's do it in another issue if you want to.

asvetlov:
Done

PR #3476, opened 2019-01-02
FileResponse works with files asynchronously

Tolmachofof:
## What do these changes do?

The web FileResponse works with files asynchronously
No

## Related issue number
#3313

## Checklist

### DIFF
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -179,22 +179,24 @@ class FileResponse(StreamResponse):
         # os.sendfile() system call. This should be used on systems
         # that don't support the os.sendfile().
 
-        # To avoid blocking the event loop & to keep memory usage low,
-        # fobj is transferred in chunks controlled by the
-        # constructor's chunk_size argument.
+        # To keep memory usage low,fobj is transferred in chunks
+        # controlled by the constructor's chunk_size argument.
 
         writer = await super().prepare(request)
         assert writer is not None
 
         chunk_size = self._chunk_size
+        loop = asyncio.get_event_loop()
 
-        chunk = fobj.read(chunk_size)
+        chunk = await loop.run_in_executor(None, fobj.read, chunk_size)
         while chunk:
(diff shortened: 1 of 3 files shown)


asvetlov:
thanks!

aiohttp/web_fileresponse.py
```diff
@@ -334,8 +337,8 @@ def __init__(self, path: Union[str, pathlib.Path],
             self.headers[hdrs.CONTENT_RANGE] = 'bytes {0}-{1}/{2}'.format(
                 real_start, real_start + count - 1, file_size)
 
-        with filepath.open('rb') as fobj:
+        with (await loop.run_in_executor(None, filepath.open, 'rb')) as fobj:
```
aamalev, line 340: @Tolmachofof blocking close of the fobj

Tolmachofof: @aamalev thanks for your comment!
I think that something like ```_FileContextManager``` would solve this problem. 
Simplified example:

```python
class _FileContextManager:

    def __init__(self, filepath, flags, start):
        self._filepath = filepath
        self._flags = flags
        self._start = start
        self._fobj = None

    async def __aenter__(self):
        self._fobj = await loop.run_in_executor(
            None, self._open, self._filepath, self._start
        )
        return self._fobj

    def _open(self, filepath, start):
        fobj = filepath.open(self._flags)
        if start:
            fobj.seek(start)
        return fobj

    async def __aexit__(self, exc_type, exc_val, exc_tb):
        await loop.run_in_executor(None, self._fobj.close)
        self._fobj = None
```
If this acceptable i can make a PR

aamalev: I think the conversion `with` to `try..finally` is more suitable. `_FileContextManager` need members loop and executor, and this is too much for a one-time code

Tolmachofof: You are right. ```try..finally``` statement sounds good. I will make a PR.

PR #3479, opened 2019-01-03
Refactor the way how Payload headers are handled

kxepal:
This change actually solves three separated, but heavy coupled
issues:

1. `Payload.content_type` may conflict with `Payload.headers[CONTENT_TYPE]`.

While in the end priority goes to the former one, it seems quite strange that
Payload object may have dual state about what content type it contains.

2. IOPayload respects Content-Disposition which comes with headers.

This is part of #3035 issue.

3. ClientRequest.skip_autoheaders now filters Payload.headers as well.

This issue was eventually found due to refactoring: Payload object
may setup some autoheaders, but those will bypass skip logic.

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -7,15 +7,13 @@ import os
 import warnings
 from abc import ABC, abstractmethod
 from itertools import chain
-from typing import (  # noqa
+from typing import (
     IO,
     TYPE_CHECKING,
     Any,
     ByteString,
-    Callable,
     Dict,
     Iterable,
-    List,
     Optional,
     Text,
     TextIO,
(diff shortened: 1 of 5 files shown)


asvetlov:
Please fix flake8

aiohttp/payload.py
```diff
@@ -47,6 +45,10 @@
 TOO_LARGE_BYTES_BODY = 2 ** 20  # 1 MB
 
 
+if TYPE_CHECKING:
```
asvetlov:
```suggestion
if TYPE_CHECKING: # pragma: no cover
```

kxepal:
@asvetlov Ok, it's ready for review.

kxepal:
Rebased with conflict fix. Also noticed that Travis CI and AppVeyor has different opinion about `.py` files MIME type. Probably, there could be other differences as well.

PR #3483, opened 2019-01-04
plain text http responses to errors

samuelcolvin:
## What do these changes do?

When using curl or httpie, error responses should not be html.

Yes, errors should be plain text not html when `text/html` is not in the "browsers" `Accept` header.

Currently this is failing due to an existing error which I inadvertently uncovered. I'll comment on the code below. I'll add `CHANGES` etc. once this is fixed.

## Checklist

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -5,6 +5,7 @@ import warnings
 from collections import deque
 from contextlib import suppress
 from html import escape as html_escape
+from http import HTTPStatus
 from logging import Logger
 from typing import (
     TYPE_CHECKING,
@@ -524,25 +525,27 @@ class RequestHandler(BaseProtocol):
         self.log_exception("Error handling request", exc_info=exc)
 
         ct = 'text/plain'
-        if status == 500:
-            msg = 'Server got itself in trouble'
+        if status == HTTPStatus.INTERNAL_SERVER_ERROR:
+            title = '500 Internal Server Error'
+            msg = HTTPStatus.INTERNAL_SERVER_ERROR.description
+            tb = None
+            if self.debug:
+                with suppress(Exception):
+                    tb = traceback.format_exc()
+
             if 'text/html' in request.headers.get('Accept', ''):
-                if self.debug:
-                    with suppress(Exception):
-                        tb = html_escape(traceback.format_exc())
-                        msg = '<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
+                if tb:
+                    tb = html_escape(tb)
+                    msg = '<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
                 message = (
                     "<html><head>"
-                    "<title>500 Internal Server Error</title>"
-                    "</head><body>\n<h1>500 Internal Server Error</h1>"
+                    "<title>{title}</title>"
+                    "</head><body>\n<h1>{title}</h1>"
                     "<br>\n{msg}\n</body></html>\n"
-                ).format(msg=msg)
+                ).format(title=title, msg=msg)
                 ct = 'text/html'
             else:
-                if self.debug:
-                    with suppress(Exception):
-                        msg = traceback.format_exc()
-                message = '500 Internal Server Error\n\n' + msg
+                message = title + '\n\n' + (tb or msg)
 
         resp = Response(status=status, text=message, content_type=ct)
         resp.force_close()
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -526,9 +526,8 @@ class RequestHandler(BaseProtocol):
 
         ct = 'text/plain'
         if status == HTTPStatus.INTERNAL_SERVER_ERROR:
-            title = '{http_status_code:d} {http_response_reason}'.format(
-                http_status_code=str(HTTPStatus.INTERNAL_SERVER_ERROR.value),
-                http_response_reason=HTTPStatus.INTERNAL_SERVER_ERROR.phrase,
+            title = '{0.value} {0.phrase}'.format(
+                HTTPStatus.INTERNAL_SERVER_ERROR
             )
             msg = HTTPStatus.INTERNAL_SERVER_ERROR.description
             tb = None
@@ -539,7 +538,7 @@ class RequestHandler(BaseProtocol):
             if 'text/html' in request.headers.get('Accept', ''):
                 if tb:
                     tb = html_escape(tb)
-                    msg += '\n<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
+                    msg = '<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
                 message = (
                     "<html><head>"
                     "<title>{title}</title>"
@@ -549,7 +548,7 @@ class RequestHandler(BaseProtocol):
                 ct = 'text/html'
             else:
                 if tb:
-                    msg += '\n' + tb
+                    msg = tb
                 message = title + '\n\n' + msg
 
         resp = Response(status=status, text=message, content_type=ct)
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -536,16 +536,18 @@ class RequestHandler(BaseProtocol):
             if 'text/html' in request.headers.get('Accept', ''):
                 if tb:
                     tb = html_escape(tb)
-                    msg = '<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
+                    msg += '\n<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
                 message = (
                     "<html><head>"
                     "<title>{title}</title>"
                     "</head><body>\n<h1>{title}</h1>"
-                    "<br>\n{msg}\n</body></html>\n"
+                    "\n{msg}\n</body></html>\n"
                 ).format(title=title, msg=msg)
                 ct = 'text/html'
             else:
-                message = title + '\n\n' + (tb or msg)
+                if tb:
+                    msg += '\n' + tb
+                message = title + '\n\n' + msg
 
         resp = Response(status=status, text=message, content_type=ct)
         resp.force_close()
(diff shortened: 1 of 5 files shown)


tests/test_web_server.py
```diff
@@ -26,13 +26,15 @@
         raise exc
 
     logger = mock.Mock()
-    server = await aiohttp_raw_server(handler, logger=logger)
+    server = await aiohttp_raw_server(handler, logger=logger, debug=False)
```
samuelcolvin, line 29: I guess these tests were assuming `debug=False` by default which is not the case as we had a separate `debug=True` test.

tests/test_web_server.py
```diff
@@ -102,10 +104,54 @@
...
+
+
+async def test_raw_server_html_exception(aiohttp_raw_server, aiohttp_client):
+    exc = RuntimeError("custom runtime error")
+
+    async def handler(request):
+        raise exc
+
+    logger = mock.Mock()
+    server = await aiohttp_raw_server(handler, logger=logger, debug=False)
```
samuelcolvin, line 124: Currently, tests are failing with `ServerDisconnectedError` for this test only. I'm not sure why?

`ServerDisconnectedError` doesn't happen with `debug=True` or if I remove the `Accept` header.

asvetlov: Sorry, I don't remember details

samuelcolvin: Fixed. 

my mistake :nerd_face: :gun:

samuelcolvin:
Maybe we should change the content-type to `text/plain` for the case of `status != 500` line 553?

Doesn't make sense for this to be html I would guess.

asvetlov:
> Maybe we should change the content-type to `text/plain` for the case of `status != 500` line 553?
> 
> Doesn't make sense for this to be html I would guess.

Agree, makes sense.
Please do

aiohttp/web_protocol.py
```diff
@@ -524,22 +524,34 @@ def handle_error(self,
...
+                        msg = traceback.format_exc()
+                else:
+                    msg = 'Server got itself in trouble'
+                msg = '500 Internal Server Error\n\n' + msg
+                resp = Response(status=status, text=msg,
+                                content_type='text/plain')
         else:
             resp = Response(status=status, text=message,
-                            content_type='text/html')
+                            content_type='text/plain')
```
Arfey: u can move this line from else if replace `msg` to `message`.
Also If u create variable content_type u can remove line 542 and 550

samuelcolvin: done.

aiohttp/web_protocol.py
```diff
@@ -523,24 +523,28 @@ def handle_error(self,
...
             else:
-                msg += "Server got itself in trouble"
-                msg = ("<html><head><title>500 Internal Server Error</title>"
-                       "</head><body>" + msg + "</body></html>")
-            resp = Response(status=status, text=msg, content_type='text/html')
-        else:
-            resp = Response(status=status, text=message,
-                            content_type='text/html')
+                if self.debug:
+                    with suppress(Exception):
```
webknjaz: I think this part should be deduplicated and put somewhere before `if 'text/html` ...` line.

aiohttp/web_protocol.py
```diff
@@ -523,24 +523,28 @@ def handle_error(self,
...
-                    msg += '</pre>'
+            msg = 'Server got itself in trouble'
+            if 'text/html' in request.headers.get('Accept', ''):
+                if self.debug:
+                    with suppress(Exception):
+                        tb = html_escape(traceback.format_exc())
+                        msg = '<h2>Traceback:</h2>\n<pre>{}</pre>'.format(tb)
+                message = (
+                    "<html><head>"
+                    "<title>500 Internal Server Error</title>"
```
webknjaz: `500 Internal Server Error` is reused 3 times, you could put it in a var

webknjaz: Actually, you could probably render it from `HTTPStatus.INTERNAL_SERVER_ERROR.description` (Ref: https://docs.python.org/3/library/http.html#http-status-codes)

aiohttp/web_protocol.py
```diff
@@ -523,24 +523,28 @@ def handle_error(self,
...
-                msg = ("<html><head><title>500 Internal Server Error</title>"
-                       "</head><body>" + msg + "</body></html>")
-            resp = Response(status=status, text=msg, content_type='text/html')
-        else:
-            resp = Response(status=status, text=message,
-                            content_type='text/html')
+                if self.debug:
+                    with suppress(Exception):
+                        msg = traceback.format_exc()
+                message = '500 Internal Server Error\n\n' + msg
```
webknjaz: You're not reusing `msg = 'Server got itself in trouble'` from above so messages are not consistent depending on the content-type. I'd probably put it into a separate var like `msg_subtitle`

aiohttp/web_protocol.py
webknjaz: `msg_title = http.HTTPStatus.INTERNAL_SERVER_ERROR.description`?

aiohttp/web_protocol.py
```diff
@@ -523,24 +523,28 @@ def handle_error(self,
         information. It always closes current connection."""
         self.log_exception("Error handling request", exc_info=exc)
 
+        ct = 'text/plain'
         if status == 500:
```
webknjaz:
```suggestion
         if status == http.HTTPStatus.INTERNAL_SERVER_ERROR:
```

aiohttp/web_protocol.py
samuelcolvin: done.

samuelcolvin:
Should all be changed as per requests.

aiohttp/web_protocol.py
```diff
@@ -523,24 +524,30 @@ def handle_error(self,
         information. It always closes current connection."""
         self.log_exception("Error handling request", exc_info=exc)
 
-        if status == 500:
-            msg = "<h1>500 Internal Server Error</h1>"
+        ct = 'text/plain'
+        if status == HTTPStatus.INTERNAL_SERVER_ERROR:
+            title = '500 Internal Server Error'
```
webknjaz:
```suggestion
            title = ' '.join((str(HTTPStatus.INTERNAL_SERVER_ERROR.value), HTTPStatus.INTERNAL_SERVER_ERROR.phrase))
```

aiohttp/web_protocol.py
```diff
@@ -523,24 +524,30 @@ def handle_error(self,
...
+                ct = 'text/html'
             else:
-                msg += "Server got itself in trouble"
-                msg = ("<html><head><title>500 Internal Server Error</title>"
-                       "</head><body>" + msg + "</body></html>")
-            resp = Response(status=status, text=msg, content_type='text/html')
-        else:
-            resp = Response(status=status, text=message,
-                            content_type='text/html')
+                message = title + '\n\n' + (tb or msg)
```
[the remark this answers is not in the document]
samuelcolvin: I don't think it adds anything, but changed to suit you.

aiohttp/web_protocol.py
webknjaz:
```python
In [1]: import http                                                                                                                                                                            

In [3]: http.HTTPStatus.INTERNAL_SERVER_ERROR                                                                                                                                                  
Out[3]: <HTTPStatus.INTERNAL_SERVER_ERROR: 500>

In [5]: http.HTTPStatus.INTERNAL_SERVER_ERROR.phrase                                                                                                                                           
Out[5]: 'Internal Server Error'

In [6]: http.HTTPStatus.INTERNAL_SERVER_ERROR.value                                                                                                                                            
Out[6]: 500

In [8]: http.HTTPStatus.INTERNAL_SERVER_ERROR.description                                                                                                                                      
Out[8]: 'Server got itself in trouble'
```

tests/test_web_server.py
```diff
@@ -26,13 +26,15 @@
...
-    server = await aiohttp_raw_server(handler, logger=logger)
+    server = await aiohttp_raw_server(handler, logger=logger, debug=False)
     cli = await aiohttp_client(server)
     resp = await cli.get('/path/to')
     assert resp.status == 500
+    assert resp.headers['Content-Type'].startswith('text/plain')
 
     txt = await resp.text()
-    assert "<h1>500 Internal Server Error</h1>" in txt
+    assert txt.startswith("500 Internal Server Error")
```
webknjaz: could you please use consistent quotes style?
```suggestion
    assert txt.startswith('500 Internal Server Error')
```

P.S. you can accept such small patch suggestions with right from GitHub UI :)

aiohttp/web_protocol.py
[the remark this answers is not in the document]
webknjaz: Thanks!

tests/test_web_server.py
samuelcolvin: Ok, but this is not related to my pr. The entire file (and project) uses inconsistent quotes.

samuelcolvin:
I've left my laptop at work so will fix on Monday.

aiohttp/web_protocol.py
asvetlov: @webknjaz honestly the original (a plain string) looks more readable for me.

asvetlov: But I can live with any variant, too negligeable reason for contestion.

webknjaz: Agreed. I'd probably go for
```suggestion
            title = str(HTTPStatus.INTERNAL_SERVER_ERROR.value) + ' ' + HTTPStatus.INTERNAL_SERVER_ERROR.phrase
```

webknjaz: would it be better readable?

webknjaz: I'm a bit surprised that `http.HTTPStatus` doesn't have a method to construct a status string. IMHO it should be smth like `title = http.HTTPStatus.INTERNAL_SERVER_ERROR.to_status_string()`

asvetlov: HTTP reason is gone in htttp/2 BTW.
I suspect that very many Web frameworks have own copies of these constants.
+0 for adding `to_status_string()` to CPython, at least I don't want to make a Pull Request (but you can count me on the PR review).

webknjaz:
There's cherry-picker, which you can use manually. No bot doing it currently.

asvetlov:
Thanks!

asvetlov:
Backport PR: #3499

PR #3484, opened 2019-01-04
Change type of Payload.size from Optional[float] to Optional[int]

kxepal:
## What do these changes do?

This changes type of `Payload.size` property from `float` to `int`. Since this property defines amount of bytes which payload contains, it's impossible to have there floating point value.

This behavior was introduced in #3294.

It's possible to break some annotations for users custom Payload instances, but no runtime issues should be caused.

### DIFF
--- a/CHANGES/3484.bugfix
+++ b/CHANGES/3484.bugfix
@@ -1,0 +1,1 @@
+``Payload.size`` type annotation changesd from `Optional[float]` to `Optional[int]`.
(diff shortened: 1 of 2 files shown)


kornicameister:
@kxepal AFAIR I have similar doubts. I think it was coming from another portion of the code that received typings earlier. @asvetlov might remember more as he reviewed the code.

PR #3485, opened 2019-01-04
Fix handling of chunked+gzipped response when first chunk does not give uncompressed data

socketpair:
## Checklist

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -248,21 +248,38 @@ class StreamReader(AsyncStreamReaderMixin):
 
     def begin_http_chunk_receiving(self) -> None:
         if self._http_chunk_splits is None:
+            if self.total_bytes:
+                raise RuntimeError("Called begin_http_chunk_receiving when"
+                                   "some data was already fed")
             self._http_chunk_splits = []
 
     def end_http_chunk_receiving(self) -> None:
         if self._http_chunk_splits is None:
             raise RuntimeError("Called end_chunk_receiving without calling "
                                "begin_chunk_receiving first")
-        if not self._http_chunk_splits or \
-                self._http_chunk_splits[-1] != self.total_bytes:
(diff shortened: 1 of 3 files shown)


aiohttp/streams.py
```diff
@@ -254,15 +254,23 @@ def end_http_chunk_receiving(self) -> None:
         if self._http_chunk_splits is None:
             raise RuntimeError("Called end_chunk_receiving without calling "
                                "begin_chunk_receiving first")
-        if not self._http_chunk_splits or \
-                self._http_chunk_splits[-1] != self.total_bytes:
-            self._http_chunk_splits.append(self.total_bytes)
```
socketpair, line 259: Previous code added zero here.

webknjaz: Do is the whole difference in `or self.total_bytes == 0` clause?

aiohttp/streams.py
```diff
@@ -254,15 +254,23 @@ def end_http_chunk_receiving(self) -> None:
...
 
-            # wake up readchunk when end of http chunk received
-            waiter = self._waiter
-            if waiter is not None:
-                self._waiter = None
-                set_result(waiter, False)
+        pos = self._http_chunk_splits[-1] if self._http_chunk_splits else 0
```
webknjaz, line 266: You might want to add a comment about the meaning of `0` here.

tests/test_streams.py
```diff
@@ -719,6 +719,34 @@ def cb():
         assert b'' == data
         assert not end_of_chunk
 
+    async def test_read_empty_chunks(self) -> None:
```
webknjaz, line 722: Could you please add a docstring here?

aiohttp/streams.py
socketpair: The problem is in adding zero as the first item of the array. Yes, total_bytes can be zero when the function is called.

Issue #2302, opened 2017-10-04
Issue with MultipartReader


thehesiod:
I'm trying to mock the google API batch endpoint which takes a multipart request.  It ends up sending a request like this:

```
--===============0167281537324890874==
Content-Type: application/http
MIME-Version: 1.0
Content-Transfer-Encoding: binary
Content-ID: <ad845523-bc65-4f77-bfad-2b8709570634+1>

GET /gmail/v1/users/me/messages/660cf5e1784d4d98b8d3adeabebb413e?format=metadata&alt=json HTTP/1.1
Content-Type: application/json
MIME-Version: 1.0
accept: application/json
accept-encoding: gzip, deflate
user-agent: google-api-python-client/1.6.2 (gzip)
Authorization: Bearer TOKEN
Host: 192.168.16.6:63191


--===============0167281537324890874==--
```

with headers:
```
{'Host': '192.168.16.6:63191', 'Content-Length': '553', 'content-type': 'multipart/mixed; boundary="===============0167281537324890874=="', 'authorization': 'Bearer TOKEN', 'user-agent': 'Python-httplib2/0.10.3 (gzip)', 'accept-encoding': 'gzip, deflate'}
```

And with a handler like the following:
```python
    async def gmail_batch(self, req: web.Request):
        reader = await req.multipart()
        async for part in reader:
            async for data in part:
                print()
```

I get the exception: 
```python
  File "/Users/amohr/dev/Operations/email_security/svc_tests/mock_google_svcs.py", line 116, in gmail_batch
    async for data in part:
  File "/Users/amohr/.pyenv/versions/3.6.2/lib/python3.6/site-packages/aiohttp/multipart.py", line 235, in __anext__
    part = yield from self.next()
  File "/Users/amohr/.pyenv/versions/3.6.2/lib/python3.6/site-packages/aiohttp/multipart.py", line 242, in next
    item = yield from self.read()
  File "/Users/amohr/.pyenv/versions/3.6.2/lib/python3.6/site-packages/aiohttp/multipart.py", line 261, in read
    data.extend((yield from self.read_chunk(self.chunk_size)))
  File "/Users/amohr/.pyenv/versions/3.6.2/lib/python3.6/site-packages/aiohttp/multipart.py", line 279, in read_chunk
    chunk = yield from self._read_chunk_from_stream(size)
[...]

PR #3492, opened 2019-01-06
Support reading multipart data with `\n` (`LF`) lines

kxepal:
## What do these changes do?

While RFC clearly says about `CRLF` newlines, there quite a lot of
implementations which uses just `LF`. Even Python's stdlib produces
multiparts with `\n` newlines by default for compatibility reasons.

We wouldn't change how we produce multipart content - here we follow
RFC. However, we can detect `\n` lines quite easily which makes their
support quite cheap.

No notable, except now they can read multipart data with "popular" newlines.

## Related issue number

#2302

### DIFF
--- a/CHANGES/2302.feature
+++ b/CHANGES/2302.feature
@@ -1,0 +1,1 @@
+Support reading multipart data with `\n` (`LF`) lines
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -243,7 +243,7 @@ class BodyPartReader:
         headers: Mapping[str, Optional[str]],
         content: StreamReader,
         *,
-        _newline: bytes = b'\r\n',
+        _newline: bytes = b'\r\n'
     ) -> None:
         self.headers = headers
         self._boundary = boundary
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -237,11 +237,16 @@ class BodyPartReader:
 
     chunk_size = 8192
 
-    def __init__(self, boundary: bytes,
-                 headers: Mapping[str, Optional[str]],
-                 content: StreamReader) -> None:
+    def __init__(
+        self,
+        boundary: bytes,
+        headers: Mapping[str, Optional[str]],
+        content: StreamReader,
+        newline: bytes = b'\r\n',
+    ) -> None:
         self.headers = headers
         self._boundary = boundary
+        self._newline = newline
         self._content = content
         self._at_eof = False
         length = self.headers.get(CONTENT_LENGTH, None)
@@ -300,8 +305,8 @@ class BodyPartReader:
         if self._read_bytes == self._length:
             self._at_eof = True
         if self._at_eof:
-            clrf = await self._content.readline()
-            assert b'\r\n' == clrf, \
+            newline = await self._content.readline()
+            assert newline == self._newline, \
                 'reader did not read all the data or it is malformed'
         return chunk
 
@@ -328,11 +333,15 @@ class BodyPartReader:
         assert self._content_eof < 3, "Reading after EOF"
         assert self._prev_chunk is not None
         window = self._prev_chunk + chunk
-        sub = b'\r\n' + self._boundary
+
+        intermeditate_boundary = self._newline + self._boundary
+
         if first_chunk:
-            idx = window.find(sub)
+            pos = 0
         else:
-            idx = window.find(sub, max(0, len(self._prev_chunk) - len(sub)))
+            pos = max(0, len(self._prev_chunk) - len(intermeditate_boundary))
+
+        idx = window.find(intermeditate_boundary, pos)
         if idx >= 0:
             # pushing boundary back to content
             with warnings.catch_warnings():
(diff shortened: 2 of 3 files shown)


asvetlov:
Thanks for the PR!
Does newline style autodetection make sense?

thehesiod (changes requested):
~~This does not fix `StreamReader.readline`, called via `MultipartReader.next`~~

kxepal:
IIRC, on merge block right below the comments there should be some button about review cancellation. As backup plan, I can just change the latest commit hash via rebase over master or just ask @asvetlov for `sudo git merge`. In anyway, no worries about (:

thehesiod:
ok so, after a few tries it works, will have to look into it more later, seems like a multithreading issue with aiohttp (one loop and server per thread with new release), but for now:

- I was calling `web.Request.multipart()`, however there's no param to set for the newline, so I now have to do: `MultipartReader(req.headers, req.content, newline=b'\n')`, should we add a param to this method?
- for automatic detection, this would be really nice because otherwise your server can only accept one type or the other, but not both.  Similar for reading multipart, somehow you'd have to detect/know what the newlines are otherwise.  I'd prefer having an auto-detect mechanism.

thanks guys! sorry about that little disruption, got my virtualenvs confused :)

thehesiod:
btw I don't see anything about cancelling my review there, perhaps asvetlov or you can dismiss my review

kxepal:
The initial idea is about to autodetect on MultipartReader side, while BodyPartReader have to be informed about what kind of newline to expect. So it should work in seconds case. But since the former one is deep in implementation details, everything should work for you without worry about what kind of newline is there.

kxepal:
Hm...I should make a test case for this. Hold on.

thehesiod:
sorry not following, from what I understand of this PR, you need to state upfront to your multipart reader what the newline is....which means if I write a server, two things are true:
1) It needs to know upfront what the multipart newline format is
2) It can only accept or format or another

These are two big negatives.  In my case it's ok, because I know upfront what it is, and I don't need to support mixed newline requests.  However in general this is not true if you have a generic server, you can't control what format clients will send their newlines.

kxepal:
No, you don't need to care about what newline is. It handles by [this part](https://github.com/aio-libs/aiohttp/pull/3492/files#diff-235eeadd1960735e1191879aa500b7b7R648) of changes for you. In other words: first boundary dictates newline format fot the rest ones.

thehesiod:
very cool! 🙇

asvetlov:
Sorry, I don't follow the design.

Looks like `newline` argument is accepted here: https://github.com/aio-libs/aiohttp/pull/3492/files#diff-235eeadd1960735e1191879aa500b7b7R538 
but overridden later here: https://github.com/aio-libs/aiohttp/pull/3492/files#diff-235eeadd1960735e1191879aa500b7b7R651

Did I miss something?

kxepal:
@asvetlov 
Multipart is recursive format. Once we detect newline at the very first boundary we have to pass it around for nested readers. So we have to have it as an argument and have to rewrite it during parsing, but only once. 
I just noticed that current implementation allows to have mix of different newlines - that's need to forbid for good.

asvetlov:
Is it ready to merge?

kxepal:
Sorry, not yet. Will be able to finish this a bit later today.

kxepal:
@asvetlov 
Ok, I think it's ready. I was tries to catch the mixed newlines case when CRLF and LF may vary, but it seems to be not issue for us. PR rebased. If everythink ok for you and @thehesiod  let's merge it.

aiohttp/multipart.py
```diff
@@ -237,11 +237,16 @@ class BodyPartReader:
...
 
-    def __init__(self, boundary: bytes,
-                 headers: Mapping[str, Optional[str]],
-                 content: StreamReader) -> None:
+    def __init__(
+        self,
+        boundary: bytes,
+        headers: Mapping[str, Optional[str]],
+        content: StreamReader,
+        newline: bytes = b'\r\n',
```
kxepal: To not cause any confusion, may be mark it somehow as internal? Like kwarg only or `_newline`. What do you think?

asvetlov: Makes sense for both keyword only and `_newline` proposals.
It will prevent users from using this *special* argument.

kxepal: Ok, will make both changes.

kxepal: Fixed in https://github.com/aio-libs/aiohttp/pull/3492/commits/5536671a48e9f72a4ff57f13c7a6ea895bbbd9fd

asvetlov (approved):

asvetlov:
Thanks!

thehesiod:
Nice work, thank you!!

kxepal:
:beers:

CHANGES/2302.feature
```diff
@@ -0,0 +1 @@
+Support reading multipart data with `\n` (`LF`) lines
```
webknjaz, line 1: Hm.. we need to watch for incorrect single quotes in RST...

PR #3504, opened 2019-01-08
Allow run_app(access_log=None)

hynek:
## What do these changes do?

The docs suggest that you can disable request logging by passing `access_log=None` to run_app.

However:

1. It breaks when loop's debug is True
2. The type stubs don't reflect that

This PR fixes both.

Documented behavior works under more conditions now. :)

## Checklist


CHANGES/3504.bugfix
```diff
@@ -0,0 +1 @@
+Fix type stubs for ``aiohttp.web.run_app(access_log=True)`` and fix edge case of ``access_log=True?? and the event loop being in debug mode.
```
kxepal: `??` need to replace with closing back quotes.

hynek: oh yeah sorry, notebook keyboard :)

aiohttp/web.py
```diff
@@ -383,15 +383,15 @@ def run_app(app: Union[Application, Awaitable[Application]], *,
...
+            access_log: Optional[logging.Logger]=access_logger,
             handle_signals: bool=True,
             reuse_address: Optional[bool]=None,
             reuse_port: Optional[bool]=None) -> None:
     """Run an app locally"""
     loop = asyncio.get_event_loop()
 
     # Configure if and only if in debugging mode and using the default logger
-    if loop.get_debug() and access_log.name == 'aiohttp.access':
+    if loop.get_debug() and access_log and access_log.name == 'aiohttp.access':
```
webknjaz, line 394: maybe
```suggestion
    if loop.get_debug() and getattr(access_log, 'name', None) == 'aiohttp.access':
```

aiohttp/web.py
```diff
@@ -269,7 +269,7 @@
                    backlog: int=128,
                    access_log_class: Type[AbstractAccessLogger]=AccessLogger,
                    access_log_format: str=AccessLogger.LOG_FORMAT,
-                   access_log: logging.Logger=access_logger,
+                   access_log: Optional[logging.Logger]=access_logger,
```
Arfey, line 272: why do u use `Optional`, and when access_log should be None?

Arfey: i just don't understand when u need to drop logging :)

hynek: Firstly, this is just an adoption of the stubs to reflect existing [docs](https://docs.aiohttp.org/en/stable/logging.html#access-logs), secondly I ask myself the vice versa question every time I try to get rid of some framework's logging. I have “dumb” HTTP logs in my load balancer and I want my applications to have only structured application logs.

Arfey: The load balancer is story about production, right? do u use debug mode for production?
Now, u can put your custom logger and this should solve your problem.

hynek: This isn't about *my* problems, this is about making the code do what the docs say. :)

Arfey: so maybe the problem is not in the code and in the documentation :)
I don't find any test for this behavior and this feature never worked. I am not sure that this feature is required.

hynek: It worked when debug was off which is the normal case. I ran into the edge case of debug=True thanks to mypy.

Arfey: You are right without mypy all is fine. I think will be cool have some test for this case.

asvetlov:
Thanks!

PR #3506, opened 2019-01-08
Skip SSL tests on Linux-32

asvetlov:


webknjaz:
Thanks, this should work for now.

webknjaz:
@asvetlov FTR the fix has been released as `trustme==0.5.0`

asvetlov:
Cool!

PR #3517, opened 2019-01-10
WIP: Add test for #3515

mihajenko:
Add test for #3515 (Send custom per-request cookies even if session jar is empty)

## What do these changes do?

Adds a test for #3515 but fix

Test suite is updated.

## Checklist


tests/test_client_session.py
```diff
@@ -493,6 +493,26 @@ class UnexpectedException(BaseException):
...
+    async def handler(_):
+        return web.Response()
+
+    app = web.Application()
+    app.router.add_route('GET', '/', handler)
+
+    # no cookies should be set at init
+    session = await aiohttp_client(app, cookies=None)
+
+    with mock.patch("http.cookies.SimpleCookie.load") as patched:
```
asvetlov: Drop mocks, please.
Making sure that `SimpleCookie.load()` is called means nothing from the library perspective.
It doesn't guarantee to keep the feature working after code refactoring.
The test purpose is not obvious.
Explicit check for sent cookies on the server side is much better.

PR #3533, opened 2019-01-13
Permit empty reason strings in ClientResponse.raise_for_status()

### LINKED ISSUE #3532: ClientResponse.raise_for_status() raises an AssertionError for empty reason
## Long story short

`ClientResponse.raise_for_status()` raises an `AssertionError` for a >= 400 status code if its `reason` is an empty string.

## Expected behaviour

I expect `ClientResponse.raise_for_status()` to raise a `ClientResponseError` if it’s invoked on a `ClientResponse` with a status >= 400, and to either:

* raise a `ClientError` (or some subclass of `ClientError`) if it’s called on a `ClientResponse` with a `reason = None`, regardless of the HTTP status code, if this can only happen when `raise_for_status()` has been called before `start()` has been called; or
* convert a `ClientResponse.reason` of `None` into `""` if `None` is a valid value for `ClientResponse.reason`

## Actual behaviour

* `ClientResponse.raise_for_status()` raises an AssertionError if it’s invoked on a `ClientResponse` when a status >= 400 and `bool(reason)` is `False`, which includes `reason == ""`
* `ClientResponse.raise_for_status()` raises a ClientResponseError if it’s invoked on a `ClientResponse` with a status >= 400 if the `reason` is not falsey
* `ClientResponse.raise_for_status()` raises no error if it’s invoked on a `ClientResponse` with a status < 400 and a `reason = None`

## Steps to reproduce

A colleague and I uncovered this using [aioresponses](https://github.com/pnuckowski/aioresponses), which currently doesn’t set a `reason` on the `ClientResponse` at all – see https://github.com/pnuckowski/aioresponses/issues/114.

You can reproduce this with the example client and server in this gist: https://gist.github.com/rhwlo/5ff3cb6bae8deb6abb4a1e527ad9583e

## Your environment

I’m using the latest `aiohttp` from the repository, but this problem shows up only in `v3.5.0` and newer.

rhwlo:
## What do these changes do?

* Allows `ClientResponse.raise_for_status` to raise a `ClientResponseError` in the case of an HTTP Reason-Phrase that is an empty string
* Raises a `TypeError` if the `ClientResponse.reason` is `None` (since we can’t make a `ClientResponseError` with a message of `None`)
* Fixes https://github.com/aio-libs/aiohttp/issues/3532

* Calling `raise_for_status` on a status with a reason phrase that’s an empty string will raise a `ClientResponseError` now instead of an `AssertionError`

## Related issue number

https://github.com/aio-libs/aiohttp/issues/3532

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -931,8 +931,10 @@ class ClientResponse(HeadersMixin):
         return noop()
 
     def raise_for_status(self) -> None:
+        # self.reason is only None if self.start() hasn't been called:
+        if self.reason is None:
+            raise TypeError("Expected a Reason-Phrase as a string")
         if 400 <= self.status:
-            assert self.reason  # always not None for started response
             self.release()
             raise ClientResponseError(
                 self.request_info,
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -931,10 +931,9 @@ class ClientResponse(HeadersMixin):
         return noop()
 
     def raise_for_status(self) -> None:
-        # self.reason is only None if self.start() hasn't been called:
-        if self.reason is None:
-            raise TypeError("Expected a Reason-Phrase as a string")
         if 400 <= self.status:
+            # reason should always be not None for a started response
+            assert self.reason is not None
             self.release()
             raise ClientResponseError(
                 self.request_info,
--- a/CHANGES/3532.bugfix
+++ b/CHANGES/3532.bugfix
@@ -1,0 +1,2 @@
+Raise a ClientResponseError instead of an AssertionError for a blank
+HTTP Reason Phrase.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -120,6 +120,7 @@ Joel Watts
 Jon Nabozny
 Joongi Kim
 Josep Cugat
+Joshu Coats
 Julia Tsemusheva
 Julien Duponchelle
 Jungkook Park
--- a/tests/test_client_response.py
+++ b/tests/test_client_response.py
@@ -669,6 +669,24 @@ def test_raise_for_status_4xx() -> None:
     assert response.closed
 
 
+def test_raise_for_status_4xx_without_reason() -> None:
+    response = ClientResponse('get', URL('http://def-cl-resp.org'),
+                              request_info=mock.Mock(),
+                              writer=mock.Mock(),
+                              continue100=None,
+                              timer=TimerNoop(),
+                              traces=[],
+                              loop=mock.Mock(),
+                              session=mock.Mock())
+    response.status = 404
+    response.reason = ''
+    with pytest.raises(aiohttp.ClientResponseError) as cm:
+        response.raise_for_status()
+    assert str(cm.value.status) == '404'
+    assert str(cm.value.message) == ''
+    assert response.closed
+
+
 def test_resp_host() -> None:
     response = ClientResponse('get', URL('http://del-cl-resp.org'),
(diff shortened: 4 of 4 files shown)


aiohttp/client_reqrep.py
```diff
@@ -931,8 +931,10 @@ def release(self) -> Any:
         return noop()
 
     def raise_for_status(self) -> None:
+        # self.reason is only None if self.start() hasn't been called:
+        if self.reason is None:
+            raise TypeError("Expected a Reason-Phrase as a string")
```
asvetlov: An exception type and especially text is meaningless for an end user.
The text describes a consequence (the `reason` field is not set) but hides an actual problem.

Moreover, assertion exists only for sake of `mypy`.
If you don't like the assertion -- please find another way to satisfy mypy.

`assert self.reason is not None` could be the simplest possible fix, even no test is needed for the case.

asvetlov: Well, a test would be nice but please use a functional test for an empty `reason`.
https://github.com/aio-libs/aiohttp/blob/master/tests/test_client_functional.py can give an inspiration how a functional client test looks like

rhwlo: I’ll switch it to `assert self.reason is not None` if that’s the desired outcome.

asvetlov (approved):

asvetlov:
Thanks!

ctg3:
Why is reason string required? I have a webserver that doesn't return any reason phrase (None). If self.reason is None, can we just change it to self.reason = '' and still satisfy mypy?

Also see a discussion here: https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase

vaultah:
Maybe make a release that contains a fix for this? No hurry though, it's only been three months.

Solvero:
I have encountered this problem after aiohttp upgrade to 3.5.4. My server doesn't return Response-Reason too. HTTP/1.1 [specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1.1) says that:

> The Status-Code is intended for use by automata and the Reason-Phrase is intended **for the human user**. The **client is not required to examine** or display the Reason- Phrase. 

Please see also:
https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase

gwerbin:
I appreciate all the work the contributors are doing on this library, but it would have been nice to merge and release this as a 3.5.5 "patch". It looks like you're waiting for 3.6 release or something like that?

In the meantime, I'm using my own implementation of this, basically copy-and-pasted from the source code:
```python
def raise_for_status(response: aiohttp.ClientResponse) -> None:
    if 400 <= response.status:
        response.release()
        raise aiohttp.ClientResponseError(
            response.request_info,
            response.history,
            status=response.status,
            message=response.reason or '',
            headers=response.headers
        )
```

webknjaz:
@gwerbin `master` branch is for 4.0. If you want, you can submit a backport PR against `3.5` branch.

gwerbin:
Thanks @webknjaz, My one concern is about this line:

> reason should always be not None for a started response

Does some other part of the application or test suite depend on this being true? I don't want to break anything else with my quick `or ''` fix.

Alternatively, why would reason be `None` here, rather than defaulting to an empty string? I feel like fixing this at the `raise_for_status` point doesn't actually address the problem, that reason can be `None` in the first place instead of an empty string.

webknjaz:
@gwerbin you should normally cherry-pick the entire PR and not try to rip off some parts of it. This includes tests.

fiendish:
raise_for_status on a response object is still giving AssertionError instead of ClientResponseError in the latest release?

PR #3539, opened 2019-01-14
Disallow creation of aiohttp objects without running event loop

asvetlov:
`ClientSession`, `Connector` and others are affected.


blueyed:
Can you explain the reason for this?
I've thought that having it on the module level would be good / comfortable way to share it / use a single instance only.

PR #3542, opened 2019-01-14
Forbid unknown web.Application attributes

asvetlov:


CHANGES/3542.removal
```diff
@@ -0,0 +1 @@
+Forbid unknown ``web.Application`` attributes access
```
kxepal: Hm...may be clarify this as "Setting ``web.Application`` custom attributes is now forbidden", or sort of, but current one is fine for me since there was a warning about.

asvetlov: Updated.
Your text is much better, thanks

PR #3545, opened 2019-01-15
Drop custom router support

asvetlov:


kxepal:
Yeah, I know. It's just quite strange to implement router thought router with the new design.

Issue #3551, opened 2019-01-17
Replace tmpdir fixture with tmp_path in tests


asvetlov:
I always stumble on `py.path` methods.
A new `pytest` supports `tmp_path` fixture https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-factory-fixture to work with `pathlib.Path`.

The issue is easy and waits for a champion.

gyermolenko:
I found it relevant to current issue, adding for context: ref https://github.com/pytest-dev/pytest/issues/4680  and two other `tmp_path` defects were fixed in pytest 4.2.
And pytest 4.2 has just beed released.

PR #3553, opened 2019-01-18
Replace tmpdir fixture with tmp_path in tests (#3551)

bruziev:
## Related issue number
#3551

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,24 +1,11 @@
 import hashlib
-import pathlib
 import platform
-import shutil
 import ssl
-import tempfile
 
 import pytest
 import trustme
 
 pytest_plugins = ['aiohttp.pytest_plugin', 'pytester']
-
-
-@pytest.fixture
-def shorttmpdir():
-    """Provides a temporary directory with a shorter file system path than the
-    tmpdir fixture.
-    """
-    tmpdir = pathlib.Path(tempfile.mkdtemp())
-    yield tmpdir
-    shutil.rmtree(tmpdir, ignore_errors=True)
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -1,26 +1,13 @@
...
-import tempfile
 
 import pytest
 import trustme
 
 pytest_plugins = ['aiohttp.pytest_plugin', 'pytester']
 
 
-@pytest.fixture
-def shorttmpdir():
```
asvetlov, line 15: IIRC thee fixture exists to satisfy cases when `path is too long` was reported by CI.
If travis/appveyor are happy without the fixture -- that's fine to me.
We can reinvent it later if needed.

tests/test_connector.py
```diff
@@ -43,8 +43,8 @@ def ssl_key():
 
 
 @pytest.fixture
-def unix_sockname(shorttmpdir):
-    sock_path = shorttmpdir / 'socket.sock'
+def unix_sockname(tmpdir):
```
asvetlov, line 46: `tmpdir` -> `tmp_path`

tests/conftest.py
blueyed: If this needs some fix later again it should be considered to get fixed in pytest also.
(Some quick thought: maybe the (temp) base dir could be shortened on Travis if that causes trouble, e.g. by exporting TMPDIR or something alike.)

asvetlov:
thanks!

tests/test_connector.py
blueyed: This was not done/changed.

asvetlov: Yes, I understand.
A task for the next PR?

blueyed: => https://github.com/aio-libs/aiohttp/pull/3555

PR #3555, opened 2019-01-19
tests: unix_sockname: use tmp_path

blueyed:
Ref: https://github.com/aio-libs/aiohttp/issues/3551

The intention from/in https://github.com/aio-libs/aiohttp/issues/3551 is not really clear to me.

Mainly via https://github.com/aio-libs/aiohttp/pull/3553#discussion_r249247018, but should be done in other places then also?

@asvetlov 
Please consider picking it up.

### DIFF
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -43,9 +43,8 @@ def ssl_key():
 
 
 @pytest.fixture
-def unix_sockname(tmpdir):
-    sock_path = tmpdir / 'socket.sock'
-    return str(sock_path)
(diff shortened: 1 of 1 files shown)


asvetlov:
Yes, the intention is replacing `tmpdir` with `tmp_path` everywhere in `tests` folder.
Consider the work as housekeeping, we need to do such tasks periodically.
`py.path` is great but `pathlib` is a part of standard batteries, it is much better known.

asvetlov:
Thanks!

PR #3557, opened 2019-01-20
add exc_info to access_logger

samuelcolvin:
## What do these changes do?

~~Add `exc_info` argument to `AccessLogger.log`.~~

call `AccessLogger.log` with the current exception available from `sys.exc_info()`.

I'm trying to implement a custom `AccessLogger` to both print useful information to the terminal during development and log information about failed requests to sentry or equivalent in production.

The problem is that `aiohttp.server` log methods are only called on an exception and without the request and response objects, while `AccessLogger.log` is called on every request but without any information about the exception.

Currently, to achieve what I want I would have to either
* print/post two log messages, one with details on the request, one with the exception. That would ok in the terminal but be very ugly in sentry or equivilant.
* use middleware which I guess it what most people do, but means ignoring `AccessLogger` completely and modifying the app. This is particularly problematic in `aiohttp-devtools` which is one of the places I want to use this.

Only if they've implemented a custom `AccessLogger`

## Checklist


asvetlov:
@samuelcolvin would you fix mypy check?

samuelcolvin:
of course, I just wanted to check you think this is a good idea before I do more work on it?

asvetlov:
Honestly, I don't know.
It changes our public API and breaks existing access loggers.
On other hand the logger can ask for `sys.exc_info()` internally if needed.
Is the exception context cleared by aiohttp protocol itself?

samuelcolvin:
It's not cleared explicitly, but I tried `sys.exc_info()` and it didn't work, I'll check again and confirm.

The option is to add `exc_info` to the response so it doesn't break existing functionality but is still available?

samuelcolvin:
Ok, I'll try that.

samuelcolvin:
Ok, all changed, but I'm not sure why tests are failing, seems to be something to do with cython.

samuelcolvin:
@asvetlov I think this is basically done.

However I have another question/feature request:

How would you feel about `access_logger.log` being a coroutine? In my use case (detailed error reporting), there is a reasonable chance that the `log` method will need to make async calls. Eg. to populate the `user` field of sentry an app might want to access the database to get the user's name from the user's ID which might be in the session, or the error details might want access to the request body even if it hasn't already been read.

I know this is currently possible with `middleware`, but I think using `AccessLogger` would be more elegant as it avoid modifying the app, so can be completely handled but an "app runner".

This could be done in a backwards compatible way by adding a `alog` method which by default just calls `log` or by inspecting `log` to see if it's a coroutine.

asvetlov:
@samuelcolvin would you fix merge conflicts?
I review the PR just after getting it done

samuelcolvin:
will do

samuelcolvin:
Hopefully conflicts are fixed. How do I sign the contributor agreement?

asvetlov:
Just ignore CLA for now.
The bot has no proper setup still, sorry.

samuelcolvin:
ok no problem, passing except for CLA.

aiohttp/web_protocol.py
```diff
@@ -342,9 +342,11 @@ def force_close(self) -> None:
     def log_access(self,
                    request: BaseRequest,
                    response: StreamResponse,
-                   time: float) -> None:
+                   request_start: Optional[float]) -> None:
```
asvetlov: Looks like reverting back `request_start` to strict float simplifies a code.

aiohttp/web_protocol.py
```diff
@@ -402,6 +404,7 @@ def _process_keepalive(self) -> None:
 
             message, payload = self._messages.popleft()
 
+            now = None
```
asvetlov: I think we can always use unconditional `now = loop.time()` here.
The call is relative cheap in comparison to all other method's logic.
Do you have evidence that conditional initialization speeds up things?

samuelcolvin: ok

samuelcolvin: fixed, I had no evidence of the speed, I just kept it the same as before for consistency.

I've also renamed the variable since `now` was a dumb name.

samuelcolvin:
> @asvetlov I think this is basically done.
> 
> However I have another question/feature request:
> 
> How would you feel about `access_logger.log` being a coroutine? In my use case (detailed error reporting), there is a reasonable chance that the `log` method will need to make async calls. Eg. to populate the `user` field of sentry an app might want to access the database to get the user's name from the user's ID which might be in the session, or the error details might want access to the request body even if it hasn't already been read.
> 
> I know this is currently possible with `middleware`, but I think using `AccessLogger` would be more elegant as it avoid modifying the app, so can be completely handled but an "app runner".
> 
> This could be done in a backwards compatible way by adding a `alog` method which by default just calls `log` or by inspecting `log` to see if it's a coroutine.

@asvetlov did you see this question?

asvetlov:
Async log method makes sense maybe. I have no objections and have no strong +1 feeling.
It should be done in a backward compatible way, sure.
It requires a separate PR IMHO.

BTW do you inherit your logger from `AccessLogger` or `AbstractAccessLogger`?

asvetlov:
The tests failure is not related to PR but a result of #3520

samuelcolvin:
`AbstractAccessLogger`, see [here](https://github.com/aio-libs/aiohttp-devtools/blob/master/aiohttp_devtools/runserver/log_handlers.py#L11) but mostly I'm not using access logger's at all because of exception context and the need for a coroutine.

asvetlov:
After thinking about the idea of `.alog()` it confuses me.
There is better (I hope) proposal:
1. Introduce a new `AbstractAsyncAccessLogger` class with the only `async def log(...)` method -- like `AbstractAccessLogger.log` but async function.
2. Provide adaptor class that implements `AbstractAsyncAccessLogger` interface and accepts synchronous `AbstractAccessLogger` instance in its constructor.
3. On `RequestHandler` initialization do check of the logger. If it is an instance of `AbstractAccessLoger` -- wrap it with adaptor. If the logger is an instance if `AbstractAsyncAccessLogger` -- pass it as is, raise an error otherwise.

What do you think?

samuelcolvin:
Agreed, I think 3 will be the easiest appraoch. Will do in another PR

asvetlov:
Bump to rebuild CI

asvetlov:
I thought that my proposal has all three bullets necessary, not only the last one :)

asvetlov:
Let's merge the PR on green

asvetlov:
Done!
Thanks, @samuelcolvin !

kxepal:
:slowpoke:,but :+1:  for async log. This will help a lot in future.

samuelcolvin:
working on it in #3767

Issue #3440, opened 2018-12-08
Streams reorganization


asvetlov:
I'd like to share my thoughts about streams architecture.

Current state
------------------

Now we have a `StreamReader` class.
`ClientSession` has an `auto_compress` parameter for optional disabling a stream compression (gzip, deflate etc.)

WebSockets have own buffer implementation for reading a data from a stream.

Proposal
---------------

Let's use `StreamReader` for everything.
On the bottom, there is a `raw` stream which is a good data source for both zgipped content and websockets data.
Unzipper provides an own class that is a wrapper around a raw stream. Unzipper's `.read()` calls `raw.read()` and pass the given data chunks into decompressor.
Websocket reader does the same.

The raw stream becomes the only TCP flow control aware structure, all descendants just use it.
Raw -> Unzipped
Raw -> WebSocket

The raw stream is accessible from Unzipped stream by it's `.raw` attribute. Something like [`io.BufferedIOBase.raw`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.raw)

I believe the proposal simplifies the design.

Optimizations
-------------------

Since asyncio 3.7 we have an advanced [`BufferedProtocol`](https://docs.python.org/3/library/asyncio-protocol.html#buffered-streaming-protocols) (old good `asyncio.Protocol` for old Python's can be supported easily on top of it, the price is more memory copy operations).

For reducing mem-copy `BufferedProtocol` *should* use mem structures like Linux Kernel [SLAB allocators](https://en.wikibooks.org/wiki/The_Linux_Kernel/Memory#Slab_Allocator), the protocol was *designed* for it.
A simplified idea is: 
- Allocate a large memory page (4Kb or more, we can start from a simple scenario but optimize an allocator for 4MB data chunks later)
- Fetch a data chunk into an allocated page
- Free a page when all page's data is consumed by a raw `StreamReader`

P.S.
IIRC `nginx` uses an interesting custom allocator strategy: it has a memory context tightly coupled with HTTP request. The memory is never released until the request is gone, then all memory allocated by the request processing goes away. We cannot steal the NGINX memory strategy as-is: a lot of Python objects are created an released during HTTP request life-cycle but we can optimize at least `StreamReader` implementation for this case.

webknjaz:
Sounds good. It looks typical for other networking protocol wrappers to do the same.

PR #3568, opened 2019-01-21
Waiters refac

socketpair:
## Checklist


socketpair:
@asvetlov Please take a look. Please invite others to look at it thoroughly.

socketpair:
@mnach Please review

aiohttp/streams.py
```diff
@@ -388,31 +388,29 @@ def end_http_chunk_receiving(self) -> None:
         of the data corresponds to the end of a HTTP chunk , otherwise it is
         always False.
```
socketpair, line 389: @asvetlov I would change the API. This function is meaningless when chunked encoding is not used. So I would raise exception if this function used on non-chunked encoding (can be checked using `self._http_chunk_splits is None`)

asvetlov:
I agree that a waiter should return `None`, boolean doesn't make any sense and looks confusing.
I'm worried about `_wait` method modification.
In my mind, we have to make streams implementation much simpler to understanding.
I've filed https://github.com/aio-libs/aiohttp/issues/3440 issue a month ago.
Skip slab allocators part when reading.

Long story short:
1. Introduce `AbstractStreamReader`. Initially, it should have all `StreamReader` public methods. We can reconsider dropping (with depreciation) a part of them later if needed.
2. Implement `RawStreamReader` for raw body data, `ChunkedStreamReader` for chunked encoding, a reader for compressed source and WebSocket etc.
3. All readers are based on the raw reader. Chunked reader use it as `.raw` attribute for example. The decompressor can be built on top of the raw reader or chunked one, pointing to the source stream be own `raw` attribute. And so on.

What do you think about this structure? Is it viable, did I miss something?

socketpair:
@asvetlov I have fixed everything. Merge ?

socketpair:
Merge ?

socketpair:
PING

asvetlov:
Hi.
Sorry, was terrible busy last week.
Will check the PR in a few hour.
I have a feeling that the PR worth a new 3.5 release, isn't it?

socketpair:
@asvetlov  regarding 3.5 release: it does not add any new features or fixes any bugs. Since I don't know rules for LTS releases, I can not answer.

socketpair:
OK, I'm sure changes are correct, so please merge.

asvetlov:
Thank you and sorry for waiting

PR #3580, opened 2019-01-28
Drop explicit loop

asvetlov:
Use `asyncio.get_event_loop()` instead if the loop instance is needed.
All aiohttp objects work with the currently running loop, a creation of aiohttp instances, e.g. `ClientSession` when the loop is not running is forbidden.

As a side effect of PR passing callables for `aiohttp_server()` and `aiohttp_client` pytest fixtures are forbidden, please call these callables explicitly.


aiohttp/cookiejar.py
```diff
@@ -334,9 +332,8 @@ class DummyCookieJar(AbstractCookieJar):
 
     """
 
-    def __init__(self, *,
-                 loop: Optional[asyncio.AbstractEventLoop]=None) -> None:
-        super().__init__(loop=loop)
+    def __init__(self) -> None:
```
kxepal: This `__init__` looks redundant.

asvetlov: It initializes `self._loop`.
I agree that `__init__` in `AbstractCookieJar` can be avoided but the change is not backward compatible a little

asvetlov: After rethinking I've decided to drop `AbstractCookieJar.__init__`.
Fixed.

PR #3585, opened 2019-01-29
Adds support for class methods in route tables.

### LINKED ISSUE #2996: Uses class methods with aiohttp.web.RouteTableDef
## Long story short
Whenever you place a route as part of a class, it does not allow this to be possible, unless you remove te response argument, leaving just self (which creates problems as self is actually response).

## Expected behaviour
Provide the webpage as it would like this: 
```py
from aiohttp import web
import asyncio

routes = web.RouteTableDef()
app = web.Application()

@routes.get("/")
async def root(response):
    return web.Response(text="5")

app.add_routes(routes)
web.run_app(app)
```

## Actual behaviour
Errors with
```py
Error handling request
  File "D:\Program Files\Python3\lib\site-packages\aiohttp\web_protocol.py", line 385, in start
    resp = await self._request_handler(request)
  File "D:\Program Files\Python3\lib\site-packages\aiohttp\web_app.py", line 338, in _handle
    resp = await handler(request)
TypeError: root() missing 1 required positional argument: 'response'
```

## Steps to reproduce
Create and run this script
```py
from aiohttp import web
import asyncio

routes = web.RouteTableDef()
app = web.Application()

class Example:
    @routes.get("/")
    async def root(self, response):
        return web.Response(text="5")

app.add_routes(routes)
web.run_app(app)
```

## Your environment
Latest version of aiohttp (3.2.1)
Windows 10

modelmat:
## What do these changes do?

Allows adding of class methods to route tables.

The change from RouteTableDef to ClassRouteTableDef and the addition of ClassRouteTableDef.add_class for adding them. No breaking changes.

## Related issue number

Fixes #2996 

## Checklist


kxepal:
Thanks for contribution, but why not to use [class based views](https://aiohttp.readthedocs.io/en/stable/web_quickstart.html#class-based-views) instead to achieve the same goal?

modelmat:
Because it achieves a different purpose. Whilst the initial reason I wanted classes has a better way of doing it, this way allows the possibility of:
```py
routes = web.ClassRouteTableDef()

class Routes:
	@routes.get("/home")
	async def home_handler(self, request):
		...
	@routes.get("/)
	async def root(self, request):
		...
	...
routes.add_class_routes(Routes())
```
~~(please don't merge this yet, I want to change one or two things)~~

asvetlov (changes requested):
Please rewrite tests in a functional style: create a test app using class route defs and make sure that registered handler is called correctly.
Something like
```
async def test_simple_get(aiohttp_client) -> None:

    async def handler(request):
        body = await request.read()
        assert b'' == body
        return web.Response(body=b'OK')

    app = web.Application()
    app.router.add_get('/', handler)

    resp = await client.get('/')
    assert 200 == resp.status
    txt = await resp.text()
    assert 'OK' == txt
```
but with `ClassRouteTableDef()` usage

kxepal:
> Because it achieves a different purpose. Whilst the initial reason I wanted classes has a better way of doing it, this way allows the possibility of...

But what's the case? You can define the same routes in the separate module to get namespace feature with the similar profit.

aiohttp/web_routedef.py
```diff
@@ -191,3 +192,28 @@ def view(self, path: str, **kwargs: Any) -> _Deco:
...
+            return handler
+        return inner
+
+    def add_class_routes(self, instance: Any) -> None:
+        def predicate(member: Any) -> bool:
+            return all((inspect.iscoroutinefunction(member),
+                        hasattr(member, "route_info")))
+        for _, handler in inspect.getmembers(instance, predicate):
+            method, path, kwargs = handler.route_info
+            super().route(method, path, **kwargs)(handler)
```
asvetlov, line 220: The code is cumbersome.
You override `route()` method a few lines above but calls a parent version here.
It confused me on code review, I believe any next code reader will stumble on this line as well.
Please find a way to avoid such magic.

kxepal:
I see this feature introduces some implicit global state between views via self-context. So your application actually will have Application context state, Request context state and group-of-views context state. The latter so far looks like an attempt to make debugging fun again.

Also this feature encourages to mix views with the logics within the same class, making things more complicated. MTV / MVC / MVVC patterns were invented just because of that.

asvetlov:
@kxepal I've seen such code organization many times, and it is different from both first-class-functions and class-based-views.
The proposed routedef makes sense for me, but implementation needs a polishing.

P.S.
Consider `Routes` class from given example as not a dynamic thing but a static part of web server configuration.
People use it to avoid storing things like database object into an application (`app['db'] = await create_db()`). 

```
# mod_a.py

routes = web.ClassRouteTableDef()

class Routes:
        def __init__(self, db):
            self.db = db

	@routes.get("/home")
	async def home_handler(self, request):
		...
	@routes.get("/)
	async def root(self, request):
		...
	...

# main.py

async def init_app():
    config = await read_config()
    db = await create_db(config)
    routes.add_class_routes(Routes())
    app = web.Application()
    app.add_routes(routes)
    return app

web.run_app(init_app())
```

Hmm. Still looks overcomplicated.

modelmat:
I just typed this all out but accidentally closed the tab. Whoops.

@asvetlov That test makes no sense, as I all changed is how the routes are added to the internal cache - none of the code relates to the actual route functionality. **edit**: Unless you are meaning to ensure that we don't get errors as seen in #2996?

@kxepal It can be used as a way to avoid global variables. Whilst a module is almost a class, you would have to rely on global variables found in the module (requiring an import if you wished to change). In classes you could have a method or just change an attribute quite easily, all you would need to do is instantiate the class before passing it to the `add_class_routes` function.

The only other alternatives I can think of rather than adding `route_info` to the handlers are:
1. Create a wrapper class which implements `__call__` and can store all the information previously in `route_info`. This can then be searched by instance of the wrapper class by `add_class_routes`.
2. Search for all methods which have the parameters `(self, request)`. Very finicky and will break if people don't name it `self` or `request`.

Also, rather than using `add_class_routes` we could of:
1. Instead, make `app.add_routes` do `isinstance` on the `routes` argument. If:
	a. `routes` is an instance of `RouteTableDef`, then continue with old code 
    b. Otherwise, (beforehand decorated the class with a @classtabledef or something decorator or maybe through inheritance that adds an attribute which signals that it contains routes) generate all the routes for the routed methods and add the routes to the application.
	c. Raise an error.
2. Have a class decorator like `@class_route_table` which takes arguments to initialise the class and then initialises the class and adds the routes to the `RouteTableDef`. Example:
```py
@class_route_table(db)
class Routes:
	def __init__(self, db):
		...
	...
```

aiohttp/web_routedef.py
modelmat: I really don't know how I could avoid it without having to duplicate code (creating problems) or introducing a new name for the subclass's @route decorator, which would be really confusing for the user.

modelmat:
Has anyone had any thoughts on this?

kxepal:
Honestly, I don't like mixing requests processing with business logic within a single class to share anything between them. This is dead end design which is hard to test, maintain, extend, replace one parts by the others. In the end, you'll decouple things putting them into different modules, so why not do that from the very start?

vindard:
Just as a sidenote, I ran into this exact issue (decorating a class method without invalidating its `self` argument) and was able to solve it using a combination of the code you've written and the different solutions [you've described here](https://github.com/aio-libs/aiohttp/pull/3585#issuecomment-458739874) @Modelmat (thanks for that!)

Is anyone else still reviewing this as a potential feature to be merged into the project still?

aiohttp/web_routedef.py
```diff
@@ -192,3 +193,28 @@ def view(self, path: str, **kwargs: Any) -> _Deco:
...
+            # Is Any rather than _HandlerType because of mypy limitations, see
+            # https://github.com/python/mypy/issues/3882 and
+            # https://github.com/python/mypy/issues/2087
+            handler.route_info = (method, path, kwargs)
+            return handler
+        return inner
+
+    def add_class_routes(self, instance: Any) -> None:
+        def predicate(member: Any) -> bool:
+            return all((inspect.iscoroutinefunction(member),
```
vindard, line 216:
```suggestion
            return all((inspect.ismethod(member),
                        hasattr(member, "route_info")))
        for _, handler in inspect.getmembers(instance, predicate):
            assert inspect.iscoroutinefunction(handler), \
                "Please ensure your decorated handler methods are all \"async\""
            method, path, kwargs = handler.route_info
            super().route(method, path, **kwargs)(handler)
```

Not sure if this will be helpful here but I refactored this a little in some personal code I have to catch cases where someone accidentally decorates a non-async function within a class.

PR #3596, opened 2019-02-06
Fix type hint for tracing Signal sequence content

stj:
The Signal is instantiated with a TraceConfig object but does not store
it in its mutable sequence. Fixed hinting to reflect what is expected to
be added as signal handler.

## What do these changes do?

Fix type hint for tracing Signal contents. They got very long with the last argument being a `Union`, maybe an `Any` would do instead?

mypy does not report an error of
```error: List item 0 has incompatible type "Callable[[ClientSession, SimpleNamespace, TraceConnectionCreateEndParams], Coroutine[Any, Any, None]]"; expected "Callable[[TraceConfig], Awaitable[None]]"```

## Related issue number

#3595

## Checklist

### CODE REVIEW

**Comment by kxepal** [comment]
Suddenly, this patch doesn't fixes the issue. I run it against your snippet and still get an error:
```
(venv) ~/projects/aio-libs/aiohttp (master) $ cat bug-3595.py 
import types

import aiohttp
import datadog

async def reuse_keepalive_connection(
    session: aiohttp.ClientSession,
    trace_config_ctx: types.SimpleNamespace,
    params: aiohttp.TraceConnectionReuseconnParams,
) -> None:
    """statsd counter; how often has a connection been reused"""
    datadog.statsd.increment("client.request.connection.reuse")

tracer = aiohttp.TraceConfig()
tracer.on_connection_reuseconn.extend([reuse_keepalive_connection])

(venv) ~/projects/aio-libs/aiohttp (master) $ mypy --ignore-missing-imports bug-3595.py 
bug-3595.py:15: error: List item 0 has incompatible type "Callable[[ClientSession, SimpleNamespace, TraceConnectionReuseconnParams], Coroutine[Any, Any, None]]"; expected "Callable[[ClientSession, SimpleNamespace, <union: 15 items>], None]"
```

stj:
I think I see the problem. Looks like I changed the return type of the `Callable` from `Awaitable[None]` to `None`. Will change an test that.

stj:
That's odd. Was sure that this passed for me but I see that it fails now again. I made it really ugly now but it passes. Type hints are still a mystery to me.

asvetlov:
thanks!

PR #3613, opened 2019-02-17
Use sanitized URL as Location header in redirects

m-burst:
## What do these changes do?

When performing HTTP redirects (e.g. using `raise HTTPFound(location)`), the URL in the Location header is now passed through `URL` constructor and thus sanitized. Most importantly, any CRLF in the location will be escaped, which will prevent [HTTP Response Splitting attacks](https://www.owasp.org/index.php/HTTP_Response_Splitting).

None unless the user's code is vulnerable.

## Related issue number

None

## Checklist

### DIFF
--- a/CHANGES/3613.bugfix
+++ b/CHANGES/3613.bugfix
@@ -1,0 +1,1 @@
+Use sanitized URL as Location header in redirects
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -152,6 +152,7 @@ Mathias Fröjdman
 Matthieu Hauglustaine
 Matthieu Rigal
 Michael Ihnatenko
+Mikhail Burshteyn
 Mikhail Kashkin
 Mikhail Lukyanchenko
 Mikhail Nacharov
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -203,8 +203,8 @@ class HTTPMove(HTTPRedirection):
             raise ValueError("HTTP redirects need a location to redirect to.")
         super().__init__(headers=headers, reason=reason,
                          text=text, content_type=content_type)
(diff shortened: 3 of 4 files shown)


m-burst:
More details about the issue:

Given the following sample aiohttp app:

```python
from aiohttp import web


def test_view(request):
    param = request.query.get('param')
    raise web.HTTPFound(f'http://example.com?param={param}')


def create_app():
    app = web.Application()
    app.router.add_get('/', test_view)
    return app


if __name__ == '__main__':
    web.run_app(create_app(), port=5000)
```


Sending a request with a urlencoded CRLF in the query parameter results in the following:
```
$ curl -v localhost:5000/?param=%0d%0aContent-Length:3%0d%0a%0d%0axxx
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 5000 (#0)
> GET /?param=%0d%0aContent-Length:3%0d%0a%0d%0axxx HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.47.0
> Accept: */*
>
< HTTP/1.1 302 Found
< Content-Type: text/plain; charset=utf-8
< Location: http://example.com?param=
< Content-Length:3
<
* Excess found in a non pipelined read: excess = 95, size = 3, maxdownload = 3, bytecount = 0
* Connection #0 to host localhost left intact
xxx
```

Though the user's code itself needs to be vulnerable in order to exploit this bug, the case is pretty common and aiohttp should protect the users from this (as most other web frameworks do).

This PR fixes the issue by essentially setting the Location header to `str(URL(location))` rather than plain location.

webknjaz:
@m-burst thanks! Would you mind sending a backport against the stable branch?

m-burst:
@webknjaz Opened the backport PR in #3614.

PR #3614, opened 2019-02-17
[3.5] Use sanitized URL as Location header in redirects

m-burst:
Backport of #3613 for the 3.5 branch

### DIFF
--- a/CHANGES/3614.bugfix
+++ b/CHANGES/3614.bugfix
@@ -1,0 +1,1 @@
+Use sanitized URL as Location header in redirects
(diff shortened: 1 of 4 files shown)


m-burst:
@webknjaz Looking at it again, I think this might be seen as breaking for some of the users who expect the location property on an HTTPFound object to be a string if they passed a string (and also I broke a test and made the mistake of not running them locally before opening the PR). Should I fix this in a follow-up PR?

webknjaz:
Odd... CI was green. I'm not sure what you mean but PR is welcome and we can go through a review process there :)

m-burst:
@webknjaz CI probably didn't fully run because the PR was merged very quickly.
Please see the fix in #3615.

PR #3615, opened 2019-02-17
Fix backport of redirect URL fix to 3.5

m-burst:
Fix issues after #3614 (backport of #3613 to 3.5).

`_HTTPMove.location` is again the `location` parameter unchanged, no longer unconditionally converted to `URL` (which broke a test and might be seen as breaking by the users).


webknjaz:
Thanks! Let's merge this once CI is green.

m-burst:
@webknjaz It looks like the CI config on 3.5 branch does not pin `pip` version to `<19.0` so the builds fail :(

m-burst:
@webknjaz OK, removed the commit pinning `pip`.
CI was green with it though, which means that the code changes are probably good to merge.

PR #3617, opened 2019-02-18
Other method for Based Class View

Cheban1996:
For me work this method app.router.add_route('*', '/path/to', MyView). 
This method  web.view('/path/to', MyView) i checked very many times but he not worked


asvetlov:
thanks

Issue #3629, opened 2019-02-27
Allow creating and connecting to named pipes in windows


hackrush01:
## Long story short

We need to use named pipes in windows in our project([lbry](https://github.com/lbryio/lbry))

## Expected behaviour

aiohttp should provide the functionality of creating and connecting to named pipes in windows

## Actual behaviour

No such functionality present

## Additional Comments

I have already "almost" implemented it. I have opened this issue for discussion regarding the feasibility and potential issues.

asvetlov:
Would you work on pull request?

hackrush01:
Yes, I should have one by the night.

asvetlov:
Fixed by #3632

PR #3632, opened 2019-02-28
Add support for Named Pipes to Site and Connector under Windows

hackrush01:
## What do these changes do?

Adds support for named pipes in windows

No

## Related issue number

#3629 

## Checklist


hackrush01:
Thanks.
As per your points, that's exactly what I'm working on right now. :-)

aiohttp/pytest_plugin.py
```diff
@@ -302,7 +303,7 @@ def aiohttp_client(loop):  # type: ignore
     async def go(__param, *args, server_kwargs=None, **kwargs):  # type: ignore
 
         if (isinstance(__param, Callable) and  # type: ignore
-                not isinstance(__param, (Application, BaseTestServer))):
```
hackrush01: Mistake, fixing....

hackrush01:
Windows proactor event policy was added only in python 3.7
https://bugs.python.org/issue33792
See last point [here](https://docs.python.org/3/whatsnew/3.7.html#asyncio)
Finding a workaround

hackrush01:
I don't understand what's up with code cov bot
Is it me?

hackrush01:
Now it works

aiohttp/web_runner.py
```diff
@@ -14,9 +14,8 @@
 except ImportError:
     SSLContext = object  # type: ignore
 
-
-__all__ = ('BaseSite', 'TCPSite', 'UnixSite', 'SockSite', 'BaseRunner',
-           'AppRunner', 'ServerRunner', 'GracefulExit')
+__all__ = ('BaseSite', 'TCPSite', 'UnixSite', 'NamedPipeSite', 'SockSite',
```
webknjaz, line 18: Return that empty line.

aiohttp/web_runner.py
```diff
@@ -58,7 +57,9 @@ def name(self) -> str:
             self._runner._unreg_site(self)
             return  # not started yet
         self._server.close()
-        await self._server.wait_closed()
+        # named pipes do not have wait_closed property
```
webknjaz, line 61: So why don't you create one?

aiohttp/web_runner.py
```diff
@@ -128,6 +129,29 @@ def name(self) -> str:
...
+    @property
+    def name(self) -> str:
+        return self._path
+
+    async def start(self) -> None:
+        await super().start()
+        loop = asyncio.get_event_loop()
+        server = self._runner.server
+        assert server is not None
+        _server = await loop.start_serving_pipe(  # type: ignore
```
webknjaz, line 154: Why ignore the type?

aiohttp/connector.py
```diff
@@ -1126,3 +1127,53 @@ def path(self) -> str:
...
+        """Path to the named pipe."""
+        return self._path
+
+    async def _create_connection(self, req: 'ClientRequest',
+                                 traces: List['Trace'],
+                                 timeout: 'ClientTimeout') -> ResponseHandler:
+        try:
+            with CeilTimeout(timeout.sock_connect):
+                _, proto = \
+                    await self._loop.create_pipe_connection(  # type: ignore
```
webknjaz: Why ignore type?

aiohttp/connector.py
```diff
@@ -1126,3 +1127,53 @@ def path(self) -> str:
...
+    def path(self) -> str:
+        """Path to the named pipe."""
+        return self._path
+
+    async def _create_connection(self, req: 'ClientRequest',
+                                 traces: List['Trace'],
+                                 timeout: 'ClientTimeout') -> ResponseHandler:
+        try:
+            with CeilTimeout(timeout.sock_connect):
+                _, proto = \
```
webknjaz: Please, use brackets instead of EOL escaping
```suggestion
                _, proto = (
```

tests/test_connector.py
```diff
@@ -1983,6 +2000,35 @@ def check_with_exc(err):
...
+        await connector.connect(req, None, ClientTimeout())
+
+
+@pytest.mark.skipif(platform.system() != "Windows",
+                    reason="Proactor Event loop present only in Windows")
+async def test_named_pipe_connector_permission(proactor_loop) -> None:
+    proactor_loop.create_pipe_connection = make_mocked_coro(
+        raise_exception=PermissionError()
+    )
+    path = r'\\.\pipe\{}'.format(uuid.uuid4().hex)
```
webknjaz: provide this via a fixture

CHANGES/3629.feature
```diff
@@ -0,0 +1 @@
+Add support for Named Pipes(Site and Connector) on windows.
```
webknjaz:
```suggestion
Add support for Named Pipes (Site and Connector) under Windows. This feature requires Proactor event loop to work.
```

aiohttp/connector.py
```diff
@@ -1126,3 +1127,53 @@ def path(self) -> str:
...
+        after each request (and between redirects).
+    limit - The total number of simultaneous connections.
+    limit_per_host - Number of simultaneous connections to one host.
+    loop - Optional event loop.
+    """
+
+    def __init__(self, path: str, force_close: bool=False,
+                 keepalive_timeout: Union[object, float, None]=sentinel,
+                 limit: int=100, limit_per_host: int=0,
+                 loop: Optional[asyncio.AbstractEventLoop]=None) -> None:
```
webknjaz, line 1165:
```suggestion
                 loop: Optional[asyncio.ProactorEventLoop]=None) -> None:
```

aiohttp/connector.py
```diff
@@ -1126,3 +1127,53 @@ def path(self) -> str:
...
+    path - Windows named pipe path.
+    keepalive_timeout - (optional) Keep-alive timeout.
+    force_close - Set to True to force close and do reconnect
+        after each request (and between redirects).
+    limit - The total number of simultaneous connections.
+    limit_per_host - Number of simultaneous connections to one host.
+    loop - Optional event loop.
+    """
+
+    def __init__(self, path: str, force_close: bool=False,
```
webknjaz, line 1162: You should validate that it runs under `ProactorEventLoop` somewhere at the beginning of this method, I think.

aiohttp/pytest_plugin.py
```diff
@@ -216,6 +217,21 @@ def loop(loop_factory, fast, loop_debug):  # type: ignore
...
+@pytest.fixture
+def proactor_loop():  # type: ignore
+    if not PY_37:
+        policy = asyncio.get_event_loop_policy()
+        policy._loop_factory = asyncio.ProactorEventLoop  # type: ignore
+    else:
+        policy = asyncio.WindowsProactorEventLoopPolicy()  # type: ignore
+        asyncio.set_event_loop_policy(policy)
+
+    loop_context(policy.new_event_loop)
```
webknjaz: Why do you invoke this twice?

hackrush01: Sorry, it was by mistake.
Fixed

aiohttp/web_runner.py
hackrush01: `start_serving_pipe` is not defined in AbstractEventLoop. So it throws the following
```
aiohttp\connector.py:1149: error: Invalid type "asyncio.ProactorEventLoop"
aiohttp\connector.py:1166: error: "AbstractEventLoop" has no attribute "create_pipe_connection"; ma
ybe "create_connection" or "create_unix_connection"?
```
Is there some other way I'm missing?

aiohttp/connector.py
hackrush01: Same reason as above

aiohttp/connector.py
hackrush01: Changing this causes(while running `mypy`)
```
aiohttp\connector.py:1149: error: Invalid type "asyncio.ProactorEventLoop"
```

aiohttp/web_runner.py
hackrush01: Named pipes in asyncio are from the class: https://github.com/python/cpython/blob/master/Lib/asyncio/windows_events.py#L241

Which comprises of only normal sync methods. `.close()` is a sync method as well. So therefore it has a `.closed()` property. Are you suggesting to add `.wait_closed()` property there for consistency(hence a PR to cython)?
Or is it something else?

aiohttp/web_runner.py
webknjaz: Dunno. Maybe @asvetlov knows a better way

aiohttp/pytest_plugin.py
webknjaz: Did you forget to push?

aiohttp/connector.py
webknjaz: Odd... are you running mypy under Python 3.7?

webknjaz: Maybe we could add a conditional for this to work under Linux?

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    try:
        from asyncio import ProactorEventLoop
    except ImportError:
        from asyncio import AbstractEventLoop
        class ProactorEventLoop(ProactorEventLoop):
            """ProactorEventLoop stub for mypy."""
```

aiohttp/web_runner.py
webknjaz: Yeah, I'd just inherit that and return immediately maybe. Let's see what @asvetlov thinks.

aiohttp/pytest_plugin.py
hackrush01: I had to run to class. So I had to defer pushing to later, just made the changes locally, and commented on the reviews. Pushing now :-D

aiohttp/connector.py
hackrush01: I am infact running mypy under python 3.7.
Also I too tried with 
```
if TYPE_CHECKING:
    from asyncio import ProactorEventLoop
```
Still the same error.
I'll try again, maybe I did something wrong in rush

aiohttp/connector.py
hackrush01: I thought about it, but creating a NamedPipeConnector, raises
`AttributeError: '_WindowsSelectorEventLoop' object has no attribute 'create_pipe_connection'`
Wouldn't that be okay?
But if there is no assertion then the error wouldn't be raised when `loop.create_pipe_connection()` is mocked out. So maybe there is a need for the assertion too.
I'd be fine with whatever seems okay to you.

hackrush01: Also, if we end up adding the assert in this case, we'd be adding it here as well, right?
https://github.com/aio-libs/aiohttp/pull/3632/files#diff-dbf24c78287ef0b964d4945f339a43a9R135

aiohttp/connector.py
hackrush01: Nope, still happening. It is odd

CHANGES/3629.feature
```diff
@@ -1 +1 @@
-Add support for Named Pipes(Site and Connector) on windows.
+Add support for Named Pipes (Site and Connector) under Windows. This feature requires Proactor event loop to work.
```
webknjaz, line 1: Do you know you can just hit "accept patch suggestion" on GitHub UI and not have to copy-paste everything manually?

aiohttp/connector.py
webknjaz: It's because you're trying to make the import under Linux and docs say it's available only under windows

aiohttp/web_runner.py
```diff
@@ -128,6 +130,31 @@ def name(self) -> str:
             ssl=self._ssl_context, backlog=self._backlog)
 
 
+class NamedPipeSite(BaseSite):
+    __slots__ = ('_path', )
+
+    def __init__(self, runner: 'BaseRunner', path: str, *,
+                 shutdown_timeout: float=60.0) -> None:
+        loop = asyncio.get_event_loop()
+        assert isinstance(loop, asyncio.ProactorEventLoop)
```
webknjaz: if-block + raising a `RuntimeError` is probably more appropriate here. `assert`s are ripped off in the optimized mode and we still want to communicate this error to users in a nice way...

CHANGES/3629.feature
hackrush01: No I didn't!
Thanks for telling me 😁

aiohttp/connector.py
hackrush01: It fails on my computer too, where I'm running it on windows 🤔

aiohttp/web_runner.py
hackrush01: Okay.
Optimized mode??

hackrush01:
I removed a test in `web_runner.py` which literally was just testing that the variable was correctly assigned.
Also, everywhere I used `asyncio.ProactorEventLoop` I had to put `# type: ignore` because mypy complains

aiohttp/connector.py
```diff
@@ -1126,3 +1127,55 @@ def path(self) -> str:
             raise ClientConnectorError(req.connection_key, exc) from exc
 
         return cast(ResponseHandler, proto)
+
+
+class NamedPipeConnector(BaseConnector):
+    """Named pipe connector.
```
webknjaz, line 1148: According to PEP257, there must be one empty line after a title line.
```suggestion
    """Named pipe connector.

```

aiohttp/connector.py
webknjaz: attribute access is not static analysis by MyPy

hackrush01:
Oy Vey! No way to cancel AppVeyor and Travis Builds...
Sorry for so many commits...
In my defense... Windows is hard :-P

hackrush01:
@webknjaz Just checking to be sure we are on the same page; The PR is awaiting comments/review from @asvetlov, right? Because I think I have done all the other non-ambiguous stuff 🤔 .

eukreign:
@asvetlov @webknjaz  Is this PR still being considered? Really looking forward to this feature!

jackrobison:
@asvetlov I'm really looking forward to using this - what's holding up merging it?

asvetlov:
I need time to review the PR carefully.
Sorry, pretty busy now, need to add several features to asyncio before Python 3.8 feature freeze date.

Do you ask for merging only or for publishing a new aiohttp release with windows named pipes functionality implemented?
The first is easy, the second requires more work to be done.

eukreign:
Ultimately, releasing a new aiohttp with this feature is the most useful thing. Although just merging it would also help in terms of knowing that this is how it will work / what the API will be like, etc, even if we can't really use it yet (until it's on pypi), we can at least start programming against a future release of it.

Would you say that at least the overall API / architecture of this PR is how it will be in the final version?

asvetlov:
I'm on the trip now. Will review and merge the PR in a few days

eukreign:
Awesome, thanks! Have a safe trip.

asvetlov:
Merged manually

asvetlov:
Thanks, @eukreign

Issue #3631, opened 2019-02-28
Chunked stream  memory leak


mvanderkroon:
## Long story short

I'm trying to (POST) stream data to an aiohttp server instance (potentially hundreds of gigabytes), however my data is not typically stored in files and can be 'generated' in client processes on the fly. I can't use a multipart upload as it does not fit my needs, nor do I use the `data = await request.post()` shorthand (which the docs are clear about in that that *will* OOM for large files).

I'm trying to use the underlying `StreamReader` (`request._payload`) to allow line by line iteration over the stream. In doing so, aiohttp (server) consumes more and more memory until the application OOMs.

## Expected behaviour

Processing a stream of data in aiohttp server should not cause OOMs

## Actual behaviour

aiohttp OOMs on large streams of data

## Steps to reproduce

aiohttp server
```python
# server.py
async def process_stream(request):
    async for line in request._payload:
        pass

    return web.json_response({"STATUS": "OK"})
```

requests client
```python
# client.py
def generate_data():
    while True:
        yield """hello world\n""".encode('utf-8')
        
r = requests.post("http://localhost:8080/", data=generate_data())
```

## Additional info
I found a [resource](https://vorpus.org/blog/some-thoughts-on-asynchronous-api-design-in-a-post-asyncawait-world/#bug-1-backpressure) relating to asyncio and `StreamReader/Writer`-backpressure. I have done my best to read through the aiohttp source but it looks like the fixes mentioned in the document are already in place so I'm not sure why this is not working.

In fact, I'm not sure whether the memory increase is due to aiohttp (or an underlying lib) holding references to elements in memory, or whether the producing process is simply pushing data in to the queue faster than aiohttp is consuming it (this latter case would suggest a problem with backpressure).

## Your environment
server
aiohttp 3.5.4
alpine 3.7.0

mvanderkroon:
Found what looks like a bonafide bug. Analysis using `tracemalloc` showed that memory was leaking in [streams:238](https://github.com/aio-libs/aiohttp/blob/v3.5.4/aiohttp/streams.py#L238). 

Confusing at first, closer inspection reveals that on [streams:276](https://github.com/aio-libs/aiohttp/blob/v3.5.4/aiohttp/streams.py#L276) this `self.total_bytes` value is appended to a list (`_http_chunk_splits`) of infinitely increasing size: hence the memory leak.

[...]

PR #3634, opened 2019-03-03
Fix memory leak

mvanderkroon:
## What do these changes do?

Fixes issue with `StreamReader.readline` infinitely growing list `self._http_chunk_splits` when iterating over streamed POST data line by line.

None, apart from aiohttp server no longer OOMing on large POST uploads.

## Related issue number

#3631

## Checklist

- [v] I think the code is well written
No, the `test_streams.py` (which looks like the obvious place to test this) always instantiate a `StreamReader` from scratch and perform tests on this class in isolation. I can confirm that the bug that this PR fixes does not occur in isolation and so I was not able to adjust the tests to verify the corrected behavior. In other words, the memory leak is the result of interactions with the `StreamReader` class with other aiohttp internals. Writing proper tests for this require a level of knowledge of aiohttp that is above my head. Tests also do not pass, but the failing tests already failed prior to the changes in this PR.
I don't believe this is applicable
- [v] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
- [v] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -328,6 +328,9 @@ class StreamReader(AsyncStreamReaderMixin):
             if not_enough:
                 await self._wait('readline')
 
+        # fixes memory leak: https://github.com/aio-libs/aiohttp/issues/3631
+        self._http_chunk_splits = []
+
         return b''.join(line)
 
     async def read(self, n: int=-1) -> bytes:
--- a/CHANGES/3631.bugfix
+++ b/CHANGES/3631.bugfix
@@ -1,0 +1,1 @@
+Fixes issue with `StreamReader.readline` infinitely growing list `self._http_chunk_splits` when iterating over streamed POST data line by line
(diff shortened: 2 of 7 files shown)


aiohttp/streams.py
```diff
@@ -328,6 +328,9 @@ def end_http_chunk_receiving(self) -> None:
             if not_enough:
                 await self._wait('readline')
 
+        # fixes memory leak: https://github.com/aio-libs/aiohttp/issues/3631
+        self._http_chunk_splits = []
```
socketpair, line 332: @asvetlov  let me deal with this bug

socketpair: https://github.com/aio-libs/aiohttp/pull/3635 it is here. @mvanderkroon good catch! Thanks for sample code. Sorry that I forced you to create PR. Anyway it helped a lot to understand what is the source of the problem.

mvanderkroon: @socketpair, happy to help!

Issue #3628, opened 2019-02-27
Session not closed in aiohttp.request after exception


fried-sausage:
## Long story short

subj
First I get ClientConnectorError (connection-related errors are expected in my application), after that i see `Unclosed client session` warning.

## Expected behaviour

ClientSession, created inside aiohttp.request, is closed after programm leaves `async with` block.

## Actual behaviour

ClientSession is not closed.

## Steps to reproduce

```
import asyncio
import aiohttp


async def main():
    async with aiohttp.request("GET", "http://127.0.0.1:8080") as resp:
        pass


loop = asyncio.get_event_loop()
try:
    loop.run_until_complete(main())
finally:
    loop.close()
```
## Your environment
aiohttp 3.5.4
python 3.5
debian stretch

## Why use aiohttp.request at all, opening new connection for every request is bad
I need to open new connection anyway:
1) server i talk to does not support keep-alive
2) i programm client that runs on a moving device using wifi network so connection failures are expected

asvetlov:
Fixed by #3640

PR #3640, opened 2019-03-06
Close session on error in aiohttp.request (#3628)

fried-sausage:
## What do these changes do?

`aiohttp.request` implicitly creates `ClientSession` object and passes it to context manager, which is then returned to the caller. If exception occurs inside `__aenter__` of this context manager, session won't be closed.
I consider this PR to be a bugfix.

## Related issue number

#3628 

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1020,8 +1020,12 @@ class _SessionRequestContextManager:
         self._session = session
 
     async def __aenter__(self) -> ClientResponse:
-        async with self._session:
+        try:
             self._resp = await self._coro
+        except BaseException:
+            await self._session.close()
+            raise
+        else:
             return self._resp
 
     async def __aexit__(self,
@@ -1030,6 +1034,7 @@ class _SessionRequestContextManager:
                         tb: Optional[TracebackType]) -> None:
         assert self._resp is not None
         self._resp.close()
+        await self._session.close()
 
 
 def request(
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1020,8 +1020,9 @@ class _SessionRequestContextManager:
         self._session = session
 
     async def __aenter__(self) -> ClientResponse:
-        self._resp = await self._coro
-        return self._resp
+        async with self._session:
+            self._resp = await self._coro
+            return self._resp
 
     async def __aexit__(self,
                         exc_type: Optional[Type[BaseException]],
@@ -1029,7 +1030,6 @@ class _SessionRequestContextManager:
                         tb: Optional[TracebackType]) -> None:
         assert self._resp is not None
         self._resp.close()
-        await self._session.close()
 
 
 def request(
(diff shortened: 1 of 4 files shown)


webknjaz:
@fried-sausage please add yourself to contributors

fried-sausage:
@webknjaz done. Just to make sure you are notified. I'm still not sure this is the best possible solution (maybe reviewers could suggest some alternatives), but it seems to work.

aiohttp/client.py
```diff
@@ -1020,8 +1020,13 @@ def __init__(self,
         self._session = session
 
     async def __aenter__(self) -> ClientResponse:
-        self._resp = await self._coro
-        return self._resp
+        try:
+            self._resp = await self._coro
+        except BaseException:
```
webknjaz, line 1025: I think that this is too broad. Best practice is to catch a specific exception class.

kxepal: It's going to be reraised shortly and used just to cleanup resources. It's legal usage for such a case. Actually, we have similar catches in other places, but reraise happens always shortly. Certainly, if there possible to limit the scope by some OSError - that would be great, but this may keep the bug for future changes when other exception types will be raised as well.

fried-sausage: Don't we also need to catch KeyboardInterrupt (or maybe even GeneratorExit) here?

fried-sausage: If yes, that would leave us with 
```
try:
    ...
except (*bunch_of_base_exceptions, *bunch_of_os_exceptions)
    ...
```
It might be nice, as mentioned above, if we could narrow down all exceptions that we expect in `session._request` to just one class (so `*bunch_of_os_exceptions` is actually just one exception).
But there is another problem: i believe we can't  make everything  foolproof and there will always be the case when user supplies some invalid arguments, which can result in unforeseen exception (TypeError, ValueError, who knows) inside aiohttp. I think in such case it is more reasonable to cleanup resources "no matter what" and re-raise in places like this.

kxepal: > Don't we also need to catch KeyboardInterrupt (or maybe even GeneratorExit) here?

It doesn't matter since you do `raise` a little bit later.

fried-sausage: I mean in case we do it the way @webknjaz suggested (replace BaseException with some  specific class). One class won't be enough, we will need to catch at least KeyboardInterrupt and may be several other classes.

fried-sausage: Going to resolve this. If anyone has further arguments against BaseException please bring them on.

kxepal: Agreed.

kxepal:
@webknjaz 
Do you have any objections? Otherwise could you please remove your block on merge? Don't want to apply sudo without a reason (:

webknjaz:
@kxepal I still don't like wholesale excepts but seems legit

asvetlov:
Thanks!

PR #3633, opened 2019-03-03
Use correct type for add_view

kornicameister:
## What do these changes do?

They are changing the type declaration for `add_view` in router. Previously it said that it was to be `AbstractView` but that required an instance of the `AbstractView` subclass. The documentation stands for it to be the class, hence changing the typing to `Type[AbstractView]`.

## Are there changes in behavior for the user?

No more mypy errors when using `add_view` method/function.

## Related issue number

Fixes #3653

## Checklist

- [x] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #3655, opened 2019-03-18
fix handler argument annotations

### LINKED ISSUE #3653: web.view should accept Type[AbstractView] as handler instead of AbstractView
## Long story short

according to [aiohttp's documents](https://docs.aiohttp.org/en/stable/web_quickstart.html#class-based-views), `web.view(path, handler)` hanlder should be a `class` inherit from `web.View`

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_routedef.py#L123 
```python
def view(path: str, handler: AbstractView, **kwargs: Any) -> RouteDef:
    return route(hdrs.METH_ANY, path, handler, **kwargs)
```
but type annotation means that `handler` should be a instance of `AbstractView` instead of a class.

## Expected behaviour

`web.view`'s handler annotation should be `Type[AbstractView]`

## Actual behaviour

`web.view`'s handler annotation should be `AbstractView`

## Steps to reproduce

no

## Your environment

server: aiohttp==3.5.4

Hanaasagi:
This pull request wants to make `handler` annotation correct.

When using Mypy to check following code,

```Python
from aiohttp import web


class MyView(web.View):
    async def get(self):
        return web.Response(text="Hello world")


app = web.Application()
app.add_routes([web.view('/', MyView)])  # line 10
web.run_app(app)
```

It will report

```
t.py:10: error: Argument 2 to "view" has incompatible type "Type[MyView]"; expected "AbstractView"                                                                                           
```

`web.view` accepts subclass of `AbstractView` type variable as `handler`, instead of an instance.

resolve #3653

### DIFF
--- a/aiohttp/web_routedef.py
+++ b/aiohttp/web_routedef.py
@@ -10,6 +10,7 @@ from typing import (
     List,
     Optional,
     Sequence,
+    Type,
     Union,
     overload,
 )
@@ -40,7 +41,7 @@ class AbstractRouteDef(abc.ABC):
 
 
 _SimpleHandler = Callable[[Request], Awaitable[StreamResponse]]
-_HandlerType = Union[AbstractView, _SimpleHandler]
+_HandlerType = Union[Type[AbstractView], _SimpleHandler]
 
 
 @attr.s(frozen=True, repr=False, slots=True)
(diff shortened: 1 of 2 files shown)


Hanaasagi:
It seems like a small change, I will add changelog if you really need.

kornicameister:
@Hanaasagi check out #3633 . Seems you've duplicated my work ;-)

Hanaasagi:
Oh, I'm Sorry for this.

kornicameister:
@Hanaasagi no biggie ;-).

PR #3657, opened 2019-03-19
fix local_addr annotation in TCPConnector.__init__

### LINKED ISSUE #3639: Wrong type annotation for local_addr in TCPConnector.__init__
https://github.com/aio-libs/aiohttp/blob/f489a927a84bf37c6c49e0c31f9228adc31d6058/aiohttp/connector.py#L717
Should not it be `Optional[Tuple[str, int]]`?

Hanaasagi:
From the [doc](https://docs.aiohttp.org/en/stable/client_reference.html#tcpconnector)

- local_addr (tuple) – tuple of `(local_host, local_port)` used to bind socket locally if specified.

resolve #3639


asvetlov:
Resolved by #3732

@Hanaasagi thank for the PR.
Sorry, I've merged the very similar code just before looking on your implementation

PR #3660, opened 2019-03-22
Suppress pytest warnings due to test util classes

cxong:
## What do these changes do?

We are using pytest; by default it collects tests in:
- files that are named `test_*.py`
- classes that are named `Test*`

Unfortunately this matches the `TestServer`/`TestClient` classes in aiohttp. Since they have `__init__()` methods they won't break tests, but there will be warnings:

```
<path>/site-packages/aiohttp/test_utils.py:128
  <path>/site-packages/aiohttp/test_utils.py:128: PytestWarning: cannot collect test class 'TestServer' because it has a __init__ constructor
    class TestServer(BaseTestServer):

<path>/site-packages/aiohttp/test_utils.py:152
  <path>/site-packages/aiohttp/test_utils.py:152: PytestWarning: cannot collect test class 'TestClient' because it has a __init__ constructor
    class TestClient:

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

The easiest way would be to exclude these tests from pytest collection via `__test__ = False`. See https://stackoverflow.com/q/24614851/2038264

N/A

## Related issue number

N/A

## Checklist

### DIFF
--- a/CHANGES/3660.bugfix
+++ b/CHANGES/3660.bugfix
@@ -1,0 +1,1 @@
+Suppress pytest warnings due to test util classes
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -56,6 +56,7 @@ Chris Moore
 Christopher Schmitt
 Claudiu Popa
 Colin Dunklau
+Cong Xu
 Damien Nadé
 Dan Xu
 Daniel García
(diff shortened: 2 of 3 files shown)


asvetlov:
Cool!
I used to think that we have to rename these classes to avoid conflicts.
If `__test__ = False` solves the issue I happy to accept the PR!

PR #3662, opened 2019-03-23
Fix graceful shutdown

amitbl:
## What do these changes do?

Graceful shutdown not working properly since commit https://github.com/aio-libs/aiohttp/commit/dd30b2af1fd0d643e02ef5023f1d8a68a75b5301
When a long running request is in progress, and the server is being gracefully closed, all handlers are  immediately stopped the client receive a empty response.

These changes make the server stop accepting new connections, process existing ones properly and then exit.

On first SIGINT/SIGTERM the server will try to gracefully close all handlers and quit.
If for some reason a handler is stuck, sending a second SIGINT/SIGTERM will force the server to quit immediately 

## Related issue number

#3638 

## Checklist

### DIFF
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -16,15 +16,15 @@ except ImportError:
 
 
 __all__ = ('BaseSite', 'TCPSite', 'UnixSite', 'SockSite', 'BaseRunner',
-           'AppRunner', 'ServerRunner', 'GracefulExit')
+           'AppRunner', 'ServerRunner', 'ForceExit')
 
 
-class GracefulExit(SystemExit):
+class ForceExit(SystemExit):
     code = 1
 
 
-def _raise_graceful_exit() -> None:
-    raise GracefulExit()
+def _force_exit() -> None:
+    raise ForceExit()
 
 
 class BaseSite(ABC):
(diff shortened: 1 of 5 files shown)


jraby:
@asvetlov Are there plans to merge this or something similar in the near future?

Graceful shutdowns, as documented in https://aiohttp.readthedocs.io/en/stable/web_advanced.html#aiohttp-web-graceful-shutdown,  do not work on 3.5.4. (clients get an empty response as soon as the aiohttp process gets a SIGTERM or SIGINT)

asvetlov:
Hi.
I'm pretty busy now by hacking asyncio for upcoming Python 3.8.
3.8 feature freeze is scheduled on the end of May.
In two weeks I will have time for reviewing pending aiohttp pull requests.

Note to myself:
As I see aiohttp 3.5 is affected, I need to backport to 3.5 as well.

MerliX:
Hello @asvetlov @amitbl , graceful shutdown is still important for me, will it be fixed sometime? Thank you

aiohttp/web_runner.py
```diff
@@ -16,17 +16,13 @@
 
 
 __all__ = ('BaseSite', 'TCPSite', 'UnixSite', 'SockSite', 'BaseRunner',
-           'AppRunner', 'ServerRunner', 'GracefulExit')
+           'AppRunner', 'ServerRunner', 'ForceExit')
 
 
-class GracefulExit(SystemExit):
+class ForceExit(SystemExit):
```
webknjaz, line 22: Don't rename it, it's exposed and can be used by end-users.

mjholtkamp:
I ran into this problem and came across this PR. I think the problem was indeed introduced in the mentioned commit, but specifically this line: https://github.com/aio-libs/aiohttp/commit/dd30b2af1fd0d643e02ef5023f1d8a68a75b5301#diff-8e4ccf3aaf43100064462a4df10952e5R214

That line cancels all running tasks in the event loop, so also tasks that were scheduled as background tasks. After cancelling everything, it waits for them to finish. Since `runner.cleanup()` is called in the `finally` of `_run_app()`, it will then proceed to call `on_shutdown` and `on_cleanup` handlers.

In my case, I tried to handle shutdown gracefully by setting a flag and letting the background task gracefully exit, but it's already cancelled because of `_cancel_all_tasks()`.

I think this PR does not fix that particular problem.

asvetlov:
I believe the issue is fixed by #3805

Issue #3669, opened 2019-03-27
broken interface normalize_path_middleware


vir-mir:
## Long story short

middleware `normalize_path_middleware` is not working properly

## Expected behaviour

I appeal to `/paymethods`, there is an endpoint `/paymethod`

`/paymethod` -> 200
`/paymethods` -> 301 -> `/paymethod`

settings: `normalize_path_middleware(remove_slash=True, append_slash=False)`

if switch off 
`/paymethod` -> 200
`/paymethods` -> 404

I think the problem is in the last `if merge_slashes and remove_slash`

```python

            if merge_slashes:
                paths_to_check.append(re.sub('//+', '/', path))
            if append_slash and not request.path.endswith('/'):
                paths_to_check.append(path + '/')
            if remove_slash and request.path.endswith('/'):
                paths_to_check.append(path[:-1])
            if merge_slashes and append_slash:
                paths_to_check.append(
                    re.sub('//+', '/', path + '/'))
            if merge_slashes and remove_slash:
                merged_slashes = re.sub('//+', '/', path)
                paths_to_check.append(merged_slashes[:-1])

```

aiohttp == 3.5.4
## Your environment
server

bigbag:
if we change last condition 
`if merge_slashes and remove_slash:`
on
`if merge_slashes and remove_slash and path.endswith('/'):`
all will be correct.

kxepal:
```
async def test_bug_3669(aiohttp_client):
    async def paymethod(request):
        return web.Response(text="OK")

    async def paymethods(request):
        raise web.HTTPPermanentRedirect('/paymethod')

    app = web.Application()
    app.router.add_route('GET', '/paymethods', paymethods)
    app.router.add_route('GET', '/paymethod', paymethod)
    app.middlewares.append(
        web.normalize_path_middleware(append_slash=False, remove_slash=True)
    )

    client = await aiohttp_client(app, server_kwargs={'skip_url_asserts': True})

    resp = await client.get('/paymethods')
    assert resp.status == 200
    assert resp.url.path == '/paymethod'
```

@vir-mir, this test case passed. How is it different from your's case?

vir-mir:
@kxepal Why did the final URL in the test change? 

you send `/paymethods`
`assert resp.url.path == '/paymethod'`

kxepal:
@vir-mir Because of redirect? Eventually, it causes new request to new location.

vir-mir:
```python
[...]

PR #3670, opened 2019-03-27
Fix normalize_path_middleware behavior for patch without slash (3669)

bigbag:
Fix for issue https://github.com/aio-libs/aiohttp/issues/3669


bigbag:
> LGFM, but could you, please, also add the test made by @vir-mir as a separate case? That one clearly demonstrates the issue and linked with real case.

I did it in https://github.com/aio-libs/aiohttp/pull/3670/commits/0292b96d727c25f2766fbbc46d323b348f1526b7

kxepal:
Thanks! :beer:

Issue #3676, opened 2019-04-03
Response reading is randomly blocked on alpine 3.8


chapurlatn:
## Long story short

Some specific call response reading are blocking on alpine 3.8 (docker python:3.7-alpine3.8).
This happens only on some URLs (and response) and the response is always read unless the last byte.

```python
async with aiohttp.ClientSession() as session:
        async with session.get(ur) as response:
            print(response.status) # ok
            print(await reponse.text()) # never ends, nothing printed!
```

This is reproduced while streaming chunks using dedicated iterator `response.content.iter_chunks()`.

```python
async with aiohttp.ClientSession() as session:
        async with session.get(ur) as response:
            print(response.status) # ok
            async for data, _ in response.content.iter_chunks(): # never go out the loop
              print(data) # all the body is printed
              #  !!! on debian, an empty chunkis printed at the end and it goes out the loop !!!
```

But it does not happen anymore by iterating "manually" over chunk and checking the end status using `response.content.readchunk()`

```python
async with aiohttp.ClientSession() as session:
        async with session.get(ur) as response:
            print(response.status) # ok
            (chunk, end) = await response.content.readchunk()
            print(chunk)
            while end:
                (chunk, end) = await response.content.readchunk()
                print(chunk)
           # this is fully working !
```

First remark: A curl call on the same alpine image works.

Second remark: I was able to fix it by updating the streams.py module in my image after the install and upgrading the chunk iterator (I will join the PR ASAP).

## Expected behaviour

All layers returns the body quickly as aiohttp does usually.

## Actual behaviour

High level chunk reading blocks sometime.

## Steps to reproduce

```
$ docker run -it python:3.7-alpine3.8 /bin/sh
# pip install aiohttp
# python3 -X dev
> 
```
Copy Paste the following python script:
```
import aiohttp
import asyncio
import logging

logging.basicConfig(level=logging.DEBUG)

async def main():
    url = '...'
    async with aiohttp.ClientSession() as session:
        async with session.get(url) as response:
            print(response.status)
            print(await reponse.text()) # blocks sometime

loop = asyncio.get_event_loop()
loop.set_debug(True)
loop.run_until_complete(main())
```

Unfortunately, I was able to reproduce the behaviour only on a request which returns really secured content at my client's office :-(

I can provide generic information like the response body size or any other information you may need to confirm a potential "bug".

Here is a curl call partially censured, hope this help.

```
[...]

PR #3677, opened 2019-04-03
Fix streamed chunk iterator ending (#3676)

chapurlatn:
Stop iterating over chunk when last chunk reading provided the end flag.

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -49,14 +49,16 @@ class ChunkTupleAsyncStreamIterator:
 
     def __init__(self, stream: 'StreamReader') -> None:
         self._stream = stream
+        self._hasnext = True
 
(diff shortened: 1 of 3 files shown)


aiohttp/streams.py
```diff
@@ -49,14 +49,16 @@ class ChunkTupleAsyncStreamIterator:
...
         return self
 
     async def __anext__(self) -> Tuple[bytes, bool]:
-        rv = await self._stream.readchunk()
-        if rv == (b'', False):
-            raise StopAsyncIteration  # NOQA
+        if self._hasnext:
+            raise StopAsyncIteration  # NOQA            
+        (rv, end) = await self._stream.readchunk()
+        self._hasnext = not end
```
socketpair, line 61: it is definitely wrong. `end` is not the end of the stream. it means that this is the last subchunk of HTTP chunk.

Issue #3681, opened 2019-04-05
StreamReader.read(n) returns less than n bytes when more data available


raees-khan:
## Long story short

If the chunk size is greater than the chunk of data received by the server, the StreamReader.read returns whatever size chunk of data was received by the server.

## Expected behaviour

StreamReader.read(n) should return n bytes when n is not -1.

## Steps to reproduce

```
async def download(large_file):
    async with aiohttp.ClientSession() as session:
        async with session.get(large_file) as resp:
            while True:
                 chunk = await resp.content.read(10 * (1024 * 1024)) # Read 10MB
                  if not chunk:
                      break
                  print(len(chunk)) # the data chunk returned by streamreader is a few hundred KBs
```

socketpair:
read vs read_exactly ?

socketpair:
I don't see any bugs here.

raees-khan:
> I don't see any bugs here.

- Go to https://docs.aiohttp.org/en/stable/streams.html.
- Read the description of `StreamReader.read`.
- Then download a large file and use `StreamReader.read` to read data in chunks of 10MB.
- You should be able to reproduce the behaviour I described in my original post.

As per the document, `StreamReader.read` should read N bytes passed when N is not -1. Either documentation is wrong or StreamReader has a bug.

socketpair:
@cpython-rocks Accroding to docs: `Read up to n bytes`. So if `n` is not `-1` it returns `n` bytes **or less**. Internally, it does not even try to accumulate buffers, but it will split big buffer to return `up to n bytes`.

Please explain why you need 10MB exactly. Internally Aiohttp receives data in small chunks. Merging them into pieces of, say, 10MB is typically useless work. Merge them in your application if you want to.

raees-khan:
@socketpair Wouldn't it read until EOF if `n is -1` and return all bytes? 10MB is just an example. What I am trying to highlight is the fact that when `n` is passed as an argument, then `read` should return `n` bytes no less (unless it reaches EOF) otherwise if it just returns the chunk received from server, then there is no point in having an optional param `n`.

```
while True:
    chunk = await resp.content.read(1048576) # Read 1 MB chunk
        if not chunk:
            break
        print(len(chunk)) # I expect `chunk` to be 1MB and it must be 1MB otherwise the `n` param is pointless.
```

socketpair:
[...]

PR #3686, opened 2019-04-11
streams: block until n bytes read (#3681)

socketpair:
## Checklist

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -346,31 +346,18 @@ class StreamReader(AsyncStreamReaderMixin):
                         'Multiple access to StreamReader in eof state, '
                         'might be infinite loop.', stack_info=True)
 
-        if not n:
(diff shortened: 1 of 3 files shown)


kxepal:
This fixes what's case?

socketpair:
@kxepal #3681

socketpair:
I will write tests if concept is accepted.

socketpair:
Well. Unittests test current implementation instead of specification, so it is expected that they will fail in many cases...

asvetlov:
I've missed those read*() methods behavior.
Could you briefly summarize that should be in the end?

Dreamsorcerer:
I think we'll close this one. The linked issue is already closed and the behaviour can be achieved with itertools, without forcing all users to have delayed chunks when they don't need full chunk sizes.

Issue #3693, opened 2019-04-15
Refactor Cython usage


asvetlov:
pip 19+ uses isolated virtual environments to install a package.
It breaks a build, please read #3581 report

The idea is:
1. We don't need Cython to be installed in venv if all `.pyx` files are already precompiled into `.c` sources.
2. `.c` are already present in distribution tarball. Binary wheels don't need to compile sources at all.
3. Precompilation step can be done in `Makefile` *before* installing `aiohttp` library.
4. Pure-python installation should require `AIOHTTP_NO_EXTENSIONS=1` env var. Otherwise, a compilation error is raised. It prevents a silent performance degradation if aiohttp installed on Alpine or OS without C compiler toolchain.

#3694 addresses points 1-3, it can be backported to aiohttp 3.5.
The 4th bullet is not backward compatible. It should be implemented in a separate PR and land on aiohttp 4.0 only.

webknjaz:
See also: https://github.com/pypa/pip/issues/6144

I think I've actually got it figured out in #3581.
https://www.python.org/dev/peps/pep-0517/#in-tree-build-backends would work for us but it's not yet implemented by Pip.

The current implementation relies on an undefined `requires` filtering behavior as per https://github.com/pypa/pip/issues/6410.

So the only thing left to do is to put it into a separate dist. (And it works!)

asvetlov:
manylinux is broken still but #3694 solves almost everything

webknjaz:
I'd keep this open anyway. I'm still thinking of improvements.

asvetlov:
Up to you, the PR will hang for years then.
I have a feeling that supporting both brand new and 2-years-old pip versions is a good idea.

asvetlov:
The issue is solved by adding explicit `make cythonize` step.

PR #3694, opened 2019-04-15
Don't depend on Cython in setup.py

### LINKED ISSUE #3581: sdist build gets crashed under pip>=19 in dev mode
## Long story short

We have `cython` as an optional dependency. That's why we install it as a pre-requisite in the CI, as a separate step.
New pip creates a separate build virtualenv which doesn't have access to the place with cython installed which causes it to crash.

## Expected behaviour

It succeeds

## Actual behaviour

It tracebacks

## Steps to reproduce

* https://travis-ci.com/aio-libs/aiohttp/jobs/172249543#L198-L219
* https://ci.appveyor.com/project/aio-libs/aiohttp/build/job/uppd0qqw2sbisqtn#L46

## Your environment

Travis CI (doesn't really matter actually)

asvetlov:
Fixes #3581 
See also #3693
The PR is an alternative to #3589


aiohttp/_helpers.pyi
```diff
@@ -1,5 +1,6 @@
 from typing import Any
 
+
```
webknjaz: This should probably get cherry-picked along with other unrelated changes...

aiohttp/__init__.py
```diff
@@ -14,11 +14,11 @@
     ClientOSError,
     ClientPayloadError,
     ClientProxyConnectionError,
-    ClientResponse,
     ClientRequest,
+    ClientResponse,
```
webknjaz: @asvetlov could you please just commit style changes into master so that it'd not show up here making it harder to review.

asvetlov: Yes, sure.
Thanks for the recall.
I don't wont to see this mess in the PR :)

webknjaz:
@asvetlov we forgot about manylinux1

asvetlov:
Do you mean that build of manylinux wheels doesn't compile cython source files?

webknjaz:
@asvetlov yes, because we don't rely on make there (and shouldn't)

PR #3703, opened 2019-04-18
Add documentation for persistent session creation & disposal

### LINKED ISSUE #3658: How can one create a persistent ClientSession
The official aiohttp doc is ambiguous, it says that creating a session per request is a **very bad idea**, but apparently doesn't tell how one can create a persistent session. There are plenty of examples all across the internet on how one can make concurrent requests, but there is unfortunately no example reusing the same session across all requests.

One would argue that the following code reuses the session which is partially true. Partially because every time I call main(), a new session will be created.

```
        return await resp.text()

async def main():
    urls = ['https://httpstat.us/200'] * 10
         tasks = [asyncio.create_task(fetch(session, url)) for url in urls]
         await asyncio.gather(*tasks)

asyncio.run(main())
```

The following example in doc is quite confusing:
```
async with session.get('...'):
    # ...
await session.close()
```

Here's what I call a persistent session in its true sense because no matter where you call fetch from or how many times you call it, the same session will be reused, but this apparently doesn't work
```
class Session:
    def __init__(self):
        self.session = aiohttp.ClientSession()
    
    async def fetch(self, url):
        async with self.session.get(url) as resp:
            return await resp.text()
```

This code doesn't work and `RuntimeError: Timeout context manager should be used inside a task` exception is caught. If you pass `timeout=None` to `self.session.get` it fixes the `RuntimeError`, but you get other warnings about `Unclosed Session`. I've learnt that you shouldn't create a `session` outside a `coroutine` and was suggested something like `async def create_session(): return aiohttp.ClientSession` which brings me back to square on how would I persist it.

I believe the author should provide examples of creating persisting `session`.

kornicameister:
## What do these changes do?

Add documentation about how to **create** and **destroy** persistent session.

Not in terms of functionality. Users will be able to find out how to create persistent session, though.

## Related issue number

Closes #3658 

## Checklist

### DIFF
--- a/CHANGES/3685.doc
+++ b/CHANGES/3685.doc
@@ -1,0 +1,1 @@
+Add documentation regarding creating and destroying persistent session.
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -535,7 +535,7 @@ advise against it. :class:`aiohttp.ClientSession` maintains a
 connection pool. Contained connections can be reused if necessary to gain some
 performance improvements. If you plan on reusing the session, a.k.a. creating
 **persistent session**, you can use either :ref:`aiohttp-web-signals` or
-:attr:`Application.cleanup_ctx`. If possible we advise using :attr:`Application.cleanup_ctx`,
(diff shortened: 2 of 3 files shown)


asvetlov:
Thanks!

Issue #3701, opened 2019-04-17
Cannot use subapps with overlapped prefixes


Mirraz:
## Long story short

If I add to main app one subapp with prefix "/qwe/" and then add another subapp with prefix "/qwerty/" then routing to second subapp does not work. But if I firstly add subapp with prefix "/qwerty/" and then subapp with prefix "/qwe/" then everything works fine. See example.

## Steps to reproduce

```python
from aiohttp.web import Application, json_response, run_app


def create_subapp01():
    app = Application()
    app.router.add_route('GET', r'/aaa', lambda req: json_response({'subapp01': 'aaa'}))
    return app


def create_subapp02():
    app = Application()
    app.router.add_route('GET', r'/aaa', lambda req: json_response({'subapp02': 'aaa'}))
    return app


def create_app():
    app = Application()

    app.add_subapp('/qwe/', create_subapp01())
    app.add_subapp('/qwerty/', create_subapp02())
    # GET /qwerty/aaa -> HTTP code 404

    # app.add_subapp('/qwerty/', create_subapp02())
    # app.add_subapp('/qwe/', create_subapp01())
    # GET /qwerty/aaa -> HTTP code 200

    return app


run_app(create_app(), host='127.0.0.1', port=8080)
```

## Your environment

aiohttp 3.5.4

asvetlov:
This is by design: URL dispatcher handles all routes in the order of adding.
The same behavior is for overlapping routes in the application, not for sub-apps only.

Nothing to do here, the issue is not a bug

socketpair:
@asvetlov what about **slash** ? if We add

```
app.add_subapp('/qwe/', create_subapp01())
app.add_subapp('/qwerty/', create_subapp02())
```

Why queries like `/qwerty/asd` match the first subapp ?

asvetlov:
Ah, I see. That's interesting.

nikie:
Hi, 
Feels like a bug. I have just created PR #3708 which supposedly fixes it.
This is my first contribution, so please be picky and cautious :)

Eugene.

asvetlov:
Fixed by #3708

PR #3708, opened 2019-04-22
Fix overshadowing of overlapped subapp prefixes (#3701).

nikie:
## What do these changes do?

Fix for issue #3701 - overshadowing of overlapped subapp prefixes.

The fix allows different subapps to have prefixes with overlapping beginning. The following didn't work properly before - the second subapp prefix was hidden by the first one as subapp prefixes matching was performed by `startswith` only, so the first added app won:
```
    app.add_subapp('/qwe/', create_subapp01())
    app.add_subapp('/qwerty/', create_subapp02())
```

Subapps with overlapping prefixes can now be used without unexpected overshadowing.

## Related issue number

#3701

## Notes

+ I have considered storing the value of `self._prefix + '/'` in an attribute for optimization but was not sure to do that because it adds complexity. Please let me know if such optimization is desired - I can dig further.

+ Without adding `request.url.raw_path != self._prefix` to the comparison the existing behavior of matching empty and root subapp routes would have changed. That didn't break any existing test though, so I have added 2 more tests to make sure this works as before - `test_urldispatch.py/test_prefixed_subapp_empty_route` and `test_urldispatch.py/test_prefixed_subapp_root_route`

## Checklist

- [ ] Documentation reflects the changes.
_Please let me know if any documentation change is required._

### DIFF
--- a/CHANGES/3701.bugfix
+++ b/CHANGES/3701.bugfix
@@ -1,0 +1,1 @@
+Fix overshadowing of overlapped subbaps prefixes.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -84,6 +84,7 @@ Eric Sheng
 Erich Healy
 Eugene Chernyshov
 Eugene Naydenov
+Eugene Nikolaiev
 Eugene Tolmachev
 Evert Lammerts
 FichteFoll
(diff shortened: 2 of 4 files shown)


asvetlov:
Thanks!
Sorry for the long delay

Issue #1925, opened 2017-05-26
Not waiting for connection_lost() when transport closed


omnicognate:
## Long story short

aiohttp calls `close()` on the transport and then immediately discards it without waiting for `connection_lost()`. This is a problem for SSL transports, which have to shut down asynchronously, and it appears aiohttp only "works" with the standard library's SSL transport because of a bug in that transport.

## Expected behaviour

aiohttp waiting for connection_lost() before discarding transport

## Actual behaviour

It not doing so

## Steps to reproduce

    import aiohttp
    import asyncio

    async def main():
        conn = aiohttp.TCPConnector(verify_ssl=False)
        async with aiohttp.ClientSession(connector=conn) as session:
            async with session.get('https://httpbin.org/') as resp:
                resp.raise_for_status()

    loop = asyncio.get_event_loop()
    try:
        loop.run_until_complete(main())
    finally:
        loop.close()

Run this program and breakpoint `connection_made()` and `connection_lost()` in client_proto.py. The former is called, but not the latter. 

The program appears to run successfully, but to convince you that something is severely amiss, add a `__del__` method to the class `_SelectorSocketTransport` in asyncio's selector_events.py and breakpoint it and that transport's `close()` method. You'll see that `close()` is never called and `__del__` gets called during shutdown, after the loop has been closed and the module finalised, because it still has a reader registered and is therefore leaked (while still waiting on I/O) in a reference cycle with the event loop.

If I'm understanding it right, the overall behaviour appears to be due to two bugs, one in asyncio and one in aiohttp:

1. When `close()` is called on the SSLProtocolTransport in asyncio's ssl_proto.py it initiates an asynchronous shutdown process, which should ultimately result in it calling `connection_lost()`. However, this doesn't progress for reasons I haven't fully figured out, so `connection_lost()` is never called. The shutdown process is clearly broken because it never results in `close()` being called on the underlying transport, which is leaked as shown above.
2. aiohttp doesn't wait for `connection_lost()`, so the hang that one would expect as a result of this never happens. By luck, the broken shutdown sequence never actually causes any errors.

I'm actually more interested in the latter, the aiohttp behaviour, because I've written my own SSL transport (based on pyopenssl), which does a correct asynchronous shutdown sequence. When aiohttp discards my transport without waiting for `connection_lost()`, it causes errors, which I don't think are my fault.

## Your environment

[...]

PR #3733, opened 2019-05-07
Fix closing client session

ayushkovskiy:
## What do these changes do?

Make `BaseConnector.close()` an asynchronous function (still `BaseConnector._close()` is synchronous and preserves the old behaviour).

Now, `await connector.close()` waits until all the connections are closed instead of just sending the closing signal to all the connections. The "Unclosed connection" warnings appeared because some connections required more time to be closed because some protocols (TLS) perform additional steps on connection-close and thus require longer time.

Yes: 
- `BaseConnector.close()` -> `await BaseConnector.close()`
- Deprecated `with Connector():` syntax is dropped

## Related issue number
https://github.com/aio-libs/aiohttp/issues/3736 https://github.com/aio-libs/aiohttp/issues/1925.
This PR removes functionality deprecated in PR https://github.com/aio-libs/aiohttp/pull/3417/.

## Checklist


aiohttp/base_protocol.py
```diff
@@ -79,3 +86,7 @@ def connection_lost(self, exc: Optional[BaseException]) -> None:
         waiter = self._loop.create_future()
         self._drain_waiter = waiter
         await waiter
+
+    @abc.abstractmethod
+    def close(self) -> None:
```
atemate: for consistency

aiohttp/connector.py
```diff
@@ -287,7 +291,7 @@ def __enter__(self) -> 'BaseConnector':
         return self
 
     def __exit__(self, *exc: Any) -> None:
-        self.close()
+        self._close()
```
atemate: so now `_close` is synchronous, while `close` is asynchronous

aiohttp/base_protocol.py
```diff
@@ -16,6 +18,7 @@ def __init__(self, loop: asyncio.AbstractEventLoop) -> None:
         self._reading_paused = False
 
         self.transport = None  # type: Optional[asyncio.Transport]
+        self.closed = self._loop.create_future()  # type: asyncio.Future[Optional[Exception]]  # noqa
```
asvetlov: Please move the future into `client_proto.py`.
Server side protocols don't need it.

atemate: sure, thank. Fixed.

aiohttp/connector.py
asvetlov: Ok. We should drop sync `__enter__` / `__exit__` in the following PR, they are not needed anymore.
Even worse: `with connector:` statement becomes dangerous.

asvetlov: Well, drop it now please

aiohttp/base_protocol.py
```diff
@@ -54,6 +57,10 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
 
     def connection_lost(self, exc: Optional[BaseException]) -> None:
         self._connection_lost = True
+        if exc is not None:
```
asvetlov: Please don't touch this file but add all required logic to `client_proto.py`

atemate: ok

aiohttp/connector.py
atemate: ok sure

asvetlov:
Please don't forget to update docs.
http://docs.aiohttp.org/en/stable/client_advanced.html and http://docs.aiohttp.org/en/stable/client_reference.html mention connectors. These pages should reflect the PR changes.

atemate:
Sorry for this long-pending PR, I will return back to it in the end of June 2019 when I have more time, thank you

aiohttp/connector.py
```diff
@@ -386,20 +389,25 @@ def _cleanup_closed(self) -> None:
...
-    def close(self) -> Awaitable[None]:
+    async def close(self) -> None:
         """Close all opened transports."""
-        self._close()
-        return _DeprecationWaiter(noop2())
+        waiters = self._close_immediately()
+        if waiters:
+            await asyncio.gather(*waiters,
+                                 loop=self._loop,
+                                 return_exceptions=True)
```
socketpair: possibly re-raise ? I'm unsure

aiohttp/connector.py
```diff
@@ -412,14 +420,19 @@ def _close(self) -> None:
             for data in self._conns.values():
                 for proto, t0 in data:
                     proto.close()
+                    waiters.append(proto.closed)
 
             for proto in self._acquired:
                 proto.close()
+                waiters.append(proto.closed)
 
+            # TODO (A.Yushovskiy, 24-May-2019) collect transp. closing futures
```
socketpair, line 432: incomplete task ?

aiohttp/connector.py
atemate: Ineed, we lost exceptions here. But maybe we should just log (as warnings) them instead?

aiohttp/connector.py
atemate: work with transports here, especially `self._cleanup_closed_transports`, seem to be a workaround of the non-closed transports for SSL only:
https://github.com/aio-libs/aiohttp/blob/dcce5ea779a580c05c2b0f0555ea5cccdb4aea29/aiohttp/connector.py#L345-L348

Also, transports in `_cleanup_closed_transports` are of type `asyncio.Transport` so we can not modify them to save closing future same way as we did in this PR with `ResponseHandler`, thus I don't see a way to get the awaitable result of `transport.abort()`.

Good news is that most likely these calls are redundant since we close all protocols (i.e., save them to the list of futures: `waiters.append(proto.closed)`)

aiohttp/connector.py
```diff
@@ -72,6 +71,8 @@
     from .client_reqrep import ConnectionKey  # noqa
     from .tracing import Trace  # noqa
 
+_ListOfFutures = List['asyncio.Future[Optional[Exception]]']
```
asvetlov: Is it a list of futures that contains None or Exception as a **value** or list of futures with `fut.set_result(None)` or `fut.set_exception(exc)`?
If later -- please use `Future[None]`.
I believe we never return an exception but set it explicitly.

aiohttp/connector.py
asvetlov: Heh, very interesting question.
Python and asyncio itself have no MultipleException yet.
Maybe we add it in Python 3.9.
I'm not sure if `warnings.warn()` is a good choice here: it is not a programming error but network issue.
Please use `logging.error()` instead.

aiohttp/connector.py
```diff
@@ -49,7 +49,6 @@
     CeilTimeout,
     get_running_loop,
     is_ip_address,
-    noop2,
```
asvetlov, line 52: Please delete `helpers.noop2` function as well.

aiohttp/connector.py
atemate: ok

PR #3518, opened 2019-01-10
tests_require: add trustme

dotlambda:
It is required since https://github.com/aio-libs/aiohttp/pull/3487.

## Checklist

- [ ] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #3752, opened 2019-05-10
Drop setup.py test support

asvetlov:
Closes #3518

We decided to not support `setup.py test` facility anymore.
Use `make test` now.
@webknjaz has a plan to add `tox` configuration -- it would be a nice addition.
`tox` can be used by CI and users who comfortable with this workflow.
For now on my daily aiohttp hacking I personally prefer old good makefiles

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,16 +83,6 @@ def read(f):
     return (here / f).read_text('utf-8').strip()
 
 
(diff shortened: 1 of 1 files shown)


webknjaz:
I'd also add that `setup.py test` is *officially* discouraged during the last several years anyway.

PR #3753, opened 2019-05-10
Enforce URL for cookie_jar.filter_cookies() call

asvetlov:
CookieJar technically supports both `str` and `URL`.
But only `URL` is pointed as an acceptable type in type hints information.
Moreover, the Jar is a kind of *internal* class. User can provide a custom implementation but the caller is always `aiohttp` itself.
The method is called with URL only by aiohttp.

Deprecation for `str` is added.
We can drop `str` eventually with a deprecation period.

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -4,6 +4,7 @@ import os  # noqa
 import pathlib
 import pickle
 import re
+import warnings
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


bdraco:
It looks like this was never back-ported so it does not warn on 3.11 or 3.10 so nobody is getting notice that it will stop working

PR #3796, opened 2019-05-23
Add debug argument to `web.run_app()`

### LINKED ISSUE #3793: Add debug argument to web.run_app()
```
500 Internal Server Error
Traceback:
   ...
NameError: name '...' is not defined
```
need a simple debug=False in web.run_app() like flask for security?

NewUserHa:
## What do these changes do?

Add debug argument to web.run_app() & new_event_loop()

no?

## Related issue number

Resolves #3793

## Checklist


aiohttp/web.py
```diff
@@ -388,7 +389,9 @@ def run_app(app: Union[Application, Awaitable[Application]], *,
             reuse_address: Optional[bool]=None,
             reuse_port: Optional[bool]=None) -> None:
     """Run an app locally"""
-    loop = asyncio.get_event_loop()
+    loop = asyncio.new_event_loop()
```
webknjaz: Why create a new one?

asvetlov: I've asked @NewUserHa to make `run_app` as close to `asyncio.run()` as possible.
Please look at https://github.com/python/cpython/blob/master/Lib/asyncio/runners.py

docs/deployment.rst
```diff
@@ -321,3 +321,11 @@ By default aiohttp uses own defaults::
...
 For more information please read :ref:`Format Specification for Access
 Log <aiohttp-logging-access-log-format-spec>`.
+
+
+Debug Mode
+---------------------
+
+use ``web.run_app(app, debug=True)`` to easily turn on debug messages.
+
+reload need <https://docs.aiohttp.org/en/latest/web_advanced.html#dev-tools>
```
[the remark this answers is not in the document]
NewUserHa: they are both `debug related` document.
putting them together?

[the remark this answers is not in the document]
asvetlov: We *mention* `aiohttp-devtools` in docs, that's it.
There is no need to describe the tool more.
If debug flag behavior affects `aiohttp-devtools` somehow the documentation for `aiohttp-devtools` should reflect this fact.
@samuelcolvin what is your opinion?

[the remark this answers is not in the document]
samuelcolvin: agreed.

We could change the docs and particularly https://demos.aiohttp.org/en/latest/tutorial.html#aiohttp-demos-polls-getting-started to use devtools more, but that's a different issue.

docs/deployment.rst
```diff
@@ -321,3 +321,11 @@ By default aiohttp uses own defaults::
 
 For more information please read :ref:`Format Specification for Access
 Log <aiohttp-logging-access-log-format-spec>`.
+
+
+Debug Mode
+---------------------
+
+Use ``web.run_app(app, debug=True)`` to easily turn on debug messages.
```
samuelcolvin: can we give some more detail here:
* explain that this enables `asyncio.debug`
* explain what `debug messages` means. I think it just refers to tracebacks, but is there any other changes to messaging/logging?

I think we also need explain that if you're using lower level components like `AppRunner` that debug is infered from `asyncio.debug`?

aiohttp/web.py
webknjaz: Alright, so perhaps a comment in the code explaining the motivation would be helpful

docs/deployment.rst
asvetlov: Agree.
It enables asyncio debug and switches on access log printing if the `aiohttp.access` logger was not configured earlier explicitly.

@samuelcolvin seems like @NewUserHa is not a native speaker, writing English texts is a complex task for him.
Would you provide a comment with an example of the proposed text to copy-paste it into the commit?

samuelcolvin: Ok, but I think this is the wrong place to document this behaviour, personally I think we should:
* add a comment to [quickstart](https://docs.aiohttp.org/en/stable/web_quickstart.html) saying "you might want to use `debug=True` to see tracebacks, there's also aiohttp-devtools if you're looking for code reloading functionality"
* give a fuller description of how debug works in the [`run_app` function documentation](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.run_app)

If you agree on that I'm happy to write the text but it sounds like another PR would be the easiest way to do it?

asvetlov: Agree, a separate PR would be easier for you to make it done.

Let's drop this section here.
Note: `debug` flag should be enumerated in parameters list for `run_app` in `web_reference` in by PR. Later @samuelcolvin provide a narrative text for the flag usage.

aiohttp/web.py
```diff
@@ -388,7 +389,10 @@ def run_app(app: Union[Application, Awaitable[Application]], *,
             reuse_address: Optional[bool]=None,
             reuse_port: Optional[bool]=None) -> None:
     """Run an app locally"""
-    loop = asyncio.get_event_loop()
+    # behavior like asyncio.run() as a main entry point for asyncio programs
```
asvetlov:
```suggestion
    # The function design follows asyncio.run() behavior
```

docs/web_reference.rst
```diff
@@ -2725,6 +2725,8 @@ Utilities
    :param app: :class:`Application` instance to run or a *coroutine*
                that returns an application.
 
+   :param bool debug: same as `loop.set_debug(True)`
```
```suggestion
   :param bool debug: enable :ref:`asyncio debug mode <asyncio-debug-mode>` if ``True``.
```

asvetlov:
The code looks good.
Please take a look at https://travis-ci.com/aio-libs/aiohttp/jobs/203039958 
`make flake` command fails :(

NewUserHa:
thanks
it passes now

asvetlov:
Good!
The next and last step is fixing tests.
`make test` hangs now :(

NewUserHa:
`tests/test_run_app.py::test_run_app_http[pyloop] ` I'm asked to add `asyncio.new_event_loop()` :(

NewUserHa:
but this seems about argument position.
what about putting debug flag at the last, again?

but that's keyword-only argument in the test hanged. how did last build success?

asvetlov:
The code from PR is correct, tests should be updated to work with it.
All tests that checks some network and thus relies on `patched_loop` can be easily switched from testing `web.run_app()` to internal `web._run_app()`

NewUserHa:
simply replacing web.run_app() with web._run_app() doesn't work

asvetlov:
@NewUserHa if you have no time/energy to finish the PR -- that's sad but ok.
I have no estimation when I'll have a time window to work on the issue.
It is not in my current very long to-do list yet, sorry.
Maybe somebody else can pick it up and make it done.

Mechanical `run_app()` -> `_run_app()` in tests is not a solution.
You should iterate over all failed tests to figure out how to fix a failure. There is no generic and easy way to make everybody happy.

NewUserHa:
no.
since I'm not familiar with asyncio test, I'm waiting if you have ideas for this. because you said `can be easily switched from testing web.run_app() to internal web._run_app()`

need I go to find why by myself?

aiohttp/web.py
```diff
@@ -388,7 +389,9 @@ def run_app(app: Union[Application, Awaitable[Application]], *,
             reuse_address: Optional[bool]=None,
             reuse_port: Optional[bool]=None) -> None:
     """Run an app locally"""
-    loop = asyncio.get_event_loop()
+    # The function design follows asyncio.run() behavior
```
webknjaz: ....that is why....?

NewUserHa: asvetlov asked it

NewUserHa:
I can't get the change of `new_event_loop` pass the test, so get back to the original issue that only change the debug argument first.

PR #3803, opened 2019-05-27
Use Brotli instead of brotlipy

felixonmars:
brotlipy is stuck at brotli 0.6 and upstream is inactive. Let's switch
to the official binding which is up-to-date.

## Checklist

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -698,8 +698,26 @@ class DeflateBuffer:
             if not HAS_BROTLI:  # pragma: no cover
                 raise ContentEncodingError(
                     'Can not decode content-encoding: brotli (br). '
-                    'Please install `brotlipy`')
-            self.decompressor = brotli.Decompressor()
+                    'Please install `Brotli`')
+
+            class BrotliDecoder:
+                # Supports both 'brotlipy' and 'Brotli' packages
+                # since they share an import name. The top branches
+                # are for 'brotlipy' and bottom branches for 'Brotli'
+                def __init__(self) -> None:
+                    self._obj = brotli.Decompressor()
+
+                def decompress(self, data: bytes) -> bytes:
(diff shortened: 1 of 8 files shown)


requirements/ci-wheel.txt
```diff
@@ -2,7 +2,7 @@
 attrs==19.1.0
 async-generator==1.10
 async-timeout==3.0.1
-brotlipy==0.7.0
+Brotlipy==1.0.7
```
kxepal: It seems should be just `Brotli==1.0.7`, right?

felixonmars: Indeed, fixed!

felixonmars:
> Please fix mypy checker

Updated. Hope I did it right.

aiohttp/http_parser.py
```diff
@@ -698,8 +698,26 @@ def __init__(self, out: StreamReader, encoding: Optional[str]) -> None:
             if not HAS_BROTLI:  # pragma: no cover
                 raise ContentEncodingError(
                     'Can not decode content-encoding: brotli (br). '
-                    'Please install `brotlipy`')
-            self.decompressor = brotli.Decompressor()
+                    'Please install `Brotli`')
+
+            class BrotliDecoder(object):
```
webknjaz: In Python 3 all classes are new-style already:
```suggestion
            class BrotliDecoder:
```

felixonmars: Updated, thanks!

asvetlov:
I need to learn brotlipy and Brotli API difference deeper.
Please don't merge.
4 days are left before Python 3.8 feature freeze.
Perhaps I can return to the PR only after this date.

romansalin:
This change has been merged to master, but  this is does not exist in 3.7.3 version as documented [here](https://docs.aiohttp.org/en/stable/changes.html). 3.7.3 version is still using brotlipy package.

Issue #3572, opened 2019-01-22
`OSError: AF_UNIX path too long` under MacOS


asvetlov:
Surprisingly Travis fails on MacOS: https://travis-ci.com/aio-libs/aiohttp/builds/98188999?utm_medium=notification&utm_source=email

Looks like it is the result of my https://github.com/aio-libs/aiohttp/pull/3548 PR.
I have no Mac available, the investigation can be slow.
If somebody can run tests locally on Mac and debug what is wrong -- it would be awesome and save my personal time a lot.

webknjaz:
I think we've already seen the issue with fs paths for unix sockets being too long in the past...

```console
[1m[31m_________________________ test_unix_connector[pyloop] __________________________[0m



unix_server = <function unix_server.<locals>.go at 0x10bfd58c8>

unix_sockname = '/private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pytest-of-travis/pytest-0/test_unix_connector_pyloop_0/socket.sock'



[1m    @pytest.mark.skipif(not hasattr(socket, 'AF_UNIX'),[0m

[1m                        reason='requires UNIX sockets')[0m

[1m    async def test_unix_connector(unix_server, unix_sockname) -> None:[0m

[1m        async def handler(request):[0m

[1m            return web.Response()[0m

[1m    [0m

[1m        app = web.Application()[0m

[1m        app.router.add_get('/', handler)[0m

[1m>       await unix_server(app)[0m



[1m[31mtests/test_connector.py[0m:2106: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[1m[31mtests/test_connector.py[0m:60: in go

[1m    await site.start()[0m

[1m[31maiohttp/web_runner.py[0m:128: in start

[1m    ssl=self._ssl_context, backlog=self._backlog)[0m

[1m[31m../../../.pyenv/versions/3.5.3/lib/python3.5/asyncio/coroutines.py[0m:210: in coro

[1m    res = func(*args, **kw)[0m

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 



self = <_UnixSelectorEventLoop running=False closed=False debug=False>

[...]

PR #3827, opened 2019-06-08
Set up CI with Azure Pipelines

asvetlov:


samuelcolvin:
Great idea. Let me know if you want any help.

asvetlov:
I'd like to get an initial working setup and iterate over in separate requests.
40 min per PR for appveyor is too much

webknjaz:
Here's a very good curated template for Azure Pipelines: https://github.com/tox-dev/azure-pipelines-template

webknjaz:
But yes, we can merge whatever works initially and iterate on that later.

tests/test_connector.py
```diff
@@ -56,7 +56,15 @@ def unix_server(loop, unix_sockname):
...
         await runner.setup()
         site = web.UnixSite(runner, unix_sockname)
-        await site.start()
+        try:
+            await site.start()
+        except OSError as exc:
+            if str(exc) == "AF_UNIX path too long":
+                if sys.platform == "linux":
+                    raise
+                pytest.skip("Some UNIXes has too strict limitation "
```
webknjaz: Actually all of them. It just happens that the default tmp path prefixed there under macOS is longer (`/private/tmp/...` vs. `/tmp/...`). I think, the right way to fix it is to substitute the temp dir.
It should be possible to do `export TMPDIR=/tmp` in CI. Try it.


Ref: https://unix.stackexchange.com/a/367012/27133

webknjaz: Also, in this case, it should be probably `xfail()` instead of `skip()` if you still want to keep it. But I don't like the idea of skipping tests when it's possible to address this properly.

webknjaz: It looks like a proper workaround it using `tmpdir_factory` fixture: https://github.com/agronholm/anyio/pull/42/files#diff-f0c1c7497545019b8489fa8bc06be6cdR208

webknjaz: Ref: #3572

asvetlov: xfail sounds good!

asvetlov: It is a workaround that doesn't guarantee short path anyway :(
Also, I prefer `tmp_path_factory` over `tmpdir_factory`

webknjaz: Once this https://github.com/aio-libs/aiohttp/pull/3832 is merged, revert the hack.

webknjaz: Yep, using `tmp_path_factory` now.

webknjaz: I've added a few improvements and it is now guaranteed to be short.

tests/test_web_runner.py
```diff
@@ -110,7 +111,15 @@ def test_non_app() -> None:
     await tcp.start()
     path = str(tmpdir / 'tmp.sock')
```
webknjaz: This will have to use a proper sock from a fixture.

setup.py
```diff
@@ -3,9 +3,6 @@
 import pathlib
 import re
 import sys
-from distutils.command.build_ext import build_ext
```
webknjaz: Can this be moved to a separate PR?

tests/test_client_functional.py
```diff
@@ -1,5 +1,3 @@
-"""HTTP client functional tests against aiohttp.web server"""
```
webknjaz, line 1: Please keep the module docstring.

tests/test_client_functional.py
```diff
@@ -1200,7 +1198,7 @@ def do_release():
     app.router.add_post('/', handler)
     client = await aiohttp_client(app)
 
-    with fname.open() as f:
```
webknjaz, line 1201: It'd be nice to move this to a separate PR.

tests/test_client_functional.py
asvetlov: Why? What is the point?

tests/test_client_functional.py
asvetlov: The test fails on azure but is passed on appveyor

setup.py
asvetlov: Yes, sure

tests/test_client_functional.py
webknjaz: When I open a python module, regardless of whether it is a test suite or not it helps me to understand what I'm about to see next, get some context and decide whether I need to read further.

tests/test_client_functional.py
webknjaz: Okay, but I still think that it'd be nice to separate this and only keep the CI config here so that it'd be more atomic.

tests/test_client_functional.py
asvetlov: I have a different position: docstrings are for public API only.
It should be just a comment otherwise.

tests/test_client_functional.py
asvetlov: I cannot test it in a separate PR, this is the problem.

webknjaz:
I think it may make sense to focus on GitHub workflows as they provide more resources.

asvetlov:
Like Windows and MacOS boxes?

webknjaz:
Yes

asvetlov:
"GitHub Actions is currently in limited public beta and is subject to change. We strongly recommend that you do not use this feature for high-value workflows and content during the beta period."

I want to use a tool, not fight with unstable yet features. aiohttp itself provides a lot of space for experimenting.

webknjaz:
I think they forgot to update that. Basically, anybody applying for access gets it instantly now and it's going to be generally available on Nov 13.
I've been playing with it for a while and I'm happy with the results.

asvetlov:
Done by separate PRs

PR #3828, opened 2019-06-08
Disable implicit switch-back to pure python mode.

asvetlov:
The build fails loudly if aiohttp cannot be compiled with C Accellerators.
Use AIOHTTP_NO_EXTENSIONS=1 to explicitly disable C Extensions complication and switch to Pure-Python mode.
Note that Pure-Python mode is significantly slower than compiled one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,7 +156,6 @@ jobs:
     install:
     - *upgrade_python_toolset
     - pip install -r requirements/cython.txt
-    - pip install -r requirements/ci.txt
     - pip install -r requirements/towncrier.txt
     script:
     - towncrier --yes
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,6 +156,7 @@ jobs:
     install:
     - *upgrade_python_toolset
     - pip install -r requirements/cython.txt
+    - AIOHTTP_NO_EXTENSIONS=1 pip install -r requirements/ci.txt
     - pip install -r requirements/towncrier.txt
     script:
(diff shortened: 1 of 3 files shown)


asvetlov:
Should apply the same to multidict and yarl eventually.

SomberNight:
Is there any chance this could be backported to the 3.7.x branch?
I have a project where performance is not critical and I would like to explicitly opt out of C extensions but the only released versions of aiohttp that already have this PR are the 4.0.0 alphas. :(

multidict and yarl already have an equivalent released :)

SomberNight:
btw part of the reason I would not like to build the extensions is to make a binary of the larger project reproducible.
As aiohttp 3.7.x uses distutils to build the extension, I would need to use at least python 3.9 (to have https://github.com/python/cpython/commit/0d30ae1a03102de07758650af9243fd31211325a) to make the aiohttp C extensions reproducible.

aiohttp master no longer uses distutils (since https://github.com/aio-libs/aiohttp/pull/4079) but that has not been released either :(

webknjaz:
I don't think so, it seems like a major change. Also, the 3.7 release stream doesn't get updates, all the new things are supposed to go into the 3.8 now.

SomberNight:
sure, 3.8.x would be great too, I just meant any branch that gets releases in the not-too-far future :)

webknjaz:
I'm not sure sure that switching this behavior in a minor release is such a good idea.

PR #3832, opened 2019-06-09
Shorten too long test UNIX socket path

### LINKED ISSUE #3572: `OSError: AF_UNIX path too long` under MacOS

webknjaz:
## What do these changes do?

Make tmp unix sock path fit the kernel limits

## Related issue number

Fixes #3572

## Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,10 +1,24 @@
-import hashlib
+import os
+import socket
 import ssl
+import sys
+from hashlib import md5, sha256
+from pathlib import Path
+from tempfile import TemporaryDirectory
 
 import pytest
 import trustme
 
 pytest_plugins = ['aiohttp.pytest_plugin', 'pytester']
+
+IS_HPUX = sys.platform.startswith('hp-ux')
+"""Specifies whether the current runtime is HP-UX."""
+IS_LINUX = sys.platform.startswith('linux')
+"""Specifies whether the current runtime is HP-UX."""
+IS_UNIX = hasattr(socket, 'AF_UNIX')
+"""Specifies whether the current runtime is *NIX."""
+
+needs_unix = pytest.mark.skipif(not IS_UNIX, reason='requires UNIX sockets')
 
 
 @pytest.fixture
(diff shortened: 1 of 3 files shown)


tests/test_connector.py
```diff
@@ -43,8 +53,37 @@ def ssl_key():
...
+
+    Ref: https://unix.stackexchange.com/a/367012/27133
+    """
+
+    sock_path = str(tmp_path / 'socket.sock')
+    sock_path_len = len(sock_path.encode())
+
+    # exit-check to verify that it's correct and simplify debugging
+    # in the future
+    assert sock_path_len <= max_sock_len, (
```
asvetlov: Final version should use `pytest.skip()` instead of `assert` I guess

tests/test_connector.py
```diff
@@ -43,8 +53,37 @@ def ssl_key():
...
+        f'Suggested UNIX socket ({sock_path}) is {sock_path_len} bytes '
```
asvetlov: We didn't drop Python 3.5 support yet, f-strings are not allowed :)

tests/test_connector.py
webknjaz: Nope, I want it to fail loudly. I'm about to change the path generation above so that this should always be true and if it fails this means that something is really wrong.

tests/test_connector.py
webknjaz: Oh, good point.

webknjaz:
@asvetlov should be good to go now. Plz test on your machine as well :)

tests/conftest.py
```diff
@@ -55,4 +69,82 @@ def tls_certificate_pem_bytes(tls_certificate):
...
+    """Amount of bytes allocated for the UNIX socket path by OS kernel.
+
+    Ref: https://unix.stackexchange.com/a/367012/27133
+    """
+
+    sock_file_name = 'unix.sock'
+
+    root_tmp_dir = Path('/tmp').resolve()
+    os_tmp_dir = Path(os.getenv('TMPDIR', '/tmp')).resolve()
+    original_base_tmp_path = Path(tmp_path_factory.getbasetemp())
```
asvetlov: Is `Path()` convertion required?

asvetlov (commented):
The whole PR can be replaced with `PYTEST_DEBUG_TEMPROOT=/tmp` in CI config, isn't it?

tests/conftest.py
webknjaz: It's just nice for the unified API and paths comparison.

webknjaz:
@asvetlov I wanted something more generic and detached from CI.

tests/conftest.py
```diff
@@ -55,4 +69,82 @@ def tls_certificate_pem_bytes(tls_certificate):
...
+    """Amount of bytes allocated for the UNIX socket path by OS kernel.
+
+    Ref: https://unix.stackexchange.com/a/367012/27133
+    """
+
+    sock_file_name = 'unix.sock'
+
+    root_tmp_dir = Path('/tmp').resolve()
+    os_tmp_dir = Path(os.getenv('TMPDIR', '/tmp')).resolve()
+    original_base_tmp_path = Path(str(tmp_path_factory.getbasetemp()))
```
asvetlov: What is the reason for Path -> str -> Path conversion?
If there is something please describe it in a comment

asvetlov (commented):
Please consider the alternative:
1. Keep `unix_sockname` fixture. Drop all the rest: `need_unix` marker, `IS_` constants etc.
2. In `unix_sockname` always create a socket path as `/tmp/{uuid}.sock` on systems with `AF_UNIX`, otherwise call `pytest.skip()`. The code should be about 5 lines long.

What do you think?

tests/conftest.py
webknjaz: Yes, I do the comparison in a few places below. So I need a few `Path()` objects.
And `str()` in this line is only needed for Python 3.5 support which apparently cannot do this by its own.

webknjaz:
I think that the pytest's default place for the socket creation is better because it's better scoped. `/tmp` should be used as a fallback because it's harder to track the location when searching for stuff in the file system, during debugging, for example. So I'd like to keep a better-designed layout while having a fallback for corner cases.

webknjaz:
P.S. In general, I like the marker more as it's cleaner.

asvetlov (approved):
Ok, please merge when green.

I don't care too much honestly but can live with the PR if you did all work already :)

PR #3833, opened 2019-06-09
Drop a test that requires 1GB of memory

asvetlov:
This is not acceptable, sorry.
I have no choice but should drop this test entirely.
It also has a mock inside which is a bad sign -- the test doesn't test too much but the own mock.

### DIFF
--- a/tests/test_payload.py
+++ b/tests/test_payload.py
@@ -1,11 +1,9 @@
-import asyncio
 from io import StringIO
-from unittest import mock
 
 import pytest
 from async_generator import async_generator
 
-from aiohttp import payload, streams
+from aiohttp import payload
 
 
 @pytest.fixture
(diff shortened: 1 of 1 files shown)


webknjaz:
@asvetlov maybe just put some marker there and never run unless explicitly asked for via command line args?

asvetlov:
The question is: who and when will run this test?

asvetlov:
Let me continue.
If the answer is "nobody" or "some person somehow" -- let's just drop dead code.

webknjaz:
I think this can be useful for manual regression checks and troubleshooting with `git bisect`.

webknjaz:
Also, if we're going for multiple CIs it could be run on unimportant ones only or maybe only on nightlies.

asvetlov:
Agree in general but not for this particular case.
I can imagine a real test that requires a lot of memory or network resources.
But this case is all about mocking.
It is very subtle itself but requires 1GB RAM. Do you feel a bad smell?

webknjaz:
Okay, I can relate to that.

Issue #3853, opened 2019-06-18
web access log show utc time instead local time with timezone.


NewUserHa:
## Long story short
in web_.log.py:
```
    def _format_t(request: BaseRequest,
                  response: StreamResponse,
                  time: float) -> str:
        now = datetime.datetime.utcnow()
        start_time = now - datetime.timedelta(seconds=time)
        return start_time.strftime('[%d/%b/%Y:%H:%M:%S +0000]')
```
the logged time is fixed and is a utc time, not a local time. there's not a easy way to setting it.

## Expected behaviour

```
INFO:aiohttp.access:127.0.0.1 [18/Jun/2019:04:56:22 -8]
```
## Actual behaviour

```
INFO:aiohttp.access:127.0.0.1 [18/Jun/2019:04:56:22 +0000]
```

asvetlov:
UTC time is intentional.

webknjaz:
It's nice to make it customizable. I see it being useful during the development in particular.

@NewUserHa feel free to research improving this and send a PR.

NewUserHa:
nginx's access.log is `[16/Jun/2019:15:25:04 +0800]`

what about `making _format_t` to obey OS or python's timezone info? I doubt adding a parameter for customing will work.

asvetlov:
What is `apache` behavior?

NewUserHa:
`127.0.0.1 - frank [10/Oct/2000:13:55:36 -0700] "GET /apache_pb.gif HTTP/1.0" 200 2326` from https://httpd.apache.org/docs/2.4/logs.html

asvetlov:
Ok, I have no objections for changing defaults to use local time instead of UTC. The output should include a timezone offset like @NewUserHa mentioned: `[16/Jun/2019:15:25:04 +0800]`

NewUserHa:
I would like to PR this too. (though the other one still is not finished)...

what about obeying OS or python's timezone info? need import like pytz module for ease?

asvetlov:
Don't follow what is the difference between OS and python TZ info?
Please no `pytz` if we can avoid the library (I think we cannot need if for local timezone-aware time).

NewUserHa:
hmm.. maybe someone has different values for OS and python TZ info.
ok.

NewUserHa:
https://github.com/aio-libs/aiohttp/pull/3860

webknjaz:
Resolved by #3860.

PR #3860, opened 2019-06-21
Make access log use local time with timezone

NewUserHa:
## What do these changes do?

change access log use localtime and timezone instead all uctime

show localtime instead utctime if turn on access log

## Related issue number

https://github.com/aio-libs/aiohttp/issues/3853

## Checklist

### DIFF
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -150,7 +150,7 @@ class AccessLogger(AbstractAccessLogger):
     def _format_t(request: BaseRequest,
                   response: StreamResponse,
                   time: float) -> str:
-        tz = datetime.timezone(datetime.timedelta(seconds=-_time.timezone))
+        tz = datetime.timezone(datetime.timedelta(seconds=-timezone))
         now = datetime.datetime.now(tz)
         start_time = now - datetime.timedelta(seconds=time)
         return start_time.strftime('[%d/%b/%Y:%H:%M:%S %z]')
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -1,4 +1,5 @@
 import datetime
+import time as _time
 import functools
 import logging
 import os
@@ -149,9 +150,9 @@ class AccessLogger(AbstractAccessLogger):
     def _format_t(request: BaseRequest,
                   response: StreamResponse,
                   time: float) -> str:
-        now = datetime.datetime.utcnow()
+        now = datetime.datetime.now(datetime.timezone(datetime.timedelta(seconds=-_time.timezone)))
         start_time = now - datetime.timedelta(seconds=time)
-        return start_time.strftime('[%d/%b/%Y:%H:%M:%S +0000]')
+        return start_time.strftime('[%d/%b/%Y:%H:%M:%S %z]')
 
     @staticmethod
     def _format_P(request: BaseRequest,
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -1,9 +1,9 @@
 import datetime
-import time as _time
 import functools
 import logging
 import os
 import re
+import time as _time
 from collections import namedtuple
 from typing import Callable, Dict, Iterable, List, Tuple  # noqa
 
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -3,8 +3,8 @@ import functools
 import logging
 import os
 import re
-from time import timezone
 from collections import namedtuple
+from time import timezone
 from typing import Callable, Dict, Iterable, List, Tuple  # noqa
 
 from .abc import AbstractAccessLogger
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -3,7 +3,7 @@ import functools
 import logging
 import os
 import re
-import time as _time
+from time import timezone
 from collections import namedtuple
 from typing import Callable, Dict, Iterable, List, Tuple  # noqa
 
--- a/aiohttp/web_log.py
+++ b/aiohttp/web_log.py
@@ -150,7 +150,8 @@ class AccessLogger(AbstractAccessLogger):
     def _format_t(request: BaseRequest,
                   response: StreamResponse,
                   time: float) -> str:
-        now = datetime.datetime.now(datetime.timezone(datetime.timedelta(seconds=-_time.timezone)))
+        tz = datetime.timezone(datetime.timedelta(seconds=-_time.timezone))
+        now = datetime.datetime.now(tz)
         start_time = now - datetime.timedelta(seconds=time)
         return start_time.strftime('[%d/%b/%Y:%H:%M:%S %z]')
 
--- a/tests/test_web_log.py
+++ b/tests/test_web_log.py
@@ -90,7 +90,6 @@ def test_access_logger_atoms(monkeypatch, log_format, expected, extra) -> None:
     response = mock.Mock(headers={}, body_length=42, status=200)
     access_logger.log(request, response, 3.1415926)
     assert not mock_logger.exception.called
-    print(mock_logger.exception.__dict__)
 
     mock_logger.info.assert_called_with(expected, extra=extra)
 
--- a/tests/test_web_log.py
+++ b/tests/test_web_log.py
@@ -75,12 +75,9 @@ def test_access_logger_atoms(monkeypatch, log_format, expected, extra) -> None:
         @staticmethod
         def now(tz):
             return datetime.datetime(1843, 1, 1, 0, 30, tzinfo=tz)
-
-    def new_get_pid():
-        return 42
-    monkeypatch.setattr('datetime.datetime', PatchedDatetime)
-    monkeypatch.setattr('time.timezone', -28800)
-    monkeypatch.setattr("os.getpid", new_get_pid)
+    monkeypatch.setattr("datetime.datetime", PatchedDatetime)
+    monkeypatch.setattr("time.timezone", -28800)
+    monkeypatch.setattr("os.getpid", lambda: 42)
     mock_logger = mock.Mock()
     access_logger = AccessLogger(mock_logger, log_format)
     request = mock.Mock(headers={'H1': 'a', 'H2': 'b'},
--- a/tests/test_web_log.py
+++ b/tests/test_web_log.py
@@ -43,7 +43,7 @@ def test_access_logger_format() -> None:
     Ref: https://stackoverflow.com/a/46102240/595220
     """,  # noqa: E501
 )
-def test_access_logger_time(mocker) -> None:
+def test_access_logger_time(monkeypatch) -> None:
     now = datetime.datetime(1843, 1, 1, 0, 30)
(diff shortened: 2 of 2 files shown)


NewUserHa:
I read the test codes about web_log.py. but how to test this since there are expected and extras dicts in it?

aiohttp/web_log.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import os
 import re
+import time as _time
```
webknjaz: I don't think that you need to use an alias here. Please follow the style of the surrounding code.

webknjaz:
@NewUserHa google `pytest monkeypatch`

asvetlov:
`tests/test_web_log.py`
Something like:
```
def test_logger_time() -> None:
    mock_logger = mock.Mock()
    access_logger = AccessLogger(mock_logger, "%t")
    access_logger.log(None, None, 0)
    mock_logger.info.assert_called_with("[1/2/3:12:13:14 +0200]", extra={'request_start_time': ...})
```

No need for monkey-patching here

aiohttp/web_log.py
NewUserHa: it's needed. the time will be that argument if no alias

NewUserHa:
I know. the problem is:
there's a test already like that:
```python
def test_access_logger_atoms(mocker) -> None:
    utcnow = datetime.datetime(1843, 1, 1, 0, 30)
    mock_datetime = mocker.patch("aiohttp.datetime.datetime")
    mock_getpid = mocker.patch("os.getpid")
    mock_datetime.utcnow.return_value = utcnow
    mock_getpid.return_value = 42
    log_format = '%a %t %P %r %s %b %T %Tf %D "%{H1}i" "%{H2}i"'
    mock_logger = mock.Mock()
    access_logger = AccessLogger(mock_logger, log_format)
    request = mock.Mock(headers={'H1': 'a', 'H2': 'b'},
                        method="GET", path_qs="/path",
                        version=(1, 1),
                        remote="127.0.0.2")
    response = mock.Mock(headers={}, body_length=42, status=200)
    access_logger.log(request, response, 3.1415926)
    assert not mock_logger.exception.called
    expected = ('127.0.0.2 [01/Jan/1843:00:29:56 +0000] <42> '
                'GET /path HTTP/1.1 200 42 3 3.141593 3141593 "a" "b"')
    extra = {
        'first_request_line': 'GET /path HTTP/1.1',
        'process_id': '<42>',
        'remote_address': '127.0.0.2',
        'request_start_time': '[01/Jan/1843:00:29:56 +0000]',
        'request_time': 3,
        'request_time_frac': '3.141593',
        'request_time_micro': 3141593,
        'response_size': 42,
        'response_status': 200,
        'request_header': {'H1': 'a', 'H2': 'b'},
    }

    mock_logger.info.assert_called_with(expected, extra=extra)
```
I meant, can it test/assert the timezone? because request_start_time is fixed, then the expected always will be the same

NewUserHa:
well. Will adding tz to `datetime.datetime(1843, 1, 1, 0, 30)` of `test_access_logger_atoms` work?

asvetlov:
I guess yes, please extract checking of `%t` format into a separate test.
Should be easy to do.
`test_access_logger_atoms()` tests too much format options.
Now I see that monkey-patching is the easiest way to check your functionality, that's fine.

aiohttp/web_log.py
webknjaz:
```suggestion
from time import timezone
```

tests/test_web_log.py
```diff
@@ -43,13 +43,31 @@ def test_access_logger_format() -> None:
...
+    mock_datetime.now.return_value = now
+    mock_timezone.return_value = 28800
+    log_format = '%t'
+    mock_logger = mock.Mock()
+    access_logger = AccessLogger(mock_logger, log_format)
+    request = mock.Mock()
+    response = mock.Mock()
+    access_logger.log(request, response, 3.1415926)
+    assert not mock_logger.exception.called
+    expected = ('[01/Jan/1843:00:29:56 +0000]')
```
webknjaz: no need for braces here
```suggestion
    expected = '[01/Jan/1843:00:29:56 +0000]'
```

tests/test_web_log.py
```diff
@@ -43,13 +43,31 @@ def test_access_logger_format() -> None:
     Ref: https://stackoverflow.com/a/46102240/595220
     """,  # noqa: E501
 )
-def test_access_logger_atoms(mocker) -> None:
-    utcnow = datetime.datetime(1843, 1, 1, 0, 30)
+def test_access_logger_time(mocker) -> None:
```
webknjaz: plz use `monkeypatch` instead
```suggestion
def test_access_logger_time(monkeypatch) -> None:
```
`pytest-mock` has certain probems which I don't want us to hit.

tests/test_web_log.py
```diff
@@ -43,13 +43,31 @@ def test_access_logger_format() -> None:
...
+    log_format = '%t'
+    mock_logger = mock.Mock()
+    access_logger = AccessLogger(mock_logger, log_format)
+    request = mock.Mock()
+    response = mock.Mock()
+    access_logger.log(request, response, 3.1415926)
+    assert not mock_logger.exception.called
+    expected = ('[01/Jan/1843:00:29:56 +0000]')
+    extra = {
+        'request_start_time': '[01/Jan/1843:00:29:56 +0000]',
```
webknjaz: How about testing non-0000 timezone too?

tests/test_web_log.py
```diff
@@ -45,8 +45,8 @@ def test_access_logger_format() -> None:
 )
 def test_access_logger_time(monkeypatch) -> None:
     now = datetime.datetime(1843, 1, 1, 0, 30)
-    mock_datetime = mocker.patch("aiohttp.datetime.datetime")
-    mock_timezone = mocker.patch("aiohttp.timezone")
+    mock_datetime = monkeypatch.patch("aiohttp.datetime.datetime")
```
webknjaz: I believe you should use `setattr()`: https://docs.pytest.org/en/latest/monkeypatch.html

tests/test_web_log.py
```diff
@@ -43,13 +43,31 @@ def test_access_logger_format() -> None:
     Ref: https://stackoverflow.com/a/46102240/595220
     """,  # noqa: E501
 )
-def test_access_logger_atoms(mocker) -> None:
-    utcnow = datetime.datetime(1843, 1, 1, 0, 30)
-    mock_datetime = mocker.patch("aiohttp.datetime.datetime")
+def test_access_logger_time(monkeypatch) -> None:
```
webknjaz: Instead of having a lot of copy-paste, just use `@pytest.mark.parametrize()`. Also, you now have the other test method without a proper skip marker.

tests/test_web_log.py
NewUserHa: mistake

tests/test_web_log.py
NewUserHa: keep all pieces, only replace `extra` and `expected` dicts?

webknjaz: yep

tests/test_web_log.py
NewUserHa: monkeypatch can't set attributes of built-in/extension type 'datetime.datetime'.
is there a nice way?

NewUserHa:
```
>       assert not mock_logger.exception.called
E       AssertionError: assert not True
```
what's wrong with this
https://docs.python.org/3/library/unittest.mock.html `mock.Mock()` dont have this class

tests/test_web_log.py
```diff
@@ -43,13 +43,42 @@ def test_access_logger_format() -> None:
...
     """,  # noqa: E501
 )
-def test_access_logger_atoms(mocker) -> None:
-    utcnow = datetime.datetime(1843, 1, 1, 0, 30)
-    mock_datetime = mocker.patch("aiohttp.datetime.datetime")
-    mock_getpid = mocker.patch("os.getpid")
-    mock_datetime.utcnow.return_value = utcnow
-    mock_getpid.return_value = 42
-    log_format = '%a %t %P %r %s %b %T %Tf %D "%{H1}i" "%{H2}i"'
+@pytest.fixture
```
webknjaz: I'm pretty sure that skip marker doesn't work on fixtures.

tests/test_web_log.py
```diff
@@ -43,13 +43,42 @@ def test_access_logger_format() -> None:
...
+def mocker(monkeypatch, expected, extra):
```
webknjaz: Please use a specific name for the fixture. Moker is useless, misleading and shadows the existing fixture from another plugin.

tests/test_web_log.py
NewUserHa: but if use parameterizing, it seems to have to use fixtrue

webknjaz: Yes, but I'm talking about that the marked is supposed to skip things depending on the condition (PyPy). And if you apply it to fixtures, not test functions it just doesn't have any effect meaning it will never skip any tests.

tests/test_web_log.py
```diff
@@ -43,13 +43,44 @@ def test_access_logger_format() -> None:
...
+    class PatchedDatetime(datetime.datetime):
+        @staticmethod
+        def now(tz):
+            return datetime.datetime(1843, 1, 1, 0, 30, tzinfo=tz)
+
+    def new_get_pid():
+        return 42
+    monkeypatch.setattr('datetime.datetime', PatchedDatetime)
+    monkeypatch.setattr('time.timezone', -28800)
+    monkeypatch.setattr("os.getpid", new_get_pid)
```
webknjaz: Don't mix up the quotation style. Also, this can be simply
```suggestion
    monkeypatch.setattr('os.getpid', lambda: 42)
```

tests/test_web_log.py
NewUserHa: ok. I need to check some documents.

webknjaz: Currently, it's okay since you removed the fixture.

NewUserHa: but...

webknjaz:
It should be good to merge once CI jobs show up as green.

webknjaz:
Thanks, @NewUserHa!

NewUserHa:
thanks for your review and merge.
but why is `test_access_logger_atoms` SKIPPED in release 3.5 with running pytest against the 3.5 `test_web_log.py`?
need change log?

webknjaz:
Because you submitted the PR against devel which is future 4.0. 3.5 is in its own stable branch.

And yes, the changelog fragment is needed. We just missed it. So please submit a follow-up PR with it.

NewUserHa:
release 3.5 against test_web_log.py 3.5.

https://github.com/aio-libs/aiohttp/pull/3863

chansonzhang:
I use version 3.7.3 and the timestamp in access log still in UTC timezone

asvetlov:
Thanks for the reminder.  We missed to backport the PR.

#5351 should land at 3.8

PR #3861, opened 2019-06-21
add kwargs into run_app func.

serjshevchenko:
Fix error "Got more than 8190 bytes (11194) when reading Header value is too long."


asvetlov:
I understand your motivation but disagree with implementation.
I think we should go in the opposite direction: get rid of `**kwargs`` usage but enumerate and pass all parameters explicitly.
It increases the code readability and provides better type annotations.

asvetlov:
Closing by the lack of interest by @serjshevchenko

PR #3889, opened 2019-07-05
Raise 400 Bad Request on server-side `await request.json()` if incorrect content-type received

### LINKED ISSUE #2174: Insecure way to handle json requests

atemate:
## What do these changes do?

make server-side request fail with `HTTPBadRequest` on `await request.json()` with wrong content-type for json.

This PR is an alternative implementation of https://github.com/aio-libs/aiohttp/pull/2177.

The following code now raises HTTP 400:
```
async def handler(request):
    await request.json()  # raises HTTP 400
```

## Related issue number

closes https://github.com/aio-libs/aiohttp/issues/2174
## Checklist


asvetlov (approved):
I'm curious why `aiohttp.svg` is changed?

asvetlov (changes requested):
LGTM except for logo file

aiohttp.svg
```diff
@@ -0,0 +1,2144 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
webknjaz: This is not used/needed anywhere. Seems totally unrelated.

aiohttp/helpers.py
```diff
@@ -106,6 +106,8 @@ def all_tasks(
 
 coroutines._DEBUG = old_debug  # type: ignore
 
+json_re = re.compile(r'^application/(?:[\w.+-]+?\+)?json')
```
webknjaz, line 109: > `+?`

This is probably not what you want.

aiohttp/helpers.py
```diff
@@ -360,6 +362,13 @@ def content_disposition_header(disptype: str,
     return value
 
 
+def is_expected_content_type(response_content_type: str,
+                             expected_content_type: str) -> bool:
+    if expected_content_type == 'application/json':
```
webknjaz, line 367: I think this should go to a constant. Now it's hardcoded all over the place which is not maintainable.

webknjaz (changes requested):

aiohttp/helpers.py
asvetlov: I disagree.
The string is better readable than constant.
Also, we never change this line in the future.
What maintenance problem are you talking about?

aiohttp/helpers.py
asvetlov: This regexp pattern is used by client code for years.
What is wrong with it?

aiohttp/helpers.py
atemate: I think, the general Python code style welcomes usage of some well-known constants as literals and not const variables. For example, everyone knows that 404 is the HTTP code for Not Found, as well as `application/json` is the name for a mime type that will not be changed in future. So I would leave it as a string literal (same as the default parameter of `content_type` in `ClientResponse.json()`)

aiohttp/helpers.py
atemate: yes, sorry for not pointing out to the fact that we simply moved `json_re` from this place: https://github.com/aio-libs/aiohttp/pull/3889/files#diff-2fa8ab44317df4fbb7864536a6794d7aL83

aiohttp.svg
atemate: I'm totally sorry for this extra file, this is a mistake of mine :/

atemate: deleted it.

asvetlov (approved):

asvetlov:
@webknjaz please review again

aiohttp/helpers.py
webknjaz: @asvetlov `+` means one or more, `?` means `0` or `1`. I don't get this syntax. Does it essentially mean `*` (`0` or more times)?

webknjaz: Oh, that's lazy mode. Nevermind.

webknjaz (approved):

asvetlov:
thanks!

Issue #3906, opened 2019-07-15
Consider removing assert about HTTP Reason-Phrase being always present


WouldYouKindly:
Hi!

`aiohttp` has the following code:

```python
def raise_for_status(self) -> None:
    if 400 <= self.status:
        assert self.reason  # always not None for started response
        self.release()
        raise ClientResponseError(
            self.request_info,
            self.history,
            status=self.status,
            message=self.reason,
            headers=self.headers)
```

This assertion fails if the server omits `Reason-Phrase`, i.e. if response looks like `HTTP/1.1 404` (no `Not Found`).

[HTTP spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html) is not clear about whether reason must be provided, but e.g. this answer suggests that it may not be: https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase

Maybe aiohttp should be more permissive in this case. For me, the problem is that [hoverfly](https://github.com/SpectoLabs/hoverfly) drops the reason phrase, which leads to aiohttp raising `AssertionError`.  Reason can be set to an empty string if it is not found in response.

asvetlov:
That's interesting.
I expected that http parsers (both cython and pure python versions) set reason to empty string if omitted.
If it is not -- the PR for parsers fix is welcome.

guyskk:
I find some website response without reason-phrase. I checked response.reason is None, is empty string OK? (assert "" still fail)

```python
import aiohttp
import asyncio

url = 'http://pyconcn.qiniudn.com/images/volunteer/zhouqi.png'

async def main():
    async with aiohttp.ClientSession() as client:
        async with client.get(url) as resp:
            print(resp.status)
            resp.raise_for_status()

loop = asyncio.get_event_loop()
loop.run_until_complete(main())
>>>
530
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/kk/.pyenv/versions/3.6.5/lib/python3.6/asyncio/base_events.py", line 468, in run_until_complete
    return future.result()
  File "<stdin>", line 5, in main
  File "/Users/kk/.pyenv/versions/rssant/lib/python3.6/site-packages/aiohttp/client_reqrep.py", line 935, in raise_for_status
    assert self.reason  # always not None for started response
AssertionError
```

achimnol:
I just encountered this issue while doing integration with one of my customer site.
Hoping to see the next release soon!

PR #3909, opened 2019-07-16
Set reason to an empty string if it is missing (_http_parser.pyx)

WouldYouKindly:
Resolve https://github.com/aio-libs/aiohttp/issues/3906

Test that if reason is missing, it is set to an empty string rather than any falsy value.

Make cython parser set it to en empty string if it is missing. `self._reason = self._reason or ''` is because `_on_status_complete` is called from two places and when it's called the second time, reason may already be set.


WouldYouKindly:
@asvetlov done.

PR #3919, opened 2019-07-19
Configure dependabot

asvetlov:
Use yaml config instead of web site settings

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,12 @@
+version: 1
+update_configs:
+  # Keep master up to date
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "live"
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -0,0 +1,12 @@
+version: 1
+update_configs:
+  # Keep master up to date
+  - package_manager: "python"
+    directory: "/"
+    update_schedule: "live"
+    target_branch: "master"
+  # Keep 3.5 branch up to date
+  - package_manager: "docker"
```
webknjaz: Docker? Why?
```suggestion
  - package_manager: "python"
```

asvetlov: Sorry, copy-paste from another project

Issue #3927, opened 2019-07-22
Update mypy to latest version and support mypy --strict


AMDmi3:
## Long story short

Following discussion in #3868, here's an issue to track a progress of cleaning up errors reported by newer mypy and/or mypy in strict mode.

## Expected behaviour

aiohttp is clean with latest mypy (0.720 ATOW) in `--strict` mode.

## Actual behaviour

There are some errors after updating mypy, and a lot of errors after enabling strict mode.

## Steps to reproduce

Run `mypy --strict aiohttp` with latest mypy

## Your environment

FreeBSD 12.0 amd64, python 3.7, mypy 0.720

AMDmi3:
(I'll try to fix at least a part of this)

PR #3928, opened 2019-07-22
Typing updates

AMDmi3:
## What do these changes do?

Bump mypy version to 0.720 and fix/silence new mypy issues.

No

## Related issue number

#3927

## Checklist

- [x] Unit tests for the changes exist (no new code or behavior change, should be covered with existing tests)
- [ ] Add a new news fragment into the `CHANGES` folder (may do after more changes are merged)

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -180,6 +180,8 @@ class ClientSession:
         '_request_class', '_response_class',
         '_ws_response_class', '_trace_configs')
 
+    _default_headers: 'CIMultiDict[str]'
+
     def __init__(self, *, connector: Optional[BaseConnector]=None,
                  cookies: Optional[LooseCookies]=None,
                  headers: Optional[LooseHeaders]=None,
@@ -239,10 +241,9 @@ class ClientSession:
 
         # Convert to list of tuples
         if headers:
-            headers = CIMultiDict(headers)
+            self._default_headers = CIMultiDict(headers)
         else:
-            headers = CIMultiDict()
-        self._default_headers = headers
+            self._default_headers = CIMultiDict()
         if skip_auto_headers is not None:
             self._skip_auto_headers = frozenset([istr(i)
                                                  for i in skip_auto_headers])
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -361,7 +361,7 @@ class ClientRequest:
             if isinstance(value, Morsel):
                 # Preserve coded_value
                 mrsl_val = value.get(value.key, Morsel())
-                mrsl_val.set(value.key, value.value, value.coded_value)  # type: ignore  # noqa
+                mrsl_val.set(value.key, value.value, value.coded_value)  # noqa
                 c[name] = mrsl_val
             else:
                 c[name] = value  # type: ignore
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -171,7 +171,7 @@ class ClientRequest:
     body = b''
     auth = None
     response = None
-    response_class = None
+    response_class: Type['ClientResponse']
 
     _writer = None  # async task for streaming data
     _continue = None  # waiter future for '100 Continue' response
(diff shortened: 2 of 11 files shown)


aiohttp/client.py
```diff
@@ -178,6 +178,8 @@ class ClientSession:
         '_request_class', '_response_class',
         '_ws_response_class', '_trace_configs')
 
+    _default_headers: CIMultiDict[str]
```
asvetlov: aiohttp still supports Python 3.5, this syntax doesn't work.

I'm open for discussing the drop of Python 3.5 in aiohttp 4.0 though.
If you want to proceed in this direction please create a new issue for discussion.

Otherwise please use Python 3.5 compatible syntax in the PR

AMDmi3:
Sorry, I'm not really interested in legacy support or participating in related discussions. If this cannot be merged right away, it should either hang until it can be merged, or have only compatible bits cherry picked, or just be discarded.

webknjaz:
@AMDmi3 it fails even under Python 3.7: https://ci.appveyor.com/project/aio-libs/aiohttp/builds/26164328/job/0nn5x41jb9ow927u#L1023

webknjaz:
Curious what Towncrier has to do with this... https://travis-ci.com/aio-libs/aiohttp/jobs/218370235#L527

asvetlov:
@webknjaz I don't know but towncrier wants to *import* the served package for some reason.
Maybe for the package version extraction?

webknjaz:
Ah right.. Package version. Makes sense. AFAIR you can use `--version` if you have an external mechanism of extracting it.
But the thing which makes me confused is that it barks at the syntax error while running in the Python 3.7 env.

AMDmi3:
> it fails even under Python 3.7

Actually I think there's problem with typing support in `multidict` module which causes this. Not sure how to fix that properly.

```
% cat 1.py
from multidict import CIMultiDict
foo: CIMultiDict[str] = CIMultiDict()
% mypy --strict 1.py
% python3 1.py
  File "1.py", line 2, in <module>
    foo: CIMultiDict[str] = CIMultiDict()
TypeError: 'type' object is not subscriptable
```
```
% cat 2.py
from multidict import CIMultiDict
foo: CIMultiDict = CIMultiDict()
% mypy --strict 2.py
2.py:2: error: Missing type parameters for generic type
% python3 2.py 
```

asvetlov:
Use forwarded type declaration (string) :
```
foo: 'CIMultiDict[str]' = CIMultiDict()
```
It is not multidict error, `asyncio.Future[int]` has the same limitation.
Python 3.7 has `from __future__ import annotations` (see https://www.python.org/dev/peps/pep-0563/ for details).
Since the feature is the mainstream I don't think that multidict should fix the issue itself by providing `__getitem__` in own metaclass.

antonagestam:
@asvetlov Isn't the issue that the C implementation of `CIMultiDict` lacks a `__class_getitem__` method? If I import it using `from multidict._multidict_py import CIMultiDict`, using `CIMultiDict[str]` does not trigger an error, since the Python implementation inherits from the abstract base class.

See https://github.com/MagicStack/immutables/pull/13 and https://gitter.im/python/typing?at=5d25bd59cdb70805c96e239f for a similar issue.

asvetlov:
`__class_getitem__` works for Python 3.7+ only, aiohttp is supposed to support older python versions

antonagestam:
@asvetlov Yes, but currently using `CIMultiDict[str]` will break at runtime. The only way around that is to make the C implementation subscriptable. Doing that doesn't break support for older Python versions, but enables the new annotation syntax for users of Python 3.5+.

asvetlov:
The way is using forward declarations: `"CIMultiDict[str]"`

antonagestam:
What do you mean with "the way"? In my opinion, a _better_ way for the multidict project would be to implement `__class_getitem__` and letting the user choose whether to use the string syntax or not? But really this discussion ought to happen in the multidict repo.

asvetlov:
> But really this discussion ought to happen in the multidict repo.

Exactly. aiohttp should use strings here.

AMDmi3:
Switched to string annotations, python 3.6+ build seem to be fixed

asvetlov:
Outdated, the master supports newer mypy without errors

PR #3932, opened 2019-07-23
Drop deprecated fixtures

asvetlov:


asvetlov:
@webknjaz thanks for review

PR #3948, opened 2019-07-26
Forbid changing frozen app properties

asvetlov:
The following code was raising deprecation and forbidden now.

```
app.freeze()
app['foo'] = 'bar'
```


kxepal:
What is the recommendation to do in this case: there is an app property, but it only could be initialized in async-way, so it has to be put in onstartup signal handler where app is frozen? Do some mutable placeholders?

PR #3955, opened 2019-08-02
Replace all tmpdir fixtures with tmp_path (#3551)

### LINKED ISSUE #3551: Replace tmpdir fixture with tmp_path in tests

vaneseltine:
tmp_path is the replacement fixture in pytest for tmpdir; tmp_path
uses the builtin pathlib.Path class. As it says on the tin, this
commit replaces every instance of tmpdir in the test suite with
tmp_path. Aside from s/tmpdir/tmp_path/ this also required changing
instances of `tmpdir.join(foo)` to `tmp_path / foo`.

This is intended to comprehensively address and close #3551, and
should have no side effects. This does not affect end users.

### DIFF
--- a/tests/test_web_sendfile_functional.py
+++ b/tests/test_web_sendfile_functional.py
@@ -319,18 +319,18 @@ def test_static_route_path_existence_check() -> None:
         web.StaticResource("/", nodirectory)
 
 
-async def test_static_file_huge(aiohttp_client, tmpdir) -> None:
+async def test_static_file_huge(aiohttp_client, tmp_path) -> None:
     filename = 'huge_data.unknown_mime_type'
 
     # fill 20MB file
-    with tmpdir.join(filename).open('w') as f:
+    with (tmp_path / filename).open('w') as f:
         for i in range(1024*20):
             f.write(chr(i % 64 + 0x20) * 1024)
 
-    file_st = os.stat(str(tmpdir.join(filename)))
+    file_st = os.stat(str((tmp_path / filename)))
 
     app = web.Application()
-    app.router.add_static('/static', str(tmpdir))
+    app.router.add_static('/static', str(tmp_path))
     client = await aiohttp_client(app)
 
     resp = await client.get('/static/'+filename)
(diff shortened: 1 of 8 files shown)


asvetlov (approved):
Thanks!

tests/test_web_sendfile_functional.py
```diff
@@ -802,7 +802,7 @@ def test_static_route_path_existence_check() -> None:
         tr = request.transport
         sock = tr.get_extra_info('socket')
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 1024)
-        ret = web.FileResponse(pathlib.Path(str(tmpdir.join(filename))))
+        ret = web.FileResponse(pathlib.Path(str((tmp_path / filename))))
```
gyermolenko, line 805: here and in a few other places now the path is turned into string and than back to Path.

asvetlov: Sorry, I've missed this.
Please feel free to make a PR for the fix.
The problem is minor though if all lines belong to tests only

vaneseltine: Good point. I'm working on a PR now to more comprehensively address path handling across the test suite.

Mostly that's replacing `os.path` calls with their appropriate `pathlib.Path` counterparts.

There are also a couple of uses of ad-hoc temporary directories that I'm updating to `tmp_path`.

vaneseltine: Ironically, to support 3.5 some `pathlib.Path(str(tmp_path)))` turns out to be necessary. 

pytest uses `pathlib2` to provide `tmp_path` for Python 3.5, which is mostly fine but it does  create a couple of corner cases of incompatibility where `pathlib2.Path` fails isinstance checks that demand `str` or `PurePath`. See pytest-dev/pytest/issues/5017

PR #3957, opened 2019-08-02
Improve test suite handling of paths, temp files

vaneseltine:
## What do these changes do?

This updates most uses of `os.path` to instead use `pathlib.Path`.
Relatedly, and following up from #3955 (which replaced pytest's `tmpdir`
fixture with `tmp_path`), this removes most ad-hoc tempfile creation in
favor of the `tmp_path` fixture. Following conversion, unnecessary `os`
and `tempfile` imports were removed.

Most pathlib changes involve straightforward changes from `os` functions
such as `os.mkdir` or `os.path.abspath` to their equivalent methods in
`pathlib.Path`.

Changing ad-hoc temporary path to `tmp_path` involved removing the
`tmp_dir_path` fixture and replacing its functionality with `tmp_path`
in `test_save_load` and `test_guess_filename_with_tempfile`.

On `test_static_route_user_home` function:

* I think that the intention of this test is to ensure that aiohttp
correctly expands the home path if passed in a string. I refactored it
to `pathlib.Path` and cut out duplication of `relative_to()` calls.
But if it's not doing anything but expanding `~`, then it's testing the
functionality of `pathlib.Path`, not aiohttp.

On `unix_sockname` fixture:

This fixture uses `tempfile.TemporaryDirectory`. Because it's a somewhat
complicated fixture used across multiple test modules, I left it as-is
for now.

On `str(tmp_path)` and even `pathlib.Path(str(tmp_path))`:

pytest uses `pathlib2` to provide `tmp_path` for Python 3.5 (only).
This is mostly fine but it fails on a couple of corner cases, such as
`os.symlink()` which blocks all but `str` and `PurePath` via isinstance
type checking. In several cases, this requires conversion to string or
conversion to string and then into `pathlib.Path` to maintain code
compatibility. See: pytest-dev/pytest/issues/5017

These changes only affect the test suite and have no impact on the end user.

## Related issue number

This is intended to address discussion following the simplistic changes from tmpdir to tmp_path of #3955.

## Checklist

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -47,7 +47,8 @@ def test_parse_mimetype(mimetype, expected) -> None:
 
 def test_guess_filename_with_tempfile(tmp_path) -> None:
     file_path = tmp_path / 'test_guess_filename'
-    assert (helpers.guess_filename(file_path, 'no-throw') is not None)
+    with file_path.open('w+b') as fp:
+        assert (helpers.guess_filename(fp, 'no-throw') is not None)
 
 
 # ------------------- BasicAuth -----------------------------------
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -3,7 +3,6 @@ import base64
 import gc
 import os
 import platform
-import tempfile
 from unittest import mock
 
 import pytest
@@ -46,9 +45,9 @@ def test_parse_mimetype(mimetype, expected) -> None:
 
 # ------------------- guess_filename ----------------------------------
 
-def test_guess_filename_with_tempfile() -> None:
-    with tempfile.TemporaryFile() as fp:
-        assert (helpers.guess_filename(fp, 'no-throw') is not None)
+def test_guess_filename_with_tempfile(tmp_path) -> None:
+    file_path = tmp_path / 'test_guess_filename'
+    assert (helpers.guess_filename(file_path, 'no-throw') is not None)
 
 
 # ------------------- BasicAuth -----------------------------------
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -45,10 +45,19 @@ def test_parse_mimetype(mimetype, expected) -> None:
 
 # ------------------- guess_filename ----------------------------------
 
-def test_guess_filename_with_tempfile(tmp_path) -> None:
+def test_guess_filename_with_file_object(tmp_path) -> None:
     file_path = tmp_path / 'test_guess_filename'
     with file_path.open('w+b') as fp:
         assert (helpers.guess_filename(fp, 'no-throw') is not None)
+
+
+def test_guess_filename_with_path(tmp_path) -> None:
+    file_path = tmp_path / 'test_guess_filename'
+    assert (helpers.guess_filename(file_path, 'no-throw') is not None)
+
+
+def test_guess_filename_with_default() -> None:
+    assert (helpers.guess_filename(None, 'no-throw') == 'no-throw')
 
 
 # ------------------- BasicAuth -----------------------------------
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -1,4 +1,3 @@
-import os
 import pathlib
 import re
 from collections.abc import Container, Iterable, Mapping, MutableMapping, Sized
@@ -427,7 +426,7 @@ def test_add_static_append_version_follow_symlink(router, tmp_path) -> None:
     """
     symlink_path = tmp_path / 'append_version_symlink'
     symlink_target_path = pathlib.Path(__file__).parent
-    os.symlink(str(symlink_target_path), str(symlink_path), True)
+    pathlib.Path(str(symlink_path)).symlink_to(str(symlink_target_path), True)
 
     # Register global static route:
     resource = router.add_static('/st', str(tmp_path), follow_symlinks=True,
@@ -450,7 +449,7 @@ def test_add_static_append_version_not_follow_symlink(router,
     symlink_path = tmp_path / 'append_version_symlink'
     symlink_target_path = pathlib.Path(__file__).parent
 
-    os.symlink(str(symlink_target_path), str(symlink_path), True)
+    pathlib.Path(str(symlink_path)).symlink_to(str(symlink_target_path), True)
 
     # Register global static route:
     resource = router.add_static('/st', str(tmp_path), follow_symlinks=False,
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -51,7 +51,7 @@ def fill_routes(router):
         route2 = router.add_route('GET', '/variable/{name}',
                                   make_handler())
         resource = router.add_static('/static',
-                                     os.path.dirname(aiohttp.__file__))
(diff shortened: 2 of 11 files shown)


vaneseltine:
Ah, I see the url issue in AppVeyor. I'll have to fire up my Windows box.

vaneseltine:
P.S. I'm not sure what I or the chronographer are doing wrong to set off Timeline Protection, but please let me know if I should do something to fix it.

webknjaz:
@vaneseltine all's fine with that check now.

tests/test_helpers.py
```diff
@@ -46,9 +45,9 @@ def test_parse_mimetype(mimetype, expected) -> None:
 
 # ------------------- guess_filename ----------------------------------
 
-def test_guess_filename_with_tempfile() -> None:
-    with tempfile.TemporaryFile() as fp:
-        assert (helpers.guess_filename(fp, 'no-throw') is not None)
+def test_guess_filename_with_tempfile(tmp_path) -> None:
+    file_path = tmp_path / 'test_guess_filename'
+    assert (helpers.guess_filename(file_path, 'no-throw') is not None)
```
webknjaz, line 56: This is wrong: the original test was checking a file-like input and the update checks a path-like object input.
Don't change the original meaning of the tests please.

tests/test_web_urldispatcher.py
```diff
@@ -221,33 +199,33 @@ def teardown():
...
 
-async def test_access_symlink_loop(tmp_dir_path, aiohttp_client) -> None:
+async def test_access_symlink_loop(tmp_path, aiohttp_client) -> None:
     """
     Tests the access to a looped symlink, which could not be resolved.
     """
-    my_dir_path = os.path.join(tmp_dir_path, 'my_symlink')
-    os.symlink(my_dir_path, my_dir_path)
+    my_dir_path = tmp_path / 'my_symlink'
+    os.symlink(str(my_dir_path), str(my_dir_path))
```
webknjaz: You could use pathlib here as well: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

tests/test_web_urldispatcher.py
```diff
@@ -85,26 +65,26 @@ def teardown():
...
-    my_file_path = os.path.join(my_dir_path, 'my_file_in_dir')
-    with open(my_file_path, 'w') as fw:
+    my_file_path = my_dir_path / 'my_file_in_dir'
+    with my_file_path.open('w') as fw:
         fw.write(data)
 
-    my_symlink_path = os.path.join(tmp_dir_path, 'my_symlink')
-    os.symlink(my_dir_path, my_symlink_path)
+    my_symlink_path = tmp_path / 'my_symlink'
+    os.symlink(str(my_dir_path), str(my_symlink_path))
```
webknjaz: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

tests/test_urldispatch.py
```diff
@@ -442,17 +441,19 @@ def test_add_static_append_version_follow_symlink(router, tmpdir) -> None:
...
     """
-    tmp_dir_path = str(tmpdir)
-    symlink_path = os.path.join(tmp_dir_path, 'append_version_symlink')
-    symlink_target_path = os.path.dirname(__file__)
-    os.symlink(symlink_target_path, symlink_path, True)
+
+    symlink_path = tmp_path / 'append_version_symlink'
+    symlink_target_path = pathlib.Path(__file__).parent
+
+    os.symlink(str(symlink_target_path), str(symlink_path), True)
```
webknjaz: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

tests/test_urldispatch.py
```diff
@@ -415,23 +415,22 @@ def test_add_static_append_version_non_exists_file(router) -> None:
...
     """
     Tests the access to a symlink, in static folder with apeend_version
     """
-    tmp_dir_path = str(tmpdir)
-    symlink_path = os.path.join(tmp_dir_path, 'append_version_symlink')
-    symlink_target_path = os.path.dirname(__file__)
-    os.symlink(symlink_target_path, symlink_path, True)
+    symlink_path = tmp_path / 'append_version_symlink'
+    symlink_target_path = pathlib.Path(__file__).parent
+    os.symlink(str(symlink_target_path), str(symlink_path), True)
```
webknjaz: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

tests/test_helpers.py
vaneseltine: I see, sorry about that. 

Would it be helpful to instead add other guess_filename tests?

tests/test_web_urldispatcher.py
vaneseltine: I have not been able to get that to work due to pytest-dev/pytest/issues/5017

In short: for Python 3.5, pytest uses `pathlib2` (https://pypi.org/project/pathlib2/) to provide `tmp_path` along with some of the 3.6-style improvements. As a side effect, symlink methods (both `os` and `pathlib`) fail due to isinstance type checking.

tests/test_web_urldispatcher.py
vaneseltine: I have not been able to get that to work in 3.5 due to pytest-dev/pytest/issues/5017

vaneseltine: See https://github.com/aio-libs/aiohttp/pull/3957#discussion_r311324192

tests/test_urldispatch.py
vaneseltine: See https://github.com/aio-libs/aiohttp/pull/3957#discussion_r311324192

tests/test_urldispatch.py
vaneseltine: See https://github.com/aio-libs/aiohttp/pull/3957#discussion_r311324192

tests/test_helpers.py
vaneseltine: I corrected this to use `.open('w+b')`, which creates a file object in the same mode that `tempfile.TemporaryFile()` did.

webknjaz: Yes. The extra test may be useful.

tests/test_web_urldispatcher.py
webknjaz: I mean
```suggestion
    pathlib.Path(str(my_dir_path)).symlink_to(str(my_dir_path))
```

tests/test_helpers.py
vaneseltine: OK -- added two tests (one for path, one a fallback to default).

tests/test_web_urldispatcher.py
vaneseltine: Understood! I've fixed all four places in 5928b41

tests/test_helpers.py
vaneseltine: Added path as a separate test, as well as a test for default fallback, in 5928b41

asvetlov:
@webknjaz please merge when ready and you have no objections/notes

Issue #3380, opened 2018-11-05
aiohttp server hangs when nginx try to mixin plain http request with failed websocket connection upgrade in one tcp stream.


zTrix:
## Long story short

aiohttp server hangs when nginx try to mixin plain http request with failed websocket connection upgrade in one tcp stream.

The following tcpdump capture shows what happens in detail.

![deepinscreenshot_select-area_20181105120935](https://user-images.githubusercontent.com/471934/47977582-df21a900-e0f3-11e8-9153-41ecec704508.png)

## Expected behaviour

Server returns something.

## Actual behaviour

Server hangs forever without closing tcp connection. Nginx return 504 Gateway timeout to end user.

## Steps to reproduce

1. write a simple api server using aiohttp (version 3.4.4 in my case), implement some /api/xxx for testing, leave /ws unimplemented.
2. using nginx to reverse proxy to this api server.
3. user open browser targeting a web page which contains both ws request and api request. Then api query will gateway timeout.

![deepinscreenshot_select-area_20181105122712](https://user-images.githubusercontent.com/471934/47978021-493b4d80-e0f6-11e8-9a8c-8cc234193467.png)

Example nginx.conf:

```
    ...
    location /api/ {
        proxy_pass http://server_backend;
        proxy_http_version 1.1;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    }

    location /ws {
        proxy_pass http://server_backend;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
    }
```

aiohttp server serves `/api/` request, but leave `/ws` unimplemented.

Then this issue could be reproduced.

## Your environment

Docker: nginx:latest
Docker: aiohttp3.4.4 on python:3.7-alpine3.7

zTrix:
The weird thing here is why nginx still sends http/1.0 request to upstream backend server while `proxy_http_version 1.1;` specified in nginx.conf.

kxepal:
Your nginx configuration works fine for me for a bit modified `web_ws.py` example. Could you share the code which reproduces the issue? And full nginx config could be useful as well.

zTrix:
[ws-api-mixed-in-one-tcp-stream-hang-bug-demo.tar.gz](https://github.com/aio-libs/aiohttp/files/2547315/ws-api-mixed-in-one-tcp-stream-hang-bug-demo.tar.gz)

[...]

PR #3971, opened 2019-08-09
Fix failed websocket handshake leaving connection hanging

gjcarneiro:
## What do these changes do?

It fixes the bug described in #3380, caused by this sequence of events:

1. clients initiates a websocket upgrade
2. server returns a normal HTTP response instead of a websocket response
3. client, with the same connection open, sends further HTTP requests.

The fix involves: if we detect a normal HTTP response instead of a websocket response, in web_protocol,  then we:
1. Reset _upgraded flag in the HttpParser to the default value (False)
2. Reset the _uprade flag in web_protocol back to False;
3. We feed any buffered data, originally intended for the websocket payload parser, back to the the http parser, so that the next normal  HTTP request can be processed.

This way, if websocket request results in a non-websocket response, we go back to normal HTTP mode, for that socket.

## Related issue number

#3380

## Checklist

### DIFF
--- a/CHANGES/3380.bugfix
+++ b/CHANGES/3380.bugfix
@@ -1,0 +1,1 @@
+Fix failed websocket handshake leaving connection hanging.
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -533,6 +533,9 @@ cdef class HttpParser:
         else:
             return messages, False, b''
 
+    def set_upgraded(self, val):
+        self._upgraded = val
+
 
 cdef class HttpRequestParser(HttpParser):
 
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -411,6 +411,13 @@ class HttpParser(abc.ABC):
 
         return (headers, raw_headers, close_conn, encoding, upgrade, chunked)
 
+    def set_upgraded(self, val: bool) -> None:
+        """Set connection upgraded (to websocket) mode.
+
+        :param bool val: new state.
+        """
+        self._upgraded = val
+
 
 class HttpRequestParser(HttpParser):
     """Read request status line. Exception .http_exceptions.BadStatusLine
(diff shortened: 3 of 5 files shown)


asvetlov:
I see the bug described here and agree that protocol should not set `upgraded` flag itself.
The proposed fix is not enough I think, sorry.

Let me describe how things work now.
HTTP parser processes incoming requests. If the parser encounters `Upgrade` header it a) saves `msg.upgrade` flag and b) sets `parser._upgraded` flag to true.

`parser._upgraded` flag is used to prevent parsing next in-flight messages when WebSocket handshake is requested but not completed yet (`parser._payload_parser` is not set).

This is a rudiment of added-but-removed-later HTTP pipelining support: the parser should stop handling incoming messages on protocol upgrade request.

Your change postpones setting `parser._upgraded` until the message is handled by user code. So, if a client sends websocket data just after sending websocket handshake request (a little strange behavior but it doesn't violate WebSocket protocol) this data is parsed as HTTP message if the data arrives before WS handshake establishing.

A possible solution is splitting `parser._upgraded` state into `_upgrading` (upgrade requested) and `_upgraded` (upgrade is done).

In `upgrading` state the transport should stop reading, collect in-flight data into internal binary buffer *before* parsing and wait for the transition into `upgraded` state (or back to `normal`). 
In `upgdaing` there is no parser, it can be HTTP or WebSocket depending on upgrading result.
The unprocessed data can be parsed only after establishing the next parser.

Sounds complicated a little :(

gjcarneiro:
I think I understand the problem.  Unlikely as it may be, I'm glad you caught it.  OK, I will work on this some more.

gjcarneiro:
Travis CI taking forever, but, assuming build passes (it passes on my devbox), what I did was start with the suggested _upgrading + _upgraded flags.  But then I realised I never used _upgraded flag.  In the end made significant simplifications and the PR is much smaller, while still fixing the original bug.

If you're happy with, I would just like to squash the commits before allowing you to merge.

gjcarneiro:
_Rebased / squashed commits._

gjcarneiro:
I think this is ready to merge, if you have no more feedback.

PR #4002, opened 2019-08-24
Make Unix socket writable for other user (says, Nginx)

hongquan:
## What do these changes do?

Fix the permission of Unix socket domain created by `web.run_app(path=...)`

No

## Related issue number

No

## Checklist

### DIFF
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -138,7 +138,7 @@ class UnixSite(BaseSite):
             ssl=self._ssl_context, backlog=self._backlog)
         # create_unix_server sets file permission as rwxr-xr-x, which is not
         # helpful in practice, so we fix it.
-        os.chmod(self._path, 0o666)
+        os.chmod(self._path, 0o664)
 
 
 class NamedPipeSite(BaseSite):
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -1,3 +1,4 @@
+import os
 import asyncio
 import signal
 import socket
@@ -135,6 +136,8 @@ class UnixSite(BaseSite):
         self._server = await loop.create_unix_server(
             server, self._path,
             ssl=self._ssl_context, backlog=self._backlog)
+        # create_unix_server sets file permission as rwxr-xr-x, which is not helpful in practice, so we fix it.
+        os.chmod(self._path, 0o666)
 
 
 class NamedPipeSite(BaseSite):
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -136,8 +136,8 @@ class UnixSite(BaseSite):
         self._server = await loop.create_unix_server(
             server, self._path,
             ssl=self._ssl_context, backlog=self._backlog)
-        # create_unix_server sets file permission as rwxr-xr-x, which is not helpful in practice,
-        # so we fix it.
+        # create_unix_server sets file permission as rwxr-xr-x, which is not
+        # helpful in practice, so we fix it.
         os.chmod(self._path, 0o666)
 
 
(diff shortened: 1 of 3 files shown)


asvetlov:
I suspect that `o666` mode is highly insecure

hongquan:
> As Andrew wrote earlier, `666` is highly insecure and we cannot accept this change as is.
> If you change it to be configurable with a user setting, however, this would be acceptable.

@asvetlov , @webknjaz 
Why is it insecure, when it is a socket?
If you replace Unix socket with TCP/IP one, every one can "send" to the port, equivalent with "write" to Unix socket.
Why in TCP/IP socket form, you don't concern about "everyone can write" fact, then suddenly in Unix socket, you worry about it?

webknjaz:
@hongquan UNIX file socket is a node in the file system. Write privileges mean that it can be unlinked/replaced/hijacked if not protected by proper privileges or more advanced tools like SELinux-based ACL attributes.

aiohttp/web_runner.py
```diff
@@ -135,6 +136,9 @@ def name(self) -> str:
         self._server = await loop.create_unix_server(
             server, self._path,
             ssl=self._ssl_context, backlog=self._backlog)
+        # create_unix_server sets file permission as rwxr-xr-x, which is not
```
webknjaz, line 139: @hongquan this assumption is incorrect. `create_unix_server()` doesn't set any privileges. It's the OS kernel which does so following the currently set `umask` value. You can modify the umask in the shell session where you run aiohttp and it file creation syscalls will respect that.

aiohttp/web_runner.py
```diff
@@ -135,6 +136,9 @@ def name(self) -> str:
...
+        # helpful in practice, so we fix it.
+        os.chmod(self._path, 0o664)
```
webknjaz, line 141: Based on the above, I still want it to be configurable. Also, the default should give write privileges only to the owner.

aiohttp/web_runner.py
```diff
@@ -135,6 +136,9 @@ def name(self) -> str:
         self._server = await loop.create_unix_server(
             server, self._path,
             ssl=self._ssl_context, backlog=self._backlog)
+        # create_unix_server sets file permission as rwxr-xr-x, which is not
+        # helpful in practice, so we fix it.
```
webknjaz, line 140: It's not broken. Nothing to fix here.

aiohttp/web_runner.py
hongquan: Thanks.

hongquan:
I'm closing this PR because "configurable file mode" is preferred.

guusbertens:
@hongquan @webknjaz You're wrong about the permissions: whether you can remove a unix socket (or any file) depends on the permission bits that are set on the containing directory. Simply put: if you own the directory, you can remove files/sockets no matter who owns them. This is not the case if the sticky bit is set; see e.g. [here](https://linux.die.net/man/1/chmod) or [here](https://www.freebsd.org/cgi/man.cgi?query=sticky&sektion=7&apropos=0&manpath=FreeBSD+12.1-RELEASE+and+Ports).
In other words: putting a socket with 0o666 in /tmp (on which most distros set the sticky bit) means that other users cannot remove it.
Is there still interest in having a "configurable file mode"? If I get positive response, I can see what I can do.

hongquan:
@guusbertens If you confirm that a Unix socket file cannot be deleted by non-owner user, then it is better to make Unix writeable for other user (convention over configuration). The sticky bit of parent directory can be set by other mean (deploy script).

My favourite deployment way is that, I let systemd create a dedicated folder for my sockets, via `WorkingDirectory` option. With this, we can control the folder sticky bit, then all my sockets inside can be safe from deleting by other users, but let Nginx "write" to it.

Issue #4018, opened 2019-08-30
i686 manylinux build fails due to missing cryptography wheel


mjpieters:
The manylinux build step fails for arch="i686", as there is no `manylinux1_i686` wheel for `cryptography` version 2.7.

I filed pyca/cryptography#4978 to ask why no wheel is available.

This ticket is mostly to track the issue on the aiohttp side, but it may be that cryptography has to be downgraded to 2.6.1 if new aiohttp wheels need to be built (the cryptography dependency is only needed to run the tests, I think).

mjpieters:
The project dropped support for 32 bit wheels on purpose. So the options are

- beef up the docker image so cryptography can be compiled on it
- find an alternative source for i686 wheels
- accept that tests are run with an older version of cryptography 
- drop i686 wheel support for aiohttp

webknjaz:
We don't depend on cryptography directly.

webknjaz:
Also, if cryptography has its metadata properly set up, Pip will just pick a previous version matching the env.

mjpieters:
> We don't depend on cryptography directly.

No, but the build process *does*. This is about the aiohttp Travis build process being affected, because the `ci-wheel.txt` requirements file pins the version:

https://github.com/aio-libs/aiohttp/blob/4a2308a64af8c4aa704286a2f490d411c84588b6/requirements/ci-wheel.txt#L16

So *at the very least* that should be unpinned, if pip should be picking up a previous version.

webknjaz:
Ah, yeah. Env deps are pinned which is controlled by dependabot. I'd use env markers for this case. Not sure whether the bot knows how to ignore certain lines.

mjpieters:
The lines with environment markers appear to be ignored already; dependabot hasn't made a suggestion to upgrade typing_extensions to 3.7.4 either, for example.

webknjaz:
Okay, mind sending a PR with markers?

https://github.com/dependabot/feedback/issues/658

PR #4020, opened 2019-08-30
Refactor travis config

mjpieters:
## What do these changes do?

Improve the travis configuration structure and fix the OS X jobs.

- Address travis configuration warnings and remove legacy keys such as `conditions` and `version`.
- Refactor the YAML components to be more effective
- Repair the OSX build matrix (untangle the `env` / `name` confusion, ensure `pyenv-virtualenv` is installed)
- Make it much easier to maintain the OSX job matrix
- Allow the manylinux1 job to run on API and cron triggers too so issues with this step can be found more easily. The OS X jobs were already running on the same condition. Actual deployment (uploading the sdist distribution and wheels to PyPI) still depends on the the revision having a tag.

## What these changes *don't* do:

- fix the 3.8-dev build, that's due to the mypy version aiohttp supports. Mypy 0.670 depends on an older `typed-ast` release that doesn't work on 3.8. See #3928 
- fix the manylinux1 i686 wheel build, that's due to the `cryptography==2.7` pin in `ci-wheel.txt` and cryptography having dropped support for 32-bit builds. See #4018 
- adjust the pypy3 version choice, it's left at pypy3.5-6.0.0.
- adjust the OSX image choices. Travis puts up a big banner for the XCode 6.4 image builds because it dropped support last year.
- generate the OSX jobs with the Travis matrix functionality. That'd require a much bigger re-architecture where the `install`, `before_script` and `script` steps can handle both Linux and OS X environments.

## Future OS X build options

This pull request makes the builds run again, but there is actually no need to switch XCode images. OS X wheels are tied to an SDK release, newer XCode versions can build code against an older SDK release.

The SDK release defaults to the SDK that was used for the Python binary that is running the build. If you download an official Python.org OS X build then you end up building a corresponding wheel version, without the need to switch XCode images. See the [*Wheel Building* page](https://github.com/MacPython/wiki/wiki/Wheel-building) on the MacPython Wiki. The [multibuild project](https://github.com/matthew-brett/multibuild) follows this approach and outputs *combined* wheels. For a given Python version, a single wheel bundling all SDK variants is created, e.g. the matplotlib project has [`..._macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl` wheels](https://pypi.org/project/matplotlib/3.1.1/#copy-hash-modal-17ab0ed9-407a-4334-bac3-1447b7ef8745).

The multibuild project would also make integrating OS X into a matrix (with or without a separate stage) easier as it already has all the scripts to distinguish between platforms. For examples of their use, see the [Pillows project](https://github.com/radarhere/pillow-wheels/blob/d5f36a1544fe0fd47e5209fe033f50abed2d9425/.travis.yml#L183-L195), or [the 

No, this is purely a CI change.

## Misc

* The `typing_extensions` pin change should perhaps be a separate pull request.
* So could the change to have the linux wheel building stage run on API and cron triggers.
* Presumably this change doesn't warrant towncrier and contributors entries

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -170,7 +170,7 @@ matrix:
 
   include:
   - <<: *mainstream_python_base
-    name: without extensions
+    name: 3.7 without extensions
     env:
       AIOHTTP_NO_EXTENSIONS: "1"
 
@@ -213,7 +213,6 @@ matrix:
 
   - <<: *lint_base
     name: Making sure that CONTRIBUTORS.txt remains sorted
-    language: shell
     install: skip
     script:
     - LC_ALL=C sort -c CONTRIBUTORS.txt
@@ -259,9 +258,9 @@ stages:
 - *doc_stage_name
 - test
 - name: *stage_test_osx_name
-  if: type IN (api, cron)
+  if: type IN (api, cron) OR tag IS present
 - name: *deploy_stage_name
-  if: type IN (api, cron)
+  if: type IN (api, cron) OR tag IS present
 
 
 cache: pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,7 +53,9 @@ _helpers:
   deploy: &_deploy_defaults
     provider: pypi
     # `skip_cleanup: true` is required to preserve binary wheels, built
-    # inside of manylinux1 docker container during `script` step above.
+    # inside of manylinux1 docker container during `script` step, and
+    # to preserve the generated .c files from the `before_script` stage
+    # in OS X builds.
     skip_cleanup: true
     # `skip_existing: true` is required to skip uploading dists, already
     # present in PYPI instead of failing the whole process.
@@ -116,7 +118,6 @@ _helpers:
   - brew --cache
   deploy:
     <<: *_deploy_defaults
-    skip_cleanup: false
     distributions: bdist_wheel
 
 # OS X test and deploy matrix helpers
@@ -159,12 +160,6 @@ _helpers:
   <<: *_generic_deploy_base
   <<: *mainstream_python_base
   stage: &deploy_stage_name Linux build and deploy
-  deploy:
-    <<: *_deploy_defaults
-    # `skip_cleanup: true` is required to preserve binary wheels, built
-    # inside of manylinux1 docker container during `script` step above.
-    skip_cleanup: true
-
 
 os: linux
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -266,8 +266,7 @@ stages:
 - name: *stage_test_osx_name
   if: type IN (api, cron)
 - name: *deploy_stage_name
-  # This will prevent deploy unless it's a tagged commit:
-  if: tag IS present
+  if: type IN (api, cron)
 
 
 cache: pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ after_success:
 _helpers:
   # anchor nodes to be included elsewhere.
   # names that start with _ are only used in other helpers
-- &_mainstream_python_base
+- &mainstream_python_base
   python: *mainstream_python
 - &_reset_steps
   env: []
@@ -38,7 +38,7 @@ _helpers:
   script: skip
 
 - &lint_base
-  <<: *_mainstream_python_base
+  <<: *mainstream_python_base
   <<: *_reset_steps
   stage: &doc_stage_name docs, linting and pre-test checks
   env:
@@ -174,8 +174,8 @@ matrix:
   - python: nightly
 
   include:
-  - name: 3.7 without extensions
-    python: 3.7
+  - <<: *mainstream_python_base
+    name: without extensions
     env:
       AIOHTTP_NO_EXTENSIONS: "1"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,11 @@ python:
 
 install:
 - &upgrade_python_toolset pip install --upgrade pip wheel setuptools
-- make cythonize
-- pip install -r requirements/ci.txt
+- pip install --upgrade -r requirements/cython.txt
+
+before_script:
+- &cythonize make cythonize
+- &ci_requirements pip install --upgrade -r requirements/ci.txt
 
 script:
 - make cov-ci-run
@@ -24,6 +27,8 @@ after_success:
  - codecov
 
 _helpers:
+  # anchor nodes to be included elsewhere.
+  # names that start with _ are only used in other helpers
 - &_mainstream_python_base
   python: *mainstream_python
 - &_reset_steps
@@ -31,48 +36,21 @@ _helpers:
   before_install: skip
   install: skip
   script: skip
-  after_success: []
-- &_lint_base
-  stage: &doc_stage_name docs, linting and pre-test checks
+
+- &lint_base
   <<: *_mainstream_python_base
   <<: *_reset_steps
-  install:
-  - *upgrade_python_toolset
-  - pip install -U -r requirements/ci.txt
-- &_doc_base
-  <<: *_lint_base
-  install:
-  - *upgrade_python_toolset
-  - pip install -U -r requirements/doc.txt -r requirements/doc-spelling.txt
-  after_failure: cat docs/_build/spelling/output.txt
-  addons:
-    apt:
-      packages:
-      - libenchant-dev
-- &osx_python_base
-  stage: &stage_test_osx_name test under OS X (last chance to fail before deploy available)
-  os: osx
-  language: shell
-  python: *pypy3
+  stage: &doc_stage_name docs, linting and pre-test checks
   env:
-  - &env_pypy3 PYTHON_VERSION=pypy3.6-7.1.1
-  - &env_pyenv PYENV_ROOT="$HOME/.pyenv"
-  - &env_path PATH="$PYENV_ROOT/bin:$PATH"
-  before_install:
-  - brew update
-  - brew install pyenv || brew upgrade pyenv
-  - &ensure_pyenv_preloaded |
-    eval "$(pyenv init -)"
-    eval "$(pyenv virtualenv-init -)"
-  - &install_python pyenv install --skip-existing --keep --verbose "$PYTHON_VERSION"
-  - &switch_python pyenv shell "$PYTHON_VERSION"
-  - &python_version python --version
-  before_cache:
-  - brew --cache
-- &generic_deploy_base
-  stage: &deploy_stage_name deploy (PYPI upload itself runs only for tagged commits)
-  <<: *_mainstream_python_base
-  deploy: &deploy_step
+    AIOHTTP_NO_EXTENSIONS: "1"
+  install:
+  - *upgrade_python_toolset
+  - *ci_requirements
+  before_script: skip
+
+- &_generic_deploy_base
+  <<: *_reset_steps
+  deploy: &_deploy_defaults
     provider: pypi
     # `skip_cleanup: true` is required to preserve binary wheels, built
     # inside of manylinux1 docker container during `script` step above.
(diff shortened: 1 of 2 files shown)


mjpieters:
I triggered an API build at https://travis-ci.com/mjpieters/aiohttp/builds/125262526 to show the full matrix.

mjpieters:
Hrm, pondering the condition to run the OSX and deploy stages some more, the condition should probably be altered to:

```
type IN (api, cron) OR tag IS present
```

Whadayathink?

.travis.yml
```diff
@@ -13,12 +10,15 @@ python:
 - 3.6
 - &mainstream_python 3.7
 - nightly
-- &pypy3 pypy3.5-6.0.0
+- pypy3.5-6.0.0
```
webknjaz, line 13: Also needs `pypy3.6-7.1.1`.

.travis.yml
```diff
@@ -1,10 +1,7 @@
-conditions: v1
```
webknjaz, line 1: :+1:

.travis.yml
```diff
@@ -1,10 +1,7 @@
...
-version: "= 0"
```
webknjaz, line 2: :+1:

.travis.yml
```diff
@@ -1,10 +1,7 @@
...
 if: >  # Forbid running non-PR pushes from pyup bot
   not (type != pull_request AND branch =~ ^pyup\-scheduled\-update\-)
 
 dist: xenial
-sudo: required
```
webknjaz, line 7: :+1:

.travis.yml
```diff
@@ -27,65 +27,40 @@ after_success:
...
+
+- &_generic_deploy_base
+  <<: *_reset_steps
+  deploy: &_deploy_defaults
     provider: pypi
     # `skip_cleanup: true` is required to preserve binary wheels, built
     # inside of manylinux1 docker container during `script` step above.
     skip_cleanup: true
-    # `skip-existing: true` is required to skip uploading dists, already
+    # `skip_existing: true` is required to skip uploading dists, already
```
webknjaz, line 60: It doesn't matter: dpl turns dashes into underscores internally. Plus I prefer dashes here.

.travis.yml
```diff
@@ -27,65 +27,40 @@ after_success:
...
     # present in PYPI instead of failing the whole process.
     # This happens when other CI (AppVeyor etc.) has already uploaded
     # the very same dist (usually sdist).
-    skip-existing: true
-    user: aio-libs-bot
+    skip_existing: true
```
webknjaz, line 64: Keep dashes.

.travis.yml
```diff
@@ -27,65 +27,40 @@ after_success:
...
+    username: aio-libs-bot
```
webknjaz, line 65: Is `user:` deprecated?

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
   env:
-  - &env_os1011_msg Build and deploy to PYPI of OS X 10.11 binary wheel=
-  - &env_py36 PYTHON_VERSION=3.6.3
-  - *env_pyenv
-  - *env_path
+  - &env_pyenv PYENV_ROOT="$HOME/.pyenv"
+  - &env_path PATH="$PYENV_ROOT/bin:$PATH"
+  before_install:
+  - brew update
+  - |
```
webknjaz, line 92: 1) This all is meant to be in `_helpers`. I put it there for a reason: big chunks of logic in the middle of matrix reduce readability.
2) If you want to help with this, please port https://github.com/amol-/dukpy/blob/69f56f3/.travis.yml#L41-L120 here, it's a speed-up for tests. But keep a brew implementation: it's needed for dev/nightly.
3) This change is too big and will need more adjustments and closer look. So please factor it out into a separate PR. Otherwise it'll be hard to review and reason about this which will delay merging and block even smaller improvements.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
+    <<: *_deploy_defaults
+    # `skip_cleanup: true` is required to preserve binary wheels, built
+    # inside of manylinux1 docker container during `script` step above.
+    skip_cleanup: true
+
+
 os: linux
 
-jobs:
+matrix:
```
webknjaz, line 166: I prefer `jobs` here. It's intentional.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
   fast_finish: true
   allow_failures:
   - python: nightly
 
   include:
-  - name: 3.7 without extensions
-    python: 3.7
+  - <<: *mainstream_python_base
+    name: without extensions
```
webknjaz: With name set, the version in the build view isn't displayed. That's why we have to put it here manually.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
     env:
-      AIOHTTP_NO_EXTENSIONS: 1
+      AIOHTTP_NO_EXTENSIONS: "1"
```
webknjaz, line 175: This is probably not important. And it's cleaner w/o double-quotes.

.travis.yml
```diff
@@ -27,65 +27,40 @@ after_success:
...
-  - &switch_python pyenv shell "$PYTHON_VERSION"
-  - &python_version python --version
-  before_cache:
-  - brew --cache
-- &generic_deploy_base
-  stage: &deploy_stage_name deploy (PYPI upload itself runs only for tagged commits)
-  <<: *_mainstream_python_base
-  deploy: &deploy_step
+  - *ci_requirements
+  before_script: skip
```
webknjaz, line 49:
```suggestion
  before_script: []
```

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
-    - AIOHTTP_NO_EXTENSIONS=1 pip install -r requirements/ci.txt -r requirements/doc.txt
     script:
-    - AIOHTTP_NO_EXTENSIONS=1 python setup.py --verbose sdist bdist_wheel
+    - python setup.py --verbose sdist bdist_wheel
     - twine check dist/*
 
-  - <<: *_lint_base
+  - <<: *lint_base
     name: Making sure that CONTRIBUTORS.txt remains sorted
-    language: minimal
```
webknjaz, line 192: Keep it minimal. This is to speed-up the check.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
-    install:
-    - skip
+    language: shell
+    install: skip
```
webknjaz, line 216: I don't think this change is necessary.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
 
 stages:
 - *doc_stage_name
 - test
 - name: *stage_test_osx_name
   if: type IN (api, cron)
 - name: *deploy_stage_name
-  # This will prevent deploy unless it's a tagged commit:
-  if: tag IS present
+  if: type IN (api, cron)
```
webknjaz: This is wrong. Tag is absolutely necessary here. We don't publish intermediate commits to PyPI.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
+    - ./tools/run_docker.sh "aiohttp"
 
 stages:
 - *doc_stage_name
 - test
 - name: *stage_test_osx_name
   if: type IN (api, cron)
 - name: *deploy_stage_name
-  # This will prevent deploy unless it's a tagged commit:
-  if: tag IS present
```
webknjaz, line 307: This part needs to be reverted.

requirements/ci-wheel.txt
```diff
@@ -15,11 +15,11 @@ tox==3.13.2
 trustme==0.5.2
 cryptography==2.7
 twine==1.13.0
+typing_extensions==3.7.4
```
webknjaz, line 18: I'd put this into a separate PR. We'd also be able to backport it while figuring out CI stuff.

.travis.yml
```diff
@@ -101,210 +76,197 @@ _helpers:
...
-  - &env_os1012_msg Build and deploy to PYPI of OS X 10.12 binary wheel=
-  - *env_py36
+  <<: *_osx_pypi_deploy_base
+  osx_image: xcode8.3
+
+# OS X Python options (the _osx_env_<version>
+# key is only needed to provide a container for
+# the &osx_env_<version> YAML anchor).
+- &osx_env_py35
+  - PYTHON_VERSION: "3.5"
```
webknjaz, line 147: You're putting a mapping into a sequence. AFAIR this isn't supposed to work. Sequences use `var=value` style while mappings use `val: value`...

webknjaz:
> Actual deployment [...] still depends on the revision having a tag

This is not true. There's no condition for this if you remove what we have.
Also, if we use if-clauses to control consumed resources: Travis CI is slow, there's limits which affect the whole org. I'd invest into making the build check work as a part of GitHub Actions workflows instead. I started experimenting with this a while back but hit certain limitations. Now that there's v2 of Actions, it seems doable.

webknjaz:
In general, I'd be happy to accept smaller patches because it's hard to spot differences between things moved back and forth in the diff.

.travis.yml
mjpieters: `shell` **is** `minimal`, see the [*minimal and generic* documentation](https://docs.travis-ci.com/user/languages/minimal-and-generic/#aliases):

> Setting the `language` key to `bash`, `sh` or `shell` is equivalent to `language: minimal`.

and travis itself, when showing the effective config for a build, starts with the message:

>  `matrix.include.language: minimal` is an alias for `shell` (value), using `shell` 

This is one of many INFO messages that make it harder to spot the important ones.

.travis.yml
mjpieters: Nope, sorry, the `language` key looks like a commit reshuffle error. Putting `skip` on the same line pulled it in line with other skips elsewhere in the file, however.

.travis.yml
mjpieters: Yes, I had come to the same conclusion and made the [same change for frozenlist](https://github.com/mjpieters/frozenlist/commit/b92ae33b292e3caf797e6a2ea0315f2cb16b09a9).

requirements/ci-wheel.txt
mjpieters: Already done, already landed. :-)

.travis.yml
mjpieters: I meant to add this to the initial post: I deliberately didn't touch the pypi version either. This can be done in a follow-up PR.

.travis.yml
mjpieters: Sure, it's just a comment.

.travis.yml
mjpieters: This is an INFO log entry in the travis config view:

>  `matrix.include.deploy` key `skip-existing` is not underscored, using `skip_existing` 

 More-over, *every other key in the config uses underscores*, using dashes here is inconsistent and confusing. This isn't the `twine` command-line, this is the `deploy` configuration.

.travis.yml
mjpieters: Travis states it has to replace `user` with `username`:

>  `matrix.include.deploy`: `user` is an alias for `username` (key), using `username` 

This is level INFO, but any message in the `view log` tab is a message too many and leads to important info being overlooked.

.travis.yml
mjpieters: This *is* in helpers.

Otherwise, happy to defer the pyenv / Python.org discussion. I did address this in the PR discussion, and agree that using a downloaded Python is a much better idea.

.travis.yml
mjpieters: But, given that the config switch is really named with an underscore, not a dash, it's more consistent to use `skip_existing`.

.travis.yml
mjpieters: Perhaps, but Travis logs it as:

>  `root: jobs` is an alias for `matrix` (key), using `matrix` 

I was trying to remove all travis config warnings and info lines, so that any future issues stand out immediately.

.travis.yml
mjpieters: I'll put it back then. I tried to remove a redundancy that could easily be overlooked when the `&mainstream_python` anchor is moved to 3.8 (eventually).

mjpieters: (if only Travis allowed for some form of templating in the `name` field..)

.travis.yml
mjpieters: Another travis *view config* message:

> `matrix.include.env.AIOHTTP_NO_EXTENSIONS`: casting value `1` (num) to `1` (str)

.travis.yml
mjpieters: Both of those work. Yes, the YAML representation is a list of single key-value mappings, but the end result is the same as having used a list of single strings with the `=` syntax. Travis doesn't care either way.

.travis.yml
mjpieters: Counter proposal: ifbthe `name` key is removed entirely then Travis lists the Python version and the `AIOHTTP_NO_EXTENSIONS` env variable. That’d also remove the redundancy b

mjpieters:
I probably over-extended with this pull request and I won't have time in the next week to attend to this. #4028 can serve as a discussion for the OS X issues and  I'll reduce this pull request to a smaller clean up.

.travis.yml
webknjaz: I've found that using kebab-case for identifiers reads better in YAML. That's why I prefer dashed options where possible. These keys correspond to CLI options of a tool called Dpl (by Travis). I even contributed one provider there, I do know how it works. So it's only a matter of documentation.

.travis.yml
webknjaz: Ah.. I must've missed that with huge chunks moved around.

.travis.yml
webknjaz: AFAIK warnings are only shown if there's `version: "= 0"` present. Since you removed that line, Travis won't show any warnings.

.travis.yml
webknjaz: With `version: "= 0"`, Travis CI doesn't show warnings at all.

I think that `minimal` carries a better semantic meaning.

webknjaz:
@mjpieters thanks for your contribution anyway :)
If I have time, I'll extract some bits of this PR myself. If not — we'll get back to this whenever you're ready.

.travis.yml
webknjaz: > if only Travis allowed for some form of templating

I suggested it, back when stages were in beta:
https://travis-ci.community/t/job-labels-name-improvement-proposal/79?u=webknjaz
Yet, it doesn't look like they're planning to add this feature...

> Counter proposal

Ah... This lack of feature in travis was very annoying back when I added "name" labels originally. I couldn't decide what's better: having a nice helpful name or seeing those matrix details there. I'm not happy with either solution. I want templating. Just env vars don't give a good context on what's going on in that job. We abused env vars to just show this information there, right until travis introduced names.
I think I still prefer names even with this ridiculous need to hardcode things :(

.travis.yml
mjpieters: Really? The `version` root key is virtually undocumented, and I find the travis-ci project a bit hard to navigate to figure out where this would be decided.

I so far have used https://config.travis-ci.org/ to test configurations, and it produces warnings and such regardless of the `version` key.

webknjaz: Yep, totally undocumented. I stole this from some of Travis' own repos.
Anyway, I'd prefer semantically informative names. Warnings aren't errors after all. WDYT?

asvetlov:
Guys, I'm curious is there a plan to land the PR?

.travis.yml
```diff
@@ -13,12 +10,15 @@ python:
 - 3.6
 - &mainstream_python 3.7
 - nightly
-- &pypy3 pypy3.5-6.0.0
+- pypy3.5-6.0.0
 
 install:
 - &upgrade_python_toolset pip install --upgrade pip wheel setuptools
+- pip install --upgrade -r requirements/cython.txt
```
asvetlov, line 17: Why do you apply `--upgrade` here and below?
Our requirement files always have strict dependency versions, updated by dependabot.
I thought that `-U` is pointless in this case.

mjpieters: I've used `--upgrade` throughout to be consistent, and to account for possible 'comes with Cython pre-installed' issues. It may be better to default to *not* using `--upgrade` and explicitly documenting when we do need to use the flag (e.g. for `pip` itself).

asvetlov: > It may be better to default to not using --upgrade and explicitly documenting when we do need to use the flag (e.g. for pip itself).

You've got my point. In the project we are *implicitly* follow this style (at least I think we do). If you want to document this fact somewhere -- you are welcome :)

mjpieters:
> Guys, I'm curious is there a plan to land the PR?

As @webknjaz stated early on, he'd like to see this split up, and I stated that I suspect I overstretched. I think the plan is now to [extract the useful bits](https://github.com/aio-libs/aiohttp/pull/4020#issuecomment-526868877).

Moreover, the OSX build config is incomplete and not an approach I think should be followed anymore:

* in `frozenlist` I had to add `pip install --upgrade pip` because the `dpl-pypi` package used by Travis [installs the newest pip with `PIP_USER=yes`](https://github.com/travis-ci/dpl/blob/2abf7b3ce8bf2c5b1b26b0df1727f194c0dc0104/lib/dpl/assets/pypi/install) and this [breaks the `bin/pip` script file](https://travis-ci.com/mjpieters/frozenlist/jobs/229766801#L11103-L11108) pre-installed with pip 9.0.1 on the OS X images.
* we really, really should use multibuild to handle OS X images, as that method is way faster. Getting to the point we can use an OS X image to start building, testing and deploying takes 12-13 minutes (the bulk of which is `brew update` followed by installing `pyenv`, the `pyenv install` command can make use of cached builds). `multibuild` on the other hand has the environment ready and primed in about a minute, and OS X builds of the numpy wheels [are slightly faster than 64-bit manylinux builds](https://travis-ci.org/MacPython/numpy-wheels).

I still plan to convert my `frozenlist` travis config into using multibuild, but I have to prep for PyCON UK this week.

.travis.yml
webknjaz: I also prefer YAML mappings because you can merge anchors with those.

asvetlov:
@mjpieters thank you for your work but the project doesn't use Travis CI anymore for the sake of Azure Pipelines.
Now we generate a file like aiohttp-3.6.2-cp37-cp37m-macosx_10_13_x86_64.whl, you can see it at https://pypi.org/project/aiohttp/#files
If you want to propose improvements -- please do. Azure configuration for wheels building is here: https://github.com/aio-libs/aiohttp/blob/master/.azure-pipelines/deploy.yml
I believe it is much easier than Travis version was.

PR #4021, opened 2019-08-30
CI: Fix i686 manylinux build

### LINKED ISSUE #4018: i686 manylinux build fails due to missing cryptography wheel

mjpieters:
## What do these changes do?

The cryptography project has stopped providing 32-bit wheels, 2.6.1 is the last version that still has a wheel available.

This makes it possible to run tests against the i686 32-bit wheel built in Travis again.

None, this is a CI build fix only.

## Related issue number

This fixes #4018

### DIFF
--- a/requirements/ci-wheel.txt
+++ b/requirements/ci-wheel.txt
@@ -13,12 +13,13 @@ pytest-cov==2.7.1
 pytest-mock==1.10.4
 tox==3.13.2
 trustme==0.5.2
-cryptography==2.7
 twine==1.13.0
 yarl==1.3.0
 
 # Using PEP 508 env markers to control dependency on runtimes:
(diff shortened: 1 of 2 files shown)


asvetlov:
Sorry, no. I rather prefer to skip `cryptography` and `trustme` installation inside docker container on tests run.

Applied this approach to 3.6 branch yesterday (see `git diff 3.6 -- tools/ tests/conftest.py requirements/ci-wheel.txt`)

PR #4031, opened 2019-08-31
Fix #4012 encoding of content-disposition parameters

### LINKED ISSUE #4012: MultipartWriter quotes field name wrong
## Long story short

My client needs to send multipart/form-data to an API that expects field names with `[]` in the name. The server does not accept the submission with default `set_content_disposition` parameters due to wrong quoting.

## Expected behaviour

`Content-Disposition: form-data; name="files[]"; filename="filename"`

## Actual behaviour

`Content-Disposition: form-data; name="files%5B%5D"; filename="filename"; filename*=utf-8''filename`

## Steps to reproduce

Client code is like

```
        with aiohttp.MultipartWriter('form-data') as mpw:
                f = mpw.append(file)
                f.set_content_disposition("form-data", name="files[]", filename="filename")

        res = await self.session.post(url, data=mpw)
```

## Your environment

aiohttp==3.5.4 async client, Ubuntu 18.04, python 3.6.8.

## Analysis

[Returning Values from Forms: multipart/form-data](https://tools.ietf.org/html/rfc7578#section-4.2) says

>    In most multipart types, the MIME header fields in each part are
>    restricted to US-ASCII; for compatibility with those systems, file
>    names normally visible to users MAY be encoded using the percent-
>    encoding method in Section 2, following how a "file:" URI
>    [URI-SCHEME] might be encoded.
> 
>    NOTE: The encoding method described in [RFC5987], which would add a
>    "filename*" parameter to the Content-Disposition header field, MUST
>    NOT be used.

It would seem the current implementation misinterpreted this to mean all field values are to be [percent-encoded](https://tools.ietf.org/html/rfc7578#section-2). But the RFC7578 is clear that the encoding is only to be used on file names. Furthermore, the `filename*=` form from [MIME Parameter Value and Encoded Word Extensions](https://tools.ietf.org/html/rfc2231) should be used only for the other fields, but as the filename is already via percent-encoding to within US-ASCII, `filename*=` is not to be used on the filename.

For converting from unicode string to bytes for the percent-encoding, user will need to specify charset in some cases, as in the RFC:

> The encoding used for the file names is typically UTF-8, although
>    HTML forms will use the charset associated with the form.

Thus, in some cases, an additional `charset` parameter is needed in `set_content_disposition`. Is it needed in other functions?

The RFCs refer to RFC822 for quoted-string definition, which is currently obsoleted by [Internet Message Format RFC5322](https://tools.ietf.org/html/rfc5322#section-3.2.4).

```
   qtext           =   %d33 /             ; Printable US-ASCII
                       %d35-91 /          ;  characters not including
                       %d93-126 /         ;  "\" or the quote character
                       obs-qtext

   qcontent        =   qtext / quoted-pair

   quoted-string   =   [CFWS]
                       DQUOTE *([FWS] qcontent) [FWS] DQUOTE
                       [CFWS]

   quoted-pair     =   ("\" (VCHAR / WSP)) / obs-qp
```

And from [Augmented BNF for Syntax Specifications: ABNF](https://tools.ietf.org/html/rfc5234#appendix-B.1)

```
  VCHAR          =  %x21-7E
                                ; visible (printing) characters

         WSP            =  SP / HTAB
                                ; white space```
```

The quoted-pair quoting of quoted-string is missing in the current implementation.

There is also a rather far-fetched case of extremely long values causing the line length limit of 998 characters to be exceeded https://tools.ietf.org/html/rfc5322#section-2.1.1 and requiring using the Folding White Space (FWS).

I can not tell if there would be any compatibility impact of just changing the percent-quoting to the correct quoted-pair quoting. Should the `quote_fields` parameter concern the percent-encoding of filename or the quoted-pair of all fields?

The current behavior seems to be result of discussion in #916 to fix #903.

kohtala:
Posting form data with field names containing "[]" were not recognized
by other multipart/form-data parsers. The percent-encoding is only to
be used on the filename parameter that follows how of a "file:" URI
may be encoded according to RFC7578.

## What do these changes do?

Only encode content-disposition filename parameter using percent-encoding.
Other parameters are encoded to quoted-string or RFC2231 extended parameter
value.

I do not know if there are some implementations that depend on the wrong encoding.

## Related issue number

Related to #4012.

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -382,13 +382,40 @@ def guess_filename(obj: Any, default: Optional[str] = None) -> Optional[str]:
     return default
 
 
+not_qtext_re = re.compile(r"[^\041\043-\133\135-\176]")
+QCONTENT = {chr(i) for i in range(0x20, 0x7F)} | {"\t"}
+
+
+def quoted_string(content: str) -> str:
+    """Return 7-bit content as quoted-string.
+
+    Format content into a quoted-string as defined in RFC5322 for
+    Internet Message Format. Notice that this is not the 8-bit HTTP
+    format, but the 7-bit email format. Content must be in usascii or
+    a ValueError is raised.
+    """
+    if not (QCONTENT > set(content)):
+        raise ValueError(f"bad content for quoted-string {content!r}")
+    return not_qtext_re.sub(lambda x: "\\" + x.group(0), content)
+
+
 def content_disposition_header(
-    disptype: str, quote_fields: bool = True, **params: str
+    disptype: str, quote_fields: bool = True, _charset: str = "utf-8", **params: str
 ) -> str:
-    """Sets ``Content-Disposition`` header.
+    """Sets ``Content-Disposition`` header for MIME.
+
+    This is the MIME payload Content-Disposition header from RFC 2183
+    and RFC 7579 section 4.2, not the HTTP Content-Disposition from
+    RFC 6266.
 
     disptype is a disposition type: inline, attachment, form-data.
     Should be valid extension token (see RFC 2183)
+
+    quote_fields performs value quoting to 7-bit MIME headers
+    according to RFC 7578. Set to quote_fields to False if recipient
+    can take 8-bit file names and field values.
+
+    _charset specifies the charset to use when quote_fields is True.
 
     params is a dict with disposition params.
     """
@@ -403,10 +430,23 @@ def content_disposition_header(
                 raise ValueError(
                     "bad content disposition parameter" " {!r}={!r}".format(key, val)
                 )
-            qval = quote(val, "") if quote_fields else val
-            lparams.append((key, '"%s"' % qval))
-            if key == "filename":
-                lparams.append(("filename*", "utf-8''" + qval))
+            if quote_fields:
+                if key.lower() == "filename":
+                    qval = quote(val, "", encoding=_charset)
+                    lparams.append((key, '"%s"' % qval))
+                else:
+                    try:
+                        qval = quoted_string(val)
+                    except ValueError:
+                        qval = "".join(
+                            (_charset, "''", quote(val, "", encoding=_charset))
+                        )
+                        lparams.append((key + "*", qval))
+                    else:
+                        lparams.append((key, '"%s"' % qval))
+            else:
+                qval = val.replace("\\", "\\\\").replace('"', '\\"')
+                lparams.append((key, '"%s"' % qval))
         sparams = "; ".join("=".join(pair) for pair in lparams)
         value = "; ".join((value, sparams))
     return value
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -1422,7 +1422,7 @@ async def test_POST_FILES_SINGLE_content_disposition(aiohttp_client, fname) -> N
             "text/x-python",
         ]
         assert request.headers["content-disposition"] == (
-            "inline; filename=\"conftest.py\"; filename*=utf-8''conftest.py"
+            'inline; filename="conftest.py"'
         )
 
         return web.Response()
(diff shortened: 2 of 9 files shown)


asvetlov:
@kxepal would you review?

aiohttp/helpers.py
```diff
@@ -373,10 +401,22 @@ def content_disposition_header(disptype: str,
             if not key or not (TOKEN > set(key)):
                 raise ValueError('bad content disposition parameter'
                                  ' {!r}={!r}'.format(key, val))
-            qval = quote(val, '') if quote_fields else val
-            lparams.append((key, '"%s"' % qval))
-            if key == 'filename':
-                lparams.append(('filename*', "utf-8''" + qval))
```
webknjaz: I'm not sure this is correct. `filename*` must be added if `filename` is not ASCII-only per RFC 6266, appendix D. The logic is supposed to be roughly https://github.com/cherrypy/cherrypy/pull/1851/files#diff-9d544371f1a0e2079e68ac3cca04bc6cd1ccaf92797f378b285f307871b8d6f0R32-R53. Note that NFKC is supposed to be used for normalization.

tests/test_client_functional.py
```diff
@@ -1439,7 +1439,7 @@ async def handler(request):
                                         'application/octet-stream',
                                         'text/x-python']
         assert request.headers['content-disposition'] == (
-            "inline; filename=\"conftest.py\"; filename*=utf-8''conftest.py")
```
webknjaz: You remove tests for `filename*` but I'd prefer to have them with actually correct examples with Unicode filenames instead.

aiohttp/helpers.py
webknjaz: Looking closer, it seems you may be right about the newer RFC...

kohtala: Hi. As this is the payload and not http Content-Disposition header, `filename*` can not be used. I believe your link was about the http header.

tests/test_client_functional.py
kohtala: `test_helpers.py` has tests with unicode filenames. They do not carry the `filename*` because RFC7578 tells not to use it here.

aiohttp/helpers.py
webknjaz: Ah, then I misread what you're doing here.

kohtala:
I rebased the commit on top of current master. I do not see the macos failure related to this PR.

What is the CLA message? The link says

> Error
> There is no CLA to sign for aio-libs/aiohttp
> (could not find linked item for owner aio-libs and repo aiohttp)

asvetlov:
Please ignore the CLA, I've switched it off back -- and forever.

asvetlov (approved):
LGTM.
Thank you!

I'd like to get an additional review from aiohttp team, otherwise I'll merge this PR in a week.
It will be a part of aiohttp 3.8

tests/test_helpers.py
```diff
@@ -442,6 +442,44 @@ def test_content_disposition() -> None:
...
+            "attachment", filename="bär", _charset="latin-1"
+        )
+        == 'attachment; filename="b%E4r"'
+    )
+    assert (
+        helpers.content_disposition_header(
+            "attachment", filename='bär "\\', quote_fields=False
+        )
+        == 'attachment; filename="bär \\"\\\\"'
+    )
```
webknjaz: It's better to have this use `@pytest.mark.parametrize` because it's many tests, not one.

asvetlov: The minor thing I believe. 
If @kohtala wants to update tests -- that's perfect.
If no -- I think we can accept tests as is.

P.S.
Our code is not ideal but just 'good enough' :)

kohtala: Ok. It looked nicer too.

kohtala: I parameterized the tests yesterday, but was in a hurry and failed to push. I pushed it now. Thanks for the idea. It also looks cleaner this way.

asvetlov (approved):

asvetlov:
Thanks!

PR #4046, opened 2019-09-07
Drop 3.5

### LINKED ISSUE #3947: Discussion: drop Python 3.5 support
Now aiohttp is Python 3.5 compatible.

The latest Debian Buster (released about 2 weeks ago) has Python 3.7 by default.
RHEL 8 (released on May 7th, 2019) has Python 3.6 by default.

I believe there is no sense in Python 3.5 support for aiohttp 4.0

In #3928 @AMDmi3 expressed an opinion that he is not interested in a contribution for legacy Python version.
In #3946 @ayushkovskiy  used Python 3.6 syntax.
IIRC @samuelcolvin asked for dropping 3.5 a half year ago.

Python 3.5 is in security-fixing mode and going to EOL soon.
 
Do somebody object for Python 3.5 support dropping?
Opinions?

asvetlov:
Fixes #3947

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,10 @@ sudo: required
 language: python
 
 python:
-- &py35 3.5
 - &py36 3.6
 - &py37 3.7
 - nightly
-- &pypy3 pypy3.5-6.0.0
+- &pypy3 pypy3
 
 install:
 - &upgrade_python_toolset pip install --upgrade pip wheel setuptools
@@ -312,14 +311,6 @@ jobs:
     deploy:
       <<: *deploy_step
 
-    # Build and deploy MacOS binary wheels for each OSX+Python combo possible
-    # OS X 10.10, Python 3.5
-  - <<: *osx_pypi_deploy_base_1010
-    name: *env_os1010_msg
-    python: *py35
-    env:
-      <<: *env_osx_base
-      PYTHON_VERSION: *py35
     # OS X 10.10, Python 3.6
   - <<: *osx_pypi_deploy_base_1010
     name: *env_os1010_msg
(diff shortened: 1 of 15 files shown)


webknjaz:
@asvetlov FYI I've rebased this

setup.py
```diff
@@ -10,8 +10,8 @@
 from setuptools import Extension, setup
 
 
-if sys.version_info < (3, 5, 3):
-    raise RuntimeError("aiohttp 3.x requires Python 3.5.3+")
+if sys.version_info < (3, 6):
```
webknjaz, line 9: It's okay to just remove this check since setuptools that supports `python_requires` is shipped with Python 3.6.

setup.py
```diff
@@ -10,8 +10,8 @@
...
+    raise RuntimeError("aiohttp 3.x requires Python 3.6+")
```
webknjaz: But if you want to keep it:
```suggestion
    raise RuntimeError("aiohttp 4.x requires Python 3.6+")
```

webknjaz:
I've triggered a full matrix build for this branch now: https://travis-ci.com/aio-libs/aiohttp/builds/126532209

setup.py
asvetlov: I want to fail loudly even on Python 3.5

asvetlov:
Thanks.
`pypy` fails on the branch on yarl encoding, I need to investigate

.travis.yml
```diff
@@ -312,14 +311,6 @@ jobs:
     deploy:
       <<: *deploy_step
 
-    # Build and deploy MacOS binary wheels for each OSX+Python combo possible
```
webknjaz, line 315: Please keep this line, it's a "header" for all other jobs.

asvetlov:
The failure source is different behavior of yarl quoting/unquoting for cythonized and pure python versions

asvetlov:
The failure reason is a different behavior of pure python and cythonized yarl versions.
The bug is fixed on yarl master, I'll move failed tests into yarl repository.

webknjaz:
Backported as https://github.com/aio-libs/aiohttp/commit/72d3550ecf2b40a8828146222057567eb997f823 and https://github.com/aio-libs/aiohttp/commit/461c98cc4eebf180abe5eaa6948a0832845e8d06 to address the compatibility issues with Python 3.11 in tests.

PR #4050, opened 2019-09-08
Port the install from python.org Travis speedup

webknjaz:


asvetlov:
Is it for MacOS only?
What is the boost?

webknjaz:
@asvetlov yes, it's about 2x:
- old: https://travis-ci.com/aio-libs/aiohttp/jobs/232085550 (13 min 19 sec)
- new: https://travis-ci.com/aio-libs/aiohttp/jobs/232164269 (6 min 57 sec)

This has been tested on a number of repos earlier by me. The speed-up happens because I'm eliminating `pyenv` install/update by `brew` which always tries to update its cache that is extremely slow.

We only need to figure out why some tests fail under some envs.

webknjaz:
Alright, `AttributeError: module 'ssl' has no attribute 'OP_NO_COMPRESSION'` happens because this build is linked against the default OpenSSL under macOS which is lower than 1.0.0. I think I'd just ignore this error but we could also emit a warning about it as well.

.travis.yml
```diff
@@ -70,6 +81,72 @@ _helpers:
   - &install_python pyenv install --skip-existing --keep --verbose "$PYTHON_VERSION"
   - &switch_python pyenv shell "$PYTHON_VERSION"
   - &python_version python --version
+  before_install: &install-from-python_org
+  - |
+    function probe_url() {
```
asvetlov, line 89: Is it possible to put this complex code into `./tools/` folder and call it from travis config?
It makes a config more understandable.

.travis.yml
```diff
@@ -26,11 +26,18 @@ script:
 after_success:
  - codecov
 
+env:
+  global:
+    GIT_INSTALLER_DIR_PATH: ${HOME}/.git-installers
```
asvetlov, line 32: Please add a comment that these "globla" vars are needed for macos only.

.travis.yml
```diff
@@ -306,7 +377,12 @@ stages:
   if: tag IS present
 
 
-cache: pip
+cache:
+  pip: true
+  directories:
+  - $HOME/Library/Caches/Homebrew
```
asvetlov: Please comment that these dirs are macos specific

tests/test_web_protocol.py
```diff
@@ -338,6 +341,11 @@ def close():
         "Error handling request", exc_info=mock.ANY)
 
 
+@pytest.mark.xfail(
```
asvetlov: Please replace `xfail` with `skipif`, apply for Python 3.5 only.
I'd like to make the test passing on Python 3.6+ at least

.travis.yml
webknjaz: I prefer it here for now because I'd like to see everything in one file. Besides, it's in a separate section anyway. Plus I want to see separate entries in the build log which is impossible with a separate script.

.travis.yml
webknjaz: It doesn't really matter. I can even move these to macOS only...

tests/test_web_protocol.py
webknjaz: In my previous commit it was ` and IS_PY35` but then I noticed that this is unstable and happens under other versions as well.
I think it must be xfail because of this. And I need to fix the reason instead.

webknjaz:
The latest build: https://travis-ci.com/aio-libs/aiohttp/builds/126421862

webknjaz:
Last build: https://travis-ci.com/aio-libs/aiohttp/builds/126421862

.travis.yml
```diff
@@ -57,11 +64,15 @@ _helpers:
...
-  - &env_pypy3 PYTHON_VERSION=pypy3.6-7.1.1
-  - &env_pyenv PYENV_ROOT="$HOME/.pyenv"
-  - &env_path PATH="$PYENV_ROOT/bin:$PATH"
-  before_install:
+  env: &env_pyenv_base
+    PYTHON_VERSION: *pypy3
+    PATH: ${PYENV_ROOT}/bin:${PATH}
+  env: &env_osx_base
+    <<: *env_pyenv_base
+    PYTHON_VERSION: *pypy3
```
mjpieters, line 75: I'd like to see a comment as to why this prefers to use a pypy3 image over other image options (such as `shell` or `generic` or `python`.

aiohttp/connector.py
```diff
@@ -849,7 +849,16 @@ def _make_ssl_context(verified: bool) -> SSLContext:
...
+            try:
+                sslcontext.options |= ssl.OP_NO_COMPRESSION
+            except AttributeError as attr_err:
+                warnings.warn(
+                    '{!s}: The Python interpreter is compiled '
+                    'against OpenSSL < 1.0.0. Ref: '
+                    'https://docs.python.org/3/library/ssl.html'
+                    '#ssl.OP_NO_COMPRESSION'.
+                    format(attr_err),
+                )
```
mjpieters: Why is this in this pull request? Can fixing the travis build and fixing the test failure on OS X be separated into two distinct pull requests?

mjpieters:
>  @mjpieters I know you prefer `multibuild` but I'm curious about your opinion for the PR

I'd still prefer to use `multibuild`, I much rather collaborate on perfecting OS X wheel building on Travis with the wider community.

.travis.yml
webknjaz: This doesn't need to be reused right now. We can talk about it separately but not in this PR. It's easier for me to copy things across projects this way right now.

webknjaz: @mjpieters let's talk on Gitter some time: I have a certain vision of where the CI stuff should go in general, architecturally and I want us to be on the same page.

.travis.yml
asvetlov: Please do

tests/test_web_protocol.py
asvetlov: The reason is OpenSSL version bundled with Python. Have no idea how to "fix" it.

webknjaz: This test doesn't seem related to TLS in any way. You're probably confusing it with the other one (actually two other tests) which actually was failing because of OpenSSL and I've patched that up.

aiohttp/connector.py
webknjaz: I can't test this w/o the CI config. I think once other comments are settled I could separate it, of course.

webknjaz:
@mjpieters multibuild is not acceptable atm. We'll think about it in other iterations. It's okay to contribute improvements there meanwhile but I've got a number of concerns that have to be settled first.

.travis.yml
webknjaz: I just remembered why I made it global: this way we don't overwrite the values in individual jobs. I'll keep it here.

PR #4059, opened 2019-09-14
Provide pytest-aiohttp namespace for pytest fixtures in the doc

### LINKED ISSUE #3723: Documentation about `aiohttp_unused_port` inaccurate?
The documentation here: https://docs.aiohttp.org/en/stable/testing.html#aiohttp.test_utils.aiohttp_unused_port

states:

"Changed in version 3.0: The fixture was renamed from unused_port to aiohttp_unused_port"

But I cannot find `aiohttp_unused_port` anywhere in the 3.5.4.

asvetlov:
Fixes #3723

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -71,6 +71,8 @@ proxy methods to the client for common operations such as
 Pytest
 ~~~~~~
 
+.. currentmodule:: pytest_aiohttp
+
 The :data:`aiohttp_client` fixture available from pytest-aiohttp_ plugin
 allows you to create a client to make requests to test your app.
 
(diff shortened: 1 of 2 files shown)


docs/testing.rst
```diff
@@ -235,6 +237,9 @@ Pytest tooling has the following fixtures:
 Unittest
 ~~~~~~~~
 
+.. currentmodule:: aiohttp.test_utils
+
```
webknjaz, line 241: Should it be two empty lines here?

asvetlov: Maybe one empty line is enough but sorry, I don't want to wait for CI again before the merge.
Two lines are not such bad.
We have no strong rule for docs, sometimes it has one line as a delimiter but sometimes it has two or even more.

Issue #4063, opened 2019-09-16
ClientSession cookies not expiring


alandtse:
## Long story short

Using a ClientSession where the server sends cookies with an `expires` date in the past results in the cookies remaining in the cookie_jar that are then sent in subsequent requests.

## Expected behaviour

If a server sends an expired cookie, it should remove that entry from the cookie_jar if it exists.

## Actual behaviour

The expired cookie remains in the cookie_jar and is sent in the header of the next response.

## Steps to reproduce

Session creation code
```python
    def _create_session(self, force=False) -> None:
        if not self._session or force:
            #  define session headers
            self._headers = {
                'User-Agent': ('Mozilla/5.0 (Windows NT 6.3; Win64; x64) '
                               'AppleWebKit/537.36 (KHTML, like Gecko) '
                               'Chrome/68.0.3440.106 Safari/537.36'),
                'Accept': ('text/html,application/xhtml+xml, '
                           'application/xml;q=0.9,*/*;q=0.8'),
                'Accept-Language': '*'
            }

            #  initiate session
            self._session = aiohttp.ClientSession(headers=self._headers)
```
Get/post example code
```python
            resp = await self._session.get(site)
            post_resp = await self._session.post(site,
                                                 data=self._data)
```
Providing debug logs from my[ use case](https://gitlab.com/keatontaylor/alexapy/blob/master/alexapy/alexalogin.py#L305) as the actual code flow is a bit complex. Hopefully the debug logs can explain the issue. I will annotate below.

Get request and headers:
```
 2019-09-16 01:19:25 DEBUG (MainThread) [alexapy.alexalogin] GET:
 https://www.amazon.es/ap/signin?showRmrMe=1&openid.return_to=https://alexa.amazon.es/&openid.identity=http://specs.openid.net/auth/2.0/identifier_select&openid.assoc_handle=amzn_dp_project_dee_es&o enid.mode=checkid_setup&openid.claimed_id=http://specs.openid.net/auth/2.0/identifier_select&openid.ns=http://specs.openid.net/auth/2.0& with
[...]

PR #4066, opened 2019-09-17
Fix handling of expired cookies so they are not stored in CookieJar

alandtse:
## What do these changes do?

Properly schedule expiration date for cookies that are provided to the CookieJar with a date in the past which will result in such cookies being immediately expired by `_do_expiration()`.  Currently, the expiration date is set to the `timestamp()` (i.e., unix timestamp) which will make almost all values far in the future since it's compared to the `event_loop.time()` which increments from the start of the loop.

Marked WIP because this fix breaks the tests in `tests/test_cookiejar.py` for `test_constructor` and `TestCookieJarSafe.test_expires`.  I believe there are incorrect assumptions about behavior but I wanted to verify the test assumptions before making changes to the test suite. Mainly, should expired cookies be stored in the CookieJar and just never sent in requests? Or should they automatically be deleted?
The current test suite assumes expired cookies will not be deleted automatically.

I can complete the PR and the remaining steps once I understand the architecture.

CookieJar will automatically reject expired cookies and not store in the CookieJar
## Related issue number

#4063 
## Checklist

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -169,9 +169,13 @@ class CookieJar(AbstractCookieJar):
             if max_age:
                 try:
                     delta_seconds = int(max_age)
-                    self._expire_cookie(datetime.datetime.now(
-                        datetime.timezone.utc) +
-                        datetime.timedelta(seconds=delta_seconds),
+                    try:
+                        max_age_expiration = (
+                            datetime.datetime.now(datetime.timezone.utc) +
+                            datetime.timedelta(seconds=delta_seconds))
+                    except OverflowError:
+                        max_age_expiration = self.MAX_TIME
+                    self._expire_cookie(max_age_expiration,
                         domain, name)
                 except ValueError:
                     cookie["max-age"] = ""
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -185,8 +185,7 @@ class CookieJar(AbstractCookieJar):
                 if expires:
                     expire_time = self._parse_date(expires)
                     if expire_time:
-                        self._expire_cookie(datetime.datetime.fromtimestamp(
-                            expire_time.timestamp(), tz=datetime.timezone.utc),
+                        self._expire_cookie(expire_time,
                             domain, name)
                     else:
                         cookie["expires"] = ""
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -55,7 +55,7 @@ class CookieJar(AbstractCookieJar):
         self._cookies = defaultdict(SimpleCookie)  #type: DefaultDict[str, SimpleCookie]  # noqa
         self._host_only_cookies = set()  # type: Set[Tuple[str, str]]
         self._unsafe = unsafe
-        self._next_expiration: datetime.datetime = self._next_whole_second()
+        self._next_expiration = self._next_whole_second()
         self._expirations = {}  # type: Dict[Tuple[str, str], datetime.datetime]
 
     def save(self, file_path: PathLike) -> None:
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -175,7 +175,9 @@ class CookieJar(AbstractCookieJar):
                 if expires:
                     expire_time = self._parse_date(expires)
                     if expire_time:
-                        self._expire_cookie(expire_time.timestamp(),
+                        diff = (expire_time.timestamp() -
+                                datetime.datetime.now().timestamp())
+                        self._expire_cookie(diff,
                                             domain, name)
                     else:
                         cookie["expires"] = ""
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -6,7 +6,6 @@ import re
 import warnings
 from collections import defaultdict
 from http.cookies import BaseCookie, Morsel, SimpleCookie  # noqa
-from math import ceil
 from typing import (  # noqa
     DefaultDict,
     Dict,
@@ -48,15 +47,16 @@ class CookieJar(AbstractCookieJar):
 
     DATE_YEAR_RE = re.compile(r"(\d{2,4})")
 
-    MAX_TIME = 2051215261.0  # so far in future (2035-01-01)
+    MAX_TIME: datetime.datetime = datetime.datetime.max.replace(
+        tzinfo=datetime.timezone.utc)
 
     def __init__(self, *, unsafe: bool=False) -> None:
         self._loop = get_running_loop()
         self._cookies = defaultdict(SimpleCookie)  #type: DefaultDict[str, SimpleCookie]  # noqa
         self._host_only_cookies = set()  # type: Set[Tuple[str, str]]
         self._unsafe = unsafe
-        self._next_expiration = ceil(self._loop.time())
-        self._expirations = {}  # type: Dict[Tuple[str, str], int]
+        self._next_expiration: datetime.datetime = self._next_whole_second()
+        self._expirations: Dict[Tuple[str, str], datetime.datetime] = {}
 
     def save(self, file_path: PathLike) -> None:
         file_path = pathlib.Path(file_path)
@@ -71,7 +71,7 @@ class CookieJar(AbstractCookieJar):
     def clear(self) -> None:
         self._cookies.clear()
         self._host_only_cookies.clear()
-        self._next_expiration = ceil(self._loop.time())
+        self._next_expiration = self._next_whole_second()
         self._expirations.clear()
 
     def __iter__(self) -> 'Iterator[Morsel[str]]':
@@ -82,8 +82,16 @@ class CookieJar(AbstractCookieJar):
     def __len__(self) -> int:
         return sum(1 for i in self)
 
+    def _next_whole_second(self) -> datetime.datetime:
+        """Return current time rounded up to the next whole second."""
+        return (
+            datetime.datetime.now(
+                datetime.timezone.utc).replace(microsecond=0) +
+            datetime.timedelta(seconds=0)
+        )
+
     def _do_expiration(self) -> None:
-        now = self._loop.time()
+        now: datetime.datetime = datetime.datetime.now(datetime.timezone.utc)
         if self._next_expiration > now:
             return
         if not self._expirations:
@@ -102,12 +110,16 @@ class CookieJar(AbstractCookieJar):
         for key in to_del:
             del expirations[key]
 
-        self._next_expiration = ceil(next_expiration)
+        try:
+            self._next_expiration = (next_expiration.replace(microsecond=0) +
+                                     datetime.timedelta(seconds=1))
+        except OverflowError:
+            self._next_expiration = self.MAX_TIME
 
-    def _expire_cookie(self, when: float, domain: str, name: str) -> None:
-        iwhen = int(when)
-        self._next_expiration = min(self._next_expiration, iwhen)
-        self._expirations[(domain, name)] = iwhen
+    def _expire_cookie(self, when: datetime.datetime, domain: str, name: str
+                       ) -> None:
+        self._next_expiration = min(self._next_expiration, when)
+        self._expirations[(domain, name)] = when
 
     def update_cookies(self,
                        cookies: LooseCookies,
@@ -165,8 +177,10 @@ class CookieJar(AbstractCookieJar):
             if max_age:
                 try:
                     delta_seconds = int(max_age)
-                    self._expire_cookie(self._loop.time() + delta_seconds,
-                                        domain, name)
+                    self._expire_cookie(datetime.datetime.now(
+                        datetime.timezone.utc) +
+                        datetime.timedelta(seconds=delta_seconds),
+                        domain, name)
                 except ValueError:
                     cookie["max-age"] = ""
 
@@ -175,10 +189,9 @@ class CookieJar(AbstractCookieJar):
                 if expires:
                     expire_time = self._parse_date(expires)
                     if expire_time:
-                        diff = (expire_time.timestamp() -
-                                datetime.datetime.now().timestamp())
-                        self._expire_cookie(diff,
-                                            domain, name)
+                        self._expire_cookie(datetime.datetime.fromtimestamp(
+                            expire_time.timestamp(), tz=datetime.timezone.utc),
+                            domain, name)
                     else:
                         cookie["expires"] = ""
 
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -91,7 +91,7 @@ class CookieJar(AbstractCookieJar):
         )
 
     def _do_expiration(self) -> None:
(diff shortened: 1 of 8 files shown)


aiohttp/cookiejar.py
```diff
@@ -175,7 +175,9 @@ def update_cookies(self,
                 if expires:
                     expire_time = self._parse_date(expires)
                     if expire_time:
-                        self._expire_cookie(expire_time.timestamp(),
+                        diff = (expire_time.timestamp() -
```
asvetlov: From my understanding the first parameter of `_expire_cookie` is called `when` and it is an absolute time (seconds since UXIN epoch), not a delta with the current time.
I like to keep the parameter meaning.

alandtse: I don't believe that's a true assumption for `_expire_cookie`. Look at line 168 where it's used for max-age expiration. It's a time relative to the loop_time().  

The first parameter is the sum of the `self._loop.time()` + seconds left from max-age. It's a relative time and not tied to UNIX epoch because `loop.time()` is [isn't tied to the UNIX epoch](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.time).  Because `loop.time()` isn't tied to the UNIX epoch, the correct answer is to pass some value <= 0 which will cause an immediate expiration.  You could include some additional math functions to force it to be 0 if you don't want it to be negative, but that isn't necessary.

alandtse: If it's supposed to be `time.time()`, then I can easily change my edit to `time.time()+diff` for consistency.  Will that resolve your concern?  However, line 86 in `do_expiration()` will break because you're comparing `loop.time` which will always be much smaller than `time.time()`.  To resolve that I could just do a `loop.time() + diff`, which will also typically be a very negative number (again since unix epochs are so much larger).

For the test suite, I still need an answer on whether CookieJar should allow storage of expired cookies as that is a core assumption in the test suite.

asvetlov: I mean using time.time() consistently in `CookieJar` for representing absolute time looks like a proper fix. It doesn't need `loop.time()` at all, at least seems so.

alandtse: Yes, I think that looks right. Do you want me to make that switch throughout?

alandtse: It can be done with datetime unless you want to import time too.

asvetlov: Please do. `datetime` is even better, we never mix absolute and relative time again.

alandtse: Ok, what about the test suite assumption?

asvetlov: Do you mean tests that become fail after the code change?
Tests should be updated.

mjpieters:
There are client functional tests in [`test_client_functional.py`](https://github.com/aio-libs/aiohttp/blob/master/tests/test_client_functional.py), including

https://github.com/aio-libs/aiohttp/blob/dc39183d4351c2fc87dbfa9bcf938567d754571b/tests/test_client_functional.py#L1959-L1983

I'd copy that test as `test_set_cookies_expired` and verify that expired cookies are handled correctly.

aiohttp/cookiejar.py
mjpieters: There are indeed more issues here. If `datetime` is desired then I'd rework the time components in the whole class and use `datetime.datetime` instances with UTC time throughout.

* `MAX_TIME` should be `datetime.datetime.max.replace(tzinfo=datetime.timezone.utc)`
* `self._next_expiration = ceil(self._loop.time())` in the `CookieJar` `__init__` `_clear()`  and  should be set to `datetime.datetime.now(datetime.timezone.utc).replace(microsecond=0) + datetime.timedelta(seconds=0)` (so microseconds cleared and a second added to align with the next whole second).
* The `_do_expirations()` method should switch to use `now = datetime.datetime.now(datetime.timezone.utc)` and `self._next_expiration = next_expiration.replace(microsecond=0) + datetime.timedelta(seconds=1)`
* alter `_expire_cookie` to make `when` a `datetime` instance:

    ```python
    def _expire_cookie(self, when: datetime.datetime, domain: str, name: str) -> None:
        self._next_expiration = min(self._next_expiration, when)
        self._expirations[(domain, name)] = when
    ```
* Change the `max-age` handling to pass in `datetime.datetime.now(timezone.utc) + delta_seconds`.

Technically speaking `datetime.datetime.now(datetime.timezone.utc).replace(microsecond=0) + datetime.timedelta(seconds=0)` is not quite the same thing as `ceil(self._loop.time())` if you *happen* to get a `datetime.datetime()` instance with `microsecond` already set to 0, but I don't think it is worth explicitly testing for that here.

It is probably worth creating a helper function for it:

```python
def _next_whole_second() -> datetime.datetime:
    """The current time rounded up to the next whole second"""
    return (
        datetime.datetime.now(datetime.timezone.utc).replace(microsecond=0) +
        datetime.timedelta(seconds=0)
    )
```

alandtse: Ok I've implemented this; time to work on the tests and make sure I haven't broken anything.

alandtse:
Can I get this reviewed please? Looking at the failed tests it's either related to 3.5 syntax which I believe you are deprecating or components I didn't touch. `aiohttp\tests\test_worker.py`

Please note I added one dev dependency for freezegun because mocking datetime is a pain because it's a builtin.

aiohttp/cookiejar.py
```diff
@@ -48,15 +47,16 @@ class CookieJar(AbstractCookieJar):
...
 
     def __init__(self, *, unsafe: bool=False) -> None:
         self._loop = get_running_loop()
         self._cookies = defaultdict(SimpleCookie)  #type: DefaultDict[str, SimpleCookie]  # noqa
         self._host_only_cookies = set()  # type: Set[Tuple[str, str]]
         self._unsafe = unsafe
-        self._next_expiration = ceil(self._loop.time())
-        self._expirations = {}  # type: Dict[Tuple[str, str], int]
+        self._next_expiration: datetime.datetime = self._next_whole_second()
+        self._expirations: Dict[Tuple[str, str], datetime.datetime] = {}
```
mjpieters: Again, keep the hints in comments. 

```suggestion
        self._expirations = {}  # type: Dict[Tuple[str, str], datetime.datetime]
```

aiohttp/cookiejar.py
```diff
@@ -48,15 +47,16 @@ class CookieJar(AbstractCookieJar):
 
     DATE_YEAR_RE = re.compile(r"(\d{2,4})")
 
-    MAX_TIME = 2051215261.0  # so far in future (2035-01-01)
+    MAX_TIME: datetime.datetime = datetime.datetime.max.replace(
```
mjpieters: Don’t use variable annotations, this is the 3.5 syntax error you see. Put type hints in a comment:

```python 
MAX_TIME = datetime.datetime.max.replace(
    tzinfo=datetime.timezone.utc)  # type: datetime.datetime
```

I don’t think a type hint is needed here however; dropping it is fine:

```suggestion
    MAX_TIME = datetime.datetime.max.replace(
```

aiohttp/cookiejar.py
```diff
@@ -82,8 +82,16 @@ def __iter__(self) -> 'Iterator[Morsel[str]]':
     def __len__(self) -> int:
         return sum(1 for i in self)
 
+    def _next_whole_second(self) -> datetime.datetime:
```
mjpieters: This helper function should just be that: a function at module level.

aiohttp/cookiejar.py
```diff
@@ -82,8 +82,16 @@ def __iter__(self) -> 'Iterator[Morsel[str]]':
...
+        """Return current time rounded up to the next whole second."""
+        return (
+            datetime.datetime.now(
+                datetime.timezone.utc).replace(microsecond=0) +
+            datetime.timedelta(seconds=0)
+        )
+
     def _do_expiration(self) -> None:
-        now = self._loop.time()
+        now: datetime.datetime = datetime.datetime.now(datetime.timezone.utc)
```
mjpieters: No hint needed. Mypy can see what the function returns. 
```suggestion
        now = datetime.datetime.now(datetime.timezone.utc)
```

aiohttp/cookiejar.py
```diff
@@ -102,12 +110,16 @@ def _do_expiration(self) -> None:
         for key in to_del:
             del expirations[key]
 
-        self._next_expiration = ceil(next_expiration)
+        try:
+            self._next_expiration = (next_expiration.replace(microsecond=0) +
+                                     datetime.timedelta(seconds=1))
+        except OverflowError:
+            self._next_expiration = self.MAX_TIME
```
mjpieters, line 109: I like how you handled the possible overflow here :-)

aiohttp/cookiejar.py
```diff
@@ -165,8 +177,10 @@ def update_cookies(self,
             if max_age:
                 try:
                     delta_seconds = int(max_age)
-                    self._expire_cookie(self._loop.time() + delta_seconds,
-                                        domain, name)
+                    self._expire_cookie(datetime.datetime.now(
+                        datetime.timezone.utc) +
+                        datetime.timedelta(seconds=delta_seconds),
+                        domain, name)
```
mjpieters: You need to handle OverflowError here. A malicious or broken server could send max_age=_more seconds than time available in the universe_.

aiohttp/cookiejar.py
```diff
@@ -175,8 +189,9 @@ def update_cookies(self,
                 if expires:
                     expire_time = self._parse_date(expires)
                     if expire_time:
-                        self._expire_cookie(expire_time.timestamp(),
-                                            domain, name)
+                        self._expire_cookie(datetime.datetime.fromtimestamp(
+                            expire_time.timestamp(), tz=datetime.timezone.utc),
+                            domain, name)
```
mjpieters: The `expire_time` value here is a perfectly serviceable UTC value already, why convert again?

aiohttp/cookiejar.py
```diff
@@ -48,15 +47,16 @@ class CookieJar(AbstractCookieJar):
...
+        tzinfo=datetime.timezone.utc)
 
     def __init__(self, *, unsafe: bool=False) -> None:
         self._loop = get_running_loop()
         self._cookies = defaultdict(SimpleCookie)  #type: DefaultDict[str, SimpleCookie]  # noqa
         self._host_only_cookies = set()  # type: Set[Tuple[str, str]]
         self._unsafe = unsafe
-        self._next_expiration = ceil(self._loop.time())
-        self._expirations = {}  # type: Dict[Tuple[str, str], int]
+        self._next_expiration: datetime.datetime = self._next_whole_second()
```
mjpieters: Drop the hint here

```suggestion
        self._next_expiration = self._next_whole_second()
```

requirements/dev.txt
```diff
@@ -1,3 +1,4 @@
 -r ci.txt
 -r towncrier.txt
 cherry_picker==1.3.2; python_version>="3.6"
+freezegun==0.3.12
```
mjpieters: This should be added to ci.txt I think, not here. Appveyor breaks because this is the wrong location for the dependency.

aiohttp/cookiejar.py
alandtse: Ok, I thought the plan was to deprecate 3.5, but will revert.

requirements/dev.txt
alandtse: Moved.

aiohttp/cookiejar.py
alandtse: Moved.

aiohttp/cookiejar.py
alandtse: Thanks.

aiohttp/cookiejar.py
```diff
@@ -48,15 +47,16 @@ class CookieJar(AbstractCookieJar):
...
 
     def __init__(self, *, unsafe: bool=False) -> None:
         self._loop = get_running_loop()
         self._cookies = defaultdict(SimpleCookie)  #type: DefaultDict[str, SimpleCookie]  # noqa
         self._host_only_cookies = set()  # type: Set[Tuple[str, str]]
         self._unsafe = unsafe
-        self._next_expiration = ceil(self._loop.time())
-        self._expirations = {}  # type: Dict[Tuple[str, str], int]
+        self._next_expiration = self._next_whole_second()
+        self._expirations = {}  # type: Dict[Tuple[str, str], datetime.datetime]
```
mjpieters: This line is now too long for flake8. Line 56 had the same issue and used a noqa flag to tell flake8 to ignore it. We could do the same here but I’d prefer it to be specific:

```suggestion
        self._expirations = {}  # type: Dict[Tuple[str, str], datetime.datetime] # noqa: E501
```

@asvetlov remind me to talk to you about flake8-bugbear and the flexible line length overflow setting. :-) if this project is to use black in future that’s a good addition anyway.

alandtse: Change made before I saw the suggestion. It's also the reason I went to inline type hints. ;)

alandtse:
Decided to just add the test for max-age overflow since it wasn't hard.

tests/test_client_functional.py
```diff
@@ -2032,6 +2033,30 @@ def dumps(obj):
...
 
+async def test_set_cookies_max_age_overflow(aiohttp_client) -> None:
+
+    async def handler(request):
+        ret = web.Response()
+        ret.headers.add('Set-Cookie',
+                        'overflow=overflow; '
+                        'HttpOnly; Path=/'
+                        " Max-Age=" +
+                        str(int(datetime.datetime.max.timestamp())) + "; ")
```
mjpieters: 👍

mjpieters: Unfortunately, the very act of asking for the time stamp can trigger an overflow exception, based on the current timezone! See the [appveyor test failures](https://ci.appveyor.com/project/aio-libs/aiohttp/builds/27616968/job/4i6eq555evqijl56#L3419):

```python
ERROR    aiohttp.server:web_protocol.py:380 Error handling request
  File "C:\projects\aiohttp\aiohttp\web_protocol.py", line 405, in _handle_request
    resp = await self._request_handler(request)
  File "C:\projects\aiohttp\aiohttp\web_app.py", line 330, in _handle
    resp = await handler(request)
  File "C:\projects\aiohttp\tests\test_client_functional.py", line 2044, in handler
    str(int(datetime.datetime.max.timestamp())) + "; ")
OverflowError: timestamp out of range
```

`datetime.max` Is a naive datetime value and has no timezone but a timestamp is always a value relative to 1970-01-01 midnight UTC, so Python uses the local timezone to interpret naive values when converting. In timezones west from UTC that requires adding hours to the datetime value, and so we get an overflow here.

alandtse: Oops. I guess I just need to choose a static large int value to avoid that issue.

alandtse:
I honestly don't know why the Travis build failed.

Edit: Nevermind, figured it out.

mjpieters:
Looks great to me; @asvetlov, @webknjaz, any remarks?

PR #4079, opened 2019-09-20
Drop custom compilation command

asvetlov:
Builtin realization does everything that we need
after strict splitting on accelerated and pure Python builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,17 +11,16 @@ if sys.version_info < (3, 5, 3):
 
 
 NO_EXTENSIONS = bool(os.environ.get('AIOHTTP_NO_EXTENSIONS'))  # type: bool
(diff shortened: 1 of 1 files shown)


asvetlov:
Extraction from #3827 by @webknjaz request

setup.py
```diff
@@ -22,7 +18,8 @@
 
 here = pathlib.Path(__file__).parent
 
-if (here / '.git').exists() and not (here / 'vendor/http-parser/README.md').exists():
+if ((here / '.git').exists() and
```
webknjaz: if you're refactoring this already, i'd do smth like 
```suggestion
IS_GIT_REPO = (here / '.git').exists()

if (IS_GIT_REPO and
```

PR #4080, opened 2019-09-21
Don't cancel web handler on disconnection

asvetlov:
Raise an error on reading/writing instead.


tests/test_client_ws.py
```diff
@@ -364,7 +363,7 @@ def read(self, decode=False):
                     await resp.close()
 
 
-async def test_send_data_after_close(ws_key, key_data, loop, mocker) -> None:
+async def test_send_data_after_close(ws_key, key_data, loop) -> None:
```
webknjaz, line 366: Can we decorate this with `@pytest.mark.parametrize` instead of looping internally? This would be better represented in the test results output then.

tests/test_web_protocol.py
```diff
@@ -821,7 +821,11 @@ def test_data_received_force_close(srv) -> None:
 async def test_client_disconnect(aiohttp_server) -> None:
 
     async def handler(request):
-        await request.content.read(10)
+        buf = b""
+        with pytest.raises(ConnectionError):
+            while len(buf) < 10:
+                buf += await request.content.read(10)
```
webknjaz, line 847: Maybe 
```suggestion
                buf += await request.content.read(10-len(buf))
```
?

tests/test_client_ws.py
asvetlov: I don't care too much.
The test did exist, I have no intention to rewrite it.

tests/test_web_protocol.py
asvetlov: The same honestly.

asvetlov:
The PR is ready for review

aiohttp/web_protocol.py
```diff
@@ -264,8 +269,10 @@ def connection_lost(self, exc: Optional[BaseException]) -> None:
         if self._keepalive_handle is not None:
             self._keepalive_handle.cancel()
 
-        if self._task_handler is not None:
-            self._task_handler.cancel()
+        if self._current_request is not None:
+            if exc is None:
+                exc = ConnectionResetError("Connetion lost")
```
nicktimko: Connection

nicktimko, line 274: Minor spelling: 'Connection'

asvetlov: Thanks! Fixed by a6af0838

PR #4090, opened 2019-09-24
Fix multipart file uploads without content type

### LINKED ISSUE #4089: File uploads without a content type are treated as ordinary fields
## Long story short

Multipart file uploads do not work correctly since aiohttp 3.6.0 unless a content type is provided. This used to work up to aiohttp 3.5.4 and broke due to #3905, [this line specifically](https://github.com/aio-libs/aiohttp/pull/3905/files#diff-6e3c6d0970cb316846993e9d988175fdR639).

## Expected behaviour

File upload data is treated as a binary data stream if no content type is provided.

## Actual behaviour

File upload data is treated as an ordinary text field, possibly causing an UnicodeDecodeError.

## Steps to reproduce

1. Send a request to an aiohttp server with a multipart/form-data payload, where one item is a file without a content type. The requests library will not provide a content type unless you specifically set one.
2. Call `await request.post()` in your aiohttp server.

## Your environment

This issue is causing [integration tests for bravado-asyncio to fail](https://travis-ci.org/sjaensch/bravado-asyncio/builds/581878297) - the build before this one [passed](https://travis-ci.org/sjaensch/bravado-asyncio/builds/581426849), the only change is the upgrade from aiohttp 3.5.4 to aiohttp 3.6.0.

sjaensch:
## What do these changes do?

Multipart form data with files without a content type were not treated correctly anymore by the aiohttp server.

This restores the behavior as it was up to aiohttp 3.5.4; for aiohttp 3.6.0 and aiohttp 3.6.1 it would treat such file uploads as normal fields, including trying to utf8-decode the data.

Fixes #4089.

## Checklist


sjaensch:
I'm noticing the `mypy` failure now, I'll work on a fix tomorrow.

aiohttp/web_request.py
```diff
@@ -613,7 +613,7 @@ def body_exists(self) -> bool:
                 field_ct = field.headers.get(hdrs.CONTENT_TYPE)
 
                 if isinstance(field, BodyPartReader):
-                    if field.filename and field_ct:
+                    if field.filename:
```
webknjaz, line 626: Could you please attach a comment pointing to the right place in the RFC to prevent this from happening in the future.

webknjaz: Sure, I saw it. I just wanted a code comment here :)

aiohttp/web_request.py
```diff
@@ -613,7 +613,10 @@ def body_exists(self) -> bool:
                 field_ct = field.headers.get(hdrs.CONTENT_TYPE)
 
                 if isinstance(field, BodyPartReader):
-                    if field.filename and field_ct:
+                    # Note that according to RFC 7578, the Content-Type header
+                    # is optional, even for files, so we can't assume it's present.
```
webknjaz: Plz fit this line in 79 chars: https://github.com/aio-libs/aiohttp/pull/4090/checks?check_run_id=235617239#step:7:7

tests/test_web_request.py
```diff
@@ -610,6 +610,26 @@ def test_clone_headers_dict() -> None:
...
 
+async def test_multipart_formdata_file(protocol) -> None:
+    # Make sure file uploads work, even without a content type
+    payload = StreamReader(protocol, loop=asyncio.get_event_loop())
+    payload.feed_data(b"""-----------------------------326931944431359\r
+Content-Disposition: form-data; name="a_file"; filename="binary"\r
+\r
+\ff\r
+-----------------------------326931944431359--\r\n""")
+    content_type = "multipart/form-data; boundary="\
```
webknjaz: Could you please avoid escaping EOL? Sometimes this causes unobvious bugs.
do smth like
```python
content_type = (
    'str1'
    'str2'
)
```

tests/test_web_request.py
```diff
@@ -610,6 +610,26 @@ def test_clone_headers_dict() -> None:
     assert dict(result) == {'a': 'b', 'c': 'd'}
 
 
+async def test_multipart_formdata_file(protocol) -> None:
+    # Make sure file uploads work, even without a content type
+    payload = StreamReader(protocol, loop=asyncio.get_event_loop())
+    payload.feed_data(b"""-----------------------------326931944431359\r
```
webknjaz: How about `\r\n` delimiters? I think, this should be tested too.

aiohttp/web_request.py
```diff
@@ -613,7 +613,10 @@ def body_exists(self) -> bool:
                 field_ct = field.headers.get(hdrs.CONTENT_TYPE)
```
webknjaz, line 619: It looks like according to https://tools.ietf.org/html/rfc7578#section-4.4 it should be
```suggestion
                 field_ct = field.headers.get(hdrs.CONTENT_TYPE, 'text/plain')
```
shouldn't it?

sjaensch: It's the default according to the RFC, but only matters to whoever processes the contents of the file. I mentioned this in #4089: Setting the content type means the user will not be able to distinguish between a file uploaded with the content type of `text/plain` and one being uploaded with no content type. I'm not sure what the value would be in setting it, especially since it might be wrong.

What users typically do is use the file name / extension for guessing the content type, or they know the expected type already. In any case, the default was meant for non-file parts. It actually mentions in the next sentence that for files, a content type should be sent, and if a more appropriate cannot be found it should be `application/octet-stream`. Which makes a lot more sense.

aiohttp/web_request.py
sjaensch: Done, thanks!

tests/test_web_request.py
sjaensch: Totally, I just copied the other test and adapted it. Fixed.

tests/test_web_request.py
sjaensch: In fact we need `\n` in there. :) Python inserts them automatically when there is a newline:

```
>>> b"""one line\r
... second line"""
b'one line\r\nsecond line'
```

asvetlov:
> Minor nit: I'd use a docstring instead of a code comment since it could be shown in the pytest report while comments get omitted by the parser.

@webknjaz please no. Feel free to apply this rule for your own projects but not here.
The reason is exactly the same: I prefer to see function names over meaningless doctsting.
Func name can be used for fast failed test lookup or filtering, e.g. `pytest -k test_pause_writing`.
Filtering by docstring is complicated.

tests/test_web_request.py
webknjaz: this should probably be made more obvious by putting them there explicitly:
```python
payload.feed_data(
    b"adf\r\n"
    b"afds\r\n"
    b"afds\r\n"
)
```

webknjaz: Also because line separators may be different under different envs.

aiohttp/web_request.py
kornicameister: Well, at least there is an agreement that RFC is there to be followed. If a part is detected to be a file part, defaulting to `application/octet-stream` is totally acceptable given that it conform to RFC.

asvetlov:
Thanks for the PR.
I need time for careful review, sorry.

sjaensch:
@asvetlov any update on this? I think the change is pretty safe, makes sure we detect file uploads correctly, and improves compatibility. This issue prevents me from using a version of aiohttp that is compatible with Python 3.8.

asvetlov:
I wanted to merge it before publishing a new release.
IIRC multipart has other problems...

asvetlov:
Superseded by #5072

PR #4116, opened 2019-09-29
Upgrade to mypy==0.730

asvetlov:


asvetlov:
Updated to mypy 0.740 in another PR

Issue #3866, opened 2019-06-25
UrlDispatcher.add_routes to return list of resource objects


ntai:
## Long story short

UrlDispatcher.add_routes returns nothing. I'd like to see a list of resources returned.

right now, 

```
    def add_routes(self, routes):
        for route_obj in routes:
            route_obj.register(self)
```
What I want - 
```
    def add_routes(self, routes):
        return [ route_obj.register(self) for route_obj in routes ]
```
Not sure route_obj.register returns a resource.

## Expected behaviour

resources = router.add_routes(routes)

## Actual behaviour

router.add_routes(routes) returns nothing

## Steps to reproduce

This is an enhancement request.

## Your environment
aiohttp 3.0.1
Python 3.6
Ubuntu 18.04 LTS

asvetlov:
Why? What is your motivation?

ntai:
I can hand it over to cors to do post-add_route. Right now, I have to get to resource by app.router._resources.
if I'm adding multiple routes to top-level but with different cors, it's a chore to separate them out after add_routes called, but if add_routes returns a list of resources from the routes, it's a piece of cake.
```
    for resource in app.router.add_routes(routes):
      cors.add(resource, { '*': aiohttp_cors.ResourceOptions(allow_credentials=True, expose_headers="*", allow_headers="*") })
```

asvetlov:
Not so easy.
First of all, `add_routes()` should return a list of *routes* if any, just by name.
*Resource* can have multiple routes, a route per supported HTTP method.
Secondly, now `router.add_resoouce()` can reuse the latest resource if possible. There is a case when `add_routes()` doesn't add a new resource at all.
Thirdly, there is `app.router.resources()` API, there is no need to access the private `_resources` method.

ntai:
Here is what I ended up doing to get around the aiohttp, aiohttp_cors and socket.io.

```
    app.router.add_routes(routes)
    app.router.add_static("/", rootdir)
    for resource in app.router._resources:
      if resource.raw_match("/socket.io/"):
        continue
      cors.add(resource, { '*': aiohttp_cors.ResourceOptions(allow_credentials=True, expose_headers="*", allow_headers="*") })
      pass
```
If I can get the resources back from add_routes, I don't have to exclude the "socket.io" to declare cors header.

if add_routes returns the newly added resources, that is perfectly fine. I didn't know it's not one-to-one match. The interaction between socket.io and aiohttp_cors is pretty cumbersome.

[...]

PR #4141, opened 2019-10-04
UrlDispatcher - add_routes returns a list of AbstractRoutes

ZlatSic:
## What do these changes do?

These changes are supposed to solve the issue #3866, this is just the basic infrastructure of the changes, some discussion about the changes is still required.

`aiohttp.web.UrlDispatcher.add_routes` and `aiohttp.web.Application.add_routes` return a list of the registered `AbstractRoute` instances. As a consequence, `RouteDef.register` now also returns the registered `AbstractRoute`

## Related issue number

#3866 

## Checklist


zlatsic:
First, thank you for the feedback at #3866, but a part of my question was still left unanswered - what about the static routes? I figured it made more sense to make a PR so we can discuss it here where we can reference the code.

aiohttp/web_routedef.py
```diff
@@ -36,7 +39,7 @@
 
 class AbstractRouteDef(abc.ABC):
     @abc.abstractmethod
-    def register(self, router: UrlDispatcher) -> None:
+    def register(self, router: UrlDispatcher) -> Optional[AbstractRoute]:
```
zlatsic: This is basically the part I have the most questions about. So far, I've made this return value `Optional` (and everywhere else where its return value is propagated). The reason for this is because of the StaticDef and its `register` implementation. Since `add_static` returns an `AbstractResource`, do you think it would be sensible for this to be a union of `AbstractRoute` and `AbstractResource` for that reason?

aiohttp/web_app.py
```diff
@@ -250,8 +251,10 @@ def add_domain(self, domain: str,
         factory = partial(MatchedSubAppResource, rule, subapp)
         return self._add_subapp(factory, subapp)
 
-    def add_routes(self, routes: Iterable[AbstractRouteDef]) -> None:
-        self.router.add_routes(routes)
+    def add_routes(
+            self, routes: Iterable[AbstractRouteDef]
+    ) -> List[Optional[AbstractRoute]]:
```
asvetlov: `Optional` is bad here. The return type should be a strict list.
```suggestion
    ) -> List[AbstractRoute]:
```

aiohttp/web_routedef.py
asvetlov: `StaticResource` internally has `self._routes` which is a dict of `ResourceRoute` instances.
Perhaps we can extend `resource.get_info()` to return routes mapping; it can avoid access to private members.

zlatsic:
All, right, I've updated the changes in the `AbstractRouteDef` and its subclasses, now `register` returns a list of registered `AbstractRoutes`, so now there's no more need for the `Optional` return values. I've also updated a test for the `get_info` so that it checks for the field `"routes"` and whether its values are `StaticResource`s.

I have another question about the tests: I've noticed that there weren't any that check the `AbstractRouteDef.register` functions, `UrlDispatcher.add_routes` and `Application.add_routes` - do you think it's sensible to add them now?

docs/web_reference.rst
```diff
@@ -1413,6 +1413,8 @@ duplicated like one using :meth:`Application.copy`.
       The table is a :class:`list` of :class:`RouteDef` items or
       :class:`RouteTableDef`.
 
+      Returns a :class:`list` of registered :class:`AbstractRoute` instances.
```
asvetlov: 1. Please use `:returns:` markup as in the method above.
2. Please add `.. versionchanged:: 3.7` markup, search in docs for the directive usage. The master is not ready for 4.0 release yet, I think we can publish 3.7 earlier.

docs/web_reference.rst
```diff
@@ -1569,6 +1571,8 @@ Router is any object that implements :class:`AbstractRouter` interface.
...
```
asvetlov: The same

docs/web_reference.rst
```diff
@@ -2080,6 +2084,8 @@ The definition is created by functions like :func:`get` or
 
       Abstract method, should be overridden by subclasses.
 
+      Returns a list of registered `AbstractRoute` objects.
```
asvetlov: And the same

zlatsic:
I've looked up how it's used, hope I got `versionchanged` right, I put it at the bottom of the method definitions (and the change description), while keeping the `returns` markup at under the function definition and input argument descriptions.

docs/web_reference.rst
```diff
@@ -1413,12 +1413,19 @@ duplicated like one using :meth:`Application.copy`.
...
+
       The method is a shortcut for
       ``app.router.add_routes(routes_table)``, see also
       :meth:`UrlDispatcher.add_routes`.
 
       .. versionadded:: 3.1
 
+      .. versionchanged:: 3.7
+
+      Return value updated from ``None`` to :class:`list` of
```
```suggestion
         Return value updated from ``None`` to :class:`list` of
```

docs/web_reference.rst
```diff
@@ -1413,12 +1413,19 @@ duplicated like one using :meth:`Application.copy`.
...
+      :class:`AbstractRoute` instances.
```
```suggestion
         :class:`AbstractRoute` instances.
```

docs/web_reference.rst
```diff
@@ -1569,8 +1576,15 @@ Router is any object that implements :class:`AbstractRouter` interface.
       The table is a :class:`list` of :class:`RouteDef` items or
       :class:`RouteTableDef`.
 
+      :returns: :class:`list` of registered :class:`AbstractRoute` instances.
+
       .. versionadded:: 2.3
 
+      .. versionchanged:: 3.7
+
+      Return value updated from ``None`` to :class:`list` of
```
```suggestion
         Return value updated from ``None`` to :class:`list` of
```

docs/web_reference.rst
```diff
@@ -1569,8 +1576,15 @@ Router is any object that implements :class:`AbstractRouter` interface.
...
+      :class:`AbstractRoute` instances.
```
```suggestion
         :class:`AbstractRoute` instances.
```

docs/web_reference.rst
```diff
@@ -2080,6 +2094,13 @@ The definition is created by functions like :func:`get` or
 
       Abstract method, should be overridden by subclasses.
 
+      :returns: :class:`list` of registered :class:`AbstractRoute` objects.
+
+      .. versionchanged:: 3.7
+
+      Return value updated from ``None`` to :class:`list` of
```
```suggestion
         Return value updated from ``None`` to :class:`list` of
```

docs/web_reference.rst
```diff
@@ -2080,6 +2094,13 @@ The definition is created by functions like :func:`get` or
...
+      :class:`AbstractRoute` instances.
```
```suggestion
         :class:`AbstractRoute` instances.
```

asvetlov:
Thanks!

zlatsic:
Don't mention it, glad to help on a library I often use 😄

PR #4184, opened 2019-10-14
Drop Travis CI and Appveyor from build matrix

asvetlov:
There is no need in them since we test and deploy everything on Azure Pipelines (which is much faster BTW)


webknjaz:
Azure is reported to have broken cron that stops working suddenly w/o any sane reason. We could limit Travis to cron jobs instead of removing it.

asvetlov:
IIRC we have daily Travis build for checking very slow MacOS test jobs.
Since mac is pretty fast on azure I don't see a reason for daily activity.

Moreover, I have a daily schedule for another project on my job.
It works well.
If we will need scheduled activities sometime and azure will fail on it -- we can manage the situation (or wait for the next daily job run).

webknjaz:
Travis provides a different env where things would fail differently from other CIs. This means that we'll no longer be able to catch those corner cases and will cause silent regressions.

asvetlov:
I can live with this.
My primary goal is saving my human time and time for other contributors.
Single CI solves the problem.
Multiple CIs can catch something potentially but I bet on users but reporting.

PR #4192, opened 2019-10-16
Drop GitHub workflows

asvetlov:


asvetlov:
We have everything working on Azure, linters included.
Why do we need a separate linters checker *plus* the same checker on Azure (which is ultimatelly needed for tests anyway).

P.S.
Azure has secrets too (and I use it now for the project).

Issue #4174, opened 2019-10-10
decorator to add per handler annotations


gjcarneiro:
## Long story short

With my custom authentication, trying to have an authentication middleware that requires login by default, except for the handlers specifically annotated with a decorator. 

```python

def login_exempt(func: F) -> F:
    func.weblogin_login_exempt = True  # type: ignore
    return func

@web.middleware
async def auth_middleware(
    request: web.Request,
    handler: Callable[[web.Request], Awaitable[web.StreamResponse]],
) -> web.StreamResponse:
    print(handler, hasattr(handler, "weblogin_login_exempt"))
    if not hasattr(handler, "weblogin_login_exempt"):
        session = await weblogin.fetch_session(request)
        if not session:
            raise web.HTTPNotFound(
                text=json.dumps(weblogin.SESSION_NOT_FOUND),
                content_type="application/json",
            )
    return await handler(request)


def create_web_app() -> web.Application:
    app = web.Application(
        middlewares=[weblogin.session_middleware, auth_middleware]
    )
    app.on_startup.append(async_init)
    app.on_shutdown.append(do_shutdown)
    app.router.add_route("GET", "/healthz", healthz)

    app.add_subapp("/s/weblogin", weblogin.create_web_app())
    app.add_subapp("/s/userprefs", userprefs.create_web_app())

```

## Expected behaviour

I expect the handler to have `weblogin_login_exempt` when it has been decorator, causing the framework to skip the usual auth checks.

## Actual behaviour

Instead, at least in the case of a subapp, the handler is a functools.partial(), which means the actual handler function is hidden away.

```
functools.partial(<function _fix_request_current_app.<locals>.impl at 0x7efbf9fb48c8>, handler=<function login at 0x7efbf9f7ba60>)
```

Only for the toplevel Application is the handler unmodified.  For a subapp, handler is wrapped as a functools.partial() object.

In general, I am either missing something obvious, or there is a lack of an API in aiohttp to achieve this sort of thing.  I would like to be able to annotate handlers, somehow, and for middleware to fetch the handler annotations. 

## Your environment

aiohttp 3.6.1, python 37.3.

asvetlov:
You are writing a middleware that is applicable to only a subset of application handlers.
I don't think that this is a good idea.

gjcarneiro:
It is only "not a good idea" because it's not possible to do it atm.  But I think I have a great use case.

[...]

PR #4195, opened 2019-10-16
Preserve handler function attributes across middlewares

gjcarneiro:
## What do these changes do?
When using middlewares (such as the hidden middleware implicitly added by subapps), it fixes the functools.partial() usage on the handler view, to copy any attributes from the view function.

Now any attributes added to a view handler function, e.g. by a decorator, are seen by all the middlewares in the chain.

Before this change, attributes we only seen by a middleware in the case of the last middleware (first to run), and only for the main Application, for subapps that wasn't true.

## Related issue number

#4174 

## Checklist

### DIFF
--- a/CHANGES/4174.bugfix
+++ b/CHANGES/4174.bugfix
@@ -1,0 +1,1 @@
+Preserve view handler function attributes across middlewares
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -1,7 +1,7 @@
 import asyncio
 import logging
 import warnings
-from functools import partial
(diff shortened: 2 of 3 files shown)


gjcarneiro:
Pushed a new test_middleware_subapp test.

gjcarneiro:
Thanks!  Well, for me, personally, there is no need to backport it, necessarily.  I have an ugly workaround in the code, it can wait a few months.

asvetlov:
Thanks.
The PR can be backported easily, there is no reason to postpone the feature delivery.

Issue #2492, opened 2017-11-09
Change web handlers cancellation policy


asvetlov:
Cancelling a web handler task on client disconnection is the most unexpected and annoying feature of aiohttp web server.
People from classic WSGI world don't expect it. 
Moreover tornado doesn't support a cancellation (I hope it will in 5.0 release).
 
The proposal is:
* *web-handler* cancellation should be opt-in, not opt-out (by aiojobs and async_armor usage).
* Application should have *allow_cancel* constructor parameter, `False` by default.
* Any record in application route table should accept *allow_cancel* to override the default.

Opinions?

webknjaz:
Yes, this should be configurable at (1) global and (2) handler level.
We could provide some decorator (`@cancel_safe`?) explicitly enabling cancellation at handler level and it would be harder to do this path-dependent (probably via middlewares).
As for constructor, I'd go for subclassing, like: `class CalcellableApp(Application):`

gjcarneiro:
Personally, I like the current policy.  Typically, inside a view only a small portion of the view code needs to be protected from cancellation.  For that, you can wrap it inside `asyncio.shield()`.

asvetlov:
PR #4080 replaces task cancellation with `OSError` raised on trying to communicate with a closed peer (read request BODY or send data back).
By this, the behavior becomes more obvious; the server code is still notified about peer disconnection but in a more predictable way on better-specified points.

We can close the issue.

gjcarneiro:
How does this handle the case?
1. We receive a request;
2. Handler does a long and complicated DB query;
3. Client loses patience and closes the connection.

With the old behaviour, the query would be automatically cancelled.  With the new PR, it doesn't get cancelled, and so the database is doing a lot of useless work.

gjcarneiro:
I think the original proposal was reasonable, making cancellation opt-in, and per route, but #4080 just removes cancellation completely.  I think #4080 removes an important functionality from aiohttp.

How can we even detect client disconnection?  There's only two points where we can: (1) at the beginning, (2) at the end.  If the client disconnects 5 seconds into a query that takes 20 seconds, we will not be notified.  The handler only finds out at the end of the query, when it's about to write the response.

asvetlov:
The previous approach required protecting user code from cancellation.
This behavior was constantly ignored by users. Even people remember to add shield/@atomic guards to their code the code becomes overcomplicated without a reason.

[...]

PR #4200, opened 2019-10-16
Add a Request.wait_for_disconnection() method

gjcarneiro:
as means of allowing request handlers to be notified of premature client
disconnections.

## What do these changes do?

Adds a new async method `Request.wait_for_disconnection()`.  This method does nothing except to wait until the connection that sent a request to be disconnected.

This may be useful, in certain cases, now that aiohttp no longer automatically cancels request handlers upon client disconnection.  E.g., if there is a slow DB query, it is beneficial to be able to cancel that DB query as soon as the client disconnects, to avoid the DB doing useless work.

This async method provides the needed "signal" for that.  Although making use of it is not pretty, it can be simplified by a utility function/wrapper, probably outside the scope of aiohttp (and outside the scope of this PR).

A new method was added, as described above.  It is opt-in, so should not affect existing code.

## Related issue number

#2492 

## Checklist

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -703,13 +703,6 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
             fut.cancel()
 
     async def wait_for_disconnection(self) -> None:
-        """Returns when the connection that sent this request closes
-
-        This can be used in handlers as a means to receive a notification of
-        premature client disconnection.
-
-        .. versionadded:: 4.0
-        """
         fut = asyncio.Future()  # type: asyncio.Future[None]
         self._disconnection_waiters.add(fut)
         try:
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -703,7 +703,8 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
             fut.cancel()
 
     async def wait_for_disconnection(self) -> None:
-        fut = asyncio.Future()  # type: asyncio.Future[None]
+        loop = asyncio.get_event_loop()
+        fut = loop.create_future()  # type: asyncio.Future[None]
         self._disconnection_waiters.add(fut)
         try:
             await fut
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -698,6 +698,10 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         for fut in self._disconnection_waiters:
             set_result(fut, None)
 
+    def _finish(self) -> None:
+        for fut in self._disconnection_waiters:
+            fut.cancel()
+
     async def wait_for_disconnection(self) -> None:
         """Returns when the connection that sent this request closes
 
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -17,6 +17,7 @@ from typing import (  # noqa
     Mapping,
     MutableMapping,
     Optional,
+    Set,
     Tuple,
     Union,
     cast,
@@ -35,6 +36,7 @@ from .helpers import (
     is_expected_content_type,
     reify,
     sentinel,
+    set_result,
 )
 from .http_parser import RawRequestMessage
 from .multipart import BodyPartReader, MultipartReader
@@ -110,7 +112,9 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         '_message', '_protocol', '_payload_writer', '_payload', '_headers',
         '_method', '_version', '_rel_url', '_post', '_read_bytes',
         '_state', '_cache', '_task', '_client_max_size', '_loop',
-        '_transport_sslcontext', '_transport_peername')
+        '_transport_sslcontext', '_transport_peername',
+        '_disconnection_waiters',
+    )
 
     def __init__(self, message: RawRequestMessage,
                  payload: StreamReader, protocol: 'RequestHandler',
@@ -142,6 +146,7 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         self._task = task
         self._client_max_size = client_max_size
         self._loop = loop
+        self._disconnection_waiters = set()  # type: Set[asyncio.Future[None]]
 
         transport = self._protocol.transport
         assert transport is not None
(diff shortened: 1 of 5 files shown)


gjcarneiro:
One check is failing because of:

> The job running on agent Azure Pipelines 7 has exceeded the maximum execution time of 10. For more information, see https://go.microsoft.com/fwlink/?linkid=2077134
> The operation was canceled.

I don't think it's my fault that it's failing.  Possibly the CI environment is broken.  Limiting to 10 minutes seems rather restrictive...

asvetlov:
Yes, tests hang on MacOS periodically.
I'm working on the fix. It is reproduced on a relative heavy load.

aiohttp/web_request.py
```diff
@@ -690,6 +694,21 @@ def __bool__(self) -> bool:
 
     def _cancel(self, exc: BaseException) -> None:
         self._payload.set_exception(exc)
+        for fut in self._disconnection_waiters:
+            fut.set_result(None)
```
asvetlov: Welcome to low-level asyncio!
This line contains a very common but annoying bug. Let me describe the failure scenario:
1. a task calls `await request.wait_for_disconnection()` and waits
2. the task gets canceled by explicit `task.cancel()` call
3. eventually the connection is broken by peer. `request._cancel()` is called; which in turn calls `fut.set_result(None)`.
4. The future is canceled by bullet 2, `set_result()` fails.

Please use `helpers.set_result()`.

asvetlov (changes requested):
Please fix my note and update the documentation.

`.. versionadded::` should use `4.0` as you use 4.0-specific `request._cancel()`

aiohttp/web_request.py
gjcarneiro: Ah, good catch.  Changed to use helpers.set_result.

gjcarneiro: versionadded...
Note: this PR needs rebase/squash commits before merging.

asvetlov (changes requested):
1. A client opens a connection in HTTP keep-alive mode
2. server code calls `await request.wait_for_disconnection()`
3. HTTP request is handled *without* socket closing (keep-alive mode is on)
4. `await request.wait_for_disconnection()` hangs forever

Please fix it

gjcarneiro:
> 1. A client opens a connection in HTTP keep-alive mode
> 2. server code calls `await request.wait_for_disconnection()`
> 3. HTTP request is handled _without_ socket closing (keep-alive mode is on)
> 4. `await request.wait_for_disconnection()` hangs forever
> 
> Please fix it

The method name clearly says "wait for disconnection".  In case "HTTP request is handled _without_ socket closing (keep-alive mode is on)" there is no disconnection. That is a use case I expected.

The way I envision this to be used, then `request.wait_for_disconnection()` coroutine call would always be cancelled at the end of each request, by the app code.  This is meant to be just a building block for something more convenient to use, like a decorator.  From the code I posted on the issue, I can create a decorator like:

```python

def cancel_on_disconnect(handler):
  @functools.wraps(handler)
  async def wrap_handler(request):
     wait = asyncio.create_task(request.wait_for_disconnection())
     bgtask = asyncio.create_task(handler())
     asyncio.wait([wait, bgtask], return_when=asyncio.FIRST_COMPLETED)
     if bgtask.done():
         wait.cancel()
         return web.Response(bgtask.result())
     bgtask.cancel()
     raise web.HTTPGatewayTimeout()
  return wrap_handler

@cancel_on_disconnect
def my_handler(request):
   ...
```

The above code probably has bugs, I didn't even test it.  But the point is that `request.wait_for_disconnection()` will _always_ be cancelled at the end of the request anyway.

I can try to make it return at request end, even if there is no disconnection due to keepalive, but honestly I don't think it matters.

asvetlov:
I have a feeling that time of `request.wait_for_disconnection()` waiting should be less than the lifetime of request object itself.
Cancelling all pending waiters after web-handler finishing makes API cleaner and safer to use, isn't it?

gjcarneiro:
OK, _fine_...  but then the documentation, and maybe method name, might need to be updated to reflect that the function also returns at the end of request, not just client disconnection.

asvetlov:
I was talking about cancellation, not a regular `None` return.
Documentation should reflect it, sure.
Regarding the method name: do you have a proposal?

gjcarneiro:
> I was talking about cancellation, not a regular None return.

I didn't understand this, but makes sense.

> Regarding the method name: do you have a proposal?

Because it cancels it, there is no need for renaming, I think.

asvetlov:
looks good. The last remaining part is documentation update

aiohttp/web_request.py
```diff
@@ -690,6 +695,27 @@ def __bool__(self) -> bool:
...
+        for fut in self._disconnection_waiters:
+            fut.cancel()
+
+    async def wait_for_disconnection(self) -> None:
+        """Returns when the connection that sent this request closes
+
+        This can be used in handlers as a means to receive a notification of
+        premature client disconnection.
+
+        .. versionadded:: 4.0
```
asvetlov: It should be in `./docs/web_reference.rst`, not in python sources.

gjcarneiro:
Note: feel free to rebase/squash commits, or else I can do it.

aiohttp/web_request.py
```diff
@@ -690,6 +695,20 @@ def __bool__(self) -> bool:
...
         self._payload.set_exception(exc)
+        for fut in self._disconnection_waiters:
+            set_result(fut, None)
+
+    def _finish(self) -> None:
+        for fut in self._disconnection_waiters:
+            fut.cancel()
+
+    async def wait_for_disconnection(self) -> None:
+        fut = asyncio.Future()  # type: asyncio.Future[None]
```
asvetlov:
```suggestion
        loop = asyncio.get_event_loop()
        fut = loop.create_future()  # type: asyncio.Future[None]
```

asvetlov (changes requested):
Please apply the suggestion.

No rebase is required; we use squash-on-merging strategy for aiohttp RPs

asvetlov (approved):

asvetlov:
Thanks!

PR #4213, opened 2019-10-17
Backport fix of noop in least invasive way

hynek:
The coroutine-style definition of noop causes a warning on Python 3.8.

On master, it has been fixed and noop2 removed. This is the least invasive way to get rid of the warning in aiohttp 3.6.x that shouldn't break anything.

I don't think this needs a changelog entry?


asvetlov:
thanks!

Issue #2502, opened 2017-11-10
Redirect not been followed


delijati:
## Long story short

Redirect not been followed. I always get a `302` back. 

## Expected behaviour

Get a `200` back with text filled.

## Steps to reproduce

```python
import aiohttp
import asyncio


async def fetch(session, url):
    async with session.get(url, allow_redirects=True) as response:
        return await response.text(), response.status


async def fetch_all(session, urls, loop):
    results = await asyncio.gather(
        *[fetch(session, url) for url in urls],
    )

    for ret in results:
        print("status: %s len: %s" % (ret[1], len(ret[0])))
    return results

if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    urls = ['https://www.gulp.de/gulp2/projekte/suche?scope=projects&query=python']
    with aiohttp.ClientSession(loop=loop) as session:
        loop.run_until_complete(fetch_all(session, urls, loop))
```

kxepal:
Try `curl -v -L 'https://www.gulp.de/gulp2/projekte/suche?scope=projects&query=python'` - it produce redirection loop. It looks like not a aiohttp, but server bug.

delijati:
with requests
```python
In [1]: import requests
In [2]: res = requests.get("https://www.gulp.de/gulp2/projekte/suche?scope=projects&query=python")
In [3]: res.status_code
Out[3]: 200
In [4]: len(res.text)
Out[4]: 171410
```

kxepal:
Not sure what requests does, but even browsers reports about improper redirection caused by this URL. I'm sure we can investigate what the request does to handle it - it's very interesting how -, but I guess it's not our problem if reference HTTP client and browsers tells that there is a redirection issue.

delijati:
Narrowed it down to setting cookies. `http.cookies.__parse_string` aka `http.cookies.SimpleCookie.load` can't parse this cookie [1] str. `requests` is using `http.client.parse_headers`.

Possible solutions:

- fix `http.cookies.__parse_string` tested to be broken in `3.5`, `3.6`
- replace `SimpleCookie` with a patched version
- None of above as the cookie is broken and `http.client.parse_headers` 

[1] JSESSION_ID_PROFILE_APP=A3C47AF767CCC88FA36B8292A6FB94A3; Path=/; HttpOnly, gp_intern=(null);path=/;Secure, gp_intern=(null);path=/;Secure

kxepal:
[...]

PR #4224, opened 2019-10-18
allow samesite in cookies

samuelcolvin:
## What do these changes do?

Allow `samesite` to be allowed in cookeis, see also:
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
* https://www.chromestatus.com/feature/5088147346030592

no, just new `samesite=None` keyword argument to `set_cookie()`

## Related issue number

* #2502 partial
* https://github.com/aio-libs/aiohttp-session/issues/376

## Checklist


asvetlov:
Hmm.
I'm biased; should we monkey-patch Python<3.8 or restrict the feature to Python 3.8+ only?

Guys, what are your opinions?

samuelcolvin:
I'm happy with either, just let me know and I'll update the PR.

samuelcolvin:
But given that chrome showing a warning for any cookie not including `SameSite`, I imagine many people will want this fairly soon - restricting it to python >= 3.8 will cause headaches for lots of people.

aiohttp/web_response.py
```diff
@@ -42,6 +42,12 @@
     BaseClass = collections.abc.MutableMapping
 
 
+if 'samesite' not in Morsel._reserved:  # type: ignore
```
asvetlov: Use `helpers.PY_38` check here instead if `Morsel._reserved` lookup.
It makes the intention explicitly clear and allows dropping the patch after dropping Python 3.7 support.

samuelcolvin:
Done, with `versionadded`, docs look like:
image

is that okay?

docs/web_reference.rst
```diff
@@ -702,6 +703,14 @@ StreamResponse
                           specification the cookie
                           conforms. (Optional, *version=1* by default)
 
+      .. versionadded:: 3.7
```
asvetlov: Please move the directive inside `:param:` block, see `web.Response` `zlib_executor_size` for inspiration.

asvetlov:
@samuelcolvin ping

samuelcolvin:
done, sorry for delay.

samuelcolvin:
:ok_hand:

AndreiPashkin:
@asvetlov, I wonder - is this going to be released anytime soon?

bitnom:
Very important security consideration. Should consider a new release just to get this feature out the door imo. I'm tempted to install from master. Any commits in master I should be afraid of at this point?

vishim:
@asvetlov  - Is there anything blocking the 3.7 release? Chrome has already started rolling out the samesite update to a selected population.

asvetlov:
This pull request has been mentioned on **aio-libs**. There might be relevant details there:

https://aio-libs.discourse.group/t/patch-samesite-cookies-to-minor-version/111/1

PR #4240, opened 2019-10-19
Added properties of default ClientSession params to ClientSession class

### LINKED ISSUE #3882: Expose ClientSession defaults such as headers and auth for introspection
## Long story short

Too many attributes of ClientSession are private without an explicit need to. The following constructor arguments can never be accessed after creating a session:

- `headers`
- `skip_auto_headers`
- `auth`
- `json_serialize`
- ~`request_class`~
- ~`response_class`~
- ~`ws_response_class`~
- `connector_owner`
- `raise_for_status`
- `timeout`
- `auto_decompress`
- `trust_env`
- `trace_configs`

## Expected behaviour

I want to be able to create a new session based on an existing session, only tweaking some aspects. Without access to these attributes, you can't properly create a new session instance with only some aspects replaced.

## Actual behaviour

You have to access API private attribute to get at this information, making it much harder than necessary to create an updated wrapper as session needs change.

For example, I'm currently implementing an OAuth 2.0 authentication handler, which automatically fetches an authentication token if required for the current request, transparently. We create a client session in a different part of the app, shared between a group of objects, some of which need to dynamically switch to support OAuth 2.0. 

Rather than have to create a new session for each of object regardless of OAuth needs (preventing connection re-use between some of these objects where the same server is involved) or having to introspect every request to see if OAuth applies to this specific request, I want to create a new session only when OAuth is actually needed, reusing the same connection pool, if and when OAuth is needed. But this requires knowing the headers, timeouts and trace config on the session used so far.

paulefoe:
## What do these changes do?

Added properties of default ClientSession params to ClientSession class so it is available for introspection now

Users can now see which default parameters were initialized with ClientSession

## Related issue number

This commit fixes  #3882

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -920,7 +920,7 @@ class ClientSession:
         return self._timeout
 
     @property
-    def default_headers(self) -> 'CIMultiDict[str]':
+    def headers(self) -> 'CIMultiDict[str]':
         """The default headers of the client session."""
         return self._default_headers
 
@@ -956,7 +956,7 @@ class ClientSession:
 
     @property
     def auto_decompress(self) -> bool:
-        """Should automatically decompress response body"""
+        """Should the body response be automatically decompressed"""
         return self._auto_decompress
 
     @property
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -918,6 +918,58 @@ class ClientSession:
         """Timeout for the session."""
         return self._timeout
 
+    @property
+    def headers(self) -> CIMultiDict[str]:
+        """The headers of the client session."""
+        return self._default_headers
+
+    @property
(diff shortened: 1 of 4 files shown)


aiohttp/client.py
```diff
@@ -918,6 +919,60 @@ def timeout(self) -> Union[object, ClientTimeout]:
         """Timeout for the session."""
         return self._timeout
 
+    @property
+    def default_headers(self) -> 'CIMultiDict[str]':
```
asvetlov:
```suggestion
    def headers(self) -> 'CIMultiDict[str]':
```
Let's keep property names in sync with constructor parameters

asvetlov:
Thanks!

PR #4263, opened 2019-10-23
Prepare response headers prior to running prepare signal

### LINKED ISSUE #1958: Possibility to disable Server header
## Long story short

By default, aiohttp server adds `Server` header. It should be possible to disable this, for security reasons.

## Expected behaviour

No `Server` header in response

## Actual behaviour

Server sends these headers:

    Content-Length: 2
    Date: Tue, 06 Jun 2017 13:52:03 GMT
    Server: Python/3.6 aiohttp/2.1.0

## Steps to reproduce

    import asyncio
    from aiohttp import web

    async def http_handler(request):
        return web.Response(text="OK", headers={})

    def main():
        loop.run_until_complete(loop.create_server(web.Server(http_handler), "127.0.0.1", 8080))
        loop.run_forever()

    if __name__ == "__main__":
        main()

## Your environment

macOS 10.12.5
Python 3.6.1

    $ pip3 freeze
    aiofiles==0.3.0
    aiohttp==2.1.0
    async-timeout==1.2.1
    asyncio==3.4.3
    chardet==3.0.3
    Cython==0.25.2
    defusedxml==0.5.0
    google-api-python-client==1.6.2
    httplib2==0.10.3
    jira==1.0.10
    llvmlite==0.16.0
    magicmemoryview==0.1.5
    multidict==2.1.6
    numba==0.31.0
    numexpr==2.6.2
    numpy==1.12.0
    oauth2client==4.1.0
    oauthlib==2.0.2
    pbr==3.0.1
    pyasn1==0.2.3
    pyasn1-modules==0.0.8
    requests==2.13.0
    requests-oauthlib==0.8.0
    requests-toolbelt==0.7.1
    rsa==3.4.2
    six==1.10.0
    tornado==4.4.2
    uritemplate==3.0.0
    websockets==3.3
    yarl==0.10.2

Qix-:
## What do these changes do?

When a response is to be sent, this prepares the headers prior to running the prepare hook instead of after.

Any uses of the response prepare signal that assumed the headers are replaced after their signal handlers will now need to take this change into account.

## Related issue number

Closes #1958.

## Checklist

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -337,19 +337,27 @@ class StreamResponse(BaseClass, HeadersMixin):
         if self._payload_writer is not None:
             return self._payload_writer
 
-        await request._prepare_hook(self)
         return await self._start(request)
 
     async def _start(self, request: 'BaseRequest') -> AbstractStreamWriter:
         self._req = request
+        writer = self._payload_writer = request._payload_writer
 
+        await self._prepare_headers(request, writer)
+        await request._prepare_hook(self)
+        await self._write_headers(request, writer)
+
+        return writer
+
+    async def _prepare_headers(
+        self, request: 'BaseRequest', writer: AbstractStreamWriter
+    ) -> None:
         keep_alive = self._keep_alive
         if keep_alive is None:
             keep_alive = request.keep_alive
         self._keep_alive = keep_alive
 
         version = request.version
-        writer = self._payload_writer = request._payload_writer
 
         headers = self._headers
         for cookie in self._cookies.values():
@@ -392,12 +400,14 @@ class StreamResponse(BaseClass, HeadersMixin):
                 if version == HttpVersion11:
                     headers[hdrs.CONNECTION] = 'close'
 
+    async def _write_headers(
+        self, request: 'BaseRequest', writer: AbstractStreamWriter
+    ) -> None:
         # status line
+        version = request.version
         status_line = 'HTTP/{}.{} {} {}'.format(
             version[0], version[1], self._status, self._reason)
-        await writer.write_headers(status_line, headers)
-
-        return writer
+        await writer.write_headers(status_line, self._headers)
 
     async def write(self, data: bytes) -> None:
         assert isinstance(data, (bytes, bytearray, memoryview)), \
--- a/tests/test_web_response.py
+++ b/tests/test_web_response.py
@@ -1090,6 +1090,31 @@ def test_response_with_immutable_headers() -> None:
                             'Content-Type': 'text/plain; charset=utf-8'}
 
 
+async def test_response_prepared_after_header_preparation() -> None:
+    req = make_request('GET', '/')
+    resp = StreamResponse()
+    await resp.prepare(req)
+
+    assert type(resp.headers['Server']) is str
+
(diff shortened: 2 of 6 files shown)


Qix-:
Compliments on the testing and CI setup for this repository by the way, quite exemplary work done here. It was a pleasure working with this codebase given that I've never touched it before. Quite rare nowadays.

aiohttp/web_response.py
```diff
@@ -337,19 +337,27 @@ def _generate_content_type_header(
...
+        writer = self._payload_writer = request._payload_writer
+
+        await self._prepare_headers(request, writer)
+        await request._prepare_hook(self)
+        await self._write_headers(request, writer)
+
+        return writer
 
+    async def _prepare_headers(
+        self, request: 'BaseRequest', writer: AbstractStreamWriter
```
webknjaz, line 353: Should it accept a `writer` argument given that the same thing is effectively available via `self._payload_writer` anyway?

aiohttp/web_response.py
```diff
@@ -392,12 +400,14 @@ def _generate_content_type_header(
                 if version == HttpVersion11:
                     headers[hdrs.CONNECTION] = 'close'
 
+    async def _write_headers(
+        self, request: 'BaseRequest', writer: AbstractStreamWriter
```
webknjaz, line 404: (same concern as above)

tests/test_web_response.py
```diff
@@ -1090,6 +1090,31 @@ def test_response_with_immutable_headers() -> None:
                             'Content-Type': 'text/plain; charset=utf-8'}
 
 
+async def test_response_prepared_after_header_preparation() -> None:
+    req = make_request('GET', '/')
+    resp = StreamResponse()
+    await resp.prepare(req)
+
+    assert type(resp.headers['Server']) is str
```
webknjaz, line 1098: I think we should also take inheritance into account:
```suggestion
    assert isinstance(resp.headers['Server'], str)
```
WDYT?

webknjaz:
@Qix- thanks for the PR! I've left a few comments and defer the rest of the review to @asvetlov.

aiohttp/web_response.py
Qix-: I can sure change it.

asvetlov: Makes sense. `request` is available as `self._req` also.

tests/test_web_response.py
asvetlov: No, the strict type check is better here.
Please keep the test as is

asvetlov:
@Qix- ping

kwuite:
Will this get merged anytime soon?

Qix-:
Hi - I don't use this library anymore as I left the team that did. Someone else is free to pick it up  if they'd like to.

aiohttp/web_response.py
ojii: The issue is `self._payload_writer` and `self._req` are `Optional[...]` so `_prepare_headers` and `_write_headers` would need to guard against them being `None` to pacify mypy. Should they raise errors if that happens? Passing the non-Optional values might be preferable.

asvetlov:
Done by #5064

PR #4285, opened 2019-10-28
A Couple of Documentation Improvements

t-mart:
## What do these changes do?

Improve some areas of the documentation that caught me up during development.

Namely:
- In `faq.rst`, the websocket example did not `await` on the websocket's `send_str`. Therefore, this coroutine will not function as intended. Also, the example included a psuedocode statement that did not compile, which is inconsistent with the other examples in the docs. I fixed up the psuedocode and added a comment.
- In `web_advanced.rst`, the "simple startup code for serving HTTP site" did not include an long (infinite?) sleep. Without that, your application will stop immediately without any indication as to why. In contrast, `web.run_app` handles this for you, so switching to the "AppRunner" style could be confusing. This documentation code is also referenced from the `whats_new_3_0.rst`, so it is probably a common landing point for many readers. I added a `while True: await asyncio.sleep(3600)` block.

No. Just documentation changes.

## Checklist


t-mart:
I'm unclear whether "code modification" in _documentation_ requires a CONTRIBUTORS.txt update. Let me know. Omitting for now.

asvetlov:
Thanks!

CONTRIBUTORS update is up to you.
We like to enumerate all the people who helped the library but don't enforce it.

Issue #3252, opened 2018-09-07
False-positive deprecation warning for partial async handlers


nicktimko:
## Long story short

Wrapping a coroutine function with a `functools.partial` causes a DeprecationWarning to trigger.

## Steps to reproduce
Something like
```
def private_route(handler):
    return functools.partial(handler, private=True)

async def hello(request, private: bool):
    return web.Response(text='boo' if private else 'hello')

handler = private_route(hello)
app.router.add_route('get', '/hello', handler)  # DepWarning
```

causes [a `DeprecationWarning`](https://github.com/aio-libs/aiohttp/blob/558519adae70795ab5c99e150e0ac10e80f68d36/aiohttp/web_urldispatcher.py#L113-L114) to be hit.

I think when called as `handler(req)` will immediately run through the partial and return the coroutine, which can then be awaited. It seems like the dep warning in this case is more of a false-positive. Maybe an expression like the ending one below could suppress it? If this is indeed just a false-positive and I'm the only ~complainer~ reporter, I could live.

```
import functools
import inspect

async def f():
    pass

fp = functools.partial(f)

inspect.iscoroutinefunction(f)  # True
inspect.iscoroutinefunction(fp) # False
isinstance(fp, functools.partial) and inspect.iscoroutinefunction(fp.func) # True
```

Tweak?:

```diff
         if asyncio.iscoroutinefunction(handler):
             pass
         elif inspect.isgeneratorfunction(handler):
             warnings.warn("Bare generators are deprecated, "
                           "use @coroutine wrapper", DeprecationWarning)
+        elif (isinstance(handler, functools.partial) and
+              asyncio.iscoroutinefunction(handler.func)):
+            pass
         elif (isinstance(handler, type) and
               issubclass(handler, AbstractView)):
             pass
         else:
             warnings.warn("Bare functions are deprecated, "
                           "use async ones", DeprecationWarning)
```

## Your environment

I'm using aiohttp 3.3.2 on Python 3.6.5 on Ubuntu 16.04, but it looks like master would still trigger the warning.

webknjaz:
Feel free to send a PR

asvetlov:
The issue is essentially the same as https://github.com/aio-libs/aiohttp/issues/1993#issuecomment-419431628
I doubt if introspection errors should be fixed on aiohttp level.

nicktimko:
[...]

PR #4302, opened 2019-10-29
Support addition of routes when handler is wrapped by functools.partial()

yuvalof:
## What do these changes do?
In aiohttp v4 a new check was added to make sure that route handlers are coroutines.
Python < v3.8 will return False for asyncio.iscoroutine(functools.partial(async_function)).
The change will iterate the wrapped function to make sure its async

No

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1,5 +1,4 @@
 import abc
-import asyncio
 import base64
 import hashlib
 import keyword
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -6,7 +6,6 @@ import keyword
 import os
 import re
 from contextlib import contextmanager
-from functools import partial
 from pathlib import Path
 from types import MappingProxyType
 from typing import (  # noqa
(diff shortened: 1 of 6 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -148,6 +149,12 @@ def __init__(self, method: str,
         self._expect_handler = expect_handler
         self._resource = resource
 
+    @staticmethod
+    def iscoroutinefunction(handler: Any) -> bool:
```
asvetlov: The method should not be a part of public API.
Please make it first-class function which name is started from underscore.

yuvalof: I exprted this function to helpers.py and changed it to only happen on Python<3.8

asvetlov:
pls fix mypy: https://dev.azure.com/aio-libs/aiohttp/_build/results?buildId=1311&view=logs&j=cc4e4887-e527-5c83-25ce-76663593f2e1&t=9d25c5b1-1b8c-5097-b25c-a7f376352755&l=215

asvetlov:
thanks!

nicktimko:
Also fixes #3252

asvetlov:
Thanks @nicktimko

Issue #4277, opened 2019-10-26
Implement HTTPException set_cookie() and del_cookie() methods


asvetlov:
See https://github.com/aio-libs/aiohttp-session/issues/396 for the context.

Implementation thoughts:
1. `.cookies` property is also needed
2. `set_cookie()` and `del_cookie()` should update headers immediately, `HTTPException` has no `.prepare()` method and should not have it.
3. Backport is not needed, in 3.x HTTPException is derived from Response and have all mentioned methods already.

derlih:
@asvetlov 
> set_cookie() and del_cookie() should update headers immediately,

Is this requirement strict? Even if `HTTPException` will have no `prepare()`?

asvetlov:
Do you mean that `headers['Cookie']` is not updated before `StreamResponse.prepare()` call?

You are right, perhaps we should keep the consistency here.

derlih:
I mean in the code:
https://github.com/aio-libs/aiohttp/blob/488b375a86e02bc55587550e2360e154acd55131/aiohttp/web_protocol.py#L447-L463

The data is copied from `HTTPException` to `Response`. Cookies can be copied as well. And `prepare()` will be called on that response instance.

asvetlov:
Ok.

derlih:
@asvetlov can this issue be closed?

PR #4308, opened 2019-10-30
Cookies for HTTPException.

prostomarkeloff:
## What do these changes do?

Add possbile to add cookies for HTTPException.

## Related issue number

#4277

## Checklist


examples/fake_server.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, *, loop):
 async def main(loop):
     token = "ER34gsSGGS34XCBKd7u"
 
-    fake_facebook = FakeFacebook()
+    fake_facebook = FakeFacebook(loop=loop)
```
asvetlov, line 98: Please sync with a fresh master; the loop parameter was removed recently.

tests/test_web_exceptions.py
```diff
@@ -9,9 +10,9 @@
 
 
 def test_all_http_exceptions_exported() -> None:
-    assert 'HTTPException' in web.__all__
+    assert "HTTPException" in web.__all__
```
asvetlov, line 13: I'm curious why did you replace all quote symbols?
The change is not related to the PR purpose.

aiohttp/web_exceptions.py
```diff
@@ -122,11 +123,80 @@ def __init__(self, *,
...
         self._text = text
         self._headers = real_headers
+        self._cookies: 'SimpleCookie[str]' = SimpleCookie()
         self.args = ()
 
     def __bool__(self) -> bool:
         return True
 
+    @property
+    def cookies(self) -> 'SimpleCookie[str]':
```
asvetlov, line 133: Please not copy this block from `web_response.py` but reuse.
Maybe extraction the common code into some mixin class in `helpers.py` can handle it pretty well.

tests/test_web_exceptions.py
prostomarkeloff: Oops, sorry me. It's black.

aiohttp/web_exceptions.py
prostomarkeloff: Sounds right for me, code duplication it's bad way.

prostomarkeloff: OK, as we can place this code in `helpers.py`? I think new class `CookiesMixin` will be good solution.

asvetlov: sounds good

tests/test_web_exceptions.py
asvetlov: Let's do black reformatting in another PR.
Since 19.10 it can be applied to Python 3.5 compatible code.
The trick is that blackifying should be done on all supported branches at the same time; otherwise backporting becomes virtually impossible.

PR #4329, opened 2019-11-08
Simplify the hello world example, explain request lifecycle

### LINKED ISSUE #4272: Improve client "hello world" in README
When looking at aiohttp README, the hello word looks like that:

```python
import aiohttp
import asyncio

        return await response.text()

async def main():
        html = await fetch(session, 'http://python.org')
        print(html)

if __name__ == '__main__':
```

For requests, it looks like this:

```python
>>> r = requests.get('https://api.github.com/user', auth=('user', 'pass'))
>>> r.status_code
200
>>> r.headers['content-type']
'application/json; charset=utf8'
>>> r.encoding
'utf-8'
>>> r.text
u'{"type":"User"...'
>>> r.json
{u'private_gists': 419, u'total_private_repos': 77, ...}
```

I can see three issues with the way aiohttp present itself:

- The snippet seems to have unnecessary complexity
- The snippet doesn't feature any quick scripting setup
- There is no explanation for the *necessary* complexity

## Expected behaviour

First, extra complexity should be removed. I understand the benefit of `if __name__ == "__main__"`, but in a hello word type of code, it's noise that makes the code seems bigger than necessary. If you have the skill to learn async, you know when to add it later.

Also, can the code go from:

```python
        return await response.text()

async def main():
        html = await fetch(session, 'http://python.org')
        print(html)
```
To:

```python
async def main():
        async with session.get('http://python.org') as response:
            return await response.text()
```

?

I'm guessing the first form exhibit some kind of best practice I'm missing but:

- since I'm missing it, I have no way to understand why it's best, and won't use it in the future
- it makes the code feels heavy, and hide the expressiveness of aiohttop behind the verbosity of the coroutine system
- the first time you grab for aiohttp, you probably just want to make a quick script out of it, not build a system with best practices

And lastly, even this is weird:

```python
async def main():
    async with aiohttp.ClientSession() as session: # why session ?
        async with session.get('http://python.org') as response: # why can't I do that in one line ?
            return await response.text()
```

This is necessary complexity. It's done for a reason, but coming from requests, it's strange and seems uncalled for. The README should acknowledge that, then offer a link leading to and explanation of why this is necessary for an async lib, and the consequences of it. It there are shortcuts to that syntax for quick scripts with less performance/best practice needs, the README should mention THAT first, give a quick warning, then follow with the best practice.

aiohttp is the `requests` of async as for now. It's the first face a newcommer see when discovering the async world nowaday (well, after playing with asyncio.sleep in tutorials).

ksamuel:
## What do these changes do?

Simplify the hello world example in the README and doc index

Add a new page to the doc (http_request_lifecycle) to explain why aiohttp's client API requires several steps for performing a request. Compare it with the "requests" library. Gives some information about the use of the ClientSession object.

No

## Related issue number

Close #4272 

## Checklist


README.rst
```diff
@@ -58,19 +57,29 @@ To get something from the web:
...
+.. code-block::
 
-  if __name__ == '__main__':
-      loop = asyncio.get_event_loop()
-      loop.run_until_complete(main())
+    Status: 200
+    Content-type: text/html; charset=utf-8
+    Body: <!doctype html> ...
 
+Comming from `requests<https://requests.readthedocs.io/>`_ ? Read :ref:`why we need so many lines <http://aiohttp.readthedocs.io/en/latest/aiohttp-request-lifecycle>`.
```
webknjaz: This makes the CI fail.

CONTRIBUTORS.txt
```diff
@@ -125,6 +125,7 @@ Igor Pavlov
 Ilya Chichak
 Ingmar Steen
 Ivan Larin
+Kevin Samuel
```
webknjaz: "K" goes after "J" in the alphabet.

ksamuel:
I see that the coverage went down, but I didn't touched the code or unit tests. 

Maybe I'm doing the PR on the wrong branch?

asvetlov:
Coverage jittering for aiohttp worries me too.
Maybe this is because we report the coverage for both *master* and *3.7* branches?
I don't know.

asvetlov:
Thanks!

PR #4350, opened 2019-11-16
Add name and backref to add_subapp

heckad:
## What do these changes do?

This pull request adds to add_subapp method two new optional arguments wich make code more cliar.

Example:
  before
```
    subapp = create_api_app()
    _app.add_subapp('/api', subapp)
    _app['api'] = subapp
    subapp['parent_app'] = _app
```
  after
```
  _app.add_subapp('/api', create_api_app(), name='api', backref='parent_app')
```

If this good change, I will do next steps.

## Checklist

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -215,18 +215,24 @@ class Application(MutableMapping[str, Any]):
         reg_handler('on_cleanup')
 
     def add_subapp(self, prefix: str,
-                   subapp: 'Application') -> AbstractResource:
+                   subapp: 'Application',
+                   name: Optional[str] = None,
+                   backref: Optional[str] = None,
+                   ) -> AbstractResource:
         if not isinstance(prefix, str):
             raise TypeError("Prefix must be str")
         prefix = prefix.rstrip('/')
         if not prefix:
             raise ValueError("Prefix cannot be empty")
         factory = partial(PrefixedSubAppResource, prefix, subapp)
-        return self._add_subapp(factory, subapp)
+        return self._add_subapp(factory, subapp, name, backref)
 
     def _add_subapp(self,
                     resource_factory: Callable[[], AbstractResource],
-                    subapp: 'Application') -> AbstractResource:
+                    subapp: 'Application',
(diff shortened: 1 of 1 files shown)


heckad:
Maybe need to add `subapp` storage for solving problems with name intersection, but it can make pipe more complex. 
For example:

before:
 `app['api']['v1.0'].router['status']`
after:
 `app.subapps['api'].subapps['v1.0'].router['status']`

asvetlov:
Can it be implemented by a third-party helper without modifying aiohttp core code?
Sure!
Is the feature useful for the vast majority of code that is built with aiohttp.web?
I very doubt. I use subapps extensively at my daily job, we never needed such functionality.
This is the first request I ever recall for aiohttp 6 years old history.

heckad:
> Can it be implemented by a third-party helper without modifying aiohttp core code?
> Sure!

We noticed that you are going to make an application class is final and add `__init_subclass__` warning. Now we inherit from the application class and override some methods. In future, we can't do this and should use composition or use procedure programming (make a function for this logic). Therefore we offer this function in the base code.

> Is the feature useful for the vast majority of code that is built with aiohttp.web?
> I very doubt. I use subapps extensively at my daily job, we never needed such functionality.
> This is the first request I ever recall for aiohttp 6 years old history.

Do you never store app in app storage? You add a special comment for this case in [fac](https://docs.aiohttp.org/en/stable/web_advanced.html#nested-applications)

asvetlov:
Maybe because my sub-applications are isolated good enough, the reverse URL lookup is almost never needed. Well, it may depend on the sort of developed application.
Maybe in your case you need the lookup constantly, and this is fine.
 
Anyway, you can write a custom top-level function that does everything that you want.

heckad:
In this case useing nested apps not the same as using bluprints. Does aiohttp have bluprints concept in future?

asvetlov:
What is the blueprint concept in your mind?
How backref converts sub-application into a blueprint?
Why do you think that blueprints are necessary?

I have a feeling that you are asking about *something* without the feature definition. Sorry, brainreading is the hard skill for me.

Issue #4345, opened 2019-11-14
Repeated client request with multipart FormData results in field duplication


ulrikjohansson:
## Long story short
We've run into a problem where a repeated client session request with the same arguments end up not sending the same request data, but duplicted fields.
Reproducible request are important to us because we're using a circuit breaker library with our client requests that do retries on certain errors (like ServerDisconnectedError and TimeoutError)

## Expected behaviour
When I repeat a call to `ClientSession.request` with the exact same arguments I expect to send a copy of the first request

## Actual behaviour

When using a FormData object with a field that gets the `_is_multipart` property set to True, the form fields are duplicated. The machinery inside `FormData._gen_form_data` just appends the form fields again on each request using it.
`
## Steps to reproduce

```
import asyncio
from aiohttp import ClientSession, FormData
from copy import deepcopy

from aiohttp import __version__ as version
print(f"aiohttp version: {version}")

async def run():
    async with ClientSession() as session:
        url = "http://httpbin.org/anything"
        data = FormData()
        data.add_field("test", "test_value", content_type="application/json")

        for n in range(2):
            await session.post(url, data=data)
            print(f"writer part count: {len(data._writer._parts)} <-- this should stay at 1")

loop = asyncio.get_event_loop()
loop.run_until_complete(run())
```
## Workaround
A workaround for this issue is passing a deepcopy of the `FormData` object to `ClientSession.request`

## Your environment
aiohttp version: tested with 3.5.4 & 3.6.2
aiohttp component: client
os: linux (ubuntu 19.10)

asvetlov:
Very interesting, thanks.
`FormData` was never designed to be reused but implicit duplication is not appropriate.

Need to dig more what exact lines in aiohttp sources produce the error.
I see two possible solutions: 
1. Don't mutate FormData on sending, which would be tricky especially for sending files.
2. Mark FormData instance as *processed*, raise an exception in the second usage. Simple and robust.

A champion is wanted for the issue, I expected that the fix for 2) is not hard at all.

AtomsForPeace:
This should be fixed now in master

asvetlov:
Fixed by #4351

PR #4351, opened 2019-11-17
Form data processed

AtomsForPeace:
## What do these changes do?

A FormData instance keeps track whether it has been processed already. If the instance is sent for a second time then it raises a ClientPayloadError.

Will raise an exception if they try to resend a FormData instance.

## Related issue number

#4345

## Checklist

### DIFF
--- a/aiohttp/formdata.py
+++ b/aiohttp/formdata.py
@@ -122,7 +122,7 @@ class FormData:
     def _gen_form_data(self) -> multipart.MultipartWriter:
         """Encode a list of fields using the multipart/form-data MIME format"""
         if self._is_processed:
-            raise ClientPayloadError('Form data has been processed already')
+            raise RuntimeError('Form data has been processed already')
         for dispparams, headers, value in self._fields:
             try:
                 if hdrs.CONTENT_TYPE in headers:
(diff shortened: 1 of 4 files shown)


AtomsForPeace:
I decided to raise a ClientPayloadError if the FormData instance has already been processed. Is this the best Exception or is there something more fitting?

I also think this code could break some people's code as even in some of the other tests code broke. 

Should I add documentation somewhere?

aiohttp/formdata.py
```diff
@@ -35,6 +37,10 @@ def __init__(self, fields:
     def is_multipart(self) -> bool:
         return self._is_multipart
 
+    @property
+    def is_processed(self) -> bool:
```
asvetlov: Please don't expose the property.
I doubt if it is useful for something else than raising an exception of the double usage.

aiohttp/formdata.py
```diff
@@ -115,6 +121,8 @@ def _gen_form_urlencoded(self) -> payload.BytesPayload:
 
     def _gen_form_data(self) -> multipart.MultipartWriter:
         """Encode a list of fields using the multipart/form-data MIME format"""
+        if self._is_processed:
+            raise ClientPayloadError('Form data has been processed already')
```
asvetlov: This is the programming error that don't depend on networking etc.
The error should be fixed by changing the source code, `raise RuntimeError` is the right exception in this context.

asvetlov:
Change-note is enough, I don't think we should document it additionally.

PR #4369, opened 2019-11-21
Avoid circular references between response handler and parser

simonschmidt:
ResponseHandler kept a reference via `_parser` to HttpResponseParser
which in turn kept a reference in `protocol` to ResponseHandler.

This breaks that reference on `close()` by setting `_parser` to
None.

---

I'm not entirely sure this is correct, I assumed that `ResponseHandler` methods `data_received` and `connection_lost` are never called after `close()`

Also I think there is a deeper issue in asyncio that I haven't been able to pin-point yet as the `ResponseHandler` is still kept around by references from `_SelectorSocketTransport` via `SSLProtocol`.

## Checklist

### DIFF
--- a/CHANGES/4369.bugfix
+++ b/CHANGES/4369.bugfix
@@ -1,0 +1,1 @@
+Avoid circular references between response handler and parser
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -64,6 +64,13 @@ class ResponseHandler(BaseProtocol,
             self._payload = None
             self._drop_timeout()
 
+        # The parser can have a reference back to this instance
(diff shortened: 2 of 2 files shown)


asvetlov:
The change looks correct.
I'll double-check it before the merge again.
Please add `./CHANGES` file, it is the project requirement.

simonschmidt:
Seems to be caused by `self._payload_parser = None`, I'll dig a little bit more to try to figure why that locks up.

Sometimes `connection_lost` is called prior to `close` and sometimes after, for example in the tests:

* test_client_ws_functional.py::test_heartbeat_no_pong
* test_web_websocket_functional.py::test_heartbeat_no_pong

Thus the assumptions of this PR are violated, will close

PR #4377, opened 2019-11-25
Link the Truth value of a ClientResponse to the underlying status

Colin-b:
## What do these changes do?

Provides the ability to use the Truth value of a response, the same we can when using [requests](https://github.com/psf/requests/blob/master/requests/models.py#L673).

In case users where relying on the Truth value instead of checking to None explicitly, this might change the behavior in case of an HTTP error.

But was it ever a proper usage ?

## Related issue number

None to my knowledge

## Checklist

### DIFF
--- a/CHANGES/4377.feature
+++ b/CHANGES/4377.feature
@@ -1,0 +1,1 @@
+Link the Truth value of a ClientResponse to the underlying status.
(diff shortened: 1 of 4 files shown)


asvetlov:
Thanks for the contribution but sorry, I cannot accept this request.

Very many existing third-party codes use `if resp: ...` as a synonym for `if resp is not None: ...`, the PR breaks this assumption.

Please check for the response status explicitly.

PR #4378, opened 2019-11-25
Add runner.serve_forever()

### LINKED ISSUE #2746: Add runner.serve_forever()
Like https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.Server.serve_forever

atemate:
## What do these changes do?
This PR adds a handy method `runner.serve_forever()` similar to [`asyncio.Server.serve_forever()`](https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.Server.serve_forever)

Yes! see changes in documentation

## Related issue number

Closes https://github.com/aio-libs/aiohttp/issues/2746

## Checklist

### DIFF
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -197,13 +197,17 @@ class SockSite(BaseSite):
 
 
 class BaseRunner(ABC):
-    __slots__ = ('_handle_signals', '_kwargs', '_server', '_sites')
+    __slots__ = (
+        '_handle_signals', '_kwargs', '_server', '_sites',
+        '_serving_forever_fut',
+    )
 
     def __init__(self, *, handle_signals: bool=False, **kwargs: Any) -> None:
         self._handle_signals = handle_signals
         self._kwargs = kwargs
         self._server = None  # type: Optional[Server]
         self._sites = []  # type: List[BaseSite]
+        self._serving_forever_fut = None  # type: Optional[asyncio.Future[None]]  # noqa: E501 line too long
 
     @property
     def server(self) -> Optional[Server]:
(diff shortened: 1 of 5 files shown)


aiohttp/web_runner.py
```diff
@@ -238,6 +242,24 @@ def sites(self) -> Set[BaseSite]:
...
+            )
+        if self._server is None:
+            raise RuntimeError('Call setup() first')
+
+        loop = asyncio.get_event_loop()
+        self._serving_forever_fut = loop.create_future()
+        try:
+            await self._serving_forever_fut
+        except asyncio.CancelledError:
+            await self.cleanup()
```
atemate, line 258: @asvetlov Am I right assuming that `runner.serve_forever()` should do `self.cleanup()` if being cancelled?

aiohttp/web_runner.py
```diff
@@ -238,6 +242,24 @@ def sites(self) -> Set[BaseSite]:
...
         self._server = await self._make_server()
 
+    async def serve_forever(self) -> None:
+        if self._serving_forever_fut is not None:
+            raise RuntimeError(
+                'Concurrent calls to serve_forever() are not allowed'
+            )
+        if self._server is None:
+            raise RuntimeError('Call setup() first')
+
```
atemate, line 252: Note: [`asyncio.Server.serve_forever()`](https://docs.python.org/3.7/library/asyncio-eventloop.html#asyncio.Server.serve_forever) actually *starts* accepting connection when `serve_forever()` was called. In contrast, current method does not start anything (though it cleans itself up if cancelled).

aiohttp/web_runner.py
asvetlov: The cleanup is questionable.
I think it deserves async context manager, e.g.
```
async with runner: 
    await runner.serve_forever()
``` 
Please remove.
Async context manager support requires a searate PR.

asvetlov (changes requested):
Almost good, please drop the cleanup call

aiohttp/web_runner.py
```diff
@@ -238,6 +242,24 @@ def sites(self) -> Set[BaseSite]:
 
         self._server = await self._make_server()
 
+    async def serve_forever(self) -> None:
+        if self._serving_forever_fut is not None:
+            raise RuntimeError(
+                'Concurrent calls to serve_forever() are not allowed'
+            )
+        if self._server is None:
+            raise RuntimeError('Call setup() first')
```
webknjaz, line 251: Any reason you don't just call `setup()` here? The name `serve_forever()` kinda implies that it'll start serving things by itself. Or perhaps it needs a better name.

asvetlov: `setup()` doesn't serve, need to register `TCPSite` first.
The initialization procedure requires explicit steps with specific parameters passing.
This is probably my design mistake, the forecast ability didn't work well here.
So, `serve_forever()` cannot start serving.
I prefer keeping this name because it looks good and asyncio already utilized it for very similar purpose.

PR #4393, opened 2019-11-27
Apply SO_REUSEADDR to test server's socket

asvetlov:
As discussed on gitter

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -5,6 +5,7 @@ import contextlib
 import functools
 import gc
 import inspect
+import os
 import socket
 import sys
 import unittest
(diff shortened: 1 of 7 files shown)


aiohttp/test_utils.py
```diff
@@ -57,12 +58,17 @@
...
+REUSE_ADDRESS = os.name == 'posix' and sys.platform != 'cygwin'
+
+
 def get_unused_port_socket(host: str) -> socket.socket:
     return get_port_socket(host, 0)
 
 
 def get_port_socket(host: str, port: int) -> socket.socket:
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    if REUSE_ADDRESS:
```
socketpair, line 70: It will be nice if you add a comment describing why it is set here.

webknjaz: +1

not everyone knows that it has different semantics under Windows.

webknjaz, line 70:
```suggestion
    if REUSE_ADDRESS:
        # Windows has different semantics for SO_REUSEADDR,
        # so don't set it. Ref:
        # https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse
```

socketpair: Well, I mean another thing ;). I don't understand why it ever needs. Even in Linux.

webknjaz: Because if you retry binding to the same socket fast enough, it'll fail. Even from the same process (after unbinding). There's some sort of timeout in place.

OTOH if you were to always use an ephemeral port 0, the kernel would allocate you a different port each time and there wouldn't be such a problem.

socketpair: Sorry, I confused. I misread as SO_REUSEPORT.

PR #4415, opened 2019-11-30
Fix a warning about unfinished task in web_protocol.py

### LINKED ISSUE #4408: Task was destroyed but it's pending
On master, I see outputs from `run_app()` like follows:
```
(Press CTRL+C to quit)
task: <Task pending name='Task-4' coro=<RequestHandler.start() done, defined at /home/andrew/.virtualenvs/multidict/lib/python3.8/site-packages/aiohttp/web_protocol.py:437> wait_for=<Future pending cb=[<TaskWakeupMethWrapper object at 0x7f8822ffef10>()]>>
```
I suspect it happens on errors and keepalive closing in `web_protocol.py`.
The message should go.

asvetlov:
Fixes #4408


webknjaz:
Is it possible to test this?

asvetlov:
Not sure.
Now I see that the fix is incorrect at least.

semiworks:
I think the problem here is, that the RequestHandler.start() function hangs when waiting for self._waiter and the start() function will never finish. In connection_lost() we should cancel self._waiter to let the RequestHandler.start() function return.

```
diff --git a/libs/aiohttp/aiohttp/web_protocol.py b/libs/aiohttp/aiohttp/web_protocol.py
index 7178781e..8404880e 100644
--- a/libs/aiohttp/aiohttp/web_protocol.py
+++ b/libs/aiohttp/aiohttp/web_protocol.py
@@ -277,6 +277,9 @@ class RequestHandler(BaseProtocol):
         if self._error_handler is not None:
             self._error_handler.cancel()

+        if self._waiter:
+            self._waiter.cancel()
+
         self._task_handler = None

```

asvetlov:
@semiworks you are right, I've updated the patch

Issue #4422, opened 2019-12-05
proxy_headers not passed to proxy


sathia-musso:
## Long story short

I'm setting headers for the proxy but they're not sent to it 

## Expected behaviour

a dict with headers should be sent to proxy 

## Actual behaviour

headers are not sent to proxy

## Steps to reproduce

    import asyncio
    import aiohttp

        
    async def main(url=None):
        async with aiohttp.ClientSession() as session:
            async with session.get(url,
                                   proxy_headers={'Foo': "bar"},
                                   proxy='http://127.0.0.1:8080',
                                   ) as resp:
                print('Foo', resp.headers['Foo'])
    
    
    asyncio.run(main(url='http://example.com'))

this is tcpdump from python: 

    ..z...z.GET http://example.com HTTP/1.1
    Host: example.com
    Accept: */*
    Accept-Encoding: gzip, deflate
    User-Agent: Python/3.7 aiohttp/3.6.2
    Proxy-Authorization: Basic blah=

this is tcpdump of curl

    curl -is --proxy-header "Foo: bar" --proxy-header "Foo: bar" --proxy 'http://127.0.0.1:8080' http://example.com/

    ........GET http://example.com/ HTTP/1.1
    Host: example.com
    Proxy-Authorization: Basic blah=
    User-Agent: curl/7.64.1
    Accept: */*
    Proxy-Connection: Keep-Alive
    Foo: bar

## Your environment

mac osx 
aiohttp  3.6.2
python 3.7

acherunilam:
Interestingly, this bug is exhibited only for HTTP URLs. The Proxy headers get passed just fine if the URL is HTTPS.

PR #4423, opened 2019-12-05
Fixed proxy_headers not being added on client_reqrep.py

sathio:
## What do these changes do?
proxy_headers are actually sent to proxy

No

## Related issue number
https://github.com/aio-libs/aiohttp/issues/4422

## Checklist

- [ x] I think the code is well written

### DIFF
--- a/CHANGES/4422.feature
+++ b/CHANGES/4422.feature
@@ -1,0 +1,1 @@
+proxy_headers not passed to proxy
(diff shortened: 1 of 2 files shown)


asvetlov:
Thanks for your contribution.
Sorry, the patch is incorrect: `proxy_headers` are processed by `connector.py` without merging with regular request headers.

sathia-musso:
Hi, yes, I saw, but from what I see they're not delivered to proxy. I'm trying to understand why

Issue #4431, opened 2019-12-09
aiohttp not honoring `no_proxy`


scirelli:
## Long story short
aiohttp not honoring `no_proxy` environment variable when `trust_env` is set.

## Expected behaviour
When there is a `no_proxy` `NO_PROXY` environment variable set, URLs listed in that variable should not use `http{s}_proxy`

## Actual behaviour
When `trust_env` is set and there are `http_proxy` and `no_proxy` set, the proxy is always used.

## Steps to reproduce
Set `http{s}_proxy` and `no_proxy`, make a request to a domain listed in `no_proxy`. The request will still go through the proxy.

## Your environment
Linux 
aiohttp 3.6.2
Company proxy server

asvetlov:
The pull request is welcome!

scirelli:
Could you point me in the right direction? 

I'm guessing this is where I'll need to check the no proxy list https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L424 and blank the proxy if found?

asvetlov:
Hmm, interesting.
The real code for extracting the proxy list from env vars is [`proxies_from_env()`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L230-L254) function.

It doesn't parse itself but reuses `getproxies()` from `urllib.request`. 
I did not investigate more.
Does `getproxies()` ignore `NO_PROXY`?

scirelli:
No. That is why I suggesting doing it manually. I could only find a reference to `no_proxy` when [urllib.request.ProxyHandler](https://docs.python.org/3/library/urllib.request.html#urllib.request.ProxyHandler) is used. I'm searching for something to compare the hostnames to the `no_proxy` list.

I could add a `is_in_no_proxy_list(url)` to the [helpers.py](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py) module.

scirelli:
Maybe a solution like 5d92cb5 might work. Would still have to write the tests. I should probably read up on the no_proxy list standards to make sure hostnames put in the no_proxy list are matching correctly.

scirelli:
Actually I looks like urllib has code to handle [no_proxy](https://github.com/python/cpython/blob/3.8/Lib/urllib/request.py#L2490). I'll use the appropriate function from there.

ghost:
Any ETA for this? We're suffering from the same problem with aiohttp not honoring no_proxy vars.

[...]

PR #4445, opened 2019-12-15
Add support for the `no_proxy` env var mechanism in the HTTP client

scirelli:
## What do these changes do?
Prior to this patch aiohttp would ignore no_proxy settings.

If `no_proxy` or `NO_PROXY` is set, and the request host matches one of the hosts from the no_proxy list, no proxy will be used.

Need help in finding where the unit tests go for this.
## Related issue number
 #4431

## Checklist


r-or:
Interested in this too!

scirelli:
Not sure why the linter is failing. Any pointers?

webknjaz:
The log is here https://github.com/aio-libs/aiohttp/pull/4445/checks?check_run_id=1827835221#step:6:60

Also, you can run the pre-commit.com tool locally to see the same.

aiohttp/helpers.py
```diff
@@ -40,7 +40,7 @@
     cast,
 )
 from urllib.parse import quote
-from urllib.request import getproxies
+from urllib.request import getproxies, proxy_bypass  # noqa: F401
```
Dreamsorcerer: Is there a reason to import this here, rather than just import it directly in client.py?

scirelli: It looked like this is where it belonged, it's where all the helpers are. And it looked like this file was meant to expose helper functions, so if in the future a new implementation of proxy_bypass was written only the code here would need to be changed. But I could be wrong, and I'll move it out that is not the case?

scirelli:
> The log is here [#4445 (checks)](https://github.com/aio-libs/aiohttp/pull/4445/checks?check_run_id=1827835221#step:6:60)
> 
> Also, you can run the pre-commit.com tool locally to see the same.

Figures, doesn't fail locally when I run pre-commit.

<details>
<summary>Github</summary>

```
mypy --show-error-codes aiohttp tests
aiohttp/client.py:78: error: Module 'aiohttp.helpers' has no attribute 'proxy_bypass'  [attr-defined]
Found 1 error in 1 file (checked 102 source files)
make: *** [mypy] Error 1
Makefile:77: recipe for target 'mypy' failed
Error: Process completed with exit code 2.
```

</details>

<details>
<summary>Local</summary>

```
pre-commit run --all-files
Check CHANGES................................................................Passed
Check for merge conflicts....................................................Passed
Strip unnecessary `# noqa`s..................................................Passed
isort........................................................................Passed
black........................................................................Passed
Fix End of Files.............................................................Passed
Fix requirements.txt.........................................................Passed
Trim Trailing Whitespace.....................................................Passed
File Contents Sorter.........................................................Passed
Check for case conflicts.....................................................Passed
Check JSON...................................................................Passed
Check Xml....................................................................Passed
Check that executables have shebangs.........................................Passed
Check Toml...................................................................Passed
Check Xml....................................................................Passed
Check Yaml...................................................................Passed
Debug Statements (Python)....................................................Passed
Check for added large files..................................................Passed
Check for broken symlinks................................(no files to check)Skipped
Debug Statements (Python)....................................................Passed
fix UTF-8 byte order marker..................................................Passed
Fix python encoding pragma...................................................Passed
Check that executables have shebangs.........................................Passed
Check for case conflicts.....................................................Passed
Detect AWS Credentials.......................................................Passed
Detect Private Key...........................................................Passed
pyupgrade....................................................................Passed
flake8.......................................................................Passed
Check .rst files with the same linter used by pypi.org.......................Passed
```

</details>

It's definitely there [proxy_bypass](https://github.com/aio-libs/aiohttp/pull/4445/files/1ba92e39a4064357fdfeffc94d3af94ef63968f1#diff-b797dd8733928df191ba2061121ab8b69976c185fcbfad4534891d3252b9ac30R43).

aiohttp/helpers.py
Dreamsorcerer: Maybe somebody else has an opinion on it, but if we're just using something from the standard library, I'm not sure there's much benefit in re-exporting it (e.g. the other imports like getproxies are used within this file, not exported elsewhere).

It just jumped out at me as a little odd (which is only highlighted by the noqa). If there's the possibility that we might want to replace this function in the future with a different implementation, then perhaps it's reasonable.

PawelPeczek:
Hi there, could anyone estimate when there will be a chance to merge and release this fix?

webknjaz:
> Figures, doesn't fail locally when I run pre-commit.

I suppose the CI pulls-in newer code from master. You should probably rebase to get the same errors locally.

> Hi there, could anyone estimate when there will be a chance to merge and release this fix?

Once merged, this will need to be backported to the 3.8 branch and then we'd need to make that branch release-worthy which requires quite a bit of work to at least stabilize the CI.

.gitignore
```diff
@@ -63,3 +63,4 @@ sources
 var/*
 venv
 virtualenv.py
+.venv
```
webknjaz: This change is unrelated to the PR, it's best to keep it out.

CHANGES/4431.bugfix
```diff
@@ -0,0 +1 @@
+Fixes requests not honoring no_proxy environment variables.
```
webknjaz:
```suggestion
Fixed HTTP client requests to honor ``no_proxy`` environment variables.
```

aiohttp/client.py
```diff
@@ -444,11 +445,12 @@ async def _request(
                     if proxy is not None:
                         proxy = URL(proxy)
                     elif self._trust_env:
-                        for scheme, proxy_info in proxies_from_env().items():
-                            if scheme == url.scheme:
-                                proxy = proxy_info.proxy
-                                proxy_auth = proxy_info.proxy_auth
-                                break
+                        if not proxy_bypass(str(url.host)):
```
webknjaz: FYI `url.host` is already a string.

aiohttp/client.py
```diff
@@ -444,11 +445,12 @@ async def _request(
...
+                            for scheme, proxy_info in proxies_from_env().items():
```
webknjaz: I feel like determining the allowed proxies should be moved to a helper function or maybe inside of `proxies_from_env()`. I suggest having a new param `for_url` that, if passed, would act as a guard. It'd look like
```python
def proxies_from_env(for_url: yarl.URL = None):
    if for_url is not None and proxy_bypass(for_url.host):
        return {}

    ...
```

webknjaz (changes requested):
I've left a few suggestions to improve the structure. And this will also need some accompanying tests and docs changes. It's important to understand how this feature is supposed to be used by the users.

Also, it looks more like a feature so we should probably rename the change fragment to `.feature` instead of `.bugfix`.

.gitignore
webknjaz:
```suggestion
```

aiohttp/client.py
scirelli: Idk, that would changes what `proxies_from_env()`'s name suggests it's doing. The way it is now it reads
if url should be proxied; get the proxy from env

Maybe pull the logic out into a new helper function?
```python
def get_env_proxy_for_url(url: yarl.URL = None):
    proxy = None
    proxy_auth = None

    if not proxy_bypass(url.host):
        for scheme, proxy_info in proxies_from_env().items():
            if scheme == url.scheme:
                proxy = proxy_info.proxy
                proxy_auth = proxy_info.proxy_auth
                break

    return (proxy, proxy_auth)
```

webknjaz: Yeah, maybe a new helper would be good. Just don't do this much nesting please (also don't forget a `None`-guard:
```python
def get_env_proxy_for_url(url: yarl.URL = None):
    if for_url is not None and proxy_bypass(for_url.host):
        return None, None

    for scheme, proxy_info in proxies_from_env().items():
        if scheme == url.scheme:
            return proxy_info.proxy, proxy_info.proxy_auth

    return None, None
```

scirelli: > Just don't do this much nesting please

What's this in reference to?
My original PR was just adding an extra if condition. I could refactor some of the code if that's what you mean?

aiohttp/client.py
scirelli: I get a mypy error if I don't cast it to a str.
```
aiohttp/helpers.py:273: error: Argument 1 to "proxy_bypass" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

aiohttp/client.py
webknjaz: I was talking about the snippet in your previous comment, it has if->for->if->break->return flow that could've been more straightforward if rewritten as shown in my last comment.

aiohttp/client.py
webknjaz: Then the function typings should be adjusted I guess... FWIW I've shown the guard expression in one of the comments so it's accounted for.

Dreamsorcerer: The type error says that it is `Optional[str]`, in other words, it could be `None`. I don't think `str(None)` would be the right approach to resolving the type error.

It suggests you should probably have an `if url.host is None` check or a sensible default value with `url.host or "default"`.

aiohttp/client.py
scirelli: Oh yeah, sorry first time contributing to this repo, I wasn't sure how much refactoring would be okay.

I changed the code to be more straightforward since it was pulled out into a new function. Let me know if you want me to change it back to what's in your last comment.

webknjaz: No problem, that's why we have a review process in place :) I've added a code suggestion in the recent review with a bunch of comments on why I think those changes are necessary.

You can accept that code suggestion via GH UI but then sync it locally and adjust the tests.

aiohttp/client.py
webknjaz: Exactly, I've incorporated this idea in the most recent bunch of the review comments.

aiohttp/helpers.py
```diff
@@ -269,6 +269,15 @@ def proxies_from_env() -> Dict[str, ProxyInfo]:
     return ret
 
 
+def get_env_proxy_for_url(url: URL) -> Tuple[Optional[URL], Optional[BasicAuth]]:
+    if url is not None and not proxy_bypass(str(url.host)):
+        proxy_info = proxies_from_env().get(url.scheme, None)
+        if proxy_info is not None:
+            return proxy_info.proxy, proxy_info.proxy_auth
+
+    return None, None
```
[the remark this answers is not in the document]
scirelli: Yeah I like this better

aiohttp/helpers.py
```diff
@@ -269,6 +269,15 @@ def proxies_from_env() -> Dict[str, ProxyInfo]:
     return ret
 
 
+def get_env_proxy_for_url(url: URL) -> Tuple[Optional[URL], Optional[BasicAuth]]:
+    if url is not None and not proxy_bypass(str(url.host)):
```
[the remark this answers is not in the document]
Dreamsorcerer: Also, if url can be `None`, shouldn't the function annotation be `Optional[URL]`? I'd have thought mypy would have picked that up though (as a condition which is always true).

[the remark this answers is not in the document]
webknjaz: It was taken from an older snippet draft that had `url: URL = None`. Now the check should actually be `url.host is not None`.

aiohttp/helpers.py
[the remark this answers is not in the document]
webknjaz: FYI it's better to commit the suggested changes from the GH UI because this preserves the authorship.

tests/test_helpers.py
```diff
@@ -515,6 +515,54 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+
+def test_get_env_proxy_for_url_when_url_scheme_does_not_match_proxy_list(
+    mocker,
+) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=False)
+    proxies = {"https": helpers.ProxyInfo("https://example.com", "")}
+    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
+    url = URL("http://aiohttp.io/path")
+
+    with pytest.raises(LookupError, match="No proxies found for `[^`]*` in the env"):
```
webknjaz: It's best to use r-strings for regexes + in tests we have predictable values for that ```[^`]*``` thingy so it's best to just use that. It will also improve readability and simplify understanding the expectations for the readers.

tests/test_helpers.py
```diff
@@ -515,6 +515,54 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+
+
+def test_get_env_proxy_for_url_when_url_matches_the_no_proxy_list(mocker) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=True)
+    mocker.patch(
+        "aiohttp.helpers.proxies_from_env",
+        return_value={"http": helpers.ProxyInfo("http://example.com", "")},
+    )
+    url = URL("http://aiohttp.io/path")
+    with pytest.raises(LookupError, match=r"Proxying is disallowed for `[^`]*`"):
```
webknjaz: It's best to actually expect a specific URL in the message.

tests/test_helpers.py
```diff
@@ -515,6 +515,54 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+
+def test_get_env_proxy_for_url_when_url_is_empty(
+    mocker,
+) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=False)
+    proxies = {"https": helpers.ProxyInfo("https://example.com", "")}
+    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
+    url = URL("")
+
+    with pytest.raises(LookupError, match="No proxies found for `[^`]*` in the env"):
```
webknjaz: Please do the same as requested in other comments.

tests/test_helpers.py
```diff
@@ -515,6 +515,54 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+
+def test_get_env_proxy_for_url_when_url_matches_the_no_proxy_list(mocker) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=True)
+    mocker.patch(
+        "aiohttp.helpers.proxies_from_env",
+        return_value={"http": helpers.ProxyInfo("http://example.com", "")},
+    )
+    url = URL("http://aiohttp.io/path")
+    with pytest.raises(LookupError, match=r"Proxying is disallowed for `[^`]*`"):
+        proxy, proxy_auth = helpers.get_env_proxy_for_url(url)
```
webknjaz: Assigning the results to variables never happens because the exception is raised before that. So it doesn't make sense to have those.
```suggestion
        helpers.get_env_proxy_for_url(url)
```

tests/test_helpers.py
```diff
@@ -515,6 +515,54 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
     assert proxy_auth.encoding == "latin1"
 
 
+# --------------------- get_env_proxy_for_url ------------------------------
```
webknjaz, line 519: Could you also add a test for the positive outcome? You could use `@pytest.mark.parametrize` to run it with several sets of params if necessary.

webknjaz (commented):
@scirelli in general, I'm quite satisfied with the implementation but there are a few things in tests that need improvements. After addressing those, it should be possible to merge this PR. Thanks for your patience!

tests/test_helpers.py
scirelli: The positive case was the last one https://github.com/aio-libs/aiohttp/pull/4445/files#diff-08a6944653efc734f4c8cb856fc854b9f21d655ca69796af2b170c53bff7919eR556

I could add more than one

scirelli:
> @scirelli in general, I'm quite satisfied with the implementation but there are a few things in tests that need improvements. After addressing those, it should be possible to merge this PR. Thanks for your patience!

Thanks learning a lot!

tests/test_helpers.py
```diff
@@ -515,6 +515,71 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    proxies = {"https": helpers.ProxyInfo("https://example.com", "")}
+    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
+    url = URL("")
+
+    with pytest.raises(LookupError, match=r"No proxies found for `` in the env"):
+        helpers.get_env_proxy_for_url(url)
+
+
+@pytest.mark.parametrize(
+    "schema,proxyInfo,url",
```
webknjaz: Apparently, pytest lets you specify the params separately for readability:
```suggestion
    ("schema", "proxyInfo", "url"),
```

tests/test_helpers.py
```diff
@@ -515,6 +515,71 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+            "https",
+            helpers.ProxyInfo("https://other.com", ""),
+            URL("https://aiohttp.com"),
+        ),
+        (
+            "http",
+            helpers.ProxyInfo("http://example.com", ""),
+            URL("http://aiohttp.com"),
+        ),
+    ],
```
webknjaz: When the params are complex or even just non-scalar, pytest tries to stringify them for the report and they don't look helpful. So it's always best to provide explicit IDs in such cases:
```suggestion
    ],
    ids=("https", "http"),
```

Besides, one can use those params to invoke a specific test as follows:
```console
$ pytest 'tests/test_helpers.py::test_get_env_proxy_for_url_when_url_scheme_matches_proxy_list[https]'
```
(it'd be less trivial with a complex structure which would need the right amount of escaping things)

tests/test_helpers.py
```diff
@@ -515,6 +515,72 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+def test_get_env_proxy_for_url_when_url_is_empty(
+    mocker,
+) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=False)
+    proxies = {"https": helpers.ProxyInfo("https://example.com", "")}
+    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
+    url = URL("")
+
+    with pytest.raises(LookupError, match=r"No proxies found for `` in the env"):
+        helpers.get_env_proxy_for_url(url)
```
webknjaz: This looks like one test with different params:
```suggestion
@pytest.mark.parametrize(
    ("bypass", "proxies", "url_input", "expected_err_msg"),
    (
        (
            True, {"http": helpers.ProxyInfo("http://example.com", "")},
            "http://aiohttp.io/path", r"Proxying is disallowed for `'aiohttp.io'`",
        ),
        (
            False, {"https": helpers.ProxyInfo("https://example.com", "")},
            "http://aiohttp.io/path",
            r"No proxies found for `http://aiohttp.io/path` in the env",
        ),
        (
            False, {"https": helpers.ProxyInfo("https://example.com", "")},
            "", r"No proxies found for `` in the env",
        ),
    ),
    ids=(
        "url_matches_the_no_proxy_list",
        "url_scheme_does_not_match_proxy_list",
        "url_is_empty",
    ),
)
def test_get_env_proxy_for_url_negative(
    bypass, expected_err_msg, mocker, proxies, url_input,
) -> None:
    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=bypass)
    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
    url = URL(url_input)
    with pytest.raises(LookupError, match=expected_err_msg):
        helpers.get_env_proxy_for_url(url)
```

webknjaz (approved):

tests/test_helpers.py
```diff
@@ -515,6 +515,68 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+        ),
+    ),
+    ids=(
+        "url_matches_the_no_proxy_list",
+        "url_scheme_does_not_match_proxy_list",
+        "url_is_empty",
+    ),
+)
+def test_get_env_proxy_for_url_negative(
+    bypass, expected_err_msg, mocker, proxies, url_input,
```
webknjaz:
```suggestion
            True,
            {"http": helpers.ProxyInfo("http://example.com", "")},
            "http://aiohttp.io/path",
            r"Proxying is disallowed for `'aiohttp.io'`",
        ),
        (
            False,
            {"https": helpers.ProxyInfo("https://example.com", "")},
            "http://aiohttp.io/path",
            r"No proxies found for `http://aiohttp.io/path` in the env",
        ),
        (
            False,
            {"https": helpers.ProxyInfo("https://example.com", "")},
            "",
            r"No proxies found for `` in the env",
        ),
    ),
    ids=(
        "url_matches_the_no_proxy_list",
        "url_scheme_does_not_match_proxy_list",
        "url_is_empty",
    ),
)
def test_get_env_proxy_for_url_negative(
    bypass,
    expected_err_msg,
    mocker,
    proxies,
    url_input,
```

tests/test_helpers.py
```diff
@@ -515,6 +515,77 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+)
+def test_get_env_proxy_for_url_negative(
+    bypass,
+    expected_err_msg,
+    mocker,
+    proxies,
+    url_input,
+) -> None:
+    mocker.patch("aiohttp.helpers.proxy_bypass", return_value=bypass)
+    mocker.patch("aiohttp.helpers.proxies_from_env", return_value=proxies)
```
webknjaz: @scirelli now that I look at this closer and I realize that it all depends on the env vars. Could you explore using `monkeypatch.setenv()` instead of `mocker.patch()`? https://docs.pytest.org/en/stable/monkeypatch.html

It's always better to have a full integration test rather than a unit test with everything mocked.

webknjaz: Setting env vars would also demonstrate how this feature is supposed to work in real life.

scirelli: Done. 

FYI:
> WARNING  aiohttp.client:helpers.py:253 HTTPS proxies https://example.com are not supported, ignoring

Which is why all proxies are http.

scirelli: Btw, am I supposed to click "Resolve conversation" on these or do you?

webknjaz: It depends. It's usually fine to resolve if you did what was asked but if there's an ongoing debate, it's better to keep it open. GH autocollapses conversations on the changed lines anyway marking them outdated. But if it's marked manually but it's not actually solved, it's harder for the reviewers to check what is the state of the PR because it adds more cognitive overhead to try to remember what was discussed compared to what was changed.

webknjaz: I recently saw this nice article on reviews, you may find it interesting: https://mtlynch.io/code-review-love/.

tests/test_helpers.py
```diff
@@ -515,6 +516,106 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    (
+        (
+            {"no_proxy": ["aiohttp.io"], "NO_PROXY": ["aiohttp.io"]},
+            "http://aiohttp.io/path",
+            r"Proxying is disallowed for `'aiohttp.io'`",
+        ),
+        (
+            {
+                "http_proxy": ["http://example.com"],
+                "HTTP_PROXY": ["http://example.com"],
```
webknjaz: I don't understand why it's necessary to have both spellings in the same test. It could be reasonable to test them separately, though. And maybe have different spellings with different values testing which one overrides another.

WDYT?

scirelli: You mean the case difference? If I remember correctly (been awhile, I'll go check) but there's no spec. for the proxy info, it's just by convention. And there's more than one; one with lower case and the other upper. I think there's even `all_proxy` `ALL_PROXY`. I looked at the source of [urllib.request.get_env_proxy_for_url](https://github.com/python/cpython/blob/7cb033c423b65def1632d6c3c747111543b342a2/Lib/urllib/request.py#L2500-L2503) and they just lower case it. So it doesn't matter, I guess I'm just used to seeing them together all the time.

webknjaz: Yes, I mean the case difference. When you write a test, you communicate the expectations of how something is supposed to be used and to work in general. Point is that when you set both `http_proxy` **and** `HTTP_PROXY`, this creates a false impression (for any third party reader) and in order for this feature to work, the end-user is supposed to set two env variables with the same value. The underlying implementation you linked implies that the upper-case variant is always preferred and the lower-case is used as a fall-back.
The only reasonable case I see for setting both variables in a test is when the values are different: for example, a test could demonstrate that with `http_proxy=http://example.org HTTP_PROXY=http://example.com` set it would choose `http://example.com`.

tests/test_helpers.py
```diff
@@ -515,6 +516,106 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    ),
+)
+def test_get_env_proxy_for_url_negative(
+    monkeypatch, proxies, url_input, expected_err_msg
+) -> None:
+    for schema in getproxies_environment().keys():
+        monkeypatch.delenv(schema, False)
+
+    for proxy_type, proxy_list in proxies.items():
+        monkeypatch.setenv(proxy_type, ",".join(proxy_list))
```
webknjaz: `str.join()` never has any effect because all the params have exactly one item in the list. I'd like to make two points here:
1) If two proxies can be present, we should actually test this case: there's no point in implying that this is the case but not checking it.
2) In tests, we must do as few runtime-dependent changes / conditional coding as possible. The cases should be statically defined because adding clever code to tests opens up more possibilities to add bugs in those tests. They should be as simple as possible. So instead of having a list of one item that is concatenated into the same one item, it's better to (1) remove `str.join()`, (2) turn lists into strings and (3) just have a string with two proxies and a comma in between in the specific param.

scirelli: 1. I used a list because the `xxx_proxy` can be a comma separated list. `no_proxy` is one that I see often with a list of domains/ips not to proxy. I'll add more test cases where there's multiple proxies in the list.
2. Okay, I'll make them a comma separated string.

tests/test_helpers.py
```diff
@@ -515,6 +516,106 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
     assert proxy_auth.encoding == "latin1"
 
 
+# --------------------- get_env_proxy_for_url ------------------------------
+
+
+@pytest.mark.parametrize(
+    ("proxies", "url_input", "expected_err_msg"),
```
webknjaz: Since the `proxies` arg is used for env vars, it's probably best to rename it to reflect this with `proxies_env`, or `proxy_env_vars`, or something similar. The intent is to communicate what this var is better without forcing the readers to try understanding the test implementation.

tests/test_helpers.py
```diff
@@ -515,6 +516,106 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    ids=("url_scheme_match_http_proxy_list", "url_scheme_match_https_proxy_list"),
+)
+def test_get_env_proxy_for_url(
+    monkeypatch, proxies, url_input, expected_err_msg
+) -> None:
+    for schema in getproxies_environment().keys():
+        monkeypatch.delenv(schema, False)
+
+    for proxy_type, proxy_list in proxies.items():
+        monkeypatch.setenv(proxy_type, ",".join(proxy_list))
```
webknjaz: Future note: normally I'd move this setup step into a pytest fixture and use "indirect" to bypass the params there (as in `@pytest.mark.parametrize(('proxy_env_vars', ...), ..., indirect=('proxy_env_vars', ))` + `@pytest.mark.usefixtures('proxy_env_vars')`).

It could look like:
```python
@pytest.fixture
def proxy_env_vars(request):
    for schema in getproxies_environment().keys():
        monkeypatch.delenv(schema, False)

    for var_name, var_value in request.param.items():
        monkeypatch.setenv(var_name, var_value)
```

tests/test_helpers.py
scirelli: Changing to `proxy_env_vars`, I like that.

tests/test_helpers.py
scirelli: I like this. Done.

tests/test_helpers.py
scirelli: Make sense. Just doing a little debugging now. Before when I ran the tests I didn't have any environment variables set for the proxies, setting them, then running the tests, the tests fail.

Looks like some other tests break:
```
FAILED tests/test_helpers.py::test_proxies_from_env[https] - AssertionError: assert dict_keys(['http', 'https']) == {'https'}
FAILED tests/test_helpers.py::test_proxies_from_env[ws] - AssertionError: assert dict_keys(['http', 'ws']) == {'ws'}
FAILED tests/test_helpers.py::test_proxies_from_env[wss] - AssertionError: assert dict_keys(['http', 'wss']) == {'wss'}
FAILED tests/test_helpers.py::test_proxies_from_env_skipped[https] - AssertionError: assert {'http': ProxyInfo(proxy=URL('http://www.example.com,http//proxy.com'), proxy_auth=None)} == {}
FAILED tests/test_helpers.py::test_proxies_from_env_skipped[wss] - AssertionError: assert {'http': ProxyInfo(proxy=URL('http://www.example.com,http//proxy.com'), proxy_auth=None)} == {}
```

To reproduce:
```
export http_proxy='http://www.example.com,http://proxy.com'
make test
```

Should I create an issue for this? I can push a PR after this one to fix it....once I figure it out.

tests/test_helpers.py
```diff
@@ -515,6 +516,97 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+        ),
+    ),
+    indirect=["proxy_env_vars"],
+    ids=(
+        "url_scheme_match_http_proxy_list",
+        "url_scheme_match_https_proxy_list",
+        "url_scheme_match_http_proxy_list_multiple",
+    ),
+)
+def test_get_env_proxy_for_url(monkeypatch, proxy_env_vars, url_input) -> None:
```
webknjaz: `monkeypatch` is no longer used here
```suggestion
def test_get_env_proxy_for_url(proxy_env_vars, url_input) -> None:
```

tests/test_helpers.py
```diff
@@ -515,6 +516,97 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+        "url_matches_the_no_proxy_list_multiple",
+        "url_scheme_does_not_match_http_proxy_list",
+        "url_scheme_does_not_match_https_proxy_list",
+        "no_proxies_are_set",
+        "url_is_empty",
+    ),
+)
+def test_get_env_proxy_for_url_negative(
+    monkeypatch, proxy_env_vars, url_input, expected_err_msg
+) -> None:
```
webknjaz: `monkeypatch`, `proxy_env_vars` aren't accessed directly anymore:
```suggestion
@pytest.mark.usefixtures('proxy_env_vars')
def test_get_env_proxy_for_url_negative(url_input, expected_err_msg) -> None:
```

tests/test_helpers.py
```diff
@@ -515,6 +516,96 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    ids=(
+        "url_scheme_match_http_proxy_list",
+        "url_scheme_match_https_proxy_list",
+        "url_scheme_match_http_proxy_list_multiple",
+    ),
+)
+def test_get_env_proxy_for_url(proxy_env_vars, url_input) -> None:
+    url = URL(url_input)
+    proxy, proxy_auth = helpers.get_env_proxy_for_url(url)
+    proxy_list = proxy_env_vars[url.scheme + "_proxy"]
```
webknjaz, line 604: Normally, it's best to avoid dynamism like this in tests because it makes it harder to reason about what's expected because one should first dig through several levels of indirection correctly.

But since I don't want an infinite number of reviews on this PR, I'll just accept it. I think it's rather good now.

webknjaz (approved):

tests/test_helpers.py
```diff
@@ -515,6 +516,96 @@ def test_proxies_from_env_http_with_auth(mocker) -> None:
...
+    ids=(
+        "url_matches_the_no_proxy_list",
+        "url_matches_the_no_proxy_list_multiple",
+        "url_scheme_does_not_match_http_proxy_list",
+        "url_scheme_does_not_match_https_proxy_list",
+        "no_proxies_are_set",
+        "url_is_empty",
+    ),
+)
+@pytest.mark.usefixtures('proxy_env_vars')
```
webknjaz:
```suggestion
@pytest.mark.usefixtures("proxy_env_vars")
```

webknjaz:
@scirelli could you send a backport PR against the `3.8` branch?
https://docs.aiohttp.org/en/stable/contributing.html#backporting

Issue #4453, opened 2019-12-19
"Line is too long" for JSON-lines from Kubernetes API


nolar:
## Long story short

Iterating over long lines in a streaming response fails in `StreamReader` with `ValueError: Line is too long`.

It affects all the APIs that return [JSON-lines format](http://jsonlines.org/), such as Kubernetes API, where a JSON-serialised object can be much more than current limit of 128 KB — e.g. 2 MB.

## Expected behaviour

Any buffer size is supported, or the buffer size is configurable per session/request/response.

## Actual behaviour

The buffer size is limited to lo-watermark 64 KB, hi-watermark 128 KB, and is not configurable except as by hacking the protected properties (`response.content._high_water` in 2019, was `response.content._limit` in 2017) or global constants (`DEFAULT_LIMIT`).

## Steps to reproduce

See https://github.com/zalando-incubator/kopf/issues/275 and https://github.com/zalando-incubator/kopf/issues/276 — a workaround with a self-made line iterator.

For example, have a Kubernetes cluster, create a secret with 2MB encrypted and base64-encoded line, and try to _watch_ over the secrets — the content is a JSON-lines streaming response.

Any other artificial ways to use JSON-lines will show this same issue.

The [JSON-lines format](http://jsonlines.org/) implies that the objects cannot be split to multiple lines anyway, i.e. it is one object per line — which is a problem for `aiohttp` as a client.

UPD: Here is an isolated example to reproduce the issue:

```python
# pip install pytest pytest-asyncio aresponses
# pytest -s -vv _thisfile.py

import aiohttp
import json
import pytest


async def client_fn():
    session = aiohttp.ClientSession()
    async with session:
        response = await session.get('http://xyz/path')
        async with response:
            async for line in response.content:
                print('===', len(line))
                assert len(line) >= 1 * 1024 * 1024


@pytest.mark.asyncio
async def test_long_line(event_loop, aresponses):
    big_objs = [
        {'spec': {'field': 'x' * 1 * 1024 * 1024}},
        {'spec': {'field': 'y' * 1 * 1024 * 1024}},
        {'spec': {'field': 'z' * 1 * 1024 * 1024}},
    ]
    content = ''.join([f'{json.dumps(obj)}\n' for obj in big_objs])
    aresponses.add('xyz', '/path', 'get', content)
    await client_fn()
```

## Your environment

`aiohttp==3.6.2` and newer.
client.

## Links

[...]

PR #4456, opened 2019-12-19
Expose the response's content buffering limit for lines longer than 128KB

nolar:
## What do these changes do?

Solve the problem with "Line is too long" for JSON-lines streaming responses (Kubernetes API and similar), as described in #4453.

An extra optional kwarg is exposed for the requests, which affects the response's content buffering:

```python
        response = await session.get('http://xyz/path', limit=1*1024*1024)
```

The value of `0` means no limit — which can affect the memory footprint:

```python
        response = await session.get('http://xyz/path', limit=0)
```

The default is `2**16`, as it was before, which leads to the buffer size limit of 128 KB (the high-watermark is 2x of the limit itself).

## PR notes

I have doubts on the following topics — it would be nice to clarify them:

* Cython & `.pyx` files — I have no experience with this, just changed like the surrounding code.
* Should it be named `limit=` for the users? Or maybe `response_buffer_limit=`?
* What's the purpose of the low-watermark, of the actual memory consumed got up to 2x of that limit, not as would be expected by the users?
* Monkey-patching of `DEFAULT_LIMIT` (if it was used by anyone) will break now, as the imports are done by-value into multiple modules. I usually use the Google-style imports (`from pkg import mod; mod.VAL`), which work with any monkey-patching quite well. I didn't find such cases in aiohttp, so I'm not sure if this will match the code style.

The code is developed and manually tested against v3.6.2. The master-based version is just a rebase, in the assumption that it works. I could not test the master/v4.0.0a1 code, as it break `aresponses` completely, and the example uses `aresponses` to have a minimalistic web server locally.

The tests (`make test`) on v3.6.2 & v4.0.0a1 fail with `ValueError: option names {'--aiohttp-fast'} already added` — some help would be needed to make them run.

## Related issue number

#4453 

## Checklist

### DIFF
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -20,7 +20,8 @@ from .http_writer import (HttpVersion as _HttpVersion,
                           HttpVersion10 as _HttpVersion10,
                           HttpVersion11 as _HttpVersion11)
 from .http_parser import DeflateBuffer as _DeflateBuffer
-from .streams import (EMPTY_PAYLOAD as _EMPTY_PAYLOAD,
+from .streams import (DEFAULT_LIMIT as _DEFAULT_LIMIT,
+                      EMPTY_PAYLOAD as _EMPTY_PAYLOAD,
                       StreamReader as _StreamReader)
 
 cimport cython
@@ -266,6 +267,7 @@ cdef class HttpParser:
         object _loop
         object _timer
 
+        size_t _limit
         size_t _max_line_size
         size_t _max_field_size
         size_t _max_headers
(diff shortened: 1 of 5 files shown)


webknjaz:
Thanks for the PR! Let's see what Andrew thinks.

asvetlov:
I think that for exceptional cases like yours you can always read the whole text and split it by lines explicitly if needed.
Passing the line size limit through the whole API is cumbersome, it is done for very little effect and affects too few people.

nolar:
That's the point: I cannot read the whole text. This is a stream of JSON lines, one line per event as soon as these events happen. The stream never ends and the connection never disconnects (in theory; in practice, there are timeouts and unstable networks).

So, I never know in advance how big is the response to read, and I should handle the incoming lines immediately, while the connection can remain alive but contentless even hours after that.

Currently, I have implemented it with my own line reader (https://github.com/zalando-incubator/kopf/pull/276/files), which is fine for me, but it looks like a job of the HTTP client library: the use-case of streaming JSON lines is not so rare, I presume. It is, however, for you to decide on should it be part of the library or not.

asvetlov:
Your solution is fine, it is only 20 lines of simple code.
Adding the feature to aiohttp makes the source much more complex and harder to understand, at least with the current design of streams.

nolar:
Okay. Then, it makes sense to close this PR. Thank you for your attention.

Issue #4269, opened 2019-10-25
Use loop.sendfile() for file sending on Python 3.7+


asvetlov:
Currently, the server code has a custom function that should be replaced with the standard `loop.sendfile()` if available.

_Originally posted by @asvetlov in https://github.com/aio-libs/aiohttp/issues/2245#issuecomment-546409157_

Issue #4268, opened 2019-10-25
Use loop.starttls() on Python 3.7+


asvetlov:
Now TCPConnector has a custom code for upgrading proxy from HTTP to HTTPS on `CONNECT` http verb.

I think we should start using `loop.starttls()` on Python 3.7+ to fix the issue.

_Originally posted by @asvetlov in https://github.com/aio-libs/aiohttp/issues/2245#issuecomment-546409157_

aivrar:
@asvetlov Hello sir.

I was wondering what you are referring to with `loop.starttls()`? Do you think this something that can help my current issue here? This is my comment link in the issues thread for aiohttp.

https://github.com/aio-libs/aiohttp/issues/4429#issuecomment-564038500

I have been getting some help:
https://pastebin.com/atLw1hC7
![image](https://user-images.githubusercontent.com/42790594/70697962-1fa30d00-1c8c-11ea-856a-ee2a4b30e506.png)

I'd really like to get this working, I'm not sure if this is a problem on my end and how to remedy it.

Thanks

Allexio:
Hi is there any update on this issue? Any planned upgrade to fix the fact that http proxies don't work anymore on windows?
This is a problem for me because it breaks the slackclient library when using a proxy :/
( https://github.com/slackapi/python-slackclient/issues/642 )

asvetlov:
It will be possible starting from aiohttp 4.0 only at least.
`loop,starttls()` doesn't work together with our home-made workaround.

aivrar:
Wow, I'm glad I stayed subscribed because I'm raring to go here.  Looking forward to update.

munding:
python 3.9 there are still such problem

HMaker:
@webknjaz I tried using HTTP proxy with aiohttp 3.7.4 and `ProactorEventLoop` on Windows 10 but it faiils with `OSError(22, 'The parameter is incorrect', None, 87, None))` on proxy connection. Same error reported by #2245.

webknjaz:
This has been implemented and will be available with hacks under aiohttp 3.8+ (and Python 3.7+). The 3.7 stream is EOL and will not get any updates.

PR #4463, opened 2019-12-20
Fix python 3.8 tests under Windows 

hh-h:
## What do these changes do?

It should fix tests for python 3.8 under Windows

No

## Related issue number

No

## Checklist


webknjaz:
Plz fill in the issue template/description/title with some meaningful explanation.

hh-h:
I'm trying to fix python 3.8 windows tests, but I don't understand how to run azure pipelines for PR, any help @webknjaz ?

webknjaz:
I think, all CI checks start up automatically when sent against master. So probably it's because you send a PR against `py38` branch. Switch it.

webknjaz:
And you should still update the PR title and the description.

hh-h:
But there's no python 3.8 in master nor 3.7. Can I temporary include py38 here https://github.com/aio-libs/aiohttp/blob/master/.azure-pipelines/ci.yml#L4 ?

webknjaz:
Your branch already contains commits from `py38`.

hh-h:
Okay, I see two ways
- create branch from `3.7`, cherry-pick CI commit from `py38` and add my changes
- update ci.yml and temporary add `py38` branch there for testing

What would you suggest?

webknjaz:
No, I think that you shouldn't recreate the PR. Just change the base branch. I'll update it myself. But for the future — just click `Edit` right to the PR title and you'll see a UI control to do it.

tests/test_web_functional.py
```diff
@@ -1443,7 +1443,7 @@ def make_handler(appname):
     assert 413 == resp.status
     resp_text = await resp.text()
     assert 'Maximum request body size 1048576 exceeded, ' \
-           'actual body size 1048591' in resp_text
+           'actual body size' in resp_text
```
webknjaz, line 1446: Why is it different?

webknjaz:
Oh. And the trick is that you should only add/change your own commits and keep those that are derived from `py38` unchanged. This will simplify merging your branch there or rebasing it avoiding merge conflicts.

aiohttp/pytest_plugin.py
```diff
@@ -230,6 +230,23 @@ def proactor_loop():  # type: ignore
         yield _loop
 
 
+@pytest.fixture
+def selector_loop():  # type: ignore
+    if not PY_37:
+        policy = asyncio.get_event_loop_policy()
```
webknjaz, line 236: Won't `3.8` also hit this branch?

aiohttp/pytest_plugin.py
```diff
@@ -230,6 +230,23 @@ def proactor_loop():  # type: ignore
...
+        policy._loop_factory = asyncio.SelectorEventLoop  # type: ignore
```
webknjaz, line 237: I don't think that it's wise to rely on private APIs.

tests/test_web_functional.py
```diff
@@ -1469,7 +1469,7 @@ def make_handler(appname):
     assert 413 == resp.status
     resp_text = await resp.text()
     assert 'Maximum request body size 2097152 exceeded, ' \
-           'actual body size 2097166' in resp_text
+           'actual body size' in resp_text
```
webknjaz, line 1475: If the number is different, we need some explanation to comment here why this is happening. Is it possible to figure out the precise number that would appear here?

tests/test_web_functional.py
hh-h: this test I wanted to discuss with @asvetlov , this case is a little bit weird and error message of `HTTPRequestEntityTooLarge` is invalid on some circumstances, proof

server.py
```python
from aiohttp import web


async def handle(request):
    await request.post()
    return web.Response()


app = web.Application(client_max_size=10)
app.add_routes([web.post("/", handle)])

web.run_app(app)
```

client.py
```python
import asyncio
import json


async def main():
        a = {"test": 1024**2 * 'x'}
        print(len(json.dumps(a)))   # prints 1048588
        async with session.post("http://0.0.0.0:8080", data=a) as resp:
            print(await resp.text())  # prints Maximum request body size 10 exceeded, actual body size 261948 

asyncio.run(main())
```
261948 != 1048588

aiohttp/pytest_plugin.py
hh-h: this is modified copypasta from `proactor_loop` :)

tests/test_web_functional.py
hh-h: see my comment https://github.com/aio-libs/aiohttp/pull/4463#discussion_r360393786

aiohttp/pytest_plugin.py
hh-h: Nope. https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L57

aiohttp/pytest_plugin.py
```diff
@@ -230,6 +230,23 @@ def proactor_loop():  # type: ignore
...
+    else:
+        if PY_38:
+            policy = asyncio.WindowsSelectorEventLoopPolicy()  # type: ignore
+        else:
+            policy = asyncio.WindowsProactorEventLoopPolicy()  # type: ignore
+        asyncio.set_event_loop_policy(policy)
+
+    with loop_context(policy.new_event_loop) as _loop:
+        asyncio.set_event_loop(_loop)
+        yield _loop
```
webknjaz, line 247: Is it possible to unset the loop after this `yield`?

aiohttp/pytest_plugin.py
webknjaz: Oh, so the var name is unfortunate.

tests/test_web_functional.py
webknjaz: So this probably means that the doesn't read the whole response which may be a bug so we shouldn't ignore it like this. Or at least we need some assertion that this number is bigger.

aiohttp/pytest_plugin.py
hh-h: Yes, at first I got confused too...

webknjaz: so maybe `... import PY_38 as AT_LEAST_PY_38` then?

tests/test_web_functional.py
hh-h: That was my first thought, but I wanted to discuss it first.

aiohttp/pytest_plugin.py
hh-h: I think this is out of scope and the general idea of PY_3* should be reworked.

aiohttp/pytest_plugin.py
hh-h: I'm not sure it makes sense since fixture executed once for each test, so there's no need to unset the loop at all, but I can be wrong. Also, see [loop fixture](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L57) and [proactor_loop fixture](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/pytest_plugin.py#L220)

tests/test_web_functional.py
hh-h: Updated tests, WDYT @webknjaz ?

asvetlov:
This pull request has been mentioned on **aio-libs**. There might be relevant details there:

https://aio-libs.discourse.group/t/aiohttp-new-release/49/5

asvetlov:
Thanks for the PR.
A part of the fixes is correct and can be applied.
The total switching from the proactor loop to select based is wrong, sorry.
The proactor event loop is the default on Python 3.8, aiohttp should work with it out of the box.
Unfortunately, tricks for emulation of `starttls` and `sendfile` don't work with proactor. #4268 and #4269 should be implemented for publishing Python 3.8 compatible release.

hh-h:
I had to add force selector_loop only for testing wrong loop, since proactor is default for windows after 3.8, as you mentioned. Only two tests should use force selector_loop, `test_named_pipe_connector_wrong_loop` and `test_named_pipe_runner_wrong_loop`, could you consider the new information or what should I do with PR?

asvetlov:
Superseded by #4513

@hh-h thanks for the PR

PR #4464, opened 2019-12-20
Added support of tempfile as payload

hh-h:
## What do these changes do?

tempfile can be used as a payload

tempfile can be used as a payload

## Related issue number

#4432 

## Checklist

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -1,4 +1,3 @@
-import abc
 import asyncio
 import enum
 import io
@@ -6,7 +5,7 @@ import json
 import mimetypes
 import os
 import warnings
-from abc import ABC, abstractmethod
+from abc import ABC, ABCMeta, abstractmethod
 from itertools import chain
 from typing import (
     IO,
@@ -441,7 +440,7 @@ class StreamReaderPayload(AsyncIterablePayload):
         super().__init__(value.iter_any(), *args, **kwargs)
 
 
-class TempFileMetaclass(abc.ABCMeta):
+class TempFileMetaclass(ABCMeta):
     def __instancecheck__(self, instance: Any) -> bool:
         return hasattr(instance, 'read') and hasattr(instance, 'close')
 
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -1,3 +1,4 @@
+import abc
 import asyncio
 import enum
 import io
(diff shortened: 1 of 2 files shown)


hh-h:
@serhiy-storchaka please review if you have some spare time

aiohttp/payload.py
```diff
@@ -440,6 +440,15 @@ def __init__(self, value: StreamReader, *args: Any, **kwargs: Any) -> None:
         super().__init__(value.iter_any(), *args, **kwargs)
 
 
+class TempFileMetaclass(ABCMeta):
```
asvetlov, line 443: I think `Protocol` and `runtime_checkable` from `typing_extensions` can serve this duty better.
See https://docs.python.org/3/library/typing.html#typing.runtime_checkable for docs.

hh-h: `Protocol` and `runtime_checkable` introduced in python 3.8, but what should we do for 3.5-3.7 then?

tests/test_client_request.py
```diff
@@ -1140,3 +1141,28 @@ def test_loose_cookies_types(loop) -> None:
...
+            req = ClientRequest(
+                meth, URL('http://python.org/'),
+                data=f, loop=loop)
+            resp = await req.send(conn)
+            assert '/' == req.url.path
+            assert isinstance(req.body, payload.IOBasePayload)
+            assert body == req.body._value.read()
+            assert 'application/octet-stream' == req.headers['CONTENT-TYPE']
+            await req.close()
+            resp.close()
```
webknjaz, line 1168: you probably need a try/finally block here

aiohttp/payload.py
```diff
@@ -448,7 +457,7 @@ def __init__(self, value: StreamReader, *args: Any, **kwargs: Any) -> None:
 PAYLOAD_REGISTRY.register(BytesIOPayload, io.BytesIO)
 PAYLOAD_REGISTRY.register(
     BufferedReaderPayload, (io.BufferedReader, io.BufferedRandom))
-PAYLOAD_REGISTRY.register(IOBasePayload, io.IOBase)
+PAYLOAD_REGISTRY.register(IOBasePayload, (io.IOBase, TempFilePayload))
```
thehesiod, line 460: shouldn't this be `BytesIOPayload` and not `IOBasePayload` as the `size` method is available

thehesiod: the other issue is that wouldn't it depend on how the file is opened?  a tempfile can be opened as bytes or string

Issue #4482, opened 2020-01-03
Client should accept a json_serialize function which returns bytes


ZacharyThomas:
## Long story short

Orjson (https://pypi.org/project/orjson/) dumps returns bytes rather than str. When using it as our serializer, we need to wrap it in a lambda to return a string, which the client library then promptly converts back to bytes. 

```
http_session = aiohttp.ClientSession(json_serialize=lambda x: orjson.dumps(x).decode())
```

## Expected behaviour

The client should accept both bytes and str 

## Actual behaviour

Client throws when using orjson serialization because it returns bytes

## Steps to reproduce

```
>>> import orjson                                                                  
>>> import aiohttp 
>>> import asyncio
>>> loop = asyncio.get_event_loop()                                                                
>>> sesh = aiohttp.ClientSession(json_serialize=orjson.dumps)                      
>>> async def do_post(): 
...      async with sesh.post("google.com", json={"body": "yes"}) as resp:
...        print("ok")

>>> loop.run_until_complete(do_post())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/3.7.5/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 579, in run_until_complete
    return future.result()
  File "<stdin>", line 2, in do_post
  File "/Users/zacharysmith/workspace/envs/lib/python3.7/site-packages/aiohttp/client.py", line 1012, in __aenter__
    self._resp = await self._coro
  File "/Users/zacharysmith/workspace/envs/lib/python3.7/site-packages/aiohttp/client.py", line 365, in _request
    data = payload.JsonPayload(json, dumps=self._json_serialize)
  File "/Users/zacharysmith/workspace/envs/lib/python3.7/site-packages/aiohttp/payload.py", line 389, in __init__
    dumps(value).encode(encoding),
AttributeError: 'bytes' object has no attribute 'encode'

```

## Your environment
client
aiohttp==3.6.2
python==3.7.5

ZacharyThomas:
@asvetlov Do you think a workable solution would be a major version change which changes the interface of both the Encoder/Decoder to take bytes?

derlih:
@ZacharyThomas 
From the documentation of `json.dumps`: `Serialize obj to a JSON formatted str`.
If I would need to get bytes directly, I would use `json.dump` with `io.BytesIO` as `fp` parameter.

[...]

PR #4483, opened 2020-01-04
client: allow for json serializer function which returns bytes

ZacharyThomas:
Allows a `json_serializer` function passed to the request return either bytes or string.

Should be transparent to users, other than accepting a new format, it should not change existing behavior

## Related issue number

#4482 

## Checklist

### DIFF
--- a/CHANGES/4482.feature
+++ b/CHANGES/4482.feature
@@ -1,0 +1,1 @@
+Allow Client to accept a json_serialize function which returns bytes
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -287,6 +287,7 @@ Yuriy Shatrov
 Yury Selivanov
 Yusuke Tsutsumi
 Yuval Ofir
+Zachary Smith
 Zlatan Sičanica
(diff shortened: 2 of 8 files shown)


asvetlov:
The PR modifies JSON encoding only without touching JSON decoding part.
I don't think that this is good design.

Another problem is the signatures: `JSONEncoder = Callable[[Any], Union[str, bytes]]` simply means that the result cannot be calculated at the static analyzer execution time. 
In turn, it means that the proposal was not cogitated carefully and have a bad smell.

I don't know how to solve this better but want to hear an opinion of other @aio-libs/aiohttp-committers

ZacharyThomas:
@asvetlov I can add the corresponding change to the Encoder if you think that is better design. Unfortunately I don't have a solution for the Union[bytes, str] - so if that's a blocker I will wait for another person to weigh in.

asvetlov:
Thanks for your contribution.
Sorry, I should decline the PR because I have a guts feeling that returning either str or bytes from json encoder is a bad idea.

jacobmonck:
sadge we need orjson support

PR #4505, opened 2020-01-16
Running aiohttp server in clustered mode

maestroprog:
## What do these changes do?

This change adds a new ability to start the server in cluster mode, using several processes instead of one (using fork). This is useful when a service needs to be running on multiple CPU cores, without manually starting each individual process.

To some extent, this is an alternative to Gunicorn.

This change only adds new lines of code, other has not changed

## Related issue number

Have not found

## Checklist

if these changes are useful for aiohttp, I will write unit tests and add documentation :)

### DIFF
--- a/aiohttp/cluster.py
+++ b/aiohttp/cluster.py
@@ -1,0 +1,406 @@
+"""Async cluster http worker for aiohttp.web"""
+
+import argparse
+import asyncio
+import fcntl
+import logging
+import os
+import signal
+import socket
+import sys
+import time
+import traceback
+from importlib import import_module
+from typing import MutableMapping, Optional, List
+
+import setproctitle
+
+from aiohttp import web
+from aiohttp.web_app import Application
+
+
+class AbstractWorker:
+    message = 'alive'
+    message_data = bytes(message, 'utf-8')
+    proctitle = None
+
+    def __init__(self, app, log: logging.Logger):
+        self.app = app
+        self.log = log
+        if AbstractWorker.proctitle is None:
+            AbstractWorker.proctitle = setproctitle.getproctitle()
+
+    def run(self) -> Optional:
+        result = asyncio.run(self._run())
+        asyncio.run(self.shutdown(close_resources=False))
+        return result
+
+    async def _run(self) -> None:
+        raise NotImplementedError
+
+    async def shutdown(self, close_resources=True) -> None:
+        raise NotImplementedError
+
+
+class ClusteredWorker(AbstractWorker):
+    def __init__(self, app, log, sock, ppid, fd, graceful_timeout=60.0):
+        super().__init__(app, log)
+
+        self._ppid = ppid
+        self._fd = fd  # write socket descriptor for child process
+        self._last_written_at = None  # last write timestamp into child socket
+        self.sock = sock
+        self.graceful_timeout = graceful_timeout
+
+        self.runner = None
+
+    async def _run(self) -> None:
+
+        setproctitle.setproctitle("%s Fork process [%d]" % (self.proctitle, os.getpid()))
+
+        self.runner = await self._run_app()
+
+        try:
+            while True:
+                await asyncio.sleep(1.0)
+                self._notify_parent()
+        except BaseException as e:
+            print(e)
+            traceback.print_tb(e)
+            pass
+
+    async def shutdown(self, close_resources=True) -> None:
+        if self.runner is not None:
+            await self.runner.cleanup()
+
+    async def _run_app(self) -> web.AppRunner:
+        runner = web.AppRunner(self.app, logger=self.log, keepalive_timeout=15.0, access_log=None)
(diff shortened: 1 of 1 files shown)


socketpair:
I’m against. The proven way is to use SO_REUSEPORT. If someone needs cpu scaling - its better to make template systemd job, or make a k8s pod with container per cpu. Anyway, using fork() instead of multiprocessing module is strange.

maestroprog:
> I’m against. The proven way is to use SO_REUSEPORT. If someone needs cpu scaling - its better to make template systemd job, or make a k8s pod with container per cpu. Anyway, using fork() instead of multiprocessing module is strange.

Thanks for the feedback.

I do not mind. I do not want to say that this is the only right way. this is only one of the approaches that is practiced in different projects, like the one used, for example, in the puma web server: https://github.com/puma/puma
It would be foolish not to use such an opportunity if it is needed. if the community does not need it - I do not mind.

about SO_REUSEPORT - in this code it is necessary in cases where the server crashed, and this is optional.

asvetlov:
1. For me, the clustered mode is something about running a server on multiple machines in a cluster.  This PR is about the multi-processed mode, not about the clustering.
2. As @socketpair mentioned there are too many ways to run multiple processes for the server. You invented another one, that's fine. Should the aiohttp team accept it into the code base and provide the feature maintenance -- this is the question. I personally don't want to get yet another maintenance burden.
3. The multi-process mode is not a drop-in replacement for single-process one, server processes have no shared memory and need a sort of communication channel for synchronizing its work. This is definitely possible but requires much more code than provided by the PR.
4. Pull Request has neither tests nor documentation. Sorry, the project cannot accept such the untested code.
5. The PR doesn't handle situation when a child is crashed for some reason. Eventually, it leads to the situation when only the main process is alive but all workers are gone. This is solvable, for example, gunicorn or supervisord handles it properly. But again, should we maintain all these things?

I suggest extracting this pull request into a separate project hosted outside of aio-libs and aiohttp.

maestroprog:
@asvetlov, thanks for the detailed consideration of this PR

> 1. For me, the clustered mode is something about running a server on multiple machines in a cluster. This PR is about the multi-processed mode, not about the clustering.
> 2. As @socketpair mentioned there are too many ways to run multiple processes for the server. You invented another one, that's fine. Should the aiohttp team accept it into the code base and provide the feature maintenance -- this is the question. I personally don't want to get yet another maintenance burden.
> 3. The multi-process mode is not a drop-in replacement for single-process one, server processes have no shared memory and need a sort of communication channel for synchronizing its work. This is definitely possible but requires much more code than provided by the PR.

The idea of this PR is that you just start the process and do not care about scaling the application horizontally within the same server by the number of cores. This is the launch of the application in a kind of "cluster" mode, but involves launching the application on only one server. For example, Docker or virtualization systems solve the problem of clustering at another level - launching several application instances. It is also very simple to configure a reverse proxy, instead of the solution that is offered in the official documentation https://docs.aiohttp.org/en/stable/deployment.html#nginx-supervisord. This is an **alternative** for using **gunicorn**, which works in exactly the same way, but only on the WSGI protocol. I even took the code from your worker as a basis, and rewrite it...

> 4. Pull Request has neither tests nor documentation. Sorry, the project cannot accept such the untested code.

I apologize, I was not sure that this feature would be accepted by the community as worthy of attention, so I wrote below in the first comment

> if these changes are useful for aiohttp, I will write unit tests and add documentation :)

5. > The PR doesn't handle situation when a child is crashed for some reason. Eventually, it leads to the situation when only the main process is alive but all workers are gone. This is solvable, for example, gunicorn or supervisord handles it properly. But again, should we maintain all these things?

If I correctly understood the remark, I dare say that this is not true. Here https://github.com/aio-libs/aiohttp/pull/4505/commits/48cb607fd9550423308d07fb0444257f137aa817#diff-202a3e0c7b116a8db3ab62171460a308R153 it is used to execute them again, if does its work, them makes up for. The supervisor is needed only to control the operation of the main process, and this, by definition, is not the concern of the library developers.
And yes, of course, this requires very good testing and refinement of the code. I would be ready to do it.

I agree that this change is not entirely suitable for this library. On this PR I close.

asvetlov:
Ok, sorry. The PR seems to handle a child crash correctly.
Anyway, I would not maintain it, please extract the code into a separate project if you want to drive it. 
Please feel free to make *documentation only* pull request for deployment instructions [here](http://docs.aiohttp.org/en/stable/deployment.html) to make your more visible to aiohttp developers.

achimnol:
I'm personally maintaining a separate project named [aiotools](https://github.com/achimnol/aiotools) where its `server` module provides a multi-process execution environment for asyncio applications (not just only aiohttp). Please have a look.

Running a multi-process version of asyncio applications is not a simple task and sometimes require a lot of application-specific stuffs, so I am also skeptical to that aiohttp's scope should include such cases.

Issue #4506, opened 2020-01-16
When compressing responses with deflate encoding, output doesn't include Zlib headers.


PaulJuliusMartinez:
## Long story short

The `wbits` argument to `zlib.compressobj` in [this method](https://github.com/aio-libs/aiohttp/blob/7a324fd46ff7dc9bb0bb1bc5afb326e04cf7cef0/aiohttp/http_writer.py#L56-L59) should be `zlib.MAX_WBITS` when using the `deflate` encoding, rather than `-zlib.MAX_WBITS`.

## Expected behaviour

The `deflate` encoding as specified in [RFC2616](https://tools.ietf.org/html/rfc2616#section-3.5) states the following:

```
   deflate
        The "zlib" format defined in RFC 1950 [31] in combination with
        the "deflate" compression mechanism described in RFC 1951 [29].
```

The [zlib website](https://zlib.net/zlib_faq.html#faq39) attempts to clear up the confusion:

> _What's the difference between the "gzip" and "deflate" HTTP 1.1 encodings?_
> "gzip" is the gzip format, and "deflate" is the zlib format. They should probably have called the second one "zlib" instead to avoid confusion with the raw deflate compressed data format. While the HTTP 1.1 RFC 2616 correctly points to the zlib specification in RFC 1950 for the "deflate" transfer encoding, there have been reports of servers and browsers that incorrectly produce or expect raw deflate data per the deflate specification in RFC 1951, most notably Microsoft. So even though the "deflate" transfer encoding using the zlib format would be the more efficient approach (and in fact exactly what the zlib format was designed for), using the "gzip" transfer encoding is probably more reliable due to an unfortunate choice of name on the part of the HTTP 1.1 authors.

[RFC7230](https://httpwg.org/specs/rfc7230.html#deflate.coding) additionally notes:

> *Note:* Some non-conformant implementations send the "deflate" compressed data without the zlib wrapper.

The [documentation](https://docs.python.org/3/library/zlib.html#zlib.compressobj) for Python zlib library explains how to produce different outputs:

> `zlib.compressobj(level=-1, method=DEFLATED, wbits=MAX_WBITS, memLevel=DEF_MEM_LEVEL, strategy=Z_DEFAULT_STRATEGY[, zdict])`
>
> ...
> 
> The wbits argument controls the size of the history buffer (or the “window size”) used when compressing data, and whether a header and trailer is included in the output. It can take several ranges of values, defaulting to 15 (MAX_WBITS):
> 
[...]

PR #4511, opened 2020-01-17
Fix deflate compression (#4506)

socketpair:
## Checklist

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -727,30 +727,36 @@ class DeflateBuffer:
             self.decompressor = BrotliDecoder()  # type: Any
         else:
             zlib_mode = (16 + zlib.MAX_WBITS
-                         if encoding == 'gzip' else -zlib.MAX_WBITS)
+                         if encoding == 'gzip' else zlib.MAX_WBITS)
             self.decompressor = zlib.decompressobj(wbits=zlib_mode)
 
     def set_exception(self, exc: BaseException) -> None:
         self.out.set_exception(exc)
 
     def feed_data(self, chunk: bytes, size: int) -> None:
+        if not size:
+            return
+
         self.size += size
+
+        # RFC1950
+        # bits 0..3 = CM = 0b1000 = 8 = "deflate"
+        # bits 4..7 = CINFO = 1..7 = windows size.
+        if not self._started_decoding and self.encoding == 'deflate' \
+                and chunk[0] & 0xf != 8:
+            # Change the decoder to decompress incorrectly compressed data
+            # Actually we should issue a warning about non-RFC-compilant data.
+            self.decompressor = zlib.decompressobj(wbits=-zlib.MAX_WBITS)
+
         try:
             chunk = self.decompressor.decompress(chunk)
         except Exception:
-            if not self._started_decoding and self.encoding == 'deflate':
-                self.decompressor = zlib.decompressobj()
-                try:
-                    chunk = self.decompressor.decompress(chunk)
-                except Exception:
-                    raise ContentEncodingError(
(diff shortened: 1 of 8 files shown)


socketpair:
I'm not sure I've done it right.

Websockets must use deflate without headers according to RFC.

I did not find anything about multipart....so I did not fix it.

socketpair:
tests failed on Cython on MAC:

image

asvetlov:
Please update against the master.
I've disabled the failed check, the buggy test was very weak anyway.

aiohttp/http_parser.py
```diff
@@ -727,30 +727,35 @@ def flush(self) -> bytes:
...
     def feed_data(self, chunk: bytes, size: int) -> None:
+        if not size:
+            return
+
         self.size += size
+
+        # RFC1950
+        # bits 0..3 = CM = 0b1000 = 8 = "deflate"
+        # bits 4..7 = CINFO = 1..7 = windows size.
+        if not self._started_decoding and self.encoding == 'deflate' and chunk[0] & 0xf != 8:
```
socketpair: on very small chunks it's not enough to trigger error in the first call to decompressor. I wrote the test on it.

So, I changed the code to check the first byte of the first chunk.

aiohttp/http_parser.py
```diff
@@ -727,30 +727,35 @@ def flush(self) -> bytes:
...
-            else:
-                raise ContentEncodingError(
-                    'Can not decode content-encoding: %s' % self.encoding)
+            raise ContentEncodingError(
+                'Can not decode content-encoding: %s' % self.encoding)
+
+        self._started_decoding = True
 
         if chunk:
-            self._started_decoding = True
```
socketpair, line 753: This was a bug. this value must be set as soon as we passed something to a decoder. Small data passed as a compressed chunk might not yeld decompressed chunk.

aiohttp/http_parser.py
socketpair: I did not check, but this code might be slightly slower. Actually, such a case can happen rarely, since I can't imagine a real situation where compressed chunk (went out of compressor) might be split in such a weird way.

socketpair:
`aiohttp/http_parser.py:746:13: E125 continuation line with same indent as next logical line`

I'm unable to fix it. It seems I tried everything.

@asvetlov please review and/or add reviewers

asvetlov:
Got you, I'll pick up the PR

aiohttp/http_parser.py
```diff
@@ -727,30 +727,36 @@ def flush(self) -> bytes:
...
     def feed_data(self, chunk: bytes, size: int) -> None:
+        if not size:
+            return
+
         self.size += size
+
+        # RFC1950
+        # bits 0..3 = CM = 0b1000 = 8 = "deflate"
+        # bits 4..7 = CINFO = 1..7 = windows size.
+        if not self._started_decoding and self.encoding == 'deflate' \
```
socketpair, line 745: @asvetlov should we decode non-RFC compliant data? I would remove such support. But unfortunately, this will break communication with deflate-enabled aiohttp server.

asvetlov: Let's keep it for a while. I have a feeling that we need to support this mode for three years at least.

asvetlov: `urllib3` used by `requests` has a similar workaround: https://github.com/urllib3/urllib3/blob/9971e27e83a891ba7b832fa9e5d2f04bbcb1e65f/src/urllib3/response.py#L55-L61

aiohttp/http_parser.py
asvetlov: Ok

PR #4513, opened 2020-01-18
Fix python 3.8 tests under Windows

asvetlov:
Co-Authored-By: hh-h


asvetlov:
This PR is almost a full copy of #4463 but `selector_loop` moved from public API to private `conftest.py` location.
Plus, tests on Python 3.8.1 are fixed.

Still need to use `loop.sendfile()` and `loop.start_tls()` if available to fully support Windows.

Thanks @hh-h for the initial PR

PR #4517, opened 2020-01-19
Use loop.sendfile() instead of custom implementation if available

### LINKED ISSUE #4269: Use loop.sendfile() for file sending on Python 3.7+

asvetlov:
Fix #4269

### DIFF
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -96,23 +96,24 @@ class SendfileStreamWriter(StreamWriter):
     async def sendfile(self) -> None:
         assert self.transport is not None
         loop = self.loop
-        ### if hasattr(loop, "sendfile"):
-        ###     # Python 3.7+
-        ###     breakpoint()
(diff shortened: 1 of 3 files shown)


aiohttp/web_fileresponse.py
```diff
@@ -94,12 +95,25 @@ def _done_fut(self, out_fd: int, fut: 'asyncio.Future[None]') -> None:
 
     async def sendfile(self) -> None:
         assert self.transport is not None
+        loop = self.loop
+        data = b''.join(self._sendfile_buffer)
```
webknjaz, line 99: Would the use of `memoryview()` be faster in this case?

asvetlov: I doubt that the speed difference is visible at all.
You can measure it if you want.

Issue #4258, opened 2019-10-23
Lost/hanging requests with 3.6.2


pvizeli:
## Long story short

Based on: https://github.com/home-assistant/hassio/pull/1342

We run into an issue that requests produce timeouts and were never logged by our middleware they protocol/monitor requests.  All instances with this issue running with an nginx proxy manager in front. Do we modify the URL with a parameter, it worked. It is very random. The only way to fix this issue was to downgrade aiohttp.

We can see the request is going to instance but we can't see it on the application layer. Maybe interesting ist that all the requests use cookies for session handling. 

## Expected behaviour

Everything work like 3.6.1

## Actual behaviour

Random connections run into a timeouts.

## Steps to reproduce

Run an nginx proxymanager in front of aiohttp 3.6.2 and use a site with cookies.

## Your environment

Python 3.7 with uvloop and without. We use aioHTTP as server application in this case.

webknjaz:
> without uvloop doesn't have this issue

did you report this to uvloop?

pvizeli:
EDIT: It look like without uvloop it doesn't have this issue.

I would do this after someone can point me to the part of code they changed on 3.6.2 and can affect this. Maybe @1st1 can assist me later.

EDIT: we had now the issue without uvloop @webknjaz. It is very random. So with 30 requests, 1-2 requests hangs.

webknjaz:
@pvizeli any chance you could `git bisect` this? Also, please post the log of `pip install`, we need to know if you use C-extensions or not and if you're installing from wheels or building stuff during install time.

webknjaz:
Source comparison: https://github.com/aio-libs/aiohttp/compare/v3.6.1...v3.6.2#files_bucket

webknjaz:
Changelog: https://docs.aiohttp.org/en/stable/changes.html#id1

webknjaz:
Looking at https://pypi.org/project/aiohttp/3.6.1/#files vs. https://pypi.org/project/aiohttp/3.6.2/#files: you might've hit a wheel with `py3-none-any` marker which would mean that you're running a pure-python version so `pip install` is really important here. 
Until you post this info, we have no idea what you're really running and consequently how to reproduce it.

pvizeli:
We build our own wheel (on Alpine) and we use the C extension. I can get a freeze from that docker if you want? The affected container is `homeassistant/amd64-hassio-supervisor:190`

pvizeli:
[...]

PR #4562, opened 2020-02-09
Fix a problem with connection waiters that are never awaited

illia-v:
## What do these changes do?
When a connector waits for multiple available connections with the same key, the described race condition can happen.

Let's say that we want to create 3 connections but there a no available ones due to a limit. Then:
1. The first future [is added](https://github.com/aio-libs/aiohttp/blob/ea8f35e80539931e864922c2214bc4a3ec7d9cf4/aiohttp/connector.py#L465), `BaseConnector.connect` [sets](https://github.com/aio-libs/aiohttp/blob/ea8f35e80539931e864922c2214bc4a3ec7d9cf4/aiohttp/connector.py#L464) a deque 1 containing the future to a local `waiters` variable, the future is awaited, the context is switched. 
2. The second future is added, `BaseConnector.connect` sets the deque 1 containing first and second futures to `waiters`, the second future is awaited, the context is switched.
3. One of the previous connections is closed, the first future [is popped](https://github.com/aio-libs/aiohttp/blob/ea8f35e80539931e864922c2214bc4a3ec7d9cf4/aiohttp/connector.py#L578) from the deque 1, its result [is set](https://github.com/aio-libs/aiohttp/blob/ea8f35e80539931e864922c2214bc4a3ec7d9cf4/aiohttp/connector.py#L580), the context is switched.
4. Another previous connection is closed, the second future is popped from the deque 1 (the deque is empty), its result is set, the context is switched.
5. We returned to the context from the first step, the deque 1 [is deleted](https://github.com/aio-libs/aiohttp/blob/ea8f35e80539931e864922c2214bc4a3ec7d9cf4/aiohttp/connector.py#L485) from `BaseConnector._waiters`, ..., the context is switched.
6. The third future is added, `BaseConnector.connect` sets a new deque 2 containing the third future to `waiters`, the third future is awaited, the context is switched.
7. Asyncio returned to the context from the second step, it checks that the local `waiters` variable containing the deque 1 is empty, and deletes `self._waiters[key]`. But `self._waiters[key]` is the deque 2 at the moment. Therefore, the waiter represented as the third future will be never released. And the program will hang if no `connect` timeout is set.

No

## Related issue number
Maybe #4258 and https://github.com/aio-libs/aiobotocore/issues/738

## Checklist


illia-v:
```python
In [1]: from collections import defaultdict, deque                                                                                                                                                                                                                                                        

In [2]: dictionary = defaultdict(deque)                                                                                                                                                                                                                                                                   

In [3]: dq = dictionary['key']                                                                                                                                                                                                                                                                            

In [4]: dq.append('future')                                                                                                                                                                                                                                                                               

In [5]: dq is dictionary['key']                                                                                                                                                                                                                                                                           
Out[5]: True

In [6]: del dictionary['key']                                                                                                                                                                                                                                                                             

In [7]: dq                                                                                                                                                                                                                                                                                                
Out[7]: deque(['future'])

In [8]: dictionary                                                                                                                                                                                                                                                                                        
Out[8]: defaultdict(collections.deque, {})

In [9]: dictionary['key'].append('future')                                                                                                                                                                                                                                                                

In [10]: dictionary                                                                                                                                                                                                                                                                                       
Out[10]: defaultdict(collections.deque, {'key': deque(['future'])})

In [11]: dq is dictionary['key']                                                                                                                                                                                                                                                                          
Out[11]: False
```

thehesiod:
this logic is getting impossible to reason with, how about something like this: https://github.com/thehesiod/aiohttp/pull/4  The issue I have right now is that `_release_waiter` is not async and doesn't appear to be easy to make async

thehesiod:
actually I think I came up with a fix in my PR

illia-v:
@thehesiod the race condition described in the first comment caused the https://github.com/aio-libs/aiobotocore/issues/738#issuecomment-581441776 problem. When I was debugging it, it was pretty clear that a different deque with a same key was deleted from the dictionary. I checked twice, and therefore I don't agree that it is impossible :)

I'll take a look at your PR when I can

illia-v:
@thehesiod I looked at your pull request and added two minor comments.

Af for me your solution looks better than the current implementation. Also, it seems to fix the problem mentioned in this PR too, I haven't checked it though.

If your changes are merged before the next release, I'm OK with closing this PR.
Otherwise, I'd like this one to be merged before the next release to fix the problem because it changes only one line and should not create any regressions

thehesiod:
@illia-v I didn't open my PR against aiohttp, up to maintainers of aiohttp to see if they want that or not ( @asvetlov )

aiohttp/connector.py
```diff
@@ -480,7 +480,10 @@ async def connect(self, req: 'ClientRequest',
 
                 raise e
             finally:
-                if not waiters:
+                # `waiters` may be deleted from `self._waiters` by another
+                # coroutine, and `self._waiters[key]` may contain another
+                # deque, the later should not be deleted.
```
pfreixes: I guess that the issue is that we are caching the `self._waiters[key]` at the local level by using the local variable `waiters` what about simply doing a check with using the none cached version, like:

```python
if key in self._waiters and not self._waiters[key]:
    ...
```
Maybe it's more clear.

WDYT?

illia-v: It's a good idea. Thanks!

asvetlov (approved):

asvetlov:
LGTM, thanks!

thehesiod:
@asvetlov welcome back :)

asvetlov:
Thanks, @thehesiod

Issue #4572, opened 2020-02-13
Server closes connection if wrong type returned by handler


Fogapod:
🐞 **Describe the bug**

Server closes session if handler returns anything else that `Response` instance.

💡 **To Reproduce**

```python
from aiohttp import web


async def handler(req):
    return 1

app = web.Application()
app.add_routes([web.get("/", handler)])
web.run_app(app)
```

```bash
~ curl localhost:8080
curl: (52) Empty reply from server
```

💡 **Expected behavior**

Error message in console and returned status 500.

📋 **Your version of the Python**

```console
$ python --version
Python 3.8.1
```

📋 **Your version of the aiohttp/yarl/multidict distributions**

Tested with aiohttp `3.6.2` and `master` branch fresh installations.

webknjaz:
@Fogapod thanks for the report! It'd be cool if you could submit a failing test marked with xfail in a PR as per https://pganssle-talks.github.io/xfail-lightning.

Fogapod:
@webknjaz alright, I'll try to make a test for this issue.

I found that returning `HTTPException` instance in aiohttp 3.6.2 resulted in returning corresponding status code by server, in master branch it results in closed connection.
Is returning exceptions deprecated? I didn't find it in documentation, there is no traceback in console.

webknjaz:
I think it is, or should be. Exceptions must be raised.

Fogapod:
`RuntimeError` is raised here https://github.com/aio-libs/aiohttp/blob/f80029b553ec041370babf75d07e2abb9b2bdca3/aiohttp/web_protocol.py#L577-L579
after calling https://github.com/aio-libs/aiohttp/blob/f80029b553ec041370babf75d07e2abb9b2bdca3/aiohttp/web_protocol.py#L480
and catched here https://github.com/aio-libs/aiohttp/blob/f80029b553ec041370babf75d07e2abb9b2bdca3/aiohttp/web_protocol.py#L523-L526
For some reason these are only logged in asyncio debug mode.

I'm not sure how this could be fixed. Returning status 500 for all Runtime errors?

webknjaz:
> Returning status 500 for all Runtime errors?

I think so

webknjaz:
> Is returning exceptions deprecated? I didn't find it in documentation, there is no traceback in console.

v3.6 deprecates returning exceptions but it still works (see the red warning box): https://docs.aiohttp.org/en/stable/web_quickstart.html#exceptions

[...]

PR #4574, opened 2020-02-13
Add an xfailing test expecting HTTP500 for invalid handler return values

Fogapod:
This change documents a contract of disallowing non-response
return values in request handlers that currently has a bug and
is therefore marked as xfail.

Issue: #4572

Created test as @webknjaz suggested.

### DIFF
--- a/tests/test_web_request.py
+++ b/tests/test_web_request.py
@@ -10,6 +10,7 @@ from multidict import CIMultiDict, CIMultiDictProxy, MultiDict
 from yarl import URL
 
 from aiohttp import HttpVersion, web
+from aiohttp.client_exceptions import ServerDisconnectedError
 from aiohttp.helpers import DEBUG
 from aiohttp.http_parser import RawRequestMessage
 from aiohttp.streams import StreamReader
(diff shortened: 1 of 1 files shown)


Fogapod:
Should test for returning `HTTPException` also be included? I'm not sure how it should be checked because it doesn't work at all currently.

tests/test_web_request.py
```diff
@@ -771,3 +772,19 @@ def test_eq() -> None:
...
+async def test_handler_return_type(aiohttp_client) -> None:
+    async def invalid_handler_1(request):
+        return 1
+
+    app = web.Application()
+    app.router.add_get('/1', invalid_handler_1)
+    client = await aiohttp_client(app)
+
+    async with client.get('/1') as resp:
+        assert 500 == resp.status
```
webknjaz, line 790: I'd also check if the text includes something like "Internal Server Error".

webknjaz:
> Should test for returning `HTTPException` also be included?

Feel free to add a separate PR.

v3.6 deprecates returning exceptions but it still works (see the red warning box): https://docs.aiohttp.org/en/stable/web_quickstart.html#exceptions

master/v4.0 doesn't support this anymore and doesn't mention returning exceptions: https://docs.aiohttp.org/en/latest/web_exceptions.html


Based on this, when returning errors under 4.0/master, check for HTTP 500 Internal Server Error.

PR #4646, opened 2020-03-18
Fix multiple calls to runner shutdown with multiple sites

achimnol:
## What do these changes do?

Move invocations to runner shutdown handlers from BaseSite to BaseRunner to avoid multiple (duplicate) shutdowns when there are multiple sites registered to a single runner and application.

Currently startup signal handlers and cleanup contexts are called only once regardless of the number
of registered sites, but the shutdown signal handlers are called multiple times.

`shutdown_timeout` argument is moved from `BaseSite` and all its derivatives to `BaseRunner`.
There are no changes in the high-level API such as `web.run_app()` and the gunicorn worker interface, but if users rely on the low-level APIs they need to move the `shutdown_timeout` argument from sites to runners.

## Example

```python3
import asyncio
from typing import AsyncIterator

from aiohttp import web


async def handle_hello(request: web.Request) -> web.Response:
    return web.Response(text='hello world')


async def my_cleanup_ctx(app: web.Application) -> AsyncIterator[None]:
    print('my_cleanup_ctx: init')
    yield
    print('my_cleanup_ctx: cleanup')


async def my_startup_func(app: web.Application) -> None:
    print('my_startup_func')


async def my_shutdown_func(app: web.Application) -> None:
    print('my_shutdown_func')


async def main() -> None:
    app = web.Application()
    app.add_routes([
        web.get('/', handle_hello)
    ])
    app.cleanup_ctx.append(my_cleanup_ctx)
    app.on_startup.append(my_startup_func)
    app.on_shutdown.append(my_shutdown_func)
    runner = web.AppRunner(app)
    await runner.setup()
    sites = [
        web.TCPSite(runner, 'localhost', 8080),
        web.TCPSite(runner, 'localhost', 8180),
    ]
    print('Starting...')
    for site in sites:
        await site.start()
    try:
        print('Started.')
        while True:
            await asyncio.sleep(1)
    finally:
        print('Terminating...')
        await runner.cleanup()


if __name__ == '__main__':
    try:
        asyncio.run(main())
    except KeyboardInterrupt:
        print('Interrupted!')
```

Before applying this PR (env: Python 3.8.2 installed using pyenv on Ubuntu 18.04):
```
my_cleanup_ctx: init
my_startup_func
Starting...
Started.
^CTerminating...
my_shutdown_func
my_shutdown_func
my_cleanup_ctx: cleanup
Interrupted!
```
Note that `my_shutdown_func` is called ***twice***.

After applying this PR:
```
my_cleanup_ctx: init
my_startup_func
Starting...
Started.
^CTerminating...
my_shutdown_func
my_cleanup_ctx: cleanup
Interrupted!
```

## Related issue number

I have mentioned this issue in the gitter chat.

## Checklist

### DIFF
--- a/CHANGES/4646.bugfix
+++ b/CHANGES/4646.bugfix
@@ -1,0 +1,1 @@
+Fix duplicate calls to shutdown signal handlers when there are multiple sites registered to the app.
--- a/CHANGES/4646.removal
+++ b/CHANGES/4646.removal
@@ -1,0 +1,1 @@
+Move shutdown_timeout arguments from BaseSite to BaseRunner
--- a/CHANGES/4646.removal
+++ b/CHANGES/4646.removal
@@ -1,1 +1,1 @@
-Move shutdown_timeout arguments from BaseSite to BaseRunner
+Move shutdown_timeout argument from BaseSite to BaseRunner
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -308,6 +308,7 @@ async def _run_app(app: Union[Application, Awaitable[Application]], *,
     app = cast(Application, app)
 
     runner = AppRunner(app, handle_signals=handle_signals,
+                       shutdown_timeout=shutdown_timeout,
                        access_log_class=access_log_class,
                        access_log_format=access_log_format,
                        access_log=access_log)
@@ -320,7 +321,6 @@ async def _run_app(app: Union[Application, Awaitable[Application]], *,
         if host is not None:
             if isinstance(host, (str, bytes, bytearray, memoryview)):
                 sites.append(TCPSite(runner, host, port,
-                                     shutdown_timeout=shutdown_timeout,
                                      ssl_context=ssl_context,
                                      backlog=backlog,
                                      reuse_address=reuse_address,
@@ -328,14 +328,12 @@ async def _run_app(app: Union[Application, Awaitable[Application]], *,
             else:
                 for h in host:
                     sites.append(TCPSite(runner, h, port,
-                                         shutdown_timeout=shutdown_timeout,
                                          ssl_context=ssl_context,
                                          backlog=backlog,
                                          reuse_address=reuse_address,
                                          reuse_port=reuse_port))
         elif path is None and sock is None or port is not None:
             sites.append(TCPSite(runner, port=port,
-                                 shutdown_timeout=shutdown_timeout,
                                  ssl_context=ssl_context, backlog=backlog,
                                  reuse_address=reuse_address,
                                  reuse_port=reuse_port))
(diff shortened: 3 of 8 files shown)


achimnol:
The CI failure of the final commit seems to be a timeout issue in the CI server. (The failed test has nothing to do with this PR)

asvetlov (commented):
The multiple shutdowns were the initial intention.
Technically, if you start the application on multiple sites -- you can shut them down independently.

I think I understand your wish to shutdown everything on the application closing. This desire makes sense but it should be implemented as a different signal (the name is the subject for discussion).

Please feel free to prepare a PR for this not-existing-yet signal support.

asvetlov (changes requested):
Sorry, I cannot accept the PR as-is, see the comment above.

achimnol:
Okay, I understand that the intention was to run shutdown handlers per site for a single app.
Then, shouldn't do the framework run the startup handlers (and maybe cleanup contexts) per site in the same way?
If so, since currently the app is unaware of site (as it's an abstraction), we may need to expose the site to the app codes to distinguish its running context.
Otherwise, we need to explicitly state which one (among cleanup contexts, startup handlers, shutdown handlers) runs once and multiple times per app when a single app instance is reused over multiple sites.
Maybe as you suggested, we could split "app-startup" and "site-startup" / "app-shutdown" and "site-shutdown" signals, and add another type of "site-cleanup context" whose invocation is bound to each site, in addition to "app-cleanup context".
Maybe again, we may need to attach signal handlers to sites, instead of apps.

The main purpose of this PR was to *clarify and make it consistent* on how many times startup/shutdown handlers are called in what context.
Please tell me what you think.

asvetlov:
In my mind current signals are:
startup [app]
cleanup_ctx [app]
shundown [site]
cleanup [app]

app signals are called once per application, site signal is called for every site.
I agree that the list of existing signals is not complete and can be extended.
Not sure that we need the full possible matrix but I'm okay to add new signals if somebody needs them.

achimnol:
Hm.. If I understand your description & intentions correctly, do the followings match with your thoughts?

 * startup: initialize/allocate the underlying resources such as DB connection pools => per-app
 * shutdown: cancel ongoing request handlers and tasks when interrupted => per-site (because request-response connections are tied with site-specific transports)
 * cleanup: release/destroy the underlying resources such as DB connection pools => per-app

If so, I think [the docs](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.Application.on_shutdown) should be clarified that "shutdown" is executed per site, with possibility of multiple executions if there are multiple sites running the same app instance. Currently it's just saying "A Signal that is fired on application shutdown.", which leads to a misunderstanding that it's just yet-another version of the cleanup signal. It would be even better to mention that the term "shutdown" has come from [`socket.shutdown()`](https://docs.python.org/3/library/socket.html#socket.socket.shutdown) which actively initiates the connection close and the aiohttp's shutdown also specifically means clearing up ongoing connections instead of resource "shutdowns".

If you are okay, I'll rewrite this PR to update the docs, instead of patching the current behavior, though I need to refactor my codes to conform with it...

achimnol:
Also, as the doc suggests: "We suggest keeping a list of long running handlers in Application dictionary.", more work are required.

For multi-site apps, we should be able to distinguish which long-running handlers belong to which site, because the shutdown handlers will be called multiple times but we have no contextual information about which site that current signal invocation is for.

To this end, I suggest to do:
* Add a reference to the site to the `BaseRequest` object so that long-running handlers can keep the reference to themselves in the app's per-site handler set.
* Pass the site object to the shutdown signal handlers so that they can know the context.
* Update [the graceful shutdown example in the docs](https://docs.aiohttp.org/en/stable/web_advanced.html#aiohttp-web-graceful-shutdown) to reflect above changes, e.g., using the site object passed to the shutdown signal handler to distinguish the set of long-running handlers to cancel.

If we implement the above changes, I think we don't need to add yet another signal.

asvetlov:
> Add a reference to the site to the BaseRequest object so that long-running handlers can keep the reference to themselves in the app's per-site handler set.
Ok

> Pass the site object to the shutdown signal handlers so that they can know the context.
Sorry, it is a breaking change. We cannot mutate the signal handler signature.
But we can add a new signal for the *fixed* shutdown and deprecate the existing one.

> Update the graceful shutdown example in the docs to reflect above changes, e.g., using the site object passed to the shutdown signal handler to distinguish the set of long-running handlers to cancel.
Ok

If you want to do this task -- you are welcome!

webknjaz:
@achimnol are you planning to resume working on this?

Dreamsorcerer:
I think I also fixed this in #7780, so I'll close this. Reopen if you think there's still something of value here.

> For multi-site apps, we should be able to distinguish which long-running handlers belong to which site, because the shutdown handlers will be called multiple times but we have no contextual information about which site that current signal invocation is for.

I think this is still relevant and something I also noticed during that work.

PR #4650, opened 2020-03-20
Implement binding to IPv6 addresses in the pytest server fixture.

tan01:
## What do these changes do?
This change allows TestServer to be instantiated with IPv6 hostnames.
Previously, the TestServer would open an IPv4 family socket regardless
of hostname, causing an error to be raised upon starting the server.
The BaseTestServer now parses the hostname and create an IPv4 or
IPv6 socket based on the hostname.

Creating a TestServer from test_utils.py with an IPv6 hostname ('::1') no longer throws
```
 socket.gaierror: [Errno -9] Address family for hostname not supported
```

## Related issue number
None

## Checklist

### DIFF
--- a/CHANGES/4650.bugfix
+++ b/CHANGES/4650.bugfix
@@ -1,0 +1,1 @@
+Modify test_utils.BaseTestServer to allow IPv6 hostnames.
--- a/CHANGES/4650.bugfix
+++ b/CHANGES/4650.bugfix
@@ -1,1 +1,1 @@
-Modify test_utils.BaseTestServer to allow IPv6 hostnames.
+Implement binding to IPv6 addresses in the pytest server fixture.
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -119,10 +119,10 @@ class BaseTestServer(ABC):
         absolute_host = self.host
         try:
             version = ipaddress.ip_address(self.host).version
-            if version == 6:
-                absolute_host = f"[{self.host}]"
         except ValueError:
             version = 4
+        if version == 6:
+            absolute_host = f"[{self.host}]"
         family = socket.AF_INET6 if version == 6 else socket.AF_INET
         _sock = get_port_socket(self.host, self.port, family=family)
         self.host, self.port = _sock.getsockname()[:2]
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -62,8 +62,10 @@ else:
 REUSE_ADDRESS = os.name == 'posix' and sys.platform != 'cygwin'
 
 
-def get_unused_port_socket(host: str) -> socket.socket:
-    return get_port_socket(host, 0)
+def get_unused_port_socket(
+        host: str,
+        family: socket.AddressFamily = socket.AF_INET) -> socket.socket:
+    return get_port_socket(host, 0, family)
 
 
 def get_port_socket(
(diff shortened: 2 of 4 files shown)


tests/test_test_utils.py
```diff
@@ -305,3 +305,11 @@ def test_noop(self) -> None:
...
 
     await client.close()
+
+
+@pytest.mark.parametrize("hostname", ["127.0.0.1", "localhost", "::1"])
+async def test_test_server_hostnames(hostname, loop) -> None:
+    app = _create_example_app()
+    server = _TestServer(app, host=hostname, loop=loop)
+    async with server:
+        pass
```
webknjaz, line 318: Maybe add some assertion?
```suggestion
        pass

    assert server.host == hostname
```

webknjaz:
Plz go through the checklist/add a change fragment.

tests/test_test_utils.py
tan01: Sounds good, added an assertion checking TestServer host against expected hostname.

CHANGES/4650.bugfix
```diff
@@ -0,0 +1 @@
+Modify test_utils.BaseTestServer to allow IPv6 hostnames.
```
webknjaz: I think that the user-facing message should be something valuable to them, not the implementation detail:
```suggestion
Implement binding to IPv6 addresses in the pytest server fixture.
```

WDYT?

aiohttp/test_utils.py
```diff
@@ -65,8 +66,11 @@ def get_unused_port_socket(host: str) -> socket.socket:
     return get_port_socket(host, 0)
```
webknjaz: Should this helper also support the family arg?

aiohttp/test_utils.py
```diff
@@ -110,7 +114,15 @@ def __init__(self,
...
         if not self.port:
             self.port = 0
-        _sock = get_port_socket(self.host, self.port)
+        absolute_host = self.host
+        try:
+            version = ipaddress.ip_address(self.host).version
+            if version == 6:
+                absolute_host = f"[{self.host}]"
+        except ValueError:
+            version = 4
```
webknjaz: try/except blocks must always wrap only one line where the exception is expected
```suggestion
        try:
            version = ipaddress.ip_address(self.host).version
        except ValueError:
            version = 4
        if version == 6:
            absolute_host = f"[{self.host}]"
```

tests/test_test_utils.py
webknjaz: FYI it's a good tone to apply suggested changes via GitHub UI since it automatically preserves the authorship by adding `Co-Authored-By:` commit message trailer. This can then be pulled and squashed manually if needed.

tests/test_test_utils.py
```diff
@@ -305,3 +305,15 @@ def test_noop(self) -> None:
     assert _hello_world_str == text
 
     await client.close()
+
+
+@pytest.mark.parametrize("hostname,expected_host",
```
webknjaz: Turns out you can separate params nicely like this:
```suggestion
@pytest.mark.parametrize(("hostname", "expected_host"),
```

CHANGES/4650.bugfix
tan01: Sounds good!

aiohttp/test_utils.py
tan01: Can't hurt. Added.

PR #4692, opened 2020-04-14
Fix bug in route name validation (#4691)

### LINKED ISSUE #4691: Cannot have python keyword in route name
🐞 **Describe the bug**

This code is rejected with 'Incorrect route name':
```python
@app_routes.get('/result/', name='check-for-result')
async def result(request):
```

This is because 'for' is a keyword, and the check rejects any keywords for some strange reason:
in https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_urldispatcher.py#L982
```python
parts = self.NAME_SPLIT_RE.split(name)
for part in parts:
      if not part.isidentifier() or keyword.iskeyword(part):
           raise ValueError
```
I think the intention was to have:
```python
if not (part.isidentifier() or keyword.iskeyword(part))
```

💡 **To Reproduce**

💡 **Expected behavior**

Should accept route names containing python keywords

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)

```

```console
$ python --version
Python 3.7.4
```

```console
$ python -m pip show aiohttp
Version: 3.6.2
```
```console
Version: 4.7.5
```
```console
$ python -m pip show yarl
Version: 1.3.0
```

📋 **Additional context**

gorogoroumaru:
fix #4691 

The register_resource function in aiohttp/web_urldispatcher.py had a bug in route name validation.
It validated each part of the splitted route name whether it's a python keyword.
I fixed the function to validate route name before splitting it and added a new error message.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,29 +1,12 @@
 # Ref: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
 blank_issues_enabled: false  # default: true
 contact_links:
-- name: 🔐 Security bug report 🔥
-  url: https://github.com/aio-libs/.github/security/policy
-  about: |
-    Please learn how to report security vulnerabilities here.
-
-    For all security related bugs, send an email
-    instead of using this issue tracker and you
-    will receive a prompt response.
-
-    For more information, see
-    https://github.com/aio-libs/.github/security/policy
 - name: 🤷💻🤦 StackOverflow
   url: https://stackoverflow.com/questions/tagged/aiohttp
   about: Please ask typical Q&A here
-- name: 💬 Gitter Chat
-  url: https://gitter.im/aio-libs/Lobby
-  about: Chat with devs and community
 - name: 💬 Discourse
   url: https://aio-libs.discourse.group/
   about: Please start usage discussions here
-- name: 💬 Mailing list
-  url: https://groups.google.com/forum/#!forum/aio-libs
-  about: Usage Q&A
-- name: 📝 Code of Conduct
-  url: https://github.com/aio-libs/.github/blob/master/CODE_OF_CONDUCT.md
-  about: ❤ Be nice to other members of the community. ☮ Behave.
+- name: 💬 Gitter Chat
+  url: https://gitter.im/aio-libs/Lobby
+  about: Chat with devs and community
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,12 +1,29 @@
 # Ref: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
 blank_issues_enabled: false  # default: true
 contact_links:
+- name: 🔐 Security bug report 🔥
+  url: https://github.com/aio-libs/.github/security/policy
+  about: |
+    Please learn how to report security vulnerabilities here.
+
+    For all security related bugs, send an email
+    instead of using this issue tracker and you
+    will receive a prompt response.
+
+    For more information, see
+    https://github.com/aio-libs/.github/security/policy
 - name: 🤷💻🤦 StackOverflow
   url: https://stackoverflow.com/questions/tagged/aiohttp
   about: Please ask typical Q&A here
-- name: 💬 Discourse
-  url: https://aio-libs.discourse.group/
-  about: Please start usage discussions here
 - name: 💬 Gitter Chat
   url: https://gitter.im/aio-libs/Lobby
   about: Chat with devs and community
+- name: 💬 Discourse
+  url: https://aio-libs.discourse.group/
+  about: Please start usage discussions here
+- name: 💬 Mailing list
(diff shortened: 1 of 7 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -977,9 +977,13 @@ def register_resource(self, resource: AbstractResource) -> None:
         name = resource.name
 
         if name is not None:
+            if keyword.iskeyword(name):
+                raise ValueError('Incorrect route name {!r}, '
+                                 'python keywords cannot be used '
+                                 'for route name'.format(name))
```
webknjaz: FYI this could be an f-string
```suggestion
                raise ValueError(f'Incorrect route name {name!r}, '
                                 'python keywords cannot be used '
                                 'for route name')
```

aiohttp/web_urldispatcher.py
```diff
@@ -977,9 +977,13 @@ def register_resource(self, resource: AbstractResource) -> None:
         name = resource.name
 
         if name is not None:
+            if keyword.iskeyword(name):
```
webknjaz: Should we trim whitespaces first?
```suggestion
            if keyword.iskeyword(name.strip()):
```

gorogoroumaru:
Thanks!
I applied your suggested change.

webknjaz:
Great, waiting for the test, then!

aiohttp/web_urldispatcher.py
hh-h: ~~Python 3.5 doesn't support f-strings and as far as I know, aiohttp supports python 3.5.~~

Nvm, got confused by README

gorogoroumaru:
Test completed!

tests/test_web_urldispatcher.py
```diff
@@ -407,3 +408,14 @@ async def test_static_absolute_url(aiohttp_client, tmp_path) -> None:
...
+def test_register_resource(aiohttp_client) -> None:
+    router = UrlDispatcher()
+    path = "/result/"
+    name = "check-for-result"
+    url = URL.build(path=path)
+    resource = PlainResource(url.raw_path, name=name)
+    try:
+        router.register_resource(resource)
+    except ValueError:
+        pytest.fail()
```
webknjaz: It's a bad idea to shadow an actual error. the pytest-way is to use
```suggestion
    with pytest.raises(
            ValueError,
            match="^Incorrect route name 'check-for-result', "
            'python keywords cannot be used for route name$',
    ):
        router.register_resource(resource)
```

webknjaz: Also, this test doesn't cover negative cases and corner cases. They should be included and the tests should contain both cases with wrong names and correct names. Wrong names should contain cases from both categories — per each if-block.

gorogoroumaru: I want to test that the route name "check-for-result" doesn't raise an error, but this way tests that the route name "check-for-result" raises an error.

webknjaz: Ah, so in this case just make the call and assert something. Doing mumbo-jumbo with try/except in tests is unidiomatic and should be avoided. The point of the test is to show that something works, not try to handle any errors that may occur — those should be reported by pytest explicitly.

gorogoroumaru: Thanks!
I finished fixing.

webknjaz: You should've just clicked `Commit suggestion` on the GitHub UI in that comment above. This preserves authorship.

tests/test_web_urldispatcher.py
```diff
@@ -407,3 +408,19 @@ async def test_static_absolute_url(aiohttp_client, tmp_path) -> None:
...
+
+    correct_name = "check-for-result"
+    resource = PlainResource(url.raw_path, name=correct_name)
+    router.register_resource(resource)
+
+    assert resource == router._named_resources[correct_name]
+
+    wrong_name = "for"
+    resource = PlainResource(url.raw_path, name=wrong_name)
+    with pytest.raises(ValueError):
```
webknjaz: You have to use `match=` arg as show @ https://github.com/aio-libs/aiohttp/pull/4692#discussion_r411475905

tests/test_web_urldispatcher.py
```diff
@@ -407,3 +408,19 @@ async def test_static_absolute_url(aiohttp_client, tmp_path) -> None:
...
+        router.register_resource(resource)
```
webknjaz: This part is a separate test and so it must be in a separate test function.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -1,29 +1,12 @@
 # Ref: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser
 blank_issues_enabled: false  # default: true
 contact_links:
-- name: 🔐 Security bug report 🔥
```
webknjaz, line 4: Things like this mustn't be a part of the PR.

gorogoroumaru: Sorry, I  wrongly operated git command and included someone else's commit to my PR.

gorogoroumaru:
I finished rebasing PR and applying suggested changes.

asvetlov:
Thanks for the PR.
I've ported your changes to a little bit more correct #5051

PR #4706, opened 2020-04-22
Introduce fixture for test client class redefinition

hoefling:
Signed-off-by: oleg.hoefling <oleg.hoefling@gmail.com>

## What do these changes do?

A new fixture named `aiohttp_client_cls` is added that serves as `TestClient` factory for the `aiohttp_client` fixture.

When writing tests against `aiohttp` apps, we often need to extend the `TestClient` with methods that hide the low-level boilerplate, e.g.

```python
class GraphqlClient(TestClient):
    async def query(self, **kwargs):
        response = await self.post(
            '/graphql', json=kwargs, headers=dict(content_type='application/json')
        )
        return response.json()
```

To replace the `TestClient` with the custom client impl, we have to copy the `aiohttp_client` fixture out as there's no safe way to inject the `GraphqlClient` class.

The changes in this PR offer a possibility to pass a custom test client impl. Example:

```python
# conftest.py

@pytest.fixture
def aiohttp_client_cls():
    return GraphqlClient

# test_module.py

def test_query(aiohttp_client):
    app: Application = app_factory()
    payload = await client.query(
        query='query Greet($who: String!) {greet($who) { message }}',
        variables={'who': 'world'}
    )
    assert 'errors' not in payload
    assert payload['data']['message'] == 'hello world'
```

A more common example could be providing a `login()`/`logout()` method pair, similar to Django/Flask's test clients.

Nothing will change for existing test suites.

## Checklist

- [x] Documentation reflects the changes - added an example in fixture docstring and a reference in sphinx docs


tmcgroul:
I think is better solution to have keyword argument in `aiohttp_client`.
```
client = await aiohttp_client(app, test_client_class=GraphqlClient)
```

In my application i need to use several `TestClass` implementation.

webknjaz:
I like it, go ahead and add tests.

hoefling:
@webknjaz done - the default case is already covered by the existing suite, added tests for custom client class passed. Also extended docs with another example on how to pass test specific client impls, should anyone need it.

PR #4711, opened 2020-04-25
Implement `aiohttp.ClientResponse.ok` property

EvanKepner:
## What do these changes do?

This adds `ClientResponse.ok` as a property, similar to `requests.Response.ok` to help people moving code to async.

Not directly. Users will get a new property for `ok` to check if the `status` is under 400.

## Checklist

### DIFF
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2202,8 +2202,9 @@ async def test_redirect_without_location_header(aiohttp_client) -> None:
     assert data == body
 
 
-@pytest.mark.parametrize("status", [200, 201, 301, 400, 403, 500],
-                         ids= lambda x: f"status_code: {x}")
+@pytest.mark.parametrize(
+    "status", [200, 201, 301, 400, 403, 500], ids=lambda x: f"status_code: {x}"
+)
 async def test_ok_from_status(aiohttp_client, status) -> None:
 
     async def handler(request):
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2203,7 +2203,7 @@ async def test_redirect_without_location_header(aiohttp_client) -> None:
 
 
 @pytest.mark.parametrize(
-    ("status", "expected_ok"),
+    "status, expected_ok",
     (
         (200, True),
         (201, True),
@@ -2211,8 +2211,7 @@ async def test_redirect_without_location_header(aiohttp_client) -> None:
         (400, False),
         (403, False),
         (500, False),
-    ),
-    ids=lambda tc: f"HTTP status code {tc[0]} is{' ' if tc[1] else ' not'} ok",
+    )
 )
 async def test_ok_from_status(aiohttp_client, status, expected_ok) -> None:
 
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2203,7 +2203,7 @@ async def test_redirect_without_location_header(aiohttp_client) -> None:
 
 
 @pytest.mark.parametrize(
-    "status, expected_ok",
+    ("status", "expected_ok"),
     (
         (200, True),
         (201, True),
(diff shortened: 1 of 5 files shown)


tests/test_client_functional.py
```diff
@@ -2202,6 +2202,24 @@ async def handler_redirect(request):
...
+
+    app = web.Application()
+    app.router.add_route('GET', '/endpoint', handler)
+    client = await aiohttp_client(app, raise_for_status=False)
+    resp = await client.get('/endpoint')
+
+    if resp.status < 400:
+        assert resp.ok
+    else:
+        assert not resp.ok
```
webknjaz: You shouldn't add such logic into tests. Either factor it out into the params matrix or split the test into two — positive and negative.

tests/test_client_functional.py
```diff
@@ -2202,6 +2202,24 @@ async def handler_redirect(request):
     assert data == body
 
 
+@pytest.mark.parametrize(
+    "status", [200, 201, 301, 400, 403, 500], ids=lambda x: f"status_code: {x}"
+)
+async def test_ok_from_status(aiohttp_client, status) -> None:
```
webknjaz: You could do
```suggestion
@pytest.mark.parametrize(
    ("status", "expected_ok"),
    (
        (200, True),
        (201, True),
        (301, True),
        (400, False),
        (403, False),
        (500, False),
    ),
    ids=lambda c, y: f"HTTP status code {c} is{' ' if o else ' not'} ok",
)
async def test_ok_from_status(aiohttp_client, status, expected_ok) -> None:
```

tests/test_client_functional.py
webknjaz:
```suggestion
    assert resp.ok is expected_ok
```

webknjaz:
@EvanKepner please update the docs @ https://github.com/aio-libs/aiohttp/blob/master/docs/client_reference.rst and I'll be able to merge this.

EvanKepner:
@webknjaz thanks for the fast review and the assist on the test structure! For docs - is this a manual edit on that file or built some other way through sphinx? I tried `make doc` from the top level and there were no files changed to commit after the pages rebuild. If it's a manual edit I'll push a new commit but if it should be auto-generated some other way let me know.

webknjaz:
Yes, just change that RST file.

webknjaz (approved):
Perfect!

tests/test_client_functional.py
```diff
@@ -2202,6 +2202,30 @@ async def handler_redirect(request):
...
+    ("status", "expected_ok"),
+    (
+        (200, True),
+        (201, True),
+        (301, True),
+        (400, False),
+        (403, False),
+        (500, False),
+    ),
+    ids=lambda tc: f"HTTP status code {tc[0]} is{' ' if tc[1] else ' not'} ok",
```
webknjaz: @EvanKepner I cannot debug how to make it write in a browser. Could you please fix this locally?

EvanKepner: Yup, no problem - pushed a commit and it should pass CI now. `ids` is an optional arg, and with the structure of the tuple for parameterization I think pytest is unpacking as a list of lists which makes the lambda function difficult. I dropped `ids` entirely to use the default values and changed the first arg to be a single string. 

The other way around this would be to do something like:

```python
from typing import NamedTuple

class OKCases(NamedTuple):
    status: int
    expected_ok: boolean
    
    @property 
    def idfn(self):
        return f"{self.status} is {self.expected_ok}"
```

For the parameterized tuples to access as a single value but that feels excessive just for ids that already print the values.

webknjaz: Well, the previous version of lambda relied on unpacking but still was breaking in the CI somehow. Thanks for the fix anyway

webknjaz: As for the param names, I prefer them to be a tuple of strings for readability

Issue #4703, opened 2020-04-21
ResourceWarning:unclosed transport


spacemanspiff2007:
🐞 **Describe the bug**
I have a long running application that creates requests with aiohttp.
Some requests have a long time between them. 
After a certain time without a request a ResourceWarning appears in the log:
```
C:\venv\lib\asyncio\sslproto.py:322: ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000003FA9A88>
```
It is server depended, my guess is that the server closes the connection.
For ``https://heise.de`` the time is at around 125secs.

💡 **To Reproduce**
```python
import asyncio
import datetime
import warnings

import aiohttp


def get_ts():
    return f'{datetime.datetime.now()} |'


def print_warnings(message, category, filename, lineno, file=None, line=None):
    print('%s %s:%s: %s:%s' % (get_ts(), filename, lineno, category.__name__, message))
    return


# Setup deprecation warnings
warnings.simplefilter('default')
warnings.showwarning = print_warnings


loop = asyncio.get_event_loop()


async def run():
    sleep = 55
    client = aiohttp.ClientSession(loop=loop)
    
    while True:
        async with client.get('https://creativecommons.tankerkoenig.de/json/list.php?lat=52.521&lng=13.438&rad=1.5&sort=dist&type=all&apikey=00000000-0000-0000-0000-000000000002') as resp:
            status = resp.status
            data = await resp.text()
        
        print(f'{get_ts()} sleep: {sleep}, Status: {status}')
        sleep += 5
        await asyncio.sleep(sleep)


loop.run_until_complete(run())
```

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
2020-04-21 15:05:10.750123 | sleep: 55, Status: 200
2020-04-21 15:06:10.857561 | sleep: 60, Status: 200
2020-04-21 15:07:16.039289 | sleep: 65, Status: 200
2020-04-21 15:08:21.037007 | C:\venv\lib\asyncio\sslproto.py:322: ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000003FA9A88>
2020-04-21 15:08:26.135299 | sleep: 70, Status: 200
```

📋 **Versions**
```console
Python 3.7.4
```
```console
Name: aiohttp
Version: 3.6.2
```
```console
Name: yarl
Version: 1.3.0
```

webknjaz:
[...]

PR #4713, opened 2020-04-26
Test for ResourceWarning:unclosed transport

spacemanspiff2007:
## What do these changes do?

Add a test for `ResourceWarning:unclosed` transport

None

## Related issue number

#4703

## Checklist


spacemanspiff2007:
```python
```
is failing and I am not sure how to fix it.

tests/test_resource_warning.py
```diff
@@ -0,0 +1,81 @@
...
+        if last_warning:
+            await client.close()
+            raise ResourceWarning(last_warning)
+
+    await client.close()
+    return None
+
+
+def test_ressource_warning(warning_setup):
+    loop = asyncio.get_event_loop()
```
webknjaz: It's failing because it should be in `async def` to work. But then, why don't you just do `await _test_warning()`?

tests/test_resource_warning.py
```diff
@@ -0,0 +1,81 @@
...
+
+            request_interval += request_increase
+            next_request = now + request_interval
+            print(f'{datetime.now()} | '
+                  f'Status: {status}, next request in {request_interval}s')
+
+        await asyncio.sleep(1)
+        if last_warning:
+            await client.close()
+            raise ResourceWarning(last_warning)
```
webknjaz: Why do you need all this complex logic here? Tests should be simple and not catch errors.

tests/test_resource_warning.py
```diff
@@ -0,0 +1,81 @@
...
+@pytest.fixture()
+def warning_setup():
+    # setup warning
+    old_func = warnings.showwarning
+    warnings.showwarning = print_warnings
+    warnings.simplefilter('default')
+
+    yield
+
+    warnings.showwarning = old_func
```
webknjaz: This can be simplified as
```suggestion
def warning_setup(monkeypatch):
    monkeypatch.setattr(warnings, 'showwarning', print_warnings)
    warnings.simplefilter('default')
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,81 @@
...
+        await asyncio.sleep(1)
+        if last_warning:
+            await client.close()
+            raise ResourceWarning(last_warning)
+
+    await client.close()
+    return None
+
+
+def test_ressource_warning(warning_setup):
```
webknjaz: Why didn't you use the [`recwarn` fixture](https://docs.pytest.org/en/latest/warnings.html#recwarn) here?

tests/test_resource_warning.py
```diff
@@ -0,0 +1,81 @@
...
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(_test_warning())
+
+
+if __name__ == '__main__':
+    warnings.showwarning = print_warnings
+    warnings.simplefilter('default')
+
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(_test_warning())
```
webknjaz: This is unnecessary, `pytest` should be running tests `pytest tests/test_resource_warning.py` would do this.
```suggestion
```

tests/test_resource_warning.py
spacemanspiff2007: I am not sure where the warning comes from. If it comes from a scheduled future recwarn will not catch it.

spacemanspiff2007:
I modified the test but when running through pytest the behavior is different.
The test does not fail and the warnings appear (at least it seems like it) once the test is finished.

```
2020-04-26 17:07:13.406819 | Status: 200, next request in 45s
2020-04-26 17:07:58.647407 | Status: 200, next request in 60s
2020-04-26 17:08:58.872852 | Status: 200, next request in 75s
2020-04-26 17:10:14.157158 | Status: 200, next request in 90s
2020-04-26 17:11:44.414320 | Status: 200, next request in 105s
2020-04-26 17:12:14.449038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C45B88>
2020-04-26 17:12:14.449038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C2D248>
2020-04-26 17:12:14.449038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C3CF08>
2020-04-26 17:12:14.449038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C485C8>
2020-04-26 17:12:14.449038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C454C8>
2020-04-26 17:12:14.450038 | C:\Progs\WinPython-64bit-3.7.4.0Ps2\python-3.7.4.amd64\lib\asyncio\sslproto.py:322 ResourceWarning:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x0000000004C3C5C8>
```

setup.cfg
```diff
@@ -34,7 +34,7 @@ source = aiohttp, tests
 omit = site-packages
 
 [tool:pytest]
-addopts = --cov=aiohttp -v -rxXs --durations 10
+addopts = --cov=aiohttp -v -rxXs --durations 15
```
webknjaz: we probably don't want to change this

webknjaz:
@spacemanspiff2007 maybe revert the test to normal `def` function but instead of `get_event_loop()` use `loop = asyncio.new_event_loop()`.

setup.cfg
spacemanspiff2007: Yes, but it took so long to trigger the warnings so I tried to increase it.
It's back to normal now.

spacemanspiff2007:
Now it fails as expected

tests/test_resource_warning.py
```diff
@@ -0,0 +1,68 @@
...
+            print(f'{datetime.now()} | '
+                  f'Status: {status}, next request in {request_interval}s')
+
+        await asyncio.sleep(5)
+
+    await client.close()
+    assert last_warning is None, last_warning
+
+
+def test_resource_warning(warning_setup):
```
webknjaz: Now that it's reproducible, try `recwarn`. I think it should work now.

tests/test_resource_warning.py
```diff
@@ -0,0 +1,68 @@
...
+    # the future has to be run like this, with the pytest async runner
+    # the warnings appear only after the test
+    loop = asyncio.new_event_loop()
+    loop.run_until_complete(_resource_warning())
+    loop.close()
```
webknjaz: How about
```suggestion
def test_resource_warning(recwarn, warning_setup):
    # the future has to be run like this, with the pytest async runner
    # the warnings appear only after the test
    loop = asyncio.new_event_loop()
    loop.run_until_complete(_resource_warning())
    loop.close()
    assert not str(recwarn.pop(RuntimeWarning).message).startswith(
        'unclosed transport '
        '<asyncio.sslproto._SSLProtocolTransport object at ',
    )
```
(but using `pytest.warns()` CM` would also work)

tests/test_resource_warning.py
```diff
@@ -0,0 +1,68 @@
...
+def warning_setup(monkeypatch):
+    # setup warning
+    monkeypatch.setattr(warnings, 'showwarning', print_warnings)
+    warnings.simplefilter('default')
+
+
+async def _resource_warning():
+    timeout = 3 * 60
+
+    client = aiohttp.ClientSession()
```
webknjaz: It's better to use an async CM style for this

tests/test_resource_warning.py
```diff
@@ -0,0 +1,68 @@
...
+    next_request = 0
+
+    end = time.time() + timeout
+    while last_warning is None:
+        now = time.time()
+        if now > end:
+            break
+
+        if now > next_request:
+            async with client.get('https://heise.de') as resp:
```
webknjaz: It's better to use a local server. Here's fixtures that will help you to set up a trusted TLS context while still using real-world SSL: https://github.com/aio-libs/aiohttp/commit/c180800a4c90dc123d05311edbec92a3a82d6317#diff-5d43d8b7848138232c802e32ab023dd3R271-R272

tests/test_resource_warning.py
spacemanspiff2007: This works
```python
    for warn in recwarn:
        assert 'unclosed transport' not in str(warn.message)
```
but provides not the fast exit on the first warning keeping the test running for some unnecessary time.

tests/test_resource_warning.py
spacemanspiff2007: How would I use the fixture?

webknjaz: Add it as an arg into fixtures that set up a web server and a client and pass as `ssl`:
>    server = await aiohttp_server(app, ssl=**ssl_ctx**)
>    conn = aiohttp.TCPConnector(ssl=**client_ssl_ctx**)

tests/test_resource_warning.py
webknjaz: What if you pass `recwarn` object as a param to `_resource_warning()` and then use it inside on each iteration? Will it work via closure? (as in recording the outer warnings that the loop emits while having access to them from the inside)
```python
    loop.run_until_complete(_resource_warning(recwarn))
```
this way you could terminate the loop early.

tests/test_resource_warning.py
spacemanspiff2007: I'm sorry I don't know how to setup a server and use it.
Can't we leave it like this, it seems to be working fine.

webknjaz: Tests should work offline. Imagine you're traveling w/o any internet connectivity. You should be able to run this on your laptop.

tests/test_resource_warning.py
```diff
@@ -0,0 +1,45 @@
...
+
+                request_interval += request_increase
+                next_request = now + request_interval
+                print(f'{datetime.now()} | '
+                      f'Status: {status}, next request in {request_interval}s')
+
+            await asyncio.sleep(5)
+
+            for warn in recwarn:
+                print(f'{datetime.now()} | {warn}')
```
webknjaz: print is unnecessary, also pytest logs local vars on failures anyway and they are visible in logs

tests/test_resource_warning.py
spacemanspiff2007: If you could make a suggestion I'd gladly copy paste it.

tests/test_resource_warning.py
```diff
@@ -0,0 +1,45 @@
...
+                assert 'unclosed transport' not in str(warn.message)
```
webknjaz: I think this reads better sematically:
```suggestion
            assert not any('unclosed transport' in str(w.message) for w in recwarn)
            recwarn.clear()
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,44 @@
...
+
+                request_interval += request_increase
+                next_request = now + request_interval
+                print(f'{datetime.now()} | '
+                      f'Status: {status}, next request in {request_interval}s')
+
+            await asyncio.sleep(5)
+
+            for warn in recwarn:
+                assert 'unclosed transport' not in str(warn.message)
```
webknjaz: I think this reads better sematically:
```suggestion
            assert not any('unclosed transport' in str(w.message) for w in recwarn)
            recwarn.clear()
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,44 @@
...
+                async with client.get('https://heise.de') as resp:
+                    status = resp.status
+                    await resp.text()
+
+                request_interval += request_increase
+                next_request = now + request_interval
+                print(f'{datetime.now()} | '
+                      f'Status: {status}, next request in {request_interval}s')
+
+            await asyncio.sleep(5)
```
webknjaz, line 42: Is this reproducible with one request if you sleep long enough?

tests/test_resource_warning.py
webknjaz: Actually, you don't have to manually copy suggestions, it's better to click `Commit suggestion` from the GitHub UI because this way it automatically preserves authorship.

tests/test_resource_warning.py
spacemanspiff2007: No

tests/test_resource_warning.py
```diff
@@ -0,0 +1,46 @@
...
+            assert not any('unclosed transport' in
+                           str(w.message) for w in recwarn)
+
+        recwarn.clear()
+
+
+def test_resource_warning(recwarn):
+    # the future has to be run like this because with the pytest async runner
+    # the warnings appear only after the test
+    loop = asyncio.new_event_loop()
```
webknjaz, line 53: Let's also enable debug output for this loop:
```suggestion
    loop = asyncio.new_event_loop()
    loop.set_debug(True)
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+async def _resource_warning(recwarn):
+    timeout = 3 * 60
+
+    request_interval = 60
+    request_increase = 30
+    next_request = 0
+
+    async with aiohttp.ClientSession() as client:
```
webknjaz: something like this should work
```suggestion

    async def handler_coro(_request):
        return aiohttp.web.Response(text='Test TLS response')
    http_server = aiohttp.test_utils.RawTestServer(handler_coro, ssl=ssl_ctx)
    await http_server.start_server()

    url = http_server.make_url('/')

    async with aiohttp.ClientSession(
        connector=aiohttp.TCPConnector(ssl=client_ssl_ctx),
    ) as client:
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,47 @@
...
+
+        end = time.time() + timeout
+        while True:
+            now = time.time()
+            if now > end:
+                break
+
+            if now > next_request:
+                async with client.get('https://heise.de') as resp:
```
webknjaz:
```suggestion
                async with client.get(url) as resp:
```
(you can add this suggestion to batch ☝️ and commit two suggestions together)

tests/test_resource_warning.py
```diff
@@ -0,0 +1,47 @@
+import asyncio
+import time
+from datetime import datetime
+
+import aiohttp
+
+
+async def _resource_warning(recwarn):
```
webknjaz:
```suggestion
async def _resource_warning(recwarn, ssl_ctx, client_ssl_ctx):
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,47 @@
...
+
+            await asyncio.sleep(5)
+
+            assert not any('unclosed transport' in
+                           str(w.message) for w in recwarn)
+
+        recwarn.clear()
+
+
+def test_resource_warning(recwarn):
```
webknjaz:
```suggestion
def test_resource_warning(recwarn, ssl_ctx, client_ssl_ctx):
```

tests/test_resource_warning.py
```diff
@@ -0,0 +1,47 @@
...
+    # the future has to be run like this because with the pytest async runner
+    # the warnings appear only after the test
+    loop = asyncio.new_event_loop()
+    loop.set_debug(True)
+    loop.run_until_complete(_resource_warning(recwarn))
```
webknjaz:
```suggestion
    loop.run_until_complete(_resource_warning(recwarn, ssl_ctx, client_ssl_ctx))
```

tests/test_resource_warning.py
spacemanspiff2007: A - too late. Sorry - it's pretty new for me.

spacemanspiff2007:
Hm - your internal server changes something so the test does not fail any more.

webknjaz:
> Hm - your internal server changes something so the test does not fail any more.

Hm... https://heise.de responds with `Connection: keep-alive`, I wonder if the test server does that. Could you please inspect this?

spacemanspiff2007:
Sorry - I don't know much about http connections and don't know how to do that.
I just observed the behavior in the logs of my application.
Maybe you can ping someone who could help?

spacemanspiff2007:
@webknjaz : Shouldn't we revert the test to the external server solution? At least then it's failing reproducible and can be fixed.

samuelcolvin:
Any update on this, I'm having the same problem.

asvetlov:
The master branch has the reimplementation of client session/connector closing which waits for the actual protocol's `connection_lost()` and graceful transport's shutdown.
Unfortunately, the work is not 100% finished, the branch is not ready for release yet.

webknjaz:
@asvetlov why did you close this? I think it's a good foundation for the regression test we need.

asvetlov:
As you wish.
Let's keep it open. 
I'm not sure that the test should be present as-is. Also, I don't like `print()` call here

Dreamsorcerer:
I'm going to close this as it's been sitting around for years and those warnings actually appeared throughout our tests anyway (previously ignored in the pytest config), so I think we have coverage just fine.

Issue #2988, opened 2018-05-10
Customize max_line_size and max_field_size in HttpParser?


alexandru:
We're using `aiohttp` 2.x with Python 3.5.2 and trying to upgrade to the latest aiohttp 3.2.0 with Python 3.6.5.

We're getting errors such as these:

```
[2018-05-10 08:31:55 +0000] [12471] [ERROR] Error handling request
Traceback (most recent call last):
  File "/project/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 235, in data_received
    messages, upgraded, tail = self._request_parser.feed_data(data)
  File "aiohttp/_http_parser.pyx", line 297, in aiohttp._http_parser.HttpParser.feed_data
  File "aiohttp/_http_parser.pyx", line 425, in aiohttp._http_parser.cb_on_header_value
aiohttp.http_exceptions.LineTooLong: 400, message='Got more than 8190 bytes (9383) when reading Header value is too long.'
```

This was happening in aiohttp 2.x as well, because I believe the standard specifies that length as the maximum, however the real world doesn't always agree with standards 😒

To fix this, with aiohttp 2.x we were monkey-patching [HttpParser](https://github.com/aio-libs/aiohttp/blob/3.2/aiohttp/_http_parser.pyx#L425) in order to override `max_line_size` and `max_field_size`. Here's the horrible things we've been doing: [gist](https://gist.github.com/alexandru/f68bc29aac2e4961f7720ecfba63afd3) — this file being imported in our `gunicorn_worker.py`.

Now that we've upgraded to 3.x, this monkey patching no longer works. Don't know why.

Any way to customize `max_line_size` and `max_field_size`?
Possibly without monkey patching, but not necessarily, atm I just want the thing to work?

Environment:

- Python 3.6.5
- aiohttp 3.2.0
- Gunicorn 19.8.1 (via `aiohttp.worker.GunicornWebWorker`)

See the full [requirements.txt](https://gist.github.com/alexandru/fa54cde5088be9fb6acb56808efb01f9) if interested.

Cheers,

asvetlov:
Nobody implemented it yet.
We have the very similar issue in backlog: #2304

pposca:
Where would be the best place to expose `max_line_size ` and `max_field_size `? Using an Application __init__ param, like with `client_max_size`?

asvetlov:
@pposca let's push params into `BaseRequest` first.
After that, we can think how to configure it the best way.
[...]

PR #4720, opened 2020-04-27
GunicornWebWorker subclasses can config AppRunner

bmbouter:
Currently there is no way to configure the Application Runner created by
the `aiohttp.GunicornWebWorker`.

This allows `GunicornWebWorker` subclasses to configure its
`ApplicationRunner` by introducing the
`GunicornWebWorker._get_application_runner_instance` method.

## What do these changes do?

This implements Design 1 from this request for comment, comment https://github.com/aio-libs/aiohttp/issues/2988#issuecomment-611124997

No

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2988

## Checklist

### DIFF
--- a/aiohttp/worker.py
+++ b/aiohttp/worker.py
@@ -73,13 +73,7 @@ class GunicornWebWorker(base.Worker):
             raise RuntimeError("wsgi app should be either Application or "
                                "async function returning Application, got {}"
                                .format(self.wsgi))
-        access_log = self.log.access_log if self.cfg.accesslog else None
-        runner = web.AppRunner(app,
-                               logger=self.log,
-                               keepalive_timeout=self.cfg.keepalive,
-                               access_log=access_log,
-                               access_log_format=self._get_valid_log_format(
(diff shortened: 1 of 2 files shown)


aiohttp/worker.py
```diff
@@ -207,6 +201,17 @@ def _get_valid_log_format(self, source_format: str) -> str:
         else:
             return source_format
 
+    def _get_application_runner_instance(self, app, **kwargs):
```
asvetlov, line 204: The method should be documented and type hinted.

asvetlov: How `**kwargs` are used? I see it is always an empty dict.

asvetlov (changes requested):
The PR doesn't pass the linter CI checks.
Sorry, I cannot accept it as-is.

bmbouter:
Our project uses a locally patched workaround. I'd still like to see this accepted so we don't have to carry it, but I'm unable to finish it so I'll close for now.

Issue #4700, opened 2020-04-19
unittest asyncio TestCase


WisdomPill:
🐣 **Is your feature request related to a problem? Please describe.**

As discussed in #4695 python 3.8 (that this library just started supporting) introduced test cases asyncio capable, for versions below 3.8 there is a library `asynctest` that is almost equal to what has been introduced in python 3.8. This would make the decorator `unittest_run_loop` useless and make it simpler to write `async` test cases.

💡 **Describe the solution you'd like**

As shown in the PR  #4697 `asynctest` is useful for people using python below 3.8 for patching but for the testcase and mock as well. I tried to include this library in case python below 3.8 and `IsolatedAsyncioTestCase` and here is an example of what should change in test_utils.

``` python
if PY_38:
    from unittest import IsolatedAsyncioTestCase as TestCase
else:
    from asynctest import TestCase  # type: ignore

class AioHTTPTestCase(TestCase):
    """A base class to allow for unittest web applications using
    aiohttp.

    Provides the following:

    * self.client (aiohttp.test_utils.TestClient): an aiohttp test client.
    * self.loop (asyncio.BaseEventLoop): the event loop in which the
        application and server are running.
    * self.app (aiohttp.web.Application): the application returned by
        self.get_application()

    Note that the TestClient's methods are asynchronous: you have to
    execute function on the test client using asynchronous methods.
    """

    async def get_application(self) -> Application:
        """
        This method should be overridden
        to return the aiohttp.web.Application
        object to test.

        """
        return self.get_app()

    def get_app(self) -> Application:
        """Obsolete method used to constructing web application.

        Use .get_application() coroutine instead

        """
        raise RuntimeError("Did you forget to define get_application()?")

    def setUp(self) -> None:
        if PY_38:
            self.loop = asyncio.get_event_loop()

        self.loop.run_until_complete(self.setUpAsync())

    async def setUpAsync(self) -> None:
        self.app = await self.get_application()
        self.server = await self.get_server(self.app)
        self.client = await self.get_client(self.server)

        await self.client.start_server()

    def tearDown(self) -> None:
        self.loop.run_until_complete(self.tearDownAsync())

    async def tearDownAsync(self) -> None:
        await self.client.close()

    async def get_server(self, app: Application) -> TestServer:
        """Return a TestServer instance."""
        return TestServer(app)

    async def get_client(self, server: TestServer) -> TestClient:
        """Return a TestClient instance."""
        return TestClient(server)

```

[...]

PR #4697, opened 2020-04-18
Add docs for using patch decorator in unittest prior to python 3.8

WisdomPill:
## What do these changes do?

As discussed in #4695 patch and make_mocked_coro do not work in a decorated unittest test case. So in the docs it is advised to use a package called `asynctest` for async method patching.

## Are there changes in behavior for the user?

It is described in the docs.

## Related issue number

#4695

Issue #4695, opened 2020-04-17
Unable to patch with decorator using unittest_run_loop


WisdomPill:
🐞 **Describe the bug**

patch using decorators is not possible while using unittest_run_loop, 
to overcome this issue patching with context manager is needed, it can be a real pain to do it multiple times in the same test case. The arguments are not passed to the unittest test case method.

💡 **To Reproduce**

```python
from unittest.mock import patch

from aiohttp.test_utils import AioHTTPTestCase, make_mocked_coro, unittest_run_loop
from aiohttp.web_app import Application
from aiohttp.web_request import Request
from aiohttp.web_response import Response
from aiohttp.web_routedef import get


async def do_something():
    print('something')


async def ping(request: Request) -> Response:
    await do_something()
    return Response(text='pong')


class TestApplication(AioHTTPTestCase):
    def get_app(self) -> Application:
        app = Application()
        app.router.add_routes([
            get('/ping/', ping)
        ])

        return app

    @unittest_run_loop
    async def test_ping(self):
        resp = await self.client.get('/ping/')

        self.assertEqual(resp.status, 200)
        self.assertEqual(await resp.text(), 'pong')

    @unittest_run_loop
    async def test_ping_mocked_do_something(self):
        with patch('test_bug.do_something', make_mocked_coro()) as do_something_patch:
            resp = await self.client.get('/ping/')

            self.assertEqual(resp.status, 200)
            self.assertEqual(await resp.text(), 'pong')

            self.assertTrue(do_something_patch.called)

    @unittest_run_loop
    @patch('test_bug.do_something', make_mocked_coro())
    async def test_ping_mocked_do_something_decorated(self, do_something_patch):
        resp = await self.client.get('/ping/')

        self.assertEqual(resp.status, 200)
        self.assertEqual(await resp.text(), 'pong')

        self.assertTrue(do_something_patch.called)

```

💡 **Expected behavior**

I expect to be able to use decorators to patch or mock a module/function

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
something


Ran 3 tests in 0.080s

FAILED (errors=1)

Error
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
[...]

PR #4732, opened 2020-05-09
AioHTTPTestCase more async friendly

WisdomPill:
## What do these changes do?

Make AioHTTPTestCase more async friendly, making a user able to run a test case with async methods without needing to use `@unittest_run_loop` decorator, adding a dependency for testing for python3.7 and below as mentioned in #4700 and #4695 while using built in asyncio test case available from python 3.8

This change is backward compatible, I did not change any name of classes or methods.

Yes, test can be written with async functions without `@unittest_run_loop`.
For python3.7 and below `asynctest` dependency must be installed.

Testing becomes more straight forward like the following
```python
from aiohttp.test_utils import AioHTTPTestCase
from aiohttp.web_app import Application
from aiohttp.web_request import Request
from aiohttp.web_response import Response
from aiohttp.web_routedef import get

async def ping(request: Request) -> Response:
    return Response(text='pong')


class TestApplication(AioHTTPTestCase):
    def get_app(self) -> Application:
        app = Application()
        app.router.add_routes([
            get('/ping/', ping)
        ])

        return app

    async def test_ping(self):
        resp = await self.client.get('/ping/')

        self.assertEqual(resp.status, 200)
        self.assertEqual(await resp.text(), 'pong')

```

## Related issue number

#4700 and #4695

#4700 will be resolved after this merge

## Checklist

I did not add any new tests, it seems to me that the old tests are passing, I would be glad to know if this kind of change is welcome so that I can proceed with the docs and more testing maybe on the patching side like I pointed out in this PR #4697

### DIFF
--- a/CHANGES/4700.feature
+++ b/CHANGES/4700.feature
@@ -1,1 +1,6 @@
 AioHTTPTestCase more async friendly
+
+for the people who use unittest and are used to use unittest.TestCase
+it will be easier to write new test cases like the sync version of the TestCase class,
+without using the decorator `@unittest_run_loop`, just `async def test_*`.
+The only difference is that for the people using python3.7 and below a new dependency is needed, it is `asynctestcase`.
--- a/CHANGES/4700.feature
+++ b/CHANGES/4700.feature
@@ -1,6 +1,6 @@
 AioHTTPTestCase is more async friendly now.
 
-for the people who use unittest and are used to use unittest.TestCase
+For people who use unittest and are used to use unittest.TestCase
 it will be easier to write new test cases like the sync version of the TestCase class,
 without using the decorator `@unittest_run_loop`, just `async def test_*`.
 The only difference is that for the people using python3.7 and below a new dependency is needed, it is `asynctestcase`.
--- a/CHANGES/4700.feature
+++ b/CHANGES/4700.feature
@@ -1,4 +1,4 @@
-AioHTTPTestCase more async friendly
+AioHTTPTestCase is more async friendly now.
 
 for the people who use unittest and are used to use unittest.TestCase
 it will be easier to write new test cases like the sync version of the TestCase class,
--- a/CHANGES/4700.feature
+++ b/CHANGES/4700.feature
@@ -1,0 +1,1 @@
+AioHTTPTestCase more async friendly
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,5 +1,5 @@
 -r lint.txt
 -r test.txt
 -r doc.txt
-cherry_picker==1.3.2; python_version>="3.6"
 asynctest==0.13.0; python_version<"3.8"
+cherry_picker==1.3.2; python_version>="3.6"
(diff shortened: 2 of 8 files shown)


WisdomPill:
I've added tests to prove the new capabilities of AioHTTPTestCase,

Maybe the next step would be to add patch tests? 

Should `@unittest_run_loop` be deprecated now?

CHANGES/4700.feature
```diff
@@ -0,0 +1 @@
+AioHTTPTestCase more async friendly
```
webknjaz: It'd be great if you could expand what such a friendliness means.

WisdomPill: Yes, sure

CHANGES/4700.feature
```diff
@@ -0,0 +1,6 @@
...
+
+for the people who use unittest and are used to use unittest.TestCase
+it will be easier to write new test cases like the sync version of the TestCase class,
+without using the decorator `@unittest_run_loop`, just `async def test_*`.
+The only difference is that for the people using python3.7 and below a new dependency is needed, it is `asynctestcase`.
```
webknjaz, line 6: Please remember that fragments are rendered as RST and not Markdown:
```suggestion
without using the decorator ``@unittest_run_loop``, just ``async def test_*``.
The only difference is that for the people using python3.7 and below a new dependency is needed, it is ``asynctestcase``.
```

CHANGES/4700.feature
```diff
@@ -0,0 +1,6 @@
+AioHTTPTestCase more async friendly
+
+for the people who use unittest and are used to use unittest.TestCase
```
webknjaz: Did you mean to start a new sentence?
```suggestion
AioHTTPTestCase more async friendly.

For the people who use unittest and are used to use unittest.TestCase
```

WisdomPill:
@asvetlov how can I solve the CLA that is not signed?

asvetlov:
Please forget about the CLA.
We had a discussion, many people don't like it and I was convinced that the CLA is not required actually.
Switched it off back.

asvetlov (approved):
Interesting, I thought that the difference between `asynctest` and `IsolatedAsyncioTestCase` is bigger.

CHANGES/4700.feature
```diff
@@ -0,0 +1,6 @@
+AioHTTPTestCase more async friendly
```
asvetlov:
```suggestion
AioHTTPTestCase is more async friendly now.
```

CHANGES/4700.feature
asvetlov:
```suggestion
For people who use unittest and are used to use unittest.TestCase
```

requirements/dev.txt
```diff
@@ -1,4 +1,5 @@
 -r lint.txt
 -r test.txt
 -r doc.txt
+asynctest==0.13.0; python_version<"3.8"
```
asvetlov:
```suggestion
```

Issue #4054, opened 2019-09-10
Implement StreamReader.readuntil() method


yapith-suplari:
Long story short:
Should be mimic to asyncio.StreamReader counterpart: https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readuntil

(Re-opening https://github.com/aio-libs/aiohttp/issues/1151)

This solution is great to have in the toolbox dealing with pulling high-load data from servers that don't declare their response content-length. Currently I run into partial content.read() intermittently under high-load situations e.g. (https://github.com/aio-libs/aiohttp/issues/3881).

asvetlov:
You can master something with `content.readany()` already without waiting for a new aiohttp version.

abeeeshek:
something i am currently dealing with, @asvetlov can you review the PR #4734 @WisdomPill, i can pitch in for the remaining task.

WisdomPill:
What is the remaining task?

asvetlov:
Fixed by #4734

PR #4734, opened 2020-05-11
Implemented readuntil in StreamResponse 

WisdomPill:
## What do these changes do?

Added `readuntil` method, heavily mimicking `asyncio.streams` and recycling the code of readline in current `StreamResponse` class. With this change `readline` uses `readuntil` with `\n` as a separator
Same for the tests that just use different separators.

New way to wait for new data

## Related issue number

#4054 

## Checklist

- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` This change is inside another PR I submitted #4732 

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -312,15 +312,15 @@ class StreamReader(AsyncStreamReaderMixin):
     async def readline(self) -> bytes:
         return await self.readuntil()
 
-    async def readuntil(self, separator: bytes=b'\n') -> bytes:
+    async def readuntil(self, separator: bytes = b"\n") -> bytes:
         seplen = len(separator)
         if seplen == 0:
-            raise ValueError('Separator should be at least one-byte string')
+            raise ValueError("Separator should be at least one-byte string")
 
         if self._exception is not None:
             raise self._exception
 
-        chunk = b''
+        chunk = b""
         chunk_size = 0
         not_enough = True
 
@@ -336,13 +336,13 @@ class StreamReader(AsyncStreamReaderMixin):
                     not_enough = False
 
                 if chunk_size > self._high_water:
-                    raise ValueError('Chunk too big')
+                    raise ValueError("Chunk too big")
 
             if self._eof:
                 break
 
             if not_enough:
-                await self._wait('readuntil')
+                await self._wait("readuntil")
 
         return chunk
 
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -310,34 +310,41 @@ class StreamReader(AsyncStreamReaderMixin):
             self._waiter = None
 
     async def readline(self) -> bytes:
+        return await self.readuntil()
+
+    async def readuntil(self, separator: bytes=b'\n') -> bytes:
+        seplen = len(separator)
+        if seplen == 0:
+            raise ValueError('Separator should be at least one-byte string')
+
         if self._exception is not None:
             raise self._exception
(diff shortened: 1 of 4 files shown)


WisdomPill:
I need to check if documentation work is needed or if it is automatic, meanwhile if somebody can clear my mind about this before I check it is welcome

aiohttp/streams.py
```diff
@@ -300,34 +300,41 @@ async def _wait(self, func_name: str) -> None:
...
+        chunk = b''
+        chunk_size = 0
         not_enough = True
 
         while not_enough:
             while self._buffer and not_enough:
                 offset = self._buffer_offset
-                ichar = self._buffer[0].find(b'\n', offset) + 1
-                # Read from current offset to found b'\n' or to the end.
+                ichar = self._buffer[0].find(separator, offset) + 1
```
socketpair, line 330: asyncio has much more effective solution. Why not to copy-paste it ?

WisdomPill: I did not want to change much the code.

What do you think should be copied from asyncio implementation?

asvetlov: I think sync the implementation with asyncio would be great!
asyncio doesn't incrementally copy data to the buffer in a loop but calculates boundaries and move bytes once.

asvetlov (changes requested):
I can live with not the-fast-possible implementation but the documentation update is required for merging.

Please update `docs/streams.rst`

aiohttp/streams.py
WisdomPill: Okay, I will fix it

WisdomPill:
Yes, okay, this weekend I will have some time to fix it

WisdomPill:
Hello, 

I tried to mimic more the asyncio implementation but it appears to me that it is quite difficult task and a lot of code will be changed.

Simply copying the data from `self._buffer` using boundaries is not possible as `self._buffer` is a deque.
To change this I can try by making `self._buffer` a `bytearray` which seems to me that it will lead to an improvement in performance but will need a lot of code rewrite.

I added the docs by the way.

WisdomPill:
@asvetlov do you think the StreamReader could be written in a better and more performant way inheriting StreamReader from asyncio?

docs/streams.rst
```diff
@@ -70,6 +70,17 @@ Reading Methods
...
+
+   Read until separator, where `separator` is a sequence of bytes.
+
+   If EOF is received, and `separator` was not found, the method will
+   return the partial read bytes.
+
+   If the EOF was received and the internal buffer is empty, return an
+   empty bytes object.
+
+   :return bytes: the given data
```
asvetlov:
```suggestion
   empty bytes object.

   .. versionadded:: 3.8 

   :return bytes: the given data
```

docs/streams.rst
```diff
@@ -70,6 +70,19 @@ Reading Methods
...
+
+   Read until separator, where `separator` is a sequence of bytes.
+
+   If EOF is received, and `separator` was not found, the method will
+   return the partial read bytes.
+
+   If the EOF was received and the internal buffer is empty, return an
+   empty bytes object.
+
+   .. versionadded:: 3.8
```
asvetlov:
```suggestion
   .. versionadded:: 3.8
```

asvetlov (approved):

asvetlov:
Thanks for the patience!

PR #4736, opened 2020-05-11
Fix @reify type hints

atugushev:
## What do these changes do?

Fixes an issue where `mypy` would incorrectly infer types from wrapped methods by `@reify` decorator (e.g., `web.Request.match_info`).

The solution inspired by [@cached_property](https://github.com/python/typeshed/blob/591522fb6a527318ea100d8adf99d91e662ba26f/stdlib/3/functools.pyi#L91-L99) type hints from [typeshed](https://github.com/python/typeshed) repository. Thanks to @asottile for the assistance!

#### Before

```python
# t.py
from typing import Callable
from typing import Any


class reify:
    def __init__(self, wrapped: Callable[..., Any]) -> None:
        ...

    def __get__(self, inst: Any, owner: Any) -> Any:
        ...

    def __set__(self, inst: Any, value: Any) -> None:
        ...


class Foo:
    @reify
    def bar(self) -> str:
        return "I'm Bar!"


foo = Foo()
print(foo.bar + 42)  # Note str + int 
```

```console
$ mypy t.py
Success: no issues found in 1 source file
```

#### After

```python
# t.py
from typing import Callable
from typing import Generic
from typing import Optional
from typing import Type
from typing import TypeVar
from typing import Any

_T = TypeVar("_T")
_S = TypeVar("_S")


class reify(Generic[_T]):
    def __init__(self, wrapped: Callable[..., _T]) -> None:
        ...

    def __get__(self, inst: _S, owner: Type[Any] = None) -> _T:
        ...

    def __set__(self, inst: _S, value: _T) -> None:
        raise AttributeError("read-only")


class Foo:
    @reify
    def bar(self) -> str:
        return "I'm Bar!"


foo = Foo()
print(foo.bar + 42)  # Note str + int 
```

```console
$ mypy t.py
t.py:30: error: Unsupported operand types for + ("str" and "int")
Found 1 error in 1 file (checked 1 source file)
```

The `mypy` users would probably notice this change since `mypy` now can infer types from various `aiohttp` objects. For example,`request.match_info["foo"]` now infers to `str` (previously `Any`).

## Related issue number

None.

## Checklist


atugushev:
Ouch, this is [much more complicated](https://dev.azure.com/aio-libs/aiohttp/_build/results?buildId=3001&view=logs&j=cc4e4887-e527-5c83-25ce-76663593f2e1&t=9d25c5b1-1b8c-5097-b25c-a7f376352755&l=204) than I would expect. Sorry for the ping! I'll fix up later.

atugushev:
It's ready for the review. I've fixed typing issues wich required by mypy due to changes to `@reify`. Feel free to bikeshed variable names I've picked. I'm not good at naming. :D

Also, it seems I've touched uncovered lines, so I presume we could ignore failed codecov checks.

atugushev:
@webknjaz gentle ping

asvetlov (approved):
Great!
Thanks!

Issue #4587, opened 2020-02-21
Cannot write to closing transport


ajay1mg:
🐞 **Describe the bug**
Every few minutes we get **Cannot write to closing transport** exception on our gateway service which is a part of our microservices infrastructure.

Gateway service calls all other services with a timeout of 10 seconds. The other service that gateway calls are also based on aiohttp framework. One such service which is giving most of the exceptions doesn't raise any on its own server.

We are on ubuntu 16.04 LTS

💡 **Expected behavior**
Our gateway service should be able to handle the timeouts.

We are getting the following exception on our server every few minutes:
📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
Traceback (most recent call last):
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/aiohttp/client.py", line 502, in _request
    resp = await req.send(conn)
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/aiohttp/client_reqrep.py", line 629, in send
    await writer.write_headers(status_line, self.headers)
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/aiohttp/http_writer.py", line 112, in write_headers
    self._write(buf)
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/aiohttp/http_writer.py", line 67, in _write
    raise ConnectionResetError('Cannot write to closing transport')
ConnectionResetError: Cannot write to closing transport

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/hydra/base_api_request.py", line 75, in request
    timeout=cls.request_timeout(timeout))
  File "<string>", line 5, in wrapper
  File "/home/ubuntu/.pyenv/versions/3.7.2/envs/gateway/lib/python3.7/site-packages/newrelic/hooks/framework_aiohttp.py", line 260, in _coro
    response = yield from wrapped(*args, **kwargs)
  File "/home/ubuntu/.pyenv/versions/gateway/lib/python3.7/site-packages/aiohttp/client.py", line 514, in _request
    raise ClientOSError(*exc.args) from exc
aiohttp.client_exceptions.ClientOSError: Cannot write to closing transport

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #4778, opened 2020-05-29
#4587 Always check transport is not closing before reuse

Dahuage:
## What do these changes do?
Always check transport is not closing before reuse a connection.

Reuse a protocol based on keepalive in headers is unreliable.
For example, uWSGI will not support keepalive even it serves a
http1.1 request, except explicitly configure uWSGI with a
`--http-keepalive` option.

Servers designed like uWSGI would cause aiohttp intermittently
raise a ConnectionResetException when the protocol poll runs
out and some protocol is reused.

Though it can be bypassed by properly configuring either in server like uWSGI or aiohttp side, this exception is confusing and annoying for new guys to aiohttp.
Yet more importantly, sometimes it can not be triggered in a development env, but it is a potential disaster when in a production env where it has high concurrence.

No

## Related issue number

#4587

## Checklist


kkonevets:
@Dahuage Hi, you said one can configure properly on aiohttp side to avoid the error, how can I do it?

Dahuage:
> @Dahuage Hi, you said one can configure properly on aiohttp side to avoid the error, how can I do it?



if the error is caused by the server you requested disable keep-alive.

then config the aiohttp send http1.0 request to the server may avoid this problem.

just init the session like this:

```
async def main():
    async with aiohttp.ClientSession(version = aiohttp.http.HttpVersion10) as sess:
        await do_sth_with_sess(sess)
```
by the way, if the server you requested is under your control, make sure your server enable HTTP/1.1 keepalive connections.

hope it helps!

kkonevets:
@Dahuage thanks for the answer. The server is not under my control, so I just enabled http1.0, looks like it works

IsaacFlos:
It's too cool！

945941192:
Thank you for your contribution to help me solve the problem for a long time.

mncu:
wow ! cool !

WoodyFleurant:
Seems that one test fail

aorumbayev:
Any updates on this ?

cdeboeser:
Would also be very interested in this fix! :) And thank you for the contribution!

asvetlov (approved):

asvetlov:
Thanks!

PR #4800, opened 2020-06-06
Ensure cleanup gets run.

### LINKED ISSUE #4799: Cleanup doesn't run when exception occurs in startup.
When an exception occurs in startup, the cleanup is never run.

Expected behaviour: Any cleanup steps added should be run, especially when using the cleanup-context.

e.g.
Task A = Setup DB as cleanup context.
Task B = Setup a remote connection as cleanup context. This throws an exception.

Actual behaviour:
  Task A is successfully setup.
  Task B creates an exception.
  No cleanup happens.

Expected behaviour:
  Task A is successfully setup.
  Task B creates an exception.
  Task A cleanup happens.

Dreamsorcerer:
This helps ensure that cleanup gets run, even when an exception occurs during startup. Fixes #4799 

This becomes particularly painful when the server can't be cleanly restarted without the cleanup running, especially when running locally with aiohttp-devtools, where exceptions during development are likely, and cause the runserver command to break or lock up.

### DIFF
--- a/CHANGES/4999.bugfix
+++ b/CHANGES/4999.bugfix
@@ -1,0 +1,1 @@
+Ensure cleanup context is run even when an exception occurs during startup.
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -302,7 +302,9 @@ class Application(MutableMapping[str, Any]):
 
         Should be called after shutdown()
         """
-        await self.on_cleanup.send(self)
+        # Avoid send() as we want to ensure cleanup is run even if not frozen.
+        for receiver in self.on_cleanup:
+            await receiver(self)
 
     def _prepare_middleware(self) -> Iterator[_Middleware]:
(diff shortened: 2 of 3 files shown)


asvetlov (commented):
I'm sorry but 'no'.
startup/cleanup works like try/finally block:
```
await startup()
try:
    await serve_http()
finally:
    await cleanup()
```
if *startup* call raises exception -- *cleanup* is not called.

Dreamsorcerer:
@asvetlov I was thinking about tweaking this PR to only run parts with a cleanup context.

The problem I am encountering is that development is hard using the devtools, as whenever an exception occurs during startup, the application is left hanging and doesn't restart as it normally would.

It seems odd to me that a cleanup context would be left in an in-between state. The result is that e.g. a DB or other process doesn't get shut down due to an unrelated part of the application.

Issue #4630, opened 2020-03-16
HttpPayloadParser fails with "Not enough data for satisfy transfer length header" on chunked transfer encoding if the data is split exactly where trailers could occur


JustAnotherArchivist:
🐞 **Describe the bug**
If the response data is fed into `HttpPayloadParser.feed_data` in a particular way, the parser is unable to successfully parse chunked data. Specifically, this happens when one call to the function contains the last `0\r\n` chunk but the following `\r\n` is supplied in a separate call.

💡 **To Reproduce**
```python
import aiohttp.http_parser
import io


class Payload: # A minimal payload implementation
        def __init__(self):
                self.data = io.BytesIO()
                self.exc = None

        def feed_data(self, data, size):
                self.data.write(data)

        def feed_eof(self):
                pass

        def set_exception(self, exc):
                self.exc = exc

        def begin_http_chunk_receiving(self):
                pass

        def end_http_chunk_receiving(self):
                pass


payload = Payload()
parser = aiohttp.http_parser.HttpPayloadParser(payload, length = None, chunked = True, compression = None, code = 200, method = 'GET')
print(repr(parser.feed_data(b'4\r\nasdf\r\n0\r\n')))
eof, data = parser.feed_data(b'\r\n')
print(repr((eof, data)))
if not eof:
        print(repr(parser.feed_eof()))
print(repr(payload.data.getvalue()))
```

I added print statements here to debug what exactly aiohttp is returning compared to the simple `feed_data(b'4\r\nasdf\r\n0\r\n\r\n')` call (which works fine).

💡 **Expected behavior**
The parser is able to process the data, and the last line produces `b'asdf'`.

📋 **Logs/tracebacks**
```python-traceback
(False, b'')
(False, b'')
Traceback (most recent call last):
  File "aiohttp-test.py", line 32, in <module>
    print(repr(parser.feed_eof()))
  File ".../lib/python3.6/site-packages/aiohttp/http_parser.py", line 575, in feed_eof
    "Not enough data for satisfy transfer length header.")
aiohttp.http_exceptions.TransferEncodingError: 400, message='Not enough data for satisfy transfer length header.'
```

📋 **Your version of the Python**
```console
$ python --version
Python 3.6.10
```

📋 **Your version of the aiohttp/yarl/multidict distributions**
```console
$ python -m pip show aiohttp | grep Version
Version: 3.6.2
```
```console
$ python -m pip show multidict | grep Version
Version: 4.7.5
```
```console
[...]

PR #4801, opened 2020-06-07
Fix HttpPayloadParser dealing with chunked response (#4630)

rhdxmr:
HttpPayloadParser waits for trailers indefinitely even if there are no trailers
at the response. This happens when only the last CRLF or the last LF are sent
via separate TCP segment.

When the connection is keep alive and if this bug occurs then users experience
response timeout. But this problem is not exposed when keep alive is disabled
because .feed_eof is called. (Instead of TimeoutError, ClientPayloadError is raised if keep alive is disabled)

## What do these changes do?

Fix a bug that HttpPayloadParser waits for data indefintely that will never come.
The bug makes caller of 'await response.read()' awaits forever or for timeout.
There are a few conditions which need to be met in order to reproduce this bug.
- keep alive enabled connection
- response body size is large so that data is splitted into many TCP segments
- Only the last CRLF or just LF is separated alone and is sent via the last TCP segment.
- asyncio tcp protocol handler for the http response receives TCP payload of the last TCP segment. In order for this the last TCP payload should not be merged with the previous data at lower level such as TCP Segmentation Offload.
- The response should not be the last response that the connection allows. The keep alive connection processes the limited number of requests, so if the response is the last thing, the connection is going to be closed by server so that .feed_eof method is called explicitly.

Improvement experience of users who are suffering from mysterous response timeout. There does not exist any log of which response time is slow in sever access log, but client writes log about response timeout. 

## Related issue number
#4630

I had a problem with intermittent response timeout so I conducted debug a lot.
And I managed to locate what caused the problem and fixed it on my own.
And then I was going to add test code that prove my modification really fix that thing. So I learned how to run test code in this project. And I ran test codes for the first time.
Surprisingly the issue 4630 just popped up with XPASS labeled. At this point I realized that the issue had been already reported 3 months ago. Sigh.. If the issue was fixed at that time, I would not spend my time for debugging this problem!

I did not mean to provide the fix on behalf of the person who reported the issue before me. I just did not know there was the issue before I started fixing the bug.

## Checklist

- [  ] Documentation reflects the changes => don't need that because this PR just fixes bug of HttpPayloadParser which has nothing to do with normal users.


rhdxmr:
I am going to do some extra works I found at the checklist. And then I will add that changes here.

rhdxmr:
Hello @asvetlov

Can you review my PR?

And I am wondering why CI fails at test_close of test_web_protocol.py
I am not sure this failure is related to my modification.

aiohttp/http_parser.py
```diff
@@ -668,12 +668,26 @@ def feed_data(self,
                 # we should get another \r\n otherwise
                 # trailers needs to be skiped until \r\n\r\n
                 if self._chunk == ChunkState.PARSE_MAYBE_TRAILERS:
```
socketpair, line 670:
```
if self._chunk == ChunkState.PARSE_MAYBE_TRAILERS:
    if chunk and chunk[:1] != '\r':
        self._chunk = ChunkState.PARSE_TRAILERS
```

Does not it ?

```
0\r\n
\rxxxx: yyy\r\n
\r\n
```

I don't know if it is valid. Need to lookup in RFC

aiohttp/http_parser.py
```diff
@@ -668,12 +668,26 @@ def feed_data(self,
...
-                    if chunk[:2] == SEP:
-                        # end of stream
-                        self.payload.feed_eof()
-                        return True, chunk[2:]
+                    if len(chunk) >= 2:
```
socketpair: Much simplier:
```python
head = chunk[:2]
if not head:
    return False, b''
if head == SEP[:1]:
    self._chunk_tail = head
    return False, b''
if head == SEP:
    # end of stream
    self.payload.feed_eof()
    return True, chunk[2:]
self._chunk = ChunkState.PARSE_TRAILERS
```

rhdxmr: Hello @socketpair
 
I agree with you. So I changed the code simpler as you suggested.
But I moved the `if head == SEP' at the top because it is more likely to happen.

Thanks,

rhdxmr:
Hello @webknjaz,
I am wondering that this PR is acceptable or not.
Can you leave a comment about that?

rhdxmr:
Hello @asvetlov, @webknjaz and @socketpair 

I think this PR is forgotten.
I hope this PR to be merged soon because I am waiting for new version of aiohttp with this patch applied.
I make use of aiohttp in production, but TimeoutError by this bug have kept occurring for several weeks.

Please review this PR positively.

Thanks,

socketpair:
@rhdxmr please rebase from ~~master~~ main branch, and re-push. I would merge... but there are failing checks.

rhdxmr:
RESEND PR: https://github.com/aio-libs/aiohttp/pull/4846

PR #4815, opened 2020-06-12
Fix exception causes in client.py

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

You'll get this: 

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!


cool-RR:
If I understand correctly, the test failure here is not related to my change. (I see it's happening on other PRs.)

asvetlov (approved):

asvetlov:
Agree, you are right.
The proposal makes value and produces better understandable exception info.
I'll merge this PR; please feel free to apply such improvements to all codebase.

PR #4828, opened 2020-06-14
Allow to send a bytestring by WebSocketResponse.send_str

7ckingBest:
## What do these changes do?

WebSocketWriter from `http_websocket.py` can actually send only the binary data and the UTF-8 string must be encoded anyway:
```python
class WebSocketWriter:
    async def send(self, message: Union[str, bytes],
                   binary: bool=False,
                   compress: Optional[int]=None) -> None:
        """Send a frame over the websocket with message as its payload."""
        if isinstance(message, str):
            message = message.encode('utf-8')
        if binary:
            await self._send_frame(message, WSMsgType.BINARY, compress)
        else:
            await self._send_frame(message, WSMsgType.TEXT, compress)
```
Sometimes we need to send a bytestring and the client must understand that this is a text data. For example:
```python
await ws.send_json({'foo': 'bar'}, dumps=orjson.dump)
```
which is equivalent to
```python
await ws.send_str(b'{"foo":"bar"}')
```
Decoding `b'{"foo":"bar"}'.decode('utf-8') ` is useless because then the WebSocketWriter encodes it back.
Using `send_bytes` instead of `send_str` is incorrect because client must understand which type of data received.

The change only adds additional functionality.

## Checklist

### DIFF
--- a/CHANGES/4828.feature
+++ b/CHANGES/4828.feature
@@ -1,1 +1,1 @@
-Allow to send JSON as bytestring by WebSocketResponse.send_json
+Allow to send a bytestring by WebSocketResponse.send_str
--- a/CHANGES/4828.feature
+++ b/CHANGES/4828.feature
@@ -1,0 +1,1 @@
+Allow to send JSON as bytestring by WebSocketResponse.send_json
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -69,6 +69,7 @@ Dan Xu
 Daniel García
 Daniel Grossmann-Kavanagh
 Daniel Nelson
+Daniil Kozhanov
 Danny Song
 David Bibb
 David Michael Brown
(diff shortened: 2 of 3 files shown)


Gelbpunkt:
@7ckingBest Same should be done in client_ws.py line 155 :smile: Would love to see this merged!

asvetlov:
I'm very sorry but not.
`send_str()` is for sending `str` value, `send_bytes()` is for bytes. Sending bytes by `send_str()` call makes a mess; it is error-prone API.

I can imagine something like `send_lowlevel(data: bytes, binary: bool)` (the name is questionable) which sends data without encoding. 
I have no motivation to implement it myself but open to reviewing. Low-level `WebSocketWriter` is not a part of public API and can be changed wildly.

P.S.
From my understanding, you use `orjson` for the sake of performance. Are there good enough alternatives that produce str output? What is the execution time difference? What part of response time is taken by json encoding/decoding. 
In my experience, the penalty is very low. The most time is spent by DB query and application code, not by the library/framework itself.

PR #4853, opened 2020-07-08
Correct the "neither/nor" double negation in `multipart.rst`

JulienPalard:
Not sure about it but I think I found a typo in the doc, but I'm french so please review carefully.

### DIFF
--- a/docs/multipart.rst
+++ b/docs/multipart.rst
@@ -55,7 +55,7 @@ body part headers: this allows you to filter parts by their attributes::
         metadata = await part.json()
         continue
 
-Nor :class:`BodyPartReader` or :class:`MultipartReader` instances does not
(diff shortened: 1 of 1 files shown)


JulienPalard:
@codecov-commenter fake :]

docs/multipart.rst
```diff
@@ -55,7 +55,7 @@ body part headers: this allows you to filter parts by their attributes::
         metadata = await part.json()
         continue
 
-Nor :class:`BodyPartReader` or :class:`MultipartReader` instances does not
+Nor :class:`BodyPartReader` or :class:`MultipartReader` instances
```
webknjaz: I think it should be 
```suggestion
Neither :class:`BodyPartReader` nor :class:`MultipartReader` instances
```

PR #4854, opened 2020-07-10
Fix missing default value in the doc

WH-2099:
Fix the missing default value in the document function signature.
According to source code [client.py](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py) line 151, the defalut value should be `None`.

    class ClientTimeout:
        total = attr.ib(type=Optional[float], default=None)
        connect = attr.ib(type=Optional[float], default=None)
        sock_read = attr.ib(type=Optional[float], default=None)
        sock_connect = attr.ib(type=Optional[float], default=None)

## What do these changes do?
Just add a simple default value in the document function signature.

NO

## Related issue number
Because it's just a samll bug, I think there's no need to open a new issue.

### DIFF
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1504,7 +1504,7 @@ ClientTimeout
 ^^^^^^^^^^^^^
 
 .. class:: ClientTimeout(*, total=None, connect=None, \
-                         sock_connect, sock_read=None)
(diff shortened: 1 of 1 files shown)


asvetlov (approved):

asvetlov:
Thanks!

PR #4894, opened 2020-08-01
TCPSite to bind on all interfaces when None is passed in

balloob:
## What do these changes do?

When passing `None` as `host` to `TCPSite`, bind to all available interfaces instead of just IPv4. This is the default behavior of [`loop.create_server`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_server). It was never triggered because `TCPSite` was rewriting `host` to `0.0.0.0` if `None` was passed in.

The server will also be available on the IPv6 interface.

## Checklist


aiohttp/web_runner.py
```diff
@@ -97,7 +95,8 @@ def __init__(self, runner: 'BaseRunner',
     @property
     def name(self) -> str:
         scheme = 'https' if self._ssl_context else 'http'
-        return str(URL.build(scheme=scheme, host=self._host, port=self._port))
+        host = "0.0.0.0" if self._host is None else self._host
```
balloob, line 98: I made it like this to be backwards compatible, but in most cases it's incorrect, as it can listen both on IPv4 and IPv6 (if available)

balloob:
Doesn't look like the failed Cython test is related to my PR?

JonasKs:
> Doesn't look like the failed Cython test is related to my PR?

Seems to be related to #4408. Had the same in my PR.

aiohttp/web_runner.py
asvetlov: Got you. Let's keep your version

asvetlov:
Thanks!

PR #4912, opened 2020-08-13
Fix the tracing module's type hints

layday:
## What do these changes do?

`_SignalArgs` is (was) a supertype of its
arguments and with `Callable ` being _contravariant_
in its arguments, the only way this would've
passed a type check is if the callback were to
(purport to) accept every single one of the
parameter types in a union,
exactly as specified - or `object`.
Either way, the parameter types were not
associated with their corresponding signals
so it would have been possible to append
a callback with a `TraceDnsCacheMissParams`
parameter to an `on_request_start` signal.

No.

## Related issue number

—

## Checklist

### DIFF
--- a/CHANGES/4912.bugfix
+++ b/CHANGES/4912.bugfix
@@ -1,0 +1,1 @@
+Fixed the type annotations in the ``tracing`` module.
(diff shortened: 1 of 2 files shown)


asvetlov:
Thanks!

PR #4966, opened 2020-09-21
Add support for int values in serializing headers

Abstract-X:
## What do these changes do?
In the current version of aiohttp, headers are serialized only if the values ​​are concatenated to a string (i.e. headers are expected to be string pairs).
In my opinion, integers can also be accepted as a value, as is done, for example, with parameters (`params` kwarg).
https://github.com/aio-libs/yarl/blob/959e8630f512d61838d20b392e3e7a48adb05849/yarl/_url.py#L907
``` python3
import asyncio

async def main():
    url = "https://httpbin.org/get"
    headers = {
        "SomeHeader": 12345  # int
    }
        async with session.get(url, headers=headers) as response:
            pass

if __name__ == "__main__":
    asyncio.run(main())

# TypeError: can only concatenate str (not "int") to str
```

This:
``` python3
headers = {
    "X-ReferralID": self._referral_id,
    "X-GroupNumber": self._group_number,
    "X-Language": self._language
}
```
Instead of this:
``` python3
headers = {
    "X-ReferralID": str(self._referral_id),
    "X-GroupNumber": str(self._group_number),
    "X-Language": self._language
}
```

## Checklist


aiohttp/http_writer.py
```diff
@@ -157,7 +157,9 @@ async def drain(self) -> None:
 def _py_serialize_headers(status_line: str,
                           headers: 'CIMultiDict[str]') -> bytes:
     line = status_line + '\r\n' + ''.join(
-        [k + ': ' + v + '\r\n' for k, v in headers.items()])
+        [k + ': ' + (str(v) if type(v) is int else v) + '\r\n'  # no subclasses like bool
```
webknjaz, line 165: It's probably cleaner to have the conditional conversion added to a helper function.

aiohttp/http_writer.py
```diff
@@ -157,7 +157,9 @@ async def drain(self) -> None:
 def _py_serialize_headers(status_line: str,
                           headers: 'CIMultiDict[str]') -> bytes:
```
webknjaz, line 163: The change violates the typing expectation here ^

webknjaz (changes requested):
I'm not sure if this is reasonable to support. But if it is, this PR must have good test coverage for the change.

Also, the C-extension based version must behave identically to its Python fallback implementation.

P.S. This goes without saying, but this PR will remain blocked as long as the corresponding checks are failing (that is CI and missing changelog fragment).

asvetlov:
Thanks for your contribution.

After careful thinking, I'm inclining to concur with @webknjaz : this feature should be declined.
If you want to pass an integer as HTTP header value -- please use `str(number)`.

Issue #4984, opened 2020-09-30
Unix connect errors use host/port which is not useful


zealws:
🐞 **Describe the bug**
A ConnectionRefusedError using UnixConnector results in an error message such as this:

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host host:80 ssl:default [Connection refused]
```

The error message specifies the host/port of the connection. However, for Unix sockets, the host/port are not useful. They're still provided as part of the HTTP request, but they do not define the connection from client to server.

For unix sockets, ideally the error message would include the socket path on disk instead.

💡 **To Reproduce**

Run this code:
Note: don't create `/tmp/server.sock` to get a Connection Refused.
```
import aiohttp
import asyncio


async def main():
    async with aiohttp.ClientSession(
        connector=aiohttp.UnixConnector(path='/tmp/server.sock'),
    ) as session:
        r = await session.get('http://host/')
        print("Status:", r.status)
        print("Text:")
        print((await r.text()).strip())


asyncio.get_event_loop().run_until_complete(main())
```

💡 **Expected behavior**

Ideally, the exception would include the socket path on disk, which (for Unix sockets) is the primary way of identifying the connection, not host/port.

📋 **Logs/tracebacks**

Output of the sample script I pasted above:

```python-traceback
Traceback (most recent call last):
  File "/home/zeal/dev/aiohttp/venv/lib/python3.8/site-packages/aiohttp-4.0.0a1-py3.8-linux-x86_64.egg/aiohttp/connector.py", line 1112, in _create_connection
    _, proto = await self._loop.create_unix_connection(
  File "/usr/lib/python3.8/asyncio/unix_events.py", line 244, in create_unix_connection
    await self.sock_connect(sock, path)
  File "/usr/lib/python3.8/asyncio/selector_events.py", line 494, in sock_connect
    return await fut
  File "/usr/lib/python3.8/asyncio/selector_events.py", line 499, in _sock_connect
    sock.connect(address)
ConnectionRefusedError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "client.py", line 15, in <module>
    asyncio.get_event_loop().run_until_complete(main())
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "client.py", line 9, in main
    r = await session.get('http://host/')
[...]

PR #4985, opened 2020-09-30
Custom error message for Unix connect errors

zealws:
## What do these changes do?

UnixConnector client errors result in an error message which could be improved.

```
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host host:80 ssl:default [Connection refused]
```

In particular, the error message notes the host/port pair, but for UnixConnector, the socket path is the primary way of identifying the connection, not host/port pair.

This is what the new error message looks like:

```
aiohttp.client_exceptions.UnixClientConnectorError: Cannot connect to unix socket /tmp/server.sock ssl:default [Connection refused]
```

[Full error and sample code in this gist](https://gist.github.com/zfjagann/3be80baf32ff5a1131b9e7982ba34659)

UnixConnector now raises `UnixClientConnectorError` instead of `ClientConnectorError`.

`UnixClientConnectorError` is a subclass of `ClientConnectorError` so any `catch ClientConnectorError` code should still function as expected.

## Related issue number

N/A

## Checklist


zealws:
This fixes #4984

zealws:
I didn't feel like adding a unit-test was essential for this, but if you'd like me to add one, I can.

aiohttp/client_exceptions.py
```diff
@@ -152,6 +152,27 @@ class ClientProxyConnectionError(ClientConnectorError):
     """
 
 
+class UnixClientConnectorError(ClientConnectorError):
+    """Unix connector error.
+
+    Raised in :class:`aiohttp.connector.UnixConnector` if
+        connection to unix socket can not be established.
```
webknjaz: Not sure if you can omit `:py:` here. Also, indentation within the same paragraph is rather unexpected.
```suggestion
    Raised in :py:class:`aiohttp.connector.UnixConnector`
    if connection to unix socket can not be established.
```

aiohttp/client_exceptions.py
```diff
@@ -152,6 +152,27 @@ class ClientProxyConnectionError(ClientConnectorError):
...
+    """
+    def __init__(self, path: str, connection_key: ConnectionKey,
+                 os_error: OSError) -> None:
+        self._path = path
+        super().__init__(connection_key, os_error)
+
+    @property
+    def path(self) -> str:
```
webknjaz, line 170: I think I'd prefer this to be a `pathlib.Path` instance that's only stringified when used.

aiohttp/client_exceptions.py
```diff
@@ -152,6 +152,27 @@ class ClientProxyConnectionError(ClientConnectorError):
     """
 
 
+class UnixClientConnectorError(ClientConnectorError):
+    """Unix connector error.
+
+    Raised in :class:`aiohttp.connector.UnixConnector` if
+        connection to unix socket can not be established.
+    """
+    def __init__(self, path: str, connection_key: ConnectionKey,
```
webknjaz, line 164: It'd be great to expect `os.PathLike` / `pathlib.Path` for `path`.

aiohttp/client_exceptions.py
```diff
@@ -152,6 +152,27 @@ class ClientProxyConnectionError(ClientConnectorError):
...
+                 os_error: OSError) -> None:
+        self._path = path
+        super().__init__(connection_key, os_error)
+
+    @property
+    def path(self) -> str:
+        return self._path
+
+    def __str__(self) -> str:
+        return ('Cannot connect to unix socket {0.path} ssl:{1} [{2}]'
```
webknjaz, line 174: I have two comments here:
1) it's better to use named placeholders instead of positional ones
2) you could do `self.path` in args like with other vars, otherwise, it looks inconsistent

webknjaz (commented):
It'd be great to add some integration tests + check how this works with abstract sockets.

zealws:
Thanks for the comments @webknjaz . I'll update this tomorrow per your suggestions.

I can add an integration test for a simple unix socket. However, I'm not sure exactly what you mean by "abstract sockets". Do you have some documentation on the topic you can refer me to?

webknjaz:
@zealws it's a Linux-only feature: https://utcc.utoronto.ca/~cks/space/blog/linux/SocketAbstractNamespace

aiohttp/client_exceptions.py
asvetlov: We don't use autodoc for documentation generation; docstrings should be plain texts.

asvetlov (commented):
Looks good in general.
@webknjaz please feel free to merge when ready.
I think the PR can be backported to 3.7 branch (label + bot works fine)

aiohttp/client_exceptions.py
asvetlov: UnixConnector historically works with `str` and has `.path` string property.
I think we should keep the type for consistency.

asvetlov:
I'll fix the black formatting error by the next commit.

PR #4986, opened 2020-10-01
Correct a spelling mistake in quickstart: "canonize"->"canonicalize"

vuzdemav:
## What do these changes do?

Fix spelling in the client-quickstart file.

"canonize" means to declare a person to be a saint - https://en.wikipedia.org/wiki/Canonization

You mean "canonicalize" - https://en.wikipedia.org/wiki/Canonicalization

no


webknjaz (approved):
Thanks!

PR #5012, opened 2020-10-14
Fix closing connection after HEAD request

### LINKED ISSUE #1769: Connections are not correctly kept after HEAD requests
When the client makes a HEAD request, it parses the response as if it was a regular message.

But the HTTP specs has a nice trick: HEAD requests `SHOULD` respond with the exact same headers than the corresponding GET request. This includes `Content-Length` or `Transfer-Encoding: chunked` headers.

This means that when the client receives a HEAD response with a `Content-Length: 5` response, it should not expect any body and `payload.is_eof()` should be True immediately, not after receiving 5 more bytes.

Currently `ResponseHandler.should_close` is True for HEAD requests with a `Content-Length: …` because `payload.is_eof()` is False when it should be True, hence the underlying connection is closed and not re-used when it could be.

This is only inefficient on the client side, not at all a bug, and does not produce any wrong results.

(also, I was mistaken a few minutes ago: returning a Content-Length header with an empty body is perfectly valid, and the `allow_head=True` parameter of the `add_get` method works perfectly well, no problem at all on the server side)

reproducer:
==========

https://github.com/aio-libs/aiohttp/issues/1769#issuecomment-614775045

leejeonghun:
## What do these changes do?

In order to prevent unintended connection close in HEAD request, make payload type to `EmptyStreamReader` on HEAD request.

After HEAD request, connection could be reuse correctly.

## Related issue number

Resolves #1769

## Checklist


tests/test_client_functional.py
```diff
@@ -54,6 +54,36 @@ async def handler(request):
...
 
+async def test_keepalive_after_head_requests_success(
+        aiohttp_client) -> None:
+    async def handler(request):
+        body = await request.read()
+        assert b'' == body
+        return web.Response(body=b'OK')
+
+    cnt_conn_reuse = 0
+    async def on_reuseconn(session, ctx, params):
```
asvetlov:
```suggestion
    cnt_conn_reuse = 0

    async def on_reuseconn(session, ctx, params):
```
This change fixes the linter.
The PR should have green checkers

asvetlov (changes requested):
Please fix linter

asvetlov (approved):
Awesome, thanks!

PR #5065, opened 2020-10-17
Customize the input buffer size

### LINKED ISSUE #4453: "Line is too long" for JSON-lines from Kubernetes API

asvetlov:
Fixes #4453

### DIFF
--- a/CHANGES/4453.feature
+++ b/CHANGES/4453.feature
@@ -1,0 +1,1 @@
+Allow configuring the sbuffer size of input stream by passing ``read_bufsize`` argument.
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -567,11 +567,10 @@ cdef class HttpParser:
 
 cdef class HttpRequestParser(HttpParser):
 
-    def __init__(self, protocol, loop, timer=None,
+    def __init__(self, protocol, loop, int limit, timer=None,
                  size_t max_line_size=8190, size_t max_headers=32768,
                  size_t max_field_size=8190, payload_exception=None,
                  bint response_with_body=True, bint read_until_eof=False,
-                 int limit=2**16,
     ):
          self._init(cparser.HTTP_REQUEST, protocol, loop, limit, timer,
                     max_line_size, max_headers, max_field_size,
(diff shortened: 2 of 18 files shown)


tseaver:
Making `limit` a required positional argument (removing the default) is a breaking change:  see https://github.com/pnuckowski/aioresponses/issues/173

asvetlov:
Sorry but no. `StreamReader.__init__` is a private API, user code should never call it.
`aioresponses` mocks aiohttp internals for its work.
I cannot prevent it (but discourage, sure).
The breakage of such libraries when aiohttp changes own internal undocumented API is the expected thing unfortunately.

tseaver:
I'm not sure I understand how the constructor of a public class (`aiohttp.streams.StreamReader` is even imported into `aiohttp/__init__.py`!) can be considered "private".  I do see the note in the docs that users aren't to instantiate the class, but that is a very unusual claim.

asvetlov:
Sorry for that

PR #5072, opened 2020-10-18
Fix multipart file uploads without content type

### LINKED ISSUE #4089: File uploads without a content type are treated as ordinary fields

asvetlov:
What do these changes do?
----------------------------------------

Multipart form data with files without a content type were not treated correctly anymore by the aiohttp server.

Are there changes in behavior for the user?
------------------------------------------------------------------------------

This restores the behavior as it was up to aiohttp 3.5.4; for aiohttp 3.6.0 and aiohttp 3.6.1 it would treat such file uploads as normal fields, including trying to utf8-decode the data.

Fixed conflicts of #4090

Fixes #4089.


asvetlov:
Thanks, @sjaensch !

PR #5095, opened 2020-10-20
Add keepalive_timeout parameter to web.run_app

### LINKED ISSUE #5094: Allow configuring the keepalive timeout in web.run_app
🐣 **Is your feature request related to a problem? Please describe.**

Adjusting the keepalive timeout is not possible without copy-pasting a lot of boilerplate code usually supplied by the `web.run_app` function.

💡 **Describe the solution you'd like**

Add a `keepalive_timeout` parameter to the `web.run_app` function.

Teyras:
## What do these changes do?

Add keepalive_timeout parameter to web.run_app. Closes #5094.

Unlikely.

## Checklist


tests/test_run_app.py
```diff
@@ -685,6 +686,24 @@ async def on_startup(app):
     exc_handler.assert_called_with(patched_loop, msg)
 
 
+def test_run_app_keepalive_timeout(patched_loop):
```
webknjaz: It's usually better to use the [`mocker`](https://pypi.org/project/pytest-mock/#usage) fixture from `pytest-mock` (it's already integrated as a test dependency) for accessing the `MagicMock` class.

tests/test_run_app.py
```diff
@@ -685,6 +686,24 @@ async def on_startup(app):
...
+    new_timeout = 1234
+    base_runner_init_mock = mock.MagicMock()
+    base_runner_init_orig = BaseRunner.__init__
+
+    def base_runner_init_spy(self, *args, **kwargs):
+        base_runner_init_mock(*args, **kwargs)
+        base_runner_init_orig(self, *args, **kwargs)
+
+    with mock.patch.object(BaseRunner, "__init__", base_runner_init_spy):
```
webknjaz: It's better to use the built-in [`monkeypatch` fixture](https://docs.pytest.org/en/stable/monkeypatch.html) for patching things.
Also, it's best to only patch the line that relies on patching, not many lines because you may accidentally create unintended side-effects with this.

tests/test_run_app.py
```diff
@@ -685,6 +686,24 @@ async def on_startup(app):
...
 
 
+def test_run_app_keepalive_timeout(patched_loop):
+    new_timeout = 1234
+    base_runner_init_mock = mock.MagicMock()
+    base_runner_init_orig = BaseRunner.__init__
+
+    def base_runner_init_spy(self, *args, **kwargs):
+        base_runner_init_mock(*args, **kwargs)
+        base_runner_init_orig(self, *args, **kwargs)
```
webknjaz, line 834: Have you tried this?
```suggestion
        super().__init__(*args, **kwargs)
```

Or maybe apply a decorator approach.

tests/test_run_app.py
```diff
@@ -685,6 +686,24 @@ async def on_startup(app):
...
+
+    with mock.patch.object(BaseRunner, "__init__", base_runner_init_spy):
+        app = web.Application()
+        web.run_app(app, keepalive_timeout=new_timeout,
+                    print=stopper(patched_loop))
+
+    base_runner_init_kwargs = base_runner_init_mock.call_args[1]
+    assert (base_runner_init_kwargs["keepalive_timeout"] == new_timeout)
```
webknjaz: assert is not a function and so it does not require parentheses: 
```suggestion
    assert base_runner_init_kwargs["keepalive_timeout"] == new_timeout
```

tests/test_run_app.py
```diff
@@ -685,6 +686,24 @@ async def on_startup(app):
     exc_handler.assert_called_with(patched_loop, msg)
 
 
+def test_run_app_keepalive_timeout(patched_loop):
+    new_timeout = 1234
+    base_runner_init_mock = mock.MagicMock()
+    base_runner_init_orig = BaseRunner.__init__
+
+    def base_runner_init_spy(self, *args, **kwargs):
+        base_runner_init_mock(*args, **kwargs)
```
webknjaz: It looks like it'd be much cleaner to just do
```suggestion
        assert kwargs["keepalive_timeout"] == new_timeout
```

docs/web_reference.rst
```diff
@@ -2793,6 +2794,10 @@ Utilities
                                 timeout but closes a server in a few
                                 milliseconds.
 
+   :param float keepalive_timeout: a delay before a TCP connection is
```
webknjaz, line 2797: The argument name is ambiguous on this level because `run_app()` doesn't suggest that it has anything to do with TCP. I'd probably call the user-facing interface `tcp_keepalive_timeout`.

webknjaz (commented):
See the initial review from me below. It's non-blocking but I don't have the confidence to accept this change on my own either and thus I defer the decision on whether it's something that we want to @asvetlov.

tests/test_run_app.py
janbuchar: The assert used to be broken across two lines and the parentheses remained there.

tests/test_run_app.py
janbuchar: Sure, I can update it. What exactly is the benefit of using the fixture?

docs/web_reference.rst
janbuchar: Well, the function already leaks this with the `port` parameter. I'm not even sure we can say that HTTP keepalive is restricted to TCP - couldn't any transport layer protocol with a handshake be used here?

tests/test_run_app.py
webknjaz: Just the consistency: if the lib is integrated, the expectation is that it would be used. Also, pytest design is more fixture-driven so external things seem foreign.

docs/web_reference.rst
webknjaz: My point is that HTTP keep-alive and TCP keep-alive are different things. And so it's not clear which one it is just based on the name.

tests/test_run_app.py
janbuchar: Got it. I can change this (and the rest of the comments) if @asvetlov agrees with the feature itself.

tests/test_run_app.py
janbuchar: Just to be clear, do you mean writing a custom decorator to encapsulate those ~three variables?

docs/web_reference.rst
janbuchar: I'm trying to add support for changing HTTP keepalive timeout, so prefixing it with `tcp_` would only make things worse.

asvetlov: Yes, the value is for http keepalive, not tcp keepalive. Let's keep it

asvetlov (approved):

docs/web_reference.rst
```diff
@@ -2793,6 +2794,10 @@ Utilities
...
+                                   closed after a HTTP request. The delay
+                                   allows for reuse of a TCP connection.
+
```
asvetlov, line 2800:
```suggestion

      .. versionadded:: 3.7

```

asvetlov: Please apply the marker

asvetlov:
@webknjaz please feel free to merge the PR when you are okay with it.

docs/web_reference.rst
```diff
@@ -2798,6 +2798,8 @@ Utilities
                                    closed after a HTTP request. The delay
                                    allows for reuse of a TCP connection.
 
+      .. versionadded:: 3.7
```
```suggestion
      .. versionadded:: 3.8
```
Sorry, the time window for 3.1 is closed.

janbuchar: Understood.

janbuchar:
@asvetlov is there any agreed-upon autoformatting tool that could help me resolve the conflicts?

asvetlov:
You can try the following:
1. Copy https://github.com/aio-libs/aiohttp/blob/master/.pre-commit-config.yaml to your local repository.
2. Install pre-commit: `pip install -U pre-commit`
3. Apply all hooks: `pre-commit run --all-files` Repeat the step several times until all checks are green.
4. I hope your local copy will be in sync after reformatting; git merge upstream/master should be succeeded. If not -- the only manual conflict resolving can help

tests/test_run_app.py
```diff
@@ -824,6 +825,19 @@ async def on_startup(app):
...
+    new_timeout = 1234
+    base_runner_init_orig = BaseRunner.__init__
+
+    def base_runner_init_spy(self, *args, **kwargs):
+        assert kwargs["keepalive_timeout"] == new_timeout
+        base_runner_init_orig(self, *args, **kwargs)
+
+    app = web.Application()
+    with mock.patch.object(BaseRunner, "__init__", base_runner_init_spy):
+        web.run_app(app, keepalive_timeout=new_timeout, print=stopper(patched_loop))
```
webknjaz:
```suggestion
    monkeypatch.setattr(BaseRunner, "__init__", base_runner_init_spy)
    web.run_app(app, keepalive_timeout=new_timeout, print=stopper(patched_loop))
```

tests/test_run_app.py
```diff
@@ -824,6 +825,19 @@ async def on_startup(app):
     exc_handler.assert_called_with(patched_loop, msg)
 
 
+def test_run_app_keepalive_timeout(patched_loop, mocker):
```
webknjaz:
```suggestion
def test_run_app_keepalive_timeout(patched_loop, mocker, monkeypatch):
```

webknjaz (approved):

asvetlov:
@webknjaz please backport manually next time if the bot fails. 
Or, at least, let me know that I should do it myself

webknjaz:
Oh, I probably missedmisse that among my notifications. I think it's also a good idea to encourage the submitters to do their backports when they are as interested in the change.

asvetlov:
> I think it's also a good idea to encourage the submitters to do their backports when they are as interested in the change.

Agree, ask for the submitter to make backport PR using `cherry_picker` tool (already installed by `requirements/dev.txt`) is a good idea.
I think the person who merges a PR should monitor Backport statuses and try to manually backport himself on failure.
It works well pretty often when conflict resolution is trivial.
In a complex case, the merger can kindly ask for manual backporting.

asvetlov:
For the note, backported to 3.8 by e88dc8a0

PR #5145, opened 2020-10-25
Simplify Makefile and requirements

asvetlov:


.pre-commit-config.yaml
```diff
@@ -27,3 +22,17 @@ repos:
...
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: 'v3.3.0'
+  hooks:
+  - id: file-contents-sorter
+    files: docs/spelling_wordlist.txt
+- repo: https://github.com/asottile/pyupgrade
+  rev: 'v2.7.3'
+  hooks:
+  - id: pyupgrade
+    args: ['--py36-plus']
```
webknjaz, line 37: Putting the formatters at the end is usually a bad idea. Because the linters will then check the contents as they were before formatting making it rather inconsistent.

asvetlov: Good point.

asvetlov: I've pushed flake8 at the last position, all previous hooks eighter modifies python source or don't check them

webknjaz: Oh, and another piece of advice: you can split the hooks from the same repo into two and put in different places if needed. Within the blocks of mutating and non-mutating checks, I normally sort the individual hooks so that the fastest ones would go first. This maximizes the responsiveness of the process.

webknjaz: One more thing: you can have a separate pre-commit config for very heavy checks and would be nice to always run in CI but users would only invoke them on-demand.

Makefile
```diff
@@ -20,59 +20,24 @@ cythonize: .install-cython $(PYXS:.pyx=.c)
 	pip install -r requirements/dev.txt
 	@touch .install-deps
 
-.PHONY: lint
-lint: isort-check black-check flake8 mypy
+.PHONY: pre-commit
+pre-commit:
+	pre-commit run --all-files
```
webknjaz: Here's a recommended invocation example: https://github.com/ansible/pylibssh/blob/bd53dd8/tox.ini#L236. It's especially useful to have `--show-diff-on-failure` in CI, less so locally but it also comes in handy to review  what mutations have been applied.
Also, dunno how it's possible with make but in tox it's common to allow end-users to bypass their own args to pre-commit so while `tox -e lint` runs `pre-commit run --all-files`, `tox -e lint -- flake8 --files x/y/z` would run `pre-commit run flake8 --files x/y/z` which is useful when trying to go through what a specfic linter reports separately in a more responsive manner.

asvetlov: Done, thanks!

.pre-commit-config.yaml
asvetlov: Got you

PR #5146, opened 2020-10-25
tests - add test for ResourceWarning: unclosed transport <asyncio.sslproto._SSLProtocolTransport object …

commonism:
## What do these changes do?

Add a test for 
`ResourceWarning: unclosed transport <asyncio.sslproto._SSLProtocolTransport object at …`
when using aiohttp.ClientSession with TLS

No

## Related issue number

#4703
#4713


tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,156 @@
...
+
+import logging
+
+import tracemalloc
+tracemalloc.start(40)
+
+
+log = logging.getLogger()
+log.setLevel(logging.DEBUG)
+logging.basicConfig(level=logging.WARNING)
```
webknjaz: It's pretty uncommon to have random logs in tests. Also, such a setup is going to influence the whole runtime which is bad.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,156 @@
...
+        while True:
+            try:
+                for s in list(runner.sites):
+                    log.info(f"Restarting Site {s.name}")
+                    await s.stop()
+                    await s.start()
+            except Exception as e:
+                log.exception(e)
+            finally:
+                await asyncio.sleep(5)
```
webknjaz: Making each test invocation 5 seconds slower doesn't sound like a great idea.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,156 @@
...
+
+def create_setUp_tearDown(s):
+    async def _setUp(app):
+        asyncio.ensure_future(s.asyncSetUp())
+    async def _tearDown(app):
+        asyncio.ensure_future(s.asyncTearDown())
+    return _setUp, _tearDown
+
+
+class TestUnclosedTransport(unittest.IsolatedAsyncioTestCase):
```
webknjaz: Inheriting from `unittest.TestCase` and similar prevents pytest from using its features and must be avoided.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,156 @@
...
+    @routes.get('/{other}', name='other')
+    async def handle_other(self, request):
+        pass
+
+    @routes.get('/', name='root')
+    async def handle_root(self, request):
+        import random
+        off = random.randint(7,17)
+        await asyncio.sleep(off/10)
+        return web.Response(text=f'{off}', status=200)
```
webknjaz: What is this supposed to do exactly?

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,156 @@
...
+        log.info(f"{self.__class__.__qualname__}.setUp")
+        self.loop = asyncio.get_event_loop()
+        self.app = app = web.Application()
+
+        self.service = Service.create(app)
+
+        runner = web.AppRunner(app)
+        self.runner = runner
+        self.server_ctx = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
+        self.server_ctx.load_cert_chain('data/service-certificates.pem', 'data/service-key.pem')
```
webknjaz: Referencing non-existent files isn't going to work. Also, hardcoding pre-generated files isn't great either. In this test suite, there's good examples of making temporary TLS certificates: we use `trustme` lib for this but you don't need to use it directly because there already a set of pytest fixtures that are ready-to-use — use them instead. Usage example: https://github.com/aio-libs/aiohttp/blob/32833c3/tests/test_client_functional.py#L295-L297.

webknjaz:
@commonism Git setup on your computer looks broken: none of the submitted commits list your GitHub account as an author. You should probably configure Git locally or verify the email that is embedded in those commits in your GitHub account.

webknjaz (changes requested):
Thanks for the effort! Unfortunately, this doesn't look ready, nor is compliant with our test writing practices.

Also, there's great prior art on the usage of pytest fixtures including proper TLS setup — https://github.com/aio-libs/aiohttp/pull/4713/files — why haven't you start there? It also uses proper loop isolation, and pytest-based warnings tracking.

Besides, adding a custom server/client implementations doesn't seem related to the problem you're attempting to reproduce — it only makes the test module messy and distracts from what's actually being tested.

asvetlov (commented):
@commonism I respect your striving to point on the very annoying issue.
It will be never fixed on aiohttp 3.x branch, sorry.
aiohttp 4.0 has partially done work for getting rid of the warning. The missing part is #5102

P.S.
aiohttp uses `pytest` with `pytest-aiohttp` plugin, not the standard `unittest`.
This is the deliberate choice. We don't accept tests that are `unittest` based.

P.P.S.
I'm aware of `unittest.IsolatedAsyncioTestCase` because I had added it to the CPython standard library.
It exists because CPython itself cannot depend on `pytest` project for running self-testing.
For third-party libs I always prefer pytest-style.

tests/test_ssl_transport_close.py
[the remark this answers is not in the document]
commonism: Will you accept the test if the defects outlined get resolved?

tests/test_ssl_transport_close.py
commonism: Force a connection_lost.

tests/test_ssl_transport_close.py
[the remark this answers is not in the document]
asvetlov: Do you mean `pytest` usage also?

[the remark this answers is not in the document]
commonism: Yes, including everything mentioned so far.

commonism (commented):
I addressed the following issues

 - pytest, not a unittest
 - no logging/print
 - no tracemalloc
 - reduce time required to 5s (was 30 seconds actually)
 - using the internal trustme CA

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+import asyncio
+import warnings
+
+
+async def handle_root(request):
+    """
+    idle 2 seconds so service is restarted meanwhile, forcing connection_lost
+    """
+    await asyncio.sleep(2)
+    return web.Response(text='', status=200)
```
webknjaz: status is redundant
```suggestion
    return web.Response(text='')
```

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    while been_restarted <= 2:
+        try:
+            await session.get('https://localhost:12345/')
+        except aiohttp.client.ClientConnectorError:
+            continue
+        finally:
+            await asyncio.sleep(0.1)
+
+        assert [warnings.formatwarning(w.message, w.category, w.filename, w.lineno, w.line)
+                for w in filter(_ssl_resource_warnings, recwarn)] == []
```
webknjaz: It seems the check you're after is
```suggestion
        assert not len(filter(_ssl_resource_warnings, recwarn))
```
I don't see any value in converting warning objects that you are only going to count and throw away.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+            await asyncio.sleep(0.5)
+
+def _ssl_resource_warnings(w):
+    if w.category == ResourceWarning and w.filename.endswith('sslproto.py') and \
+        w.message.args[0].startswith('unclosed transport <asyncio.sslproto._SSLProtocolTransport object at'):
+        return w
+    return None
+
+async def _run(loop, recwarn, ssl_ctx, client_ssl_ctx):
+    global been_restarted
```
webknjaz: Globals must be avoided per the same concerns as before: mutating global state makes tests unisolated.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+        except Exception as e:
+            print(e)
+        finally:
+            await asyncio.sleep(0.5)
+
+def _ssl_resource_warnings(w):
+    if w.category == ResourceWarning and w.filename.endswith('sslproto.py') and \
+        w.message.args[0].startswith('unclosed transport <asyncio.sslproto._SSLProtocolTransport object at'):
+        return w
+    return None
```
webknjaz: Aren't filter predicates [supposed to return bool](https://docs.python.org/3/library/functions.html#filter)?

webknjaz: So the proper implementation would be
```suggestion
    unclosed_transport_msg = 'unclosed transport <asyncio.sslproto._SSLProtocolTransport object at'
    return (
        w.category == ResourceWarning
        and w.filename.endswith('sslproto.py')
        and w.message.args[0].startswith(unclosed_transport_msg)
    )
```

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    idle 2 seconds so service is restarted meanwhile, forcing connection_lost
+    """
+    await asyncio.sleep(2)
+    return web.Response(text='', status=200)
+
+
+async def _prepare(ssl_ctx, client_ssl_ctx):
+    app = web.Application()
+    app.router.add_get('/', handle_root)
+    runner = web.AppRunner(app)
```
webknjaz: It's better not to reimplement the server initialization. Just use our native [`aiohttp_server` fixture](https://github.com/aio-libs/aiohttp/blob/f8df85d/aiohttp/pytest_plugin.py#L261). It at least will get you a server with an app runner set up and you can probably extend it with a tcp site on top.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+        finally:
+            await asyncio.sleep(0.1)
+
+        assert [warnings.formatwarning(w.message, w.category, w.filename, w.lineno, w.line)
+                for w in filter(_ssl_resource_warnings, recwarn)] == []
+
+    await _close(app, runner, session)
+
+
+def test_run(recwarn, ssl_ctx, client_ssl_ctx):
```
webknjaz: This test must have some meaningful name telling the reader that it is actually for.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    global been_restarted
+    while True:
+        try:
+            for s in list(runner.sites):
+                await s.stop()
+                await s.start()
+            been_restarted += 1
+        except Exception as e:
+            print(e)
```
webknjaz: What is suppressing all the exceptions for?

commonism (commented):
addressing the issues.

webknjaz:
@commonism we also need the CI to actually execute this test. Red linting job prevents that from happening.

webknjaz:
@commonism P.S. your local Git setup is still broken

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    return web.Response(text='')
+
+
+async def _prepare(aiohttp_server, ssl_ctx, client_ssl_ctx, cq, dq):
+    app = web.Application()
+    app['cq'] = cq
+    app['dq'] = dq
+    app.router.add_get('/', handle_root)
+    server = await aiohttp_server(app, port=12345, ssl=ssl_ctx)
+    asyncio.ensure_future(_restart(server.runner, ssl_ctx, cq))
```
webknjaz: I don't get why you need to schedule a task on the loop and not wait for it. You're basically creating a synthetic race condition that will likely make the testing indeterministic.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    await dq.put(i)
+    return web.Response(text='')
+
+
+async def _prepare(aiohttp_server, ssl_ctx, client_ssl_ctx, cq, dq):
+    app = web.Application()
+    app['cq'] = cq
+    app['dq'] = dq
+    app.router.add_get('/', handle_root)
+    server = await aiohttp_server(app, port=12345, ssl=ssl_ctx)
```
webknjaz: It's better to use an ephemeral port 0. You can then get a client to whatever the actual port ends up being via [`aiohttp_client` fixture](https://github.com/aio-libs/aiohttp/blob/f8df85d/aiohttp/pytest_plugin.py#L334) and you don't need to care about managing the client session manually.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+    await session.close()
+
+async def _restart(runner, ssl_ctx, cq):
+    """
+    restart service to force connection_lost
+    """
+    while True:
+        for s in list(runner.sites):
+            await s.stop()
+            s = web.TCPSite(runner, '127.0.0.1', port=12345, ssl_context=ssl_ctx)
```
webknjaz: What is the point of creating a new site object on the same port?

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            await session.get('https://localhost:12345/')
+        except aiohttp.client.ClientConnectorError:
+            continue
+        finally:
+            await asyncio.sleep(0.1)
+
+        assert not len(list(filter(_ssl_resource_warnings, recwarn)))
+
+    await _close(server, session)
```
webknjaz: Since this is not executed in a `finally:` block, it won't get executed in case of assertion errors potentially messing up external tests or anything that hardcodes its port to 12345.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,83 @@
...
+
+async def _run(loop, aiohttp_server, recwarn, ssl_ctx, client_ssl_ctx, cq, dq):
+    """
+    run for two processed client requests
+    """
+    server, session = await _prepare(aiohttp_server, ssl_ctx, client_ssl_ctx, cq, dq)
+
+    while dq.qsize() <= 2:
+        try:
+            await session.get('https://localhost:12345/')
```
webknjaz: Don't hardcode hostname/port. Just use a built-in test client from the fixture I've mentioned before.

tests/test_ssl_transport_close.py
commonism: closing the site and re-opening the site to force a client side connection_lost

tests/test_ssl_transport_close.py
commonism: using port 0 now

commonism:
> @commonism we also need the CI to actually execute this test. Red linting job prevents that from happening.

The Red Linter is gone, now the tests fail after 2 minutes - timeout?

commonism:
Rebased, addressed the authorship issues to have CLAassistant compliance, force pushed.

"Timeline protection" is missing, following the Details, the link https://camo.githubusercontent.com/81006649ffdf2304b123fbe39fca080c91f9aef3e8e4d33a0958e88c04d41a16/68747470733a2f2f7468656576656e746368726f6e69636c652e636f6d2f77702d636f6e74656e742f75706c6f6164732f323031342f31302f7661746c6962372e6a7067 is "Non-Image content-type returned"

Any advice?

commonism:
This is still an issue.
How to proceed to get this test merged.
Maybe as pytest.mark.fail?

Dreamsorcerer:
We've got some stricter type checking these days. If you could fix the mypy errors and add an xfail, then I'll make sure this gets merged.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,118 @@
+import asyncio
+
+import aiohttp
```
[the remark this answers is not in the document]
Dreamsorcerer: Could probably just import the 2 things we need from the module here.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+async def _prepare(
+    aiohttp_server: Any,
+    aiohttp_client: Any,
+    ssl_ctx: Any,
+    client_ssl_ctx: Any,
+    cq: asyncio.Queue[int],
+    dq: asyncio.Queue[int],
+) -> Tuple[TestServer, TestClient]:
```
Dreamsorcerer, line 36: I know many of our tests are still labelled with `Any`, but I'm in the process of cleaning it up. Could you add in the correct annotations here? Let me know if there's anything you can't find the type of.

Off the top of my head, TestClient/TestServer are in aiohttp.test_utils, ssl_ctx is `ssl.SSLContext`, and any other fixtures can be checked in the source code (they should have the return type annotated correctly).

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,144 @@
...
+            await asyncio.sleep(0.1)
+
+        assert not len(
+            list(filter(_ssl_resource_warnings, recwarn))
+        ), "unclosed transport"
+
+    await _close(server, session)
+
+
+@pytest.mark.xfail
```
Dreamsorcerer:
```suggestion
@pytest.mark.xfail(reason="Depends on #5102")
```

tests/test_ssl_transport_close.py
commonism: try regexxer for the cleanup.

Current
```
AiohttpClient = Callable[[Union[Application, BaseTestServer]], Awaitable[TestClient]]
AiohttpServer = Callable[[Application], Awaitable[TestServer]]
```
do not have the kwargs parameters for port & ssl_ctx, which is why I have to declare them as Any in one location.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,151 @@
...
+    cq: asyncio.Queue[int],
+    dq: asyncio.Queue[int],
+) -> None:
+    """run for two processed client requests"""
+    server, session = await _prepare(
+        aiohttp_server, aiohttp_client, ssl_ctx, client_ssl_ctx, cq, dq
+    )
+    if server.runner is None or session.port is None:
+        # mypy | None required
+        return
```
Dreamsorcerer:
```suggestion
    assert server.runner is not None and session.port is not None
```

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,149 @@
...
+        try:
+            jobs: Any = []
+            jobs.append(session.get("/"))
+            jobs.append(_restart(server.runner, ssl_ctx, session.port, cq, dq))
+            await asyncio.gather(*jobs)
+        except ServerDisconnectedError:
+            """
+            restarting the service will cause the client connections to fail
+            this is expected and no failure
+            """
```
Dreamsorcerer:
```suggestion
            # Restarting the service will cause the client connections to fail
            # this is expected and not a failure.
```

Dreamsorcerer:
```
E               Traceback (most recent call last):
E                 File "/opt/hostedtoolcache/Python/3.11.4/x64/lib/python3.11/warnings.py", line 537, in _warn_unawaited_coroutine
E                   warn(msg, category=RuntimeWarning, stacklevel=2, source=coro)
E               RuntimeWarning: coroutine 'aiohttp_server.<locals>.finalize' was never awaited
```

Because it's a cleanup thing, the error appears in the next test, but it's caused by this test.

Dreamsorcerer:
Seemingly only on Python 3.11+ for some reason though.

commonism:
Should be fixed

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,148 @@
...
+    await _restart(server.runner, ssl_ctx, session.port, cq, dq)
+    await cq.get()
+
+    return server, session
+
+
+async def _close(server: TestServer, session: TestClient) -> None:
+    await server.close()
+    await session.close()
+    await server
```
Dreamsorcerer: I don't think that fixes the issue, but somehow masks it.
```
TypeError: object TestServer can't be used in 'await' expression
```

Dreamsorcerer: I'm also seeing:
```
Exception ignored in: <coroutine object aiohttp_client.<locals>.finalize at 0x7fe61a202340>
  File "<string>", line 1, in <lambda>
KeyError: '__import__'
Exception ignored in: <coroutine object aiohttp_client.<locals>.finalize at 0x7fe61a202340>
  File "<string>", line 1, in <lambda>
KeyError: '__import__'
```
Which must be related to the problem.

Dreamsorcerer: Think I've got a solution. The server/client already auto-close in the fixture, but the test was mixing up loops.

Dreamsorcerer: So, actually, the test passes on 3.11+, which is why the warning didn't happen on the other tests.

Dreamsorcerer:
So, if the test is passing on 3.11+, then there's probably nothing for us to fix here (given the complexity of the problem is going to make it not worth fixing on older versions).

Dreamsorcerer:
Hmm, there's a TimeoutError in 3.12 though...
Not sure what's happening there. It also passes on pypy.

Dreamsorcerer:
Don't suppose you have any ideas about the TimeoutError on 3.12?
Once we figure that out we can merge.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,141 @@
...
+    await dq.get()
+    dq.task_done()
+    for s in list(runner.sites):
+        await s.stop()
+        await cq.put(0)
+        s = web.TCPSite(
+            runner, "127.0.0.1", port=port, ssl_context=ssl_ctx, shutdown_timeout=1.0
+        )
+        await s.start()
+        break
```
Dreamsorcerer: Just noticed this one in the codecov.
The break means this never iterates, so there's no point of a loop here.

Dreamsorcerer: If the intention is just to use any site from the set, then try `site = next(iter(runner.sites))`.

commonism: We only have 1 site.

`site = next(iter(runner.sites))` would be limited to 1 site as well?

Dreamsorcerer: Yes, my point is that, if we are not iterating, then we don't need a loop. Which will also clear up coverage, which highlights the for loop as a partial, because it only runs once.

Dreamsorcerer:
```suggestion
    site = next(iter(runner.sites))
    await site.stop()
    await cq.put(0)
    site = web.TCPSite(
        runner, "127.0.0.1", port=port, ssl_context=ssl_ctx, shutdown_timeout=1.0
    )
    await site.start()
```

commonism:
> Hmm, there's a TimeoutError in 3.12 though... Not sure what's happening there. It also passes on pypy.

Actually the timer got canceld?

```
self = <aiohttp.helpers.TimerContext object at 0x7ff31767dac0>
exc_type = <class 'asyncio.exceptions.CancelledError'>
exc_val = CancelledError(), exc_tb = <traceback object at 0x7ff3192e1d00>

    def __exit__(
        self,
        exc_type: Optional[Type[BaseException]],
        exc_val: Optional[BaseException],
        exc_tb: Optional[TracebackType],
    ) -> Optional[bool]:
        if self._tasks:
            self._tasks.pop()  # type: ignore[unused-awaitable]
    
        if exc_type is asyncio.CancelledError and self._cancelled:
>           raise asyncio.TimeoutError from None
E           TimeoutError

exc_tb     = <traceback object at 0x7ff3192e1d00>
exc_type   = <class 'asyncio.exceptions.CancelledError'>
exc_val    = CancelledError()
self       = <aiohttp.helpers.TimerContext object at 0x7ff31767dac0>

aiohttp/helpers.py:723: TimeoutError
```

https://github.com/aio-libs/aiohttp/actions/runs/5978392347/job/16220898181

Dreamsorcerer:
> Actually the timer got canceld?

Hmm, we switched to asyncio.timeout() recently. So, I've merged master and will see if that helps.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,140 @@
...
+            jobs: Any = []
+            jobs.append(session.get("/"))
+            jobs.append(_restart(server.runner, ssl_ctx, session.port, cq, dq))
+            await asyncio.gather(*jobs)
+        except ServerDisconnectedError:
+            # Restarting the service will cause the client connections to fail
+            # this is expected and not a failure.
+            pass
+        finally:
+            await asyncio.sleep(0.1)
```
commonism, line 106: I guess it's the 300s ClientTimeout

        real_timeout = ClientTimeout(total=300, connect=None, sock_read=None, sock_connect=None, ceil_threshold=5)

300.26s call     tests/test_ssl_transport_close.py::test_unclosed_transport_asyncio_sslproto_SSLProtocolTransport[pyloop]


increasing the sleep to something larger to have to server restart in time?

Dreamsorcerer: I'm not seeing how the sleep relates to it, but we can try it if you think that's the problem.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,140 @@
...
+    ssl_ctx: ssl.SSLContext,
+    port: int,
+    cq: "asyncio.Queue[int]",
+    dq: "asyncio.Queue[int]",
+) -> None:
+    """restart service to force connection_lost"""
+    await dq.get()
+    dq.task_done()
+    site = next(iter(runner.sites))
+    await site.stop()
```
commonism, line 64: In 3.12 it hangs here - you do not get 

> DEBUG    aiohttp.server:web_protocol.py:434 Ignored premature client disconnection

as you get in 3.11

Dreamsorcerer: I don't see anything suspicious in there, unless the asyncio server is not closing:
https://github.com/commonism/aiohttp/blob/ebe2a48c282f2484ee1766af5c4d52ff3431b43a/aiohttp/web_runner.py#L83

Everything else looks like it has a 60s timeout, so would fail before the 5 min timeout we're seeing.

Dreamsorcerer: If it is the asyncio server not closing, then we should get a cpython bug filed for the regression.

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,140 @@
...
+    await cq.put(0)
+    site = web.TCPSite(
+        runner, "127.0.0.1", port=port, ssl_context=ssl_ctx, shutdown_timeout=1.0
```
Dreamsorcerer:
```suggestion
        runner, "127.0.0.1", port=port, ssl_context=ssl_ctx
```

tests/test_ssl_transport_close.py
```diff
@@ -0,0 +1,138 @@
...
+
+        assert not len(
+            list(filter(_ssl_resource_warnings, recwarn))
+        ), "unclosed transport"
+
+
+@pytest.mark.xfail(
+    sys.version_info < (3, 11) and platform.python_implementation() != "PyPy",
+    reason="Working on 3.11+ and Pypy.",
+)
```
Dreamsorcerer:
```suggestion
```

Dreamsorcerer:
This is actually working now. But, I don't know if we want to merge such a complex test after realising that these warnings were already visible in many tests (just ignored in the pytest config), so we already have coverage of this.

PR #5166, opened 2020-10-28
Use md5 checksums instead of file timestamps

asvetlov:
It allows more reliable rebuilding C extensions on jumping between aiohttp versions


asvetlov:
Backported manually

PR #5184, opened 2020-10-31
Return Request.

Dreamsorcerer:
A small improvement to typing annotation. Noticed this while trying to update typing on aiohttp-jinja2.


asvetlov (approved):
Thanks!

PR #5197, opened 2020-11-03
Http exception cookies support

derlih:
## What do these changes do?

Add methods `set_cookie()` an `del_cookie()` to `HTTPException`
Add `cookies` property to `HTTPException`

User can set the cookies for `HTTPException`

## Related issue number

#4277 

## Checklist


derlih:
@asvetlov if this approach is OK, I will update the documentation.

asvetlov (commented):
Looks good.
Could you please update web_excetions.rst to reflect new methods?

derlih:
What `versionadded` should I use to reflect these changes in documentation?

asvetlov:
4.0
In 3.x branch HTTPException is derived from Response, thus it has cookie methods already.
Let's add breaking changes in major releases only.

derlih:
That was exactly the reason why I've raised my question. If it doesn't break the API from 3.x  in theory it can be backported to 3.x.

asvetlov:
Dropping the inheritance from `web.Response` is not backward compatible change.

While we keep the inheritance -- there Is no need to backport this PR because `web.Response` already provides the cookies API

asvetlov (approved):

asvetlov:
Thanks!

PR #4771, opened 2020-05-24
[3.7] Don't cancel web handler on disconnection (#4080)

jack1142:
## What do these changes do?

Backport #4080 to aiohttp 3.7. I'm not sure whether there's gonna be another 3.6 release or if 3.7 is still planned but this allows to handle connections issues properly so would be good to have.

## Are there changes in behavior for the user?

`ConnectionResetError` is raised on writing to closing transport instead of a warning.

## Related issue number

#4080

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #5233, opened 2020-11-14
[3.8] Backport fix for setting cookies

Dreamsorcerer:
I see this fix in master, but this needs to be released on the 3.7 branch, as aiohttp-session is broken without this.

(cherry-picked from https://github.com/aio-libs/aiohttp/commit/54afac769c44054d7b9458c8fe0e42278ec70766)


asvetlov:
Could you please refer to the master's commit with the fix?

Dreamsorcerer:
It's this one: https://github.com/aio-libs/aiohttp/commit/54afac769c44054d7b9458c8fe0e42278ec70766

Appears to be part of a more substantial change in aiohttp4 (where HTTPException is no longer subclassing Response). This single line I've taken from there is enough to fix the behaviour in v3.7.

Dreamsorcerer:
That test appears to have passed on the previous commit, so I think it's just flaky (given that I only added a CHANGE file in this commit).

Dreamsorcerer:
I've added a simplified version of the test in aiohttp-session, which should catch regressions.

asvetlov:
Cool! I'll take a review soon

aiohttp/web_protocol.py
```diff
@@ -426,6 +426,7 @@ async def _handle_request(
             resp = Response(
```
asvetlov: I'm sorry but the fix is incorrect.
This line was backported by the incident.
Since in aiohttp 3.x HttpException is derived from Response we should just use it as-is without the creation of intermediate response object.
I believe this line was incorrectly backported by https://github.com/aio-libs/aiohttp/pull/4771

I apologize for the regression.

asvetlov (changes requested):

aiohttp/web_protocol.py
Dreamsorcerer: No problem, I'll just update this with the reverted code then.

Appears to actually have been this commit, which moved some of the logic into a new _handle_request() method:
https://github.com/aio-libs/aiohttp/commit/29eccad84e8200b5c90856c8732da0fdbbcef904#diff-2126b277e07e3fdd05e7a81da456accf24e5515a46c78c48a79db4eb166043e4L373

Dreamsorcerer: Apparently, not so easy. Now getting ServerDisconnectedError in the tests.
Will take a look later in the week, unless you beat me to it.

Dreamsorcerer:
OK, reverting back to using the exception as the response seems to require moving the exception handler back to the parent function in order to handle the deprecation warning correctly. I think that's working now though.

Dreamsorcerer:
Flaky test? Seems odd that the changes would fail on macos only. Also the CI seems to be struggling, it fails to load the test log half of the time I try to look at it.

thijstriemstra:
+1! Also ran into this issue and reverted aiohttp to 3.6.3 for now.

Dreamsorcerer:
Looking back at my previous comments, the current test failure is the same one from earlier, which is just flaky, as it failed on one commit and passed on the next despite only changing documentation.

aiohttp/web_protocol.py
```diff
@@ -483,17 +478,21 @@ async def start(self) -> None:
                 task = self._loop.create_task(self._handle_request(request, start))
                 try:
                     resp, reset = await task
+                except HTTPException as exc:
```
asvetlov: Please move this `except` block and `if getattr(resp, "__http_exception__", False):` into `_handle_request method`.

asvetlov (changes requested):
The PR is correct but please move your changes into `_handle_request`; it reduces the diff between the master and backport branches.

Also, I would freeze 3.7 branch but publish aiohttp 3.8 soon.
Please use 3.8 as the base branch for PR.

thijstriemstra:
maybe change the pr title as well.

Dreamsorcerer:
Tests passed this time. Good to go.

Dreamsorcerer:
First time I've seen a failure from lgtm. :P
Appears to be from the merge though, not related to the PR (in fact, it's stopped failing since the target branch has been changed).

asvetlov:
Then I see lgtm alerts I usually open "view on LGTM.com" link, analyze it, and close as "I don't care" or "false alarm".
But sometimes it is helpful as a point at "bad smelling" code.

asvetlov (approved):

asvetlov:
Thanks!

PR #5263, opened 2020-11-21
Ensure cleanup contexts are completed when an exception occurs on startup

### LINKED ISSUE #4799: Cleanup doesn't run when exception occurs in startup.

Dreamsorcerer:
Fixes #4799 

Second attempt to fix this, focusing only on cleanup_ctx this time. This will ensure any cleanup context that has been successfully started will get cleaned up, even if an exception occurs later in the startup process.

This can also be merged into 3.8.

### DIFF
--- a/CHANGES/4799.bugfix
+++ b/CHANGES/4799.bugfix
@@ -1,0 +1,1 @@
+Ensure a cleanup context is cleaned up even when an exception occurs during startup.
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -318,7 +318,11 @@ class Application(MutableMapping[str, Any]):
 
         Should be called after shutdown()
         """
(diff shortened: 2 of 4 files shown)


asvetlov (approved):
LGTM
Sorry for the non-instant reply for such straightforward pull request.
I was busy with other things

Issue #5220, opened 2020-11-11
Server returns 400: "invalid character in chunk size header" for valid request


romasku:
🐞 **Describe the bug**
When server ignores content of request with `Transfer-Encoding: chunked`, parsing of next request can fail.

💡 **To Reproduce**

The following script:

```
import asyncio

from aiohttp import web, ClientSession


async def start_server():
    async def hello(request):
        return web.Response(text="OK", status=200)

    app = web.Application()
    app.add_routes([web.get('/', hello)])
    app.add_routes([web.put('/', hello)])

    runner = web.AppRunner(app)
    await runner.setup()
    site = web.TCPSite(runner, 'localhost', 8080)
    await site.start()


async def main():
    await start_server()

    async def data_gen():
        for _ in range(10):
            yield b'just data'
            await asyncio.sleep(0.1)  # Make sending of content slow

    async with ClientSession() as session:
        while True:
            # Do put
            async with session.put("http://localhost:8080", data=data_gen()):
                pass
            # Do get
            async with session.get("http://localhost:8080") as resp:
                if resp.status != 200:
                    print("failed with status: ", resp.status)
                    print("content is: ", await resp.content.read())
                    return
            await asyncio.sleep(0.1)

asyncio.run(main())
```

Prints the following:

```
$ python3 reproduce.py 
Error handling request
Traceback (most recent call last):
  File ".../venv/lib/python3.7/site-packages/aiohttp/web_protocol.py", line 315, in data_received
    messages, upgraded, tail = self._request_parser.feed_data(data)
  File "aiohttp/_http_parser.pyx", line 546, in aiohttp._http_parser.HttpParser.feed_data
aiohttp.http_exceptions.BadHttpMessage: 400, message='invalid character in chunk size header'
failed with status:  400
content is:  b'invalid character in chunk size header'
```

If I add `await request.content.read()`, the problem disappears.

💡 **Expected behavior**

Server should not fail to parse valid request.

📋 **Your version of the Python**

```console
$ python --version
Python 3.7.9
```

📋 **Your version of the aiohttp/yarl/multidict distributions**

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.2
...
```
```console
$ python -m pip show multidict
Name: multidict
Version: 4.7.6
...
```
```console
$ python -m pip show yarl
Name: yarl
Version: 1.6.2
...
```

asvetlov:
Thanks for the reproducer.
[...]

PR #5270, opened 2020-11-22
Refactor web error handling

asvetlov:
A part of #5220 refactoring


aiohttp/web_protocol.py
```diff
@@ -639,6 +646,13 @@ def handle_error(
         information. It always closes current connection."""
         self.log_exception("Error handling request", exc_info=exc)
 
+        # some data already got sent, connection is broken
+        if request.writer.output_size > 0:
+            raise ConnectionError(
+                "Response is sent already, cannot send another response "
+                "with the error message"
+            )
```
webknjaz, line 642: How about
```suggestion
            ) from exc
```

tests/test_web_protocol.py
```diff
@@ -1,805 +0,0 @@
-# Tests for aiohttp/server.py
```
webknjaz, line 2: Why are you removing these tests completely?

aiohttp/streams.py
```diff
@@ -539,7 +539,7 @@ def read_nowait(self) -> bytes:
         return b""
 
 
-EMPTY_PAYLOAD = EmptyStreamReader()
+EMPTY_PAYLOAD: StreamReader = EmptyStreamReader()  # type: ignore
```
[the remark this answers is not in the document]
asvetlov: Good point.  I've decided to explicitly inherit `EmptyStreamReader` from `StreamReader`, it is the less confusing change.

aiohttp/web_protocol.py
```diff
@@ -95,41 +97,48 @@ async def log(
...
-    connection gets closed.
+        RequestHandler handles errors in incoming request, like bad
+        status line, bad headers or incomplete payload. If any error occurs,
+        connection gets closed.
 
-    :param keepalive_timeout: number of seconds before closing
-                              keep-alive connection
-    :type keepalive_timeout: int or None
+        :param keepalive_timeout: number of seconds before closing
+    o                              keep-alive connection
```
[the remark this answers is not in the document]
asvetlov: Done

aiohttp/web_protocol.py
asvetlov: No, `exc` is not related to the raised ConnectionError exception.
This ConnectionError is not caught by a user but intercepted by web_protocol internals and silently ignored.
I can send bare ConnectionError technically but prefer attaching a string annotation which can help in debugging in the future.

Plus, `exc` is logged already.

tests/test_web_protocol.py
asvetlov: Because they are mock-based totally.
These tests test their own mocks, nothing else.
I wrote these tests striving to get 100% formal coverage but I see later that tests don't help find any error.
The code is covered by functional tests more or less, which is a good enough compromise.

BTW, the same situation exists for asyncio test suite: a part of tests a pure mock-based one. They were useful at the very beginning but now when a mocked test fails by a change of internal logic or refactoring -- we just kick off this subtle useless test.

aiohttp/web_protocol.py
```diff
@@ -201,11 +204,12 @@ def __init__(
         self._keepalive_timeout = keepalive_timeout
         self._lingering_time = float(lingering_time)
 
-        self._messages: Deque[Tuple[RawRequestMessage, StreamReader]] = deque()
+        self._messages: Deque[
+            Tuple[Union[_ErrInfo, RawRequestMessage], StreamReader]
```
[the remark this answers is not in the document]
asvetlov: Done

tests/test_web_protocol.py
webknjaz: It looks like the coverage is reduced. I guess my point is that the functional tests should make sure to cover the same lines this suite used to cover.

aiohttp/streams.py
```diff
@@ -490,7 +492,10 @@ def _read_nowait(self, n: int) -> bytes:
         return b"".join(chunks) if chunks else b""
 
 
-class EmptyStreamReader(AsyncStreamReaderMixin):
+class EmptyStreamReader(StreamReader):
```
webknjaz: Add this
```suggestion
class EmptyStreamReader(StreamReader):  # lgtm [py/missing-call-to-init]
```
to ignore https://lgtm.com/projects/g/aio-libs/aiohttp/snapshot/0e21998af5e6fc75741c4f9e1c43696c349dd279/files/aiohttp/streams.py?sort=name&dir=ASC&mode=heatmap#x797aa799000fbf74:1

P.S. LGTM will take the comment into account only upon merge.

tests/test_web_protocol.py
asvetlov: Fair point!
The coverage was reduced because handling the base `Exception` thrown from HttpParser was not covered by tests.
Ironically, the code that returns 500 on the internal parser error was invalid.
The correct behavior in this case -- do nothing but catch the exception by asyncio code, report about an unexpected exception, and close the transport.
The wrong code is dropped, thank you.
Previously we had a mock-based test that emulated the error-ness behavior without careful thinking.

asvetlov:
Done. The next part of the changes I want to do in a separate PR.
It should be dropping `force_close()` method for the sake of half-close/linger/close sequence always.

PR #5274, opened 2020-11-23
Fix CI link and typo in setup.py

asvetlov:


setup.py
```diff
@@ -100,7 +100,7 @@ def read(f):
     url="https://github.com/aio-libs/aiohttp",
     project_urls={
         "Chat: Gitter": "https://gitter.im/aio-libs/Lobby",
-        "CI: Azure Pipelines": "https://dev.azure.com/aio-libs/aiohttp/_build",
+        "CI: GitHub Actions": "https://github.com/aio-libs/aiohttp/actions?query=workflow%3ACI",  # noqa
```
webknjaz, line 103: I normally add a filter for the master branch, otherwise, it'll also list runs from PR branches that may originate from upstream but aren't ready..

PR #5278, opened 2020-11-23
Drop Python 3.6 support

asvetlov:
Python 3.6 is close to EOL, aiohttp 4.0 is not released yet.
Dropping Python 3.6 for aiohttp 4.0 allows dropping a bunch of backward compatibility shims.


setup.py
```diff
@@ -5,8 +5,8 @@
 
 from setuptools import Extension, setup
 
-if sys.version_info < (3, 6):
-    raise RuntimeError("aiohttp 4.x requires Python 3.6+")
+if sys.version_info < (3, 7):
```
webknjaz, line 8: NOTE: I believe that with this change, we'll be able to migrate most of `setup.py` to `setup.cfg` since [CPython 3.7.0 ships `pip==10.0.1` and `setuptools==39.0.1`](https://github.com/python/cpython/tree/v3.7.0/Lib/ensurepip/_bundled).
[`setup.cfg` support was added in setuptools 30.3.0](https://setuptools.readthedocs.io/en/latest/history.html#v30-3-0) and [`Requires-Python` support has been added in pip 9.0.0](https://pip.pypa.io/en/stable/news/#id418).
Meaning that it's virtually impossible to hit this condition under the supported Python version. Add the platform wheels that we ship and you'll see that in 99% of cases users won't even run `setup.py`.

P.S. This is a topic for a separate PR, I just wanted to post some thoughts here.

webknjaz:
Looks like you also need to drop 3.6 from GHA

setup.py
asvetlov: I support the proposal (in a separate PR) but don't want to apply the change myself.
Regarding binary wheels -- the case number 1 when people build from tarball are Alpine Linux users.

asvetlov:
> Looks like you also need to drop 3.6 from GHA

Ooops. Done

PR #5284, opened 2020-11-24
Replace attrs with dataclasses

asvetlov:
aiohttp 4.0 supports Python 3.7+.
There is no need to use `attrs` but better to switch to the standard `dataclasses` module.
`attrs` is great but the standard is even better.

The only downside is: for `attrs` you should use `dataclasses.replace(ds, name=val)` instead of `attr.evolve(ds, name=val)`.
I think it is a good compromise.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -1,5 +1,6 @@
 import asyncio
 import codecs
+import dataclasses
 import functools
 import io
 import re
@@ -23,7 +24,6 @@ from typing import (
     cast,
 )
 
-import attr
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
 from yarl import URL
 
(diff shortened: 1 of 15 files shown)


aiohttp/client_reqrep.py
```diff
@@ -86,23 +86,19 @@ def _gen_default_accept_encoding() -> str:
...
 class RequestInfo:
     url: URL
     method: str
     headers: "CIMultiDictProxy[str]"
-    real_url: URL = attr.ib()
-
-    @real_url.default
-    def real_url_default(self) -> URL:
-        return self.url
+    real_url: URL
```
derlih, line 101: You can use `init=False` and write an `__init__` with the same behavior as it was with attrs.

asvetlov: really no.
aiohttp always call `RequestInfo` with `real_url` argument; only tests use 3-arguments form.
Changing the signature is ok if there is no user code that instantiates `RequestInfo` explicitly.

gleb-chipiga:
As far as I understand, dataclasses does not support `__slots__`, which were used with attrs. Have you decided to stop using `__slots__`?

asvetlov:
I just don't care.
Slots save some memory. 
The difference was huge for old Python, with a new compact dict (which split keys and values structure) implemented a gap is negligible.

Tinche:
@asvetlov You are free to do as you like, obviously, but dataclasses aren't better than attrs. Their functionality is a strict subset; it's mostly for use within the standard library and if you don't have access to third party packages. Everyone using aiohttp obviously has access to third party packages.

PR #5287, opened 2020-11-25
Use a distinct type for sentinel

asvetlov:
Before `sentinel` was processed as either `object` or `Any`, both variants are far from perfectness.

Now `sentinel` has a dedicated type which is not equal to anything.


asvetlov:
The missing part is the Ellipsis (`...`) handling.
I very much appreciate if somebody bears this change. 
Sure, I can help with the review etc.

asvetlov:
Give up backpoting, should be ok as is.

PR #5328, opened 2020-12-09
Replace AbstractMatchInfo with UrlMappingMatchInfo.

### LINKED ISSUE #4748: mypy AbstractMatchInfo has no attribute "route"
When running mypy, I get the error:
`error: "AbstractMatchInfo" has no attribute "route"  [attr-defined]`

From this code:
```
        match_info = await request.app.router.resolve(request_clone)
        if isinstance(match_info.route, web.ResourceRoute):
```

But, looking at the code, it looks to me like everything is actually returning a `UrlMappingMatchInfo`, which does have this attribute.

Would it make sense to just change the return to `UrlMappingMatchInfo`? I can't see why you would get another type returned from this function.

Dreamsorcerer:
Fixes #4748 

If I understood the comment on that issue correctly, this is the change that should be made.

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -79,7 +79,7 @@ PATH_SEP = re.escape("/")
 
 _WebHandler = Callable[[Request], Awaitable[StreamResponse]]
 _ExpectHandler = Callable[[Request], Awaitable[None]]
-_Resolve = Tuple[Optional[AbstractMatchInfo], Set[str]]
+_Resolve = Tuple[Optional["UrlMappingMatchInfo"], Set[str]]
 
 
 class _InfoDict(TypedDict, total=False):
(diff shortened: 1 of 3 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -987,11 +987,11 @@ async def resolve(self, request: Request) -> AbstractMatchInfo:
                 return match_dict
             else:
                 allowed_methods |= allowed
+
+        if allowed_methods:
+            return MatchInfoError(HTTPMethodNotAllowed(method, allowed_methods))
         else:
```
Dreamsorcerer, line 998: I dropped the else here, as it is redundant, and makes it a little more difficult to see that the function will always return.
(The diff is a little awkward to look at, as it's split it weirdly, but all I've done is remove the `else` following the `for`).

asvetlov:
Superseded by #6135

PR #5344, opened 2020-12-14
Fix StreamResponse.prepared not returning True after EOF is sent

### LINKED ISSUE #5343: `web_response.Response.prepared` is `False` after `write_eof()`
`web_response.Response.prepared` is `False` after `write_eof()` (the payload writer goes back to None).

This is a bit mis-leading - any reason not to change `Response.prepared` to `self._eof_sent or self._payload_writer is not None`?

arthurdarcet:
Fixes #5343

I renamed a few tests (i think the `prepared` attribute used to be named `started` and the tests were not renamed?)
And I removed some asserts around prepare in a `test___repr__` test - to move them to a dedicated test


Dreamsorcerer:
This PR seems to have been missed. Thanks @arthurdarcet.

bdraco (approved):

PR #5374, opened 2020-12-29
Fix imports of typing_extensions

### LINKED ISSUE #5107: ModuleNotFoundError: No module named 'typing_extensions'
🐞 **Describe the bug**

After updating aiohttp from 3.6.2 to 3.7.0, the [gql](https://github.com/graphql-python/gql) tests are failing on python 3.8 with the Exception `ModuleNotFoundError: No module named 'typing_extensions'`

```
tests/conftest.py:15: in <module>
    from aiohttp.test_utils import TestServer as AIOHTTPTestServer
.tox/py38/lib/python3.8/site-packages/aiohttp/test_utils.py:43: in <module>
    from .web import (
.tox/py38/lib/python3.8/site-packages/aiohttp/web.py:23: in <module>
    from .web_app import Application as Application
.tox/py38/lib/python3.8/site-packages/aiohttp/web_app.py:39: in <module>
    from .web_middlewares import _fix_request_current_app
.tox/py38/lib/python3.8/site-packages/aiohttp/web_middlewares.py:7: in <module>
    from .web_urldispatcher import SystemRoute
.tox/py38/lib/python3.8/site-packages/aiohttp/web_urldispatcher.py:36: in <module>
    from typing_extensions import TypedDict
E   ModuleNotFoundError: No module named 'typing_extensions'

```

💡 **To Reproduce**

Check the PR [#156](https://github.com/graphql-python/gql/pull/156)

Dreamsorcerer:
Fixes #5107


derlih:
@Dreamsorcerer 
may be it is better to move all of this kind of importing
```python
if sys.version_info >= (3, 8):
    from typing import Final
else:
    from typing_extensions import Final
```
to `typedefs.py` and then import it from there. What do you think?

Dreamsorcerer:
> to `typedefs.py` and then import it from there. What do you think?

I'm not overly opposed to it, but I think it could make maintenance slightly more difficult. e.g. When we drop support for Python 3.7, I can currently grep for "3, 8" or "typing_extensions" to find every place that should be updated. I don't really want to leave the code importing from typedefs when it's no longer needed.

derlih:
Then it could be still present in `typedefs.py` but without version checking.
I mean to find all occurrences is not a big problem. Just run the tests or mypy.

Dreamsorcerer:
Yep, OK, we'll do it that way.

Issue #5389, opened 2021-01-08
[maintenance] Adjust test job timeouts in GitHub Actions CI/CD jobs


webknjaz:
🐞 **Describe the bug**

The CI doesn't complete currently because the test jobs are capped to run under 15 minutes and they now get killed on timeout. I believe this is called by the new pip resolver doing backtracking which is rather time-consuming.

💡 **To Reproduce**

It's visible in any recent CI build.

💡 **Expected behavior**

The jobs should complete. Carefully bumping the timeout in the workflow file should be enough.

📋 **Logs/tracebacks**

Example: https://github.com/aio-libs/aiohttp/runs/1669008218?check_suite_focus=true

📋 **Your version of the Python**

N/A

📋 **Your version of the aiohttp/yarl/multidict distributions**

repo master

📋 **Additional context**

N/A

greshilov:
Increased timeout will not solve the problem completely, I guess. Stricter `requirements.txt` could help though.

webknjaz:
I think increasing the timeout should be enough to get it to run completely for the first time and then many things will probably get picked up from cache. One of the problems there is that when pip has to download many package versions, it makes the backtracking substantially slower when those packages don't have wheels.

PR #5393, opened 2021-01-10
Fix content-type on empty string.

### LINKED ISSUE #5392: Setting empty str as data does not set text/plain header
Setting empty str as data does not set text/plain header
```
await session.post(url, data='')
```
will set
```
'Content-Type': 'application/octet-stream'
```
in the header.

Expected would be
```
'Content-Type': 'text/plain'
```

This is also described in the [docs](https://docs.aiohttp.org/en/stable/client_advanced.html#custom-request-headers).
It shows ``data`` as a shortcut for ``text/plain``.

📋 **Versions**

```console
Python/3.8
aiohttp/3.7.3
```

Dreamsorcerer:
Fixes #5392

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -245,7 +245,7 @@ class ClientRequest:
         self.update_proxy(proxy, proxy_auth, proxy_headers)
 
         self.update_body_from_data(data)
-        if data or self.method not in self.GET_METHODS:
+        if data is not None or self.method not in self.GET_METHODS:
             self.update_transfer_encoding()
         self.update_expect_continue(expect100)
         if traces is None:
@@ -383,7 +383,7 @@ class ClientRequest:
 
     def update_content_encoding(self, data: Any) -> None:
         """Set request content encoding."""
-        if not data:
+        if data is None:
             return
 
         enc = self.headers.get(hdrs.CONTENT_ENCODING, "").lower()
@@ -433,7 +433,7 @@ class ClientRequest:
         self.headers[hdrs.AUTHORIZATION] = auth.encode()
 
     def update_body_from_data(self, body: Any) -> None:
-        if not body:
+        if body is None:
             return
 
         # FormData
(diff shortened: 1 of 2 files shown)


aiohttp/client_reqrep.py
```diff
@@ -433,7 +433,7 @@ def update_auth(self, auth: Optional[BasicAuth]) -> None:
         self.headers[hdrs.AUTHORIZATION] = auth.encode()
 
     def update_body_from_data(self, body: Any) -> None:
-        if not body:
+        if body is None:
```
Dreamsorcerer, line 436: This one fixes the error reported. If somebody more familiar with the code can double check that the other 2 are sensible changes, that would be great.

tests/test_client_session.py
```diff
@@ -271,6 +271,18 @@ def test_http_DELETE(session: Any, params: Any) -> None:
...
+    with mock.patch.object(session, "_connector", autospec=True) as connector_mock:
+        connector_mock.closed = False
+
+        # ClientOSError occurs after req has been produced, so we can just give up there.
+        with pytest.raises(aiohttp.client_exceptions.ClientOSError):
+            await session.post("https://foo.example", data="")
+
+        req = connector_mock.connect.await_args.args[0]
+        assert req.headers["Content-Type"] == "text/plain; charset=utf-8"
+
```
Dreamsorcerer: Maybe there's a cleaner way to test this? I didn't see an obvious way, but this is a little messy with the mock causing that exception and a couple of warnings.

Dreamsorcerer:
Anybody else seeing the same thing as me? Some checkes were not successful, but none of them failed. It appears to have just cancelled 6 for no apparent reason.

Harmon758:
> Anybody else seeing the same thing as me? Some checkes were not successful, but none of them failed. It appears to have just cancelled 6 for no apparent reason.

They timed out after 15 minutes (#5389).

Dreamsorcerer:
I'm lost trying to figure out how to make a test for this.

aiohttp/client_reqrep.py
asvetlov: IMHO the fix is correct

tests/test_client_session.py
asvetlov: I have no better idea

PR #5398, opened 2021-01-11
Fix quoted cookie

### LINKED ISSUE #5397: [Client] Inconsistent quoted cookie handling
🐞 **Describe the bug**

Quoted cookies are unquoted if domain attribute is not set.

💡 **To Reproduce**
Run this code:
```python
import aiohttp
import asyncio


async def request_with_cookie(session_cookie: str, request_cookie: str):
    session_cookie = SimpleCookie(session_cookie)
    request_cookie = SimpleCookie(request_cookie)

    async with aiohttp.ClientSession(cookies=session_cookie) as session:
        async with session.get(
            'https://httpbin.org/headers',
            cookies=request_cookie,
        ) as response:
            html = await response.text()
            print("Body:", html)


async def main():
    await request_with_cookie(
        'sess="quoted_value"; domain=.httpbin.org',
        'req="quoted_value"'
    )

```

Outputs
```
Body: {
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "gzip, deflate", 
    "Cookie": "req=quoted_value; sess=\"quoted_value\"", 
    "Host": "httpbin.org", 
    "User-Agent": "Python/3.8 aiohttp/3.7.3", 
    "X-Amzn-Trace-Id": "Root=1-5ffc692a-7ee8f7946bd59b5f3dc67407"
  }
}
```
Cookie with `domain` attribute is quoted, and cookie without `domain` attribute is not quoted

💡 **Expected behavior**

I think if user inputs a quoted cookie, the quotes should not be dropped.
At least it should be consistent, and not dependent on `domain` attribute.

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)

```

Python 3.8.5

```console
$ python -m pip show aiohttp
Version: 3.7.3
...
```
```console
Version: 4.7.6
...
```
```console
$ python -m pip show yarl
Version: 1.4.2

...
```

📋 **Additional context**

Additional code to what I think is the cause
```python
from aiohttp import CookieJar
import yarl


def process_cookie_string(cookie_str: str) -> None:
    cookies = SimpleCookie(cookie_str)
    # Done internally by aiohttp - filtering cookies by domain.
    # https://github.com/aio-libs/aiohttp/blob/v3.7.3/aiohttp/client.py#L477
    tmp_cookie_jar = CookieJar()
    tmp_cookie_jar.update_cookies(cookies)
    req_cookies = tmp_cookie_jar.filter_cookies(yarl.URL('https://www.example.com/'))

    for name, value in req_cookies.items():
        print(f"{name=}, {value.value=}, {value.coded_value=}")

if __name__ == '__main__':
    process_cookie_string('name="value"')
    process_cookie_string('name="value"; Domain=example.com')

```

```
name='name', value.value='value', value.coded_value='value'
name='name', value.value='value', value.coded_value='"value"'
```

Also, these example might be a bit of a stretch because official docs say that cookies should be passed as `dict`.
But is there any reason why I shouldn't be able to send quoted cookies?

Dreamsorcerer:
Fixes #5397


asvetlov:
Looks like a draft, converting

Dreamsorcerer:
Yes, but I also don't see how to fix this, without breaking another of the test cases:
https://github.com/aio-libs/aiohttp/issues/5397#issuecomment-865260094

Dreamsorcerer:
I'm giving up on this one. I think at some point we should probably drop http.cookies, as it seems to give too many issues. We should also lose the interpretation of quoted cookies altogether as this concept doesn't exist in either latest or oldest cookie spec and was never adopted by browsers.

Issue #3450, opened 2018-12-13
pytest-aiohttp + pytest-xdist cause RuntimeError: set_wakeup_fd only works in main thread


JoseKilo:
## Long story short

Combining `pytest-aiohttp` and `pytest-xdist` an error is triggered sometimes from the `loop` fixture.

Let me know if you think the issue is more on the pytest(-xdist) side.

## Expected behaviour

No error.

## Actual behaviour

```
=================================================================================================================== ERRORS ====================================================================================================================
_____________________________________________________________________________________________________ ERROR at setup of test_foo[pyloop] ______________________________________________________________________________________________________
[gw6] linux -- Python 3.7.0 /home/merlin/workspace/pytest-test/.venv/bin/python3.7

loop_factory = <class 'asyncio.unix_events._UnixDefaultEventLoopPolicy'>, fast = False, loop_debug = False

    @pytest.fixture
    def loop(loop_factory, fast, loop_debug):
        """Return an instance of the event loop."""
        policy = loop_factory()
        asyncio.set_event_loop_policy(policy)
>       with loop_context(fast=fast) as _loop:

.venv/lib/python3.7/site-packages/aiohttp/pytest_plugin.py:205:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/local/lib/python3.7/contextlib.py:112: in __enter__
    return next(self.gen)
.venv/lib/python3.7/site-packages/aiohttp/test_utils.py:404: in loop_context
    loop = setup_test_loop(loop_factory)
.venv/lib/python3.7/site-packages/aiohttp/test_utils.py:427: in setup_test_loop
    watcher.attach_loop(loop)
/usr/local/lib/python3.7/asyncio/unix_events.py:868: in attach_loop
    loop.add_signal_handler(signal.SIGCHLD, self._sig_chld)
[...]

PR #5431, opened 2021-01-25
Re-introduce `pytest-xdist` in supported envs

### LINKED ISSUE #5849: Use pytest-xdist?

While developing #5847, I noticed that the pytests were taking quite a while to run on my machine (which is a pretty weak old desktop).

I thought that one easy way to speed up the testing time would be parallelization. I gave using [`pytest-xdist`](https://pypi.org/project/pytest-xdist/) a quick try and it cut the execution time from around ~4 minutes to ~1.5 minutes. And somewhat surprisingly all the tests passed

The only changes I made were the following:
- Add `pytest-xdist==2.3.0` to `requirements/test.txt`
- Run the `pip-compile` command
- Re-build and activate the venv
- Add the flag `-n auto` to the test command in the Makefile

Then I ran `make test`.

- Do nothing
- Use `pytest-parallel`

### Related component

Server, Client

### Additional context

_No response_

### Code of Conduct

webknjaz:
## What do these changes do?

This change resurrects parallel test execution in most of the test envs. On my laptop, the full test run is almost 3x faster: 36.04s vs 105.53s (0:01:45).

Refs:
  * https://github.com/pytest-dev/pytest-xdist/issues/620
  * https://stackoverflow.com/a/58614689/595220
  * https://bugs.python.org/issue35621
  * https://github.com/python/cpython/pull/14344

Nope.

## Related issue number

https://github.com/aio-libs/aiohttp/pull/3419 + https://github.com/aio-libs/aiohttp/issues/3450

## Checklist

### DIFF
--- a/tests/test_client_request.py
+++ b/tests/test_client_request.py
@@ -736,7 +736,8 @@ async def test_urlencoded_formdata_charset(
         data=aiohttp.FormData({"hey": "you"}, charset="koi8-r"),
         loop=loop,
     )
-    await req.send(conn)
+    async with await req.send(conn):
+        await asyncio.sleep(0)
     assert "application/x-www-form-urlencoded; charset=koi8-r" == req.headers.get(
         "CONTENT-TYPE"
     )
@@ -755,7 +756,8 @@ async def test_formdata_boundary_from_headers(
             headers={"Content-Type": f"multipart/form-data; boundary={boundary}"},
             loop=loop,
         )
-        await req.send(conn)
+        async with await req.send(conn):
+            await asyncio.sleep(0)
         assert req.body._boundary == boundary.encode()
 
 
@@ -1088,13 +1090,13 @@ async def test_data_stream_exc(
 
     t = loop.create_task(throw_exc())
 
-    await req.send(conn)
-    assert req._writer is not None
-    await req._writer
-    await t
-    # assert conn.close.called
-    assert conn.protocol is not None
-    assert conn.protocol.set_exception.called
+    async with await req.send(conn):
+        assert req._writer is not None
+        await req._writer
+        await t
+        # assert conn.close.called
+        assert conn.protocol is not None
+        assert conn.protocol.set_exception.called
     await req.close()
 
 
@@ -1118,9 +1120,9 @@ async def test_data_stream_exc_chain(
 
     t = loop.create_task(throw_exc())
 
-    await req.send(conn)
-    assert req._writer is not None
-    await req._writer
+    async with await req.send(conn):
+        assert req._writer is not None
+        await req._writer
     await t
     # assert conn.close.called
     assert conn.protocol.set_exception.called
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -291,6 +291,8 @@ async def test_create_conn() -> None:
     with pytest.raises(NotImplementedError):
         await conn._create_connection(object(), [], object())  # type: ignore[arg-type]
 
+    await conn.close()
+
 
 async def test_async_context_manager(loop: asyncio.AbstractEventLoop) -> None:
     conn = aiohttp.BaseConnector()
@@ -581,6 +583,8 @@ async def test_release_close(key: ConnectionKey) -> None:
     assert not conn._conns
     assert proto.close.called
 
+    await conn.close()
+
 
 async def test_release_proto_closed_future(
     loop: asyncio.AbstractEventLoop, key: ConnectionKey
@@ -591,6 +595,8 @@ async def test_release_proto_closed_future(
     # See PR #6321
     assert protocol.closed.result() is None
 
+    await conn.close()
+
 
 async def test__release_acquired_per_host1(
     loop: asyncio.AbstractEventLoop, key: ConnectionKey
@@ -598,6 +604,8 @@ async def test__release_acquired_per_host1(
     conn = aiohttp.BaseConnector()
     conn._release_acquired(key, create_mocked_conn(loop))
     assert len(conn._acquired_per_host) == 0
+
+    await conn.close()
 
 
 async def test__release_acquired_per_host2(
@@ -608,6 +616,8 @@ async def test__release_acquired_per_host2(
     conn._acquired_per_host[key].add(handler)
     conn._release_acquired(key, handler)
     assert len(conn._acquired_per_host) == 0
+
+    await conn.close()
 
 
 async def test__release_acquired_per_host3(
@@ -621,6 +631,8 @@ async def test__release_acquired_per_host3(
     conn._release_acquired(key, handler)
     assert len(conn._acquired_per_host) == 1
     assert conn._acquired_per_host[key] == {handler2}
+
+    await conn.close()
 
 
 async def test_tcp_connector_certificate_error(
@@ -643,6 +655,8 @@ async def test_tcp_connector_certificate_error(
         assert isinstance(ctx.value.certificate_error, ssl.CertificateError)
         assert isinstance(ctx.value, aiohttp.ClientSSLError)
 
+    await conn.close()
+
 
 async def test_tcp_connector_server_hostname_default(
     loop: asyncio.AbstractEventLoop, start_connection: mock.AsyncMock
@@ -658,6 +672,8 @@ async def test_tcp_connector_server_hostname_default(
 
         with closing(await conn.connect(req, [], ClientTimeout())):
             assert create_connection.call_args.kwargs["server_hostname"] == "127.0.0.1"
+
+    await conn.close()
 
 
 async def test_tcp_connector_server_hostname_override(
@@ -676,6 +692,8 @@ async def test_tcp_connector_server_hostname_override(
 
         with closing(await conn.connect(req, [], ClientTimeout())):
             assert create_connection.call_args.kwargs["server_hostname"] == "localhost"
+
+    await conn.close()
 
 
 async def test_tcp_connector_multiple_hosts_errors(
@@ -832,6 +850,8 @@ async def test_tcp_connector_multiple_hosts_errors(
 
     established_connection.close()
 
+    await conn.close()
+
 
 @pytest.mark.parametrize(
     ("happy_eyeballs_delay"),
@@ -921,6 +941,8 @@ async def test_tcp_connector_happy_eyeballs(
 
                 established_connection.close()
 
+    await conn.close()
+
 
 async def test_tcp_connector_interleave(loop: asyncio.AbstractEventLoop) -> None:
     conn = aiohttp.TCPConnector(interleave=2)
@@ -1002,6 +1024,8 @@ async def test_tcp_connector_interleave(loop: asyncio.AbstractEventLoop) -> None
     assert success_ips == [ip4]
     assert interleave_val == 2
     established_connection.close()
+
+    await conn.close()
 
 
 async def test_tcp_connector_family_is_respected(
@@ -1191,6 +1215,8 @@ async def test_tcp_connector_multiple_hosts_one_timeout(
 
     established_connection.close()
 
+    await conn.close()
+
 
 async def test_tcp_connector_resolve_host(loop: asyncio.AbstractEventLoop) -> None:
     conn = aiohttp.TCPConnector(use_dns_cache=True)
@@ -1210,6 +1236,8 @@ async def test_tcp_connector_resolve_host(loop: asyncio.AbstractEventLoop) -> No
                 assert rec["host"] in ("::1", "fe80::1", "fe80::1%lo0")
             else:
                 assert rec["host"] == "::1"
+
+    await conn.close()
 
 
 @pytest.fixture
@@ -1232,6 +1260,8 @@ async def test_tcp_connector_dns_cache_not_expired(
         await conn._resolve_host("localhost", 8080)
         m_resolver().resolve.assert_called_once_with("localhost", 8080, family=0)
 
+        await conn.close()
+
 
 async def test_tcp_connector_dns_cache_forever(
     loop: asyncio.AbstractEventLoop, dns_response: Callable[[], Awaitable[List[str]]]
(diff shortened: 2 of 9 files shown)


webknjaz:
@Dreamsorcerer @bdraco any ideas why the http parser tests might be failing in with pytest-xdist?

Dreamsorcerer:
> @Dreamsorcerer @bdraco any ideas why the http parser tests might be failing in with pytest-xdist?

It's not the parser tests, it's dev_mode. For some reason it must be disabling dev mode, thus all those tests are failing.

Dreamsorcerer:
Ah, is it because pytest-xdist is running new processes and not passing the same python arguments to the new processes?

Maybe easiest to just disable pytest-xdist on dev_mode run. There's only a dozen tests taking about a second.

webknjaz:
> and not passing the same python arguments to the new processes?

What do you mean? Aren't they initialized the same way?

Dreamsorcerer:
> > and not passing the same python arguments to the new processes?
> 
> What do you mean? Aren't they initialized the same way?

No, it runs `python -X dev`. That's why it's in a separate run.

webknjaz:
> Maybe easiest to just disable pytest-xdist on dev_mode run. There's only a dozen tests taking about a second.

Sure, it's going to be much faster without the plugin with such a small amount of tests. The overhead is not worth it.

However, I think that it's still a good idea to understand and fix the underlying issue.

Dreamsorcerer:
https://github.com/aio-libs/aiohttp/blob/04b1212fe231a7e827959aa9a648d5c847e57c0a/.github/workflows/ci-cd.yml#L232

webknjaz:
> python -X dev

I'd expect this to set `sys.flags` in the subprocesses.. Although, you might be right there. Hm... Can't we set the same setting via env vars that would leak into the subprocesses automatically?

Dreamsorcerer:
> Can't we set the same setting via env vars that would leak into the subprocesses automatically?

Sure:
> It can be enabled using the [-X dev](https://docs.python.org/3/using/cmdline.html#cmdoption-X) command line option or by setting the [PYTHONDEVMODE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONDEVMODE) environment variable to 1.

Need to check that the `--cov-append` also gets passed down.

webknjaz:
I added a commit for logging the mode just to be sure.

> (actually, I suppose that's already handled by pytest).

`pytest-cov` and `pytest-xdist` have a weird relationship — by default `pytest` would load any of them first and this order is not really predictable. So they have some code inside to detect that the other one is loaded and perform some extra initialization for tracking subprocesses...


> Need to check that the `--cov-append` also gets passed down

This should be visible in the codecov report, I suppose.

Dreamsorcerer:
Attempting to fix the Ubuntu failure in #8391.
Mac networking seems rather flaky, been seeing a lot of errors recently where the networking just doesn't seem to work.

webknjaz:
Oh, nice! Thanks for confirming that those are not related to this patch!

tests/test_connector.py
```diff
@@ -1397,6 +1397,7 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(0.1)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(0.2)  # Mitigate race condition on slow systems
```

webknjaz:
@Dreamsorcerer looks like there's even more flakiness coming out with the parallelism now, eh?

webknjaz:
@bdraco could you check https://github.com/aio-libs/aiohttp/actions/runs/8887163170/job/24401975076#step:11:8577 ? I think you were the last one to dig into the proxy tests, right? From the log
```python-traceback
ERROR tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http] - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <socket.socket fd=-1, family=2, type=1, proto=6>

  File "/home/runner/work/aiohttp/aiohttp/aiohttp/test_utils.py", line 481, in teardown_test_loop
    gc.collect()
ResourceWarning: unclosed <socket.socket fd=21, family=2, type=1, proto=6, laddr=('127.0.0.1', 58090), raddr=('127.0.0.1', 39771)>
```
it seems like there might be a live connection hanging around post-test.. WDYT?

tests/test_connector.py
```diff
@@ -1397,10 +1459,13 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(0.5)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(5.5)  # Mitigate race condition on slow systems
```

tests/test_client_request.py
```diff
@@ -715,7 +719,8 @@ async def test_formdata_boundary_from_headers(loop: Any, conn: Any) -> None:
             headers={"Content-Type": f"multipart/form-data; boundary={boundary}"},
             loop=loop,
         )
-        await req.send(conn)
+        async with await req.send(conn):
+            pass
```
webknjaz:
```suggestion
            await asyncio.sleep(5.5)
```

tests/test_client_request.py
```diff
@@ -715,7 +719,8 @@ async def test_formdata_boundary_from_headers(loop: Any, conn: Any) -> None:
             headers={"Content-Type": f"multipart/form-data; boundary={boundary}"},
             loop=loop,
         )
-        await req.send(conn)
+        async with await req.send(conn):
+            await asyncio.sleep(5.5)
```
webknjaz:
```suggestion
            await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -698,7 +701,8 @@ async def test_urlencoded_formdata_charset(loop: Any, conn: Any) -> None:
         data=aiohttp.FormData({"hey": "you"}, charset="koi8-r"),
         loop=loop,
     )
-    await req.send(conn)
+    async with await req.send(conn):
+        pass
```
webknjaz:
```suggestion
        await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -594,18 +594,21 @@ async def test_connection_header(loop: Any, conn: Any) -> None:
...
     assert req.headers.get("CONNECTION") == "keep-alive"
 
     req.keep_alive.return_value = False
     req.version = HttpVersion(1, 1)
     req.headers.clear()
-    await req.send(conn)
+    async with await req.send(conn):
+        pass
```
webknjaz:
```suggestion
        await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -594,18 +594,21 @@ async def test_connection_header(loop: Any, conn: Any) -> None:
...
     assert req.headers.get("CONNECTION") is None
 
     req.version = HttpVersion(1, 0)
     req.headers.clear()
-    await req.send(conn)
+    async with await req.send(conn):
+        pass
```
webknjaz:
```suggestion
        await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -594,18 +594,21 @@ async def test_connection_header(loop: Any, conn: Any) -> None:
     req.keep_alive.return_value = True
     req.version = HttpVersion(1, 1)
     req.headers.clear()
-    await req.send(conn)
+    async with await req.send(conn):
+        pass
```
webknjaz:
```suggestion
        await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -1046,7 +1051,8 @@ async def throw_exc():
 
     loop.create_task(throw_exc())
 
-    await req.send(conn)
+    async with await req.send(conn):
+        pass
```
webknjaz:
```suggestion
        await asyncio.sleep(0)
```

tests/test_client_request.py
```diff
@@ -1046,7 +1051,8 @@ async def throw_exc():
 
     loop.create_task(throw_exc())
 
-    await req.send(conn)
+    async with await req.send(conn):
+        await asyncio.sleep(0)
     await req._writer
```
webknjaz:
```suggestion
    if req._writer is not None:
        await req._writer
```

tests/test_connector.py
```diff
@@ -1397,10 +1459,13 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(5.5)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(2)  # Mitigate race condition on slow systems
```

tests/test_client_request.py
```diff
@@ -1046,8 +1051,10 @@ async def throw_exc():
 
     loop.create_task(throw_exc())
 
-    await req.send(conn)
-    await req._writer
+    async with await req.send(conn):
+        await asyncio.sleep(0)
+    if req._writer is not None:
```
webknjaz:
```suggestion
```

tests/test_connector.py
```diff
@@ -1397,10 +1459,13 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(2)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(3)  # Mitigate race condition on slow systems
```

tests/test_connector.py
```diff
@@ -1397,10 +1459,13 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(3)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(4)  # Mitigate race condition on slow systems
```

tests/test_connector.py
```diff
@@ -1397,10 +1459,13 @@ async def test_cleanup(key: Any) -> None:
     existing_handle = conn._cleanup_handle = mock.Mock()
 
     conn._cleanup()
+    await asyncio.sleep(4)  # Mitigate race condition on slow systems
```
webknjaz:
```suggestion
    await asyncio.sleep(5)  # Mitigate race condition on slow systems
```

tests/_pytest_plugin.py
webknjaz, line 1: this is currently disabled so I'll get rid of it before merging

webknjaz:
@Dreamsorcerer @bdraco could you check the changes in tests especially carefully? Looks like I made it work, and it makes the CI complete 3x faster. Some cleanup pending. But I'm mostly worried about the changes in tests — perhaps you'd be able to point out better ways of dealing with them.

tests/test_proxy_functional.py
```diff
@@ -118,6 +119,7 @@ async def test_secure_https_proxy_absolute_path(
 
     await sess.close()
     await conn.close()
+    await asyncio.sleep(0.1)
```
bdraco, line 163: `0.1` seems a bit brittle. Does this need one iteration of the event loop or is it waiting for a specific timer?
```suggestion
    await asyncio.sleep(0)
```

tests/test_proxy_functional.py
```diff
@@ -190,6 +192,8 @@ async def test_https_proxy_unsupported_tls_in_tls(
     await sess.close()
     await conn.close()
 
+    await asyncio.sleep(0.1)
```
bdraco, line 237: https://github.com/aio-libs/aiohttp/pull/5431/files#r1685841873

bdraco:
> @Dreamsorcerer @bdraco could you check the changes in tests especially carefully? Looks like I made it work, and it makes the CI complete 3x faster. Some cleanup pending. But I'm mostly worried about the changes in tests — perhaps you'd be able to point out better ways of dealing with them.

The test changes seem fine. In Home Assistant we have a call_soon at the end of each test to give aiohttp one more event loop iteration to shutdown https://github.com/home-assistant/core/blob/4eb096cb0a593bc35f9cd9a2b3c797d2700a4aff/tests/common.py#L369

tests/test_proxy_functional.py
webknjaz: I'm not sure. I was experimenting w/o really debugging in many of the cases. And sometimes `0` wasn't helping… I'll need to take a look again, I suppose.

tests/test_connector.py
```diff
@@ -1906,6 +2012,8 @@ def check_with_exc(err):
 
         assert not conn._waiters
 
+        yield from conn.close()
```
Dreamsorcerer: I wonder if this function should just be converted to an async function? Maybe the test just predates async syntax...

The codecov comment says that it's not running though, which makes sense given all we are doing is creating a generator object. So, we never actually run any of the function body...

Dreamsorcerer:
> In Home Assistant we have a `call_soon` at the end of each test to give aiohttp one more event loop iteration to shutdown https://github.com/home-assistant/core/blob/4eb096cb0a593bc35f9cd9a2b3c797d2700a4aff/tests/common.py#L369 for similar reasons

I was under the impression this should not be needed on master (v4) anymore, so shouldn't be a factor in these tests. Unless it is something different from: https://docs.aiohttp.org/en/stable/client_advanced.html#graceful-shutdown

tests/test_connector.py
webknjaz: That's an interesting idea… I'll have to check. I'm tired, working on this patch on and off for years. So at this point, I'm almost ready to accept any suggestion that would get this merged. I just want to be careful not to break anything. I think the last couple of times (like 6–7 years ago) I attempted integrating xdist here, and it brought a lot of flakiness that would become evident in Travis CI. So I'm both invested and annoyed at this point :)

webknjaz:
Perhaps, somebody will be able to pair with me on the sprint day. Or just take care of rebasing+researching the solution.

tests/test_connector.py
```diff
@@ -2732,6 +2834,8 @@ async def check_with_exc(err: Exception) -> None:
 
             assert not conn._waiters
 
+            yield from conn.close()
```
bdraco:
```suggestion
        await conn.close()
```

bdraco (approved):

Issue #4818, opened 2020-06-12
deny injection of custom http headers


xppt:
Hello!

I've noticed, that aiohttp is simply concatenating server-response (or client-request) header like this, w/o any validation:
```python
    line = status_line + '\r\n' + ''.join(
        [k + ': ' + v + '\r\n' for k, v in headers.items()])
```

Which may be not okay, if some of the header values were based on user input.

Consider this example:
```python
import aiohttp.web
async def handler(req: aiohttp.web.Request):
    return aiohttp.web.Response(headers={
        'X-Debug-Param': req.query.get('param', ''),
    })
app = aiohttp.web.Application()
app.add_get('/', handler)
```

This code seems to be fine. Unfortunately it is not, since an attacker can craft urls that will force this handler to return any custom http-headers, or skip some of the existing ones, or broke http payload:
``` 
/?param=%0d%0aLocation:%20https://malware.host/  # open redirect
/?param=%0d%0aSet-Cookie:%20...                  # set some cookie
/?param=%0d%0aContent-Length:%2040%0d%0a         # skip next headers
```
and so on.

I think that aiohttp should raise an exception for any http-reason, header-name or header-value that contains \r or \n characters, instead of breaking http payload silently.

Actually this is what flask/werkzeug do for header-value:
```python
	if u"\n" in value or u"\r" in value:
		raise ValueError(
			"Detected newline in header value.  This is "
			"a potential security problem"
		)
```

asvetlov:
Sounds good.
We need a champion for this issue!

webknjaz:
@xppt this sounds like a security issue. Posting such things on public forums is considered a bad tone security-wise: https://github.com/aio-libs/aiohttp/security/policy. Please consider practicing responsible disclosure next time.

asvetlov:
Not such bad I think.
I rather consider it a minor issue.
aiohttp is a library. If a code that uses the library doesn't check what is passed to aiohttp -- many very bad things are possible.
But I agree: If we can protect sending malformed HTTP header values -- let's do it.

PR #5452, opened 2021-01-29
Add validation of HTTP status line, header keys and values

franekmagiera:
## What do these changes do?
HTTP status line, header keys and values are checked in order to prevent custom header injection. If any of those contain a carriage return or newline character a ValueError is raised. The checks are done in `_serialize_headers`.

Changes where made and tested for pure python version and cython version.

Also imported the `_serialize_headers` function to `test_web_response`. I wanted to prevent code duplication and make sure that future changes to the way the headers are created are reflected in the tests. Let me know what you think about it.

The documentation was not updated since I could not find a suitable place to find describe the changes made. Suggestions are welcome.

No significant changes, except an exception may be raised.

## Related issue number
#4818 

## Checklist


franekmagiera:
@asvetlov @webknjaz what do you think about this?

franekmagiera:
@asvetlov @webknjaz would you be interested in merging something like this?

aiohttp/http_writer.py
```diff
@@ -171,11 +171,26 @@ async def drain(self) -> None:
             await self._protocol._drain_helper()
 
 
+def _check_string(string: str) -> str:
```
Dreamsorcerer: I'm not sure how I feel about a function which just returns the string unchanged, but it should atleast have a better name. Maybe `_checked_header()` or `_safe_header()`?

I'm also wondering if it's not better to check the values earlier, where the user sets them, rather than when they are being sent.

Dreamsorcerer (changes requested):

aiohttp/http_writer.py
```diff
@@ -171,11 +171,26 @@ async def drain(self) -> None:
...
+    if "\r" in string or "\n" in string:
+        raise ValueError(
+            "Newline or carriage return character detected in HTTP status message or "
+            "header. This is a potential security issue."
+        )
+    else:
```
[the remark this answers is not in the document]
franekmagiera: Unindented

aiohttp/http_writer.py
```diff
@@ -171,11 +171,26 @@ async def drain(self) -> None:
...
         + "\r\n"
-        + "".join([k + ": " + v + "\r\n" for k, v in headers.items()])
+        + "".join(
+            [
+                _check_string(k) + ": " + _check_string(v) + "\r\n"
+                for k, v in headers.items()
+            ]
+        )
     )
     return line.encode("utf-8") + b"\r\n"
```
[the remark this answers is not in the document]
franekmagiera: Yep, definately a good idea. Broke it down as suggested.

aiohttp/http_writer.py
franekmagiera: I left it as `_check_string` because I use it to check status line as well. Not sure what would be a better name, maybe I should just add a docstring explaining what this function is intended for?

As for checking the values when user sets them - in this case I am not sure where in the codebase the valuation would take place - would veryfing the headers in `client.py` and `web_response.py` be enough or did you have another approach in mind?

I think it makes sense to check the values earlier, when they are being set. But I worry it would be easy to miss some case like for example when `charset`is passed (might be more cases like that). Checking the values when they are being sent seems more bulletproof, given that the headers may be mutated by some other code. Also I am wondering if this approach could potentially break existing code. Let me know what you think about it, but for now I would be more comfortable with checking the values when they are being sent.

Dreamsorcerer: > I left it as `_check_string` because I use it to check status line as well.

I think the status line is still part of the head, so not sure I'm too bothered by that. But, do we even need to check the status line? I've not run through the code yet, but I'd have thought the status line does not involve any user input.

> As for checking the values when user sets them - in this case I am not sure where in the codebase the valuation would take place - would veryfing the headers in `client.py` and `web_response.py` be enough or did you have another approach in mind?

@webknjaz Do you think there's a better place for this check to live?

franekmagiera: I think you are right, there is no need to validate the status line. I changed the name of the function and removed validation for status line.

Dreamsorcerer: So, just taking a look through the codebase, the alternative way I could imagine this working, is that `ClientRequest.headers` and `StreamResponse._headers` are both set to a `CIMultiDict`. If we changed this class to perform the check when a value is set in the dict, that could work. Though we'd need to also catch cases where the user passes in their own dict, so maybe it's not worth the complexity.

aiohttp/http_writer.py
```diff
@@ -171,13 +171,19 @@ async def drain(self) -> None:
...
+
 def _py_serialize_headers(status_line: str, headers: "CIMultiDict[str]") -> bytes:
-    line = (
-        status_line
-        + "\r\n"
-        + "".join([k + ": " + v + "\r\n" for k, v in headers.items()])
-    )
-    return line.encode("utf-8") + b"\r\n"
+    headers = (_safe_header(k) + ": " + _safe_header(v) for k, v in headers.items())
+    line = status_line + "\r\n" + "\r\n".join(headers) + "\r\n"
```
Dreamsorcerer:
```suggestion
    headers_gen = (_safe_header(k) + ": " + _safe_header(v) for k, v in headers.items())
    line = status_line + "\r\n" + "\r\n".join(headers_gen) + "\r\n"
```
`headers` is already defined as `CIMultiDict`.

Dreamsorcerer:
Tests without extensions are stuck. Looks like they get stuck on `tests/test_client_functional.py::test_keepalive_two_requests_success[pyloop]`

franekmagiera:
Sorry, missed that, forgot to run the code without extensions. There should be two `\r\n` signs after the headers.

Dreamsorcerer:
> Sorry, missed that, forgot to run the code without extensions. There should be two `\r\n` signs after the headers.

Ah, yes, I missed that change.

Dreamsorcerer (approved):
@webknjaz Unless you've got an idea on a different place to include the check, I think this should be merged.

aiohttp/http_writer.py
```diff
@@ -171,13 +171,19 @@ async def drain(self) -> None:
             await self._protocol._drain_helper()
 
 
+def _safe_header(string: str) -> str:
+    if "\r" in string or "\n" in string:
+        raise ValueError(
+            "Newline or carriage return character detected in HTTP status message or "
+            "header. This is a potential security issue."
```
Dreamsorcerer:
```suggestion
            "Newline or carriage return detected in headers. "
            "Potential header injection attack."
```

Dreamsorcerer:
If there's no response from @webknjaz, I'll merge this at the weekend.

PR #5457, opened 2021-02-07
Mypy coverage

Dreamsorcerer:
Provide some better Mypy coverage. Enables some more options that are useful and adds examples/ into the checker (which helps ensure our public API is correct, and helps keep the examples themselves up-to-date).

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -20,6 +20,9 @@ warn_unused_ignores = True
 disallow_any_unimported = True
 warn_return_any = True
 
+[mypy-tests.*]
+disallow_untyped_defs = False
+
 [mypy-aiodns]
 ignore_missing_imports = True
 
@@ -32,7 +35,7 @@ ignore_missing_imports = True
 [mypy-cchardet]
 ignore_missing_imports = True
 
-[mypy-gunicorn]
+[mypy-gunicorn.*]
 ignore_missing_imports = True
 
 [mypy-tokio]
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -74,7 +74,7 @@ fmt format:
 
 .PHONY: mypy
 mypy:
-	mypy --show-error-codes aiohttp tests
+	mypy
```
Dreamsorcerer, line 77: With no arguments, it uses the files and options in the config file.

examples/fake_server.py
```diff
@@ -50,7 +51,7 @@ def __init__(self) -> None:
                 web.get("/v2.7/me/friends", self.on_my_friends),
             ]
         )
-        self.runner = None
+        self.runner = web.AppRunner(self.app)
```
Dreamsorcerer, line 54: This example has clearly not been run in several years. :P

.mypy.ini
```diff
@@ -1,24 +1,45 @@
 [mypy]
-warn_unused_configs = True
-strict = True
+files = aiohttp, examples, tests
+check_untyped_defs = True
+follow_imports_for_stubs = True
+#disallow_any_decorated = True
```
Dreamsorcerer, line 5: These 2 commented out, we should consider in the future, but will need a ~50 errors to be cleaned up or ignored first.

webknjaz (approved):

webknjaz:
@Dreamsorcerer would you take care of backporting plz?

PR #5448, opened 2021-01-28
3.8 test run fix

derlih:
## What do these changes do?

Fix test run for python 3.6 on 3.8 branch

## Are there changes in behavior for the user?

no

## Related issue number

#5447 

## Checklist

- [x] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #5464, opened 2021-02-09
Backport #5457

Dreamsorcerer:
Improve Mypy coverage

I've skipped checking tests on this one as there are a lot of type errors coming through.


webknjaz:
Looks like this needs patching...

Dreamsorcerer:
Problem is already in the 3.8 branch (see failed tests on last commit to 3.8).

It looks like the `CIMultiDictProxy[str]` syntax should work because of this code:
https://github.com/aio-libs/multidict/blob/master/multidict/_abc.py#L7

Is it possible that the functionality is missing from the C version?
https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.c

Or any other ideas @webknjaz?

Dreamsorcerer:
Problematic annotation was added in: https://github.com/aio-libs/aiohttp/commit/383016b4f9b88b5eaf02f4bf649248827210dd24#diff-197153342549adb4b25580f8b842876841eec42dc48ae12cb72ba2ae4f7ced10R79

I can annotate it another way if needed, but I'd expect it to work if the class supports `__getitem__`

webknjaz:
Dunno, maybe quote it?

derlih:
I've created the PR #5448 to fix the problem on a CI. But I don't know how to solve the mypy issue there :-(

PR #5470, opened 2021-02-11
Update Makefile pip command

Dreamsorcerer:


Makefile
```diff
@@ -62,7 +62,7 @@ aiohttp/%.c: aiohttp/%.pyx $(call to-hash,$(CYS)) aiohttp/_find_header.c
 cythonize: .install-cython $(PYXS:.pyx=.c)
 
 .install-deps: .install-cython $(PYXS:.pyx=.c) $(call to-hash,$(CYS) $(REQS))
-	pip install -r requirements/dev.txt
+	@pip install -r requirements/dev.in -c requirements/dev.txt
```
webknjaz, line 65: It seems like this may need a more intrusive fix: https://github.com/aio-libs/aiohttp/pull/5470/checks?check_run_id=1879584882#step:7:41

CHANGES/5470.bugfix
```diff
@@ -0,0 +1 @@
+Fix CI
```
webknjaz, line 1: Could you add something that'd be useful in the changelog here?

webknjaz:
I think it's also necessary to do this update in master.

Dreamsorcerer:
Looks like this is superseeded by #5486.

Issue #4446, opened 2019-12-16
Why isn't aiodns used?


vangheem:
## Long story short

The documentation recommends to install aiodns. Reality is that even though aiodns might be installed, it's forced to not be used: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/resolver.py#L11

## Expected behaviour

aiodns to be used

## Actual behaviour

aiodns is not used

## workaround

Every time you create a `ClientSession`...

```python
from aiohttp.resolver import AsyncResolver

async with aiohttp.ClientSession(
        connector=aiohttp.TCPConnector(resolver=AsyncResolver())) as session:
    pass

```

gjcarneiro:
See #559

asvetlov:
I think @gjcarneiro answered the question.

#4451 can help with enabling aiodns by default on Python 3.8.
If somebody wants to champion the issue -- you are welcome!

PR #5471, opened 2021-02-13
Add Happy Eyeballs support to TCPConnector

### LINKED ISSUE #4451: Support Happy Eyeballs from Python 3.8
Starting from Python 3.8 asyncio supports https://tools.ietf.org/html/rfc6555

It allows working better in dual-host configurations.
It would be nice to support it by aiohttp if the feature is available by Python version.
Perhaps passing `happy_eyeballs_delay` and `interleave` to `loop.create_connection()` is enough.

A volunteer is welcome!

P.S.
Ooops, parameters are even not documented by asyncio docs, need to fix it.

illia-v:
## What do these changes do?

They add Happy Eyeballs support to `TCPConnector` and enable it by default.

aiohttp should become more responsive to users having dual-stack configuration.

For example, Go enabled Happy Eyeballs (Fast Fallback) by default in 2018 https://github.com/golang/go/commit/efc185029bf770894defe63cec2c72a4c84b2ee9.

Also, in https://github.com/aio-libs/aiohttp/issues/4446#issuecomment-567071597 @asvetlov wrote that Happy Eyeballs may help with enabling aiodns by default. But I would not do so for now because pycares that is a dependency of aiodns uses an old version of c-ares that has security vulnerabilities.

## Related issue number

Closes #4451 

## Checklist

### DIFF
--- a/CHANGES/4451.feature
+++ b/CHANGES/4451.feature
@@ -1,0 +1,2 @@
+Add Happy Eyeballs support to ``TCPConnector`` and enable it by default.
+Note, if Python 3.7 is used, the feature is disabled and related arguments are ignored.
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -45,7 +45,7 @@ from .client_exceptions import (
 )
 from .client_proto import ResponseHandler
 from .client_reqrep import SSL_ALLOWED_TYPES, ClientRequest, Fingerprint
-from .helpers import _SENTINEL, ceil_timeout, is_ip_address, sentinel
+from .helpers import _SENTINEL, PY_38, ceil_timeout, is_ip_address, sentinel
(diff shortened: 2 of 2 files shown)


webknjaz (commented):
Adding runtime-dependent behavior defaults is not the best idea: it should be consistent across envs. I don't think we should enable this by default but rather make it opt-in for as long as we support CPython versions below 3.8. If we were to have it on by default, it needs to be supported across all runtimes equally, otherwise, end-users would be confused as to what magic happens in some envs but not others, for example, by relying on a backport.

Also, this sort of change needs to be bundled with updates to tests and to docs.

illia-v:
I am closing this because the change has no effect 🙁. See https://github.com/aio-libs/aiohttp/issues/4451#issuecomment-786890416

PR #5486, opened 2021-02-20
Complete pip-tools setup

greshilov:
## What do these changes do?

Solve problems with inconsistent approaches to installing dependencies in Makefile.

No.

## Related issue number

Related PR: #5470.

## Checklist


greshilov:
@webknjaz I've marked this PR with `WIP` because I'm not sure about `pip-sync` usage. It seemed logical to use it since we moved to `pip-compile`, what do you think?

greshilov:
Looks like some tests are failing because of new warnings introduced in 77d0366b276fc8f7e22d10d157e47fa96a1e2de8.

requirements/cython.txt
```diff
@@ -1,3 +1,2 @@
--r multidict.txt
+-c dev.txt
```
webknjaz: dependabot does not support this

requirements/cython.txt
```diff
@@ -1,3 +1,2 @@
...
 cython==0.29.21
-typing_extensions==3.7.4.3  # required for parsing aiohttp/hdrs.py by tools/gen.py
```
webknjaz, line 3: This looks wrong.

webknjaz (commented):
I don't use pip-sync and it's a good idea to make sure that this all works with vanilla pip.

Also, for dependabot to work properly, we need to turn all the direct dependency files into `.in`. The pip-tools setup is incomplete and it needs polishing first before we even consider integrating extra things.
Take a look at pypa/warehouse — they have a good example of the file pairs for dependabot.

greshilov:
Summarizing the changes:

1. Files `doc-spelling,lint,cython,dev` are split into `in,txt` pairs. All of them are used as a direct requirement somewhere (`lint` in workflow script e.t.c.). 
Note the order, `lint` file must be always compiled before `dev`.

2. Unfortunately we can't use `-r dev.in -c dev.txt` anymore, because of the [new pip resolver implementation](https://github.com/pypa/pip/issues/8210). That's not a problem, I guess. `pip-compile` is designed to produce ready-to-go `.txt` dependencies.

3. I added the `.update-pip` step before all installation steps.

requirements/cython.txt
greshilov: Yep, fixed.

requirements/cython.txt
greshilov: Done!

webknjaz:
> * Unfortunately we can't use `-r dev.in -c dev.txt` anymore, because of the [new pip resolver implementation](https://github.com/pypa/pip/issues/8210). That's not a problem, I guess. `pip-compile` is designed to produce ready-to-go `.txt` dependencies.
> *

Why is that? What exactly causes problems?

greshilov:
> > * Unfortunately we can't use `-r dev.in -c dev.txt` anymore, because of the [new pip resolver implementation](https://github.com/pypa/pip/issues/8210). That's not a problem, I guess. `pip-compile` is designed to produce ready-to-go `.txt` dependencies.
> 
> Why is that? What exactly causes problems?

We have a requirement `pyjwt[crypto]==2.0.0`, which is declared with the extra dependency `[crypto]`. `pip-compile` carefully includes this requirement in `dev.txt`, but new pip resolver rules strictly prohibit any extras in a constraint file.

Generally speaking any dependency with `extra` will break the approach with  `-r dev.in -c dev.txt` :(
Read more about the reasons behind this behavior [here](https://github.com/pypa/pip/issues/6628).

webknjaz:
@greshilov sounds like `pip-tools` should be fixed to generate a proper constraints file...

webknjaz:
Meanwhile, we'll have to use `-r req.txt` syntax and probably generate separate requirements for different envs

greshilov:
> Meanwhile, we'll have to use `-r req.txt` syntax and probably generate separate requirements for different envs

This PR is alright for now then? I also wanted to suggest adding hashes to lock files, like in `pypa/warehouse` and `mozilla/ichnaea`.

webknjaz:
> This PR is alright for now then?

Alright. Let's merge it then, but we still need to raise this issue in the pip-tools repo, I think.
Also, I suppose you'll backport this, right?

> I also wanted to suggest adding hashes to lock files, like in `pypa/warehouse` and `mozilla/ichnaea`.

That's a great idea! I do this in all of my projects by default.

CHANGES/5470.bugfix
```diff
@@ -0,0 +1 @@
+Complete pip-tools setup.
```
webknjaz, line 1: Urgh... This is not a bugfix but a misc change because it's not a user-facing fix but something that would only affect the core devs and contributors.

greshilov: Sorry, my bad!

webknjaz: It's not a big deal, I guess this can be fixed by a follow-up PR with `git mv`.

PR #5495, opened 2021-02-25
[backport][3.8][PR#5494] Release all forgotten resources in tests

webknjaz:
## What do these changes do?

Ever since pytest got updated to v6.2, it started emitting resource
warnings under new Pythons 3.8+ which results in a failed status.

This change addresses this issue by properly most of the resources
so they don't even appear and ignoring the rest in cases of race
conditions and other cases that are hard to track down.

No.

## Related issue number

N/A

## Checklist


tests/test_client_session.py
```diff
@@ -514,6 +515,7 @@ async def handler(request):
 async def test_session_default_version(loop) -> None:
     session = aiohttp.ClientSession(loop=loop)
     assert session.version == aiohttp.HttpVersion11
+    await session.close()
```
luckydenis: Good afternoon. Maybe it's better to use a context manager for the session? 
```python
async with aiohttp.ClientSession(loop=loop) as session:
    assert session.version == aiohttp.HttpVersion11
```

webknjaz: Yes, it's better in general but I wanted to keep the patch as small as possible. Besides, if this fails in tests, we'll notice it and if the test succeeds, it does not really matter if an async CM was used or a regular close was called.

Besides, it's a backport PR, it should be kept as close to the original one as possible.

I'm more concerned about all the warnings that I failed to fix and had to ignore for now because they are either flaky or it is unclear what causes them.

That said, feel free to contribute improvements. I have a lot on my plate so I've only made a bare minimum of the changes necessary to get the new pytest up and running. Strictly speaking, I could just fill the config with ignores but I've made an extra effort to improve things which required substantially more involvement than one would expect at glance.

webknjaz:
Superseded by https://github.com/aio-libs/aiohttp/pull/6104.

PR #5502, opened 2021-02-26
Use `client_error` instead of a hard-coded exception in `TCPConnector`

illia-v:
## What do these changes do?

Use `client_error` that is an existing method argument instead of a hard-coded exception.

I doubt.
Although, `_create_proxy_connection` passes a different `client_error` and it won't be handled correctly in case of a problem. So the changes may prevent a crash.
Note, I didn't face the problem myself, I found the issue when I was working on #4451. 

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -985,7 +985,7 @@ class TCPConnector(BaseConnector):
         except OSError as exc:
             # in case of proxy it is not ClientProxyConnectionError
             # it is problem of resolving proxy ip itself
-            raise ClientConnectorError(req.connection_key, exc) from exc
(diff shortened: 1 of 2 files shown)


aiohttp/connector.py
```diff
@@ -985,7 +985,7 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
         except OSError as exc:
             # in case of proxy it is not ClientProxyConnectionError
```
webknjaz, line 986: this comment probably needs to be updated

webknjaz (commented):
Please add regression tests for this change.

PR #5507, opened 2021-02-28
Fix 5486.misc

greshilov:
## What do these changes do?

Fixes for #5486 

No.

## Checklist


greshilov:
@webknjaz looks like the only way to add hashes is to create a separate requirement file for each environment.

For example:
```
linux-py3.7-dev.txt
mac-py3.8-lint.txt
win32-py3.9-dev.txt
```

It will be necessary to create a `python` script `tools/deps.py` to automate installation and pin processes on different platforms.
Seems like a huge change, but it also will solve problem in `3.8` branch.

greshilov:
Yep, you've already suggested this approach in #5496, #5486.

webknjaz:
> ```
> linux-py3.7-dev.txt
> mac-py3.8-lint.txt
> win32-py3.9-dev.txt
> ```

I think that we don't need it per-python but per-OS. If you use the lowest supported Python to generate the constraints, it should be fine.

webknjaz (approved):

PR #5516, opened 2021-03-04
Remove @unittest_run_loop

Dreamsorcerer:
Removes the now unneeded @unittest_run_loop from master.

### DIFF
--- a/CHANGES/5516.misc
+++ b/CHANGES/5516.misc
@@ -1,0 +1,1 @@
+Removed @unittest_run_loop. This is now the default behaviour.
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -2,7 +2,6 @@
 
 import asyncio
 import contextlib
-import functools
 import gc
 import inspect
 import ipaddress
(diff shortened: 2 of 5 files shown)


webknjaz:
@Dreamsorcerer is the new default on regardless of the Python version?

Dreamsorcerer:
> @Dreamsorcerer is the new default on regardless of the Python version?

Appears to have a fallback on asynctest, so looks like it:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L60

webknjaz:
> Appears to have a fallback on asynctest, so looks like it:
> [https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py#L60](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/test_utils.py?rgh-link-date=2021-03-04T14%3A28%3A17Z#L60)

Could you make sure that there's a test for this then?

Dreamsorcerer:
> Could you make sure that there's a test for this then?

Indeed, the test_test_utils.py covers this already:
https://github.com/aio-libs/aiohttp/blob/unittest1/tests/test_test_utils.py#L105-L109

webknjaz (approved):

PR #5529, opened 2021-03-09
Fix sock_connect timeout being applied when 0.

### LINKED ISSUE #5527: ClientTimeout values of 0 don't disable timeouts
🐞 **Describe the bug**
`ClientTimeout` values of 0 and `None` are treated differently. The docs [here](https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts) state:

> All fields are floats, None or 0 disables a particular timeout check

and versions 3.6.3 and earlier do work this way. After #5091 this behavior was changed.

💡 **To Reproduce**
This simple example illustrates the issue:

```python
timeout = aiohttp.ClientTimeout(
    total=90,
    connect=0,
    sock_connect=0,
    sock_read=0
)

async with aiohttp.ClientSession(timeout=timeout) as session:
    print(await session.get("https://google.com"))
```

With v3.6.3:

```bash
ryan@debian:~$ python -m pip show aiohttp
Name: aiohttp
Version: 3.6.3

ryan@debian:~$ python test.py
<ClientResponse(https://www.google.com/) [200 OK]>
```

With v3.7.4:

```bash
ryan@debian:~$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.4

ryan@debian:~$ python test.py
```

```python
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/client.py", line 520, in _request
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/connector.py", line 535, in connect
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/connector.py", line 892, in _create_connection
    _, proto = await self._create_direct_connection(req, traces, timeout)
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/connector.py", line 999, in _create_direct_connection
    hosts = await asyncio.shield(host_resolved)
asyncio.exceptions.CancelledError

  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/client.py", line 520, in _request
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/async_timeout/__init__.py", line 45, in __exit__
    self._do_exit(exc_type)
  File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/async_timeout/__init__.py", line 92, in _do_exit
    raise asyncio.TimeoutError
asyncio.exceptions.TimeoutError

...

File "/home/ryan/.virtualenvs/aio-test/lib/python3.9/site-packages/aiohttp/client.py", line 524, in _request
    raise ServerTimeoutError(
aiohttp.client_exceptions.ServerTimeoutError: Connection timeout to host https://google.com
```
This timeout happens immediately.

With 3.7.4 if I change the timeout values from 0 to `None` it works correctly.

```
ryan@debian:~$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.4

ryan@debian:~$ python test.py
<ClientResponse(https://www.google.com/) [200 OK]>
```

💡 **Expected behavior**
Timeouts with values of 0 should be ignored (or the docs updated to say 0 doesn't disable the timeouts).

Dreamsorcerer:
Fixes #5527 

This is actually already on master, so this just matches what is there already.

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -671,15 +671,15 @@ class TimerContext(BaseTimerContext):
 
 
 def ceil_timeout(delay: Optional[float]) -> async_timeout.Timeout:
-    if delay is None:
-        return async_timeout.timeout(None)
-    else:
+    if delay is not None and delay > 0:
         loop = get_running_loop()
         now = loop.time()
         when = now + delay
         if delay > 5:
             when = ceil(when)
         return async_timeout.timeout_at(when)
+    else:
+        return async_timeout.timeout(None)
 
 
 class HeadersMixin:
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -671,15 +671,15 @@ class TimerContext(BaseTimerContext):
 
 
 def ceil_timeout(delay: Optional[float]) -> async_timeout.Timeout:
-    if delay is not None and delay > 0:
-        loop = get_running_loop()
-        now = loop.time()
-        when = now + delay
-        if delay > 5:
-            when = ceil(when)
-        return async_timeout.timeout_at(when)
-    else:
+    if delay is None or delay <= 0:
         return async_timeout.timeout(None)
+
+    loop = get_running_loop()
+    now = loop.time()
+    when = now + delay
+    if delay > 5:
+        when = ceil(when)
+    return async_timeout.timeout_at(when)
 
 
 class HeadersMixin:
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -714,8 +714,11 @@ async def test_client_session_timeout_argument() -> None:
 
 async def test_client_session_timeout_zero() -> None:
     timeout = client.ClientTimeout(total=10, connect=0, sock_connect=0, sock_read=0)
-    async with ClientSession(timeout=timeout) as session:
-        await session.get("http://example.com")
+    try:
+        async with ClientSession(timeout=timeout) as session:
+            await session.get("http://example.com")
+    except asyncio.TimeoutError:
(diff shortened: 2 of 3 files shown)


tests/test_client_session.py
```diff
@@ -712,6 +712,12 @@ async def test_client_session_timeout_argument() -> None:
     assert session.timeout == 500
 
 
+async def test_client_session_timeout_zero() -> None:
+    timeout = client.ClientTimeout(total=10, connect=0, sock_connect=0, sock_read=0)
+    async with ClientSession(timeout=timeout) as session:
+        await session.get("http://example.com")
```
webknjaz: I'd add some assertion because without an `assert` it's totally unclear what side-effect is being checked.

aiohttp/helpers.py
```diff
@@ -671,15 +671,15 @@ def timeout(self) -> None:
 
 
 def ceil_timeout(delay: Optional[float]) -> async_timeout.Timeout:
-    if delay is None:
-        return async_timeout.timeout(None)
-    else:
+    if delay is not None and delay > 0:
```
webknjaz: I'd prefer a guard expression here because this way you can eliminate the `else:` block entirely.
```suggestion
    if delay is None or delay <= 0:
        return async_timeout.timeout(None)
```
and then dedent the rest of this function.

Dreamsorcerer: Yeah, I'd probably have done it this way if I wasn't copying the code from master. Will update the other PR for master to match.

PR #5534, opened 2021-03-10
StaticResource only matches folder-like URL prefix

### LINKED ISSUE #5250: router.add_static is greedy
🐞 **Describe the bug**

The first argument for `add_static` matches on any path that starts with that string. For instance, if you specify "h", then `add_static` will match on "h" or "help" or "hoover".

💡 **To Reproduce**

Run this code:

```
from aiohttp import web

async def handle(request):
    text = "Hello, " + name
    return web.Response(text=text)

app = web.Application()

router = app.router

router.add_static("/h", "./")

router.add_routes([web.get('/', handle),
                web.get('/{name}', handle)])

if __name__ == '__main__':
    web.run_app(app)
```

Add a text file in the current working directory named `test.txt` and put some text in there.

Then navigate to the following URLs to demonstrate the issue:

* http://0.0.0.0:8080/
* http://0.0.0.0:8080/groot
* http://0.0.0.0:8080/h/test.txt
* http://0.0.0.0:8080/hoover

The last link unexpectedly returns a 404.

💡 **Expected behavior**

I would have expected http://0.0.0.0:8080/hoover to have similar behavior to http://0.0.0.0:8080/groot, and render "Hello, hoover".

3.8

aiohttp 3.7.2

danking:
## What do these changes do?

I think this fixes https://github.com/aio-libs/aiohttp/issues/5250.

A static resource like:
```
routes.static('/foo', '/www/foo')
```
currently matches `/foobar` but returns 404. This PR ensures that aforementioned static route *does not match* `/foobar`

If a user expected `/foobar` to load the file `/www/foo/ar`, this PR will violate their expectation.

## Related issue number

Fixes #5250

## Checklist

I'm not aware of any docs that need to change, but I didn't tick that box because I am not certain.


Dreamsorcerer:
The fact that `/foobar` is translated to `/ar` seems surprising, so I'd be inclined to say that it's unlikely that people are depending on this behaviour. Though it could be possible that someone uses paths like `/static-filename`...

I wonder if the `prefix` argument should be renamed to `path` though, to avoid confusion if we do change it...

danking:
OK, I’m fairly certain this is ready for review. Again, I don’t think the docs need to change because I think of this as a bug fix, but I’m not sure.

Dreamsorcerer:
@webknjaz Thoughts on this? Seems to resolve some surprising behaviour.
But, looks like it also adds a little overhead, which I assume is running on every request.

Maybe a more efficient solution would be to ensure that that static resource is always at the end of the list, so it's the last one checked? This sounds reasonable to me, as a web server should be handling static assets on a production system where performance matters, therefore the static resource would never be useful.

CHANGES/5250.bugfix
```diff
@@ -0,0 +1 @@
+StaticResource prefixes no longer match URLs with a non-folder prefix. For example `routes.static('/foo', '/foo')` no longer matches the URL `/foobar`. Previously, this would attempt to load the file `/foo/ar`.
```
webknjaz: This note should use RST, not Markdown: please use double backticks.

tests/test_web_urldispatcher.py
```diff
@@ -458,3 +458,18 @@ async def test_static_absolute_url(aiohttp_client: Any, tmp_path: Any) -> None:
...
+
+
+async def test_for_issue_5250(aiohttp_client: Any, tmp_path: Any) -> None:
+    app = web.Application()
+    app.router.add_static("/foo", tmp_path)
+
+    async def get_foobar(request):
+        return web.Response(body="success!")
+
+    app.router.add_get("/foobar", get_foobar)
```
webknjaz, line 471: I feel like there should be more tested cases and maybe some corner cases identified.

webknjaz:
@Dreamsorcerer I'm not sure about that. Making the priority assumptions should probably be on the user. Besides, this would be a breaking change for people who used to rely on the order.

CHANGES/5250.bugfix
asvetlov:
```suggestion
StaticResource prefixes no longer match URLs with a non-folder prefix. For example ``routes.static('/foo', '/foo')`` no longer matches the URL ``/foobar``. Previously, this would attempt to load the file ``/foo/ar``.
```

asvetlov (approved):

PR #5576, opened 2021-03-31
Add cov-dev makefile target

weastur:
## What do these changes do?

Add `cov-dev` target to makefile to run tests with coverage on dev host. Was removed in #5145, but stayed in [docs](https://docs.aiohttp.org/en/stable/contributing.html#tests-coverage). Just thought it's useful because now there is no way to get coverage except explicit pytest run

No.

## Related issue number

#5145 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ vvtest: .develop
 .PHONY: cov-dev
 cov-dev: .develop
 	@pytest --cov-report=html
-	@echo "open file://`pwd`/htmlcov/index.html"
+	@echo "xdg-open file://`pwd`/htmlcov/index.html"
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -97,6 +97,11 @@ vtest: .develop
 vvtest: .develop
 	@pytest -vv
 
+.PHONY: cov-dev
+cov-dev: .develop
+	@pytest --cov-report=html
+	@echo "open file://`pwd`/htmlcov/index.html"
```
webknjaz: This is macOS-only. There should be a fallback to `xdg-open` for GNU/Linux if we want this to be useful in general.

psapezhka: fixed

webknjaz (approved):

PR #5578, opened 2021-03-31
Added support orjson for dumps of response

KorsaR-ZN:
## What do these changes do?

Added support for JSONEncoder implementations that return bytes instead of string, for example, such as orjson

No

## Related issue number

No

## Checklist

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -173,7 +173,9 @@ class ClientWebSocketResponse:
         *,
         dumps: JSONEncoder = DEFAULT_JSON_ENCODER,
     ) -> None:
-        await self.send_str(dumps(data), compress=compress)
+        data = dumps(data)
+        data = data if isinstance(data, str) else data.decode("utf-8")
+        await self.send_str(data, compress=compress)
 
     async def close(self, *, code: int = WSCloseCode.OK, message: bytes = b"") -> bool:
         # we need to break `receive()` cycle first,
--- a/CHANGES/5578.feature
+++ b/CHANGES/5578.feature
@@ -1,0 +1,1 @@
+Added support for JSONEncoder implementations that return bytes instead of string, for example, such as orjson
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -262,6 +262,7 @@ Serhiy Storchaka
 Simon Kennedy
 Sin-Woo Bang
 Stanislas Plum
+Stanislav Kiryukhin (korsar-zn)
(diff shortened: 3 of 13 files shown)


webknjaz (changes requested):
I'm not sure this is a good idea. What stops you from making the encoder compliant? Are there any other non-compliant implementations or is it just this one?

aiohttp/client_ws.py
```diff
@@ -173,7 +173,9 @@ async def send_json(
         *,
         dumps: JSONEncoder = DEFAULT_JSON_ENCODER,
     ) -> None:
-        await self.send_str(dumps(data), compress=compress)
+        data = dumps(data)
+        data = data if isinstance(data, str) else data.decode("utf-8")
```
webknjaz, line 177: Assuming an arbitrarily chosen encoding is not a good idea.

KorsaR-ZN:
> I'm not sure this is a good idea. What stops you from making the encoder compliant? Are there any other non-compliant implementations or is it just this one?

I have not seen any other similar implementations.

In fact, the changes is required only in this [line](https://github.com/aio-libs/aiohttp/pull/5578/files#diff-f6f586b9b1a764ad0fc87850c8a70387dd1ac03f62c9c4ff69b7266303a20dcfL399-R398), 
so as not to do a double conversion, I originally did just that, see [previous](https://github.com/KorsaR-ZN/aiohttp/commit/78b7a29e32096748be4812ac9edd08cc04f6e83f) commit.

All other changes can be undone and a compatible encoder can be used there.

asvetlov:
Sorry, no. 
You support `orjson` *encoder* only without *decoding*. It looks weird.

In fact, you can write a simple function that accepts JSON dict and returns `orjson` encoded bytes for passing as `data` argument to `client.get()` and family.

There is no need for special support on the aiohttp side.

HMaker:
How to set the content-type header @asvetlov ? by also passing the `headers` dict?

I think a custom data payload is better
```python
import orjson
import aiohttp
from aiohttp.payload import BytesPayload

class ORJSONData(BytesPayload):
    def __init__(self, value) -> None:
        super().__init__(
            orjson.dumps(value),
            content_type="application/json",
            encoding="utf-8"
        )

    async with session.post(
        "http://json.example.com/",
        data=ORJSONData({
            "foo": "bar"
        })
     ) as response:
        jsonbody = await response.json(loads=orjson.loads)
 
 ```
 This reduces usability though, that's why builtin encoder/decoder support for orjson would be helpful.

PR #5583, opened 2021-04-02
Return a readily processed `FormData` instead of raising an exception

### LINKED ISSUE #5577: Incorrect reuse of FormData on certain redirects
🐞 **Describe the bug**

If a POST request results in a redirect response with response code 307 or 308, `aiohttp.client` may attempt to construct another POST request with the redirected URL using the same `data` as the first request. If the value passed as the `data` argument is an instance of `aiohttp.FormData`, this will fail with a somewhat cryptic exception: ["RuntimeError: Form data has been processed already"](https://github.com/aio-libs/aiohttp/blob/0c36ab6350f8c82ac467d23183eb42ead515754d/aiohttp/formdata.py#L132). 

💡 **To Reproduce**

1. Send a POST request with FormData to a server that will respond with code 307 or 308

💡 **Expected behavior**

Instead of reconstructing the body for each request, perhaps the request body could be reused. This would also resolve #5530, which appears to have a similar root cause.

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
  File "--", line --, in --
    async with session.post(url, data=data) as response:
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/client.py", line 1118, in __aenter__
    self._resp = await self._coro
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/client.py", line 493, in _request
    req = self._request_class(
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/client_reqrep.py", line 313, in __init__
    self.update_body_from_data(data)
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/client_reqrep.py", line 507, in update_body_from_data
    body = body()
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/formdata.py", line 169, in __call__
    return self._gen_form_data()
  File "/usr/local/lib/python3.8/dist-packages/aiohttp/formdata.py", line 133, in _gen_form_data
    raise RuntimeError("Form data has been processed already")
RuntimeError: Form data has been processed already
```

```console
$ python --version
Python 3.8.5
```

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.4.post0
Author: Nikolay Kim
License: Apache 2
Location: /usr/local/lib/python3.8/dist-packages
Requires: attrs, chardet, yarl, multidict, typing-extensions, async-timeout
```
```console
Version: 5.1.0
Author: Andrew Svetlov
License: Apache 2
Location: /usr/local/lib/python3.8/dist-packages
Requires: 
Required-by: yarl, aiohttp
```
```console
$ python -m pip show yarl
Name: yarl
Version: 1.6.3
Author: Andrew Svetlov
License: Apache 2
Location: /usr/local/lib/python3.8/dist-packages
Requires: idna, multidict
Required-by: aiohttp
```

📋 **Additional context**

This is a client issue.

For anyone else who has encountered this issue: Skipping the redirect by sending the original request to the redirected URL may be an adequate workaround at least in the case of a permanent redirect.

paulefoe:
## What do these changes do?

This changes the behavior of already processed data. Instead of raising an exception when trying to access already processed data -- return the data.

The POST redirects with FormData no longer raise exception.

## Related issue number

This commit fixes #5577 issue

## Checklist

### DIFF
--- a/CHANGES/5577.bugfix
+++ b/CHANGES/5577.bugfix
@@ -1,0 +1,1 @@
+Instead of raising ClientPayloadError, return already processed FormData.
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -15,7 +15,7 @@ import pytest
 from multidict import MultiDict
 
 import aiohttp
-from aiohttp import Fingerprint, ServerFingerprintMismatch, hdrs, web
+from aiohttp import Fingerprint, ServerFingerprintMismatch, hdrs, web, FormData
 from aiohttp.abc import AbstractResolver
 from aiohttp.client_exceptions import TooManyRedirects
 from aiohttp.test_utils import unused_port
@@ -980,6 +980,60 @@ async def test_HTTP_308_PERMANENT_REDIRECT_POST(aiohttp_client: Any) -> None:
     client = await aiohttp_client(app)
 
     resp = await client.post("/redirect", data={"some": "data"})
+    assert 200 == resp.status
+    assert 1 == len(resp.history)
+    txt = await resp.text()
+    assert txt == "POST"
+    resp.close()
+
+
+async def test_HTTP_307_REDIRECT_POST_FORMDATA(aiohttp_client: Any) -> None:
+    async def handler(request):
+        return web.Response(text=request.method)
+
+    async def redirect(request):
+        await request.read()
+        raise web.HTTPTemporaryRedirect(location="/")
+
+    app = web.Application()
+    app.router.add_post("/", handler)
+    app.router.add_post("/redirect", redirect)
+    client = await aiohttp_client(app)
+
+    data = FormData()
+    data.add_field(
+        'file',
+        io.BytesIO(),
+        filename='bytes_file',
+    )
+    resp = await client.post("/redirect", data=data)
+    assert 200 == resp.status
+    assert 1 == len(resp.history)
+    txt = await resp.text()
+    assert txt == "POST"
+    resp.close()
+
+
+async def test_HTTP_308_PERMANENT_REDIRECT_POST_FORMDATA(aiohttp_client: Any) -> None:
+    async def handler(request):
+        return web.Response(text=request.method)
+
+    async def redirect(request):
+        await request.read()
+        raise web.HTTPPermanentRedirect(location="/")
+
+    app = web.Application()
+    app.router.add_post("/", handler)
+    app.router.add_post("/redirect", redirect)
+    client = await aiohttp_client(app)
+
+    data = FormData()
+    data.add_field(
+        'file',
+        io.BytesIO(),
+        filename='bytes_file',
+    )
+    resp = await client.post("/redirect", data=data)
     assert 200 == resp.status
     assert 1 == len(resp.history)
     txt = await resp.text()
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -987,40 +987,15 @@ async def test_HTTP_308_PERMANENT_REDIRECT_POST(aiohttp_client: Any) -> None:
     resp.close()
 
 
-async def test_HTTP_307_REDIRECT_POST_FORMDATA(aiohttp_client: Any) -> None:
+@pytest.mark.xfail
+@pytest.mark.parametrize("redirect_type", [web.HTTPPermanentRedirect, web.HTTPTemporaryRedirect])
+async def test_HTTP_REDIRECT_POST_FORMDATA(redirect_type, aiohttp_client: Any) -> None:
     async def handler(request):
         return web.Response(text=request.method)
 
     async def redirect(request):
         await request.read()
-        raise web.HTTPTemporaryRedirect(location="/")
-
-    app = web.Application()
-    app.router.add_post("/", handler)
-    app.router.add_post("/redirect", redirect)
-    client = await aiohttp_client(app)
-
-    data = FormData()
-    data.add_field(
-        'file',
-        io.BytesIO(),
-        filename='bytes_file',
-    )
-    resp = await client.post("/redirect", data=data)
-    assert 200 == resp.status
-    assert 1 == len(resp.history)
-    txt = await resp.text()
-    assert txt == "POST"
-    resp.close()
-
-
-async def test_HTTP_308_PERMANENT_REDIRECT_POST_FORMDATA(aiohttp_client: Any) -> None:
-    async def handler(request):
-        return web.Response(text=request.method)
-
-    async def redirect(request):
-        await request.read()
-        raise web.HTTPPermanentRedirect(location="/")
+        raise redirect_type(location="/")
 
     app = web.Application()
     app.router.add_post("/", handler)
--- a/tests/test_formdata.py
+++ b/tests/test_formdata.py
@@ -97,6 +97,24 @@ async def test_mark_formdata_as_processed() -> None:
         await session.post(url, data=data)
         assert len(data._writer._parts) == 1
 
-        # second request doesn't append values to existing FormData
+        with pytest.raises(RuntimeError):
+            await session.post(url, data=data)
+
+
+@pytest.mark.xfail
+async def test_reuse_same_formadata() -> None:
+    async with ClientSession() as session:
+        url = "http://httpbin.org/anything"
+        data = FormData()
+        data.add_field("test", "test_value", content_type="application/json")
+
         await session.post(url, data=data)
+        assert len(data._writer._parts) == 1
+
+        with pytest.raises(RuntimeError):
+            # once the FormData is processed do not allow to add fields
+            data.add_field('other', 'value', content_type='application/json')
+
+        # second request doesn't append values to existing FormData
+        await session.post(f'{url}/some', data=data)
         assert len(data._writer._parts) == 1
--- a/tests/test_formdata.py
+++ b/tests/test_formdata.py
@@ -97,5 +97,6 @@ async def test_mark_formdata_as_processed() -> None:
         await session.post(url, data=data)
         assert len(data._writer._parts) == 1
 
-        with pytest.raises(RuntimeError):
-            await session.post(url, data=data)
+        # second request doesn't append values to existing FormData
+        await session.post(url, data=data)
+        assert len(data._writer._parts) == 1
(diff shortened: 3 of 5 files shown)


CHANGES/5577.bugfix
```diff
@@ -0,0 +1 @@
+Instead of raising ClientPayloadError, return already processed FormData.
```
webknjaz, line 1:
```suggestion
Reuse a pre-cached ``FormData`` instance when following HTTP POST redirects instead of raising ``ClientPayloadError``.
```

tests/test_formdata.py
```diff
@@ -97,5 +97,6 @@ async def test_mark_formdata_as_processed() -> None:
         await session.post(url, data=data)
         assert len(data._writer._parts) == 1
 
-        with pytest.raises(RuntimeError):
-            await session.post(url, data=data)
+        # second request doesn't append values to existing FormData
+        await session.post(url, data=data)
+        assert len(data._writer._parts) == 1
```
webknjaz, line 112: The test name suggests that the form data is checked for being processed and now it is not. Besides, this was probably a regression test that should remain unchanged under certain circumstances at least.

webknjaz (changes requested):
Having reflected on the related issue for a while, I'm not sure that this change is the best solution. We need more information about [the implications of switching this user-facing behavior](https://xkcd.com/1172/) in cases differing from handling HTTP 307/308 redirects.

First of all, let's invite the people involved with adding this check in the first place — @ulrikjohansson and @AtomsForPeace. In order to make the right decision, we need to first understand why it was added and make sure that the problem this check solves remain solved.
Refs:
* https://github.com/aio-libs/aiohttp/pull/4351
* https://github.com/aio-libs/aiohttp/issues/4345

Second, this is going to be a breaking change for non-redirect scenarios. Is it worth it? What cases will this affect?

Third, the issue is complaining about HTTP 307/308 specifically and the bugfix should fix those without breaking other flows that may rely on an exception being raised. This point deserves a lot of exploration.
We need to know how aiohttp behaves when different redirection types happen. Are all of these cases covered with tests? If not, we should write those tests first and check any corner cases that may come up in a similar fashion.
Then, we need to compare those behaviors to other HTTP client implementations and standards. Pay extra attention to cases where the client is allowed to change the method from POST to GET or similar. Refs:
* https://stackoverflow.com/a/5129140/595220
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307
* https://tools.ietf.org/html/rfc7231#section-6.4.7
* https://tools.ietf.org/html/rfc7538
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
* https://softwareengineering.stackexchange.com/a/99966/75899

All of the above should be carefully researched, summarized on GitHub, and handled.

paulefoe:
Good points! Thank you for your review. I'll address your comment in a period over a few days if you don't mind.

I think the original issue by @ulrikjohansson asked exactly for this, at least this is the expected behavior from the issue:

>When I repeat a call to ClientSession.request with the same arguments I expect to send a copy of the first request

>this is going to be a breaking change for non-redirect scenarios. Is it worth it? What cases will this affect?

This a nice guard mechanism, so I think the cases which are going to be affected are where the developer unintentionally made a mistake and sent the same request twice.


>the issue is complaining about HTTP 307/308 specifically and the bugfix should fix those without breaking other flows

In general, I can see how this fix could be useful not only for redirects but for any requests. Suppose you want to retry the POST request. Right now you will need to create a new FormData. 
Handling specifically 307/308 wouldn't probably be that hard, we can add the property `_is_redirected` to FormData and either raise an exception or return processed data.

On the other hand, retrying POST requests might not seem like a good idea, so maybe handling just redirects here is better.

I'll address the third point separately when I will dig a little bit through the links you've shared, implementation in other libraries and standards. This might take some time, but I think I will come back in 2-3 days at most

paulefoe:
Okay, so about 307

> If the 307 status code is received in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user, since this might change the conditions under which the request was issued. 
https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.8

However, on MDN:
>The method and the body of the original request are reused to perform the redirected request. The only difference between 307 and 302 is that 307 guarantees that the method and the body will not be changed when the redirected request is made
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307
same in spec
https://tools.ietf.org/html/rfc7231#section-6.4.7

Although this is not really contradicting w3, w3 just says that they shouldn't be performed automatically, meaning that the user should confirm resending POST request. 

For 308 everyone seems to agree.
>The request method and the body will not be altered
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308

>The user agent MAY use the Location field value for automatic redirection
https://tools.ietf.org/html/rfc7538#section-3

So my guess is that we should implement correct form data redirects definitely for 308 and maybe for 307 as well.

I'll do library comparison tomorrow

webknjaz:
> Although this is not really contradicting w3, w3 just says that they shouldn't be performed automatically, meaning that the user should confirm resending POST request.

There is no conflict here: RFC 7231 obsoletes RFC 2616, I only mentioned the link to the latter for the historic context mostly (just look at the header of https://tools.ietf.org/html/rfc2616 and you'll see). MDN references the newer standard and, as such, it does not contradict what the relevant RFC states. Also, often newer RFCs refer to things defined in earlier ones so it's useful to have both open, just in case.

paulefoe:
>There is no conflict here: RFC 7231 obsoletes RFC 2616, I only mentioned the link to the latter for the historic context mostly (just look at the header of https://tools.ietf.org/html/rfc2616 and you'll see). MDN references the newer standard and, as such, it does not contradict what the relevant RFC states. Also, often newer RFCs refer to things defined in earlier ones so it's useful to have both open, just in case.

Ah, I see. I'm kinda new to reading these rfc docs, but that seems really interesting, good to know. Thanks.

With regards to the libraries:
`Requests`:
on 308 would post the same data to the redirected URL
on 307 would post the same data to the redirected URL
on 301 would issue a get request to the redirected URL

`http.client` quite expectedly doesn't do any redirects.

`httpx`:
on 308 would post the same data to the redirected URL
on 307 would post the same data to the redirected URL
on 301 would issue a get request to the redirected URL

`axios` quite unexpectedly doesn't do any redirects on 301, 307, 308.

ulrikjohansson:
I'll chime in here as I was explicitly mentioned earlier in the thread.
For our use case (automatically retrying requests if they fail) this change would not have any adverse effects. Like @paulefoe noted, if this change makes it possible to re-send a POST request without having to recreate the FormData (with a deep copy for example), we could just remove that workaround code.

webknjaz:
@paulefoe all of those are libraries. What do web browsers do? FWIW we should probably ensure to align with httpx and requests. For this, could you analyze what tests are present and try to cover all the cases so that it's explicit what we support and what we don't?

@ulrikjohansson do you know how other async libs handle resending the data? @paulefoe it'd be interesting to have this in the comparison too.

paulefoe:
Yeah, sure, here's what browsers do.
Firefox:
on 301 would issue a get request to the redirected URL
on 307 would post the same data to the redirected URL
on 308 would post the same data to the redirected URL

Chromium:
on 301 would issue a get request to the redirected URL
on 307 would post the same data to the redirected URL
on 308 would post the same data to the redirected URL

I've added 2 test cases that fail on master [here](https://github.com/aio-libs/aiohttp/pull/5583/files#diff-0aecf79a65ea580060d8c239330193e86bd1463d05aa7e87611d8abca1c1e8ebR990)

aiohttp actually already has test cases for 308 and 307 [here](https://github.com/aio-libs/aiohttp/pull/5583/files#diff-0aecf79a65ea580060d8c239330193e86bd1463d05aa7e87611d8abca1c1e8ebR948), but they don't take into account the form data.

I could probably write a few more test cases tomorrow, that could be missing.

paulefoe:
One thing also to consider, is that issue that @ulrikjohansson had in my understanding is not really ok. The other libraries don't have such issues, for example, neither httpx, not requests, doesn't have a concept as a FormData exposed to the end-user, they just accept dictionaries. This might be less efficient, but you can send them as many times as you want.

That is just my 5 cents, maybe it doesn't have that much application in the real world.

webknjaz:
Based on our conversation, we need to have a (negative?) test where one would create a FormData instance, send it once, add a field or two to it, send it again.

tests/test_client_functional.py
```diff
@@ -987,6 +987,60 @@ async def redirect(request):
...
+        'file',
+        io.BytesIO(),
+        filename='bytes_file',
+    )
+    resp = await client.post("/redirect", data=data)
+    assert 200 == resp.status
+    assert 1 == len(resp.history)
+    txt = await resp.text()
+    assert txt == "POST"
+    resp.close()
```
webknjaz: Please merge these tests into one with `@pytest.mark.parametrize()`

paulefoe: Okay, I'll add a few more to describe how I see the desired behavior.

tests/test_formdata.py
```diff
@@ -99,3 +99,22 @@ async def test_mark_formdata_as_processed() -> None:
...
+        url = "http://httpbin.org/anything"
+        data = FormData()
+        data.add_field("test", "test_value", content_type="application/json")
+
+        await session.post(url, data=data)
+        assert len(data._writer._parts) == 1
+
+        with pytest.raises(RuntimeError):
+            # once the FormData is processed do not allow to add fields
+            data.add_field('other', 'value', content_type='application/json')
```
paulefoe, line 116: Alternatively, adding a new field could clear the writer and change `self._is_processd` to `False`

inyutin:
Is this PR still in development stage strictly because of lack of tests?
@paulefoe are you planning to finish this PR ? Can I help somehow ?

brewbytes-dev:
Is this work on hold and any chance to proceed?

Dreamsorcerer:
Given the lack of response, I'd suggest just forking from that branch and submitting another PR. We'll then close this one.

Issue #5518, opened 2021-03-06
Fix missing references in docs


webknjaz:
**:wave:  THIS IS A GOOD FIRST ISSUE**

🐞 **Describe the bug**

I noticed that the docs have some RST references to internal and external intersphinx objects. We should fix those and enable `nitpicky = True` by default.

_P.S. It is okay to send small PRs to fix references gradually_

💡 **To Reproduce**

Uncomment `nitpicky = True` in `docs/conf.py` and run `make doc`.

💡 **Expected behavior**

No errors.

📋 **Logs/tracebacks**

```console
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:22: WARNING: py:class reference target not found: AbstractRouter
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:22: WARNING: py:meth reference target not found: AbstractRouter.resolve
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:22: WARNING: py:class reference target not found: AbstractMatchInfo
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:26: WARNING: py:meth reference target not found: AbstractMatchInfo.handler
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:26: WARNING: py:attr reference target not found: AbstractMatchInfo.http_exception
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:30: WARNING: py:attr reference target not found: AbstractMatchInfo.http_exception
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:30: WARNING: py:meth reference target not found: AbstractMatchInfo.handler
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:30: WARNING: py:attr reference target not found: AbstractMatchInfo.http_exception
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:52: WARNING: py:class reference target not found: AbstractMatchInfo
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:57: WARNING: py:meth reference target not found: AbstractRouter.resolve
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:105: WARNING: py:attr reference target not found: ClientSession.cookie_jar
/home/wk/src/github/aio-libs/aiohttp/docs/abc.rst:169: WARNING: py:class reference target not found: RequestHandler
[...]

PR #5579, opened 2021-03-31
Fix missing references in client_advanced.rst

l1storez:
## What do these changes do?

This change corrects multiple unrendered intersphinx class reference in the `client_advanced.rst` document.

## Are there changes in behavior for the user?

No

## Related issue number

#5518 

## Checklist

- [x] I think the code is well written
- [ ] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #5588, opened 2021-04-04
Fix `StreamReader` reference @ `client_quickstart`

sgr81:
## What do these changes do?

This change corrects multiple unrendered intersphinx class reference in the `client_quickstart.rst` document.

No.

## Related issue number

#5518

## Checklist

### DIFF
--- a/docs/client_quickstart.rst
+++ b/docs/client_quickstart.rst
@@ -318,7 +318,7 @@ You can set the ``filename`` and ``content_type`` explicitly::
     await session.post(url, data=data)
 
 If you pass a file object as data parameter, aiohttp will stream it to
-the server automatically. Check :class:`~aiohttp.streams.StreamReader`
+the server automatically. Check :class:`~aiohttp.StreamReader`
 for supported format information.
 
 .. seealso:: :ref:`aiohttp-multipart`
@@ -327,7 +327,7 @@ for supported format information.
 Streaming uploads
 =================
 
-:mod:`aiohttp` supports multiple types of streaming uploads, which allows you to
+:mod:`~aiohttp.web` supports multiple types of streaming uploads, which allows you to
 send large files without reading them into memory.
 
 As a simple case, simply provide a file-like object for your body::
@@ -366,7 +366,7 @@ can chain get and post requests together::
 WebSockets
 ==========
 
-:mod:`aiohttp` works with client websockets out-of-the-box.
+:mod:`~aiohttp.web` works with client websockets out-of-the-box.
 
 You have to use the :meth:`aiohttp.ClientSession.ws_connect` coroutine
 for client websocket connection. It accepts a *url* as a first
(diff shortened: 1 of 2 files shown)


docs/client_quickstart.rst
```diff
@@ -327,7 +327,7 @@ for supported format information.
 Streaming uploads
 =================
 
-:mod:`aiohttp` supports multiple types of streaming uploads, which allows you to
+:mod:`~aiohttp.web` supports multiple types of streaming uploads, which allows you to
```
webknjaz: I don't think this is the right solution:
1) This will render as a link with an anchor `web` but the original one was `aiohttp`. To keep this, you could make it explicit:
```suggestion
:mod:`aiohttp <aiohttp.web>` supports multiple types of streaming uploads, which allows you to
```
2) I'm not sure that the original intent was to link `aiohttp.web` module. It looks like it was to refer to the `aiohttp` package.

The problem is that the module is not documented. So I feel like it should be documented and then, it could be linked. But meanwhile, it could be added to the ignore list as shown in https://github.com/aio-libs/aiohttp/pull/5579 and dealt with later, after all the existing references are corrected.

docs/client_quickstart.rst
```diff
@@ -318,7 +318,7 @@ You can set the ``filename`` and ``content_type`` explicitly::
     await session.post(url, data=data)
 
 If you pass a file object as data parameter, aiohttp will stream it to
-the server automatically. Check :class:`~aiohttp.streams.StreamReader`
+the server automatically. Check :class:`~aiohttp.StreamReader`
```
webknjaz, line 321: :+1:

docs/client_quickstart.rst
```diff
@@ -366,7 +366,7 @@ can chain get and post requests together::
 WebSockets
 ==========
 
-:mod:`aiohttp` works with client websockets out-of-the-box.
+:mod:`~aiohttp.web` works with client websockets out-of-the-box.
```
webknjaz: see the comment above wrt this

docs/client_quickstart.rst
webknjaz: Or maybe add `.. currentmodule:: aiohttp` at the top of `index.rst` instead...

sgr81: Alright, I will add it to the ignore list and change it to the way it was `~aiohttp.web`.

docs/conf.py
```diff
@@ -349,3 +349,5 @@
 
 # -------------------------------------------------------------------------
 # nitpicky = True
+
+nitpick_ignore = [("py:mod", "aiohttp")]
```
webknjaz: Let's leave comments next to each ignore so that in the future we'd know how to fix this:
```suggestion
nitpick_ignore = [
    ("py:mod", "aiohttp"),  # undocumented, no `.. currentmodule:: aiohttp` in docs
]
```

webknjaz (approved):

PR #5625, opened 2021-04-17
Remove legacy examples

Dreamsorcerer:
Does legacy mean these can be removed?

They appear to no longer work and are causing a blocker to upgrading Mypy.


webknjaz:
@Dreamsorcerer have you checked if there are any equivalent examples elsewhere? Like the new ones in this repo, in the docs, or in https://github.com/aio-libs/aiohttp-demos?

Dreamsorcerer:
I didn't really spend any time understanding the code, beyond noticing all the errors that mypy was reporting.
There's not many comments or anything to explain what these examples actually do, so it's unclear to me whether they are useful or still relevant, or whether alternative versions already exist.

I was hoping you might have a better idea why they are under a `legacy` directory. I thought maybe they were only relevant to older versions of aiohttp or something.

webknjaz:
> I thought maybe they were only relevant to older versions of aiohttp or something.

Probably, but I never touched files in the example dirs IIRC. If you want to know more, you'd have to resort to Git archeology.

FWIW my only concern is removing app examples that may otherwise be unavailable. Can't mypy just ignore that subdir?
Also, I think it's acceptable to just move the legacy dir under the demo repo (and preserve it as `legacy/`) — just add a good commit message for future reference.

Dreamsorcerer:
> Probably, but I never touched files in the example dirs IIRC. If you want to know more, you'd have to resort to Git archeology.

Unfortunately, that just shows they were considered 'obsolete', but doesn't explain why they were kept in the first place:
https://github.com/aio-libs/aiohttp/commit/e10fc28f9c2ec6458b633633884dfc4efbcac7f5#diff-6a83e9296f6b57991c897cd0fcbbdfe0efebde8e9a64868467e5aa1acb9ba535

That was also 5 years ago, so I feel it's unlikely to still be useful now.

> Can't mypy just ignore that subdir?

Yes, we can do something like that, I'm just not sure if there's any point in keeping around broken code. If they're useful, we should probably fix them, if not, we should remove them. The advantage of Mypy is that it can ensure that the examples are kept in sync with any changes to the codebase, so we don't end up with broken examples like this.

Dreamsorcerer:
> That was also 5 years ago, so I feel it's unlikely to still be useful now.

I mean, if these were obsolete before v1, what are the chances they will work at all in v4? :P

webknjaz:
If the ideas they demonstrate aren't easily discoverable, I suppose they need to be fixed. Could you explore this more?

asvetlov:
Agree, legacy examples are broken; they use dropped API

PR #5637, opened 2021-04-20
Update client/server implementation in the autobahn tests

### LINKED ISSUE #5606: Autobahn test needs updating
The `_make_handler()` method was removed in this PR: https://github.com/aio-libs/aiohttp/pull/3939

But, is still used at: https://github.com/aio-libs/aiohttp/blob/master/tests/autobahn/server.py#L37

This needs updating, if anybody would like to take on this small task.

References to `aiohttp.ws_connect()` appear to also need updating:
https://github.com/aio-libs/aiohttp/blob/master/tests/autobahn/client.py#L9

wefatherley:
## What do these changes do?
Alters `autobahn` client and server tests by updating deprecated syntax

No, there should not be

## Related issue number
Resolves #5606

## Checklist

### DIFF
--- a/tests/autobahn/client.py
+++ b/tests/autobahn/client.py
@@ -6,31 +6,32 @@ import aiohttp
 
 
 async def client(loop, url, name):
-    ws = await aiohttp.ws_connect(url + "/getCaseCount")
-    num_tests = int((await ws.receive()).data)
-    print("running %d cases" % num_tests)
-    await ws.close()
+    async with aiohttp.ClientSession() as session:
+        async with session.ws_connect(url + "/getCaseCount") as ws:
+            num_tests = int((await ws.receive()).data)
+            print("running %d cases" % num_tests)
+            await ws.close()
 
-    for i in range(1, num_tests + 1):
-        print("running test case:", i)
-        text_url = url + "/runCase?case=%d&agent=%s" % (i, name)
-        ws = await aiohttp.ws_connect(text_url)
-        while True:
-            msg = await ws.receive()
+            for i in range(1, num_tests + 1):
+                print("running test case:", i)
+                text_url = url + "/runCase?case=%d&agent=%s" % (i, name)
+                ws = await session.ws_connect(text_url)
+                while True:
+                    msg = await ws.receive()
 
-            if msg.type == aiohttp.WSMsgType.TEXT:
-                await ws.send_str(msg.data)
-            elif msg.type == aiohttp.WSMsgType.BINARY:
-                await ws.send_bytes(msg.data)
-            elif msg.type == aiohttp.WSMsgType.CLOSE:
-                await ws.close()
-                break
-            else:
-                break
+                    if msg.type == aiohttp.WSMsgType.TEXT:
+                        await ws.send_str(msg.data)
+                    elif msg.type == aiohttp.WSMsgType.BINARY:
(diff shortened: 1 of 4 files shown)


wfatherley:
Used the `crosbario/autobahn-testsuite:0.8.2` Docker image.

tests/autobahn/client.py
```diff
@@ -6,31 +6,32 @@
...
-                break
-
-    url = url + "/updateReports?agent=%s" % name
-    ws = await aiohttp.ws_connect(url)
-    await ws.close()
+    async with aiohttp.ClientSession() as session:
+        async with session.ws_connect(url + "/getCaseCount") as ws:
+            num_tests = int((await ws.receive()).data)
+            print("running %d cases" % num_tests)
+            await ws.close()
```
greshilov: `ws` object automatically closes after the exit from the `session.ws_connect` context manager. So maybe it makes sense to replace all of the `ws_connect / ws.close()` pairs with separate context managers, like so:

```python3
async with session.ws_connect(url + "/getCaseCount") as ws:
    num_tests = int((await ws.receive()).data)
    print("running %d cases" % num_tests)
    
for i in range(1, num_tests + 1):
    ...
    async with session.ws_connect(text_url) as ws:
        ...
 ```

File: tests/autobahn/client.py [hunk 2]
```diff
@@ -6,31 +6,32 @@
...
+
+            for i in range(1, num_tests + 1):
+                print("running test case:", i)
+                text_url = url + "/runCase?case=%d&agent=%s" % (i, name)
+                ws = await session.ws_connect(text_url)
+                while True:
```
greshilov: This `while` loop can be replaced with the  `__aiter__` call on the `ws` object. The `CLOSE` message will be handled by the library :) 

```python3
async for msg in ws:
        await ws.send_str(msg.data)
    elif msg.type == aiohttp.WSMsgType.BINARY:
        await ws.send_bytes(msg.data)
    else:
        break
```

greshilov:
@webknjaz PTAL

tests/autobahn/client.py
wfatherley: Not sure why I didn't notice the assignment to `ws` nested in the outer `ws` context, or why the testclient returned OK on all the test cases, but I've made these changes :)

tests/autobahn/client.py
wfatherley: That is much cleaner, thank you

greshilov (approved):
Thanks!

webknjaz (approved):

Dreamsorcerer:
What's happening with these tests? There was a build failure for Windows, so I tried rerunning them, but now they aren't running at all...

webknjaz:
I'm curious if that's related with today's GHA outage

webknjaz:
@wefatherley mind backporting this?

wfatherley:
Sure thing

wfatherley:
It seems swapping out the deprecated `_make_handler` syntax can be backported to 3.2, while swapping out deprecated `ws_connect` syntax can be backported to 3.0? Does the cherry picker handle this on it's own?

webknjaz:
It only does `git cherry-pick -x` and that's about it. Normally, if automatic backporting fails, it means that there's manual adjustments to be made.

Issue #5641, opened 2021-04-23
Add RawHeaders to BodyPartReader and MultipartReader 


ikrivosheev:
I need to get RawHeaders from BodyPartReader and from MultipartReader

PR #5642, opened 2021-04-23
Add RawHeaders to BodyPartReader and to MultipartReader

ikrivosheev:
## What do these changes do?

Add RawHeader to BodyPartReader and to MultipartReader

## Related issue number

https://github.com/aio-libs/aiohttp/issues/5641

## Checklist

### DIFF
--- a/CHANGES/5641.feature
+++ b/CHANGES/5641.feature
@@ -1,0 +1,1 @@
+Add RawHeaders to BodyPartReader and MultipartReader.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -139,6 +139,7 @@ Illia Volochii
 Ilya Chichak
 Ilya Gruzinov
 Ingmar Steen
+Ivan Krivosheev
 Ivan Larin
 Jacob Champion
 Jaesung Lee
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -36,7 +36,6 @@ from .hdrs import (
 )
 from .helpers import CHAR, TOKEN, parse_mimetype, reify
 from .http import HeadersParser
-from .typedefs import RawHeaders
 from .payload import (
     JsonPayload,
     LookupError,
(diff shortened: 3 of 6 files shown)


paulefoe:
What is the context for this change? Why do you need it?

ikrivosheev:
@paulefoe, I have ICAP server. I need to generate ICAP Response. ICAP packet consist of HTTP request or (and) HTTP response. In order not to encode headers, i need raw headers.

ikrivosheev:
@asvetlov , @webknjaz can you review this PR?

webknjaz (changes requested):
I appreciate the effort but I'd recommend consulting on the design and desirability of such an impactful change.

In the issue, you wrote
> I need to get RawHeaders from BodyPartReader and from MultipartReader

But as maintainers, we must consider what an average user needs, not one in a million. Any change like this will affect all the users and, of course, we wouldn't want such a penalty on them.

Moreover, it's best to explain your motivation with something more than "I want this" because the solution you invented may be architecturally wrong and nobody would be able to judge whether what you're proposing is even the right thing to do. Start with what you're attempting to solve with all the details and constraints, and code snippets. And only then follow up with how you envision that, including an explanation of why you think this proposal is good. It's also recommended to think about whether anybody besides you would benefit from this and whether it would be harmful in some cases. Of course, we'll help with the latter during the discussion/reviews but it's always preferable to learn such things before investing much time/effort into the implementation.

> I need to generate ICAP Response. ICAP packet consist of HTTP request or (and) HTTP response.

aiohttp is an HTTP server (and a client for that matter). Supporting other protocols is usually out of scope. And since we care a lot about the performance and memory consumption penalties, this is something we take into account when deciding whether the proposal is worthy.

I've marked some problematic changes in the comments below but besides that, I'm going to leave this for @asvetlov to accept or reject.

P.S. When filing issues and PR descriptions, it's a good tone to actually fill them out. Otherwise, you're forcing the reviewers to re-ask those same questions, increasing the latency of getting to the final decision.

Issue #5571, opened 2021-03-29
Client: ClientSession will not send secure cookie to localhost on unsecure connection.


DanielDewberry:
A ClientSession with an unsafe cookiejar will not send a secure cookie to a localhost endpoint, when that connection is over http (i,e. unsecure).

According to [Mozilla documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#restrict_access_to_cookies):

> A cookie with the Secure attribute is sent to the server only with an encrypted request over the HTTPS protocol, never with unsecured HTTP (except on localhost), and therefore can't easily be accessed by a man-in-the-middle attacker. Insecure sites (with http: in the URL) can't set cookies with the Secure attribute

Note the **(except on localhost)** clause.

💡 **To Reproduce**
The following example sets up a test (`TestIntegration::test_cookie_is_sent_server`) which fails.  Supplementary tests are included to demonstrate the setup operates as expected.  Scroll to the end of the codeblock to see that particular test.

```py
#! /usr/bin/env pytest
# -*- coding: utf-8 -*-

"""Test module."""

import base64
import http
from typing import cast

from aiohttp import ClientSession
from aiohttp import web
from aiohttp.cookiejar import CookieJar
from aiohttp.test_utils import TestClient

from aiohttp_session import get_session, setup
from aiohttp_session.cookie_storage import EncryptedCookieStorage

from cryptography import fernet

import pytest


fernet_key = fernet.Fernet.generate_key()


async def index_get_handler(request: web.Request) -> web.Response:
    session = await get_session(request)
    secret_key = session.get('session_id') or 'not set'
    key = 'key: {}'.format(secret_key)

    response = web.Response(
        status=http.HTTPStatus(200),
        headers={'Content-Type': 'text/plain; charset=UTF-8',
                 },
        body=key,
    )

    return response


async def login_post_handler(request: web.Request) -> web.Response:
    session = await get_session(request)
    session['session_id'] = 'some secret data'

    response = web.Response(
        status=http.HTTPStatus(303),
        headers={'Content-Type': 'text/plain; charset=UTF-8',
                 'Location': '/',
                 },
    )

    return response


@pytest.fixture
def application_instance() -> web.Application:
    """Generate the application instance."""
    application = web.Application()

    application.add_routes(
        [
            web.get('/', index_get_handler, name='index'),
            web.post('/login', login_post_handler, name='login'),
        ]
    )

    setup(
        app=application,
[...]

PR #5652, opened 2021-04-26
tests/test_cookiejar.py: add test

DanielDewberry:
## What do these changes do?

In accordance with https://github.com/aio-libs/aiohttp/issues/5571, this
commit adds a test to ensure secure cookies are not filtered from an unsafe
cookiejar when given an unsecured endpoint.

Rationale:

According to Mozilla documentation:
> A cookie with the Secure attribute is sent to the server only with an
> encrypted request over the HTTPS protocol, never with unsecured HTTP
> (except on localhost), and therefore can't easily be accessed by a
> man-in-the-middle attacker. Insecure sites (with http: in the URL) can't
> set cookies with the Secure attribute

Note the "(except on localhost)".  In addition, RFC 6265 section-4.1.2.5 states:

> The Secure attribute limits the scope of the cookie to "secure"
> channels (where "secure" is defined by the user agent). When a
> cookie has the Secure attribute, the user agent will include the
> cookie in an HTTP request only if the request is transmitted over a
> secure channel (typically HTTP over Transport Layer Security (TLS)
> [RFC2818]).
> Although seemingly useful for protecting cookies from active network
> attackers, the Secure attribute protects only the cookie's
> confidentiality

Note "(where "secure" is defined by the user agent)".  The behaviour
this commit tests for is therefore an engineer's decision, not an IETF
standard.

An addition to the test suite, adding a failing test

## Related issue number

 https://github.com/aio-libs/aiohttp/issues/5571

## Checklist

### DIFF
--- a/tests/test_cookiejar.py
+++ b/tests/test_cookiejar.py
@@ -756,3 +756,36 @@ async def test_cookie_jar_clear_domain() -> None:
     assert morsel.value == "bar"
     with pytest.raises(StopIteration):
         next(iterator)
+
+
+async def test_secure_cookie_not_filtered_from_unsafe_cookiejar_when_given_unsecured_endpoint() -> None:
+    """Secure SimpleCookie should not be filtered from unsafe CookieJar when given an unsecured endpoint.
+
+    There are times when sending a secure cookie to an unsecured endpoint is desireable.  Such an
+    occasion is during testing. RFC 6265 section-4.1.2.5 states that this behaviour is a decision
+    based on the trust of a network by the user agent.
+    """
+    endpoint = 'http://127.0.0.1/'
+
+    secure_cookie = SimpleCookie(
+        "cookie-key=cookie-value; HttpOnly; Path=/; Secure",
+    )
+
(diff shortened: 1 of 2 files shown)


tests/test_cookiejar.py
```diff
@@ -756,3 +756,36 @@ async def test_cookie_jar_clear_domain() -> None:
     assert morsel.value == "bar"
     with pytest.raises(StopIteration):
         next(iterator)
+
+
+async def test_secure_cookie_not_filtered_from_unsafe_cookiejar_when_given_unsecured_endpoint() -> None:
```
Dreamsorcerer, line 761: unsecured -> insecure

DanielDewberry: I chose the word unsecured as it has the following meaning, which insecure seems not to
> not made secure or safe.

Just letting you know that it was intentional :)

Dreamsorcerer: Yeah, but insecure is more standard. insecure means not secure, liable to risk or danger. HTTP connections are inherently insecure, not just unsecured.

PR #5678, opened 2021-05-07
Use an SVG logo across the docs

himanshu007-creator:
## What do these changes do?

This patch switches the png logo with an SVG one.

## Are there changes in behavior for the user?

The logo in the docs looks nicer when zoomed in.

## Related issue number

Resolves #5677

## Checklist

- [ ] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

Issue #5677, opened 2021-05-07
[TODO] Use an SVG logo in the Sphinx docs


webknjaz:
🐣 **Is your feature request related to a problem? Please describe.**

I noticed that the sphinx docs use a non-scalable png file as a logo. It needs to be changed.

💡 **Describe the solution you'd like**

Use https://github.com/aio-libs/aiohttp/blob/master/docs/aiohttp-plain.svg in sphinx instead.

❓ **Describe alternatives you've considered**

N/A

📋 **Additional context**

N/A

PR #5709, opened 2021-05-14
Add scalable version of the logo

The-Compiler:
I needed a scalable version for a (non-public) presentation, and 128x128 is pretty small.

I couldn't find a source for the logo anywhere, so I decided to quickly redraw it based on the original:

image

and arrived at something which looks pretty close:

aiohttp-svg

I thought I'd add it to the repo in case anyone else needs it too - even though it's not actually used in the docs, next to the .png is the best place to put it, IMHO. If someone had the original logo in a scalable format, that'd be even better, of course.

### DIFF
--- a/docs/_static/aiohttp.svg
+++ b/docs/_static/aiohttp.svg
@@ -1,0 +1,157 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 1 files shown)


webknjaz:
Hey Florian, I recently discovered that the SVG logo is actually present in this repo, just isn't used in sphinx for no reason :man_shrugging:, and it is located poorly — https://github.com/aio-libs/aiohttp/blob/master/docs/aiohttp-plain.svg.
I've filed an issue to make it actually used (https://github.com/aio-libs/aiohttp/issues/5677) and somebody even attempted to address it (https://github.com/aio-libs/aiohttp/pull/5678).

That said, I'm going to close this PR. Thanks for caring!

The-Compiler:
Ah, turns out I had searched for "logo" in the repo, but not for "svg"... Even better, thanks! :+1:

PR #5731, opened 2021-05-24
Fix a test for HTTPUnauthorized with a Unicode string body argument

### LINKED ISSUE #5657: Raising a HTTPUnauthorized caused error after upgrading from 3.6.2 to 3.7.4.post0
🐞 **Describe the bug**

Raising a HTTPUnauthorized with a body="string" is giving an exception, as you can see in the traceback. This started to happen after I've upgraded from 3.6.2 to 3.7.4.post0. Looks to me that, as the body I return is **str**, it creates a StringPayload and still it tries to decode() it as it was **bytes-like** object

My piece of code that works in 3.6.2: `raise web.HTTPUnauthorized(body='Not logged in')`
Workaround to work in 3.7.4.post0: `raise web.HTTPUnauthorized(body=b'Not logged in')`

💡 **To Reproduce**

1. Try to raise a HTTPUnauthorized with a string body

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
  File "aiohttp/web_protocol.py", line 422, in _handle_request
  File "aiohttp/web_app.py", line 499, in _handle
  File "aiohttp/web_middlewares.py", line 119, in impl
  File "aiohttp_session/__init__.py", line 172, in factory
  File "aiohttp_session/__init__.py", line 154, in factory
  File "http_server.py", line 1869, in route_get_info
  File "http_server.py", line 461, in validate_identity
aiohttp.web_exceptions.HTTPUnauthorized: Unauthorized

  File "aiohttp/web_protocol.py", line 485, in start
  File "aiohttp/web_protocol.py", line 427, in _handle_request
  File "aiohttp/web_response.py", line 650, in text
AttributeError: 'StringPayload' object has no attribute 'decode'
```

```console
$ python --version
Python 3.8.0
```

```console
aiohttp==3.7.4.post0
aiohttp-jinja2==1.4.2
aiohttp-security==0.4.0
aiohttp-session==2.9.0
multidict==5.1.0
yarl==1.6.3
```

paulefoe:
## What do these changes do?

Allows previous behavior when constructing HTTPUnauthorized with a body argument as a string

Users are now able to send HTTPUnauthorized response with a string

https://github.com/aio-libs/aiohttp/issues/5657

This commit fixes #5657

## Checklist

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -697,6 +697,8 @@ class Response(StreamResponse):
     def text(self) -> Optional[str]:
         if self._body is None:
             return None
+        if isinstance(self._body, Payload):
+            return self._body.value.decode(self._body.encoding)
         return self._body.decode(self.charset or "utf-8")
 
     @text.setter
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -576,9 +576,6 @@ class StreamResponse(BaseClass, HeadersMixin):
     def __eq__(self, other: object) -> bool:
         return self is other
 
-        # if isinstance(self._body, Payload):
-        #     return self._body.value.decode(self._body.encoding)
-
 
 class Response(StreamResponse):
     def __init__(
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -576,6 +576,9 @@ class StreamResponse(BaseClass, HeadersMixin):
     def __eq__(self, other: object) -> bool:
         return self is other
 
+        # if isinstance(self._body, Payload):
+        #     return self._body.value.decode(self._body.encoding)
+
 
 class Response(StreamResponse):
     def __init__(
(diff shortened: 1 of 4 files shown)


paulefoe:
I'm not sure how to fix mypy here, should I move value property to StringPayload and say that it returns str?

aiohttp/web_response.py
```diff
@@ -697,6 +697,8 @@ def body(
     def text(self) -> Optional[str]:
         if self._body is None:
             return None
+        if isinstance(self._body, Payload):
```
webknjaz: Could you explain this bit? I expected the fix to be in the exception initializer.

paulefoe:
Now that I look at it, I think that the test is wrong and everything works correctly on 3.8.

The reason is, on master, the response from `HTTPUnauthorized` uses `exc.text` which I tested here, but in 3.8 it's not used, the response forms without using the text property of exception.

I'll look tomorrow if the test can be rewritten or should be deleted entirely.

paulefoe:
I think I should remove the last line for the test, where it access `text` test just the initialization and status, or maybe just the initialization.

```python
def test_unicode_text_body_unauthorized() -> None:
    """
    Test that HTTPUnauthorized can be initialized with a string.
    """
    with pytest.warns(
        DeprecationWarning, match="body argument is deprecated for http web exceptions"
    ):
        resp = web.HTTPUnauthorized(body="text")
    assert resp.status == 401
```

webknjaz (approved):

PR #5763, opened 2021-06-03
add test, typedef and session implementation

shobu13:
- new test solution
- new type def
- modification of existing session class to implement params argument similar to actual header implementation.

Why this pull request:
to have a simple and elegant way to share common query parameter with session object, like actual implementation of header.

typical usage:

```python
with session(headers=headers, params={apikey=someRandomKey}) as session:
    response = await session.get("http://randomsecurewebsite")
    response2 = await session.post("http://otherSecureWebsite")
```
instead of

```python
some_random_key = ''
with session(headers=headers) as session:
    response = await session.get("http://randomsecurewebsite", params={some_random_key})
    response2 = await session.post("http://otherSecureWebsite", params={some_random_key})
```

thank for reading, feel free to respond.

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -89,7 +89,7 @@ from .http import WS_KEY, HttpVersion, WebSocketReader, WebSocketWriter
 from .http_websocket import WSHandshakeError, WSMessage, ws_ext_gen, ws_ext_parse
 from .streams import FlowControlDataQueue
 from .tracing import Trace, TraceConfig
-from .typedefs import JSONEncoder, LooseCookies, LooseHeaders, StrOrURL
+from .typedefs import JSONEncoder, LooseCookies, LooseHeaders, LooseParams, StrOrURL
 
 __all__ = (
     # client_exceptions
@@ -129,7 +129,6 @@ __all__ = (
     "ClientTimeout",
     "request",
 )
-
 
 try:
     from ssl import SSLContext
@@ -184,6 +183,7 @@ class ClientSession:
         "_trust_env",
         "_default_headers",
         "_skip_auto_headers",
+        "_default_params",
         "_request_class",
         "_response_class",
         "_ws_response_class",
@@ -197,6 +197,7 @@ class ClientSession:
         connector: Optional[BaseConnector] = None,
         cookies: Optional[LooseCookies] = None,
         headers: Optional[LooseHeaders] = None,
+        params: Optional[LooseParams] = None,
         skip_auto_headers: Optional[Iterable[str]] = None,
         auth: Optional[BasicAuth] = None,
         json_serialize: JSONEncoder = json.dumps,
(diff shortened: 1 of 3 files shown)


aiohttp/client.py
```diff
@@ -130,7 +130,6 @@
     "request",
 )
 
-
```
webknjaz, line 133: Please don't include formatting changes in the same PR that has functional changes.

aiohttp/client.py
```diff
@@ -854,7 +865,9 @@ async def _ws_connect(
             )
 
     def _prepare_headers(self, headers: Optional[LooseHeaders]) -> "CIMultiDict[str]":
-        """Add default headers and transform it to CIMultiDict"""
+        """
+        Add default headers and transform it to CIMultiDict
+        """
```
webknjaz, line 870: Why did you change this?

aiohttp/client.py
```diff
@@ -1042,7 +1073,6 @@ async def __aexit__(
 
 
 class _BaseRequestContextManager(Coroutine[Any, Any, _RetType], Generic[_RetType]):
-
```
webknjaz, line 1045: This is a wrong formatting change.

aiohttp/client.py
```diff
@@ -1099,7 +1129,6 @@ async def __aexit__(
 
 
 class _SessionRequestContextManager:
-
```
webknjaz, line 1102: Don't do this either.

tests/test_client_session.py
```diff
@@ -570,7 +574,6 @@ async def on_request_headers_sent(session, context, params):
     async with session.post(
         "/", data=body, trace_request_ctx=trace_request_ctx
     ) as resp:
-
```
webknjaz, line 573: Please don't make random formatting changes in unrelated places.

tests/test_client_session.py
```diff
@@ -663,7 +666,6 @@ async def new_headers(session, trace_config_ctx, data):
 
 def test_client_session_inheritance() -> None:
     with pytest.raises(TypeError):
-
```
webknjaz, line 666: Please don't make random formatting changes in unrelated places.

tests/test_client_session.py
```diff
@@ -706,3 +708,13 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+
+
+async def test_http_GET_with_params_argument(session_params: Any) -> None:
+    try:
+        async with session_params as session:
+            response: aiohttp.ClientResponse = await session.get("http://example.com")
+
+            assert response.url, f"http://example.com/x=2"
+    except Exception:
+        pass
```
webknjaz, line 720: It's a very bad idea to ignore errors in tests. This way you don't test anything at all.

tests/test_client_session.py
```diff
@@ -706,3 +708,13 @@ async def test_requote_redirect_url_default_disable() -> None:
...
     assert not session.requote_redirect_url
     await session.close()
+
+
+async def test_http_GET_with_params_argument(session_params: Any) -> None:
+    try:
+        async with session_params as session:
+            response: aiohttp.ClientResponse = await session.get("http://example.com")
+
+            assert response.url, f"http://example.com/x=2"
```
webknjaz, line 718: You probably wanted to do
```suggestion
        assert response.url == "http://example.com/x=2"
```

webknjaz (changes requested):
Not sure if we need this sort of change in the framework shipped to everybody. One could just inherit the objects if they need something like this in their project.

Do you have any examples of web frameworks having this feature?

PR #5764, opened 2021-06-04
Allow keep-alive with proxy for message with appropriate header

avderin:
### What do these changes do?

Once ago there was a merged pr - https://github.com/aio-libs/aiohttp/pull/3070 that disabled keep-alive connections through the proxy. There was a suggestion to make this functionality switchable in discussion.

Keep-alive with the proxy sometimes is need. E.g. to establish ntlm authorization the client must send 2 requests in one connection (a handshake). Current aiohttp logic restricts it, closing the connection on the first request.

I suggest to allow keep-alive connection when client explicitly sets keep-alive header.

### Related issue number

https://github.com/aio-libs/aiohttp/issues/5765


webknjaz:
A change like this would require regression tests.

Dreamsorcerer:
Superseded by #8920.

PR #5793, opened 2021-06-13
Pass timeout to initial websocket request

quthla:
## What do these changes do?
Pass timeout to initial websocket request

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -743,6 +743,7 @@ class ClientSession:
             proxy_auth=proxy_auth,
             ssl=ssl,
             proxy_headers=proxy_headers,
+            timeout=timeout,
         )
 
         try:


paulefoe:
Hey, thank you for your contribution!

Make sure you read the docs on how to contribute here 
https://docs.aiohttp.org/en/stable/contributing.html

Specifically, in this PR we would need at least [changelog update](https://docs.aiohttp.org/en/stable/contributing.html#changelog-update) and my guess is that a small test would be nice to have too.

quthla:
The section reads

`The CHANGES.rst file is managed using towncrier tool and all non trivial changes must be accompanied by a news entry.`

I do actually consider this a very trivial one line change so I guess a news entry is not needed here?

webknjaz:
> I do actually consider this a very trivial one line change so I guess a news entry is not needed here?

Change notes are there to be included in the changelog which will be consumed by the end-users. If the change affects anything user-facing, it should be documented even if the code change was one byte. When a user upgrades to the version containing this patch, what will change for them? Will it make their app behave differently? If so, they should be able to figure out what happened.

quthla:
Thanks, added changelog

quthla:
So this can be merged now?

asvetlov:
Thanks!

asvetlov:
Sorry, the PR breaks tests and should be reverted. 47ab6dfd

quthla:
Which test?

https://github.com/aio-libs/aiohttp/actions/runs/1377574239

Issue #4247, opened 2019-10-21
aiohttp doesn't comply autobahn test suite


asvetlov:
Thanks @Nothing4You for raising the question and @njsmith for the fix suggestion.
```
are autobahn tests not used anymore?

Richard Schwab @Nothing4You Oct 20 21:30
also, is ws_connect only supported using a ClientSession?

Andrew Svetlov @asvetlov 01:26
autobahn workflow required manual steps. If you want to integrate them into aiohttp test suite -- you are welcome!
is ws_connect only supported using a ClientSession

Sorry, I don't follow. What did you expect? ClientSession is the root for client-side API


Richard Schwab @Nothing4You 01:52
forgot that a session is also required for other stuff
thought it was just for convenience
i just noticed that the autobahn tests have code that doesn't work anymore

Nathaniel J. Smith @njsmith 06:19
running the autobahn tests is kind of annoying, but wsproto has an automated harness that you might be able to steal parts of: https://github.com/python-hyper/wsproto/tree/master/compliance
```

At least we should run autobahn manually and fix it; ideally automate it.

webknjaz:
> ```
> running the autobahn tests is kind of annoying, but wsproto has an automated harness that you might be able to steal parts of: https://github.com/python-hyper/wsproto/tree/master/compliance
> ```

@anesabml I think this could be a starting point for automating the test runs in our CIs :point_up:

anesabml:
@webknjaz it's a great example, and it got me thinking if we could just use [docker-py](https://github.com/docker/docker-py) / [python-on-whales](https://pypi.org/project/python-on-whales/) instead of [pytest-docker-compose](https://pypi.org/project/pytest-docker-compose/) to start the docker containers.
(I am sorry if this is taking too long but I just want to make sure that I am doing the right thing)

webknjaz:
Yeah, maybe. Honestly, I'd be happy to avoid a vendor lock and make sure that it also works with a rootless podman too. But if the libs don't support it, hitting up docker directly is fine by me.

Dreamsorcerer:
What's still needed to get these tests passing? It's rather a shame that >10% of our CI time is spent running xfail tests. Maybe it's better to disable these tests for now? Or, only run them in specific scenarios (like push to master, so it doesn't run on PRs)?

adi928:
I started tracking down what is needed to fix the tests. Apparently, the tests were never failing. Working with a docker setup, the tests do pass. 
[...]

PR #5809, opened 2021-06-18
Integrate autobahn tests with pytest

anesabml:
## What do these changes do?

Automate running autobahn test suite.

No.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/4247

## Checklist


.github/workflows/ci.yml
```diff
@@ -116,6 +116,10 @@ jobs:
         path: ${{ steps.pip-cache.outputs.dir }}
         restore-keys: |
             pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
+    - name: Install Docker Compose V2
+      if: ${{ matrix.os == 'ubuntu' }}
+      run: |
+        curl -L https://raw.githubusercontent.com/docker/compose-cli/main/scripts/install/install_linux.sh | sh
```
webknjaz: It's discouraged to pipe scripts on the internet right into interpreters. Why didn't you use apt or pip? They are preferable.

CHANGES/4247.feature
```diff
@@ -0,0 +1 @@
+Automated running autobahn test suit by integrating with pytest.
```
webknjaz: "suit" has a different meaning :)
```suggestion
Automated running autobahn test suite by integrating with pytest.
```

webknjaz: Also, it's not a user-facing feature. It doesn't change anything for users, only for devs/maintainers. So it's probably ".misc".

requirements/dev.txt
```diff
@@ -290,3 +290,5 @@ setuptools==51.3.1
     #   blockdiag
     #   gunicorn
     #   sphinx
+
+python-on-whales==0.19.0
```
webknjaz: This file is managed by `pip-tools` with the source in the neighboring `.in` file. Don't edit it manually. It will be overwritten and this change will be lost.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,32 @@
+import os
+import shutil
+
+import pytest
+from python_on_whales import DockerClient
+
+
+@pytest.fixture(scope="session", autouse=True)
+def create_report_directory():
+    path = "reports"
```
webknjaz: This would probably have to be gitignored. Maybe it's best to have a `.test-results/` dir in the repo with everything gitignored as a designated location for various reports.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,32 @@
...
+    if os.path.isdir(path):
```
webknjaz: It's best to use `pathlib` APIs, not low-level `os.path`.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,32 @@
...
+        shutil.rmtree(path)
+    os.mkdir(path)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def build_aiohttp_docker_image():
+    docker = DockerClient(compose_files=["tests/autobahn/docker-compose.yml"])
+    yield docker.compose.build()
+    docker.compose.rm()
```
webknjaz: You don't use the result of this fixture so don't yield it.
```suggestion
    docker.compose.build()
    try:
        yield
    finally:
        docker.compose.rm()
```

.github/workflows/ci.yml
anesabml: it's the only way that is mentioned in the [official docs](https://docs.docker.com/compose/cli-command/#install-on-linux).

tests/autobahn/test_autobahn.py
anesabml: The `report` directory is added to .gitignore in the tests/autobahn directory, should I remove it and create a `test-results` instead?

.github/workflows/ci.yml
anesabml: I just found out that the script specified in the documentation doesn't actually install docker compose v2 according to this [issue](https://github.com/docker/compose-cli/issues/1820#issuecomment-867659395).

webknjaz: Sadly, the official docs don't always advise secure solutions.

CHANGES/4247.feature
webknjaz: Make it `4247.1.misc` then. Towncrier understands multiple fragments per issue with an extra number like that.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,38 @@
+import shutil
+from pathlib import Path
+
+import pytest
+from python_on_whales import DockerClient
+
+
+@pytest.fixture(scope="session", autouse=True)
+def create_report_directory():
```
webknjaz: Instead of doing this in-tree, which may be problematic in parallel envs, consider using `tmp_path` fixture (pytest builtin) that provides you with an empty dir dedicated to your current test run.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,38 @@
...
+    try:
+        yield
+    finally:
+        # default is the name of the default builder
+        docker.buildx.remove("default")
+
+
+def test_client():
+    docker = DockerClient(compose_files=["tests/autobahn/client/docker-compose.yml"])
+    docker.compose.up(abort_on_container_exit=True)
```
webknjaz: I wonder if you could inject some assertion checking for the presence of the logs, for example, or maybe even analyzing their contents for errors.

tests/autobahn/test_autobahn.py
anesabml: I didn't know about such a thing, thank you for the suggestion I will look into it.

tests/autobahn/test_autobahn.py
anesabml: I think we could extract some information from the generated reports and make assertions on that.

tests/autobahn/test_autobahn.py
anesabml: I couldn't figure out how to bind the created directory using `tmp_path` to the `crossbario/autobahn-testsuite` report volume.

.github/workflows/ci.yml
```diff
@@ -116,6 +116,10 @@ jobs:
         path: ${{ steps.pip-cache.outputs.dir }}
         restore-keys: |
             pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
+    - name: Install Docker Compose
+      if: ${{ matrix.os == 'macos' }}
```
webknjaz:
```suggestion
      if: runner.os == 'macOS'
```

.github/workflows/ci.yml
```diff
@@ -116,6 +116,10 @@ jobs:
         path: ${{ steps.pip-cache.outputs.dir }}
         restore-keys: |
             pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
+    - name: Install Docker Compose
```
webknjaz: The install command just says "docker", not compose

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    subprocess.run(
+        ["docker-compose", "-f", "tests/autobahn/client/docker-compose.yml", "down"]
+    )
+
+    failed_messages = get_failed_tests("tests/autobahn/reports/clients", "aiohttp")
+
+    if failed_messages:
+        pytest.xfail(
```
webknjaz: why not "fail"?

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,117 @@
...
+        ]
+    )
+
+    subprocess.run(
+        ["docker-compose", "-f", "tests/autobahn/client/docker-compose.yml", "down"]
+    )
+
+    failed_messages = get_failed_tests("tests/autobahn/reports/clients", "aiohttp")
+
+    if failed_messages:
```
webknjaz:
```suggestion
    if not failed_messages:
        return
```

tests/autobahn/test_autobahn.py
anesabml: because it causes CI to fail.

anesabml:
@webknjaz Here's a quick update:
1. I was trying to get docker running on ci and manually install Compose V2 or use docker-compose, but I couldn't figure it out, due to the following reasons:
    a.  There's no official way to install Compose V2, you have to manually download the plugin and copy it to `~/.docker/cli-plugins/docker-compose`.
    b. GitHub actions macOS runner doesn't have docker installed, I tried installing manually but it's a bit complicated.
2. I am now trying to figure out how to run autobahn tests using the built-in command tool `wstest`, I have managed to do so and the tests work fine on my laptop but failing on CI due to missing python2 headers because autobahn test suites require python2.

tests/autobahn/test_autobahn.py
webknjaz: If testing fails, the CI must fail as well. But if you want to mark that the test is currently non-passing and it's expected, better use an xfail decorator instead but keep the failure here.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,163 @@
...
+                [
+                    f"case: {msg['case']}"
+                    f"\ndescription: {msg['description']}"
+                    f"\nexpectation: {msg['expectation']}"
+                    f"\nexpected: {msg['expected']}"
+                    f"\nreceived: {msg['received']}"
+                    for msg in failed_messages
+                ]
+            )
+        )
```
webknjaz: It's best to use an actual `assert` statement and add some indents for subfields:
```suggestion
    assert not failed_messages, "\n".join(
        "\n\t".join(
            f"{field}: {msg[field]}"
            for field in ("case", "description", "expectation", "expected", "received")
        )
        for msg in failed_messages
    )
```

webknjaz:
@anesabml for now, just run this under Linux jobs (use `skipif` marker in tests).
Also, you don't have to run docker-compose — just using docker could be enough. Why do we need another level of indirection?

tests/autobahn/server/docker-compose.yml
```diff
@@ -14,5 +14,7 @@ services:
     command: ["wstest", "--mode", "fuzzingclient", "--spec", "/config/fuzzingclient.json"]
 
   aiohttp:
```
webknjaz: Does aiohttp itself really need to run in container? Maybe just autobahn would be enough.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,163 @@
...
+def wait_for_server(host: str, port: int) -> None:
+    while True:
+        sock = socket.socket()
+        try:
+            sock.connect((HOST, PORT))
+        except OSError as sock_err:
+            if sock_err.errno == errno.ECONNREFUSED:
+                time.sleep(0.01)
+            else:
+                raise
```
webknjaz: Would this be cleaner?
```suggestion
        except ConnectionRefusedError:
            time.sleep(0.01)
```
https://docs.python.org/3/library/exceptions.html#ConnectionRefusedError

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def get_failed_tests(report_path: str, name: str) -> List[Dict[str, Any]]:
+    with open(Path(f"{report_path}/index.json")) as f:
+        result_summary = json.load(f)[name]
+    failed_messages = []
+    PASS = {"OK", "INFORMATIONAL"}
+    for results in result_summary.values():
+        if results["behavior"] not in PASS or results["behaviorClose"] not in PASS:
+            with open(Path(f"{report_path}/{results['reportfile']}")) as f:
```
webknjaz: When working with pathlib, don't concatenate strings with an os-specific separator (it's different under windows, for example.

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,163 @@
...
+                report = json.load(f)
+                failed_messages.append(
+                    {
+                        "case": report["case"],
+                        "description": report["description"],
+                        "expectation": report["expectation"],
+                        "expected": report["expected"],
+                        "received": report["received"],
+                    }
+                )
```
webknjaz:
```suggestion
    report_path = Path(report_path)
    result_summary = json.loads((report_path / "index.json").read_text())[name]

    failed_messages = []
    PASS = {"OK", "INFORMATIONAL"}
    entry_fields = {"case", "description", "expectation", "expected", "received"}

    for results in result_summary.values():
        if results["behavior"] in PASS and results["behaviorClose"] in PASS:
            continue

        report = json.loads((report_path / results['reportfile']).read_text())
        failed_messages.append({field: report[field] for field in entry_fields})
```

anesabml: That's a lot better 😅.

tests/autobahn/test_autobahn.py
anesabml: Thank you for the tip 🙏.

Dreamsorcerer: The new code is definitely better than using an f-string, but I believe it's actually fine using `/` in strings for pathlib (it was a problem back in the `os.path` days). The docs do it in many examples, and it is automatically converted to the system separator:
```
>>> str(pathlib.PureWindowsPath("a/b"))
'a\\b'
>>> str(pathlib.PurePosixPath("a/b"))
'a/b'
```

Even when constructing it programmatically, it uses `/` in the `repr` version as a universal representation:
```
>>> p = pathlib.PureWindowsPath("a") / "b"
>>> p
PureWindowsPath('a/b')
>>> str(p)
'a\\b'
```

tests/autobahn/server/docker-compose.yml
anesabml: Yes no need to run aiohttp in a container.

anesabml:
@webknjaz is there anything I should add to this PR?

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,131 @@
...
+    else:
+        docker.build(
+            file="tests/autobahn/Dockerfile.autobahn",
+            tags=["autobahn-testsuite"],
+            context_path=".",
+        )
+        try:
+            yield
+        finally:
+            docker.image.remove(x="autobahn-testsuite")
```
webknjaz: Skip interrupts the test so there's no need to indent this block.
```suggestion

    docker.build(
        file="tests/autobahn/Dockerfile.autobahn",
        tags=["autobahn-testsuite"],
        context_path=".",
    )
    try:
        yield
    finally:
        docker.image.remove(x="autobahn-testsuite")
```

tests/autobahn/test_autobahn.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+@pytest.fixture(scope="session")
+def report_dir(tmp_path_factory: TempPathFactory) -> Path:
+    return tmp_path_factory.mktemp("reports")
+
+
+@pytest.fixture(scope="session", autouse=True)
+def build_autobahn_testsuite() -> Generator[None, None, None]:
+    if shutil.which("docker") is None:
```
webknjaz: This check would only check if the docker client executable is present, not if the server side is running. We should really be checking whether it's possible to reach out to the service (which may be stopped on the machine).

Doesn't this python-on-whales have any API to verify this? If not, consider wrapping that `docker.build()` invocation in try-except and skipping on exception it emits when docker is stopped.

anesabml: python-on-whales will install docker if it's not installed, however, if it's not running it will throw a `DockerException`

anesabml:
@webknjaz is there anything I should add to this PR to close it?

webknjaz:
Sorry I haven't had time to review this sooner. I suppose it's good enough, no need to iterate on improvements forever. Those could be added separately when needed.

webknjaz:
@anesabml mind backporting?

anesabml:
@webknjaz sure, I will backport it

webknjaz:
Wonderful, thanks!

anesabml:
Here's the PR #6074

PR #5812, opened 2021-06-19
Removed duplicated hook ids from .pre-commit-config.yaml

cansarigol:
Are those duplicated lines put delibaretely? I've created this PR to learn the purpose.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,10 +6,6 @@ repos:
     language: system
     entry: ./tools/check_changes.py
     pass_filenames: false
-- repo: https://github.com/pre-commit/pre-commit-hooks
-  rev: 'v3.3.0'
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -6,10 +6,6 @@ repos:
     language: system
     entry: ./tools/check_changes.py
     pass_filenames: false
-- repo: https://github.com/pre-commit/pre-commit-hooks
```
webknjaz, line 9: I believe this one may be deliberate. The reason is that it's best to run fast checks before the heavy ones and if one repo has both heavy and light checks, it's okay to split them.

webknjaz: The rest are probably accidental.

asvetlov:
Superseded by #6123

PR #5828, opened 2021-06-22
Remove incorrect default from docs

### LINKED ISSUE #5827: `StreamResponse.set_cookie` `version` attribute wrong documentation
### Describe the bug

The documentation for [`StreamResponse.set_cookie`](https://docs.aiohttp.org/en/stable/web_reference.html?highlight=set_cookie#aiohttp.web.StreamResponse.set_cookie) shows that the default value for `version` is `None` in the signature (`version=None`) but that it is `1` under the bullet point ("version=1 by default"). It appears that the actual default is `None`. The bullet point should be fixed to reflect this.

Dreamsorcerer:
Fixes #5827

### DIFF
--- a/CHANGES/5727.bugfix
+++ b/CHANGES/5727.bugfix
@@ -1,0 +1,1 @@
+ Remove incorrect default from docs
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
Not sure the optional/default actually needs to be said, given that it can be seen in the function signature. But, that's a separate task if someone wants to clean that up.

AustinScola:
Any reason not to merge this?

PR #5848, opened 2021-07-01
Keep auth header during http->https redirect

### LINKED ISSUE #5783: [ClientSession behavior] Authorization header dropped during HTTP redirect
🐞 **Describe the behavior**

When using `aiohttp.ClientSession` to make a GET request that carries an `Authorization` header, the header is silently dropped whenever the remote server responds with HTTP 3xx redirections to the initial request.

📋 **Versions**

`python 3.8.1 aiohttp 3.7.4.post0 multidict 4.7.6 yarl 1.5.1`

💡 **To Reproduce**

Script to reproduce the behavior using the Discord HTTP API and a non-sensitive OAuth2 access token (note that the access tokens expires on Jun 14 2021, although the behavior should still be observable through `Response.request_info.headers`):

```py
import asyncio, logging, sys
import aiohttp, multidict, yarl

# Non-sensitive, expires on Jun 14 2021
OAUTH2_ACCESS_TOKEN = 'UHwmFVVCnXiywoXp4WMAauaPLkIDzx'
log = logging.getLogger('debug')

async def make_request(url: str):
        async with session.get(url, headers={
            'Authorization': f'Bearer {OAUTH2_ACCESS_TOKEN}',  # Set Authorization header
        }) as res:
            log.info(f'Requested URL: {url}')
            log.info(f'Real URL: {res.real_url}')
            log.info(f'Request headers: { {**res.request_info.headers} }')
            log.info(f'Status: {res.status}')

async def main(no_redirect: str, redirected: str):
    logging.basicConfig(
        level=logging.DEBUG, format='%(levelname)-8s [%(name)s] %(message)s',
    )
    log.info(f'aiohttp {aiohttp.__version__}')
    log.info(f'multidict {multidict.__version__}')
    log.info(f'yarl {yarl.__version__}')
    await make_request(no_redirect)
    await asyncio.sleep(1)
    await make_request(redirected)

if __name__ == '__main__':
    asyncio.run(main(sys.argv[1], sys.argv[2]))
```

Save as `redirect.py` and run

```
python3 redirect.py https://discord.com/api/users/@me http://discord.com/api/users/@me
```

Output:

```
INFO     [debug] aiohttp 3.7.4.post0
INFO     [debug] multidict 4.7.6
INFO     [debug] yarl 1.5.1
INFO     [debug] Requested URL: https://discord.com/api/users/@me
INFO     [debug] Real URL: https://discord.com/api/users/@me
INFO     [debug] Request headers: {'Host': 'discord.com', 'Authorization': 'Bearer UHwmFVVCnXiywoXp4WMAauaPLkIDzx', 'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'User-Agent': 'Python/3.8 aiohttp/3.7.4.post0'}
INFO     [debug] Status: 200
INFO     [debug] Requested URL: http://discord.com/api/users/@me
INFO     [debug] Real URL: https://discord.com/api/users/@me
INFO     [debug] Request headers: {'Host': 'discord.com', 'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'User-Agent': 'Python/3.8 aiohttp/3.7.4.post0'}
INFO     [debug] Status: 401
```

💡 **Expected behavior**

Provided that the authorization info is valid, I expect that the remote server not return an HTTP 401 Unauthorized response due to `aiohttp` dropping the auth header.

📋 **Additional context**

This was raised a year ago in #4568.

These are the specific lines causing this behavior for `3.7.4.post0`: https://github.com/aio-libs/aiohttp/blob/184274d9b28bbfa06ac60e48bf286a761c6a6cb0/aiohttp/client.py#L610-L612

For the test script, the first URL is HTTPS, for which Discord returns HTTP 200 directly, but for the second, HTTP version of the API, Discord issues an HTTP 301 first to redirect the client to the HTTPS version. `aiohttp` then proceeds to drop the `Authorization` header because it is being redirected to a different origin (from http://discord.com to http**s**://discord.com).

This seems to be done for security considerations, to prevent sending the authorization info to a different location. Notably this is also the behavior for the `requests` library (see [this SO](https://stackoverflow.com/a/60451840/10896407) and this [PR on `requests`](https://github.com/psf/requests/pull/5375)).

However, for `aiohttp`, this is undocumented and is therefore surprising for developers, especially in this scenario when the remote server would like to simply upgrade from HTTP to HTTPS but stay on the same host. It is also surprising because it is different than the behaviors seen on major browsers (even though the goal of `aiohttp` is not to emulate browser behaviors). For browsers, the Fetch API spec decides that the same request info including non-forbidden headers should be reused for all redirections (see [issue 553 at whatwg/fetch](https://github.com/whatwg/fetch/issues/553)).

---

For `aiohttp`, my opinion is that it seems less surprising to at least keep the header when the redirect is to the same host/authority as in the previous request, and is done to upgrade from HTTP to HTTPS (`http://discord.com` to `https://discord.com`), instead of dropping it whenever the origin changes (emulating `curl`, see below):

```py
if (
    url.authority() != parsed_url.authority()
    or url.scheme == 'https' and parsed_url.scheme == 'http'  # downgrade
):
    auth = None
    headers.pop(hdrs.AUTHORIZATION, None)
```

and to perhaps provide some warning when it happens, or to document it somewhere.

---

I set up a test server at `https://aiohttp-issue4568.tonywu.org/` that unconditionally redirects any request to `https://discord.com/api/users/@me`. It is setup to respond correctly to CORS preflights.

On Chrome and Firefox, open the console while at `https://discord.com`, then run the following:

```js
await fetch('https://aiohttp-issue4568.tonywu.org/', {mode: 'cors', headers: {'Authorization': 'Bearer UHwmFVVCnXiywoXp4WMAauaPLkIDzx'}})
```

The fetch will succeed without receiving an HTTP 401, meaning the Authorization header survived through redirections.

Safari on the other hand has the same behavior and drops the Authorization header.

`curl` seems to be more specific: it seems to throw out the header when it is redirected to a different host, but keep it when it is upgrading on the same host:

```
$ curl -V
curl 7.77.0 (x86_64-apple-darwin20.4.0) libcurl/7.77.0 ...
```

```
$ curl -Lv 'http://discord.com/api/users/@me' -H 'Authorization: Bearer UHwmFVVCnXiywoXp4WMAauaPLkIDzx'
...
< HTTP/1.1 301 Moved Permanently
...
* Issue another request to this URL: 'https://discord.com/api/users/@me'
...
> GET /api/users/@me HTTP/2
> Host: discord.com
> user-agent: curl/7.77.0
> accept: */*
> authorization: Bearer UHwmFVVCnXiywoXp4WMAauaPLkIDzx
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 256)!
< HTTP/2 200
```

```
$ curl -Lv 'https://aiohttp-issue4568.tonywu.org/' -H 'Authorization: Bearer UHwmFVVCnXiywoXp4WMAauaPLkIDzx'
...
< HTTP/1.1 301 Moved Permanently
...
* Issue another request to this URL: 'https://discord.com/api/users/@me'
...
> GET /api/users/@me HTTP/2
> Host: discord.com
> user-agent: curl/7.77.0
> accept: */*
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 256)!
< HTTP/2 401
```

greshilov:
## What do these changes do?

`http->https` redirect is a common thing today. It's safe to keep `Authorization` header in that situation if the host remains the same.
I'll create a separate PR for the `3.8` branch without this feature, but with the test and doc update (to emphasize, that this feature is `4+` only).

 `Authorization` header wouldn't be dropped during the `http->https` redirect if the host remains the same.

## Related issue number

Resolves #5783

## Checklist

### DIFF
--- a/CHANGES/5783.feature
+++ b/CHANGES/5783.feature
@@ -1,1 +1,1 @@
-Keep ``Authorization`` header during the http->https redirect if the host remains the same.
+Started keeping the ``Authorization`` header during http->https redirects when the host remains the same.
--- a/CHANGES/5783.feature
+++ b/CHANGES/5783.feature
@@ -1,0 +1,1 @@
+Keep ``Authorization`` header during the http->https redirect if the host remains the same.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -572,7 +572,10 @@ class ClientSession:
                             and url.scheme == "http"
                         )
 
-                        if url.origin() != parsed_url.origin() and not https_redirect:
+                        if (
+                            url.origin() != parsed_url.origin()
+                            and not is_same_host_https_redirect
+                        ):
                             auth = None
                             headers.pop(hdrs.AUTHORIZATION, None)
 
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -566,7 +566,7 @@ class ClientSession:
                         elif not scheme:
                             parsed_url = url.join(parsed_url)
 
-                        https_redirect = (
+                        is_same_host_https_redirect = (
                             url.host == parsed_url.host
                             and parsed_url.scheme == "https"
                             and url.scheme == "http"
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -566,7 +566,13 @@ class ClientSession:
                         elif not scheme:
                             parsed_url = url.join(parsed_url)
 
-                        if url.origin() != parsed_url.origin():
+                        https_redirect = (
+                            url.host == parsed_url.host
+                            and parsed_url.scheme == "https"
+                            and url.scheme == "http"
+                        )
+
+                        if url.origin() != parsed_url.origin() and not https_redirect:
                             auth = None
                             headers.pop(hdrs.AUTHORIZATION, None)
 
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -62,6 +62,10 @@ For *text/plain* ::
    to a different host or protocol, except the case when  ``HTTP -> HTTPS``
    redirect is performed on the same host.
 
+.. versionchanged:: 4.0
+
+   Started keeping the ``Authorization`` header during ``HTTP -> HTTPS``
+   redirects when the host remains the same.
 
 Custom Cookies
 --------------
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -56,6 +56,14 @@ For *text/plain* ::
 
     await session.post(url, data='Привет, Мир!')
 
+.. note::
+   ``Authorization`` header will be removed if you get redirected
+   to a different host or protocol.
+
+.. versionchanged:: 4.0
+   ``Authorization`` header will **not** be removed during the
+   ``HTTP -> HTTPS`` redirect if the host remains the same.
+
 Custom Cookies
 --------------
 
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -57,6 +57,7 @@ For *text/plain* ::
     await session.post(url, data='Привет, Мир!')
 
 .. note::
+
    ``Authorization`` header will be removed if you get redirected
    to a different host or protocol, except the case when  ``HTTP -> HTTPS``
    redirect is performed on the same host.
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -58,11 +58,9 @@ For *text/plain* ::
 
 .. note::
    ``Authorization`` header will be removed if you get redirected
-   to a different host or protocol.
+   to a different host or protocol, except the case when  ``HTTP -> HTTPS``
+   redirect is performed on the same host.
 
-.. versionchanged:: 4.0
-   ``Authorization`` header will **not** be removed during the
-   ``HTTP -> HTTPS`` redirect if the host remains the same.
 
 Custom Cookies
 --------------
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2350,9 +2350,7 @@ def create_server_for_url_and_handler(
             )
             ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
             cert.configure_cert(ssl_ctx)
-            kwargs = dict(ssl=ssl_ctx)
-        else:
-            kwargs = {}
+            kwargs['ssl'] = ssl_ctx
         return await aiohttp_server(app, **kwargs)
 
     return create
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2350,7 +2350,7 @@ def create_server_for_url_and_handler(
             )
             ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
             cert.configure_cert(ssl_ctx)
-            kwargs['ssl'] = ssl_ctx
+            kwargs["ssl"] = ssl_ctx
         return await aiohttp_server(app, **kwargs)
 
     return create
@@ -2363,10 +2363,15 @@ def create_server_for_url_and_handler(
             "http://host1.com/path1",
             "http://host2.com/path2",
             True,
-        ],  # entirely different hosts
-        ["http://host1.com/path1", "https://host1.com/path1", False],  # http -> https
-        ["https://host1.com/path1", "http://host1.com/path2", True],  # https -> http
+        ],
+        ["http://host1.com/path1", "https://host1.com/path1", False],
+        ["https://host1.com/path1", "http://host1.com/path2", True],
     ],
+    ids=(
+        "entirely different hosts",
+        "http -> https",
+        "https -> http",
+    ),
 )
 async def test_drop_auth_on_redirect_to_other_host(
     create_server_for_url_and_handler: Any,
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -8,11 +8,13 @@ import io
 import json
 import pathlib
 import socket
+import ssl
 from typing import Any
 from unittest import mock
 
 import pytest
 from multidict import MultiDict
+from yarl import URL
 
 import aiohttp
 from aiohttp import Fingerprint, ServerFingerprintMismatch, hdrs, web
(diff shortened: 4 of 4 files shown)


tests/test_client_functional.py
```diff
@@ -2333,25 +2335,74 @@ async def test_creds_in_auth_and_url() -> None:
         await session.close()
 
 
-async def test_drop_auth_on_redirect_to_other_host(aiohttp_server: Any) -> None:
+@pytest.mark.parametrize(
+    ["url_a", "url_b", "drop_header"],
```
webknjaz: It's better to use meaningful variable names. What are "a" and "b"? `url_from`/`url_to`?

webknjaz: s/drop_header/is_drop_header_expected/

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,74 @@ async def test_creds_in_auth_and_url() -> None:
...
+            cert = tls_certificate_authority.issue_cert(
+                url.host, "localhost", "127.0.0.1"
+            )
+            ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            cert.configure_cert(ssl_ctx)
+            kwargs = dict(ssl=ssl_ctx)
+        else:
+            kwargs = {}
+
+        return await aiohttp_server(app, **kwargs)
```
webknjaz: This should probably be a fixture?

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,74 @@ async def test_creds_in_auth_and_url() -> None:
...
+
+    server_a = await create_server(yarl_a, srv1)
+    server_b = await create_server(yarl_b, srv2)
```
webknjaz: same problem as above with a/b

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,74 @@ async def test_creds_in_auth_and_url() -> None:
...
+
+    assert (
+        yarl_a.host != yarl_b.host or server_a.scheme != server_b.scheme
+    ), "Invalid test case, host or scheme must differ"
+
+    schemes = dict(http=80, https=443)
```
webknjaz: plz use a dict literal here, not a constructor.

webknjaz: Also, it's s/schemes/schemas/ in this case.

webknjaz: I think it's better to reflect what this variable is in its name: `protocol_port_map`

docs/client_advanced.rst
```diff
@@ -56,6 +56,14 @@ For *text/plain* ::
 
     await session.post(url, data='Привет, Мир!')
 
+.. note::
+   ``Authorization`` header will be removed if you get redirected
+   to a different host or protocol.
```
webknjaz: I think this should mention the redirect case, not just in the versionchanged directive.

aiohttp/client.py
```diff
@@ -566,7 +566,13 @@ async def _request(
                         elif not scheme:
                             parsed_url = url.join(parsed_url)
 
-                        if url.origin() != parsed_url.origin():
+                        https_redirect = (
```
webknjaz: It's probably better to be verbose
```suggestion
                        is_same_host_https_redirect = (
```

CHANGES/5783.feature
```diff
@@ -0,0 +1 @@
+Keep ``Authorization`` header during the http->https redirect if the host remains the same.
```
webknjaz:
```suggestion
Started keeping the ``Authorization`` header during http->https redirects when the host remains the same.
```

tests/test_client_functional.py
greshilov: Fair enough! Done.

tests/test_client_functional.py
greshilov: Done.

tests/test_client_functional.py
greshilov: Fixed!

tests/test_client_functional.py
greshilov: Done.

By the way, why dict literal is preferable?

docs/client_advanced.rst
greshilov: Ok, moved to the note.

tests/test_client_functional.py
webknjaz: It's slightly faster but also it's because of the consistent style considerations. Do you do `list()` everywhere instead of `[]`? How about `tuple()` vs `()`? The only case I'd use `dict` is when I need to pass callable as an argument.

docs/client_advanced.rst
webknjaz: I'd have a copy, I think. Not a move. Or not a copy but rather some mention about when things changed.

docs/client_advanced.rst
```diff
@@ -56,6 +56,12 @@ For *text/plain* ::
 
     await session.post(url, data='Привет, Мир!')
 
+.. note::
+   ``Authorization`` header will be removed if you get redirected
+   to a different host or protocol, except the case when  ``HTTP -> HTTPS``
+   redirect is performed on the same host.
```
webknjaz, line 63: Plz keep that versionchanged with a very small note.

docs/client_advanced.rst
```diff
@@ -56,6 +56,12 @@ For *text/plain* ::
 
     await session.post(url, data='Привет, Мир!')
 
+.. note::
+   ``Authorization`` header will be removed if you get redirected
```
webknjaz, line 61:
```suggestion

   ``Authorization`` header will be removed if you get redirected
```

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,82 @@ async def test_creds_in_auth_and_url() -> None:
...
-async def test_drop_auth_on_redirect_to_other_host(aiohttp_server: Any) -> None:
+@pytest.fixture
+def create_server_for_url_and_handler(
+    aiohttp_server: Any, tls_certificate_authority: Any
+):
+    async def create(url: URL, srv: Any):
+        app = web.Application()
+        app.router.add_route("GET", url.path, srv)
+
+        if url.scheme == "https":
```
webknjaz, line 2347:
```suggestion
        kwargs = {}
        if url.scheme == "https":
```

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,82 @@ async def test_creds_in_auth_and_url() -> None:
...
+            cert = tls_certificate_authority.issue_cert(
+                url.host, "localhost", "127.0.0.1"
+            )
+            ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            cert.configure_cert(ssl_ctx)
+            kwargs = dict(ssl=ssl_ctx)
+        else:
+            kwargs = {}
```
webknjaz:
```suggestion
            kwargs['ssl'] = ssl_ctx
```

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,82 @@ async def test_creds_in_auth_and_url() -> None:
...
+    [
+        [
+            "http://host1.com/path1",
+            "http://host2.com/path2",
+            True,
+        ],  # entirely different hosts
+        ["http://host1.com/path1", "https://host1.com/path1", False],  # http -> https
+        ["https://host1.com/path1", "http://host1.com/path2", True],  # https -> http
+    ],
+)
```
webknjaz, line 2375:
```suggestion
    ids=(
        'entirely different hosts',
        'http -> https',
        'https -> http',
    ),
)
```

docs/client_advanced.rst
greshilov: I added `versionchanged` with text from `CHANGES/5783.feature`

tests/test_client_functional.py
greshilov: Done.

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,86 @@ async def test_creds_in_auth_and_url() -> None:
...
+
+        kwargs = {}
+        if url.scheme == "https":
+            cert = tls_certificate_authority.issue_cert(
+                url.host, "localhost", "127.0.0.1"
+            )
+            ssl_ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            cert.configure_cert(ssl_ctx)
+            kwargs["ssl"] = ssl_ctx
+        return await aiohttp_server(app, **kwargs)
```
webknjaz: nit: you could drop `await` here and `async` at the top and the interface would remain the same.

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,86 @@ async def test_creds_in_auth_and_url() -> None:
...
+    url_from: str,
+    url_to: str,
+    is_drop_header_expected: bool,
+) -> None:
+    yarl_a = URL(url_from)
+    yarl_b = URL(url_to)
+
     async def srv1(request):
-        assert request.host == "host1.com"
+        assert request.host == yarl_a.host
```
webknjaz: same problem with naming: yarl_a? b? c? it needs to have words referring to things it represents.

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,86 @@ async def test_creds_in_auth_and_url() -> None:
...
+        "https -> http",
+    ),
+)
+async def test_drop_auth_on_redirect_to_other_host(
+    create_server_for_url_and_handler: Any,
+    url_from: str,
+    url_to: str,
+    is_drop_header_expected: bool,
+) -> None:
+    yarl_a = URL(url_from)
```
webknjaz: Why did you use a lib name to describe a URL? why not just a URL? You could even just reuse the same variable name, right? Maybe you could even just wrap params in the decorator with `yarl.URL()` and keep it out of the test body to reduce the cognitive complexity of this test.

tests/test_client_functional.py
```diff
@@ -2333,25 +2335,86 @@ async def test_creds_in_auth_and_url() -> None:
...
+    yarl_b = URL(url_to)
+
     async def srv1(request):
-        assert request.host == "host1.com"
+        assert request.host == yarl_a.host
         assert request.headers["Authorization"] == "Basic dXNlcjpwYXNz"
-        raise web.HTTPFound("http://host2.com/path2")
+        raise web.HTTPFound(url_to)
```
webknjaz, line 2387: Does this really need a string? I bet it could just accept that `yarl.URL` object you create anyway.

greshilov: Fixed.

tests/test_client_functional.py
greshilov: Yep, you're right.

tests/test_client_functional.py
greshilov: I've renamed all redirect related variables using `from,to` convention!

tests/test_client_functional.py
greshilov: Indeed, thanks!

greshilov:
@webknjaz PTAL 
It feels like PR is almost there.

webknjaz (approved):

PR #5853, opened 2021-07-03
Don't cancel tasks when entering a timer context

Dreamsorcerer:
Fixes #5851

Issue #5851, opened 2021-07-03
unexpected asyncio.exceptions.CancelledError


vitidev:
### Describe the bug

Sometimes after raising `asyncio.exceptions.TimeoutError` next awaits of coroutines causes `asyncio.exceptions.CancelledError`

### To Reproduce

Emulate server with random timeouts

```python
import asyncio
import datetime
import random
from aiohttp import web

async def uptime_handler(request):
    chance = random.randint(1, 4)
    if chance == 0:
        await asyncio.sleep(50)
        return
    response = web.StreamResponse()
    response.headers['Content-Type'] = 'text/html'
    await response.prepare(request)

    for i in range(100000):
        formatted_date = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
        message = f'{formatted_date}<br>'
        await response.write(message.encode('utf-8'))
        await asyncio.sleep(0.01)

if __name__ == '__main__':
    app = web.Application()
    app.add_routes([ web.get('/', uptime_handler),])
    web.run_app(app)
```

and a client that "downloads files" in parallel

```python
import asyncio
import random
import traceback

import aiohttp

URL = 'http://127.0.0.1:8080/'

__http__: aiohttp.ClientSession = None
__semaphore__: asyncio.Semaphore = None

async def download_file(client, url: str):
    try:
        r = await client.get(url, timeout=random.randint(5, 15))
        async for data in r.content.iter_chunked(8192):
            await asyncio.sleep(0.01)
        stub = 2 + 2
        foo = "bar"  # garbage code
    except asyncio.exceptions.TimeoutError:
        try:
            print(f'catch TIMEOUT in download_file')
            await asyncio.sleep(5)
        except asyncio.exceptions.CancelledError:
            print('ALARM ' + traceback.format_exc())

async def _download(client, idx) -> int:
    try:
        for i in range(10):
            print(f'{idx} start')
            try:
                await download_file(client, URL)
            except aiohttp.client.ClientPayloadError:
                break
    except BaseException as e:
        print('**********' + traceback.format_exc())
    finally:
        print(f'{idx} finish')
        __semaphore__.release()
    return idx

async def main():
    global __semaphore__
    __semaphore__ = asyncio.BoundedSemaphore(10)
    async with aiohttp.ClientSession()as client:
        i = 0
        while True:
            i += 1
            await __semaphore__.acquire()
            asyncio.ensure_future(_download(client, i % 10))

if __name__ == '__main__':
[...]

PR #5854, opened 2021-07-04
Allow timeout to work when reading with nowait

Dreamsorcerer:
(Note this depends on and extends #5853)

When reading in a loop while the buffer is being constantly filled, the timeout does not work as there are no calls to `_wait()` where the timer is used.

I don't know if this edge case is enough to be worried about, but have put together an initial attempt at fixing it.
I'm not sure if this is really the right solution, but can atleast be used as as a discussion on ways to improve this.

This can't be backported as this changes the public API (one of the functions is now async).

Related #5851.


asvetlov:
`def read_nowait()` -> `async def read_nowait()` signature change is not backward compatible.

Honestly, I'm not sure if nowait read should handle timeouts; it is a nonblocking call by definition.

webknjaz:
This won't make it into the 3.8 release stream. Turning it into a draft.

Dreamsorcerer:
OK, after looking at it some more, I realised the timer I was using to fix this was just raising TimeoutError in the `__enter__`. It's unlikely that the inner loop will ever take long enough to block the event loop too long. So, I've just copied that raise logic to a new method and called it directly. There's no need to make anything async now.

aiohttp/streams.py
```diff
@@ -122,7 +122,7 @@ def __init__(
         self._waiter: Optional[asyncio.Future[None]] = None
         self._eof_waiter: Optional[asyncio.Future[None]] = None
         self._exception: Optional[BaseException] = None
-        self._timer = timer
+        self._timer = timer if timer else TimerNoop()
```
Dreamsorcerer: Updated this for simplicity and consistency, as mypy required me to update some related parts anyway.

aiohttp/helpers.py
```diff
@@ -676,7 +676,8 @@ def __call__(self) -> None:
 
 
 class BaseTimerContext(ContextManager["BaseTimerContext"]):
-    pass
+    def check_timeout(self) -> None:
```
webknjaz: "check" implies that it'll return a boolean. Let's be explicit about the side effects:
```suggestion
    def assert_timeout_exceeded(self) -> None:
```

aiohttp/helpers.py
```diff
@@ -700,6 +701,11 @@ def __init__(self, loop: asyncio.AbstractEventLoop) -> None:
         self._tasks: List[asyncio.Task[Any]] = []
         self._cancelled = False
 
+    def check_timeout(self) -> None:
+        """Raise TimeoutError if timer has already been cancelled."""
+        if self._cancelled:
+            raise asyncio.TimeoutError from None
```
webknjaz, line 707: Let's add a reason string to the exception, maybe?

aiohttp/streams.py
webknjaz: Plz make it `is not None` for accuracy, since our sentinel is `None` and not an arbitrary falsy value..

tests/test_client_functional.py
```diff
@@ -3022,6 +3022,30 @@ async def handler(request):
...
+        resp = web.StreamResponse()
+        await resp.prepare(request)
+        while True:
+            await resp.write(b"1" * 1000)
+            await asyncio.sleep(0.01)
+
+    async def request(client):
+        timeout = aiohttp.ClientTimeout(total=0.5)
+        with pytest.raises(asyncio.TimeoutError):
+            async with await client.get("/", timeout=timeout) as resp:
```
webknjaz: Swap these two lines to scope the expectation to a narrow code block/line.

tests/test_client_functional.py
```diff
@@ -3022,6 +3022,30 @@ async def handler(request):
...
+                async for data in resp.content.iter_chunked(1):
+                    await asyncio.sleep(0.01)
+
+    app = web.Application()
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+    # wait_for() timeout should not be reached.
```
webknjaz: If it shouldn't be reached, use `raise RuntimeError` or `pytest.fail()` instead..

aiohttp/helpers.py
Dreamsorcerer: Though you might come up with a better name.

aiohttp/helpers.py
Dreamsorcerer: Can do, but the other one doesn't, and what is the reason other than the timer reached the timeout?

tests/test_client_functional.py
Dreamsorcerer: Sorry, I meant the timeout shouldn't be reached. I used wait_for() just to ensure that a failing test doesn't hang forever.

PR #5860, opened 2021-07-06
Fix the `ClientWebSocketResponse.receive` stuck for a long time if network is suddenly unavaiable.

### LINKED ISSUE #2309: WebSocketResponse does not throw/close when connection is abruptly cut

Hanaasagi:
## What do these changes do?

This PR fix the websocket client will get stuck when a the internect is unavaiable.

## How to reproduce it

Run following code, and turn off the internet after seeing the `Connected` text in the console.

```Python
from asyncio import run

async def main():
        async with session.ws_connect('https://echo.websocket.org/',
                                      heartbeat=10) as ws:
            print('Connected', ws)
            print('Received', await ws.receive())

run(main())
```

In my local machine, it will block ablout 8 min and get a `CLOSED` message.

## Explain Why 

`receive` method has a `timeout` argument, and the default value is `None`. It means no timeout for reading from a connection. The websocket pong frame and user data frame uses the same timeout. So if `aiohttp` send a ping frame successfully, then we turn off the network. The `receive` coroutine will be block at the `msg = await self._reader.read()`. Because there are no data comes from the server.

https://github.com/aio-libs/aiohttp/blob/184274d9b28bbfa06ac60e48bf286a761c6a6cb0/aiohttp/client_ws.py#L213-L238

Even if we call `_pong_not_received`, it doesn't cancel the `receive` coroutine.

https://github.com/aio-libs/aiohttp/blob/184274d9b28bbfa06ac60e48bf286a761c6a6cb0/aiohttp/client_ws.py#L101-L106

The `receive` coroutine finished until a `EofStream` error from the TCP Stack.

## What did this PR do 

Add a simple flag(`asyncio.Event`) `_is_heartbeat_failed` in `ClientWebSocketResponse`. And when call `ClientWebSocketResponse.receive`, it both await `self._reader.read()` and this flag by `asyntio.wait` . So if the heartbeat is failed, the `ClientWebSocketResponse.receive` will be awake.

## Related issue number

Fix https://github.com/aio-libs/aiohttp/issues/2309

## Checklist

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -78,6 +78,8 @@ class ClientWebSocketResponse:
         self._compress = compress
         self._client_notakeover = client_notakeover
 
+        # A flag to indicate whether the latest heartbeat failed.
+        self._is_heartbeat_failed = asyncio.Event()
         self._reset_heartbeat()
 
     def _cancel_heartbeat(self) -> None:
@@ -91,6 +93,7 @@ class ClientWebSocketResponse:
 
     def _reset_heartbeat(self) -> None:
         self._cancel_heartbeat()
+        self._is_heartbeat_failed.clear()
 
         if self._heartbeat is not None:
             self._heartbeat_cb = call_later(
@@ -116,6 +119,7 @@ class ClientWebSocketResponse:
             self._close_code = WSCloseCode.ABNORMAL_CLOSURE
             self._exception = asyncio.TimeoutError()
             self._response.close()
+            self._is_heartbeat_failed.set()
 
     @property
     def closed(self) -> bool:
@@ -235,11 +239,32 @@ class ClientWebSocketResponse:
 
             try:
                 self._waiting = self._loop.create_future()
+
+                read_task: asyncio.Task[WSMessage] = asyncio.create_task(
+                    self._reader.read()
+                )
+                is_heartbeat_failed_task: asyncio.Task[bool] = asyncio.create_task(
+                    self._is_heartbeat_failed.wait()
+                )
                 try:
                     async with async_timeout.timeout(
                         timeout or self._timeout.ws_receive
                     ):
-                        msg = await self._reader.read()
+                        # Check the heartbeat status when waiting data from server
+                        done, pending = await asyncio.wait(
+                            (read_task, is_heartbeat_failed_task),
+                            return_when=asyncio.FIRST_COMPLETED,
+                        )
+                        # If server doesn't pong, but return data normally,
+                        # supress the exception.
+                        if read_task in done:
+                            is_heartbeat_failed_task.cancel()
+                            msg = read_task.result()
+                        elif is_heartbeat_failed_task in done:
+                            read_task.cancel()
+                            assert self._exception is not None
+                            raise self._exception
+
                     self._reset_heartbeat()
                 finally:
                     waiter = self._waiting
--- a/CHANGES/2309.bugfix
+++ b/CHANGES/2309.bugfix
@@ -1,0 +1,1 @@
+Fix the `ClientWebSocketResponse.receive` stuck for a long time if network is suddenly interrupted.
--- a/tests/test_client_ws_functional.py
+++ b/tests/test_client_ws_functional.py
@@ -1,6 +1,8 @@
 # type: ignore
 import asyncio
+import time
 from typing import Any
+from unittest import mock
 
 import async_timeout
 import pytest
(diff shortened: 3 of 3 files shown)


webknjaz:
Please fill out the PR template with the details on what this is supposed to do.

aiohttp/client_ws.py
```diff
@@ -235,11 +239,32 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
+                        )
+                        # If server doesn't pong, but return data normally,
+                        # supress the exception.
+                        if read_task in done:
+                            is_heartbeat_failed_task.cancel()
+                            msg = read_task.result()
+                        elif is_heartbeat_failed_task in done:
+                            read_task.cancel()
+                            assert self._exception is not None
+                            raise self._exception
```
Hanaasagi, line 266: I'm not sure that raise ` TimeoutError` or just return a `WSMessage(WSMsgType.ERROR, ..)` message which is better.

gil-cohen:
any update?

aiohttp/client_ws.py
```diff
@@ -235,11 +239,32 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
 
             try:
                 self._waiting = self._loop.create_future()
+
+                read_task: asyncio.Task[WSMessage] = asyncio.create_task(
+                    self._reader.read()
+                )
+                is_heartbeat_failed_task: asyncio.Task[bool] = asyncio.create_task(
+                    self._is_heartbeat_failed.wait()
+                )
```
astrand, line 248: Do we really need two additional tasks in order to solve this bug? Is it at all necessary to check heartbeat status, or is it sufficient to check transport.is_closing()?

Dreamsorcerer:
The linked issue was closed, would be good if @bdraco or @Hanaasagi could confirm if this PR is no longer relevant.

bdraco:
The receive resolves immediately when we know the connection is dropped now

bdraco:
Very likely this issue is fixed. Please let us know if you still experience this behavior on 3.10.5 or later and we can reopen.

PR #2075, opened 2017-07-10
Add automatically a SafeChildWatcher to the test loop

cecton:
The fact that on Unix system we don't create automatically a child
watcher makes all call to asyncio.create_subprocess_* to fail.

## What do these changes do?

It adds a default child watcher to the test loop.

## Are there changes in behavior for the user?

Any `asyncio.create_subprocess_*` call in the application or during a test will work because the ChildWatcher exist (on Unix).

No change for Windows.

## Related issue number

#2058 #2059 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new news fragment into the `changes` folder
  * name it `<issue_id>.<type>` for example (588.bug)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #5862, opened 2021-07-07
Use `MultiLoopChildWatcher` in tests where available

sweatybridge:
## What do these changes do?

Python 3.8 adds a new child watcher class that doesn't require running loop in the main thread: https://docs.python.org/3/library/asyncio-policy.html#asyncio.MultiLoopChildWatcher. This PR uses the new watcher if it is available in asyncio module.

It should reduce test flakiness when parallelized with `pytest-xdist`.

No change for end users. Existing tests in `test_loop.py` covers this change given the current text matrix.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/3450
https://github.com/aio-libs/aiohttp/pull/2075

## Checklist

### DIFF
--- a/CHANGES/3450.bugfix
+++ b/CHANGES/3450.bugfix
@@ -1,0 +1,1 @@
+Started using `MultiLoopChildWatcher` when it's available under POSIX while setting up the test I/O loop.
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -497,7 +497,7 @@ def setup_test_loop(
             # * https://stackoverflow.com/a/58614689/595220
             # * https://bugs.python.org/issue35621
             # * https://github.com/python/cpython/pull/14344
-            watcher = asyncio.ThreadedChildWatcher()
+            watcher = asyncio.MultiLoopChildWatcher()
         except AttributeError:  # Python < 3.8
             watcher = asyncio.SafeChildWatcher()
         watcher.attach_loop(loop)
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -490,7 +490,16 @@ def setup_test_loop(
     asyncio.set_event_loop(loop)
     if sys.platform != "win32" and not skip_watcher:
         policy = asyncio.get_event_loop_policy()
-        watcher = asyncio.SafeChildWatcher()
+        watcher: asyncio.AbstractChildWatcher
+        try:  # Python >= 3.8
+            # Refs:
+            # * https://github.com/pytest-dev/pytest-xdist/issues/620
+            # * https://stackoverflow.com/a/58614689/595220
+            # * https://bugs.python.org/issue35621
+            # * https://github.com/python/cpython/pull/14344
+            watcher = asyncio.ThreadedChildWatcher()
+        except AttributeError:  # Python < 3.8
+            watcher = asyncio.SafeChildWatcher()
         watcher.attach_loop(loop)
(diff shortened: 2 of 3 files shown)


webknjaz:
This is already being introduced in https://github.com/aio-libs/aiohttp/pull/5431 but with a threaded watcher. It's also being discussed in https://github.com/aio-libs/aiohttp/pull/5852. Why do you think that multiloop would be a better fit? Do you want to join those discussions instead?

sweatybridge:
ok, will join the other thread. Wasn't sure where to look since there are many related tickets.

webknjaz:
Let's keep this open for now. I think we could accept it after some minor polishing. But I still need an answer to my question above, okay?

I encourage you to participate in other discussions, though. And it's indeed better to merge small PRs. I only mentioned other efforts so that people interested in achieving the common goal could collaborate more effectively and not repeat each other's work twice.

CHANGES/3450.bugfix
```diff
@@ -0,0 +1 @@
+Use MultiLoopChildWatcher when setting up test loop
```
webknjaz:
```suggestion
Started using `MultiLoopChildWatcher` when it's available under POSIX while setting up the test I/O loop.
```

aiohttp/test_utils.py
```diff
@@ -490,7 +490,10 @@ def setup_test_loop(
     asyncio.set_event_loop(loop)
     if sys.platform != "win32" and not skip_watcher:
         policy = asyncio.get_event_loop_policy()
-        watcher = asyncio.SafeChildWatcher()
+        if hasattr(asyncio, "MultiLoopChildWatcher"):
+            watcher = asyncio.MultiLoopChildWatcher()
+        else:
+            watcher = asyncio.SafeChildWatcher()
```
webknjaz: I have a few comments on this block: it is preferable to use the EAFP style in Python and also we need to have comments about the motivation of this.

To solve this, I suggest you cherry-pick my commit [`bea959a` (#5431)](https://github.com/aio-libs/aiohttp/pull/5431/commits/bea959a4ce195405939c40f184c5189206064447) (same patch: https://github.com/aio-libs/aiohttp/commit/bea959a4ce195405939c40f184c5189206064447.patch) and then change `ThreadedChildWatcher` with `MultiLoopChildWatcher` in a follow-up commit. If you know how to use interactive rebase in Git, it should be easy for you :)

Then, we could merge this change and backport it.

After that, I think it would be useful to vendor a backported watcher right in this repo and make use of it under older supported Pythons too. But this is a separate effort that should go into a separate PR.

sweatybridge: Thanks for the review.

I'm reading this ticket https://bugs.python.org/issue38591 and there seems to be many instabilities with `MultiLoopChildWatcher`. I think `ThreadedChildWatcher` might actually be a better choice here...

Regarding the performance concerns of creating a new thread, how about additionally exposing `skip_watcher: bool` as a param? `SafeChildWatcher` was originally introduced to support `asyncio.create_subprocess_*`. In my use case, this support is not needed. If we expose the param, no one pays the cost unnecessarily.

webknjaz: Oh, maybe that's why I chose `ThreadedChildWatcher`. I think I talked to Andrew and he said to use it. It was still problematic under the heavy load on my laptop, IIRC.

As for `skip_watcher`, please file a proposal issue or a draft PR to demonstrate how you see it being exposed, API-wise.

PR #5871, opened 2021-07-11
Replace if statement with if expression

monosans:

No.

## Checklist

- [X] Add myself to `CONTRIBUTORS.txt`


webknjaz (changes requested):
Thanks for the PR but I don't believe this change adds any value to the project.

PR #5877, opened 2021-07-14
Switch to ThreadedChildWatcher and test

sweatybridge:
## What do these changes do?

Switch back to using `ThreadedChildWatcher` (originally introduced in #5862). Adds a unit test to ensure loop can be setup from non-main thread by `pytest-xdist`.

NA

## Related issue number

#5852

## Checklist

### DIFF
--- a/CHANGES/5877.bugfix
+++ b/CHANGES/5877.bugfix
@@ -1,0 +1,1 @@
+Uses `ThreadedChildWatcher` under POSIX to allow setting up test loop in non-main thread.
--- a/CHANGES/5877.bugfix
+++ b/CHANGES/5877.bugfix
@@ -1,1 +1,1 @@
-Uses `ThreadedChildWatcher` under POSIX to allow setting up test loop in non-main thread.
+Uses :py:class:`~asyncio.ThreadedChildWatcher` under POSIX to allow setting up test loop in non-main thread.
--- a/tests/test_loop.py
+++ b/tests/test_loop.py
@@ -13,7 +13,6 @@ from aiohttp.test_utils import AioHTTPTestCase, loop_context
     platform.system() == "Windows", reason="the test is not valid for Windows"
 )
 async def test_subprocess_co(loop: Any) -> None:
-    assert threading.current_thread() is threading.main_thread()
     proc = await asyncio.create_subprocess_shell(
         "exit 0",
         stdin=asyncio.subprocess.DEVNULL,
@@ -47,8 +46,9 @@ def test_default_loop(loop: Any) -> None:
 
 def test_setup_loop_non_main_thread() -> None:
     def target() -> None:
-        with loop_context():
-            pass
+        with loop_context() as loop:
+            assert asyncio.get_event_loop() is loop
+            loop.run_until_complete(test_subprocess_co(loop))
 
     # Ensures setup_test_loop can be called by pytest-xdist in non-main thread.
     t = threading.Thread(target=target)
--- a/tests/test_loop.py
+++ b/tests/test_loop.py
@@ -14,6 +14,7 @@ from aiohttp.test_utils import AioHTTPTestCase, loop_context
     platform.system() == "Windows", reason="the test is not valid for Windows"
 )
 async def test_subprocess_co(loop: Any) -> None:
+    assert PY_38 or threading.current_thread() is threading.main_thread()
     proc = await asyncio.create_subprocess_shell(
         "exit 0",
         stdin=asyncio.subprocess.DEVNULL,
--- a/tests/test_loop.py
+++ b/tests/test_loop.py
@@ -45,7 +45,7 @@ def test_default_loop(loop: Any) -> None:
     assert asyncio.get_event_loop() is loop
 
 
-def test_setup_loop_non_main_thread():
+def test_setup_loop_non_main_thread() -> None:
     # Ensures setup_test_loop can be called by pytest-xdist in non-main thread.
     t = threading.Thread(target=setup_test_loop)
     t.start()
--- a/tests/test_loop.py
+++ b/tests/test_loop.py
@@ -6,6 +6,7 @@ from typing import Any
 import pytest
 
 from aiohttp import web
+from aiohttp.helpers import PY_38
 from aiohttp.test_utils import AioHTTPTestCase, loop_context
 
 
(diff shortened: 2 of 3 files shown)


AustinScola:
Hey @sweatybridge, can you help me understand this change please? My understanding is that `pytest-xdist` uses sub-processes and each of the sub-processes run their own event loop. And `asyncio.ThreadedChildWatcher` handles this scenario better than `asyncio.MultiLoopChildWatcher` does? Is that an accurate description?

sweatybridge:
Hi @AustinScola, `pytest-xdist` uses execnet to run workers in subprocesses. According to this issue https://github.com/pytest-dev/pytest-xdist/issues/469, modern versions of execnet no longer guarantees that every worker is run on the main thread of its own subprocess. As a result, when these tests are scheduled occasionally on non-main thread by `pytest-xdist`, the test will crash due to `MultiLoopChildWatcher` trying to register signal handlers, which requires main thread. Using `ThreadedChildWatcher` avoids this issue because it doesn't register signal handlers.

AustinScola:
@sweatybridge, that makes sense. Thank you for the wonderful explanation!

sweatybridge:
Happy to help! I'm looking forward to xdist support.

AustinScola:
Is this ready to be merged?

sweatybridge:
@webknjaz could you take a look at this and merge it if possible?

CHANGES/5877.bugfix
```diff
@@ -0,0 +1 @@
+Uses `ThreadedChildWatcher` under POSIX to allow setting up test loop in non-main thread.
```
webknjaz: Change notes should use RST syntax, not MD.
```suggestion
Uses ``ThreadedChildWatcher`` under POSIX to allow setting up test loop in non-main thread.
```

webknjaz: Even better, we could link the CPython docs:
```suggestion
Uses :py:class:`~asyncio.ThreadedChildWatcher` under POSIX to allow setting up test loop in non-main thread.
```

tests/test_loop.py
```diff
@@ -43,3 +43,14 @@ def test_default_loop(self) -> None:
 
 def test_default_loop(loop: Any) -> None:
     assert asyncio.get_event_loop() is loop
+
+
+def test_setup_loop_non_main_thread() -> None:
+    def target() -> None:
+        with loop_context():
+            pass
```
webknjaz: Could we run something in this loop and maybe process some signal? This is what may go wrong potentially.

sweatybridge: Setting up the loop context in python 3.7 still uses `SafeChildWatcher`, which in turn registers signals. I've updated the test to expect failure on python 3.7 and success in 3.8+. This way we have both code paths covered.

Subsequently after merging `pytest-xdist`, we might see occasional failures on python 3.7 tests in CI for the same reason. This could happen with any tests that use the asyncio event loop, not specific to this test alone. This will be ok once we deprecate 3.7.

webknjaz: I'd rather vendor a backport and use it under Python 3.7. Also, it's not yet time to drop it anyway. aiohttp 3.8 still supports Python 3.6 even. It'll be EOL in 5 months but for CPython 3.7 there are 2 years to go.
Being a framework/library forces us to support a wider range of versions, unlike projects that are just apps bound to a single env.

tests/test_loop.py
```diff
@@ -6,14 +6,14 @@
...
-from aiohttp.test_utils import AioHTTPTestCase
+from aiohttp.helpers import PY_38
+from aiohttp.test_utils import AioHTTPTestCase, loop_context
 
 
 @pytest.mark.skipif(
     platform.system() == "Windows", reason="the test is not valid for Windows"
 )
 async def test_subprocess_co(loop: Any) -> None:
-    assert threading.current_thread() is threading.main_thread()
```
webknjaz, line 16: Let's keep this check for Python 3.7 and below. We can drop it only when the watcher is vendored.

tests/test_loop.py
```diff
@@ -43,3 +43,24 @@ def test_default_loop(self) -> None:
...
+
+@pytest.mark.xfail(not PY_38, reason="ThreadedChildWatcher is only available in 3.8+")
+def test_setup_loop_non_main_thread() -> None:
+    child_exc = None
+
+    def target() -> None:
+        try:
+            with loop_context() as loop:
+                assert asyncio.get_event_loop() is loop
+                loop.run_until_complete(test_subprocess_co(loop))
```
webknjaz, line 57: I see this calls subprocesses but what about registering signal handlers? This is what actually fails on old watchers.

tests/test_loop.py
```diff
@@ -6,14 +6,14 @@
...
 from aiohttp import web
-from aiohttp.test_utils import AioHTTPTestCase
+from aiohttp.helpers import PY_38
+from aiohttp.test_utils import AioHTTPTestCase, loop_context
 
 
 @pytest.mark.skipif(
     platform.system() == "Windows", reason="the test is not valid for Windows"
 )
 async def test_subprocess_co(loop: Any) -> None:
```
webknjaz, line 16:
```suggestion
async def test_subprocess_co(loop: Any) -> None:
    assert PY_38 or threading.current_thread() is threading.main_thread()
```

tests/test_loop.py
sweatybridge: Registering signal handlers will always fail in non-main thread, regardless of whether we are using old or new watcher. The new watcher avoids this issue by not registering signal handlers at all (using waitpid as alternative).

What we want to ensure in this test is that child processes can be watched, not whether signals can be successfully registered. In other words, since the library no longer registers signal handler on python 3.8, it's not relevant to cover signal registering in tests.

Am I misunderstanding something?

tests/test_loop.py
sweatybridge: Yup, agree that backporting is better.

tests/test_loop.py
webknjaz: Yes. The original problem with xdist was exactly the fact that there's a problem with the signal handlers that happens in an inconsistent manner. This is what blocks https://github.com/aio-libs/aiohttp/pull/5431 and all the previous attempts. I was sure that the new watchers were supposed to fix this based on what @asvetlov mentioned to me privately. Are you sure it's still problematic?

webknjaz (approved):
I'll merge this "as is" for now. It's an improvement over what we had before. But it'd be really great to tackle that problem with the signals in order to make xdist enabled by default.

webknjaz:
@sweatybridge could you backport this manually?

tests/test_loop.py
sweatybridge: What I'm saying is signal handling is an independent problem from watching child process. The old watcher uses signal handling and will crash when mixed with xdist. The new watcher doesn't use signal handling and therefore is not problematic. Our test only need to verify that child process is watched, it doesn't matter whether it's implemented signals or not.

For reference, here's cpython's implementation for `ThreadedChildWatcher` https://github.com/python/cpython/blob/3.8/Lib/asyncio/unix_events.py#L1296. It starts a new thread and calls `waitpid`.

Compare that with `SafeChildWatcher` which inherits from `BaseChildWatcher` https://github.com/python/cpython/blob/3.8/Lib/asyncio/unix_events.py#L927. It tries to register signal handler and hence will raise an exception in non-main thread.

tests/test_loop.py
sweatybridge: here's the backport with manually resolved merge conflicts https://github.com/aio-libs/aiohttp/pull/5919

PR #5894, opened 2021-07-18
 Do not restrict +json media type suffix to application/ 

scop:
## What do these changes do?

Fix +json media type suffix handling to apply to other main types besides application/.

An example of a registered media type triggering this issue is https://www.iana.org/assignments/media-types/model/gltf+json

Media types with main type other than `application` using the '+json` suffix are accepted as JSON now.

## Checklist

### DIFF
--- a/CHANGES/5894.bugfix
+++ b/CHANGES/5894.bugfix
@@ -1,0 +1,1 @@
+Fix JSON media type suffix matching with main types other than application.
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -120,7 +120,7 @@ else:
         return asyncio.iscoroutinefunction(func)
 
 
-json_re = re.compile(r"^application/(?:[\w.+-]+?\+)?json")
+json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json")
 
 
 class BasicAuth(namedtuple("BasicAuth", ["login", "password", "encoding"])):
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -730,6 +730,22 @@ def test_is_expected_content_type_json_match_partially():
     )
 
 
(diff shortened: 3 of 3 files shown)


asvetlov:
Sorry, I think that the PR should be rejected.
There is no guarantee that the content type like `xxx/yyy+json` is fully JSON compatible.

If in your particular case `model/gltf+json` can be processed by `.json()` call -- pass the content type explicitly:
`await resp.json(content_type="model/gltf+json")`.

scop:
Can you cite a reference that says +json might not be JSON compatible? The suffix registration is not restricted to the application/ main type.

https://www.rfc-editor.org/rfc/rfc6839.html#section-3.1 says:
> The suffix "+json" MAY be used with any media type whose representation follows that established for "application/json".

Additionally, from https://www.rfc-editor.org/rfc/rfc6839.html#section-2
>   At the same time, using the suffix allows receivers of the media
>   types to do generic processing of the underlying representation in
>   cases where
>
>      they do not need to perform special handling of the particular
>      semantics of the exact media type, and
>
>      there is no special knowledge needed by such a generic processor
>      in order to parse that underlying representation other than what
>      would be needed to parse any example of that underlying
>      representation.

This is exactly the case for aiohttp.

I think that's guarantee enough, as much as there is guarantee that anything served as application/json is JSON compatible (there is no _guarantee_ about that either, but both that and +json are reasonable expectations).

asvetlov:
Thank you for clarification. 
Let's keep the PR open until I re-read the mentioned RFC

asvetlov (approved):

asvetlov:
I've changed my mind, the PR is correct. 
Thanks!

asvetlov:
Sorry, the backporting is very hard since previous pull requests that introduced and updated `json_re` were not backported too.

PR #5895, opened 2021-07-18
Add test case for too lax application/json prefix matching

scop:
## What do these changes do?

Add a test case demonstrating too lax application/json prefix matching.

## Checklist


asvetlov:
Thanks for the PR.
Sorry, I should reject it, the reason is described here: https://github.com/aio-libs/aiohttp/pull/5894#issuecomment-950289207

scop:
There is some confusion here. #5894 is a completely different thing.

This PR demonstrates a bug where aiohttp treats anything starting with `application/json` as an appropriate type when exoecting `application/json`. The example used here is `application/json-seq` -- that is NOT JSON compatible, and should not be accepted as a JSON type. Similarly, `application/jsonfoofoofoofoofoo` should not be accepted as JSON, but that is what happens:
```
>>> from aiohttp.helpers import is_expected_content_type
>>> is_expected_content_type("application/jsonfoofoofoofoofoo", "application/json")
True
```

IMO this PR should be kept open until the corresponding problem is fixed, and then merged as a test case for the fix.

Issue #4686, opened 2020-04-12
Easy way to get common types for annotations


Dreamsorcerer:
🐣 **Is your feature request related to a problem? Please describe.**
I'm unclear what the best way currently is to make use of type annotations in my own aiohttp projects.

💡 **Describe the solution you'd like**
One of the aiohttp-demos to be updated with full type hinting to serve as an example.

❓ **Describe alternatives you've considered**
The only way I can see at the moment, is with importing protected objects from various modules.

e.g. `from aiohttp.web_app import _Handler`
though this is also in `aiohttp.web_urldispatcher._WebHandler`

Types to add:

- [x] Handler
- [x] Middleware
- [ ] CleanupContext (https://github.com/aio-libs/aiohttp/issues/5883#issuecomment-881895531)

hyzyla:
You can define all types in your code using public classes:
```python3
from aiohttp import web

Handler = Callable[[web.Request], Awaitable[web.StreamResponse]]
```
But, I think, that defining a common combination of types will improve the user experience of working with a library

Dreamsorcerer:
That is also possible, but seems even less maintainable than importing protected variables.

Hardcoded types in my code is more difficult to ensure they are correct and that they get updated if the signatures change in future.

playpauseandstop:
Can add +1 here.

In my aiohttp libraries, such as [aiohttp-middlewares](https://github.com/playpauseandstop/aiohttp-middlewares/blob/a0366c04dabcfa7cb430425e41d71c66860fb994/aiohttp_middlewares/annotations.py#L20) I’m importing `_Handler` and `_Middleware` from `aiohttp.web_middlewares` and redefine it to `Handler` & `Middleware` for later reuse across the library.

While, in reality I would like to have ability to import `aiohttp` type annotations such as,

```python
from aiohttp.web_typing import Handler, Middleware  # and so on
```

Or from `aiohttp.annotations` module, or from `aiohttp.typing` module, doesn’t matter. Main point to have single entry point with all reusable type annotations to use in other aiohttp libraries and projects.

ps. As implementation reference there is a [pyrsistent.typing](https://github.com/tobgu/pyrsistent/blob/v0.16.0/pyrsistent/typing.py) module, which provides type annotations for `pyrsistent` library in very clear way for 3rd party users
pps. I found out, that there is `aiohttp.typedefs` module already, but from quick overview it missed many usable type annotations, such as Handler, Middleware

[...]

PR #5898, opened 2021-07-18
Ascola/add middleware type alias

AustinScola:
## What do these changes do?
Add a middleware type alias and change some code to use it.

Users can now use the `Middleware` type alias in their code.

## Related issue number
#4686

## Checklist

### DIFF
--- a/CHANGES/4686.feature
+++ b/CHANGES/4686.feature
@@ -1,1 +1,1 @@
-Add a request handler type alias ``aiohttp.typedefs.Handler``.
+Add a request handler and a middleware type alias ``aiohttp.typedefs.Handler``.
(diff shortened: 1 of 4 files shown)


tests/test_typedefs.py
```diff
@@ -0,0 +1,12 @@
...
+
+from aiohttp.typedefs import Handler, Middleware
+
+if TYPE_CHECKING:
+    from aiohttp.web import Request, StreamResponse
+
+
+def test_middleware() -> None:
+    """Test aiohttp.typedefs.Middleware."""
+    assert Middleware == Callable[["Request", Handler], Awaitable["StreamResponse"]]
```
Dreamsorcerer: I'm not sure this is really testing anything.

I think a better option would be to add the return types throughout test_web_middlewares.py and then this function can be annotated with the new `Middleware` type as well:
https://github.com/aio-libs/aiohttp/blob/master/tests/test_web_middleware.py#L62

Any problems with the type in future should then be caught by Mypy from this usage.

asvetlov: Agree, type hints don't need unit tests

Issue #5901, opened 2021-07-19
resolver.py/ThreadResolver: address[3] does not exist


dilyanpalauzov:
### Describe the bug

As outlined at https://github.com/pimutils/vdirsyncer/issues/916, I use python 3.8 with aiohttp 3.8 branch (v3.7.4.post0-209-g22f8be99) .  In aiohttp/resolver.py:ThreadedResolver the line
>    if family == socket.AF_INET6 and address[3]:  # type: ignore[misc]

fails, as address has only two elements and address[3] is out of range.

The values of adress are:
> AddressFamily.AF_INET ('142.250.185.238', 443)
> AddressFamily.AF_INET6 (10, b'\x01\xbb\x00\x00\x00\x00*\x00\x14P@\x01\x08\x13')

The system is not connected to IPv6, no routes, no iPv6 addressese etc, but the kernel understands IPv6.  I fixed it locally with:
```diff
diff --git a/aiohttp/resolver.py b/aiohttp/resolver.py
index 2161c8fa..4e57b319 100644
--- a/aiohttp/resolver.py
+++ b/aiohttp/resolver.py
@@ -38,16 +38,7 @@ class ThreadedResolver(AbstractResolver):
 
         hosts = []
         for family, _, proto, _, address in infos:
-            if family == socket.AF_INET6 and address[3]:  # type: ignore[misc]
-                # This is essential for link-local IPv6 addresses.
-                # LL IPv6 is a VERY rare case. Strictly speaking, we should use
-                # getnameinfo() unconditionally, but performance makes sense.
-                host, _port = socket.getnameinfo(
-                    address, socket.NI_NUMERICHOST | socket.NI_NUMERICSERV
-                )
-                port = int(_port)
-            else:
-                host, port = address[:2]
+            host, port = address[:2]
             hosts.append(
                 {
                     "hostname": hostname,
```

### To Reproduce

See above.

### Expected behavior

Shall not read address[3], when address[3] does not exist.

### Logs/tracebacks

```python-traceback
Not applicable.
```

### Python Version

```console
$ python --version

Python 3.8.8+
```

### aiohttp Version

```console
$ python -m pip show aiohttp

3.8.0a0
```

### multidict Version

```console
$ python -m pip show multidict
5.1.0
```

### yarl Version

```console
$ python -m pip show yarl

1.6.3
```

### OS

Linux from Scratch, python 3.8.8+

### Related component

Client

### Code of Conduct

[...]

Issue #5156, opened 2020-10-26
Tuple index out of range with 3.7.1 and no IPv6


danielnelson:
🐞 **Describe the bug**
After upgrading to aiohttp==3.7.1, dns resolution fails with an exception.

💡 **To Reproduce**

1. I believe this is due to not have IPV6 enabled in my kernel, which is custom compiled with.
   `CONFIG_IPV6=n`
2. Run:
   ```python
   import asyncio
   import aiohttp


   async def main():
       async with aiohttp.ClientSession() as session:
           async with session.get('http://python.org') as response:
               print(response.status)


   asyncio.run(main())
   ```
3. Actual behavior
   ```python-traceback
   $ python bug.py
   Traceback (most recent call last):
     File "bug.py", line 11, in <module>
       asyncio.run(main())
     File "/usr/lib/python3.7/asyncio/runners.py", line 43, in run
       return loop.run_until_complete(main)
     File "/usr/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
       return future.result()
     File "bug.py", line 7, in main
       async with session.get('http://python.org') as response:
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/client.py", line 1124, in __aenter__
       self._resp = await self._coro
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/client.py", line 528, in _request
       req, traces=traces, timeout=real_timeout
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/connector.py", line 541, in connect
       proto = await self._create_connection(req, traces, timeout)
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/connector.py", line 898, in _create_connection
       _, proto = await self._create_direct_connection(req, traces, timeout)
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/connector.py", line 1005, in _create_direct_connection
       hosts = await asyncio.shield(host_resolved)
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/connector.py", line 871, in _resolve_host
       addrs = await self._resolver.resolve(host, port, family=self._family)
     File "/home/dbn/usr/py37/lib/python3.7/site-packages/aiohttp/resolver.py", line 37, in resolve
       if family == socket.AF_INET6 and address[3]:  # type: ignore
   IndexError: tuple index out of range
   ```

[...]

PR #5906, opened 2021-07-21
[WIP]Check if ipv6 is enabled, when processing socket address info

Hanaasagi:
## What do these changes do?

Related to #5901; `getaddrinfo` will return an `(int, bytes)` tuple, if
CPython could not handle the address family. It will cause a index out
of range error in aiohttp.

I think not.

## Related issue number

#5901 and #5156.

## Checklist


Hanaasagi:
Is there any good way to test this change? It needs a CPython that compiled with `--disable-ipv6` and enable ipv6 in Linux.

Dreamsorcerer (approved):
Looks good to me, but I'm not really familiar with this part.

tests/test_resolver.py
```diff
@@ -119,6 +119,30 @@ async def test_threaded_negative_lookup() -> None:
...
+            (
+                socket.AF_INET6,
+                socket.SOCK_STREAM,
+                6,
+                "",
+                (10, b"\x01\xbb\x00\x00\x00\x00*\x04NB\x00\x1a\x00\x00"),
+            )
+        ]
+
+    loop.getaddrinfo = unknown_addrinfo
```
derlih, line 138: I would switch to unittest.MagicMock in this case.
Something like
```
loop.getaddrinfo = MagicMock(return_value=[
            (
                socket.AF_INET6,
                socket.SOCK_STREAM,
                6,
                "",
                (10, b"\x01\xbb\x00\x00\x00\x00*\x04NB\x00\x1a\x00\x00"),
            )
        ])
```

Hanaasagi: Need to use `AsyncMock` here, but it was added in Python 3.8. There are Python 3.7 tests on the current CI.

webknjaz:
FTR this may have caused a regression, according to https://github.com/aio-libs/aiohttp/issues/6195.

asvetlov:
Thanks for the reminder!
I've missed the fact that the auto-backporting has failed.
Backported by 13b64003

PR #5912, opened 2021-07-26
Do not require typing_extensions on Python 3.8+

mgorny:
## What do these changes do?

All the types imported from typing_extensions are present in Python 3.8.
Support using the built-in types and restrict typing_extensions
requirement to Python < 3.8.

No.

## Related issue number

n/a

## Checklist

- [ ] Unit tests for the changes exist (NFC)

### DIFF
--- a/CHANGES/5912.misc
+++ b/CHANGES/5912.misc
@@ -1,0 +1,1 @@
+Fix unnecessary dependency on typing_extensions on Python 3.8+.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -30,8 +30,12 @@ from typing import (
     Union,
 )
 
+try:
+    from typing import Final, final
+except ImportError:  # pragma: no cover
+    from typing_extensions import Final, final
+
 from multidict import CIMultiDict, MultiDict, MultiDictProxy, istr
-from typing_extensions import Final, final
 from yarl import URL
 
 from . import hdrs, http, payload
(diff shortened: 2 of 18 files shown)


mgorny:
I don't know how to fix mypy :-(.

webknjaz:
IIRC it's a known bug of mypy forcing us to use if-else paradigm instead of try-except. This is probably one of the reasons to keep typing-extensions — it makes imports more straightforward.

webknjaz:
Also, if we're to have conditional imports, this should probably live in a private compat module rather than be checked in every import section. typing-extensions provide this sort of shim.

mgorny:
> Also, if we're to have conditional imports, this should probably live in a private compat module rather than be checked in every import section. typing-extensions provide this sort of shim.

While your point makes sense, the distros that no longer support Python < 3.8 would like to work towards removing backwards compatibility packages. Does `aiohttp.typedefs` sound like a good place for compatibility shims, or would you prefer a dedicated module?

Dreamsorcerer:
@mgorny Before you waste more time on this, I'd like to point out that I've already done this for 3.8:
https://github.com/aio-libs/aiohttp/pull/5374

I was probably too lazy to repeat it for master, as it seems likely that Python 3.7 will be dropped before aiohttp 4 is ready for release anyway. Though feel free to copy the things done in there if you do feel like making the changes to master as well.

mgorny:
Oh, thanks for letting me know. I didn't notice the 3.8 branch, and I probably want that one more anyway ;-).

mgorny:
Ok, I'm going to close this for now. I'm sorry for the noise.

PR #5914, opened 2021-07-26
Further optimize the logo/icon

ShadiestGoat:
## What do these changes do?

The only thing this does is make further optimizations to the svg of the icon. It also changes the viewbox from 76 to 24, so that it can be used easier for projects that use tools like simple-icons

Not really, unless the extract the raw path data. It still has the same height & width, so should not affect users.

## Related issue number

None

## Checklist

Unrelated: 
TODO: 


CHANGES/5914.misc
```diff
@@ -0,0 +1 @@
+Changes the icon to be more optimized, and use the 0 0 24 24 viewbox, while keeping the same height and width
```
webknjaz:
```suggestion
Changed the SVG logos to be more optimized and the viewbox to 0 0 24 24, while keeping the same height and width -- :user:`ShadiestGoat`.
```

PR #5927, opened 2021-08-02
[WIP] Add support for Python 3.10 experimental builds

Hanaasagi:
## Checklist


Hanaasagi:
pycares is not ready for Python 3.10. See the action log https://github.com/aio-libs/aiohttp/pull/5927/checks?check_run_id=3223285551 and upstream issue https://github.com/saghul/pycares/issues/134

Dreamsorcerer:
Indeed, we can fix the deprecation warnings in another PR in the meantime though.

.github/workflows/ci.yml
```diff
@@ -93,8 +93,15 @@ jobs:
         #   - pyver: pypy3
         #     no-extensions: 'Y'
         #     os: ubuntu
+        experimental: [false]
+        include:
+          - os: Ubuntu
+            pyver: "3.10.0-alpha - 3.10.0"
+            experimental: true
+            no-extensions: 'Y'
```
webknjaz: It's also important to test with the extensions too.

webknjaz:
Hey @Hanaasagi,
@asvetlov and I want to produce a new release soon. Mind updating this PR?

Hanaasagi:
I rebased the master branch. But I don't figure out why there are two failed test.

- FAILED tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http]
- FAILED tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https]
- FAILED tests/test_proxy_functional.py::test_https_proxy_unsupported_tls_in_tls[pyloop-https]

Looks like they were introduced in PR #5992 .

## About the changes 

### Dependency update
1. `pytest` is supporting Python 3.10 since 6.2.5, see https://docs.pytest.org/en/6.2.x/changelog.html#pytest-6-2-5-2021-08-29 .
2. `six` should be updated to 1.16.0, see https://github.com/benjaminp/six/blob/master/CHANGES
3. `setuptools` should be updated to 56.0.0 or above, see https://setuptools.pypa.io/en/latest/history.html.


### Fixing warnings

1. `ssl.PROTOCOL_SSLv23` is deprecated since 3.6, and `ssl.PROTOCOL_TLS` is deprecated since 3.10. See https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS.
2. `asyncio.get_event_loop()` will emit deprecation warning, if there is no running event loop,  See: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

webknjaz:
> I rebased the master branch. But I don't figure out why there are two failed test.
> 
> * FAILED tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http]
> * FAILED tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https]
> * FAILED tests/test_proxy_functional.py::test_https_proxy_unsupported_tls_in_tls[pyloop-https]
> 
> Looks like they were introduced in PR #5992 .

Yep. And they weren't tested against Python 3.10. Here's what's happening in the CI:
```console
E           ConnectionAbortedError: SSL handshake is taking longer than 60.0 seconds: aborting the connection
```
I wonder if this is caused by proxy.py not supporting Python 3.10 or something on our side... Any ideas? I think it may be okay to temporarily mark the specific test cases with `xfail` under the affected envs if it helps get this PR green.

CHANGES/5927.feature
```diff
@@ -0,0 +1 @@
+Add support for Python 3.10 in Github Action and fix the warning.
```
webknjaz:
```suggestion
Added support for Python 3.10 to Github Actions CI/CD workflows and fix the related deprecation warnings -- :user:`Hanaasagi`.
```

docs/third_party.rst
```diff
@@ -290,6 +290,6 @@ ask to raise the status.
 
 - `aiohttp-retry <https://github.com/inyutin/aiohttp_retry>`_
   Wrapper for aiohttp client for retrying requests. Python 3.6+ required.
-  
+
```
webknjaz, line 293: Could you submit this separately?

Hanaasagi: Emm, the pre-commit hooks will check this...

Hanaasagi:
The secure proxy related test will be passed in 3.10, macos, but failed in other os platform. I marked them with xfail.

docs/third_party.rst
webknjaz: This change is already on master so rebase should take care of it.

tests/test_proxy_functional.py
```diff
@@ -113,6 +117,10 @@ def _pretend_asyncio_supports_tls_in_tls(
     )
 
 
+@pytest.mark.xfail(
+    PY_310 and platform.system() != "Darwin",
+    reason="we need to fix the secure proxy fixture in 3.10",
```
webknjaz: Let's be more precise here:
```suggestion
    raises=ConnectionAbortedError,
    reason="we need to fix the secure proxy fixture in 3.10",
```

tests/test_proxy_functional.py
```diff
@@ -139,6 +147,10 @@ async def test_secure_https_proxy_absolute_path(
     await conn.close()
 
 
+@pytest.mark.xfail(
+    PY_310 and platform.system() != "Darwin",
+    reason="we need to fix the secure proxy fixture in 3.10",
```
webknjaz: Let's be more precise here:
```suggestion
    raises=ConnectionAbortedError,
    reason="we need to fix the secure proxy fixture in 3.10",
```

tests/test_proxy_functional.py
```diff
@@ -15,6 +17,8 @@
 from aiohttp import web
 from aiohttp.client_exceptions import ClientConnectionError
 
+PY_310 = sys.version_info >= (3, 10)
```
webknjaz: I think we could have a reusable marker here
```suggestion
PY_310 = sys.version_info >= (3, 10)
xfail_under_py310_except_macos = pytest.mark.xfail(
    PY_310 and platform.system() != "Darwin",
    raises=ConnectionAbortedError,
    reason="The secure proxy fixture does not seem to work under Python 3.10 on Linux or Windows. See https://github.com/abhinavsingh/proxy.py/issues/622.",
)
```
and apply it on tests as `@xfail_under_py310_except_macos`.

webknjaz: Also, the best place for the version constant seems to be https://github.com/aio-libs/aiohttp/blob/c9c9bec/aiohttp/helpers.py#L57. Consider moving it there.

tests/test_proxy_functional.py
Hanaasagi: Yes, but I think it would be better if we make the decrator a partial function, only reuse the condition and reason arguments. Because the exception behavior is dependent on the logic of test function. This will increase coupling.

tests/test_proxy_functional.py
Hanaasagi: Done.

tests/test_proxy_functional.py
```diff
@@ -139,6 +153,7 @@ async def test_secure_https_proxy_absolute_path(
     await conn.close()
 
 
+@secure_proxy_xfail_under_py310_except_macos()
```
webknjaz: @Hanaasagi and what is the exception here? Even if it's `AssertionError`, set it plz.

Hanaasagi: Done.

Dreamsorcerer:
@Hanaasagi Do you think it's worth readding the experimental matrix back in for 3.11? I see there is already an alpha release available, so would be good to have an early warning of any issues this time. https://github.com/actions/python-versions/releases

asvetlov:
IFAIK cython for py3.11 is broken; we just cannot compile aiohttp extensions.

Dreamsorcerer:
If it's just cython though, then we would still get something useful from the no_extensions tests. We can also just leave the cython tests in, so they should start running successfully at some point in the future.

webknjaz:
@Dreamsorcerer feel free to add a limited matrix extension to cover this but we probably don't need it for the v3.8 release so it can wait.

Issue #4112, opened 2019-09-27
Memory efficient encoding detection


decaz:
At #2549 @asvetlov proposed useful improvement of detection of content encoding to prevent reading the whole response in memory by allowing to specify max data size for sniffing. 

`UniversalDetector` can be used: https://chardet.readthedocs.io/en/latest/usage.html#example-detecting-encoding-incrementally.

asvetlov:
Yes, sure.
Would you make a PR?

decaz:
Unfortunately right now I don't have time to, maybe later. I've created issue just not to forget about it.

icmpecho:
@asvetlov @decaz I was taking a quick look into this and have some open questions. As far as I see the `read` method of `ClientResponse` has already read the entire body and put it into the memory. the `get_encoding` method just make use of that in-memory body right?
So what is your expectation of this PR? Is it about changing the behaviour of `read` so it return a stream or just splitting the already in-memory body into chunks and feed that into the `UniversalDetactor` ?

asvetlov:
We can replace 'resp.read()` with `resp.content.read(2048)` for fetching first 2 Kb buffer (I have no idea what is the best size of sniffing uncoding though).
`content.read()` is not stored in `resp._body`, so we need more complex buffering strategy.

Transfusion:
> We can replace 'resp.read()`with`resp.content.read(2048)`for fetching first 2 Kb buffer (I have no idea what is the best size of sniffing uncoding though).`content.read()`is not stored in`resp._body`

At the time of @decaz 's [original PR](https://github.com/aio-libs/aiohttp/pull/2549/files#diff-2fa8ab44317df4fbb7864536a6794d7a), `await self.read()` is **always** called if the encoding is not detected, i.e. there is no `charset` in the `Content-Type` response header from the web server, so `_content` (aka `_body`) will always be populated when `chardet.detect` is called,

However, as of now, `get_encoding()` always assumes `_body is not None`, i.e. `read()`, `json()`, or `text()` must be called first before it will work, or else it will throw an error if `charset` isn't in the `Content-Type` response header.

https://github.com/aio-libs/aiohttp/blob/a54956d94c21bfbea74153867e182fe47cf12c6d/aiohttp/client_reqrep.py#L927-L944

> so we need more complex buffering strategy.

- If `ClientResponse.read()` has already been called then we can just do detection on the whole body, no issue.
- If it hasn't (`_body is None`)...
[...]

PR #5930, opened 2021-08-04
Improve the default fallback encoding detection

Ousret:
## What do these changes do?

Switch **Chardet** dependency to **Charset-Normalizer** for the fallback encoding detection.

This change is mostly backward-compatible, exception of a thing:

- This new library support way more code pages (x3) than its counterpart Chardet.
  - Based on the 30-ich charsets that Chardet support, expect roughly 90% BC results https://github.com/Ousret/charset_normalizer/pull/77/checks?check_run_id=3244585065

## Why should you bother with such a change? Is it worth it?  

Short answer, absolutely.

Long answer:

- On average x4 faster than Chardet and countless times faster in the larger payload cases (>1MB)
  - https://github.com/Ousret/charset_normalizer/pull/77/checks?check_run_id=3244584863 
- Put a definitive end to the licensing debate around Chardet and LGPL. Remove ANY license ambiguity/restriction for projects bundling aiohttp.
- Never return an encoding if not suited for the given decoder. Eg. Never get UnicodeDecodeError!
- The package size is X4 lower than Chardet’s (4.0)!
- Binary files should never be detected as potential texts.
  - https://github.com/Ousret/char-dataset/blob/master/None/sample-1.gif 
    - Chardet and cChardet sees it as `Windows-1252`
    - None for charset-normalizer
- Especially, because the Unicode detection is far better than Chardet (or cChardet).
  - Some representative examples:
    - https://github.com/Ousret/char-dataset/blob/master/utf_8/reddit_wsb.csv
    - https://github.com/Ousret/char-dataset/blob/master/utf_8/_ude_1.md
      - Chardet sees it as `Windows-1254`
      - cChardet sees it as `ISO-8859-7`
      - `utf_8` for charset-normalizer
    - https://github.com/Ousret/char-dataset/blob/master/utf_8/_ude_1.rst
      - Chardet and cChardet sees it as `Windows-1252`
      - `utf_8` for charset-normalizer
- Actively maintained, watching closely for any concerns.
  - Many workflows are set up to ensure quality continuous delivery.
- `requests` did integrate it first and for total transparency, the lib needed some minors adjustments. But it is going well so far.

It's still a heuristic lib, therefore cannot be trusted blindly of course.

## Is UTF-8 everywhere already?

Not really, that is a dangerous assumption. Looking at https://w3techs.com/technologies/overview/character_encoding may seem like encoding detection is a thing of the past but not really. Solo based on 33k websites, you will find 3,4k responses without predefined encoding. (Top 1000 sites from 80 countries in the world according to Data for SEO) https://github.com/potiuk/test-charset-normalizer

First of all, neither aiohttp, chardet or charset-normalizer are dedicated to HTML content. The detection concern every text document (SubRip Subtitle for ex.). 
It is so hard to find any stats at all regarding this matter. Users' usages can be very dispersed, so making assumptions is unwise.

The real debate is to state if the detection is an HTTP client matter or not. That is more complicated and not my field.

## Related issue number

No related issue.

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -70,7 +70,7 @@ except ImportError:  # pragma: no cover
 try:
     import cchardet as chardet
 except ImportError:  # pragma: no cover
-    import chardet  # type: ignore[no-redef]
+    import charset_normalizer as chardet  # type: ignore[no-redef]
 
 
 __all__ = ("ClientRequest", "ClientResponse", "RequestInfo", "Fingerprint")
--- a/CHANGES/5928.feature
+++ b/CHANGES/5928.feature
@@ -1,0 +1,1 @@
+Switch chardet to charset-normalizer when aiohttp need to guess the body encoding
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -335,3 +335,4 @@ Zeal Wierslee
 Zlatan Sičanica
 Марк Коренберг
 Семён Марьясин
+Ahmed Tahri
--- a/README.rst
+++ b/README.rst
@@ -165,14 +165,14 @@ Requirements
 - Python >= 3.7
 - async-timeout_
 - attrs_
-- chardet_
+- charset-normalizer_
 - multidict_
 - yarl_
 
 Optionally you may install the cChardet_ and aiodns_ libraries (highly
 recommended for sake of speed).
 
-.. _chardet: https://pypi.python.org/pypi/chardet
+.. _charset-normalizer: https://pypi.org/project/charset-normalizer
 .. _aiodns: https://pypi.python.org/pypi/aiodns
 .. _attrs: https://github.com/python-attrs/attrs
 .. _multidict: https://pypi.python.org/pypi/multidict
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1374,10 +1374,10 @@ Response object
       specified *encoding* parameter.
 
       If *encoding* is ``None`` content encoding is autocalculated
-      using ``Content-Type`` HTTP header and *chardet* tool if the
+      using ``Content-Type`` HTTP header and *charset-normalizer* tool if the
       header is not provided by server.
 
-      :term:`cchardet` is used with fallback to :term:`chardet` if
+      :term:`cchardet` is used with fallback to :term:`charset-normalizer` if
       *cchardet* is not available.
 
       Close underlying connection if data reading gets an error,
@@ -1389,14 +1389,14 @@ Response object
 
       :return str: decoded *BODY*
 
-      :raise LookupError: if the encoding detected by chardet or cchardet is
-                          unknown by Python (e.g. VISCII).
+      :raise LookupError: if the encoding detected by charset-normalizer or
+                          cchardet is unknown by Python (e.g. VISCII).
 
       .. note::
 
          If response has no ``charset`` info in ``Content-Type`` HTTP
-         header :term:`cchardet` / :term:`chardet` is used for content
-         encoding autodetection.
+         header :term:`cchardet` / :term:`charset-normalizer` is used for
+         content encoding autodetection.
 
          It may hurt performance. If page encoding is known passing
          explicit *encoding* parameter might help::
@@ -1411,7 +1411,7 @@ Response object
       a ``read`` call will be done,
 
       If *encoding* is ``None`` content encoding is autocalculated
-      using :term:`cchardet` or :term:`chardet` as fallback if
+      using :term:`cchardet` or :term:`charset-normalizer` as fallback if
       *cchardet* is not available.
 
       if response's `content-type` does not match `content_type` parameter
@@ -1449,11 +1449,11 @@ Response object
       Automatically detect content encoding using ``charset`` info in
       ``Content-Type`` HTTP header. If this info is not exists or there
       are no appropriate codecs for encoding then :term:`cchardet` /
-      :term:`chardet` is used.
+      :term:`charset-normalizer` is used.
 
       Beware that it is not always safe to use the result of this function to
       decode a response. Some encodings detected by cchardet are not known by
-      Python (e.g. VISCII).
+      Python (e.g. VISCII). *charset-normalizer* is not concerned by that issue.
 
       :raise RuntimeError: if called before the body has been read,
                            for :term:`cchardet` usage
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -45,11 +45,12 @@
       Any object that can be called. Use :func:`callable` to check
       that.
 
-   chardet
+   charset-normalizer
 
-       The Universal Character Encoding Detector
+       The Real First Universal Charset Detector.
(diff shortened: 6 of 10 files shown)


Ousret:
Ping. 

Would you be interested in such a change in the near future? @Dreamsorcerer @webknjaz 

Regards,

Dreamsorcerer:
I'm not really familiar with either library, so I don't think I'm in a place to make a decision on this. Although it does sound promising.

I've enabled the CI though, so looks like it needs some minor reformatting for flake8.

aiohttp/client_reqrep.py
```diff
@@ -70,7 +70,9 @@
 try:
     import cchardet as chardet
 except ImportError:  # pragma: no cover
-    import chardet  # type: ignore[no-redef]
+    import charset_normalizer as chardet  # type: ignore[no-redef]
+
+    warnings.filterwarnings('ignore', 'Trying to detect', module='charset_normalizer')
```
Dreamsorcerer: I'm curious about this. Why are we ignoring a warning? If there's nothing the user can do in regards to the warning, then it probably shouldn't be a warning in the upstream library (but, logging.warning instead).

Ousret: Good question. 

> Why are we ignoring a warning? If there's nothing the user can do in regards to the warning, then it probably shouldn't be a warning.

You are right about the main logic behinds warnings.

But its a little bit trickier in that case.

The warning targeted here is when a tiny content (less than 32o) is given for detection. Sometime its reliable, sometime not. Those cases are often tiny JSON / ASCII / u8 content in my experiences.

In some ways, the end user can do something about it but often wont. eg. server not adding the explicit `charset` when it obviously should.

It is entirely up to aiohttp maintainers to choose whether or not a warning should be raised in that context.
It could potentially spam the stdout and raises complaints.

I am happy to debate around the necessity of the raised warning inside this library or even in `charset_normalizer`.

Ousret:
The rtd build failed with a syntax error at runtime. I cannot reboot the task.

aiohttp/client_reqrep.py
Dreamsorcerer: > In some ways, the end user can do something about it but often wont. eg. server not adding the explicit charset when it obviously should.

But, if it's mainly going to be used for decoding data that is sent to the user, in most cases, the user will have no control to fix the source.

If we wanted to warn users about cases where the charset is not set, then we'd do that in our code every time charset_normalizer is invoked. Making it dependant on the size of the input seems less useful.

If you have control over the source, in order to fix the warning, why don't you know the encoding already?

Ousret: Frankly I am tempted to remove the warning in favour of a straight forward logging.warning instead.
In a that environment, I would totally agree. I am considering to remove the warning in favour of a simple logging output.
Thanks for your inputs.

Meanwhile, would this be a blocker?

Dreamsorcerer: No, the blocker is somebody else more familiar with this giving an approval.

Ousret: OK. Will wait upon that.

Ousret: I have removed the "silent" warning set for `charset_normalizer`. The release 2.0.5 no longer raise any warning.
From what I have seen in the project (commits&code owners), the best candidate to address the approval is asvetlov.

Dreamsorcerer: I agree, but might be awhile before he has a chance to look over these PRs.

Dreamsorcerer:
> The rtd build failed with a syntax error at runtime. I cannot reboot the task.

Should be fixed now. Just need to merge master.

Ousret:
Merged.
Thanks.

Ousret:
@asvetlov is there anything that could be done to ease the road toward the approval? I would gladly help in any way I can.
From any specific angle to run tests if needed.

Ousret:
I ran some basic tests and got some measurements to share.
You may find the sources of what I produced to get those results here https://github.com/Ousret/aiohttp-tests

Everything went well as expected. No crashes.

### Context

- `aiohttp.TCPConnector(limit=16)`
- alpine linux
- nginx 1.21
- python 3.9
- aiohttp _dev-master_
- chardet 4.0.0 (_aiohttp-chardet_)
- charset-normalizer 2.0.6 (_aiohttp-next_)

Prepare the environment 

```
docker-compose build aiohttp-next server
docker-compose build --no-cache aiohttp-chardet

docker-compose up -d server
```

Collect results

```
docker-compose up aiohttp-chardet
docker-compose up aiohttp-next
```

Memory sampling (to be run in parallel of `aiohttp-*` services)

```
docker stats
```

### RAW performance

This comes with no surprise as it was already proven previously.

- using charset-normalizer
  - 37.7s with 467 files --> 81ms 
- using chardet
  - 189.3s with 467 files --> 405ms

5 times faster while keeping 88% _(410 / 464 files)_ full backward-compatible results.
- see https://github.com/Ousret/charset_normalizer/runs/3637139029 

### RAM footprint

Following issue https://github.com/aio-libs/aiohttp/issues/4112 it could be a good idea to showcase that aspect.
Using `docker stats` for the sampling.

Bellow, the peak usage for each flavor.

- using charset-normalizer
  - 38MiB
- using chardet
  - 67MiB

1.7 times less memory consumption (on peaks). For 5 times faster guess output (on avg).

### Other

Keep in mind that the actual delays may vary depending on your CPU capabilities. 
The factor between chardet/charset-normalizer flavors should remain intact.

You may find the RAW console outputs for both flavors at https://gist.github.com/Ousret/ec22b8842a42fb74276b052b9178dcc4

CHANGES/5930.feature
```diff
@@ -0,0 +1 @@
+Switch chardet to charset-normalizer when aiohttp need to guess the body encoding
```
webknjaz:
```suggestion
Switched ``chardet`` to ``charset-normalizer`` for guessing the HTTP payload body encoding -- :user:`Ousret`.
```

docs/spelling_wordlist.txt
```diff
@@ -126,6 +126,8 @@ cchardet
 ceil
 charset
 charsetdetect
+normalizer
+Chardet
```
webknjaz: Nit: the words in this list are typically ordered alphabetically.

webknjaz:
Hi @Ousret, the change is good. Though, could you apply those minor improvements I've pointed out?

docs/spelling_wordlist.txt
Ousret: Fixed.

Ousret:
I have applied the recommendations.

webknjaz:
@Ousret I've enabled auto-merge which should get this patch into `master` once the tests pass. After that, we'll need to get it backported to the 3.8 branch so that it'll get into the upcoming release of aiohttp v3.8. If automatic backporting fails, you'll need to follow the bot's instructions to do this manually.

webknjaz:
Backport PR: https://github.com/aio-libs/aiohttp/pull/6108.

Ousret:
Thanks for your time. I can see that the backport PR has already been merged.

Issue #2304, opened 2017-10-07
Cannot easily access max_field_size in client response


rckclmbr:
## Long story short

The http client doesn't really support overriding the default header value length of 8190, even though the parser parameter is there https://github.com/aio-libs/aiohttp/blob/e7c9390111932dd6dbb642170d7a0da1876271ec/aiohttp/http_parser.py#L59

This results in the error message:

```
aiohttp.client_exceptions.ClientResponseError: 400, message='Got more than 8190 bytes when reading Header value is too long.'
```

Stack trace from real app:
```
<snip>
File "/usr/local/lib/python3.6/site-packages/pyportify/google.py", line 113, in _http_get
params=merged_params,
File "/usr/local/lib/python3.6/site-packages/aiohttp/helpers.py", line 97, in iter
ret = yield from self._coro
File "/usr/local/lib/python3.6/site-packages/aiohttp/client.py", line 241, in _request
yield from resp.start(conn, read_until_eof)
File "/usr/local/lib/python3.6/site-packages/aiohttp/client_reqrep.py", line 564, in start
message=exc.message, headers=exc.headers) from exc
aiohttp.client_exceptions.ClientResponseError: 400, message='Got more than 8190 bytes when reading Header value is too long.'
```

## Expected behaviour

I can either set the parser in the session/get request, or set the max_field_size.

## Actual behaviour

I can't do either

## Steps to reproduce
I believe this will do it:

```
import aiohttp
import asyncio

@asyncio.coroutine
def main():
        with aiohttp.ClientSession() as session:
                resp = yield from session.get('http://test.xr6.me')
                print(len(resp.headers['X-TEST-HEADER']))
                resp.close()

if __name__ == "__main__":
        loop = asyncio.get_event_loop()
        loop.run_until_complete(
            asyncio.gather(main())
        )
        loop.close()
```

## Your environment

Ubuntu jessie (bash for windows, reproduces on others though)
python3.4 (also reproduces on others)

asvetlov:
Unfortunately current client design have no place for requested parameter.
After #2019 done we could return to the issue.

wumpus:
Isn't our goal to be more generous than a browser? I don't know if browsers discard these long lines or not, but they certainly don't error out. In which case 8190 is obviously too small. If we aren't going to make this limit settable, raising it a lot is a great solution.

asvetlov:
IFAIK browsers silently discard things that are longer that the browser expects.
[...]

PR #5931, opened 2021-08-05
Set more reasonable default header limits (64kb)

mvanderlee:
## What do these changes do?

Set more reasonable default header limits (64kb) and make them configurable via environment variables.
I opted to go with environment variables because it was a lot quicker and easier than traversing the entire callstack and adding it to the client code. 

My use-case is to process responses from Trino, which has a default limit of 2MB, so configurability is a must.

No

## Related issue number

#2304, at least partially. 

## Checklist

### DIFF
--- a/CHANGES/2304.bugfix
+++ b/CHANGES/2304.bugfix
@@ -1,0 +1,1 @@
+Set more reasonable default header limits and make them configurable via environment variables.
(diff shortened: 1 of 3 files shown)


asvetlov:
NGINX has [8kb limit](https://nginx.org/en/docs/http/ngx_http_core_module.html#large_client_header_buffers), aiohttp often follows nginx defaults.

Regarding dynamic limits configuration -- I agree, the feature is useful.
But the configuration should be controlled by Connector class, not globally by env vars.
aiohttp is a library, not a  program.
If user wants to use env vars for configuration -- that is perfectly fine but out of aiohttp project scope.

I'm closing this PR but open for another proposal that configures HttpResponseParser in a way similar to timeouts, read_bufsize and auto_decompress configuration.

PR #5932, opened 2021-08-06
Dependabot auto-merge

Dreamsorcerer:
All remaining PRs on aiohttp-jinja2 auto-merged successfully, so looks like this is the solution.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,22 @@
+name: Dependabot auto-merge
+on: pull_request_target
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.1.0
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs
+        run: gh pr merge --auto --squash "$PR_URL"
+        env:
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

.github/workflows/auto-merge.yml
```diff
@@ -0,0 +1,22 @@
...
+on: pull_request_target
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
```
webknjaz, line 11: Strictly speaking, if-clauses do not need interpolation syntax and should only contain expressions.

Dreamsorcerer: Tell that to Microsoft. :P
That's a verbatim copy/paste from the docs.
https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request

webknjaz:
@Dreamsorcerer this patch merges the PRs with the GH-provided token. As a result, merge commits don't produce updates and the CI doesn't run after the merge.
The consequence is that this means that the CI also doesn't upload the coverage report to Codecov. Meaning that Codecov shows a drop in coverage on each of such merges: https://app.codecov.io/gh/aio-libs/aiohttp/branch/master/commits?page=1.

Dreamsorcerer:
OK, this suggests using a personal access token:
https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token

Dreamsorcerer:
Although I'm curious that those merges seem to have a pre-commit status check still run against them.

PR #5811, opened 2021-06-19
Remove leftover usage of the attrs library in runtime

RawatDevanshu:
fix #5806

PR #5934, opened 2021-08-07
Add flake8-requirements

Dreamsorcerer:


setup.cfg
```diff
@@ -64,67 +71,3 @@ minversion = 3.8.2
 testpaths = tests/
 junit_family=xunit2
 xfail_strict = true
-
-[mypy]
```
webknjaz: This doesn't look related. Submit it separately?

.github/workflows/ci.yml
```diff
@@ -47,9 +47,12 @@ jobs:
         python setup.py install
       env:
         AIOHTTP_NO_EXTENSIONS: 1
-    - name: Run linters
+    - name: Mypy
       run: |
         make mypy
+    - name: Flake8
+      run: |
+        flake8
```
webknjaz: This should be integrated into pre-commit instead.

Dreamsorcerer: Hmm, it already is. So, why does flake8 pass without errors on my first commit?

webknjaz: Maybe it runs in a slightly different env (because it has its own venv set up by pre-commit).

webknjaz: It's because you have to add this plugin as an additional dependency to the pre-commit hook config. It doesn't use these pip-tools' lock files. Well, unless you remake how the hook is set up it won't.

webknjaz: This is why it's hard to integrate the pin files into pre-commit w/o an external wrapper: https://github.com/pre-commit/pre-commit/issues/1963.

Dreamsorcerer:
Blocked by #5811.

Dreamsorcerer:
Should we add tokio to the base requirements file, like uvloop is? I think both are optional, so not sure why one is currently included and the other is not..

webknjaz:
> Should we add tokio to the base requirements file, like uvloop is? I think both are optional, so not sure why one is currently included and the other is not..

Not sure, I think it may've had problems on some CIs or envs in the past. I don't remember.

Dreamsorcerer:
> > Should we add tokio to the base requirements file, like uvloop is? I think both are optional, so not sure why one is currently included and the other is not..
> 
> Not sure, I think it may've had problems on some CIs or envs in the past. I don't remember.

Added it in, and doesn't seem to be complaining. If a problem does occur, then we can just `# noqa` it in future.

tests/test_connector.py
```diff
@@ -26,6 +25,8 @@
 from aiohttp.test_utils import make_mocked_coro, unused_port
 from aiohttp.tracing import Trace
 
+from .conftest import needs_unix
```
webknjaz: Plz don't add `__init__`. In fact, we shouldn't import from `conftest` at all. I'd rewrite this to be `pytest.mark.usefixtures()` itstead.

Dreamsorcerer: It's not a fixture though. The pytest docs suggest doing it this way (but not from conftest, and not a relative import, which seems questionable to me): https://docs.pytest.org/en/stable/skipping.html#id1

It is only used in this single test file though, so I could just move it into this file?

webknjaz: I'd still refactor it as a fixture. Having `__init__` is the test dir is usually discouraged, hence no relative import.

Dreamsorcerer: This came up again at work today, and after looking at the docs a bit more, I think you might be wrong on that. It appears the lack of `__init__.py` is only supported via a legacy feature:
> This is the classic mechanism, dating back from the time Python 2 was still supported.

While having `__init__.py` appears to be a requirement in future:
> Users which require this should turn their tests into proper packages instead.
>
> We intend to make importlib the default in future releases.

https://docs.pytest.org/en/6.2.x/pythonpath.html#import-modes

So, adding `__init__.py` now will just save us time updating it in future. If we're worried about the possible side effects that can be caused by this, we can enable `--import-mode=importlib` and start using the new approach now, which doesn't suffer from those side effects.

webknjaz: I haven't checked if adding `__init__.py` became recommended but I'm pretty sure that importing things from `conftest.py` goes against the idea of fixtures which was my primary concern.

Dreamsorcerer: I couldn't figure out any way to use the skip as a fixture. It doesn't make any sense to me, unless I'm misunderstanding something.

I can just move it into this file though instead. It is only used in one place. Though this is arguably outside the scope of the PR anyway, as it is already being imported from conftest. I've merely changed an absolute import to a relative one.

webknjaz: Yeah, I guess moving it is one way to do it.

webknjaz: > I couldn't figure out any way to use the skip as a fixture.

There are several ways I can think of off the top of my head to do this:
1) use `pytest.skip()` (not `pytest.mark.skip`!) inside a fixture and that will skip the test
2) use https://pypi.org/project/pytest-platform-markers/
3) declare a custom marker in `pytest.ini`, use it as `pytest.mark.needs_unix` and use a `pytest_runtest_setup` hook in `conftest` to add a skip as documented at https://docs.pytest.org/en/6.2.x/example/markers.html#marking-platform-specific-tests-with-pytest

webknjaz: So yeah... I'd submit this change as a separate PR I suppose, making this one atomic too. Sounds good?

PR #5935, opened 2021-08-07
Enable flake8

Dreamsorcerer:
For some reason flake8 is not enabled in CI?


webknjaz:
Context: https://github.com/aio-libs/aiohttp/pull/5934#discussion_r685333220.

Issue #5634, opened 2021-04-20
[server] Add warning when a cookie's length would exceed the RFC 6265 minimum user-agent support


panagiks:
🐣 **Is your feature request related to a problem? Please describe.**
We had a report in `aiohttp-session` (see aio-libs/aiohttp-session#574) that a user submitting a 'large' cookie had the cookie dropped by the receiving browser.

According to [RFC 6265 - section 6.1](https://tools.ietf.org/html/rfc6265#section-6.1):

> General-use user agents SHOULD
   provide each of the following minimum capabilities:
>   o  At least 4096 bytes per cookie (as measured by the sum of the
      length of the cookie's name, value, and attributes).

This means that cookies that exceed 4096 bytes in storage (not transmission) have a chance that the user-agent will drop them which is a case that is hard to 'guess' for a developer.

💡 **Describe the solution you'd like**
A warning log should be emitted if a cookie that would exceed the RFC minimum support is created in `aiohttp` server.

A challenge here is the definition of 'exceed' because `aiohttp` knows the transmitted size but the RFC refers to (and the user-agents seem to implement) a check on the final stored size of the cookie string, including all field names and values regardless of whether they were transmitted or filled in as defaults by the user-agent.

❓ **Describe alternatives you've considered**
It was briefly considered implementing this check in `aiohttp-session` but was rejected since this is a general issue with cookies and not specific to `aiohttp-session` so such a check could benefit all `aiohttp` users.

📋 **Additional context**
None.

Dreamsorcerer:
The warning would go somewhere around:
https://github.com/aio-libs/aiohttp/blob/1a4126a2c0e52b331aa953288e844bac680290fc/aiohttp/helpers.py#L852

Maybe requires a little trial and error with a browser forcing this limit, to help figure out exactly what is included in the size limit (the linked issue mentions this being reproducible in Chrome 89).

But, at a guess, this might look something like: `len(name) + len(value) + sum(len(k) + len(v) for k, v in c.items())`

anesabml:
@Dreamsorcerer I came across a similar bug report in `Django`, eventually, they didn't implement a warning or anything like that (they added a warning to the documentation), because of performance reasons (calculating the whole length of every cookie in the cookies set can introduce a performance overhead).

https://code.djangoproject.com/ticket/22242

Dreamsorcerer:
Well, that's what dev mode is for, then atleast it will catch errors for people during development.
[...]

PR #5959, opened 2021-08-20
Add a warning when a cookie's length exceeds 4096 bytes

anesabml:
## What do these changes do?

Add warning when a cookie's length exceeds the RFC 6265 minimum client support.

The user might see a warning if a cookie size exceeds the RFC 6265 minimum client support.

## Related issue number

#5634 

## Checklist


aiohttp/helpers.py
```diff
@@ -852,6 +853,15 @@ def set_cookie(
         self._cookies[name] = value
         c = self._cookies[name]
 
+        if DEBUG:
+            cookie_length = (
+                len(name) + len(value) + sum(len(k) + len(v) for k, v in c.items())
```
Dreamsorcerer: Have you reproduced the issue in Chrome and tested the different lengths? I think I saw somewhere recently that the actual limit is on the total length of the string that is output in the header. If so, this probably misses some spaces or joining characters.

aiohttp/helpers.py
```diff
@@ -852,6 +853,15 @@ def set_cookie(
...
+            )
+            if cookie_length > COOKIE_MAX_LENGTH:
+                server_logger.warning(
```
Dreamsorcerer: Codecov is now adding inline comments, that's cool. As it suggests, we should add a test for this case. Maybe also a test that our logic matches 4096 bytes in the output.

Something like:
```
input_4096...
assert no_warning_raised
assert len(cookie_header) == 4096

input_4097...
assert warning_raised
assert len(cookie_header) == 4097
```

Dreamsorcerer: Also, I think this should be `warnings.warn()`, see: https://docs.python.org/3/howto/logging.html#when-to-use-logging

CHANGES/5634.misc
```diff
@@ -0,0 +1 @@
+Add warning when a cookie's length exceeds the RFC 6265 minimum client support
```
webknjaz: Sphinx has built-in RST roles for linking RFCs:
```suggestion
A warning was added, when a cookie's length exceeds the :rfc:`6265` minimum client support.
```

aiohttp/helpers.py
```diff
@@ -852,6 +853,15 @@ def set_cookie(
         self._cookies[name] = value
         c = self._cookies[name]
 
+        if DEBUG:
+            cookie_length = (
```
webknjaz: The cookie length is being modified later on, after this if-block. Should it be moved?

webknjaz:
cc @panagiks

aiohttp/helpers.py
anesabml: Yes, it should be moved.

tests/test_helpers.py
```diff
@@ -840,3 +840,20 @@ def test_populate_with_cookies():
...
+    sut = helpers.CookieMixin()
+
+    assert sut.cookies == {}
+
+    with pytest.warns(None) as record:
+        sut.set_cookie("name", "value")
+
+    assert len(record) == 0
+
+    value = "a" * 4097
```
Dreamsorcerer: I'll try and test out the limits in a bit, but I'm wondering if the 4096 limit may be based on the output of `len(output) for output in sut.cookies.output(header="").split("\r\n")`.

aiohttp/helpers.py
anesabml: Nice note, what about characters that are more than 1 byte like ÿ that is as far as I know stored as %C3%BF which is 6 bytes. doing `len('ÿ')` results into 1.

Dreamsorcerer: That is a good point, maybe they should be converted to `bytes`?

Dreamsorcerer:
So, in Firefox it only counts key + value, limited to 4096. But, in Chromium it counts the length of the raw cookie header, regardless of how much data is actually stored.

So, to be on the safe side, we should limit the total output to 4096 bytes. A good test might be roughly:

```python
async def test_warn_large_cookie(buf: Any, writer: Any) -> None:
    resp = Response()
    resp.set_cookie("foo", "8"*4070, max_age=2600)
    req = make_request("GET", "/", writer=writer)

    await resp.prepare(req)  # No warning
    await resp.write_eof()

    cookie = re.search(b"Set-Cookie: (.*?)\r\n", buf).group(1)
    assert len(cookie) == 4096
    buf[:] = b""

    resp = Response()
    resp.set_cookie("foo", "8"*4071, max_age=2600)
    req = make_request("GET", "/", writer=writer)

    with pytest.warns(UserWarning, match="size of the cookie"):
        await resp.prepare(req)
    await resp.write_eof()

    cookie = re.search(b"Set-Cookie: (.*?)\r\n", buf).group(1)
    assert len(cookie) == 4097
```

I think this would also need a new way to run tests with dev mode. I don't see any examples online about how to do this, but maybe just marking tests which need dev mode would work, then we can add an extra pytest invocation which only runs the dev mode tests with something like `python -X dev -m pytest -m dev_mode`.

Dreamsorcerer:
Although, if the check becomes trivial at runtime, then maybe we don't need dev mode after all.

aiohttp/helpers.py
webknjaz: Consider using `sys.getsizeof()` instead.

webknjaz: Note that for complex structures you'd have to go for recursion that may be expensive: https://code.activestate.com/recipes/577504/.

webknjaz:
> But, in Chromium it counts the length of the raw cookie header, regardless of how much data is actually stored.

We should be RFC-compliant. Deviating from the standards leads to chaos: https://xkcd.com/927/.

Dreamsorcerer:
> > But, in Chromium it counts the length of the raw cookie header, regardless of how much data is actually stored.
> 
> We should be RFC-compliant. Deviating from the standards leads to chaos: https://xkcd.com/927/.

The RFC isn't that clear on precisely what counts. But, Chrome has the most pessimistic approach, so if we warn on this case, we can be sure to catch all possible issues. i.e. No browser is going to drop cookies that are smaller than what Chrome is accepting (and if they did, it would definitely be violating the RFC, whereas Chrome is more of a grey-area due to the imprecise explanation in the RFC).

On this occasion, I think it makes more sense to take the pessimistic approach and play it safe. Maybe to avoid confusion to users, we shouldn't mention the 4096 bytes in the warning and just say something like "Sending a large cookie. Some browsers may ignore this."

aiohttp/helpers.py
anesabml: I think that won't be a problem because we are computing the length of the raw cookie header which is a string.

webknjaz: A length of a Unicode string may be smaller than the length of its underlying byte representation since some char would be represented as more than one byte:
```pycon
$ ipython
Python 3.9.0 (default, Oct 26 2020, 13:08:59) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: len('ї')
Out[1]: 1

In [2]: len('ї'.encode())
Out[2]: 2
```

webknjaz: Or are you implying that only ASCII is allowed there? Is it actually checked anywhere?

aiohttp/helpers.py
```diff
@@ -876,6 +878,14 @@ def set_cookie(
         if samesite is not None:
             c["samesite"] = samesite
 
+        if DEBUG:
+            cookie_length = len(c.output(header="")[1:])
+            if cookie_length > COOKIE_MAX_LENGTH:
+                warnings.warn(
+                    "The size of is too large, it might get ignored by the client.",
+                    UserWarning,
+                )
```
webknjaz, line 888: You want `stacklevel=2` almost always:
```suggestion
                    stacklevel=2,
                )
```
(this will make the warning point at the calling code, meaning the user app, rather than aiohttp internals)

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
...
+        match="The size of is too large, it might get ignored by the client.",
+    ) as record:
+        resp.set_cookie("foo", "8" * 4071, max_age=2600)
+    req = make_request("GET", "/", writer=writer)
+
+    await resp.prepare(req)
+    await resp.write_eof()
+
+    cookie = re.search(b"Set-Cookie: (.*?)\r\n", buf).group(1)
+    assert len(cookie) == 4097
```
webknjaz, line 1201: Could this be something like
```suggestion
    assert re.match(b"Set-Cookie: [^\s]{4097}\r\n", buf)
```
?

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
...
+    assert len(cookie) == 4096
+    buf[:] = b""
+
+    resp = Response()
+
+    with pytest.warns(
+        UserWarning,
+        match="The size of is too large, it might get ignored by the client.",
+    ) as record:
+        resp.set_cookie("foo", "8" * 4071, max_age=2600)
```
webknjaz: It's probably better to do all parts of the check in the same place rather than in other part of the test
```suggestion
    ) as warnings_caught:
        resp.set_cookie("foo", "8" * 4071, max_age=2600)
    assert len(warnings_caught) == 1

```

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
...
+    resp = Response()
+    resp.set_cookie("foo", "8" * 4070, max_age=2600)  # No warning
+    req = make_request("GET", "/", writer=writer)
+
+    await resp.prepare(req)
+    await resp.write_eof()
+
+    cookie = re.search(b"Set-Cookie: (.*?)\r\n", buf).group(1)
+    assert len(cookie) == 4096
+    buf[:] = b""
```
webknjaz: Why do you reuse a variable like this? Maybe create a new one?

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
     def test_content_type_is_overrideable(self) -> None:
         resp = json_response({"foo": 42}, content_type="application/vnd.json+api")
         assert "application/vnd.json+api" == resp.content_type
+
+
+@pytest.mark.skipif(
+    not bool(os.environ.get("PYTHONASYNCIODEBUG")),
+    reason="Only run the test with debug mode",
```
webknjaz: Why? Is it possible to enable the debug mode for this test temporarily?

aiohttp/helpers.py
```diff
@@ -876,6 +878,14 @@ def set_cookie(
         if samesite is not None:
             c["samesite"] = samesite
 
+        if DEBUG:
+            cookie_length = len(c.output(header="")[1:])
```
webknjaz, line 882: What exactly are you cutting off with `[1:]`? A space? Then, it'd be more readable to use `str.lstrip()` to trim it.

webknjaz: If not, then at least add a code comment explaining this magic.

aiohttp/helpers.py
Dreamsorcerer: He was responding to your comment about needing recursion, which shouldn't be needed as they are just strings.

We'll need to ensure we are checking the size of the bytes and not the string length as mentioned previously.

This thread is also referencing old code, so I'll resolve it to avoid confusion. We'll make sure that a 2-byte character is included in the tests to cover this case.

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
     def test_content_type_is_overrideable(self) -> None:
         resp = json_response({"foo": 42}, content_type="application/vnd.json+api")
         assert "application/vnd.json+api" == resp.content_type
+
+
+@pytest.mark.skipif(
+    not bool(os.environ.get("PYTHONASYNCIODEBUG")),
```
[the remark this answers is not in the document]
Dreamsorcerer: Doesn't appear to work for me:
```
>>> async def foo():
...  asyncio.get_running_loop().set_debug(True)
...  print(sys.flags.dev_mode)
... 
>>> asyncio.run(foo())
False
```

tests/test_web_response.py
Dreamsorcerer: `sys.flags.dev_mode` is read-only, so not that I can see...

tests/test_web_response.py
Dreamsorcerer: That was me, it comes from a fixture, so seemed weird to create a second one manually.
At the same time, I'm not really sure why there is a fixture just to create a `bytearray()` in the first place...

tests/test_web_response.py
Dreamsorcerer: I don't think there's really any need to check the record anyway. `with pytest.warns(..):` will fail if no warning happens, so there's no real need to check how many warnings were caught.

tests/test_web_response.py
Dreamsorcerer: It could, though I wonder if that's being a little too smart. I feel like the original code is clearer.

tests/test_web_response.py
```diff
@@ -1171,3 +1173,36 @@ def test_text_is_json_encoded(self) -> None:
...
+    reason="Only run the test with debug mode",
+)
+async def test_warn_large_cookie(buf: Any, writer: Any) -> None:
+    resp = Response()
+    resp.set_cookie("foo", "8" * 4070, max_age=2600)  # No warning
```
Dreamsorcerer: To ensure we are testing bytes. Let's do something like:
`value = "ÿ" + "8" * 4068`
And same again for the second test below.

tests/test_web_response.py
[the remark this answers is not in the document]
Dreamsorcerer: I think this should also not be skipping based on the environment variable. We could just import `DEBUG` from `helpers.py` and then skip on `not DEBUG`, which would actually match the code then.

aiohttp/helpers.py
Dreamsorcerer: I think `lstrip()` could potentially cause issues if a cookie started with whitespace or something. I suspect the output code produces something like "{header} {cookie}", so there's an extra dividing character that doesn't count towards the cookie.

Could also just count 1 less: `len(c.output(header="")) - 1`.

tests/test_web_response.py
Dreamsorcerer: Unless anybody has any other ideas, then I think we'll just need to add an extra pytest invocation to CI for dev mode as I mentioned before, using something like `python -X dev -m pytest`.

aiohttp/helpers.py
anesabml: Noted, Thanks for the tip.

tests/test_web_response.py
anesabml: I think what @webknjaz suggests is more clear.

tests/test_web_response.py
anesabml: But we still need to check that no warnings happened.

tests/test_web_response.py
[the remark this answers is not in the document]
webknjaz: I'm pretty sure it doesn't enable debug globally, just for the specific I/O loop.

tests/test_web_response.py
webknjaz: While flags are provisioned at the beginning of the interpreter startup, I was under impression that it's the loop mode that needed to be set.

tests/test_web_response.py
webknjaz: Probably for reuse and consistency. Anyway, I don't think there should be more than one test in a test.

tests/test_web_response.py
webknjaz: Probably for reuse and consistency. Anyway, I don't think there should be more than one test in a test.

tests/test_web_response.py
webknjaz: This helper fails if there's no warnings.

tests/test_web_response.py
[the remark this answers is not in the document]
Dreamsorcerer: But, `DEBUG` uses `sys.flags.dev_mode`, so I think that's what we need to have working.

tests/test_web_response.py
anesabml: I am not sure why `"ÿ"` takes 6 bytes when calculating the length of the cookie header but `len('ÿ'.encode("utf-8"))` gives 2, maybe because `len('ÿ'.encode("utf-8"))` returns a byte string.

tests/test_web_response.py
anesabml: Or we could replace the test invocation with `python -X dev -m pytest` instead of re-running all the tests in debug mode.

webknjaz: I think that may be expensive. I'd rather not slow down the whole test run. Why does this need to depend on the global debug, anyway?

Dreamsorcerer: > Or we could replace the test invocation with `python -X dev -m pytest` instead of re-running all the tests in debug mode.

Yes, sorry, back to my original suggestion. Replace the `skipif` with a `pytest.mark.dev_mode`, and then we can run `python -X dev -m pytest -m dev_mode`. This will only run the dev_mode tests in the extra invocation.

Dreamsorcerer: And we then need to disable tests marked as dev_mode by default for the other runs.

tests/test_web_response.py
Dreamsorcerer: It's not encoded to utf-8:
```
>>> from http.cookies import _Translator
>>> "ÿ".translate(_Translator)
'\\377'
```
It's 4 bytes, but I think it may also result in adding `"` around the cookie value for another 2 bytes.

Dreamsorcerer: Not sure if Chrome counts it this way, or after translating it back to utf-8 or something. But, following the pessimistic approach, counting this encoded value is the safe option. It is only a warning after all.

tests/test_web_response.py
Dreamsorcerer: > Why does this need to depend on the global debug, anyway?

https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L75

aiohttp/helpers.py
anesabml: I agree with @Dreamsorcerer, if the cookie has more than one extra space character `lstrip()` will remove them which is not what we want here.

Makefile
```diff
@@ -91,11 +91,13 @@ test: .develop
 
 .PHONY: vtest
 vtest: .develop
-	@pytest -s -v
+	@pytest -s -v -m "not dev_mode"
+	@python -X dev -m pytest -s -v -m dev_mode
```
Dreamsorcerer: Can we try adding the 'not dev_mode' option by default, so a simple `pytest` will still run correctly?
https://stackoverflow.com/a/64729160

anesabml: Sure.

Dreamsorcerer (approved):

webknjaz:
Thanks, everyone! I've set it to auto-merge on CI success.

CHANGES/5634.misc
```diff
@@ -0,0 +1 @@
+A warning was added, when a cookie's length exceeds the :rfc:`6265` minimum client support.
```
webknjaz:
```suggestion
A warning was added, when a cookie's length exceeds the :rfc:`6265` minimum client support -- :user:`anesabml`.
```

webknjaz:
This change cannot be backported to `3.8` because it relies on the code that hasn't been (and can't be) backported: https://github.com/aio-libs/aiohttp/pull/5197#issuecomment-725654319.

Dreamsorcerer:
I don't see why, the code just needs to be added to the `Response` class instead of `CookieMixin`, at the end of this function:
https://github.com/aio-libs/aiohttp/blob/3.8/aiohttp/web_response.py#L255

webknjaz:
Maybe. But my point is that it wouldn't be the same thing exactly. But if anybody wants to make a manual effort, feel free to do so.

PR #5963, opened 2021-08-23
Stop codecov emailing bad reports

Dreamsorcerer:
On every PR, we get a notification email from codecov telling us that the coverage has dropped. This is only because the other tests have not yet finished running.

This should stop codecov reporting until the other tests are complete.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,6 @@
 codecov:
   branch: master
+  require_ci_to_pass: true
 
 coverage:
   range: "95..100"


Dreamsorcerer:
Well, that didn't work inside the PR, but might work after merging?

As an alternative, there is also the `codecov.notify.wait_for_ci` option, which the documentation makes sound like it does exactly the same thing.

I'm also not clear on whether it waits for non-required statuses like LGTM as well, which is probably not desired. If so, we could also try the `codecov.notify.after_n_builds` option, although it might be a little finicky as the required number of statuses may change over time.

https://docs.codecov.com/docs/codecovyml-reference#codecovrequire_ci_to_pass

PR #5992, opened 2021-09-08
Add secure proxy support in the client

### LINKED ISSUE #3816: Implement HTTPS Proxy Support
## Long story short

Once TLS in TLS support is added to asyncio (https://bugs.python.org/issue37179) is landed to (*hopefully*) Python 3.8 we should plan including support for HTTPS Proxy servers if Python >= 3.8.

## Expected behaviour

Have the ability to use no authentication HTTPS proxies and Authentication via using custom SSL Contexts: [Example Code](https://pastebin.com/rA2edLrD)

## Actual behaviour

Today we get explicitly told aiohttp can't do HTTPS proxies:
```
[cooper:~:( (aiohttp_auth_proxy))]$ python3.7 aioclient.py
HTTPS proxies https://proxy:1443 are not supported, ignoring
```

## Steps to reproduce

Set via the environment or other means to use an HTTPS proxy.
```python
#! /usr/bin/env python3
 
import asyncio
import logging
import os
import ssl
import sys
import time
 
import aiohttp
 
 
PROXY_PORT = 1443
PROXY_HOSTNAME = "https://proxy.company.com"
EXTERNAL_ENDPOINT = "https://www.google.com"
CA_BUNDLE = "/var/certs/ca.pem"
 
 
logger = logging.getLogger(__name__)
logger.setLevel(logging.INFO)
logger.addHandler(logging.StreamHandler(sys.stdout))
 
 
async def run_example():
    cert = os.environ["TLS_CL_CERT_PATH"]
    key = os.environ["TLS_CL_KEY_PATH"]
 
    if not cert or not key:
        raise ValueError("Missing key TLS cert settings.")
 
    # For the example lets ensure HTTPS_PROXY is set
    os.environ["HTTPS_PROXY"] = f"{PROXY_HOSTNAME}:{PROXY_PORT}"
 
    # Setup SSL Fun
    ssl_ctx = ssl.create_default_context(cafile=CA_BUNDLE)
    ssl_ctx.load_cert_chain(cert, key)
    conn = aiohttp.TCPConnector(ssl=ssl_ctx)
 
    start_time = time.time()
    # trust_env allows HTTP(s)_PROXY vars to work
    async with aiohttp.ClientSession(connector=conn, trust_env=True) as session:
        async with session.get(EXTERNAL_ENDPOINT) as response:
            logger.info(
                "Received response with status code "
                + f"{response.status} in {time.time() - start_time}s"
            )
 
 
if __name__ == "__main__":
    asyncio.run(run_example())
```

## Your environment
- Python 3.7.3
- aiohttp 3.5.4

### LINKED ISSUE #4268: Use loop.starttls() on Python 3.7+

bmbouter:
## What do these changes do?

This patch opens up the code path and adds the implementation that allows end-users to start sending HTTPS requests through HTTPS proxies.

The support for TLS-in-TLS (needed for this to work) in the stdlib is kinda available since Python 3.7 but is disabled for `asyncio` with an attribute/flag/toggle. When the upstream CPython enables it finally, aiohttp v3.8+ will be able to work with it out of the box.

Currently the tests monkey-patch `asyncio` in order to verify that this works. The users who are willing to do the same, will be able to take advantage of it right now. Eventually (hopefully starting Python 3.11), the need for monkey-patching should be eliminated.

Refs:
* https://bugs.python.org/issue37179
* https://github.com/python/cpython/pull/28073
* https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support
* https://github.com/aio-libs/aiohttp/discussions/6044

<details>
<summary>
  Original details
</summary>

I test with this script:

```python
import aiohttp
import asyncio


url = "http://example.com"
url = "https://example.com"

proxy_url = "http://localhost:3128/"
proxy_url = "https://localhost:3130/"

async def main():
    import pydevd_pycharm
    # pydevd_pycharm.settrace('localhost', port=29437, stdoutToServer=True, stderrToServer=True, suspend=False)
    async with aiohttp.ClientSession(headers={"Cache-Control": "no-cache"}) as session:
        async with session.get(url, proxy=proxy_url, verify_ssl=False) as resp:
            print(resp.status)
            print(resp)
            print(await resp.text())


```

And I get

```python-traceback
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 22, in <module>
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 15, in main
    async with session.get(url, proxy=proxy_url, verify_ssl=False) as resp:
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 1117, in __aenter__
    self._resp = await self._coro
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 520, in _request
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 535, in connect
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 890, in _create_connection
    _, proto = await self._create_proxy_connection(req, traces, timeout)
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 1139, in _create_proxy_connection
    transport, proto = await self._wrap_create_connection(
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 969, in _wrap_create_connection
    return await self._loop.create_connection(*args, **kwargs)  # type: ignore  # noqa
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 1081, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 1111, in _create_connection_transport
    await waiter
ConnectionResetError
```

The squid proxy does show `1631129861.057     45 ::1 TCP_TUNNEL/200 39 CONNECT example.com:443 - HIER_DIRECT/93.184.216.34 -` and I can see the 200 OK from the direct connection.

What fails is when taking the TCP socket from the direct connection and TLS wrapping it I get a `ConnectionResetError`.

</details>

They get the ability to send out HTTPS queries through HTTPS proxies if they monkey-patch the stdlib `asyncio`.

## Related issue number

Resolves #3816
Resolves #4268

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -481,8 +481,6 @@ class ClientRequest:
         proxy_auth: Optional[BasicAuth],
         proxy_headers: Optional[LooseHeaders],
     ) -> None:
-        if proxy and not proxy.scheme == "http":
-            raise ValueError("Only http proxies are supported")
         if proxy_auth and not isinstance(proxy_auth, helpers.BasicAuth):
             raise ValueError("proxy_auth must be None or BasicAuth() tuple")
         self.proxy = proxy
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -951,6 +951,100 @@ class TCPConnector(BaseConnector):
         except OSError as exc:
             raise client_error(req.connection_key, exc) from exc
 
+    def _warn_about_tls_in_tls(
+        self,
+        underlying_transport: asyncio.Transport,
+        req: "ClientRequest",
+    ) -> None:
+        """Issue a warning if the requested URL has HTTPS scheme."""
+        if req.request_info.url.scheme != "https":
+            return
+
+        asyncio_supports_tls_in_tls = getattr(
+            underlying_transport,
+            "_start_tls_compatible",
+            False,
+        )
+
+        if asyncio_supports_tls_in_tls:
+            return
+
+        warnings.warn(
+            "An HTTPS request is being sent through an HTTPS proxy. "
+            "This support for TLS in TLS is known to be disabled "
+            "in the stdlib asyncio. This is why you'll probably see "
+            "an error in the log below.\n\n"
+            "It is possible to enable it via monkeypatching under "
+            "Python 3.7 or higher. For more details, see:\n"
+            "* https://bugs.python.org/issue37179\n"
+            "* https://github.com/python/cpython/pull/28073\n\n"
+            "You can temporarily patch this as follows:\n"
+            "* https://docs.aiohttp.org/en/stable/client_advanced.html#proxy-support\n"
+            "* https://github.com/aio-libs/aiohttp/discussions/6044\n",
+            RuntimeWarning,
+            source=self,
+            # Why `4`? At least 3 of the calls in the stack originate
+            # from the methods in this class.
+            stacklevel=3,
+        )
+
+    async def _start_tls_connection(
+        self,
+        underlying_transport: asyncio.Transport,
+        req: "ClientRequest",
+        timeout: "ClientTimeout",
+        client_error: Type[Exception] = ClientConnectorError,
+    ) -> Tuple[asyncio.BaseTransport, ResponseHandler]:
+        """Wrap the raw TCP transport with TLS."""
+        tls_proto = self._factory()  # Create a brand new proto for TLS
+
+        # Safety of the `cast()` call here is based on the fact that
+        # internally `_get_ssl_context()` only returns `None` when
+        # `req.is_ssl()` evaluates to `False` which is never gonna happen
+        # in this code path. Of course, it's rather fragile
+        # maintainability-wise but this is to be solved separately.
+        sslcontext = cast(ssl.SSLContext, self._get_ssl_context(req))
+
+        try:
+            async with ceil_timeout(timeout.sock_connect):
+                try:
+                    tls_transport = await self._loop.start_tls(
+                        underlying_transport,
+                        tls_proto,
+                        sslcontext,
+                        server_hostname=req.host,
+                        ssl_handshake_timeout=timeout.total,
+                    )
+                except BaseException:
+                    # We need to close the underlying transport since
+                    # `start_tls()` probably failed before it had a
+                    # chance to do this:
+                    underlying_transport.close()
+                    raise
+        except cert_errors as exc:
+            raise ClientConnectorCertificateError(req.connection_key, exc) from exc
+        except ssl_errors as exc:
+            raise ClientConnectorSSLError(req.connection_key, exc) from exc
+        except OSError as exc:
+            raise client_error(req.connection_key, exc) from exc
+        except TypeError as type_err:
+            # Example cause looks like this:
+            # TypeError: transport <asyncio.sslproto._SSLProtocolTransport
+            # object at 0x7f760615e460> is not supported by start_tls()
+
+            raise ClientConnectionError(
+                "Cannot initialize a TLS-in-TLS connection to host "
+                f"{req.host!s}:{req.port:d} through an underlying connection "
+                f"to an HTTPS proxy {req.proxy!s} ssl:{req.ssl or 'default'} "
+                f"[{type_err!s}]"
+            ) from type_err
+        else:
+            tls_proto.connection_made(
+                tls_transport
+            )  # Kick the state machine of the new TLS protocol
+
+        return tls_transport, tls_proto
+
     async def _create_direct_connection(
         self,
         req: "ClientRequest",
@@ -1028,7 +1122,7 @@ class TCPConnector(BaseConnector):
 
     async def _create_proxy_connection(
         self, req: "ClientRequest", traces: List["Trace"], timeout: "ClientTimeout"
-    ) -> Tuple[asyncio.Transport, ResponseHandler]:
+    ) -> Tuple[asyncio.BaseTransport, ResponseHandler]:
         headers = {}  # type: Dict[str, str]
         if req.proxy_headers is not None:
             headers = req.proxy_headers  # type: ignore[assignment]
@@ -1063,7 +1157,8 @@ class TCPConnector(BaseConnector):
                 proxy_req.headers[hdrs.PROXY_AUTHORIZATION] = auth
 
         if req.is_ssl():
-            sslcontext = self._get_ssl_context(req)
+            self._warn_about_tls_in_tls(transport, req)
+
             # For HTTPS requests over HTTP proxy
             # we must notify proxy to tunnel connection
             # so we send CONNECT command:
@@ -1083,7 +1178,11 @@ class TCPConnector(BaseConnector):
             try:
                 protocol = conn._protocol
                 assert protocol is not None
-                protocol.set_response_params()
+
+                # read_until_eof=True will ensure the connection isn't closed
+                # once the response is received and processed allowing
+                # START_TLS to work on the connection below.
+                protocol.set_response_params(read_until_eof=True)
                 resp = await proxy_resp.start(conn)
             except BaseException:
                 proxy_resp.close()
@@ -1104,21 +1203,19 @@ class TCPConnector(BaseConnector):
                             message=message,
                             headers=resp.headers,
                         )
-                    rawsock = transport.get_extra_info("socket", default=None)
-                    if rawsock is None:
-                        raise RuntimeError("Transport does not expose socket instance")
-                    # Duplicate the socket, so now we can close proxy transport
-                    rawsock = rawsock.dup()
-                finally:
+                except BaseException:
+                    # It shouldn't be closed in `finally` because it's fed to
+                    # `loop.start_tls()` and the docs say not to touch it after
+                    # passing there.
                     transport.close()
+                    raise
 
-                transport, proto = await self._wrap_create_connection(
-                    self._factory,
-                    timeout=timeout,
-                    ssl=sslcontext,
-                    sock=rawsock,
-                    server_hostname=req.host,
+                return await self._start_tls_connection(
+                    # Access the old transport for the last time before it's
+                    # closed and forgotten forever:
+                    transport,
                     req=req,
+                    timeout=timeout,
                 )
             finally:
                 proxy_resp.close()
--- a/CHANGES/5992.feature
+++ b/CHANGES/5992.feature
@@ -1,0 +1,3 @@
+Added support for HTTPS proxies to the extent CPython's
+:py:mod:`asyncio` supports it -- by :user:`bmbouter`,
+:user:`jborean93` and :user:`webknjaz`.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -57,6 +57,7 @@ Boris Feld
 Borys Vorona
 Boyi Chen
 Brett Cannon
+Brian Bouterse
 Brian C. Lane
 Brian Muller
 Bruce Merry
@@ -165,6 +166,7 @@ Jonas Obrist
 Jonathan Wright
 Jonny Tan
 Joongi Kim
+Jordan Borean
 Josep Cugat
 Josh Junon
 Joshu Coats
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -533,9 +533,11 @@ DER with e.g::
 Proxy support
 -------------
 
-aiohttp supports plain HTTP proxies and HTTP proxies that can be upgraded to HTTPS
-via the HTTP CONNECT method. aiohttp does not support proxies that must be
-connected to via ``https://``. To connect, use the *proxy* parameter::
+aiohttp supports plain HTTP proxies and HTTP proxies that can be
+upgraded to HTTPS via the HTTP CONNECT method. aiohttp has a limited
+support for proxies that must be connected to via ``https://`` — see
+the info box below for more details.
+To connect, use the *proxy* parameter::
 
    async with aiohttp.ClientSession() as session:
        async with session.get("http://python.org",
@@ -569,6 +571,33 @@ variables* (all are case insensitive)::
 
 Proxy credentials are given from ``~/.netrc`` file if present (see
 :class:`aiohttp.ClientSession` for more details).
+
+.. attention::
+
+   CPython has introduced the support for TLS in TLS around Python 3.7.
+   But, as of now (Python 3.10), it's disabled for the transports that
+   :py:mod:`asyncio` uses. If the further release of Python (say v3.11)
+   toggles one attribute, it'll *just work™*.
+
+   aiohttp v3.8 and higher is ready for this to happen and has code in
+   place supports TLS-in-TLS, hence sending HTTPS requests over HTTPS
+   proxy tunnels.
+
+   ⚠️ For as long as your Python runtime doesn't declare the support for
+   TLS-in-TLS, please don't file bugs with aiohttp but rather try to
+   help the CPython upstream enable this feature. Meanwhile, if you
+   *really* need this to work, there's a patch that may help you make
+   it happen, include it into your app's code base:
+   https://github.com/aio-libs/aiohttp/discussions/6044#discussioncomment-1432443.
+
+.. important::
+
+   When supplying a custom :py:class:`ssl.SSLContext` instance, bear in
+   mind that it will be used not only to establish a TLS session with
+   the HTTPS endpoint you're hitting but also to establish a TLS tunnel
+   to the HTTPS proxy. To avoid surprises, make sure to set up the trust
+   chain that would recognize TLS certificates used by both the endpoint
+   and the proxy.
 
 .. _aiohttp-persistent-session:
 
--- a/tests/test_client_request.py
+++ b/tests/test_client_request.py
@@ -119,11 +119,6 @@ def test_version_err(make_request: Any) -> None:
         make_request("get", "http://python.org/", version="1.c")
 
 
-def test_https_proxy(make_request: Any) -> None:
-    with pytest.raises(ValueError):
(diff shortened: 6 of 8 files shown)


bmbouter:
Note of the 4 cases (http or https proxy and http or https url being requested) 3 of the 4 work.

http proxy - http url - works
http proxy - https url - works
https proxy - http url - works
https proxy - https url - does not work

bmbouter:
Here are some notes from a discussion with @webknjaz 

* An actual PR fix should go against master which will become aiohttp 4.0. The 3.7.z line very likely won't have another release. Any PR would be merged to master and then possibly backported to 3.8.
* aiohttp can't support this because asyncio itself doesn't support TLS in TLS. See this [bug against python](https://bugs.python.org/issue37179).
* [`loop.starttls()` seems incompatible](https://github.com/aio-libs/aiohttp/issues/4268#issuecomment-732285417) with how aiohttp handshakes TLS.
* it is [claimed](https://github.com/aio-libs/aiohttp/issues/3816#issuecomment-655825192) that uvloop does support TLS in TLS correctly.
* Here's [another reproducer](https://github.com/aio-libs/aiohttp/issues/3816#issue-453181982)

Things for me to try:
* retest this against master
* attempt my reproducer with uvloop
* try the other reproducer to try to learn more

bmbouter:
Here's an interesting post summarizing how httpcore (another asyncio based http client) is persuing resolving this https://github.com/encode/httpcore/issues/254#issuecomment-818433859  It has details about why this doesn't work.

Also here's what I believe is the only remaining fix needed for cPython:  https://github.com/python/cpython/pull/28073

webknjaz:
> The 3.y line very likely won't have another release.

Correction: 3.7.x won't have a release but 3.8 will. But master will be 4.0+ w/o a specific plan currently.

bmbouter:
When I try using `uvloop` as the event loop with my reproducer I get this:

```python-traceback
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 23, in <module>
  File "uvloop/loop.pyx", line 1501, in uvloop.loop.Loop.run_until_complete
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 16, in main
    async with session.get(url, proxy=proxy_url, verify_ssl=False) as resp:
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 1117, in __aenter__
    self._resp = await self._coro
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 520, in _request
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 535, in connect
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 890, in _create_connection
    _, proto = await self._create_proxy_connection(req, traces, timeout)
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 1139, in _create_proxy_connection
    transport, proto = await self._wrap_create_connection(
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 969, in _wrap_create_connection
    return await self._loop.create_connection(*args, **kwargs)  # type: ignore  # noqa
  File "uvloop/loop.pyx", line 2069, in create_connection
  File "uvloop/loop.pyx", line 2064, in uvloop.loop.Loop.create_connection
  File "uvloop/sslproto.pyx", line 517, in uvloop.loop.SSLProtocol._on_handshake_complete
ConnectionResetError
```

jborean93:
I've had a play around with this today and while I haven't found a solution I think I may know why the connection is being reset. This process https://github.com/aio-libs/aiohttp/blob/0fe5b62c552cc0c464580d58cf0be3f9a9296697/aiohttp/connector.py#L1076-L1124 will send the `CONNECT` request to the proxy and then duplicates the socket for the newly wrapped transport. The closing of the original transport on line 1113 seems to bring down the HTTPS context and thus `_wrap_create_connection` raises the `ConnectionResetError`. If I was to comment out the `transport.close()` line and change the `self._wrap_create_connection` to

```python
setattr(transport, "_start_tls_compatible", True)  # for https://github.com/python/cpython/pull/28073
proto = self._factory()
transport = await self._loop.start_tls(transport, proto, sslcontext, server_hostname=req.host)
proto.connection_made(transport)
```

I am able to complete the TLS handle for a HTTPS connection over a HTTPS proxy but once the scope of the function is exited the garbage collector then closes the connection. The whole asyncio protocol and transport mechanism is something I don't have much knowledge on but there seems to be some sort of relationship between the original connection that the `CONNECT` request was sent over and the TLS context which needs to be kept alive. Unfortunately at this point it's beyond my understanding.

bmbouter:
@jborean93 can you post a diff of this? I was applying it, but since it doesn't fully resolve the issue due to GC I wanted to make sure I was getting as far as you got. Thanks!

bmbouter:
To make testing easier I was able to reproduce curl working with secure proxy and aiohttp not working with secure proxy with:

```
# Make your keys
sudo openssl genrsa -out /etc/ssl/key.pem 4096
sudo openssl req -new -x509 -key /etc/ssl/key.pem -out /etc/ssl/cert.pem -days 1826

# Install proxy and run it
pip install proxy.py
proxy --hostname 0.0.0.0 --cert-file /etc/ssl/cert.pem --key-file /etc/ssl/key.pem

# Test it with curl in another tty
curl -v --proxy-insecure --proxy https://127.0.0.1:8899 https://example.com
```

webknjaz:
By the way, `pip install trustme` should give you an easier command for generating the CA certs (it's integrated in our tests and recently gained a CLI too).

bmbouter:
Trustme was easier!

```
# Make your keys
pip install trustme
python -m trustme

# Install proxy and run it
pip install proxy.py
python -m proxy --hostname 0.0.0.0 --cert-file ./server.pem --key-file ./server.key

# Test it with curl in another tty
curl -v --proxy-cacert client.pem --proxy https://127.0.0.1:8899 https://example.com
```

bmbouter:
Interestingly `proxy.py` shows the same "30-ish" second hang going through squid as the secure proxy was doing for me. A few days ago I figured out that if I asked curl to hangup the connection by specifying http 1.0 it would return immediately all the time. So the curl for that is:

`curl -v --http1.0 --proxy-cacert client.pem --proxy https://127.0.0.1:8899 https://example.com`

bmbouter:
I switched my branch to `aio-list/aiohttp:master` and now I'm using this reproducer:

```python
import aiohttp
import asyncio
import ssl

external_url = "https://httpbin.org/get"
proxy_url = "https://localhost:8899/"
client_key = "/home/vagrant/client.pem"

async def main():
    # Setup SSL Fun
    ssl_ctx = ssl.create_default_context(cafile=client_key)

    async with aiohttp.ClientSession(headers={"Cache-Control": "no-cache"}) as session:
        async with session.get(external_url, proxy=proxy_url, ssl=ssl_ctx) as resp:
            print(resp.status)
            print(resp)
            print(await resp.text())

```

For me (with proxy.py) it produces:

```python-traceback
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 946, in _wrap_create_connection
    return await self._loop.create_connection(*args, **kwargs)  # type: ignore[return-value]  # noqa
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 1081, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 1112, in _create_connection_transport
    await waiter
ConnectionResetError

  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 20, in <module>
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 14, in main
    async with session.get(external_url, proxy=proxy_url, ssl=ssl_ctx) as resp:
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 1077, in __aenter__
    self._resp = await self._coro
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 479, in _request
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 508, in connect
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 867, in _create_connection
    _, proto = await self._create_proxy_connection(req, traces, timeout)
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 1115, in _create_proxy_connection
    transport, proto = await self._wrap_create_connection(
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 952, in _wrap_create_connection
    raise client_error(req.connection_key, exc) from exc
aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host httpbin.org:443 ssl:<ssl.SSLContext object at 0x7f0fcabcc4c0> [None]
```

webknjaz:
So with the recent commit (e76ef53) the following repro works for me:
```python
# tls-proxy-repro.py 
import aiohttp
import asyncio


url = "https://httpbin.org/get"

proxy_url = "https://localhost:8899/"

async def main():
            connector=aiohttp.TCPConnector(ssl=False),
            headers={"Cache-Control": "no-cache"},
            trust_env=True,
    ) as session:
        async with session.get(url, proxy=proxy_url) as resp:
            print(resp.status)
            print(resp)
            print(await resp.text())


__name__ == "__main__" and asyncio.run(main())
```

bmbouter:
I tried this patch (e76ef53)) and your reproducer on my dev machine and also on a fresh F34 install (which has Python 3.9.7). It didn't work for me in either environment and yielded the same traceback.

```python-traceback
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 22, in <module>
    __name__ == "__main__" and asyncio.run(main())
  File "/usr/lib64/python3.9/asyncio/runners.py", line 44, in run
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/home/vagrant/devel/aiohttp_proxy_test.py", line 16, in main
    async with session.get(url, proxy=proxy_url) as resp:
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 1077, in __aenter__
    self._resp = await self._coro
  File "/home/vagrant/devel/aiohttp/aiohttp/client.py", line 479, in _request
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 508, in connect
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 867, in _create_connection
    _, proto = await self._create_proxy_connection(req, traces, timeout)
  File "/home/vagrant/devel/aiohttp/aiohttp/connector.py", line 1125, in _create_proxy_connection
    tls_transport.get_protocol().connection_made(
AttributeError: 'NoneType' object has no attribute 'get_protocol'
```

webknjaz:
That's weird. I wonder in what cases `start_tls()` returns `None`. IIRC F34 enforces a more strict TLS setup so maybe if something's not right on the TLS level, it returns `None` while failing to create a new transport. I'm on Gentoo and I don't have this problem. OTOH I applied the patch to the install from PyPI, not to the recent code on master so there's that...

jborean93:
The issue is the `CONNECT` resp is still closing the original connection so the EOF is still being handled causing the failure down the line. @webknjaz you might still have https://github.com/aio-libs/aiohttp/blob/0fe5b62c552cc0c464580d58cf0be3f9a9296697/aiohttp/client_reqrep.py#L846 commented out allowing it to still run for you. If I was to do that then things start working again but with it running the connection is still `FIN` once the `CONNECT` response is processed.

bmbouter:
I did receive the 200 OK response and the proxy shows a payload being returned correctly when I apply this patch to the 3.7 branch and commenting out the `payload.on_eof(self._response_eof)`, but it still errors just later on. I think it's what you're saying @jborean93 that the connection is `FIN` once the `CONNECT` response is processed. Here's what it shows for me:

```python-traceback
python ~/devel/aiohttp_proxy_test.py 
200
<ClientResponse(https://httpbin.org/get) [200 OK]>
<CIMultiDictProxy('Date': 'Mon, 13 Sep 2021 15:19:07 GMT', 'Content-Type': 'application/json', 'Content-Length': '349', 'Connection': 'keep-alive', 'Server': 'gunicorn/19.9.0', 'Access-Control-Allow-Origin': '*', 'Access-Control-Allow-Credentials': 'true')>

{
  "args": {}, 
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "gzip, deflate", 
    "Cache-Control": "no-cache", 
    "Host": "httpbin.org", 
    "User-Agent": "Python/3.9 aiohttp/3.7.4.post0", 
    "X-Amzn-Trace-Id": "Root=1-613f6beb-3790eb756b11edfb70c500a5"
  }, 
  "origin": "174.99.22.188", 
  "url": "https://httpbin.org/get"
}

Fatal error on SSL transport
protocol: <asyncio.sslproto.SSLProtocol object at 0x7ff27bce1430>
transport: <_SelectorSocketTransport closing fd=6>
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 918, in write
    n = self._sock.send(data)
OSError: [Errno 9] Bad file descriptor

  File "/usr/lib64/python3.9/asyncio/sslproto.py", line 684, in _process_write_backlog
    self._transport.write(chunk)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 924, in write
    self._fatal_error(exc, 'Fatal write error on socket transport')
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 719, in _fatal_error
    self._force_close(exc)
  File "/usr/lib64/python3.9/asyncio/selector_events.py", line 731, in _force_close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 746, in call_soon
    self._check_closed()
  File "/usr/lib64/python3.9/asyncio/base_events.py", line 510, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

So what is the right way to solve this? What can be done?

jborean93:
When testing offline changing https://github.com/aio-libs/aiohttp/pull/5992/files#diff-c143bd72e7f6748d879e5a0466a2872e05dc2d0b4a1157e9221702cc3c5516bbR1124 to the below was able to delay the connection being closed.

```python
protocol.set_response_params(read_until_eof=True)
```

This works as it switches the behaviour in the raw HttpPayloadParser to not return a `EmptyStreamReader` but rather a generic `StreamReader`. In the former's case `EOF` is always marked thus processing the `CONNECT` response will bring down the connection whereas for a `StreamReader` it's not marked as EOF and the connection will stay active for the START_TLS step.

aiohttp/connector.py
```diff
@@ -1104,22 +1108,29 @@ async def _create_proxy_connection(
...
-
-                transport, proto = await self._wrap_create_connection(
-                    self._factory,
-                    timeout=timeout,
-                    ssl=sslcontext,
-                    sock=rawsock,
+                    raise
+
+                setattr(
+                    transport, "_start_tls_compatible", True
```
webknjaz: @asvetlov do you think it'd be okay to stick this conditionally somewhere in aiohttp or should we rather recommend the end-users to add this patch `setattr(asyncio.sslproto._SSLProtocolTransport, '_start_tls_compatible', True)` to their apps until the upstream CPython PR is accepted and shipped?

webknjaz: I've decided to recommend keeping this monkey-patching in the user app.

aiohttp/connector.py
```diff
@@ -1028,7 +1028,7 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
 
     async def _create_proxy_connection(
         self, req: "ClientRequest", traces: List["Trace"], timeout: "ClientTimeout"
-    ) -> Tuple[asyncio.Transport, ResponseHandler]:
+    ) -> Tuple[asyncio.BaseTransport, ResponseHandler]:
```
webknjaz, line 1125: (mypy-requested change)

aiohttp/connector.py
```diff
@@ -1063,7 +1063,13 @@ async def _create_proxy_connection(
                 proxy_req.headers[hdrs.PROXY_AUTHORIZATION] = auth
 
         if req.is_ssl():
-            sslcontext = self._get_ssl_context(req)
+            # Safety of the `cast()` call here is based on the fact that
+            # internally `_get_ssl_context()` only returns `None` when
+            # `req.is_ssl()` evaluates to `False` which is never gonna happen
+            # in this code path. Of course, it's rather fragile
+            # maintainability-wise but this is to be solved separately.
+            sslcontext = cast(ssl.SSLContext, self._get_ssl_context(req))
```
webknjaz: (mypy knows that this may return `None` but we know that this won't happen under `if req.is_ssl():`)

aiohttp/client_reqrep.py
```diff
@@ -481,8 +481,6 @@ def update_proxy(
         proxy_auth: Optional[BasicAuth],
         proxy_headers: Optional[LooseHeaders],
     ) -> None:
-        if proxy and not proxy.scheme == "http":
-            raise ValueError("Only http proxies are supported")
```
webknjaz, line 485: I think we need to add a warning here if asyncio doesn't yet support TLS-in-TLS with pointers to how to monkeypatch it.

webknjaz:
I've made the tests green but the PR still needs polishing:
- [x] a change note
- [x] monkeypatching docs
- [x] warning
- [x] fix the PR description
- [x] improve the commit message

webknjaz (approved):

PR #6002, opened 2021-09-14
Add xfailing integration tests against ``proxy.py``

bmbouter:
I was running this with `pytest 'tests/test_proxy_functional.py::test_secure_proxy_http_absolute_path'`

I tried using [the TestCase from proxy.py](https://github.com/abhinavsingh/proxy.py/blob/c09fc8627727fed98dd1bdde8321a6efa36a8fcd/proxy/testing/test_case.py#L22) but I want these as pytest fixtures and I couldn't find a way to call the trustme fixture and set the class attributes I need to [here](https://github.com/abhinavsingh/proxy.py/blob/c09fc8627727fed98dd1bdde8321a6efa36a8fcd/proxy/testing/test_case.py#L30). So instead I'm "doing it like their TestCase" only as a fixture. This probably should go into proxy.py eventually, but here it is for now.

### DIFF
--- a/tests/test_proxy_functional.py
+++ b/tests/test_proxy_functional.py
@@ -5,11 +5,139 @@ import pathlib
 from typing import Any
 from unittest import mock
 
+import proxy
 import pytest
 from yarl import URL
 
 import aiohttp
 from aiohttp import web
+
+ASYNCIO_SUPPORTS_TLS_IN_TLS = hasattr(
+    asyncio.sslproto._SSLProtocolTransport,
+    "_start_tls_compatible",
+)
+
+
+@pytest.fixture
+def secure_proxy_url(monkeypatch, tls_certificate_pem_path):
+    """Return the URL of an instance of a running secure proxy.
+
+    This fixture also spawns that instance and tears it down after the test.
+    """
+    proxypy_args = [
+        "--threadless",  # use asyncio
+        "--num-workers",
+        "1",  # the tests only send one query anyway
+        "--hostname",
+        "127.0.0.1",  # network interface to listen to
+        "--port",
+        0,  # ephemeral port, so that kernel allocates a free one
+        "--cert-file",
+        tls_certificate_pem_path,  # contains both key and cert
+        "--key-file",
+        tls_certificate_pem_path,  # contains both key and cert
+    ]
+
+    class PatchedAccetorPool(proxy.core.acceptor.AcceptorPool):
+        def listen(self):
+            super().listen()
+            self.socket_host, self.socket_port = self.socket.getsockname()[:2]
+
+    monkeypatch.setattr(proxy.proxy, "AcceptorPool", PatchedAccetorPool)
+
+    with proxy.Proxy(input_args=proxypy_args) as proxy_instance:
+        yield URL.build(
+            scheme="https",
+            host=proxy_instance.acceptors.socket_host,
+            port=proxy_instance.acceptors.socket_port,
+        )
+
+
+@pytest.fixture
+def web_server_endpoint_payload():
+    return "Test message"
+
+
+@pytest.fixture(params=("http", "https"))
+def web_server_endpoint_type(request):
+    return request.param
+
+
+@pytest.fixture
+async def web_server_endpoint_url(
+    aiohttp_server,
+    ssl_ctx,
+    web_server_endpoint_payload,
+    web_server_endpoint_type,
+):
+    server_kwargs = (
+        {
+            "ssl": ssl_ctx,
+        }
+        if web_server_endpoint_type == "https"
+        else {}
+    )
+
+    async def handler(*args, **kwargs):
+        return web.Response(text=web_server_endpoint_payload)
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+    server = await aiohttp_server(app, **server_kwargs)
+
+    return URL.build(
+        scheme=web_server_endpoint_type,
+        host=server.host,
+        port=server.port,
+    )
+
+
+@pytest.fixture
+def _pretend_asyncio_supports_tls_in_tls(
+    monkeypatch,
+    web_server_endpoint_type,
+):
+    if web_server_endpoint_type != "https" or ASYNCIO_SUPPORTS_TLS_IN_TLS:
+        return
+
+    # for https://github.com/python/cpython/pull/28073
+    # and https://bugs.python.org/issue37179
+    monkeypatch.setattr(
+        asyncio.sslproto._SSLProtocolTransport,
+        "_start_tls_compatible",
+        True,
+        raising=False,
+    )
+
+
+@pytest.mark.xfail(
+    reason="https://github.com/aio-libs/aiohttp/pull/5992",
+    raises=ValueError,
+)
+@pytest.mark.parametrize("web_server_endpoint_type", ("http", "https"))
+@pytest.mark.usefixtures("_pretend_asyncio_supports_tls_in_tls", "loop")
+async def test_secure_https_proxy_absolute_path(
+    client_ssl_ctx,
+    secure_proxy_url,
+    web_server_endpoint_url,
+    web_server_endpoint_payload,
+) -> None:
+    """Test urls can be requested through a secure proxy."""
+    conn = aiohttp.TCPConnector()
+    sess = aiohttp.ClientSession(connector=conn)
+
+    response = await sess.get(
+        web_server_endpoint_url,
+        proxy=secure_proxy_url,
+        ssl=client_ssl_ctx,  # used for both proxy and endpoint connections
+    )
+
+    assert response.status == 200
+    assert await response.text() == web_server_endpoint_payload
+
+    response.close()
+    await sess.close()
+    await conn.close()
 
 
 @pytest.fixture
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,8 @@ jobs:
         path: ${{ steps.pip-cache.outputs.dir }}
         restore-keys: |
             pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
+    - name: Upgrade wheel  # Needed for proxy.py install not to explode
+      run: pip install -U wheel
     - name: Cythonize
       if: ${{ matrix.no-extensions == '' }}
       run: |
--- a/CHANGES/6002.misc
+++ b/CHANGES/6002.misc
@@ -1,0 +1,2 @@
+Implemented end-to-end testing of sending HTTP and HTTPS requests
+via ``proxy.py``.
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -163,6 +163,8 @@ pluggy==0.13.1
     #   pytest
 pre-commit==2.15.0
     # via -r requirements/lint.txt
+proxy.py==2.3.1
+    # via -r requirements/test.txt
 py==1.10.0
     # via
     #   -r requirements/lint.txt
@@ -277,6 +279,7 @@ typing-extensions==3.7.4.3
     #   -r requirements/lint.txt
     #   async-timeout
     #   mypy
+    #   proxy.py
 uritemplate==3.0.1
     # via gidgethub
 urllib3==1.26.5
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -6,6 +6,7 @@ cryptography==3.3.1; platform_machine!="i686" and python_version<"3.9" # no 32-b
 freezegun==1.1.0
 mypy==0.910; implementation_name=="cpython"
 mypy-extensions==0.4.3; implementation_name=="cpython"
+proxy.py==2.3.1
 pytest==6.2.2
 pytest-cov==2.12.1
 pytest-mock==3.6.1


setup.cfg
```diff
@@ -53,6 +53,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
+    ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning
```
webknjaz: Note for the future: This will probably be fixed in #6004. We shouldn't include it in the final patch.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+    input_args = [
+        "--num-workers",
+        "1",
+        "--threadless",
+        "--hostname",
+        "0.0.0.0",
+        "--port",
+        str(port),
+        "--cert-file",
+        trustme_certs[0],
```
webknjaz: Don't reference magic indexes in a tuple. Instead, they should be in named vars. Better yet — in fixtures for each with names relaying their purpose.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+import trustme
 from yarl import URL
 
 import aiohttp
 from aiohttp import web
 
 
+@pytest.fixture
+def trustme_certs(tmp_path):
+    ca = trustme.CA()
```
webknjaz: Why didn't you reuse the trustme fixtures that already exist? Do these tests *need* an additional CA? Do you expect to have certificates for the proxy and the web server from different CA?

I think the simplest test could just rely on existing fixtures.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+        "--hostname",
+        "0.0.0.0",
+        "--port",
+        str(port),
+        "--cert-file",
+        trustme_certs[0],
+        "--key-file",
+        trustme_certs[1],
+    ]
+    proxy_url = "http://0.0.0.0:" + str(port)
```
webknjaz: `0.0.0.0` is a special address with undefined semantics in terms of routing. Don't use it for anything except for binding to network interfaces. Always use `127.0.0.1` in URLs. While some clients under GNU/Linux may apply hacks to route the traffic locally, it'll fail under Windows, for example.
Also, just use an f-string to add a port.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+    with proxy.Proxy(input_args=input_args) as proxy_instance:
```
webknjaz: 1) If you don't use `proxy_instance`, no need to declare it
2) Does `proxy_instance` have any attribute exposing its real port? It'd be useful for ephemeral port binds

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+def secure_proxy_url(trustme_certs):
+    port = proxy.common.utils.get_available_port()
+    input_args = [
+        "--num-workers",
+        "1",
+        "--threadless",
+        "--hostname",
+        "0.0.0.0",
+        "--port",
+        str(port),
```
webknjaz: Is it possible to just bind to an ephemeral port and let the kernel allocate it atomically?
```suggestion
        "0",
```
(this would require extracting the actual port from the proxy later)

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+        "--key-file",
+        trustme_certs[1],
+    ]
+    proxy_url = "http://0.0.0.0:" + str(port)
+    with proxy.Proxy(input_args=input_args) as proxy_instance:
+        yield proxy_url
+
+
+async def test_secure_proxy_http_absolute_path(secure_proxy_url, get_request) -> None:
+    url = "https://httpbin.org/get"
```
webknjaz: It's okay for a PoC but the actual tests would need to test against a local server as the regular functional client tests do.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+    await get_request(url=url, proxy=secure_proxy_url, ssl=False)
```
webknjaz: We'll need to supply an actual SSLContext, not `False`.

This should work if you'll rely on the same CA from the existing fixtures:
```suggestion
    await get_request(url=url, proxy=secure_proxy_url, ssl=client_ssl_ctx)
```

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+        "1",
+        "--threadless",
+        "--hostname",
+        "0.0.0.0",
+        "--port",
+        str(port),
+        "--cert-file",
+        trustme_certs[0],
+        "--key-file",
+        trustme_certs[1],
```
webknjaz: It seems like you could've just used
```suggestion
        tls_certificate_pem_path,
```

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+    ca.issue_cert("localhost")
+    ca_cert_path = tmp_path / "ca.pem"
+    ca.cert_pem.write_to_path(ca_cert_path)
+    ca_key_path = tmp_path / "server.key"
+    ca.private_key_pem.write_to_path(ca_key_path)
+    return (str(ca_cert_path), str(ca_key_path))
+
+
+@pytest.fixture
+def secure_proxy_url(trustme_certs):
```
webknjaz: Please add a PEP 257 compatible docstrings. Besides the normal usefulness of docstrings, it'll also be displayed in the list that `pytest --fixtures` outputs.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,52 @@
...
+    port = proxy.common.utils.get_available_port()
```
webknjaz: Even if you don't end up relying on an ephemeral port, you could just use the `aiohttp_unused_port` fixture — no need to use a helper from `proxy.py` because it's not a public API (I suppose) and may break over time.

setup.cfg
bmbouter: I'll remove this when I undraft it.

tests/test_proxy_functional.py
bmbouter: I'm not familiar with the existing fixtures. I will look at replacing mine with those now.

tests/test_proxy_functional.py
bmbouter: I'll see if I can find one. I looked briefly but didn't see it. I'll look again.

tests/test_proxy_functional.py
bmbouter: Got a link to a place in the tests that sets one of those up with a fixture?

tests/test_proxy_functional.py
bmbouter: Agreed. This is what I plan to fix now.

setup.py
```diff
@@ -89,6 +89,7 @@ def read(f):
     "pytest-timeout",
```
bmbouter: @webknjaz  If `aiohttp` already has tests that use `trustme` why isn't it listed as ` tests_require`?

webknjaz: `tests_require` is an obsolete setuptools concept and should be eliminated.

webknjaz: In fact, it's best to put test deps in requirements files — it's their use-case to describe venvs (like the one for testing)

tests/test_proxy_functional.py
webknjaz: They are in conftest and are also listed @ `pytest --collect-only --fixtures`

tests/test_proxy_functional.py
webknjaz: https://github.com/aio-libs/aiohttp/blob/c8e2e2c/tests/test_client_functional.py#L316

tests/test_proxy_functional.py
bmbouter: I looked into this again, but I didn't see a way to do it. When I interactively debug [here](https://github.com/abhinavsingh/proxy.py/blob/develop/proxy/core/acceptor/pool.py#L110) I can see the port, but when I access the `proxy_instance.acceptors.socket` it's closed...  Unless you have a pointer I'm going to leave it at `proxy.common.utils.get_available_port()` which I also don't like, but it's what proxy.py's code itself also does. Any ideas are welcome.

The only other way I know is to call to netstat and parse the response, but that seems unreliable.

aiohttp/client_reqrep.py
```diff
@@ -547,8 +547,8 @@ def update_proxy(
         proxy_auth: Optional[BasicAuth],
         proxy_headers: Optional[LooseHeaders],
     ) -> None:
-        if proxy and not proxy.scheme == "http":
-            raise ValueError("Only http proxies are supported")
+        # if proxy and not proxy.scheme == "http":
+        #     raise ValueError("Only http proxies are supported")
```
webknjaz: This will also need to be reverted before merge.

setup.py
```diff
@@ -89,6 +89,7 @@ def read(f):
...
     "async-generator",
     "pytest-xdist",
+    "proxy.py",
```
webknjaz: Don't forget to add this to https://github.com/aio-libs/aiohttp/blob/master/requirements/test.txt.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,47 @@
...
+    proxy_url = f"https://{address}:{port}"
+    with proxy.Proxy(input_args=input_args):
+        yield proxy_url
+
+
+async def test_secure_proxy_http_absolute_path(secure_proxy_url, get_request) -> None:
+    """Test HTTP urls can be requested through a secure proxy."""
+    import pdb
+
+    pdb.set_trace()
```
webknjaz: don't forget to drop this. Also, starting Python 3.7 it's easier to just do
```suggestion
    breakpoint()
```

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,47 @@
...
+        "--key-file",
+        tls_certificate_private_key_path,
+        "--threadless",
+    ]
+    proxy_url = f"https://{address}:{port}"
+    with proxy.Proxy(input_args=input_args):
+        yield proxy_url
+
+
+async def test_secure_proxy_http_absolute_path(secure_proxy_url, get_request) -> None:
```
webknjaz: If you can, you could try sticking both HTTP and HTTPS into a single test. Just use `@pytest.mark.parametrize` for this.

tests/conftest.py
```diff
@@ -75,6 +75,12 @@ def tls_certificate_pem_path(tls_certificate):
         yield cert_pem
 
 
+@pytest.fixture
+def tls_certificate_private_key_path(tls_certificate_authority):
```
webknjaz: Could you move this closer to `tls_ca_certificate_pem_path`? And maybe use a fixture name that is similar. Like `tls_ca_private_key_pem_path`.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,47 @@
...
+        "127.0.0.1",
+        "--port",
+        port,
+        "--cert-file",
+        tls_ca_certificate_pem_path,
+        "--key-file",
+        tls_certificate_private_key_path,
+        "--threadless",
+    ]
+    proxy_url = f"https://{address}:{port}"
```
webknjaz: Since we have a hard dependency on `yarl`, maybe wrap this with `yarl.URL()`?

tests/test_proxy_functional.py
webknjaz: Well, at least use our own fixture for this.

tests/test_proxy_functional.py
```diff
@@ -3,13 +3,47 @@
...
+    port = str(proxy.common.utils.get_available_port())
+    address = "127.0.0.1"
+    input_args = [
+        "--num-workers",
+        "1",
+        "--threadless",
+        "--hostname",
+        "127.0.0.1",
+        "--port",
+        port,
```
webknjaz: Per earlier discussion:
```suggestion
def secure_proxy_url(aiohttp_unused_port, tls_ca_certificate_pem_path, tls_certificate_private_key_path):
    """Return the an instance of a running secure proxy."""
    address = "127.0.0.1"
    input_args = [
        "--num-workers",
        "1",
        "--threadless",
        "--hostname",
        "127.0.0.1",
        "--port",
        str(aiohttp_unused_port),
```

tests/test_proxy_functional.py
webknjaz:
```suggestion
    proxy_url = yarl.URL.build(scheme="https", host=address, port=aiohttp_unused_port)
```

aiohttp/client_reqrep.py
bmbouter: Agreed. I think my next push will remove it after I can see that even the http endpoints through https work. That way the xfail can be confident it's due to actual breakage and not this check failing the test.

setup.cfg
bmbouter: I can rebase my PR after https://github.com/aio-libs/aiohttp/pull/6004/files merges.

setup.py
bmbouter: I'll put it in there, but should I just remove this section altogether?

setup.py
bmbouter: I'll put it in there, but should I just remove this section altogether?

tests/test_proxy_functional.py
bmbouter: Nice! I'm removing it now.

tests/test_proxy_functional.py
bmbouter: Will do.

setup.py
webknjaz: Looks like you commented here and deleted those comments. Let's not do any section removals or packaging refactoring in this PR.

setup.py
webknjaz: Let's not do any section removals or packaging refactoring in this PR. It may be reasonable to do this separately, though.
Ref: https://github.com/aio-libs/aiohttp/pull/3518#issuecomment-453128368.

tests/test_proxy_functional.py
webknjaz: Wait until you hear about the `$PYTHONBREAKPOINT` env var :)

tests/test_proxy_functional.py
```diff
@@ -1,15 +1,84 @@
...
+
+    # url = URL.build(scheme=protocol, host=server.host, port=server.port)
+    url = URL.build(scheme="http", host=server.host, port=server.port)
+    response = await get_request(url=url, proxy=secure_proxy_url, ssl=False)
+    assert response.status == 200
+
+
+# @pytest.mark.xfail
+# @pytest.mark.parametrize("protocol", ["http", "http"])
+async def test_secure_proxy_https_absolute_path(
```
bmbouter, line 61: These are still separate because the ssl contexts are different, but I plan to deduplicate.

setup.cfg
bmbouter: I rebased this and removed this change from my PR.

tests/test_proxy_functional.py
```diff
@@ -1,14 +1,137 @@
...
+#         "--port",
+#         unused_port,
+#         "--cert-file",
+#         tls_ca_certificate_pem_path,
+#         "--key-file",
+#         tls_ca_private_key_pem_path,
+#     ]
+#     proxy_url = URL.build(scheme="https", host=address, port=unused_port)
+#     with proxy.Proxy(input_args=input_args):
+#         yield proxy_url
```
bmbouter: This is what I wanted to use originally, and what I think we ultimately should use. I made the section below as an alternate version that very closely follows what trustme does here https://github.com/python-trio/trustme/blob/master/trustme/_cli.py#L50-L86

Ultimately I'm not sure the version of the fixture below actually improved it any so I'm leaving both versions for @webknjaz to see.

tests/test_proxy_functional.py
```diff
@@ -1,14 +1,137 @@
...
+                "--key-file",
+                server_key,
+            ]
+            proxy_url = URL.build(scheme="https", host=address, port=unused_port)
+            with proxy.Proxy(input_args=input_args):
+                yield proxy_url
+
+
+# @pytest.mark.xfail
+async def test_secure_proxy_http_absolute_path(
```
bmbouter: This test is fully working and is more-or-less in its final form. It just needs the xfail uncomented from above, and possibly paramterize'd on the protocol to be `http` versus `https` if the tests can be consolidated.

tests/test_proxy_functional.py
```diff
@@ -1,14 +1,137 @@
...
+    response_body = "Test message"
+
+    async def handler(*args, **kwargs):
+        return web.Response(text=response_body)
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+    server = await aiohttp_server(app)
+
+    conn = aiohttp.TCPConnector(loop=loop)
```
bmbouter: I tried finding another fixture to use instead of instantiating the `TCPConnector` and `ClientSession` here, but those all configured the SSL context more than I needed when making the `get()` so I ended up writing this.

tests/test_proxy_functional.py
bmbouter: This test is also passing, but when I try to switch it against the local server it starts failing. This is the part I really struggled with.

What I noticed is that if I switch to use the local aiohttp server as the endpoint AND I used `secure_proxy_url = "https://127.0.0.1:8899"` and configured the TLS to use certs corresponding with an externally started proxy.py with trustme certs it also works! So basically I can use proxy.py started by the tests, or an endpoint that isn't aiohttp.Server, but not both. :( Most of the code here is me switching back and forth between these approaches.

bmbouter:
I was running these with:

`pytest --log-level=INFO 'tests/test_proxy_functional.py::test_secure_proxy_http_absolute_path'`
`pytest --log-level=INFO 'tests/test_proxy_functional.py::test_secure_proxy_https_absolute_path'`

bmbouter:
@webknjaz this is as far as I could take these tests. I spent a long time messing with it. I'm hoping you can take it from here. Also FYI @dkliban.

webknjaz (approved):

PR #6003, opened 2021-09-14
Quick fix: Ignore python internal deprecation warning

Dreamsorcerer:
A change to deprecation warnings in the last python 3.9 release has resulted in warnings appearing in some internal asyncio code. We'll just ignore them for now (think they are fixed in the next release).

For some bizarre reason, they only seem to happen on Windows though...


Dreamsorcerer:
Tests seem to be passing fine on master actually... Just need to do this on the 3.8 branch.

PR #6004, opened 2021-09-14
Quick fix: Ignore python internal deprecation warning

Dreamsorcerer:
A change to deprecation warnings in the last python 3.9 release has resulted in warnings appearing in some internal asyncio code. We'll just ignore them for now (think they are fixed in the next release).

For some bizarre reason, they only seem to happen on Windows though...

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,6 +53,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
+    ignore:The loop argument is deprecated:DeprecationWarning:asyncio.base_events
 junit_suite_name = aiohttp_test_suite
 norecursedirs = dist docs build .tox .eggs
 minversion = 3.8.2
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,7 +53,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:The loop argument is deprecated:DeprecationWarning:asyncio.base_events
+    ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning:asyncio.base_events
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
Not such a simple fix, as there are tests which assert the number of DeprecationWarnings...

setup.cfg
```diff
@@ -53,6 +53,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
+    ignore:The loop argument is deprecated:DeprecationWarning:asyncio.base_events
```
webknjaz: Let's use the full message
```suggestion
    ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning:asyncio.base_events
```

webknjaz:
> For some bizarre reason, they only seem to happen on Windows though...

In fact, this was happening on F34 as well so @bmbouter temporarily ignored this in #6002.

Dreamsorcerer:
Weird that it doesn't happen in CI unless it's Windows.

webknjaz:
@bmbouter what's your CPython version, again?

bmbouter:
> @bmbouter what's your CPython version, again?

Python 3.9.7

asvetlov (approved):

Dreamsorcerer:
Alternative (because this still fails some tests): #6006

setup.cfg
```diff
@@ -53,6 +53,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
+    ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning:asyncio.base_events
```
webknjaz, line 57: Could you add a comment with the explanation right above this line? Maybe with links too. And then, go ahead and merge it.

Dreamsorcerer: Is it worth merging this when it doesn't fix all the tests? Or should we just merge #6006 instead? Or both?

I'll probably just open an issue though to remind myself to revert the changes in a few weeks.

webknjaz: I think if it improves things at least a little, we should merge it. And yes, we could revert later. Plus I'd combine with #6006 too.

Dreamsorcerer: Right, you'll need to tweak the required status to get them merged.

PR #6006, opened 2021-09-14
Use python 3.9.6

Dreamsorcerer:


webknjaz:
I don't think this is helpful because the problem happens on the contributor machines. It's best to work on an actual fix.

Dreamsorcerer:
The actual fix is to use Python 3.9.8, scheduled for release November 1st.

I can see it has been fixed upstream:
https://github.com/python/cpython/commit/c967bd523caabb05bf5988449487d7c1717f3ac6

webknjaz:
Ah, cool. But still, we could ignore this for as long as it's scoped to stdlib.

webknjaz:
Also, I want this not to appear in the envs of the contributors. Putting a band-aid on the CI level doesn't really facilitate this.

Dreamsorcerer:
Yep, but I don't have a fix for the failing tests, and am unlikely to spend time figuring something out within the next 6 weeks before it fixes itself.

webknjaz (approved):

asvetlov:
We decided to ignore DeprecationWarning instead

Issue #2934, opened 2018-04-12
Assert in _drain_helper when sending websocket message


willmcgugan:
## Long story short

We see bursts of the following exception on our websocket server.

```python
Traceback (most recent call last):
  File "", line 378, in _on_packet
    await self.dispatch(packet)
  File "", line 67, in dispatch
    await method(self, packet)
  File "", line 550, in handle_request_send
    packet.data
  File "", line 216, in route
    packets.Route(route_no, data)
  File ", line 289, in send
    if not await self.send_bytes(packet_bytes):
  File "", line 297, in send_bytes
    await self.ws.send_bytes(packet_bytes)
  File "lib/python3.6/site-packages/aiohttp/web_ws.py", line 263, in 
    await self._writer.send(data, binary=True, compress=compress)
  File "lib/python3.6/site-packages/aiohttp/http_websocket.py", line 607, 
    return await self._send_frame(message, WSMsgType.BINARY, compress)
  File "lib/python3.6/site-packages/aiohttp/http_websocket.py", line 588, 
    await self.protocol._drain_helper()
  File "lib/python3.6/asyncio/streams.py", line 214, in _drain_helper
    assert waiter is None or waiter.cancelled()
AssertionError
```
(paths censored at client's request)

I'm guessing its a flow control issue? We're essentially copying data from one websocket to another and rely on back pressure to keep things running.

I can't reproduce it easily, but happy to investigate if it would help.

## Your environment

A websocket server running aiohttp==3.1.0 on Linux

dangusev:
Let's bump this topic. I faced the same problem and found some related issues. 
Long story short, `_drain_helper` method is not coroutine-safe and may fail if it has been called from different coroutines. The easiest fix is to wrap it with a lock. See more here:
https://github.com/aaugustin/websockets/issues/16
https://bugs.python.org/issue29930
https://github.com/Polyconseil/aioamqp/pull/137
Hope it will help to resolve the issue.

crusaderky:
I'm getting exactly the same on a websocket *client* using aiohttp 3.6.1.
```
    await ws.send_bytes(payload)
  File "site-packages/aiohttp/client_ws.py", line 155, in send_bytes
    await self._writer.send(data, binary=True, compress=compress)
  File "site-packages/aiohttp/http_websocket.py", line 644, in send
    await self._send_frame(message, WSMsgType.BINARY, compress)
  File "site-packages/aiohttp/http_websocket.py", line 623, in _send_frame
[...]

PR #6028, opened 2021-10-01
Modify _drain_helper() to handle parallel calls without race-condition

Traktormaster:
## What do these changes do?

These changes fix a bug/race-condition on the BaseProtocol's drain-waiter.

There are no API or behavioral changes. This should exclusively handle the case of the race-condition when multiple coroutines produce messages to a websocket for example.

## Related issue number

This fix is aimed at the #2934 issue directly.

## Checklist

- [ ] Documentation reflects the changes - Not applicable AFAIK

### DIFF
--- a/CHANGES/2934.bugfix
+++ b/CHANGES/2934.bugfix
@@ -1,0 +1,1 @@
+Modify _drain_helper() to handle parallel calls of _send_frame() without race-condition.
--- a/aiohttp/base_protocol.py
+++ b/aiohttp/base_protocol.py
@@ -81,7 +81,7 @@ class BaseProtocol(asyncio.Protocol):
         if not self._paused:
             return
         waiter = self._drain_waiter
-        assert waiter is None or waiter.cancelled()
-        waiter = self._loop.create_future()
-        self._drain_waiter = waiter
-        await waiter
+        if waiter is None:
+            waiter = self._loop.create_future()
+            self._drain_waiter = waiter
+        await asyncio.shield(waiter)
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -225,6 +225,7 @@ Navid Sheikhol
 Nicolas Braem
 Nikolay Kim
 Nikolay Novik
+Nándor Mátravölgyi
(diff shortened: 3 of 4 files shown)


aiohttp/base_protocol.py
```diff
@@ -81,7 +81,7 @@ async def _drain_helper(self) -> None:
...
             return
         waiter = self._drain_waiter
-        assert waiter is None or waiter.cancelled()
-        waiter = self._loop.create_future()
-        self._drain_waiter = waiter
-        await waiter
+        if waiter is None:
+            waiter = self._loop.create_future()
+            self._drain_waiter = waiter
+        await asyncio.shield(waiter)
```
asvetlov, line 87: Why is `shield` required?
If the connection is closed by another task -- the current task should get the exception as well.

asvetlov (changes requested):

aiohttp/base_protocol.py
Traktormaster: The exception handling is not affected by shield at all. It only protects the inner future from cancellation. https://docs.python.org/3/library/asyncio-task.html#shielding-from-cancellation

The problem with the implementation was that it was not safe to be called from multiple coroutines. The issue could happen when the protocol was in paused state and multiple coroutines would send larger messages then the WebSocketWriter's limit. The `assert waiter is None or waiter.cancelled()` would fail in that case.

The shield is a solution for this because it provides independency of all coroutines that call send_frame by shielding the inner future from cancellation. Each "worker" coroutine can cancel itself with a timeout for example, without affecting the other waiters or the drain-helper future. Exception or result can only be set for the future from within the drain helper, while all coroutines awaiting on it will get the result.

CHANGES/2934.bugfix
```diff
@@ -0,0 +1 @@
+Modify _drain_helper() to handle parallel calls of _send_frame() without race-condition.
```
asvetlov:
```suggestion
Modify _drain_helper() to handle concurrrent `await resp.write(...)` or `ws.send_json(...)` calls without race-condition.
```

Issue #5281, opened 2020-11-23
AttributeError: object has no attribute 'decode' for PAYLOAD_REGISTRY class


folt:
🐞 **Describe the bug**
While working with `PAYLOAD_REGISTRY` I came across a problem. After examining the information, I found an error. The fact is that the Response class implements a property that calls the [self._body.decode](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_response.py#L587) method expecting a string there. But when using `PAYLOAD_REGISTRY` (I used `JsonPayload`) in `self._body` there is an instance of the payload class where there is no implementation of the `decode` method.

💡 **To Reproduce**
```
from typing import Mapping, Optional
from types import MappingProxyType
from http import HTTPStatus

from aiohttp import web
from aiohttp import payload
from aiohttp import web_request
from aiohttp import web_middlewares
from aiohttp import PAYLOAD_REGISTRY
from aiohttp import web_exceptions


def format_http_error(http_error_cls, message: Optional[str] = None,
                      ) -> web_exceptions.HTTPException:
    status = HTTPStatus(http_error_cls.status_code)

    error = {
        'code': status.name.lower(),
        'message': message or status.description
    }

    return http_error_cls(body={'error': error})


async def handle():
    raise web_exceptions.HTTPNotFound()


@web_middlewares.middleware
async def error_middleware(request: web_request.Request, handler):
    try:
        return await handler(request)
    except web_exceptions.HTTPException as err:
        err = format_http_error(err.__class__, err.text)
        raise err


app = web.Application(
    middlewares=[error_middleware]
)

PAYLOAD_REGISTRY.register(payload.JsonPayload, (Mapping, MappingProxyType))

app.add_routes([web.get('/', handle)])

if __name__ == '__main__':
    web.run_app(app)
```

📋 **Logs/tracebacks**
```
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/folt/Documents/Work/geodb/.venv/lib/python3.8/site-packages/aiohttp/web_protocol.py", line 422, in _handle_request
    resp = await self._request_handler(request)
  File "/Users/folt/Documents/Work/geodb/.venv/lib/python3.8/site-packages/aiohttp/web_app.py", line 499, in _handle
    resp = await handler(request)
  File "/Users/folt/Documents/Work/geodb/.venv/lib/python3.8/site-packages/aiohttp/web_middlewares.py", line 118, in impl
    return await handler(request)
[...]

PR #6086, opened 2021-10-15
Fix AttributeError: object has no attribute 'decode' for PAYLOAD_REGISTRY class #5281

folt:
Implement the decode method in the JsonPayload class #5281
Added test for JsonPayload

## Checklist


Dreamsorcerer:
#5281 appears to be caused by a separate regression in 3.7, doesn't look like this is needed. The test case also seems to pass for me without the other changes in this PR.

PR #6119, opened 2021-10-23
Fix MacOS tests

asvetlov:
I see trustme/cryptography initialization problems on MacOS for Python 3.10, the PR disables such tests.


asvetlov:
Backported by 4f26ddae

PR #6129, opened 2021-10-24
Add a `base_url` option to `ClientSession`

### LINKED ISSUE #6013: Add a base_url option to ClientSession like httpx does

Yes, I would like to have the option to declare a `base_url` to a ClientSession, such that every request use the same `base_url` for that session.

Instead of writing:

```python
    async with session.get('http://python.org') as response:
        print(await response.text())
    async with session.get('http://python.org/foobar') as response:
        print(await response.text())
   async with session.get('http://python.org/foo/foo/bar') as response:
        print(await response.text())
```

Write:

```python
async with aiohttp.ClientSession(base_url='http://python.org') as session:
    async with session.get('/') as response:
        print(await response.text())
    async with session.get('/foobar') as response:
        print(await response.text())
   async with session.get('/foo/foo/bar') as response:
        print(await response.text())
```

Add an `base_url` parameter to ClientSession, like httpx does.

An alternative is having a constant `BASE = 'http://python.org'` and then say:

```python
async with session.get(f'{BASE}/foobar') as response:
        print(await response.text())
```

### Related component

Client

### Additional context

_No response_

### Code of Conduct

derlih:
## What do these changes do?

Add `base_url` to `ClientSession` constructor that makes several requests to the same server look simpler.

New optional parameter in `ClientSession` constructor. If not set, the behavior is not changed.

## Related issue number

Fixes #6013 

## Checklist


tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
     session = ClientSession(requote_redirect_url=False)
     assert not session.requote_redirect_url
     await session.close()
+
+
+@pytest.mark.parametrize(
+    "base_url,url",
```
webknjaz:
```suggestion
    ("base_url", "url"),
```

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+    ],
+)
+async def test_build_url(base_url, url) -> None:
+    session = ClientSession(base_url)
+    assert session._build_url(url) == URL("http://example.com/test")
+
+
+async def test_build_url_raises() -> None:
+    session = ClientSession("http://example.com")
+    with pytest.raises(ValueError):
```
webknjaz: Always use `match=` in `raises()` checks.

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+    "base_url,url",
+    [
+        (None, "http://example.com/test"),
+        ("http://example.com", "/test"),
+        (URL("http://example.com"), "/test"),
+        ("http://example.com/", "/test"),
+        (URL("http://example.com/"), "/test"),
+    ],
+)
+async def test_build_url(base_url, url) -> None:
```
webknjaz: Each test needs a PEP 257 compliant docstring with an explanation of what's going on.

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+async def test_build_url_raises() -> None:
+    session = ClientSession("http://example.com")
+    with pytest.raises(ValueError):
+        session._build_url("test")
+
+
+async def test_request_uses_base_url_when_url_is_str() -> None:
+    request_class = mock.MagicMock()
+    session = ClientSession("http://example.com", request_class=request_class)
+    with contextlib.suppress(Exception):
```
webknjaz: Why not be more precise here?

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+
+
+async def test_build_url_raises() -> None:
+    session = ClientSession("http://example.com")
+    with pytest.raises(ValueError):
+        session._build_url("test")
+
+
+async def test_request_uses_base_url_when_url_is_str() -> None:
+    request_class = mock.MagicMock()
```
webknjaz: Is it possible to avoid mocking stuff?

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+    session = ClientSession("http://example.com", request_class=request_class)
+    with contextlib.suppress(Exception):
+        await session.get("/test")
+
+    args, _ = request_class.call_args
```
webknjaz: Why not use `called_with_args()`?

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+    [
+        (None, "http://example.com/test"),
+        ("http://example.com", "/test"),
+        (URL("http://example.com"), "/test"),
+        ("http://example.com/", "/test"),
+        (URL("http://example.com/"), "/test"),
+    ],
+)
+async def test_build_url(base_url, url) -> None:
+    session = ClientSession(base_url)
```
webknjaz: This session is unclosed. Each session must be closed to prevent resource warnings in tests.

tests/test_client_session.py
derlih: I want to make sure that this mock will be called with expected url

tests/test_client_session.py
derlih: What do you mean?

tests/test_client_session.py
derlih: I need to check only the url param. The others I don't care.
For `called_with_args` you have to specify all of them.

tests/test_client_session.py
derlih: Done

tests/test_client_session.py
derlih: Is it a new rule in aiohttp codebase?

tests/test_client_session.py
derlih: Done

tests/test_client_session.py
derlih: Fixed

tests/test_client_session.py
webknjaz: Yeah, you could specify mocks for those IIRC.

tests/test_client_session.py
webknjaz: `Exception` seems too broad. What sort of error do you expect really?

tests/test_client_session.py
webknjaz: Sort of. It used to be like this but then because of Azure test reports, docstrings became comments and now it is possible again.
https://hynek.me/articles/document-your-tests/

CHANGES/6013.feature
```diff
@@ -0,0 +1 @@
+Add `base_url` parameter to ClientSession constructor
```
webknjaz: It's RST, not Markdown. Also, https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme has instructions on composing the change notes properly.
```suggestion
Added ``base_url`` parameter to the initializer of :class:`ClientSession`.
```

(`__new__()` is a constructor, `__init__()` is an initializer, it accepts a pre-created object as `self`)

aiohttp/client.py
```diff
@@ -304,6 +311,14 @@ def request(
         """Perform HTTP request."""
         return _RequestContextManager(self._request(method, url, **kwargs))
 
+    def _build_url(self, url: str) -> URL:
+        if self._base_url is None:
+            return URL(url)
+        elif url.startswith("/"):
+            return self._base_url / url[1:]
+        else:
+            raise ValueError("url must start with /")
```
webknjaz: I think it'd be helpful to reference user-facing arg names. They don't pass a `url` param. They pass `base_url`.

docs/client_reference.rst
```diff
@@ -56,6 +57,11 @@ The client session supports the context manager protocol for self closing.
    The class for creating client sessions and making requests.
 
 
+   :param base_url: Base part of the URL (optional)
+      If set it allows to skip the base part in request calls.
+
+      .. versionadded:: 3.8
```
webknjaz, line 63: Since we're targeting 3.8, I've added the corresponding milestone so @asvetlov notices this.

tests/test_client_session.py
```diff
@@ -706,3 +706,46 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+        ("http://example.com/", "/test"),
+        (URL("http://example.com/"), "/test"),
+    ],
+)
+async def test_build_url(create_session, base_url, url) -> None:
+    session = await create_session(base_url)
+    assert session._build_url(url) == URL("http://example.com/test")
+
+
+async def test_build_url_raises(create_session) -> None:
```
webknjaz: Mind adding the docstrings after all?

tests/test_client_session.py
webknjaz: The reason I dislike the current approach is that having a check to "extract an element from a list and compare it with some value" is semantically ambiguous. It doesn't *mean* anything. While `called_with_args()` communicates the intent of the test better.

aiohttp/client.py
asvetlov: Technically the argument of `session.request()` or `session.get()` is called `url`.
I think the error message is correct here.

aiohttp/client.py
```diff
@@ -304,6 +311,14 @@ def request(
         """Perform HTTP request."""
         return _RequestContextManager(self._request(method, url, **kwargs))
 
+    def _build_url(self, url: str) -> URL:
+        if self._base_url is None:
+            return URL(url)
+        elif url.startswith("/"):
```
asvetlov: Why do you merge an url started with '/' instead of URL path replacement?

As I see in [URL merging algorithm](https://datatracker.ietf.org/doc/html/rfc3986#section-5.2.2) we should perform something like `merge(base_url, url)`. It means if `url` doesn't contain authority and started with `/` the `base_url` path is replaced with `url`.

Does it sound reasonable? Any reason for non-RFC design?

aiohttp/client.py
webknjaz: Ah okay, I got confused then.

tests/test_client_session.py
asvetlov: > Each test needs a PEP 257 compliant docstring

Sorry, I disagree. Docstrings *may* be used but not enforced.
Otherwise, somebody (maybe you) should iterate over all existing tests and add docstrings. This is a reasonable pre-requirement for the enforcement. I personally have no spare time for this effort. Sure, we have much more important things in todo shortlist.

webknjaz: The problem is that it's hard to understand the intent behind the tests sometimes. Which means that I often cannot judge whether the test is correct and then cannot approve the PR without full understanding. Docstrings could help with this a lot.

docs/client_reference.rst
derlih: I've put 3.8 here because I thought it would be a next release version. I can put 3.7.something instead.

webknjaz: There will be no more 3.7 stream releases. My concern was whether this will get merged in the nearest future before 3.8.0 is out or not. If not, this feature would need to be moved to 3.9 maybe, or to a more specific patch release of 3.8.x. This is why I wanted to keep an eye on this.

tests/test_client_session.py
derlih: @webknjaz I've changed the names of those two tests. Is it more understandable what is actually tested?

CHANGES/6013.feature
```diff
@@ -0,0 +1 @@
+Added ``base_url`` parameter to the initializer of :class:`ClientSession`.
```
webknjaz: Linter fix:
```suggestion
Added ``base_url`` parameter to the initializer of :class:`~aiohttp.ClientSession`.
```

tests/test_client_session.py
derlih: That was exactly the point. In this test I don't care about the exception, I care only about the call of the `request_class`.
Actually quite many things happen inside `ClientSession._request`, and I don't want this test to fail because something unrelated to it has been changed.

tests/test_client_session.py
derlih: Same idea as with `supress(Exception)`. Here the only thing that matters is the `url` param of the `request_class` constructor.

tests/test_client_session.py
webknjaz: I disagree with this perception. There are things that we'd want to fail the test. For example, we make pytest error out on any warnings (namely deprecation warnings). This will silently shadow those while they must bubble up and show up as an error.

docs/client_reference.rst
derlih: I liked the feature, that's why I wanted to jump onto 3.8 release.

tests/test_client_session.py
webknjaz: It's better now, but sometimes it's hard to come up with good function names. This is where an old-good docstring with a paragraph-long description would come in handy.

docs/client_reference.rst
webknjaz: Sure, but it's not a matter of liking it. It's mostly whether you'll manage to polish the PR to an acceptable state and get approvals+backport before the deadline.

tests/test_client_session.py
```diff
@@ -706,3 +706,48 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+
+
+@pytest.mark.parametrize(
+    ("base_url", "url", "expected_url"),
+    [
+        (None, "http://example.com/test", URL("http://example.com/test")),
+        ("http://example.com", "/test", URL("http://example.com/test")),
+        (URL("http://example.com"), "/test", URL("http://example.com/test")),
+        ("http://example.com/", "/test", URL("http://example.com/test")),
+        (URL("http://example.com/"), "/test", URL("http://example.com/test")),
```
webknjaz: Note that for such complex parameter combos, it's best to assign the pytest IDs manually.

tests/test_client_session.py
webknjaz: It's not the only thing that matters. Shadowing the underlying warnings shouldn't happen because it's important for us to see any deprecations as soon as any runtime deps get updated: https://github.com/aio-libs/aiohttp/pull/6129/files#r738312860.

tests/test_client_session.py
derlih: How can I assign pytest IDs?

aiohttp/client.py
derlih: Do you know if the proper merge is supported by `yarl`?

tests/test_client_session.py
asvetlov: [pytest docs](https://docs.pytest.org/en/6.2.x/example/parametrize.html#different-options-for-test-ids) has several examples.

I've found the version 3 with embedded id for each parametrization a little verbose but the most readable:
```
@pytest.mark.parametrize(
    "a,b,expected",
    [
        pytest.param(
            datetime(2001, 12, 12), datetime(2001, 12, 11), timedelta(1), id="forward"
        ),
        pytest.param(
            datetime(2001, 12, 11), datetime(2001, 12, 12), timedelta(-1), id="backward"
        ),
    ],
)
def test_timedistance_v3(a, b, expected):
    diff = a - b
    assert diff == expected
````

File: aiohttp/client.py [hunk 15]
```diff
@@ -216,6 +218,11 @@ def __init__(
         trace_configs: Optional[List[TraceConfig]] = None,
         read_bufsize: int = 2 ** 16,
     ) -> None:
+        if isinstance(base_url, str):
+            self._base_url: Optional[URL] = URL(base_url)
+        else:
+            self._base_url = base_url
```
asvetlov:
```suggestion
        if base_url is None or isinstance(base_url, URL):
            self._base_url: Optional[URL] = base_url
        else:
            self._base_url = URL(base_url)
```

The reversed check handles more errors in runtime, e.g. passing a number or list as `base_url`

aiohttp/client.py
```diff
@@ -363,7 +378,10 @@ async def _request(
         proxy_headers = self._prepare_headers(proxy_headers)
 
         try:
-            url = URL(str_or_url)
+            if isinstance(str_or_url, URL):
+                url = str_or_url
+            else:
+                url = self._build_url(str_or_url)
```
asvetlov:
```suggestion
            url = self._build_url(URL(str_or_url))
```
URL.join() supports URL argument, this version can handle relative URLs (`URL('rel/path/to?a=1&b=2')`) as a bonus.

aiohttp/client.py
asvetlov: Good question.
`URL(base).join(URL(url))` should do the trick; it uses `urllib.parse.urljoin()`.

`URL(base) / url` supports only `url` *not started* with `/` (it doesn't conflict with `.join()` behavior at least).

In `yarl`, potentially we can accept both `str` and `URL` for the `join(...)` argument and make `__truediv__` an alias for `.join(...)`.  I see no objections but need to try, sure.

tests/test_client_session.py
webknjaz: Yep, either wrap with `pytest.param(..., id=...)` or add a detached argument `ids=('id1', 'id2', ...)`.

tests/test_client_session.py
derlih: It's also hard to write the proper description for parametrized test.:-(

aiohttp/client.py
derlih: Adjusted to `URL.join`

tests/test_client_session.py
derlih: I didn't know about this nice feature. Thanks for the hint!

tests/test_client_session.py
webknjaz: It is easy if a parametrized test checks one high-level thing.
And the params are often just IDs w/o whitespaces. They can be used to select a specific test variant to run via `pytest` CLI w/o having to execute all of them which is useful during testing. It's also useful in the test report output. Param IDs are not supposed to be complex, vice versa — the autogenerated param IDs for complex combos are not very readable and the manually chosen IDs are usually short and precise.

aiohttp/client.py
asvetlov: Sorry, edited str->URL.
Now the snipped is correct IMHO

tests/test_client_session.py
derlih: Is it better now?

docs/client_quickstart.rst
```diff
@@ -55,6 +55,18 @@ Other HTTP methods are available as well::
     session.options('http://httpbin.org/get')
     session.patch('http://httpbin.org/patch', data=b'data')
 
+To make several requests to the same site more simple, the parameter ``base_url``
+of :class:`ClientSession` constructor can be used. For example to request different
+endpoints of ``http://httpbin.org`` can be used the following code::
+
+    async with aiohttp.ClientSession('http://httpbin.org') as session:
+        async with session.get('/get'):
```
asvetlov, line 63: This code is correct, sure.

I think an example (or `.. note::`) that demonstrates the result of joining base URL with partial path and a relative path can be helpful also.
httpbin.org supports `/status/{codes}` and `/image/jpeg` / `/image/png` urls.

derlih: It is an example that uses `base_url` feature of the code above.

tests/test_client_session.py
webknjaz: A little. But it seems to still produce quite complex IDs.

tests/test_client_session.py
```diff
@@ -706,3 +706,73 @@ async def test_requote_redirect_url_default_disable() -> None:
...
+
+
+@pytest.mark.parametrize(
+    ("base_url", "url", "expected_url"),
+    [
+        pytest.param(
+            None,
+            "http://example.com/test",
+            URL("http://example.com/test"),
+            id="base_url(None) + str(http://example.com/test)",
```
webknjaz: I think I'd go for object reprs like this
```suggestion
            id="base_url=None url='http://example.com/test'",
```

docs/client_quickstart.rst
webknjaz: Something like a doctest could be more visual here.
```python
>>> invocation_block()
result
```

derlih: Actually this is the problem.
If I call `URL.join` for `http://httpbin.org/status` and `200` I get `http://httpbin.org/200` which is probably not expected result :-)

asvetlov: Oooh.

Tests for the standard `urllib.parse.urljoin()`:

```
In [1]: from urllib.parse import urljoin

In [2]: urljoin("http://a.b.c:/path", "to")
Out[2]: 'http://a.b.c:/to'

In [3]: urljoin("http://a.b.c:/path/", "to")
Out[3]: 'http://a.b.c:/path/to'

In [4]: urljoin("http://a.b.c:/path/thing", "to")
Out[4]: 'http://a.b.c:/path/to'

In [5]: urljoin("http://a.b.c:/path/thing/foo", "to")
Out[5]: 'http://a.b.c:/path/thing/to'

```

Sorry, I should go now. 
Tomorrow I'll check the compliance of behavior above with RFC. I suspect it matches :(

aiohttp/client.py
derlih: Already done

tests/test_client_session.py
derlih: Adjusted to your variant of naming

tests/test_client_session.py
derlih: In this particular case `_build_url` method can be threaten as a pure function. It gets one part of required data from `__init__`, another one from its parameter and does its job. In this case splitting tests doesn't make sense. It's easier to have one parametrized with input values and expected result.

Sometimes for pure functions I have two parametrized tests: one as I described above, another is for failing cases with expected exception instead of expected result.

docs/client_quickstart.rst
asvetlov: Ok, works as expected by RFC.

derlih: Checked one more time:
```
>>> from yarl import URL
>>> URL("http://httpbin.org/status").join(URL("/200"))
URL('http://httpbin.org/200')
```

derlih: @asvetlov 
It works as expected by RFC, but I think the feature wasn't about RFC implementation.
May be it would be better if `base_url='http://httpbin.org/status'` and `url='/200'` build `http://httpbin.org/status'/200`.
Not RFC complaint, but I would expect such behavior.

asvetlov: Another option is following the current `url / path` behavior: 
`base_url='http://httpbin.org/status'` and `url='/200'` is forbidden
`base_url='http://httpbin.org/status'` and `url='200'` builds `http://httpbin.org/status/200`

BTW, should we forbid query/fragment parts in `base_url`? They are ignored anyway.

aiohttp/client.py
```diff
@@ -304,6 +311,12 @@ def request(
         """Perform HTTP request."""
         return _RequestContextManager(self._request(method, url, **kwargs))
 
+    def _build_url(self, str_or_url: StrOrURL) -> URL:
+        if self._base_url is None:
+            return URL(str_or_url)
+        else:
+            return self._base_url.join(URL(str_or_url))
```
asvetlov:
```suggestion
    def _build_url(self, str_or_url: StrOrURL) -> URL:
        url = URL(str_or_url)
        if self._base_url is None:
            return url
        else:
            assert not url.is_absolute() and url.path.startswith('/')
            return self._base_url.join(url)
```
Let's restrict the code to conforming provided docs examples and avoiding edge cases.

asvetlov: We can extend the support later

aiohttp/client.py
```diff
@@ -216,6 +218,11 @@ def __init__(
         trace_configs: Optional[List[TraceConfig]] = None,
         read_bufsize: int = 2 ** 16,
     ) -> None:
+        if base_url is None or isinstance(base_url, URL):
+            self._base_url: Optional[URL] = base_url
+        else:
+            self._base_url = URL(base_url)
```
asvetlov, line 224:
```suggestion
            self._base_url = URL(base_url)
            assert self._base_url.origin() == self._base_url, "Only absolute URLs without path part are supported"
```

asvetlov:
I'm merging as-is.
We can improve the feature support in future

asvetlov:
Backported by 59bbcac0

docs/client_quickstart.rst
derlih: Let's wait for the user feedback on it.

Issue #6131, opened 2021-10-25
Helper to run tests in different python versions


derlih:
### Is your feature request related to a problem?

Add helper targets to the Makefile that allow to run the test using the specific version of python and environment (AIOHTTP_NO_EXTENSIONS)

### Describe the solution you'd like

Introduce new targets:
- `test-3.7`
- `test-3.7-no-extensions`

and so on.

It will make easier to check locally the problems occurred during CI on the python versions that are not installed locally.

### Describe alternatives you've considered

Run the required steps manually.

### Related component

Server, Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

webknjaz:
Personally, I'd prefer to have tox for this but I know that Andrew doesn't like it.

derlih:
I just need a single command to run it. I don't care much how it would work. Would it use docker, tox or other tools.
For instance I had some failing tests for 3.7 and I'm not able to install it on my PC via package manager, I've started to search how to run it. Then I realized, that it's not the first time I'm doing it :-)
That's why I've created this ticket and PR #6132 .

PR #6132, opened 2021-10-25
Run test in docker

derlih:
## What do these changes do?

Add new test targets to the Makefile

No

## Related issue number

#6131 

## Checklist

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -150,7 +150,7 @@ Any extra texts (print statements and so on) should be removed.
      make test-<python-version>[-no-extensions]
 
    For example, if you want to run tests for python3.7
-   without extentions, you can run this command::
+   without extensions, you can run this command::
 
      make test-3.7-no-extensions
 


asvetlov:
Are you committed to supporting this mode for years, adding python 3.11 etc?

I personally use  pyenv for such things:
```
cd aiohttp
pyenv install 3.10
pyenv virtualenv 3.10 aiohttp-3.10
pyenv local aiohttp-3.10
```
`pyenv local ...` call can be used for switching aiohttp environments.
Maybe this approach will work for you?

webknjaz:
I'm also not sure I'd want to support yet another test invocation method. Using pyenv is more than enough normally.

derlih:
This mode would be only used by people who contribute to aiohttp. If such a person sees that a new python version is missing, they can just add with a couple of lines, isn't it?

derlih:
@asvetlov with backporting to 3.8?

asvetlov:
Up to you. Probably having the PR backported can be handy

Issue #6130, opened 2021-10-25
Subprocess returncode is not detected when running Gunicorn with aiohttp


harrylyx:
### Describe the bug

When starting Gunicorn with aiohttp worker(s), if the app uses `subprocess` to start other processes and captures the output, their `returncode` is in most cases `0`, even if the actual exit code was `1`.

### To Reproduce

```python
import asyncio
import subprocess
from aiohttp import web
from concurrent.futures import ProcessPoolExecutor

EXECUTOR = ProcessPoolExecutor(4)

def create():
    subprocess_run = subprocess.run(
        "pip install numpy==123",
        shell=True,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
    )
    print(subprocess_run.returncode)
    print(subprocess_run.stderr.decode("utf-8"))
    return f"{subprocess_run.returncode}, {subprocess_run.stderr.decode('utf-8')}"


async def index(request):
    loop = asyncio.get_event_loop()
    return web.Response(text=await loop.run_in_executor(EXECUTOR, create))


async def my_web_app():
    app = web.Application()
    app.router.add_get("/", index)
    return app


if __name__ == "__main__":
    web.run_app(my_web_app())
```

Then run it with:
```
gunicorn start:my_web_app --bind localhost:8080 --worker-class aiohttp.GunicornWebWorker
```

### Expected behavior

The detected `returncode` should always be `1`, as the subprocess always exits with `1`.

### Logs/tracebacks

```python-traceback
❯ gunicorn start:my_web_app --bind localhost:8080 --worker-class aiohttp.GunicornWebWorker
[2021-10-25 16:53:27 +0800] [4190] [INFO] Starting gunicorn 20.1.0
[2021-10-25 16:53:27 +0800] [4190] [INFO] Listening at: http://127.0.0.1:8080 (4190)
[2021-10-25 16:53:27 +0800] [4190] [INFO] Using worker: aiohttp.GunicornWebWorker
[2021-10-25 16:53:27 +0800] [4193] [INFO] Booting worker with pid: 4193
0
[...]

PR #6134, opened 2021-10-25
Fix gunicorn signal bug

harrylyx:
## What do these changes do?

🐛 Fix resetting SIGCHLD  signals in Gunicorn aiohttp Worker to fix subprocesses that capture output having an incorrect returncode.

No

## Related issue number

Fix https://github.com/aio-libs/aiohttp/issues/6130

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
         #     os: ubuntu
       fail-fast: true
     runs-on: ${{ matrix.os }}-latest
-    timeout-minutes: 15
+    timeout-minutes: 20
     steps:
     - name: Checkout
       uses: actions/checkout@v2.3.5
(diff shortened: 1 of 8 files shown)


aiohttp/worker.py
```diff
@@ -167,6 +167,9 @@ def init_signals(self) -> None:
         # by interrupting system calls
         signal.siginterrupt(signal.SIGTERM, False)
         signal.siginterrupt(signal.SIGUSR1, False)
+        # Reset signals so Gunicorn doesn't swallow subprocess return codes
+        # See: https://github.com/aio-libs/aiohttp/issues/6130
+        signal.signal(signal.SIGCHLD, signal.SIG_DFL)
```
asvetlov:
```suggestion
        if sys.version_info < (3, 8):
            # Starting from Python 3.8, the default child watcher is ThreadedChildWatcher.
            # The watcher doesn't depend on SIGCHLD signal, there is no need to reset it.
            signal.signal(signal.SIGCHLD, signal.SIG_DFL)
```

asvetlov: IMHO the change is not needed for Python 3.8+.

I'm ok to apply it for Python 3.6-3.7 though.

PR #6139, opened 2021-10-27
Setup binary builds for Alpine and MacOS arm

asvetlov:


.github/workflows/ci.yml
```diff
@@ -153,13 +153,13 @@ jobs:
     needs: pre-deploy
     steps:
     - name: Checkout
-      uses: actions/checkout@v2.3.5
-      with:
-        submodules: true
-    - name: Setup Python 3.8
-      uses: actions/setup-python@v2
+      uses: actions/checkout@v2.3.4
```
webknjaz: Why did you downgrade this?

asvetlov: Ooops. Maybe because yarl has no dependabot settings for github actions.
Perhaps we need https://github.com/asottile/all-repos org repo for applying configuration sync changes to all desired repos.
I use such a system on my job. Sorry, cannot share a link to the private repo for demonstration of the idea.

asvetlov:
@webknjaz  Any idea why the chronographer didn't catch CHANGES files?

CHANGES/6139.1.feature
```diff
@@ -0,0 +1 @@
+Started shipping platform-specific arm64 wheels for Apple Silicon.
```
webknjaz: I've recently enabled the `:user:` role in this repo — try it out:
```suggestion
Started shipping platform-specific arm64 wheels for Apple Silicon — :user:`asvetlov`.
```

CHANGES/6139.2.feature
```diff
@@ -0,0 +1 @@
+Started shipping platform-specific wheels with the ``musl`` tag targeting typical Alpine Linux runtimes.
```
webknjaz:
```suggestion
Started shipping platform-specific wheels with the ``musl`` tag targeting typical Alpine Linux runtimes — :user:`asvetlov`.
```

.github/workflows/ci.yml
webknjaz: Personally, I've been meaning to try a little bit different approach from https://github.com/jaraco/skeleton which is Git-based. I wonder we they could be combined...

webknjaz:
@asvetlov

> @webknjaz Any idea why the chronographer didn't catch CHANGES files?

Because you named them wrong. The number should go after the type.
`CHANGES/6139.1.feature` -> `CHANGES/6139.feature.1`.

By the way, we now have a Sphinx integration that also shows unreleased change notes. See https://aiohttp--6139.org.readthedocs.build/en/6139/changes.html#to-be-included-in-vrelease-if-present.

Look closer and you'll see that your names confused Towncrier and it linked issues #1 and #2 to your change notes and not #6139.

asvetlov:
Aha, thanks!

.github/workflows/ci.yml
asvetlov: Anyway, out of the PR scope

asvetlov:
Backported by d8efd321

PR #5364, opened 2020-12-22
WIP http-parser -> llhttp

derlih:
## What do these changes do?

Migrate from [http-parser](https://github.com/nodejs/http-parser) to [llhttp](https://github.com/nodejs/llhttp)

## Are there changes in behavior for the user?

No

## Related issue number

#3561 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

Issue #5621, opened 2021-04-16
3.7.0 regression: aiohttp route matching is now pre url decode instead of post decode


thehesiod:
Pre 3.7.0 (3.6.3 and before) you could write a route regex match to work on the post-decoded url, however now it swapped to pre-decode causing our route handlers to 404

💡 **To Reproduce**
Running the following server
```python
import aiohttp
from aiohttp import web
import yarl


async def hello(request: web.Request):
    return web.Response(text=str(dict(request.match_info)))


def main():
    app = web.Application()
    app.add_routes([web.get('/hello', hello)])
    app.add_routes([web.get('/{user_ids:(([0-9]+)|(u_[0-9a-f]{16}))(,(([0-9]+)|(u_[0-9a-f]{16})))*}/hello', hello)])
    web.run_app(app, port=8888)


if __name__ == '__main__':
    print(f'aiohttp version: {aiohttp.__version__}')
    print(f'yarl version: {yarl.__version__}')
    main()
```
with aiohttp 3.6.3 + yarl 1.5.1

and hitting `curl 'http://0.0.0.0:8888/467%2C802%2C24834%2C24952%2C25362%2C40574/hello'`

results in a 200

However running the server w/ 3.7.0-3.7.4.post0 + same yarl results in 404. 

📋 **Your version of the Python**
3.8.5

📋 **Your version of the aiohttp/yarl/multidict distributions**
various, see above

If I revert the changes to `_http_parser.pyx` introduced in https://github.com/aio-libs/aiohttp/commit/d3219238447f09162f2c38c63ae136a94f39c594 it works again.

thehesiod:
I found that by setting `AIOHTTP_NO_EXTENSIONS=1` it also works.  So there's a discrepancy between the cython and python code

thehesiod:
so python version just does [URL(path)](https://github.com/aio-libs/aiohttp/blob/v3.7.0/aiohttp/http_parser.py#L472) whereas cython does [_parse_url](https://github.com/aio-libs/aiohttp/blob/v3.7.0/aiohttp/_http_parser.pyx#L584)

webknjaz:
[...]

PR #6144, opened 2021-10-27
http-parser -> llhttp

### LINKED ISSUE #3561: Switch http_parser to llhttp
Nodejs decide to move from their hard-coded C parser to this one.

https://github.com/nodejs/http-parser/pull/285#issuecomment-456025694

*Upd:* 
proposed parser is https://llhttp.org

asvetlov:
The resurrection of #5364

Closes #3561

Co-authored-by: Dmitry Erlikh <derlih@gmail.com>

### DIFF
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -324,13 +324,15 @@ cdef class HttpParser:
         PyMem_Free(self._cparser)
         PyMem_Free(self._csettings)
 
-    cdef _init(self, cparser.llhttp_type mode,
-                   object protocol, object loop, int limit,
-                   object timer=None,
-                   size_t max_line_size=8190, size_t max_headers=32768,
-                   size_t max_field_size=8190, payload_exception=None,
-                   bint response_with_body=True, bint read_until_eof=False,
-                   bint auto_decompress=True):
+    cdef _init(
+        self, cparser.llhttp_type mode,
+        object protocol, object loop, int limit,
+        object timer=None,
+        size_t max_line_size=8190, size_t max_headers=32768,
+        size_t max_field_size=8190, payload_exception=None,
+        bint response_with_body=True, bint read_until_eof=False,
+        bint auto_decompress=True,
+    ):
         cparser.llhttp_settings_init(self._csettings)
         cparser.llhttp_init(self._cparser, mode, self._csettings)
         self._cparser.data = <void*>self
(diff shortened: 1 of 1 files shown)


asvetlov:
@derlih the initial #5364 had about 8 failed tests when merged to master.
I've reverted the merge and created this PR to don't miss your work.
Not sure when I can find time to work on failed tests; my current target is publishing 3.8 with python 3.10 support, llhttp is nice to have but not mandatory.

If we cannot find time to fit in 3.8 time window -- we can publish 3.9 relatively soon, there is no need to wait another year if we will get a value soon.

P.S.
The branch is available for pushing to every aiohttp committer, feel free to help. Please don't be confused with `revert-6143-revert-5364-llhttp` branch name (revert of reversion :)

derlih:
I've updated the `llhttp` to the latest stable and fixed the spelling error in CI

derlih:
`test_http_request_parser_two_slashes` fails for `llhttp`. Looks line the behavior has been changed in comparison to `http_parser`.
Do we really need this test? It checks that python parser behaves the same as a native one. But this case tests wrong path with `//` inside. Also it doesn't cover many other cases with `//`. For example `llhttp` for path `//path/a` returns `/a`. Should this strange result be also implemented by python http parser and covered by tests?

Update:
This behavior is not covered by `llhttp` [tests](https://github.com/nodejs/llhttp/blob/v6.0.6/test/url.md).

derlih:
The next problem is `aiohttp.web.normalize_path_middleware`.
This middleware has the `merge_slashes` parameter:

> If merge_slashes is True, merge multiple consecutive slashes in the path into one.

But any middleware is called after the HTTP parser. And since the behavior of double slashes in path is changed in `llhttp`, it makes sense to remove this `merge_slashes` parameter from this middleware. It's an API change -> should be shifted to the next major release.

Also I think it means that we can't have a transition phase with both native parsers at the same time.

derlih:
Also I found that `llhttp` fixes regression #5621 :-)

asvetlov:
> `test_http_request_parser_two_slashes` fails for `llhttp`...

We cannot change llhttp upstream (at least it requires long enough conversation I think).
But we can change the behavior of our pure-python parser to get the same result for both Python and llhttp parsing results.
Please update the failed test.

asvetlov:
> The next problem is `aiohttp.web.normalize_path_middleware`...

If we really cannot support `merge_slashes=False` let's make the argument no-op and raise `DeprecationWarning` if `False` was passed.

asvetlov:
> Also I found that `llhttp` fixes regression #5621 :-)

Awesome!

asvetlov:
@derlih I think I've found the reason for broken double-slash tests.

llhttp is correct but `URL("//abc")` assumes that 'abc' is an authority, not path.
`URL.build()` should be used instead.
I'm working on it.

derlih:
@asvetlov the middleware test is checking the [security issue](https://github.com/advisories/GHSA-v6wp-4m6f-gcjg) inside aiohttp. I don't want to reintroduce it )

asvetlov:
I didn't touch tests except `test_http_parser.py`, where only fixture ids are added without changing the actual test code.

asvetlov:
Tests are green, merging

PR #6163, opened 2021-10-28
Add clarification to configuring app

### LINKED ISSUE #4137: docs: explain why all setups we set to app
Add in documentation part (maybe into tutorial) where explaining why all connection, settings and etс, we add to the instance of application.

example:

```python
async def init_pg(app):
    conf = app['config']['postgres']
    engine = await aiopg.sa.create_engine(
        database=conf['database'],
        user=conf['user'],
        password=conf['password'],
        host=conf['host'],
        port=conf['port'],
        minsize=conf['minsize'],
        maxsize=conf['maxsize'],
    )
    app['db'] = engine
```

Please don't forget:

- add screenshot with your part of documentation
- write below that u want to take it to work

sha016:
## What do these changes do?

Added clarification on configuring the app object with setings such as a db connection

No

## Related issue number

Closes #4137 

## Checklist

### DIFF
--- a/CHANGES/4137.doc
+++ b/CHANGES/4137.doc
@@ -1,0 +1,1 @@
+Added clarification on configuring the app object with setings such as a db connection
--- a/CHANGES/4137.doc
+++ b/CHANGES/4137.doc
@@ -1,1 +1,1 @@
-Added clarification on configuring the app object with setings such as a db connection
+Added clarification on configuring the app object with settings such as a db connection
--- a/CHANGES/4137.doc
+++ b/CHANGES/4137.doc
@@ -1,1 +1,1 @@
-Added clarification on configuring the app object with settings such as a db connection
+Added clarification on configuring the app object with settings such as a db connection.
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -278,7 +278,8 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
-The app object can be used in this way to reuse a database connection or anything else needed throughout the application.
+The app object can be used in this way to reuse a database connection or anything
+else needed throughout the application.
 
 See this reference section for more detail: :ref:`aiohttp-web-app-and-router`.
 
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -239,6 +239,13 @@ Data Sharing aka No Singletons Please
 :mod:`aiohttp.web` discourages the use of *global variables*, aka *singletons*.
 Every variable should have its own context that is *not global*.
 
+Global variables are generally considered bad practice due to the complexity
+they add in keeping track of state changes to variables.
+
+*aiohttp* does not use globals by design, which will reduce the number of bugs
+and/or unexpected behaviors for its users. For example, an i18n translated string
+being written for one request and then being served to another.
+
 So, :class:`Application` and :class:`Request`
 support a :class:`collections.abc.MutableMapping` interface (i.e. they are
 dict-like objects), allowing them to be used as data stores.
@@ -271,6 +278,9 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
+You may use the app object to connect your database to your web application.
+
+See this reference section for more detail: :ref:`aiohttp-web-app-and-router`.
 
 Request's storage
 ^^^^^^^^^^^^^^^^^
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -278,7 +278,7 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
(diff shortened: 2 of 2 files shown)


docs/web_advanced.rst
```diff
@@ -271,6 +271,10 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
+When connecting a web application to a database, a common convention is to
+ reference the connection object through the key 'db' in the global app object.
```
webknjaz:
```suggestion
reference the connection object through the key ``'db'`` in the global app object.
```

webknjaz:
> Closes #4137

I don't believe this solves that issue. The issue asks for clarification and explanation about _why_, not _how_.

sha016:
Would you agree that by saving the reference to the database connection in `app['db']` it allows you to access it easily later in the request context via `request.app['db']`?

https://docs.aiohttp.org/en/stable/web_reference.html#application

Is that mainly why one should do this or are there other sections of code that look in `app['db']` for a connection? I'm not intimately familiar with the codebase.

Thanks in advance,

Arfey:
Now i don't think so that is the best solution for share connection with storage or something like that. Current pattern produce a lot of drilling arguments between functions

```python
async def get_statistic(db, redis):
    data = await get_statistic(db)
    await inc_stats_view(redis)
    return data

async def index(req):
   await get_data(req.app['db'], req.app['redis])
   ...
```

it's not okey. in my opinion the best way is to create a context variable for db and redis connections and then use them in `get_statistic ` and `inc_stats_view `.

I think so, problem #4137 in 2021 is not relevant, but I could be wrong

Dreamsorcerer:
If you compare to other projects like flask, they often use global variables (`flask.g` etc.). This approach is less explicit and can be very error prone (my previous company's flask site was randomly serving Chinese translations to all visitors due to this, which could never have happened with the aiohttp approach).

So, I feel that the benefit is making sure everything is explicit, thus reducing the chance of bugs, by having the request/app object passed directly into the handler function and never using globals. The app object just gives us a convenient place to store these objects, but the main point is that everything is passed to the function explicitly.

It is also worth noting that we encourage the same approach with handling the app object itself. In many other frameworks the app object is created as a global variable in their examples. In all the aiohttp examples, the app object is always created as a local variable in a function (`create_app()` or similar). Again, this helps people to avoid problems which I've often seen on flask projects where the app object is getting imported into nearly every module (e.g. to use `@app.get()` decorators for handlers) and then ends up with circular imports and import ordering issues.

asvetlov:
@Dreamsorcerer the perfect summary of my (never written maybe) thoughts when I designed this part of `aiohttp.web` API.

sha016:
Maybe that was poor word choice calling the  app object global. So the _why_ to this issues is simply convenience?

Dreamsorcerer:
> Maybe that was poor word choice calling the app object global. So the _why_ to this issues is simply convenience?

No, reread my comment, I was not responding to you calling it a global, but explaining the _why_. We should be explaining why we use the app object in this way, as opposed to the way other popular frameworks do it (i.e. avoiding globals etc.). If convenience was the only factor, we would just be lazy and use globals like flask.

The app object is a convenient location for storing these things _given the design parameters of aiohttp_. But, it's those design parameters which are the _reason_ we end up using it. So, any documentation for this should probably mention globals and explicitness, and how this reduces the chances of bugs in code.

CHANGES/4137.doc
```diff
@@ -0,0 +1 @@
+Added clarification on configuring the app object with setings such as a db connection
```
asvetlov:
```suggestion
Added clarification on configuring the app object with settings such as a db connection
```

sha016:
@Dreamsorcerer Thanks for explaining your thoughts @asvetlov I've amended the commit and it should cover what we've discussed here.

CHANGES/4137.doc
```diff
@@ -0,0 +1 @@
+Added clarification on configuring the app object with settings such as a db connection
```
webknjaz:
```suggestion
Added clarification on configuring the app object with settings such as a db connection.
```

webknjaz:
@Arfey WDYT?

docs/web_advanced.rst
```diff
@@ -271,6 +278,9 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
+You may use the app object to connect your database to your web application.
```
```suggestion
The app object can be used in this way to reuse a database connection or anything else needed throughout the application.
```

Dreamsorcerer (approved):
Looks OK to me.

docs/web_advanced.rst
```diff
@@ -271,6 +278,9 @@ For this please use :attr:`Request.config_dict` read-only property::
     async def handler(request):
         data = request.config_dict['my_private_key']
 
+The app object can be used in this way to reuse a database connection or anything else needed throughout the application.
```
webknjaz:
```suggestion
The app object can be used in this way to reuse a database connection or anything
else needed throughout the application.
```

PR #6176, opened 2021-10-29
Bump the `proxy.py` test dep to v3.4.0

webknjaz:
## What do these changes do?

This is the version that started supporting ephemeral port as a
first-class citizen so this patch also uses that.

(Note that this is currently a draft because the recent `proxy.py` has a dependency conflict with us — they put a precise pin on the `typing-extensions` requirement; to be bumped once the upstream issue is solved)

No.

## Related issue number

https://github.com/abhinavsingh/proxy.py/issues/617

## Checklist

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -5,7 +5,7 @@ cryptography==3.3.1; platform_machine!="i686" and python_version<"3.9" # no 32-b
 freezegun==1.1.0
 mypy==0.910; implementation_name=="cpython"
 mypy-extensions==0.4.3; implementation_name=="cpython"
-proxy.py==2.3.1
+proxy.py==3.4.0
 pytest==6.2.5
 pytest-cov==3.0.0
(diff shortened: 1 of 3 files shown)


requirements/test.txt
```diff
@@ -5,7 +5,7 @@ cryptography==3.3.1; platform_machine!="i686" and python_version<"3.9" # no 32-b
 freezegun==1.1.0
 mypy==0.910; implementation_name=="cpython"
 mypy-extensions==0.4.3; implementation_name=="cpython"
-proxy.py==2.3.1
+proxy.py==3.4.0
```
webknjaz, line 8: Looks like this was supposed to be v2.4.0 but it's already `proxy.py ~= 2.4.4rc3`.

webknjaz (commented):
This change has already been implemented in another commit. Nothing to do here.

Issue #5896, opened 2021-07-18
Any media type starting with application/json treated as JSON


scop:
### Describe the bug

Any media type that starts with `application/json` is treated as JSON. This treatment should be limited to `application/json` and types with `+json` media type suffix. An example of a registered media type that triggers this problem is https://www.iana.org/assignments/media-types/application/json-seq

### To Reproduce

PR #5895 has a failing test case.

### Expected behavior

Not anything that begins with `application/json` should be treated as JSON.

### Logs/tracebacks

```python-traceback
-
```

### Python Version

```console
$ python --version
Python 3.9.6
```

### aiohttp Version

```console
$ python -m pip show aiohttp
WARNING: Package(s) not found: aiohttp

(current git master)
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 5.1.0
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /home/scop/.pyenv/versions/3.9.6/envs/aiohttp/lib/python3.9/site-packages
Requires: 
Required-by: yarl
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.6.3
Summary: Yet another URL library
Home-page: https://github.com/aio-libs/yarl/
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /home/scop/.pyenv/versions/3.9.6/envs/aiohttp/lib/python3.9/site-packages
Requires: multidict, idna
Required-by:
```

### OS

Any

### Related component

Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

asvetlov:
Fixed by #5894

scop:
As mentioned in https://github.com/aio-libs/aiohttp/pull/5895#issuecomment-950321559, this is a different thing, not fixed by #5894. The test case in #5895 still fails.

scop:
Real fix in #6180.

PR #6180, opened 2021-10-31
Fix too lax application/json prefix-only matching

scop:
## What do these changes do?

Fix too lax application/json matching.

For example, `application/jsonfoo` or [`application/json-seq`](https://www.iana.org/assignments/media-types/application/json-seq) should not be treated as valid for JSON.

## Related issue number

Fixes https://github.com/aio-libs/aiohttp/issues/5896

## Checklist

### DIFF
--- a/CHANGES/6180.bugfix
+++ b/CHANGES/6180.bugfix
@@ -1,0 +1,1 @@
+Fix JSON media type matching to not accept everything starting with application/json.
--- a/CHANGES/6180.bugfix
+++ b/CHANGES/6180.bugfix
@@ -1,1 +1,1 @@
-Fix JSON media type matching to not accept everything starting with application/json.
+Fixed matching the JSON media type to not accept arbitrary characters after ``application/json`` or the ``+json`` media type suffix.
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -425,6 +425,7 @@ def is_expected_content_type(
 ) -> bool:
     """Checks if received content type is processable as an expected one.
 
+
     Both arguments should be given without parameters.
     """
     if expected_content_type == "application/json":
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -124,9 +124,7 @@ else:
         return asyncio.iscoroutinefunction(func)
 
 
-json_re = re.compile(
-    r"(?:application/|[\w.-]+/[\w.+-]+?\+)json(?:\s*;.*)?$", re.IGNORECASE
-)
+json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json$", re.IGNORECASE)
 
 
 class BasicAuth(namedtuple("BasicAuth", ["login", "password", "encoding"])):
@@ -425,6 +423,9 @@ def content_disposition_header(
 def is_expected_content_type(
     response_content_type: str, expected_content_type: str
 ) -> bool:
+    """Checks if received content type is processable as an expected one.
+    Both arguments should be given without parameters.
+    """
     if expected_content_type == "application/json":
         return json_re.match(response_content_type) is not None
     return expected_content_type in response_content_type
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -124,7 +124,9 @@ else:
         return asyncio.iscoroutinefunction(func)
 
 
-json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json", re.IGNORECASE)
+json_re = re.compile(
+    r"(?:application/|[\w.-]+/[\w.+-]+?\+)json(?:\s*;.*)?$", re.IGNORECASE
+)
 
 
 class BasicAuth(namedtuple("BasicAuth", ["login", "password", "encoding"])):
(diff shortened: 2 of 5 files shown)


scop:
Note that this will conflict with #6181, will fix the other after one of these is merged.

asvetlov:
#6181 is merged, please fix conflicts

scop:
Done/rebased.

tests/test_helpers.py
```diff
@@ -789,6 +789,22 @@ def test_is_expected_content_type_json_non_lowercase():
     )
 
 
+def test_is_expected_content_type_json_with_charset():
```
webknjaz: Nit: instead of copy-pasting, I'd prefer having the same test be consolidated with `@pytest.mark.parametrize()`

tests/test_helpers.py
```diff
@@ -789,6 +789,22 @@ def test_is_expected_content_type_json_non_lowercase():
...
+        response_content_type=response_ct, expected_content_type=expected_ct
+    )
+
+
+def test_is_expected_content_type_json_match_not_just_prefix():
+    expected_ct = "application/json"
+    response_ct = "application/json-seq"
+    assert not is_expected_content_type(
+        response_content_type=response_ct, expected_content_type=expected_ct
+    )
```
webknjaz:
```suggestion
@pytest.mark.paratetrize(
    ("expected_ct", "response_ct", "should_match"),
    (
        ("application/json", "application/json; charset=UTF-8", True),
        ("application/json", "application/json-seq", False),
    ),
    ids=("with-charset", "non-plus-suffix-should-not-match"),
)
def test_is_expected_content_type_json(expected_ct, response_ct, should_match):
    assert is_expected_content_type(
        response_content_type=response_ct, expected_content_type=expected_ct
    ) is should_match
```

CHANGES/6180.bugfix
```diff
@@ -0,0 +1 @@
+Fix JSON media type matching to not accept everything starting with application/json.
```
webknjaz:
```suggestion
Fixed matching the JSON media type to only accept ``application/json`` when it's plus-delimited -- :user:`scop`.
```

scop: I don't think this is what this change actually does -- there's nothing plus-delimited here. The only thing we do to the regex is to add the `(?:\s*;.*)?$` in the end (plus the resulting `black` formatting change).

Perhaps "Fixed matching the JSON media type to not accept arbitrary characters after ``application/json``."

webknjaz: Okay, that would do.

scop: Added that plus "or the ``+json`` media type suffix" as it applies to those types as well.

scop:
Done.

aiohttp/helpers.py
```diff
@@ -423,6 +423,9 @@ def content_disposition_header(
 def is_expected_content_type(
     response_content_type: str, expected_content_type: str
 ) -> bool:
+    """Checks if received content type is processable as an expected one.
+    Both arguments should be given without parameters.
```
webknjaz, line 429: Per PEP 257, there must always be an empty line between the docstring title and the description part.
```suggestion

    Both arguments should be given without parameters.
```

aiohttp/helpers.py
```diff
@@ -124,7 +124,7 @@ def iscoroutinefunction(func: Any) -> bool:
         return asyncio.iscoroutinefunction(func)
 
 
-json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json", re.IGNORECASE)
+json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json$", re.IGNORECASE)
```
webknjaz, line 127: Maybe mark the start as well?
```suggestion
json_re = re.compile(r"^(?:application/|[\w.-]+/[\w.+-]+?\+)json$", re.IGNORECASE)
```

webknjaz:
I don't have any additional comments beyond those few nitpicks so I'll let @asvetlov decide when to merge this.

aiohttp/helpers.py
scop: This regex is used with `.match` and is not a `re.MULTILINE` one, so the start marker does not make a difference. The majority regexps in aiohttp that are used in this kind of setup don't have it either. IMO adding it would just be a potential source of confusion.

(And IMO the whole `re.match` method shouldn't even exist, _it_ is just a source of confusion and `re.search` could be used instead of it everywhere. Maybe some day it annoys me enough that I'll start the journey trying to get it deprecated and eventually removed from Python. </rant>)

aiohttp/helpers.py
scop: BTW if PEP 257 is desired here, I suggest adding flake8-docstrings to pre-commit. I can submit a PR for that, let me know.

asvetlov: Maybe it is a good idea.
To the node: I'm not volunteered to fix all broken docstrings if the check is applied :)

aiohttp/helpers.py
asvetlov: Good to me

asvetlov (approved):

asvetlov:
LGTM

PR #6182, opened 2021-10-31
Raise '400: Content-Length can't be present with Transfer-Encoding' if both Content-Length and Transfer-Encoding are sent by peer

asvetlov:
Fixes minor CVE


asvetlov:
Backported by f016f068

PR #6193, opened 2021-11-01
Bump async-timeout to >=4.0

asvetlov:


gitpushdashf:
Would it be possible for this to support async-timeout 3.0 at the same time?

asvetlov:
No, sorry

PR #6205, opened 2021-11-02
Set the minimum required `setuptools` to v46.4.0

jameshilliard:
In order to avoid a recursive dependency issue when parsing the aiohttp version during install setuptools needs:
https://github.com/pypa/setuptools/commit/c457e68319555b7266e5e0802946c5b1b63e4d61

This implies we need setuptools 46.4.0 or newer.

This should probably also be back-ported as this setuptools version requirement was introduced in `3.8`.

## What do these changes do?

Adds `setuptools >= 46.4.0` requirement to pyproject.toml and readme.

Should ensure setuptools version is correct instead of cryptic build dependency errors.

## Related issue number

N/A

## Checklist

- [N/A] I think the code is well written
- [N/A] Unit tests for the changes exist
- [N/A] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`


asvetlov (changes requested):
I believe we should specify setuptools version in `pyproject.toml`, not `setup.py`.

Please add something like:
```
[build-system]
requires = [
    "setuptools>=46.4.0",
]
build-backend = "setuptools.build_meta"

```

asvetlov:
@webknjaz please check the required minimal setuptools version.
I believe you are aware of setuptools features and changelog better than me.

jameshilliard:
> I believe we should specify setuptools version in `pyproject.toml`, not `setup.py`.

Yeah, I wasn't sure if something like that worked properly, I thought checking there required a fairly recent setuptools version or something as well or something.

jameshilliard:
> I believe you are aware of setuptools features and changelog better than me.

FYI for 3.8 it appears this is the change that bumped the setuptools requirement:
https://github.com/aio-libs/aiohttp/commit/4e39c6bcaceaa869a5ae299acdb8a930838bd8c0#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R3

asvetlov:
From my understanding, a relatively modern `pip` (about 3 years old or even older) looks at `pyproject.toml` to select build tool version used for installation.
The check from `setup.py` is redundant.

Did I miss something?

jameshilliard:
> From my understanding, a relatively modern `pip` (about 3 years old or even older) looks at `pyproject.toml` to select build tool version used for installation.

Would that work for installs that don't use pip?

pyproject.toml
```diff
@@ -11,3 +11,9 @@ issue_format = "`#{issue} <https://github.com/aio-libs/aiohttp/issues/{issue}>`_
 test-command = ""
 # don't build PyPy wheels, install from source instead
 skip = "pp*"
+
+[build-system]
+requires = [
+    "setuptools>=46.4.0",
+]
+build-backend = "setuptools.build_meta"
```
jameshilliard: Added `build-system` section.

webknjaz (changes requested):
TL;DR I don't see how this patch improves anything we specifically care about, it only seems to make the maintenance burden worse if anything. So I'm strongly opposed to adding the check but updates to `pyproject.toml` are fine.

PEP 517 build front-end uses `pyproject.toml` to populate the build env. When `setuptools` can be specified uncapped, in practice, the front-ends (like pip) will pull in the very latest version of it. So we are covered here.

Invoking `setuptools` directly is deprecated. Here's why https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html.
And also `distutils` has been deprecated for quite a while now so importing it is rather discouraged.

So here's my request:
1) Drop the changes from `setup.py`
2) Move the `[build-system]` to the top of `pyproject.toml`
3) Update the change note to reflect this and use the past tense there (you may find a more comprehensive guide in `CHANGES/README.rst`

webknjaz:
> Would that work for installs that don't use pip?

The ecosystem is standardized around PEP 517 now. We won't support weird installs and it's up to the users to abuse this in unsupported ways. This standard brings a lot of convenience and even distro packagers follow it (for example, Fedora has their own PEP 517 helper macros, some of the others invoke pip).

There is no point for us to support non-standardized things.

CHANGES/6205.bugfix
```diff
@@ -0,0 +1 @@
+Added check for compatible ``setuptools`` version.
```
webknjaz: This is not true anymore, nor is the commit message or the PR title. It needs to reflect the actual effect on the project when the patch is applied.

webknjaz: See https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme for examples.

webknjaz:
@jameshilliard please fill out the PR template, it seems you've deleted it so here's where you can recover it from https://github.com/aio-libs/aiohttp/blob/master/.github/PULL_REQUEST_TEMPLATE.md.

jameshilliard:
> Invoking setuptools directly is deprecated. Here's why https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html.
> And also distutils has been deprecated for quite a while now so importing it is rather discouraged.

Hmm, I wonder what the recommended way for handling cross compilation without direct invocation would be.

> This standard brings a lot of convenience and even distro packagers follow it (for example, Fedora has their own PEP 517 helper macros, some of the others invoke pip).

We're still using `setup.py` at the moment, I ran into this issue since we still in the process of fully removing python2 and were sharing the setuptools version:
https://github.com/buildroot/buildroot/blob/76c7fed85f7a822663cf0c8828e7d282198d0803/package/pkg-python.mk#L245-L249

> There is no point for us to support non-standardized things.

Oh, I assumed supporting those was why there was an explicit python version check in `setup.py`, was there a different reason for doing it that way?

CHANGES/6205.bugfix
jameshilliard: Isn't that what this does still?:
```
requires = [
    "setuptools>=46.4.0",
]
```

jameshilliard: I changed it to:
```
Added requirement for compatible ``setuptools`` version.
```
Is that better?

jameshilliard:
> please fill out the PR template, it seems you've deleted it so here's where you can recover it from

Ok, re-added, figured it was a bit verbose since most sections didn't seem applicable here for a trivial fix like this.

webknjaz:
> > please fill out the PR template, it seems you've deleted it so here's where you can recover it from
> 
> Ok, re-added, figured it was a bit verbose since most sections didn't seem applicable here for a trivial fix like this.

When it's missing, we still need to ask clarification questions which is basically extracting the same info from you but slower.

webknjaz:
> > Invoking setuptools directly is deprecated. Here's why [blog.ganssle.io/articles/2021/10/setup-py-deprecated.html](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).
> > And also distutils has been deprecated for quite a while now so importing it is rather discouraged.
> 
> Hmm, I wonder what the recommended way for handling cross compilation without direct invocation would be.

I guess it depends on what you need. There's also an option to have an in-tree PEP 517 build backend, for example.

> > This standard brings a lot of convenience and even distro packagers follow it (for example, Fedora has their own PEP 517 helper macros, some of the others invoke pip).
> 
> We're still using `setup.py` at the moment, I ran into this issue since we still in the process of fully removing python2 and were sharing the setuptools version: [buildroot/buildroot@`76c7fed`/package/pkg-python.mk#L245-L249](https://github.com/buildroot/buildroot/blob/76c7fed85f7a822663cf0c8828e7d282198d0803/package/pkg-python.mk#L245-L249)

I suppose you could participate in the discussion around the semantics of the arbitrary args that PEP 517 is able to pass from the front-end to the back-end. AFAIR setuptools doesn't accept those yet because they couldn't agree on the semantics. But once it's done, you should be able to just invoke `python -m build`.
I think the issue on the pip side is https://github.com/pypa/pip/issues/5771.

> > There is no point for us to support non-standardized things.
> 
> Oh, I assumed supporting those was why there was an explicit python version check in `setup.py`, was there a different reason for doing it that way?

It's mostly a legacy artifact, pre PEP 517. It could probably be removed now.

jameshilliard:
> I guess it depends on what you need. There's also an option to have an in-tree PEP 517 build backend, for example.

Yeah, it's def not an in-tree backend that we need.

We're basically just cross compiling python packages from source and installing them into a target rootfs, we've def had to carry a good number of downstream patches/hacks to the python infrastructure to make this work though.

We use existing buildroot infrastructure for downloading and dependency resolution(needed for system package dependency resolution that python packages often depend on) so pip I think kinda gets in the way there.

asvetlov:
@webknjaz I agree with the current PR changes.

Please feel free to merge when PR is ready from *your* perspective.

asvetlov (approved):

CHANGES/6205.bugfix
```diff
@@ -0,0 +1 @@
+Added requirement for compatible ``setuptools`` version.
```
webknjaz:
```suggestion
Declared the minimum required version of ``setuptools`` in the :pep:`517`
configuration file -- :user:`jameshilliard`.
```

README.rst
```diff
@@ -163,6 +163,7 @@ Requirements
 ============
 
 - Python >= 3.7
+- setuptools >= 46.4.0
```
webknjaz: This is a list of runtime requirements, don't add build requirements here.

webknjaz:
> > I guess it depends on what you need. There's also an option to have an in-tree PEP 517 build backend, for example.
> 
> Yeah, it's def not an in-tree backend that we need.
> 
> We're basically just cross compiling python packages from source and installing them into a target rootfs, we've def had to carry a good number of downstream patches/hacks to the python infrastructure to make this work though.
> 
> We use existing buildroot infrastructure for downloading and dependency resolution(needed for system package dependency resolution that python packages often depend on) so pip I think kinda gets in the way there.

So maybe you need to be a front-end, then, in terms of PEP 517. It's pretty well-documented how to call the hooks. You could start by exploring how the Fedora's macros work, as an example: https://src.fedoraproject.org/rpms/pyproject-rpm-macros/blob/rawhide/f/pyproject_buildrequires.py.

README.rst
jameshilliard: Hmm, where do you think this should go in the docs?

CHANGES/6205.bugfix
webknjaz: > Isn't that what this does still?:
> 
> ```
> requires = [
>     "setuptools>=46.4.0",
> ]
> ```

TBF, this specifies a dependency to install, it doesn't do any checks :man_shrugging:

README.rst
webknjaz: Honestly, I don't see any need for this to be documented because PEP 517 is standardized, hence it's well-known that the source of truth is in `pyproject.toml`.

pyproject.toml
```diff
@@ -1,3 +1,9 @@
+[build-system]
+requires = [
+    "setuptools>=46.4.0",
```
webknjaz: Nitpick: it reads a bit better with spacing.
```suggestion
    "setuptools >= 46.4.0",
```

README.rst
webknjaz: The only place close enough to mentioning the build process is https://docs.aiohttp.org/en/stable/contributing.html. I don't think we have any other dedicated docs.

jameshilliard: Ok, well I removed it.

CHANGES/6205.bugfix
```diff
@@ -0,0 +1,2 @@
+Declared the minimum required version of ``setuptools`` in the :pep:`517`
```
webknjaz: I just realized that this is marked as a bugfix. It's not a runtime bug. So I'd probably use `misc` in the file name.

jameshilliard: changed to misc

CHANGES/6205.misc
```diff
@@ -0,0 +1,2 @@
+Declared the minimum required version of ``setuptools`` in the :pep:`517`
+configuration file -- :user:`jameshilliard`.
```
webknjaz:
```suggestion
Declared the minimum required version of ``setuptools`` v46.4.0
in the :pep:`517` configuration file -- :user:`jameshilliard`.
```

webknjaz (approved):
Thanks for your contribution! I've set this PR to `auto-merge` meaning that it'll get merged as soon as the CI passes. After that, the backporting bot will pick up the change and create PRs against branches `3.8` and `3.9` so it'll get released with the next v3.8.x.

PR #6238, opened 2021-11-03
Requirements cleanup

asvetlov:
Less aggressive alternative that still uses `Makefile` instead of `pre-commit` hooks


Makefile
```diff
@@ -183,10 +184,11 @@ doc:
 doc-spelling:
 	@make -C docs spelling SPHINXOPTS="-W --keep-going -n -E"
 
+requirements/%.txt: requirements/%.in $(filter-out $(REQIN2),$(REQS))
+	pip-compile --no-header --allow-unsafe -q -o $@ $(@:.txt=.in)
```
webknjaz: Why `--no-header`?

asvetlov: `header` contains Python version, files generated by Python 3.9 based environment are different from 3.10 for example.

Looks weird.

asvetlov:
Hmm, all this exists to prevent long dependency installation time on CI, especially if dependencies are not cached previously.

There is an alternative approach based on the constraints file.
1. All requirement `*.in` files are converted back to plain `txt`.
2. `constraints.in` is added with a content that includes all `*.txt` file
3. `constraints.txt` is generated from `constraints.in` by `pre-compile` run.
4. in GitHub CI, `pip install -r requirements/lint.txt` is replaced with `pip install -r requirements/lint.txt -c requiremments/constraints.txt`
5. new workflow is added to regenerate constraints file on every dependabot commit and push it back into PR if necessary

Currently, dependabot looks on `txt` files only and ignores `.in` templates; this breaks our dependency auto-update system.
Also, `pip-compile` is not auto-applied. We have de-sync between `txt` and `in` files.

By the proposed change, dependabot should keep working, installation is reasonable fast, constraints are regenerated automatically.

asvetlov:
Merging the PR, dependabot post-update part may require additional tuning -- will do it by a separate PR if needed.

PR #6251, opened 2021-11-04
Drop types-chardet mypy dependency

asvetlov:


requirements/constraints.txt
```diff
@@ -230,9 +230,7 @@ trustme==0.9.0 ; platform_machine != "i686"
 typer==0.4.0
     # via python-on-whales
 types-chardet==4.0.1
-    # via
-    #   -r requirements/lint.txt
-    #   -r requirements/test.txt
+    # via -r requirements/test.txt
```
webknjaz, line 233: Why didn't you drop it from `test.txt`, then? Also, I'd expect such PRs to only update `.in` files and re-generate the constraints w/o manually editing them.

asvetlov: Because it *was* in `test.txt`.

Dependabot doesn't look into `.in` files and cannot be configured to do this.
That's why I've replaced all `requirements/*.in` with `.txt`, and added `requirements/constraints.in` with the following context:
```
-r cython.txt
-r dev.txt
-r doc-spelling.txt
-r lint.txt
```

constraints are regenerated on every dependabot PR. 
The installation instruction is `pip install -r requirements/lint.txt  -c requirements/constraints.txt`

Dreamsorcerer: I added it to test.txt. If chardet is not used in the project anymore, it can be removed.

webknjaz: > Dependabot doesn't look into `.in` files and cannot be configured to do this.

This is not entirely true. It does look into them when there are pairs with the same filename: `{{name}}.in` and `{{name}}.txt`. But this only works if both files are committed to Git. And, of course, you can include both `.in` and `.txt` in other `.in` files as needed.

One of my concerns with the current setup is that there's just one lockfile for everything. But the envs are different and should have different constraints. This is because you don't want your test env to influence how the docs are built or how the linters are run (most of the time).

Note that we should have non-pinned direct deps in the `.in` files (with a few exceptions) and have pip-tools (this is what Dependabot actually runs under the hood) figure out the deptree updates.

PR #6256, opened 2021-11-05
Added "strategy" argument to StreamResponse.enable_compression() method

shootkin:
## What do these changes do?

This PR adds "strategy" argument to StreamResponse.enable_compression() method
The end users can significantly speed up compression using this parameter.
Example of usage:
```
async def get_png(request):
    response = web.FileResponse("./content.png")
    response.enable_compression(strategy=zlib.Z_RLE)
    return response
```

## Checklist

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -88,6 +88,7 @@ class StreamResponse(BaseClass, HeadersMixin):
         self._keep_alive = None  # type: Optional[bool]
         self._chunked = False
         self._compression = False
+        self._strategy = None  # type: int
         self._compression_force = None  # type: Optional[ContentCoding]
         self._cookies = SimpleCookie()  # type: SimpleCookie[str]
 
@@ -177,7 +178,9 @@ class StreamResponse(BaseClass, HeadersMixin):
             warnings.warn("Chunk size is deprecated #1615", DeprecationWarning)
 
     def enable_compression(
-        self, force: Optional[Union[bool, ContentCoding]] = None
+        self,
+        force: Optional[Union[bool, ContentCoding]] = None,
+        strategy: int = zlib.Z_DEFAULT_STRATEGY,
     ) -> None:
         """Enables response compression encoding."""
         # Backwards compatibility for when force was a bool <0.17.
@@ -193,6 +196,7 @@ class StreamResponse(BaseClass, HeadersMixin):
 
         self._compression = True
         self._compression_force = force
+        self._strategy = strategy
 
     @property
     def headers(self) -> "CIMultiDict[str]":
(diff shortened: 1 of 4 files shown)


Dreamsorcerer:
We don't merge new features etc. straight into 3.9, you'll need to make a PR to master, which will then be backported to 3.9 (automatically, if we are lucky).

Unless I'm really confused, the PR looks completely broken to me though.

aiohttp/web_response.py
```diff
@@ -88,6 +88,7 @@ def __init__(
         self._keep_alive = None  # type: Optional[bool]
         self._chunked = False
         self._compression = False
+        self._strategy = None  # type: int
```
Dreamsorcerer, line 91: The comments for types is really old, let's use annotations for new code (`self._strategy: int = ...`).

But, you can't assign `None` to a type of `int`, use `Optional[]`, like the other types.

aiohttp/web_response.py
```diff
@@ -398,7 +402,7 @@ async def _do_start_compression(self, coding: ContentCoding) -> None:
         if coding != ContentCoding.identity:
             assert self._payload_writer is not None
             self._headers[hdrs.CONTENT_ENCODING] = coding.value
-            self._payload_writer.enable_compression(coding.value)
+            self._payload_writer.enable_compression(coding.value, self._strategy)
```
Dreamsorcerer, line 405: Type of `self._payload_writer` is `AbstractStreamWriter`, which doesn't support this 2nd parameter:
https://github.com/aio-libs/aiohttp/blob/3.9/aiohttp/abc.py#L184

shootkin:
No problem, I will create another PR to master.
It is not completely broken, you can test it with the code I provided.
When I added these changes to 3.8 and  downloaded .png files with Z_REL
strategy, I got files 2x faster than earlier.

сб, 6 нояб. 2021 г., 00:03 Sam Bull ***@***.***>:

aiohttp/web_response.py
Dreamsorcerer: So, my question is where is the parameter actually used?

shootkin: No, in reality it is the instance of StreamWriter class: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_writer.py#L65
And enable_compression() method of this class supports strategy parameter.
That's why my code works - I tested it, and you can test it as well.

Dreamsorcerer: Regardless of whether it works in your test, nothing can be merged until mypy passes successfully, and currently this will cause a type error.

I'm not too familiar with this code, so I'm not sure if the type can be changed to `StreamWriter`, or this parameter added to `AbstractStreamWriter`, or if we would need an `isinstance(StreamWriter, ...)` check. Maybe @asvetlov can answer that one.

shootkin:
Moved to https://github.com/aio-libs/aiohttp/pull/6257

PR #6260, opened 2021-11-06
Bump typing-extesions in constraints.txt

Dreamsorcerer:
Not sure why dependabot didn't bump it in this file...

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -233,7 +233,7 @@ types-chardet==4.0.1
     # via
     #   -r requirements/lint.txt
     #   -r requirements/test.txt
-typing-extensions==3.7.4.3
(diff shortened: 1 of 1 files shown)


webknjaz:
Don't do this. It'll create more conflicts (because of proxy.py).

After Andrew's recent changes Dependabot got misconfigured with all of the other parts of the automation. This needs to be fixed instead.


Honestly, it needs to be redesigned and have multiple lockfiles per env (type/python version/os/arch) to really work properly. One of us will probably work on it soon'ish. Unless you want to take this task.

Dreamsorcerer:
Then let's go the other way and revert the previous change. But, the required checks need to be re-enabled to stop dependabot breaking it again.

PR #6264, opened 2021-11-07
Switch to called workflows solution

asvetlov:
I want to split `ci.yml` into several top-level entry points, the single github event in each.
These top-level workflows should call sub-workflows if needed for testing, wheels generation, deployment, requirements recompilation etc.


.github/workflows/ci.yml
```diff
@@ -19,168 +19,12 @@ concurrency:
 
 
 jobs:
-
+  # TODO: replace to @master after the PR merging
```
webknjaz: No, not master, just use the local in-repo path.

webknjaz:
I was considering using this feature but while exploring it and experimenting in other places, it's pretty visible that it's full of imperfections. So be careful with this.

.github/workflows/ci.yml
asvetlov: Does it work? 
https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#calling-a-reusable-workflow tells about absolute paths only.
If local paths work -- I'm happy to switch.

Sorry, should go now, return to the PR in the evening.

webknjaz: I haven't experimented with that part yet. But since it's in beta, I'd assume that some of the features are undocumented and need discovery.

webknjaz: OTOH I've been meaning to have these in a separate repository for reuse across projects.

asvetlov: workflow version (`@version`) is mandatory now.

asvetlov:
Maybe local path will be supported eventually: https://github.com/actions/runner/discussions/1419

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+    name: Lint
+    needs: compile-deps
+    uses: aio-libs/aiohttp/.github/workflows/lint.yml@0c8a36f3383a4b27ba2facda47944f314d0cf8c2
+  test:
+    name: Test
+    needs: lint
+    uses: aio-libs/aiohttp/.github/workflows/test.yml@0c8a36f3383a4b27ba2facda47944f314d0cf8c2
+
+  check:  # This job does nothing and is only used for the branch protection
+    needs: test
```
webknjaz: and lint?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,39 @@
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # TODO: replace to @master after the PR merging
+  lint:
+    name: Lint
+    needs: compile-deps
+    uses: aio-libs/aiohttp/.github/workflows/lint.yml@0c8a36f3383a4b27ba2facda47944f314d0cf8c2
+  test:
+    name: Test
+    needs: lint
```
webknjaz: I think this shouldn't be blocked on linting.

asvetlov: Ok, agree

asvetlov:
@webknjaz I've updated workflows.
Please review again if you have time.

asvetlov:
Conflicts are fixed

.github/workflows/cron.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  # No compile-deps on cron job.
+  # TODO: Apply run to all active branches by keeping workflows like cron-38.yml etc.
+
+  lint:
+    name: Lint
+    uses: aio-libs/aiohttp/.github/workflows/sub-lint.yml@master
+    with:
+      ref: ${{ env.ref }}
```
webknjaz: Would something like
```suggestion
      ref: ${{ github.event.ref }}
```
work?

webknjaz: Or maybe
```suggestion
      ref: ${{ github.event.repository.default_branch }}
```

.github/workflows/dependabot.yml
```diff
@@ -0,0 +1,48 @@
...
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  lint:
+    name: Lint
+    needs: compile-deps
+    uses: aio-libs/aiohttp/.github/workflows/sub-lint.yml@master
+  test:
+    name: Test
+    needs: compile-deps
+    uses: aio-libs/aiohttp/.github/workflows/sub-test.yml@master
+  check:  # This job does nothing and is only used for the branch protection
```
webknjaz, line 27: You missed the `needs:` definition. It'll always be green.

.github/workflows/sub-compile-deps.yml
```diff
@@ -48,9 +43,18 @@ jobs:
           echo "::set-output name=changed::true"
         else
           echo "::set-output name=changed::false"
-    - name: Commit and push if needed
+        fi
+    - name: Commit and push if needed\
```
webknjaz:
```suggestion
    - name: Commit and push if needed
```

webknjaz (commented):
Can you pass artifacts between reusable workflows? I wanted to implement proper testing of the artifacts as opposed to testing just a Git checkout and this would be necessary to make that happen.

asvetlov:
> Can you pass artifacts between reusable workflows? 

Why not? An artifact is bound to workflow run number, sub-workflows share it as I see.

webknjaz:
> > Can you pass artifacts between reusable workflows?
> 
> Why not? An artifact is bound to workflow run number, sub-workflows share it as I see.

Yes, I think they artificially inject the jobs into the calling workflow. I just haven't attempted testing some of the possible corner cases, that's why I'm asking.

webknjaz (commented):
You need to apply https://github.com/aio-libs/aiohttp/pull/6369 here.

.github/workflows/cron.yml
asvetlov: Thanks!
I think the `master` constant should work here.
Not existing yet `cron-39.yml` file will contain `'3.9'` instead

.github/workflows/dependabot.yml
asvetlov: Good point!

.github/workflows/sub-compile-deps.yml
asvetlov: Done

asvetlov:
Notes are fixed

Dreamsorcerer:
There's no way this PR will still be good over 2 years later, if it's still wanted, it'll probably be easier to begin anew.

webknjaz:
@Dreamsorcerer i have a vision to make something similar, but haven't gotten to it

PR #6286, opened 2021-11-10
Proposal for configuration of ceiling timeout value threshold

hlecnt:
## What do these changes do?

Proposal to restore former aiohttp version (<3.7) regarding the timeout handling.
More details are available here [this discussion](https://github.com/aio-libs/aiohttp/discussions/6008)

By default, no behavior changes compared to aiohttp versions >=3.7

## Related issue number

No issue tracked.

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -74,6 +74,12 @@ _SENTINEL = NewType("_SENTINEL", object)
 sentinel: _SENTINEL = _SENTINEL(object())
 NO_EXTENSIONS = bool(os.environ.get("AIOHTTP_NO_EXTENSIONS"))  # type: bool
 
+CEIL_TIMEOUT_THRESHOLD: float = 5
+try:
+    CEIL_TIMEOUT_THRESHOLD = float(os.environ.get("AIOHTTP_CEIL_TIMEOUT_THRESHOLD", 5))
+except ValueError:
+    pass
+
 # N.B. sys.flags.dev_mode is available on Python 3.7+, use getattr
 # for compatibility with older versions
 DEBUG = getattr(sys.flags, "dev_mode", False) or (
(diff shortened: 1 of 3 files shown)


aiohttp/helpers.py
```diff
@@ -74,6 +74,12 @@
 sentinel: _SENTINEL = _SENTINEL(object())
 NO_EXTENSIONS = bool(os.environ.get("AIOHTTP_NO_EXTENSIONS"))  # type: bool
 
+CEIL_TIMEOUT_THRESHOLD: float = 5
+try:
+    CEIL_TIMEOUT_THRESHOLD = float(os.environ.get("AIOHTTP_CEIL_TIMEOUT_THRESHOLD", 5))
```
Dreamsorcerer, line 79: I don't think we want to configure anything with environment variables in the core library. That's the domain of applications.

hlecnt: As the cause of the performance drop was introduced to achieve tests faster in a testing environment (different of production environment), I supposed that an environment variable would be acceptable as a global parameter. Performing tests faster should be done for sure, modifying the default core behavior for that ... Im' not sure. 
Nevertheless I understand your requirements and I'll try to find another way to do it.

Functions handling timeouts are present in the `helpers` module and don't know anything about the app and its context. Adding app level configuration will need more work.  

Maybe you have some tricks or ideas to easily pass app context info to the `helpers` module functions ?

asvetlov: I concur with @Dreamsorcerer 
We use env wars for disabling C Extensions only IIRC.
Env wars are handy for application development (and I use them on my job extensively, sure).
But a library should configure everything explicitly.
Yes, it requires much more work but this requirement is one of the aiohttp design rules.
Other libraries can have other rules, that's fine. Every software engineer is free to choose tools that one wants to use.

There is no trick to pass `ClientSession` settings down to `helpers`. Maybe [contextvars](https://docs.python.org/3/library/contextvars.html) can be used for such *global-like* things but I strongly ask for using old-good explicit arguments passing in aiohttp.
Please leave the *implicitness* to aiohttp *users*, not to the library itself.
Thanks for the understanding.

hlecnt: Thanks for your answer. 

I'm working on a second version using old-good explicit arguments passing in aiohttp. When finished, I'll close this draft PR and I'll create a new one.

asvetlov: Cool! Thanks!

PR #6313, opened 2021-11-15
docs: Fix statement about `setUpAsync()` and `tearDownAsync()`

H--o-l:
Since https://github.com/aio-libs/aiohttp/pull/4732, it's wrong to says that
`setUpAsync()` and `tearDownAsync()` do nothing and can be overridden without
calling `super.setUpAsync()` and `super.tearDownAsync()`.

## What do these changes do?

Fix documentation

No

## Related issue number

None

## Checklist

Only docs.

----

Remark:
My rst preview wasn't working, I hope there is no rst mistakes.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -369,17 +369,33 @@ functionality, the AioHTTPTestCase is provided::
 
     .. comethod:: setUpAsync()
 
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``setUp`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``setUp`` stage of the ``TestCase``::
+
+           async def setUpAsync(self):
+               await super().setUpAsync()
+               await foo()
 
        .. versionadded:: 2.3
+
+       .. versionchanged:: 3.8
+
+          ``await super().setUpAsync()`` call is required.
 
     .. comethod:: tearDownAsync()
 
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``tearDown`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``tearDown`` stage of the ``TestCase``::
+
+           async def tearDownAsync(self):
+               await super().tearDownAsync()
+               await foo()
 
        .. versionadded:: 2.3
+
+       .. versionchanged:: 3.8
+
+          ``await super().tearDownAsync()`` call is required.
 
     .. method:: setUp()
 


docs/testing.rst
```diff
@@ -369,15 +369,23 @@ functionality, the AioHTTPTestCase is provided::
 
     .. comethod:: setUpAsync()
 
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``setUp`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``setUp`` stage of the ``TestCase``::
+
+           async def setUpAsync(self)
```
Dreamsorcerer:
```suggestion
           async def setUpAsync(self):
```

docs/testing.rst
```diff
@@ -369,15 +369,23 @@ functionality, the AioHTTPTestCase is provided::
...
        .. versionadded:: 2.3
 
     .. comethod:: tearDownAsync()
 
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``tearDown`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``tearDown`` stage of the ``TestCase``::
+
+           async def tearDownAsync(self)
```
Dreamsorcerer:
```suggestion
           async def tearDownAsync(self):
```

Dreamsorcerer:
Looks like a backwards-incompatible change we made by mistake.
We should probably point this out in the change fragment.

docs/testing.rst
H--o-l: Thanks for catching this. I fixed it.

docs/testing.rst
H--o-l: Thanks for catching this. I fixed it.

H--o-l:
> Looks like a backwards-incompatible change we made by mistake.

I agree.

> We should probably point this out in the change fragment.

It's something I should do in this PR?

Thanks @Dreamsorcerer for your first review!

H--o-l:
Open question, maybe
```py
           async def tearDownAsync(self):
               await foo()
               await super().tearDownAsync()
```
Would be more suited than
```py
           async def tearDownAsync(self):
               await super().tearDownAsync()
               await foo()
```
?

docs/testing.rst
```diff
@@ -369,15 +369,23 @@ functionality, the AioHTTPTestCase is provided::
...
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``setUp`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``setUp`` stage of the ``TestCase``::
+
+           async def setUpAsync(self):
+               await super().setUpAsync()
+               await foo()
 
        .. versionadded:: 2.3
```
asvetlov, line 379:
```suggestion
       .. versionadded:: 2.3
       
       .. versionchanged:: 3.8
       
          ``await super().setUpAsync()`` call is required.
```

docs/testing.rst
```diff
@@ -369,15 +369,23 @@ functionality, the AioHTTPTestCase is provided::
...
-       This async method do nothing by default and can be overridden to execute
-       asynchronous code during the ``tearDown`` stage of the ``TestCase``.
+       This async method can be overridden to execute asynchronous code during
+       the ``tearDown`` stage of the ``TestCase``::
+
+           async def tearDownAsync(self):
+               await super().tearDownAsync()
+               await foo()
 
        .. versionadded:: 2.3
```
asvetlov, line 394:
```suggestion
       .. versionadded:: 2.3
       
       .. versionchanged:: 3.8
       
          `await super().setUpAsync()` call is required.
```

docs/testing.rst
H--o-l: Thanks, done!

docs/testing.rst
H--o-l: Thanks, done! (with `super().tearDownAsync()`, not `super().setUpAsync()`)

asvetlov (approved):
LGTM, thanks!

PR #6321, opened 2021-11-17
Ignore protocol exceptions after it is closed

### LINKED ISSUE #4526: "application data after close notify on" 4.0.0a1
See: https://github.com/aio-libs/aiohttp/issues/3477

I'm experiencing this issue on the 4.0.0a1 release.

It can be reproduced with this script (Thanks @Cadair):
```python
import asyncio
import aiohttp

print(aiohttp.__version__)

url = "https://github.com/sunpy/sample-data/blob/master/sunpy/v1/AIA20110607_063303_0335_lowres.fits?raw=true"


async def run_example():
            print(resp)

asyncio.run(run_example())
```

Python version: 3.7.3, on Linux (WSL).

greshilov:
## Issue

I've recently started to experience strange ssl exceptions with `aiohttp 4+` (master branch) with the following message:
```
Future exception was never retrieved
future: <Future finished exception=SSLError(1, '[SSL: APPLICATION_DATA_AFTER_CLOSE_NOTIFY] application data after close notify (_ssl.c:2756)') created at /usr/lib/python3.9/asyncio/base_events.py:424>
source_traceback: 
 ...
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1081, in create_connection
    transport, protocol = await self._create_connection_transport(
  File "/usr/lib/python3.9/asyncio/base_events.py", line 1099, in _create_connection_transport
    protocol = protocol_factory()
  File "/home/s/python/aiohttp/aiohttp/client_proto.py", line 39, in __init__
    self.closed = self._loop.create_future()  # type: asyncio.Future[None]
  File "/usr/lib/python3.9/asyncio/base_events.py", line 424, in create_future
    return futures.Future(loop=self)
  File "/usr/lib/python3.9/asyncio/sslproto.py", line 528, in data_received
    ssldata, appdata = self._sslpipe.feed_ssldata(data)
  File "/usr/lib/python3.9/asyncio/sslproto.py", line 206, in feed_ssldata
    self._sslobj.unwrap()
  File "/usr/lib/python3.9/ssl.py", line 948, in unwrap
    return self._sslobj.shutdown()
ssl.SSLError: [SSL: APPLICATION_DATA_AFTER_CLOSE_NOTIFY] application data after close notify (_ssl.c:2756
```
## Reproducer
Code to reproduce the error.
```
import asyncio
import aiohttp

async def main():

        async with session.get(f'http://drive.google.com') as resp:
            print(resp.status)

if __name__ == '__main__':
    asyncio.run(main(), debug=True)
```

## Description

After some digging I found out a related issue in python bug tracker:
https://bugs.python.org/issue39951

It describes possible scenario for this type of SSL exception to happen:
* client closes the ssl connection and sends `close notify`.
* however server sends some data before receiving that `close notify`.
* client receives the data, but the connection is already closed, thus SSL exception.

In `aiohttp` this exception happens when `BaseConnector._release` is called with `should_close=True`
https://github.com/aio-libs/aiohttp/blob/d149eff62e1e110d2eafac8a79219dd2e56b8581/aiohttp/connector.py#L628-L633

After execution of this branch `protocol` is on its way to be removed by GC. But if the `protocol.connection_lost` is called with exception **before** the removal (which is exactly what is happening) the `protocol.closed` future will store this exception.
https://github.com/aio-libs/aiohttp/blob/d149eff62e1e110d2eafac8a79219dd2e56b8581/aiohttp/client_proto.py#L73-L79

During the removal of the `protocol` object, the following exception will be produced: `Future exception was never retrieved`.

## Solution
I solved it by setting the result of `self.closed` to `None` explicitly during the `_release` call with `should_close=True`.

## Related issue number

Fixes #4526 

## Checklist

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -45,7 +45,7 @@ from .client_exceptions import (
 )
 from .client_proto import ResponseHandler
 from .client_reqrep import SSL_ALLOWED_TYPES, ClientRequest, Fingerprint
-from .helpers import _SENTINEL, ceil_timeout, is_ip_address, sentinel
+from .helpers import _SENTINEL, ceil_timeout, is_ip_address, sentinel, set_result
 from .http import RESPONSES
 from .locks import EventResultOrError
 from .resolver import DefaultResolver
@@ -628,6 +628,9 @@ class BaseConnector:
         if should_close or protocol.should_close:
             transport = protocol.transport
             protocol.close()
+            # Ignore the following protocol exceptions
+            # See PR #6321
+            set_result(protocol.closed, None)
 
             if key.is_ssl and not self._cleanup_closed_disabled:
                 self._cleanup_closed_transports.append(transport)
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -628,7 +628,7 @@ class BaseConnector:
         if should_close or protocol.should_close:
             transport = protocol.transport
             protocol.close()
-            # Ignore the following protocol exceptions
+            # TODO: Remove once fixed: https://bugs.python.org/issue39951
             # See PR #6321
             set_result(protocol.closed, None)
 
--- a/CHANGES/4526.bugfix
+++ b/CHANGES/4526.bugfix
@@ -1,0 +1,1 @@
+Ignore protocol exceptions after it is closed.
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -508,6 +508,14 @@ async def test_release_close(key: Any) -> None:
     assert proto.close.called
 
 
+async def test_release_proto_closed_future(loop: Any, key: Any):
+    conn = aiohttp.BaseConnector()
+    protocol = mock.Mock(should_close=True, closed=loop.create_future())
+    conn._release(key, protocol)
+    # See PR #6321
+    assert protocol.closed.result() is None
+
+
 async def test__drop_acquire_per_host1(loop: Any) -> None:
     conn = aiohttp.BaseConnector()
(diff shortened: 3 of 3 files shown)


aiohttp/connector.py
```diff
@@ -628,6 +628,9 @@ def _release(
         if should_close or protocol.should_close:
             transport = protocol.transport
             protocol.close()
+            # Ignore the following protocol exceptions
```
Dreamsorcerer: This should only happen because of that bug, right? So, we can remove it once the bug is fixed.

```suggestion
            # TODO: Remove once fixed: https://bugs.python.org/issue39951
```

Dreamsorcerer (approved):
Confirmed that this fixes #4526

aiohttp/connector.py
greshilov: I hope so! I didn't fully understand from that discussion (bpo-39951) whether it's a bug or a lack of `ssl` module functionality. Christian Heimes said:

> Typical HTTP clients like curl send a close_notify and then shut down the TCP connection. HTTP servers may not reply with close_notify or may not wait for the client to confirm the server-side close notify.

And then:
> Python's ssl module does not support one-way close yet. The unwrap() methods always performs a blocking two-way shutdown. It requires cooperation from both parties.

Seems it might be solved by implementing `one-way` close `unwrap` in `ssl` module and then by adding corresponding functionality to `asyncio.sslproto.SSLProtocol`.

asvetlov (approved):

asvetlov:
Backporting doesn't make sense, `protocol` has no `closed` property

PR #6329, opened 2021-11-20
Replace sentinel definition from object() with enum

asvetlov:
The recipe is borrowed from https://discuss.python.org/t/pep-661-sentinel-values/9126/83


asvetlov:
Backported by 57cbc3ed

PR #6345, opened 2021-11-24
Use str.partition() instead of .split() with a limit and a check

wbolster:
It's shorter, more idiomatic, and doesn't need noisy type casting.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
         twine check --strict dist/*
     - name: Making sure that CONTRIBUTORS.txt remains sorted
       run: |
-        LC_ALL=C sort -c CONTRIBUTORS.txt
+        LC_ALL=C sort --check --ignore-case CONTRIBUTORS.txt
 
   gen_llhttp:
     name: Generate llhttp sources
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,7 @@ repos:
       ^requirements/constraints[.]txt$
   - id: trailing-whitespace
   - id: file-contents-sorter
+    args: ['--ignore-case']
     files: |
       CONTRIBUTORS.txt|
       docs/spelling_wordlist.txt|
--- a/CHANGES/6345.misc
+++ b/CHANGES/6345.misc
@@ -1,0 +1,1 @@
+Use str.partition() instead of .split() with a limit and a check.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -331,6 +331,7 @@ Willem de Groot
 William Grzybowski
 William S.
 Wilson Ong
+wouter bolsterlee
 Yang Zhou
 Yannick Koechlin
 Yannick Péroux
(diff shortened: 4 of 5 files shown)


wbolster:
_update_: see next comment

re `CONTRIBUTORS.rst`: my name uses only lowercase characters

the pre-commit `file-contents-sorter` is not case insensitive, and the CI checks are not case insensitive (and using `sort -f` there would make it clash with pre-commit)

wbolster:
i've tweaked the pre-commit config and CI check to sort contributors in a case-insensitive way

Dreamsorcerer:
> my name uses only lowercase characters

Interesting, I'm curious what language/locale this comes from that case matters?

In English, names are case insensitive and in most situations will be capitalised, as in this file. For example, many written forms require you to fill them in block capitals (so, I'd write my name as SAM, and it would likely end up being displayed as Sam when someone views the name), there is no differentiation between case. I think many systems in wide use in English-speaking countries would not be able to handle your name correctly.

wbolster:
hi @Dreamsorcerer 

> > my name uses only lowercase characters

> Interesting, I'm curious what language/locale this comes from that case matters?

heh, glad you ask 😬

tl;dr: lowercasing my name matters to me, and since it is my name, i get to decide. in general, names have no case, but that's very different from ‘case does not matter’. 

i've personally not used capital letters to spell my name (or most prose for that matter) for many years now, in both official and informal settings, and since name capitalisation is not codified in (dutch) law, this means it's essentially up to the bearer of the name. while others sometimes do use capitals, it is not my preferred spelling of my (!) name and hence i will correct it if possible. i guess you would do the same if i were to refer to you as ‘saM bUlL’ (ok, contrived) or ‘BULL sam’ (less contrived than you may think; see below).

some background, since you're asking 🙃

names are generally considered to be case insensitive indeed. in fact, many languages don't even have the concept of letter case. in reality, billions of people have names that do not have any case at all, e.g. in chinese, in arabic, etc.

of course there are linguistic conventions, but those are merely that, and highly language specific. as an example, dutch conventions are ‘Guido van Rossum’ and ‘G. van Rossum’, but without the given name it becomes ‘Van Rossum’, and that will be abbreviated to ‘v. Rossum’, never as ‘V. Rossum’, and it's always sorted with ‘r’, while in the dutch speaking part of belgium it would be ‘Guido Van Rossum’ instead and historically this could even have transformed into ‘Vanrossum’ in belgium, sorted with ‘v’, but never in the netherlands, roughly speaking, and this goes back to before these countries even existed; [details](https://nl.wikipedia.org/wiki/Tussenvoegsel#Hoofdletters). meanwhile, [in france](https://www.quora.com/Why-do-French-people-write-their-surnames-in-capital-letters) it's customary to use ALL CAPS for part of a name, but only in some situations, while in japan, many people spell part of the name in ALL CAPS, but only in its romanized form.

here's a quite popular article that will undoubtedly surprise you at least once: [falsehoods programmers believe about names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)

here are some ‘unconventional’ examples for anglophones:

- dana boyd: [wikipedia](https://en.wikipedia.org/wiki/danah_boyd), [own website](https://www.danah.org/name.html)
- bell hooks: [wikipedia](https://en.wikipedia.org/wiki/bell_hooks)
- Jerald terHorst: [wikipedia](https://en.wikipedia.org/wiki/Jerald_terHorst)

PR #6374, opened 2021-12-04
Revert assert

Dreamsorcerer:
@asvetlov added this assert, but we have a test which explicitly says this should work. So, this would be a backwards-incompatible change.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -136,6 +136,7 @@ filterwarnings =
     # Temporarily ignore warnings internal to Python 3.9.7, can be removed again in 3.9.8.
     ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning:asyncio
     ignore:Creating a LegacyVersion has been deprecated and will be removed in the next major release:DeprecationWarning::
+    ignore::ResourceWarning
 junit_suite_name = aiohttp_test_suite
 norecursedirs = dist docs build .tox .eggs
 minversion = 3.8.2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
           - os: windows
             no-extensions: 'Y'
         include:
-          - pyver: pypy3
+          - pyver: pypy-3
             no-extensions: 'Y'
             os: ubuntu
       fail-fast: true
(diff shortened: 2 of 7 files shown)


asvetlov:
Backport error I guess, my wrong.
I'm curious, what is the real type?
If it is `None`, maybe adding `None` type to the union solves the issue.

Dreamsorcerer:
> If it is `None`, maybe adding `None` type to the union solves the issue.

No, I can do an `or x is None` if you'd like to keep the assert.

We also have another problem with the CI though, with the Pypy version being 3.6, which we no longer support.

asvetlov (approved):
LGTM.
I can live without the assert, not a big deal.

Dreamsorcerer:
Seems like there are more issues with pypy...

asvetlov:
Initialized some `ClientSession` attrs early to allow `__del__` destroy not fully built object by 63217fc.

asvetlov:
Finally, everything is green.
Merging.
@Dreamsorcerer thank you very much for your help!

setup.cfg
```diff
@@ -136,6 +136,7 @@ filterwarnings =
     # Temporarily ignore warnings internal to Python 3.9.7, can be removed again in 3.9.8.
     ignore:The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.:DeprecationWarning:asyncio
     ignore:Creating a LegacyVersion has been deprecated and will be removed in the next major release:DeprecationWarning::
+    ignore::ResourceWarning
```
webknjaz, line 139: This lacks a comment on why this is necessary and under what conditions it can be removed.

webknjaz:
This has been partially ported to 3.8 as https://github.com/aio-libs/aiohttp/commit/78be64bd7e1f3961a4bab75434a997d21a30dcc9.

setup.cfg
webknjaz: This has now been removed.

webknjaz:
> This has been partially ported to 3.8 as [78be64b](https://github.com/aio-libs/aiohttp/commit/78be64bd7e1f3961a4bab75434a997d21a30dcc9).

And https://github.com/aio-libs/aiohttp/commit/6243204a6a6a0e5ff84ac754218381b44a841e72

PR #6408, opened 2021-12-10
Retrieve Future exception and break circular reference to fix memory leaks

kfur:
## What do these changes do?

Manly it fix memory leaks due to circular reference and ignoring exception in Future
for clients when exception occurred like `ClientPayloadError`

No

## Related issue number

Maybe some about memory leaks, need testing...

## Checklist


asvetlov:
Looking at the PR changes, I'm not convinced that the memory leak exists.
Could you provide proof?
is the leak even not GC collectible?

kfur:
Main problem here is that `Exception` store `traceback` while `traceback` store reference to it origin so `StreamReader._exception` refer to `Exception` that refer to `traceback` that refer to `StreamReader ` , so loop successfully created.

Here is simple example:
```
import gc

gc.set_debug(gc.DEBUG_UNCOLLECTABLE | gc.DEBUG_LEAK)

class A:
    exc = None
    def run_exception(self):
        raise self.exc

i = A()
i.exc = Exception('foo')
try:
    i.run_exception()
except:
    pass

del i
gc.collect()
print(gc.garbage)
```

You will see that `del i` has no effect and `A()` instance still in memory due to circular reference.
Here is backrefs plot:
a


The same problem in the `aiohttp` code, we don't need to store `StreamReader._exception` because we can retrieve it from `Future`, by the way, as i understand, exception in `Future` doesn't used at this moment.

Here is example with `aiohttp` code:
Just install `objgraph`, then create empty file `dd if=/dev/zero of=data.bin bs=1m count=500`,
and serve it with `python3 -m http.server` to simulate `ClientPayloadError`.
Run code for several seconds and stop `http.server` and you will get `StreamReader00.png`
```

import asyncio
import aiohttp
import sys
import gc
import objgraph
import logging
from aiohttp import  ClientPayloadError


async def test():
    async with aiohttp.ClientSession() as ses:
        async with ses.get('http://127.0.0.1:8000/data.bin') as resp:
            for _ in range(200):
                a = await resp.content.read(512 *1024)
                await asyncio.sleep(1)


async def main():
    try:
        await test()
    except ClientPayloadError as e:
        logging.exception(e)
    gc.collect()
    print(f"Garbage len {len(gc.garbage)}")
    for o in gc.garbage:
        if type(o) is aiohttp.streams.StreamReader:
            print(id(o))
            objgraph.show_backrefs(o, max_depth=10, filename=f"StreamReader00.png")
            break
    sys.stdout.flush()

gc.set_debug(gc.DEBUG_UNCOLLECTABLE | gc.DEBUG_LEAK)
asyncio.get_event_loop_policy().get_event_loop().run_until_complete(main())
```

Here what i got(extracted only significant part, you can see that `StreamReader` and thus `ClientResponse` still eat memory):
aiml

aiohttp/streams.py
```diff
@@ -307,11 +310,14 @@ async def _wait(self, func_name: str) -> None:
...
-                    await waiter
+                    await asyncio.wait((waiter,))
             else:
-                await waiter
+                await asyncio.wait((waiter,))
         finally:
             self._waiter = None
+            fut_exc = waiter.exception()
+            if fut_exc:
+                raise fut_exc
```
asvetlov: How is new version behavior different from original source?

aiohttp/streams.py
```diff
@@ -321,9 +327,6 @@ async def readuntil(self, separator: bytes = b"\n") -> bytes:
         if seplen == 0:
             raise ValueError("Separator should be at least one-byte string")
 
-        if self._exception is not None:
-            raise self._exception
```
asvetlov, line 325: I think you cannot just drop the check here and below but should lookup for `self._waiter` has an exception and re-raise it if present.

asvetlov (changes requested):
The second iteration looks better but still needs some polishing.

Also, please add CHANGES/6408.bugfix file

aiohttp/streams.py
kfur: In original code `waiter.exception()` always empty, replacing `await waiter` with `await asyncio.wait((waiter,))` fix it, honestly i don't know why, i will try to read docs about `Future` more

aiohttp/streams.py
kfur: Yep, but if we leave `self._exception` circular reference will be still exist

kfur: `weakref` can help, but as i understood it doesn't work well with `Exception`, because we can't raise `weakref`

kfur:
And `DataQueue` in `streams.py` has the same architecture, by the way.
Should i fix it here or create another PR?

aiohttp/streams.py
kfur: ok, will fix it

kfur: Done

kfur:
It seems that i misunderstood what `gc.garbage` contain and even with circular reference objects are freed, but original code show more objects in `gc.garbage` than with this PR, so i think applying this changes have a sense

aiohttp/streams.py
```diff
@@ -143,18 +142,22 @@ def __repr__(self) -> str:
             info.append("low=%d high=%d" % (self._low_water, self._high_water))
         if self._waiter:
             info.append("w=%r" % self._waiter)
-        if self._exception:
-            info.append("e=%r" % self._exception)
+        if self._waiter:
```
Dreamsorcerer: This is the same condition as 2 lines above..

aiohttp/streams.py
Dreamsorcerer: `waiter.exception()` was not called in the original code. It would always be `None` because an exception would already have been raised and would never reach this code.

These changes appear to do the same thing as the original code (wait for a future and raise any exceptions from it), but also break the code if `waiter` is not completed (due to the timeout), at which point it raises `InvalidStateError` as can be seen in the failing tests (in the previous test run, Github seems to be failing to run pytest on the latest commit).

aiohttp/streams.py
```diff
@@ -143,18 +142,22 @@ def __repr__(self) -> str:
...
+            wait_exc = self._waiter.exception()
```
Dreamsorcerer: This...

aiohttp/streams.py
```diff
@@ -143,18 +142,22 @@ def __repr__(self) -> str:
...
+                info.append("e=%r" % wait_exc)
         return "<%s>" % " ".join(info)
 
     def get_read_buffer_limits(self) -> Tuple[int, int]:
         return (self._low_water, self._high_water)
 
     def exception(self) -> Optional[BaseException]:
-        return self._exception
+        if self._waiter is not None:
+            return self._waiter.exception()
```
Dreamsorcerer: ...and this, also probably have issues with `InvalidStateError`.

Dreamsorcerer (commented):
Unless we can demonstrate a real benefit, I'm inclined to avoid these changes. They currently break things due to not fully understanding the changes, and I'm not seeing any benefit from them.

aiohttp/streams.py
kfur: i added calling `waiter.exception()` and raising exception from `waiter` also removed  `raise self._exception` from `StreamReader.read` and can't catch any exception

kfur: So as i can see `set_exception` here https://github.com/aio-libs/aiohttp/blob/15941aa1d941716d6fefe67f19369139bc3d0843/aiohttp/streams.py#L163
have no sense, because it never retrieved, at least for `ClientPayloadError`, because `await waiter` doesn't raise it, so i think thats why `StreamReader._exception` was added

aiohttp/streams.py
kfur: Fixed

aiohttp/streams.py
kfur: Fixed

asvetlov:
@kfur blames that `reader._exception` makes a circular reference.
It may be true, but not fatal: the next GC collection should break the loop.
But we should keep the reference also in `reader._waiter._exception` which makes another reference.

Maybe `objgraph` is not aware of this second reference because it is implemented by C-written `asyncio.Future` by default but it doesn't change anything; the reference does exist and it is on the same level as direct `reader._exception` ref.

I doubt if something could be improved here.

kfur:
Ok, but for what we need `StreamReader._exception` if `StreamReader. _waiter` store the same exception ?

asvetlov:
It exists already for 7 years at least :)
I agree, it looks redundant a little.

asvetlov:
I'm curious how is `self.exception()` call slower than `self._exception` attribute lookup.
The performance can win the memory footprint in this contest.

PR #6410, opened 2021-12-12
Convert type comments to type annotations

asvetlov:


aiohttp/client_reqrep.py
```diff
@@ -654,12 +652,12 @@ class ClientResponse(HeadersMixin):
```
Dreamsorcerer, line 652:
```suggestion

# Some of these attributes are None when created, but will be set by the start() method.
# As the end user will likely never see the None values, we cheat the types below.
```

I don't see a better solution, so probably just add a comment like this.

PR #6445, opened 2021-12-20
Fix content-length header being set before prepare hooks

### LINKED ISSUE #6443: Content-Length header not set correctly after on_response_prepare
### Describe the bug

I'm struggling to figure out exactly what is happening here, but in aiohttp-devtools we append a script to the end of the body using on_response_prepare: https://github.com/aio-libs/aiohttp-devtools/blob/master/aiohttp_devtools/runserver/serve.py#L52

I can print out `response.body` on the next line and get the expected output, something ending with:
```
</html>
<script src="http://10.0.3.106:8001/livereload.js"></script>
```

But, the output received by the client is without the modification, i.e. curl output ends with:
```
</html>⏎ 
```

### To Reproduce

Run aiohttp-devtools master and observe that there is no script tag included in the output.

### aiohttp Version

3.8

Dreamsorcerer:
Fixes #6443.

Move creation of default Content-Length headers to immediately before writing the headers out.


Dreamsorcerer:
@asvetlov Let me know if this looks right to you and I'll finish it off.

Test failures appear to be caused by the order of headers changing, so tests can easily be updated. Plus 1 test failure caused by an unconditional `del`: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_response.py#L400

dzhalaevd:
Hi, @Dreamsorcerer, is this issue still relevant?

Dreamsorcerer:
Closed in favour of #7283.

Issue #6446, opened 2021-12-21
Error when handling HTTP requests with HTTP2 related "Upgrade: h2c" in header


pwuertz:
### Describe the bug

I'm experiencing problems when trying to handle HTTP requests sent by Qt6 clients. It looks like Qt now sends additional HTTP2 related headers by default. HTTP servers based on `aiohttp 3.6` and python `http.server` are handling those requests just fine, but `aiohttp 3.7` and `3.8` are showing irregularities in the parsed headers, are unable to read request bodies, and fail to send replies.

Test code and examples for good/bad observations are attached to this issue.

### To Reproduce

Qt6 / Qt5 client test code:
```c++
#include <QtCore/QCoreApplication>
#include <QtNetwork/QNetworkAccessManager>
#include <QtNetwork/QNetworkReply>


int main(int argc, char *argv[])
{
  QCoreApplication a(argc, argv);
  QNetworkAccessManager qnam;

  QUrl requestUrl("http://localhost:9000/");
  QNetworkRequest request(requestUrl);
  // --- WORKAROUND: aiohttp only processes Qt6 requests with Http2Allowed set to false ---
  // request.setAttribute(QNetworkRequest::Attribute::Http2AllowedAttribute, false);
  request.setHeader(QNetworkRequest::ContentTypeHeader, "text/plain;charset=UTF-8");
  QByteArray queryData("Hello World");
  auto* reply = qnam.post(request, queryData);

  QObject::connect(reply, &QNetworkReply::finished, [&]() {
    if(reply->error() == QNetworkReply::NoError) {
      QString contents = QString::fromUtf8(reply->readAll());
      qDebug() << "Reply Ok" << contents;
    }
    else {
      QString err = reply->errorString();
      qDebug() << "Reply Error" << err;
    }
    reply->deleteLater();
    a.quit();
  });

  return a.exec();
}
```

Aiohttp test server:
```python
from aiohttp import web

async def post_handler(request: web.Request):
    body = await request.read()
    n_expected, n_read = request.headers.get('Content-Length'), len(body)
    print(f"Received request, body expected {n_expected}, body read {n_read}")
    headers = "\n".join(f"  {k}: {v}" for k, v in request.headers.items())
    print(f"Headers\n{headers}")
    return web.Response(body=body)

app = web.Application()
app.add_routes([web.post('/', post_handler)])
web.run_app(app, port=9000)
```

### Expected behavior

aiohttp server should be able to handle the requests.

[...]

PR #6451, opened 2021-12-23
Regression test for handling unsupported Upgrade requests (issue #6446)

pwuertz:
## What do these changes do?

`aiohttp` server test for handling client requests with unsupported Upgrade headers (issue #6446).

### DIFF
--- a/tests/test_web_server.py
+++ b/tests/test_web_server.py
@@ -3,7 +3,7 @@ from unittest import mock
 
 import pytest
 
-from aiohttp import client, web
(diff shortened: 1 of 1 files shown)


webknjaz:
I've added the xfail marker right from the browser so let's see if I didn't make any mistakes.

webknjaz:
Cherry-picked to `3.9` and `master` as https://github.com/aio-libs/aiohttp/pull/6455 and https://github.com/aio-libs/aiohttp/pull/6456.

PR #6452, opened 2021-12-23
Upgrade mypy to 0.930

asvetlov:


Dreamsorcerer:
@asvetlov I was already fixing this on #6423.
Just wanted to know what this `asyncio.AbstractServer.sockets` attribute is about. You've just `type: ignore`d it here, but I can find no evidence that these attributes actually exist...

aiohttp/test_utils.py
```diff
@@ -133,7 +133,7 @@ async def start_server(self, **kwargs: Any) -> None:
         await site.start()
         server = site._server
         assert server is not None
-        sockets = server.sockets
+        sockets = server.sockets  # type: ignore[attr-defined]
```
Dreamsorcerer, line 136: I still think there is something wrong here, I can't see where this attribute is coming from. But, otherwise, this is ready to merge.

asvetlov: It's from the [asyncio.Server](https://github.com/python/cpython/blob/main/Lib/asyncio/base_events.py#L321) class.

asvetlov:
@Dreamsorcerer Sorry.
I saw your PR while creating the came change.

aiohttp/test_utils.py
asvetlov: Maybe AbstactServer should have `sockets` abstract property.
It is a part of de-facto public API.

asvetlov:
Thank for your help, @Dreamsorcerer

aiohttp/test_utils.py
Dreamsorcerer: Ah, I didn't see the property method. Yes, if you'd like to make it part of `AbstractServer`, that would work. Otherwise, we can change the type to `asyncio.Server`: https://github.com/aio-libs/aiohttp/blob/d4a2ee6c1fab60f0c86d83aefd516f5d56395353/aiohttp/web_runner.py#L62

PR #6458, opened 2021-12-26
Speeding up PlainResource's (_PlainResourceGroup)

aamalev:
## What do these changes do?

Reduces the complexity of resolving routes.
Resolve route on router with many PlainResource's in some cases O(n) -> O(1)

Improved detection of invalid and unreachable routes.

## Checklist

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1120,15 +1120,16 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         if path and not path.startswith("/"):
             raise ValueError("path should be started with / or be empty")
         # Reuse named resource
-        resource = self._named_resources.get(name)
-        if resource is None:
-            pass
-        elif resource.raw_match(path):
-            return cast(Resource, resource)
-        else:
-            raise ValueError(
-                f"Duplicate {name!r}, already handled by {resource!r}"
-            )
+        if name:
+            resource = self._named_resources.get(name)
+            if resource is None:
+                pass
+            elif resource.raw_match(path):
+                return cast(Resource, resource)
+            else:
+                raise ValueError(
+                    f"Duplicate {name!r}, already handled by {resource!r}"
+                )
         if not ("{" in path or "}" in path or ROUTE_RE.search(path)):
             if self._resources and isinstance(self._resources[-1], GroupPlainResource):
                 resource = self._resources[-1]
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -420,6 +420,10 @@ class GroupPlainResource(Resource):
         self._resources: Dict[str, PlainResource] = {}
 
     @property
+    def resources(self) -> List[AbstractResource]:
+        return list(self._resources.values())
+
+    @property
     def canonical(self) -> str:
         return "/"
 
@@ -1004,7 +1008,12 @@ class View(AbstractView):
 
 class ResourcesView(Sized, Iterable[AbstractResource], Container[AbstractResource]):
     def __init__(self, resources: List[AbstractResource]) -> None:
-        self._resources = resources
+        self._resources = []
+        for resource in resources:
+            if isinstance(resource, GroupPlainResource):
+                self._resources.extend(resource.resources)
+            else:
+                self._resources.append(resource)
 
     def __len__(self) -> int:
         return len(self._resources)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -425,7 +425,7 @@ class GroupPlainResource(Resource):
 
     @property
     def canonical(self) -> str:
-        return "/"
+        return f"group_{id(self)}"
 
     def freeze(self) -> None:
         self._routes.clear()
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -431,7 +431,8 @@ class GroupPlainResource(Resource):
         self._routes.clear()
         for resource in self._resources.values():
             resource.freeze()
-            self._routes.extend(iter(resource))
+            for route in resource:
+                self._routes.append(route)
 
     def add_prefix(self, prefix: str) -> None:
         assert prefix.startswith("/")
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -426,7 +426,7 @@ class GroupPlainResource(Resource):
     def freeze(self) -> None:
         self._routes.clear()
         for resource in self._resources.values():
-            self._routes.extend(iter(resource))
+            self._routes.extend(cast(Iterator[ResourceRoute], resource))
 
     def add_prefix(self, prefix: str) -> None:
         assert prefix.startswith("/")
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1088,7 +1088,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
     def named_resources(self) -> Mapping[str, AbstractResource]:
         return MappingProxyType(self._named_resources)
 
-    def register_named_resource(self, resource: AbstractResource) -> None:
+    def _register_named_resource(self, resource: AbstractResource) -> None:
         assert isinstance(
             resource, AbstractResource
         ), f"Instance of AbstractResource class is required, got {resource!r}"
@@ -1124,7 +1124,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
 
     def register_resource(self, resource: AbstractResource) -> None:
         if resource.name:
-            self.register_named_resource(resource)
+            self._register_named_resource(resource)
         self._resources.append(resource)
 
     def add_resource(self, path: str, *, name: Optional[str] = None) -> Resource:
@@ -1147,7 +1147,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                 self.register_resource(grp)
             resource = grp.add_resource(_requote_path(path), name=name)
             if name:
-                self.register_named_resource(resource)
+                self._register_named_resource(resource)
             return resource
         # Reuse last added resource if path and name are the same
         if self._resources:
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -84,6 +84,7 @@ PATH_SEP: Final[str] = re.escape("/")
 
 _ExpectHandler = Callable[[Request], Awaitable[None]]
 _Resolve = Tuple[Optional["UrlMappingMatchInfo"], Set[str]]
+_InnerResource = Union["AbstractResource", "_PlainResourceGroup"]
 
 
 class _InfoDict(TypedDict, total=False):
@@ -150,10 +151,6 @@ class AbstractResource(Sized, Iterable["AbstractRoute"]):
     @abc.abstractmethod
     def raw_match(self, path: str) -> bool:
         """Perform a raw match against path"""
-
-    @abc.abstractmethod
-    def resources(self) -> List["AbstractResource"]:
-        """Return representing resources"""
 
 
 class AbstractRoute(abc.ABC):
@@ -374,9 +371,6 @@ class Resource(AbstractResource):
     def __iter__(self) -> Iterator["ResourceRoute"]:
         return iter(self._routes)
 
-    def resources(self) -> List[AbstractResource]:
-        return [self]
-
     # TODO: implement all abstract methods
 
 
@@ -421,17 +415,12 @@ class PlainResource(Resource):
         return f"<PlainResource {name} {self._path}>"
 
 
-class GroupPlainResource(Resource):
+class _PlainResourceGroup:
     def __init__(self) -> None:
-        super().__init__()
         self._resources: Dict[str, PlainResource] = {}
 
-    def resources(self) -> List[AbstractResource]:
-        return list(self._resources.values())
-
-    @property
-    def canonical(self) -> str:
-        return f"group_{id(self)}"
+    def resources(self) -> Iterator[PlainResource]:
+        return iter(self._resources.values())
 
     def freeze(self) -> None:
         for resource in self._resources.values():
@@ -451,25 +440,13 @@ class GroupPlainResource(Resource):
 
         return await resource.resolve(request)
 
-    def _match(self, path: str) -> Optional[Dict[str, str]]:
-        if path in self._resources:
-            return {}
-        else:
-            return None
-
     def raw_match(self, path: str) -> bool:
         return path in self._resources
-
-    def get_info(self) -> _InfoDict:
-        return {}
-
-    def url_for(self) -> URL:  # type: ignore[override]
-        raise RuntimeError(".url_for() is not allowed for GroupPlainResource")
 
     def __repr__(self) -> str:
         return f"<GroupPlainResource count={len(self._resources)}>"
 
-    def add_resource(self, path: str, name: Optional[str] = None) -> Resource:
+    def add_resource(self, path: str, name: Optional[str] = None) -> PlainResource:
         result = self._resources.get(path)
         if result is None:
             result = PlainResource(path, name=name)
@@ -574,9 +551,6 @@ class PrefixResource(AbstractResource):
 
     def raw_match(self, prefix: str) -> bool:
         return False
-
-    def resources(self) -> List[AbstractResource]:
-        return [self]
 
     # TODO: impl missing abstract methods
 
@@ -1010,9 +984,7 @@ class View(AbstractView):
 
 class ResourcesView(Sized, Iterable[AbstractResource], Container[AbstractResource]):
     def __init__(self, resources: List[AbstractResource]) -> None:
-        self._resources = []
-        for resource in resources:
-            self._resources.extend(resource.resources())
+        self._resources = resources
 
     def __len__(self) -> int:
         return len(self._resources)
@@ -1028,9 +1000,8 @@ class RoutesView(Sized, Iterable[AbstractRoute], Container[AbstractRoute]):
     def __init__(self, resources: List[AbstractResource]):
         self._routes: List[AbstractRoute] = []
         for resource in resources:
-            for resource in resource.resources():
-                for route in resource:
-                    self._routes.append(route)
+            for route in resource:
+                self._routes.append(route)
 
     def __len__(self) -> int:
         return len(self._routes)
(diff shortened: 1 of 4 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -1052,20 +1113,36 @@ def register_resource(self, resource: AbstractResource) -> None:
                     "already handled by {!r}".format(name, self._named_resources[name])
                 )
             self._named_resources[name] = resource
-        self._resources.append(resource)
+        if append:
+            self._resources.append(resource)
```
Dreamsorcerer: I feel like this would make more sense being split into a separate function (and then not using `self.register_resource()` below), rather than using an awkward keyword argument. Like:

```
def register_resource(...):
    self._register_named_resource(resource)
    self._resources.append(resource)
```

aiohttp/web_urldispatcher.py
```diff
@@ -1052,20 +1113,36 @@ def register_resource(self, resource: AbstractResource) -> None:
...
             raise ValueError("path should be started with / or be empty")
+        # Reuse named resource
+        if name:
+            resource = self._named_resources.get(name)
+            if resource is None:
+                pass
+            elif resource.raw_match(path):
+                return cast(Resource, resource)
+            else:
+                raise ValueError(f"Duplicate {name!r}, already handled by {resource!r}")
```
Dreamsorcerer: This looks like a separate change, it doesn't seem related to the rest of the PR. It also appears to make the below section (line 1142) redundant.

aiohttp/web_urldispatcher.py
```diff
@@ -1052,20 +1113,36 @@ def register_resource(self, resource: AbstractResource) -> None:
...
+        if not ("{" in path or "}" in path or ROUTE_RE.search(path)):
+            if self._resources and isinstance(self._resources[-1], GroupPlainResource):
+                grp = self._resources[-1]
+            else:
+                grp = GroupPlainResource()
+                self.register_resource(grp)
+            resource = grp.add_resource(_requote_path(path), name=name)
+            if name:
+                self.register_resource(resource, append=False)
+            return resource
```
Dreamsorcerer: It's not obvious to me why this has moved before the below block of code?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+            return None
+
+    def raw_match(self, path: str) -> bool:
+        return path in self._resources
+
+    def get_info(self) -> _InfoDict:
+        return {}
+
+    def url_for(self) -> URL:  # type: ignore[override]
+        raise RuntimeError(".url_for() is not allowed for GroupPlainResource")
```
Dreamsorcerer: Could this cause issues elsewhere?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+
+    def __repr__(self) -> str:
+        return f"<GroupPlainResource count={len(self._resources)}>"
+
+    def add_resource(self, path: str, name: Optional[str] = None) -> Resource:
+        result = self._resources.get(path)
```
Dreamsorcerer, line 449: Maybe just use `.setdefault()` and make this entire thing 1 line?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+        if path in self._resources:
+            return {}
+        else:
+            return None
+
+    def raw_match(self, path: str) -> bool:
+        return path in self._resources
+
+    def get_info(self) -> _InfoDict:
+        return {}
```
Dreamsorcerer: Could this cause issues as well? The typing information doesn't say that any of the keys are required, so maybe it's fine...

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+        if resource is None:
+            return None, allowed_methods
+
+        return await resource.resolve(request)
+
+    def _match(self, path: str) -> Optional[Dict[str, str]]:
+        if path in self._resources:
+            return {}
+        else:
+            return None
```
Dreamsorcerer:
```suggestion
        return {} if path in self._resources else None
```

Could make the same change to `PlainResource._match()` while we're here.

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+        assert not prefix.endswith("/")
+        assert len(prefix) > 1
+        old = self._resources.copy()
+        self._resources.clear()
+        for path, resource in old.items():
+            resource.add_prefix(prefix)
+            self._resources[prefix + path] = resource
+
+    async def resolve(self, request: Request) -> _Resolve:
+        allowed_methods: Set[str] = set()
```
Dreamsorcerer: This isn't used.

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+    @property
+    def canonical(self) -> str:
+        return "/"
+
+    def freeze(self) -> None:
+        self._routes.clear()
+        for resource in self._resources.values():
+            self._routes.extend(cast(Iterator[ResourceRoute], resource))
+
+    def add_prefix(self, prefix: str) -> None:
```
Dreamsorcerer, line 431: Given that the code below returns the inner resource, and not the group, when would this method ever get used?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
+        self._resources: Dict[str, Resource] = {}
+
+    @property
+    def canonical(self) -> str:
+        return "/"
+
+    def freeze(self) -> None:
+        self._routes.clear()
+        for resource in self._resources.values():
+            self._routes.extend(cast(Iterator[ResourceRoute], resource))
```
Dreamsorcerer: Why is the `cast()` needed? The definition just above says this should be a `Resource`, so how is it `Iterator[ResourceRoute]`?

aiohttp/web_urldispatcher.py
Dreamsorcerer, line 431: What happens when calling `add_prefix()` on the `PlainResource` in a group? How is the group going to update its keys?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
...
 
 
+class GroupPlainResource(Resource):
+    def __init__(self) -> None:
+        super().__init__()
+        self._resources: Dict[str, Resource] = {}
+
+    @property
+    def canonical(self) -> str:
+        return "/"
```
Dreamsorcerer: Is this useful... Are there cases where this might cause problems?

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,65 @@ def __repr__(self) -> str:
         return f"<PlainResource {name} {self._path}>"
 
 
+class GroupPlainResource(Resource):
+    def __init__(self) -> None:
+        super().__init__()
+        self._resources: Dict[str, Resource] = {}
```
Dreamsorcerer:
```suggestion
        self._resources: Dict[str, PlainResource] = {}
```

Dreamsorcerer (commented):
Would have been useful to add a description (or fill out anything in the PR template)...

Looks like this could provide an improvement in speed, but only when many plain resources are added together in sequence. There also appear to be a number of edge cases which I'm not sure will work safely with these changes.

aiohttp/web_urldispatcher.py
aamalev: To catch the last GroupPlainResource

aiohttp/web_urldispatcher.py
aamalev: https://github.com/aio-libs/aiohttp/runs/4636862493?check_suite_focus=true

```
Run make mypy
mypy
aiohttp/web_urldispatcher.py:429: error: Argument 1 to "iter" has incompatible type "Resource"; expected "Iterable[ResourceRoute]"  [arg-type]
aiohttp/web_urldispatcher.py:429: note: Following member(s) of "Resource" have conflicts:
aiohttp/web_urldispatcher.py:429: note:     Expected:
aiohttp/web_urldispatcher.py:429: note:         def __iter__(self) -> Iterator[ResourceRoute]
aiohttp/web_urldispatcher.py:429: note:     Got:
aiohttp/web_urldispatcher.py:429: note:         def __iter__(self) -> Iterator[AbstractRoute]
Found 1 error in 1 file (checked 124 source files)
make: *** [Makefile:89: mypy] Error 1
Error: Process completed with exit code 2.
```

Dreamsorcerer: Fixed: #6460

aiohttp/web_urldispatcher.py
```diff
@@ -414,6 +414,69 @@ def __repr__(self) -> str:
         return f"<PlainResource {name} {self._path}>"
 
 
+class GroupPlainResource(Resource):
+    def __init__(self) -> None:
+        super().__init__()
+        self._resources: Dict[str, PlainResource] = {}
+
+    @property
+    def resources(self) -> List[AbstractResource]:
```
Dreamsorcerer:
```suggestion
    def resources(self) -> List[PlainResource]:
```

aiohttp/web_urldispatcher.py
aamalev: fixed

aiohttp/web_urldispatcher.py
aamalev: Method modifies inner resources and returns None. Used in router.add_prefix

aamalev: The method will be called along the chain  router.add_prefix => grp.add_prefix => plain.add_prefix

aiohttp/web_urldispatcher.py
aamalev: The group is almost completely encapsulated in a router

aiohttp/web_urldispatcher.py
aamalev: In this section reuse all resources by name.
Next section for new PlainResource or reuse without name PlainResource.
And next section for another new or reuse without name last resource.
Finally section create DynamicResource

aiohttp/web_urldispatcher.py
aamalev: `.setdefault()` requires creation PlainResource. it's expensive

aiohttp/web_urldispatcher.py
aamalev: I have no other idea

aiohttp/web_urldispatcher.py
aamalev: then the readability of the code will decrease

aiohttp/web_urldispatcher.py
Dreamsorcerer: So, 3rd section does `if resource.name == name and resource.raw_match(path):`. But, if it has a name, wouldn't it already exist in `self._named_resources` and therefore already have done this check above? I'm struggling to see a situation where this 3rd check will ever run and return true.

tests/test_client_functional.py
```diff
@@ -876,8 +876,8 @@ async def redirect(request):
         raise web.HTTPFound(location="/")
 
     app = web.Application()
-    app.router.add_get("/", handler)
-    app.router.add_get("/redirect", redirect)
+    app.router.add_get("/", handler, allow_head=False)
+    app.router.add_get("/redirect", redirect, allow_head=False)
```
Dreamsorcerer, line 880: Is this now required? That potentially makes it a backwards-incompatible change. Although, it looks weird that it was previously using `add_head()` separately anyway...

asvetlov (changes requested):
The PR *might* add a performance boost in some cases (I'm still curious about numbers though).
I support the idea if the performance increase is visible but would see a different implementation.

`UrlDispatcher._resources` could be modified from `List[AbstractResource]` to `List[Union[AbstractResource, _PlainResourceGroup]]`.

The `_PlainResourceGroup` is a hidden container for `PlainResource` mapping that is never exposed to end user. It can have `_PlainResourceGroup.resolve()` and `_PlainResourceGroup.freeze()` methods but have no other `AbstractResource`'s methods and properties.

`UrlDispatcher.resources`, `UrlDispatcher.routes`, `UrlDispatcher.register_resource`, and `UrlDispatcher.add_resource` can be modified to work with `_PlainResourceGroup` directly if needed.

What do you think?

aiohttp/web_urldispatcher.py
```diff
@@ -1028,44 +1099,65 @@ def register_resource(self, resource: AbstractResource) -> None:
         if self.frozen:
             raise RuntimeError("Cannot register a resource into frozen router.")
```
Dreamsorcerer: These checks can be skipped if adding a resource, when there is already a group present. In that case, it only runs `_register_named_resource()`.

Maybe it would be better to put these checks in `_register_named_resource()` and call it unconditionally? Changing to `if not name: return` within that function.

aiohttp/web_urldispatcher.py
```diff
@@ -1028,44 +1099,65 @@ def register_resource(self, resource: AbstractResource) -> None:
...
+                raise ValueError(f"Duplicate {name!r}, already handled by {resource!r}")
+        if not ("{" in path or "}" in path or ROUTE_RE.search(path)):
+            if self._resources and isinstance(self._resources[-1], GroupPlainResource):
+                grp = self._resources[-1]
+            else:
+                grp = GroupPlainResource()
+                self.register_resource(grp)
+            resource = grp.add_resource(_requote_path(path), name=name)
+            if name:
+                self._register_named_resource(resource)
```
Dreamsorcerer: I think there's an edge case here where the `frozen` check doesn't happen. I'd like to see a test for that, I think something like:
```
add_resource("/foo", "foo")
freeze()
add_resource("/bar", "bar")
```

I suspect a simple solution to then get the test passing would be to ensure the group is frozen. It looks like `grp.add_resource()` currently will allow adding new resources after `grp.freeze()`, which seems incorrect. It freezes the resources, but not itself.

aiohttp/web_urldispatcher.py
Dreamsorcerer: Personally, I find the latter more readable, it's faster for me to see that the function returns correctly in all situations. Plus these simple 1-word literals are exactly the case these inline `if` expressions were made for.

aiohttp/web_urldispatcher.py
Dreamsorcerer: Problem is that `.add_resource()` appears to be part of the public API: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.UrlDispatcher.add_resource

So, if a user calls it, then it returns the `PlainResource`. The `.add_prefx()` method is not listed in the documentation, so not sure if it is considered part of the public API or not: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.PlainResource

Probably a question for @asvetlov.

Dreamsorcerer:
> The PR _might_ add a performance boost in some cases (I'm still curious about numbers though). I support the idea if the performance increase is visible but would see a different implementation.
> 
> `UrlDispatcher._resources` could be modified from `List[AbstractResource]` to `List[Union[AbstractResource, _PlainResourceGroup]]`.
> 
> The `_PlainResourceGroup` is a hidden container for `PlainResource` mapping that is never exposed to end user. It can have `_PlainResourceGroup.resolve()` and `_PlainResourceGroup.freeze()` methods but have no other `AbstractResource`'s methods and properties.
> 
> `UrlDispatcher.resources`, `UrlDispatcher.routes`, `UrlDispatcher.register_resource`, and `UrlDispatcher.add_resource` can be modified to work with `_PlainResourceGroup` directly if needed.
> 
> What do you think?

I think that might resolve some of the more odd quirks with the current proposal.

aiohttp/web_urldispatcher.py
aamalev: For example DynamicResource without name

aamalev:
> The PR _might_ add a performance boost in some cases (I'm still curious about numbers though). I support the idea if the performance increase is visible but would see a different implementation.
> 
> `UrlDispatcher._resources` could be modified from `List[AbstractResource]` to `List[Union[AbstractResource, _PlainResourceGroup]]`.
> 
> The `_PlainResourceGroup` is a hidden container for `PlainResource` mapping that is never exposed to end user. It can have `_PlainResourceGroup.resolve()` and `_PlainResourceGroup.freeze()` methods but have no other `AbstractResource`'s methods and properties.
> 
> `UrlDispatcher.resources`, `UrlDispatcher.routes`, `UrlDispatcher.register_resource`, and `UrlDispatcher.add_resource` can be modified to work with `_PlainResourceGroup` directly if needed.
> 
> What do you think?

Good. So there will be a clearer border between public and private. But the group will also need _PlainResourceGroup.add_prefix

Bench script:
```
import time
import asyncio

from aiohttp.test_utils import make_mocked_request
from aiohttp.web import UrlDispatcher, HTTPOk


async def handler(request):
    return HTTPOk()


async def main():
    router = UrlDispatcher()
    for i in range(10):
        router.add_get(f'/_{i}', handler)
    router.add_get('/a', handler)

    on_grp = len(router._resources) == 1

    request = make_mocked_request('GET', '/a')

    n = 2000000
    start = time.time_ns()
    for i in range(n):
        await router.resolve(request)

    print('on_grp', on_grp, 'with n', n, 'time', (time.time_ns() - start) / n)


asyncio.run(main())
```

Numbers:
on_grp True with n 2000000 time 3613.75
on_grp False with n 2000000 time 10493.992

tests/test_client_functional.py
aamalev: Now required. But in any case next add_head creates an unreachable resource

aiohttp/web_urldispatcher.py
aamalev: Maybe raise ValueError?

aiohttp/web_urldispatcher.py
aamalev: Covered by test_frozen_plain_group

aiohttp/web_urldispatcher.py
aamalev: lost relevance

aiohttp/web_urldispatcher.py
aamalev: lost relevance

aiohttp/web_urldispatcher.py
aamalev: fixed

Dreamsorcerer:
I believe this is no longer relevant with the new routing algorithm which also improves routing for all resources.

PR #6495, opened 2022-01-06
Support credentials in URL with empty user, http://:password@host (#6494)

### LINKED ISSUE #6494: Auth provided in URL skipped if user field is blank
### Describe the bug

Some services publish a URL like:

    https://:mypassword@ropsten.infura.io/v3/..."

You can see the user field is an empty length string. When a request is made using aiohttp, the credentials are ignored and the server responds with a 403. This differs from requests and also curl, which parse this URL as having a zero-length username and a valid password.

### To Reproduce

The request is sent without credentials

### Expected behavior

Credentials should be sent

### Logs/tracebacks

```python-traceback
N.A.
```

### Python Version

```console
$ python --version
Version: 3.7.4.post0
```

### aiohttp Version

```console
$ python -m pip show aiohttp
```

### multidict Version

```console
Version: 5.1.0
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.6.3
...
```

### OS

macOS

### Related component

Client

### Additional context

In `asyncio/helpers.py` specifically invoking `strip_auth_from_url` calls `BasicAuth.from_url` line, which contains:

    @classmethod
    def from_url(cls, url: URL, *, encoding: str = "latin1") -> Optional["BasicAuth"]:
        """Create BasicAuth from url."""
        if not isinstance(url, URL):
            raise TypeError("url should be yarl.URL instance")
        if url.user is None:
            return None
        return cls(url.user, url.password or "", encoding=encoding)

The underlying cause appears to be that yarl returns `url.user=None` and `url.password=mypassword`. You can see that `url.password` is not checked in this case. I will open a PR with suggested fix and test.

### Code of Conduct

shuckc:
## What do these changes do?
Per issue #6494 this brings our handling of credentials in the URL where the username is blank in line with the behavior of curl, requests and perhaps other client libraries. 

Potentially - although vanishingly remote. One could imagine a client to be providing a password in this style, and for aiohttp to be ignoring it, and for the service to allow anonymous access, but not with a Basic authentication header provided. However this is a hidden misconfiguration, so perhaps we are helping them out. 
 
## Related issue number

Fixes #6494 

## Checklist


shuckc:
For the test case added to `test_basic_auth_no_user_from_url` I ran the same fictitious URL with curl and verified the Basic auth header value matched.

CHANGES/6494.bugfix
```diff
@@ -0,0 +1 @@
+Support credentials in URL with empty user, e.g. http://:password@host
```
webknjaz: Plz use the past tense here.

webknjaz (changes requested):
Blocking until we can decide if this change is correct. What cURL does is not indicative as it relies on the user/invoker constructing parts of the request properly. And the RFC seems to indicate that having an empty username is incorrect.

This needs more research and we must remain RFC-compliant.

Ref: https://github.com/aio-libs/aiohttp/issues/6494#issuecomment-1006646632

shuckc:
If we look at the collected [ABNF for a URL from Appendix-A of the RFC3986](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A) in particular:
 
    authority     = [ userinfo "@" ] host [ ":" port ]
    userinfo      = *( unreserved / pct-encoded / sub-delims / ":" )

You can see that `:password` is permitted as a `userinfo` value, indeed the use of repeated colons is allowed. However the RFC mentions nothing about how to actually _parse_ the extracted userinfo into what we commonly call user/password, for which reference to other implementations is probably helpful. 

The regexp given in Appendex-B of the same document also permits the authority to contain an empty username.

Looking at other relevant RFCs, Basic Auth has to join back together the user/password (recreating the `userinfo` field) and here they are more specific about that process:

[RFC2617 section 2 (Basic Auth)](https://datatracker.ietf.org/doc/html/rfc2617#section-2) specifies an ABNF that allows zero length usernames (and passwords)

    2 Basic Authentication Scheme

    The "basic" authentication scheme is based on the model that the
    client must authenticate itself with a user-ID and a password for
    each realm (...)
    ...
    To receive authorization, the client sends the userid and password,
    separated by a single colon (":") character, within a base64 [7]
    encoded string in the credentials.

      basic-credentials = base64-user-pass
      base64-user-pass  = <base64 [4] encoding of user-pass,
                       except not limited to 76 char/line>
      user-pass   = userid ":" password
      userid      = *<TEXT excluding ":">
      password    = *TEXT


[RFC7617  (Basic Auth)](https://datatracker.ietf.org/doc/html/rfc7617) does not mandate the username be of any particular length.

I cannot see anything in relevant RFCs that prohibit a zero-length username.

CHANGES/6494.bugfix
shuckc: How about:
"URLs providing credentials with zero-length user are now supported (http://:password@host)" ?
"URL credentials with empty user field are supported (http://:password@host)" ?

Dreamsorcerer:
I don't really mind either way. Looks fine.

As mentioned, the whole basic auth thing is a rather old deprecated thing, but that's more a concern for the server-side implementations still making the mistake of using it, rather than the client-side code which is forced to use it.

shuckc:
Updated commit and `CHANGES` message - @webknjaz anything outstanding?

CHANGES/6494.bugfix
```diff
@@ -0,0 +1 @@
+URL credentials with zero-length user now supported, e.g. http://:password@host
```
webknjaz: This fragment is actually rendered using an RST markup so let's take advantage of that:
```suggestion
Added support for URL credentials with empty (zero-length) username, e.g. ``https://:password@host`` -- by :user:`shuckc`
```

webknjaz:
> Updated commit and `CHANGES` message - @webknjaz anything outstanding?

I'd like to get feedback from @asvetlov too and dig into the standards a little bit more myself before making this change permanent.

webknjaz:
WHATWG seems to think that this corner case is valid: https://github.com/whatwg/url/issues/139. OTOH I haven't found any explicit tests in the `requests` repo that would indicate that they intentionally support this.

shuckc:
Any chance of re-review and/or merge?

Dreamsorcerer:
To summarise:
https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.1
This does not say anything about not allowing empty usernames and the ABNF seems to allow it (the "optional password" I would interpret as the `:` being optional to specific a password, which again can be empty).

WHATWG also seems to allow it: https://url.spec.whatwg.org/#authority-state

RFC7617 doesn't appear to disallow it.

So, I think we can just proceed with this.

PR #6498, opened 2022-01-09
Introduce AppKey for type safety

Dreamsorcerer:
Will create some PRs in other projects shortly to test it all out properly. But, this looks promising so far. It's less intrusive than the previous proposal and still works with `Request.config_dict`.

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -812,10 +812,7 @@ class AppKey(Generic[_T]):
             frame = frame.f_back
 
         self._name = module + "." + name
-        try:
-            self._t = t.__qualname__  # type: ignore[union-attr]  # If concrete class
-        except AttributeError:
-            self._t = t
+        self._t = t
 
     def __lt__(self, other: object) -> bool:
         if isinstance(other, AppKey):
@@ -823,16 +820,22 @@ class AppKey(Generic[_T]):
         return True  # Order AppKey above other types.
 
     def __repr__(self) -> str:
-        if self._t:
-            t: object = self._t
-        else:
-            t = "<<Unknown>>"
+        t = self._t
+        if t is None:
             with suppress(AttributeError):
                 # Set to type arg.
-                t = cls = get_args(self.__orig_class__)[0]
-                # If a class, set to the qualname.
-                t = cls.__qualname__
-        return f"<AppKey({self._name}, type={t})>"
+                t = get_args(self.__orig_class__)[0]
+
+        if t is None:
+            t_repr = "<<Unkown>>"
+        elif isinstance(t, type):
+            if t.__module__ == "builtins":
+                t_repr = t.__qualname__
+            else:
+                t_repr = f"{t.__module__}.{t.__qualname__}"
+        else:
+            t_repr = repr(t)
+        return f"<AppKey({self._name}, type={t_repr})>"
 
 
 @final
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -797,7 +797,7 @@ class AppKey(Generic[_T]):
         frame = inspect.currentframe()
         while frame:
             if frame.f_code.co_name == "<module>":
-                module = frame.f_globals["__name__"]
+                module: str = frame.f_globals["__name__"]
                 break
             frame = frame.f_back
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -41,6 +41,7 @@ from typing import (
     TypeVar,
     Union,
     cast,
+    get_args,
     overload,
 )
 from urllib.parse import quote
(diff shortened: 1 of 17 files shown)


aiohttp/helpers.py
```diff
@@ -784,11 +786,41 @@ def set_exception(fut: "asyncio.Future[_T]", exc: BaseException) -> None:
         fut.set_exception(exc)
 
 
+@functools.total_ordering
+class AppKey(Generic[_T]):
```
Dreamsorcerer, line 795: I initially put this in web_app.py, but that looks like a circular import to reference it here. Maybe I can still move it back though and do a `if TYPE_CHECKING` thing here.

asvetlov: I think this is fine

asvetlov:
If it makes linter problems, we can drop inheritance from `Mapping` / `MutableMapping`.
Strictly speaking, the proposed structure is a heterogeneous container; `typing.Mapping[K, V]` is more likely a homogenous version.
Mypy should still be smart enough to infer types from `__getitem__` / `__setitem__`, isn't it?
The support of runtime `collections.abc` check is still desirable. It can be done either by inheritance or by direct `collections.abc.mapping.register()` call -- both works to me.

Dreamsorcerer:
> If it makes linter problems, we can drop inheritance from `Mapping` / `MutableMapping`.

I'll just drop a `type: ignore` in, should work otherwise. I was just hoping to get a better approach from https://github.com/python/mypy/issues/11949

Dreamsorcerer:
So, it doesn't work as nicely as I'd like.

`AppKey("foo", Iterable[int])` won't work as it's not a concrete class. So, instead we probably have to do it like the Stash implementation as `AppKey[Iterable[int]]("foo")`.

Another issue is that the typing errors from mypy are rather difficult to work with. For example, this code in aiohttp-jinja2:
```
app[aiohttp_jinja2.APP_CONTEXT_PROCESSORS_KEY] = (
    aiohttp_jinja2.request_processor,
    processor,
)
```
Produces:
```
Cannot infer type argument 1 of "__setitem__" of "Application"  [misc]
```

But, it works if you write it as:
```
f: Tuple[aiohttp_jinja2._ContextProcessor, ...] = (
    aiohttp_jinja2.request_processor,
    processor,
)
app[aiohttp_jinja2.APP_CONTEXT_PROCESSORS_KEY] = f
```


Even with simple examples like:
```
k = web.AppKey[int]("k")
app[k] = "foo"
```
Mypy just produces: `Cannot infer type argument 1 of "__setitem__" of "Application"  [misc]`

Which is rather awkward for the user to figure out the problem. The type of both items is defined as a `TypeVar` (`def __setitem__(self, key: AppKey[_T], value: _T)`), so mypy only tells you if they are not the same type, but it doesn't tell you that the value type is the problem or what the expected type is.

Dreamsorcerer:
Anyway, changes for aiohttp-jinja2: https://github.com/aio-libs/aiohttp-jinja2/pull/534/files

Dreamsorcerer:
> `AppKey("foo", Iterable[int])` won't work as it's not a concrete class.

Weirdly, mypy stop complaining about this once the parameter is defined with `Optional`. I'd rather like to keep this syntax, so I've added support for both.

asvetlov (approved):
LGTM.
@Dreamsorcerer please feel free to merge when you decide that the PR is ready.

Dreamsorcerer:
Right, I'm going to merge this and hope that mypy improves its error message in future. I opened an issue for it: https://github.com/python/mypy/issues/12012

eoghanmurray:
What if I don't care about types (because I'm using Python, not Typescript)? 
I want to do 
```
app['redis'] = RedisCluster()
app['port'] = 30000
```
Without having to figure out why my app is warning about AppKeys etc.?
Specifically IMO the warning `UserWarning: It is recommended to use web.AppKey instances for keys.` would be an appropriate warning given by mypy or some tool as it is running type checking over an app, it is inappropriate for an application startup warning.

Dreamsorcerer:
Firstly, it's a warning, just filter it out if you decide it's not relevant for you. Secondly, it's not possible to create a warning in mypy for something like this (especially retroactively, while maintaining backwards-compatibility).

I'll add a note to that section with examples to filter out the warning, but it's just a standard warning. We added a custom class, so you can filter out that specific warning very easily.

Dreamsorcerer:
Also, even if you don't use mypy, a lot of users still want annotations etc. in their IDE via pyright or similar.

Dreamsorcerer:
https://aiohttp--7839.org.readthedocs.build/en/7839/web_advanced.html#application-s-config

eoghanmurray:
Thanks for your response!

> Just filter it out

To do that I have to figure out what the warning means, and whether it's something that I need to fix. In this case the fix (using a WebKey class when both setting and retrieving, and figuring out types for values which would then need to be maintained) seems quite un-pythonic and would make the code more mysterious.

I did already try to filter it out before posting here (I didn't previously know if filtering warnings was possible in Python) but the following code taken from SO didn't work

```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    app['redis'] = RedisCluster()
    app['port'] = 30000
```

As I now have learned that you need to set up the 'ignore' before `app = web.Application()` is initialized.

I just think that this is something that should be highlighted as best practices in the documentation and doesn't merit a warning.
With thanks for your time.

Dreamsorcerer:
```
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    app["foo"] = 3
```

Works fine for  me.

Anyway, if you don't care at all, filter it globally, examples in the above link. I'd use catch_warnings() if you're only trying to maintain backwards-compatibility with something (which is likely what homeassistant will be doing).

eoghanmurray:
My version is 3.9.0b0 which doesn't appear to have #7677 in it "Only display warning when using -Wdefault, -We, -X dev or similar."

So basically my 'observations' are redundant I think (I didn't have warnings explicitly turned on).

Dreamsorcerer:
Ah, yes, I didn't realise your were on b0. I did mention that the warning shouldn't appear unless you're explicitly enabling them with `-Wall` or similar. That was a bug, as I thought it would be ignored by default.

rc0 is building now.

eoghanmurray:
I apologize for wasting your time, `3.9.0b1` works perfect in this respect.

PR #6505, opened 2022-01-11
Add error-handling to web-advanced docs

yuvalherziger:
## What do these changes do?

Following [this gitter discussion](https://gitter.im/aio-libs/Lobby?at=61dd3a6e742c3d4b218fe89c), this PR:

* Adds a new section to the web-advanced documentation page for error handling
* References aiohttp-catcher as a possible third-party candidate library for error handling, as well as a reference to middleware example.

N/A

## Related issue number

N/A

## Checklist

### DIFF
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -397,6 +397,20 @@ An example of context vars usage::
 
 .. versionadded:: 3.5
 
+.. _aiohttp-web-error-handling:
+
+Error Handling
+--------------
+
+*aiohttp* deliberately doesn't have a built-in error handling
+mechanism.  While the framework does raise :exc:`HTTPException` exceptions,
+it leaves it up to you to handle those errors per your specific requirements.
+
+You can leverage a third-party middleware library such as
+`aiohttp-catcher <https://github.com/yuvalherziger/aiohttp-catcher>`_. to
+handle errors in a centralized way, or alternatively build your own error-handling
+middleware if **aiohttp-catcher** doesn't satisfy your requirements (see the
+**Examples** section in :ref:`aiohttp-web-middlewares`).
 
 .. _aiohttp-web-middlewares:
 


Dreamsorcerer:
Not sure if this was just forgotten in that conversation or not, but we already have a simple catcher in the tutorial:
https://demos.aiohttp.org/en/latest/tutorial.html#middlewares

Personally, I'd prefer to point people to that than another library for most cases. It's ~10 lines of code and probably enough for the vast majority of cases.

Also, a piece of advice for aiohttp-catcher would be to highlight the ordering of middlewares in the readme, in particular when using something like aiohttp-debugtoolbar for development. As both middlewares catch all exceptions, it's important that debugtoolbar is inserted after (and therefore catches before) catcher.

webknjaz:
@Dreamsorcerer it was @asvetlov who suggested adding this section, we could wait until he has time to review it.

yuvalherziger:
Great points, @Dreamsorcerer. Addressing each point separately:

* You're 100% right that I should highlight the ordering of middleware. That's an oversight on my part and I'll make that change in the library's README file.
* I wasn't aware of that tutorial page, and it's very useful - thank you.  That being said, I think that this example is a little different.  aiohttp-catcher is aimed at making error handling consistent at scale.  In fact, it's already being used in production in a REST API that throws dozens of different exceptions, mapped to different JSON responses.  Its approach is a little different from the example in the tutorial, as aiohttp-catcher is meant to avoid throwing random 500's with a response containing the wrong MIME type (e.g., an HTML 500 error page where a JSON response is expected from a REST API).  I'd really understand if it's preferable to reference the tutorial's example instead of a new third-party library like aiohttp-catcher, I just thought I'd point out that the example and aiohttp-catcher aim for two different goals.

One way or the other, I appreciate the feedback!

Edit: re:first point, note added:

image

Dreamsorcerer:
> aiohttp-catcher is meant to avoid throwing random 500's with a response containing the wrong MIME type (e.g., an HTML 500 error page where a JSON response is expected from a REST API).

I've only glanced at the code, but it looks like it always returns a JSON response: https://github.com/yuvalherziger/aiohttp-catcher/blob/main/aiohttp_catcher/catcher.py#L64

Which would be less flexible, as the tutorial expects you to define a handler, so you can choose what the response should look like. If I have a website and an API, I might create the API as a subapp, then have a Jinja handler for the website and a JSON handler for the API.

I'm also not sure the need to handle a lot of different exceptions is a general use case. The general approach that I use for most projects would be the tutorial catcher for custom status pages, plus a second middleware that catches trafaret/pydantic exceptions and generates a response to the user explaining what validation failed. Any other exceptions should be logged and fixed (exposing them to the user can be a security risk).

So, I feel the use case is probably a little niche for this documentation. However, if others say that they think it's useful I don't mind adding it in, as long as it's reworded to something more like "see the tutorial for a simple implementation. For more complex exception handlers you may want to check out aiohttp-catcher".

Dreamsorcerer:
There doesn't seem to have been much interest in this one, so I think we'll close it. I'll probably look at a bigger cleanup of the docs at some point and make sure it's covered better then.

PR #6515, opened 2022-01-14
Add way to read http response for failed ws connect

romasku:
## What do these changes do?

It adds `session.try_ws_connect` methods that return an intermediate object that allows reading HTTP response body in
cases WS handshake failed.

Old  `session.ws_connect` should continue to work same it did before.

## Checklist

### DIFF
--- a/CHANGES/6515.feature
+++ b/CHANGES/6515.feature
@@ -1,0 +1,1 @@
+Add `session.try_ws_connect` method that allows access to ClientResponse in case WS handshake failed.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -66,6 +66,7 @@ from .client_reqrep import (
 )
 from .client_ws import (
     DEFAULT_WS_CLIENT_TIMEOUT,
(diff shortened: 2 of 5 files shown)


Dreamsorcerer:
I'm not sure this is necessarily the right API, but we should certainly come back to this some time. #8572 is also related, as it wouldn't be needed if we went with something like this.

Dreamsorcerer:
@bdraco What do you think about this? Should we have a way to get the failed response?

If so, do you think just adding a response attribute as in #8572 makes the most sense?

bdraco:
I wonder if we could improve the error reporting in the normal path instead.

I think just getting the failed `response` makes more sense. This proposal adds more complexity and requires the dev implementing it to expect the failure mode in advance.

Dreamsorcerer:
I'm going to close this one, but a new approach may be welcome. The PR is mentioned won't work though, as it just makes the response available after a successful connection. I'm not sure how much it's worth looking at currently.

PR #6533, opened 2022-01-19
Add HTTP method validation

asvetlov:


CHANGES/6533.feature
```diff
@@ -0,0 +1 @@
+Add HTTP method validation
```
webknjaz: Could you use the past tense, add a trailing period and be more specific about what this validation does?

tests/test_client_request.py
```diff
@@ -88,6 +88,11 @@ def test_method3(make_request: Any) -> None:
     assert req.method == "HEAD"
 
 
+def test_method3_invalid(make_request: Any) -> None:
+    with pytest.raises(ValueError):
```
webknjaz: The best practice is to always use the `match=` argument with `pytest.raises()`.

CHANGES/6533.feature
asvetlov: I rather recommend relaxing the 'past tense' rule. It is a spike in change-note writing, many people make this mistake.
Why should we have it? `pytest` accepts both present and past tenses in their changenotes.

asvetlov: The method validation is... the validation.
Are you asking for mentioning allowed/forbidden characters? Why?

webknjaz: There are several reasons for doing this:
1) Changelog has a certain audience — it's humans/end-users that need to understand what changed when updating to the given version from the previous one, hence the past tense is usually the best fit since it describes what has already happened since that previous version. While they describe a high-level effect for the end-users, they still need to be precise and include just enough context for the users to understand what they are even about without having to dig deep into the actual patch. It's different from Git commits that traditionally use the imperative mood of the present tense and focus on describing the change at hand, going deeper into internal developer-oriented details.
2) I don't think that accepting differing styles of change notes is good because this contributes to the inconsistency in the compiled things. Just look at https://docs.aiohttp.org/en/stable/changes.html#id1 — it's a zoo of styles of inconsistent phrases. Ideally, this whole list should look as if it was written by a single individual in the same consistent style.
3) It's the best guidance I can give at the moment. Unfortunately, there are not enough recommendations on the internet. But I've recently started collecting such recommendations from different sources/communities/tech writers in order to come up with better instructions for FOSS projects. I'll probably update the contributing guide at some point, once that information is discovered.

webknjaz: > The method validation is... the validation.
> Are you asking for mentioning allowed/forbidden characters? Why?

From a generic note, it seems like (1) there was no validation whatsoever and it was just added, (2) it's unclear if this is a validation of the allowed list of method names or maybe some RFC compliance implementation. It is not really clear if anything changed compared to what was in the previous version (remember that this text is disconnected from the source code and is consumed on a dedicated web page, not as a part of a PR or a Git commit or any other context).

PR #6536, opened 2022-01-19
Fix missing AttributeError

Dreamsorcerer:
Found this test in one of my PRs. Must have been removed at some point, so just putting it back in.


asvetlov:
If you would backport it -- please go ahead!

PR #6547, opened 2022-01-22
Remove overlapping slots and fix broken slots inheritance

ariebovenberg:
## What do these changes do?

There were a few `__slots__` related mistakes:

```shell
python -m slotscheck aiohttp -v
ERROR: 'aiohttp.web_protocol:RequestHandler' defines overlapping slots.
       - _reading_paused (aiohttp.base_protocol:BaseProtocol)
ERROR: 'aiohttp.web_response:StreamResponse' has slots but superclass does not.
       - aiohttp.helpers:CookieMixin
```

Fixing the second error was more problematic than it seems. The straightforward solution of adding `__slots__ = ('_cookies', )` was not possible, because this triggered a `TypeError: multiple bases have instance lay-out conflict` in `HTTPException` since non-emptly slots cannot be combined with an `Exception` case class:

```python
>>> class CookieMixin:
...  __slots__ = ('_cookies', )
...
>>> class HTTPException(CookieMixin, Exception):
...  pass
...
TypeError: multiple bases have instance lay-out conflict
```

There are several possible solutions:
1. put empty slots in `CookieMixin`, since it's an abstract class. This is the currently implemented solution.
2. leave the situation as-is. The `__slots__` in `StreamResponse` will be mostly ineffective.
3. remove `__slots__` from `StreamResponse`. They aren't saving memory or preventing `__dict__` from being created. 

`StreamResponse` will no longer have a `__dict__`, so arbitrary attributes cannot be set. This is not documented, so likely not a problem

## Related issue number

No issue. This is a small change.

## Checklist

- [x] ~~Unit tests for the changes exist~~ n/a (see questions below)
- [x] ~~Documentation reflects the changes~~ n/a

## Questions

- regarding tests: probably not needed. I discovered the slots issue with [slotscheck](https://github.com/ariebovenberg/slotscheck), a tool I wrote. Of course, If you like, I can add it to tox/CI as I've done for [instagram/LibCST](https://github.com/Instagram/LibCST/pull/615) and [dry-python/returns](https://github.com/dry-python/returns/pull/1233).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,12 @@ jobs:
         python -m pip install .
       env:
         AIOHTTP_NO_EXTENSIONS: 1
-    - name: Run linters
+    - name: Run mypy
       run: |
         make mypy
+    - name: Run slotscheck
+      run: |
+        slotscheck -m aiohttp
     - name: Install libenchant-dev
       run: |
         sudo apt install libenchant-dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         make mypy
     - name: Run slotscheck
       run: |
-        slotscheck -m aiohttp
+        python -m slotscheck -m aiohttp
     - name: Install libenchant-dev
       run: |
         sudo apt install libenchant-dev
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,10 @@ jobs:
         make mypy
     - name: Run slotscheck
       run: |
-        python -m slotscheck -m aiohttp
+        # Some extra requirements are needed to ensure all modules
+        # can be scanned by slotscheck.
+        pip install -r requirements/base.txt -c requirements/constraints.txt
+        slotscheck -v -m aiohttp
     - name: Install libenchant-dev
       run: |
         sudo apt install libenchant-dev
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -831,9 +831,15 @@ class ChainMapProxy(Mapping[str, Any]):
 
 
 class CookieMixin:
+    # The `_cookies` slots is not defined here because non-empty slots cannot
+    # be combined with an Exception base class, as is done in HTTPException.
+    # CookieMixin subclasses with slots should
+    # define the `_cookies` slot themselves.
+    __slots__ = ()
+
     def __init__(self) -> None:
         super().__init__()
-        self._cookies: SimpleCookie[str] = SimpleCookie()
+        self._cookies: SimpleCookie[str] = SimpleCookie()  # type: ignore
 
     @property
     def cookies(self) -> "SimpleCookie[str]":
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -833,13 +833,15 @@ class ChainMapProxy(Mapping[str, Any]):
 class CookieMixin:
     # The `_cookies` slots is not defined here because non-empty slots cannot
     # be combined with an Exception base class, as is done in HTTPException.
-    # CookieMixin subclasses with slots should
-    # define the `_cookies` slot themselves.
+    # CookieMixin subclasses with slots should define the `_cookies`
+    # slot themselves.
     __slots__ = ()
 
     def __init__(self) -> None:
         super().__init__()
-        self._cookies: SimpleCookie[str] = SimpleCookie()  # type: ignore
+        # Mypy doesn't like that _cookies isn't in __slots__.
+        # See the comment on this class's __slots__ for why this is OK.
+        self._cookies: SimpleCookie[str] = SimpleCookie()  # type: ignore[misc]
 
     @property
     def cookies(self) -> "SimpleCookie[str]":
--- a/CHANGES/6547.bugfix
+++ b/CHANGES/6547.bugfix
@@ -1,2 +1,2 @@
-Remove overlapping slots in :py:class:`~aiohttp.web_protocol.RequestHandler`, 
+Remove overlapping slots in :py:class:`~aiohttp.web_protocol.RequestHandler`,
 fix broken slots inheritance in :py:class:`~aiohttp.web_response.StreamResponse`.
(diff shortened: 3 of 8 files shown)


Dreamsorcerer:
Would be great if you could add this test to the linter step in CI. Then we can avoid introducing new overlaps in future.

aiohttp/helpers.py
```diff
@@ -831,9 +831,15 @@ def __repr__(self) -> str:
...
+    # The `_cookies` slots is not defined here because non-empty slots cannot
+    # be combined with an Exception base class, as is done in HTTPException.
+    # CookieMixin subclasses with slots should
+    # define the `_cookies` slot themselves.
+    __slots__ = ()
+
     def __init__(self) -> None:
         super().__init__()
-        self._cookies: SimpleCookie[str] = SimpleCookie()
+        self._cookies: SimpleCookie[str] = SimpleCookie()  # type: ignore
```
Dreamsorcerer: Please add the code, and ideally, a comment on why we are ignoring. `# type: ignore[specific-code]`

Though might be worth waiting to see if someone else has a suggestion on the `__slots__` issue.

Although, I'm confused what actually causes the error when inheriting from `Exception`. It works if using single inheritance, or even by inheriting from `Exception` twice:
```
>>> class A:
...  __slots__ = ("a",)
... 
>>> class AE(Exception):
...  __slots__ = ("a",)
... 
>>> class B(AE):  # Fine
...  __slots__ = ("b",)
... 
>>> class B(AE, Exception):  # Fine
...  __slots__ = ("b",)
... 
>>> class B(A, Exception):  # Borked
...  __slots__ = ("b",)
... 
TypeError: multiple bases have instance lay-out conflict
```

I can't seem to find any documentation explaining this behaviour.

Dreamsorcerer: This is potentially a backward-incompatible change as well, so not sure if this is safe to backport.

ariebovenberg:
@Dreamsorcerer You mean in `workflows/ci.yaml`?

```yaml
    - name: Run linters
      run: |
        make mypy
        make slotscheck  # after adding slotscheck to Makefile
```

Alternatively I can add the [hook](https://slotscheck.readthedocs.io/en/latest/advanced.html#pre-commit-hook) to the `.pre-commit-config.yaml`. Whichever you prefer.

Dreamsorcerer:
> @Dreamsorcerer You mean in `workflows/ci.yaml`?
> 
> ```yaml
>     - name: Run linters
>       run: |
>         make mypy
>         make slotscheck  # after adding slotscheck to Makefile
> ```

That looks fine to me. Not sure it needs to go in the Makefile. But, I'd make it a separate step (and probably rename that existing step to 'Run Mypy'). I meant lint job earlier, rather than linter step..

aiohttp/helpers.py
ariebovenberg: I don't know either what this 'lay-out' conflict is, exactly. ~It may have to do with the fact that `Exception` is a builtin class, but its instances _do_ have a `__dict__`.~

edit: nevermind, the case is exactly the same when inheriting from `int`, or `zip`

ariebovenberg: From the documentation there are some _hints_:

> Multiple inheritance with multiple slotted parent classes can be used, but only one parent is allowed to have attributes created by slots (the other bases must have empty slot layouts) - violations raise TypeError.

ariebovenberg:
> That looks fine to me. Not sure it needs to go in the Makefile. But, I'd make it a separate step (and probably rename that existing step to 'Run Mypy'). I meant lint job earlier, rather than linter step.

Can do. I'm curious though: what's the reasoning behind what goes in pre-commit vs GitHub workflow?

aiohttp/helpers.py
Dreamsorcerer: The error is supposed to indicate an issue where it's unclear which class the attribute would come from:
```
>>> class A:            
...   __slots__ = ("a",)
... 
>>> class B:            
...   __slots__ = ("a",)
... 
>>> class C(A, B):
...  __slots__ = ()
... 
TypeError: multiple bases have instance lay-out conflict
```

It's interesting that the same error occurs with `int`. This doesn't support non-empty slots due to being variable length, which would normally give an error like:
```
>>> class C(int):
...  __slots__ = ("c",)
... 
TypeError: nonempty __slots__ not supported for subtype of 'int'
```

Maybe there is some other restriction with C classes that I couldn't find documented.

Dreamsorcerer:
> > That looks fine to me. Not sure it needs to go in the Makefile. But, I'd make it a separate step (and probably rename that existing step to 'Run Mypy'). I meant lint job earlier, rather than linter step.
> 
> Can do. I'm curious though: what's the reasoning behind what goes in pre-commit vs GitHub workflow?

Probably a question for someone else. I'd assume that pre-commit is being run as a git hook, so it makes sense to have primarily fast checks which are likely to catch mistakes. If it only catches an error for 1 in 500 PRs or something, it's probably best not delaying a developer's commit. But, not sure if that's what has been done so far, I didn't set up anything in pre-commit and don't really mind which way it goes in.

ariebovenberg:
> > > That looks fine to me. Not sure it needs to go in the Makefile. But, I'd make it a separate step (and probably rename that existing step to 'Run Mypy'). I meant lint job earlier, rather than linter step.
> > 
> > 
> > Can do. I'm curious though: what's the reasoning behind what goes in pre-commit vs GitHub workflow?
> 
> Probably a question for someone else. I'd assume that pre-commit is being run as a git hook, so it makes sense to have primarily fast checks which are likely to catch mistakes. If it only catches an error for 1 in 500 PRs or something, it's probably best not delaying a developer's commit. But, not sure if that's what has been done so far, I didn't set up anything in pre-commit and don't really mind which way it goes in.

Indeed, problems with slots probably won't occur in most PRs. I'll add it to the GH workflow for now.

aiohttp/helpers.py
ariebovenberg: I've added an extra explanation on the mypy ignore. I'll consider it resolved unless anybody else comments :)

asvetlov (approved):

asvetlov:
Thanks!

webknjaz:
@Dreamsorcerer @ariebovenberg does anybody remember why this never got backported to stable branches?

.github/workflows/ci.yml
```diff
@@ -51,9 +51,15 @@ jobs:
...
-    - name: Run linters
+    - name: Run mypy
       run: |
         make mypy
+    - name: Run slotscheck
+      run: |
+        # Some extra requirements are needed to ensure all modules
+        # can be scanned by slotscheck.
+        pip install -r requirements/base.txt -c requirements/constraints.txt
+        slotscheck -v -m aiohttp
```
webknjaz, line 62: @Dreamsorcerer I think we need to figure this out: https://github.com/aio-libs/aiohttp/pull/13466

PR #6551, opened 2022-01-23
Cleanup errors with new multidict

Dreamsorcerer:


aiohttp/client_reqrep.py
```diff
@@ -343,11 +343,8 @@ def update_headers(self, headers: Optional[LooseHeaders]) -> None:
                     self.headers.add(key, value)
 
     def update_auto_headers(self, skip_auto_headers: Iterable[str]) -> None:
-        self.skip_auto_headers = CIMultiDict(
-            (hdr, None) for hdr in sorted(skip_auto_headers)
-        )
-        used_headers = self.headers.copy()
-        used_headers.extend(self.skip_auto_headers)  # type: ignore[arg-type]
+        self.skip_auto_headers = set(skip_auto_headers)
+        used_headers = self.skip_auto_headers | self.headers.keys()
```
Dreamsorcerer, line 347: Please sanity check this change.

There was a type error, but I see no reason to be using a dict here, all references to it only use an `in` test. The same name attribute in client.py is also a set().

Dreamsorcerer:
OK, I've no idea what has just happened with multidict 6.0.1.

On this branch I get:
`aiohttp/typedefs.py:21: error: "CIMultiDict" expects 1 type argument, but 2 given  [type-arg]`
On master I get:
`aiohttp/typedefs.py:21: error: Type application has too few types (2 expected)  [misc]`

asvetlov:
Sorry, 2-arguments in 6.0.0 were introduced by a mistake.
6.0.1 reverts types to expected state.

Dreamsorcerer:
> Sorry, 2-arguments in 6.0.0 were introduced by a mistake. 6.0.1 reverts types to expected state.

I'll close this, but 6.0.1 is actually worse. Lots of type errors with no way to resolve them. Take a look at results in: https://github.com/aio-libs/multidict/pull/684

asvetlov:
I see, sorry for that.
Working on the new release. Multidict typing is hard :(

PR #6552, opened 2022-01-23
Minor fix

Dreamsorcerer:
There is a type error here, but I see no reason to be using a dict here, all references to it only use an `in` test. The same name attribute in client.py is also a set().


Dreamsorcerer:
Seems to actually rely on the case-insensitive feature of CIMultiDict. Probably not worth the extra work to make it work in a set.

PR #6555, opened 2022-01-23
Start rechecking types on tests

Dreamsorcerer:
Reenabling type checking in the tests.

This is the first batch, still a bunch of other files currently disabled.


tests/test_worker.py
```diff
@@ -121,32 +133,24 @@ def test_run_not_app(worker, loop) -> None:
     assert loop.is_closed()
 
 
-def test_handle_quit(worker, loop) -> None:
-    worker.loop = mock.Mock()
-    worker.handle_quit(object(), object())
-    assert not worker.alive
-    assert worker.exit_code == 0
-    worker.loop.call_later.asset_called_with(0.1, worker._notify_waiter_done)
```
Dreamsorcerer, line 123: This test has been silently failing for a long time (maybe it never even worked). I can't see any reason in the code that `loop.call_later()` would be called. If anyone knows anything about this test, please fix, otherwise I'll just remove it.

PR #6588, opened 2022-02-03
Make pip use `--user` in GHA CI/CD workflows

### LINKED ISSUE #6578: [TODO] Adjust the GHA setup compatibility with `pip >= 22`
## Long story short

The newly released version requires an explicit `--user` in envs where there is a risk of damaging the interpreter distribution failing loudly.

Example: https://github.com/aio-libs/aiohttp/runs/4998902391?check_suite_focus=true#step:10:166.

Dreamsorcerer:
Fixes #6578.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,7 @@ jobs:
       env:
         COLOR: 'yes'
         AIOHTTP_NO_EXTENSIONS: ${{ matrix.no-extensions }}
-        INSTALL_ARG: '--user'
+        PIP_USER: 1
       run: |
         make vvtest
         python -m coverage xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,7 @@ jobs:
       env:
         COLOR: 'yes'
         AIOHTTP_NO_EXTENSIONS: ${{ matrix.no-extensions }}
+        INSTALL_ARG: '--user'
       run: |
         make vvtest
         python -m coverage xml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -74,7 +74,7 @@ generate-llhttp: .llhttp-gen
 cythonize: .install-cython $(PYXS:.pyx=.c)
 
 .install-deps: .install-cython $(PYXS:.pyx=.c) $(call to-hash,$(CYS) $(REQS))
-	@pip install -r requirements/dev.txt -c requirements/constraints.txt
+	@pip install -r requirements/dev.txt -c requirements/constraints.txt --user
```
webknjaz: This probably won't work for the contributors as they usually run under venvs and using it with error out for them. I would probably try using either an env var or the config file just in the CI.

Dreamsorcerer: How about something like that?

.github/workflows/ci.yml
```diff
@@ -178,6 +178,7 @@ jobs:
       env:
         COLOR: 'yes'
         AIOHTTP_NO_EXTENSIONS: ${{ matrix.no-extensions }}
+        INSTALL_ARG: '--user'
```
webknjaz: All of pip CLI options have env var counterparts prefixed with `PIP_`. For example, here we could use
```suggestion
        PIP_USER: 1
```
which is the same as `--user`: https://pip.pypa.io/en/latest/topics/configuration/#environment-variables.

Dreamsorcerer: I'll give that a try then. Not sure if all commands should be --user or not, so we'll see.

PR #6591, opened 2022-02-03
Improve import time

Dreamsorcerer:
See https://github.com/aio-libs/aiohttp-debugtoolbar/issues/394#issuecomment-998280496


Dreamsorcerer:
This one is still failing even in user mode...

CHANGES/6591.misc
```diff
@@ -0,0 +1 @@
+Lazy load Gunicorn workers on import, decreasing import time by ~53%.
```
webknjaz: Let's use the past tense here to keep the resulting changelog consistent, writing style-wise.
```suggestion
Changed importing Gunicorn to happen on-demand, decreasing import time by ~53% -- :user:`Dreamsorcerer`.
```

tests/test_imports.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+
+def test___all__(pytester: pytest.Pytester) -> None:
+    """See https://github.com/aio-libs/aiohttp/issues/6197"""
+    pytester.makepyfile(
+        test_a="""
+            from aiohttp import *
```
webknjaz, line 13: Suppose this could be more accurate with:
```suggestion
            from aiohttp import *
            assert 'GunicornWebWorker' in globals()
```

tests/test_imports.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def test_import_time(pytester: pytest.Pytester) -> None:
+    """Check that importing aiohttp doesn't take too long.
+
+    Obviously, the time may vary on different machines and may need to be adjusted
+    from time to time, but this should provide an early warning if something is
+    added that significantly increases import time.
+    """
+    r = pytester.run("python3", "-c", "import aiohttp", timeout=0.3)
```
webknjaz: Why not use a subprocess?
Also, instead of using some abstract-maybe-different interpreter, I'd rely on `sys.executable`.
Plus, it'd be nice to bake in `-Werror`.

Dreamsorcerer: I was looking at the above tests as an example of how to run an external test, and then ended up looking at the other methods of `pytester`. But, I guess using `subprocess` would be just the same.

Dreamsorcerer: Seems that the way `pytester` handles it is faster and more reliable (I think sending stderr/stdout to a temp file instead of PIPE may be one reason, although that doesn't make much sense to me). It seems safer to stick with this.

Dreamsorcerer:
Any idea what this internal error from coverage is about? Seems to have just started happening when I switched it to `sys.executable`.

.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
```
Dreamsorcerer, line 1: I don't fully understand why, but this needs to be moved to make it run correctly:
https://github.com/nedbat/coveragepy/issues/512

Dreamsorcerer:
OK, think we're good now for a final review. Not sure why codecov thinks that the entire test I added is not being run, when it is clearly passing in the logs (in fact, it looks rather like it's only logged codecov for the Windows tests or something...).

In fact, the general codecov report looks a bit weird. For example, it claims that no code in `ClientSession` has run at all, which is not possible.

aiohttp/__init__.py
```diff
@@ -203,11 +210,24 @@
...
     "TraceResponseChunkReceivedParams",
+    # workers (imported lazily with __getattr__)
+    "GunicornUVLoopWebWorker",
+    "GunicornWebWorker",
 )
 
-try:
-    from .worker import GunicornUVLoopWebWorker, GunicornWebWorker
 
-    __all__ += ("GunicornWebWorker", "GunicornUVLoopWebWorker")
```
asvetlov, line 211: I guess `__dir__` should be overloaded also to keep the backward compatibility.

Dreamsorcerer: How exactly should we do that? I notice here, they just add `__doc__` and `__author__` to `__all__`: https://github.com/python/cpython/blob/main/Lib/concurrent/futures/__init__.py

But, that's a lot less than what it currently returns. It currently includes a bunch of things we are not exporting (e.g. modules), plus several more things like `__builtins__` and `__path__`.

hauntsaninja:
Thanks for fixing this! You may also be interested in https://github.com/benoitc/gunicorn/pull/2539 , which fixes the slowness in importing gunicorn

I've had to have hacky workarounds for this like https://github.com/hauntsaninja/boostedblob/blob/78986043e676806a1e2f8bc46606083b8000f332/boostedblob/__init__.py#L5 so really glad to see this change :-)

.coveragerc
webknjaz: Personally, I prefer separate config files for different tools. This way, I know where to go on the file system and within a config, instead of trying out many possible locations.

Also, https://github.com/nedbat/coveragepy/issues/512#issuecomment-1246595148 mentions that it's best to specify `--cov-config=.coveragerc` explicitly.

webknjaz: I think it'd be reasonable to have adjusting the coverage config in a separate PR.

webknjaz: Here's a more comprehensive config, by the way: https://github.com/ansible/pylibssh/blob/devel/.coveragerc.

tests/test_imports.py
```diff
@@ -0,0 +1,45 @@
+import platform
+import sys
+
+import pytest
+
+
+def test___all__(pytester: pytest.Pytester) -> None:
+    """See https://github.com/aio-libs/aiohttp/issues/6197"""
```
webknjaz, line 10: It'd be useful to see an actual docstring here, not a reference to something external.

aiohttp/__init__.py
Dreamsorcerer: @webknjaz Any opinions on what we return from `__dir__`? Just do the same as the linked code above?

tests/test_imports.py
Dreamsorcerer: I've not changed this test, the diff is just confused due to the file rename.

Dreamsorcerer:
> Thanks for fixing this! You may also be interested in [benoitc/gunicorn#2539](https://github.com/benoitc/gunicorn/pull/2539) , which fixes the slowness in importing gunicorn

Shame it's still not been merged.

Dreamsorcerer:
Hmm, seems the tests can't import aiohttp anymore. Maybe need to do something with PYTHONPATH...

aiohttp/__init__.py
webknjaz: Yeah, I guess that'd do.

tests/test_imports.py
```diff
@@ -0,0 +1,43 @@
...
+            from aiohttp.web import *
+        """
+    )
+    result = pytester.runpytest("-vv")
+    result.assert_outcomes(passed=0, errors=0)
+
+
+@pytest.mark.skipif(
+    not sys.platform.startswith("linux") or platform.python_implementation() == "PyPy",
+    reason="Timing is less reliable on other platforms",
```
webknjaz: Plz. Specify which platforms. Take into account that this string shows up in the test report.
Imaging seeing "SKIPPED: Timing is less reliable on other platforms" in the logs under macos — would it be helpful? Would you immediately guess that you're on an "other" platform?

tests/test_imports.py
webknjaz: So maybe do this in a separate PR, then?

tests/test_imports.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+    )
+    result = pytester.runpytest("-vv")
+    result.assert_outcomes(passed=0, errors=0)
+
+
+def test_web___all__(pytester: pytest.Pytester) -> None:
+    pytester.makepyfile(
+        test_b="""
+            from aiohttp.web import *
```
webknjaz, line 24: Does this test anything https://github.com/aio-libs/aiohttp/blob/77d3d68/tests/test_circular_imports.py doesn't? Maybe just parametrize that test to check both `import ...` and `from ... import ...` syntaxes?

Dreamsorcerer:
OK, I'm totally lost on why we always get an import error. Anybody got any ideas how to fix this?

webknjaz:
Maybe because you're patching PYTHONPATH? Old tests didn't do that. And I don't know if pytester changes the directory (it probably does).

webknjaz:
Also, you modify environ for the whole test process, not just one test. This will have weird side-effects.

Dreamsorcerer:
> Maybe because you're patching PYTHONPATH? Old tests didn't do that. And I don't know if pytester changes the directory (it probably does).

That was me trying to fix the issue, not the cause.

I think it might have been working before merging master, let's see if it works on an older pytest release, maybe some change there broke it..

Dreamsorcerer:
OK, I've even pushed a separate branch from the time the tests were successful, but they are failing now: https://github.com/aio-libs/aiohttp/pull/7060

Seems like something outside of our code has changed the CI behaviour, so it no longer imports...

webknjaz:
Does it pass locally for you?

Dreamsorcerer:
OK, seem to have got it working now, even if it seems a bit awkward. The Python invocation seems to end up with half as many entries in `sys.path` than the main process does, so I've had to copy it into PYTHONPATH to get it to run with the same path values.

Dreamsorcerer:
OK, full CI run passed twice with a threshold of 410ms. Think we can merge this and take a look at the other 2 PRs relating to import time.

PR #6592, opened 2022-02-03
Add is_multipart parameter to FormData

### LINKED ISSUE #6587: Set is_multipart for FormData

I would like to send list of key-value strings as multipart/form-data using client library. For that, I tried to use FormData:

```python
form = aiohttp.FormData()
form.add_field('key', 'value')
```

But server that receive request, receive `application/x-www-form-urlencoded` string, instead of `multipart/from-data`. By looking into source code of `FormData` I found that `is_multipart` flag is set to `True` when I pass bytes or file as value. One way to force FormData to form multipart-form/data is to set private argument manually.
```python
form._is_multipart = True
```

I would like to have non-private option for `FormData` class to use `multipart-form/data`. I may be an option to `__init__.py` or it may be a setter for `is_multipart` property.

Using:

```python
form._is_multipart = True
```

### Related component

Client

### Additional context

_No response_

### Code of Conduct

hyzyla:
## What do these changes do?

Allows to set `is_multipart` parameter to `FormData`. It can be helpful in situation when you have only values as a string and whant to send it as multipart/form-data

## Related issue number

Fixes #6587

## Checklist

### DIFF
--- a/CHANGES/6587.feature
+++ b/CHANGES/6587.feature
@@ -1,0 +1,1 @@
+Add is_multipart parameter to FormData
--- a/aiohttp/formdata.py
+++ b/aiohttp/formdata.py
@@ -23,11 +23,12 @@ class FormData:
         quote_fields: bool = True,
         charset: Optional[str] = None,
         boundary: Optional[str] = None,
+        is_multipart: bool = False,
     ) -> None:
         self._boundary = boundary
         self._writer = multipart.MultipartWriter("form-data", boundary=self._boundary)
         self._fields: List[Any] = []
-        self._is_multipart = False
+        self._is_multipart = is_multipart
         self._is_processed = False
         self._quote_fields = quote_fields
         self._charset = charset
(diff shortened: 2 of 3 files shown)


hyzyla:
It's a quick solution to the problem that I have, but I'm open to the ideas how it could be solved in another way. One question that I have, what we should do when some sets `FormData(is_multipart=False)` and then adds binary file that change internal state of attribute `self._is_multipart` to False? Maybe raise an error?

Dreamsorcerer:
@hyzyla If you'd like to reopen this PR, this looks reasonable to me.

> One question that I have, what we should do when some sets `FormData(is_multipart=False)` and then adds binary file that change internal state of attribute `self._is_multipart` to False? Maybe raise an error?

I'd suggest renaming the parameter that better describes the current behaviour. Maybe `default_to_multipart` or similar. So, setting it to False doesn't imply that it won't produce a multipart response.

Dreamsorcerer:
@hyzyla Just checking if you're going to get back to this one?

hyzyla:
@Dreamsorcerer go ahead! I don't mind if you reopen this pull request 🤝

PR #6594, opened 2022-02-05
Export HTTPMove

Dreamsorcerer:
Nitpick: If you catch `HTTPRedirection` then mypy complains that it doesn't have a `location` attribute (this is used in aiohttp-debugtoolbar, for example).

Proposal to export `HTTPMove`, in order to catch any exceptions with location.


Dreamsorcerer:
For some reason this is a private class in the 3.9 branch, so it'll need to be made public like master in the backport.

PR #6601, opened 2022-02-09
Update copyright year from 2020 to 2022

meitham:
## What do these changes do?
A simple update to copyright year from 2020 to 2022

No change in behaviour, but simply giving the docs reader a sense that the project is up-to-date.  

## Related issue number
None

## Checklist

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-   Copyright aio-libs collaboration.
+   Copyright aio-libs contributors.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-   Copyright 2013-2022 aio-libs collaboration.
+   Copyright aio-libs collaboration.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-   Copyright 2013-2020 aio-libs collaboration.
+   Copyright 2013-2022 aio-libs collaboration.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -99,7 +99,7 @@ github_repo_url = f"{github_url}/{github_repo_slug}"
 github_sponsors_url = f"{github_url}/sponsors"
 
 project = github_repo_name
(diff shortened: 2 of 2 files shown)


webknjaz:
It's better to drop the years or just keep the first one.

Dreamsorcerer:
> It's better to drop the years or just keep the first one.

I think from a legal perspective, this range is technically correct. Copyright is from the date the code was written, so this says parts of the project are copyrighted in 2013, 2014, ..., and 2022. Although it is a bit of a hassle needing to keep it updated..

meitham:
> It's better to drop the years or just keep the first one.

I saw a discussion recently on HN around dropping the copyright year but I'm failing to find the link.  The copyright year came in a discussion around picking up a Python web framework and it was pointed that aiohttp might not be as maintained as FastAPI purely based on the date in that copyright notice!

webknjaz:
> I think from a legal perspective, this range is technically correct.

Not exactly. Copyright is associated with the year of content creation and remains valid throughout the future years. When the content is updated, there's some part of the content that is new so this update copyright gets a new year. The copyright is held by the author regardless of the statement in the files. That line can make it easier to track it but is not really required. It only creates maintenance burden.

Here's a good guidance to follow https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/. Plus there's Hynek's note at https://hynek.me/til/copyright-years/.

Dreamsorcerer:
Yep, exactly. I think this is the technically more accurate approach. But, I think removing the years altogether makes sense, so I'd vote for that.

meitham:
> > I think from a legal perspective, this range is technically correct.
> 
> Not exactly. Copyright is associated with the year of content creation and remains valid throughout the future years. When the content is updated, there's some part of the content that is new so this update copyright gets a new year. The copyright is held by the author regardless of the statement in the files. That line can make it easier to track it but is not really required. It only creates maintenance burden.
> 
> Here's a good guidance to follow https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/. Plus there's Hynek's note at https://hynek.me/til/copyright-years/.

Thanks, yes it was Hynek's post on HN that I have failed to find earlier.  I have now removed the years from the copyright notice.

webknjaz:
Yeah, I suggested to add that LF post to the article when he posted it on Twitter :) This is how I remembered where to look.

webknjaz:
@meitham while on it, could you update the text to "aio-libs contributors"? It seems weird that they are different.

meitham:
> aio-libs contributors"

done

webknjaz (approved):
Brilliant, thanks!

CryptoGodfatherVA2 (commented):
Yes

PR #6607, opened 2022-02-10
[TODO] Fix Windows tests

Dreamsorcerer:
Windows tests have been failing silently.

Feel free to add any fixes directly to this PR.


webknjaz:
Another solution would be to add `set -Euo pipefail` + add `shell: bash` for consistency. Also, why can't pytest-cov generate xml right away?

Dreamsorcerer:
> Another solution would be to add `set -Euo pipefail` + add `shell: bash` for consistency. Also, why can't pytest-cov generate xml right away?

Feel free to make any changes. When the CI steps are run on Windows, it seems that the step passes if the last command succeeded, whereas the others will pass if all commands succeed. The coverage command always succeeds, therefore the tests always pass regardless of the pytest result. Running one command per step is a simple solution.

Dreamsorcerer:
I just realised you're asking why `python -m coverage xml` needs to be run at all. No idea, I didn't write it.

PR #3627, opened 2019-02-27
Store cookies with the same name if they have different paths

jsagarribay:
## What do these changes do?

Do not overwrite cookies when they have different paths, even if they have the same name and domain

## Are there changes in behavior for the user?

Without this change, if the server sets the same cookie (for example JSESSIONID) in several paths the cookie will be overwritten after each request and the session will get lost.

With this change aiohttp will do the same as browsers and store both cookies.

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #6636, opened 2022-02-25
Fix cookie handling

bratao:
## What do these changes do?

There is an error in cookie handling that override the existing cookie if the path is different. This is a huge bug in aiohttp, as it makes requests to some sites impossible to work.

The RFC 6265 is clear on this point (https://datatracker.ietf.org/doc/html/rfc6265) , for example:

```
Finally, to remove a cookie, the server returns a Set-Cookie header
   with an expiration date in the past.  The server will be successful
   in removing the cookie only **if the Path and the Domain attribute** in
   the Set-Cookie header match the values used when the cookie was
   created.

```

Actually aiohttp only consider domain and name. Pull request #3627 tried to fix this, but the code has been updated vastly since then. 

No

## Checklist


tests/test_cookiejar.py
```diff
@@ -664,6 +664,28 @@ async def make_jar():
...
+        )
+        self.assertEqual(len(jar), 2)
+
+        jar_filtred = jar.filter_cookies(URL("http://pathtest.com/"))
+        self.assertEqual(len(jar_filtred), 1)
+        self.assertEqual(jar_filtred["path-cookie"].value, "zero")
+
+        jar_filtred = jar.filter_cookies(URL("http://pathtest.com/one"))
+        self.assertEqual(len(jar_filtred), 1)
+        self.assertEqual(jar_filtred["path-cookie"].value, "one")
```
Dreamsorcerer, line 687: filtred -> filtered

Dreamsorcerer (changes requested):
Afraid you'll need to rebase or something to merge into master branch. From there we can backport to 3.9 (not sure we should include this in 3.8).

PR #6638, opened 2022-02-28
Fix cookie handling

bratao:
## What do these changes do?

There is an error in cookie handling that override the existing cookie even if the path is different. This is a huge bug in aiohttp, as it makes requests to some sites impossible to work.

This is an updated PR as asked by @Dreamsorcerer 

The RFC 6265 is clear on this point (https://datatracker.ietf.org/doc/html/rfc6265) , for example:

```
Finally, to remove a cookie, the server returns a Set-Cookie header
   with an expiration date in the past.  The server will be successful
   in removing the cookie only **if the Path and the Domain attribute** in
   the Set-Cookie header match the values used when the cookie was
   created.

```

Actually aiohttp only consider domain and name. Pull request #3627 tried to fix this, but the code has been updated vastly since then. 

Pay attention to another library, such as requests. It always consider the path as a key to the cookie jar.
https://github.com/psf/requests/blob/79f60274f7e461b8fd2f579e741f748438d7eadb/requests/cookies.py#L189

No

## Checklist

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -65,7 +65,7 @@ class CookieJar(AbstractCookieJar):
         treat_as_secure_origin: Union[StrOrURL, List[StrOrURL], None] = None
     ) -> None:
         self._loop = asyncio.get_running_loop()
-        self._cookies: DefaultDict[str, SimpleCookie[str]] = defaultdict(SimpleCookie)
+        self._cookies: DefaultDict[Tuple[str, str], SimpleCookie[str]] = defaultdict(SimpleCookie)
         self._host_only_cookies: Set[Tuple[str, str]] = set()
         self._unsafe = unsafe
         self._quote_cookie = quote_cookie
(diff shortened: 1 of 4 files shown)


aiohttp/cookiejar.py
```diff
@@ -110,20 +112,21 @@ def clear(self, predicate: Optional[ClearCookiePredicate] = None) -> None:
...
+                key = (domain, path, name)
                 if (
                     key in self._expirations and self._expirations[key] <= now
                 ) or predicate(morsel):
                     to_del.append(key)
 
-        for domain, name in to_del:
+        for domain, path, name in to_del:
             key = (domain, name)
             self._host_only_cookies.discard(key)
```
```suggestion
            self._host_only_cookies.discard((domain, name))
```

As this is not being reused...

Dreamsorcerer: Should also resolve your mypy error.

Dreamsorcerer: Should also resolve your mypy error.

Dreamsorcerer (approved):
If no one else reviews, I'll merge it at the weekend.

Dreamsorcerer:
If you could handle the backport to 3.9 (as per instructions above), that would be great.

PR #6658, opened 2022-03-18
Support chunked=False

eerimoq:
## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -585,7 +585,7 @@ class ClientRequest:
         if self.compress:
             writer.enable_compression(self.compress)
 
-        if self.chunked is not None:
+        if self.chunked:
             writer.enable_chunking()
 
         # set default content-type


Dreamsorcerer:
As far as I can tell the `None` usage dates back to the start of the git history, 9 years ago. So, I'm inclined to say that it was probably a mistake.

If so, we should also change the type of the parameters to `bool`. It doesn't look like there's any reason for the argument to be `None` (though maybe we should leave it in the 3.x branch to keep backwards compatibility).

eerimoq:
Maybe `None` should mean that `aiohttp` shall set it to `True / False` based on other request parameters, but if given as `True / False` to `get() / post() / ...` by the user `aiohttp` should do as told. Just an idea...

My use case is to start POST request and leave the connection open. Data is written to this requests body many times during the programs execution. Tunneling data over HTTP in other words. My initial testing showed that chunking is not supported by the server. You can see this document for more info: https://opensource.apple.com/source/QuickTimeStreamingServer/QuickTimeStreamingServer-412.42/Documentation/RTSP_Over_HTTP.pdf

eerimoq:
I just found out that if `Content-Length` is part of `headers` it will not be chunked in my use case. So this PR is not needed for me personally anymore, but could possibly be useful to others? If not, feel free to close it.

PR #6682, opened 2022-04-08
fix: CookieJar quote_cookie / unsafe args lost when input cookies to _request

exherb:
## What do these changes do?

the result of those two code blocks are not the same.

```python
import aiohttp
jar = aiohttp.CookieJar(quote_cookie=False)
async with aiohttp.ClientSession(cookie_jar=jar, cookies=cookies) as session:
  async with session.post(url) as r:
    print(await r.json())  
```

```python
import aiohttp
jar = aiohttp.CookieJar(quote_cookie=False)
async with aiohttp.ClientSession(cookie_jar=jar) as session:
  async with session.post(url, cookies=cookies) as r:  # cookies value got quoted even current cookie_jar quote_cookie is False
    print(await r.json())  
```

no

## Checklist

### DIFF
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -13,6 +13,7 @@ from typing import (
     List,
     Optional,
     Tuple,
+    TypeVar,
 )
 
 from multidict import CIMultiDict
@@ -134,10 +135,15 @@ else:
 
 
 ClearCookiePredicate = Callable[["Morsel[str]"], bool]
+_CookieJar = TypeVar("_CookieJar", bound="AbstractCookieJar")
 
 
 class AbstractCookieJar(Sized, IterableBase):
     """Abstract Cookie Jar."""
+
+    @abstractmethod
+    def clone(self: _CookieJar) -> _CookieJar:
+        """Clone."""
 
     @abstractmethod
     def clear(self, predicate: Optional[ClearCookiePredicate] = None) -> None:
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -92,6 +92,13 @@ class CookieJar(AbstractCookieJar):
         except OverflowError:
             self._max_time = self.MAX_32BIT_TIME
 
+    def clone(self) -> "CookieJar":
+        return self.__class__(
+            unsafe=self._unsafe,
+            quote_cookie=self._quote_cookie,
+            treat_as_secure_origin=self._treat_as_secure_origin,
+        )
+
     def save(self, file_path: PathLike) -> None:
         file_path = pathlib.Path(file_path)
         with file_path.open(mode="wb") as f:
@@ -398,6 +405,9 @@ class DummyCookieJar(AbstractCookieJar):
     It can be used with the ClientSession when no cookie processing is needed.
 
     """
+
+    def clone(self) -> "DummyCookieJar":
+        return self.__class__()
 
     def __iter__(self) -> "Iterator[Morsel[str]]":
         while False:
--- a/CHANGES/6682.bugfix
+++ b/CHANGES/6682.bugfix
@@ -1,0 +1,1 @@
+CookieJar quote_cookie / unsafe args lost when input cookies to _request
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -119,6 +119,7 @@ Eugene Nikolaiev
 Eugene Tolmachev
 Evan Kepner
 Evert Lammerts
+Exherb
 Felix Yan
 Fernanda Guimarães
 FichteFoll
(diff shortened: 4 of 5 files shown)


Dreamsorcerer:
It's unclear what this PR is supposed to do, and most of the changes are unrelated. Please don't start changing a random selection of annotations, they won't be accepted.

exherb:
> It's unclear what this PR is supposed to do, and most of the changes are unrelated. Please don't start changing a random selection of annotations, they won't be accepted.

it's changed by robot, not me. https://github.com/aio-libs/aiohttp/pull/6682/commits/fc26cd7cde87caef91bb5e7c3a4ecd9507dacb6d

Dreamsorcerer:
> > It's unclear what this PR is supposed to do, and most of the changes are unrelated. Please don't start changing a random selection of annotations, they won't be accepted.
> 
> it's changed by robot, not me. [fc26cd7](https://github.com/aio-libs/aiohttp/commit/fc26cd7cde87caef91bb5e7c3a4ecd9507dacb6d)

Ah, sorry. I'll look into that, not sure why that's changed.

exherb:
ok, thanks. And I just updated why submit this pr.

aiohttp/cookiejar.py
```diff
@@ -92,6 +92,13 @@ def __init__(
         except OverflowError:
             self._max_time = self.MAX_32BIT_TIME
 
+    def clone(self) -> AbstractCookieJar:
+        return CookieJar(
```
Dreamsorcerer: Should probably use `self.__class__` or similar, so it works as expected for subclasses.

aiohttp/abc.py
```diff
@@ -139,6 +139,10 @@ async def close(self) -> None:
 class AbstractCookieJar(Sized, IterableBase):
     """Abstract Cookie Jar."""
 
+    @abstractmethod
+    def clone(self) -> "AbstractCookieJar":
```
Dreamsorcerer: Return type would need to be a TypeVar, so that this should return the same class, rather than AbstractCookieJar.
e.g. https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html?highlight=TypeVar#decorators

aiohttp/cookiejar.py
```diff
@@ -92,6 +92,13 @@ def __init__(
         except OverflowError:
             self._max_time = self.MAX_32BIT_TIME
 
+    def clone(self) -> AbstractCookieJar:
```
Dreamsorcerer: TypeVar again.

Dreamsorcerer (changes requested):
At an initial look, it might be reasonable.
It will need a new test though, ideally one that reproduces the problem described in the PR description.

Dreamsorcerer:
Weird that the pre-commit check is now passing without making those changes...

Dreamsorcerer:
Ah, I think you added the future import in your commit, which appears to trigger it:
https://github.com/asottile/pyupgrade#pep-585-typing-rewrites

You've correctly removed that import in the new code.

aiohttp/abc.py
exherb: could you explain more about this? like this?

```python
CookieJar = TypeVar("CookieJar", bound="AbstractCookieJar")
def clone(self: CookieJar) -> CookieJar:
```

Dreamsorcerer: Looks about right. I'd name it `_CookieJar` to keep it private.

exherb:
> Ah, I think you added the future import in your commit, which appears to trigger it: https://github.com/asottile/pyupgrade#pep-585-typing-rewrites
> 
> You've correctly removed that import in the new code.

maybe import annotation  from future will make typing clear?

aiohttp/abc.py
exherb: I have updated my commit, please review when got time

exherb:
> At an initial look, it might be reasonable.
> It will need a new test though, ideally one that reproduces the problem described in the PR description.

Ok, I will ad some test latter

Dreamsorcerer:
> maybe import annotation from future will make typing clear?

The annotations are correct for Python 3.7+, which is what we support. They will get updated as we drop support for old versions.

If we were to make those changes, then it would cause problems with anything that tried to inspect annotations at runtime.

exherb:
> At an initial look, it might be reasonable.
> It will need a new test though, ideally one that reproduces the problem described in the PR description.

I found quote_cookie = False already tested, I have no idea to how to test aiohttp.Client, any suggestions?

exherb:
@Dreamsorcerer

Dreamsorcerer:
I think you should be able to figure it out by copying one of the tests in: https://github.com/aio-libs/aiohttp/blob/master/tests/test_client_functional.py

But, if you struggle to get hold of the cookies, you could maybe use `mock.patch.object()` to mock out self._request_class, and then check what cookies get passed to that function?

PR #6683, opened 2022-04-09
[doc] clarify ClientTimeout behavior

rafzi:
## What do these changes do?

Be clear that the 5 second timeout is also rounded. see: https://github.com/aio-libs/aiohttp/blob/master/aiohttp/helpers.py#L638

No

### DIFF
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1713,7 +1713,7 @@ ClientTimeout
 
    .. note::
 
-      Timeouts larger than 5 seconds are rounded for scheduling on the next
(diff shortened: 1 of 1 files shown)


docs/client_reference.rst
```diff
@@ -1713,7 +1713,7 @@ ClientTimeout
 
    .. note::
 
-      Timeouts larger than 5 seconds are rounded for scheduling on the next
+      Timeouts larger than or equal to 5 seconds are rounded for scheduling on the next
```
```suggestion
      Timeouts of 5 seconds or more are rounded for scheduling on the next
```

Maybe this is a little more succinct.

PR #6700, opened 2022-04-13
fix ClientResponseError docstring

ntextreme3:
## What do these changes do?

ClientResponseError is currently described as:

- [/aiohttp/client_exceptions.py](https://github.com/aio-libs/aiohttp/blob/6b35a466e99eabda70befb90cf8545f3b987ab18/aiohttp/client_exceptions.py#L49) - "Connection error during reading response."
- [/docs/client_reference.rst](https://github.com/aio-libs/aiohttp/blame/918fe4c520812b4b8b5e6f927dd6554ad3917031/docs/client_reference.rst#L1433) - "These exceptions could happen after we get response from server."

Since ClientResponseError isn't a connection error, this PR reconciles the docstring with the `.rst`. 

I figured this would be a good place to make the docstring the "Source of Truth" and use `autodoc`, so that these docs won't diverge again, however I noticed that [autodoc was dropped in a previous commit](https://github.com/aio-libs/aiohttp/commit/2f78996d8e2ae55d018a2dc37c4f20c234be99cd), so I left it alone. If you're interested in adding it back, I can look into adding that here. 

No

## Checklist


Dreamsorcerer:
Deduplicating the documentation seems like a good idea, but I'd like to hear from @asvetlov if there is a reason that wasn't done first. And, it make take a while to get a reply at the moment.

Dreamsorcerer:
If it's a small task to make a start, then leaving an open PR around would be the best way to make sure it gets looked at eventually.

PR #6708, opened 2022-04-20
Stop using the cgi module, deprecated in Python 3.11 per PEP 594

scop:
## What do these changes do?

Removes use of the `cgi` module which is deprecated in 3.11+ per https://peps.python.org/pep-0594/#cgi

No

## Related issue number

No

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -3,7 +3,6 @@
 import asyncio
 import base64
 import binascii
-import cgi
 import dataclasses
 import datetime
 import enum
@@ -19,6 +18,7 @@ import warnings
 import weakref
 from collections import namedtuple
 from contextlib import suppress
+from email.parser import HeaderParser
 from email.utils import parsedate
 from http.cookies import SimpleCookie
 from math import ceil
@@ -748,7 +748,11 @@ class HeadersMixin:
             self._content_type = "application/octet-stream"
             self._content_dict = {}
         else:
-            self._content_type, self._content_dict = cgi.parse_header(raw)
+            msg = HeaderParser().parsestr("Content-Type: " + raw)
+            self._content_type = msg.get_content_type()
+            params = msg.get_params()
+            del params[0]  # content type duplicated here
+            self._content_dict = dict(params)
 
     @property
     def content_type(self) -> str:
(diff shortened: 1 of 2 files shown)


scop:
Untested and mostly for discussion purposes for now, quite likely some refinements would be in order.

aiohttp/helpers.py
```diff
@@ -748,7 +748,11 @@ def _parse_content_type(self, raw: str) -> None:
             self._content_type = "application/octet-stream"
             self._content_dict = {}
         else:
-            self._content_type, self._content_dict = cgi.parse_header(raw)
+            msg = HeaderParser().parsestr("Content-Type: " + raw)
+            self._content_type = msg.get_content_type()
```
scop, line 752: One difference compared to cgi is that if the actual type is invalid (for example `foo`), this returns `text/plain`.

Dreamsorcerer: Documentation here suggest to use `Message`: https://peps.python.org/pep-0594/#cgi
Though the result looks much the same.

aiohttp/helpers.py
```diff
@@ -748,7 +748,11 @@ def _parse_content_type(self, raw: str) -> None:
...
+            params = msg.get_params()
+            del params[0]  # content type duplicated here
+            self._content_dict = dict(params)
```
```suggestion
            self._content_dict = dict(params[1:])  # First element is content type again
```
Maybe a list slice is simpler?

aiohttp/helpers.py
scop: Yeah, I know.

But to me the _example_ (it's labeled as such, not a suggestion) in the PEP isn't that great -- to parse a HTTP header one would create an e-mail Message, set a header to it and then grab the header's parameters back... quite ugly in my opinion. The approach I took here to me reads much better: parse a header using a header parser.

Another reason using a Message doesn't sound that great to me in new code is that it is labeled as a compat/legacy thing:
- https://docs.python.org/3/library/email.compat32-message.html#compat32-message
- https://docs.python.org/3/library/email.message.html#id2

aiohttp/helpers.py
scop: Yes, definitely, thanks!

Dreamsorcerer:
If you could also handle the backport, that would be great.

scop:
https://github.com/aio-libs/aiohttp/pull/6734

webknjaz:
Backported to 3.8 as https://github.com/aio-libs/aiohttp/commit/cb79f0dc1597b683ab5a5a9fed8262954a086d35

Issue #2335, opened 2017-10-17
Support request_info and history in ClientError


asvetlov:
Otherwise the context is missing for exceptions like `ClientConnectionError`.
It makes using `asyncio.gather` very complicated

webknjaz:
Ref: #2507

PR #6722, opened 2022-04-27
Implement granular URL error hierarchy in the HTTP client

### LINKED ISSUE #2507: Implement a better way of handling non-HTTP urls in client.ClientSession
## Long story short

This is a follow-up for https://github.com/aio-libs/aiohttp/issues/2479#issuecomment-343559915.
Currently, if one uses `ClientSession` and it reaches URL with custom schema the exception with insufficient metadata is being raised.

## Expected behaviour

There should be some way to give control back to the caller once we reach **valid URL**, which we cannot follow because of unsupported schema. @asvetlov suggested to implement a custom exception for this.

## Actual behaviour

```python
ValueError: Can redirect only to http or https
```

## Steps to reproduce

Query `https://steamcommunity.com/mobileconf/conf` or `steammobile://lostauth/`, or `twitter://webknjaz`

### LINKED ISSUE #2630: aiohttp raises ValueError for invalid Location: in redir
## Long story short

I am crawling lots of sites, and I don't expect aiohttp to be raising ValueError.

## Expected behaviour

Raise an exception that's documented

## Actual behaviour

Yarl's ValueError is passed up to the top

## Steps to reproduce

```python
import aiohttp
import asyncio
import sys

hostname = None

            return await response.text()

async def main():
        html = await fetch(session, hostname)
        print(html)

hostname = sys.argv[1]
```

```
$ python ./bug.py http://www.oberoithreesixtywest.in/
...
  File "/home/lindahl/.pyenv/versions/3.6.4/lib/python3.6/site-packages/yarl/__init__.py", line 348, in origin
    raise ValueError("URL should be absolute")
ValueError: URL should be absolute
```
Note:
```
$ curl -D /dev/tty http://www.oberoithreesixtywest.in/
HTTP/1.1 301 Moved Permanently
Server: nginx
Date: Thu, 28 Dec 2017 06:35:33 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
X-Frame-Options: GOFORIT
Location: http://
```

## Your environment

aiohttp==2.3.6
yarl==0.16.0

### LINKED ISSUE #3315: ValueError when redirecting to URL with incorrect schema
## Long story short

When the Client handles redirection, if the redirected URL contains an incorrect schema, it throws a `ValueError`:

https://github.com/aio-libs/aiohttp/blob/ced10580c1f8f0673ba8b59986f44f8d4605e35d/aiohttp/client.py#L453

While this is technically correct, it is more sensible to expect some kind of more specific error.

## Expected behaviour

I would expect the Client to throw an `aiohttp`-specific error, preferably `InvalidURL`.

## Actual behaviour

The Client throws a generic `ValueError`, which is difficult to specifically handle.

## Steps to reproduce

Create a redirection URL, e.g. using an URL shortener like https://tinyurl.com/, and have it redirect to any URL with an incorrect schema, for example `file://foo/bar`. Then, call it with `aiohttp`

    import aiohttp, asyncio

    async def get_url(url):
        async with aiohttp.ClientSession() as s, s.get(url) as resp:
            print(await resp.text())

    asyncio.get_event_loop().run_until_complete(get_url('https://tinyurl.com/y9rf58n8'))
    
Output:

      File "/usr/lib/python3.6/asyncio/base_events.py", line 468, in run_until_complete
        return future.result()
      File "<stdin>", line 2, in get_url
      File "/home/berislav/.virtualenvs/foobar/lib/python3.6/site-packages/aiohttp/client.py", line 855, in __aenter__
        self._resp = await self._coro
      File "/home/berislav/.virtualenvs/foobar/lib/python3.6/site-packages/aiohttp/client.py", line 456, in _request
        'Can redirect only to http or https')
    ValueError: Can redirect only to http or https

## Your environment

Python 3.6.6, and the installed packages:

    Package       Version
    ------------- -------
    aiohttp       3.4.4  
    async-timeout 3.0.0  
    attrs         18.2.0 
    chardet       3.0.4  
    idna          2.7    
    idna-ssl      1.1.0  
    multidict     4.4.2  
    pip           18.0   
    setuptools    40.4.3 
    wheel         0.32.0 
    yarl          1.2.6

setla:
## What do these changes do?

Added new exception type to separate cases with invalid redirect url during request. Previously ValueError or InvalidURL was raised and screening out was hard (valid url that redirects to invalid one raised the same error as invalid url).

Slightly improved test coverage.

Docs changes previews:
* https://aiohttp--6722.org.readthedocs.build/en/6722/client_reference.html#hierarchy-of-exceptions
* https://aiohttp--6722.org.readthedocs.build/en/6722/changes.html#features

In case of invalid redirect url during request - new InvalidRedirectUrl will be raised instead of ValueError or InvalidURL, it derives from InvalidURL so it's backward compatibile

## Related issue number

Resolves #2507
Resolves #2630
Resolves #3315

## Checklist


aiohttp/client.py
```diff
@@ -574,34 +576,40 @@ async def _request(
...
-                            and parsed_url.scheme == "https"
+                            url.host == parsed_redirect_url.host
+                            and parsed_redirect_url.scheme == "https"
                             and url.scheme == "http"
                         )
 
+                        try:
+                            redirect_origin = parsed_redirect_url.origin()
+                        except ValueError as e:
+                            raise InvalidRedirectUrl(parsed_redirect_url) from e
```
```suggestion
                            raise InvalidRedirectUrl(parsed_redirect_url)
```
from e is the default behaviour.

Dreamsorcerer: Actually, they are subtly different, this is probably fine.

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,31 @@ class InvalidURL(ClientError, ValueError):
...
 
-    def __init__(self, url: Any) -> None:
+    def __init__(self, url: Any, description: Optional[str] = None) -> None:
         # The type of url is not yarl.URL because the exception can be raised
         # on URL(url) call
-        super().__init__(url)
+        super_args = [url]
+        if description:
+            super_args.append(description)
+        super().__init__(*super_args)
```
Dreamsorcerer: Think I'd prefer the more direct approach here, personally.
```suggestion
        if description:
            super().__init__(url, description)
        else:
            super().__init__(url)
```

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,31 @@ class InvalidURL(ClientError, ValueError):
...
 
     @property
     def url(self) -> Any:
         return self.args[0]
 
+    @property
+    def description(self) -> Any:
+        return self.args[1]
```
Dreamsorcerer: This would error if description wasn't passed.
Also return type should not be `Any`.

tests/test_client_functional.py
```diff
@@ -2166,6 +2166,36 @@ async def handler_redirect(request):
...
+    ),
+)
+async def test_invalid_redirect_url(
+    aiohttp_client: Any, invalid_redirect_url: Any
+) -> None:
+    body = b"redirect"
+    headers = {hdrs.LOCATION: invalid_redirect_url}
+
+    async def handler_request(request):
+        return web.Response(status=301, body=body, headers=headers)
```
Dreamsorcerer: Body is irrelevant, right? Can we just drop it?

aiohttp/client_exceptions.py
setla: Sure, updated.

aiohttp/client_exceptions.py
setla: Right, I lost KeyError here - added. There is problem with typing:
`aiohttp/client_exceptions.py:257: error: Returning Any from function declared to return "Optional[str]"`
That's why `def url` has Any return as well. I updated both to match it's real type and added `# type: ignore[no-any-return]` - this seems to be better

tests/test_client_functional.py
setla: Right, it's not needed - removed.

aiohttp/client_exceptions.py
Dreamsorcerer: For url, the argument type is `Any`, so the return type would need to match that. Why it doesn't have a more precise type, I'm not sure, but it's unclear to me whether it should always be `str` or not (I'd need to take a closer look to check).

Dreamsorcerer: A type ignore comment is fine. Another option would be to cast it, or maybe even annotate the type in the class like `args: Tuple[str, str]`

aiohttp/client.py
setla: Yup, traceback seems to be just a little bit more clear with `from`.

aiohttp/client_exceptions.py
setla: I updated `url` type to `StrOrURL` - `yarl.URL` can be passed to this exception as well. Regarding args - there is no access to overwrite type as its `Exception` thing

setla:
@Dreamsorcerer can I get some clue if (and when if so) it's going to be merged?

webknjaz:
@setla looks like a few files diverged since your last update — could you fix the merge conflicts?

@Dreamsorcerer has this slipped off of your review queue?

CHANGES/6722.feature
```diff
@@ -0,0 +1 @@
+Added ``InvalidRedirectURL`` which is now raised instead ``ValueError`` or ``InvalidURL`` when redirect url is invalid
```
webknjaz:
```suggestion
Added an ``InvalidRedirectURL`` exception which is now raised instead of
:py:exc:`ValueError` or ``InvalidURL`` when the redirect URL
is invalid -- by :user:`setla`.

The ``InvalidURL`` exception (and
``InvalidRedirectURL``, that inherits it), now exposes
a ``description`` property with the text explanation of
the error details.
```

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,34 @@ class InvalidURL(ClientError, ValueError):
 
     # Derive from ValueError for backward compatibility
 
-    def __init__(self, url: Any) -> None:
+    def __init__(self, url: StrOrURL, description: Optional[str] = None) -> None:
```
webknjaz: Pls use `Union` for clarity.
```suggestion
    def __init__(self, url: StrOrURL, description: Union[str, None] = None) -> None:
```

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,34 @@ class InvalidURL(ClientError, ValueError):
...
         # The type of url is not yarl.URL because the exception can be raised
         # on URL(url) call
-        super().__init__(url)
+        if description:
+            super().__init__(url, description)
```
webknjaz, line 263: I think, this could be unconditional.

tests/test_client_exceptions.py
```diff
@@ -280,10 +283,22 @@ def test_pickle(self) -> None:
             assert err2.url == ":wrong:url:"
             assert err2.foo == "bar"
 
-    def test_repr(self) -> None:
+    def test_repr_no_description(self) -> None:
         err = client.InvalidURL(url=":wrong:url:")
         assert repr(err) == "<InvalidURL :wrong:url:>"
```
webknjaz, line 289: Check that description is None, maybe?

docs/client_reference.rst
```diff
@@ -2271,3 +2283,5 @@ Hierarchy of exceptions
   * :exc:`ClientPayloadError`
 
   * :exc:`InvalidURL`
+
+    * :exc:`InvalidRedirectURL`
```
webknjaz: This looks overindented
```suggestion
  * :exc:`InvalidRedirectURL`
```

docs/client_reference.rst
```diff
@@ -2077,6 +2077,18 @@ All exceptions are available as members of *aiohttp* module.
 
       Invalid URL, :class:`yarl.URL` instance.
 
+    .. attribute:: description
+
+      Invalid URL description, :class:`str` instance.
+
+.. exception:: InvalidRedirectUrl
+
+   Redirect URL used for fetching is malformed, e.g. it does not contain host
```
webknjaz: It'd say it's the URL in the server's location header, not used for fetching.

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,34 @@ class InvalidURL(ClientError, ValueError):
...
+        else:
+            super().__init__(url)
+
+    @property
+    def url(self) -> StrOrURL:
+        return self.args[0]  # type: ignore[no-any-return]
```
webknjaz:
```suggestion
        return self.args[0]
```

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,34 @@ class InvalidURL(ClientError, ValueError):
...
+            return self.args[1]  # type: ignore[no-any-return]
+        except KeyError:
+            return None
 
     def __repr__(self) -> str:
-        return f"<{self.__class__.__name__} {self.url}>"
+        return f"<{self.__class__.__name__} {self}>"
+
+    def __str__(self) -> str:
+        return " - ".join(map(str, self.args))
```
webknjaz: I'd probably make this colon- or period-separated instead.

aiohttp/client_exceptions.py
```diff
@@ -238,17 +239,34 @@ class InvalidURL(ClientError, ValueError):
...
+
+
+class InvalidRedirectURL(InvalidURL):
+    """Invalid redirect url error"""
```
webknjaz:
```suggestion
    """Invalid redirect URL error."""
```

Don't forget to add a long description below this too.

tests/test_client_functional.py
```diff
@@ -2166,6 +2166,35 @@ async def handler_redirect(request):
...
+    "invalid_redirect_url",
+    (
+        # yarl.URL.__new__ raises ValueError
+        "http://:/",
+        "http://www.example.com/../filename.html",
+        "http://example.org:non_int_port/",
+        # yarl.URL.origin raises ValueError
+        "http:/",
+        "http:/example.com",
+        "http:///example.com",
```
webknjaz: Let's also test schemas used to invoke custom non-browser apps:
```suggestion
        "http:///example.com",
        "bluesky://d:i:d",
        "call:+3801234567",
        "skype:handle",
        "slack://instance/room",
        "steam:code",
        "twitter://handle",
```

aiohttp/client_exceptions.py
Dreamsorcerer: These don't produce the same result, due to the parent class:
`InvalidURL('url', None)` vs `InvalidURL('url')`

docs/client_reference.rst
Dreamsorcerer: It's a subclass, indentation is correct.

Dreamsorcerer:
> @Dreamsorcerer has this slipped off of your review queue?

Seems so. I probably still have a tab open to come back to it somewhere..
This looks like it's basically ready.

aiohttp/client_exceptions.py
Dreamsorcerer: If the ignore wasn't needed, it'd error in mypy. In this case args are from the base exception and are therefore `Any`. Maybe... we can just define the type in the subclass with `args: tuple[StrOrUrl, str]` or similar.

aiohttp/client_exceptions.py
setla: Can You post an example? I'm not sure what You mean

aiohttp/client_exceptions.py
setla: Can we resolve this conversation?

docs/client_reference.rst
setla: Can we resolve this conversation?

tests/test_client_functional.py
setla: added

docs/client_reference.rst
setla: updated

tests/test_client_exceptions.py
setla: added

aiohttp/client_exceptions.py
setla: Can we resolve this conversation?

setla:
@Dreamsorcerer @webknjaz please let me know if there is anything else to change and resolve conversations if changes cover it

aiohttp/client_exceptions.py
Dreamsorcerer: `" - "` -> `" : "`

I don't really mind regardless.

CHANGES/6722.feature
```diff
@@ -0,0 +1,8 @@
+Added an ``InvalidRedirectURL`` exception which is now raised instead of
```
webknjaz:
```suggestion
Added an :py:exc:`~aiohttp.InvalidRedirectURL` exception which is now raised instead of
```

CHANGES/6722.feature
```diff
@@ -0,0 +1,8 @@
...
+:py:exc:`ValueError` or ``InvalidURL`` when the redirect URL
+is invalid -- by :user:`setla`.
+
+The ``InvalidURL`` exception (and
+``InvalidRedirectURL``, that inherits it), now exposes
```
webknjaz:
```suggestion
:py:exc:`~aiohttp.InvalidRedirectURL`, that inherits it), now exposes
```

CHANGES/6722.feature
```diff
@@ -0,0 +1,8 @@
+Added an ``InvalidRedirectURL`` exception which is now raised instead of
+:py:exc:`ValueError` or ``InvalidURL`` when the redirect URL
+is invalid -- by :user:`setla`.
+
+The ``InvalidURL`` exception (and
```
webknjaz:
```suggestion
The :py:exc:`~aiohttp.InvalidURL` exception (and
```

CHANGES/6722.feature
```diff
@@ -0,0 +1,8 @@
+Added an ``InvalidRedirectURL`` exception which is now raised instead of
+:py:exc:`ValueError` or ``InvalidURL`` when the redirect URL
```
webknjaz:
```suggestion
:py:exc:`ValueError` or :py:exc:`~aiohttp.InvalidURL` when the redirect URL
```

CHANGES/6722.feature
webknjaz, line 1: Please, commit a symlink called `2507.feature.rst` additionally.

docs/client_reference.rst
webknjaz: Yes, marking it as resolved.

aiohttp/client_exceptions.py
webknjaz: Okay, marking it as resolved.

aiohttp/client_exceptions.py
webknjaz: I was thinking maybe
```suggestion
        return ". ".join(map(str, self.args))
```

aiohttp/client.py
```diff
@@ -611,34 +613,40 @@ async def _request(
...
                         except ValueError as e:
-                            raise InvalidURL(r_url) from e
+                            raise InvalidRedirectURL(r_url) from e
 
-                        scheme = parsed_url.scheme
+                        scheme = parsed_redirect_url.scheme
                         if scheme not in ("http", "https", ""):
                             resp.close()
-                            raise ValueError("Can redirect only to http or https")
+                            raise InvalidRedirectURL(
```
webknjaz: @setla So I dug up an old discussion — #2507 — and have an improvement request.

I think that this specific case requires a subclass of `InvalidRedirectURL` for specificity, or perhaps a direct subclass of `ClientError`. Call it a `NonHttpURL` with a `yarl.URL` instance always exposed.

I suppose, this check should also be reproduced after line 407/409 for consistency — this way, the clients would be able to rely on `NonHttpURL` for both the initial request and redirects the same way.

The cases above and below are equivalent between each other but differ from the case here.

Such a change would allow clients to break out of the HTTP land and invoke other apps through `webbrowser.open()`, for example. It's something I've been wanting to have for half a decade and seems natural to do in this PR. Moreover, since it adds new exceptions, it would be weird to change them later if implemented separately. And it could facilitate #2335 later on.

aiohttp/client_exceptions.py
webknjaz: Okay, fair. Marking as resolved.

aiohttp/client_exceptions.py
```diff
@@ -248,17 +249,34 @@ class InvalidURL(ClientError, ValueError):
...
+            super().__init__(url)
+
+    @property
+    def url(self) -> StrOrURL:
+        return self.args[0]  # type: ignore[no-any-return]
 
     @property
-    def url(self) -> Any:
-        return self.args[0]
+    def description(self) -> Optional[str]:
```
webknjaz: Let's avoid `Optional`:
```suggestion
    def description(self) -> "str | None":
```

aiohttp/client_exceptions.py
```diff
@@ -248,17 +249,34 @@ class InvalidURL(ClientError, ValueError):
...
+        try:
+            return self.args[1]  # type: ignore[no-any-return]
+        except KeyError:
+            return None
```
webknjaz: @setla could you add test coverage for this line?

docs/client_reference.rst
```diff
@@ -2099,6 +2099,18 @@ All exceptions are available as members of *aiohttp* module.
 
       Invalid URL, :class:`yarl.URL` instance.
 
+    .. attribute:: description
+
+      Invalid URL description, :class:`str` instance.
```
webknjaz:
```suggestion
      Invalid URL description, :class:`str` instance or :data:`None`.
```

docs/client_reference.rst
```diff
@@ -2099,6 +2099,18 @@ All exceptions are available as members of *aiohttp* module.
...
+
+.. exception:: InvalidRedirectURL
+
+   Redirect URL in the server location header is malformed,
+   e.g. it does not contain host part.
```
webknjaz:
```suggestion
   e.g. it does not contain host part or is not a URL at all.
```

tests/test_client_functional.py
```diff
@@ -2453,6 +2457,40 @@ async def handler_redirect(request):
...
+        "call:+3801234567",
+        "skype:handle",
+        "slack://instance/room",
+        "steam:code",
+        "twitter://handle",
+    ),
+)
+async def test_invalid_redirect_url(
+    aiohttp_client: Any, invalid_redirect_url: Any
+) -> None:
```
webknjaz: With the request @ https://github.com/aio-libs/aiohttp/pull/6722/files#r1472189734, please add test cases for these passed into `client.get()` directly that make sure that the exceptions bubbling up from redirect and direct ones raise `NonHttpURL` in the same manner.

I suppose, it's also useful to make these test cases reusable via a fixture:

```suggestion
@pytest.fixture(
    params=(
        # yarl.URL.__new__ raises ValueError
        "http://:/",
        "http://example.org:non_int_port/",
        # # yarl.URL.origin raises ValueError
        "http:/",
        "http:/example.com",
        "http:///example.com",
        "bluesky://d:i:d",
        "call:+3801234567",
        "skype:handle",
        "slack://instance/room",
        "steam:code",
        "twitter://handle",
    ),
)
def invalid_redirect_url(request: pytest.FixtureRequest) -> str:
    return request.param


async def test_invalid_redirect_url(
    aiohttp_client: Any, invalid_redirect_url: str,
) -> None:
```

aiohttp/client_exceptions.py
```diff
@@ -248,17 +249,34 @@ class InvalidURL(ClientError, ValueError):
 
     # Derive from ValueError for backward compatibility
 
-    def __init__(self, url: Any) -> None:
+    def __init__(self, url: StrOrURL, description: Union[str, None] = None) -> None:
         # The type of url is not yarl.URL because the exception can be raised
         # on URL(url) call
```
webknjaz, line 258: Let's just store the params in attributes and avoid jumping through the hoops of accessing `self.args` + error handling in properties. Those would just return straightforward values.

```suggestion
        # on URL(url) call
        self._url = url
        self._description = description

```

aiohttp/client.py
```diff
@@ -611,34 +613,40 @@ async def _request(
...
-                            and parsed_url.scheme == "https"
+                            url.host == parsed_redirect_url.host
+                            and parsed_redirect_url.scheme == "https"
                             and url.scheme == "http"
                         )
 
+                        try:
+                            redirect_origin = parsed_redirect_url.origin()
+                        except ValueError as e:
+                            raise InvalidRedirectURL(parsed_redirect_url) from e
```
webknjaz: Why not have a description so that this case is distinguishable for the one a few checks above?
```suggestion
                        except ValueError as origin_val_err:
                            raise InvalidRedirectURL(
                                parsed_redirect_url,
                                "Invalid redirect URL origin",
                            ) from origin_val_err
```

aiohttp/client.py
```diff
@@ -611,34 +613,40 @@ async def _request(
...
 
                         try:
-                            parsed_url = URL(
+                            parsed_redirect_url = URL(
                                 r_url, encoded=not self._requote_redirect_url
                             )
-
                         except ValueError as e:
-                            raise InvalidURL(r_url) from e
+                            raise InvalidRedirectURL(r_url) from e
```
webknjaz: Similarly, assign a description here:
```suggestion
                            raise InvalidRedirectURL(
                                r_url,
                                "Server attempted redirecting to a location that does not look like a URL",
                            ) from e
```

tests/test_client_functional.py
```diff
@@ -2453,6 +2457,40 @@ async def handler_redirect(request):
...
+    headers = {hdrs.LOCATION: invalid_redirect_url}
+
+    async def handler_request(request):
+        return web.Response(status=301, headers=headers)
+
+    app = web.Application()
+    app.router.add_get("/redirect", handler_request)
+    client = await aiohttp_client(app)
+
+    with pytest.raises(InvalidRedirectURL):
```
webknjaz: It is the best practice to use the `match=` regex together with `pytest.raises()`.

webknjaz:
@setla another round of reviews complete. I remembered some old discussions and made some structural requests based on them. Plus this would make the different cases even more easier to distinguish.

wumpus:
Thanks for keeping this issue alive. I stopped using aiohttp's redirect code because I had no idea what the correct behavior should be, thus I did not know what a patch should do.

aiohttp/client_exceptions.py
setla: `" : "` looks better than dot for me but `" - "` seems to be more general (in case of more args). I don't really mind too so just let me know which version you will accept so I will keep current one or change

webknjaz: With the suggestions to actually save the args into private attributes, I'd drop the loop here and use an f-string for clarity.

CHANGES/6722.feature
setla: added

aiohttp/client_exceptions.py
setla: added

aiohttp/client.py
setla: I probably wanted to keep exception the same, with just class updated but adding description does not break anything :+1:

aiohttp/client.py
setla: added

tests/test_client_functional.py
setla: sure but in some cases this error arg gets a bit modified arg e.g. http:/ is changed into http:/// - probably during some parsing/fix attempt

aiohttp/client.py
setla: I see some conflict here. If `NonHttpURL` was subclass of `InvalidRedirectURL` I couldn't use it in lines 407/409 - it's not redirect. If it wasn't subclass of `InvalidRedirectURL` - it won't be catched if someone would like repair redirect url on his own (if raised) or will catch it if was raised in both of mentioned code places (again only one of them is redirect). This might require 2 new exceptions - `NonHttpRedirectURL` (subclass of `InvalidRedirectURL`) and `NonHttpURL` (subclass of `InvalidURL`) but in this case it's not scope of this issue and could be done in separate PR/issue. What do You think?

aiohttp/client_exceptions.py
setla: updated

tests/test_client_functional.py
webknjaz: This can be addressed by making a partial match, just the beginning of the string, for example. Additionally, it's possible to have another input in parametrize for exact messages, which would be useful to reuse too.

aiohttp/client.py
webknjaz: @setla I think this should still happen in this PR so that there's no remaking of a feature with breaking changes later on.

To address your concern, I suggest using multiple inheritance as follows:
```mermaid

---
title: Exception hierarchy for invalid and non-HTTP URLs, direct or redirect
---

graph BT;

    base-client-error{{"
        ClientError
    "}};

    redirect-client-error["
        RedirectClientError
    "];

    invalid-url-client-error["
        InvalidUrlClientError
    "];

    non-http-url-client-error["
        NonHttpUrlClientError
    "];

    invalid-url-redirect-client-error[["
        InvalidUrlRedirectClientError
    "]];

    non-http-url-redirect-client-error[["
        NonHttpUrlRedirectClientError
    "]];

    subgraph "Base classes for the end-users to rely on";
        invalid-url-client-error;
        redirect-client-error;
        non-http-url-client-error;
    end;

    invalid-url-client-error --> base-client-error;
    redirect-client-error --> base-client-error;
    non-http-url-client-error --> base-client-error;

    non-http-url-redirect-client-error --> redirect-client-error;
    non-http-url-redirect-client-error --> non-http-url-client-error;

    invalid-url-redirect-client-error --> invalid-url-client-error;
    invalid-url-redirect-client-error --> redirect-client-error;

```

With that, the end-users would have enough interfaces to catch specific error classes granularly.

CHANGES/6722.feature
webknjaz: Also symlink `3315.feature.rst` as I realized it needs to show up in the changelog too.

aiohttp/client_exceptions.py
```diff
@@ -248,17 +249,36 @@ class InvalidURL(ClientError, ValueError):
...
+    def description(self) -> "str | None":
+        return self._description
 
     def __repr__(self) -> str:
-        return f"<{self.__class__.__name__} {self.url}>"
+        return f"<{self.__class__.__name__} {self}>"
+
+    def __str__(self) -> str:
+        if self._description:
+            return f"{self._url} - {self._description}"
```
webknjaz, line 280: This is much better, thanks! Having a reference to arbitrary `args` iterable in the previous version didn't make it obvious that the message started with a URL which is why I wanted a dot-based separator originally. Now, it's clear that it was not a good idea.

CHANGES/6722.feature
setla: added

aiohttp/client.py
setla: Isn't enough to have `NonHttpInvalidRedirectURL(InvalidRedirectURL, [NonHttpInvalidUrl])` and `NonHttpInvalidUrl(InvalidURL)`?  We also need to inherit `InvalidURL` for backward compatibility. Common class for redirect related errors can be added anytime (as one of parent class of redirect url errors) if needed but there is no usecase for it now

tests/test_client_functional.py
setla: Okay, I added partial match with additional arg

Dreamsorcerer (approved):
Once a decision is made on the inheritance (I personally don't mind either way), then I think this is ready to merge.

setla:
@Dreamsorcerer @webknjaz is it expected that windows tests fail? https://github.com/setla/aiohttp/actions/runs/7765565978/job/21180923429?pr=1 - it shouldn't be related to changes in this PR.

Dreamsorcerer:
> @Dreamsorcerer @webknjaz is it expected that windows tests fail? https://github.com/setla/aiohttp/actions/runs/7765565978/job/21180923429?pr=1 - it shouldn't be related to changes in this PR.

You only need to worry if the required statuses fail (in this case, `check`).

aiohttp/client.py
webknjaz: There should be no thing called `NonHttpInvalidUrl` — non-HTTP URLs are valid, aiohttp just can't process them because they need to be process through something external to HTTP. So this naming is a misrepresentation of what's actually happening.

webknjaz: As for having less exceptions in the hierarchy, I think having a full tree is well within the scope of this PR, in the context of #2507 at least.

tests/test_client_functional.py
webknjaz: Marking as resolved.

aiohttp/client.py
webknjaz: @setla here's an updated proposal:

```mermaid

---
title: Exception hierarchy for invalid and non-HTTP URLs, direct or redirect
---

graph BT;

    base-client-error{{"
        ClientError
    "}};

    base-value-error{{"
        ValueError
    "}};

    base-invalid-url-error{{"
        InvalidURL
    "}};

    redirect-client-error["
        RedirectClientError
    "];

    invalid-url-client-error["
        InvalidUrlClientError
    "];

    non-http-url-client-error["
        NonHttpUrlClientError
    "];

    invalid-url-redirect-client-error[["
        InvalidUrlRedirectClientError
    "]];

    non-http-url-redirect-client-error[["
        NonHttpUrlRedirectClientError
    "]];

    subgraph "Base classes for the end-users to rely on";
        invalid-url-client-error;
        redirect-client-error;
        non-http-url-client-error;
    end;

    base-invalid-url-error --> base-value-error;
    base-invalid-url-error --> base-client-error;

    invalid-url-client-error --> base-invalid-url-error;
    redirect-client-error --> base-client-error;
    non-http-url-client-error --> base-client-error;

    non-http-url-redirect-client-error --> redirect-client-error;
    non-http-url-redirect-client-error --> non-http-url-client-error;

    invalid-url-redirect-client-error --> invalid-url-client-error;
    invalid-url-redirect-client-error --> redirect-client-error;

```

tests/test_client_functional.py
```diff
@@ -2453,6 +2457,40 @@ async def handler_redirect(request):
...
+    headers = {hdrs.LOCATION: invalid_redirect_url}
+
+    async def handler_request(request):
+        return web.Response(status=301, headers=headers)
+
+    app = web.Application()
+    app.router.add_get("/redirect", handler_request)
+    client = await aiohttp_client(app)
+
+    with pytest.raises(InvalidRedirectURL, match=f"^{error_message_url} - [a-zA-Z]+"):
```
webknjaz: Regexps are traditionally raw-strings:
```suggestion
    with pytest.raises(InvalidRedirectURL, match=fr"^{error_message_url} - [a-zA-Z]+"):
```

aiohttp/client.py
setla: Okay, just one more change - `NonHttpUrlClientError` needs to inherit `InvalidURL` to have the same init args. It seems reasonable to inherit `InvalidURL` anyway because this exception is related to some invalid url.

tests/test_client_functional.py
setla: I will use exception class as parameter so it won't need extraction to fixture, I will resolve this conversation

setla:
@webknjaz please take a look, i really want to merge this PR :)

aiohttp/client.py
webknjaz: > Okay, just one more change - `NonHttpUrlClientError` needs to inherit `InvalidURL` to have the same init args. It seems reasonable to inherit `InvalidURL` anyway because this exception is related to some invalid url.

No, those URLs aren't necessarily invalid. Just non-HTTP. It would be wrong to imply such a thing.

tests/test_client_functional.py
```diff
@@ -2453,6 +2458,53 @@ async def handler_redirect(request):
     assert data == body
 
 
+@pytest.mark.parametrize(
+    "invalid_redirect_url, error_message_url, expected_exception_class",
```
webknjaz:
```suggestion
    ("invalid_redirect_url", "error_message_url", "expected_exception_class"),
```

tests/test_client_functional.py
webknjaz: > I will use exception class as parameter so it won't need extraction to fixture, I will resolve this conversation

Without a dedicated test, it's hard to judge whether those exceptions are caused by the initial redirect or now. Maybe, we need several endpoints, redirecting to one another before redirecting to an exceptional URL. Separately, from an endpoint that does such a redirect from the very first response. And separate from an error happening before even attempting an HTTP request.

tests/test_client_functional.py
```diff
@@ -2453,6 +2458,53 @@ async def handler_redirect(request):
...
+        (
+            "http://example.org:non_int_port/",
+            "http://example.org:non_int_port/",
+            InvalidUrlRedirectClientError,
+        ),
+        # # yarl.URL.origin raises ValueError
+        ("http:/", "http:///", InvalidUrlRedirectClientError),
+        ("http:/example.com", "http:///example.com", InvalidUrlRedirectClientError),
+        ("http:///example.com", "http:///example.com", InvalidUrlRedirectClientError),
+        ("bluesky://d:i:d", "bluesky://d:i:d", InvalidUrlRedirectClientError),
```
webknjaz: This URL is valid. It's just non-HTTP.

tests/test_client_functional.py
```diff
@@ -2453,6 +2458,53 @@ async def handler_redirect(request):
...
+    error_message_url: str,
+    expected_exception_class: Any,
+) -> None:
+    headers = {hdrs.LOCATION: invalid_redirect_url}
+
+    async def handler_request(request):
+        return web.Response(status=301, headers=headers)
+
+    app = web.Application()
+    app.router.add_get("/redirect", handler_request)
```
webknjaz: Let's follow the structure of other test here:
```suggestion
    async def generate_redirecting_response(request):
        raise web.HTTPFound(location=invalid_redirect_url)

    app = web.Application()
    app.router.add_get("/redirect", generate_redirecting_response)
```

tests/test_client_exceptions.py
webknjaz, line 1: This test module needs something to cover the new exception classes too.

docs/client_reference.rst
```diff
@@ -2315,3 +2350,17 @@ Hierarchy of exceptions
...
 
   * :exc:`InvalidURL`
+
+    * :exc:`InvalidUrlClientError`
+
+      * :exc:`InvalidUrlRedirectClientError`
+
+    * :exc:`NonHttpUrlClientError`
+
+      * :exc:`NonHttpUrlRedirectClientError`
```
webknjaz: Per https://github.com/aio-libs/aiohttp/pull/6722#discussion_r1479985833 I maintain that these shouldn't be advertised as invalid. It sends a wrong message.

@Dreamsorcerer care to share an opinion?

docs/client_reference.rst
```diff
@@ -2315,3 +2350,17 @@ Hierarchy of exceptions
...
+
+  * :exc:`RedirectClientError`
```
webknjaz, line 2362: So Sphinx supports mermaid, just like GitHub does. Can we paste the diagram from https://github.com/aio-libs/aiohttp/pull/6722#discussion_r1477103562 here? It might require enabling an extra Sphinx plugin in `docs/conf.py`.

webknjaz:
@setla I also really want to merge it. Though, I insist on the exception hierarchy to be corrected first. Non-HTTP URLs are not necessarily invalid, just different. The clients should use `ClientError` if they want to have a catch-all. I think @Dreamsorcerer mentioned that the users expect this anyway.

tests/test_client_functional.py
```diff
@@ -2453,6 +2458,53 @@ async def handler_redirect(request):
...
+        (
+            "slack://instance/room",
+            "slack://instance/room",
+            NonHttpUrlRedirectClientError,
+        ),
+        ("steam:code", "steam:code", NonHttpUrlRedirectClientError),
+        ("twitter://handle", "twitter://handle", NonHttpUrlRedirectClientError),
+    ),
+)
+async def test_invalid_redirect_url(
```
webknjaz, line 2527: Per https://github.com/aio-libs/aiohttp/pull/6722#discussion_r1480006621 I want to see these clearly visible test scenarios:
* [x] No HTTP handler present in the test, an exception is raised prior to making an HTTP request due to the initial input validation (the cases of a non-HTTP URL and an invalid URL)
* [x] An HTTP handler is present in the test and it immediately redirects somewhere (non-HTTP URL and invalid URL scenarios)
* [x] Several HTTP handlers are present in the test. The first one redirects to the second one through HTTP, without any problems, the second one redirects to the third one. And only the fourth handler redirects to something exceptional (non-HTTP URL and invalid URL scenarios)

setla:
> @setla I also really want to merge it. Though, I insist on the exception hierarchy to be corrected first. Non-HTTP URLs are not necessarily invalid, just different. The clients should use `ClientError` if they want to have a catch-all. I think @Dreamsorcerer mentioned that the users expect this anyway.

@webknjaz so should I replace `InvalidUrl` by `ClientError` as`NonHttpUrlClientError` parent class?

docs/client_reference.rst
Dreamsorcerer: Yeah, I don't mind too much. I think some users will expect it your way and other users another way (i.e. InvalidUrl meaning any URL that can't be processed by aiohttp). Renaming InvalidURL, but keeping it as the parent could meet both expectations, but would break backwards-compatibility.

webknjaz:
> so should I replace `InvalidUrl` by `ClientError` as`NonHttpUrlClientError` parent class?

Yes, make it follow my last graph, please. Sam said that some people may expect it to be one way and the others would expect it the other way, so I'm going to insist on my way since it's impossible to satisfy all parties in such cases. The important thing is that in both cases, it's a client error that remains a catchable base.

tests/test_client_functional.py
setla: @webknjaz I applied this but it's not response - it's error and some of test cases don't raise expected exceptions. I Restored it for now

tests/test_client_functional.py
setla: It's result of `yarl`'s `ValueError` here:

```python
try:
    parsed_redirect_url = URL(
        r_url, encoded=not self._requote_redirect_url
    )
except ValueError as e:
    raise InvalidUrlRedirectClientError(
        r_url,
        "Server attempted redirecting to a location that does not look like a URL",
    ) from e
```

```python-traceback
...
>               raise ValueError(message) from None
E               ValueError: Port could not be cast to integer value as 'i:d'
```

it seems to be invalid

tests/test_client_exceptions.py
setla: There is no code in new exceptions, just inheritance. What kind of tests do You expect? Is it some kind of duplication of `TestInvalidURL`?

docs/client_reference.rst
setla: I updated inheritance, it's not subclass of `ClientError`

docs/client_reference.rst
setla: You want me to paste this diagram just in this file or instead of some lines?

tests/test_client_functional.py
setla: I added new tests - please take a look and mark tasks as resolved if are ok.

CHANGES/6722.feature
```diff
@@ -0,0 +1,11 @@
...
+:py:exc:`~aiohttp.NonHttpUrlClientError`, :py:exc:`~aiohttp.InvalidUrlRedirectClientError`,
+:py:exc:`~aiohttp.NonHttpUrlRedirectClientError`
+
+:py:exc:`~aiohttp.InvalidUrlRedirectClientError`, :py:exc:`~aiohttp.NonHttpUrlRedirectClientError`
+are raised instead of :py:exc:`ValueError` or :py:exc:`~aiohttp.InvalidURL` when the redirect URL is invalid. Classes
+:py:exc:`~aiohttp.InvalidUrlClientError`, :py:exc:`~aiohttp.RedirectClientError`,
+:py:exc:`~aiohttp.NonHttpUrlClientError` are base for them.
+
+The :py:exc:`~aiohttp.InvalidURL` now exposes a ``description`` property with the text explanation of the error details.
+by :user:`setla`.
```
webknjaz: Since it's not a complete sentence, let's reformat this to match the surrounding style:
```suggestion

-- by :user:`setla`
```

The preview is visible @ https://aiohttp--6722.org.readthedocs.build/en/6722/changes.html#features, by the way.

aiohttp/client.py
```diff
@@ -406,6 +417,9 @@ async def _request(
         except ValueError as e:
             raise InvalidURL(str_or_url) from e
```
webknjaz: This lacks specificity slightly, `InvalidURL` is a base now:
```suggestion
            raise 
InvalidUrlClientError(str_or_url) from e
```

docs/client_reference.rst
webknjaz: Yep, just paste it below. Are you able to extract the mermaid source from that comment?

OTOH, we could skip this if it turns out too complicated, and circle back in a follow-up. Your call. I just thought that since I've made it, perhaps it shouldn't go to waste.

tests/test_client_exceptions.py
webknjaz: Yes, some duplication, taking into account the specifics of each. That's why I was suggesting a fixture parametrization improvement earlier.
I suppose I meant these scenarios should cover using all of the new specific exceptions: https://github.com/aio-libs/aiohttp/pull/6722#discussion_r1480045777.

tests/test_client_functional.py
```diff
@@ -2453,6 +2460,126 @@ async def handler_redirect(request):
...
 
+INVALID_URL_WITH_ERROR_MESSAGE = (
+    # yarl.URL.__new__ raises ValueError
+    ("http://:/", "http://:/"),
+    ("http://example.org:non_int_port/", "http://example.org:non_int_port/"),
+    # # yarl.URL.origin raises ValueError
+    ("http:/", "http:///"),
+    ("http:/example.com", "http:///example.com"),
+    ("http:///example.com", "http:///example.com"),
+    ("bluesky://d:i:d", "bluesky://d:i:d"),
```
webknjaz:
```suggestion
    ("bluesky://profile/d:i:d", "bluesky://profile/d:i:d"),
```

tests/test_client_functional.py
```diff
@@ -2453,6 +2460,126 @@ async def handler_redirect(request):
...
+)
+
+NON_HTTP_URL_WITH_ERROR_MESSAGE = (
+    ("call:+3801234567", r"call:\+3801234567"),
```
webknjaz: Let's use the realistic length here.
```suggestion
    ("call:+380123456789", r"call:\+380123456789"),
```

tests/test_client_functional.py
webknjaz: @setla you're right. I came up an incorrect URL example. I suggested changes here: https://github.com/aio-libs/aiohttp/pull/6722/files#r1485732854.

tests/test_client_exceptions.py
webknjaz: Scratch that. I confused it with that other test module. Since the new classes don't add extra behavior, this shouldn't be necessary. And the mentioned scenarios will be enough. Marking as resolved.

tests/test_client_functional.py
```diff
@@ -2453,6 +2460,126 @@ async def handler_redirect(request):
...
+    url: Any, error_message_url: Any, expected_exception_class: Any
+) -> None:
+    session = aiohttp.ClientSession()
+    try:
+        with pytest.raises(
+            expected_exception_class, match=rf"^{error_message_url}( - [A-Za-z ]+)?"
+        ):
+            await session.get(url)
+    finally:
+        await session.close()
```
webknjaz: Using an async CM should be simpler:
```suggestion
    async with aiohttp.ClientSession() as http_session:
        with pytest.raises(
            expected_exception_class, match=rf"^{error_message_url}( - [A-Za-z ]+)?"
        ):
            await http_session.get(url)
```

tests/test_client_functional.py
```diff
@@ -2453,6 +2460,126 @@ async def handler_redirect(request):
...
+    ("steam:code", "steam:code"),
+    ("twitter://handle", "twitter://handle"),
+)
+
+
+@pytest.mark.parametrize(
+    ("url", "error_message_url", "expected_exception_class"),
+    (
+        *(
+            (url, message, InvalidURL)
```
webknjaz:
```suggestion
            (url, message, InvalidUrlClientError)
```

tests/test_client_functional.py
webknjaz: Marking this as resolved. Pending the suggested fixes.

aiohttp/__init__.py
```diff
@@ -25,6 +25,7 @@
     ContentTypeError,
     Fingerprint,
     InvalidURL,
+    InvalidUrlRedirectClientError,
```
webknjaz, line 29: Why expose this one specifically but not the others?

aiohttp/client.py
```diff
@@ -404,7 +415,11 @@ async def _request(
         try:
             url = self._build_url(str_or_url)
         except ValueError as e:
-            raise InvalidURL(str_or_url) from e
+            raise
+InvalidUrlClientError(str_or_url) from e
```
webknjaz:
```suggestion
            raise InvalidUrlClientError(str_or_url) from e
```

tests/test_client_functional.py
```diff
@@ -21,7 +21,14 @@
 import aiohttp
 from aiohttp import Fingerprint, ServerFingerprintMismatch, hdrs, web
 from aiohttp.abc import AbstractResolver
-from aiohttp.client_exceptions import SocketTimeoutError, TooManyRedirects
+from aiohttp.client_exceptions import (
+    InvalidURL,
```
webknjaz:
```suggestion
    InvalidURL,
    InvalidUrlClientError,
```

webknjaz:
@setla I was hoping that applying small changes via GH UI would be enough to get this going but it seems like it needs some editing still. Plz pull the new commits from here before attempting to fix the tests.

aiohttp/__init__.py
setla: added

docs/client_reference.rst
setla: To be honest I don't know what You expect here - I can copy this graph
(

---
title: Exception hierarchy for invalid and non-HTTP URLs, direct or redirect
---

graph BT;

    base-client-error{{"
        ClientError
    "}};

    base-value-error{{"
        ValueError
    "}};

    base-invalid-url-error{{"
        InvalidURL
    "}};

    redirect-client-error["
        RedirectClientError
    "];

    invalid-url-client-error["
        InvalidUrlClientError
    "];

    non-http-url-client-error["
        NonHttpUrlClientError
    "];

    invalid-url-redirect-client-error[["
        InvalidUrlRedirectClientError
    "]];

    non-http-url-redirect-client-error[["
        NonHttpUrlRedirectClientError
    "]];

    subgraph "Base classes for the end-users to rely on";
        invalid-url-client-error;
        redirect-client-error;
        non-http-url-client-error;
    end;

    base-invalid-url-error --> base-value-error;
    base-invalid-url-error --> base-client-error;

    invalid-url-client-error --> base-invalid-url-error;
    redirect-client-error --> base-client-error;
    non-http-url-client-error --> base-client-error;

    non-http-url-redirect-client-error --> redirect-client-error;
    non-http-url-redirect-client-error --> non-http-url-client-error;

    invalid-url-redirect-client-error --> invalid-url-client-error;
    invalid-url-redirect-client-error --> redirect-client-error;

)
^ like this but pasting it in this doc does not look well :thinking:

aiohttp/__init__.py
setla: added all

docs/client_reference.rst
setla: Resolvig for now as it's last unresolved conversation - You can reopen it later or just commit changes directly

setla:
@webknjaz fixed, it seems like everything is addressed now, please let me know if there is anything else to do

docs/client_reference.rst
webknjaz: @setla did you enable the mermaid extension in Sphinx, though?

docs/client_reference.rst
```diff
@@ -2099,6 +2099,41 @@ All exceptions are available as members of *aiohttp* module.
...
+
+   Base class for all errors related to non http client urls.
+
+   Derived from :exc:`ClientError`
+
+.. exception:: InvalidUrlRedirectClientError
+
+   Redirect URL is malformed, e.g. it does not contain host part.
+
+   Derived from :exc:`InvalidUrlClientError` and `RedirectClientError`
```
webknjaz:
```suggestion
   Derived from :exc:`InvalidUrlClientError` and :exc:`RedirectClientError`
```

docs/client_reference.rst
```diff
@@ -2099,6 +2099,41 @@ All exceptions are available as members of *aiohttp* module.
...
+
+.. exception:: NonHttpUrlRedirectClientError
+
+   Redirect URL does not contain http schema.
+
+   Derived from :exc:`RedirectClientError` and `NonHttpUrlClientError`
```
webknjaz:
```suggestion
   Derived from :exc:`RedirectClientError` and :exc:`NonHttpUrlClientError`
```

docs/client_reference.rst
webknjaz: (single backticks would link to an arbitrary object with this name, not just an exception)

tests/test_client_functional.py
```diff
@@ -2453,6 +2461,132 @@ async def handler_redirect(request):
...
+
+@pytest.mark.parametrize(
+    ("url", "error_message_url", "expected_exception_class"),
+    (
+        *(
+            (url, message, InvalidUrlClientError)
+            for (url, message) in INVALID_URL_WITH_ERROR_MESSAGE_YARL_NEW
+        ),
+        *(
+            (url, message, InvalidURL)
```
webknjaz: We shouldn't be checking for the broad error classes. It seems like the actual specific exception is
```suggestion
            (url, message, InvalidUrlRedirectClientError)
```

tests/test_client_functional.py
```diff
@@ -21,7 +21,15 @@
 import aiohttp
 from aiohttp import Fingerprint, ServerFingerprintMismatch, hdrs, web
 from aiohttp.abc import AbstractResolver
-from aiohttp.client_exceptions import SocketTimeoutError, TooManyRedirects
+from aiohttp.client_exceptions import (
+    InvalidURL,
```
webknjaz:
```suggestion
```

webknjaz:
@setla to get this released, backport PRs should be submitted per instructions above.

setla:
@webknjaz right, should I create 2 PRs (3.9 and 3.10)? or just one (3.9)? 3.9 is already present https://github.com/aio-libs/aiohttp/pull/8154

webknjaz:
Two — the branches are separate. The backports are to made in reversed order. Besides, I know that 3.10 is definitely necessary, but not sure about 3.9 — I'll wait to see if anybody objects since bits of this might be perceived as breaking/new behaviors and not just a bugfix. Though I'll be able to merge into 3.10 right away.

Dreamsorcerer:
> I'll wait to see if anybody objects since bits of this might be perceived as breaking/new behaviors and not just a bugfix.

I'm not sure it's important enough to bother going into 3.9. I think we have enough improvements in 3.10 that we should be planning a release in a couple of months anyway.

webknjaz:
> I'm not sure it's important enough to bother going into 3.9. I think we have enough improvements in 3.10 that we should be planning a release in a couple of months anyway.

I want to make a 3.9 release as soon as https://github.com/aio-libs/aiohttp/pull/8089 is in. So if there's anything that could piggyback with that for free, I'd rather include it for as long as it's eligible. OTOH, the new exceptions are a new API, so per SemVer it should go into a minor release, not patch. I lost sight of it, but I just checked that the change notes are of the feature type, which makes more sense. I suppose, a 3.9 backport isn't needed indeed.

setla:
@webknjaz okay - I opened PR to 3.10 https://github.com/aio-libs/aiohttp/pull/8158

webknjaz:
@setla thank you! This improvement is what I've been wanting to get done for years and now it's mere months away.. Watch for new releases of the 3.10 series coming up this year!

setla:
thanks :)

bdraco:
It looks like this introduced a regression with websocket urls

```
2024-02-16 09:45:14.978 ERROR (MainThread) [pyunifiprotect.websocket] Websocket disconnect error: %s
  File "/usr/local/lib/python3.11/site-packages/pyunifiprotect/websocket.py", line 95, in _websocket_loop
    self._ws_connection = await session.ws_connect(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 884, in _ws_connect
    resp = await self.request(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 472, in _request
aiohttp.client_exceptions.NonHttpUrlClientError: wss://192.168.209.164/proxy/protect/ws/updates?lastUpdateId=80f54055-84de-4e15-95d7-80cac46aafbd
```

webknjaz:
@bdraco does this mean there's no test coverage for this case? Would you look into it? What's the expected behavior? What's supposed to handle this?

aiohttp/client.py
```diff
@@ -167,6 +177,7 @@ class ClientTimeout:
 
 # https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
+HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
```
webknjaz, line 180: @bdraco @setla would augmenting this make it work?

bdraco: I will analyze how its being used in the case I posted above when I get back home next week and provide additional detail

webknjaz: As a possible future improvement, I'd rename the constant to something like `SUPPORTED_URI_SCHEMAS`. Especially, if it'll stop doing listing http only.

setla: @webknjaz  @bdraco  Mentioned exception is related because it's raised here https://github.com/aio-libs/aiohttp/pull/6722/files#diff-f334e752b4894ef951105572ab8b195aeb8db90eb6e48b1dfbd9a01da4c854f5R420 but it's also used in redirects check https://github.com/aio-libs/aiohttp/pull/6722/files#diff-f334e752b4894ef951105572ab8b195aeb8db90eb6e48b1dfbd9a01da4c854f5R647. If we don't want regression, probably first mentioned place should be deleted

webknjaz: We could have two constants where one lists ws/wss and the check asserts against both in the first place but not the second one. Are websockets supposed to be disallowed in redirects?

Dreamsorcerer: > Are websockets supposed to be disallowed in redirects?

Seems that if we initiate a websocket request, then no redirects are allowed (of any protocol, including http):
> Let requestURL be a copy of url, with its [scheme](https://url.spec.whatwg.org/#concept-url-scheme) set to "http", if url’s [scheme](https://url.spec.whatwg.org/#concept-url-scheme) is "ws"; otherwise to "https".
> [...]
> The reason redirects are not followed and this handshake is generally restricted is because it could introduce serious security problems in a web browser context. For example, consider a host with a WebSocket server at one path and an open HTTP redirector at another. Suddenly, any script that can be given a particular WebSocket URL can be tricked into communicating to (and potentially sharing secrets with) any host on the internet, even if the script checks that the URL has the right hostname. 
https://websockets.spec.whatwg.org/#websocket-opening-handshake

Dreamsorcerer: Which means there should probably be an `allow_redirects=False` in here:
https://github.com/aio-libs/aiohttp/blob/dc38630b168a169139974617d75e176530c91696/aiohttp/client.py#L848-L860

webknjaz: @bdraco did you end up deciding how this should be fixed?

bdraco: I haven't had a free moment to get back to this. After the Home Assistant beta tomorrow and initial feedback I should be able to come back to it

bdraco: In this case the call that is failing is:

`await session.ws_connect('wss://.....')` 

Docs say
```
url – Websocket server url, [URL](https://yarl.aio-libs.org/en/stable/api/#yarl.URL) or [str](https://docs.python.org/3/library/stdtypes.html#str) that will be encoded with [URL](https://yarl.aio-libs.org/en/stable/api/#yarl.URL) (see [URL](https://yarl.aio-libs.org/en/stable/api/#yarl.URL) to skip encoding).
```

So we definitely have a regression.

looking how to fix it now

bdraco: Since  `ws_connect` always calls into `self.request` think the simplest solution to avoid a breaking change is:

```diff
diff --git a/aiohttp/client.py b/aiohttp/client.py
index 8d8d13f2..d608c009 100644
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -178,7 +178,7 @@ DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 
 # https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
-HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
+HTTP_SCHEMA_SET = frozenset({"http", "https", "ws", "wss", ""})
 
 _RetType = TypeVar("_RetType")
 _CharsetResolver = Callable[[ClientResponse, bytes], str]
```

Dreamsorcerer: I think we probably want that in the _request() check, but probably still want to limit it to http/https in the redirects, and disable redirects in _ws_connect() (as mentioned above).

bdraco: That makes sense to me.  So it seems we need to different constants

AraHaan, line 180: this should include wss:// as well as this breaks discord.py entirely.

AraHaan: this should include wss:// as well as this breaks discord.py entirely.

Dreamsorcerer: I probably lost track of this, maybe nobody got round to making the change discussed above.

PR #6726, opened 2022-04-29
Fix FileResponse fallback code

mdellweg:
In case the FileResponse is using _sendfile_fallback and the requested
range is smaller then the chunk size, we need to only read and send
count bytes.

## Checklist

### DIFF
--- a/CHANGES/6726.bugfix
+++ b/CHANGES/6726.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in ``FileResponse`` fallback code when the requested range is smaller than the chunk size.
--- a/tests/test_web_sendfile_functional.py
+++ b/tests/test_web_sendfile_functional.py
@@ -1179,9 +1179,15 @@ async def test_a_ranged_request_serves_no_additional_bytes(
     app.router.add_get("/", handler)
     client = await aiohttp_client(app)
 
-    resp = await client.get("/", headers={"Range": "bytes=20-40"})
-    assert resp.status == 206
-    body = await resp.read()
+    async with await client.get("/", headers={"Range": "bytes=20-40"}) as resp:
+        assert resp.status == 206
+        assert resp.headers["content-length"] == "21"
+        body = await resp.read()
+
+    async with await client.get("/", headers={"Range": "bytes=20-40"}) as resp:
+        assert resp.status == 206
+        assert resp.headers["content-length"] == "21"
+        body2 = await resp.read()
 
     # 21 bytes, because http range headers are including both fenceposts.
     assert len(body) == 21
@@ -1191,3 +1197,5 @@ async def test_a_ranged_request_serves_no_additional_bytes(
         f.seek(20)
         content = f.read(21)
         assert content == body
+
+    assert body == body2
--- a/tests/test_web_sendfile_functional.py
+++ b/tests/test_web_sendfile_functional.py
@@ -1164,3 +1164,30 @@ async def test_static_file_huge_error(
 
     resp.release()
     await client.close()
+
+
+@pytest.mark.parametrize("chunk_size", (15, 31))
+async def test_a_ranged_request_serves_no_additional_bytes(
+    aiohttp_client: AiohttpClient, sender: _Sender, chunk_size: int
+) -> None:
+    filepath = pathlib.Path(__file__).parent / "aiohttp.jpg"
+
+    async def handler(request: web.Request) -> web.FileResponse:
+        return sender(filepath, chunk_size=chunk_size)
+
+    app = web.Application()
+    app.router.add_get("/", handler)
+    client = await aiohttp_client(app)
+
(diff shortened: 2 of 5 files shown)


CHANGES/6726.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug in ``FileResponses`` fallback code when the requested range is smaller than the chunk size.
```
ipanova: s/FileResponses/FileResponse/

dzhalaevd:
Hi, @mdellweg, this issue is still relevant?

mdellweg:
I would guess so.

dzhalaevd:
@mdellweg, Since there has been no activity over the past year can you send an code example in which such a issue occurs. Also you need fill CONTRIBUTORS.txt.and change unit test.

Dreamsorcerer:
As mentioned, this still needs a regression test. It should go in test_sendfile_functional.py and use the `sender` fixture to show that the behaviour is the same with/without the fallback.

mdellweg:
I added a test.
But the weird thing is, I could not make the test fail. Even though I _think_ I found another place where this content length seems to be limited (weirdly so, after compression...) [0]. I even broke that (by commenting out line 126 [0]) and tracked the (now too large) chunk with a debugger to line 82 where it still wastoo large. But the test did not fail. Is it possible that the `client` used for testing here is more forgiving than it should be (for the purpose of testing anyway)?

How should I proceed?

[0] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_writer.py#L126
[1] https://github.com/aio-libs/aiohttp/blob/master/aiohttp/http_writer.py#L82

tests/test_web_sendfile_functional.py
```diff
@@ -1164,3 +1164,30 @@ async def handler(request: web.Request) -> web.FileResponse:
...
+    async def handler(request: web.Request) -> web.FileResponse:
+        return sender(filepath, chunk_size=chunk_size)
+
+    app = web.Application()
+    app.router.add_get("/", handler)
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/", headers={"Range": "bytes=20-40"})
+    assert resp.status == 206
+    body = await resp.read()
```
Dreamsorcerer: It may also be important to use the context manager correctly if we want to test 2 requests in a row.
```suggestion
    async with client.get("/", headers={"Range": "bytes=20-40"}) as resp:
        assert resp.status == 206
        body = await resp.read()
```

Dreamsorcerer:
> If there's actually more data sent than the header says, then reading the _next_ response should produce an error. So, you could adapt the test to try sending and receiving 2 range requests in a row.

I should note that this error may only happen with the test client. If you try it outside of a test, then the client may silently retry the second request and succeed.

Dreamsorcerer:
Actually, looking at the code you've edited. Isn't it already doing that in the next couple of lines:
https://github.com/aio-libs/aiohttp/blob/237d46722a0e56363c4a07dcf31e73fe4c2ea392/aiohttp/web_fileresponse.py#L121

I think that code cuts it to the correct length. So, your min() is probably not making any difference functionally. It might save reading a little bit of data when it doesn't need it, but wastes a function call when it does, so I'm not clear if it would be more efficient or not.

mdellweg:
> Actually, looking at the code you've edited. Isn't it already doing that in the next couple of lines:

That is the line right after reading and handing down the too big chunk.
I think I don't quite understand what you intend with the next response.
Also i could no longer reproduce the bug in the environment i saw it originally.
As you say, now it's "just" the extra cost of the min call against reading a few more bytes than needed from the IO object.
I'd be fine with just closing this, given that i believe no extra bytes are actually transfered over the wire.

Dreamsorcerer:
> That is the line right after reading and handing down the too big chunk.

Ah, yes, it writes the existing chunk first. So, actually still seems like your fix might be relevant. Did you try the test with 2 range requests in a row?

mdellweg:
I'll have a shot.

mdellweg:
Not sure if that's what you meant (looking at the test).
I also broke all the other places that I think prevent sending too many bytes, but still the test would not capture that. The client seems to comply to the content length more than to the tcp stream.

Dreamsorcerer:
> The client seems to comply to the content length more than to the tcp stream.

I think the client should definitely be hitting an error there if there's actually more data sent. Have you tried curl or any other clients and seen it sending more data?

bdraco:
All Python versions we support implement sendfile fallback internally now. So you would likely have to set the env var to reach the fallback code

bdraco:
Also we have been discussing dropping the aiohttp implementation now that asyncio does it for us

https://github.com/aio-libs/aiohttp/issues/10096

Dreamsorcerer:
I think from the last few comments, it sounds like this should just be closed now.

PR #6727, opened 2022-04-29
Fix asyncio.CancelledError again (#6719)

### LINKED ISSUE #6719: CancelledError
### Describe the bug

Using the aiohttp web server, I struggle with `asyncio.CancelledError()` when a client disconnect for some time.
I shielded what needed to be but it has its limits.
Today I found this old PR:
https://github.com/aio-libs/aiohttp/pull/4080

It looks like my issue.
If I understand correctly  `asyncio.CancelledError()` was fixed at the beginning of v3.7.0 (1th October 2019).
But I still have frequent `asyncio.CancelledError()` in my server.

I have investigated a bit more and I find another PR, at the end of v3.7.0 (just before v3.7.0b0 I think) that for me re-introduce the issue:
https://github.com/aio-libs/aiohttp/pull/4415

For me the wrong code is the following:
```py
        if self._task_handler is not None:
            self._task_handler.cancel()
```    

It was removed by https://github.com/aio-libs/aiohttp/pull/4080 (and with all the documentation regarding this issue) but re-introduced silently by https://github.com/aio-libs/aiohttp/pull/4415.

I'm clearly not an expert so I'm not sure of what I'm saying, but does it make sense for you?

### To Reproduce

NA

### Expected behavior

NA

### Logs/tracebacks

```python-traceback
NA
```

### Python Version

```console
3.10.4
```

### aiohttp Version

```console
3.8.1
```

### multidict Version

```console
5.1.0
```

### yarl Version

```console
1.6.3
```

### OS

Fedora 35

### Related component

Server

### Additional context

_No response_

### Code of Conduct

H--o-l:
`asyncio.CancelledError()` on peer disconnection have been removed by #4080,
but #4415 re-introduced it silently.

`self._waiter.cancel()` and `self._task_handler.cancel()` were added by #4415,
but #4415 in fact only needed `self._waiter.cancel()` (proof below).

So I propose to remove `self._task_handler.cancel()`, both #4080 and #4415 will
be fixed.

To test that I re-resolved #4080 I used:
```py
async def handle(request):
    try:
        await asyncio.sleep(5)
        await request.text()
    except BaseException as e:
        print(f'base exception {type(e).__name__}')
    return web.Response(text='toto')
```
```console
curl -X POST -H "Content-Type: text/plain" --data "bouh" localhost:8080
```
I kill the curl request before the 5 seconds of sleep.
Before this PR I have the following error right after killing the curl:
```console
$ python -Wall -X dev test.py
(Press CTRL+C to quit)
base exception CancelledError
```

After this commit I have the following error, but only after the 5 seconds
sleep:
```console
$ python -Wall -X dev test.py
(Press CTRL+C to quit)
base exception ConnectionResetError
```

To test that I didn't re-introduce #4415 I use a basic `handle` and 30
`curl localhost:8080`:

- Before this commit no issue
- If I remove `self._task_handler.cancel()` no issue
- If I remove both `self._task_handler.cancel()` and `self._waiter.cancel()`:

  ```console
  $ python -Wall -X dev test.py
  (Press CTRL+C to quit)
  task: <Task pending name='Task-4' coro=<RequestHandler.start() done, defined at ...
  ...
  ```

So it's OK to remove only `self._task_handler.cancel()`.

There is no documentation or tests to be updated because #4080 already did that
part.

However I guess that for a few people it might be seen as a breaking change,
I'm not sure.

Tested on master and on v3.8.1.

fixes #6719

### DIFF
--- a/CHANGES/6719.bugfix
+++ b/CHANGES/6719.bugfix
@@ -1,1 +1,1 @@
-Fix asyncio.CancelledError again
+Fix regression where ``asyncio.CancelledError`` occurs on client disconnection.
--- a/CHANGES/6719.bugfix
+++ b/CHANGES/6719.bugfix
@@ -1,0 +1,1 @@
+Fix asyncio.CancelledError again
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -331,8 +331,6 @@ class RequestHandler(BaseProtocol):
                 exc = ConnectionResetError("Connection lost")
             self._current_request._cancel(exc)
 
-        if self._task_handler is not None:
-            self._task_handler.cancel()
         if self._waiter is not None:
             self._waiter.cancel()
 


H--o-l:
Hey!

Here is an available reviewer for this PR?

PS: CI seems to fail but I don't understand what is failing (I'm not sure it's related to this PR).

H--o-l:
Hey @Dreamsorcerer!

Any chance you have a bit of time to look at this PR? It's the one for the issue https://github.com/aio-libs/aiohttp/issues/6719 we discussed together.

Have a great day!

Dreamsorcerer:
Excellent, thanks.

webknjaz:
@Dreamsorcerer I was under impression that handler cancellation was supposed to have stopped happening only with aiohttp 4+. Is this something else?

webknjaz:
OTOH this seems to have been intended for 3.7, after all: https://github.com/aio-libs/aiohttp/pull/4771.

Dreamsorcerer:
Yeah, it was in 3.7, then got accidentally disabled in 3.8, causing a regression.

PR #6740, opened 2022-05-07
Fix chunked eof

freshness79:
I found some web sites whose last chunk is "/r/n0/r/n/r/n". The parser matches the head "/r/n" but returns chunk[2:], which is "0/r/n/r/n", throwing at last a 400 error.
Given the fact we do know it's the last chunk, this can be easily solved by making sure we return "/r/n". I used last 2 bytes of data (which works) just to highlight badly configured services, but you could well use directly SEP if you think its better.
I tested it against some websites which were giving 400 error and with this fix the work fine.

None

## Related issue number

None

## Checklist

- [-] Unit tests for the changes exist
- [-] Documentation reflects the changes
- [-] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -823,7 +823,7 @@ class HttpPayloadParser:
                     if head == SEP:
                         # end of stream
                         self.payload.feed_eof()
-                        return True, chunk[2:]
+                        return True, chunk[-2:]
                     # Both CR and LF, or only LF may not be received yet. It is
                     # expected that CRLF or LF will be shown at the very first
                     # byte next time, otherwise trailers should come. The last


Dreamsorcerer:
I'm not too familiar with this, so I'll leave it open for @asvetlov or @webknjaz to look at, but it might be some time.

Does this occur both with and without extensions (`AIOHTTP_NO_EXTENSIONS` env variable)?
I think the file you edited is only used without extensions, _http_parser.pyx relates to the compiled extensions which should be the default if you pip installed the library.

freshness79:
@Dreamsorcerer
Honestly I didn't notice this specific module was available in both Python and compiled code.
I checked and I can confirm that the patch I'm submitting fixes only the pure python version of the library (AIOHTTP_NO_EXTENSIONS=1).
I'll try to spend some time on pyx version: should I be able to fix it as well, I'll update the pull request.
Then again, maybe my current fix can be a good tip for you to quickly catch the bug in Cython code.

webknjaz:
> I'll try to spend some time on pyx version: should I be able to fix it as well, I'll update the pull request.

I'd also ask to add tests covering this case and showing that both work the same way.

Dreamsorcerer:
There's been no movement for two years and I'm not convinced this is correct. We can reopen if a test is provided to reproduce the issue, though you may find it already fixed by one of the other parser fixes.

Issue #3862, opened 2019-06-21
Fix error with too long Header


serjshevchenko:
I got error "Got more than 8190 bytes (11194) when reading Header value is too long."
It would be nice to be able to pass additional args into RequestHandler (max_line_size, max_field_size)  via `web.run_app`

serjshevchenko:
Fixed in PR #3861

Dreamsorcerer:
Judging by the linked PR, interest was lost for this.

PR #6744, opened 2022-05-09
Client session new args max_line_size and max_field_size

starflows:
## What do these changes do?

Add new arguments `max_line_size` and `max_field_size` to ClientSession.

No, both arguments are optional and will use the same default values as previously when unset.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2304
https://github.com/aio-libs/aiohttp/issues/2988
https://github.com/aio-libs/aiohttp/issues/3862

## Checklist


Dreamsorcerer:
I'll leave this one for @asvetlov to evaluate when he comes back. It looks like he has some kind of plan for this: https://github.com/aio-libs/aiohttp/issues/2304#issuecomment-335099602

But, #2988 and #3862 appear to be unrelated. They both refer to setting limits in an `Application` in regards to incoming requests.

sumeshmurali:
Is this feature available to use?

fili:
It looks like this is in the most recent codebase:

https://github.com/aio-libs/aiohttp/blob/bc37e44f233ab8f5ed543ba842b1a4866cd77252/aiohttp/client.py#L222

added using:

https://github.com/aio-libs/aiohttp/commit/ecd9c72697ca1529e7cae4693775612987a2a6d2

However, trying to use this leads to the following error: 

```
TypeError: ClientSession.__init__() got an unexpected keyword argument 'max_line_size'
```

And it is also missing from the official documentation:
https://docs.aiohttp.org/en/stable/_modules/aiohttp/client.html#ClientSession
https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession

Dreamsorcerer:
Ah, didn't notice the duplicate. There are tests in that code, so if you can change one to reproduce the failure and fix the problem, then we can merge it.

Issue #6655, opened 2022-03-11
EmptyStreamReader instances raise an exception when printed


wmsmith:
### Describe the bug

I have code that uses `aiohttp` and is attempting to log the body of responses.  If the response is an instance of `EmptyStreamReader`, an exception is raised.  

This happens because `EmptyStreamReader` does not define `__str__`, and thus falls back to the `__repr__` it inherits from its parent (`StreamReader`), which reads attributes defined in the parent's `__init__` but are not defined for `EmptyStreamReader` since version 3.8.0 (commit https://github.com/aio-libs/aiohttp/pull/5270/commits/dac7994d9afce3dd43948a47162f9c7633ecf38a) when code was added to prevent executing the parent's `__init__`.

### To Reproduce

1.  Create an instance of `EmptyStreamReader`
2. Try to print the instance

Eg:
```
>>> from aiohttp.streams import EmptyStreamReader
>>> es = EmptyStreamReader()
>>> print(es)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/wayne/development/aiohttp/aiohttp/streams.py", line 130, in __repr__
    if self._size:
AttributeError: 'EmptyStreamReader' object has no attribute '_size'
```

### Expected behavior

The instance of `EmptyStreamReader` should not raise when printed.  

### Logs/tracebacks

```python-traceback
[ERROR] AttributeError: 'EmptyStreamReader' object has no attribute '_size'
Traceback (most recent call last):
  File "/var/task/main.py", line 115, in handler
    event_loop.run_until_complete(asyncio.gather(*tasks))
  File "/var/lang/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "/var/task/aiobotocore/client.py", line 89, in _make_api_call
    operation_model, request_dict, request_context)
  File "/var/task/aiobotocore/client.py", line 110, in _make_request
    request_dict)
  File "/var/task/aiobotocore/endpoint.py", line 73, in _send_request
    request, operation_model, context)
  File "/var/task/aiobotocore/endpoint.py", line 106, in _get_response
    request, operation_model)
  File "/var/task/aiobotocore/endpoint.py", line 164, in _do_get_response
    response_dict, operation_model.output_shape)
  File "/var/task/botocore/parsers.py", line 235, in parse
    LOG.debug('Response body:\n%s', response['body'])
[...]

PR #6748, opened 2022-05-12
Fixed __repr__ call in EmptyStreamReader

WindowGenerator:
## What do these changes do?

Fixed an error when calling the \_\_repr\_\_ method on an **EmptyStreamReader**

**Briefly about the solution:**

Removes the problem associated with calling the \_\_repr\_\_ method on the super class. 
The problem comes from missing methods in the child class. 
So the decision was made to override the lower class's \_\_repr\_\_ method

The user will no longer get an error when calling \_\_repr\_\_ on an **EmptyStreamReader**

## Related issue number

#6655

## Checklist

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -492,6 +492,9 @@ class EmptyStreamReader(StreamReader):  # lgtm [py/missing-call-to-init]
     def __init__(self) -> None:
         pass
 
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}>"
+
     def exception(self) -> Optional[BaseException]:
         return None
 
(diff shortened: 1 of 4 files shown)


aiohttp/streams.py
```diff
@@ -492,6 +492,9 @@ class EmptyStreamReader(StreamReader):  # lgtm [py/missing-call-to-init]
     def __init__(self) -> None:
         pass
 
+    def __repr__(self) -> str:
+        return "<%s>" % self.__class__.__name__
```
Dreamsorcerer: Should probably use .format()

WindowGenerator: Corrected. I'm sorry I didn't write properly

PR #6756, opened 2022-05-17
Fix pip reinstallation on windows due to constraints

greshilov:
## What do these changes do?

Fix for constantly failing CI on windows agents.

## Description

We have pinned pip version in constraints.txt file:
https://github.com/aio-libs/aiohttp/blob/08ca779e9abdb6d558cdd1647fdca17cfe979711/requirements/constraints.txt#L260

So when you call `pip install ... -c constraints.txt` it will downgrade/upgrade pip depending on the currently installed version. This works fine on all platforms except windows, which forces you to explicitly use `python -m pip install ...` when dealing with the pip package itself.

Since `python -m pip install ...` command is a completely legit replacement for `pip install ...` ([examples ](https://pip.pypa.io/en/latest/getting-started/)from the pip wiki), I decided to update `Makefile` with it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ endif
 	@python -m pip install --upgrade pip
 
 .install-cython: .update-pip $(call to-hash,requirements/cython.txt)
-	@pip install -r requirements/cython.txt -c requirements/constraints.txt
+	@python -m pip install -r requirements/cython.txt -c requirements/constraints.txt
 	@touch .install-cython
 
 aiohttp/_find_header.c: $(call to-hash,aiohttp/hdrs.py ./tools/gen.py)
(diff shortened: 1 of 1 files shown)


greshilov:
@Dreamsorcerer PTAL

Dreamsorcerer:
Weird. Fine by me.

Dreamsorcerer:
Hmm, Timeline protection seems to have broken. I guess it's a remote call, so maybe the server is down?

greshilov:
> Hmm, Timeline protection seems to have broken. I guess it's a remote call, so maybe the server is down?

@webknjaz should know, I suppose.

webknjaz:
Urgh... Must be the same problem as with Patchback.. https://github.com/sanitizers/patchback-github-app/issues/27

webknjaz:
Should be fixed now.

PR #6761, opened 2022-05-19
Feature: Add trace_request_ctx for WS connections

mario-sangar:
## What do these changes do?

Adding the possibility to pass `trace_request_ctx` when opening a Websocket connection. It's supported for all HTTP requests, but for some reason it was not an option for this method.

The argument is optional, so it shouldn't affect the current behavior of the method.

## Related issue number

None

## Checklist

### DIFF
--- a/CHANGES/6761.feature
+++ b/CHANGES/6761.feature
@@ -1,0 +1,1 @@
+Add `trace_request_ctx` argument for `ws_connect` function replicating the behavior of the other requests.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -211,6 +211,7 @@ Manuel Miranda
 Marat Sharafutdinov
 Marco Paolini
 Mariano Anaya
+Mario Sánchez García
 Mariusz Masztalerczuk
 Marko Kohtala
 Martijn Pieters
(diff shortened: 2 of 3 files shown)


cakiralpx:
I also need this feature. Because if there's an exception to the websocket request, there's no way to retrieve the response body. The only way is to send a mutable object with `trace_request_ctx` and retrieve the data from it.
Plus, since the trace configs that are passed with `ClientSession` work the same with `ws_connect` requests, `ws_connect` function not having `trace_request_ctx` param is clearly a missing implementation detail which needs a bug fix as soon as possible.

CHANGES/6761.feature
```diff
@@ -0,0 +1 @@
+Add `trace_request_ctx` argument for `ws_connect` function replicating the behavior of the other requests.
```
webknjaz, line 1: https://docs.aiohttp.org/en/stable/contributing.html#adding-change-notes-with-your-prs

webknjaz:
@bdraco I think you work with websockets regularly? Mind reviewing?

bdraco:
I'm traveling this week but I'll take a look at the end of the week or if I get some free time in between flights

bdraco (changes requested):
I think this will be fine, but it could use a test to exercise this path.

PR #6810, opened 2022-07-04
Fix readuntil work 

### LINKED ISSUE #6701: Error in StreamReader#readuntil
### Describe the bug

Hi there, it looks like StreamReader#readuntil function [here](https://github.com/aio-libs/aiohttp/blob/3.8/aiohttp/streams.py#L328) has an error.

My original use case was to read CSV files directly from ClientResponse line-by-line when the line terminator is `\r\n`.
The example I've provided is very similar.

When the separator has more than one symbol, it splits lines in a strange way.
For instance, when separator is `b'ab'`, it splits as follows:
```
b'11a'
b'b\n22a'
b'b\n33a'
b'b'
```
e.g. at the middle of the separator.

I think that this part wasn't updated at the time when it was `readline` function with only `\n` (one symbol length) separator available.

Probably this line should be:
```python
ichar = self._buffer[0].find(separator, offset) + len(separator) 
```

so that it will contain the whole separator instead of 'splitting' it.

### To Reproduce

Run this thing https://gist.github.com/vitaliihonta/b9be77deb0dfea1e9a411e695e6f15c9

### Expected behavior

Each line should contain the whole separator
```
b'11ab'
b'\n22ab'
b'\n33ab'
```

### Logs/tracebacks

```python-traceback
None
```

### Python Version

```console
Python 3.8.9
```

### aiohttp Version

```console
3.8.1
```

### multidict Version

```console
Name: multidict
Version: 6.0.2
```

### yarl Version

```console
Name: yarl
Version: 1.7.2
```

### OS

macOS

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Tingaev:
## What do these changes do?

Fix work `readuntil()` with separator length more one character

No

## Related issue number
Fixes #6701

## Checklist

### DIFF
--- a/tests/test_streams.py
+++ b/tests/test_streams.py
@@ -388,8 +388,8 @@ class TestStreamReader:
 
         def cb():
             stream.feed_data(b"chunk2 ")
-            stream.feed_data(b"chunk3 " + separator)
-            stream.feed_data(b" chunk4")
+            stream.feed_data(b"chunk3 ")
+            stream.feed_data(separator + b" chunk4")
 
         loop.call_soon(cb)
 
@@ -400,22 +400,24 @@ class TestStreamReader:
         data = await stream.read()
         assert b" chunk4" == data
 
-    async def test_readuntil_limit_with_existing_data(self) -> None:
+    @pytest.mark.parametrize("separator", [b"*", b"**"])
+    async def test_readuntil_limit_with_existing_data(self, separator: bytes) -> None:
         # Read one chunk. The data is in StreamReader's buffer
         # before the event loop is run.
 
         stream = self._make_one(limit=2)
         stream.feed_data(b"li")
-        stream.feed_data(b"ne1&line2&")
+        stream.feed_data(b"ne1" + separator + b"line2" + separator)
 
         with pytest.raises(ValueError):
-            await stream.readuntil(b"&")
+            await stream.readuntil(separator)
         # The buffer should contain the remaining data after exception
         stream.feed_eof()
         data = await stream.read()
-        assert b"line2&" == data
+        assert b"line2" + separator == data
 
-    async def test_readuntil_limit(self) -> None:
+    @pytest.mark.parametrize("separator", [b"*", b"**"])
+    async def test_readuntil_limit(self, separator: bytes) -> None:
         loop = asyncio.get_event_loop()
         # Read one chunk. StreamReaders are fed with data after
         # their 'readuntil' methods are called.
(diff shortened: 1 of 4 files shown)


tests/test_streams.py
```diff
@@ -377,23 +377,24 @@ async def test_readline_exception(self) -> None:
...
         stream.feed_data(b"chunk1 ")
-        read_task = loop.create_task(stream.readuntil(b"*"))
+        read_task = loop.create_task(stream.readuntil(separator))
 
         def cb():
             stream.feed_data(b"chunk2 ")
-            stream.feed_data(b"chunk3 ")
-            stream.feed_data(b"* chunk4")
+            stream.feed_data(b"chunk3 " + separator)
+            stream.feed_data(b" chunk4")
```
Dreamsorcerer: This one seems to have changed. Deliberate, or should we prefix with `separator`?

Tingaev: I'll do it like before
What do you think about the other tests?
Should I use pytest's parametrization decorator in other tests?

Dreamsorcerer: Would probably be nice, just to have that extra coverage, but it's up to you.

tests/test_streams.py
```diff
@@ -377,116 +377,127 @@ async def test_readline_exception(self) -> None:
...
         line = await read_task
-        assert b"chunk1 chunk2 chunk3 *" == line
+        assert b"chunk1 chunk2 chunk3 " + separator == line
 
         stream.feed_eof()
         data = await stream.read()
         assert b" chunk4" == data
 
-    async def test_readuntil_limit_with_existing_data(self) -> None:
+    @pytest.mark.parametrize("separator", [b"*", b"**"])
```
Dreamsorcerer: Thanks, looks good, but I'd reuse the separators used originally in each test (e.g. "&" and "&&" in this test).

tests/test_streams.py
```diff
@@ -377,116 +377,127 @@ async def test_readline_exception(self) -> None:
...
+    @pytest.mark.parametrize("separator", [b"&", b"&&"])
+    async def test_readuntil_limit(self, separator: bytes) -> None:
         loop = asyncio.get_event_loop()
         # Read one chunk. StreamReaders are fed with data after
         # their 'readuntil' methods are called.
         stream = self._make_one(limit=4)
 
         def cb():
             stream.feed_data(b"chunk1")
-            stream.feed_data(b"chunk2$")
```
Dreamsorcerer, line 428: I meant, look at each test. :P
This one was `$`. Every test appears to use different characters, which may help get a more thorough testing.

Tingaev:
@Dreamsorcerer ping

PR #6826, opened 2022-07-14
Allow filename content disposition to have space in accordance with RFC6266 Appendix D

ben-cutler-datarobot:
## What do these changes do?

According to RFC 6266 Appendix D:
https://datatracker.ietf.org/doc/html/rfc6266#appendix-D

```
   o  Avoid including the percent character followed by two hexadecimal
      characters (e.g., %A9) in the filename parameter, since some
      existing implementations consider it to be an escape character,
      while others will pass it through unchanged.
```
The current implementation of aiohttp does encode spaces of the filename, and this implementation should be avoided,.

Yes, filenames will no longer be encoded.

This is not exactly a new change for users on aiohttp 3.7 though. The filename was not encoded until this Pr https://github.com/aio-libs/aiohttp/issues/4012 encoded it. 
## Checklist

- [✅] I think the code is well written
- [✅] Unit tests for the changes exist
- [⚠️] Documentation reflects the changes
- [✅] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
- [✅] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/CHANGES/6826.bugfix
+++ b/CHANGES/6826.bugfix
@@ -1,0 +1,1 @@
+Don't encode ' ' as '%20' in a content-disposition's filename
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -52,6 +52,7 @@ Artem Yushkovskiy
 Arthur Darcet
 Austin Scola
 Ben Bader
+Ben Cutler
 Ben Greiner
 Ben Timby
 Benedikt Reinartz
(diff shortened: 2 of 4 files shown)


madduck:
> I realized there's a better solution. I realized the FormData has a constructor param `(quote_fields=False)` that I should be using instead

Doesn't this potentially have side-effects we don't want? After all, fields *should* be quoted in general, no? Just the `filename` should not be quoted…

PR #6829, opened 2022-07-17
Allow linking to request()

Dreamsorcerer:
noindex actually removes the anchor, so there is no way to create a link to this method in the documentation.


webknjaz:
@Dreamsorcerer you may want to also drop https://github.com/aio-libs/aiohttp/blob/c34eccb/docs/conf.py#L393 then.

Issue #6839, opened 2022-07-22
`web.run_app(path=)` should accept `pathlib.Path`


Dreamsorcerer:
The `path` parameter to `web.run_app()`: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.run_app

This should accept a `Path` object, to make things simpler.

SamirPS:
Hello, when this function is define in the source code ? @Dreamsorcerer

webknjaz:
> This should accept a `Path` object, to make things simpler.

How about `os.PathLike`?

_(https://github.com/aio-libs/aiohttp/pull/6843/files#r940309064)_

Dreamsorcerer:
#6843 Done.

PR #6842, opened 2022-07-23
Add path web app run

SamirPS:
## What do these changes do?

I modify the web.py file and add Union[str,pathlib.Path] for path

For the option of path, we can have str or pathlib.path

## Related issue number

#6839 

## Checklist


Dreamsorcerer:
Need to see a test covering this. Could probably just remove `str()` from these 2:
https://github.com/aio-libs/aiohttp/blob/master/tests/test_run_app.py#L521
https://github.com/aio-libs/aiohttp/blob/master/tests/test_run_app.py#L535

SamirPS:
> Need to see a test covering this. Could probably just remove `str()` from these 2: https://github.com/aio-libs/aiohttp/blob/master/tests/test_run_app.py#L521 https://github.com/aio-libs/aiohttp/blob/master/tests/test_run_app.py#L535

Okay i remove str() for this two test and all test in this file passed

aiohttp/web.py
```diff
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import pathlib
```
Dreamsorcerer: Minor nitpick. In most places we do `from pathlib import Path`.

Dreamsorcerer:
Also, test failing. Looks like there's another place that needs a `Union[]` annotation.

SamirPS:
@Dreamsorcerer For me all test passed or skipped because i working on ubuntu (WSL)

aiohttp/client_exceptions.py
```diff
@@ -175,13 +176,16 @@ class UnixClientConnectorError(ClientConnectorError):
...
+        path: Optional[Union[str, Path]],
+        connection_key: ConnectionKey,
+        os_error: OSError,
     ) -> None:
         self._path = path
         super().__init__(connection_key, os_error)
 
     @property
-    def path(self) -> str:
+    def path(self) -> Optional[Union[str, Path]]:
```
Dreamsorcerer, line 188: I don't understand why you've just added `Optional` everywhere. You've now gone from 1 error to 32 errors.

Dreamsorcerer:
We run a lot of tests in CI, so check the output. You should be able to reproduce the mypy errors on your machine, by simply running `mypy`.

aiohttp/client_exceptions.py
Dreamsorcerer: If you revert these changes, back to where you were before. There was only 1 error, it should be easy to then add the right annotation to pass the tests, you were almost there.

SamirPS: i create a new pull request

PR #6843, opened 2022-07-23
Add type Path on web app run

SamirPS:
## What do these changes do?

I modify the web.py file and add Union[str,pathlib.Path] for path

For the option of path, we can have str or pathlib.path

## Related issue number

#6839 

## Checklist

### DIFF
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+from pathlib import Path
 import socket
 import sys
 from argparse import ArgumentParser
@@ -290,7 +291,7 @@ async def _run_app(
     *,
     host: Optional[Union[str, HostSequence]] = None,
     port: Optional[int] = None,
-    path: Optional[str] = None,
+    path: Optional[Union[str, Path]] = None,
     sock: Optional[Union[socket.socket, TypingIterable[socket.socket]]] = None,
     shutdown_timeout: float = 60.0,
     keepalive_timeout: float = 75.0,
@@ -366,7 +367,7 @@ async def _run_app(
             )
 
         if path is not None:
-            if isinstance(path, (str, bytes, bytearray, memoryview)):
+            if isinstance(path, (str, bytes, bytearray, memoryview, Path)):
                 sites.append(
                     UnixSite(
                         runner,
@@ -464,7 +465,7 @@ def run_app(
     debug: bool = False,
     host: Optional[Union[str, HostSequence]] = None,
     port: Optional[int] = None,
-    path: Optional[str] = None,
+    path: Optional[Union[str, Path]] = None,
     sock: Optional[Union[socket.socket, TypingIterable[socket.socket]]] = None,
     shutdown_timeout: float = 60.0,
     keepalive_timeout: float = 75.0,
--- a/aiohttp/web.py
+++ b/aiohttp/web.py
@@ -1,11 +1,11 @@
 import asyncio
 import logging
-from pathlib import Path
 import socket
 import sys
 from argparse import ArgumentParser
 from collections.abc import Iterable
 from importlib import import_module
+from pathlib import Path
 from typing import (
     Any,
     Awaitable,
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -167,7 +167,8 @@ class UnixSite(BaseSite):
         server = self._runner.server
         assert server is not None
         self._server = await loop.create_unix_server(
-            server, self._path, ssl=self._ssl_context, backlog=self._backlog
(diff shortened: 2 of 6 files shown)


SamirPS:
@Dreamsorcerer Here the new

aiohttp/web_runner.py
```diff
@@ -166,15 +167,19 @@ async def start(self) -> None:
         server = self._runner.server
         assert server is not None
         self._server = await loop.create_unix_server(
-            server, self._path, ssl=self._ssl_context, backlog=self._backlog
+            server, str(self._path), ssl=self._ssl_context, backlog=self._backlog
```
Dreamsorcerer: Test is failing currently, looks like because of this `str()` call. It shouldn't be necessary, the docs say it accepts a Path object since Python 3.7.

SamirPS: @Dreamsorcerer if i remove the str call here, i have this error with mypy aiohttp/web_runner.py : 

aiohttp/web_runner.py:170: error: Argument 2 to "create_unix_server" of "AbstractEventLoop" has incompatible type "Union[str, Path]"; expected "Optional[str]"  [arg-type]

using version Python 3.8.10

but i try the commit

Dreamsorcerer: Fixed: https://github.com/python/typeshed/pull/8376

Just add a `# type: ignore[arg-type]` for now.

SamirPS: where i add this in my code ?

SamirPS:
The previous error was due to a python library ?

Dreamsorcerer:
> The previous error was due to a python library ?

Mypy gets its type information from typeshed, which is just a collection of annotations, because cpython doesn't have any annotations itself.

Dreamsorcerer:
So, that error will disappear when there's a new mypy release which will include the updated typeshed.

SamirPS:
Oh okay thanks for the information 

now we have this error : aiohttp/web_runner.py:199: error: Incompatible return value type (got "Union[str, Path]", expected "str")  [return-value]

I need to change str to Union[str, Path] for the name(self) function ?

Dreamsorcerer:
You had partial changes to NamedPipeSite, I was just playing with them. I've reverted the change, as it's not clear to me whether someone would actually want to use a `Path` object, and the function it uses is undocumented.

SamirPS:
Oh okay, use Path permit to avoid doing some convertion between Path and str i thinks

SamirPS:
For the doc web_reference.rst how can i says str or Path  :param str path: ?

Dreamsorcerer:
> For the doc web_reference.rst how can i says str or Path :param str path: ?

Ah, good point. Take a look at add_static(), it mentions Path there: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.UrlDispatcher.add_static

SamirPS:
> > For the doc web_reference.rst how can i says str or Path :param str path: ?
> 
> Ah, good point. Take a look at add_static(), it mentions Path there: https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.UrlDispatcher.add_static

I just commit the change, i hope all test will pass

SamirPS:
The test which fail it's because of a gcc error

Dreamsorcerer:
If you could follow the steps above to get this in to 3.9, that'd be great.

SamirPS:
i will do it tomorrow

SamirPS:
I just do it @Dreamsorcerer and now i wait

aiohttp/web.py
```diff
@@ -290,7 +291,7 @@ async def _run_app(
     *,
     host: Optional[Union[str, HostSequence]] = None,
     port: Optional[int] = None,
-    path: Optional[str] = None,
+    path: Optional[Union[str, Path]] = None,
```
webknjaz, line 294: @Dreamsorcerer I wonder if this should've been `os.PathLike` instead...

Dreamsorcerer: Hmm, seems we even have a PathLike in typedefs.py.

SamirPS: i didn't see it, you want me to use it or keep path from Pathlib ?

Dreamsorcerer: I'm already on it. Updating a couple of other places too.
#6876

SamirPS: You update this pull request ? https://github.com/aio-libs/aiohttp/pull/6874

webknjaz: > You update this pull request ? #6874

No, backports must be as close to the original patch as possible, for traceability.

webknjaz:
3.9 backport PR: https://github.com/aio-libs/aiohttp/pull/6957.

PR #6872, opened 2022-08-06
ignore the ResourceWarnings before they become unraisable exceptions

graingert:
## What do these changes do?

> @webknjaz:matrix.org btw you're using PytestUnraisableExceptionWarning incorrectly - you need to ignore the ResourceWarning not the PytestUnraisableExceptionWarning. Because pytest will only collect one unraisable per test and if you allow a ResourceWarning to raise then you defer the cleanup of the underlying resource eg:
> ```
> _warn("...", ResourceWarning)  # this raises 
> self._sock.close()  # this never gets called
> ```

nope

## Related issue number

#6872

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -132,11 +132,11 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:Exception ignored in. <function _SSLProtocolTransport.__del__ at.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object BaseConnector.close at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object ClientSession._request at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <function ClientSession.__del__ at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <_io.FileIO .closed.>:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
+    ignore:unclosed transport <_SSLProtocolTransport.*:ResourceWarning
+    ignore:coroutine 'BaseConnector.close' was never awaited:RuntimeWarning
+    ignore:coroutine 'ClientSession._request' was never awaited:RuntimeWarning
+    ignore:Unclosed client session <aiohttp.client.ClientSession object at 0x:ResourceWarning
+    ignore:unclosed file <_io.FileIO.*:ResourceWarning
     ignore:The loop argument is deprecated:DeprecationWarning:asyncio
     ignore:Creating a LegacyVersion has been deprecated and will be removed in the next major release:DeprecationWarning::
     # The following deprecation warning is triggered by importing
--- a/setup.cfg
+++ b/setup.cfg
@@ -132,7 +132,7 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:unclosed transport <_SSLProtocolTransport.*:ResourceWarning
+    ignore:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x.*:ResourceWarning
     ignore:coroutine 'BaseConnector.close' was never awaited:RuntimeWarning
     ignore:coroutine 'ClientSession._request' was never awaited:RuntimeWarning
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -132,11 +132,11 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:Exception ignored in. <function _SSLProtocolTransport.__del__ at.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object BaseConnector.close at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object ClientSession._request at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <function ClientSession.__del__ at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <_io.FileIO .closed.>:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
+    ignore:unclosed transport <_SSLProtocolTransport.*:ResourceWarning
```
graingert:
```suggestion
    ignore:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x.*:ResourceWarning
```

tests/test_run_app.py
```diff
@@ -21,16 +21,15 @@
...
-    try:
-        _abstract_path_sock.bind(b"\x00" + uuid4().hex.encode("ascii"))
-    except FileNotFoundError:
-        _abstract_path_failed = True
-    else:
-        _abstract_path_failed = False
-    finally:
-        _abstract_path_sock.close()
-        del _abstract_path_sock
+    with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as _abstract_path_sock:
```
webknjaz, line 24: Interesting... I'm pretty sure that `finally:` took care for this.

But yeah, at least it looks nicer.

tests/test_run_app.py
```diff
@@ -48,7 +47,8 @@
     # support, but the target system still may not have it.
     # So let's ensure that we really have IPv6 support.
     try:
-        socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+        with socket.socket(socket.AF_INET6, socket.SOCK_STREAM):
```
webknjaz, line 50: Nice catch!

tests/test_tcp_helpers.py
```diff
@@ -11,7 +11,8 @@
     # support, but the target system still may not have it.
     # So let's ensure that we really have IPv6 support.
     try:
-        socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
+        with socket.socket(socket.AF_INET6, socket.SOCK_STREAM):
```
webknjaz, line 14: :eyes:

webknjaz (approved):

setup.cfg
```diff
@@ -132,11 +132,11 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:Exception ignored in. <function _SSLProtocolTransport.__del__ at.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object BaseConnector.close at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <coroutine object ClientSession._request at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <function ClientSession.__del__ at 0x.:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
-    ignore:Exception ignored in. <_io.FileIO .closed.>:pytest.PytestUnraisableExceptionWarning:_pytest.unraisableexception
+    ignore:unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x.*:ResourceWarning
```
webknjaz, line 135: @graingert any suggestions on how to make this scoped to a single module?

webknjaz:
Backported as e4266b1eaa09fefa0f4d487dc8f06128584c0537 + 8201922ddddea3cb958971d8511b27f61a8bbfd4.

setup.cfg
graingert: You'll have to talk to @vstinner on how to get per module resource warnings

You can dig out the original owner of the object via tracemalloc but it slows down python considerably, The issue being tracemalloc is great and gives you all the info you need, but it's too slow to run all the time, also resources warnings are flakey which means by the time you see one it's too late and you weren't running tracemalloc

What I think should happen is that:

* When any object with a custom `__del__` is being constructed
* If the ResourceWarning filter is configured as "error"
* the frame module, filename and line number 2 frames above the last `__init__` call is recorded, as a sort of tracemalloc of last-resort

Then when the ResourceWarning is raised it's associated with those lines

webknjaz:
Also backported to 3.8 as https://github.com/aio-libs/aiohttp/commit/4d10cc9dc9d55bbde392164eda46a1f6683b8e64

PR #6876, opened 2022-08-08
Use PathLike

Dreamsorcerer:


aiohttp/web.py
```diff
@@ -367,7 +369,7 @@ async def _run_app(
             )
 
         if path is not None:
-            if isinstance(path, (str, bytes, bytearray, memoryview, Path)):
+            if isinstance(path, (str, bytes, bytearray, memoryview, os.PathLike)):
```
Dreamsorcerer: This conditional doesn't actually make sense, given the annotations. It should be `None | str | Path`. But, the `else` tries to iterate through it.

Dreamsorcerer: Documentation only mentions `str`, so maybe we can drop this?
https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.run_app

Dreamsorcerer: Looks like, long time ago, these were iterables:
https://github.com/aio-libs/aiohttp/blame/52dc7f0400984a8f5ee1a3b8ad2d6220fc4ab1a4/aiohttp/web.py#L330-L336

But, it doesn't look like that's expected today.

Dreamsorcerer: It's explicitly tested, and I missed that it is mentioned in the documentation.

PR #6878, opened 2022-08-10
Fix readexactly eof

SafaSafari:
## What do these changes do?
fix readexactly last chunk with eof and make the user decide to select between readexactly and read

maybe yes

## Checklist

### DIFF
--- a/CHANGES/6878.bugfix
+++ b/CHANGES/6878.bugfix
@@ -1,0 +1,1 @@
+fix ``readexaclty`` last chunk eof and make the user decide between ``read`` and ``readexactly`` in ``iter_chunked``
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -267,6 +267,7 @@ Robert Lu
 Robert Nikolich
 Roman Markeloff
 Roman Podoliaka
+Safa Safari
 Samir Akarioh
 Samuel Colvin
 Sean Hunt
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -64,8 +64,19 @@ class AsyncStreamReaderMixin:
     def __aiter__(self) -> AsyncStreamIterator[bytes]:
         return AsyncStreamIterator(self.readline)  # type: ignore[attr-defined]
 
-    def iter_chunked(self, n: int) -> AsyncStreamIterator[bytes]:
-        """Returns an asynchronous iterator that yields chunks of size n."""
+    def iter_chunked(self, n: int, exactly: bool = True) -> AsyncStreamIterator[bytes]:
+        """Returns an asynchronous iterator that yields chunks of size n
+
+        exactly n by default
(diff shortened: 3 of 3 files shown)


webknjaz:
Hi, such a change would require a test and a more detailed explanation of the effect of applying it. Saying that it fixes something and not explaining what was broken and how the behavior is affected, and what's the difference, makes it hard for us to judge how dangerous this patch is and if it's even necessary.
Could you please fix this?

SafaSafari:
> Hi, such a change would require a test and a more detailed explanation of the effect of applying it. Saying that it fixes something and not explaining what was broken and how the behavior is affected, and what's the difference, makes it hard for us to judge how dangerous this patch is and if it's even necessary.
> Could you please fix this?

yeah
let me explain the problem in this video

output

SafaSafari:
> Hi, such a change would require a test and a more detailed explanation of the effect of applying it. Saying that it fixes something and not explaining what was broken and how the behavior is affected, and what's the difference, makes it hard for us to judge how dangerous this patch is and if it's even necessary. Could you please fix this?

yeah, you're right
let me explain in a video

output

SafaSafari:
it's for the main problem about readexactly function
I also modified iter_chunked function to make the user decide between read and readexaclty
because when we use iter_chunked function, we expect every chunk size exactly equal to input int variable

Dreamsorcerer:
> yeah, you're right let me explain in a video

You just show the change, not the reason. Most often, a chunk size is to set an upper limit on how much data is read into memory, so the application doesn't use too much. Why do we want to stop the application processing data if it is available? Your change merely delays any available data from reaching the application. Why do we want to return something at EOF instead of producing an error? Your change results in `readexactly()` returning data which is _not_ exactly the requested length.

In other words, what are the use cases that justify this change?

SafaSafari:
So let me explain the problem from the beginning
I need to divide a large file into equal chunks using iter_chunked
But it depends on network speed and yielding chunks lower than chunk_size
I opened 2 pull requests before and changed the read function to return exactly the input size
A person said we have readexactly and I've seen that rating an exception
So do this commit
I think I should create another function without editing readexactly 🤔

Dreamsorcerer:
> I need to divide a large file into equal chunks using iter_chunked

I'm not sure why you would need to use aiohttp's iter_chunked to do this, it's still unclear what the overall goal is. It's sounding like a rather obscure use case. We probably don't want to merge something into this library unless it's going to be useful for a significant number of users (or atleast is very difficult to do without changes to the library).

For a use case like this, it is very easy to handle in the application code. e.g. Something like:
```
result = b""
async for chunk in iter_chunked():
    result += chunk
    while len(result) > part_size:
        save_part(result[:part_size])
        result = result[part_size:]
if result:
    save_part(result)
```
So, we're just not really seeing a reason this should be added to the library for everyone, if it might only be used by you.

SafaSafari:
Sure

PR #6894, opened 2022-08-23
Drop Python 3.6

Dreamsorcerer:
Think we've created a clash here by adding 3.11 support. But, the pinned version of frozenlist no longer supports Python 3.6.

Maybe we can switch versions conditionally to pass the tests, without dropping it in the 3.8 branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
     needs: gen_llhttp
     strategy:
       matrix:
-        pyver: [3.6, 3.7, 3.8, 3.9, '3.10']
+        pyver: [3.7, 3.8, 3.9, '3.10']
         no-extensions: ['', 'Y']
         os: [ubuntu, macos, windows]
         exclude:


Dreamsorcerer:
Well, the 3.6 test was failing due to pinning frozenlist, so we can either drop the testing for 3.6 or figure out how to pin an older version for 3.6.

No idea what this other error is though.

webknjaz:
> Well, the 3.6 test was failing due to pinning frozenlist, so we can either drop the testing for 3.6 or figure out how to pin an older version for 3.6.

Yep, I wouldn't drop it in 3.8.

webknjaz:
> No idea what this other error is though.

Try running `pytest --no-cov` locally, it helped me with `master`. The fix there was to update the pytest-related transitive deps. I thought I'd cherry-picked it into 3.8 already but maybe there's more to it. While you're on it, maybe start debugging with 3.9 — it looks like the problem might be the same but could be easier to debug on that branch.

Dreamsorcerer:
I recall spending hours to fix some issue at the beginning of the year. But, for some bizarre reason, the fix didn't work on the 3.9 branch (but, fixed master and 3.8). I don't think it's been looked at since (or maybe I misremember).

Dreamsorcerer:
I'm currently rebuilding the (archived) aiohttp-admin library for a client project. Once I've got that sorted, I should be able to spend some more time looking at any remaining issues here.

PR #6907, opened 2022-09-06
docs: Update redis to follow the latest spec

rapsealk:
## What do these changes do?

I changed some Redis code from the documentation to follow the latest API. It is based on redis-py>=4.2.x on which aioredis has been merged and started to be maintained.

Nope!

## Related issue number

There is no issue related to this PR.

## Checklist

Thank you for reading and any feedback is always welcomed!

### DIFF
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -906,7 +906,7 @@ below::
                   msg = await pubsub.get_message(ignore_subscribe_messages=True)
                   if msg is not None:
                       for ws in app['websockets']:
-                          ws.send_str('{}: {}'.format(channel, msg))
+                          await ws.send_str('{}: {}'.format(channel, msg))
               except asyncio.CancelledError:
                   break
 
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -896,26 +896,19 @@ background tasks could be registered as an :attr:`Application.on_startup`
 signal handler or :attr:`Application.cleanup_ctx` as shown in the example
 below::
 
-  import redis.asyncio as redis
-  from aiohttp import web
-
-
   async def listen_to_redis(app: web.Application):
       client = redis.from_url('redis://localhost:6379')
-      pubsub = client.pubsub()
       channel = 'news'
-      await pubsub.subscribe(channel)
-      while True:
-          try:
-              msg = await pubsub.get_message(ignore_subscribe_messages=True, timeout=1.0)
-              if msg is not None:
-                  for ws in app['websockets']:
-                      ws.send_str('{}: {}'.format(channel, msg))
-              await asyncio.sleep(0.01)
-          except asyncio.CancelledError:
-              break
-      await pubsub.unsubscribe(channel)
-      await pubsub.close()
+      async with client.pubsub() as pubsub:
+          await pubsub.subscribe(channel)
+          while True:
+              try:
+                  msg = await pubsub.get_message(ignore_subscribe_messages=True)
+                  if msg is not None:
+                      for ws in app['websockets']:
+                          ws.send_str('{}: {}'.format(channel, msg))
+              except asyncio.CancelledError:
+                  break
 
 
   async def background_tasks(app):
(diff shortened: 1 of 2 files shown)


docs/web_advanced.rst
```diff
@@ -896,20 +896,26 @@ background tasks could be registered as an :attr:`Application.on_startup`
...
-                  ws.send_str('{}: {}'.format(ch.name, msg))
-      except asyncio.CancelledError:
-          pass
-      finally:
-          await sub.unsubscribe(ch.name)
-          await sub.quit()
+  import redis.asyncio as redis
+  from aiohttp import web
+
+
```
```suggestion
```

docs/web_advanced.rst
```diff
@@ -896,20 +896,26 @@ background tasks could be registered as an :attr:`Application.on_startup`
...
+      pubsub = client.pubsub()
+      channel = 'news'
+      await pubsub.subscribe(channel)
+      while True:
+          try:
+              msg = await pubsub.get_message(ignore_subscribe_messages=True, timeout=1.0)
+              if msg is not None:
+                  for ws in app['websockets']:
+                      ws.send_str('{}: {}'.format(channel, msg))
+              await asyncio.sleep(0.01)
```
Dreamsorcerer: I know this is in the documentation examples, but this looks a little odd to me.
If we just need to yield back to the event loop, then `asyncio.sleep(0)` is better.
But, I'm not clear why we even want that.

docs/web_advanced.rst
```diff
@@ -896,20 +896,26 @@ background tasks could be registered as an :attr:`Application.on_startup`
...
+
+
+  async def listen_to_redis(app: web.Application):
+      client = redis.from_url('redis://localhost:6379')
+      pubsub = client.pubsub()
+      channel = 'news'
+      await pubsub.subscribe(channel)
+      while True:
+          try:
+              msg = await pubsub.get_message(ignore_subscribe_messages=True, timeout=1.0)
```
Dreamsorcerer: Another thing I'm not getting from the documentation. Why do we want a timeout? If we just retry immediately, shouldn't we just wait until actually get a response...?

I might have to open an issue to discuss these examples on the redis-py repo.

docs/web_advanced.rst
```diff
@@ -896,20 +896,26 @@ background tasks could be registered as an :attr:`Application.on_startup`
...
-      try:
-          sub = await aioredis.create_redis(('localhost', 6379))
-          ch, *_ = await sub.subscribe('news')
-          async for msg in ch.iter(encoding='utf-8'):
-              # Forward message to all connected websockets:
-              for ws in app[websockets]:
-                  ws.send_str('{}: {}'.format(ch.name, msg))
-      except asyncio.CancelledError:
-          pass
-      finally:
```
Dreamsorcerer, line 910: Would it not be a good idea to keep the `finally`? In case there are any errors.
I guess an even better approach would be if the API supported `async with pubsub:`.

docs/web_advanced.rst
Dreamsorcerer: Actually, I'd suggest removing the imports as well. This is meant to be a concise example of how to do something, not a fully working demo.

rapsealk:
@Dreamsorcerer Deeply appreciated your feedback! I made some changes following your advice.
Also, I found that Redis supports async context manager, it resets and releases a connection when `__aexit__` is called.
https://github.com/redis/redis-py/blob/510d8fe089addaea65055aae1e9606d0f40829d4/redis/asyncio/client.py#L651-L668

docs/web_advanced.rst
rapsealk: Applied `async with client.pubsub() as pubsub:` instead.

Dreamsorcerer:
https://github.com/redis/redis-py/issues/2419

docs/web_advanced.rst
```diff
@@ -896,20 +896,19 @@ background tasks could be registered as an :attr:`Application.on_startup`
...
+      client = redis.from_url('redis://localhost:6379')
+      channel = 'news'
+      async with client.pubsub() as pubsub:
+          await pubsub.subscribe(channel)
+          while True:
+              try:
+                  msg = await pubsub.get_message(ignore_subscribe_messages=True)
+                  if msg is not None:
+                      for ws in app['websockets']:
+                          ws.send_str('{}: {}'.format(channel, msg))
```
```suggestion
                          await ws.send_str('{}: {}'.format(channel, msg))
```

Dreamsorcerer:
Upstream examples have been updated, so this looks correct.

Issue #6917, opened 2022-09-17
ClientSession.timeout has an incorrect typing


DevilXD:
### Describe the bug

The `aiohttp.ClientSession.timeout` attribute has a type of `Union[object, aiohttp.ClientTimeout]`, however the code logic will never actually assign a bare `object` type to the `self._timeout` attribute, making this typing quite over-inclusive. Trying to use this attribute in typed code results in having to use `cast(aiohttp.ClientTimeout, session.timeout)`, which is far from ideal considering one can just fix the typing in the library.

I ran into this while using Python 3.8.10, but the exact same explanation above applies to the current master branch (and the version I'm using of course), as shown by the snippets below.

3.8 branch `__init__` parameter:
https://github.com/aio-libs/aiohttp/blob/6243204a6a6a0e5ff84ac754218381b44a841e72/aiohttp/client.py#L217

3.8 branch `self._timeout` assignment:
https://github.com/aio-libs/aiohttp/blob/6243204a6a6a0e5ff84ac754218381b44a841e72/aiohttp/client.py#L261-L290
Note the `# type: ignore` comment on `L278` there - it's because the `timeout is sentinel` check does not narrow down the `timeout` type. The correct way to go about this would be to use a `cast` there instead of ignoring the issue like that.

3.8 branch `timeout` attribute declaration:
https://github.com/aio-libs/aiohttp/blob/6243204a6a6a0e5ff84ac754218381b44a841e72/aiohttp/client.py#L1029-L1032

Master branch `__init__` parameter:
https://github.com/aio-libs/aiohttp/blob/52fa599c5637dd1a38761afb6829b0439b1cf505/aiohttp/client.py#L215

Master branch `self._timeout` assignment:
https://github.com/aio-libs/aiohttp/blob/52fa599c5637dd1a38761afb6829b0439b1cf505/aiohttp/client.py#L260-L263
Due to a different handling of the `sentinel` value via an `Enum` member, no `cast` is needed here.

Master branch `timeout` attribute declaration:
https://github.com/aio-libs/aiohttp/blob/52fa599c5637dd1a38761afb6829b0439b1cf505/aiohttp/client.py#L1008-L1011
The attribute type is still over-inclusive here though.

The solution would be quite simple:

```py
    @property
    def timeout(self) -> ClientTimeout:
        """Timeout for the session."""
        return self._timeout
````

[...]

PR #6923, opened 2022-09-20
Exclude the `object` possibility from the `ClientSession.timeout` type

DevilXD:
## What do these changes do?

This is a follow-up PR to #6917, which fixes the overly inclusive typing of `ClientSession.timeout` attribute, which can never actually end up as an `object` instance. This leftover probably exists from the time when there was the `_sentinel` value being assigned to the attribute (see "3.8 branch `self._timeout` assignment" from the linked issue), hence why it ended up this way.

For untyped code, there's absolutely no change - the `self._timeout` attribute was never actually assigned the sentinel value to. Most of the explanation already exists in #6917, here's the test code snippet from there again:

```py
import asyncio
import aiohttp

async def main:
    session = aiohttp.ClientSession(timeout=aiohttp.ClientTimeout(total=10))
    # read back the total time attribute
    total_time = session.timeout.total  # "object" type of "Union[object, ClientTimeout]" has no attribute "total"
    print(total_time)

asyncio.run(main())
```

For typed code, the current solution would be to `assert isinstance(session.timeout, aiohttp.ClientTimeout)` everywhere the attribute is being accessed. This PR removes this "unnecessary necessity".

## Related issue number

#6917

## Checklist

- [x] Documentation reflects the changes (documentation never mentioned an `object` instance being accessible from there even)


DevilXD:
I'm not sure if I deserve a `CONTRIBUTORS` entry for such a simple typing-only change (it specifically asks for "code modification" and this changes nothing other than typing), so I left it out for now (and take no issue with that). If it would be needed though, please let me know and I'll add a commit for it.

CHANGES/6917.misc
```diff
@@ -0,0 +1 @@
+Fix ClientSession.timeout having incorrect typing.
```
webknjaz: Plz try following the recommendations at https://github.com/aio-libs/aiohttp/tree/master/CHANGES#alright-so-how-to-add-a-news-fragment. Also, use RST formatting to highlight/link identifiers.

Dreamsorcerer:
> I'm not sure if I deserve a `CONTRIBUTORS` entry...

I've never even added my own name. :P
Not sure that file is really useful, personally

webknjaz:
> Merge branch 'master' into patch-1

@Dreamsorcerer FYI you can now select a rebase mode in that GH button (just click on the drop-down arrow)

Dreamsorcerer:
> @Dreamsorcerer FYI you can now select a rebase mode in that GH button (just click on the drop-down arrow)

Ah, it all get squash committed anyway though, so I'm not it makes any difference.

Dreamsorcerer (approved):
This seems obviously correct to me, probably didn't need the long story in the bug ticket. :P

Dreamsorcerer:
@webknjaz Seems like a simple fix for 3.8.2 if we've not already passed the point of no return?

webknjaz:
> This seems obviously correct to me, probably didn't need the long story in the bug ticket. :P

A good description is useful, since what's obvious for one person would be confusing for another. Especially, this applies to a “future you”, or a “trying to decide on backports you”, or a “Git archeologist-you”.

CHANGES/6917.misc
webknjaz: Also, this is probably a bugfix + maybe docs type.

webknjaz:
> @webknjaz Seems like a simple fix for 3.8.2 if we've not already passed the point of no return?

I guess I could get it in. But it still needs brushing up the change note.

CHANGES/6917.misc
webknjaz: I've made the updates myself.

Issue #6600, opened 2022-02-08
Fails to build on Python 3.11 - longintrepr.h: No such file or directory


gaborbernat:
### Describe the bug

Because of https://github.com/python/cpython/pull/28968, cython fixed it in https://github.com/cython/cython/pull/4428 and is released with 0.29.5

### To Reproduce

pip install aiohttp

### Expected behavior

To build the wheel.

### Logs/tracebacks

```console
Failed to build aiohttp yarl
  error: subprocess-exited-with-error

  × Building wheel for aiohttp (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [31 lines of output]
      *********************
      * Accelerated build *
      *********************
      running bdist_wheel
      running build
      running build_py
      running egg_info
      writing aiohttp.egg-info/PKG-INFO
      writing dependency_links to aiohttp.egg-info/dependency_links.txt
      writing requirements to aiohttp.egg-info/requires.txt
      writing top-level names to aiohttp.egg-info/top_level.txt
      reading manifest file 'aiohttp.egg-info/SOURCES.txt'
      reading manifest template 'MANIFEST.in'
      warning: no files found matching 'aiohttp' anywhere in distribution
      warning: no previously-included files matching '*.pyc' found anywhere in distribution
      warning: no previously-included files matching '*.pyd' found anywhere in distribution
      warning: no previously-included files matching '*.so' found anywhere in distribution
      warning: no previously-included files matching '*.lib' found anywhere in distribution
      warning: no previously-included files matching '*.dll' found anywhere in distribution
      warning: no previously-included files matching '*.a' found anywhere in distribution
      warning: no previously-included files matching '*.obj' found anywhere in distribution
      warning: no previously-included files found matching 'aiohttp/*.html'
      no previously-included directories found matching 'docs/_build'
      adding license file 'LICENSE.txt'
      running build_ext
      building 'aiohttp._websocket' extension
      aiohttp/_websocket.c:198:12: fatal error: longintrepr.h: No such file or directory
        198 |   #include "longintrepr.h"
            |            ^~~~~~~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
      [end of output]

```

### Python Version

```console
3.11
```

### aiohttp Version

```console
latest
```

### multidict Version

```console
n/a
```

### yarl Version

```console
latest
```

### OS

Any

### Related component

Server, Client

### Code of Conduct

[...]

PR #6550, opened 2022-01-23
Pin down multidict

Dreamsorcerer:
With the new errors coming in, it might be a good idea to pin this down in the 3.8 release.

PR #6950, opened 2022-09-21
Increase the upper boundary of the multidict dependency allowing v6

hugovk:
## What do these changes do?

The multidict dependency used to be `multidict >=4.5, < 7.0` in aiohttp 3.8.1, but was pinned down to `multidict >=4.5, < 6.0` in aiohttp 3.8.2 (https://github.com/aio-libs/aiohttp/pull/6550) in an attempt to help avoid new errors.

However, https://github.com/python/bedevere/ uses `multidict==6.0.2` so the requirements resolution fails with `aiohttp==3.8.2`:

```console
ERROR: Cannot install -r requirements.txt (line 1) and multidict==6.0.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested multidict==6.0.2
    aiohttp 3.8.2 depends on multidict<6.0 and >=4.5

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

https://github.com/hugovk/bedevere/actions/runs/3095794573/jobs/5010600772

None that I know of, other than they can install multidict v6 as with aiohttp 3.8.1.

## Related issue number

Partially related to closed https://github.com/aio-libs/aiohttp/issues/6600 and merged https://github.com/aio-libs/aiohttp/pull/6550.

## Checklist

- [n/a] Unit tests for the changes exist
- [n/a] Documentation reflects the changes
- [already there] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

### DIFF
--- a/CHANGES/6950.bugfix
+++ b/CHANGES/6950.bugfix
@@ -1,1 +1,7 @@
-Re-allow multidict version 6 as a dependency.
+Increased the upper boundary of the :doc:`multidict:index` dependency
+to allow for the version 6 -- by :user:`hugovk`.
+
+It used to be limited below version 7 in :doc:`aiohttp <index>` v3.8.1 but
+was lowered in v3.8.2 via :pr:`6550` and never brought back, causing
+the upgrading users problems with their dependency pins. :doc:`aiohttp
+<index>` v3.8.3 fixes that by recovering the original boundary of ``< 7``.
--- a/CHANGES/6950.bugfix
+++ b/CHANGES/6950.bugfix
@@ -1,0 +1,1 @@
+Re-allow multidict version 6 as a dependency.
--- a/CHANGES/6950.bugfix
+++ b/CHANGES/6950.bugfix
@@ -3,5 +3,5 @@ to allow for the version 6 -- by :user:`hugovk`.
 
 It used to be limited below version 7 in :doc:`aiohttp <index>` v3.8.1 but
 was lowered in v3.8.2 via :pr:`6550` and never brought back, causing
-the upgrading users problems with their dependency pins. :doc:`aiohttp
-<index>` v3.8.3 fixes that by recovering the original boundary of ``< 7``.
+problems with dependency pins when upgrading. :doc:`aiohttp <index>` v3.8.3
+fixes that by recovering the original boundary of ``< 7``.
--- a/setup.cfg
+++ b/setup.cfg
@@ -50,7 +50,7 @@ include_package_data = True
 install_requires =
   attrs >= 17.3.0
   charset-normalizer >=2.0, < 3.0
-  multidict >=4.5, < 6.0
+  multidict >=4.5, < 7.0
   async_timeout >= 4.0.0a3, < 5.0
   asynctest == 0.13.0; python_version<"3.8"
   yarl >= 1.0, < 2.0
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -196,9 +196,9 @@ class WebSocketResponse(StreamResponse):
         accept_val = base64.b64encode(
             hashlib.sha1(key.encode() + WS_KEY).digest()
         ).decode()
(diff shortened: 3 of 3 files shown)


Rongronggg9:
Facing exactly the same issue
https://github.com/Rongronggg9/RSS-to-Telegram-Bot/actions/runs/3094555753/jobs/5008056472

See also
https://pypi.org/pypi/aiohttp/json

I think this can be merged quickly since
https://github.com/aio-libs/aiohttp/pull/6550#issuecomment-1020960474

A new version (maybe v3.8.2.post0?) should be quickly released after merging or it may be a wide impact.

hugovk:
The CI passes on my fork: https://github.com/hugovk/aiohttp/actions/runs/3095913542 (I had to add a workaround to allow it to run on feature branches) ✅

webknjaz:
> should be quickly released

Unfortunately, the process isn't quick. So I've just yanked the release and will make a new ".3" with this instead.

setup.cfg
```diff
@@ -50,7 +50,7 @@ include_package_data = True
 install_requires =
   attrs >= 17.3.0
   charset-normalizer >=2.0, < 3.0
-  multidict >=4.5, < 6.0
+  multidict >=4.5, < 7.0
```
webknjaz, line 53: Based on https://github.com/aio-libs/aiohttp/pull/6550#issuecomment-1020960474, I'd also exclude the problematic versions, I suppose:
```suggestion
  multidict >=4.5, < 7.0, != 6.0.0, != 6.0.1
```

webknjaz:
So I verified that the constraints files can't be updated. Besides that, it seems like a few cosmetic improvements are needed here and it'll be good to go.

CHANGES/6950.bugfix
```diff
@@ -0,0 +1 @@
+Re-allow multidict version 6 as a dependency.
```
webknjaz:
```suggestion
Increased the upper boundary of the :doc:`multidict:index` dependency
to allow for the version 6 -- by :user:`hugovk`.

It used to be limited below version 7 in :doc:`aiohttp <index>` v3.8.1 but
was lowered in v3.8.2 via :pr:`6550` and never brought back, causing
the upgrading users problems with their dependency pins. :doc:`aiohttp
<index>` v3.8.3 fixes that by recovering the original boundary of ``< 7``.
```

setup.cfg
webknjaz: _(decided not to add the exclusions)_

CHANGES/6950.bugfix
```diff
@@ -0,0 +1,7 @@
+Increased the upper boundary of the :doc:`multidict:index` dependency
+to allow for the version 6 -- by :user:`hugovk`.
+
+It used to be limited below version 7 in :doc:`aiohttp <index>` v3.8.1 but
+was lowered in v3.8.2 via :pr:`6550` and never brought back, causing
+the upgrading users problems with their dependency pins. :doc:`aiohttp
+<index>` v3.8.3 fixes that by recovering the original boundary of ``< 7``.
```
webknjaz: @hugovk WDYT about the change note update? The preview is here: https://aiohttp--6950.org.readthedocs.build/en/6950/changes.html#to-be-included-in-vrelease-if-present.

hugovk: Looks good, how about changing:

> causing the upgrading users problems with their dependency pins.

to:

> causing problems with dependency pins when upgrading.

?

Rongronggg9: I agree with the below one too. Not only user-specified dependency pins but also other dependencies' (e.g. `bedevere`) dependency pins.

hugovk: Updated!

webknjaz: Thanks!

webknjaz (approved):

Issue #6915, opened 2022-09-15
Support dynamically updating authentication headers


faph:
### Is your feature request related to a problem?

See #6908 for discussion with @Dreamsorcerer 

We are have a continuously running service that makes HTTP requests using `aiohttp`'s client. These requests are sent with a bearer token in the authorization header. This token needs to be refreshed from time to time.

Therefore we cannot set the header on the client object. We need to evaluate the header prior to making individual requests. This is cumbersome for an app developer.

### Describe the solution you'd like

We would like to configure the client with the required data which can be used at each request to populate an authorization header. In its simplest form, this could be a simple co-routine that evaluates a token cache, generates a new token if required, and populate the header as required.

For the avoidance of doubt, the proposal is for wherever `aiohttp` supports an `auth` argument (client session, request), it would support the "enhanced" auth handler.

Currently, `aiohttp` supports a `BasicAuth` object which is a customized namedtuple (fields: username, password) with an `encode` method which returns the base64 encoded header value. Runtime checks are performed ensuring the `auth` value is an instance of that class.

~~TODO: expand on preferred option~~See options in comment below. 

### Describe alternatives you've considered

Both the `requests` library and the `httpx` library have similar functionality. For reference:

- https://requests.readthedocs.io/en/latest/user/authentication/#new-forms-of-authentication
- https://www.python-httpx.org/advanced/#customizing-authentication

The `httpx` approach supports a more defined base class/interface for an auth handler, whereas `requests` just requires a callable. Fundamentally they do the same thing and inject an authorization header in the request headers.

### Related component

Client

### Additional context

I am probably able to contribute a PR with an implementation, depending on the agreed scope of the changes.

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

faph:
### Implementation options

1. Do nothing. App developers can pass in headers manually as and when required. Or alternatively they could subclass ~~`ClientSession`~~ or `ClientRequest` (`ClientSession` is marked as "final" and should not be subclassed).

2. Add support for a generic coroutine **function** to be passed in for the `auth` argument. This would be in addition to the currently supported `BasicAuth` class. The coroutine function would accept a single argument, the `ClientRequest` instance and the coroutine would be free to modify the request (with the intention to just modify headers...). `ClientRequest.send` would call/await this coroutine. 

[...]

PR #6954, opened 2022-09-21
[WIP] Support dynamically updating authentication headers

faph:
[Work in progress]

Initial basic refactor of `BasicAuth` object, generalized as an async callable/coroutine.

## What do these changes do?

TODO

TODO

## Related issue number

#6915 

## Checklist


aiohttp/helpers.py
```diff
@@ -132,7 +136,14 @@ def iscoroutinefunction(func: Any) -> bool:
 json_re = re.compile(r"(?:application/|[\w.-]+/[\w.+-]+?\+)json$", re.IGNORECASE)
 
 
-class BasicAuth(namedtuple("BasicAuth", ["login", "password", "encoding"])):
+class Auth:
+    """Authorization handler base class"""
+
+    async def __call__(self, request) -> None:
```
webknjaz, line 137: `__call__()` is for normal callables, it'd be weird for it to be async. There's an `__await__()` that exists for awaitables. Maybe don't go for magic methods? It looks like you're reinventing a protocol pattern.

Also, it feels like for the use-case you described using an async generator interface could be a better fit then an async function.

faph: Thanks for the feedback @webknjaz. 

It might be useful to collect these high-level design considerations in the corresponding issue #6915. I was using this PR largely just to give an example what an implementation *could* look like.

PR #6967, opened 2022-09-23
Compute XOR table lazily

hauntsaninja:
This speeds up import time by a little bit:
```
In [2]: %timeit [bytes(a ^ b for a in range(256)) for b in range(256)]
8.41 ms ± 656 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
This is about 5% of aiohttp's import time


Dreamsorcerer:
Thanks

PR #6968, opened 2022-09-25
Make GunicornWebWorker exit with APP_LOAD_ERROR in case of a startup error

HallBregg:
Make GunicornWebWorker exit with WORKER_BOOT_ERROR (3) in case of an boot error.

I've found out that `GunicornWebWorker` exits with `0` error code even when an error occurs during startup. This makes Gunicorn endlessly spawn new workers.

With this pull request I would like to show the problem and ask if there is any other solution. I am aware of the comment `# ignore all finalization problems` and would like to ask about the context of this comment.

## What do these changes do?
These changes set `GunicornWebWorker.exit_code` to the value expected by the Gunicorn, which is `3` and set  `GunicornWebWorker.booted` to false.
```
class Arbiter(object):
    """
    Arbiter maintain the workers processes alive. It launches or
    kills them if needed. It also manages application reloading
    via SIGHUP/USR2.
    """

    # A flag indicating if a worker failed to
    # to boot. If a worker process exist with
    # this error code, the arbiter will terminate.
    WORKER_BOOT_ERROR = 3
```

## Checklist


aiohttp/worker.py
```diff
@@ -55,6 +56,9 @@ def run(self) -> None:
             self.loop.run_until_complete(self._task)
         except Exception:
             self.log.exception("Exception in gunicorn worker")
+            self.booted = False
+            self.exit_code = Arbiter.WORKER_BOOT_ERROR
```
Dreamsorcerer: Hmm, the gunicorn code seems to set `self.booted = True` before it calls `.run()`. This suggests to me that it is about recording whether the process was created, rather than whether the application startup succeeded. This seems to me like it should probably exit 1 instead.

If this was how gunicorn was supposed to work, then I'd expect `booted` to still be `False` and then we would set it to `True` in `_run()` after the setup is complete (line 95).

Looking at some of the other workers, I'm also noticing gtornado using `alive` and `server_alive`, not sure if something like that is more appropriate:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gtornado.py#L91-L92

Also wondering if we should actually be using the `AsyncWorker` as our base class:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base_async.py

Dreamsorcerer: In fact there is an APP_LOAD_ERROR, which sounds a lot more accurate for this situation:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L34

HallBregg: I've been thinking about using APP_LOAD_ERROR before, but decided to stay with WORKER_BOOT_ERROR (I didn't know when we can say the app is loaded, but the same problem I have got with a worker). Nevertheless you are probably right about setting `booted` to `False`after it has already been set to `True`, so APP_LOAD_ERROR seems to be correct.

I will take a look at the other workers to have a bit more knowledge.

Dreamsorcerer: > (I didn't know when we can say the app is loaded, but the same problem I have got with a worker).

When the setup has completed on line 95, the application has been initialised. If we go this way, then I'd suggest adding a `self.started = False` and setting it to `True` after `site.start()` on line 105. Then only set this exit status if `not self.started`, otherwise a `1` is probably a reasonable exit code (likely indicating that an exception happened in the app's cleanup).

aiohttp/worker.py
```diff
@@ -55,6 +57,11 @@ def run(self) -> None:
             self.loop.run_until_complete(self._task)
         except Exception:
             self.log.exception("Exception in gunicorn worker")
+            if self.started:
+                self.exit_code = 1
+            else:
+                self.exit_code = Arbiter.APP_LOAD_ERROR
```
Dreamsorcerer:
```suggestion
            self.exit_code = 1 if if self.started else Arbiter.APP_LOAD_ERROR
```

Personally, I think this is a little cleaner, avoiding the nesting.

Dreamsorcerer (approved):
I'm hoping one of the gunicorn maintainers might provide some feedback, but if we don't hear anything back in a few days, then this looks reasonable to me.

aiohttp/worker.py
HallBregg: I agree. But I didn't know which syntax is preferred in aiohttp project. That's why I chose that. But of course I will update it.

Dreamsorcerer:
Seems they've now closed the issue without providing any useful feedback. So, I guess we'll just go with this then.
https://github.com/benoitc/gunicorn/issues/2867

Dreamsorcerer:
OK, they have provided feedback now. Sounds like this is not the desired approach:
https://github.com/benoitc/gunicorn/issues/2867#issuecomment-1537554827

Dreamsorcerer:
#12879

PR #6983, opened 2022-09-28
Make `ClientSession.headers` a settable property

faph:
## What do these changes do?

After a client session has been initialized, default session-wide headers may be updated. This allows application developers to update authorization headers periodically, for example.

No, except that in addition to reading `ClientSession.headers`, users may now set this property. Existing behavior does not change.

## Related issue number

Issue #6954 

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -267,12 +267,7 @@ class ClientSession:
         self._requote_redirect_url = requote_redirect_url
         self._read_bufsize = read_bufsize
 
-        # Convert to list of tuples
-        if headers:
-            real_headers: CIMultiDict[str] = CIMultiDict(headers)
-        else:
-            real_headers = CIMultiDict()
-        self._default_headers: CIMultiDict[str] = real_headers
+        self.headers = headers  # See `headers` property setter function
         if skip_auto_headers is not None:
             self._skip_auto_headers = frozenset(istr(i) for i in skip_auto_headers)
         else:
@@ -1014,6 +1009,12 @@ class ClientSession:
     def headers(self) -> "CIMultiDict[str]":
         """The default headers of the client session."""
         return self._default_headers
+
+    @headers.setter
+    def headers(self, value: Optional[LooseHeaders] = None) -> None:
+        """Set or update the default headers of the client session."""
+        value = value or {}
+        self._default_headers = CIMultiDict(value)
 
     @property
     def skip_auto_headers(self) -> FrozenSet[istr]:
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -267,7 +267,10 @@ class ClientSession:
         self._requote_redirect_url = requote_redirect_url
         self._read_bufsize = read_bufsize
 
-        self.headers = headers  # See `headers` property setter function
+        # See `headers` property setter function. Mypy struggling with the property
+        # returning a stricter type than the setter function allows.
+        # https://github.com/python/mypy/issues/3004
+        self.headers = headers  # type: ignore
         if skip_auto_headers is not None:
             self._skip_auto_headers = frozenset(istr(i) for i in skip_auto_headers)
         else:
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -124,8 +124,10 @@ async def test_init_cookies_with_list_of_tuples(create_session: Any) -> None:
 async def test_headers_can_be_updated(create_session: Any) -> None:
     session = await create_session()
     session.headers = {"h1": "header1", "h2": "header2"}
-    assert session.headers == session._default_headers == CIMultiDict(
-        [("h1", "header1"), ("h2", "header2")]
+    assert (
+        session.headers
+        == session._default_headers
+        == CIMultiDict([("h1", "header1"), ("h2", "header2")])
     )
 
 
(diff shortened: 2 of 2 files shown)


teuneboon:
I was writing a pull request for this exact feature, decided to check open pull requests to see how this project likes its pull requests and ran across this one as the first open one(hey, at least this means I'm not the only one running into this).

My only suggestion would be adding this file: `CHANGES/6983.feature`:

    ``ClientSession.headers`` is now a settable property.

faph:
@teuneboon Yeah, once this is deemed acceptable and pipelines authorized I will update the changelog etc.

@Dreamsorcerer gentle nudge...

Dreamsorcerer:
My inclination is that you shouldn't be able to set the dict like that. If you really want to completely replace the headers arbitrarily, then just use the dict methods:
```
client.headers.clear()
client.headers.update(new_headers)
```

But, the typical use case would likely just need the `.update()` alone, or just a `client.headers["Authorization"] = value`.

faph:
@Dreamsorcerer ~~I dont understand. `ClientSession.headers` property not having a setter function, one should not be able to *modify* the dict either. I'll have to try this to be sure. Since there's a getter only and the actually underlying attribute is `_default_headers`. I don't know how modifying *could* work. Even with a setter function, updating a dict would not work I guess?~~

Oh I see, the getter returns a pointer to the dict from attribute `_default_headers` and one can just update that in place. Mmh, I guess. In that case, this PR is not needed at all...

I agree the test should be whether one can do ``client.headers["Authorization"] = ...``. Let me add that test.

faph:
@Dreamsorcerer @webknjaz If someone could authorize the pipelines to be run, that would be great, btw.

Dreamsorcerer:
> @Dreamsorcerer I dont understand. `ClientSession.headers` property not having a setter function, one should not be able to _modify_ the dict either. I'll have to try this to be sure.

No, they are unrelated. When you type `client.headers` that is the getter, then the `[key] = value` is executed on the returned object.

```
>>> class A:
...  def __init__(self):
...   self._a = {}
...  @property
...  def a(self):
...   return self._a
... 
>>> a = A()
>>> a.a["foo"] = 1
>>> a.a
{'foo': 1}
```

faph:
Yeah, took me a minute to realize...

Dreamsorcerer:
The point is that you shouldn't be able to replace the actual dict object, but you can modify the values in the dict.

aiohttp/client.py
```diff
@@ -267,12 +267,10 @@ def __init__(
...
-        # Convert to list of tuples
-        if headers:
-            real_headers: CIMultiDict[str] = CIMultiDict(headers)
-        else:
-            real_headers = CIMultiDict()
-        self._default_headers: CIMultiDict[str] = real_headers
+        # See `headers` property setter function. Mypy struggling with the property
+        # returning a stricter type than the setter function allows.
+        # https://github.com/python/mypy/issues/3004
+        self.headers = headers  # type: ignore
```
webknjaz, line 273: Type ignores must never be wholesale. They should include the specific rule they are meant to suppress, and should not suppress all the rules existing now or in the future.

teuneboon:
To me modifying the dict felt a little dirty when you can't set the dict, but I can't exactly point my finger to why I find that "dirty". If this is the "officially" intended way to do this I'm fine with it and I'll just write my code so it modifies the dict instead.

EDIT: and to clarify the usecase: I need to modify the headers because I have a long-lived session object that sends an Authorization header with an access token that gets refreshed every 15 minutes.

Dreamsorcerer:
> To me modifying the dict felt a little dirty when you can't set the dict, but I can't exactly point my finger to why I find that "dirty". If this is the "officially" intended way to do this I'm fine with it and I'll just write my code so it modifies the dict instead.

The idea is that `client.headers` is a mutable value, you can change it. But, the exact type and objects used for that is internal to the session object, you shouldn't be able to mess around with that.

Having a setter which converts it to the type doesn't seem terrible, but it still feels like an antipattern, and the one thing that could cause problems is if something is expecting to keep a reference to that object. e.g. Totally vague made-up example:
```
def auth1_update(headers):
    # On some event
    headers["Auth1"] = foo

def auth2_update(session):
    # On some other event
    session.headers = {"Auth2": foo}

auth1_update(session.headers)
auth2_update(session)
```
Obviously, in this example auth1 would keep updating the headers in a dict which is no longer referenced by the session because auth2 has removed them.

> EDIT: and to clarify the usecase: I need to modify the headers because I have a long-lived session object that sends an Authorization header with an access token that gets refreshed every 15 minutes.

If you only have 1 header to update, then `session.headers["Authorization"] = value` seems clearer to me than `session.headers = {"Authorization": foo}`.

teuneboon:
Alright, clearly reasoned. I've seen code where something that's coded like this was meant to be truly read-only(should've been an immutable dict instead in those projects damnit!), but I'll just rewrite my code to update the headers then.

Dreamsorcerer:
> Alright, clearly reasoned. I've seen code where something that's coded like this was meant to be truly read-only(should've been an immutable dict instead in those projects damnit!), but I'll just rewrite my code to update the headers then.

Fair enough. We have immutable dicts where we actually want them to be read-only (e.g. `FrozenList` for `Application.on_startup`).

faph:
I think knowing whether something mutable should or should not be mutated is always a thing.

PR #7014, opened 2022-10-12
Cover ttl=0 (expire always) _DNSCacheTable case

laky55555:
## What do these changes do?
Fix bug when using ttl_dns_cache=0

When using `ttl_dns_cache=0`, hostname is added to cache and on expire check there is a KeyError as it is never added into _timestamp dictionary.
Added a short description to docs ClientSession advising to use `use_dns_cache` instead of `ttl_dns_cache=0` for always expiring cache. 

Alternative to this change is to enforce user to use positive none 0 float or None for `ttl_dns_cache` during init time and in case of using 0 implicitly change `use_dns_cache` to False or raise exception.
No.

## Checklist

### DIFF
--- a/CHANGES/7014.bugfix
+++ b/CHANGES/7014.bugfix
@@ -1,0 +1,1 @@
+Fixed bug when using TCPConnector with ttl_dns_cache=0.
--- a/CHANGES/7014.bugfix
+++ b/CHANGES/7014.bugfix
@@ -1,1 +1,1 @@
-Fixed bug when using TCPConnector with ttl_dns_cache=0.
+Fixed bug when using ``TCPConnector`` with ``ttl_dns_cache=0``.
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -678,13 +678,13 @@ class _DNSCacheTable:
     def add(self, key: Tuple[str, int], addrs: List[Dict[str, Any]]) -> None:
         self._addrs_rr[key] = (cycle(addrs), len(addrs))
 
-        if self._ttl:
+        if self._ttl is not None:
             self._timestamps[key] = monotonic()
 
     def remove(self, key: Tuple[str, int]) -> None:
         self._addrs_rr.pop(key, None)
 
-        if self._ttl:
+        if self._ttl is not None:
             self._timestamps.pop(key, None)
 
     def clear(self) -> None:
@@ -701,8 +701,6 @@ class _DNSCacheTable:
     def expired(self, key: Tuple[str, int]) -> bool:
         if self._ttl is None:
             return False
-        elif self._ttl == 0:
-            return True
 
         return self._timestamps[key] + self._ttl < monotonic()
 
@@ -719,7 +717,6 @@ class TCPConnector(BaseConnector):
         resolver
     use_dns_cache - Use memory cache for DNS lookups.
     ttl_dns_cache - Max seconds having cached a DNS entry, None forever.
-        0 expire immediately (use use_dns_cache=False instead). 10 by default.
     family - socket address family
     local_addr - local tuple of (host, port) to bind socket to
 
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -701,6 +701,8 @@ class _DNSCacheTable:
     def expired(self, key: Tuple[str, int]) -> bool:
         if self._ttl is None:
             return False
+        elif self._ttl == 0:
+            return True
 
         return self._timestamps[key] + self._ttl < monotonic()
 
@@ -717,6 +719,7 @@ class TCPConnector(BaseConnector):
         resolver
     use_dns_cache - Use memory cache for DNS lookups.
     ttl_dns_cache - Max seconds having cached a DNS entry, None forever.
+        0 expire immediately (use use_dns_cache=False instead). 10 by default.
     family - socket address family
     local_addr - local tuple of (host, port) to bind socket to
 
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -2190,7 +2190,9 @@ class TestDNSCacheTable:
 
     def test_never_expire(self) -> None:
         dns_cache_table = _DNSCacheTable(ttl=None)
+        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 1)
         dns_cache_table.add("localhost", ["127.0.0.1"])
+        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 10000000)
         assert not dns_cache_table.expired("localhost")
 
     def test_always_expire(self, monkeypatch: pytest.MonkeyPatch) -> None:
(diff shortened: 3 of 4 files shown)


tests/test_connector.py
```diff
@@ -2185,6 +2185,17 @@ async def test_expired_ttl(self, loop: Any) -> None:
         await asyncio.sleep(0.02)
         assert dns_cache_table.expired("localhost")
 
+    async def test_never_expire(self, loop: Any) -> None:
+        dns_cache_table = _DNSCacheTable(ttl=None)
+        dns_cache_table.add("localhost", ["127.0.0.1"])
+        await asyncio.sleep(0.02)
```
Dreamsorcerer: Not sure sleeping for 20ms is a fair test (or efficient). Can we just patch the monotonic() function to move the clock forward by a day or similar?

aiohttp/connector.py
```diff
@@ -701,6 +701,8 @@ def next_addrs(self, key: Tuple[str, int]) -> List[Dict[str, Any]]:
     def expired(self, key: Tuple[str, int]) -> bool:
         if self._ttl is None:
             return False
+        elif self._ttl == 0:
+            return True
```
Dreamsorcerer: Hmm, I feel that this case is not common (I'm not sure anyone else has reported the bug before), so I'd be more inclined to do an `is None` check in the other methods, rather than add additional logic here which would reduce performance in the typical case.

Dreamsorcerer: Also, it should be pretty obvious that `use_dns_cache=False` is the more efficient option.

aiohttp/connector.py
```diff
@@ -717,6 +719,7 @@ class TCPConnector(BaseConnector):
         resolver
     use_dns_cache - Use memory cache for DNS lookups.
     ttl_dns_cache - Max seconds having cached a DNS entry, None forever.
+        0 expire immediately (use use_dns_cache=False instead). 10 by default.
```
Dreamsorcerer: I'd revert this given the above comments. Also the docstring doesn't get output into the docs (You'll see the default value is already mentioned in the actual docs: https://docs.aiohttp.org/en/stable/client_reference.html#tcpconnector).

Dreamsorcerer (changes requested):

tests/test_connector.py
laky55555: Fixed

aiohttp/connector.py
laky55555: Removed

aiohttp/connector.py
laky55555: I agree. Change to `is not None` per suggestion to fix crashing on `ttl = 0`.

Also, do you think it makes sense in case when user passes number <=0 to `ttl_dns_cache` to override `use_dns_cache` to `False` as behaviour will be the same but performance would be better?

laky55555:
Thank you for reviewing the PR :smiley:

aiohttp/connector.py
Dreamsorcerer: I think if we were designing it from scratch it would make sense to have only 1 option, rather than both. Maybe it's worth removing the `use_dns_cache` option in v4...

tests/test_connector.py
```diff
@@ -2179,10 +2179,23 @@ def test_not_expired_ttl(self) -> None:
         dns_cache_table.add("localhost", ["127.0.0.1"])
         assert not dns_cache_table.expired("localhost")
 
-    async def test_expired_ttl(self, loop: Any) -> None:
+    async def test_expired_ttl(self, loop: Any, monkeypatch: Any) -> None:
```
Dreamsorcerer:
```suggestion
    async def test_expired_ttl(self, loop: Any, monkeypatch: pytest.MonkeyPatch) -> None:
```

tests/test_connector.py
```diff
@@ -2179,10 +2179,23 @@ def test_not_expired_ttl(self) -> None:
...
         dns_cache_table = _DNSCacheTable(ttl=0.01)
+        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 1)
+        dns_cache_table.add("localhost", ["127.0.0.1"])
+        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 1.02)
+        assert dns_cache_table.expired("localhost")
```
Dreamsorcerer:
```suggestion
        dns_cache_table = _DNSCacheTable(ttl=1)
        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 1)
        dns_cache_table.add("localhost", ["127.0.0.1"])
        monkeypatch.setattr("aiohttp.connector.monotonic", lambda: 3)
        assert dns_cache_table.expired("localhost")
```
Think we can just do simple numbers here, given that we are patching the clock.

CHANGES/7014.bugfix
```diff
@@ -0,0 +1 @@
+Fixed bug when using TCPConnector with ttl_dns_cache=0.
```
Dreamsorcerer:
```suggestion
Fixed bug when using ``TCPConnector`` with ``ttl_dns_cache=0``.
```

Think that'll fix the spell check problem.

tests/test_connector.py
laky55555: Added typehint + removed now obsolete `loop` fixture

tests/test_connector.py
laky55555: Makes sense, added "nicer" numbers.

CHANGES/7014.bugfix
laky55555: Thanks for the hint. :crossed_fingers:

Dreamsorcerer:
It would be preferable if you didn't keep force-pushing branches, it makes it impossible to view diffs of new changes.
Let me know when you're finished.

laky55555:
Sorry for force push, I though it would be easier to check.

From my side I think it should be ok. 
Do I need to do something about the 3.11 tests that are failing or is that expected?

Dreamsorcerer:
Looks all good to me. I'd just like to figure out what codecov is complaining about.
Looks like it says only partial coverage of the monkey patches in the `None` test. Presumably, this is because the clock functions are never called in this case? Which means we should be able to remove those 2 lines.

> Sorry for force push, I though it would be easier to check.

All good. We squash commit, so there's no point doing anything other than regular commits on the branch.

laky55555:
> Looks like it says only partial coverage of the monkey patches in the `None` test.

Thanks for the hint, hopefully that should do the trick.

Dreamsorcerer:
I remembered that we were testing that it never expires though. So, best to keep it in. We'll just ignore the partial codecov.

Dreamsorcerer:
@laky55555 If you've got time to do the backport as above, that'd be great.

laky55555:
@Dreamsorcerer thanks you for fixing the test and merging.
I followed the instructions of pathcback and hopefully backport is ok

PR #7025, opened 2022-10-16
Fix expect handler when returning a response

Dreamsorcerer:


tests/test_web_functional.py
```diff
@@ -576,6 +576,31 @@ async def expect_handler(request):
...
+    cache = {"foo": "bar"}
+
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response(text="handler")
+
+    async def expect_handler(request: web.Request) -> Optional[web.Response]:
+        k = request.headers.get("X-Key")
+        cached_value = cache.get(k)
+        if cached_value:
+            return web.Response(text=cached_value)
```
Dreamsorcerer, line 589: I was trying to think of a vaguely realistic use case for this. So, my idea was something that uses a header to look for a cached response, without needing to read the body.

PR #7043, opened 2022-10-23
Add clarification about `GracefulExit` when using `handle_signals=True`

### LINKED ISSUE #4414: Document the need to catch `GracefulExit` if `handle_signals=True`
## Long story short

I've been using low-level machinery exposed by aiohttp for my framework.
When I learned about `handle_signals=True` I started testing it and it turned out that my process always exited with return code 1 on `SIGTERM` and `SIGINT`.

This was confusing and I thought that it's a bug. But then, after diving into the code deeper, I realized that with this option enabled, aiohttp binds the corresponding handler to process these signals and the only thing they do is [raising `GracefulExit`](https://github.com/aio-libs/aiohttp/blob/95e548c31efe1f53cc0a5eeb919fa8d9e666f4d9/aiohttp/web_runner.py#L29-L34) (which is a subclass of `SystemExit` with the return code defaulting to 1).

Turns out that it works but only if one uses `aiohttp.web.run_app()`. This was [the last bit of the puzzle](https://github.com/aio-libs/aiohttp/blob/95e548c31efe1f53cc0a5eeb919fa8d9e666f4d9/aiohttp/web.py#L434-L435) for me: using low-level APIs implies that I mostly don't rely on high-level interfaces and thus have to catch the `GracefulExit` exception in my own runner code.

So my request here is that this should be documented.

*I'm marking this as a "good first issue" so that new contributors could have something easy to start with.*

## Expected behaviour

Docs clarifying this.

## Actual behaviour

Docs don't mention this detail.

## Steps to reproduce

N/A

## Your environment

N/A

Daste745:
## What do these changes do?

Clarify that `GracefulExit` needs to be handled manually when using `handle_signals=True` in `AppRunner` and `ServerRunner`.

N/A

## Related issue number

Fixes #4414

## Checklist


Daste745:
The issue #4414 is quite old, but I think this still needed addressing. I added a short mention about the need to handle `GracefulExit` signals in the `handle_signals` param description. If this is not enough, I could expand it to a 'Note' or 'Warning' section.

Dreamsorcerer:
I'll let @webknjaz look over it, as it was his issue.

CHANGES/4414.doc
```diff
@@ -0,0 +1 @@
+Clarify that `GracefulExit` needs to be handled in `AppRunner` and `ServerRunner` when using `handle_signals=True`.
```
webknjaz: This should use RST syntax, not Markdown: https://aiohttp--7043.org.readthedocs.build/en/7043/changes.html#improved-documentation.

webknjaz: Also, please use the past tense and optionally add
```rst
-- by :user:`Daste745`
```

webknjaz (approved):
Other than the necessary change note improvements, this LGTM.

CHANGES/4414.doc
Daste745: Should be fixed in [e075e56](https://github.com/aio-libs/aiohttp/pull/7043/commits/e075e56959937c454ec26d08a003ad766e604c0d).

webknjaz:
Cool, thanks!

Daste745:
We've got some CI errors, probably because `GracefulExit` doesn't have any section in the docs and the linter can't create a reference. I don't see an obvious place to put more information about `GracefulExit`, as it's only used for this one purpose.

Dreamsorcerer:
Hmm, should it be added to the reference section then? Previously, it was considered a private inner working of the library, and not meant to be used by a user. But, if we are now expecting users to catch the exception, then maybe it needs to be documented as part of the public API now..

Daste745:
I agree. I'll try to find a place to document the `GracefulExit` API somewhere in the docs.

Daste745:
I added it under the same section that `AppRunner` and `ServerRunner` live. I'm not sure if it's the best spot though.

Dreamsorcerer (approved):

docs/web_reference.rst
```diff
@@ -2577,7 +2577,8 @@ application on specific TCP or Unix socket, e.g.::
    :param bool handle_signals: add signal handlers for
                                :data:`signal.SIGINT` and
                                :data:`signal.SIGTERM` (``False`` by
-                               default).
+                               default). These signals will raise
```
webknjaz:
```suggestion
                               default). These handlers will raise
```

docs/web_reference.rst
```diff
@@ -2648,7 +2649,8 @@ application on specific TCP or Unix socket, e.g.::
...
```
webknjaz:
```suggestion
                               default). These handlers will raise
```

docs/web_reference.rst
```diff
@@ -2773,6 +2775,15 @@ application on specific TCP or Unix socket, e.g.::
 
                        ``128`` by default.
 
+.. exception:: GracefulExit
+
+   Raised by classes which inherit from :class:`BaseRunner` whenever
+   the ``handle_signals`` parameter is set to ``True``.
+
+   Inherited from :exc:`SystemExit`,
+   which exits with error code 1 if not handled.
```
webknjaz:
```suggestion
   which exits with error code ``1`` if not handled.
```

docs/web_reference.rst
```diff
@@ -2773,6 +2775,15 @@ application on specific TCP or Unix socket, e.g.::
 
                        ``128`` by default.
 
+.. exception:: GracefulExit
+
+   Raised by classes which inherit from :class:`BaseRunner` whenever
```
webknjaz: I don't like that it says that classes raise exceptions. They are containers for stuff, the classes don't do anything short of existing. Methods could raise exceptions but not classes.
Any chance to rephrase this?

Daste745: Hmm, agreed, I used the wrong words here. How about "Raised by signal handlers registered in classes which inherit from :class:`BaseRunner` whenever the ``handle_signals`` parameter is set to ``True``."?

Daste745: Or maybe we should skip the abstraction of "classes which inherit from `BaseRunner`" and just specify that this applies to `AppRunner` and `ServerRunner`? It would be more explicit this way.

webknjaz: I suppose we could try the latter.

Issue #5269, opened 2020-11-22
invalid character in header


zhaoxuan:
🐞 **Describe the bug**
```
[2020-11-22 21:24:53 +0800] [55277] [ERROR] Error handling request
Traceback (most recent call last):
  File "/opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages/aiohttp/web_protocol.py", line 314, in data_received
    messages, upgraded, tail = self._request_parser.feed_data(data)
  File "aiohttp/_http_parser.pyx", line 546, in aiohttp._http_parser.HttpParser.feed_data
aiohttp.http_exceptions.BadHttpMessage: 400, message='invalid character in header'
```

💡 **To Reproduce**

I do not have any idea about it, the CPU pressure was not hight.

```
aiohttp == 3.7.3
gunicorn == 20.0.4
uvloop == 0.14.0
```

📋 **Your version of the Python**

```console
Python == 3.6.9
```

📋 **Your version of the aiohttp/yarl/multidict distributions**

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.3
Summary: Async http client/server framework (asyncio)
Home-page: https://github.com/aio-libs/aiohttp
Author: Nikolay Kim
Author-email: fafhrd91@gmail.com
License: Apache 2
Location: /opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages
Requires: multidict, typing-extensions, idna-ssl, chardet, attrs, yarl, async-timeout
Required-by: aiohttp-devtools
```
```console
$ python -m pip show multidict
Name: multidict
Version: 4.7.6
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages
Requires:
Required-by: yarl, aiohttp
```
```console
$ python -m pip show yarl
Name: yarl
Version: 1.5.1
Summary: Yet another URL library
Home-page: https://github.com/aio-libs/yarl/
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Location: /opt/rh/rh-python36/root/usr/lib64/python3.6/site-packages
Requires: multidict, typing-extensions, idna
Required-by: aiohttp
```

webknjaz:
[...]

PR #7044, opened 2022-10-24
Fix unicode errors

ret2libc:
## What do these changes do?

Fix few errors related to Unicode decoding.
* multipart forms with invalid utf-8 characters as data can cause a UnicodeDecodeError to be raised. This should be raised as ValueError, like what is done in other parts of the code base.
* HTTP request parser (pure-python) tries to decode the header name with utf-8/xmlcharrefreplace, which cannot decode bytes such as `\xd9`. 

I don't think so.

## Checklist

- [ ] ~Documentation reflects the changes~
- [x] ~Add a new news fragment into the `CHANGES` folder~

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -154,7 +154,7 @@ class HeadersParser:
             if len(bname) > self.max_field_size:
                 raise LineTooLong(
                     "request header name {}".format(
-                        bname.decode("utf8", "xmlcharrefreplace")
+                        bname.decode("utf8", "backslashreplace")
                     ),
                     str(self.max_field_size),
                     str(len(bname)),
@@ -176,7 +176,7 @@ class HeadersParser:
                     if header_length > self.max_field_size:
                         raise LineTooLong(
                             "request header field {}".format(
-                                bname.decode("utf8", "xmlcharrefreplace")
+                                bname.decode("utf8", "backslashreplace")
                             ),
                             str(self.max_field_size),
                             str(header_length),
@@ -197,7 +197,7 @@ class HeadersParser:
                 if header_length > self.max_field_size:
                     raise LineTooLong(
                         "request header field {}".format(
-                            bname.decode("utf8", "xmlcharrefreplace")
+                            bname.decode("utf8", "backslashreplace")
                         ),
                         str(self.max_field_size),
                         str(header_length),
--- a/tests/test_http_parser.py
+++ b/tests/test_http_parser.py
@@ -117,6 +117,14 @@ test2: data\r
     assert not msg.upgrade
 
 
+def test_parse_headers_longline(parser: Any) -> None:
+    invalid_unicode_byte = b"\xd9"
+    header_name = b"Test" + invalid_unicode_byte + b"Header" + b"A" * 8192
+    text = b"GET /test HTTP/1.1\r\n" + header_name + b": test\r\n" + b"\r\n" + b"\r\n"
+    with pytest.raises((http_exceptions.LineTooLong, http_exceptions.BadHttpMessage)):
+        parser.feed_data(text)
+
+
 def test_parse(parser: Any) -> None:
     text = b"GET /test HTTP/1.1\r\n\r\n"
     messages, upgrade, tail = parser.feed_data(text)
--- a/tests/test_multipart.py
+++ b/tests/test_multipart.py
@@ -599,6 +599,21 @@ class TestPartReader:
             result = await obj.form()
         assert [("foo", "bar"), ("foo", "baz"), ("boo", "")] == result
 
+    async def test_read_form_invalid_utf8(self, newline: Any) -> None:
+        invalid_unicode_byte = b"\xff"
+        data = invalid_unicode_byte + b"%s--:--" % newline
+        with Stream(data) as stream:
+            obj = aiohttp.BodyPartReader(
+                BOUNDARY,
+                {CONTENT_TYPE: "application/x-www-form-urlencoded"},
+                stream,
+                _newline=newline,
+            )
+            with pytest.raises(
+                ValueError, match="data cannot be decoded with utf-8 encoding"
+            ):
+                await obj.form()
+
     async def test_read_form_encoding(self, newline: Any) -> None:
         data = b"foo=bar&foo=baz&boo=%s--:--" % newline
         with Stream(data) as stream:
--- a/CHANGES/7044.bugfix
+++ b/CHANGES/7044.bugfix
@@ -1,0 +1,1 @@
+Avoid raising UnicodeDecodeError in multipart and in HTTP headers parsing.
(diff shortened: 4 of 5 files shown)


webknjaz:
Thanks for the PR @ret2libc!

> Fix few errors related to Unicode decoding.

Could you spare a few more details on how the problem occurs? Where is it visible to the end-users? Is it different for the Cython-based parser vs. the pure Python one you're editing here?

Also, I'd expect to see some regression tests bundled with PRs modifying such core functionality. Since we have two parser implementations, this would also help us make sure that their behavior is equivalent.

ret2libc:
I added unit tests, hope they are ok!

While doing so, I noticed the Cython-based parser does not have the same problem, but it raises a InvalidHeader error. What would be the best way forward? The python-based parser accepts the header in the test, thus I had to replace `xmlcharrefreplace` with `backslackreplace`, however it might be better to just reject the header?

Thanks for your review!

(and happy to see you again ;) )

aiohttp/http_parser.py
```diff
@@ -154,7 +154,7 @@ def parse_headers(
             if len(bname) > self.max_field_size:
                 raise LineTooLong(
                     "request header name {}".format(
-                        bname.decode("utf8", "xmlcharrefreplace")
+                        bname.decode("utf8", "backslashreplace")
```
Dreamsorcerer, line 157: To clarify what the change does, this results in any invalid utf-8 bytes being transformed into literal characters:
```
>>> b"\xd9".decode("utf-8", "backslashreplace")
'\\xd9'
```
So, the invalid continuation byte becomes a literal `\xd9`, i.e. now 4 bytes/characters instead of 1 byte.

No idea if this is the best option, but seems better than the error produced by the current code.

webknjaz:
> however it might be better to just reject the header?

This sounds like a good idea. I can't imagine anybody wanting to process an invalid header in their web application.
But it'd also be reasonable to double-check with the RFCs what it allows/suggests/recommends, just in case.

tests/test_http_parser.py
```diff
@@ -117,6 +117,20 @@ def test_parse_headers(parser: Any) -> None:
     assert not msg.upgrade
 
 
+def test_parse_headers_longline(parser: Any) -> None:
+    text = (
+        b"""GET /test HTTP/1.1\r
```
webknjaz: Why is it `\r` and not `\r\n`

webknjaz: Oh.. that `\n` is probably represented by the actual line feed in the literal :exploding_head:

tests/test_http_parser.py
```diff
@@ -117,6 +117,20 @@ def test_parse_headers(parser: Any) -> None:
...
+Test\xd9Header"""
```
webknjaz: Could you put `b'\xd9'` into a dedicated variable (`invalid_unicode_byte`?) and interpolate it into this string? This way, it'd be cleaner what the intent is.

tests/test_http_parser.py
```diff
@@ -117,6 +117,20 @@ def test_parse_headers(parser: Any) -> None:
...
+        + b"A" * 8192
+        + b": test\r"
+        + b"""\r
+\r
+"""
+    )
+    with pytest.raises(http_exceptions.LineTooLong):
```
webknjaz: It is the best practice to always use the `match=` argument with `pytest.raises()` (and other similar helpers like `pytest.warns()` or `pytest.deprecated()`). Could you add it?

tests/test_multipart.py
```diff
@@ -599,6 +599,18 @@ async def test_read_form(self, newline: Any) -> None:
             result = await obj.form()
         assert [("foo", "bar"), ("foo", "baz"), ("boo", "")] == result
 
+    async def test_read_form_invalid_utf8(self, newline: Any) -> None:
+        data = b"\xff%s--:--" % newline
```
webknjaz: plz separate this byte into a var too.

tests/test_multipart.py
```diff
@@ -599,6 +599,18 @@ async def test_read_form(self, newline: Any) -> None:
...
+        with pytest.raises(ValueError):
```
webknjaz: Same here: plz add a `match=` arg.

tests/test_multipart.py
```diff
@@ -599,6 +599,18 @@ async def test_read_form(self, newline: Any) -> None:
...
+            with Stream(data) as stream:
+                obj = aiohttp.BodyPartReader(
+                    BOUNDARY,
+                    {CONTENT_TYPE: "application/x-www-form-urlencoded"},
+                    stream,
+                    _newline=newline,
+                )
+                await obj.form()
```
webknjaz, line 615: What is actually supposed to raise a `ValueError`? This line? If yes, move the `pytest.raises()` CM here to isolate where we expect the exception to arise. Otherwise, the test will be misleading and will pass even if `aiohttp.BodyPartReader()` raises a `ValueError` not even calling `.form()`.

tests/test_http_parser.py
ret2libc: Yes! I saw it's done similarly in other tests as well.

Dreamsorcerer:
> > however it might be better to just reject the header?
> 
> This sounds like a good idea. I can't imagine anybody wanting to process an invalid header in their web application. But it'd also be reasonable to double-check with the RFCs what it allows/suggests/recommends, just in case.

I can refer you to the dozen issues complaining about the strict header parsing... :P

I suspect this will actually not reject if we enable lenient headers in the future: https://github.com/aio-libs/aiohttp/issues/5269#issuecomment-1215686989

Dreamsorcerer:
I think if we just update the test to catch either LineTooLong or InvalidHeaderToken, that will then ensure that we catch any regression caused by the lenient headers being enabled in future.

webknjaz:
> either LineTooLong or InvalidHeaderToken

But these should be two distinct cases. Tests must be kept simple.

Dreamsorcerer:
> > either LineTooLong or InvalidHeaderToken
> 
> But these should be two distinct cases. Tests must be kept simple.

Well, if lenient headers are enabled, then they should both cause LineTooLong, and the test can be updated then. We probably don't want to try and make the Python parser produce the same error, when it may change in llhttp parser later anyway.

Lots of things trip the llhttp parser, which don't trip the Python parser currently (which is why people are disabling extensions as a workaround).

tests/test_multipart.py
ret2libc: I haven't added a `match=` here because I now catch two exceptions: the one raised by the C parser and the one raised by the Python parser, which have different texts.

ret2libc:
I have just adjusted the tests for now and I didn't reject the header.

Dreamsorcerer:
Looks good to me. Just some black formatting issue (thought the CI automatically fixed those...).

Dreamsorcerer:
@ret2libc Could you complete the backport for us?

ret2libc:
@Dreamsorcerer tried to do that in https://github.com/aio-libs/aiohttp/pull/7099 .

PR #7048, opened 2022-10-26
align with OSI template

lsmith77:
## What do these changes do?

Align with the OSI template to hopefully get Github to detect the license properly:
https://twitter.com/Stof70/status/1585287286786330625

Will help anyone that wants to use automation to check license compatibility

## Checklist

N/A

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,10 +1,10 @@
-   Copyright aio-libs contributors.
+  Copyright 2014-2022 aio-libs contributors.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
 
-       http://www.apache.org/licenses/LICENSE-2.0
+     http://www.apache.org/licenses/LICENSE-2.0
 
    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


lsmith77:
then again the issue might be related to https://github.com/licensee/licensee/issues/534#issuecomment-1054040286, ie. reference to the license in the README

Dreamsorcerer:
The copyright year was removed deliberately, as it adds maintenance burden (and apparently, there are some users who think the entire library is unmaintained if the copyright year has not been updated).

If you can play with that licensee library and figure out exactly what the cause is, maybe we can do something else.

lsmith77:
yeah. I am trying to find out more about how Github determines the license and come back.

webknjaz:
> I am trying to find out more about how Github determines the license

It uses `licensee` to determine how similar the found licenses are to their canonical forms. It's a Ruby app. You can install it and call via CLI in project checkouts to learn how GH "sees" them.

LICENSE.txt
```diff
@@ -1,10 +1,10 @@
-   Copyright aio-libs contributors.
+  Copyright 2014-2022 aio-libs contributors.
```
webknjaz, line 1: The Linux Foundation recommends dropping the year or using just the start date. Updating the year is just busy work.

PR #7054, opened 2022-10-27
Third-Part/Other addition: RSocket

jell-o-fishi:
Third-Part/Other addition: RSocket

Supports aiohttp as a transport layer.


jell-o-fishi:
Is there anything i need to do to help merge this PR ?

webknjaz:
@Dreamsorcerer looks like the linting job started failing after https://github.com/aio-libs/aiohttp/commit/1337c452723629f6f70d5f114e49c4202c32d38c (probably due to not strict enough dependency constraints?)

webknjaz:
> probably due to not strict enough dependency constraints

The only difference (compared to the previous commit) I noticed in the CI logs is that the successful run had `multidict-6.0.2 yarl-1.8.1` in the env while the failing one got `multidict-6.0.3 yarl-1.8.2`.

Dreamsorcerer:
@webknjaz Yes, I did try to ping you elsewhere. The latest multidict release is causing some typing errors (and seems to get installed as part of the install self step regardless of the previous pinning).

The changes in multidict are just syntax changes, there shouldn't be any change in what mypy understands. I also don't get any errors locally, so I'm not really sure how to debug. Do you get any errors when running mypy on your machine? (I even created https://github.com/python/mypy/issues/14282 hoping one of the mypy devs might have an idea, as I'm completely lost).

webknjaz:
@Dreamsorcerer ah, I must've missed that. It's probably best to create dedicated tracking issues in such cases so it's not mixed up with other things that I might ignore. A separate issue would also mean a separate thread with notifications under a proper title in the inbox that would be harder to overlook.

As for debugging, I recommend https://github.com/marketplace/actions/debugging-with-tmate — it creates gets you an ssh connection into the actual runner, so you can inspect the env interactively. It often helps me track down the env differences, inject a `breakpoint()` right into various packages in `site-packages` etc.

I've also noticed that the Python version between those jobs is the same, so my last idea is broken cache invalidation. Both jobs restore their pip cache using the same key `pip-lint-4cde58ecca631ddf7a4f6a74e927540bff4eac5c92238c252e1b41757ea37b4b` so it's entirely possible that the resulting env differs from what one would get in a freshly populated venv.
Earlier this year, it became possible to remove caches via GH UI at https://github.com/aio-libs/aiohttp/actions/caches and will restart.

I'm going to restart the successful job to see if it starts breaking.

P.S. I haven't tried reproducing this locally.

webknjaz:
> I'm going to restart the successful job to see if it starts breaking.

Yep. It started failing. Now I'm going to wipe the specific cache entry to see if this happens in a fresh env.

webknjaz:
> Now I'm going to wipe the specific cache entry to see if this happens in a fresh env.

Nope. It wasn't caching...

webknjaz:
Locally, I'm experiencing different issues:
```console
$ mypy
tests/autobahn/test_autobahn.py:9: error: Module "python_on_whales" does not explicitly export attribute "DockerException"; implicit reexport disabled  [attr-defined]
tests/autobahn/test_autobahn.py:76: error: Item "str" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
tests/autobahn/test_autobahn.py:76: error: Item "Iterable[Tuple[str, bytes]]" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
aiohttp/helpers.py:121: error: unused "type: ignore" comment
aiohttp/helpers.py:886: error: Argument 1 to "union" of "set" has incompatible type "*Tuple[Mapping[Union[str, AppKey[Any]], Any], ...]"; expected "Iterable[<nothing>]"  [arg-type]
aiohttp/helpers.py:917: error: unused "type: ignore" comment
aiohttp/connector.py:1306: error: Incompatible return value type (got "BaseProtocol", expected "ResponseHandler")  [return-value]
aiohttp/client_reqrep.py:826: error: Incompatible types in assignment (expression has type "Union[str, URL]", variable has type "str")  [assignment]
aiohttp/web_ws.py:232: error: Need type annotation for "response_headers"  [var-annotated]
aiohttp/web_ws.py:233: error: Argument 1 to "CIMultiDict" has incompatible type "Dict[istr, str]"; expected "Union[Mapping[Union[str, istr], <nothing>], Dict[Union[str, istr], <nothing>], MultiMapping[<nothing>], Iterable[Tuple[Union[str, istr], <nothing>]]]"  [arg-type]
aiohttp/web_runner.py:171: error: Argument 2 to "create_unix_server" of "AbstractEventLoop" has incompatible type "Union[str, PathLike[str]]"; expected "Optional[str]"  [arg-type]
aiohttp/web_runner.py:268: error: unused "type: ignore" comment
aiohttp/web.py:314: error: Incompatible types in "await" (actual type "Union[Application, Awaitable[Application]]", expected type "Awaitable[Any]")  [misc]
tests/test_circular_imports.py:52: error: Module has no attribute "__path__"  [attr-defined]
examples/curl.py:33: error: unused "type: ignore" comment
aiohttp/test_utils.py:136: error: unused "type: ignore" comment
Found 16 errors in 10 files (checked 124 source files)
```
But I haven't updated the venv deps in a while so I'll do that now.

webknjaz:
> But I haven't updated the venv deps in a while so I'll do that now.

Aha! So it upgraded mypy from 0.910 to 0.982, and now I see this:

```console
$ mypy
aiohttp/web_ws.py:232: error: Need type annotation for "response_headers"  [var-annotated]
aiohttp/web_ws.py:233: error: Argument 1 to "CIMultiDict" has incompatible type "Dict[istr, str]"; expected "Union[Mapping[Union[str, istr], <nothing>], Dict[Union[str, istr], <nothing>], MultiMapping[<nothing>], Iterable[Tuple[Union[str, istr], <nothing>]]]"  [arg-type]
tests/autobahn/test_autobahn.py:9: error: Module "python_on_whales" does not explicitly export attribute "DockerException"; implicit reexport disabled  [attr-defined]
tests/autobahn/test_autobahn.py:76: error: Item "str" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
tests/autobahn/test_autobahn.py:76: error: Item "Iterable[Tuple[str, bytes]]" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
Found 5 errors in 2 files (checked 124 source files)
$ pyenv which mypy 
~/.pyenv/versions/aiohttp-pyenv-py3.9.0/bin/mypy
```

Note that it's Python 3.9.0 (and not 3.9.15 like in the failing CI job).

webknjaz:
I've also updated the editable installation in the env and got:
```console
$ mypy
tests/autobahn/test_autobahn.py:9: error: Module "python_on_whales" does not explicitly export attribute "DockerException"; implicit reexport disabled  [attr-defined]
tests/autobahn/test_autobahn.py:76: error: Item "str" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
tests/autobahn/test_autobahn.py:76: error: Item "Iterable[Tuple[str, bytes]]" of "Union[Container, str, Iterable[Tuple[str, bytes]]]" has no attribute "stop"  [union-attr]
aiohttp/web_ws.py:232: error: Need type annotation for "response_headers"  [var-annotated]
aiohttp/web_ws.py:233: error: Argument 1 to "CIMultiDict" has incompatible type "Dict[istr, str]"; expected "Union[Mapping[Union[str, istr], <nothing>], Dict[Union[str, istr], <nothing>], MultiMapping[<nothing>], Iterable[Tuple[Union[str, istr], <nothing>]]]"  [arg-type]
Found 5 errors in 2 files (checked 124 source files)
```

This is closer to what's in the CI.

webknjaz:
Passing different `--python-version` values doesn't help.

Dreamsorcerer:
It's weird, because I can't reproduce with mypy 0.982 or 0.991.

Dreamsorcerer:
Although the only one in your output, which is also in CI is:
`aiohttp/web_ws.py:232: error: Need type annotation for "response_headers"  [var-annotated]`

So, still rather weird...

Dreamsorcerer:
I suspect that python-on-whales isn't installed in CI or something, so those errors don't appear. It has an `ignore_missing_imports` in the config. So, ignore those ones, I can look at that another time.

webknjaz:
@Dreamsorcerer downgrading multidict-6.0.3 to multidict-6.0.2 makes errors disappear, except for the autobahn ones.

Dreamsorcerer:
Yep, those errors are definitely from multidict upgrade, I just don't get them on my machine. I don't think anything changed semantically though, it was just upgrading the syntax, so I'm not sure why there are errors now.

As mentioned above, ignore the autobahn ones (or uninstall python-on-whales and it'll probably disappear).

Dreamsorcerer:
Aha, it doesn't happen when running mypy with Python 3.10+: https://github.com/aio-libs/aiohttp/actions/runs/3734373324/jobs/6336357081

So, some of the changes I was lead to believe were safe to update in a stubs file, clearly weren't...
I suspect it's the change from typing to collections.abc for some of the imports.

webknjaz:
@Dreamsorcerer

> So, some of the changes I was lead to believe were safe to update in a stubs file, clearly weren't...

This is why, it's important to test types against multiple Python versions, at least the lowest and the highest. Here's an example of how I implemented it in Cheroot, some time ago: https://github.com/cherrypy/cheroot/blob/4aa647d/.pre-commit-config.yaml#L120-L182.

PR #7056, opened 2022-10-28
Added a configuration flag for enable request task handler cancelling when client connection closing.

mosquito:
## Related to #6719 #6727. Added a configuration flag for enable request task  handler cancelling when client connection closing.

After changes in version 3.8.3, there is no longer any way to enable this behaviour. In our services, we want to handle protocol-level errors, for example for canceling the execution of a heavy query in the DBMS if the user's connection is broken.

Now I created this PR in order to discuss my solution, of course if I did everything well I will add tests changelog, etc.

## I guess this breakdown can be solved using the configuration flag that is passed to the Server instance.

Of course `AppRunner` and `SiteRunner` can pass this through `**kwargs` too.

## Checklist


tzoiker:
I'd rename `cancel_when_connection_lost` -> `cancel_handler_on_connection_lost`

Dreamsorcerer:
Seems simple enough. First though, can we get some tests to validate this behaviour? Then we can be sure not to break it again in future.

Second, how do I use this as a user? Can we pass the argument to `run_app()`? If so, please also update the docs to include the new argument.

mosquito:
@Dreamsorcerer I started adding information to the documentation, and found [this example](https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks) this made me think that the current default behaviour is unsafe, and that's what I mean.

I imagined a simpler example that worked well before the change where tasks weren't cancelled, and didn't work so well now.

```python
from aioredis import Redis
from aiohttp import web


async def publish(request: web.Request) -> web.Response:
    redis = await Redis(host='localhost', port=6379)
    await redis.lpush("events", await request.read())
    return web.json_response(data={"status": "success"})


async def subscribe(_: web.Request) -> web.Response:
    redis = await Redis(host='localhost', port=6379)
    _, event = await redis.brpop("events")
    return web.json_response(data={"event": event.decode()})


app = web.Application()
app.router.add_route("POST", "/pub", publish)
app.router.add_route("GET", "/sub", subscribe)
web.run_app(app)
```

In the previous version, when closing a client connection, an exception was thrown that canceled the connection to the redis too. Now, after the closing the client connection the task is not cancelled and waiting continuously **until an event arrives**, and will fail only when you try to write to a closed connection.

Thus, in this example, you can arrange a denial of service attack, simply create connections and close them without waiting for completion.

From a user experience point of view, I understand that the default when tasks are not canceled is very intuitive and simple, but from a security point of view, I am scared.

Dreamsorcerer:
> @Dreamsorcerer I started adding information to the documentation, and found [this example](https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks) this made me think that the current default behaviour is unsafe, and that's what I mean.

I don't see how that's related. There is no view/handler in that example.

> I imagined a simpler example that worked well before the change where tasks weren't cancelled, and didn't work so well now.

Creating a Redis instance and waiting on an event that might never come, in a handler, just seems crazy. I've not seen any evidence that people are doing anything like this. I think we are still going to have more issues with the old behaviour than the new one (remember that 1 reason we changed it is because every other framework also does this, and they don't seem to be having major issues).

> Thus, in this example, you can arrange a denial of service attack, simply create connections and close them without waiting for completion.

DoS is exactly why I'd suggest advanced users look at enabling this option again. It probably won't make a huge difference, but atleast it means an attacker must leave connections open to DoS the server, rather than finding a slow endpoint and briefly opening a connection there many times.

> From a user experience point of view, I understand that the default when tasks are not canceled is very intuitive and simple, but from a security point of view, I am scared.

As mentioned, it seems every other framework already does this, and there aren't security warnings in Django, Flask etc. because of it. Ultimately, if someone really wants to DoS the server, this will probably only make it slightly harder, and you'll be best off with Cloudflare protection or something if there is a serious threat.

mosquito:
>Creating a Redis instance and waiting on an event that might never come, in a handler, just seems crazy.

The redis protocol is designed so that it can only process one command at a time, so one way or another you'll have to do it like this, in this example it's just done explicitly, without using connection pooling.

> I've not seen any evidence that people are doing anything like this. I think we are still going to have more issues with the old behaviour than the new one (remember that 1 reason we changed it is because every other framework also does this, and they don't seem to be having major issues).

No, after applying my patch, there is no problem, I already checked.

> DoS is exactly why I'd suggest advanced users look at enabling this option again. It probably won't make a huge difference, but atleast it means an attacker must leave connections open to DoS the server, rather than finding a slow endpoint and briefly opening a connection there many times.

I already sent the email to @asvetlov with a working exploit.

> As mentioned, it seems every other framework already does this, and there aren't security warnings in Django, Flask etc. because of it.

Django and especially Flask are synchronous frameworks, and the task of stopping a workflow is the task of gunicorn or another program that implements a bridge between HTTP and WSGI.

> Ultimately, if someone really wants to DoS the server, this will probably only make it slightly harder, and you'll be best off with Cloudflare protection or something if there is a serious threat.

Cloudflare will not protect against this. I think you are confusing DDoS and DoS.

It seems to me that you did not take into account all the negative consequences of disabling this early, especially in a ~minor~ patch release.

Dreamsorcerer:
> > Creating a Redis instance and waiting on an event that might never come, in a handler, just seems crazy.
> 
> The redis protocol is designed so that it can only process one command at a time, so one way or another you'll have to do it like this, in this example it's just done explicitly, without using connection pooling.

Regardless (still inefficient), it's the waiting on an event that may never happen, which is the main issue. That's not code I've seen in realistic applications, there is either some guarantee that an event will happen or a timeout, I've never seen anything that blindly blocks forever in a handler, and most developers would realise that, whereas many don't realise that a handler can be cancelled midway.

> > I've not seen any evidence that people are doing anything like this. I think we are still going to have more issues with the old behaviour than the new one (remember that 1 reason we changed it is because every other framework also does this, and they don't seem to be having major issues).
> 
> No, after applying my patch, there is no problem, I already checked.

I meant more problems in general, not your specific example. So, unless you've tested every developer's code in the world...

> I already sent the email to @asvetlov with a working exploit.

You'd be better off sending to @webknjaz. But, it's pretty trivial to come up with an example.

> > As mentioned, it seems every other framework already does this, and there aren't security warnings in Django, Flask etc. because of it.
> 
> Django and especially Flask are synchronous frameworks, and the task of stopping a workflow is the task of gunicorn or another program that implements a bridge between HTTP and WSGI.

Yes, and a quick search suggests that they behave exactly the same as aiohttp does today. They only error when trying to write to the socket (e.g. https://stackoverflow.com/questions/4142012/how-to-find-the-socket-connection-state-in-c).

Can you find a single framework which implements the cancelling behaviour we had before?
The fact that the feature was unique to aiohttp, is the reason it was causing problems for users.

> Cloudflare will not protect against this. I think you are confusing DDoS and DoS.

No, I'm just considering that any WAF or similar that can stop an attacker hitting a slow endpoint a large number of times is going to be more effective. If an attacker has resources then they can go full DDoS and it wouldn't make much difference, if they don't have the resources then they'd probably still struggle to take down the server without auto-cancelling (unless you've actually created handlers which block forever on an event that never occurs, in which case you need to rethink your code...).

> It seems to me that you did not take into account all the negative consequences of disabling this early, especially in a ~minor~ patch release.

I considered it, but the change was already made when I was first getting involved. Unless you can convince @asvetlov or @webknjaz that the default behaviour should be changed back (or provide evidence that the auto-cancelling is actually normal for most frameworks), I'll be leaving it as is.


Anyway, I'll take a look through the new changes in a bit (would be preferable to not force-push, it messes up the diff to look at), and we can hopefully get this feature back in, whether default or not.

docs/web_reference.rst
```diff
@@ -2904,6 +2905,11 @@ Utilities
                            this flag when being created. This option is not
                            supported on Windows.
 
+   :param bool cancel_handler_on_connection_lost: tells the runner whether to
+                                                  cancel the execution of the
+                                                  handler task if the client
+                                                  connection has been closed.
```
Dreamsorcerer: I think we probably want to add a section to the advanced docs and link to it here. That section should then highlight the reasons to enable it (save resources, DoS) and highlight the tools/techniques needed to write safe code.

I can probably do it in a separate PR, but if you want to get started that would be great. Looks like Github is starving us of workers right now, so I can't get the tests to run just yet anyway.

tests/test_web_server.py
```diff
@@ -5,6 +5,7 @@
 
 import pytest
 
+import aiohttp
 from aiohttp import client, helpers, web
```
Dreamsorcerer: I'd just add `ClientSession` to this import.

tests/test_web_server.py
```diff
@@ -207,3 +207,53 @@ async def handler(request):
...
+        assert event.is_set(), "Request handler hasn't been cancelled"
+    finally:
+        await asyncio.gather(runner.shutdown(), site.stop())
+
+
+async def test_cancel_handler_on_connection_lost_flag_on_runner() -> None:
+    runner = web.AppRunner(web.Application(), cancel_handler_on_connection_lost=True)
+    await runner.setup()
+    assert runner.server.cancel_handler_on_connection_lost, "Flag was not propagated"
+    await runner.shutdown()
```
Dreamsorcerer: This appears to already be asserted in the above test. Not sure if there's a reason for this second test?

tests/test_web_server.py
```diff
@@ -207,3 +207,53 @@ async def handler(request):
...
+                await asyncio.wait_for(request, timeout=0.1)
+
+    try:
+        assert (
+            runner.server.cancel_handler_on_connection_lost
+        ), "Flag was not propagated"
+        await client_request_maker()
+
+        await asyncio.wait_for(event.wait(), timeout=1)
+        assert event.is_set(), "Request handler hasn't been cancelled"
```
Dreamsorcerer, line 248: If you could add a negative test as well, which verifies the behaviour when the setting is `False`, that would be perfect.

Dreamsorcerer (approved):

tzoiker:
> I considered it, but the change was already made when I was first getting involved. Unless you can convince @asvetlov or @webknjaz that the default behaviour should be changed back (or provide evidence that the auto-cancelling is actually normal for most frameworks), I'll be leaving it as is.

I agree with the argumentation regarding the user expectations due to the other frameworks. The only problem with it is that this behaviour was changed with the patch release. Probably, reverting it back to cancellation on disconnect by default would cause even more damage now.

However, I suggest considering it for the next major release (4.0) if there'll be more evidence in favour.

mosquito:
> Regardless (still inefficient), it's the waiting on an event that may never happen, which is the main issue. That's not code I've seen in realistic applications, there is either some guarantee that an event will happen or a timeout, I've never seen anything that blindly blocks forever in a handler, and most developers would realise that, whereas many don't realise that a handler can be cancelled midway.

Regardless but http protocol has been designed similar. This code will be completely inefficient too:

```python


async def proxy_handler(request: web.Request) -> web.Response:
    session: ClientSession = request.app['client']

    async with session.get(f"/search?q={request.query.get('q')}") as response:
        # This code will wait for response but client connection already closed.
        # Of course the safe way is cancel this task after that.
        payload = await response.read()
    return web.Response(status=response.status, body=payload)


app = web.Application()
app.router.add_route("*", "/", proxy_handler)
app['client'] = ClientSession(base_url="https://stackoverflow.com/")
web.run_app(app)
```

This still inefficient code that you may not have seen in practice before will work exactly the same. So it will receive the entire page, although the client is no longer waiting for this. In the above example, you will quickly run into a TCPConnector limitation, and your service will no longer serve "good" clients (so thats a classical DoS). Worst of all, you will be downloading kilobytes over the network, and the attacker will be limited to transmitting a few lines of headers.

The HTTP 1.x protocol is not much different from the redis protocol, it is exactly the same text protocol that cannot asynchronously interact in one connection with many requests at once. You will exhaust the connection pool through the fault of the client that mimic a very bad connection (close the connection after headers sent each request). So the CloudFlare wount guard from it on 100%.

docs/web_reference.rst
mosquito: I still hope that my facts will be taken into account and we will change the default value to `True`.

docs/web_reference.rst
```diff
@@ -2904,6 +2905,16 @@ Utilities
...
+
+   .. warning::
+
+      Always set the ``cancel_handler_on_connection_lost`` option to ``True``
```
mosquito: It looks a little bit harsh, but it really is.

Dreamsorcerer:
> You will exhaust the connection pool through the fault of the client that mimic a very bad connection (close the connection after headers sent each request).

Yes, but you can still do that by keeping the connections alive. Even a single machine may be able to exhaust the connections of the average server (as most people don't change the default connection limits).

I agree that any developer deploying a high-performance service should look at enabling this option, but there also needs to be an understanding of how to write safe asyncio code when this option is enabled, and we've already seen that many developers simply don't know how to do that and don't understand the cancelling behaviour. Making it opt-in ensures they've looked at the documentation and understood how things work. Again, unless you can show that other frameworks are doing this as standard, I don't think anyone here will change the default behaviour back.

mosquito:
@Dreamsorcerer @asvetlov let me sum it up a bit, I think that aiohttp was incredibly close to the protocol, which is its cardinal difference from the competition, and that was great. It was possible to do TLS services using only aiohttp, however, what is being discussed now narrows the scope, and now it is better for us not to do anything without Cloudflare and WAF.

The right way, in my opinion, would be to make it safe by default without relying on CloudFlare. Explaining to users why it's done this way, and how to disable this behavior, and the most important thing that follows (any script kiddie will make you a DoS), it would make the Internet a little more secure.

It is also worth explain the user about `try: ... finally: ...` about `aiojobs`, about manual task management, except from the documentation, he is unlikely to find out about it. I understand the argument - "others have done it another one". So let me go and find out from them how protected they are from attacks like this? Maybe we need to fix it there and not break it here? I don't mind doing this in principle, but I should start here.

mosquito:
> when this option is enabled, and we've already seen that many developers simply don't know how to do that and don't understand the cancelling behaviour.

There have been several support requests among the thousands of people who use aiohttp. At the same time, earlier the user did not have the ability to manage this, and even more so the user does not know what it threatens.

It seems to me that the default behavior of `True` and explaining what exactly the user is doing wrong is better than making insecure thousands of services that already work and are already written as they are, when those migrate from version 3.8.1.

Tinche:
I think this discussion has gotten a little too heated for its own good.

FWIW I think Uvicorn's default behavior is to cancel on disconnect.

mosquito:
> FWIW I think Uvicorn's default behavior is to cancel on disconnect.

This code just printing `Done` after sleep is ending despite to client disconnections:

```python
import asyncio
from fastapi import FastAPI

app = FastAPI()

@app.get("/")
async def root():
    try:
        await asyncio.sleep(5)
        print("Done")
        print("Cancelled")
        raise
    return {}
```

So attacker will make DoS when `uvicorn` limiting concurrency:

```
$ uvicorn --limit-concurrency 2 fastapi_example:app
INFO:     Started server process [19554]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
WARNING:  Exceeded concurrency limit.
INFO:     127.0.0.1:53861 - "GET / HTTP/1.1" 503 Service Unavailable
WARNING:  Exceeded concurrency limit.
INFO:     127.0.0.1:53863 - "GET / HTTP/1.1" 503 Service Unavailable
Done
Done
```

The first 2 clients was terminated before response. After that service just respond 503 while **useless tasks** is alive.

I again urge you to think about security by default, not to repeat other people's problems.

mosquito:
@Dreamsorcerer I took the liberty of changing the default behavior and also described the reasons for this change.

H--o-l:
@mosquito On a neutral point of view If you re-change the aiohttp default behavior then you need to revert this documentation deletion:
https://github.com/aio-libs/aiohttp/pull/4080/files#diff-af5cbca68a75093b494497c857162462f30fc2f981c4b1104c57bd6a99d8db6cL23

From a personal point of view, I think this feature should be opt-in, not opt-out, so I'm against the current version of the PR. 

I understand that opt-out is the most secure behavior against DDoS but it's the worse behavior for data consistency in database. I think most users want consistency first, and think DDoS after, as an improvement.

Also, this feature is not compatible with the aiohttp middleware feature. The documentation should say clearly that if this feature is enabled, you should not expect your middlewares to behave like you think they wil.

mosquito:
@H--o-l

> I understand that opt-out is the most secure behavior against DDoS but it's the worse behavior for data consistency in database.

If your database can break data consistency by unexpectedly disconnecting from it, it's no ACID and probably not production ready solution.

> Also, this feature is not compatible with the aiohttp middleware feature. The documentation should say clearly that if this feature is enabled, you should not expect your middlewares to behave like you think they will.

If the middleware contains `try: ... finally: ...` block then the behaviour will be as expected.

Unexpected terminations of TCP connections will still occur, including between the HTTP server and the DBMS server, and you still have to take care of error handling. CancelledError can come not only from the fact that the client has fallen off, but also if the control program sends a SIGINT (oomd for example). Just because you're covering up a problem doesn't mean it doesn't exist.

mosquito:
@asvetlov @webknjaz I don't know if I need to [roll it back](https://github.com/aio-libs/aiohttp/pull/4080/files#diff-af5cbca68a75093b494497c857162462f30fc2f981c4b1104c57bd6a99d8db6cL23) or not

H--o-l:
> @H--o-l
> 
> > I understand that opt-out is the most secure behavior against DDoS but it's the worse behavior for data consistency in database.
> 
> If your database can break data consistency by unexpectedly disconnecting from it, it's no ACID and probably not production ready solution.
> 

```py
@app.get("/invoice/paid")
async def user_paid_his_invoice():
    external_service.invoice.paid()
    # CancelledError
    database.save("unlock user") # Due to CancelledError my system is not consistent,
                                 # user has paid his invoice but is not unlocked
```


> > Also, this feature is not compatible with the aiohttp middleware feature. The documentation should say clearly that if this feature is enabled, you should not expect your middlewares to behave like you think they will.
> 
> If the middleware contains `try: ... finally: ...` block then the behaviour will be as expected.
> 

```py
@app.get("/something_that_will_be_cancelled")
async def foo():
    try:
       shield:
            sleep(10)
            raise Exception("toto")
    finally:  # executed on the CancelledError
        print("the exception toto occured in the shield block after the finally block execution"
               "so I can't get it. That programming pattern is crap")
```

> Unexpected terminations of TCP connections will still occur, including between the HTTP server and the DBMS server, and you still have to take care of error handling. CancelledError can come not only from the fact that the client has fallen off, but also if the control program sends a SIGINT (oomd for example). Just because you're covering up a problem doesn't mean it doesn't exist.

I didn't understand.

tzoiker:
> @app.get("/invoice/paid")
> async def user_paid_his_invoice():
>     external_service.invoice.paid()
>     # CancelledError
>     database.save("unlock user") # Due to CancelledError my system is not consistent,
>                                  # user has paid his invoice but is not unlocked

Anything apart from the user disconnection can happen between `external_service.invoice.paid()` and `database.save("unlock user")`:
- Server dies
- Database connection breaks
- etc

If a service is designed in the way that the consistency is guaranteed only under the 100% stability of every component, then, something is done fundamentally wrong.

H--o-l:
> 
> Anything apart from the user disconnection can happen between `external_service.invoice.paid()` and `database.save("unlock user")`:
> 
> * Server dies
> * Database connection breaks
> * etc
> 
> If a service is designed in the way that the consistency is guaranteed only under the 100% stability of every component, then, something is done intrinsicly wrong.

Designing a system for covering 99% of cases is way easier than 100% of case. 
An asteroid can hit your system at the wrong point in time, does it mean your system should include asteroid protection?

Here CancelledError by user disconnection represents 99.9% of what can go wrong between both calls. Having CancelledError in the middle of the execution increases the risk of something going wrong by one or two orders of magnitude.

(But it depends on the use case of course. The question is what is the most natural use case for aiohttp users, and does a satisfying solution exists).

mosquito:
@H--o-l The asteroid fell to the planet a couple of times, but client disconnections happen all the time. You are manipulating facts. We are not solving a far-fetched problem, but a problem that makes any service written in aiohttp vulnerable by default.

tzoiker:
> An asteroid can hit your system at the wrong point in time, does it mean your system should include asteroid protection?

My point is that something wrong shall definitely happen between those calls at some point. If the system is designed in the way that it can not be handled automatically (e.g., by adding idempotence), breaks the flow and requires manual intrusions, then, it is a bad design, independent on the framework that is used. The more servers (or VM containers), users, components in the system you have, the sooner and more frequently it shall happen.

But as I mentioned before, I am up for keeping the default as it is now, and leave this discussion for the next major release in order to not making it even more messy and painful for everyone :)

H--o-l:
> @H--o-l The asteroid fell to the planet a couple of times, but client disconnections happen all the time. You are manipulating facts. We are not solving a far-fetched problem, but a problem that makes any service written in aiohttp vulnerable by default.

With that sentence I was only answering that other sentence :
```
If a service is designed in the way that the consistency is guaranteed only under the 100% stability
of every component, then, something is done fundamentally wrong.
```
If you are suspicious of every component of your architecture, then you are programming only in a statistical ways, because there is always something that can go wrong.
Developers working on airplanes or rockets or trains work like that, they suspect everything. It's not the same job as web development.

> > An asteroid can hit your system at the wrong point in time, does it mean your system should include asteroid protection?
> 
> My point is that something wrong shall definitely happen between those calls at some point. If the system is designed in the way that it can not be handled automatically (e.g., by adding idempotence), breaks the flow and requires manual intrusions, then, it is a bad design, independent on the framework that is used. The more servers, users, components in the system you have, the sooner and more frequently it shall happen.
> 

I understand the scaling argument. If it can happen it will at some point, sooner or later. But that doesn't mean everyone should think about these types of issues when they are only starting their company or their web server for the first time.

Having CancelledError on user disconnection requires you to think about these type of issues a lot sooner than you will do otherwise (and maybe you will never have to).

mosquito:
> But that doesn't mean everyone should think about these types of issues when they are only starting their company or their web server for the first time.

You pick a technology, you start coding on it, you launch a company, and then your competitor makes you a DoS using one python-script running on his laptop, you decide that the technology is bad, and it was a bad idea to choose it. Win-Win!

Another scenario, you encounter this problem during development, do `pip install aiojobs` and after 30 minutes, the problem is solved.

tzoiker:
> If you are suspicious of every component of your architecture, then you are programming only in a statistical ways, because there is always something that can go wrong.

Exactly, and you should always keep it in mind when designing the system. If a service just fetches or atomically updates data in the database, then, it is not a problem you are required to care, of course.

> Developers working on airplanes or rockets or trains work like that, they suspect everything. It's not the same job as web development.

IMHO, services that involve financial transactions (as in your example) impose stronger requirements comparing to the other ones.

H--o-l:
> > But that doesn't mean everyone should think about these types of issues when they are only starting their company or their web server for the first time.
> 
> You pick a technology, you start coding on it, you launch a company, and then your competitor makes you a DoS using one python-script running on his laptop, you decide that the technology is bad, and it was a bad idea to choose it. Win-Win!
> 

In the DDoS case, your application is down for a few hours.
In the "CancelledError poping in the middle of your sequential code" case, you have forgotten if your users paid their invoices or not, and you have no way to find out.

I hardly see how, from your users' point of view, the first is worse than the second. (hopefully it's _just_ invoices).

> Another scenario, you encounter this problem during development, do `pip install aiojobs` and after 30 minutes, the problem is solved.

The CancelledError issue is hardly noticeable during development. The exception was caught by aiohttp and made completely silent. Your usual exception tracking or logging tool integrations would not catch it at all. And I'm ready to bet that 99% of developers don't have tests to test real-life "client disconnection" cases. 

We ourselves noticed it maybe 2 years after live production when we stumbled upon a half-filled database by chance.
The simple _idea_ that aiohttp could behave like that is a lot less natural and known than DDoS principle, and thus a lot harder to understand and debug for a usual developers.

But It is an awesome idea that aiohttp documentation talks about DDoS, explains the thing, and shows its options to deal with it. 
Interested developers can then _choose_ to invest time on the subject, they will not be _forced_ to investigate why the hell their databases are completely wrong, many months after said databases were filled.

mosquito:
> In the DDoS case, your application is down for a few hours.

You are confusing DoS and DDoS, they are fundamentally different things.

H--o-l:
> > In the DDoS case, your application is down for a few hours.
> 
> You are confusing DoS and DDoS, they are fundamentally different things.

Neat, it means that I need good documentation to explain it to me _if_ I'm interested. 
At the same time, I need my code to be executed, and if it's not, I need an exception to be raised and not caught.

mosquito:
@H--o-l You are ready to post a link to your production project here, I will run my exploit to demonstrate the vulnerability, and yes, by the way, share a link to your bugs bounty at the same time. There will be no large traffic, as is usually the case with DDoS, this is a classic vulnerability of loss of control over resources, 100kb/s is enough for me. Perhaps this will be enough for you to reconsider your position on this issue.

H--o-l:
> @H--o-l You are ready to post a link to your production project here, I will run my exploit to demonstrate the vulnerability, and yes, by the way, share a link to your bugs bounty at the same time. There will be no large traffic, as is usually the case with DDoS, this is a classic vulnerability of loss of control over resources, 100kb/s is enough for me. Perhaps this will be enough for you to reconsider your position on this issue.

I understand your point of view since the beginning of the conversation, there is no need to threaten me.

If you think the risk is high, and aiohttp users should be warned, it's kind of you. I thank you for making me aware of the risk that I'm facing. I'm sure your warning can be included in the aiohttp documentation at a favorable place so every aiohttp users can also benefice from your advice.

At the same time, the default behavior, that most of us expect I think, is that the code scheduled to be executed is really executed, not dumped in the middle of its execution, without warning, and with consequences that can be seen only month or year after the issue happens. 
The worse for me in this is the timeline. At least a DoS attack I'm aware of it soon, and it's temporary. Client disconnections on the contrary slowly and silently accumulate, with various hard-to-debug consequences, for months.

Coming back to the `shield + try .. finally` solution, I showed you that I cannot catch exceptions in that case. 
If your PR improves the aiohttp DoS behavior I'd be happy to use it, but that  `shield + try .. finally` issue is a deal breaker for me. Could your PR solution account for it?

mosquito:
@H--o-l Basically you're suggesting that you don't lock the car when you leave it, and only if it gets stolen, start locking it up, am I right?

H--o-l:
> @H--o-l Basically you're suggesting that you don't lock the car when you leave it, and only if it gets stolen, start locking it up, am I right?

I'm not against locking the car, but here the car locking mechanism, thanks to a weird and hard-to-track side effect, also slowly accumulates dirt in the car engine that, possibly, will be impossible to remove.
Besides your analogy is wrong, because, in the case of a DoS, the car is not stolen, it's temporarily unusable.
Between a temporary unusable car and an engine that slowly gets full of dirt without any warning before it's too late, the choice is hard.

I'm not saying my choice is better than yours, only that your solution is not great for a few reasons:
- hard to understand for a developer without a lot of also hard-to-understand explanations
- hard to debug 
- doesn't account for aiohttp middleware tracking integrations.

So in its current state, I think the feature should be opt-in, or maybe "experimental" 

Ideas to improve the feature :

- Don't catch CancelledError silently but instead raise them like any other errors. A user-defined middleware can then catch them and sort/filter them.
- Find a way to make `try .. finally` block works even when the code is protected against cancellation. Maybe the cancellation mechanism shouldn't use native `task.cancel()`? If not maybe the documentation can talk about that issue? 
- Maybe the subject should be discussed upstream at the Python project directly? Indeed catching exceptions from shielded code in the outside code is a native Python issue.
- Enable or disable the feature per route? So it can be fined tune.
- Write more documentation about it that clearly states what is at stake.

Dreamsorcerer:
Jeez, we are *not* going to change the default in *this* PR.
Please open a new issue to discuss changing the default, tag asvetlov and he can weigh in on the discussion when he's back around.

mosquito:
@Dreamsorcerer fine, the now default is `False`

mosquito:
@H--o-l Is it really not enough in your case to make such a middleware:

```python
@middleware
async def shield_request_middleware(request, handler):
    task = asyncio.create_task(handler(request))
    return await asyncio.shield(task)
```

The behavior it provides is identical to what is being just right now, your task will run until it tries to read something from request or return a response, at which point it will throw a `ConnectionResetError`.

The protocol object no longer exists (the task property is set to `None`), and no one except the event-loop controls its execution, just like it works now. But this middleware should be the very first, it's not good.

If you look at it more broadly, then it would be enough to wrap the task with a decorator that would apply `shield` to the handler's coroutine.

H--o-l:
@mosquito hey!

I tried it once, and [other people on the internet have to](https://aiohttp-middlewares.readthedocs.io/en/latest/_modules/aiohttp_middlewares/shield.html). Right now I don't remember why it didn't work for me but more generally I think:
- middleware shouldn't depend on each other, and especially the chain of exceptions shouldn't be broken like `asyncio.shield()` do I think.
-  `asyncio.shield()` also breaks the chain of Python contextvars so it's yet another issue.
- If it's done like that, I think the official aiohttp documentation should explain that workaround, and aiohttp tests should test it, in order to be sure the workaround is maintained in the long run, and to reduce the number of issues opened on the subject.

I also have in mind that there is probably something to be done to improve the behavior of aiohttp against DoS without using `task.cancel()` which brings a lot of hard-to-workaround side effects. Maybe aiohttp could raise a more conventional exception? Or at least it could be activated/deactivated route by route?

ikrivosheev:
Any updates?) In my case, a huge number of tasks are accumulating, inside which I do not understand that there is no one to answer.

Dreamsorcerer:
> Any updates?) In my case, a huge number of tasks are accumulating, inside which I do not understand that there is no one to answer.

Waiting till I have a full afternoon available to look at this properly. Though I don't know if it's related to your particular issue or not.

Dreamsorcerer:
OK, I've fixed the tests so they fail correctly now (the 2nd test seemed to even pass with both options).

I've renamed the parameter to `handler_cancellation` to be a little more concise (and being a little more vague should encourage users to check the documentation, I'd rather they didn't _think_ they know what this does without looking at the documentation).

I've also stripped most of the documentation changes, I'm going to do this in more detail in another PR.

I also removed the Gunicorn env var. We tend to avoid using env vars in the library, with that being something for the application to handle. It also seems to me that the person deploying an app would be the one deciding whether the option is enabled or not, when it should be the developer who wrote the app.
    So, let's come back to this specific case in another PR which can focus exclusively on how to enable this in a Gunicorn deployment.

PR #7131, opened 2022-12-16
Support using netrc for non-proxy HTTP credentials

yuvipanda:
## What do these changes do?

Currently, setting `trust_env=True` in `ClientSession` will only use the `.netrc`
file if present for authentication to proxies. With this change, the credentials will
also be used for regular HTTP hosts, and Basic Auth will be sent if an appropriate
entry is present in the user's `.netrc` file.

YES! Currently, their .netrc file will be ignored when not making proxy requests. This
will start sending any credentials in netrc files to hosts.

The behavior of this PR seems to have possibly been the suggested behavior when
netrc support was initially added, based on this comment: https://github.com/aio-libs/aiohttp/pull/2584#issuecomment-348935249.

The behavior with this PR applied also matches requests and other libraries. In fact, this
was a deep dive output after trying to figure out why accessing NASA open data with requests
works but aiohttp does not :D  (https://github.com/nsidc/earthaccess/issues/188 has the trail crumbs).

If this is considered a breaking change, I'm happy to instead introduce a *new* param that
toggles this behavior instead. However, `trust_env` is what requests and other libraries use,
so may be ok.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/2581
https://github.com/aio-libs/aiohttp/pull/2584#issuecomment-348935249
(outside aiohttp) https://github.com/nsidc/earthaccess/issues/188

## Checklist

### DIFF
--- a/CHANGES/7131.feature
+++ b/CHANGES/7131.feature
@@ -1,1 +1,1 @@
-Support reading HTTP Basic Auth credentials from .netrc file when ``trust_env`` is ``True``.
+Added support for using Basic Auth credentials from :file:`.netrc` file when making HTTP requests with the :py:class:`~aiohttp.ClientSession` ``trust_env`` argument is set to ``True`` -- by :user:`yuvipanda`.
--- a/CHANGES/7131.feature
+++ b/CHANGES/7131.feature
@@ -1,0 +1,1 @@
+Support reading HTTP Basic Auth credentials from .netrc file when ``trust_env`` is ``True``.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -436,8 +436,7 @@ class ClientRequest:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
-        URL_HAS_A_HOST = self.url.host is not None
-        if auth is None and trust_env and URL_HAS_A_HOST:
+        if auth is None and trust_env and self.url.host is not None:
             netrc_obj = netrc_from_env()
             with contextlib.suppress(LookupError):
                 auth = basicauth_from_netrc(netrc_obj, self.url.host)
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -44,6 +44,7 @@ from .helpers import (
     BasicAuth,
     HeadersMixin,
     TimerNoop,
+    basicauth_from_netrc,
     is_expected_content_type,
     netrc_from_env,
     noop,
@@ -440,11 +441,8 @@ class ClientRequest:
 
                 netrc_obj = netrc_from_env()
                 if netrc_obj:
-                    auth_from_netrc = netrc_obj.authenticators(self.url.host)
-                    if auth_from_netrc:
-                        *logins, password = auth_from_netrc
-                        login = logins[0] if logins[0] else logins[-1]
-                        auth = BasicAuth(cast(str, login), cast(str, password))
+                    auth = basicauth_from_netrc(netrc_obj, self.url.host)
+                    if auth:
                         self.headers[hdrs.AUTHORIZATION] = auth.encode()
 
             return
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -1,5 +1,6 @@
 import asyncio
 import codecs
+import contextlib
 import dataclasses
 import functools
 import io
@@ -437,7 +438,7 @@ class ClientRequest:
             auth = self.auth
         if auth is None and trust_env:
             netrc_obj = netrc_from_env()
-            if netrc_obj:
+            with contextlib.suppress(LookupError):
                 auth = basicauth_from_netrc(netrc_obj, self.url.host)
         if auth is None:
             return
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -45,6 +45,7 @@ from .helpers import (
     HeadersMixin,
     TimerNoop,
     is_expected_content_type,
+    netrc_from_env,
     noop,
     parse_mimetype,
     reify,
@@ -433,6 +434,17 @@ class ClientRequest:
         if auth is None:
             auth = self.auth
         if auth is None:
+            # If no auth is specified, we read from netrc.
+            # FIXME: This should only be the case if a `trust_env` flag is passed
+            netrc_obj = netrc_from_env()
+            if netrc_obj:
+                auth_from_netrc = netrc_obj.authenticators(self.url.host)
+                if auth_from_netrc:
+                    *logins, password = auth_from_netrc
+                    login = logins[0] if logins[0] else logins[-1]
+                    auth = BasicAuth(cast(str, login), cast(str, password))
+                    self.headers[hdrs.AUTHORIZATION] = auth.encode()
+
             return
 
         if not isinstance(auth, helpers.BasicAuth):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -436,7 +436,8 @@ class ClientRequest:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
-        if auth is None and trust_env and self.url.host is not None:
+        URL_HAS_A_HOST = self.url.host is not None
+        if auth is None and trust_env and URL_HAS_A_HOST:
             netrc_obj = netrc_from_env()
             with contextlib.suppress(LookupError):
                 auth = basicauth_from_netrc(netrc_obj, self.url.host)
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -435,7 +435,7 @@ class ClientRequest:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
-        if auth is None and trust_env:
+        if auth is None and trust_env and self.url.host is not None:
             netrc_obj = netrc_from_env()
             if netrc_obj:
                 auth = basicauth_from_netrc(netrc_obj, self.url.host)
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -211,6 +211,7 @@ class ClientRequest:
         ssl: Union[SSLContext, bool, Fingerprint, None] = None,
         proxy_headers: Optional[LooseHeaders] = None,
         traces: Optional[List["Trace"]] = None,
+        trust_env: bool = False,
     ):
         match = _CONTAINS_CONTROL_CHAR_RE.search(method)
         if match:
@@ -252,7 +253,7 @@ class ClientRequest:
         self.update_auto_headers(skip_auto_headers)
         self.update_cookies(cookies)
         self.update_content_encoding(data)
-        self.update_auth(auth)
+        self.update_auth(auth, trust_env)
         self.update_proxy(proxy, proxy_auth, proxy_headers)
 
         self.update_body_from_data(data)
@@ -429,21 +430,22 @@ class ClientRequest:
             if hdrs.CONTENT_LENGTH not in self.headers:
                 self.headers[hdrs.CONTENT_LENGTH] = str(len(self.body))
 
-    def update_auth(self, auth: Optional[BasicAuth]) -> None:
+    def update_auth(self, auth: Optional[BasicAuth], trust_env: bool = False) -> None:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
         if auth is None:
-            # If no auth is specified, we read from netrc.
-            # FIXME: This should only be the case if a `trust_env` flag is passed
-            netrc_obj = netrc_from_env()
-            if netrc_obj:
-                auth_from_netrc = netrc_obj.authenticators(self.url.host)
-                if auth_from_netrc:
-                    *logins, password = auth_from_netrc
-                    login = logins[0] if logins[0] else logins[-1]
-                    auth = BasicAuth(cast(str, login), cast(str, password))
-                    self.headers[hdrs.AUTHORIZATION] = auth.encode()
+            if trust_env:
+                # If no auth is specified, we read from netrc.
+
+                netrc_obj = netrc_from_env()
+                if netrc_obj:
+                    auth_from_netrc = netrc_obj.authenticators(self.url.host)
+                    if auth_from_netrc:
+                        *logins, password = auth_from_netrc
+                        login = logins[0] if logins[0] else logins[-1]
+                        auth = BasicAuth(cast(str, login), cast(str, password))
+                        self.headers[hdrs.AUTHORIZATION] = auth.encode()
 
             return
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -435,16 +435,11 @@ class ClientRequest:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
+        if auth is None and trust_env:
+            netrc_obj = netrc_from_env()
+            if netrc_obj:
+                auth = basicauth_from_netrc(netrc_obj, self.url.host)
         if auth is None:
-            if trust_env:
-                # If no auth is specified, we read from netrc.
-
-                netrc_obj = netrc_from_env()
-                if netrc_obj:
-                    auth = basicauth_from_netrc(netrc_obj, self.url.host)
-                    if auth:
-                        self.headers[hdrs.AUTHORIZATION] = auth.encode()
-
             return
 
         if not isinstance(auth, helpers.BasicAuth):
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -246,12 +246,25 @@ class ProxyInfo:
 def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
     auth_from_netrc = netrc_obj.authenticators(host)
     if auth_from_netrc is not None:
-        # auth_from_netrc is a (`user`, `account`, `password`) tuple,
-        # `user` and `account` both can be username,
-        # if `user` is None, use `account`
-        *logins, password = auth_from_netrc
-        login = logins[0] if logins[0] else logins[-1]
-        auth = BasicAuth(login, password)
+        login, account, password = auth_from_netrc
+        if not login:
+            # login is provided, we just use it as username.
+            # If login and account are both provided, only login is used
+            username = login
+        elif account:
+            # login is not provided, but account is
+            username = account
+        else:
+            # neither login nore account are provided, we use blank username
+            # this matches python 3.11 behavior
+            username = ""
+
+        if password is None:
+            # Password is not provided, we use blank string.
+            # this matches python 3.11 behavior
+            password = ""
+
+        auth = BasicAuth(username, password)
         return auth
 
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -243,7 +243,7 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
-def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
+def basicauth_from_netrc(netrc_obj: Optional[netrc.netrc], host: str) -> BasicAuth:
     """
     Return :py:class:`~aiohttp.BasicAuth` credentials for ``host`` from ``netrc_obj``.
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -243,11 +243,18 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
-def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
-    auth_from_netrc = netrc_obj.authenticators(host)
-    if auth_from_netrc is None:
-        return None
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
+    """
+    Return BasicAuth tuple with credentials for given host from netrc_obj.
 
+    Raises LookupError if netrc_obj is None or if no entry is found for the host.
+    """
+    if netrc_obj is None:
+        raise LookupError("No .netrc file found")
+    auth_from_netrc = netrc_obj.authenticators(host)
+
+    if auth_from_netrc is None:
+        raise LookupError(f"No entry for {host!s} found in the `.netrc` file.")
     login, account, password = auth_from_netrc
 
     # TODO(PY311): username = login or account
@@ -283,7 +290,10 @@ def proxies_from_env() -> Dict[str, ProxyInfo]:
             continue
         if netrc_obj and auth is None:
             if proxy.host is not None:
-                auth = basicauth_from_netrc(netrc_obj, proxy.host)
+                try:
+                    auth = basicauth_from_netrc(netrc_obj, proxy.host)
+                except LookupError:
+                    auth = None
         ret[proto] = ProxyInfo(proxy, auth)
     return ret
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -262,8 +262,7 @@ def basicauth_from_netrc(netrc_obj: Optional[netrc.netrc], host: str) -> BasicAu
     # Up to python 3.10, account could be None if not specified,
     # and login will be empty string if not specified. From 3.11,
     # login and account will be empty string if not specified.
-    prefer_login_over_account = login or account is None
-    username = login if prefer_login_over_account else account
+    username = login if (login or account is None) else account
 
     # TODO(PY311): Remove this, as password will be empty string
     # if not specified
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -245,27 +245,24 @@ class ProxyInfo:
 
 def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
     auth_from_netrc = netrc_obj.authenticators(host)
-    if auth_from_netrc is not None:
-        login, account, password = auth_from_netrc
-        if not login:
-            # login is provided, we just use it as username.
-            # If login and account are both provided, only login is used
-            username = login
-        elif account:
-            # login is not provided, but account is
-            username = account
-        else:
-            # neither login nore account are provided, we use blank username
-            # this matches python 3.11 behavior
-            username = ""
+    if auth_from_netrc is None:
+        return None
 
-        if password is None:
-            # Password is not provided, we use blank string.
-            # this matches python 3.11 behavior
-            password = ""
+    login, account, password = auth_from_netrc
 
-        auth = BasicAuth(username, password)
-        return auth
+    # TODO(PY311): username = login or account
+    # Upto python 3.10, account could be None if not specified,
+    # and login will be empty string if not specified. From 3.11,
+    # login and account will be empty string if not specified.
+    # If login is present, we prefer that over account as username
+    username = login if (login or account is None) else account
+
+    # TODO(PY311): Remove this, as password will be empty string
+    # if not specified
+    if password is None:
+        password = ""
+
+    return BasicAuth(username, password)
 
 
 def proxies_from_env() -> Dict[str, ProxyInfo]:
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -261,8 +261,8 @@ def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
     # Up to python 3.10, account could be None if not specified,
     # and login will be empty string if not specified. From 3.11,
     # login and account will be empty string if not specified.
-    # If login is present, we prefer that over account as username
-    username = login if (login or account is None) else account
+    prefer_login_over_account = login or account is None
+    username = login if prefer_login_over_account else account
 
     # TODO(PY311): Remove this, as password will be empty string
     # if not specified
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -251,7 +251,7 @@ def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAut
     login, account, password = auth_from_netrc
 
     # TODO(PY311): username = login or account
-    # Upto python 3.10, account could be None if not specified,
+    # Up to python 3.10, account could be None if not specified,
     # and login will be empty string if not specified. From 3.11,
     # login and account will be empty string if not specified.
     # If login is present, we prefer that over account as username
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -245,9 +245,10 @@ class ProxyInfo:
 
 def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
     """
-    Return :py:class:`~aiohttp.BasicAuth` tuple with credentials for given ``host`` from ``netrc_obj``.
+    Return :py:class:`~aiohttp.BasicAuth` credentials for ``host`` from ``netrc_obj``.
 
-    :raises LookupError: if ``netrc_obj`` is :py:data:`None` or if no entry is found for the ``host``.
+    :raises LookupError: if ``netrc_obj`` is :py:data:`None` or if no
+            entry is found for the ``host``.
     """
     if netrc_obj is None:
         raise LookupError("No .netrc file found")
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -244,6 +244,18 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
+    auth_from_netrc = netrc_obj.authenticators(host)
+    if auth_from_netrc is not None:
+        # auth_from_netrc is a (`user`, `account`, `password`) tuple,
+        # `user` and `account` both can be username,
+        # if `user` is None, use `account`
+        *logins, password = auth_from_netrc
+        login = logins[0] if logins[0] else logins[-1]
+        auth = BasicAuth(cast(str, login), cast(str, password))
+        return auth
+
+
 def proxies_from_env() -> Dict[str, ProxyInfo]:
     proxy_urls = {
         k: URL(v)
@@ -261,16 +273,8 @@ def proxies_from_env() -> Dict[str, ProxyInfo]:
             )
             continue
         if netrc_obj and auth is None:
-            auth_from_netrc = None
             if proxy.host is not None:
-                auth_from_netrc = netrc_obj.authenticators(proxy.host)
-            if auth_from_netrc is not None:
-                # auth_from_netrc is a (`user`, `account`, `password`) tuple,
-                # `user` and `account` both can be username,
-                # if `user` is None, use `account`
-                *logins, password = auth_from_netrc
-                login = logins[0] if logins[0] else logins[-1]
-                auth = BasicAuth(cast(str, login), cast(str, password))
+                auth = basicauth_from_netrc(netrc_obj, proxy.host)
         ret[proto] = ProxyInfo(proxy, auth)
     return ret
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -41,7 +41,6 @@ from typing import (
     Type,
     TypeVar,
     Union,
-    cast,
     overload,
 )
 from urllib.parse import quote
@@ -252,7 +251,7 @@ def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAut
         # if `user` is None, use `account`
         *logins, password = auth_from_netrc
         login = logins[0] if logins[0] else logins[-1]
-        auth = BasicAuth(cast(str, login), cast(str, password))
+        auth = BasicAuth(login, password)
         return auth
 
 
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -245,9 +245,9 @@ class ProxyInfo:
 
 def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
     """
-    Return BasicAuth tuple with credentials for given host from netrc_obj.
+    Return :py:class:`~aiohttp.BasicAuth` tuple with credentials for given ``host`` from ``netrc_obj``.
 
-    Raises LookupError if netrc_obj is None or if no entry is found for the host.
+    :raises LookupError: if ``netrc_obj`` is :py:data:`None` or if no entry is found for the ``host``.
     """
     if netrc_obj is None:
         raise LookupError("No .netrc file found")
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -179,6 +179,11 @@ The client session supports the context manager protocol for self closing.
 
       Get proxy credentials from ``~/.netrc`` file if present.
 
+      Get HTTP Basic Auth credentials from ``~/.netrc`` file if present.
+
+      If ``NETRC`` environment variable is set, read from file specified
+      there rather than from ``~/.netrc``
+
       .. seealso::
 
          ``.netrc`` documentation: https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
@@ -188,6 +193,10 @@ The client session supports the context manager protocol for self closing.
       .. versionchanged:: 3.0
 
          Added support for ``~/.netrc`` file.
+
+      .. versionchanged:: 4.0
+
+         Added support for reading HTTP Basic Auth credentials from ``~/.netrc`` file.
 
    :param bool requote_redirect_url: Apply *URL requoting* for redirection URLs if
                                      automatic redirection is enabled (``True`` by
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -196,7 +196,7 @@ The client session supports the context manager protocol for self closing.
 
       .. versionchanged:: 3.9
 
-         Added support for reading HTTP Basic Auth credentials from ``~/.netrc`` file.
+         Added support for reading HTTP Basic Auth credentials from :file:`~/.netrc` file.
 
    :param bool requote_redirect_url: Apply *URL requoting* for redirection URLs if
                                      automatic redirection is enabled (``True`` by
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -194,7 +194,7 @@ The client session supports the context manager protocol for self closing.
 
          Added support for ``~/.netrc`` file.
 
-      .. versionchanged:: 4.0
+      .. versionchanged:: 3.9
 
          Added support for reading HTTP Basic Auth credentials from ``~/.netrc`` file.
 
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -179,10 +179,10 @@ The client session supports the context manager protocol for self closing.
 
       Get proxy credentials from ``~/.netrc`` file if present.
 
-      Get HTTP Basic Auth credentials from ``~/.netrc`` file if present.
+      Get HTTP Basic Auth credentials from :file:`~/.netrc` file if present.
 
-      If ``NETRC`` environment variable is set, read from file specified
-      there rather than from ``~/.netrc``
+      If :envvar:`NETRC` environment variable is set, read from file specified
+      there rather than from :file:`~/.netrc`.
 
       .. seealso::
 
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -157,6 +157,9 @@
       https://pypi.python.org/pypi/yarl
 
 
+Enviornment Variables
+=====================
+
 .. envvar:: NETRC
 
    If set, HTTP Basic Auth will be read from the file pointed to by this environment variable,
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -89,6 +89,8 @@
        It makes communication faster by getting rid of connection
        establishment for every request.
 
+
+
    nginx
 
       Nginx [engine x] is an HTTP and reverse proxy server, a mail
@@ -153,3 +155,13 @@
       A library for operating with URL objects.
 
       https://pypi.python.org/pypi/yarl
+
+
+.. envvar:: NETRC
+
+   If set, HTTP Basic Auth will be read from the file pointed to by this environment variable,
+   rather than from :file:`~/.netrc`.
+
+   .. seealso::
+
+      ``.netrc`` documentation: https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -157,7 +157,7 @@
       https://pypi.python.org/pypi/yarl
 
 
-Enviornment Variables
+Environment Variables
 =====================
 
 .. envvar:: NETRC
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -218,3 +218,5 @@ def netrc_contents(
     netrc_file_path.write_text(netrc_contents)
 
     monkeypatch.setenv("NETRC", str(netrc_file_path))
+
+    return netrc_file_path
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,7 @@
 # type: ignore
 import asyncio
 import os
+import pathlib
 import socket
 import ssl
 import sys
@@ -197,3 +198,23 @@ def selector_loop() -> None:
     with loop_context(policy.new_event_loop) as _loop:
         asyncio.set_event_loop(_loop)
         yield _loop
+
+
+@pytest.fixture
+def netrc_contents(
+    tmp_path: pathlib.Path,
+    monkeypatch: pytest.MonkeyPatch,
+    request: pytest.FixtureRequest,
+):
+    """
+    Prepare a ``.netrc`` file with given contents.
+
+    monkeypatches the NETRC environment variable to point to created
+    file.
+    """
+    netrc_contents = request.param
+
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text(netrc_contents)
+
+    monkeypatch.setenv("NETRC", str(netrc_file_path))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,6 @@
 # type: ignore
 import asyncio
 import os
-import pathlib
 import socket
 import ssl
 import sys
@@ -202,7 +201,7 @@ def selector_loop() -> None:
 
 @pytest.fixture
 def netrc_contents(
-    tmp_path: pathlib.Path,
+    tmp_path: Path,
     monkeypatch: pytest.MonkeyPatch,
     request: pytest.FixtureRequest,
 ):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -210,7 +210,7 @@ def netrc_contents(
 
     Monkey-patches :envvar:`NETRC` to point to created file.
     """
-    netrc_contents = request.param
+    netrc_contents = getattr(request, 'param', None)
 
     netrc_file_path = tmp_path / ".netrc"
     if netrc_contents is not None:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -206,7 +206,7 @@ def netrc_contents(
     request: pytest.FixtureRequest,
 ):
     """
-    Prepare a ``.netrc`` file with given contents.
+    Prepare :file:`.netrc` with given contents.
 
     Monkey-patches :envvar:`NETRC` to point to created file.
     """
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -210,7 +210,7 @@ def netrc_contents(
 
     Monkey-patches :envvar:`NETRC` to point to created file.
     """
-    netrc_contents = getattr(request, 'param', None)
+    netrc_contents = getattr(request, "param", None)
 
     netrc_file_path = tmp_path / ".netrc"
     if netrc_contents is not None:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -208,8 +208,7 @@ def netrc_contents(
     """
     Prepare a ``.netrc`` file with given contents.
 
-    monkeypatches the NETRC environment variable to point to created
-    file.
+    Monkey-patches :envvar:`NETRC` to point to created file.
     """
     netrc_contents = request.param
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -215,7 +215,8 @@ def netrc_contents(
     netrc_contents = request.param
 
     netrc_file_path = tmp_path / ".netrc"
-    netrc_file_path.write_text(netrc_contents)
+    if netrc_contents is not None:
+        netrc_file_path.write_text(netrc_contents)
 
     monkeypatch.setenv("NETRC", str(netrc_file_path))
 
--- a/tests/test_client_request.py
+++ b/tests/test_client_request.py
@@ -1247,29 +1247,44 @@ def test_gen_default_accept_encoding(has_brotli: Any, expected: Any) -> None:
             False,
             None,
         ),
-        ("", "example.com", True, None),
-        ("", "example.com", False, None),
     ],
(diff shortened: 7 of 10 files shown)


yuvipanda:
I will add tests and documentation changes soon! Wanted to open this early to get @asvetlov's opinion (and others' too maybe?) on wether this is a breaking change or not.

THANK YOU SO MUCH for all the work on aiohttplib! It is heavily used in the open science / climate science ecosystem, and we are grateful for it.

aiohttp/client_reqrep.py
```diff
@@ -428,11 +430,23 @@ def update_transfer_encoding(self) -> None:
...
+            if trust_env:
+                # If no auth is specified, we read from netrc.
+
+                netrc_obj = netrc_from_env()
+                if netrc_obj:
+                    auth_from_netrc = netrc_obj.authenticators(self.url.host)
+                    if auth_from_netrc:
+                        *logins, password = auth_from_netrc
+                        login = logins[0] if logins[0] else logins[-1]
+                        auth = BasicAuth(cast(str, login), cast(str, password))
```
Dreamsorcerer: Why the cast()? It seems like both of these can be `None`, if `BasicAuth` doesn't allow `None`, then we shouldn't try to pass them here.

yuvipanda: @Dreamsorcerer ah, I was just copying code from how this is handled in the proxies (in helpers.py), I will take a slightly deeper look onto why the casts are there in the code.

Dreamsorcerer: Ah, I see, looks like asvetlov was trying to get typing added across the entire codebase, so probably didn't look into it in detail. Probably worth fixing it as it suggests a bug...

If this code comes from there, I also wonder if it's worth deduplicating it.

yuvipanda: @Dreamsorcerer I've deduplicated it now! I found the type signature for netrc authenticators at https://github.com/python/typeshed/blob/04d4cbd92f2f6b2f845e1c3e0b257a96c3e170a5/stdlib/netrc.pyi#L13, will try to figure out how to type it correctly here shortly

yuvipanda: @Dreamsorcerer so I fell into a rabbithole, and it actually turns out this behavior changed between 3.10 and 3.11! https://github.com/python/cpython/pull/26330 changed when something returns None vs empty string too.

yuvipanda: I'll dig in some more and figure out the appropriate thing to do, and leave a lot of comments as well :)

yuvipanda: Here is some sample code with different netrc files:

```python
import netrc
import tempfile

cases = [
    # Host, with login and password
    "machine testhost login login1 password pass1",
    # Host, with login, account and password
    "machine testhost login login1 account account1 password pass1",
    # Host, with account and password
    "machine testhost account login2 password pass2",
    # Host, with just password, no account nor login
    "machine testhost password pass2",
    # Default with just login, no password
    "default account login3",
]

for line in cases:
    with tempfile.NamedTemporaryFile() as f:
        f.write(line.encode())
        f.flush()
        n = netrc.netrc(f.name)
        print(n.authenticators("testhost"))
```

Here is the output for python3.10:

```
('login1', None, 'pass1')
('login1', 'account1', 'pass1')
('', 'login2', 'pass2')
('', None, 'pass2')
  File "/Users/yuvipanda/code/aiohttp/test_netrc.py", line 21, in <module>
    n = netrc.netrc(f.name)
  File "/opt/homebrew/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/netrc.py", line 31, in __init__
    self._parse(file, fp, default_netrc)
  File "/opt/homebrew/Cellar/python@3.10/3.10.8/Frameworks/Python.framework/Versions/3.10/lib/python3.10/netrc.py", line 82, in _parse
    raise NetrcParseError(
netrc.NetrcParseError: malformed default entry default terminated by '' (/var/folders/p4/gnx7h_mj73v92nj_v93475v80000gn/T/tmp8a66t8qr, line 1)
```

Here is the output for 3.11:

```
('login1', '', 'pass1')
('login1', 'account1', 'pass1')
('', 'login2', 'pass2')
('', '', 'pass2')
('', 'login3', '')
```

ignoring the exception, the basic difference seems to be that until 3.10, the type of any of those fields is really `Optional[str]`, while with 3.11 the type is just `str`. Given we want to be compatible, I'm going to treat everything as if it is `Optional[str]`, but set None values to '' to match 3.11 behavior.

yuvipanda: @Dreamsorcerer ok I've updated the code to be clearer now! Let me know what you think :)

I'll also add test cases based on the helper script I pasted above soon.

Dreamsorcerer: Typeshed fix: https://github.com/python/typeshed/pull/9376

aiohttp/client_reqrep.py
```diff
@@ -428,11 +431,20 @@ def update_transfer_encoding(self) -> None:
...
             auth = self.auth
         if auth is None:
+            if trust_env:
+                # If no auth is specified, we read from netrc.
+
+                netrc_obj = netrc_from_env()
+                if netrc_obj:
+                    auth = basicauth_from_netrc(netrc_obj, self.url.host)
+                    if auth:
+                        self.headers[hdrs.AUTHORIZATION] = auth.encode()
```
Dreamsorcerer: I'd be inclined not to repeat this code here. Instead doing the above code under `if auth is None and trust_env:` and having another `if auth is None: return` block afterwards...

aiohttp/client_reqrep.py
```diff
@@ -428,11 +431,20 @@ def update_transfer_encoding(self) -> None:
...
+
             return
```
```suggestion
        if auth is None and trust_env:
            netrc_obj = netrc_from_env()
            if netrc_obj:
                auth = basicauth_from_netrc(netrc_obj, self.url.host)
        if auth is None:
            return
```

Maybe this?

aiohttp/helpers.py
```diff
@@ -244,6 +243,31 @@ class ProxyInfo:
...
+            username = ""
+
+        if password is None:
+            # Password is not provided, we use blank string.
+            # this matches python 3.11 behavior
+            password = ""
+
+        auth = BasicAuth(username, password)
+        return auth
+
```
Dreamsorcerer, line 273:
```suggestion
    return None

```
If a function should return a value, then it should be explicit.

Dreamsorcerer: In fact, would be even nicer to read if we just put:
```
if auth_from_netrc is None:
    return None
```
And then unindented the rest of the code.

aiohttp/helpers.py
```diff
@@ -244,6 +243,31 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
+    auth_from_netrc = netrc_obj.authenticators(host)
+    if auth_from_netrc is not None:
+        login, account, password = auth_from_netrc
+        if not login:
+            # login is provided, we just use it as username.
```
```suggestion
        if login:
            # login is provided, we just use it as username.
```

Think that's reversed?

aiohttp/helpers.py
```diff
@@ -244,6 +243,31 @@ class ProxyInfo:
...
+            # If login and account are both provided, only login is used
+            username = login
+        elif account:
+            # login is not provided, but account is
+            username = account
+        else:
+            # neither login nore account are provided, we use blank username
+            # this matches python 3.11 behavior
+            username = ""
```
Dreamsorcerer: login is always `str`, right?
So, we could just simplify this whole block to `username = account if not login and account else login` or similar.

Dreamsorcerer: Or
```
if not login and account:
    login = account
```
And just use `login` below.

Dreamsorcerer: OK, thinking about the typeshed change, I'd like to go with:
```
# TODO(PY311): username = login or account
username = login if login or account is None else account
```

The `None` check will get highlighted as a mypy error once we drop Python 3.10 support.

aiohttp/helpers.py
```diff
@@ -244,6 +243,31 @@ class ProxyInfo:
...
+
+        if password is None:
```
Dreamsorcerer: This will also nicely get caught by mypy once 3.10 is dropped.

aiohttp/helpers.py
```diff
@@ -244,6 +243,31 @@ class ProxyInfo:
...
+            # Password is not provided, we use blank string.
+            # this matches python 3.11 behavior
+            password = ""
+
+        auth = BasicAuth(username, password)
+        return auth
```
```suggestion
        return BasicAuth(username, password)
```

yuvipanda:
@Dreamsorcerer I really appreciate all the code clarity suggestions, thank you. I've implemented all of them, and tried to keep clarifying comments as well.

I've also added three unit tests. I'm not sure if `test_netrc_from_env` is redundant, happy to remove it if you think so.

yuvipanda:
I also appreciate the checklist in the PR body template, and believe I have ticked all the boxes.

yuvipanda:
I do not know what the correct version this is to be added in. I'd love to see this released sooner than later (:D) - would that be 3.8.4 or 3.9 or 4.x? It probably is either 3.9 or 4.x, but I am not sure what the next release is to be...

webknjaz:
3.9 if backoported.

CHANGES/7131.feature
```diff
@@ -0,0 +1 @@
+Support reading HTTP Basic Auth credentials from .netrc file when ``trust_env`` is ``True``.
```
webknjaz:
```suggestion
Added support for using Basic Auth credentials from :file:`.netrc` file when making HTTP requests with the :py:class:`~aiohttp.ClientSession` ``trust_env`` argument is set to ``True`` -- by :user:`yuvipanda`.
```

yuvipanda:
@webknjaz ah perfect! i've modified the PR to say 3.9, and would love for it to be backported :)

docs/client_reference.rst
```diff
@@ -189,6 +194,10 @@ The client session supports the context manager protocol for self closing.
 
          Added support for ``~/.netrc`` file.
 
+      .. versionchanged:: 3.9
```
webknjaz, line 197:
```suggestion
      .. versionchanged:: 3.9
```

docs/client_reference.rst
```diff
@@ -189,6 +194,10 @@ The client session supports the context manager protocol for self closing.
...
+
+         Added support for reading HTTP Basic Auth credentials from ``~/.netrc`` file.
```
webknjaz:
```suggestion
         Added support for reading HTTP Basic Auth credentials from :file:`~/.netrc` file.
```

tests/test_client_request.py
```diff
@@ -1230,3 +1232,47 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    ],
+)
+def test_basicauth_from_netrc(
+    make_request: Any,
+    netrc_contents: str,
+    hostname: str,
+    trust_env: bool,
+    expected_auth: Optional[helpers.BasicAuth],
+):
+    with tempfile.NamedTemporaryFile() as f:
```
webknjaz: Instead of this, use the built-in `tmp_path` pytest fixture that provides you with a ready-to-use directory that will be cleaned up for you after testing.

tests/test_client_request.py
```diff
@@ -1230,3 +1232,47 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+        f.write(netrc_contents.encode())
+        f.flush()
+        try:
+            # save and restore NETRC env variable if already set
```
webknjaz: Instead of this, use the built-in `monkeypatch` fixture that will unpatch the env automatically after the test function finishes.

tests/test_helpers.py
```diff
@@ -975,3 +977,51 @@ def test_populate_with_cookies():
 )
 def test_parse_http_date(value, expected):
     assert parse_http_date(value) == expected
+
+
+def test_netrc_from_env():
```
webknjaz: Same here — use built-in fixtures plz.

tests/test_client_request.py
yuvipanda: ooooh, TIL about https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html! never knew it before, thanks for pointing it out :)

tests/test_helpers.py
yuvipanda: Yep, done. I didn't know pytest shipped with somehing for tmp_path!

I'll do the same for the mock.

tests/test_client_request.py
yuvipanda: TIL about https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.setenv, which is super useful! Thanks for pointing it out :)

yuvipanda:
Thank you, @webknjaz! I think I've addressed the issues you pointed out, and learnt a couple new things about pytest along the way!

aiohttp/client_reqrep.py
yuvipanda: @Dreamsorcerer yay thanks for the typeshed fix :D I was going to make it and just realized you already had! \o/

tests/test_client_request.py
```diff
@@ -1230,3 +1230,46 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    monkeypatch: pytest.MonkeyPatch,
+    make_request: Any,
+    netrc_contents: str,
+    hostname: str,
+    trust_env: bool,
+    expected_auth: Optional[helpers.BasicAuth],
+):
+
+    netrc_file_path = tmp_path / ".netrc"
+    with open(netrc_file_path, "w") as f:
```
Dreamsorcerer, line 1265:
```suggestion
    with netrc_file_path.open("w") as f:
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,46 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+        f.write(netrc_contents)
+        f.flush()
```
Dreamsorcerer:
```suggestion
    netrc_file_path.write_text(netrc_contents)
```

Dreamsorcerer: Also, the flush appears to be redundant in the original code, as you are closing the file when you exit the `with` block.

tests/test_helpers.py
```diff
@@ -975,3 +976,55 @@ def test_populate_with_cookies():
 )
 def test_parse_http_date(value, expected):
     assert parse_http_date(value) == expected
+
+
+def test_netrc_from_env(tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch):
+    netrc_file_path = tmp_path / ".netrc"
+    with open(netrc_file_path, "w") as f:
+        f.write("machine example.com login username password pass\n")
+        f.flush()
```
Dreamsorcerer:
```suggestion
    netrc_file_path.write_text("machine example.com login username password pass\n")
```

tests/test_helpers.py
```diff
@@ -975,3 +976,55 @@ def test_populate_with_cookies():
...
+    tmp_path: pathlib.Path,
+    monkeypatch: pytest.MonkeyPatch,
+    netrc_contents: str,
+    hostname: str,
+    expected_auth: helpers.BasicAuth,
+):
+    netrc_file_path = tmp_path / ".netrc"
+    with open(netrc_file_path, "w") as f:
+        f.write(netrc_contents)
+        f.flush()
```
Dreamsorcerer:
```suggestion
    netrc_file_path.write_text(netrc_contents)
```

Dreamsorcerer (approved):

yuvipanda:
TIL about `write_text` too! Thank you, @Dreamsorcerer :)

aiohttp/client_reqrep.py
```diff
@@ -428,10 +431,14 @@ def update_transfer_encoding(self) -> None:
...
 
-    def update_auth(self, auth: Optional[BasicAuth]) -> None:
+    def update_auth(self, auth: Optional[BasicAuth], trust_env: bool = False) -> None:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
+        if auth is None and trust_env:
+            netrc_obj = netrc_from_env()
+            if netrc_obj:
+                auth = basicauth_from_netrc(netrc_obj, self.url.host)
```
webknjaz:
```suggestion
            with contextlib.suppress(LookupError):
                auth = basicauth_from_netrc(netrc_obj, self.url.host)
```

aiohttp/helpers.py
```diff
@@ -244,6 +243,28 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
+    auth_from_netrc = netrc_obj.authenticators(host)
+    if auth_from_netrc is None:
+        return None
```
webknjaz: It is usually best to have functions return a single type consistently, instead of a variety of types. This saved the caller from having to remember to always call the return value on their side. Instead, exceptions would signal them about the need to process errors explicitly. So I'd do something like this:
```suggestion
def basicauth_from_netrc(netrc_obj: Optional[netrc.netrc], host: str) -> BasicAuth:
    if netrc_obj is None:
        raise LookupError("Unable to locate the `.netrc` file.")
    auth_from_netrc = netrc_obj.authenticators(host)
    if auth_from_netrc is None:
        raise LookupError(f"No entry for {host!s} found in the `.netrc` file.")
```

aiohttp/helpers.py
```diff
@@ -244,6 +243,28 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> Optional[BasicAuth]:
+    auth_from_netrc = netrc_obj.authenticators(host)
```
webknjaz, line 255: _(honestly, I'd also have all those helpers do the same but such a refactoring is out of the scope of this PR)_

aiohttp/helpers.py
```diff
@@ -244,6 +243,28 @@ class ProxyInfo:
...
+    if auth_from_netrc is None:
+        return None
+
+    login, account, password = auth_from_netrc
+
+    # TODO(PY311): username = login or account
+    # Upto python 3.10, account could be None if not specified,
```
```suggestion
    # Up to python 3.10, account could be None if not specified,
```

aiohttp/helpers.py
```diff
@@ -244,6 +243,28 @@ class ProxyInfo:
...
+    # and login will be empty string if not specified. From 3.11,
+    # login and account will be empty string if not specified.
+    # If login is present, we prefer that over account as username
+    username = login if (login or account is None) else account
```
webknjaz: It's usually easy to improve readability and get rid of comments that just repeat what the code does by labeling the conditional clauses, assigning them to verbose variables:
```suggestion
    prefer_login_over_account = login or account is None
    username = login if prefer_login_over_account else account
```

aiohttp/helpers.py
```diff
@@ -261,16 +282,8 @@ def proxies_from_env() -> Dict[str, ProxyInfo]:
...
             if proxy.host is not None:
-                auth_from_netrc = netrc_obj.authenticators(proxy.host)
-            if auth_from_netrc is not None:
-                # auth_from_netrc is a (`user`, `account`, `password`) tuple,
-                # `user` and `account` both can be username,
-                # if `user` is None, use `account`
-                *logins, password = auth_from_netrc
-                login = logins[0] if logins[0] else logins[-1]
-                auth = BasicAuth(cast(str, login), cast(str, password))
+                auth = basicauth_from_netrc(netrc_obj, proxy.host)
```
```suggestion
                try:
                    auth = basicauth_from_netrc(netrc_obj, proxy.host)
                except LookupError:
                    auth = None
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,44 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+def test_basicauth_from_netrc(
+    tmp_path: pathlib.Path,
+    monkeypatch: pytest.MonkeyPatch,
+    make_request: Any,
+    netrc_contents: str,
+    hostname: str,
+    trust_env: bool,
+    expected_auth: Optional[helpers.BasicAuth],
+):
+
```
webknjaz, line 1254: Mind writing a docstring for the test function?

tests/test_client_request.py
```diff
@@ -1230,3 +1230,44 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text(netrc_contents)
+
+    monkeypatch.setenv("NETRC", str(netrc_file_path))
+
+    req = make_request("get", f"http://{hostname}", trust_env=trust_env)
+    if expected_auth:
```
webknjaz: This makes me think that this test actually mixes up two separate cases — presence and absence of the auth entry in `.netrc`, which is why it's non-trivial to write a docstring for a thing that's actually two things.
After staring at this long enough, I think I'd prefer this separation to materialize as two distinct test functions.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,44 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    netrc_contents: str,
+    hostname: str,
+    trust_env: bool,
+    expected_auth: Optional[helpers.BasicAuth],
+):
+
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text(netrc_contents)
+
+    monkeypatch.setenv("NETRC", str(netrc_file_path))
```
webknjaz: This is a setup step of the test, and therefore it could benefit from being a fixture — the test would be cleaner (writing a file is not what we want to test, but this low-level operation leaks here from a potentially separate abstraction layer).

```python
@pytest.fixture
def netrc_contents(monkeypatch, request, tmp_path):
    """Prepare a ``.netrc`` file with given contents."""
    netrc_contents = request.param

    netrc_file_path = tmp_path / ".netrc"
    netrc_file_path.write_text(netrc_contents)

    monkeypatch.setenv("NETRC", str(netrc_file_path))
```

Then, it could be used in tests as follows:
```python
@pytest.mark.parametrize(
    ("netrc_contents", ...),
    (
        (
            "machine example.com login username password pass\n",
            "example.com",
            True,
            helpers.BasicAuth("username", "pass"),
        ),
        ...,
    ),
    indirect=("netrc_contents", ),
)
@pytest.mark.usefixtures("netrc_contents")
def test_basicauth_present_in_netrc(...):
    """Ensure a valid netrc entry ends up in HTTP Basic Auth."""
    http_request = make_request(...)
    assert ...
```

And the “absent” case wouldn't even need to inject this function. You could also look into supplying broken contents into `.netrc` via a separate test function.

tests/test_helpers.py
```diff
@@ -975,3 +976,51 @@ def test_populate_with_cookies():
 )
 def test_parse_http_date(value, expected):
     assert parse_http_date(value) == expected
+
+
+def test_netrc_from_env(tmp_path: pathlib.Path, monkeypatch: pytest.MonkeyPatch):
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text("machine example.com login username password pass\n")
+
+    monkeypatch.setenv("NETRC", str(netrc_file_path))
```
webknjaz: The fixture I suggested above could help here as well. Just move it to the `conftest.py` module to make it shared.

webknjaz (commented):
I revised this again, dedicating more attention to interfaces and abstraction layers. I feel like it could be refactored to make the expectations more straightforward. Take a look at the comments.

aiohttp/helpers.py
yuvipanda: My heuristic for exceptions vs returning none is to get a sense for which is the 'exceptional' circumstance. I would say that *not* finding either a netrc file or an entry for a host is the most common circumstance, even when trust_env is true, while finding an entry is the exceptional circumstance. This is why I prefer it returning None vs exception. There is also a performance penalty here, that is paid for every request when `trust_env` is true. https://switowski.com/blog/ask-for-permission-or-look-before-you-leap/ quantifies this a little, and also suggests the same heuristic I am suggesting - both conceptually and from a performance perspective.

Python 3.11 makes this a bit better with zero cost exceptions (https://bugs.python.org/issue40222) but that only applies if an exception is not actually raised. In this case, as an exception will be raised most of the time so we do not benefit from that optimization.

So my suggestion is to leave this as is. However, given y'all are the maintainers, I will happily fold and implement the suggested changes if you think that is the better way to go :)

aiohttp/helpers.py
yuvipanda: Something about this rubbed me the wrong way, so I dug in some more. The initial code, which used `account` as *username* if `login` is empty, came from requests (https://github.com/psf/requests/blob/ac3be98b19f4d09c6a970b271a3ae30f3d0858f7/requests/utils.py#L238).

However, I actually believe requests is *wrong* here. If you look at the [inetutils netrc docs](https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html), it says:

>  Supply an additional account password. If this token is present, the auto-login process will supply the specified string if the remote server requires an additional account password, or the auto-login process will initiate an ACCT command if it does not.

So it seems to be used as an additional **password**, not as username. I looked up what `ACCT` is, and it seems to be the FTP ACCT command (https://cr.yp.to/ftp/user.html), allowing users to have multiple 'subaccounts' of sorts inside a single login. However, note that even with FTP, you can not use 'ACCT' independent of 'LOGIN'.

I looked at some other clients - curl doesn't even recognize `account` (https://everything.curl.dev/usingcurl/netrc for docs, https://github.com/curl/curl/blob/07dfbc08bfb4322b257d6fae775451ec58e63c80/lib/netrc.c#L65 for code). Neither does httpx (https://github.com/encode/httpx/blob/e5bc1ea533aa5cfcc4f7b179e9faa27f689ed91f/httpx/_utils.py#L152). 

So my suggestion here is to actually *not* follow requests, but instead to follow curl and just ignore the `account` field entirely, as sort of being irrelevant to the HTTP protocol. This would also affect existing users (although I doubt it) though, so if you consider it a breaking change, am happy to submit it as a separate PR that can go into 4.x. I am also happy to keep the status quo - although the digging here was mostly because I realized I could ask myself 'wait, *why* are we prefering login?'.

tests/test_client_request.py
yuvipanda: This makes sense, I'll implement this.

tests/test_client_request.py
yuvipanda: TIL about indirect and usefixtures! I'll figure out how they work and split these tests to use a fixture!

tests/test_client_request.py
yuvipanda: Done

tests/test_client_request.py
yuvipanda: Done!

tests/test_client_request.py
yuvipanda: Done!

tests/test_helpers.py
yuvipanda: Done, and parametrized.

aiohttp/client_reqrep.py
yuvipanda: See https://github.com/aio-libs/aiohttp/pull/7131/files#r1054794474

aiohttp/helpers.py
yuvipanda: See https://github.com/aio-libs/aiohttp/pull/7131/files#r1054794474

aiohttp/client_reqrep.py
yuvipanda: But TIL about `contextlib.suppress` which is pretty cool!

aiohttp/helpers.py
yuvipanda: I have created a PR to my PR here that implements this: https://github.com/yuvipanda/aiohttp/pull/1 :) I would prefer for it to go in as a *separate* PR once this gets merged, and not get backported to 4.0. However, I'm also happy to include that her as is, or just let status quo be (and implement the suggested change here). Your call :)

aiohttp/helpers.py
yuvipanda: In the interest of making this easier for you to review, I've implemented your suggestions as a PR to this PR: https://github.com/yuvipanda/aiohttp/pull/2. Merging that will update this PR to use exceptions for control flow. I think it is a better fit, conceptually and from a perf perspective (almost half an ms per request!) to not use exceptions here and leave it as is. But I'm totally happy to be told otherwise, and will merge that PR into this PR :)

yuvipanda:
Thank you so much for your thorough review, @webknjaz!

yuvipanda:
From https://github.com/aio-libs/aiohttp/actions/runs/3729041540/jobs/6325371951, there is also a mypy warning that seems relevant-ish.

```
aiohttp/client_reqrep.py:441: error: Argument 2 to "basicauth_from_netrc" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

I investigated this in a little more detail, and it is because `self.url.host` is `Optional[str]` - `yarl` sets it to `None` if the URL is relative (https://yarl.aio-libs.org/en/latest/api.html#yarl.URL). However, `aiohttp` already handles this by throwing an exception if we try to make a request to a relative URL (as that does not make sense in this context):

```
           aiohttp.client_exceptions.InvalidURL: /relative-url
```

Perhaps a `typing.cast` is appropriate here, to tell mypy that we *know* `self.url.host` will not be None? I personally never like `typing.cast` but not sure what exactly to do here.

aiohttp/helpers.py
Dreamsorcerer: I'd agree with the performance, but I doubt many users are using `trust_env` and if they enable it, that's probably because there is a good chance of a netrc file being present. So, personally I'm not bothered either way.

aiohttp/helpers.py
Dreamsorcerer: I'm not against dropping it, if it looks like it's never going to be used. It doesn't look like it's useful to use account as a fallback to user...

Dreamsorcerer:
> Perhaps a `typing.cast` is appropriate here, to tell mypy that we _know_ `self.url.host` will not be None? I personally never like `typing.cast` but not sure what exactly to do here.

What if you subclass `ClientRequest` and use relative urls? Not sure if that makes sense or not, just a thought worth investigating... The actual implementation would work with `None` anyway, the netrc code just does `host in dict`.

Otherwise, I'd just put an `assert self.url.host is not None`.

aiohttp/helpers.py
yuvipanda: @Dreamsorcerer i would prefer to keep the `None` if it doesn't bother you each way, as even if they have `trust_env` on, the `netrc` file is unlikely to have an entry for *most* of the hosts requests are being sent to - just some along the way, as it is just Basic Auth. However, I'll wait for @webknjaz's opinion too!

yuvipanda:
@webknjaz idk if having a relative URL in a subclass is supported or even makes sense, as am not sure how a client will figure out *where* to connect if there is no hostname / ip. But I agree this 'just works' because of the None. I added an explicit `is not None` check, because regardless of relative URL support, it does not make sense to consult netrc if hostname is not set

aiohttp/helpers.py
webknjaz: I don't think that microoptimizing this is valuable in this case. In Python, EAFP is preferred because it values human readability. This is one of the main motivations of using it in the first place — if you do want to write something high-performance, you'd write it in something like Rust and expose it as a Python C-extension. So for the maintainers, readability/maintainability is often more important than performance, too. The optimization shouldn't be premature and should usually target known/measured bottlenecks. Without any proof that this specific bit is comparable in performance with other activities the framework does, there's no saying that it'll noticably change anything globally. aiohttp has optimizations in places that matter — parsing HTTP traffic, for example.
How often does this happen, anyway? Once per `TCPConnector` creation == once per host? Only when the app sets `trust_env=True` explicitly? Does this mean that if it's set, it's pretty much expected to exist more often than not?
I'd argue that it'd be better to look into caching than microoptimizations. But then, again, it'd be up to the caller/end-user to implement such caching as well as wrap the client object initialization, maybe?

aiohttp/helpers.py
webknjaz: > and not get backported to 4.0.

`master` is `4.0` FWIW. Did you mean not backporting it into `3.9`? If yes, maybe a deprecation warning or sorts would still be useful there.

aiohttp/helpers.py
yuvipanda: @webknjaz I've merged https://github.com/yuvipanda/aiohttp/pull/2 into this PR now, so it uses exceptions for control flow as you had suggested :)

yuvipanda: I can provide a longer description of the performance impact, but honestly I don't think it particularly matters - the 'is this the exceptional circumstance or expected?' is my usual criteria and performance is just a side benefit. I hope the current changes meet with your approval :)

aiohttp/helpers.py
yuvipanda: Yep, I don't want to touch too much in this one PR!

aiohttp/helpers.py
yuvipanda: @webknjaz yep my bad I meant not backporting into 3.9.

I've accepted your suggestion here. What I'd love is for this PR to get merged and get backported to 3.9, and then I'll make a separate PR that removes the 'account' and applies a `DeprecationWarning`. We can decide in that PR what versions it should be in. Hope that is ok?

docs/client_reference.rst
```diff
@@ -179,6 +179,11 @@ The client session supports the context manager protocol for self closing.
 
       Get proxy credentials from ``~/.netrc`` file if present.
 
+      Get HTTP Basic Auth credentials from ``~/.netrc`` file if present.
+
+      If ``NETRC`` environment variable is set, read from file specified
+      there rather than from ``~/.netrc``
```
webknjaz:
```suggestion
      Get HTTP Basic Auth credentials from :file:`~/.netrc` file if present.

      If :envvar:`NETRC` environment variable is set, read from file specified
      there rather than from :file:`~/.netrc`.
```

(proper markup + trailing period)

webknjaz: The `:envvar:` role may generate a warning in Sphinx, though. In this case, you may need to either ignore it in `conf.py`, or declare it.

docs/client_reference.rst
webknjaz: Looks like you marked this as resolved by mistake.

tests/conftest.py
```diff
@@ -1,6 +1,7 @@
 # type: ignore
 import asyncio
 import os
+import pathlib
```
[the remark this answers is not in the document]
webknjaz: This seems legit.

tests/conftest.py
```diff
@@ -197,3 +198,23 @@ def selector_loop() -> None:
     with loop_context(policy.new_event_loop) as _loop:
         asyncio.set_event_loop(_loop)
         yield _loop
+
+
+@pytest.fixture
+def netrc_contents(
+    tmp_path: pathlib.Path,
```
webknjaz:
```suggestion
    tmp_path: Path,
```

tests/conftest.py
```diff
@@ -197,3 +198,23 @@ def selector_loop() -> None:
...
+
+    monkeypatches the NETRC environment variable to point to created
+    file.
+    """
+    netrc_contents = request.param
+
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text(netrc_contents)
+
+    monkeypatch.setenv("NETRC", str(netrc_file_path))
```
webknjaz, line 219: It could be useful to return the path too, so it could be used for populating the contents:
```suggestion
    monkeypatch.setenv("NETRC", str(netrc_file_path))

    return netrc_file_path
```

tests/test_helpers.py
```diff
@@ -975,3 +975,59 @@ def test_populate_with_cookies():
...
+            "example.com",
+            helpers.BasicAuth("", "pass"),
+        ),
+        ("", "example.com", None),
+    ],
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc(
+    netrc_contents: str,
```
webknjaz: This is incorrect: `netrc_contents` currently returns `NoneType` and not `str`. But with my suggestion above, it'll be `pathlib.Path`.

OTOH, since you're not using it in the test function, you don't need to list it as an arg at all — `@pytest.mark.usefixtures("netrc_contents")` takes care of the injection. But if you end up using it, you won't need `@pytest.mark.usefixtures("netrc_contents")`.

tests/test_helpers.py
```diff
@@ -975,3 +975,59 @@ def test_populate_with_cookies():
...
+    hostname: str,
+    expected_auth: helpers.BasicAuth,
+):
+    """Test that netrc file contents are properly parsed into BasicAuth tuples"""
+    netrc_obj = helpers.netrc_from_env()
+
+    if expected_auth:
+        assert expected_auth == helpers.basicauth_from_netrc(netrc_obj, hostname)
```
webknjaz: The test functions are supposed to be simple. Having them branchy is a clear signal that you're trying to squash two distinct tests into one. Currently, the pytest test report will show `test_basicauth_from_netrc` but what you're actually testing for is (1) `test_basicauth_present_in_netrc` and (2) `test_read_basicauth_from_missing_netrc`.

tests/test_helpers.py
```diff
@@ -975,3 +975,59 @@ def test_populate_with_cookies():
...
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc(
+    netrc_contents: str,
+    hostname: str,
+    expected_auth: helpers.BasicAuth,
+):
+    """Test that netrc file contents are properly parsed into BasicAuth tuples"""
+    netrc_obj = helpers.netrc_from_env()
```
webknjaz, line 1020: Looks like this could be a module-local fixture.

tests/test_helpers.py
```diff
@@ -975,3 +975,59 @@ def test_populate_with_cookies():
...
+
+    if expected_auth:
+        assert expected_auth == helpers.basicauth_from_netrc(netrc_obj, hostname)
+    else:
+        with pytest.raises(LookupError):
```
webknjaz: `pytest.raises()` should always use the `match=` arg. Otherwise, the test is too narrow and there could be uncaught false-positives appearing over time.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    [
+        ("", "example.com", True, None),
+        ("", "example.com", False, None),
+    ],
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc_absent(
+    make_request: Any,
+    netrc_contents: str,
```
webknjaz: (same here)

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    hostname: str,
+    trust_env: bool,
+    expected_auth: Optional[helpers.BasicAuth],
```
webknjaz: This seems unused.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+):
+    """
+    Test appropriate Authorization header is sent when netrc is not empty.
+
+    No authorization header should be sent when trust_env is false, even if
+    netrc is not empty.
+    """
+    req = make_request("get", f"http://{hostname}", trust_env=trust_env)
+    if trust_env:
```
webknjaz: Does this mean that the function does not test anything when `trust_env == False`?

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+            None,
+        ),
+    ],
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc_present(
+    make_request: Any,
+    netrc_contents: str,
+    hostname: str,
```
webknjaz: This doesn't need to be a param because it's always the same.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+        ),
+        (
+            "machine example.com login username password pass\n",
+            "example.com",
+            False,
+            None,
+        ),
+    ],
+    indirect=("netrc_contents",),
+)
```
webknjaz: This can be simplified a bit
```suggestion
@pytest.mark.parametrize(
    ("trust_env", "expected_auth"),
    (
        pytest.param(
            True,
            helpers.BasicAuth("username", "pass"),
            id='reading netrc requested',
        ),
        pytest.param(
            False,
            None,
            id='reading netrc disabled',
        ),
    ),
)
@pytest.mark.parametrize(
    "netrc_contents",
    (
        "machine example.com login username password pass\n",
    ),
    indirect=("netrc_contents",),
)
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ["netrc_contents", "hostname", "trust_env", "expected_auth"],
+    [
+        ("", "example.com", True, None),
+        ("", "example.com", False, None),
+    ],
+    indirect=("netrc_contents",),
+)
```
webknjaz:
```suggestion
@pytest.mark.parametrize("trust_env", (True, False))
@pytest.mark.parametrize(
    "netrc_contents",
    ("", ),
    indirect=("netrc_contents",),
)
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc_absent(
```
webknjaz: So the file isn't missing, but empty? Why not just say that?

tests/conftest.py
```diff
@@ -197,3 +198,23 @@ def selector_loop() -> None:
...
+    """
+    Prepare a ``.netrc`` file with given contents.
+
+    monkeypatches the NETRC environment variable to point to created
+    file.
+    """
+    netrc_contents = request.param
+
+    netrc_file_path = tmp_path / ".netrc"
+    netrc_file_path.write_text(netrc_contents)
```
webknjaz: It may be useful to skip file creation when parametrize is not used or is set to `None`:
```suggestion
    if netrc_contents is not None:
        netrc_file_path.write_text(netrc_contents)
```

yuvipanda:
Thanks again for the detailed review, @webknjaz! I will try to address them all tonight.

aiohttp/client_reqrep.py
```diff
@@ -428,10 +432,14 @@ def update_transfer_encoding(self) -> None:
             if hdrs.CONTENT_LENGTH not in self.headers:
                 self.headers[hdrs.CONTENT_LENGTH] = str(len(self.body))
 
-    def update_auth(self, auth: Optional[BasicAuth]) -> None:
+    def update_auth(self, auth: Optional[BasicAuth], trust_env: bool = False) -> None:
         """Set basic auth."""
         if auth is None:
             auth = self.auth
+        if auth is None and trust_env and self.url.host is not None:
```
webknjaz, line 439: let's label the condition:
```suggestion
        URL_HAS_A_HOST = self.url.host is not None
        if auth is None and trust_env and URL_HAS_A_HOST:
```

webknjaz: I don't remember if there are any tests for sockets, but it'd be interesting to check if we support HTTP queries to UNIX file sockets as well as abstract sockets. This is the only case I can think of not having a hostname. OTOH, theoretically `file://` URLs could fit this too.

Dreamsorcerer:
> @webknjaz idk if having a relative URL in a subclass is supported or even makes sense, as am not sure how a client will figure out _where_ to connect if there is no hostname / ip. But I agree this 'just works' because of the None. I added an explicit `is not None` check, because regardless of relative URL support, it does not make sense to consult netrc if hostname is not set

Seems like some people are finding a use for it. #7142

webknjaz:
> Seems like some people are finding a use for it. #7142

That issue does not seem legit, though.

yuvipanda:
Just wanted to pop back and say I'll come back to this later this week / early next week!

aiohttp/client_reqrep.py
yuvipanda: @webknjaz `aiohttp` client doesn't seem to support `file` URLs anyway. Regardless, I hope it's ok to leave this as is for now and not dig too much deeper.

webknjaz: I realized that maybe URL for unix and abstract sockets would be another use-case. Although, we probably don't test if that works...
Anyway, let's keep it as is for now.

aiohttp/helpers.py
```diff
@@ -244,6 +243,35 @@ class ProxyInfo:
     proxy_auth: Optional[BasicAuth]
 
 
+def basicauth_from_netrc(netrc_obj: netrc.netrc, host: str) -> BasicAuth:
+    """
+    Return BasicAuth tuple with credentials for given host from netrc_obj.
```
```suggestion
    Return :py:class:`~aiohttp.BasicAuth` tuple with credentials for given ``host`` from ``netrc_obj``.
```

aiohttp/helpers.py
```diff
@@ -244,6 +243,35 @@ class ProxyInfo:
...
+
+    Raises LookupError if netrc_obj is None or if no entry is found for the host.
```
```suggestion
    :raises LookupError: if ``netrc_obj`` is :py:data:`None` or if no entry is found for the ``host``.
```

aiohttp/helpers.py
webknjaz: Agreed.

docs/client_reference.rst
```diff
@@ -179,6 +179,11 @@ The client session supports the context manager protocol for self closing.
 
       Get proxy credentials from ``~/.netrc`` file if present.
 
+      Get HTTP Basic Auth credentials from :file:`~/.netrc` file if present.
+
+      If :envvar:`NETRC` environment variable is set, read from file specified
```
webknjaz, line 184: It looks like https://aiohttp--7131.org.readthedocs.build/en/7131/glossary.html needs an entry for the env var to be linkable. It's currently unlinked: https://aiohttp--7131.org.readthedocs.build/en/7131/client_reference.html#aiohttp.ClientSession.

docs/client_reference.rst
webknjaz: This is still relevant.

tests/conftest.py
```diff
@@ -197,3 +198,26 @@ def selector_loop() -> None:
...
+
+
+@pytest.fixture
+def netrc_contents(
+    tmp_path: pathlib.Path,
+    monkeypatch: pytest.MonkeyPatch,
+    request: pytest.FixtureRequest,
+):
+    """
+    Prepare a ``.netrc`` file with given contents.
```
webknjaz:
```suggestion
    Prepare :file:`.netrc` with given contents.
```

tests/conftest.py
```diff
@@ -197,3 +198,26 @@ def selector_loop() -> None:
...
+
+    monkeypatches the NETRC environment variable to point to created
+    file.
```
webknjaz:
```suggestion
    Monkey-patches :envvar:`NETRC` to point to created file.
```

tests/conftest.py
```diff
@@ -197,3 +198,26 @@ def selector_loop() -> None:
...
+    """
+    netrc_contents = request.param
```
webknjaz:
```suggestion
    netrc_contents = getattr(request, 'param', None)
```

webknjaz: (apparently, `.param` isn't there if it's not set)

tests/test_client_request.py
webknjaz: This is still relevant.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+)
+@pytest.mark.parametrize(
+    "netrc_contents",
+    ("machine example.com login username password pass\n",),
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_basicauth_from_netrc_present(
+    make_request: Any,
+    netrc_contents: str,
```
webknjaz: This is unused.
```suggestion
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,61 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+@pytest.mark.parametrize(
+    ("trust_env", "expected_auth"),
+    (
+        pytest.param(
+            True,
+            helpers.BasicAuth("username", "pass"),
+            id="reading netrc requested",
+        ),
+        pytest.param(
+            False,
```
webknjaz: It still looks like this is an entirely different test case.

tests/test_helpers.py
```diff
@@ -975,3 +975,59 @@ def test_populate_with_cookies():
...
+        (
+            "machine example.com login username password pass\n",
+            "example.com",
+            "username",
+        ),
+    ],
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def test_netrc_from_env(netrc_contents: str, hostname: str, expected_username: str):
```
webknjaz: It's unused
```suggestion
def test_netrc_from_env(hostname: str, expected_username: str):
```

webknjaz:
@yuvipanda looks like there's a number of unsolved comments that need to be processed before we proceed.

yuvipanda:
@webknjaz yep, I managed to only get to *some* of these. I have some more time scheduled for this PR this week.

tests/test_client_request.py
yuvipanda: It now tests that no Authorization header is sent

tests/test_client_request.py
yuvipanda: Removed

tests/test_client_request.py
yuvipanda: Removed

tests/test_client_request.py
yuvipanda: Renamed test case

tests/test_client_request.py
yuvipanda: I'll split it up.

tests/test_helpers.py
yuvipanda: Given that it is only used twice, I'd *love* to keep it as is. My brain already hurts from the amount of indirection going on here with the parametrization, I'd love to let this one be.

tests/test_helpers.py
yuvipanda: GitHub had automatically marked this as resolved before, I am not sure why. I've split this up into two tests named after your suggestions

tests/test_helpers.py
yuvipanda: Done

tests/test_client_request.py
yuvipanda: Split

docs/client_reference.rst
yuvipanda: I've added an `..envvar` directive there, but it is now at the bottom of the page rather than in line. I can instead switch the `:envvar:` here to `:term:` and just put `NETRC` into the glossary if you would prefer

yuvipanda:
I believe I've addressed everything. You might still have opinions on https://github.com/aio-libs/aiohttp/pull/7131#discussion_r1071693726 and https://github.com/aio-libs/aiohttp/pull/7131/files/4eadd65dcf102cd523aa40812fb1fafe61e4e4a4#r1071666671 @webknjaz

yuvipanda:
I see two relevant errors perhaps in `make mypy`:

```
aiohttp/helpers.py:273: error: Argument 1 to "BasicAuth" has incompatible type "Optional[str]"; expected "str"  [arg-type]
aiohttp/client_reqrep.py:443: error: Argument 2 to "basicauth_from_netrc" has incompatible type "Optional[str]"; expected "str"  [arg-type]
```

Both of these are things we guard against with an `if` and an *intermediate variable*, but mypy doesn't seem to detect that. I've this PR to my PR open that fixes these https://github.com/yuvipanda/aiohttp/pull/3.

However, these intermediate bools were added at your explicit request, @webknjaz (see https://github.com/aio-libs/aiohttp/pull/7131#discussion_r1053328611). I personally find what mypy want, without intermediate boolean variables, a lot more readable as well. However, at this juncture I'll basically do whatever it is that you want to get this PR merged - so if you can point me to a way to make mypy happy while also keeping your intermediate variables I'll do so :)

Dreamsorcerer:
> However, these intermediate bools were added at your explicit request, @webknjaz (see [#7131 (comment)](https://github.com/aio-libs/aiohttp/pull/7131#discussion_r1053328611)). I personally find what mypy want, without intermediate boolean variables, a lot more readable as well.

Yeah, I don't think the second one improves readability at all. The first one is also scheduled for removal in a few years anyway, so I personally am not bothered about keeping them.

yuvipanda:
@Dreamsorcerer i've hence merged in https://github.com/yuvipanda/aiohttp/pull/3 so mypy should pass now!

yuvipanda:
yay, all tests seem to pass now!

yuvipanda:
Just wanted to check-in on this again, @Dreamsorcerer and @webknjaz :)

docs/glossary.rst
```diff
@@ -153,3 +155,16 @@
       A library for operating with URL objects.
 
       https://pypi.python.org/pypi/yarl
+
+
+Enviornment Variables
```
Dreamsorcerer:
```suggestion
Environment Variables
```

tests/test_helpers.py
```diff
@@ -975,3 +975,66 @@ def test_populate_with_cookies():
...
+        ),
+        (
+            "machine example.com password pass\n",
+            helpers.BasicAuth("", "pass"),
+        ),
+    ],
+    indirect=("netrc_contents",),
+)
+@pytest.mark.usefixtures("netrc_contents")
+def est_basicauth_present_in_netrc(
```
yuvipanda:
```suggestion
def test_basicauth_present_in_netrc(
```

yuvipanda:
The windows test failures seem unrelated? And the missing code coverage result is due to differences in python versions, so I hope that's ok.

docs/glossary.rst
yuvipanda: Whoops, thanks for catching this, @Dreamsorcerer

yuvipanda:
This time there's a macos test failure :) So i'd imagine those are flakes.

the codecov warning is one missed line, and that's due to differing behavior in versions of python. I hope that's ok to ignore.

Dreamsorcerer:
> This time there's a macos test failure :) So i'd imagine those are flakes.

Not flaky, they just don't work yet. If check has passed, then that means the tests are passing correctly.

> the codecov warning is one missed line, and that's due to differing behavior in versions of python. I hope that's ok to ignore.

Hmm, that's not right. We run on Python 3.7-3.11, so if there was test coverage, it would have run. However, that code doesn't seem important and will be removed in a future version, so I'm not going to bother investigating.

Dreamsorcerer:
If @webknjaz is happy to resolve his comments, then I'll merge it in.

yuvipanda:
> We run on Python 3.7-3.11, so if there was test coverage, it would have run.

Ah, hmm. I'm not sure why that's happening then :|



> However, that code doesn't seem important and will be removed in a future version, so I'm not going to bother investigating.

I agree wholeheartedly!

yuvipanda:
I'm going on vacation from tuesday for 15 days, and wanted to check-in and see if there is anything I can do before I go :)

yuvipanda:
Just another gentle ping, as I slowly pick up back from vacation @Dreamsorcerer @webknjaz :)

webknjaz:
> However, at this juncture I'll basically do whatever it is that you want to get this PR merged - so if you can point me to a way to make mypy happy while also keeping your intermediate variables I'll do so :)

I suppose you're talking about this [`524c04d` (#7131)](https://github.com/aio-libs/aiohttp/pull/7131/commits/524c04d61a3a4384f314df23bbd5445604678fc7). I don't like complex clauses but if this bothers mypy, it's probably fine to do what you did for now. Wonder if explicit annotations on the vars would help...

webknjaz (approved):

webknjaz:
@yuvipanda looks like manual cherry-picking is needed

Issue #6383, opened 2021-12-06
support setting client network interface by name


SobhanMP:
### Is your feature request related to a problem?

While setting the target interface's ip should work in most cases, it's possible that multiple devices have the same ip.

 Also i'm having trouble using a wireguard vpn when i set Table=off. Table=off disables setting up the routes for the vpn and it justs shows up as a regular device. The connection is fine as pycurl can access it when i set the device but aiohttp fails. probably related but not directly.

### Describe the solution you'd like

Users should be able to set the inet device directly.

### Describe alternatives you've considered

UnixConnector seems like a good starting place. 

loop.create_connection with (socket.SOL_SOCKET, socket.SO_BINDTODEVICE, iface) [inspired from here](https://stackoverflow.com/a/68906789/9181037) also seems like a good idea

### Related component

Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

asvetlov:
Interesting idea, thanks!

webknjaz:
> ### Related component
> Client

What you're describing sounds like a server-side thing, no?

SobhanMP:
no?

i'm trying to do request with `aiohttp.ClientSession`. by definition that's a client side thing, right?

Muno459:
> > ### Related component
> > Client
> 
> What you're describing sounds like a server-side thing, no?

Not exactly. You can have multiple interfaces connected to a linux machine all with the same subnet, and DHCP ranges. Adding the abillity to choose exactly what interface to be used for the communication adds a lot of flexbility.

I have created a workaround, but i believe this should be implemented natively. Please note that `socket.SO_BINDTODEVICE` is only a valid attribute if the OS supports it. Futhermore chosing the source ip and a interface should not be possible if this was to be implemented. 

I have not investigated how the workaround works with proxies, but should not be a problem.


```python
async def _wrap_create_connection(
        self,
        *args: Any,
        req: "ClientRequest",
        timeout: "ClientTimeout",
        client_error: Type[Exception] = ClientConnectorError,
        **kwargs: Any,
) -> Tuple[asyncio.Transport, ResponseHandler]:
    try:
        async with aiohttp.helpers.ceil_timeout(timeout.sock_connect):
            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
            s.setsockopt(socket.SOL_SOCKET, socket.SO_BINDTODEVICE, 'en0'.encode())
            # nonblocking as we are async
[...]

PR #7132, opened 2022-12-17
Binding to network interfaces

Muno459:
## What do these changes do?

Finally added support to bind directly to network interfaces.

No, a optional keyword argument has been added. This will allow end-users to specify the network interface, a feature many related projects lacks without monkey-patching.

## Related issue number

#6383 
## Checklist


aiohttp/connector.py
```diff
@@ -770,6 +774,13 @@ def __init__(
         self._throttle_dns_events: Dict[Tuple[str, int], EventResultOrError] = {}
         self._family = family
         self._local_addr = local_addr
+        self._network_interface = network_interface
+        if self._local_addr and self._network_interface:
+            raise ValueError(
+                "local_addr and network_interface can't be defined at the same time."
```
Dreamsorcerer: Why are these incompatible? Couldn't you just pass the local_addr arguments to .connect_ex() (line 986)?

aiohttp/connector.py
```diff
@@ -719,6 +720,8 @@ class TCPConnector(BaseConnector):
     ttl_dns_cache - Max seconds having cached a DNS entry, None forever.
     family - socket address family
     local_addr - local tuple of (host, port) to bind socket to
+    network_interface - String of the network interface to bind socket to.
```
Dreamsorcerer: If you want it to appear in the documentation, you'll need to update docs/reference as well.

It would also be a good idea to give an example. By network interface, do you mean e.g. "eth0"?

Dreamsorcerer: In which case, it may be better as:
```suggestion
    network_interface - Name of network interface to bind socket to (e.g. "eth0").
```

aiohttp/connector.py
Muno459: No, local_addr is the source ip. You don't wanna connect to yourself. The reason why they are incompatible is because we have created our own socket. We can still respect the original local_addr values by adding 
```python 
socket.bind((ipaddr, port)) 
```
but that is not necessary because you **should** only bind to the IP the interface has. The reason you would bind to interface is because if you have multiple internet connections all with same DHCP range e.g 192.168.1.0/24, then you would not have control over the routing of the packet, thats why it's much more acceptable to bind to the interface rather the local ip.

aiohttp/connector.py
Mostafa-Mahdi: Agreed.

aiohttp/connector.py
Muno459: However, i agree the user should have full control and for future compatibility I will remove the check and allow binding.

Muno459: I have allowed support for custom source ip on custom nic and removed the check.

aiohttp/connector.py
Muno459: I have added examples, found the docs and changed that as well.

Dreamsorcerer:
Looks good to me, though I'm not too familiar with it. @webknjaz Want to double check it?

aiohttp/connector.py
```diff
@@ -770,6 +774,9 @@ def __init__(
         self._throttle_dns_events: Dict[Tuple[str, int], EventResultOrError] = {}
         self._family = family
         self._local_addr = local_addr
+        self._network_interface = network_interface
+        if self._network_interface and not hasattr(socket, "SO_BINDTODEVICE"):
```
webknjaz: I'd put the feature-detection into a dedicated *named* module-level constant like
```python
SOCKET_SUPPORTS_BINDING_TO_DEVICE = hasattr(socket, "SO_BINDTODEVICE")
```
Maybe even store that in the `aiohttp.helpers` module.

aiohttp/connector.py
```diff
@@ -770,6 +774,9 @@ def __init__(
...
+            raise OSError("Binding to interface is not supported by your OS.")
```
webknjaz: Isn't it more of a `RuntimeError` for us? Or rather its subclass `NotImplementedError`? `OSError` tends to represent low-level syscall interactions, while the current check is more of an assertion happening closer to the application level.
@Dreamsorcerer WDYT?

aiohttp/connector.py
```diff
@@ -964,6 +971,18 @@ async def _wrap_create_connection(
             async with ceil_timeout(
                 timeout.sock_connect, ceil_threshold=timeout.ceil_threshold
             ):
+                if self._network_interface:
+                    kwargs["sock"] = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
webknjaz: It's best to prepare the socket through a dedicated variable and only then assign it. Moreover, I'd move the init+connect to a helper function and just assign the result here.

aiohttp/connector.py
```diff
@@ -964,6 +971,18 @@ async def _wrap_create_connection(
...
+                    kwargs["sock"].setsockopt(
+                        socket.SOL_SOCKET,
+                        socket.SO_BINDTODEVICE,
+                        self._network_interface.encode(),
+                    )
+                    if self._local_addr:
+                        kwargs["sock"].bind(self._local_addr)
+                    kwargs["sock"].setblocking(False)
+                    kwargs["sock"].connect_ex((args[-2], args[-1]))
+                    args = args[:-2]  # only send socket instead host/ip
```
webknjaz: This feels like sketchy hacking. At least, use a named var to split this like
```python
*args, sock_hostname, sock_port = args
```
and use pass those vars to the socket initialization helper.

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
     conn = aiohttp.TCPConnector(resolver=resolver)
 
 
+Changing source interface
+^^^^^^^^^^^^^^^^^^^
```
webknjaz: The underline length must match the title text length.

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
...
+
+By default, if you have multiple internet connections, aiohttp will choose the
```
webknjaz: I don't think it's correct to imply that it's `aiohttp` that chooses the route, it's not true — the OS kernel chooses it. Moreover, even if you bind to an interface, I'm pretty sure this can still be intercepted via something like iptables and rerouted outside the application.
So let's try to rephrase this to be accurate in what actually happens.

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
...
+  conn = aiohttp.TCPConnector(network_interface="eth0")
+  session = aiohttp.ClientSession(connector=conn)
+
+
+If you get a message that your OS does not support the function, you may
+specify the source IP instead::
+
+  conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", 0))
+  session = aiohttp.ClientSession(connector=conn)
 Unix domain sockets
```
webknjaz, line 475: Don't forget about the paragraph separator empty line.
```suggestion

Unix domain sockets
```

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
...
+Changing source interface
+^^^^^^^^^^^^^^^^^^^
+
+By default, if you have multiple internet connections, aiohttp will choose the
+default route, to change the source interface, you'll have to specifiy it.
+This requires that you OS supports it.
+:class:`~aiohttp.TCPConnector`::
+
+  conn = aiohttp.TCPConnector(network_interface="eth0")
+  session = aiohttp.ClientSession(connector=conn)
```
webknjaz: Follow the code block indentation of the rest of the document.
```suggestion
    conn = aiohttp.TCPConnector(network_interface="eth0")
    session = aiohttp.ClientSession(connector=conn)
```

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
...
+
+
+If you get a message that your OS does not support the function, you may
+specify the source IP instead::
+
+  conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", 0))
+  session = aiohttp.ClientSession(connector=conn)
```
webknjaz:
```suggestion
    conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", 0))
    session = aiohttp.ClientSession(connector=conn)
```

docs/client_advanced.rst
webknjaz: Ah, that's not entirely correct. https://stackoverflow.com/a/68906789/595220 says that this would guarantee that the traffic would go out of the said interface. But still, it's not routing since routing is an L3 concept and interfaces are lower on the OSI stack.

aiohttp/connector.py
```diff
@@ -738,6 +741,7 @@ def __init__(
         family: int = 0,
         ssl: Union[None, bool, Fingerprint, SSLContext] = None,
         local_addr: Optional[Tuple[str, int]] = None,
+        network_interface: Optional[str] = None,
```
webknjaz, line 750: Looking at this, it seems like an overkill to add an extra argument (right?). I'd reuse `local_addr` since it'd be mutually exclusive with the `network_interface` anyway, right?

cURL does the same, it seems — they have an `--interface` CLI option that accepts the interface definition in both formats — name or IP: https://everything.curl.dev/usingcurl/connections/interface.

webknjaz: Ah... Looking at dnsmasq, it seems like there's a legit use case for accepting both local address and network interface.
I wonder, though, is it possible to pass an IP address for `network_interface`?

docs/faq.rst
```diff
@@ -210,8 +210,14 @@ be used used to bind a socket locally::
...
 
-.. seealso:: :class:`aiohttp.TCPConnector` and ``local_addr`` parameter.
+If your OS supports it, you may also bind directly to the interface ::
 
+    conn = aiohttp.TCPConnector(network_interface="eth0")
+    async with aiohttp.ClientSession(connector=conn) as session:
+        ...
+
+
+.. seealso:: :class:`aiohttp.TCPConnector`, ``local_addr`` and ``network_interface`` parameter.
```
webknjaz:
```suggestion
.. seealso::
   ``local_addr`` and ``network_interface`` arguments of :class:`~aiohttp.TCPConnector`.
```

webknjaz:
In general, this seems mostly acceptable, except for the comments I've made and missing tests. Also, the CI must pass before it's mergeable, including linting.

aiohttp/connector.py
Muno459: The best way would be to have two different arguments. Because `local_addr` requires a tuple of `(ip, port)`. Therefore, I think it's best to let it be.

docs/client_advanced.rst
Muno459: I will rephrase that.

docs/client_advanced.rst
Muno459: My bad, first time writing docs in this RST format. Will fix that

aiohttp/connector.py
Muno459: Looks cleaner agreed. I was unsure what the overall coding style was for aiohttp, so i tried to make code that's uses atleast resources / inexpensive as possible.

aiohttp/connector.py
Muno459: You are right.

> This exception is raised when a system function returns a system-related error.

We could also make our own exception, but depends on what you see fit.

Muno459:
I looked through your comments, and will create a new commit. In regards to tests, I'm not sure how I would implement that as it's highly specific for each system.  In regards to the CI check failing, the check is failing from other files that hasn't been changed in this PR.

Muno459:
I have futhermore fixed the socket family to be used from the request and not be fixed to `AF_INET` which should add support for IPv6, futhermore I removed the `connect_ex` call and replaced it with async `loop.sock_connect`. I'm not sure how it worked before, but i tested the `socket.connect_ex` function just for fun on macOS, and it was throwing errors that the socket wasn't connected unless i set it blocking, but on my Ubuntu machine it was working fine. Anyways, after awaiting, it worked out just fine.

Muno459:
I noticed my latest commit has introduced a linting issue, however i'm unsure what caused it. The typing used for `local_addr` is the same one defined in the built-in library socket.

webknjaz:
> i'm unsure what caused it.

You used lowercase `tuple` instead of the imported capitalized version. That would only work on Python 3.10+ IIRC. To make it work the same in the earlier versions, you could add `from __future__ import annotations` and it should satisfy the linter because mypy runs in a separate runtime but the future import will turn the annotations into regular strings making the regular runtime ignore what's inside.

webknjaz:
Also, there are typing problems present on master, they need to be addressed separately so they don't pollute the checks in this PR with confusing unrelated things.

Dreamsorcerer:
> you could add `from __future__ import annotations` and it should satisfy the linter because mypy runs in a separate runtime but the future import will turn the annotations into regular strings making the regular runtime ignore what's inside.

My understanding is that this merely delays evaluation of annotations. If someone wants the type hints at runtime (e.g. with `get_type_hints()`) then you'd still hit an error at that point. So, I think it's best we stick to `Tuple` (pyupgrade will presumably replace all instances once we drop support for 3.8).

Muno459:
I made it uppercase now, but it's still causing a linting issue.

aiohttp/connector.py
```diff
@@ -964,6 +975,15 @@ async def _wrap_create_connection(
             async with ceil_timeout(
                 timeout.sock_connect, ceil_threshold=timeout.ceil_threshold
             ):
+                if self._network_interface:
+                    *args, sock_hostname, sock_port = args
```
Dreamsorcerer: That's a little weird. mypy seems to think `args` on the right-hand side is `List`, but `args` on the left-hand side is `Tuple` (unless I'm reading that error wrong)...

One solution is to assign to a new variable name, but looks like that'd make the code more complex. I'm fine with a type ignore here. Maybe we should file a mypy bug too (or check if one already exists).

Dreamsorcerer: Actually, `*args` creates a list here, while `*args` as a parameter creates a tuple. That's why it's complaining.

Muno459: I will do a type ignore. I'm a novice python coder, so bear with me.

Muno459: Ohh.. and how should it be mitigated?

aiohttp/connector.py
```diff
@@ -964,6 +975,15 @@ async def _wrap_create_connection(
...
+                if self._network_interface:
+                    *args, sock_hostname, sock_port = args  # type: ignore
+                    s = connect_socket_with_interface(
+                        self._network_interface,
+                        family=kwargs["family"],
+                        local_addr=self._local_addr,
+                    )
+                    await self._loop.sock_connect(s, (sock_hostname, sock_port))
+                    kwargs["sock"] = s
                 return await self._loop.create_connection(*args, **kwargs)  # type: ignore[return-value]  # noqa
```
Dreamsorcerer:
```suggestion
                conn_args: Sequence[Any]
                if self._network_interface:
                    *conn_args, sock_hostname, sock_port = args
                    s = connect_socket_with_interface(
                        self._network_interface,
                        family=kwargs["family"],
                        local_addr=self._local_addr,
                    )
                    await self._loop.sock_connect(s, (sock_hostname, sock_port))
                    kwargs["sock"] = s
                else:
                    conn_args = args
                return await self._loop.create_connection(*conn_args, **kwargs)  # type: ignore[return-value]  # noqa
```

Could be one possible approach. I'm not sure if there's an easy way to just declare the type of `args` itself, as the `tuple` part is inferred and not available.

Muno459: I think it's more simple to just pass `(args[-2], args[-1])` to sock_connect, no? and redefine the tuple, like my first implementation.

webknjaz:
> My understanding is that this merely delays evaluation of annotations.

Not exactly. What's happening here is a combination of factors. So let me try to explain various aspects a bit deeper.
1) Type annotations can be specified in two ways:
    `var1: type1` — the AST node for `type1` is an object that must be previously imported and available at runtime
    `var1: 'type1'` — the AST node for `'type1'` is an `'str'` literal
    For the second case, it won't get `eval()`ed at import time like the first one. Both cases will produce metadata available in runtime via the `__annotations__` variable/attribute in the local scope.
    ```console
    $ ipython3                                                                        
    Python 3.9.0 (default, Dec  9 2022, 20:45:53) 
    Type 'copyright', 'credits' or 'license' for more information
    IPython 7.19.0 -- An enhanced Interactive Python. Type '?' for help.

    In [1]: var: asdf
    ---------------------------------------------------------------------------
    NameError                                 Traceback (most recent call last)
    <ipython-input-1-71b1d907417d> in <module>
    ----> 1 var: asdf

    NameError: name 'asdf' is not defined

    In [2]: __annotations__
    Out[2]: {}

    In [3]: var: 'asdf'

    In [4]: __annotations__
    Out[4]: {'var': 'asdf'}

    In [5]: var: str

    In [6]: __annotations__
    Out[6]: {'var': str}
    ```
    So, as we see, what you type in ends up in `__annotations__` exactly as you specify it. Thanks to `tuple` having the `__slice__()` method in Python 3.10+, you can use `tuple[thing]` in that runtime. In older runtimes, only `typing.Tuple` has it.
2) Now, there's [PEP 563](https://peps.python.org/pep-0563/) that was supposed to become the default behavior in Python 3.10 but got a strong pushback from the Pydantic community very last-minute, about 3 weeks before the release. Turned out that even though it was planned like 4 years before that, many folks didn't realize the impact. It created a lot of noise on Twitter and GitHub, the SC had to get involved and ended up reverting the change: https://github.com/pydantic/pydantic/issues/2678#issuecomment-823569522 / https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/. [PEP 649](https://peps.python.org/pep-0649/) which is currently a draft, was created to try to mitigate the concerns raised: https://dev.to/tiangolo/the-future-of-fastapi-and-pydantic-is-bright-3pbm#pep-649-deferred-evaluation-of-annotations-using-descriptors.
What was it about? This feature, that is currently only enabled via `from __future__ import annotations`, makes it so that the parser turns anything you type in as an annotation into strings — that `var: asdf` example above would become an equivalent of `var: 'asdf'`. This is exactly what's happening with my proposal to add an import — it basically stops the interpreter from looking up any references you may have in your annotations, since they are all just strings in runtime now.
3) With that in mind, my `'tuple[thing]'` **does not** become `tuple[thing]` in runtime and therefore, it won't cause any problems with linting and testing for us.
4) When `typing.TYPE_CHECKING is True`, though, it's a different story. MyPy does evaluate those strings to verify that the referenced objects exist. But it also includes a couple of shims that make it able to evaluate “syntax” from newer Pythons which is how `tuple[thing]` would work in MyPy regardless of the target Python version.

aiohttp/connector.py
webknjaz: Maybe use `conn_args: Sequence[Any] = args` before the if-block and let it rewrite this value conditionally. No else-block needed.

webknjaz:
> If someone wants the type hints at runtime (e.g. with `get_type_hints()`) then you'd still hit an error at that point.

You're right, that's a valid concern.

aiohttp/connector.py
webknjaz: > more simple to just pass `(args[-2], args[-1])` to sock_connect, no?

That does not contribute to the readability/maintainability of the code base. It shouldn't be only understood by a person adding changes in the first 5 minutes, but also 5 years from now. Things like `args[-2]` are confusing and don't have a named label to suggest what's the semantic meaning of the thing being passed. We could've made calls like `thing(x, y, z)` but tell be what this means instantly without having to inspect what `x`, `y` and `z` represent without reading another 100 lines of code that may have complex transformations too.
Naming pieces of data in code is beyond important — it allows humans to reason about the higher-level logic without an unnecessary level of mental overhead which increases the maintenance burden because of unreadable code.

webknjaz: @Mostafa-Mahdi note that you can use GH `Commit suggestion` buttons to make sure you credit the co-authors of the patch. https://hynek.me/til/easier-crediting-contributors-github/ explains why this is important.

CHANGES/6383.feature
```diff
@@ -0,0 +1 @@
+Added ``network_interface`` parameter to ``TCPConnector`` to allow binding directly to an interface.
```
webknjaz: In our project, you are also welcome to have a byline in the changelog, like this:
```suggestion
Added the ``network_interface`` parameter to :class:`~aiohttp.TCPConnector` to allow binding it directly to a source interface — by :user:`Mostafa-Mahdi`.
```
Also, let's try to link `TCPConnector` from the changelog to make it clickable. You can preview this after making the change via https://aiohttp--7132.org.readthedocs.build/en/7132/changes.html#features.

aiohttp/connector.py
```diff
@@ -58,18 +66,15 @@
     ssl = None  # type: ignore[assignment]
     SSLContext = object  # type: ignore[misc,assignment]
 
-
```
webknjaz: Could you exclude the formatting changes from the PR? It's a bad tone to mix them with any behavior changes.

aiohttp/connector.py
```diff
@@ -719,6 +724,8 @@ class TCPConnector(BaseConnector):
     ttl_dns_cache - Max seconds having cached a DNS entry, None forever.
     family - socket address family
     local_addr - local tuple of (host, port) to bind socket to
+    network_interface - Name of network interface to bind socket to (e.g. "eth0").
+
```
webknjaz, line 731: It doesn't seem like there's a case for having a double-empty-line separator here.
```suggestion
```

aiohttp/connector.py
webknjaz: Yes, and it's best not to resolve conversations before the other party sees new comments. It's rather hard to find them when they are hidden, or even know those new comments exist if they are collapsed.

aiohttp/connector.py
```diff
@@ -770,6 +778,9 @@ def __init__(
         self._throttle_dns_events: Dict[Tuple[str, int], EventResultOrError] = {}
         self._family = family
         self._local_addr = local_addr
+        self._network_interface = network_interface
+        if self._network_interface and not SOCKET_SUPPORTS_BINDING_TO_DEVICE:
+            raise RuntimeError("Binding to interface is not supported by your OS.")
```
webknjaz: I think it's best to avoid referring to a “you” in exceptions. It's not clear who that is and how they relate to the app being run. Maybe this would be more accurate?
```suggestion
            raise RuntimeError("Binding to interface is not supported by the current runtime (kernel or OS).")
```

aiohttp/connector.py
```diff
@@ -770,6 +778,9 @@ def __init__(
         self._throttle_dns_events: Dict[Tuple[str, int], EventResultOrError] = {}
         self._family = family
         self._local_addr = local_addr
+        self._network_interface = network_interface
+        if self._network_interface and not SOCKET_SUPPORTS_BINDING_TO_DEVICE:
```
webknjaz, line 784: @Dreamsorcerer what are your thoughts on having this as a hard assertion vs emitting a warning and falling back to a regular `bind()`? I think that `curl` does that implicitly. OTOH, it's a CLI app and not a framework (not sure if libcurl behaves the same, though).

aiohttp/connector.py
```diff
@@ -964,6 +975,15 @@ async def _wrap_create_connection(
             async with ceil_timeout(
                 timeout.sock_connect, ceil_threshold=timeout.ceil_threshold
             ):
+                if self._network_interface:
+                    *args, sock_hostname, sock_port = args  # type: ignore
```
webknjaz: It is important to avoid adding ignores when it's possible to make proper fixes. But if ignoring is inevitable, it shouldn't include ignoring any possible future rule violations, but only the specific rule that needs to be suppressed.

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,19 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
             with suppress(ValueError):
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+# Addresses can be either tuples of varying lengths (AF_INET, AF_INET6,
+# AF_NETLINK, AF_TIPC) or strings (AF_UNIX).
```
webknjaz: Looks like this could be a part of the docstring. The use case for code comments is to explain motivation. The use case for docstrings is to explain implementation/expectations — treat these like docs for the developers, as they will also show up in REPL and code editors in places where the function is used.

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,19 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
...
+def connect_socket_with_interface(
+    network_interface: str,
+    family: int = socket.AF_INET,
+    local_addr: Optional[Union[Tuple[Any, ...], str]] = None,
+) -> socket.socket:
+    """Prepares a socket to asyncio"""
```
webknjaz: The docstrings must follow PEP 257. This means having a full stop after the title-sentence. Also, the title must be in an imperative mood ("Do smth", not "Does smth") — similar to Git commit best practices.
Additionally, the docstring should be accurate including what's the effect of calling the function — not an abstract action on an abstract thing.

aiohttp/connector.py
Dreamsorcerer: Hmm, I'm not sure. I think if I'm trying to bind to a specific interface, I probably want an error to tell me it's not going to happen. If I'd _prefer_ a specific interface, then a `try ... catch ...` to fall back would be fine, and it's explicit then too.

aiohttp/connector.py
Dreamsorcerer: Our mypy config disallows it anyway, it would've been sorted if we weren't fixing the code.

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,19 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
             with suppress(ValueError):
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+# Addresses can be either tuples of varying lengths (AF_INET, AF_INET6,
+# AF_NETLINK, AF_TIPC) or strings (AF_UNIX).
+def connect_socket_with_interface(
```
webknjaz: This name seems misleading. Connecting a socket implies that it's pre-existing while I see that it's not true — it creates a socket and configures it in a specific way. Just like variable names, function names should tell a story about the effect the function has. And because it's a function, it represents an action and not a static object, keep using an imperative verb at the beginning.

How about something like this?
```suggestion
def make_an_interface_bound_socket(
```
Does it convey the purpose? Does it advertise why a caller would want to make use of it? If yes, it's a good name. If not, what could be done to improve it?

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,19 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
...
+    network_interface: str,
+    family: int = socket.AF_INET,
+    local_addr: Optional[Union[Tuple[Any, ...], str]] = None,
+) -> socket.socket:
+    """Prepares a socket to asyncio"""
+    s = socket.socket(family, socket.SOCK_STREAM)
```
webknjaz: Please, use an actual name for the variable, not just some random letter.

docs/client_reference.rst
```diff
@@ -1126,6 +1126,10 @@ TCPConnector
    :param tuple local_addr: tuple of ``(local_host, local_port)`` used to bind
       socket locally if specified.
 
+   :param str network_interface: Name of the network interface used to bind
+      socket locally if specified (e.g. "eth0"). Note that this feature is not
```
webknjaz: I think, we can highlight the example as code:
```suggestion
      socket locally if specified (e.g. ``eth0``). Note that this feature is not
```

docs/faq.rst
```diff
@@ -200,7 +200,7 @@ and call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in
         web.run_app(app, host='localhost', port=8080)
 
 
-How do I make a request from a specific IP address?
+How do I make a request from a specific IP address or network interface?
 ---------------------------------------------------
```
webknjaz: Changing the title length, please align the underline too.

aiohttp/connector.py
webknjaz: Right, makes sense. @Mostafa-Mahdi would you mind incorporating this in the docs? A try-except example seems a natural for that FAQ doc you're already changing.

aiohttp/connector.py
webknjaz: A custom exception is probably an overkill here.

webknjaz:
@Mostafa-Mahdi since you mentioned being a novice, here's a curated collection of materials related to FOSS contributions and Python you may like:
* https://gist.github.com/webknjaz/b623cb5eb941ed17617afa831f4a7fb0#idiomatic-python
* https://gist.github.com/webknjaz/a5a4fb374b7579de827e6bedb93a5220
* https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79
* https://gist.github.com/webknjaz/a7362787a80067af8621a85a71746ca1
* https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3

(these are not specific to our project and can be applied pretty much universally)

aiohttp/connector.py
Muno459: Sure.

Muno459:
> @Mostafa-Mahdi since you mentioned being a novice, here's a curated collection of materials related to FOSS contributions and Python you may like:
> 
> * https://gist.github.com/webknjaz/b623cb5eb941ed17617afa831f4a7fb0#idiomatic-python
> * https://gist.github.com/webknjaz/a5a4fb374b7579de827e6bedb93a5220
> * https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79
> * https://gist.github.com/webknjaz/a7362787a80067af8621a85a71746ca1
> * https://gist.github.com/webknjaz/2aa9c7a43b51c1385260ff87e0cbb9e3
> 
> (these are not specific to our project and can be applied pretty much universally)

I appreciate you taking the time to link these. I will definitely read up on these.

aiohttp/helpers.py
Muno459: Agreed, and co-authored you for the suggestion 👍

aiohttp/connector.py
Muno459: Aplogies, these are caused by the IDE. I will revert it.

Muno459:
I believe I have addressed all the issues, futhermore found a fix to the tuple assignment without using type ignore :)

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,24 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
...
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+def make_an_interface_bound_socket(
+    network_interface: str,
+    family: int = socket.AF_INET,
+    local_addr: Optional[Union[Tuple[Any, ...], str]] = None,
+) -> socket.socket:
+    """Returns a socket, bounded to a network interface.
```
```suggestion
    """Return a socket, bounded to a network interface.
```

docs/faq.rst
```diff
@@ -210,8 +210,19 @@ be used used to bind a socket locally::
...
+If your OS supports it, you may also bind directly to the interface ::
+
+    try:
+        conn = aiohttp.TCPConnector(network_interface="eth0")
+        async with aiohttp.ClientSession(connector=conn) as session:
+            ...
+    except RuntimeError:
+        # fallback to local_addr as the runtime kernel is unsupported.
+        conn = aiohttp.TCPConnector(local_addr=('127.0.0.1', 0))
+        ...
```
Dreamsorcerer:
```suggestion
If your OS supports it, you may also bind directly to the interface ::

    conn = aiohttp.TCPConnector(network_interface="eth0")
    async with aiohttp.ClientSession(connector=conn) as session:
        ...

It is also possible to have a fallback by catching the :exc:`RuntimeError` on creation:

    try:
        conn = aiohttp.TCPConnector(network_interface="eth0")
    except RuntimeError:
        # fallback to local_addr as the runtime kernel is unsupported.
        conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", 0))

    async with aiohttp.ClientSession(connector=conn) as session:
        ...
```

docs/client_advanced.rst
```diff
@@ -451,6 +451,23 @@ In order to specify the nameservers to when resolving the hostnames,
     conn = aiohttp.TCPConnector(resolver=resolver)
 
 
+Changing source interface
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you wish to specify where the packets should originate from, you should
+specify the interface you wish to use. This requires that you OS supports it.
```
Dreamsorcerer:
```suggestion
specify the interface you wish to use.
Raises :exc:`RuntimeError` if the OS kernel does not support this feature.
```

Let's include the exception information.

aiohttp/connector.py
```diff
@@ -964,6 +980,15 @@ async def _wrap_create_connection(
             async with ceil_timeout(
                 timeout.sock_connect, ceil_threshold=timeout.ceil_threshold
             ):
+                if self._network_interface:
+                    args, sock_hostname, sock_port = args
```
Dreamsorcerer: This isn't right, it will error if `*args` is not of length 3.
If we can rely on it always having 3 items, then we should just turn them into positional parameters and drop this line entirely.

At a brief glance this does appear to be a private function that is always passed `factory, host, port`, so it may be a good option to change the parameter.

Muno459: Is this acceptable? This will allow more parameters in the future if they are added in the front.
```python
*_, sock_hostname, sock_port = args
args = args[:-2]
```

aiohttp/helpers.py
webknjaz: bounded->bound?

docs/faq.rst
webknjaz: @Dreamsorcerer you need to specify a role for the `RuntimeError` reference in Markdown. It may look like this:
```rst
:exc:`RuntimeError`
```

Without it, the role is `any` which essentially matches the first found thing and if it breaks, it may be non-trivial to an inexperienced person to debug.

webknjaz: @Mostafa-Mahdi explanation on Sam's suggestion: try-except blocks should always wrap only one instruction for which you'd like to process an exception. Wrapping more may cause unintended side effects that might appear in the future silently and would be hard to notice and track down.

docs/client_advanced.rst
webknjaz: @Dreamsorcerer I've edited your patch a bit.

docs/faq.rst
```diff
@@ -210,8 +210,25 @@ be used used to bind a socket locally::
...
         ...
 
-.. seealso:: :class:`aiohttp.TCPConnector` and ``local_addr`` parameter.
+If your OS supports it, you may also bind directly to the interface ::
+
+    conn = aiohttp.TCPConnector(network_interface="eth0")
+    async with aiohttp.ClientSession(connector=conn) as session:
+        ...
+
+It is also possible to have a fallback by catching the ``RuntimeError`` on creation:
```
webknjaz:
```suggestion
It is also possible to have a fallback by catching the :exc:`RuntimeError` on creation::
```
(the double trailing colon is needed for implicit code blocks)

docs/faq.rst
```diff
@@ -210,8 +210,25 @@ be used used to bind a socket locally::
     async with aiohttp.ClientSession(connector=conn) as session:
         ...
 
-.. seealso:: :class:`aiohttp.TCPConnector` and ``local_addr`` parameter.
+If your OS supports it, you may also bind directly to the interface ::
```
webknjaz:
```suggestion
If your OS supports it, you may also bind directly to the interface::
```

docs/faq.rst
```diff
@@ -210,8 +210,25 @@ be used used to bind a socket locally::
...
+It is also possible to have a fallback by catching the ``RuntimeError`` on creation:
+
+    try:
+        conn = aiohttp.TCPConnector(network_interface="eth0")
+    except RuntimeError:
+        # fallback to local_addr as the runtime kernel is unsupported.
+        conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", 0))
+
+    async with aiohttp.ClientSession(connector=conn) as session:
+        ...
```
webknjaz: I'd merge two code examples into one and have a combined paragraph with common explanation.

aiohttp/connector.py
webknjaz: Use an actual name for the data. `args` is too generic. Also, it's better to have an atomic instruction rather than multiple, manipulating unidentified bits of information.

aiohttp/helpers.py
```diff
@@ -727,7 +729,6 @@ def ceil_timeout(
 
 
 class HeadersMixin:
```
```suggestion
class HeadersMixin:

```

aiohttp/helpers.py
```diff
@@ -1038,3 +1039,24 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
             with suppress(ValueError):
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+def make_an_interface_bound_socket(
+    network_interface: str,
+    family: int = socket.AF_INET,
+    local_addr: Optional[Union[Tuple[Any, ...], str]] = None,
```
webknjaz, line 1049: Shouldn't this be
```suggestion
    local_addr: Optional[Union[Tuple[str, int], str]] = None,
```
?

docs/client_advanced.rst
```diff
@@ -451,6 +451,24 @@ In order to specify the nameservers to when resolving the hostnames,
     conn = aiohttp.TCPConnector(resolver=resolver)
 
 
+Changing source interface
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you wish to specify where the packets should originate from, you should
+specify the interface you wish to use.
+Raises ``RuntimeError`` if the OS does not support this feature.
```
webknjaz:
```suggestion
Raises :exc:`RuntimeError` if the OS does not support this feature.
```

aiohttp/helpers.py
Muno459: Socket library has it like that, presumably because there are varying lengths to different socket families

aiohttp/connector.py
Muno459: I decided to go back to your original code suggestion and adding the specific type ignore rule. I have geninuely no idea how to make it more clean and atomic (without introducing unnecessary if-statements and extra vars)  than what I suggested til this point.

docs/faq.rst
```diff
@@ -200,18 +200,25 @@ and call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in
...
+------------------------------------------------------------------------
 
 If your system has several IP interfaces, you may choose one which will
-be used used to bind a socket locally::
+be used used to bind a socket locally, either by specifying the source IP, or
+source network interface. It is also possible to have a
+fallback by catching the :exc:`RuntimeError` on creation::
+
+    try:
+        conn = aiohttp.TCPConnector(network_interface="eth0")
```
webknjaz, line 212: Should we use the `local_addr` argument here too so it's closer to the fallback?

docs/faq.rst
```diff
@@ -200,18 +200,25 @@ and call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in
...
+    except RuntimeError:
+        # fallback to local_addr as the runtime kernel is unsupported.
```
webknjaz:
```suggestion
        # Fall back to local_addr when the runtime kernel does not support interface binding.
```

docs/faq.rst
```diff
@@ -200,18 +200,25 @@ and call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in
         web.run_app(app, host='localhost', port=8080)
 
 
-How do I make a request from a specific IP address?
----------------------------------------------------
+How do I make a request from a specific IP address or network interface?
+------------------------------------------------------------------------
 
 If your system has several IP interfaces, you may choose one which will
-be used used to bind a socket locally::
```
webknjaz, line 207: Apparently, both the original and the updated text have the double `used used` bit.

docs/faq.rst
```diff
@@ -200,18 +200,25 @@ and call :meth:`aiohttp.web.WebSocketResponse.close` on all of them in
...
+be used used to bind a socket locally, either by specifying the source IP, or
```
webknjaz:
```suggestion
be used to bind a socket locally, either by specifying the source IP, or
```

docs/client_advanced.rst
```diff
@@ -451,6 +451,24 @@ In order to specify the nameservers to when resolving the hostnames,
     conn = aiohttp.TCPConnector(resolver=resolver)
 
 
+Changing source interface
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you wish to specify where the packets should originate from, you should
+specify the interface you wish to use.
+Raises :exc:`RuntimeError` if the OS does not support this feature.
+:class:`~aiohttp.TCPConnector`::
```
webknjaz: This looks like a disconnected reference that got separated from some sentence ^^^

webknjaz:
```suggestion
specify the interface you wish to use when initializing a :class:`~aiohttp.TCPConnector` instance.

.. warning::
   This may cause a :exc:`RuntimeError` if the OS kernel does not support this feature.
   
Here's an example of how to do this::
```

docs/client_advanced.rst
```diff
@@ -451,6 +451,24 @@ In order to specify the nameservers to when resolving the hostnames,
...
+
+    conn = aiohttp.TCPConnector(network_interface="eth0")
+    session = aiohttp.ClientSession(connector=conn)
+
+
+If your OS does not support the function, you may specify
```
webknjaz:
```suggestion
If the runtime OS does not support the function, you may specify
```

docs/client_advanced.rst
webknjaz:
```suggestion
If it is necessary to specify where the packets should originate from, it is possible to
set a specific network interface when initializing the :class:`~aiohttp.TCPConnector` instance.

.. warning::
   This may cause a :exc:`RuntimeError` if the OS kernel does not support this feature.
   
Here's an example of how to do this::
```

aiohttp/helpers.py
```diff
@@ -80,7 +81,6 @@
     not sys.flags.ignore_environment and bool(os.environ.get("PYTHONASYNCIODEBUG"))
 )
```
```suggestion


```

aiohttp/connector.py
webknjaz: Resolving this thread.

aiohttp/connector.py
webknjaz: Resolving.

aiohttp/connector.py
```diff
@@ -964,6 +980,15 @@ async def _wrap_create_connection(
             async with ceil_timeout(
                 timeout.sock_connect, ceil_threshold=timeout.ceil_threshold
             ):
+                if self._network_interface:
+                    *args, sock_hostname, sock_port = args  # type: ignore[assignment]
```
webknjaz: Honestly, I'd prefer having a new named variable rather than some generic `args` that is not clear what it contains at different points.


@Dreamsorcerer WDYT?

docs/faq.rst
Dreamsorcerer: The original also duplicates the 2nd line unnecessarily.

aiohttp/connector.py
Dreamsorcerer: See the 2nd part of my previous comment: https://github.com/aio-libs/aiohttp/pull/7132#discussion_r1052622495

If you agree with my analysis, then we can just replace `*args: Any` from the parameter list and have 3 positional parameters instead. (line 973)

Muno459: So which one is it?

webknjaz: @Dreamsorcerer if we're confident that it's always 3 args, enumerating the m explicitly seems like a great idea.

Dreamsorcerer: Well, it's definitely only used in one place in our code. I'd only be concerned if a user subclasses this and then passes extra arguments or something. But, it's not documented anywhere or anything, so I'd assume people are not doing that.

So, I'd say it's safe to convert to fixed arguments.

Muno459:
Changed it to static

aiohttp/connector.py
```diff
@@ -954,7 +970,9 @@ def _get_fingerprint(self, req: "ClientRequest") -> Optional["Fingerprint"]:
 
     async def _wrap_create_connection(
         self,
-        *args: Any,
+        factory: functools.partial[ResponseHandler],
```
Dreamsorcerer, line 973: The actual requirement is this: https://github.com/python/typeshed/blob/main/stdlib/asyncio/base_events.pyi#L117

So, I'd suggest doing the same thing with `Callable[[], _ResponseHandlerT]`. Then we can also update the return type so this becomes generic.

Dreamsorcerer:
To fix the typing error, I'd suggest renaming the first `host` variable to `raw_host`:
```python
        raw_host = req.url.raw_host
        assert raw_host is not None
        port = req.port
        assert port is not None
        host_resolved = asyncio.ensure_future(
            self._resolve_host(raw_host, port, traces=traces), loop=self._loop
        )
```
https://github.com/aio-libs/aiohttp/blob/f0073afe3cc66f8cbc951522d3cacf46523bd818/aiohttp/connector.py#L1112-L1118

aiohttp/helpers.py
```diff
@@ -1038,3 +1041,24 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
...
+    family: int = socket.AF_INET,
+    local_addr: Optional[Union[Tuple[Any, ...], str]] = None,
+) -> socket.socket:
+    """Return a socket, bounded to a network interface.
+
+    Addresses can be either tuples of varying lengths
+    (AF_INET, AF_INET6, AF_NETLINK, AF_TIPC)
+    or strings (AF_UNIX).
+    """
+    sock = socket.socket(family, socket.SOCK_STREAM)
```
webknjaz, line 1057: It probably makes sense to add
```suggestion
    assert SOCKET_SUPPORTS_BINDING_TO_DEVICE
    sock = socket.socket(family, socket.SOCK_STREAM)
```

bertybuttface:
Sorry to bump an old PR but is this stale / is there another way to select a particular network interface when using aiohttp.ClientSession?

Dreamsorcerer:
Looks like the PR didn't get finished off. Feel free to fork it and create a new PR with the final changes.

Muno459:
Hey @Dreamsorcerer Exactly, what is needed to finish this PR?

Dreamsorcerer:
The tests are failing, there are conflicts, and there are unresolved conversations above.

sheldygg:
When this PR will be merged?

webknjaz:
When somebody addresses the requests above.

PR #7134, opened 2022-12-19
Fix CI

Dreamsorcerer:


.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.9
+        python-version: '3.9'
```
webknjaz: This is unnecessary, it's cleaner w/o quoting, plus the only exception is 3.10 anyway 
```suggestion
        python-version: 3.9
```

PR #7143, opened 2022-12-24
[3.9] Fix CI

Dreamsorcerer:


webknjaz (approved):

webknjaz:
Looks like new CI failures appeared

Dreamsorcerer:
@webknjaz Does XPASS mean the test is expected to fail, but has now started passing?

Not sure if it should just be updated, as ubuntu 3.11 passed, but have got this XPASS on ubuntu 3.10 and mac 3.11. Seems a bit flaky maybe..

Dreamsorcerer:
We have quite a bit of flakiness regardless at the moment. CI is failing 50+% of the time on master, and that Windows 3.11 test holds everything up as it seems to timeout after 15 mins.

Dreamsorcerer:
Hmm, these ones appear to relate to `ClientRequest.loop`. I'd be inclined to remove this attribute, and then remove those tests too.

webknjaz:
> Does XPASS mean the test is expected to fail, but has now started passing?

Yes, which might mean some changes in stdlib when GHA updates the Python patch version or the deps updates. But since there were no deps updates, I'd blame GH updating their VMs or Python builds.
Note that the `(strict)` marker means that the xpass will fail the test session. It exists so that if known bugs get fixed (sometimes accidently), this would become evident immediately.

webknjaz:
Hm, I didn't realize it's a PR against 3.9. So maybe you're right that the behavior might be not the same.

Dreamsorcerer:
I've just noticed these tests aren't in master. I don't think they really make sense either, e.g. `ClientRequest` shouldn't be instantiated by the user, and would only be done internally within a running loop, so a test that verifies instantiating it outside of a running loop seems pointless.

webknjaz:
The reason they may be missing in master might be that it might've added a breaking change, modifying the behavior, though. In this case, it'd be reasonable to have them in 3.9 but not in master.

Dreamsorcerer:
Hmm, seems that setting it to pass in <=3.10 works fine. No idea why that's suddenly changed in 3.10. I still think the tests are pointless, but if they're passing, might as well leave them for now.

PR #7150, opened 2022-12-28
client: fix upload timeouts with sock_read set

### LINKED ISSUE #7149: client: unexpected timeouts with chunked uploads when setting sock_read timeout
### Describe the bug

Uploading large files using `put`/`post` with chunked encoding AND setting `sock_read` timeout results in an unexpected timeout error. 

Traceback is provided below.

### To Reproduce

Server code (aiohttp):

```python
import asyncio
import aiohttp

PORT = 8472
HOST = "localhost"


async def handler(request: aiohttp.web.Request) -> aiohttp.web.StreamResponse:
    print(f"[{loop.time()}] got request")

    response = aiohttp.web.StreamResponse()
    response.enable_chunked_encoding()
    writer = await response.prepare(request)

    read_data = 0
    print(f"[{loop.time()}] started streaming body")

    async for data, _ in request.content.iter_chunks():
        read_data += len(data)
        print(f"[{loop.time()}] {read_data=}", end="\r")
        await asyncio.sleep(1e-4)    # NOTE: tweaking this and/or the above print, along with `sock_read` in the client might yield different results (successful upload instead of timeout)

    await writer.write_eof()
    print(f"\n[{loop.time()}] finished")

    print(f"[{loop.time()}] finished reading body {read_data=}")
    return response


async def main():
    """ listens and streams the body of a chunked transfer upload """

    server = aiohttp.web.Server(handler)
    await loop.create_server(server, HOST, PORT)
    print(f"Listening on http://{HOST}:{PORT}")

    try:
        while True:
            await asyncio.sleep(1)
    except KeyboardInterrupt:
        print("Shutting down")
        await server.shutdown(10)


if __name__ == "__main__":
    asyncio.run(main())
```

Client code:

```python
import asyncio
import aiohttp

PORT = 8472
HOST = "localhost"


SOCK_READ_TIMEOUT = 5 # tweaking this and/or the sleep in the server implementation might yield different results


async def gen_chunks(
    payload_size: int,
    chunk_size: int = 0x1000,  # 4k
):
    processed = 0
    while processed < payload_size:
        if (processed + chunk_size) > payload_size:
            # final chunk
            chunk_size = payload_size - processed
        yield b"Z" * chunk_size
        processed += chunk_size


async def main(payload_size: int, method: str = "put", **kwargs):
    url = f"http://{HOST}:{PORT}"
    remote_path = "file"

    timeout = aiohttp.ClientTimeout(
        total=None,
        connect=0,
        sock_connect=0,
        sock_read=SOCK_READ_TIMEOUT,
    )
    async with aiohttp.ClientSession(timeout=timeout) as session:
        meth = getattr(session, method)
        async with meth(
            f"{url}/{remote_path}",
            data=gen_chunks(payload_size),
            **kwargs,
        ) as resp:
            resp.raise_for_status()
            print(f"{resp.status=}, {await resp.read()}")


if __name__ == "__main__":
    payload_1gb = 1024 * 0x100000

    asyncio.run(main(payload_1gb))
```

### Expected behavior

Upload should succeed with no timeout.

In particular, the `ServerTimeoutError` exception is raised because we're waiting to read from the protocol while `StreamWriter` is still sending data.

When calling [`_request`](https://github.com/aio-libs/aiohttp/blob/7d1ada8ae598c61dcb087516358db3803ae105a6/aiohttp/client.py#L326), we call [`conn.protocol.set_response_params(..., read_timeout=real_timeout.sock_read, ...)`](https://github.com/aio-libs/aiohttp/blob/7d1ada8ae598c61dcb087516358db3803ae105a6/aiohttp/client.py#L515-L524) which sets up a timeout handler callback that fires in `real_timeout.sock_read`  seconds, although this is rescheduled whenever data is received ( see [`ResponseHandler.data_received`](https://github.com/aio-libs/aiohttp/blob/8e92c4c7c305f914c17f41539da20822ea7baddf/aiohttp/client_proto.py#L202), [`ResponseHandler._reschedule_timeout`](https://github.com/aio-libs/aiohttp/blob/8e92c4c7c305f914c17f41539da20822ea7baddf/aiohttp/client_proto.py#L185-L194). 

Since nothing will be read from the `ResponseHandler` until all of the chunks are sent to the server, any upload that takes longer than `sock_read` s to upload will fail with a timeout.

A simple workaround for this is to reschedule the read timer after each chunk is sent through `StreamWriter`, something like this:
```diff
diff --git a/aiohttp/http_writer.py b/aiohttp/http_writer.py
index db3d6a04..a5b241d8 100644
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -114,6 +114,7 @@ class StreamWriter(AbstractStreamWriter):
                 chunk = chunk_len_pre + chunk + b"\r\n"
 
             self._write(chunk)
+            self.protocol._reschedule_timeout()
 
             if self.buffer_size > LIMIT and drain:
                 self.buffer_size = 0
```

Or, if we wish to avoid calling `_reschedule_timeout` directly:

```diff
diff --git a/aiohttp/http_writer.py b/aiohttp/http_writer.py
index db3d6a04..7a274c56 100644
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -114,6 +114,7 @@ class StreamWriter(AbstractStreamWriter):
                 chunk = chunk_len_pre + chunk + b"\r\n"
 
             self._write(chunk)
+            self.protocol.data_received(b"")
 
             if self.buffer_size > LIMIT and drain:
                 self.buffer_size = 0
```
 

### Logs/tracebacks

```python-traceback
  File "aiohttp/timeouts_investigation.py", line 291, in <module>
    asyncio.run(main_client(payload_size))
  File "/usr/lib/python3.10/asyncio/runners.py", line 44, in run
  File "/usr/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "aiohttp/timeouts_investigation.py", line 140, in main_client
    async with meth(
  File "aiohttp/aiohttp/client.py", line 1141, in __aenter__
    self._resp = await self._coro
  File "aiohttp/aiohttp/client.py", line 560, in _request
    await resp.start(conn)
  File "aiohttp/aiohttp/client_reqrep.py", line 899, in start
  File "aiohttp/aiohttp/streams.py", line 616, in read
    await self._waiter
aiohttp.client_exceptions.ServerTimeoutError: Timeout on reading data from socket
```

### Python Version

```console
3.10.8
```

### aiohttp Version

```console
3.8.3
```

### multidict Version

```console
5.2.0
```

### yarl Version

```console
1.8.1
```

### OS

Linux

### Related component

Client

### Additional context

_No response_

### Code of Conduct

dtrifiro:
## What do these changes do?

Prevent the `sock_read` timeout callback from firing by only scheduling it afterthe payload (if any) has been fully written.

No

## Related issue number

Fixes #7149

## Checklist

### DIFF
--- a/CHANGES/7149.bugfix
+++ b/CHANGES/7149.bugfix
@@ -1,0 +1,1 @@
+changed ``sock_read`` timeout to start after writing has finished, to avoid read timeouts caused by an unfinished write. -- by :user:`dtrifiro`
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -673,6 +673,25 @@ async def test_read_timeout_on_reading_chunks(aiohttp_client: Any, mocker: Any) 
             await resp.content.read()
 
 
+async def test_read_timeout_on_write(aiohttp_client: Any) -> None:
+    async def gen_payload() -> AsyncIterator[str]:
+        # Delay writing to ensure read timeout isn't triggered before writing completes.
+        await asyncio.sleep(0.5)
+        yield b"foo"
+
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response(body=await request.read())
+
+    app = web.Application()
+    app.router.add_put("/", handler)
+
+    timeout = aiohttp.ClientTimeout(total=None, sock_read=0.1)
+    client = await aiohttp_client(app)
+    async with client.put("/", data=gen_payload(), timeout=timeout) as resp:
+        result = await resp.read()  # Should not trigger a read timeout.
+    assert result == b"foo"
+
+
 async def test_timeout_on_reading_data(aiohttp_client: Any, mocker: Any) -> None:
     loop = asyncio.get_event_loop()
 
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -9,7 +9,7 @@ import json
 import pathlib
 import socket
 import ssl
-from typing import Any
+from typing import Any, AsyncIterator
 from unittest import mock
 
 import pytest
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -157,7 +157,6 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         self._skip_payload = skip_payload
 
         self._read_timeout = read_timeout
-        self._reschedule_timeout()
 
         self._timeout_ceil_threshold = timeout_ceil_threshold
 
@@ -192,6 +191,9 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
             )
         else:
             self._read_timeout_handle = None
+
+    def start_timeout(self) -> None:
+        self._reschedule_timeout()
 
     def _on_read_timeout(self) -> None:
         exc = ServerTimeoutError("Timeout on reading data from socket")
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -551,6 +551,8 @@ class ClientRequest:
                 protocol.set_exception(exc)
         except Exception as exc:
             protocol.set_exception(exc)
+        else:
+            protocol.start_timeout()
         finally:
             self._writer = None
 
--- a/tests/test_client_proto.py
+++ b/tests/test_client_proto.py
@@ -109,12 +109,15 @@ async def test_empty_data(loop: Any) -> None:
 async def test_schedule_timeout(loop: Any) -> None:
     proto = ResponseHandler(loop=loop)
     proto.set_response_params(read_timeout=1)
+    assert proto._read_timeout_handle is None
+    proto.start_timeout()
     assert proto._read_timeout_handle is not None
 
 
 async def test_drop_timeout(loop: Any) -> None:
     proto = ResponseHandler(loop=loop)
     proto.set_response_params(read_timeout=1)
+    proto.start_timeout()
     assert proto._read_timeout_handle is not None
     proto._drop_timeout()
     assert proto._read_timeout_handle is None
@@ -123,6 +126,7 @@ async def test_drop_timeout(loop: Any) -> None:
 async def test_reschedule_timeout(loop: Any) -> None:
     proto = ResponseHandler(loop=loop)
     proto.set_response_params(read_timeout=1)
+    proto.start_timeout()
     assert proto._read_timeout_handle is not None
     h = proto._read_timeout_handle
     proto._reschedule_timeout()
@@ -133,6 +137,7 @@ async def test_reschedule_timeout(loop: Any) -> None:
 async def test_eof_received(loop: Any) -> None:
     proto = ResponseHandler(loop=loop)
     proto.set_response_params(read_timeout=1)
+    proto.start_timeout()
     assert proto._read_timeout_handle is not None
     proto.eof_received()
     assert proto._read_timeout_handle is None


dtrifiro:
TODO: add a tests

Dreamsorcerer:
Think you can write a test that reproduces the original problem?

tests/test_client_functional.py
```diff
@@ -673,6 +673,47 @@ async def handler(request):
             await resp.content.read()
 
 
+async def test_read_timeout_on_sending_chunks(aiohttp_client: Any) -> None:
```
dtrifiro: This test will fail without the changes in `StreamWriter.write`, `StreamWriter.write_eof`, as reported in #7149

Dreamsorcerer:
I'm wondering if this is the correct behaviour. If the timeout relates to reading data, shouldn't we start the timeout after we've finished writing? Or do we need to handle cases where reading happens simultaneously to writing?

Currently, it looks like we start writing as a task:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L621

Then we start the response immediately after, which initialises the timer:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L529

Also, wouldn't this issue also occur if you tried to send a large payload without chunking? I'm not sure this fix works in that case.

@webknjaz Thoughts?

dtrifiro:
> If the timeout relates to reading data, shouldn't we start the timeout after we've finished writing? Or do we need to handle cases where reading happens simultaneously to writing?

I also asked myself that question. I guess one of the use cases is to start reading the response while the file is being uploaded in order to catch potential errors early and abort the upload, although there's currently no mechanism for that: whole file is uploaded even if the server immediately returns an error (e.g. 401). In my limited testing, `curl` does this:
```bash
curl -X PUT -H "Transfer-Encoding: chunked" http://localhost:8472/file -T <large file>
```
Fails immediately if the server returns a 401:
```python
async def handler(request: aiohttp.web.Request) -> aiohttp.web.Response:
    return aiohttp.web.Response(status=401)
```
Also note that the sock read timeout is actually started before we send the request (via `_reschedule_timeout`):

https://github.com/aio-libs/aiohttp/blob/743c21f4f683e11ecef588a0955b433a3b3102e9/aiohttp/client_proto.py#L146-L160


--- 


> Also, wouldn't this issue also occur if you tried to send a large payload without chunking? I'm not sure this fix works in that case.

Good point, that's still broken, and the current fix is also broken when the time to upload a single chunk is comparable to the timeout value.

Another proposal to fix this could be something like this:
```diff
diff --git a/aiohttp/client.py b/aiohttp/client.py
index c4074577..8ceeeb09 100644
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -525,6 +525,9 @@ class ClientSession:
                     try:
                         try:
                             resp = await req.send(conn)
+                            conn.protocol.pause_reading()
+                            await resp.wait_for_close()
+                            conn.protocol.resume_reading()
                             try:
                                 await resp.start(conn)
                             except BaseException:
```

would fix the issue for all cases although I'm not sure it makes sense to completely pause reading while uploading.

Any thoughts?

Dreamsorcerer:
> I'm not sure it makes sense to completely pause reading while uploading.

I think you're probably right. I wonder if there's a reasonable way to only start the read timeout once the write has completed, even if we're already waiting on the read...

Dreamsorcerer:
If we can use that `resp.wait_for_closed()` and it's not too messy to get a reference to that in the right place, then maybe we can modify the timeout code.

e.g. The timeout might be modified to do something like:
```
async def timeout():
    await resp.wait_for_closed()
    await asyncio.sleep(...)
    task.cancel()
```

CHANGES/7149.bugfix
```diff
@@ -0,0 +1 @@
+fix chunked upload timeouts with sock_read timeout set
```
webknjaz: This needs to be formatted with RST. https://github.com/aio-libs/aiohttp/tree/master/CHANGES#alright-so-how-to-add-a-news-fragment.

Make sure it's consistent with other entries and describes the difference from the previous release.
Preview here: https://aiohttp--7150.org.readthedocs.build/en/7150/changes.html#bugfixes.

aiohttp/http_writer.py
```diff
@@ -114,6 +114,7 @@ async def write(
                 chunk = chunk_len_pre + chunk + b"\r\n"
 
             self._write(chunk)
+            self.protocol.data_received(b"")
```
webknjaz: How does this correspond to https://github.com/aio-libs/aiohttp/blob/master/aiohttp/_http_writer.pyx?

dtrifiro:
@Dreamsorcerer 

I think it might be easier to do something like [this](https://github.com/aio-libs/aiohttp/commit/78efee121e8b842d691c381d6686468f922a8767#diff-720910fd022b9ba8b7ddf0cb446ce5e1afe443f4287ad26a1eb4dc4b5b5f4508L527):

```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index 9478e8d3..a11a600f 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -525,6 +525,7 @@ class ClientRequest:
 
         protocol = conn.protocol
         assert protocol is not None
+        protocol.drop_timeout()
         try:
             if isinstance(self.body, payload.Payload):
                 await self.body.write(writer)
@@ -551,6 +552,8 @@ class ClientRequest:
                 protocol.set_exception(exc)
         except Exception as exc:
             protocol.set_exception(exc)
+        else:
+            protocol.reschedule_timeout()
         finally:
             self._writer = None
```

This works in every case we discussed above and does not block reading (note that I renamed `_drop_timeout` and `_reschedule_timeout`).

Somewhat related: if the total timeout is not set, this could result in hangs while uploading since there's no `write_timeout` (could be added to `StreamWriter`?)

Dreamsorcerer:
> This works in every case we discussed above and does not block reading (note that I renamed `_drop_timeout` and `_reschedule_timeout`).
> 
> Somewhat related: if the total timeout is not set, this could result in hangs while uploading since there's no `write_timeout` (could be added to `StreamWriter`?)

This looks promising to me. Might be cleaner to skip the `reschedule_timeout()` call in `set_response_params()` so the timeout is not started initially, and then have a `start_timeout()` call or something. Or maybe an extra kwarg `start_timeout=False`?
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_proto.py#L160

The only other code I see using this is in connector.py, and appears that it should adopt the same logic.

Dreamsorcerer:
@webknjaz There's no reason to expect `ResponseHandler.set_response_params()` to be used outside of the library, is there?

So, should be safe to change the timeout logic in that class?

aiohttp/http_writer.py
Dreamsorcerer: Seems like we're going to make the changes elsewhere, so this should be irrelevant once updated.

dtrifiro:
> This looks promising to me. Might be cleaner to skip the `reschedule_timeout()` call in `set_response_params()` so the timeout is not started initially, and then have a `start_timeout()` call or something.

If we want to modify `set_response_params`, we can just get rid of the `_reschedule_timeout` and schedule it only when we are finished writing, see the latest changes.

tests/test_client_functional.py
```diff
@@ -673,6 +673,29 @@ async def handler(request):
...
+
+    app = web.Application()
+    app.router.add_put("/", handler)
+    client = await aiohttp_client(app)
+
+    timeout = aiohttp.ClientTimeout(total=None, sock_read=sock_read_timeout)
+    resp = await client.put("/", data=gen_payload(), timeout=timeout)
+    txt = await resp.text()
+    assert txt == f"{len(payload)}"
+    resp.close()
```
```suggestion
async def test_read_timeout_on_write(aiohttp_client: Any) -> None:
    async def gen_payload() -> AsyncIterator[str]:
        # Delay writing to ensure read timeout isn't triggered before writing completes.
        await asyncio.sleep(0.5)
        yield b"foo"

    async def handler(request: web.Request) -> web.Response:
        return web.Response(body=await request.read())

    app = web.Application()
    app.router.add_put("/", handler)

    timeout = aiohttp.ClientTimeout(total=None, sock_read=0.1)
    client = await aiohttp_client(app)
    async with client.put("/", data=gen_payload(), timeout=timeout) as resp:
        result = await resp.read()  # Should not trigger a read timeout.
    assert result == b"foo"
```

Minor suggestions to make this more concise and clear.

Dreamsorcerer (approved):
I'll wait for @webknjaz to confirm as well, but this looks good to me.

Dreamsorcerer:
Adding a write timeout, as you mentioned before, is probably a good idea. But, maybe best to do it in a second PR.

CHANGES/7149.bugfix
```diff
@@ -0,0 +1,2 @@
+fixed timeouts when writing payload bodies when ``sock_read`` is set.
+-- by :user:`dtrifiro`
```
Dreamsorcerer:
```suggestion
changed ``sock_read`` timeout to start after writing has finished, to avoid read timeouts caused by an unfinished write. -- by :user:`dtrifiro`
```

dtrifiro:
Thanks @Dreamsorcerer. I'll try and find time to implement the write timeout in the next few weeks. ~~I'm still concerned about the whole file being uploaded even if the server immediately replies with an error, but I haven't yet figured it out. Will probably open an issue for that too.~~ edit: this is not an issue

Dreamsorcerer:
> Thanks @Dreamsorcerer. I'll try and find time to implement the write timeout in the next few weeks. I'm still concerned about the whole file being uploaded even if the server immediately replies with an error, but I haven't yet figured it out. Will probably open an issue for that too.

Might need a bit of testing, but I'd expect the server to close the connection in that case, which should cause the write code to error.

dtrifiro:
> Might need a bit of testing, but I'd expect the server to close the connection in that case, which should cause the write code to error.

That's indeed the behaviour, a simpler test setup shows this. The error I was seeing was due to a blocking call in my code.

dtrifiro:
Hi, is there anything I can do to help this get merged? Thanks!

Dreamsorcerer:
Think this is good now. Would be great to get that write_timeout in a new PR.

Dreamsorcerer:
#7206

PR #7152, opened 2022-12-30
Add response proxy headers to ClientResponse

### LINKED ISSUE #6078: Expose CONNECT response headers when using a proxy

When using a forward proxy, the client sends a [CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) request, to which you get a 200 response code, following which you make the proxied request. The AIOHTTP client however, doesn't expose the proxy's CONNECT response headers - which is often used to return debug information - as part of the final [ClientResponse](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse).

Can we?

```bash
# assumes that you've Docker setup locally
# let's start a forward proxy 
$ docker run -d --name=tinyproxy -p 8888:8888 dannydirect/tinyproxy ANY

# the 'proxy-agent' CONNECT response header is visible when using curl
$ curl 'https://www.facebook.com' -x http://localhost:8888 -sSD - -k -o /dev/null
HTTP/1.0 200 Connection established
Proxy-agent: tinyproxy/1.8.4

HTTP/2 200
vary: Accept-Encoding
set-cookie: cm_j=eu; expires=Mon, 18-Oct-2021 06:31:08 GMT; Max-Age=604800; path=/; domain=.facebook.com; secure; httponly; SameSite=Strict
...

$ cat << EOF >aio.py
import asyncio
import aiohttp


async def send_request(url, proxy):
        async with session.get(url, proxy=proxy) as resp:
            print(resp)


asyncio.run(send_request('https://www.facebook.com', "http://localhost:8888"))
EOF

# assumes that you've Python3 + AIOHTTP set up locally
# CONNECT response headers are missing
$ python3 aio.py | command grep -i 'proxy-agent'
```

The ClientResponse object should have a field name proxy_headers of the type CIMultiDictProxy.

The only way I have of getting this right now is to monkey patch TCPConnector.

### Related component

Client

### Additional context

_No response_

### Code of Conduct

galaxyfeeder:
## What do these changes do?

Add to the ClientResponse the headers and raw headers of the underlying CONNECT call in a proxied HTTPS request.

The headers and raw headers from the response are added to the ResponseHandler and later when processing the request are added to the response.

No.

## Related issue number

Closes #6078 

## Checklist


galaxyfeeder:
I'm unsure about:
- when this will be able to be released?
- better test the solution with the current tests for the proxy

dzhalaevd:
Hello, @galaxyfeeder this problem is still relevant?

tests/test_proxy_functional.py
```diff
@@ -881,6 +882,21 @@ async def xtest_proxy_from_env_https_with_auth(
...
 
 
+@pytest.mark.xfail
+async def xtest_proxy_https_proxy_headers(proxy_test_server: Any, loop: Any) -> None:
+    sess = aiohttp.ClientSession()
+    proxy = await proxy_test_server()
+    proxy.return_value = {"status": 200, "body": b"1" * (2**20)}
+    url = "https://www.google.com.ua/search?q=aiohttp proxy"
+
+    resp = await sess.get(url, proxy=proxy.url)
```
dzhalaevd, line 849: I think it's better to use the `async with` context manager because it ensures that resources will be properly closed after the `async with` block completes, even if an exception occurs. The code will also become more readable, because it does not require explicit invocation of closing methods (`close()`, `release()`, etc.).

Here's what the test implementation might look like:
```py
async def test_proxy_https_proxy_headers(make_proxy_test_server: Any) -> None:
        async with make_proxy_test_server() as proxy:
            proxy.return_value = {"status": 200, "body": b"1" * (2 ** 20)}
            url = "https://www.google.com.ua/search?q=aiohttp proxy"
            async with session.get(url, proxy=proxy.url) as resp:
                assert resp.proxy_headers == {"proxy-agent": "TestProxy/1.0"}
                assert resp.headers == {}
```

galaxyfeeder:
Hi @DavidRomanovizc, it is still relevant for us. Right now we are currently running a fork which includes this changes. Reading the proxy headers coming from underlying `CONNECT` allows us to debug the connection with proxy provider, mainly it allows us to receive information about the connection done through the proxy (outgoing IP address and performance data).

tests/test_proxy_functional.py
```diff
@@ -881,6 +882,21 @@ async def xtest_proxy_from_env_https_with_auth(
     assert r2.headers["Proxy-Authorization"] == auth.encode()
 
 
+@pytest.mark.xfail
```
Dreamsorcerer, line 842: What's with the xfail? Surely the test should be working?

tests/test_proxy_functional.py
Dreamsorcerer: Agreed (note you can make code suggestions with the first toolbar button, which can then just be comitted).

aiohttp/client_reqrep.py
```diff
@@ -871,6 +882,13 @@ async def start(self, connection: "Connection") -> "ClientResponse":
         self._headers = message.headers  # type is CIMultiDictProxy
         self._raw_headers = message.raw_headers  # type is Tuple[bytes, bytes]
 
+        # proxy headers
+        protocol = self._protocol
+        self._proxy_headers = protocol.proxy_headers if protocol is not None else None
+        self._raw_proxy_headers = (
+            protocol.raw_proxy_headers if protocol is not None else None
+        )
```
```suggestion
        if protocol is None:
            self._proxy_headers = None
            self._raw_proxy_headers = None
        else:
            self._proxy_headers = protocol.proxy_headers
            self._raw_proxy_headers = protocol.raw_proxy_headers
```

Dreamsorcerer:
I'd like @webknjaz to glance over this first, as I believe he's more familiar with the proxy stuff. I only have some minor comments.

tests/test_proxy_functional.py
dzhalaevd: Sure, thank you, I'll take it note

galaxyfeeder:
@Dreamsorcerer I've missed your review in my notifications. I will review your comments as soon as I have some spare time and I will sync with master.

tests/test_proxy_functional.py
galaxyfeeder: Thanks for the suggestion! Applying it!

tests/test_proxy_functional.py
galaxyfeeder: I don't remember why I put that annotation, I will remove it or try to understand why I add it.

galaxyfeeder: After trying to make that test work, I'm getting an SSL error trying to connect to the HTTPs website and get the response. Haven't been able to find any example on how should I be able to test this.

I was able to make it work with the fixture that uses proxy.py, but in that case I'm not able to modify the headers that I receive on the CONNECT reply.

I tried also to make it work with the mock, but I think that that implementation is not able to process HTTPS requests.

Any idea?

Dreamsorcerer: I think we really want a test that doesn't rely on an external server. @webknjaz @bdraco Any ideas how to rewrite the test?

aiohttp/client_proto.py
```diff
@@ -318,3 +332,9 @@ def data_received(self, data: bytes) -> None:
                         self.data_received(tail)
                     else:
                         self._tail = tail
+
+    def connection_proxied(
```
webknjaz, line 336: Why is it in the past tense? Methods should have imperative mood.

docs/client_reference.rst
```diff
@@ -1353,6 +1353,20 @@ Response object
       Unmodified HTTP headers of response as unconverted bytes, a sequence of
       ``(key, value)`` pairs.
 
+   .. attribute:: proxy_headers
+
+      A case-insensitive multidict proxy with HTTP headers of the CONNECT
+      response in a proxied HTTPS request, :class:`~multidict.CIMultiDictProxy` (optional).
+
+      .. versionadded:: 3.9
```
webknjaz, line 1361: This wouldn't be in 3.9.

docs/client_reference.rst
```diff
@@ -1353,6 +1353,20 @@ Response object
...
+
+   .. attribute:: raw_proxy_headers
+
+      Unmodified HTTP headers of the CONNECT response in a proxied HTTPS request
+      as unconverted bytes, a sequence of ``(key, value)`` pairs (optional).
+
+      .. versionadded:: 3.9
```
webknjaz, line 1368: This wouldn't be in 3.9.

tests/test_client_proto.py
```diff
@@ -155,3 +155,13 @@ async def test_connection_lost_sets_transport_to_none(loop: Any, mocker: Any) ->
     proto.connection_lost(OSError())
 
     assert proto.transport is None
+
+
+async def test_connection_proxied(loop: Any) -> None:
```
webknjaz, line 160: Plz add a docstring and loose the past tense.

tests/test_proxy_functional.py
```diff
@@ -838,6 +839,21 @@ async def xtest_proxy_from_env_https_with_auth(
...
+async def xtest_proxy_https_proxy_headers(proxy_test_server: Any, loop: Any) -> None:
```
webknjaz, line 843: @Dreamsorcerer @galaxyfeeder this also shouldn't start with `xtest` because this hides the function from the pytest collection process.
```suggestion
async def test_proxy_https_proxy_headers(proxy_test_server: Any, loop: Any) -> None:
```

tests/test_proxy_functional.py
webknjaz: Should we just follow the suite of `test_secure_https_proxy_absolute_path`?

webknjaz: It is possible to stick a unique header into `proxy.py` via https://github.com/abhinavsingh/proxy.py#modifyrequestheaderplugin. Here's an example of setting a “sentinel” header that my colleague came up with: https://github.com/ansible/ansible/pull/83085/files#diff-5b303b7467efd4c7529d1e171e39e472261e198fb172a45de566c9ec614824d2R3-R32.

PR #7175, opened 2023-01-18
Drop Python 3.6

Dreamsorcerer:
Obviously didn't bother to do this on the backport branch last time.


webknjaz:
@Dreamsorcerer pro tip: it's best to have branches in forks so when other people fork upstream, they wouldn't get copies of this repository polluted with garbage references in Git. This would be a nice thing to do for potential contributors.

PR #7180, opened 2023-01-22
Fix ConnectionResetError not being raised when the transport is closed

### LINKED ISSUE #7172: StreamResponse write() does not raise an exception if client disconnects
### Describe the bug

Hi!

I have an application that uses a `StreamResponse` to continuously write data to the client by calling the `write()` method. The server stops sending data when the client disconnects. In version 3.8.1 the `write()` method raises an exception when the client disconnects, but in later version it does not. The commit that introduces the bug is 20c93656d18b50c17c85b7a1e234276fb1444f27.

I've tested on Python 3.7, 3.8 and 3.10.

### To Reproduce

Add later if really needed.

### Expected behavior

An exception is raised when the client disconnects.

### Logs/tracebacks

```python-traceback
-
```

### Python Version

```console
$ python --version
Python 3.10.7
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.8.3
...
```

### multidict Version

```console
-
```

### yarl Version

```console
-
```

### OS

Linux

### Related component

Server

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

`ConnectionResetError` will always be raised when `StreamWriter.write` is called after `connection_lost` has been called on the `BaseProtocol`

Restores pre 3.8.3 behavior

## Related issue number

fixes #7172

## Checklist


tests/test_http_writer.py
```diff
@@ -262,6 +262,18 @@ async def test_write_to_closing_transport(
         await msg.write(b"After closing")
 
 
+async def test_write_to_closed_transport(
+    protocol: Any, transport: Any, loop: Any
+) -> None:
+    msg = http.StreamWriter(protocol, loop)
+
+    await msg.write(b"Before transport close")
+    protocol.transport = None
```
Dreamsorcerer, line 276: The only place I can see where transport is set to `None` is in `write_eof()`. I'm pretty sure this function is not going to have been called when the client connection drops...

bdraco: Its called here https://github.com/aio-libs/aiohttp/blob/master/aiohttp/base_protocol.py#L64

bdraco: I added a test for this and tweaked it a bit so there is a single source of truth for the transport

Dreamsorcerer: Right, it's a different object. So, `self._protocol.transport` is set to `None`, but `self._transport` is not.

I'd still like to know the root cause, as to why that gets set to `None` on a disconnection, but the actual transport object still returns `False` from `.is_closing()`. Seems like they should both happen...

bdraco: I expect it has something to do with all the ssl sends being buffered and the plaintext transport tries to send immediately so it knows the connection has dropped.

Dreamsorcerer: I've tried to track down the issue in the code, but not figured it out.
Seems like _SSLProtocolTransport.is_closing() is the one being used when doing SSL, but it seems to wrap the non-SSL object, so the regular transport would be available under _SSLProtocolTransport._ssl_protocol._transport, and I'm wondering if it should be proxying the call to that transport, which would then given us the same result as non-SSL connections.

CHANGES/7180.bugfix
```diff
@@ -0,0 +1 @@
+`ConnectionResetError` will always be raised when `StreamWriter.write` is called after `connection_lost` has been called on the `BaseProtocol`
```
webknjaz: Remember to use RST syntax here, not markdown.

tests/test_client_proto.py
```diff
@@ -136,3 +136,18 @@ async def test_eof_received(loop: Any) -> None:
...
     assert proto._read_timeout_handle is None
+
+
+async def test_connection_lost_sets_transport_to_none(loop: Any) -> None:
+    """Ensure that the transport is set to None when the connection is list.
+
+    This ensures the writer knows that the connection is closed.
+    """
+    proto = ResponseHandler(loop=loop)
+    transport = mock.Mock()
```
webknjaz: Could you get this from the `mocker` fixture?

aiohttp/http_writer.py
```diff
@@ -71,10 +70,10 @@ def _write(self, chunk: bytes) -> None:
         size = len(chunk)
         self.buffer_size += size
         self.output_size += size
-
-        if self._transport is None or self._transport.is_closing():
+        transport = self._protocol.transport
```
webknjaz: Sounds like now it's just
```suggestion
        transport = self.transport
```

tests/test_http_writer.py
```diff
@@ -262,6 +262,18 @@ async def test_write_to_closing_transport(
...
+
+    with pytest.raises(ConnectionResetError):
```
webknjaz: Remember to always use `match=` with `pytest.raises()`.

tests/test_http_writer.py
```diff
@@ -262,6 +262,18 @@ async def test_write_to_closing_transport(
         await msg.write(b"After closing")
 
 
+async def test_write_to_closed_transport(
+    protocol: Any, transport: Any, loop: Any
+) -> None:
```
webknjaz, line 267: A docstring here wouldn't hurt.

CHANGES/7180.bugfix
webknjaz: https://github.com/aio-libs/aiohttp/tree/master/CHANGES#examples-for-adding-changelog-entries-to-your-pull-requests

bdraco: 860fde8aea3c78b3c5a769c38ce8308ac6787d22

tests/test_http_writer.py
bdraco: 516a5c1aeb79fc3ad0b434a18338b0eaae5f904c

tests/test_http_writer.py
bdraco: 21bb53bbaed029df234a3961a7489d1b2977e4a6
0ff872e60bc46ad5378825b4b46b4f5b09baa478

tests/test_client_proto.py
bdraco: 5833948074ba7b3eab752ea4c594e8279aabff24

tests/test_http_writer.py
Dreamsorcerer: OK, I don't think we're going to work that out. But, having another look, I think I've found a clean solution.

It seems that `BaseProtocol.connection_lost()` is called reliably and sets an attribute:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/base_protocol.py#L62

Maybe we should just add a property method to expose that publically, and check for that attribute instead?

bdraco: 90f42571b78cea86165adbab392218a7104d0d60
471fa2e200009f1463e5743d6d986eec5e2cd89e

aiohttp/http_writer.py
```diff
@@ -71,10 +70,9 @@ def _write(self, chunk: bytes) -> None:
         size = len(chunk)
         self.buffer_size += size
         self.output_size += size
-
-        if self._transport is None or self._transport.is_closing():
+        if not self._protocol.connected:
             raise ConnectionResetError("Cannot write to closing transport")
-        self._transport.write(chunk)
+        self.transport.write(chunk)
```
Dreamsorcerer: You could do something funky with a `TypeGuard`, but you'd probably need to make the protocol generic in order to make it work. Probably easier to just add an assert here.

```suggestion
        assert self.transport is not None  # Must be true if connected.
        self.transport.write(chunk)
```

Dreamsorcerer: Saying that, there is actually a test failing because this is None. Not sure how that happens..

bdraco: Looks like a mocking issue in the test. I'll revert the None check to make mypy happy.

aiohttp/http_writer.py
```diff
@@ -71,10 +70,10 @@ def _write(self, chunk: bytes) -> None:
         size = len(chunk)
         self.buffer_size += size
         self.output_size += size
-
-        if self._transport is None or self._transport.is_closing():
+        transport = self.transport
+        if not self._protocol.connected or transport is None or transport.is_closing():
```
Dreamsorcerer, line 74: Surely the other 2 checks would always produce the same result as the first? Do we actually need them anymore?

aiohttp/base_protocol.py
```diff
@@ -76,8 +79,8 @@ def connection_lost(self, exc: Optional[BaseException]) -> None:
             waiter.set_exception(exc)
 
     async def _drain_helper(self) -> None:
-        if self._connection_lost:
-            raise ConnectionResetError("Connection lost")
+        if not self.connected:
+            raise ConnectionResetError("Not connected")
```
Dreamsorcerer:
```suggestion
            raise ConnectionResetError("Connection lost")
```

Think we should stick to the original, this method doesn't look like it would get called if we didn't create a connection in the first place.

tests/test_client_proto.py
```diff
@@ -136,3 +136,17 @@ async def test_eof_received(loop: Any) -> None:
     assert proto._read_timeout_handle is not None
     proto.eof_received()
     assert proto._read_timeout_handle is None
+
+
+async def test_connection_lost_sets_transport_to_none(loop: Any, mocker: Any) -> None:
+    """Ensure that the transport is set to None when the connection is list.
```
Dreamsorcerer:
```suggestion
    """Ensure that the transport is set to None when the connection is lost.
```

aiohttp/base_protocol.py
```diff
@@ -18,11 +18,15 @@ def __init__(self, loop: asyncio.AbstractEventLoop) -> None:
...
         self._drain_waiter: Optional[asyncio.Future[None]] = None
-        self._connection_lost = False
         self._reading_paused = False
 
         self.transport: Optional[asyncio.Transport] = None
 
+    @property
+    def connected(self) -> bool:
+        """Return True if the connection is open."""
+        return bool(self.transport)
```
Dreamsorcerer:
```suggestion
        return self.transport is not None
```

Probably better to be explicit.

Dreamsorcerer (approved):
Looks good to me, any last comments @webknjaz?

aiohttp/http_writer.py
Dreamsorcerer: Still not clear if these are needed, but will leave them to be safe.

bdraco:
Do you want me to take care of the backport commits?  Home Assistant is blocked from upgrading to Python 3.11 because of this so we are quite keen to get a new 3.8.x

Dreamsorcerer:
I already did them, but thanks. #7198 #7199

bdraco:
Thank you very much.  Is there anything blocking getting a new 3.8.x release? 

edit: I assume https://github.com/aio-libs/aiohttp/pull/7200 is

Dreamsorcerer:
@webknjaz Are you happy to push one more 3.8 release?

This is a moderate severity issue, so I think it's worth getting this fix out. Will also get people to stop complaining about the charset-normalizer pin.

I've cleared the errors on CI, so should be good to go.

webknjaz:
@Dreamsorcerer okay, I'll try to take a look... unless you want to take this on personally?

Dreamsorcerer:
I can take a look tomorrow. Is it just adding a tag like other repos, or are there manual steps to create the changelog?

webknjaz:
@Dreamsorcerer the changelog should be generated and the hardcoded version number should be bumped too. Then, the commit with these changes should be tagged with the same version. The rest of the automation is the same. After the release, the 3.8 branch should be carefully merged into 3.9 and master so the changelog fragment removal would be reflected there. But some things shouldn't get into the merge. And it should be a proper merge, not squash, for Git to track things properly.

bdraco:
Thank you.

ZupoLlask:
> @webknjaz Are you happy to push one more 3.8 release?
> 
> This is a moderate severity issue, so I think it's worth getting this fix out. Will also get people to stop complaining about the charset-normalizer pin.
> 
> I've cleared the errors on CI, so should be good to go.

Sorry guys but... Am I wrong or you forgot to remove/bump that charset-normalizer?

Thanks!

PR #7181, opened 2023-01-23
Add ClientResponse.certificate and populate it, if appropriate

### LINKED ISSUE #2816: Feature request: add functionality to retrieve the the server certificate in a client request
## Rationale
This request is related to this [SO Question](https://stackoverflow.com/questions/49154697/aiohttp-how-do-i-retrieve-peer-certificate/49166837#49166837) where I asked whether or not it is possible to get the peer certificate when running a client request.

It is quite logical that the implemented hack is not going to be supported. I do, however, have use-cases in which it is required to retrieve the peer certificate. And I was hoping on working together to find a clean solution.

In protocols like Kerberos and NTLM the server certificate is used in followup steps of the authentication protocol.

## A possible solution

If there is a better solution that would be great as well. Just trying to offer something workable

An extra parameter like `retrieve_peer_certificate` to the method `aiohttp.Client._request` which is default `False` and would make sure that a property like `peer_certificate` would be set in the response object. It would it be possible internally use a `_get_transport` that would return the SSLSocket if the connection is SSL/TLS on which you would be able to call `getpeercert`

auxsvr:
Fixes #2816.

## What do these changes do?
Add `ClientResponse.certificate` and populate it, if appropriate.

Yes, the `ClientResponse.certificate` property contains the server certificate, if available.

## Related issue number
#2816

## Checklist

### DIFF
--- a/CHANGES/2816.feature
+++ b/CHANGES/2816.feature
@@ -1,1 +1,1 @@
-Add `ClientResponse.certificate` and populate it, if appropriate.
+Added ``ClientResponse.certificate``, which is populated, if appropriate, with the certificate of the server.
--- a/CHANGES/2816.feature
+++ b/CHANGES/2816.feature
@@ -1,0 +1,1 @@
+Add `ClientResponse.certificate` and populate it, if appropriate.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -861,11 +861,7 @@ class ClientResponse(HeadersMixin):
                     set_result(self._continue, True)
                     self._continue = None
 
-        # certificate
-        is_tls_transport = (
-            self._connection is not None and self._connection.transport is not None
-        )
-        if is_tls_transport:
+        if self._connection is not None and self._connection.transport is not None:
             self.certificate = self._connection.transport.get_extra_info("peercert")
 
         # payload eof handler
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -882,7 +882,9 @@ class ClientResponse(HeadersMixin):
                 client_logger.warning("Can not load response cookies: %s", exc)
 
         # certificate
-        is_tls_transport = self._connection is not None and self._connection.transport is not None
+        is_tls_transport = (
+            self._connection is not None and self._connection.transport is not None
+        )
         if is_tls_transport:
             self.certificate = self._connection.transport.get_extra_info("peercert")
         return self
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -882,7 +882,8 @@ class ClientResponse(HeadersMixin):
                 client_logger.warning("Can not load response cookies: %s", exc)
 
         # certificate
-        if self._connection is not None and self._connection.transport is not None:
+        is_tls_transport = self._connection is not None and self._connection.transport is not None
+        if is_tls_transport:
             self.certificate = self._connection.transport.get_extra_info("peercert")
         return self
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -862,7 +862,9 @@ class ClientResponse(HeadersMixin):
                     self._continue = None
 
         # certificate
-        is_tls_transport = self._connection is not None and self._connection.transport is not None
+        is_tls_transport = (
+            self._connection is not None and self._connection.transport is not None
+        )
         if is_tls_transport:
             self.certificate = self._connection.transport.get_extra_info("peercert")
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -861,6 +861,11 @@ class ClientResponse(HeadersMixin):
                     set_result(self._continue, True)
                     self._continue = None
 
+        # certificate
+        is_tls_transport = self._connection is not None and self._connection.transport is not None
+        if is_tls_transport:
+            self.certificate = self._connection.transport.get_extra_info("peercert")
+
         # payload eof handler
         payload.on_eof(self._response_eof)
 
@@ -882,13 +887,6 @@ class ClientResponse(HeadersMixin):
                 self.cookies.load(hdr)
             except CookieError as exc:
                 client_logger.warning("Can not load response cookies: %s", exc)
-
-        # certificate
-        is_tls_transport = (
-            self._connection is not None and self._connection.transport is not None
-        )
-        if is_tls_transport:
-            self.certificate = self._connection.transport.get_extra_info("peercert")
         return self
 
     def _response_eof(self) -> None:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -697,6 +697,7 @@ class ClientResponse(HeadersMixin):
 
         self.method = method
         self.cookies: SimpleCookie[str] = SimpleCookie()
+        self.certificate: Optional[Dict[str, Union[Tuple[Any, ...], int, str]]] = None
 
         self._real_url = url
         self._url = url.with_fragment(None)
@@ -879,6 +880,10 @@ class ClientResponse(HeadersMixin):
                 self.cookies.load(hdr)
             except CookieError as exc:
                 client_logger.warning("Can not load response cookies: %s", exc)
+
+        # certificate
+        if self._connection is not None and self._connection.transport is not None:
+            self.certificate = self._connection.transport.get_extra_info("peercert")
         return self
 
     def _response_eof(self) -> None:
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1360,6 +1360,10 @@ Response object
       objects of preceding requests (earliest request first) if there were
       redirects, an empty sequence otherwise.
 
+   .. attribute:: certificate
+
+      The server certificate, if available.
+
    .. method:: close()
 
       Close response and underlying connection.
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1362,7 +1362,9 @@ Response object
 
    .. attribute:: certificate
 
-      The server certificate, if available.
+      The server certificate. This is available with TLS connections,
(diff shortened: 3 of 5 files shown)


CHANGES/2816.feature
```diff
@@ -0,0 +1 @@
+Add `ClientResponse.certificate` and populate it, if appropriate.
```
webknjaz: This should use RST and be in the past tense. It should also be an accurate description of what effect the end-users will see when upgrading.

aiohttp/client_reqrep.py
```diff
@@ -880,6 +881,10 @@ async def start(self, connection: "Connection") -> "ClientResponse":
                 self.cookies.load(hdr)
             except CookieError as exc:
                 client_logger.warning("Can not load response cookies: %s", exc)
+
+        # certificate
+        if self._connection is not None and self._connection.transport is not None:
+            self.certificate = self._connection.transport.get_extra_info("peercert")
```
webknjaz: What happens if it's not a TLS connection?

aiohttp/client_reqrep.py
```diff
@@ -880,6 +881,10 @@ async def start(self, connection: "Connection") -> "ClientResponse":
                 self.cookies.load(hdr)
             except CookieError as exc:
                 client_logger.warning("Can not load response cookies: %s", exc)
+
+        # certificate
+        if self._connection is not None and self._connection.transport is not None:
```
webknjaz: Labeling complex conditions improves maintainability:
```suggestion
        is_tls_transport = self._connection is not None and self._connection.transport is not None:
        if is_tls_transport:
```

docs/client_reference.rst
```diff
@@ -1360,6 +1360,10 @@ Response object
       objects of preceding requests (earliest request first) if there were
       redirects, an empty sequence otherwise.
 
+   .. attribute:: certificate
+
+      The server certificate, if available.
```
webknjaz: It's be a good idea to indicate when it's available.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,9 @@ def test_loose_cookies_types(loop: Any) -> None:
 def test_gen_default_accept_encoding(has_brotli: Any, expected: Any) -> None:
     with mock.patch("aiohttp.client_reqrep.HAS_BROTLI", has_brotli):
         assert _gen_default_accept_encoding() == expected
+
+
+async def test_certificate(loop: Any, conn: Any) -> None:
+    req = ClientRequest("get", URL("https://www.python.org"), loop=loop)
```
webknjaz: Instead of reaching out to the network, it's best to use a local test server, extracting it's certificate.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,9 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    resp = await req.send(conn)
+    assert resp.certificate is not None
```
webknjaz: With a local server, you can check/inspect the exact certificate here.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,9 @@ def test_loose_cookies_types(loop: Any) -> None:
 def test_gen_default_accept_encoding(has_brotli: Any, expected: Any) -> None:
     with mock.patch("aiohttp.client_reqrep.HAS_BROTLI", has_brotli):
         assert _gen_default_accept_encoding() == expected
+
+
+async def test_certificate(loop: Any, conn: Any) -> None:
+    req = ClientRequest("get", URL("https://www.python.org"), loop=loop)
+    resp = await req.send(conn)
```
webknjaz: Don't forget to close both the response and the request to prevent resource warnings. It's useful to use CMs for this.

tests/test_client_request.py
```diff
@@ -1230,3 +1230,9 @@ def test_loose_cookies_types(loop: Any) -> None:
 def test_gen_default_accept_encoding(has_brotli: Any, expected: Any) -> None:
     with mock.patch("aiohttp.client_reqrep.HAS_BROTLI", has_brotli):
         assert _gen_default_accept_encoding() == expected
+
+
+async def test_certificate(loop: Any, conn: Any) -> None:
```
webknjaz: This test name should be more specific. A docstring wouldn't hurt too.

aiohttp/client_reqrep.py
auxsvr: With HTTP connections, nothing happens.

tests/test_client_request.py
auxsvr: Not sure about this. A real server allows one to test in various conditions that are not easy to recreate in a controlled environment, such as premature loss of transport or connection, redirects, proxy servers, and is closest to the experience of the end user. On the other hand, if the connection is closed immediately due to a transient issue, this test will fail. As many tests depend on a real server, this should not be much of an issue.

tests/test_client_request.py
auxsvr: Yes, but why should I? This is a test to check if `self.certificate` is populated, not to check if `.get_extra_info("peercert")` extracts the certificate correctly.

Another useful test might be to connect to an HTTP server, such as http://example.com, but  these are getting rare these days.

Dreamsorcerer: Checking that is it populated _correctly_ would still be preferable.

tests/test_client_request.py
Dreamsorcerer: > As many tests depend on a real server, this should not be much of an issue.

What tests? I'm not sure our test suite needs an external internet connection at all.
Adding a test which relies on an internet connection and a working server at python.org is not a good way to go. We have many packagers etc. who run our tests as well, so it's not just us this would affect.

auxsvr: Much of the code in the test was copied from another test, such as https://github.com/aio-libs/aiohttp/blob/b0a679f9247fce4e3600d7ef5ea796b9d09b6862/tests/test_client_request.py#L562 There are several such tests in the suite.

auxsvr: Anyway, the test now uses a local test server. The assertion now is about the keys in the certificate dictionary, which makes the test dependent on the certificates `trustme` produces.

tests/test_client_request.py
auxsvr: Currently, the assertion is about the keys of `self.certificate`.

auxsvr:
The linter failure in mypy is spurious. Should I silence this instance or is there a better way?

Dreamsorcerer:
> The linter failure in mypy is spurious. Should I silence this instance or is there a better way?

mypy is pretty much deterministic, the failure will appear every time.
In this case, the variable is defined as `None` in the class definition:
https://github.com/aio-libs/aiohttp/pull/7181/files#diff-720910fd022b9ba8b7ddf0cb446ce5e1afe443f4287ad26a1eb4dc4b5b5f4508R677

aiohttp/client_reqrep.py
```diff
@@ -860,6 +861,13 @@ async def start(self, connection: "Connection") -> "ClientResponse":
                     set_result(self._continue, True)
                     self._continue = None
 
+        # certificate
+        is_tls_transport = (
+            self._connection is not None and self._connection.transport is not None
+        )
```
Dreamsorcerer: Ah, but this variable trips it up. mypy can't infer whether the next line passes based on the connection variable. Just stick to putting this logic in the `if` without a variable.

aiohttp/client_reqrep.py
```diff
@@ -860,6 +861,13 @@ async def start(self, connection: "Connection") -> "ClientResponse":
                     set_result(self._continue, True)
                     self._continue = None
 
+        # certificate
```
```suggestion
```

docs/client_reference.rst
```diff
@@ -1360,6 +1360,12 @@ Response object
       objects of preceding requests (earliest request first) if there were
       redirects, an empty sequence otherwise.
 
+   .. attribute:: certificate
+
+      The server certificate. This is available with TLS connections,
+      if the certificate is retrieved before the connection or the TLS session are
```
```suggestion
      if the certificate is retrieved before the connection or the TLS session is
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,31 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+            "subject",
+            "issuer",
+            "version",
+            "serialNumber",
+            "notBefore",
+            "notAfter",
+            "subjectAltName",
+        )
+    )
+    resp.close()
```
Dreamsorcerer:
```suggestion
        async with session.get("https://127.0.0.1:8443", ssl=client_ssl_ctx) as resp:
            assert frozenset(resp.certificate) >= frozenset(
                (
                    "subject",
                    "issuer",
                    "version",
                    "serialNumber",
                    "notBefore",
                    "notAfter",
                    "subjectAltName",
                )
            )
```

tests/test_client_request.py
```diff
@@ -1230,3 +1230,31 @@ def test_loose_cookies_types(loop: Any) -> None:
...
+    async def handler(request):
+        return web.Response(text="ok")
+
+    server = web.Server(handler)
+    runner = web.ServerRunner(server)
+    await runner.setup()
+    site = web.TCPSite(runner, ssl_context=ssl_ctx)
+    await site.start()
+
+    async with aiohttp.ClientSession() as session:
```
Dreamsorcerer: Isn't this what the aiohttp_client fixture is for?
https://docs.aiohttp.org/en/stable/testing.html#pytest

aiohttp/client_reqrep.py
```diff
@@ -699,6 +699,7 @@ def __init__(
 
         self.method = method
         self.cookies: SimpleCookie[str] = SimpleCookie()
+        self.certificate: Optional[Dict[str, Union[Tuple[Any, ...], int, str]]] = None
```
Dreamsorcerer, line 702: The type appears to be wrong. Typeshed defines it as: https://github.com/python/typeshed/blob/9d6fc1dafdf9d27a9ea87e210ff6e287d00a61d5/stdlib/ssl.pyi#L11

We could probably do even better and define a TypedDict if we wanted to.

Issue #7185, opened 2023-01-28
ClientSesssion's request methods don't display acceptable parameters.


jujumilk3:
### Describe the solution you'd like

I just created PR of it.
I created more explicit parameters for the http methods used in the ClientSession class.

### Describe alternatives you've considered

Create http methods as classes and inherit parameters from `_request` method

### Related component

Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
An issue pointing to a PR with no description is not helpful, we'll just look at the PR..

PR #7186, opened 2023-01-28
Feature/expose request interface

jujumilk3:
## What do these changes do?
Display ClientSession methods' parameters in the help window.

Before
image
After
image

Yes

## Related issue number
https://github.com/aio-libs/aiohttp/issues/7185

## Checklist

### DIFF
--- a/CHANGES/7185.feature
+++ b/CHANGES/7185.feature
@@ -1,0 +1,1 @@
+Display ClientSession methods' parameters in the help window.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -104,6 +104,7 @@ Dmitry Shamov
 Dmitry Trofimov
 Dmytro Bohomiakov
 Dmytro Kuznetsov
+Donggyu Damon Lee
 Dustin J. Mitchell
 Eduard Iskandarov
 Eli Ribble
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -4,6 +4,7 @@ import asyncio
 import base64
 import dataclasses
 import hashlib
+import inspect
 import json
 import os
 import sys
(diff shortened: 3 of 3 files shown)


Dreamsorcerer:
I don't think we want so much copy/paste.

I think there is a better alterative coming up though: https://peps.python.org/pep-0692/
Seems we'll need to wait until Python 3.12 to use it.

dzhalaevd:
> I don't think we want so much copy/paste.
> 
> I think there is a better alterative coming up though: https://peps.python.org/pep-0692/ Seems we'll need to wait until Python 3.12 to use it.

I think it may not be necessary to wait for version 3.12, as precise type annotations for **kwargs are now supported using TypedDict starting from MyPy version 0.981 (https://github.com/python/mypy/pull/13471). This means that **kwargs: Any can now be rewritten as **kwargs: Unpack[HttpClientRequest], for example.

We need to update `mypy` to 0.981 and create possibly `types.py` which will contain our custom types

Here is an example that works on python 3.10
```python
import sys
from typing import (
    Any,
    Optional,
    TypedDict,
    Union
)

if sys.version_info >= (3, 11):
    from typing import (  # noqa: F401
        Unpack,
    )
else:
    from typing_extensions import (  # noqa: F401
        Unpack,
    )

StrOrURL = Union[str, bytes, "URL"]


class HttpClientRequest(TypedDict):
    data: Optional[Union[str, bytes, dict]]
    json: Optional[Union[dict, list]]
    cookies: Optional[dict[str, str]]
    headers: Optional[dict[str, str]]
    allow_redirects: Optional[bool]


def get(url: StrOrURL, *, allow_redirects: bool = True, **kwargs: Unpack[HttpClientRequest]) -> Any:
    """
    A function that sends an HTTP GET request.

    :param url: The URL to send the request to.
    :param allow_redirects: Whether or not to follow redirects.
    :param kwargs: Additional parameters to pass to the request.
    :return: The response from the server.
    """


# Example usage
param = {"data": None, "json": None, "cookies": None, "headers": None, "allow_redirects": True}
get(url="https://example.com", **param)

get(url="https://example.com", allow_redirects=True, data=None, json=None)
```

Dreamsorcerer:
> > I don't think we want so much copy/paste.
> > I think there is a better alterative coming up though: https://peps.python.org/pep-0692/ Seems we'll need to wait until Python 3.12 to use it.
> 
> I think it may not be necessary to wait for version 3.12, as precise type annotations for **kwargs are now supported using TypedDict starting from MyPy version 0.981 ([python/mypy#13471](https://github.com/python/mypy/pull/13471)). This means that **kwargs: Any can now be rewritten as **kwargs: Unpack[HttpClientRequest], for example.

That is PEP 692. It's still not stable and disabled by default (https://github.com/python/mypy/issues/14697) and requires Unpack from Python 3.12. We can use typing_extensions if needed, but would be nice to start dropping typing_extensions as a dependency if possible. So, I'd probably still wait till 3.12 is released, though we can start preparing a PR beforehand if anyone wants to start looking at it.

Dreamsorcerer:
This has been done with Unpack now.

PR #7191, opened 2023-02-04
Make tests pass after year 2039

bmwiedemann:
## What do these changes do?

Make tests pass after year 2039

none

## Related issue number

N/A

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future. The usual offset is +16 years, because that is how long I expect some software will be used in some places. This showed up failing tests in our `python-aiohttp` package build. See https://reproducible-builds.org/ for why this matters.

## Checklist

### DIFF
--- a/CHANGES/7191.misc
+++ b/CHANGES/7191.misc
@@ -1,0 +1,1 @@
+Made tests pass after the year 2039.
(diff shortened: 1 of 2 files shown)


CHANGES/7191.misc
```diff
@@ -0,0 +1 @@
+Make tests pass after 2039
```
webknjaz: It took me a while to understand that this is not a PR/issue number, but a year. A changelog fragment should be specifica about that, use the past tense, contain complete sentences (with a trailing period) and be more accurate in what it described — generic phrases are pointless in the changelog.

bmwiedemann: I updated the file. Feel free to rewrite it as you see appropriate.
I'm just a one-time contributor.

Dreamsorcerer: Personally, I'd just have skipped the changelog, I doubt any other users are going to care about this when checking what changed in a release.

PR #7221, opened 2023-02-21
Test if CookieJar pickle format breaks

### LINKED ISSUE #7216: Loading saved CookieJar breaks in 3.8.4
### Describe the bug

I save the contents of a CookieJar to persist a session. Loading a file saved in <3.8.4 will not work in aiohttp 3.8.4 as expected. When I'd iterate through the cookies the expiration update will raise an unhandled exception.

This issue completely breaks the authentication flow and bricks the session of a user in my application.

### To Reproduce

1. Create and save a simple CookieJar in 3.8.3:
```python
from aiohttp import CookieJar
cj = CookieJar()
cooks = SimpleCookie({"coo": "kie"})
cj.update_cookies(cooks)
cj.save("/tmp/aiohttp383s.cj")
```
2. Load it in 3.8.4 and trigger expiration-update by iterating over it:
```python
from aiohttp import CookieJar
cj = CookieJar()
cj.load("/tmp/aiohttp383s.cj")
list(cj)
```

### Expected behavior

I expected the save/load to work across versions, but at least when upgrading to a newer version. The documentation does not mention anything to be afraid of either.

### Logs/tracebacks

```python-traceback
  File "<input>", line 1, in <module>
  File "/home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages/aiohttp/cookiejar.py", line 147, in __len__
    return sum(1 for i in self)
  File "/home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages/aiohttp/cookiejar.py", line 147, in <genexpr>
    return sum(1 for i in self)
  File "/home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages/aiohttp/cookiejar.py", line 142, in __iter__
    self._do_expiration()
  File "/home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages/aiohttp/cookiejar.py", line 150, in _do_expiration
    self.clear(lambda x: False)
  File "/home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages/aiohttp/cookiejar.py", line 115, in clear
    for (domain, path), cookie in self._cookies.items():
ValueError: not enough values to unpack (expected 2, got 0)
```

### Python Version

```console
Python 3.10.9
```

### aiohttp Version

```console
Name: aiohttp
Version: 3.8.4
Author: 
Author-email: 
License: Apache 2
Location: /home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages
Requires: aiosignal, async-timeout, attrs, charset-normalizer, frozenlist, multidict, yarl
```

### multidict Version

```console
Name: multidict
Version: 6.0.4
Author: Andrew Svetlov
License: Apache 2
Location: /home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
Name: yarl
Version: 1.8.2
Author: Andrew Svetlov
License: Apache 2
Location: /home/gg/PycharmProjects/.venv/x310/lib/python3.10/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

Arch Linux

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #7216 (well, stops it from happening accidentally again).


Traktormaster:
Do I understand correctly that the save/load functionality will not be patched to be backward compatible with itself?

That would seem antithetical to their purpose of persisting the value and being able to use it later. In fact, breaking changes would require the using application to migrate the data at which point it could be easier to write a custom save/load for itself.

I feel like this is something the library should handle and not each application itself that uses it.

Dreamsorcerer:
> Do I understand correctly that the save/load functionality will not be patched to be backward compatible with itself?

I'm not sure exactly how we'd handle this. Possibly adding a version number to the dict or something?
If you'd like to have a go at introducing a backwards compatible approach, please go ahead.

Dreamsorcerer:
Having said that, it's worth noting that this is a pickle format, so it seems highly unlikely that we'd guarantee that it would never break.
It might break if you save from a newer version of Python and load in an older version (that doesn't support the newer pickle protocol). It's likely to break if we move or change any of the classes involved. It's likely to break if Python changes any of the classes involved (in http.cookies).

So, I don't think we can really provide much of a guarantee here regardless. The fact that the implementation uses pickle suggests to me that it's a convenience function, rather than a file format that should be relied upon from version to version. i.e. I think we should only guarantee that a file saved and loaded from the same machine in the same version of Python and the same x.y version of aiohttp will work.

Traktormaster:
It sounds like I've not understood truly what the save/load methods provide. The documentation does state it's using a "pickled representation", but I did not think much of it. Maybe a warning at the load would be nice that "Loading a saved value from a different version might not work" or something to make it explicit.

I might look into making a PR later with a save/load serialization option that provides backward compatibility... it seems I have to do it anyways after all.

Dreamsorcerer:
Feel free to change that test alongside any other PRs to improve backwards compatibility.

Issue #7192, opened 2023-02-06
Enabling of responses compression causes blocking of the async loop


mykola-mokhnach:
### Describe the bug

Enabling  `response.enable_compression()` on a `web.Response` instance causes async loop blocking delays up to 2 seconds long on our server. 

I assume the original issue is coming from the fact that the compression is synchronous and eats lots of CPU cycles in case of large chunks/high core load: https://github.com/aio-libs/aiohttp/blob/2a1024ff1e8f210c5d735f87cd3b408c9ceedd8e/aiohttp/http_writer.py#L97

Perhaps it would be possible to "offload" the compression to another CPU core in order to unblock the main event loop.

### To Reproduce

Create a simple GET endpoint, which returns a JSON response like 

```python
            response = web.Response(
                text=json.dumps(result_json, ensure_ascii=False), status=200, content_type='application/json'
            )
            response.enable_compression()
            return response
```

Then fetch the endpoint multiple times with json payload large enough (up to 10 megabytes in my case) and deflate compression accepted in the request headers. Async event loop debug must be enabled on the server side to see blocking events occurrences

### Expected behavior

No async loop blocks should be observed longer as 100ms, although our charts show delays up to 2 seconds. This is not observed if I disable the compression.

### Logs/tracebacks

```python-traceback
[EventLoopBlockingTimeEvent] function RequestHandler._handle_request (<Handle <TaskWakeupMethWrapper object at 0x7fb12e614ee0>(<Task finishe...> result=None>)>) was blocking the loop for 168.54695370420814 ms, called with args self: <RequestHandler connected>, request: <Request GET /api/v1/files >, start_time: 2369932.481492733, request_handler: <bound method Application._handle of <Application 0x7fb12f8bd190>>
```

### Python Version

```console
$ python --version
Python 3.8.16
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.8.1
```

### multidict Version

```console
$ python -m pip show multidict
Version: 6.0.4
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.8.2
```

### OS

python:3.8-slim-bullseye Docker image

### Related component

Server

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

mykola-mokhnach:
[...]

PR #7223, opened 2023-02-24
refactor: Extract zlib-related logic into a single module

mykola-mokhnach:
## What do these changes do?

Addresses issue https://github.com/aio-libs/aiohttp/issues/7192
Refactors the logic to have the zlib-related stuff concentrated into a single module

No

## Related issue number

https://github.com/aio-libs/aiohttp/issues/7192

## Checklist

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -902,7 +902,7 @@ class DeflateBuffer:
             self.out.feed_data(chunk, len(chunk))
             if (
                 self.encoding == "deflate"
-                and isinstance(self.decompressor, ZLibDecompressor)
+                and hasattr(self.decompressor, 'eof')
                 and not self.decompressor.eof
             ):
                 raise ContentEncodingError("deflate")
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -902,7 +902,7 @@ class DeflateBuffer:
             self.out.feed_data(chunk, len(chunk))
             if (
                 self.encoding == "deflate"
-                and hasattr(self.decompressor, 'eof')
+                and hasattr(self.decompressor, "eof")
                 and not self.decompressor.eof
             ):
                 raise ContentEncodingError("deflate")
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -902,8 +902,7 @@ class DeflateBuffer:
             self.out.feed_data(chunk, len(chunk))
             if (
                 self.encoding == "deflate"
-                and hasattr(self.decompressor, "eof")
-                and not self.decompressor.eof
+                and not self.decompressor.eof  # type: ignore
             ):
                 raise ContentEncodingError("deflate")
 
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -6,7 +6,6 @@ import string
 from contextlib import suppress
 from enum import IntEnum
 from typing import (
-    Any,
     Generic,
     List,
     NamedTuple,
@@ -21,7 +20,7 @@ from typing import (
 )
 
 from multidict import CIMultiDict, CIMultiDictProxy, istr
-from typing_extensions import Final
+from typing_extensions import Final, Protocol
 from yarl import URL
 
 from . import hdrs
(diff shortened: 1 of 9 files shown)


aiohttp/zlib_utils.py
```diff
@@ -0,0 +1,144 @@
...
+        ...
+
+    def flush(self, mode: int = ...) -> bytes:
+        ...
+
+    def copy(self) -> "Compressor":
+        ...
+
+
+class Decompressor(Protocol):
```
Dreamsorcerer: Given that the other use case is the custom BrotliDecoder class, why don't we just make both classes inherit from this, rather than using a Protocol?

Also, this class seems to be declared twice. It should probably just be declared in typedefs.py or similar.

mykola-mokhnach: > Also, this class seems to be declared twice. It should probably just be declared in typedefs.py or similar.

The other protocol is different even though it has the same name. Also, they are not supposed to be public and are suited for internal usage in classes where they are declared on the module level. Having a declared protocol is anyway better that just Any type.

> Given that the other use case is the custom BrotliDecoder class, why don't we just make both classes inherit from this, rather than using a Protocol?

It is probably just a matter of taste. I like protocols because they just help to understand that a particular instance implements given methods or properties without applying too much of restrictions.

Dreamsorcerer: > The other protocol is different even though it has the same name.

OK, I see these ones are just used to annotate the internal compressobj(). But, I don't see the point of doing that, mypy should already know the type of those objects, meaning that these 2 classes can be removed.

> I like protocols because they just help to understand that a particular instance implements given methods or properties without applying too much of restrictions.

Protocols are useful when you don't control the classes, or don't know the classes ahead of time. But, these protocols are being used where we are assigning hardcoded internal classes, so inheritance makes more sense.

mykola-mokhnach: makes sense, removed

aiohttp/web_response.py
```diff
@@ -706,26 +707,26 @@ async def _do_start_compression(self, coding: ContentCoding) -> None:
...
+            assert self._body is not None
+            if self._zlib_executor_size is None and len(self._body) > 1024 * 1024:
+                warnings.warn(
+                    "Synchronous compression of large response bodies "
+                    f"({len(self._body)} bytes) might block the async event loop. "
+                    "Consider providing a custom value to zlib_executor_size/"
+                    "zlib_executor response properties or disabling compression on it."
                 )
-            else:
-                self._compress_body(zlib_mode)
```
Dreamsorcerer, line 722: Can we delete the `_compress_body()` method now?

mykola-mokhnach: oh, yes. Good catch

aiohttp/zlib_utils.py
```diff
@@ -0,0 +1,146 @@
...
+            encoding is None and wbits is None
+        ), "Either encoding or wbits must be provided"
+        super().__init__(
+            encoding=encoding,
+            suppress_deflate_header=suppress_deflate_header,
+            executor=executor,
+            max_sync_chunk_size=max_sync_chunk_size,
+        )
+        if wbits is not None:
+            self._mode = wbits
```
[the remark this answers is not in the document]
Dreamsorcerer: This does look a little odd to me.

aiohttp/http_parser.py
```diff
@@ -50,6 +49,18 @@
     HAS_BROTLI = False
 
 
+class Decompressor(Protocol):
```
Dreamsorcerer: I'd still like to see this replaced with inheritance.
Might also be nice to move the inline BrotliDecoder to the same module (renaming from zlib_utils to compress or similar).

aiohttp/zlib_utils.py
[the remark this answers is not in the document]
mykola-mokhnach: np, changed

aiohttp/http_parser.py
mykola-mokhnach: should it be a part of the same PR?

Dreamsorcerer: Yeah, still feels like it's part of refactoring the compressing logic.

mykola-mokhnach: done

mykola-mokhnach:
@Dreamsorcerer Thanks for checking and fine-tuning the PR. I appreciate your help

Dreamsorcerer:
Thanks a lot. If you could handle the backport as well, that'd be great.

PR #7251, opened 2023-04-08
Reduce access log overhead when logging is disabled with a logging filter

bdraco:
alternative to #7240

## What do these changes do?

If the logger was disabled the log message would be formatted and thrown away. Avoid formatting the log message when logging is not enabled at info level.

If you are using multiple `aiohttp.web.AppRunner` in the same python process you must also now pass a unique logger to each because `AccessLogger` will now attach a filter which affects the state of any logger object passed in and is not re-entrant.

Consumers of `AccessLogger` MUST use a new unique logger as the `AccessLogger`  will now add logging filter to any existing logger that is passed in which will affect logging for all handlers attached to that logger. Be sure to generate a unique logger object for each instance of AccessLogger. It is important the logger is unique because if other code is using the same logger, the logging filter will likely raise an exception or mutate the log messages in an unexpected way.

```python
        log = logging.getLogger("my_unique_logger_per_instance_of_aiohttp")
        log_format = "%a %{User-Agent}i"
        access_logger = AccessLogger(log, log_format)
        access_logger.log(request, response, time)
```

If multiple aiohttp instances use the same logger object then the log filter will be attached multiple times which will lead to undefined behavior.

## Checklist


bdraco:
After writing out `Are there changes in behavior for the user?` it seems like the side effects of this change are not going to be received well so I think this path doesn't have a way forward.

Issue #7070, opened 2022-11-07
Backslashes stripped from multipart form filenames


barneygale:
### Describe the bug

I'm dealing with a slightly wonky HTTP client that sends _full file paths_ in the 'filename' field of a form data part. Like:

```
POST /test.html HTTP/1.1
Host: example.org
Content-Type: multipart/form-data;boundary="boundary"

--boundary
Content-Disposition: form-data; name="uploadfile"; filename="C:\Users\me\example.txt"

value2
--boundary--
```

However, when I read the request with `await request.post()`, I find that the filename has been mangled to `C:Usersmeexample.txt`, i.e. the backslashes have been removed.

[The standards don't agree](https://github.com/curl/curl/issues/7789#issuecomment-930138204) on how backslashes should be supplied, but the [WHATWG spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart-form-data) seemed to be followed by the CURL folks, and it *doesn't* suggest that backslashes need to be doubled, nor escape-encoded. Therefore I think this is a bug in aiohttp.

### To Reproduce

1. Implement an HTTP client that sends a full Windows-style file path as the filename in a multipart form data request
2. Call `data = await request.post()` and print `data['uploadfield'].filename`

### Expected behavior

The original filename is reproduced

### Logs/tracebacks

```python-traceback
Actual behaviour: the backslashes are stripped.
```

### Python Version

```console
$ python --version
Python 3.10.4
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.8.1
Summary: Async http client/server framework (asyncio)
Home-page: https://github.com/aio-libs/aiohttp
Author: 
Author-email: 
License: Apache 2
Requires: aiosignal, async-timeout, attrs, charset-normalizer, frozenlist, multidict, yarl
Required-by: aiobotocore
```

### multidict Version

```console
$ python -m pip show multidict
Name: multidict
Version: 6.0.2
Summary: multidict implementation
Home-page: https://github.com/aio-libs/multidict
Author: Andrew Svetlov
Author-email: andrew.svetlov@gmail.com
License: Apache 2
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.8.1
Summary: Yet another URL library
Home-page: https://github.com/aio-libs/yarl/
Author: Andrew Svetlov
[...]

PR #7257, opened 2023-04-18
Fix stripping of backslashes in multipart filenames

barneygale:
## What do these changes do?

Adjust `aiohttp.multipart` to not "unescape" backslashes in multipart filenames.

Undoubled backslashes in multipart form data will no longer be stripped. Well-behaved clients will not send these anyway (they send only the basename, not the full path).

## Related issue number

- #7070

## Checklist

### DIFF
--- a/CHANGES/7070.bugfix
+++ b/CHANGES/7070.bugfix
@@ -1,0 +1,1 @@
+Fix issue where undoubled backslashes in multipart filenames were stripped.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -51,6 +51,7 @@ Arseny Timoniq
 Artem Yushkovskiy
 Arthur Darcet
 Austin Scola
+Barney Gale
 Ben Bader
 Ben Greiner
 Ben Timby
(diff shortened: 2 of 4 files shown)


Dreamsorcerer:
I've had another look, and I don't think this is right. It seem the escaping comes from the HTTP/1.1 spec.

The original code appears to be based off https://www.rfc-editor.org/rfc/rfc2616#section-2.2 (see quoted-string at the end of the section).

The newer RFC which obsoletes that one is https://www.rfc-editor.org/rfc/rfc7230#section-3.2.6

Where the old one defined the quoted-pair as "\\" followed by any ASCII character, the new one limits it to "\\" followed by tab, space or any visible character (and the "\\" is not allowed in any other place). So, I think the behaviour in these tests should remain the same according to the HTTP spec. The only change is that `CHAR` should be reduced to exclude control characters etc.

barneygale:
Yep I think you're right. RFC 7230 particularly:

> Recipients that process the value of a quoted-string MUST handle a quoted-pair as if it were replaced by the octet following the backslash.

I'll close this PR and the issue.

Dreamsorcerer:
No problem. If you'd like to make the small changes for RFC7230 mentioned above, I'd be happy to merge that though.

PR #7260, opened 2023-04-26
Update update_query calls to work with latest yarl

### LINKED ISSUE #7259: Tracing method "on_request_start" does not receive full URL
### Describe the bug

After a release in `yarl` which had a change in behaviour for URL's `update_query` https://github.com/aio-libs/yarl/issues/854#issuecomment-1521943597 method, the URL received in the `on_request_start` tracing hook lacks the query string part.

The problem only happens with yarl version 1.9.1. It works fine with previous version which is 1.8.2

### To Reproduce

1. Add a tracer
2. Implement `on_request_start` for the tracer
3. Issue a request that has a query string
4. Observe that the `params.url` does not have the query string.

### Expected behavior

I expect that the URL passed within `params` to the tracer hook contains the query string part as well and resembles the original URL exactly.

### Logs/tracebacks

```python-traceback
See https://github.com/aio-libs/yarl/issues/854#issuecomment-1521670598 please for the exact issue that's causing this problem
```

### Python Version

```console
$ python --version
Python 3.9.16 (main, Mar 15 2023, 15:55:54) 
[Clang 14.0.0 (clang-1400.0.29.202)] on darwin
```

### aiohttp Version

```console
$ python -m pip show aiohttp
3.8.4
```

### multidict Version

```console
6.0.4
```

### yarl Version

```console
$ python -m pip show yarl
1.9.1 (the issue is not seen with version 1.8.2 as explained above)
```

### OS

macOS

### Related component

Client

### Additional context

In this piece of the code https://github.com/aio-libs/aiohttp/blob/315ae9037438351b48a712734173020374e5ff33/aiohttp/client.py#L417 `params` is `None` and the url actually has the query string in it. But the call to `url.update_query(params)` is removing the query string which is passed to the tracing hook method.

### Code of Conduct

danigm:
This patch pass "{}" when params is "None" to the url.update_query to avoid setting the url params to None.

Related to this change in yarl:
https://github.com/aio-libs/yarl/commit/dd86b3435093b9ca251ecb7831346b92a3f16b25

Fix https://github.com/aio-libs/aiohttp/issues/7259

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -374,6 +374,7 @@ class ClientSession:
         redirects = 0
         history = []
         version = self._version
+        params = params or {}
 
         # Merge with default headers and transform to CIMultiDict
         headers = self._prepare_headers(headers)
@@ -613,7 +614,7 @@ class ClientSession:
                             headers.pop(hdrs.AUTHORIZATION, None)
 
                         url = parsed_url
-                        params = None
+                        params = {}
                         resp.release()
                         continue
 
--- a/CHANGES/7259.bugfix
+++ b/CHANGES/7259.bugfix
@@ -1,0 +1,1 @@
+Fixed missing query in tracing method URLs when using ``yarl`` 1.9+.
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -11,4 +11,4 @@ charset-normalizer==2.0.12
 frozenlist==1.3.1
 gunicorn==20.1.0
 uvloop==0.14.0; platform_system!="Windows" and implementation_name=="cpython" and python_version<"3.9" # MagicStack/uvloop#14
-yarl==1.8.1
+yarl==1.9.2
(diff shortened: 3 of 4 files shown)


aiohttp/client.py
```diff
@@ -428,7 +428,9 @@ async def _request(
         ]
 
         for trace in traces:
-            await trace.send_request_start(method, url.update_query(params), headers)
+            await trace.send_request_start(
+                method, url.update_query(params or {}), headers
```
Dreamsorcerer: Does it make more sense to just do `params = params or {}` further up, rather than repeating the same code 4 times?

aiohttp/client.py
Dreamsorcerer, line 1: If you could also merge in #7261, that'd be perfect.

aiohttp/client.py
danigm: done

PR #7265, opened 2023-04-29
♻️ Delete `size` arg from `StreamReader.feed_data`

dzhalaevd:
## What do these changes do?
Removed the unused `size` parameter from the `feed_data` method to simplify the code and usage of the method.

If the user is using the feed_data method without the size parameter, there should be no changes in the behavior of the method for them. However, if the size parameter was used, there may be changes in the behavior of the method.

## Checklist


CHANGES/7265.feature
```diff
@@ -0,0 +1 @@
+The unused size parameter in the feed_data method has been removed
```
webknjaz: It's not a feature but a breaking change in the public API, probably.

aiohttp/streams.py
```diff
@@ -228,7 +228,7 @@ def unread_data(self, data: bytes) -> None:
         self._eof_counter = 0
 
     # TODO: size is ignored, remove the param later
```
Dreamsorcerer:
```suggestion
```

aiohttp/http_parser.py
```diff
@@ -729,13 +729,13 @@ def feed_data(
...
+                self.payload.feed_data(chunk)
                 if self._length == 0:
                     self.payload.feed_eof()
                     return True, b""
             else:
                 self._length = 0
-                self.payload.feed_data(chunk[:required], required)
+                self.payload.feed_data(chunk[:required])
                 self.payload.feed_eof()
                 return True, chunk[required:]
```
Dreamsorcerer: Maybe some further improvement to do this:
```
self._length = max(required - len(chunk), 0)
self.payload.feed_data(chunk[:required])
if self._length == 0:
    self.payload.feed_eof()
    return True, chunk[required:]
```

And remove the chunk_len variable on the line above.

aiohttp/http_parser.py
```diff
@@ -783,11 +783,11 @@ def feed_data(
...
-                        self.payload.feed_data(chunk, chunk_len)
+                        self.payload.feed_data(chunk)
                         return False, b""
                     else:
                         self._chunk_size = 0
-                        self.payload.feed_data(chunk[:required], required)
+                        self.payload.feed_data(chunk[:required])
                         chunk = chunk[required:]
                         self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
                         self.payload.end_http_chunk_receiving()
```
Dreamsorcerer: Same again here:
```
self._chunk_size = max(required - len(chunk), 0)
self.payload.feed_data(chunk[:required])
if self._chunk_size:
    return False, b""

chunk = chunk[required:]
...
```

And another chunk_len variable can be removed from the line above.

webknjaz:
It wouldn't hurt to write actual full commit messages and PR titles. When I get a notification about "something happened with todo" or "some improvement", those don't look important enough to compell me not to skip the notifications. I have hundreds of those and will choose things that look meaningful and related to what I care about.

aiohttp/http_parser.py
```diff
@@ -779,18 +771,12 @@ def feed_data(
...
-                        self._chunk_size = required - chunk_len
-                        self.payload.feed_data(chunk, chunk_len)
+                    if self._chunk_size:
                         return False, b""
-                    else:
-                        self._chunk_size = 0
-                        self.payload.feed_data(chunk[:required], required)
-                        chunk = chunk[required:]
-                        self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
-                        self.payload.end_http_chunk_receiving()
```
Dreamsorcerer, line 793: I think these 2 lines are still needed.

aiohttp/http_parser.py
```diff
@@ -779,18 +771,14 @@ def feed_data(
...
-                        return False, b""
-                    else:
-                        self._chunk_size = 0
-                        self.payload.feed_data(chunk[:required], required)
-                        chunk = chunk[required:]
+                    if self._chunk_size:
                         self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
                         self.payload.end_http_chunk_receiving()
+                        return False, b""
+                    chunk = chunk[required:]
```
```suggestion
                        return False, b""
                    chunk = chunk[required:]
                    self._chunk = ChunkState.PARSE_CHUNKED_CHUNK_EOF
                    self.payload.end_http_chunk_receiving()
```

This was in the else condition before...

CHANGES/7265.feature
dzhalaevd: So, I should use `7265.removal`?

dzhalaevd:
CI also crashes due to the missing of the `size` arg and for some other reasons, however I do not know how to fix that. And I would really appreciate it if you could take care of it. Thanks!

Dreamsorcerer:
Error seems pretty clear:
`aiohttp/http_parser.py:729: error: Missing positional argument "size" in call to "feed_data" of "DeflateBuffer"  [call-arg]`

Method is defined at: https://github.com/DavidRomanovizc/aiohttp/blob/master/aiohttp/http_parser.py#L854

Dreamsorcerer:
Presumably, the same changes can be made, using `len()` instead of the `size` parameter.

tests/test_http_parser.py
```diff
@@ -921,13 +921,14 @@ async def test_parse_eof_payload(self, stream: Any) -> None:
...
         assert out.is_eof()
-        assert [(bytearray(b"data"), 4)] == list(out._buffer)
+        assert [(b"data",)] == [(data,) for (data, _) in out._buffer]
 
     async def test_parse_no_body(self, stream: Any) -> None:
         out = aiohttp.FlowControlDataQueue(
             stream, 2**16, loop=asyncio.get_event_loop()
         )
         p = HttpPayloadParser(out, method="PUT")
+        p.feed_eof()
```
Dreamsorcerer, line 931: I'm unclear why these 2 need to change?

Dreamsorcerer: I'm unclear why these 2 need to change?

tests/test_streams.py
```diff
@@ -1135,7 +1135,7 @@ def test_at_eof(self, buffer: Any) -> None:
 
     def test_feed_data(self, buffer: Any) -> None:
         item = object()
-        buffer.feed_data(item, 1)
+        buffer.feed_data(item)
         assert [(item, 1)] == list(buffer._buffer)
```
Dreamsorcerer: Looks like this becomes `0`, as the default behaviour appears to skip parsing (which is why the `object()` doesn't trigger any errors).

@webknjaz Do you think there's a chance this might break anything?

tests/test_http_parser.py
dzhalaevd: This tests  have failed, and I thought, since we removed the `size` arg, then we need to little changed the assert in the test 🤔

Dreamsorcerer: Maybe, but I'd like to understand why these were not needed before. It's not obvious how these changes relate to a size parameter being removed, and could be an indication that some behaviour may break for users.

dzhalaevd: I can suggest that we have something wrong  in `HttpPayloadParser.feed_data`

Dreamsorcerer: Maybe, but what? We need to understand exactly why the change is needed.

Dreamsorcerer: Uhh, well, you've reverted the second change, and the test still seems to be passing. So, doesn't look like anything was broken anyway...

What about the first change from bytearray(), are you sure that's needed? If so, why?

dzhalaevd: So, earlier in `HttpPayloadParser.feed_data`, the `feed_data.payload` method was called which saved `chunk` and `len(chunk)`. Now that we have removed the `size` arg, we cannot save `len(chunk)` anymore. I believe to resolve this issue, we can modify the assertion like this:

```python
assert [bytearray(b"data")] == [item[0] for item in out._buffer]
```
(This is another option, not the one currently in my branch. If everything goes well, I can push it)

In two cases we compare data, but in old test we compared and len(data). But I think if the data somehow changes, then so does the size. 

Similarly, in `TestDataQueue.test_feed_data`, we need to replace the assertion with:

```python
assert [item] == [data[0] for data in buffer._buffer]
```

The reason is the same: we need to modify the assertion to compare only the first element of each tuple in `buffer._buffer` with the expected value of `item`.

Therefore, I am confident in changing the test. And also on my pc, all tests passed with these changes.

aiohttp/http_parser.py
```diff
@@ -863,10 +851,11 @@ def __init__(self, out: StreamReader, encoding: Optional[str]) -> None:
...
         self.out.set_exception(exc)
 
-    def feed_data(self, chunk: bytes, size: int) -> None:
-        if not size:
+    def feed_data(self, chunk: bytes) -> None:
+        if not chunk:
             return
 
+        size = len(chunk)
         self.size += size
```
```suggestion
        self.size += len(chunk)
```

tests/test_http_parser.py
Dreamsorcerer: I've looked into this, and that doesn't seem to be correct.
The buffer these tests refer to are `streams.DataQueue`, where the size parameter is still expected:
https://github.com/DavidRomanovizc/aiohttp/blob/master/aiohttp/streams.py#L559

So, maybe we want to drop the size parameter from that class as well? Then these changes can be reverted (only dropping the `4` from the test comparison).

tests/test_streams.py
Dreamsorcerer: This is again because of the DataQueue. If that is updated, then this test should be fine as well.

aiohttp/streams.py
```diff
@@ -635,10 +632,10 @@ def __init__(
         self._protocol = protocol
         self._limit = limit * 2
 
-    def feed_data(self, data: _T, size: int = 0) -> None:
-        super().feed_data(data, size)
+    def feed_data(self, data: _T) -> None:
+        super().feed_data(data)
 
-        if self._size > self._limit and not self._protocol._reading_paused:
+        if self._size < self._limit and not self._protocol._reading_paused:
```
Dreamsorcerer: Why is this changed? Surely we don't want to pause reading if the size is still less than the limit...

aiohttp/streams.py
```diff
@@ -580,9 +579,8 @@ def set_exception(self, exc: BaseException) -> None:
             self._waiter = None
             set_exception(waiter, exc)
 
-    def feed_data(self, data: _T, size: int = 0) -> None:
-        self._size += size
```
Dreamsorcerer, line 582: Why is this removed? This will surely break the limit functionality. Just use `len(data)` to calculate the actual size?

aiohttp/streams.py
```diff
@@ -608,8 +606,7 @@ async def read(self) -> _T:
                 raise
 
         if self._buffer:
-            data, size = self._buffer.popleft()
-            self._size -= size
```
Dreamsorcerer, line 610: ^

dzhalaevd: Because if we don't delete it we get error in almost all tests the `TypeError error: unsupported operand type(s) for -=: 'int' and 'StreamReader'`. In tests/test_client_functional.py and he refers to this block of code

aiohttp/streams.py
dzhalaevd: And if we don't delete this line, the `TestFlowControlDataQueue.test_resume_on_read` test will fall

aiohttp/streams.py
Dreamsorcerer: `len(data)`? Obviously you can't subtract a StreamReader from an int...

Though if it's possible for `data` to be something other than bytes, then maybe this won't work.

dzhalaevd: OK, but we will still have tests falling in TestDataQueue. With TypeError error: object of type 'object' has no len(). Now it seems that we need to use `b"something"` instead of `object()` or handle error, but I'm not sure about that.

Dreamsorcerer: Well, object() is probably unreasonable, so we could maybe update that test... I'll have a look through later.

dzhalaevd: I added types checking to feed_data and now those tests don't crash, but I doubt the correctness of such a solution

aiohttp/streams.py
dzhalaevd: Under this condition, for some reason, the `test_feed_pause` test began falled

aiohttp/streams.py
DavidRomanovizc: @Dreamsorcerer could we resume work on this PR?

PR #7272, opened 2023-05-07
Fix Read The Docs config

Dreamsorcerer:


webknjaz:
What was broken?

Dreamsorcerer:
Wrong version of ssl. Seems it needs the OS set explicitly to the latest version (maybe that `build: latest` is a setting that has been removed?).

webknjaz:
Oh, that's because urllib3 v2 is finally out (and unpinned!)
Fair enough..

Dreamsorcerer:
Also, did you disable the admin bypass for branch protection recently? I wouldn't care, but the CI currently always waits 30 mins for the Windows test to timeout and fail before it merges things..

webknjaz:
I think I did at some point. OTOH, you shouldn't have to babysit the CI if you use automerge. If the windows jobs time out and you know what a typical success timeline looks like, I'd recommend setting that timeout for said jobs to improve responsiveness..

Dreamsorcerer:
It's just the 3.11 broken one. While the Mac one fails quickly, the Windows one has a test which takes 5 mins to timeout (and then retries that failed test).

PR #7275, opened 2023-05-08
Preserve default total client timeout value

### LINKED ISSUE #7274: Preserve default total timeout when other timeout is changed in ClientSession.__init__()

Original discussion #7268

It's easy to unintentionally overwrite the default **total** timeout value when changing another, more specific, client timeout:

```
timeout = ClientTimeout()
print(timeout.total) # 300, default value

timeout = ClientTimeout(connect=5.0)
print(timeout.total) # None, default value is lost
```

`ClientTimeout.total` should be preserved when any other timeout is changed, but the total timeout value is not explicitly specified.

As suggested in the original discussion, the default 5-minute timeout value should be set directly in the `ClientTimeout` class, not later in the `ClientSession`.

An alternative solution could've been changing the documentation and describing the behavior there, leaving the responsibility to explicitly specify the total timeout value to the user.

### Related component

Client

### Additional context

The issue is created in preparation for the PR.

### Code of Conduct

wombatonfire:
## What do these changes do?

Changed default value for ``ClientTimeout.total`` so that the default is always present, even if not explicitly specified.

If a user wants to adjust some specific client timeout, i.e. `connect` or `sock_connect`, they won't need to explicitly specify `total` timeout value in order to preserve it.

## Related issue number

Closes #7274

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -31,7 +31,7 @@ from typing import (
 )
 
 from multidict import CIMultiDict, MultiDict, MultiDictProxy, istr
-from typing_extensions import Final, final
+from typing_extensions import final
 from yarl import URL
 
 from . import hdrs, http, payload
@@ -134,7 +134,7 @@ except ImportError:  # pragma: no cover
 
 @dataclasses.dataclass(frozen=True)
 class ClientTimeout:
-    total: Optional[float] = None
+    total: Optional[float] = 5 * 60  # 5 minute default timeout
     connect: Optional[float] = None
     sock_read: Optional[float] = None
     sock_connect: Optional[float] = None
@@ -153,9 +153,6 @@ class ClientTimeout:
     # - or use https://docs.python.org/3/library/dataclasses.html#dataclasses.replace
     # to overwrite the defaults
 
-
-# 5 Minute default read timeout
-DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 
 _RetType = TypeVar("_RetType")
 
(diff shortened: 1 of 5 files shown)


wombatonfire:
Not sure if the documentation should be updated. Currently, it does not describe the behavior that this PR is changing, and the new behavior is somewhat expected, IMHO.

Dreamsorcerer:
> Currently, it does not describe the behavior that this PR is changing

It says it defaults to None:
https://docs.aiohttp.org/en/stable/client_reference.html?highlight=ClientTimeout#clienttimeout

wombatonfire:
> It says it defaults to None:
> https://docs.aiohttp.org/en/stable/client_reference.html?highlight=ClientTimeout#clienttimeout

I guess I need glasses :) Updated.

aiohttp/client.py
```diff
@@ -134,7 +134,7 @@
 
 @dataclasses.dataclass(frozen=True)
 class ClientTimeout:
-    total: Optional[float] = None
+    total: Optional[float] = 5 * 60  # 5 minute default timeout
```
Dreamsorcerer, line 137: On second thought, is this the expected behaviour if a user were to do something like
`ClientTimeout(sock_read=600)`?

i.e. They set sock_read to 10 mins, but total is still set to 5 mins, so it achieves nothing.

I'm not sure now if having this default value makes things simpler or more complex...
@webknjaz Any thoughts?

webknjaz: Yeah, I'm not sure. Perhaps, there should be some sort of validation against causing such cases accidentally. We should get more comprehensive coverage for these situations before going forward.

Dreamsorcerer: I think it probably makes the most sense to make the total a max() of all the timeouts.

Dreamsorcerer:
Continuing in #7275

PR #7319, opened 2023-06-10
Refactor WSMessage to use tagged unions

### LINKED ISSUE #7313: websocket: message.json() fails with a TypeError when the message exceeds the size limit
### Describe the bug

When a websocket message exceeds the size limit, calling message.json() raises a TypeError instead of an informative exception

### To Reproduce

Send a large websocket message to an aiohttp server.

In the handler, calling message.json() raises TypeError:

    TypeError: the JSON object must be str, bytes or bytearray, not WebSocketError

`message.data` is set to a WebSocketError instance:

    aiohttp.http_websocket.WebSocketError: Message size 29818 exceeds limit 8192

(I lowered the limit to more easily trigger the problem)

### Expected behavior

I would expect calling `message.json()` to raise the WebSocketError.

### Logs/tracebacks

```python-traceback
  File "/Users/just/code/git/BlackFoundry/fontra/src/fontra/core/remote.py", line 47, in _handleConnection
    message = message.json()
              ^^^^^^^^^^^^^^
  File "/Users/just/code/git/BlackFoundry/fontra/venv/lib/python3.11/site-packages/aiohttp/http_websocket.py", line 98, in json
    return loads(self.data)
           ^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/json/__init__.py", line 339, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not WebSocketError
```

### Python Version

```console
Python 3.11.2
```

### aiohttp Version

```console
Name: aiohttp
Version: 3.8.4
```

### multidict Version

```console
Name: multidict
Version: 6.0.4
```

### yarl Version

```console
Name: yarl
Version: 1.8.2
```

### OS

macOS 10.15

### Related component

Server

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
This reworks WSMessage to be a union of dataclasses which can provide much better type safety using tagged unions (on the `type` attribute).

I need to go over a couple of details (like what should be exported and from where, imports are a little messy currently). But, my main question before I finish it off, is whether it makes sense to have `.json()` only on the TEXT/BINARY messages? If we do this, it becomes easy for mypy to validate the code, requiring a `msg.type is WSMsgType.TEXT` check before calling `.json()`.

Fixes #7313


bdraco:
Creating the dataclasses will be significantly slower than the NamedTuple, but the property accesses will be faster...probably need to benchmark it to see what the tradeoff will be

Dreamsorcerer:
> Creating the dataclasses will be significantly slower than the NamedTuple, but the property accesses will be faster...probably need to benchmark it to see what the tradeoff will be

A quick test locally suggests that NamedTuple will also work with unions. Let me know which you want to go with and I'll update (and then try to revert the micro-optimisations that got caught in conflict resolutions).

bdraco:
I'll write a benchmark for all the options later today

aiohttp/http_websocket.py
```diff
@@ -428,25 +496,23 @@ def _feed_data(self, data: bytes) -> None:
                         raise WebSocketError(
                             WSCloseCode.INVALID_TEXT, "Invalid UTF-8 text message"
                         ) from exc
-                    msg = tuple.__new__(
```
bdraco, line 431: It would be some much nicer if we could lazy decode these since it results in a lot of cycles that could be avoided

bdraco:
Options

```
from dataclasses import dataclass
from typing import NamedTuple
from aiohttp.http_websocket import WSMsgType
from typing import Optional, Literal
import timeit

BINARY_TYPE = WSMsgType.BINARY


@dataclass
class _WSMessageDataClass:
    data: object
    type: WSMsgType
    extra: Optional[str] = None


@dataclass
class WSMessageBinaryDataClass(_WSMessageDataClass):
    data: bytes
    type: Literal[WSMsgType.BINARY] = BINARY_TYPE


@dataclass(frozen=True)
class _WSMessageFrozenDataClass:
    data: object
    type: WSMsgType
    extra: Optional[str] = None


@dataclass(frozen=True)
class WSMessageBinaryFrozenDataClass(_WSMessageFrozenDataClass):
    data: bytes
    type: Literal[WSMsgType.BINARY] = BINARY_TYPE


class _WSMessageNamedTuple(NamedTuple):
    data: object
    type: WSMsgType
    extra: Optional[str] = None


class WSMessageBinaryNamedTuple(_WSMessageNamedTuple):
    data: bytes
    type: Literal[WSMsgType.BINARY]


data = b"Hello, world!"

named_tuple_direct = timeit.timeit(
    "x=WSMessageBinaryNamedTuple(data, BINARY_TYPE);x.data;x.type",
    globals=locals(),
    number=1000000,
)
named_tuple_tuple_new = timeit.timeit(
    "x=tuple.__new__(WSMessageBinaryNamedTuple, (data, BINARY_TYPE));x.data;x.type",
    globals=locals(),
    number=1000000,
)
dataclass_direct = timeit.timeit(
    "x=WSMessageBinaryDataClass(data);x.data;x.type", globals=locals(), number=1000000
)
frozen_dataclass_direct = timeit.timeit(
    "x=WSMessageBinaryFrozenDataClass(data);x.data;x.type",
    globals=locals(),
    number=1000000,
)

print(f"named_tuple_direct: {named_tuple_direct}")
print(f"named_tuple_tuple_new: {named_tuple_tuple_new}")
print(f"dataclass_direct: {dataclass_direct}")
print(f"frozen_dataclass_direct: {frozen_dataclass_direct}")
```


named_tuple_direct: 0.13957041699904948
named_tuple_tuple_new: 0.09394508300465532
dataclass_direct: 0.11717991600744426
frozen_dataclass_direct: 0.26404679199913517

`WSMessageBinaryNamedTuple(data, BINARY_TYPE)`: 0.13957041699904948
`tuple.__new__(WSMessageBinaryNamedTuple, (data, BINARY_TYPE))`: 0.09394508300465532
`WSMessageBinaryDataClass(data)`: 0.11717991600744426
`WSMessageBinaryFrozenDataClass(data)`: 0.26404679199913517

bdraco:
If you rebase, you should get a better idea of the performance impact, however from looking at the https://codspeed.io/aio-libs/aiohttp/branches/websocket_benchmarks it looks like changes here may be shadowed by the performance of `parse_frame`
Screenshot 2024-10-25 at 1 04 29 PM

Dreamsorcerer:
Think we're all good now.

bdraco (approved):

PR #7333, opened 2023-07-01
Fix TLS transport is `None` error

### LINKED ISSUE #3355: Request.transport is sometimes None
## Long story short

The transport attribute on requests is sometimes (like 1/month on a rather idle system!?) None – I cannot figure out why?

## Expected behaviour

Either never be None or an explanation in http://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.transport

Could it be some kind of cancellation behavior!?

## Actual behaviour

I get an email from Sentry that `'NoneType' object has no attribute 'get_extra_info'`.

## Steps to reproduce

I have no idea. I don’t even have an idea whether it’s a behavioral or documentation bug. :(

## Your environment

Bionic Beaver in a Docker container; aiohttp 3.4.4. This doesn’t appear to be a recent regression though, I just lived with it because I had no clue what to do about it.

Dreamsorcerer:
Fixes #3355

The type has been fixed in `typeshed` as well, so the bug here would be highlighted in mypy 1.5+.
https://github.com/python/typeshed/pull/10346#issuecomment-1603046653

### DIFF
--- a/CHANGES/3355.bugfix
+++ b/CHANGES/3355.bugfix
@@ -1,0 +1,1 @@
+Fixed a transport is ``None`` error. -- by :user:`Dreamsorcerer`
--- a/CHANGES/3355.bugfix
+++ b/CHANGES/3355.bugfix
@@ -1,1 +1,1 @@
-Fixed a transport is ``None`` error. -- by :user:`Dreamsorcerer`
+Fixed a transport is :data:`None` error -- by :user:`Dreamsorcerer`.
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
@webknjaz Can you just double check that looks like the right behaviour?

Dreamsorcerer:
@webknjaz This one could probably go in the next release, if you think the change looks reasonable.

CHANGES/3355.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a transport is ``None`` error. -- by :user:`Dreamsorcerer`
```
webknjaz:
```suggestion
Fixed a transport is :data:`None` error -- by :user:`Dreamsorcerer`.
```

PR #7364, opened 2023-07-12
Strip trailing dot from FQDNs in Host and TLS

### LINKED ISSUE #3636: aiohttp raises SSLError when requesting URLs with FQDN
## Long story short

It seems that aiohttp doesn't work properly when requesting URLs with Fully Qualified Domain Name. This was fixed in urllib3, see https://github.com/urllib3/urllib3/pull/1255 It should probably by fixed also in aiohttp. May be related to https://github.com/aio-libs/aiohttp/issues/3171

## Expected behaviour

aiohttp works when requesting URLs with FQDN

## Actual behaviour

aiohttp raises SSL error:

```python-traceback
  File "/home/pavel/.pyenv/versions/3.7.1/lib/python3.7/asyncio/base_events.py", line 573, in run_until_complete
    return future.result()
  File "<stdin>", line 3, in main
  File "<stdin>", line 2, in fetch
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/client.py", line 1005, in __aenter__
    self._resp = await self._coro
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/client.py", line 476, in _request
    timeout=real_timeout
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/connector.py", line 522, in connect
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/connector.py", line 854, in _create_connection
    req, traces, timeout)
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/connector.py", line 992, in _create_direct_connection
    raise last_exc
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/connector.py", line 974, in _create_direct_connection
    req=req, client_error=client_error)
  File "/home/pavel/Projects/_lab/aiohttp/lib/python3.7/site-packages/aiohttp/connector.py", line 927, in _wrap_create_connection
    req.connection_key, exc) from exc
aiohttp.client_exceptions.ClientConnectorCertificateError: Cannot connect to host github.com.:443 ssl:True [SSLCertVerificationError: (1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch, certificate is not valid for 'github.com.'. (_ssl.c:1051)")]
```

## Steps to reproduce

Run the following code:

```python
import aiohttp
import asyncio


        return await response.text() 
 
async def main(): 
        html = await fetch(session, 'https://github.com.') 
        print(html)


```

## Your environment

Python 3.7.1
Ubuntu 18.04

pip freeze

```console
aiohttp==3.5.4
async-timeout==3.0.1
attrs==19.1.0
chardet==3.0.4
idna==2.8
multidict==4.5.2
yarl==1.3.0
```

martin-sucha:
The TLS verification fails with an exception if the client uses a fully-qualified domain name with a trailing dot, like https://github.com./ :
```console
aiohttp.client_exceptions.ClientConnectorCertificateError: Cannot connect to host github.com.:443 ssl:True
[SSLCertVerificationError: (1, "[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: Hostname mismatch,
certificate is not valid for 'github.com.'. (_ssl.c:1051)")]
```
The reason is that TLS certificates do not contain the trailing dot, as per RFC 6066:

"HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client.  The hostname is represented as a byte
   string using ASCII encoding without a trailing dot.

We need to strip the trailing dot for TLS context and Host header, where trailing dots are not present.
For DNS resolution, we need to include the trailing dot as it signifies a fully-qualified domain name (FQDN).
DNS lookups of FQDNs are faster as the resolver does not need to check DNS search path, like for relative DNS names.

Closes https://github.com/aio-libs/aiohttp/issues/3636

## What do these changes do?

Strip trailing dot from hostname in Host header and TLS context.

Allows client to connect to server if URL has dot at the end of the hostname: `https://example.com./`

## Related issue number

https://github.com/aio-libs/aiohttp/issues/3636

## Checklist


martin-sucha:
I tried looking into adding a test for this but it was not clear to me if there is some test I could easily modify to check. Please advise what is the best way to write a test for this.

Dreamsorcerer:
> I tried looking into adding a test for this but it was not clear to me if there is some test I could easily modify to check. Please advise what is the best way to write a test for this.

I'd assume tests in test_client_request.py and test_connector.py could be added that atleast asserts that something is called with the truncated domain?

CHANGES/3636.bugfix
```diff
@@ -0,0 +1,2 @@
+Strip trailing dot in Host header and TLS context from fully-qualified domain names.
```
webknjaz:
```suggestion
Implemented stripping the trailing dots from fully-qualified domain names in ``Host`` headers and TLS context when acting as an HTTP client.
```

CHANGES/3636.bugfix
```diff
@@ -0,0 +1,2 @@
...
+This allows the client to connect to URLs with FQDN hostname like `https://example.com./`
```
webknjaz: This is not markdown but RST.
```suggestion
This allows the client to connect to URLs with FQDN hostname like ``https://example.com./``.
-- by :user:`martin-sucha`.
```

aiohttp/client_reqrep.py
```diff
@@ -340,6 +340,9 @@ def update_headers(self, headers: Optional[LooseHeaders]) -> None:
         netloc = cast(str, self.url.raw_host)
         if helpers.is_ipv6_address(netloc):
             netloc = f"[{netloc}]"
+        elif netloc.endswith("."):
+            # Strip trailing dot. See https://github.com/aio-libs/aiohttp/issues/3636.
+            netloc = netloc[:-1]
```
webknjaz: This needs coverage, but also it'd make sense to apply `str.rstrip()` like urllib3 does, to handle multiple trailing dots like Google Chrome.

Refs:
* https://github.com/urllib3/urllib3/pull/1255
* https://github.com/urllib3/urllib3/issues/1254#issuecomment-347786089
* https://github.com/aio-libs/aiohttp/issues/3636#issuecomment-469641380

aiohttp/connector.py
```diff
@@ -1117,6 +1117,13 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
             port = hinfo["port"]
 
             try:
+                server_hostname = None
+                if sslcontext:
+                    server_hostname = hinfo["hostname"]
+                    if server_hostname.endswith("."):
+                        # Strip trailing dot, certificates contain FQDN without dot.
+                        # See https://github.com/aio-libs/aiohttp/issues/3636
+                        server_hostname = server_hostname[:-1]
```
webknjaz: Same, let's strip all the dots.

aiohttp/connector.py
```diff
@@ -1117,6 +1117,13 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
             port = hinfo["port"]
 
             try:
+                server_hostname = None
+                if sslcontext:
+                    server_hostname = hinfo["hostname"]
+                    if server_hostname.endswith("."):
```
webknjaz: You can save some nesting by doing stripping unconditionally.

aiohttp/connector.py
webknjaz: Let's improve readability by getting rid of two unnecessary nesting levels:
```suggestion
                # Strip trailing dot, certificates contain FQDN without dot.
                # See https://github.com/aio-libs/aiohttp/issues/3636
                server_hostname = hinfo["hostname"].rstrip(".") if sslcontext else None
```

webknjaz:
> I tried looking into adding a test for this but it was not clear to me if there is some test I could easily modify to check. Please advise what is the best way to write a test for this.

You'll probably want to use these fixtures https://github.com/aio-libs/aiohttp/blob/7911f1e/tests/conftest.py#L50-L61 to test that this patch works well with TLS.

martin-sucha:
Thank you for the review. I've applied the suggestions and added tests.

webknjaz:
@martin-sucha hey, it looks like some linting checks are failing. Could you fix those so we could proceed?

aiohttp/client_reqrep.py
```diff
@@ -340,6 +340,9 @@ def update_headers(self, headers: Optional[LooseHeaders]) -> None:
         netloc = cast(str, self.url.raw_host)
         if helpers.is_ipv6_address(netloc):
             netloc = f"[{netloc}]"
+        else:
+            # See https://github.com/aio-libs/aiohttp/issues/3636.
+            netloc = netloc.rstrip(".")
```
webknjaz: Just put it before the if-block, no need for it to be conditional.

tests/test_client_request.py
```diff
@@ -279,6 +279,23 @@ def test_host_header_ipv6_with_port(make_request: Any) -> None:
     assert req.headers["HOST"] == "[::2]:99"
 
 
+def test_host_header_fqdn(make_request: Any) -> None:
+    req = make_request("get", "http://python.org.:99")
+    assert req.headers["HOST"] == "python.org:99"
+
+
+def test_host_header_fqdn_multiple_dots(make_request: Any) -> None:
+    req = make_request("get", "http://python.org...:99")
```
webknjaz: I think these tests could be collapsed into one with `@pytest.mark.parametrize`. Use custom IDs to label the test report.

Also, I'd like to see some cases with `https://` and also with `localhost` (or anything without a dot in the middle). Include cases with and without an explicit port.

As a separate negative test, let's check for cases with only dots, like `https://...`.

tests/test_connector.py
```diff
@@ -2064,6 +2064,39 @@ async def handler(request):
...
+    app.router.add_get("/", handler)
+
+    srv = await aiohttp_server(app, ssl=ssl_ctx)
+    port = unused_port()
+    conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", port))
+
+    session = aiohttp.ClientSession(connector=conn)
+    url = srv.make_url("/")
+
+    r = await session.get(url.with_host("localhost."), ssl=client_ssl_ctx)
```
webknjaz: Let's parametrize this too, just like I asked in the other test module...

martin-sucha: Parametrized the `test_tcp_connector_do_not_raise_connector_ssl_error` test.

* Can't add `example.org` because it does not resolve to local host, so the connection fails.
* Can't add `localhost...` because the resolution fails during IDNA parsing (that processes each component separately).

I don't think there is any value in supporting multiple dots at the end of the host name, so I did not change the IDNA code. Please let me know your thoughts about this.

Dreamsorcerer (approved):

aiohttp/connector.py
```diff
@@ -1117,6 +1117,9 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
             port = hinfo["port"]
 
             try:
+                # Strip trailing dots, certificates contain FQDN without dots.
+                # See https://github.com/aio-libs/aiohttp/issues/3636
+                server_hostname = hinfo["hostname"].rstrip(".") if sslcontext else None
```
webknjaz: I just realized this violates the rule of having only one instruction in a try-block. Let's fix this:
```suggestion
            # Strip trailing dots, certificates contain FQDN without dots.
            # See https://github.com/aio-libs/aiohttp/issues/3636
            server_hostname = hinfo["hostname"].rstrip(".") if sslcontext else None
            try:
```

tests/test_client_request.py
```diff
@@ -279,6 +279,34 @@ def test_host_header_ipv6_with_port(make_request: Any) -> None:
     assert req.headers["HOST"] == "[::2]:99"
 
 
+@pytest.mark.parametrize(
+    "url,headers,expected",
```
webknjaz: It's visually nicer to have param names as separate string literals:
```suggestion
    ("url", "headers", "expected"),
```

tests/test_connector.py
webknjaz: > I don't think there is any value in supporting multiple dots at the end of the host name, so I did not change the IDNA code. Please let me know your thoughts about this.

I think this is valuable in terms of consistency with what other popular projects do, so long as it doesn't violate RFCs.

@martin-sucha could you fix this too? I think we should be consistent across the entire code base, rather than supporting this only in some places. Where/how does this fail exactly?

tests/test_connector.py
```diff
@@ -2031,8 +2031,21 @@ async def handler(request):
...
+    [
+        "127.0.0.1",
+        "localhost",
+        "localhost.",
+    ],
+    ids=(
+        "ip address",
+        "domain name",
+        "fully-qualified domain name",
+    ),
```
webknjaz: It's usually better to match ids with params when they are coupled together. When the list grows, it's harder to visually match them...
```suggestion
    (
        pytest.param("127.0.0.1", id="ip address"),
        pytest.param("localhost", id="domain name"),
        pytest.param("localhost.", id="fully-qualified domain name"),
    ),
```

tests/test_client_request.py
```diff
@@ -279,6 +279,34 @@ def test_host_header_ipv6_with_port(make_request: Any) -> None:
...
+    ],
+    ids=(
+        "dot only at the end",
+        "single dot",
+        "single dot with port",
+        "multiple dots with port",
+        "explicit host header",
+        "https",
+        "only dots",
+    ),
```
webknjaz: let's match these ids better too
```suggestion
    (
        pytest.param("http://localhost.", None, "localhost", id="dot only at the end"),
        pytest.param("http://python.org.", None, "python.org", id="single dot"),
        pytest.param("http://python.org.:99", None, "python.org:99", id="single dot with port"),
        pytest.param("http://python.org...:99", None, "python.org:99", id="multiple dots with port"),
        pytest.param("http://python.org.:99", {"host": "example.com.:99"}, "example.com.:99", id="explicit host header"),
        pytest.param("https://python.org.", None, "python.org", id="https"),
        pytest.param("https://...", None, "", id="only dots"),
    ),
```

webknjaz:
@martin-sucha I've completed another round of review. Sorry it too so long. Could you address it?

tests/test_connector.py
martin-sucha: That looks better, thanks!

tests/test_connector.py
martin-sucha: > so long as it doesn't violate RFCs.

I can't say it does not as I haven't verified it. Personally I prefer the more strict/conservative handling of not allowing multiple dots at the end.

> could you fix this too? I think we should be consistent across the entire code base, rather than supporting this only in some places.

Okay, added the stripping to `_create_dirrect_connection`.

> Where/how does this fail exactly?

```
../aiohttp/connector.py:856: in _resolve_host
    addrs = await self._resolver.resolve(host, port, family=self._family)
        host       = 'localhost...'
        key        = ('localhost...', 35905)
        port       = 35905
        self       = <aiohttp.connector.TCPConnector object at 0x7f542fc5c790>
        traces     = []
```

raises

```
UnicodeError: encoding with 'idna' codec failed (UnicodeError: label empty or too long)
```

webknjaz: @martin-sucha thanks! Let's fix that IDNA thing too, for consistency.

tests/test_connector.py
```diff
@@ -2031,8 +2031,17 @@ async def handler(request):
     await session.close()
 
 
+@pytest.mark.parametrize(
+    "host",
+    (
+        pytest.param("127.0.0.1", id="ip address"),
+        pytest.param("localhost", id="domain name"),
+        pytest.param("localhost.", id="fully-qualified domain name"),
+        pytest.param("localhost...", id="fully-qualified domain name"),
```
webknjaz:
```suggestion
        pytest.param("localhost...", id="fully-qualified domain name with multiple trailing dots"),
```

aiohttp/connector.py
```diff
@@ -1085,6 +1085,10 @@ async def _create_direct_connection(
 
         host = req.url.raw_host
         assert host is not None
+        # Replace multiple trailing dots with a single one.
+        # See https://github.com/aio-libs/aiohttp/pull/7364.
+        while host.endswith(".."):
+            host = host[:-1]
```
webknjaz: Multiple string assignments and recreations seem rather expensive. How about:
```suggestion
        if host.endswith("..") and (trailing_dots_cutoff := -2):
            while test_host[trailing_dots_cutoff - 1] == ".":
                trailing_dots_cutoff -= 1
            host = host[:trailing_dots_cutoff]
```

Dreamsorcerer: Or just `host = host.rstrip(".") + "."`?

Dreamsorcerer:
```
> python3 -m timeit 'host = "foo.bar....."' 'host = host.rstrip(".") + "."'
2000000 loops, best of 5: 144 nsec per loop

> python3 -m timeit 'host = "foo.bar....."' 'while host.endswith(".."):
 host = host[:-1]'

500000 loops, best of 5: 702 nsec per loop

> python3 -m timeit 'host = "foo.bar....."' 'if host.endswith("..") and (trailing_dots_cutoff := 2):
 while host[trailing_dots_cutoff - 1] == ".":
  trailing_dots_cutoff -= 1
 host = host[:trailing_dots_cutoff]'
1000000 loops, best of 5: 217 nsec per loop
```

webknjaz: Wow, that's even better! I also tested your variant with an f-string instead of concatenation, and it's a bit slower. So let's go with your suggestion:
```suggestion
        host = host.rstrip(".") + "."
```

martin-sucha: `host = host.rstrip(".") + "."` adds a dot at the end if it was not present, so we can't use it directly. (Names without the dot at the end are resolved using the domain search path, with dot at the end the resolver does not use search path). With if block it is a bit slower:

```
> python3 -m timeit 'host = "foo.bar....."' 'host = host.rstrip(".") + "."'
5000000 loops, best of 5: 86.7 nsec per loop
> python3 -m timeit 'host = "foo.bar....."' 'if host.endswith("."): host = host.rstrip(".") + "."'
1000000 loops, best of 5: 189 nsec per loop
> python3 -m timeit 'host = "foo.bar....."' 'host = host[:len(host.rstrip("."))+1]'
2000000 loops, best of 5: 134 nsec per loop

> python3 -m timeit 'host = "foo.bar"' 'host = host.rstrip(".") + "."'
5000000 loops, best of 5: 54.4 nsec per loop
> python3 -m timeit 'host = "foo.bar"' 'if host.endswith("."): host = host.rstrip(".") + "."'
5000000 loops, best of 5: 72.7 nsec per loop
> python3 -m timeit 'host = "foo.bar"' 'host = host[:len(host.rstrip("."))+1]'
5000000 loops, best of 5: 74.9 nsec per loop
```

I opted for `if host.endswith("."): host = host.rstrip(".") + "."` anyway because it is more explicit about the missing dot case. Please let me know if you want to change it.

aiohttp/connector.py
```diff
@@ -1085,6 +1085,11 @@ async def _create_direct_connection(
 
         host = req.url.raw_host
         assert host is not None
+        # Replace multiple trailing dots with a single one.
+        # A trailing dot is only present for fully-qualified domain names.
+        # See https://github.com/aio-libs/aiohttp/pull/7364.
+        if host.endswith("."):
```
webknjaz: This makes sense if there's at least two dots:
```suggestion
        if host.endswith(".."):
```

aiohttp/connector.py
webknjaz: Right, I thought I included that check. Thanks for spotting that!

tests/test_connector.py
webknjaz: Looks like this is the last request that still needs to be implemented.

martin-sucha: This one is already implemented by the replacement of multiple trailing dots with a single dot in `_create_direct_connection` that we discussed in https://github.com/aio-libs/aiohttp/pull/7364#discussion_r1299426737

The UnicodeError will not be raised because there won't be empty labels in the DNS name.

Dreamsorcerer:
@martin-sucha There's a black formatting error (I'd update it myself, but you've disabled maintainer access to your branch).

martin-sucha:
> There's a black formatting error (I'd update it myself, but you've disabled maintainer access to your branch).

Oops. Fixed.

tests/test_connector.py
webknjaz: @martin-sucha sounds like we need a test for that, then? Maybe with some actual Unicode chars? Both decoded and encoded with Punycode?

tests/test_client_request.py
```diff
@@ -279,6 +279,37 @@ def test_host_header_ipv6_with_port(make_request: Any) -> None:
...
+        pytest.param(
+            "http://python.org...:99",
+            None,
+            "python.org:99",
+            id="multiple dots with port",
+        ),
+        pytest.param(
+            "http://python.org.:99",
+            {"host": "example.com.:99"},
+            "example.com.:99",
```
webknjaz, line 299: I wonder if this would work well:
```suggestion
            {"host": "tomáš.masaryk.cz.:99"},
            "tomáš.masaryk.cz.:99",
```
Also, let's think of other places where we might need IDN coverage.

tests/test_connector.py
martin-sucha: I don't understand why we'd need another test.

`test_tcp_connector_do_not_raise_connector_ssl_error`/`fully-qualified domain name with multiple trailing dots` covers the case.

We just won't pass a string with multiple trailing dots to `self._resolve_host` so we don't need to make any changes in `_resolve_host`.

tests/test_client_request.py
martin-sucha: @webknjaz I've added a different IDNA test case.

tests/test_connector.py
webknjaz: I was just trying to cover the cases that might happen if someone introduces a but in the future. You implied that there may be problems with IDN in some places so I thought it's a good idea to have regression tests covering such corner cases and expecting the same across the board. But if you think it's too much, I won't insist.

tests/test_client_request.py
webknjaz: Thank you 🙏

Marking this thread as resolved.

webknjaz (approved):
Love it! Thanks for working on this! I think it's ready now.

webknjaz:
@martin-sucha looks like there are some legitimate failures in the CI, could you take a look?

martin-sucha:
@webknjaz Seems that on macOS the resolver won't resolve `something.localhost`. I've mocked _resolve_host to always return the loopback address. I think that should be okay since the test is checking TLS.

webknjaz:
@martin-sucha thank you! I've enabled auto-merge and there's a backporting bot label. If it fails to produce a backport PR, please, follow the instructions to do this manually.

tests/test_connector.py
webknjaz: Marking this thread as resolved.

webknjaz:
RTD build was stuck for some reason. I restarted it so GH could perform merging.

martin-sucha:
@webknjaz I opened https://github.com/aio-libs/aiohttp/pull/7601 for the backport.

PR #7366, opened 2023-07-15
Improve error messages from C parser

Dreamsorcerer:
This adds information to the error messages to show what causes the error.

Exception output looks like:
```
aiohttp.http_exceptions.BadHttpMessage: 400, message:
  Invalid header value char:

    b'Set-Cookie: abc\x01def\r'
                     ^
```


vendor/README
```diff
@@ -0,0 +1,19 @@
+LLHTTP
```
webknjaz: Looks like this lacks `.rst` in the name.

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "vendor/llhttp"]
     path = vendor/llhttp
     url = https://github.com/nodejs/llhttp.git
-    branch = v8.1.1
+    branch = v8.x
```
webknjaz: We shouldn't have this unpinned, it'd make it impossible to track what version aiohttp is compiled against..

vendor/llhttp
```diff
@@ -1 +1 @@
-Subproject commit 69d6db2008508489d19267a0dcab30602b16fc5b
+Subproject commit 7e18596bae8f63692ded9d3250d5d984fe90dcfb
```
webknjaz: I would probably make the bump separate...

.gitmodules
Dreamsorcerer: Surely the pin is the commit of the submodule. This is just the branch to update from, and there is no v8.1.1 branch.

Dreamsorcerer: Hmm, actually, looks like we can do `tags/v8.1.1`.

Dreamsorcerer: Nevermind, that doesn't work...

PR #7346, opened 2023-07-07
 Bump vendored llhttp to v8.1.1

webknjaz:
It hasn't been updated in two years... It's time to upgrade to the latest version!

PR #7367, opened 2023-07-16
Fix bump llhttp to v8.1.1

Dreamsorcerer:


vendor/README.rst
```diff
@@ -0,0 +1,20 @@
...
+    cd vendor/llhttp/
+    npm install
+    make
+
+Then build our parser::
+
+    cd -
+    make cythonize
+
+Then you can build or install it with ``python -m build`` or ``pip install .``
```
webknjaz: I wonder if `-e` is needed in the example..

vendor/README.rst
```diff
@@ -0,0 +1,20 @@
+LLHTTP
```
webknjaz, line 1: Can we try including this in the docs? Via `.. include::`.

tests/test_http_parser.py
```diff
@@ -694,6 +694,7 @@ def test_http_response_parser_bad(response: Any) -> None:
         response.feed_data(b"HTT/1\r\n\r\n")
 
 
+@pytest.mark.skipif(not NO_EXTENSIONS, reason="Behaviour has changed in C parser")
```
webknjaz, line 697: Better xfail instead..

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "vendor/llhttp"]
     path = vendor/llhttp
     url = https://github.com/nodejs/llhttp.git
-    branch = v8.1.1
+    branch = v8.x
```
webknjaz, line 4: This shouldn't be a moving target but a more specific thing like a tag or even a commit sha...

webknjaz:
@Dreamsorcerer I think there was a change note file that needs updating. One of the recent ones.

tests/test_http_parser.py
webknjaz: Or.. Can we include the minimum change needed to process the behavior difference?

.gitmodules
Dreamsorcerer: Again, from previous PR, this is just the branch that it updates from, there is no v8.1.1 branch. The pin is the commit of the actual submodule at vendor/llhttp.

Dreamsorcerer:
> @Dreamsorcerer I think there was a change note file that needs updating. One of the recent ones.

I don't think anything needs updating from your previous note: https://github.com/aio-libs/aiohttp/blob/master/CHANGES/7346.feature

The change you applied just didn't work, this PR fixes it (as evidenced by the test not failing in your PR). The version is still 8.1.1.

tests/test_http_parser.py
Dreamsorcerer: Could do, but I was planning to come back to this and figure out if we should make the behaviour consistent between parsers later. If there's a test here that explicitly allows it, then it seems like something that was wanted, so ideally we can tweak llhttp later to allow the behaviour again. So, I'd suggest leaving it like this for the moment, and I'll come back to it later.

vendor/README.rst
Dreamsorcerer: Could do, anywhere you have in mind to put it?

Dreamsorcerer: OK, took a go at adding it after the dependencies list.

.gitmodules
webknjaz: Oh, I didn't realize that, somehow. Just checked out https://stackoverflow.com/a/18797720/595220, and I now understand better how this works and why https://github.com/aio-libs/aiohttp/pull/7346 didn't work…

tests/test_http_parser.py
webknjaz: Okay, agreed. And in the terms of synchronizing the behaviors of C- and Python-based HTTP parsers, we should really invest into wiring Hypothesis.

vendor/README.rst
webknjaz: @Dreamsorcerer I was kinda expecting this @ https://aiohttp--7367.org.readthedocs.build/en/7367/contributing.html. Not sure which one is better. llhttp could be mentioned in the deps, but the end-users wouldn't compile from Git anyway..

docs/index.rst
```diff
@@ -140,9 +140,6 @@ Please feel free to file an issue on the `bug tracker
 <https://github.com/aio-libs/aiohttp/issues>`_ if you have found a bug
 or have some suggestion in order to improve the library.
 
-The library uses `Azure Pipelines <https://dev.azure.com/aio-libs/aiohttp/_build>`_ for
```
webknjaz, line 143: Can this be changed separately and mention GHA?

webknjaz:
So I suppose this is mergeable, but it'd be nice to think about those nitpicks...

docs/index.rst
Dreamsorcerer: I don't think mentioning what CI system is used is something relevant to the front page. People can easily find that by navigating to the source code on Github.

Dreamsorcerer: We also have the "CI: passing" badge in the navigation, which links directly to the CI runs. So, I'm just going to drop this.

PR #7383, opened 2023-07-19
Replace sync `open()` calls within async functions with `aiofiles.open()` in code examples

cclauss:
% [`ruff --select=ASYNC --statistics .`](https://beta.ruff.rs/docs/rules/open-sleep-or-subprocess-in-async-function)

2	ASYNC101	[ ] Async functions should not call `open`, `time.sleep`, or `subprocess` methods

% [`ruff rule ASYNC101`](https://beta.ruff.rs/docs/rules/open-sleep-or-subprocess-in-async-function/)
# open-sleep-or-subprocess-in-async-function (ASYNC101)

Derived from the **flake8-async** linter.

## What it does
Checks that async functions do not contain calls to `open`, `time.sleep`, or `subprocess` methods.

## Why is this bad?
Blocking an async function via a blocking call will block the entire event loop, preventing it from executing other tasks while waiting for the call to complete, negating the benefits of asynchronous programming.

Instead of making a blocking call, use an equivalent asynchronous library or function.

## Example
```python
async def foo():
    time.sleep(1000)
```

Use instead:
```python
async def foo():
    await asyncio.sleep(1000)
```

## What do these changes do?
Screenshot 2023-07-19 at 17 44 51

## Checklist

    * `.misc`: A ticket has been closed, but it is not interesting to users.

### DIFF
--- a/tools/bench-asyncio-write.py
+++ b/tools/bench-asyncio-write.py
@@ -123,10 +123,11 @@ async def main(loop):
         base = await bench(t, writes[0], c)
         for w in writes[1:]:
             await bench("", w, c, base)
-    with open("bench.md", "w") as f:
-        for line in res:
-            f.write("| {} |\n".format(" | ".join(line)))
+    return res
 
 
 loop = asyncio.get_event_loop()
-loop.run_until_complete(main(loop))
+results = loop.run_until_complete(main(loop))
+with open("bench.md", "w") as f:
+    for line in results:
+        f.write("| {} |\n".format(" | ".join(line)))


webknjaz (changes requested):
This looks quite invasive and combines several separate pointless changes. For example, adding a dependency that is not actually used in the library.
Also, ruff can only be added to pre-commit, not a random workflow that only multiplies the maintenance burden.
I'm 👎 on the change as it is. Some small bits could be added if discussed and implemented correctly, but not as a bunch of commits that have no visible purpose.

requirements/test.txt
```diff
@@ -1,4 +1,5 @@
 -r base.txt
+aiofiles
```
webknjaz: Why do you think this is needed in the tests?

cclauss: Reverted.  This will cause flake8 to raise an I900 error.
* https://github.com/Arkq/flake8-requirements#readme
* https://results.pre-commit.ci/run/github/13258039/1689858060.9-g2zv78Q8ejon_od5hMtw

examples/web_ws.py
```diff
@@ -18,8 +20,8 @@ async def wshandler(request: web.Request) -> Union[web.WebSocketResponse, web.Re
     resp = web.WebSocketResponse()
     available = resp.can_prepare(request)
     if not available:
-        with open(WS_FILE, "rb") as fp:
-            return web.Response(body=fp.read(), content_type="text/html")
+        async with aiofiles.open(WS_FILE, "rb") as fp:
+            return web.Response(body=await fp.read(), content_type="text/html")
```
Dreamsorcerer: I think this is reasonable (although I'd note that we seem to use `to_thread()` or similar in our internal code).

tools/bench-asyncio-write.py
```diff
@@ -123,9 +125,9 @@ async def bench(job_title, w, body, base=None):
         base = await bench(t, writes[0], c)
         for w in writes[1:]:
             await bench("", w, c, base)
-    with open("bench.md", "w") as f:
+    async with aiofiles.open("bench.md", "w") as f:
         for line in res:
-            f.write("| {} |\n".format(" | ".join(line)))
+            await f.write("| {} |\n".format(" | ".join(line)))
```
Dreamsorcerer: But, I don't think we should change this script and add an extra dependency. The blocking call clearly doesn't affect performance as it is the only thing running at this point.

If you wanted to be pedantic, I'd suggest moving it out of the async function instead.

e.g.
```
results = asyncio.run(run_benchmark())
with open(...):
   for l in results:
       ....
```

Dreamsorcerer:
Not sure why the change to examples has disappeared?

cclauss:
I removed it because the current CI requires adding both `aiofiles` and `aiofiles-types` to various requirements files which became part of comments in https://github.com/aio-libs/aiohttp/pull/7383#pullrequestreview-1538108058  After this PR lands, I can re-create the other so the steps are clear.

PR #7439, opened 2023-07-27
Remove sphinxcontrib-asyncio

Dreamsorcerer:
I don't this is needed anymore, and may be causing issues with newer sphinx releases.

### DIFF
--- a/docs/abc.rst
+++ b/docs/abc.rst
@@ -39,7 +39,8 @@ Not Allowed*. :meth:`aiohttp.abc.AbstractMatchInfo.handler` raises
    *router* parameter and returns as
    :attr:`aiohttp.web.Application.router`.
 
-   .. coroutinemethod:: resolve(request)
+   .. method:: resolve(request)
+      :async:
 
       Performs URL resolving. It's an abstract method, should be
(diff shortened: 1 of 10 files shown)


webknjaz:
...maybe? I don't know what it does anyway.

webknjaz (approved):

Dreamsorcerer:
Judging by the errors, it adds the `coroutine` directives. Updating them to `:async:`, which is what Sphinx itself uses these days.

Dreamsorcerer:
@webknjaz If you've got a moment spare, could you take a look at the toctree and help decide what it should look like. In client.rst the default view now looks like:
image
I tried setting maxdepth to 3, which still looks a bit odd with duplicate headers:
https://aiohttp--7439.org.readthedocs.build/en/7439/client.html#aiohttp-client

~~We could set it to 2, but that will then remove some nested headers from the above sections, which is probably not desirable.~~
Setting it to 2 would eliminate some of the actual subsections shown currently.

Dreamsorcerer:
Actually, I think that was caused by the previous Sphinx version bump, not related to this change.

PR #7441, opened 2023-07-27
Tidy up toctree

Dreamsorcerer:
Attempt to tidy up the toctree

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -10,6 +10,7 @@ The page contains all information about aiohttp Client API:
 
 .. toctree::
    :name: client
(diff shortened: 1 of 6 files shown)


Dreamsorcerer:
That's an improvement. But, I don't know how to make a class be a top-level item, instead of a sublevel. For example, 'Response object' may be a redundant heading, but if I just remove it, then `ClientResponse` will appear under the 'Connectors' header when it needs to be 1 level higher.
https://aiohttp--7441.org.readthedocs.build/en/7441/client.html#aiohttp-client

PR #7442, opened 2023-07-27
Use setup-python for cache

Dreamsorcerer:


webknjaz:
I remember avoiding this because of some quirks. But can't recall why exactly.. I'll try to look into this.

Dreamsorcerer:
OK, on other repos the cache wasn't working correctly with the old cache action, and setup-python worked first time (I think it's aiohttp-demos which takes about 10 mins extra without caching).

webknjaz:
@Dreamsorcerer

> OK, on other repos the cache wasn't working correctly with the old cache action

One common mistake is making checkout prior to invoking cache. Check for that first.

> I remember avoiding this because of some quirks. But can't recall why exactly.. I'll try to look into this.

I tried to recall what were the corner cases. And I found two:
1. Skipping caching for release CI/CD invocations — this allows to catch weird cases that don't come up in PRs because they use older versions of packages, for example: https://github.com/cherrypy/cheroot/blob/0fd16f0/.github/workflows/ci-cd.yml#L246-L247
2. Skipping caching for unstable/nightly/dev Python jobs, especially in the period when CPython transitions between alpha/beta/rc. This is because older package versions may be picked up by pip earlier and the, they'd break when GitHub updates their CPython alpha build but the cached packages aren't updated. Example: https://github.com/cherrypy/cheroot/blob/0fd16f0/.github/workflows/ci-cd.yml#L717-L775.

webknjaz:
FTR, the caching works as it is configured today: https://github.com/aio-libs/aiohttp/actions/runs/7688751163/job/20950541567#step:5:25.

It doesn't implement those stable ABI-conditional hacks, though. Would be nice to have them here too.

.github/workflows/ci-cd.yml
```diff
@@ -173,18 +168,8 @@ jobs:
...
-      run: |
-        echo "dir=$(pip cache dir)" >> "${GITHUB_OUTPUT}"
-      shell: bash
-    - name: Cache PyPI
-      uses: actions/cache@v4.1.2
-      with:
-        key: pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-${{ hashFiles('requirements/*.txt') }}
-        path: ${{ steps.pip-cache.outputs.dir }}
-        restore-keys: |
-            pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
```
bdraco, line 187: I'm worried it won't take `matrix.no-extensions` into consideration

bdraco: When it runs I think we can look at the cache key it generates and see if it has no-ext

bdraco:
`Test (3.13, Y, ubuntu, false)`
`setup-python-Linux-x64-22.04-Ubuntu-python-3.13.0-pip-16df3627115a202aec77821077e7df01ad7900a81e1096da766b5297341a202b`

`Test (3.13, ubuntu, false)`
setup-python-Linux-x64-22.04-Ubuntu-python-3.13.0-pip-16df3627115a202aec77821077e7df01ad7900a81e1096da766b5297341a202b

webknjaz:
> no-extensions isn't considered in the cache key

Is that important, though? Wouldn't the deps be mostly the same in both cases?

webknjaz:
@Dreamsorcerer I finally have the preferred solution to show: https://github.com/aio-libs/frozenlist/pull/622. Let's replicate it instead.

Yedfree (approved):

PR #7484, opened 2023-08-06
Increase stacklevel for UserWarning about AppKey usage

cdce8p:
## What do these changes do?

The `UserWarning` to use `web.AppKey` should be emitted on the call site.

## Checklist

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -215,6 +215,7 @@ Lukasz Marcin Dobrzanski
 Makc Belousow
 Manuel Miranda
 Marat Sharafutdinov
+Marc Mueller
 Marco Paolini
 Mariano Anaya
 Mariusz Masztalerczuk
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
Thanks for trying this out. Do you think it's possible to tweak the test to catch this? https://github.com/aio-libs/aiohttp/blob/master/tests/test_web_app.py#L147

cdce8p:
> Thanks for trying this out. Do you think it's possible to tweak the test to catch this? https://github.com/aio-libs/aiohttp/blob/master/tests/test_web_app.py#L147

Pytest doesn't really support it unfortunately https://github.com/pytest-dev/pytest/issues/4343. I've tried a small workaround to at least check the file name where the warning is emitted but this is not something I would recommend for everything. Let me know what you think.

Dreamsorcerer:
That's probably fine.

PR #7485, opened 2023-08-06
Update to LLHTTP 9

Dreamsorcerer:
We already have a new version to upgrade to.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "vendor/llhttp"]
     path = vendor/llhttp
     url = https://github.com/nodejs/llhttp.git
-    branch = v8.x
+    branch = main
--- a/aiohttp/_cparser.pxd
+++ b/aiohttp/_cparser.pxd
@@ -1,13 +1,4 @@
-from libc.stdint cimport (
-    int8_t,
-    int16_t,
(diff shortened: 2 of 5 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "vendor/llhttp"]
     path = vendor/llhttp
     url = https://github.com/nodejs/llhttp.git
-    branch = v8.x
+    branch = main
```
Dreamsorcerer, line 4: There's no v9 branch. So, it looks like we'll be best just leaving it on main, and being sure to checkout a particular tag when updating.

aiohttp/_cparser.pxd
Dreamsorcerer, line 1: Cleaning up unused parts here.

tests/test_http_parser.py
```diff
@@ -872,19 +872,6 @@ def test_partial_url(parser: Any) -> None:
     assert payload.is_eof()
 
 
-def test_url_parse_non_strict_mode(parser: Any) -> None:
```
Dreamsorcerer, line 875: This doesn't test anything more than the below test. There is also no strict/non-strict mode as the name suggests...

PR #7490, opened 2023-08-06
Use lenient headers for response parser

Dreamsorcerer:


tests/test_http_parser.py
```diff
@@ -730,6 +730,21 @@ def test_http_response_parser_no_reason(response: Any) -> None:
...
+        b"HTTP/1.1 200 test\r\nFoo: abc\x01def\r\n\r\n"
+    )
+    msg = messages[0][0]
+
+    assert msg.headers["Foo"] == "abc\x01def"
+
+
+@pytest.mark.dev_mode
+def test_http_response_parser_strict_headers(response: Any) -> None:
+    with pytest.raises(http_exceptions.BadHttpMessage):
```
webknjaz, line 746: It's highly recommended to always use the `match=` argument with `pytest.raises()`. Use a full regex, starting with `^` and ending with `$`.

tests/test_http_parser.py
```diff
@@ -730,6 +730,21 @@ def test_http_response_parser_no_reason(response: Any) -> None:
     assert msg.reason == ""
 
 
+def test_http_response_parser_lenient_headers(response: Any) -> None:
```
webknjaz, line 733: Could you replace `Any` with an actual type?

webknjaz (commented):
I like the idea! Could you address the test concerns?

tests/test_http_parser.py
Dreamsorcerer: Yep, but for the parser this seems too complex, as the messages might not be the same between py/c parsers. Plus, the contents of the message from llhttp are not our concern, we might end up needing to update the regex with every version of llhttp (I'm sure they have their own tests for error messages).

tests/test_http_parser.py
Dreamsorcerer: I've got an in-progress PR to add more typing to all the tests, so I will update all of them in the file in that PR.

Dreamsorcerer: Also note the type: ignore on line 1, which means this file is currently not checked at all (I'm going through every file that has been ignored and updating them one at a time).

webknjaz: Isn't it easier to start adding typing to new things right away?

tests/test_http_parser.py
webknjaz: But if this starts erroring out for a different reason, our test would still fail. Shouldn't we normalize those, then? Besides, since we always update llhttp to a pinned version, it's okay to update the error message.

tests/test_http_parser.py
Dreamsorcerer: Figuring out the pytest fixture annotations always takes me a bit of time, and I don't want to duplicate that effort. I also can't test it without introducing 100+ mypy errors. So, in this case, it's easier to do it all at once with a find/replace.

tests/test_http_parser.py
Dreamsorcerer: Well, BadHttpMessage is a parse error. I don't really care what it thinks the error is beyond that, I just care that it thinks this is bad in dev mode, but not normally.

Normalising the exception messages and keeping them in sync seems like a big job to me, and not one I currently want to take on.

> Besides, since we always update llhttp to a pinned version, it's okay to update the error message.

My concern is the maintenance burden of updating llhttp. I'm not going to be happy if I have to manually update 106 tests (so far), just to bump the version...

webknjaz: okay, let's keep it as is. resolving

tests/test_http_parser.py
webknjaz: marking as resolved, then.

.github/workflows/ci-cd.yml
```diff
@@ -238,6 +238,13 @@ jobs:
       run: >-  # `exit 1` makes sure that the job remains red with flaky runs
         pytest --no-cov -vvvvv --lf && exit 1
       shell: bash
+    - name: Run dev_mode tests
```
webknjaz, line 241: Maybe put this before the normal test run?

webknjaz: It's not that important, I suppose...

PR #7503, opened 2023-08-10
Skip uvloop install on Python 3.12

cdce8p:
## What do these changes do?

Skip `uvloop` install on Python 3.12.
MagicStack/uvloop#547

Let's see if that unblocks the [failing 3.12 tests](https://github.com/aio-libs/aiohttp/actions/runs/5779615304/job/15662155865#step:7:210).

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -2,4 +2,4 @@
 -r runtime-deps.in
 
 gunicorn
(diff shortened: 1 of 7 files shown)


Dreamsorcerer:
Thanks, looks like it could be some time before there's a new uvloop, so seems good to me.

PR #7504, opened 2023-08-10
Replace typing-extensions imports

cdce8p:
## What do these changes do?
Replace import from `typing_extensions` with `typing` now that Python 3.8 is the oldest supported version.

_No_

Technically, this removes the need for the `typing_extension` dependency. However, it isn't even listed as a direct dependency at the moment.
https://github.com/aio-libs/aiohttp/blob/c9176acb672b90717914d232ecca24e22ed42325/setup.cfg#L49-L56 

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -18,6 +18,7 @@ from typing import (
     Callable,
     Collection,
     Coroutine,
+    Final,
     FrozenSet,
     Generator,
     Generic,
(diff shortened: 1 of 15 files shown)


Dreamsorcerer:
Ah, thanks. I had updated the 3.x branch to use `sys.version_info` checks properly, so they all got cleaned up there when we dropped Python 3.7. But, I forgot about these on the master branch.

PR #7518, opened 2023-08-12
Fix GunicornWebWorker max_requests_jitter not work

phyng:
## What do these changes do?

`GunicornWebWorker` use `self.cfg.max_requests` which is not add jitter, from https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L56-L60, the correct way is to use `sef.max_requests`

After the PR is merged, the max-requests-jitter parameter of Gunicorn can take effect.

## Checklist

### DIFF
--- a/CHANGES/7518.bugfix
+++ b/CHANGES/7518.bugfix
@@ -1,0 +1,1 @@
+Fix GunicornWebWorker max_requests_jitter not work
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
Appears to break some tests, so will need updating.

PR #7541, opened 2023-08-19
Support passing a custom server name parameter on HTTPS connection

### LINKED ISSUE #7114: Specify SNI server_hostname when performing request with asyncio/aiohttp
I also posted this [HERE](https://stackoverflow.com/questions/74647236/specify-sni-server-hostname-when-performing-request-with-asyncio-aiohttp)

The context :
```
            load-balancer.example.com:443 (TCP passthrough)
                               /\
                              /  \
                             /    \
                            /      \
          s1.example.com:443        s2.example.com:443
              (SSL/SNI)                  (SSL/SNI)              
```

I am able to launch a huge batch of async `ClientSession.get` requests on both upstreams in parallel but for each request I need to somehow tell `asyncio` or `aiohttp` to use `load-balancer.example.com` as `server_hostname`, otherwise the SSL handshake fails.

Is there an easy way to setup the `ClientSession` to use a specific `server_hostname` when setting up the SSL socket ? 
Does someone have already done something like that ?

The goal is to stress-test the upstreams `s1` and `s2` directly using `aiohttp` with certificate-validation enable. Since the load-balancer does not belong to me I don't want to do the stress-test over it.
- the code is not supposed to run on other platforms than GNU Linux with at least Python-v3.7 (but I can use any recent version if needed)
- all servers serve a valid certificate for `load-balancer.example.com`
- `openssl` validates the certificate from the upstreams when using `openssl s_connect s1.example.com:443 -servername load-balancer.example.com`
- `cURL` needs `curl 'https://load-balancer.example.com/' --resolve s1.example.com:443:load-balancer.example.com` and also validates successfully

here is the most simple snippet with just a single request :

```python
import aiohttp
import asyncio


async def main_async(host, port, uri, params=[], headers={}, sni_hostname=None):
    if sni_hostname is not None:
        print('Setting SNI server_name field ')
        #
        # THIS IS WHERE I DON'T KNOW HOW TO TELL ASYNCIO
        # TO SET THE SERVER_NAME FIELD TO sni_hostname
        # IN THE SSL SOCKET BEFORE PERFORMING THE SSL HANDSHAKE
        #
    try:
        async with aiohttp.ClientSession(raise_for_status=True) as session:
            async with session.get(f'https://{host}:{port}/{uri}', params=params, headers=headers) as r:
                body = await r.read()
                print(body)
    except Exception as e:
        print(f'Exception while requesting ({e}) ')


if __name__ == "__main__":
    asyncio.run(main_async(host='s1.example.com', port=443,
                           uri='/api/some/endpoint',
                           params={'apikey': '0123456789'},
                           headers={'Host': 'load-balancer.example.com'},
                           sni_hostname='load-balancer.example.com'))

```

When running it with real hosts, it throws
```python
Cannot connect to host s1.example.com:443 ssl:True 
[SSLCertVerificationError: (1, '[SSL: CERTIFICATE_VERIFY_FAILED] '
certificate verify failed: certificate has expired (_ssl.c:1131)')]) 
```

Note that the error `certificate has expired` indicates that the certificate proposed to the client is the default certificate since the SNI hostname is `s1.example.com` which is unknow by the webserver running there.

When running it against the load-balancer it works just fine, the SSL handshake happens with the upstreams which serve the certificate and everything is valid.

Also note that
 - [`sni_callback`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.sni_callback) does not help since it is called after the handshake has started and the certificate was received (and at this point `server_hostname` is a read-only property anyway)
 - it does not seem to be possible to set `server_hostname` when creating an `SSLContext` allthough [SSLContext.wrap_socket](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket) does support `server_hostname` but I was not able to make that work

I hope someone knows how to fill the comment block in that snippet ;-]

### Related component

Client

multani:
## What do these changes do?

This adds the missing support to set the `server_hostname` setting when creating TCP connection, when the underlying connection is authenticated using TLS.

See the documentation for the 2 stdlib functions:

* https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection
* https://docs.python.org/3/library/asyncio-eventloop.html#opening-network-connections

This would be needed to support features in clients using aiohttp, such as https://github.com/tomplus/kubernetes_asyncio/issues/267

The default behavior should not change, but this would allow on a per-connection basis to specify a custom server name to check the certificate name against.

## Related issue number

Closes: https://github.com/aio-libs/aiohttp/issues/7114

(for reference, similar implementation in urllib3: https://github.com/urllib3/urllib3/pull/1397)

## Checklist


aiohttp/connector.py
```diff
@@ -1085,14 +1085,18 @@ async def _start_tls_connection(
         # maintainability-wise but this is to be solved separately.
         sslcontext = cast(ssl.SSLContext, self._get_ssl_context(req))
 
+        server_hostname = req.host
+        if req.server_hostname is not None:
+            server_hostname = req.server_hostname
```
Dreamsorcerer: Should just be able to do:
```suggestion
        server_hostname = req.server_hostname or req.host
```
Maybe worth moving it back inline as well.

aiohttp/connector.py
```diff
@@ -1174,6 +1178,10 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
             host = hinfo["host"]
             port = hinfo["port"]
 
+            server_hostname = hinfo["hostname"] if sslcontext else None
+            if req.server_hostname is not None:
+                server_hostname = req.server_hostname
```
Dreamsorcerer: Should this be set regardless of `sslcontext`? Seems like a change in behaviour there.

tests/test_connector.py
```diff
@@ -557,6 +557,40 @@ async def certificate_error(*args, **kwargs):
     assert isinstance(ctx.value, aiohttp.ClientSSLError)
 
 
+async def test_tcp_connector_server_hostname_default(loop) -> None:
+    conn = aiohttp.TCPConnector(loop=loop)
+
+    conn._loop.create_connection = mock.AsyncMock()
+    conn._loop.create_connection.return_value = mock.Mock(), mock.Mock()
```
Dreamsorcerer: Would be great if you can get this to work with something like:

```suggestion
    with mock.patch.object(conn._loop, "create_connection", autospec=True, spec_set=True):
```

tests/test_connector.py
```diff
@@ -557,6 +557,40 @@ async def certificate_error(*args, **kwargs):
...
+
+    req = ClientRequest("GET", URL("https://127.0.0.1:443"), loop=loop)
+
+    established_connection = await conn.connect(req, [], ClientTimeout())
```
Dreamsorcerer: Can probably use `async with` here, rather than needing to close explicitly.

Dreamsorcerer:
Also, if the tests still don't work after those updates on 3.8 due to missing AsyncMock, feel free to add a `pytest.skipif` for 3.8.

aiohttp/connector.py
Dreamsorcerer: > Should only be passed if ssl is not None.

So, looks like not.

Dreamsorcerer:
```suggestion
            server_hostname = (req.server_hostname or hinfo["hostname"]) if sslcontext else None
```

docs/client_reference.rst
```diff
@@ -487,6 +488,13 @@ The client session supports the context manager protocol for self closing.
 
             Use ``ssl=aiohttp.Fingerprint(digest)``
 
+      :param str server_hostname: Sets or overrides the hostname that the
+         target server’s certificate will be matched against.
+         See `upstream documentation <https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection>`_
```
Dreamsorcerer: You can use intersphinx:
```suggestion
         See :method:`asyncio.loop.create_connection`
```

aiohttp/connector.py
multani: Implemented the one-liner suggestion below :+1:

tests/test_connector.py
multani: Sadly, `.connect` returns a [`Connection`](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/connector.py#L72) which doesn't implement the context management protocol :/

I can use [`contextlib.closing`](https://docs.python.org/3/library/contextlib.html#contextlib.closing) though :+1:

multani:
@Dreamsorcerer thanks for all the good comments! :bow: 

I made all the changes requested, it's better now! :+1: I'll keep an eye on the CI.

Dreamsorcerer:
Oh, I just realised you're pointing to the wrong branch. You need to target a PR to master.

multani:
> Oh, I just realised you're pointing to the wrong branch. You need to target a PR to master.

Arghhh, sorry about that, I did my tests through [another library](https://github.com/tomplus/kubernetes_asyncio/pull/270) that was using 3.8 and I used that branch for my personal PR before publishing it here :facepalm: 

Sorry about the inconvenience, I just rebased/squashed all my changes on top of the master branch and changed the PR to point on that branch instead :bow:

multani:
@Dreamsorcerer thanks a lot for the merge! :tada: 

I can take care of the failing backport and create the related pull request if you want :+1:

Dreamsorcerer:
Thanks, that would be great. Just follow the instructions and create a PR to 3.9 branch.

multani:
The backport is in https://github.com/aio-libs/aiohttp/pull/7543!

PR #7552, opened 2023-08-25
Skip default headers

SkyTheCodeMaster:
## What do these changes do?
These changes allows users to specify whether or not the `ClientSession.headers` are included in this request.
I added this because I was writing a library where I add an `Authorization` header in `ClientSession.headers`, however I needed 
to do another request where I didn't want to pass that `Authorization` header. I've added a simple boolean to turn it on and off.

No significant changes, the default behaviour is still the same.

## Checklist

### DIFF
--- a/CHANGES/7552.feature
+++ b/CHANGES/7552.feature
@@ -1,0 +1,1 @@
+Add `skip_default_headers` to client session requests
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -347,7 +347,8 @@ The client session supports the context manager protocol for self closing.
                          timeout=sentinel, ssl=None, \
                          verify_ssl=None, fingerprint=None, \
                          ssl_context=None, proxy_headers=None, \
-                         server_hostname=None, auto_decompress=None)
+                         server_hostname=None, auto_decompress=None, \
+                         skip_default_headers = False)
       :async:
       :noindexentry:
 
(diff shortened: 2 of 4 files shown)


docs/client_reference.rst
```diff
@@ -347,7 +347,8 @@ The client session supports the context manager protocol for self closing.
                          timeout=sentinel, ssl=None, \
                          verify_ssl=None, fingerprint=None, \
                          ssl_context=None, proxy_headers=None, \
-                         server_hostname=None, auto_decompress=None)
+                         server_hostname=None, auto_decompress=None, \
+                         skip_default_headers = False)
```
webknjaz, line 351:
```suggestion
                         skip_default_headers=False)
```

CHANGES/7552.feature
```diff
@@ -0,0 +1 @@
+Add `skip_default_headers` to client session requests
```
webknjaz, line 1: This must use RST, and follow the writing style as explained in https://docs.aiohttp.org/en/stable/contributing.html#adding-change-notes-with-your-prs.

webknjaz (commented):
Honestly, I'm not sure if this use case is worth complicating the API. Why not just use an additional client session instance with different default headers?

Dreamsorcerer:
> Honestly, I'm not sure if this use case is worth complicating the API. Why not just use an additional client session instance with different default headers?

I think that's 3 of us in agreement, so I'll just close this.

PR #7555, opened 2023-08-25
Make AppKey compatible with string keys

cdce8p:
## What do these changes do?
Allow the simultaneous use of `AppKey` and string keys. During my early testing for [Home Assistant](https://github.com/home-assistant/core) I found it quite challenging to start using `AppKey`. At least in this particular case, a variable was stored once but read all throughout the code base. Just missing one update, would cause a test failure. By making both compatible, users will be able to update them gradually.

_The only downside being that the names of `AppKey` will no longer be able to contain the module name and loose some of their uniqueness. However, I don't believe that is an issue with the current string keys so it should be fine. It might even help avoid some additional confusions if AppKey instances have the same name + same variable name but are in different modules -> so not compatible._

Overall, I found that without this change, I wasn't able to introduce `AppKey` during testing in a larger code base.

/CC @Dreamsorcerer 

The feature hasn't been shipped yet.

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -8,7 +8,6 @@ import dataclasses
 import datetime
 import enum
 import functools
-import inspect
 import netrc
 import os
 import platform
@@ -817,19 +816,15 @@ class AppKey(Generic[_T]):
     __orig_class__: Type[object]
 
     def __init__(self, name: str, t: Optional[Type[_T]] = None):
-        # Prefix with module name to help deduplicate key names.
-        frame = inspect.currentframe()
-        while frame:
-            if frame.f_code.co_name == "<module>":
-                module: str = frame.f_globals["__name__"]
-                break
-            frame = frame.f_back
-        else:
-            raise RuntimeError("Failed to get module name.")
-
-        # https://github.com/python/mypy/issues/14209
-        self._name = module + "." + name  # type: ignore[possibly-undefined]
+        self._name = name
         self._t = t
+
+    # Add __hash__ and __eq__ to make AppKey compatible with string keys
+    def __hash__(self) -> int:
+        return hash(self._name)
+
+    def __eq__(self, other: object) -> bool:
+        return hash(self) == hash(other)
 
     def __lt__(self, other: object) -> bool:
         if isinstance(other, AppKey):
--- a/tests/test_web_app.py
+++ b/tests/test_web_app.py
@@ -1,4 +1,5 @@
 import asyncio
+import warnings
 from typing import Any, AsyncIterator, Callable, Iterator, NoReturn
 from unittest import mock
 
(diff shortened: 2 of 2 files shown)


Dreamsorcerer:
Thanks for testing this out.

> During my early testing for [Home Assistant](https://github.com/home-assistant/core) I found it quite challenging to start using `AppKey`. At least in this particular case, a variable was stored once but read all throughout the code base. Just missing one update, would cause a test failure.

Could you show an example of what the problem is? I'd have thought that it would just be a find/replace job...

> _The only downside being that the names of `AppKey` will no longer be able to contain the module name and loose some of their uniqueness. However, I don't believe that is an issue with the current string keys so it should be fine.

The problem this solves is that libraries with string keys have needed to use awkward naming conventions to reduce the likelihood (but not guarantee) that the keys won't clash with a user's own keys or those of another library. With AppKey, this is no longer a problem, but displaying the module name is needed to tell which key we're talking about. e.g. Your application might have a "db" key, aiohttp-admin might have a "db" key and aiohttp-session might have a "db" key. That's 3 different keys that could share the same name (currently, they use awkward names like "__aiohttp_session_db", but a library from another developer not aware of the issue could easily use a simpler string without realising that it's likely to clash with a portion of their user's own keys).

> It might even help avoid some additional confusions if AppKey instances have the same name + same variable name but are in different modules -> so not compatible._

As above, that's rather the point, they are different keys. It would be more confusing to find your key now refers to aiohttp-admin's database connection and not the one you created in your application.

cdce8p:
> > During my early testing for [Home Assistant](https://github.com/home-assistant/core) I found it quite challenging to start using `AppKey`. At least in this particular case, a variable was stored once but read all throughout the code base. Just missing one update, would cause a test failure.
> 
> Could you show an example of what the problem is? I'd have thought that it would just be a find/replace job...

Find and replace technically, however we don't control all parts. There are a lot of external component which we can't update directly / need to be updated by their developers **after** a release. Just starting to use `AppKey` would thus be a breaking change.
```diff
# -- we control this --
class HomeAssistant: ...
KEY_HASS = AppKey[HomeAssistant]("hass")
hass = HomeAssistant()

 app = web.Application()
-app["hass"] = hass
+app[KEY_HASS] = hass
```

```py
# -- external components --
# this would break without an update
hass = app["hass“]
```

That's where the idea to make `str` compatible with `AppKey` came from.

> The problem this solves is that libraries with string keys have needed to use awkward naming conventions to reduce the likelihood (but not guarantee) that the keys won't clash with a user's own keys or those of another library. [...]

> As above, that's rather the point, they are different keys. It would be more confusing to find your key now refers to aiohttp-admin's database connection and not the one you created in your application.

Is this such a prevent issue? For `db` sure, but other keys as well? I would consider it good practice to choose a unique name and if that means to prefix it with the application name, so be it. Is it that nobody is doing it atm, so AppKey should do the name mangling automatically?

--
🤔  Not an easy solution here. I'd have liked to be able to do a one-to-one replacement, but maybe that isn't possible given the circumstances. To avoid the breaking change for Home Assistant then, I should probably assign both a `str` key and an `AppKey` (and deprecated the `str` key as per usual).
```py
app[KEY_HASS] = app["hass"] = hass
```

Dreamsorcerer:
> so AppKey should do the name mangling automatically?

It's not mangling the name, the object itself is the key. It's just the repr() that adds the module name in to help debugging.

> -- 🤔 Not an easy solution here. I'd have liked to be able to do a one-to-one replacement, but maybe that isn't possible given the circumstances. To avoid the breaking change for Home Assistant then, I should probably assign both a `str` key and an `AppKey` (and deprecated the `str` key as per usual).
> 
> ```python
> app[KEY_HASS] = app["hass"] = hass
> ```

Yeah, I feel like duplicating it on the string key makes the most sense for transitioning.

cdce8p:
> > so AppKey should do the name mangling automatically?
> 
> It's not mangling the name, the object itself is the key. It's just the repr() that adds the module name in to help debugging.

The repr name is mangled with the the module name but you're right the key comparison is based on object id.

> > -- 🤔 Not an easy solution here. I'd have liked to be able to do a one-to-one replacement, but maybe that isn't possible given the circumstances. To avoid the breaking change for Home Assistant then, I should probably assign both a `str` key and an `AppKey` (and deprecated the `str` key as per usual).
> > ```python
> > app[KEY_HASS] = app["hass"] = hass
> > ```
> 
> Yeah, I feel like duplicating it on the string key makes the most sense for transitioning.

Not as beautiful as I would've liked but I the reasoning for keeping it the way it is makes sense. I'll go ahead and close this then.

Dreamsorcerer:
> The repr name is mangled with the the module name but you're right the key comparison is based on object id.

Sure, but really, this is just adding more information than we had before with string keys. If you print() the key object (e.g. when iterating through all keys in the Application), it should give you a very good idea of where that key was created and help you to find it a lot easier.

PR #7556, opened 2023-08-25
Only install async-timeout for Python < 3.11

### LINKED ISSUE #7502: Switch to asyncio.timeout()
We should look at whether it is easy to start using `asyncio.timeout()` in Python 3.11+ and drop the async-timeout dependency on newer Python releases.

We could always make changes to async-timeout if it is not trivial to switch between them.

cdce8p:
## What do these changes do?
Replace `async-timeout` with `asyncio.timeout` for Python 3.11+.

Overall the change isn't as clean as I would have liked it to be. That's mostly a result of the name conflict with `timeout`. I'm open for suggestions on how to improve it.

`async-timeout` won't be a dependency for Python 3.11+ any longer.

## Related issue number

Fixes: #7502


aiohttp/helpers.py
```diff
@@ -50,17 +50,22 @@
...
 
+if sys.version_info >= (3, 11):
+    import asyncio as async_timeout
+else:
+    import async_timeout
+
 __all__ = ("BasicAuth", "ChainMapProxy", "ETag")
 
 PY_310 = sys.version_info >= (3, 10)
+PY_311 = sys.version_info >= (3, 11)
```
Dreamsorcerer: Personally, I'm in favour of removing these variables (or atleast not adding new ones). mypy doesn't seem to work well with them, and they seem to actually end up being more work for me when I come to remove them after dropping support for a given Python version.

cdce8p: Works for me! Yeah, mypy was the reason I used `sys.version_info` for the import guard already. Just wasn't sure which you prefer.

cdce8p:
I think, this could be tagged for a backport to `3.9` as well.

Dreamsorcerer:
Great, thanks.

cdce8p:
Backport PR:
- https://github.com/aio-libs/aiohttp/pull/7558

PR #7561, opened 2023-08-27
Remove chardet/charset-normalizer. Add fallback_charset_resolver ClientSession parameter.

### LINKED ISSUE #7126: Use cChardet fork

john-parton:
## What do these changes do?

Continuing work from https://github.com/aio-libs/aiohttp/pull/7559

~This replaces cchardet with chardetng and also paves a path towards deprecating character detection completely. It will issue a DeprecationWarning if the character detection method is invoked~

Need updates.

~Previously, encoding detection was done on every request without a content-type header. This was a bottleneck in some code I wrote. This changes it so that first utf-8 is tried, and then encoding detection happens aftewards.~

Need updates.

## Related issue number

Should close https://github.com/aio-libs/aiohttp/issues/7126

## Checklist


john-parton:
Note: When we decide on which approach to go with, I'll close the other two, and polish the remaining pull request. I haven't done docs or release notes quite as well as I would like.

Dreamsorcerer:
I think the best approach would be to implement the new callable in master and backport to 3.9.
Then do a backport to 3.8, where we set the default callable to the existing chardet approaches, so behaviour should work the same in 3.8, while still allowing people to make the change that will be needed for 3.9.

john-parton:
@Dreamsorcerer 

I've updated this pull request with your suggestions, with a few tweaks

1. ClientSession now takes the argument `default_encoding`. I added this to the `__init__` function and the corresponding private property to `__slots__`.
2. I chose `default_encoding` over `fallback_encoding` because the word `default` is used elsewhere and is a more familiar word
3. The valid types for `default_encoding` are `None`, `str`, `Callable[[ClientRequest], str]`. I chose only to pass the ClientRequest and not the body because in other places in the code where callables like this are passed around, that's usually how it's done. However, as I'm typing this I realize that a user will have to access `_body` directly, and won't be able to use `read()` because it's a coro. We can change the signature for this to include the body, or we can make it return `Awaitable[str]` if we want to enforce that the only parameter is `ClientRequest`, but also allow accessing of `_body` via a public method.

If this general approach looks good, I'll start updating the docs.

Dreamsorcerer:
> I chose `default_encoding` over `fallback_encoding` because the word `default` is used elsewhere and is a more familiar word

I feel like default makes it sound more like it should be a constant than a function, though I realise fallback isn't a lot better. Maybe something else like `charset_func` or something..?

> The valid types for `default_encoding` are `None`, `str`, `Callable[[ClientRequest], str]`.

I think this adds unnecessary complexity and performance overhead. If we just accept a callable only, it is trivial for someone to do `lamba *args: "encoding"`, and likewise for the default value.

> I chose only to pass the ClientRequest and not the body because in other places in the code where callables like this are passed around, that's usually how it's done.

The one thing that a charset guesser will need is the body (or part of it) and the text() method will already have read the body, so I think we should just pass it through and same any trouble.

john-parton:
Charset guesser might also need the tld, as we already discussed. Removing `str` seems reasonable as well.

So  I think `Callable[[ClientResponse, bytes], str]` and `None` would cover it?

john-parton:
We could literally just make the function called `encoding_guesser` or `charset_guesser`.

Dreamsorcerer:
> Charset guesser might also need the tld, as we already discussed.

It's available in the Response, which I why I included it in the original examples.

> So I think `Callable[[ClientResponse, bytes], str]` and `None` would cover it?

We don't need `None`, just default to `lambda: r, b: "utf-8"`.

Dreamsorcerer:
> We could literally just make the function called `encoding_guesser` or `charset_guesser`.

Yeah, I guess that's fine, just might be a little confusing given that we default to a function which doesn't do any guessing.

john-parton:
Okay, I changed the variable/callback to 'fallback_encoding', I change the type signature to just be Optional[Callable], and I updated code in appropriate places.

As far as code snippets are concerned for instructing a user on how to operate the character set detection once the feature is removed, here are two options

```python3
# chardetng_py: fast charset detection written in rust
from chardetng_py import detect

def fallback_encoding(response, body):
    tld = response.url.host.split(".")[-1]
    return detect(body, tld=tld)

# cChardet, chardet, charset_normalizer
from charset_normalizer import chardet

def fallback_encoding(response, body):
    return chardet(body)["encoding"] or "utf-8"

async with ClientSession(fallback_encoding=fallback_encoding) as client:
    ....
```

Dreamsorcerer:
I think things are getting mixed up a little.

Can we start with a PR that removes chardet and supports the new parameter? This will go into master. Once that's in we can look at backports and including chardetng in the 3.8 backport.

john-parton:
No problem.

john-parton:
@Dreamsorcerer So this removes the character set detection now and introduces the `fallback_encoding` callback we talked about.

Do you want me to update the `ClientSession` documentation and the relevant `text()` method documentation to refer to that?

aiohttp/client.py
```diff
@@ -221,6 +222,7 @@ def __init__(
         read_bufsize: int = 2**16,
         max_line_size: int = 8190,
         max_field_size: int = 8190,
+        fallback_encoding: Optional[Callable[[ClientResponse, bytes], str]],
```
```suggestion
        fallback_encoding: Callable[[ClientResponse, bytes], str] = lambda r, b: "utf-8",
```

aiohttp/client_reqrep.py
```diff
@@ -1011,36 +1006,33 @@ def get_encoding(self) -> str:
...
-            ):
-                # RFC 7159 states that the default encoding is UTF-8.
-                # RFC 7483 defines application/rdap+json
-                encoding = "utf-8"
-            elif self._body is None:
-                raise RuntimeError(
-                    "Cannot guess the encoding of " "a not yet read body"
-                )
-            else:
-                encoding = chardet.detect(self._body)["encoding"]
```
Dreamsorcerer, line 1031: Given this is a public, documented method, I think we should call self._fallback_encoding() here, and keep the logic similar to before. That probably means not using the try/except logic in text()..

aiohttp/client_reqrep.py
```diff
@@ -1011,36 +1006,33 @@ def get_encoding(self) -> str:
...
-        if encoding is None:
-            encoding = self.get_encoding()
+        # If the user specified the encoding, do not attempt to guess the
+        # encoding. May raise UnicodeDecodeError if user specified an invalid
+        # encoding
+        if encoding is not None:
+            return self._body.decode(encoding, errors=errors)  # type: ignore[union-attr]
+
+        try:
+            return self._body.decode(self.get_encoding(), errors=errors)  # type: ignore[union-attr]
```
Dreamsorcerer: The old code achieves the same thing here without code duplication.

```suggestion
        if encoding is None:
            encoding = self.get_encoding()

        try:
            return self._body.decode(encoding, errors=errors)  # type: ignore[union-attr]
```

setup.cfg
```diff
@@ -64,8 +64,6 @@ speedups =
   # required c-ares (aiodns' backend) will not build on windows
   aiodns >= 1.1; sys_platform=="linux" or sys_platform=="darwin"
   Brotli
-  # cchardet is unmaintained: aio-libs/aiohttp#6819
-  cchardet; python_version < "3.10"
```
Dreamsorcerer, line 68: charset-normalizer needs to be removed as well.

Dreamsorcerer:
> Do you want me to update the `ClientSession` documentation and the relevant `text()` method documentation to refer to that?

Yep, we'll want that included too with examples of correctly using it for chardetng and charset-normalizer.

aiohttp/client_reqrep.py
john-parton: If I keep the code structured the way that it was before, `_fallback_encoding()` will _always_ get called, even if the content type header is populated or if the response body is already valid utf8.

Structuring it closer to how I have makes it a true fallback.

Dreamsorcerer: No, the old code only calls it if `not encoding`, after getting it from the header. The only difference between your proposal is that it won't try to decode as utf-8 first, which is not something I'm strongly tied to, I think it could damage performance is some cases and improve it in others.

john-parton: OK, I think I understand what you're asking for. I'll see if I can deliver.

john-parton: @Dreamsorcerer I have the structure of the method largely preserved here. 

Somewhat related: Should `get_encoding()` return the value from the header _or_ should it return Python's value?

For instance, if the header says the encoding is "windows-1252", should this function return "windows-1252" or "cp1252"?

Dreamsorcerer: Probably whatever it did before, which I guess is Python's? Although both those values appear to be valid in .encode().

aiohttp/client_reqrep.py
```diff
@@ -1013,26 +1008,23 @@ def get_encoding(self) -> str:
         encoding = mimetype.parameters.get("charset")
         if encoding:
             try:
-                codecs.lookup(encoding)
+                return codecs.lookup(encoding).name
             except LookupError:
```
```
with suppress(LookupError):
    return codecs.lookup(encoding).name
```

tests/test_client_response.py
Dreamsorcerer, line 1: We'll probably want a new test for the method. Probably just pass something like `lambda r, b: "windows-1252"` and test that it uses that encoding as the fallback.

docs/client_reference.rst
Dreamsorcerer, line 1: I think under client_advanced.rst we'll also want a new section that demonstrates the usage. It should also mention the exception that users are likely to see if they do need this (so when they search for their exception they are likely to find these docs).

aiohttp/client_reqrep.py
john-parton: Neat trick. I've not seen that yet.

aiohttp/client_reqrep.py
john-parton: The previous behavior appeared to return the header value, actually. It calls codecs.lookup(), but it doesn't actually return from it, nor assign the output to any variable. Then flow control continues to the bottom where it will return the header value unchanged

aiohttp/client_reqrep.py
Dreamsorcerer: It's in `contextlib`.

aiohttp/client_reqrep.py
Dreamsorcerer: Right, I don't think it really matters, as both appear to work. But, I guess the Python version is what I'd expect for the most part.

tests/test_client_response.py
john-parton: I did a test like that. I chose cp1252 to prefer Python's encoding, but I could just as well do windows-*.

I also removed the two tests that were specific to encoding detection.

tests/test_client_response.py
```diff
@@ -453,35 +454,7 @@ def side_effect(*args, **kwargs):
...
-        loop=loop,
-        session=session,
-    )
-
-    def side_effect(*args, **kwargs):
-        fut = loop.create_future()
-        fut.set_result('{"тест": "пройден"}'.encode("cp1251"))
-        return fut
-
-    response._headers = {"Content-Type": "text/plain;charset=invalid"}
```
Dreamsorcerer, line 476: This one tests when charset=invalid, which we probably still want to test. Though you could probably just parametrize the above test, so it tests with no charset and charset=invalid (rather than needing to duplicate all the code).

john-parton: Yeah, I see that. In the docs I'm updating I keep having to say "no charset, or a charset that python doesn't understand."

john-parton: @Dreamsorcerer I parameterized the test. I haven't done that before, so do make sure I did it correctly.

docs/client_reference.rst
Dreamsorcerer: If we can get this one included for tomorrow, then I'll do a final run-through and get it merged.

john-parton:
@Dreamsorcerer I did my best. I'm not great at writing documentation. I tried to make some of the finer points clear, and I documented that `text()` will raise `UnicodeDecodeError` if not configured correctly, and I added a similar note in the "advanced usage" section that I added, so hopefully that will help with google-abilty-of-error.

webknjaz:
I see that `fallback_encoding` argument is a callable. Functions/methods should have names hinting at the action they do, starting with a verb. A noun in the name suggests that it's a pre-defined piece of static data. Anybody can fairly confidently guess that `fallback_encoding` must mean "an encoding to be used as a backup, like „utf-8”", not "a function to call if encoding is unknown".

A better name could be `detect_encoding`, for example.

Dreamsorcerer:
> A better name could be `detect_encoding`, for example.

Great, we were debating a good name further up.

john-parton:
This is somewhat unrelated, but I just realized there's no support for <meta charset="XXX">

Shouldn't we look for this tag?

Assuming the content type makes sense, obviously. (Not present, text/html)

Dreamsorcerer:
> This is somewhat unrelated, but I just realized there's no support for <meta charset="XXX">
> 
> Shouldn't we look for this tag?

Probably not, this is an HTTP library, not an HTML one (I'm not completely against the idea, but I think it'd add a little too much complexity for little gain, so you'd have to convince another maintainer first). The user can always add support for this in their custom function (or you could even add support for it in chardetng?).

john-parton:
> > This is somewhat unrelated, but I just realized there's no support for 
> > Shouldn't we look for this tag?
> 
> Probably not, this is an HTTP library, not an HTML one (I'm not completely against the idea, but I think it'd add a little too much complexity for little gain, so you'd have to convince another maintainer first). The user can always add support for this in their custom function (or you could even add support for it in chardetng?).

Note: chardetng_py is a Pythonic binding to the chardetng rust library. It has a *very* limited scope. If I wanted to add a convenient plugin to do better decoding of html resposnes, I would create yet-another-library.

Regarding "this is an HTTP library, not an HTML one", HTTP and HTML are acronyms for Hyper Text Transfer Protocol and Hyper Text Markup Language. HTML is literally the lingua franca of the web, despite recent changes. Following that line of "this is a http library" logic, why have a `json()` method? Furthermore, why have a `text()` method at all? Can't users just do `.read().decode()`?

I understand not wanting to increase the scope of the project or add unnecessary complexity, but I really think this is where the complexity *ought* to be.

john-parton:
I would also like to point out in terms of "adding complexity":

This pull request's diff is +93 −171
This pull request to just swap out cchardet https://github.com/aio-libs/aiohttp/pull/7559 is  +29 −24 

So in the sense of "adding complexity", maybe it's better to just take that approach.

docs/client_advanced.rst
```diff
@@ -740,3 +740,32 @@ HTTP Pipelining
...
+    def fallback_encoding(response, body):
+        return chardet(body)["encoding"] or "utf-8"
+
+    async with ClientSession(fallback_encoding=fallback_encoding) as client:
+        ....
+
+.. note:: Attempting to call :meth:`ClientResponse.text()` without setting
+          `fallback_encoding` will result in a :exc:`UnicodeDecodeError` exception
+          if the charset is not specified in the Content-Type header and the response
+          body is not valid UTF-8.
```
Dreamsorcerer: Just some tweaks that I think will be a little better:

```suggestion
If you encounter a :exc:`UnicodeDecodeError` when using :meth:`ClientResponse.text()`
this may be because the response does not include the charset needed
to decode the body.

If you know the correct encoding for a request, you can simply specify
the encoding as a parameter (e.g. ``resp.text("windows-1252")``).

Alternatively, :class:`ClientSession` accepts a ``fallback_encoding`` parameter which
can be used to introduce charset guessing functionality. When a charset is not found
in the Content-Type header, this function will be called to get the charset encoding. For
example, this can be used with the chardetng_py library.::

    from chardetng_py import detect

    def fallback_encoding(resp: ClientResponse, body: bytes) -> str:
        tld = resp.url.host.rsplit(".", maxsplit=1)[-1]
        return detect(body, allow_utf8=True, tld=tld)

    ClientSession(fallback_encoding=fallback_encoding)

Or, if ``chardetng_py`` doesn't work for you, then ``charset-normalizer`` is another option::

    from charset_normalizer import chardet

    ClientSession(fallback_encoding=lamba r, b: chardet(b)["encoding"] or "utf-8")
```

docs/client_reference.rst
```diff
@@ -208,6 +209,14 @@ The client session supports the context manager protocol for self closing.
                          disabling.  See :ref:`aiohttp-client-tracing-reference` for
                          more information.
 
+   :param Callable[[ClientResponse, bytes], str] fallback_encoding:
+      any :term:`callable` that takes a :class:`client response <ClientResponse>` object
+      and its :type:`contents <bytes>` and returns a :class:`str` encoding to use if
+      the response does not specify one.  By default, the encoding is assumed to be
+      ``utf-8``.
+
+      .. versionadded:: 4.0
```
Dreamsorcerer: We'll backport the feature.

```suggestion
      .. versionadded:: 3.8.6
```

docs/client_reference.rst
```diff
@@ -208,6 +209,14 @@ The client session supports the context manager protocol for self closing.
                          disabling.  See :ref:`aiohttp-client-tracing-reference` for
                          more information.
 
+   :param Callable[[ClientResponse, bytes], str] fallback_encoding:
+      any :term:`callable` that takes a :class:`client response <ClientResponse>` object
+      and its :type:`contents <bytes>` and returns a :class:`str` encoding to use if
+      the response does not specify one.  By default, the encoding is assumed to be
+      ``utf-8``.
```
```suggestion
      A :term:`callable` that accepts a :class:`ClientResponse` and its
      :type:`contents <bytes>` and returns a :class:`str` which will be used as
      the encoding parameter to :meth:`bytes.decode()`.
      
      This function will be called when the charset is not known (e.g. not specified in the
      Content-Type header). The default function simply defaults to ``utf-8``.
```

docs/client_reference.rst
```diff
@@ -1406,12 +1415,8 @@ Response object
...
       specified *encoding* parameter.
 
-      If *encoding* is ``None`` content encoding is autocalculated
-      using ``Content-Type`` HTTP header and *charset-normalizer* tool if the
-      header is not provided by server.
-
-      :term:`cchardet` is used with fallback to :term:`charset-normalizer` if
-      *cchardet* is not available.
+      If *encoding* is ``None`` content encoding is determined using
+      :meth:`get_encoding`.
```
```suggestion
      If *encoding* is ``None`` content encoding is determined from the
      Content-Type header, or using the ``fallback_encoding`` function.
```

docs/client_reference.rst
```diff
@@ -1420,35 +1425,21 @@ Response object
                            ``None`` for encoding autodetection
                            (default).
 
-      :return str: decoded *BODY*
-
-      :raise LookupError: if the encoding detected by cchardet is
-                          unknown by Python (e.g. VISCII).
 
-      .. note::
+      :raises: :exc:`UnicodeDecodeError` if encoding is unknown. See also
```
```suggestion
      :raises: :exc:`UnicodeDecodeError` if decoding fails. See also
```

docs/glossary.rst
```diff
@@ -50,16 +50,18 @@
...
+
        https://pypi.org/project/charset-normalizer/
 
    cchardet
 
        cChardet is high speed universal character encoding detector -
        binding to charsetdetect.
 
-       https://pypi.python.org/pypi/cchardet/
+       Prior versions of aiohttp used cchardet as an optional speedup.
```
Dreamsorcerer: Can probably just delete these 2 sections.

john-parton:
> > A better name could be `detect_encoding`, for example.
> 
> Great, we were debating a good name further up.

Should I go ahead and update it to be `detect_encoding`? It sounds better to my ear.

Dreamsorcerer:
> Should I go ahead and update it to be `detect_encoding`? It sounds better to my ear.

Yeah, go ahead. Still sounds a bit more like a bool to me, but I've got nothing better.

john-parton:
> > This is somewhat unrelated, but I just realized there's no support for 
> > Shouldn't we look for this tag?
> 
> Probably not, this is an HTTP library, not an HTML one (I'm not completely against the idea, but I think it'd add a little too much complexity for little gain, so you'd have to convince another maintainer first). The user can always add support for this in their custom function (or you could even add support for it in chardetng?).

I've changed my mind on the subject of looking for the meta tag.

If you pass the entire body directly as bytes to lxml, lxml does this anyways, and what else would be doing with the html body if you actually cared about that tag?

Not worth it. I agree.

john-parton:
LXML is rather smart.

```python
>>> doc = etree.fromstring(b"""<html><head><meta charset="cp1251"></head><body><p>\xef\xf0\xee\xe9\xe4\xe5\xed</p></body></html>""", parser)
>>> doc.xpath("//p/text()")
['пройден']
>>> doc = etree.fromstring(b"""<html><head></head><body><p>\xef\xf0\xee\xe9\xe4\xe5\xed</p></body></html>""", parser)
>>> doc.xpath("//p/text()")
['ïðîéäåí']
```

john-parton:
> > Should I go ahead and update it to be `detect_encoding`? It sounds better to my ear.
> 
> Yeah, go ahead. Still sounds a bit more like a bool to me, but I've got nothing better.

@Dreamsorcerer Done. I've had a lot of commits. I've been squashing some commits as I go when we get to reasonable stopping points. Let me know if you need anything else.

Dreamsorcerer:
> I understand not wanting to increase the scope of the project or add unnecessary complexity, but I really think this is where the complexity _ought_ to be.

Going back to performance, with a really simple solution (looking for a literal `<meta charset=`), I got around 3-7 times slower than doing the default callback for utf-8. Using regex instead was atleast 9 times slower. With worse results if the tag is not there.

That barely covers a minimal use case. If you want to support `'` and `"`, or support http-equiv, or arbitrary ordering of attributes etc., then it's going to quickly get much slower and likely take just as long as the charset guesser (your lxml example was 3x slower than chardetng for me).

john-parton:
Right, so I think we're on the same page.

I think that using the charset meta tag is in some cases going to be more "correct" than just guessing the encoding, and most likely the user can make a better judgement call than any default implementation we put in here.

chardetng being faster than lxml [for this specific function] is nice! There's a reason I spent so many hours writing that binding. :laughing:

Dreamsorcerer:
> @Dreamsorcerer Done. I've had a lot of commits. I've been squashing some commits as I go when we get to reasonable stopping points.

Whole PRs are squash committed, so no need to mess with it.

webknjaz:
> > Should I go ahead and update it to be `detect_encoding`? It sounds better to my ear.
> 
> Yeah, go ahead. Still sounds a bit more like a bool to me, but I've got nothing better.

I think it sounds like a bool because it's used in the context of a function argument. We could add disambiguate it by appending `_callable` at the end. This would take into account the abstraction layer of the caller, but would be a little weird in the framework code. Though, we can rename it on the class instance itself so that it'd be assigned like this: `self._detect_encoding = detect_encoding_callable`. This way, it would correspond to abstraction layers where it's used.

john-parton:
> > > Should I go ahead and update it to be `detect_encoding`? It sounds better to my ear.
> > 
> > 
> > Yeah, go ahead. Still sounds a bit more like a bool to me, but I've got nothing better.
> 
> I think it sounds like a bool because it's used in the context of a function argument. We could add disambiguate it by appending `_callable` at the end. This would take into account the abstraction layer of the caller, but would be a little weird in the framework code. Though, we can rename it on the class instance itself so that it'd be assigned like this: `self._detect_encoding = detect_encoding_callable`. This way, it would correspond to abstraction layers where it's used.

Searching through the code, usually passing around callables are "callback" or "handler". If you want it to be consistent with the rest of the codebase, `detect_encoding_callback` or `detect_encoding_handler` are closer to other parameters.

See 

handlers

AbstractMatchInfo.handler
AbstractMatchInfo.expect_handler
RequestHandler._handle_request

callbacks

Connection._callbacks
TimeoutHandler.register(callback, ...)
TimeoutHandler._callbacks

EDIT I kind of hate discussion about naming things, insert Phil Karlton quote.

webknjaz:
Handler is not a good option since it's usually used in the context of events and errors. Callbacks are usually used in the context of side effects that the calling code won't be waiting for.

So I'd stick with a callable since it's a legit Python term that is generic enough and relays the exact properly that is expected — implementing the `__call__()` protocol.

The rest of the name should be specific to what it does.

Dreamsorcerer:
I'm not sure what's wrong with the docs. I assume something is wrong with the formatting of the new param, which is causing things after it to be treated as global, but can't figure out what is wrong with it...

Dreamsorcerer:
> I'm not sure what's wrong with the docs. I assume something is wrong with the formatting of the new param, which is causing things after it to be treated as global, but can't figure out what is wrong with it...

Ugh, was just a missing \ right at the top.

docs/client_advanced.rst
```diff
@@ -740,3 +740,33 @@ HTTP Pipelining
...
+
+Alternatively, :class:`ClientSession` accepts a ``detect_encoding`` parameter which
+can be used to introduce charset guessing functionality. When a charset is not found
+in the Content-Type header, this function will be called to get the charset encoding. For
+example, this can be used with the ``chardetng_py`` library.::
+
+    from chardetng_py import detect
+
+    def detect_encoding(resp: ClientResponse, body: bytes) -> str:
+        tld = resp.url.host.rsplit(".", maxsplit=1)[-1]
```
john-parton, line 763: nice with the maxsplit. You could also use `rpartition` if you want to be even fancier.

Dreamsorcerer:
> Handler is not a good option since it's usually used in the context of events and errors. Callbacks are usually used in the context of side effects that the calling code won't be waiting for.
> 
> So I'd stick with a callable since it's a legit Python term that is generic enough and relays the exact properly that is expected — implementing the `__call__()` protocol.
> 
> The rest of the name should be specific to what it does.

OK, I threw the question into ChatGPT, to see if it had any good ideas. How does `charset_resolver` sound?

webknjaz:
> How does charset_resolver sound?

`fallback_charset_resolver`? That's probably fine as a public API.
But the private instance attribute should still have a verb in the name since it'll be called.

Do we need to add a `Co-Authored-By: ChatGPT` trailer to the Git commit message? 😂

docs/client_reference.rst
```diff
@@ -208,6 +209,16 @@ The client session supports the context manager protocol for self closing.
                          disabling.  See :ref:`aiohttp-client-tracing-reference` for
                          more information.
 
+   :param Callable[[ClientResponse,bytes],str] fallback_charset_resolver:
```
Dreamsorcerer, line 212: Not sure if there is a way to get Sphinx to recognise this as the type while including spaces? Otherwise it takes the type as `Callable[[ClientResponse,` and the parameter name as `bytes, str] fallback_charset_resolver`...

Dreamsorcerer: I'm going to merge, if someone has a fix, it can be done in a followup PR.

Dreamsorcerer:
@john-parton If you could handle the backports, that'd be great. We'll need to:

- [x] Follow the instructions above for 3.9 backport. This should be a straight backport.
- [x] Follow the instructions above for 3.8 backport.
  - [x] Modify the 3.8 backport to change the default function to emulate the current chardet behaviour (preferably without chardetng).
  - [x] Add a DeprecationWarning in the default function when it is called (or maybe when it returns a non "utf-8" value?).
- [ ] Once they are both merged, create a new PR on 3.8 branch that replaces cchardet with chardetng.

docs/client_advanced.rst
```diff
@@ -740,3 +740,33 @@ HTTP Pipelining
 ---------------
 
 aiohttp does not support HTTP/HTTPS pipelining.
+
+
+Character Set Detection
+-----------------------
+
+If you encounter a :exc:`UnicodeDecodeError` when using :meth:`ClientResponse.text()`
```
webknjaz, line 748: Does this actually work? The correct syntax AFAIK doesn't include `()`.

docs/client_reference.rst
webknjaz: @Dreamsorcerer I think that this should work
```rst
:type fallback_charset_resolver: Callable[[ClientResponse, bytes], str]
```

docs/client_advanced.rst
webknjaz: And so it doesn't: https://aiohttp--7561.org.readthedocs.build/en/7561/client_advanced.html#character-set-detection.

cc @Dreamsorcerer

Dreamsorcerer: Ah, sorry, I actually copied from the section above, which makes the same mistake with .close(). I wonder why Sphinx doesn't produce a warning...

webknjaz: Dunno. Maybe because of this ignore https://github.com/aio-libs/aiohttp/blob/6755796/docs/conf.py#L384 or some Sphinx but, or a corner case. Somebody needs to go through all those ignores and fix them.

webknjaz: @Dreamsorcerer so it looks like we don't run the normal docs build @ GHA and spellcheck that we do run doesn't emit those warnings: https://github.com/aio-libs/aiohttp/actions/runs/6018473896/job/16326723444#step:13:18.

OTOH, RTD _does have_ those warnings in the log but isn't set up to turn them into errors: https://readthedocs.org/projects/aiohttp/builds/21761448/.

I think the CI was set up to fail on warnings at some point. Maybe that got removed, or I'm just confusing the repos…

Dreamsorcerer: Well, other warnings seem to fail CI in the docs spelling step (I'm not aware of another part that produces warnings):
https://github.com/aio-libs/aiohttp/actions/runs/6017616744/job/16324167311#step:13:86

john-parton:
> @john-parton If you could handle the backports, that'd be great. We'll need to:

I don't have the time to dedicate this week.

docs/client_reference.rst
Dreamsorcerer: That's not working, unfortunately.

Dreamsorcerer:
3.9 backport done: #7586
If you have some time to handle the 3.8 backport and finish https://github.com/aio-libs/aiohttp/pull/7561#issuecomment-1698287181 soon, that'd be great. Hoping to get a new release out shortly.

john-parton:
I'll take a crack at it later today.

john-parton:
@Dreamsorcerer PR opened here: https://github.com/aio-libs/aiohttp/pull/7589

PR #7577, opened 2023-09-04
CookieJar - return 'best-match' and not LIFO

marq24:
## What do these changes do?
The CookieJAR can contain multiple cookies with the identical name (e.g. JSESSIONID) for different paths of the same domain

the `filter_cookies` have to make sure that the cookie with the best-matching path will be returned. In the current implementation that last matching cookie that was inserted into the JAR will return... 

I came across this issue in the situation, that a request to `/auth` set a SESSION_A, then redirected to `/` set another (different SESSION_B), redirected me again to `/auth` where the current implementation returned the SESSION_B for `/` (instead of the SESSION_A for `/auth`

I might do things here completely wrong - please excuse - I am far from being a python expert (I consider myself more a newbie)

## Tests?
I added an additional test in the cookiejar.py `test_path_filter_diff_folder_same_name_return_best_match_independent_from_put_order` that fail with the current implementation, but with the change/fix it test is passed

IMHO - nope

## Related issue number
PRNUM: #7577

## Checklist

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -256,6 +256,7 @@ class CookieJar(AbstractCookieJar):
             and request_origin not in self._treat_as_secure_origin
         )
 
+        # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
         for cookie in sorted(self, key=lambda c: len(c["path"])):
             name = cookie.key
             domain = cookie["domain"]
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -282,12 +282,10 @@ class CookieJar(AbstractCookieJar):
 
             # we need to check, if the found cookie is a better match then the
             # previously found
-            if name in filtered:
-                existing_cookie = filtered[name]
-                if len(existing_cookie.get("path")) < len(cookie.get("path")):
-                    filtered[name] = cookie
-            else:
-                filtered[name] = cookie
+            if name in filtered and len(cookie["path"]) < len(filtered[name]["path"]):
+                continue
+
+            filtered[name] = cookie
 
         # It's critical we use the Morsel so the coded_value
         # (based on cookie version) is preserved
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -256,7 +256,7 @@ class CookieJar(AbstractCookieJar):
             and request_origin not in self._treat_as_secure_origin
         )
 
-        for cookie in self:
+        for cookie in sorted(self, key=lambda c: len(c["path"])):
             name = cookie.key
             domain = cookie["domain"]
 
@@ -280,23 +280,10 @@ class CookieJar(AbstractCookieJar):
             if is_not_secure and cookie["secure"]:
                 continue
 
-            # we need to check, if the found cookie is a better match then the
-            # previously found
-            if name in filtered and len(cookie["path"]) < len(filtered[name]["path"]):
-                continue
-
-            filtered[name] = cookie
-
-        # It's critical we use the Morsel so the coded_value
-        # (based on cookie version) is preserved
-        for name in filtered:
-            filtered_cookie = filtered[name]
-            mrsl_val = cast(
-                "Morsel[str]", filtered_cookie.get(filtered_cookie.key, Morsel())
-            )
-            mrsl_val.set(
-                filtered_cookie.key, filtered_cookie.value, filtered_cookie.coded_value
-            )
+            # It's critical we use the Morsel so the coded_value
+            # (based on cookie version) is preserved
+            mrsl_val = cast("Morsel[str]", cookie.get(cookie.key, Morsel()))
+            mrsl_val.set(cookie.key, cookie.value, cookie.coded_value)
             filtered[name] = mrsl_val
 
         return filtered
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -284,7 +284,7 @@ class CookieJar(AbstractCookieJar):
             # previously found
             if name in filtered:
                 existing_cookie = filtered[name]
-                if len(existing_cookie.get('path')) < len(cookie.get('path')):
+                if len(existing_cookie.get("path")) < len(cookie.get("path")):
                     filtered[name] = cookie
             else:
                 filtered[name] = cookie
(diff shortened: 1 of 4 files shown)


aiohttp/cookiejar.py
```diff
@@ -280,10 +280,25 @@ def filter_cookies(
...
+        # It's critical we use the Morsel so the coded_value
+        # (based on cookie version) is preserved
+        for name in filtered:
+            filtered_cookie = filtered[name]
+            mrsl_val = cast(
+                "Morsel[str]", filtered_cookie.get(filtered_cookie.key, Morsel())
+            )
+            mrsl_val.set(
+                filtered_cookie.key, filtered_cookie.value, filtered_cookie.coded_value
+            )
```
Dreamsorcerer: Can't this whole change be reduced to:
```
if name in filtered and len(cookie["path"]) < len(filtered[name]["path"]):
    continue
```

Alternatively, I'm wondering if we should do it like the spec says and simply order the cookies by path length. At which point we don't need to add the comparisons in here, just take the first cookie that comes through.

Dreamsorcerer: The spec also suggests that we should actually be sending both cookie values through (e.g. `Cookie: path-cookie=one; path-cookie=zero`).

But, I don't see a way to represent that with SimpleCookie/Morsel.

Dreamsorcerer: http.cookiejar does appear to do something like this actually:
https://github.com/python/cpython/blob/main/Lib/http/cookiejar.py#L1306

So, e.g. I get a result like (regardless of the order of the 2 cookies) `'Cookie: $Version=1; foo=one; $Path="/bar"; $Port; foo=zero; $Path="/"; $Port'`.

Though, that is using http.cookiejar.Cookie, not http.cookies.SimpleCookie. Looking through, I think the latter is meant more for server-side code, so maybe aiohttp is using the wrong objects here...

marq24: first of all thanks spending time with this PR

I agree with all of your points - specially the the implementation should follow the spec... But I have to admit, that my current knowledge of python is way to limited in order to provide an spec-valid implementation.

I hope you can agree, that the (based on your first comment adjusted) PR will improve the current situation. For me it was already a tough challenge to create this PR (trying to fulfill all the requirements mentioned in https://docs.aiohttp.org/en/stable/contributing.html) - but looks like I made it :-)

If it's better in your opinion to withdraw this PR, create a Issue describing the situation/bug and then work on a complete  solution for the situation this is totally fine for me - using this this PR as some sort of starting point. But then I can't be part in this process (simply cause of my knowledge - and I don't want to break things that had been working for years). So then 'somebody-else' has to provide a valid PR (I don't have to become a contributor).

So it's totally up to the maintainer-team to make a decision here - I came across this (IMHO serious) bug last Friday afternoon when I was working on a migration of an home assistant integration (from 'request' to 'aiohttp') - and it took me the half Saturday to find the root cause - so I wanted to let you know, that 'there is something' (but also already include a 'working-fix').

Dreamsorcerer: I suspect sending multiple values would need a bigger change, possibly backwards-incompatible, so this is certainly the correct first step.

aiohttp/cookiejar.py
```diff
@@ -280,10 +280,23 @@ def filter_cookies(
...
```
Dreamsorcerer: In my first comment, I meant that this whole section could also be reverted, we only need 2 lines of code added. Unless I'm misunderstanding something.

marq24: My understanding (why this last lines are needed)

a)
in order to be able to check the "path" len in the code a cookie object was/is needed in the filtered list object

b)
original code convert cookie object to Morsel object before adding it to the filtered list (but that Mosel does not have a 'path' member - so I store temporary the cookie object itself)

c)
I did not wanted to change the return type of the method from Mosel to Cookie...

so that was/is the reason why I thought, the final convert from all entries in the filtered object from cookie to Mosel are required...

Dreamsorcerer: Right, in this case, I think it might make more sense to just sort the cookies before iterating through them (as suggested by the spec, and done in the upstream CookieJar: https://github.com/python/cpython/blob/a0773b89dfe5cd2190d539905dd89e7f6455668e/Lib/http/cookiejar.py#L1307).

Dreamsorcerer: Then we won't need to check the cookies here. If we do reverse sorting, then we don't need to do anything at all here, the last (longest path) cookie will be the one which is set.

marq24: Again I fully agree with opinion, that sorting the "cookie" storage is a better solution of the problem - but I have to admit, with my current (python) knowledge I am not able to implement such a functionality... If you have an idea, how to do that - please do so and just cancel this PR.

Dreamsorcerer: I would've thought copy/paste the sort function would suffice.
Probably just: `sorted(self, key=lambda c: len(c["path"]))`

I noticed there is also a tie-breaker in the spec too:
```
       *  Among cookies that have equal-length path fields, cookies with
          earlier creation-times are listed before cookies with later
          creation-times.
```
But, I don't think we store creation time currently, so will have to ignore for now.

aiohttp/cookiejar.py
Dreamsorcerer: I've noted these improvements in #7583, maybe someone will pick them up.

aiohttp/cookiejar.py
Dreamsorcerer: Actually, the `iter()` shouldn't even be needed. Updated above.

marq24: I tried to include the line `sorted(self, key=lambda c: len(c["path"]))` after the `self._do_expiration()` - but the content of the self._cookies does not change it's order... [inspecting in the debugger]

The result/return-type of the sorted method is IMHO also different type then the Jar itself... I tried also sorting something else then 'self' (self._cookies or self.cookies.values()) - but all that was not successful... please don't laugh too loud (it might be totally stupid what I am trying to do based on my java knowledge)

Dreamsorcerer: _cookies is a dict (nested dicts as well...), so it would be tricky to resort it in-place. That function just returns a sorted list of all the cookies. Just use the return result in the `for` loop.

Dreamsorcerer: When you iterate over `self` (which the sorted() function is doing), it just calls the `__iter__()` method:
https://github.com/aio-libs/aiohttp/blob/4639b36852f3b4939e61311d7863f1d7450d500d/aiohttp/cookiejar.py#L141-L142

As you can see from that (assuming you can follow the `yield` syntax), it basically iterates over every cookie in the nested dicts. So, that's why you see a difference in types (it's basically flattening the dict structure into a list).

marq24: should I (or have to) squash now all commits? - since "now" with the latest commit I guess we are done?

Dreamsorcerer: It squashes on merge. Force pushes only make it more difficult to review changes.

aiohttp/cookiejar.py
```diff
@@ -256,7 +256,7 @@ def filter_cookies(
             and request_origin not in self._treat_as_secure_origin
         )
 
-        for cookie in self:
+        for cookie in sorted(self, key=lambda c: len(c["path"])):
```
Dreamsorcerer, line 260:
```suggestion
        # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
        for cookie in sorted(self, key=lambda c: len(c["path"])):
```

Dreamsorcerer:
Thanks a lot. I've sorted out the backports, so it'll be in the next release.

PR #7589, opened 2023-09-07
Remove chardet/charset-normalizer.

john-parton:
Add fallback_charset_resolver ClientSession parameter. (#7561)

Co-authored-by: Sam Bull <git@sambull.org>
(cherry picked from commit 675579699422680607108a7dd68c85ec5284220c)

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -167,6 +167,21 @@ _RetType = TypeVar("_RetType")
 _CharsetResolver = Callable[[ClientResponse, bytes], str]
 
 
+def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) -> str:
+
+    ret: str = chardet.detect(body)["encoding"] or "utf-8"
+
+    if ret != "utf-8":
+        warnings.warn(
+            "Automatic character set detection is deprecated, use "
+            "fallback_charset_resolver instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+
+    return ret
+
+
 class ClientSession:
     """First-class interface for making HTTP requests."""
 
@@ -226,8 +241,8 @@ class ClientSession:
         requote_redirect_url: bool = True,
         trace_configs: Optional[List[TraceConfig]] = None,
         read_bufsize: int = 2**16,
-        fallback_charset_resolver: _CharsetResolver = lambda r, b: (
-            chardet.detect(b)["encoding"] or "utf-8"
+        fallback_charset_resolver: _CharsetResolver = (
+            _default_fallback_charset_resolver
         ),
     ) -> None:
         if loop is None:
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -173,8 +173,8 @@ def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) ->
 
     if ret != "utf-8":
         warnings.warn(
-            "Automatic character set detection is deprecated, use "
-            "fallback_charset_resolver instead.",
+            "Automatic charset detection will be removed in 3.9, see: "
+            "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",
             DeprecationWarning,
             stacklevel=2,
         )
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -174,7 +174,7 @@ def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) ->
     if ret != "utf-8":
         warnings.warn(
             "Automatic charset detection will be removed in 3.9, see: "
-            "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",
+            "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",  # noqa: E501
             DeprecationWarning,
             stacklevel=3,
         )
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -176,7 +176,7 @@ def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) ->
             "Automatic charset detection will be removed in 3.9, see: "
             "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",
             DeprecationWarning,
-            stacklevel=2,
+            stacklevel=3,
         )
 
     return ret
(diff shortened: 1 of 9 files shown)


john-parton:
@Dreamsorcerer Here's that PR we discussed.

Looks like this isn't quite ready. The code and tests should be correct, but I haven't got the docs updated.

aiohttp/client.py
```diff
@@ -220,6 +226,9 @@ def __init__(
         requote_redirect_url: bool = True,
         trace_configs: Optional[List[TraceConfig]] = None,
         read_bufsize: int = 2**16,
+        fallback_charset_resolver: _CharsetResolver = lambda r, b: (
+            chardet.detect(b)["encoding"] or "utf-8"
+        ),
```
Dreamsorcerer: We're also missing the warning in the second sub-point:
https://github.com/aio-libs/aiohttp/pull/7561#issuecomment-1698287181

So, probably won't be able to use a lambda here.

Then we can just tweak the docs slightly to match.

john-parton: Yeah, I'll factor it out into an underscored/private method and then pass it

john-parton: @Dreamsorcerer I added a basic deprecation warning. I'm not sure the logic for how to determine what the appropriate `stacklevel` kwarg should be, however. I think `3` might be the "correct" value here.

aiohttp/client.py
```diff
@@ -159,6 +164,22 @@ class ClientTimeout:
...
+
+
+def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) -> str:
+
+    ret: str = chardet.detect(body)["encoding"] or "utf-8"
+
+    if ret != "utf-8":
+        warnings.warn(
+            "Automatic character set detection is deprecated, use "
+            "fallback_charset_resolver instead.",
```
```suggestion
            "Automatic charset detection will be removed in 3.9, see: "
            "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",
```

docs/client_advanced.rst
```diff
@@ -640,3 +640,33 @@ are changed so that aiohttp itself can wait on the underlying
...
 <https://github.com/aio-libs/aiohttp/issues/1925>`_ for the progress
 on this.
+
+
+Character Set Detection
+-----------------------
+
+If you encounter a :exc:`UnicodeDecodeError` when using :meth:`ClientResponse.text()`
+this may be because the response does not include the charset needed
+to decode the body.
```
Dreamsorcerer:
```suggestion
If you encounter an 'Automatic charset detection will be removed' warning
when using :meth:`ClientResponse.text()` this may be because the response
does not include the charset needed to decode the body.
```

docs/client_advanced.rst
```diff
@@ -640,3 +640,33 @@ are changed so that aiohttp itself can wait on the underlying
...
+
+If you know the correct encoding for a request, you can simply specify
+the encoding as a parameter (e.g. ``resp.text("windows-1252")``).
+
+Alternatively, :class:`ClientSession` accepts a ``fallback_charset_resolver`` parameter which
+can be used to introduce charset guessing functionality. When a charset is not found
```
Dreamsorcerer:
```suggestion
can be used to reintroduce charset guessing functionality. When a charset is not found
```

docs/client_reference.rst
```diff
@@ -51,7 +51,8 @@ The client session supports the context manager protocol for self closing.
                          read_bufsize=2**16, \
                          requote_redirect_url=False, \
                          trust_env=False, \
-                         trace_configs=None)
+                         trace_configs=None, \
+                         fallback_charset_resolver=lambda r, b: "utf-8")
```
```suggestion
                         fallback_charset_resolver=_chardet_resolver)
```

Not really sure what we're supposed to put here..

docs/client_reference.rst
```diff
@@ -200,6 +201,16 @@ The client session supports the context manager protocol for self closing.
                          disabling.  See :ref:`aiohttp-client-tracing-reference` for
                          more information.
 
+   :param Callable[[ClientResponse,bytes],str] fallback_charset_resolver:
+      A :term:`callable` that accepts a :class:`ClientResponse` and the
+      :class:`bytes` contents, and returns a :class:`str` which will be used as
+      the encoding parameter to :meth:`bytes.decode()`.
+
+      This function will be called when the charset is not known (e.g. not specified in the
+      Content-Type header). The default function simply defaults to ``utf-8``.
```
```suggestion
      Content-Type header). The default function in 3.8.6 calls chardetng
      or charset-normaliser. In 3.9+ this be replaced with a function that
      simply defaults to ``utf-8``.
```

docs/client_reference.rst
```diff
@@ -1414,10 +1421,9 @@ Response object
                            ``None`` for encoding autodetection
                            (default).
 
-      :return str: decoded *BODY*
 
-      :raise LookupError: if the encoding detected by cchardet is
-                          unknown by Python (e.g. VISCII).
+      :raises: :exc:`UnicodeDecodeError` if decoding fails. See also
+               :meth:`get_encoding`.
```
```suggestion
```

docs/client_reference.rst
```diff
@@ -1430,18 +1436,14 @@ Response object
 
             await resp.text('ISO-8859-1')
 
-   .. comethod:: json(*, encoding=None, loads=json.loads, \
+   .. method:: json(*, encoding=None, loads=json.loads, \
                       content_type='application/json')
```
Dreamsorcerer, line 1440:
```suggestion
   .. method:: json(*, encoding=None, loads=json.loads, \
                      content_type='application/json')
      :async:
```

aiohttp/client.py
```diff
@@ -159,6 +164,22 @@ class ClientTimeout:
...
+def _default_fallback_charset_resolver(response: ClientResponse, body: bytes) -> str:
+
+    ret: str = chardet.detect(body)["encoding"] or "utf-8"
+
+    if ret != "utf-8":
+        warnings.warn(
+            "Automatic charset detection will be removed in 3.9, see: "
+            "https://docs.aiohttp.org/en/stable/client_advanced.html#character-set-detection",
+            DeprecationWarning,
+            stacklevel=2,
```
```suggestion
            stacklevel=3,
```

Dreamsorcerer:
OK, looks like it's ready. That should auto-merge, then we just need the final PR.

PR #7612, opened 2023-09-17
Resolve last 3.12 issues

cdce8p:
## What do these changes do?

Some last changes to be able to release `3.9` with support for Python 3.12.
* Ignore DeprecationWarning for datetime.utcnow

~* Add classifier for Python 3.11 and 3.12~
~* Remove experimental mark from 3.12 (ubuntu) config [ci]~

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,8 +36,6 @@ classifiers =
   Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
   Programming Language :: Python :: 3.10
-  Programming Language :: Python :: 3.11
-  Programming Language :: Python :: 3.12
 
   Topic :: Internet :: WWW/HTTP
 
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -36,6 +36,8 @@ classifiers =
   Programming Language :: Python :: 3.8
   Programming Language :: Python :: 3.9
   Programming Language :: Python :: 3.10
+  Programming Language :: Python :: 3.11
+  Programming Language :: Python :: 3.12
```
Dreamsorcerer: It was previously decided not to put these in, as the CI is still not passing on Windows/Mac, so 3.11+ support is still considered experimental.

cdce8p:
@Dreamsorcerer Do you have an idea when `3.9` will be ready? Others have already mentioned it on the `3.12` issue but a new release would really help to test both `aiohttp` and Python 3.12.

Issue #7381, opened 2023-07-18
Authentication header is removed with a legal redirect (port in host)


Alex-Izquierdo:
### Describe the bug

According to the documentation:
```
Authorization header will be removed if you get redirected to a different host or protocol.
```
But I've found a case where the header was removed without meet the criteria. 

### To Reproduce

Perform a GET request (with follow_redirects enabled) against an HTTP server that returns a 301 of the following form:

https://example.com:443/resource -> https://example.com/resource/

In this case, just the trailing slash was added and the port was removed. The library is considering the port as part of the host so it removes the Authentication header. 

### Expected behavior

Authentication header should not be removed because neither the host nor the protocol have really changed. 

### Logs/tracebacks

```python-traceback
_
```

### Python Version

```console
_
```

### aiohttp Version

```console
3.8.4
```

### multidict Version

```console
_
```

### yarl Version

```console
_
```

### OS

_

### Related component

Client

### Additional context

_

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
Hmm, that does seem a little off, given that it would actually be using the same port for both requests.
If you could open a PR with a test to reproduce the issue, that'd go a long way to getting it fixed (and if you'd like to have a go at fixing it too, even better).

Alex-Izquierdo:
Hello @Dreamsorcerer Thanks for your response. 
This sample script is enough to reproduce the problem. (you need to create the certs)  
You can see that without change the protocol or the host the Authorization header is not received by the server. 
```python
from aiohttp import web, ClientSession
import ssl
import asyncio

# port must be 443 to be cross compatible with host with/without port. It might require admin rights
# Also it can be served behind some http server like nginx
PORT = 443
HOST = "localhost"
URL_REQUEST = f"https://{HOST}:{PORT}/example"


routes = web.RouteTableDef()


@routes.get(r"/example{none:/?}")
async def example(request: web.Request):
    if str(request.rel_url).endswith("/"):
        return web.Response(text=f"Received headers: {request.headers}")
    raise web.HTTPMovedPermanently(f"{request.scheme}://{HOST}/example/")


async def server():
    print("RUNNING SERVER")
    app = web.Application()
    app.add_routes(routes)
    ssl_context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)

[...]

PR #7636, opened 2023-10-01
Authentication header removed although same origin host #7381

eyalhei:
Adding to is_same_host_https_redirect check case both origin and redirect are same host on https, as one might include port and the other not, or the different port

Also adding test_dont_drop_auth_on_redirect_to_same_host_https_scheme to check this

## What do these changes do?
change is_same_host_https_redirect to support if previous scheme was "https" - this way the port itself doesn't matter

Adding to is_same_host_https_redirect check case both origin and redirect are same host on https, as one might include port and the other not, or the different port

## Related issue number
#7381 

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -608,9 +608,7 @@ class ClientSession:
                             parsed_url = url.join(parsed_url)
 
                         is_same_host_https_redirect = (
-                            url.host == parsed_url.host
-                            and parsed_url.scheme == "https"
-                            and url.scheme in ("http", "https")
+                            url.host == parsed_url.host and parsed_url.scheme == "https"
                         )
 
                         if (
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -610,7 +610,7 @@ class ClientSession:
                         is_same_host_https_redirect = (
                             url.host == parsed_url.host
                             and parsed_url.scheme == "https"
-                            and url.scheme == "http"
+                            and url.scheme in ("http", "https")
                         )
 
                         if (
--- a/CHANGES/7636.bugfix
+++ b/CHANGES/7636.bugfix
@@ -1,0 +1,1 @@
+Authentication header shouldn't disappear in same host redirection
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2473,6 +2473,58 @@ async def test_drop_auth_on_redirect_to_other_host(
         assert resp.status == 200
 
 
+async def test_dont_drop_auth_on_redirect_to_same_host_https_scheme(
+    aiohttp_server: Any, tls_certificate_authority: Any
+) -> None:
+    app1 = web.Application()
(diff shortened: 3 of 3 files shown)


aiohttp/client.py
```diff
@@ -610,7 +610,7 @@ async def _request(
                         is_same_host_https_redirect = (
                             url.host == parsed_url.host
                             and parsed_url.scheme == "https"
-                            and url.scheme == "http"
+                            and url.scheme in ("http", "https")
                         )
 
                         if (
```
Dreamsorcerer, line 614: @greshilov @webknjaz Do you think this could maybe just be reduced to:
```
if url.host != parsed_url.host or parsed_url.scheme != "https":
```

i.e. Keep the Authorization header if it's the same host and the redirect is to https. Do we actually need to care whether it's the same port or not?

If we do care about the port, then maybe this needs to be fixed in yarl. We get 2 different origins for 'https://foo.com', and 'https://foo.com:443', despite both reporting the same port.

eyalhei: 1. the original bug mentioned a URL with port vs one without port, so we do care
2. Maybe it should be fixed in yarl, but that will probably constitute a breaking change, and will take time to pass through
3. I do agree on simplifying the statement

Dreamsorcerer: > the original bug mentioned a URL with port vs one without port, so we do care

Maybe you misread that. If we don't care, as per my suggested code, then it works regardless of the port number. If we do care, then we should make this code _more_ restrictive, not less, and the problem can only be solved in yarl at that point.

aiohttp/client.py
```diff
@@ -608,9 +608,7 @@ async def _request(
                             parsed_url = url.join(parsed_url)
 
                         is_same_host_https_redirect = (
-                            url.host == parsed_url.host
-                            and parsed_url.scheme == "https"
-                            and url.scheme == "http"
+                            url.host == parsed_url.host and parsed_url.scheme == "https"
```
Dreamsorcerer, line 611: My suggestion was on the if statement below. This variable can be completely removed if we go that way.

aiohttp/client.py
Dreamsorcerer: > Maybe it should be fixed in yarl, but that will probably constitute a breaking change, and will take time to pass through

@mjpieters Do you think a change like this would be acceptable in a minor release?

i.e. We want to make this test pass: `URL("https://foo.com").origin() == URL("https://foo.com:443").origin()`

Dreamsorcerer: > separating protection spaces by using a different host name (or port number) for each party.
https://www.rfc-editor.org/rfc/rfc9110.html#section-17.16.3-2

This suggests to me that we should care about the port number. In which case, it would be best to fix in yarl.
I guess http->https will (almost) always change the port number, so we probably can't be more restrictive in that specific case.

PR #7656, opened 2023-10-05
replace freezegun with time_machine

### LINKED ISSUE #7655: Migrate to time-machine
It looks like freezegun is considered an abandoned project at this point.
We should look at migrating the tests to time-machine instead:
https://pypi.org/project/time-machine/

See, for example: https://github.com/spulec/freezegun/pull/511

gruebel:
## What do these changes do?

- replaced the `freezegun` dependency with `time-machine`, because it seems like `freezegun` became an unmaintained project

## Related issue number

Fixes #7655 

## Checklist

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -206,7 +206,7 @@ sphinxcontrib-spelling==8.0.0 ; platform_system != "Windows"
     # via -r requirements/doc-spelling.in
 sphinxcontrib-towncrier==0.3.2a0
     # via -r requirements/doc.in
-time-machine==2.13.0
+time-machine==2.13.0 ; implementation_name == "cpython"
     # via -r requirements/test.in
 toml==0.10.2
     # via cherry-picker
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -67,8 +67,6 @@ exceptiongroup==1.1.2
     # via pytest
 filelock==3.12.2
     # via virtualenv
-freezegun==1.2.2
-    # via -r requirements/test.in
 frozenlist==1.4.0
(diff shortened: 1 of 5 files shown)


gruebel:
@Dreamsorcerer looks like `time-machine` doesn't support `pypy` 😢 Do you have a specific way on handling such cases or should it be an other library altogether?

Dreamsorcerer:
Ah, that's annoying. Not really sure what the best approach is...

Dreamsorcerer:
Although, it looks like it's only used in 3 tests, so maybe we can just skip the tests on pypy for now...

gruebel:
definitely a possibility, but I would also need to remove the dependency from the `test.in` file. Should I create a separate file, like `test_pypy.in` with the same constraint file as `test.in` and then I can do an if/else block in the workflow file to use the `pypy` flavour only, when `pypy` is tested?

Dreamsorcerer:
> definitely a possibility, but I would also need to remove the dependency from the `test.in` file.

No, look at the mypy line in that file. You can do the same kind of thing to exclude installation on pypy.

gruebel:
@Dreamsorcerer I think I added the needed changes 😅

gruebel:
@Dreamsorcerer if you have time to take a look again, it would be great 🙂

gruebel:
@Dreamsorcerer anything left to do here?

Dreamsorcerer:
Sorry, just busy with 3.9 release tasks.

Dreamsorcerer:
If you could handle that backport, that'd be great.

gruebel:
sure, I will give it a try

Issue #7078, opened 2022-11-10
Implement `WebSocketResponse.get_extra_info`


xdustinface:
### Is your feature request related to a problem?

There is no way to access `get_extra_info` from a `WebSocketResponse`, like its possible in `ClientWebSocketResponse`, there was a related issue before this was implemented: https://github.com/aio-libs/aiohttp/issues/1717.

### Describe the solution you'd like

Im thinking about just having `WebSocketResponse.get_extra_info` implemented like `ClientWebSocketResponse.get_extra_info`.

### Describe alternatives you've considered

As workaround we just access the private member like in 

```python
response._writer.transport.get_extra_info("the info")
```

### Related component

Server

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
#7658

Dreamsorcerer:
#7658

PR #7658, opened 2023-10-06
Implement WebSocketResponse.get_extra_info method to avoid using private member

ttsia:
## What do these changes do?

This change introduces new method to access get_extra_info from a WebSocketResponse, like its possible in ClientWebSocketResponse to avoid accessing the private member like in

`response._writer.transport.get_extra_info("the info")`

## Related issue number

https://github.com/aio-libs/aiohttp/issues/7078

## Checklist

### DIFF
--- a/CHANGES/7078.feature
+++ b/CHANGES/7078.feature
@@ -1,0 +1,1 @@
+Implement ``WebSocketResponse.get_extra_info`` method to avoid accessing the private member ``_writer``
--- a/CHANGES/7078.feature
+++ b/CHANGES/7078.feature
@@ -1,1 +1,1 @@
-Implement ``WebSocketResponse.get_extra_info`` method to avoid accessing the private member ``_writer``
+Added ``WebSocketResponse.get_extra_info()`` to access a protocol transport's extra info.
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -324,7 +324,10 @@ class WebSocketResponse(StreamResponse):
         return self._compress
 
     def get_extra_info(self, name: str, default: Any = None) -> Any:
-        """extra info from writer transport"""
+        """Get optional transport information.
+
+        If no value associated with ``name`` is found, ``default`` is returned.
+        """
         writer = self._writer
         if writer is None:
             return default
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -323,6 +323,16 @@ class WebSocketResponse(StreamResponse):
     def compress(self) -> bool:
         return self._compress
 
+    def get_extra_info(self, name: str, default: Any = None) -> Any:
+        """extra info from writer transport"""
+        writer = self._writer
+        if writer is None:
+            return default
+        transport = writer.transport
+        if transport is None:
+            return default
+        return transport.get_extra_info(name, default)
+
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -1049,6 +1049,10 @@ and :ref:`aiohttp-web-signals` handlers::
       May be ``None`` if server and client protocols are
       not overlapping.
 
+   .. method:: get_extra_info(name, default=None)
+
+      Reads extra info from writer's transport
+
    .. method:: exception()
 
       Returns last occurred exception or None.
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -379,6 +379,8 @@ and :ref:`aiohttp-web-signals` handlers.
       Reads extra information from the protocol's transport.
       If no value associated with ``name`` is found, ``default`` is returned.
 
+      See :meth:`asyncio.BaseTransport.get_extra_info`
+
       :param str name: The key to look up in the transport extra information.
 
       :param default: Default value to be used when no value for ``name`` is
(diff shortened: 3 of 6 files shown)


docs/web_reference.rst
```diff
@@ -1049,6 +1049,10 @@ and :ref:`aiohttp-web-signals` handlers::
       May be ``None`` if server and client protocols are
       not overlapping.
 
+   .. method:: get_extra_info(name, default=None)
+
+      Reads extra info from writer's transport
```
webknjaz: It'd be useful to see more details here. Like, what's that info? Why is it extra? What's non-extra? What's the format? Where's the data coming from? What is it related to?

Saying that it's some arbitrary data coming from the transport layer is not really helpful and is too generic.

Oh, and I see a missing trailing period that should be used to terminate sentences.

aiohttp/web_ws.py
```diff
@@ -323,6 +323,16 @@ def ws_protocol(self) -> Optional[str]:
     def compress(self) -> bool:
         return self._compress
 
+    def get_extra_info(self, name: str, default: Any = None) -> Any:
+        """extra info from writer transport"""
```
webknjaz: Per PEP 257, the docstring title is to be a sentence. Hence, it should have a trailing period and start with a capital letter. There are also semantic recommendations, like starting with a verb for functions/methods/callables — so that it describes what they do as opposed to what they are, which constants, vars, modules, packages, classes typically require.

Besides, such a generic phrase is not helpful. A docstring should accurately set the expectations. It is a public interface to the method usage, after all, and it shows up in REPL and editors.

tests/test_web_websocket.py
```diff
@@ -412,3 +412,33 @@ async def test_no_transfer_encoding_header(make_request: Any, mocker: Any) -> No
...
+    transp_extra_info = ws._writer.transport.get_extra_info(valid_key, default_value)
+    assert ws_extra_info == transp_extra_info
+
+    ws._writer.transport = None
+    extra_info = ws.get_extra_info(valid_key, default_value)
+    assert extra_info == default_value
+
+    ws._writer = None
+    extra_info = ws.get_extra_info(valid_key, default_value)
+    assert extra_info == default_value
```
webknjaz: Looks like you included 3 separate tests into the same test function. This is not a good idea as they wouldn't show up in the test report. Also, if one assert fails, the following ones won't be checked, meaning that it's impossible to know if other cases pass, from the test report output.

To fix this, use parametrize. You can also use param ids to label individual tests for human readability.

aiohttp/web_ws.py
```diff
@@ -323,6 +323,16 @@ def ws_protocol(self) -> Optional[str]:
     def compress(self) -> bool:
         return self._compress
 
+    def get_extra_info(self, name: str, default: Any = None) -> Any:
```
webknjaz, line 326: Can the ambiguous "extra info" be surfaced as something more specific? It seems weird to me, putting this up as a public API.

CHANGES/7078.feature
```diff
@@ -0,0 +1 @@
+Implement ``WebSocketResponse.get_extra_info`` method to avoid accessing the private member ``_writer``
```
webknjaz: Change notes should normally use the past tense as their audience is the end-users seeking info on the differences from the previous version. This also means that they don't need to see any internal implementation details, unlike devs/contributors that would be using Git commits for context.
A trailing period wouldn't hurt either.

aiohttp/web_ws.py
ttsia: Good point, but I decided to be consistent with already existing https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientWebSocketResponse.get_extra_info

webknjaz: Fair enough. Marking this thread as resolved.

docs/web_reference.rst
Dreamsorcerer: Well, this is just proxying the call to asyncio, right?
So, both this one and the original it was copied from should be changed to reference `asyncio.BaseTransport.get_extra_info()`. Then users can just look at the Python documentation for details.

Dreamsorcerer: Maybe just:

```suggestion
      Return information about the transport or underlying resources it uses.
      
      See :meth:`asyncio.BaseTransport.get_extra_info`
```

ttsia: I've added reference to python docs just in case

Dreamsorcerer: We use Intersphinx, so my suggestion below is a link to the docs.

ttsia: resolved

aiohttp/web_ws.py
ttsia: fixed

tests/test_web_websocket.py
ttsia: fixed

CHANGES/7078.feature
ttsia: fixed

docs/web_reference.rst
```diff
@@ -1049,6 +1049,19 @@ and :ref:`aiohttp-web-signals` handlers::
       May be ``None`` if server and client protocols are
       not overlapping.
 
+   .. method:: get_extra_info(name, default=None)
```
Dreamsorcerer, line 1054: While we're here, can we update the other docs, where you originally copied this from?

Dreamsorcerer: Appears to be further up in BaseRequest as well as in client_reference.rst,

ttsia: python docs reference has been included for all _get_extra_info_ implementations

PR #7673, opened 2023-10-07
Document release process

Dreamsorcerer:
@webknjaz We really need to document this somewhere. Is this an acceptable place?


webknjaz:
Best put it under `docs/` and include somewhere via MyST.

Dreamsorcerer:
> Best put it under `docs/` and include somewhere via MyST.

We don't really want it on the main website, right? It's only relevant to aiohttp-admins, so being accessible only in the codebase makes more sense to me.

webknjaz:
Honestly, I've seen people having this in a separate docs section, somewhere close to the contributing guides so I typically prefer it. It's not something secret. We document how to add changelog fragments there too, which is kinda related.

webknjaz:
I also know that some people dislike polluting the repo root with many files, which is sometimes a source of complaints..

docs/contributing-admins.rst
```diff
@@ -0,0 +1,53 @@
...
+     the target branch (as tests have passed on the merge commit now).
+  #. This should automatically consider the PR merged and delete the temporary branch.
+
+Back on the original release branch, append ``.dev0`` to the version number in ``__init__.py``.
+
+If doing a minor release:
+
+#. Create a new release branch for future features to go to: e.g. ``git checkout -b 3.10 3.9 && git push``
+#. Delete the older backport label (e.g. backport-3.8): https://github.com/aio-libs/aiohttp/labels
+#. Add a new backport label (e.g. backport-3.10).
```
Dreamsorcerer: Anything else we need to do, e.g. to configure patchback-bot?

webknjaz: `patchback` uses existing labels but doesn't have an ability to create new ones, atm. Could be something to request @ https://github.com/sanitizers/patchback-github-app/issues/new. Ideally, the app should auto-create branches on `git push` (`created` action of the `push` event @ GH platform) and on GitHub App `repository_installation` or `unsuspend` events.

For now, yes, creating backport labels manually is enough. I wonder if we need another FR to teach the bot to auto-create a new stable branch if a label got created for a non-existing branch…

docs/contributing-admins.rst
```diff
@@ -0,0 +1,53 @@
+:orphan:
+
```
webknjaz, line 2: Can we include this into the ToC instead of orphaning? Doesn't linking it from the other document not make this automatically?

```suggestion
```

For inspiration, there's some release sections in public docs:

https://pip.pypa.io/en/stable/development/release-process/#id5 / https://setuptools.pypa.io/en/latest/development/releases.html#release-process / https://ansible-pylibssh.readthedocs.io/en/latest/contributing/release_guide/#release-guide / https://scikit-hep.org/pyhf/development.html#publishing / https://devguide.python.org/developer-workflow/development-cycle/

docs/contributing-admins.rst
```diff
@@ -0,0 +1,53 @@
...
+   :local:
+
+Creating a new release
+----------------------
+
+To create a new release:
+
+#. Start on the branch for the release you are planning (e.g. ``3.8`` for v3.8.6): ``git checkout 3.8 && git pull``
+#. Update the version number in ``__init__.py``.
+#. Run ``towncrier``.
```
webknjaz, line 21: No args?

docs/contributing-admins.rst
```diff
@@ -0,0 +1,53 @@
...
+#. Check and cleanup the changes in ``CHANGES.rst``.
+#. Checkout a new branch: e.g. ``git checkout -b release/v3.8.6``
+#. Commit and create a PR. Once PR is merged, continue.
+#. Go back to the release branch: e.g. ``git checkout 3.8 && git pull``
+#. Add a tag: e.g. ``git tag -a v3.8.6 -m 'Release 3.8.6'``
+#. Push the tag: e.g. ``git push origin v3.8.6``
```
webknjaz, line 27: If this assumes that `origin` is the upstream repo, we should document it. Some people use `fork` for their forks and `origin` for upstream, while others use `upstream` for the upstream and `origin` for their fork. Also, there's people using their account name for their fork (as in `username` for `https://github.com/username/aiohttp.git`).

It's a good idea to clearly declare such expectations.

Here's an example of a such doc: https://docs.ansible.com/ansible-core/devel/community/development_process.html#backporting-merged-prs-in-ansible-core. Here's one project using the `upstream` convention: https://github.com/python-attrs/attrs/blob/bd443b0/.github/CONTRIBUTING.md#local-development-environment.

docs/contributing-admins.rst
```diff
@@ -0,0 +1,53 @@
...
+#. Monitor CI to ensure release process completes without errors.
+
+Once released, we need to complete some cleanup steps (no further steps are needed for
+non-stable releases though). If doing a patch release, we need to do the below steps twice,
+first merge into the newer release branch (e.g. 3.8 into 3.9) and then to master
+(e.g. 3.9 into master). If a new minor release, then just merge to master.
+
+#. Switch to target branch: e.g. ``git checkout 3.9 && git pull``
+#. Start a merge: e.g. ``git merge 3.8 --no-commit``
```
webknjaz: I usually do `--no-ff` and `--gpg-sign`.

webknjaz (approved):
@Dreamsorcerer I left some notes. They are not mandatory and can be fixed in follow-ups. That's up to you.

docs/contributing-admins.rst
Dreamsorcerer: Apparently, the meaning of orphan is just that it is not included it a toctree, not that it is undiscoverable (something which is confusing, as most users expect orphan to mean the latter)...

I'm not really sure we want to include it in a toctree? The sidebar has a link directly to contributing, I'm not sure it makes sense to include another link from the sidebar for something almost nobody needs to see. And I don't really see any other toctrees to use.

docs/contributing-admins.rst
Dreamsorcerer: I didn't use any. Do you have any arguments when you run it? I was just guessing.

Dreamsorcerer:
@webknjaz Would it be fine to delete some of the oldest 3.x branches? Not sure if there's a reason they've been kept around, but they just seem so old now that they are not useful, and it's a bit of a nuisance that I now have to scroll to select the correct branch in a backport PR (all upto 3.8 are selectable without scrolling).

docs/contributing-admins.rst
webknjaz: @Dreamsorcerer I think I was testing with `--draft` and always passing `--version`. Though, I suppose Towncrier has some ability to import the version from a Python module. That doesn't always work, not for C-extension based codebases, anyway.

webknjaz:
> @webknjaz Would it be fine to delete some of the oldest 3.x branches? Not sure if there's a reason they've been kept around, but they just seem so old now that they are not useful, and it's a bit of a nuisance that I now have to scroll to select the correct branch in a backport PR (all upto 3.8 are selectable without scrolling).

@Dreamsorcerer I think it's fine for as long as there are tags on the tops of those branches.

docs/contributing-admins.rst
webknjaz: @Dreamsorcerer in other projects, I started making it a separate TOC, see:
* https://ansible-pylibssh.readthedocs.io/en/latest/contributing/release_guide/
* https://cheroot.cherrypy.dev/en/latest/devguide/
* https://yarl.aio-libs.org/en/latest/contributing/release_guide/

I'd like to implement a similar structure here, eventually.

docs/contributing-admins.rst
```diff
@@ -0,0 +1,57 @@
+:orphan:
```
webknjaz, line 1: @Dreamsorcerer also, since you put a link to this doc from another doc, Sphinx won't consider it orphan anymore. So it's not really needed either way.

docs/contributing-admins.rst
Dreamsorcerer: Right, I was wondering where it kept getting the version from. I just corrected it in the file manually.

docs/contributing-admins.rst
Dreamsorcerer: I mentioned above, that's not what orphan means to Sphinx. It literally just means that it is not in a toctree, not that it is unreachable. So, it is needed in its current state.

PR #7676, opened 2023-10-07
Update FUNDING.yml

Dreamsorcerer:
Am I good to add this?

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -3,6 +3,7 @@
 github: # Replace with up to 4 GitHub Sponsors-enabled usernames e.g., [user1, user2]
 - asvetlov
 - webknjaz
+- Dreamsorcerer
 # patreon: # Replace with a single Patreon username
 # open_collective: # Replace with a single Open Collective username
 # ko_fi: # Replace with a single Ko-fi username


webknjaz (approved):
Absolutely! Should've done this long ago!

By the way, try registering @ https://thanks.dev — it's some new platform that Sentry uses to distribute funding. There been some weird changes in how they do it so maybe this'll get you something too (I haven't been able to figure out why they donated to me last year and reduced it this year, nor what projects they're using at all..)

webknjaz:
Ah, I only now realized that I've been trying to keep the list sorted. But it's too late so maybe we'll fix that next time.

Dreamsorcerer:
> By the way, try registering @ https://thanks.dev

Not sure what I'm doing wrong, but I just get an empty dashboard (2 days after signing up).

image

webknjaz:
Same for me. I was hoping to track down what deps they see per request here https://twitter.com/chadwhitacre_/status/1710318405398470942, but haven't gotten anything so far.

Dreamsorcerer:
The home page still looks like that, but it's finally updated and lists potential dependents now. I guess that means the project is atleast registered now and should be available for sponsors in the next round (I notice in the sponsor screen, half the projects are greyed out, probably because they aren't registered to accept payouts).

webknjaz:
Yeah, I guess. They didn't get into much detail on Twitter. I also got my scanning finished yesterday..

webknjaz:
I also don't understand if they want multiple separate users with own payout bank accounts or one account per project/org. I know that Tidelift does an account per person and spreads money equally or per arranged percentage..

Dreamsorcerer:
Yeah, there is a button to book a call to talk about funding. Maybe worth doing that to ask them questions?

Dreamsorcerer:
> Yeah, there is a button to book a call to talk about funding. Maybe worth doing that to ask them questions?

Oh, I just looked at the calendar times, and their hours are about 10pm till 9am in my timezone, which is a bit awkward.

webknjaz:
I've just submitted like 4 of their feedback forms with stuff. Hopefully, they'll end up putting the answers into their FAQ...

Dreamsorcerer:
Seems to be a few donations coming in now, not sure if you've figured out the details yet. I see payments being made to gh/aio-libs, but then it doesn't seem to have separate balances for gh/aio-libs and gh/dreamsorcerer or anything, which makes me think that the balance is in my account, but really not sure...

Also, I think you filled out the profile information under gh/aio-libs with your details. :P
There's a drop-down to switch between the different profiles.

webknjaz:
> Also, I think you filled out the profile information under gh/aio-libs with your details. :P
> There's a drop-down to switch between the different profiles.

Yeah, by accident and never fixed it..

webknjaz:
> I see payments being made to gh/aio-libs, but then it doesn't seem to have separate balances for gh/aio-libs and gh/dreamsorcerer or anything, which makes me think that the balance is in my account, but really not sure...

Oh, I got a reply by email and thought I shared it, but that was in some other conversation..
All the admins have access to the org account which is shared. It is possible to forward some of it into personal accounts. I also sent them a bug report about GH Sponsors UI bugs which they fixed as a result.. Here's that email:

Dreamsorcerer:
Ah, OK, se we do still need to split it then.

Still a little confusing in the UI, as there's no dropdown or anything for the balance under the payouts page. So, if donations come through another account, it will surely be confusing to tell how much is in each account.

webknjaz:
When you click on transfer or re-donate, it lets you select one of the balances. That's probably the only place in the UI where you see what's fully yours. And I also gave them the confusion feedback. They are pretty responsive so I encourage you to write them too so they understand the scale of the UX impact.

webknjaz:
@Dreamsorcerer did you perform any transfers on thanks.dev on Nov 15? The UI has no way of tracking who performs what actions...

thanks-dev-nov-15-2023-payout

Dreamsorcerer:
image

Dreamsorcerer:
So, no, not me.

webknjaz:
@Dreamsorcerer yeah, I saw the target, but I wasn't sure if that entry displays an actor. I think this is a transfer from the org account to a separate user account. Probably, Andrew did it, then. But I haven't talked to him in a while.

webknjaz:
It has "to gh/<username>", and that "to" suggests the direction. I think anybody can forward money to other entities, just no idea how it'd look.

Dreamsorcerer:
I was just going to leave it you to figure out what to do with it. I imagine he hasn't realised yet that it's a shared pool.

webknjaz:
Maybe. I've been postponing this myself. And I think half a dozen org members have similar access too, by the way.

We probably need to implement this as a stopgap: https://github.com/typescript-eslint/.thanks.dev/blob/main/THANKS.yaml. It's an example they sent me via email.

Dreamsorcerer:
Yeah, looks reasonable.

webknjaz:
Looks like there's a per-repo config too: https://news.ycombinator.com/item?id=35521396. Although, that post suggests it's lower-cased. There's no docs or other resources on the internet related to this.

webknjaz:
@Dreamsorcerer I also want to get more maintainers grab money from Tidelift: https://github.com/orgs/aio-libs/discussions/33.

PR #7677, opened 2023-10-08
Fix warning displayed by default

Dreamsorcerer:


aiohttp/web_exceptions.py
```diff
@@ -71,6 +71,10 @@
 )
 
 
+class NotAppKeyWarning(UserWarning):
```
webknjaz, line 74: @Dreamsorcerer so I would probably try moving warnings into a separate importable package that does not have import-time side effects.

The problem with having it here is that `from 
aiohttp.web_exceptions import NotAppKeyWarning` would trigger importing a number of modules under the package. This effectively means that whenever this warning starts being used anywhere in one of the modules or aiohttp's `__init__`, it will be impossible for the end-users to suppress/filter out such warning by its name using `-W` or the corresponding env var because Python would attempt importing that exception which would trigger importing the top-level package and everything it imports before loading the warning completes. This causes a chicken-egg type of issue.

I documented this problem @ https://cryptography.io/en/3.4/faq.html#i-cannot-suppress-the-deprecation-warning-that-cryptography-emits-on-import at some point.

Basically, this is why the WPS linter has a rule forbidding to have anything inside any of the `__init__.py` files. And I've adopted a practice of making a non-init files entrypoints (going for `api.py` or [`ext.py`](https://github.com/sphinx-contrib/sphinxcontrib-towncrier/blob/729b600/src/sphinxcontrib/towncrier/ext.py#L328) instead).

One possibility is making a new `aiohttp_warnings/` package in the repo root, and packaing into the same dist. So both `aiohttp` and `aiohttp_warnings` importables would be available. Then, we'd make something like `aiohttp_warnings.types` for the warning hierarchy and prevent any use of `__init__.py` there (maybe, with the `import-linter` tool?).

Dreamsorcerer: I'll try rereading this tomorrow to try and understand this. But, I can say that I can filter it out at runtime with `warnings.filterwarnings("ignore", category=web.NotAppKeyWarning)`. And, `-W` won't work generally, because it tries to import `aiohttp` _before_ Python has finished setting up the `sys.path` or something like that (there's an old bug report about it), so it's not importable (I think you can get it to work, by adding `aiohttp` directly to `PYTHONPATH`, but that's a pretty useless solution).

The cryptography one appears to be about emitting the warning on import, we're just defining it and adding an ignore filter. It's not emitted until much later.

Dreamsorcerer: > it will be impossible for the end-users to suppress/filter out such warning by its name using `-W` or the corresponding env var

OK, I've reread this, but I still can't see how there would be an issue unless we actually emit a warning on import.

Even if there was such an issue, I don't think it'd apply here, as `__init__.py` will never import any of the `web` modules.

Dreamsorcerer: @webknjaz Any last concerns? Going to need to merge this to unblock the 3.9 release.
Also, if you have any chance to give a brief review of #7718, that is the other blocker.
I'd like to merge and do a second beta release by the weekend.

webknjaz: Well, I've been thinking of a more general pattern. Yes, we don't import `web`, but it's not impossible that other import paths may start causing import-time warnings in the future if we start using this technique. And that won't work with the normal Python's import suppression mechanism through CLI/env vars.
That's my only concern right now. It's probably fine to ignore it now, we just need to remember this in the future.

Dreamsorcerer: OK, but there's still no import-time warnings. We're only modifying the filters here.

webknjaz: Totally. But if we're going to use warnings in the future, we'll probably put the classes into the same location. And we may then end up using those for import-time deprecations. At some point in the future. That's when the problem might manifest itself.

webknjaz:
@Dreamsorcerer you don't have to merge in master anymore. The problem is with the checks not being reported.

PR #7679, opened 2023-10-08
Change headers to a dict that parses comma-separated values

Dreamsorcerer:
This is a proposal to change the headers from a CIMultiDict to a more regular dict (in v4). The problem with the multidict approach is that list headers (i.e. headers that can have multiple values) can have values combined in single headers and/or split over multiple headers.

----

Basically, these 2 payloads should be considered equivalent:
```
Foo: 1
Foo: 2
```
```
Foo: 1, 2
```

But, currently aiohttp will produce in the first case:
```
headers["Foo"]  # "1"
headers.getall("Foo")  # ["1", "2"]
```
And, in the second case:
```
headers["Foo"]  # "1, 2"
headers.getall("Foo")  # ["1, 2"]
```

-----

The spec recommends concatenating duplicate headers together with `", "`. This is also what the vast majority of existing software does (including requests).

The only problem with this, is that if the user wants the values as a list, they are left to parse the value themselves, which when accounting for quoted values becomes quite complex and easy to get wrong on edge cases. So, in this proposal I've concatenated the values as recommended, but added a `.getall()` method which parses the final value to get the list.

With the code in this PR, both of the previous payloads produce the same output:
```
headers["Foo"]  # "1, 2"
headers.getall("Foo")  # ["1", "2"]
```

------

>    "Field value" now refers to the value after multiple field lines are combined with commas -- by far the most common use.
From [RFC 9110 appendix B2](https://www.rfc-editor.org/rfc/rfc9110#appendix-B.2)

From kenballus's testing:
> Servers that join the duplicate headers by default: Apache httpd, Caddy, Gunicorn, H2O, IIS, Lighttpd, Nginx, Node.js, Puma.
> Servers that accept the duplicate headers without joining them: aiohttp (currently), Boost::Beast, Mongoose, Tornado

I've also seen that requests combines into a regular dict. The only other library I've seen that uses a multidict for this is Starlette.

Also: https://www.rfc-editor.org/rfc/rfc9110.html#name-recipient-requirements


steverep:
Some questions to consider here:

1. Should the `getAll` list be converted to lowercase when the values are case-insensitive (which is true for most if not all I think)?
2. If yes, should the list then also be de-duplicated?
3. If yes, should content negotiation headers that can have [quality values per RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#quality.values) be parsed and assigned, e.g. by returning a dictionary instead of a list or tuple as `{"<value>": <quality>, ...}`?

I think all of these being done directly in the parsers is best for performance, and would make for easier and less error-prone usage.

aiohttp/http_parser.py
```diff
@@ -65,9 +67,56 @@
 METHRE: Final[Pattern[str]] = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
 VERSRE: Final[Pattern[str]] = re.compile(r"HTTP/(\d).(\d)")
 HDRRE: Final[Pattern[bytes]] = re.compile(rb"[\x00-\x1F\x7F()<>@,;:\[\]={} \t\"\\]")
+QUOTEHDRRE = re.compile(r'(".*?(?:[^\\]"))[ \t]*(?:,|$)')
 HEXDIGIT = re.compile(rb"[0-9a-fA-F]+")
 
 
+class HeadersDictProxy(Mapping[str, str]):
```
webknjaz: Have you compared performance compared to inheriting multidict and overriding the behaviors of storing the data and outputting the combined values?

Dreamsorcerer: Implementation details are not a concern yet, so will come back to that. First, I want to get consensus that this is the correct approach and should be changed in v4, given that it is likely to cause backwards-compatibility breakages for atleast a small proportion of users.

Dreamsorcerer:
> Some questions to consider here:

If you've got any information in the specs to answer those questions, that'd be great to have.

steverep:
> If you've got any information in the specs to answer those questions, that'd be great to have.

Just by their nature, I think it's certainly safe to deduplicate the content negotiation fields defined in [Section 12.5 of RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-negotiation-fields).  However, AFAICT, the RFC has no guidance on what quality value to assign if the duplicates happen to disagree.  Seems like server's choice in that edge case would be conformant.

Other list headers should not be deduplicated because duplicates can actually mean something.  For example, `Content-Encoding: "gzip, gzip"` means the content was double compressed with gzip.

Dreamsorcerer:
OK, now having time to think over this, I think it's all a level above what we should be doing here. I think we're just providing a list based on the definition of a general HTTP field. So, I feel the answer is no to all 3 questions. Maintaining logic for all the different kind of headers seems out of scope to me (and if we did, it'd be through dedicated attributes, like cookies).

steverep:
> OK, now having time to think over this, I think it's all a level above what we should be doing here. I think we're just providing a list based on the definition of a general HTTP field. So, I feel the answer is no to all 3 questions. Maintaining logic for all the different kind of headers seems out of scope to me (and if we did, it'd be through dedicated attributes, like cookies).

After reading the spec a bit more, I guess I agree with you on the first two, but parameters are actually generically defined in [section 5.6.6](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.6).  The syntax and case-insensitivity of parameter names is defined there (the content negotiation headers just happen to use "q" as the name).  I think the parsers should provide a way to access them as a dictionary (maybe by returning a list subclass?).

tests/test_http_parser.py
```diff
@@ -167,6 +167,41 @@ def test_cve_2023_37276(parser: Any) -> None:
...
+        ),
+        (
+            ('"Sat, 04 May 1996", "Wed, 14 Sep 2005"',),
+            ("Sat, 04 May 1996", "Wed, 14 Sep 2005"),
+        ),
+        (("foo,bar,baz",), ("foo", "bar", "baz")),
+        (('"applebanna, this',), ('"applebanna', "this")),
+        (('fooo", "bar"',), ('fooo"', "bar")),
+        ((" spam , eggs ",), ("spam", "eggs")),
+        ((" spam ", " eggs "), ("spam", "eggs")),
```
Dreamsorcerer, line 287: TODO: Add tests for escaped quotes (e.g. `"foo\"bar"`), maybe also escaped backslash, if that's valid (e.g. `"foo\\"` or `"foo\\\""`).

aiohttp/_http_parser.pyx
```diff
@@ -393,6 +394,7 @@ cdef class HttpParser:
         self._limit = limit
 
     cdef _process_header(self):
+
```
Dreamsorcerer:
```suggestion
```

aiohttp/http_parser.py
```diff
@@ -73,9 +74,56 @@
...
+    def __len__(self):
+        return self._d.__len__()
+
+    def has_key(self, key: str):
+        return self._d.has_key(key.title())
+
+    def get(self, key: str, default=None):
+        return self._d.get(key.title(), default)
+
+    def getall(self, key: str) -> Tuple[str]:
```
```suggestion
    def getall(self, key: str) -> tuple[str]:
```

aiohttp/http_parser.py
```diff
@@ -73,9 +74,56 @@
...
+        return self._split_on_commas(self._d.get(key, ""))
+
+    def _split_on_commas(self, val: str) -> Tuple[str]:
```
```suggestion
    def _split_on_commas(self, val: str) -> tuple[str]:
```

tests/test_http_parser.py
```diff
@@ -218,6 +218,41 @@ def test_bad_header_name(
...
+        ((",   , ",), ()),
+        (("",), ()),
+    ),
+)
+def test_list_headers(parser: Any, hdr_vals: List[str], expected: List[str]) -> None:
```
```suggestion
def test_list_headers(parser: HttpRequestParser, hdr_vals: tuple[str], expected: tuple[str]) -> None:
```

Dreamsorcerer:
As nobody has given any reason not to do this, I think we can continue with finishing this PR now.

aiohttp/client_reqrep.py
```diff
@@ -461,14 +455,14 @@ async def start(self, connection: "Connection") -> "ClientResponse":
...
-        self._raw_headers = message.raw_headers  # type is Tuple[bytes, bytes]
+        self._headers = message.headers
+        self._raw_headers = message.raw_headers
 
         # payload
         self.content = payload
 
         # cookies
-        if cookie_hdrs := self.headers.getall(hdrs.SET_COOKIE, ()):
+        if cookie_hdrs := (v for k, v in self.raw_headers if k.title() == hdrs.SET_COOKIE):
```
Dreamsorcerer: This is inefficient. To improve this, we probably need a multidict back. My thinking is that it might make sense to change raw_headers from a tuple to a CIMultiDict. However, CIMultiDict only allows str keys, so we'd need to update multidict to allow bytes.

Although the cookie_hdrs here then need to be decoded back to str....

aiohttp/client_reqrep.py
```diff
@@ -461,16 +456,18 @@ async def start(self, connection: "Connection") -> "ClientResponse":
...
+        self._headers = message.headers
+        self._raw_headers = message.raw_headers
 
         # payload
         self.content = payload
 
         # cookies
-        if cookie_hdrs := self.headers.getall(hdrs.SET_COOKIE, ()):
+        if cookie_hdrs := tuple(
+            v for k, v in self.raw_headers if k.title() == hdrs.SET_COOKIE.encode()
```
```suggestion
            v.decode() for k, v in self.raw_headers if k.title() == hdrs.SET_COOKIE.encode()
```

aiohttp/http_parser.py
```diff
@@ -73,14 +76,60 @@
 )
 VERSRE: Final[Pattern[str]] = re.compile(r"HTTP/(\d)\.(\d)", re.ASCII)
 DIGITS: Final[Pattern[str]] = re.compile(r"\d+", re.ASCII)
+QUOTEHDRRE = re.compile(r'(".*?(?:[^\\]"))[ \t]*(?:,|$)')
 HEXDIGITS: Final[Pattern[bytes]] = re.compile(rb"[0-9a-fA-F]+")
 
 
+class HeadersDictProxy(Mapping[str, str]):
+    def __init__(self, d: Mapping[str, str]):
+        self._d = d
```
```suggestion
        self._d = {k.title(): v for k, v in d}
```

aiohttp/http_parser.py
```diff
@@ -73,14 +76,60 @@
 )
 VERSRE: Final[Pattern[str]] = re.compile(r"HTTP/(\d)\.(\d)", re.ASCII)
 DIGITS: Final[Pattern[str]] = re.compile(r"\d+", re.ASCII)
+QUOTEHDRRE = re.compile(r'(".*?(?:[^\\]"))[ \t]*(?:,|$)')
 HEXDIGITS: Final[Pattern[bytes]] = re.compile(rb"[0-9a-fA-F]+")
 
 
+class HeadersDictProxy(Mapping[str, str]):
+    def __init__(self, d: Mapping[str, str]):
+        self._d = {k.title(): v for k, v in d}
```
```suggestion
        self._d = {k.title(): v for k, v in d.items()}
```

aiohttp/multipart.py
```diff
@@ -755,7 +755,7 @@
 
     def _get_part_reader(
         self,
-        headers: "CIMultiDictProxy[str]",
+        headers: HeadersDictProxy",
```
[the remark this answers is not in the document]
Dreamsorcerer:
```suggestion
        headers: HeadersDictProxy,
```

tests/test_benchmarks_web_urldispatcher.py
```diff
@@ -38,7 +38,7 @@ def _mock_request(method: str, path: str) -> web.Request:
         method,
         path,
         HttpVersion(1, 1),
-        CIMultiDictProxy(CIMultiDict()),
+        HeadersDictProxy({}),
```
Dreamsorcerer:
```suggestion
        HeadersDictProxy(CIMultiDict()),
```

tests/test_web_request.py
```diff
@@ -34,7 +35,7 @@ def test_base_ctor() -> None:
         "GET",
         "/path/to?a=1&b=2",
         HttpVersion(1, 1),
-        CIMultiDictProxy(CIMultiDict()),
+        HeadersDictProxy({}),
```
Dreamsorcerer:
```suggestion
        HeadersDictProxy(CIMultiDict()),
```

aiohttp/http_parser.py
```diff
@@ -123,10 +178,8 @@ def __init__(self, max_field_size: int = 8190, lax: bool = False) -> None:
         self.max_field_size = max_field_size
         self._lax = lax
 
-    def parse_headers(
-        self, lines: list[bytes]
-    ) -> tuple["CIMultiDictProxy[str]", RawHeaders]:
-        headers: CIMultiDict[str] = CIMultiDict()
+    def parse_headers(self, lines: list[bytes]) -> tuple[HeadersDictProxy, RawHeaders]:
+        headers: CIMultiDict[str] = {}
```
```suggestion
        headers: CIMultiDict[str] = CIMultiDict()
```

aiohttp/_http_parser.pyx
```diff
@@ -442,11 +443,11 @@ cdef class HttpParser:
...
 
         raw_headers = tuple(self._raw_headers)
-        headers = CIMultiDictProxy(CIMultiDict(self._headers))
+        headers = HeadersDictProxy(CIMultiDict(self._headers))
 
         # https://www.rfc-editor.org/rfc/rfc9110.html#name-collected-abnf
         bad_hdr = next(
-            (h for h in SINGLETON_HEADERS if len(headers.getall(h, ())) > 1),
+            (h for h in SINGLETON_HEADERS if len(headers._md.getall(h, ())) > 1),
             None,
```
bdraco: seems like this would be better to do at parse time

bdraco: might be able to avoid the perf regression with something like
```diff
diff --git a/aiohttp/_http_parser.pyx b/aiohttp/_http_parser.pyx
index f22073da6..f352f9b9f 100644
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -312,6 +312,7 @@ cdef class HttpParser:
         str     _path
         str     _reason
         object  _headers
+        set     _seen_singletons
         list    _raw_headers
         bint    _upgraded
         list    _messages
@@ -406,6 +407,10 @@ cdef class HttpParser:
             if "\x00" in value:
                 raise InvalidHeader(self._raw_value)
 
+            if name in SINGLETON_HEADERS:
+                if name in self._seen_singletons:
+                    raise BadHttpMessage(f"Duplicate '{name}' header found.")
+                self._seen_singletons.add(name)
             self._headers.append((name, value))
             if len(self._headers) > self._max_headers:
                 raise BadHttpMessage("Too many headers received")
@@ -445,14 +450,6 @@ cdef class HttpParser:
         raw_headers = tuple(self._raw_headers)
         headers = HeadersDictProxy(CIMultiDict(self._headers))
 
-        # https://www.rfc-editor.org/rfc/rfc9110.html#name-collected-abnf
-        bad_hdr = next(
-            (h for h in SINGLETON_HEADERS if len(headers._md.getall(h, ())) > 1),
-            None,
-        )
-        if bad_hdr is not None:
-            raise BadHttpMessage(f"Duplicate '{bad_hdr}' header found.")
-
         if self._cparser.type == cparser.HTTP_REQUEST:
             h_upg = headers.get("upgrade", "")
             allowed = upgrade and h_upg.isascii() and h_upg.lower() in ALLOWED_UPGRADES
@@ -703,6 +700,7 @@ cdef int cb_on_message_begin(cparser.llhttp_t* parser) except -1:
 
     pyparser._started = True
     pyparser._headers = []
+    pyparser._seen_singletons = set()
     pyparser._raw_headers = []
     PyByteArray_Resize(pyparser._buf, 0)
     pyparser._path = None
diff --git a/aiohttp/http_parser.py b/aiohttp/http_parser.py
index be5c3fe61..685c2e557 100644
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -78,6 +78,25 @@ VERSRE: Final[Pattern[str]] = re.compile(r"HTTP/(\d)\.(\d)", re.ASCII)
 DIGITS: Final[Pattern[str]] = re.compile(r"\d+", re.ASCII)
 HEXDIGITS: Final[Pattern[bytes]] = re.compile(rb"[0-9a-fA-F]+")
 
+# https://www.rfc-editor.org/rfc/rfc9110.html#name-field-order
+# https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5-8
+# https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.2-13.1
+# https://www.rfc-editor.org/rfc/rfc9110.html#name-collected-abnf
+SINGLETON_HEADERS: Final[frozenset[istr]] = frozenset(
+    {
+        hdrs.CONTENT_LENGTH,
+        hdrs.CONTENT_LOCATION,
+        hdrs.CONTENT_RANGE,
+        hdrs.CONTENT_TYPE,
+        hdrs.ETAG,
+        hdrs.HOST,
+        hdrs.MAX_FORWARDS,
+        hdrs.SERVER,
+        hdrs.TRANSFER_ENCODING,
+        hdrs.USER_AGENT,
+    }
+)
+
 
 class RawRequestMessage(NamedTuple):
     method: str
@@ -195,6 +214,8 @@ class HeadersParser:
             elif _FIELD_VALUE_FORBIDDEN_CTL_RE.search(value):
                 raise InvalidHeader(bvalue)
 
+            if name in SINGLETON_HEADERS and name in headers:
+                raise BadHttpMessage(f"Duplicate '{name}' header found.")
             headers.add(name, value)
             raw_headers.append((bname, bvalue))
 
@@ -501,28 +522,6 @@ class HttpParser(abc.ABC, Generic[_MsgT]):
         upgrade = False
         chunked = False
 
-        # https://www.rfc-editor.org/rfc/rfc9110.html#name-field-order
-        # https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5-8
-        # https://www.rfc-editor.org/rfc/rfc9110.html#section-10.2.2-13.1
-        # https://www.rfc-editor.org/rfc/rfc9110.html#name-collected-abnf
-        singletons = (
-            hdrs.CONTENT_LENGTH,
-            hdrs.CONTENT_LOCATION,
-            hdrs.CONTENT_RANGE,
-            hdrs.CONTENT_TYPE,
-            hdrs.ETAG,
-            hdrs.HOST,
-            hdrs.MAX_FORWARDS,
-            hdrs.SERVER,
-            hdrs.TRANSFER_ENCODING,
-            hdrs.USER_AGENT,
-        )
-        bad_hdr = next(
-            (h for h in singletons if len(headers._md.getall(h, ())) > 1), None
-        )
-        if bad_hdr is not None:
-            raise BadHttpMessage(f"Duplicate '{bad_hdr}' header found.")
-
         # keep-alive and protocol switching
         # RFC 9110 section 7.6.1 defines Connection as a comma-separated list.
         conn_values = headers.get(hdrs.CONNECTION)
```

Dreamsorcerer: > might be able to avoid the perf regression with something like

There shouldn't be a regression here, as it's doing the same thing it's always done. The ._md is the multidict we've always used here.

If we can get an improvement, then that's great too.

aiohttp/helpers.py
```diff
@@ -749,10 +751,62 @@ def ceil_timeout(
     return async_timeout.timeout_at(when)
 
 
+class HeadersDictProxy(Mapping[str, str]):
+    def __init__(self, md: CIMultiDict[str]):
+        self._md = CIMultiDictProxy(md)
```
Dreamsorcerer: The proxy redirection is probably unneeded.
```suggestion
        self._md = md
```

aiohttp/helpers.py
```diff
@@ -44,7 +45,7 @@
 from urllib.parse import quote
 from urllib.request import getproxies, proxy_bypass
 
-from multidict import CIMultiDict, MultiDict, MultiDictProxy, MultiMapping
+from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
```
```suggestion
from multidict import CIMultiDict, MultiDict, MultiDictProxy
```

aiohttp/helpers.py
```diff
@@ -753,10 +785,57 @@ def ceil_timeout(
...
+        for k in self._md.__iter__():
+            if k in seen:
+                continue
+            seen.add(k)
+            yield k
+
+    def __len__(self) -> int:
+        return len(set(self._md.keys()))
+
+    @reprlib.recursive_repr()
```
Dreamsorcerer: Shouldn't be needed as it only holds str.
```suggestion
```

aiohttp/helpers.py
```diff
@@ -13,6 +13,7 @@
 import os
 import platform
 import re
+import reprlib
```
```suggestion
```

aiohttp/_http_parser.pyx
Dreamsorcerer: Think this was done in another PR.

aiohttp/http_parser.py
Dreamsorcerer: It now wraps multidict. There's no obvious performance impact in the benchmarks.

Dreamsorcerer:
This should be ready to merge now, if anyone would like to do a final review.

aiohttp/helpers.py
```diff
@@ -753,10 +784,56 @@ def ceil_timeout(
...
 
 
+class HeadersDictProxy(Mapping[str, str]):
+    def __init__(self, md: CIMultiDict[str]):
+        self._md = md
+
+    def getall(self, key: str) -> tuple[str, ...]:
+        return self._split_on_commas(self.get(key, ""))
+
+    def _split_on_commas(self, val: str) -> tuple[str, ...]:
```
```suggestion
        val = self.get(key, "")
```

PR #7684, opened 2023-10-09
Make setup.cfg packages explicit

Dreamsorcerer:
Avoids accidentally including other packages.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,7 @@ classifiers =
 
 [options]
 python_requires = >=3.8
-packages = find:
+packages = aiohttp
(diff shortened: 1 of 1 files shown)


webknjaz:
Do we include tests into sdists, by the way?

webknjaz:
I'd like to integrate https://github.com/marketplace/actions/checkout-python-sdist at some point, to make sure everything required for testing is included..

Dreamsorcerer:
> Do we include tests into sdists, by the way?

Should be handled by MANIFEST.in, I believe.
https://github.com/aio-libs/aiohttp/blob/139069af539b5687edfd6ad2790dc7373c0c515b/MANIFEST.in#L9

PR #7688, opened 2023-10-09
 Add runner.serve_forever()

vsmaxim:
## What do these changes do?
This is a continuation of work from #4378. I've dropped `cleanup` call on `serve_forever()` task cancellation 
and updated the documentation accordingly. 

No.

## Related issue number
#2746 

Original attempt from asvetlov: https://github.com/aio-libs/aiohttp/compare/master...server_forever

## Checklist

### DIFF
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -2555,10 +2555,10 @@ application on specific TCP or Unix socket, e.g.::
       Stop handling all registered sites and cleanup used resources.
 
    .. method:: serve_forever()
+      :async:
 
-      Wait forever. Cancellation of serve_forever task causes the runner to be closed.
-      This method can be called only if the runner is already initialized.
-      Only one serve_forever task can exist per one runner object.
+      Wait forever. Make sure to call :meth:`setup` prior calling this method.
+      Only one :meth:`serve_forever` task is allowed per one runner object.
 
 
 .. class:: AppRunner(app, *, handle_signals=False, **kwargs)
--- a/docs/web_reference.rst
+++ b/docs/web_reference.rst
@@ -2511,7 +2511,7 @@ application on specific TCP or Unix socket, e.g.::
     site = web.TCPSite(runner, 'localhost', 8080)
     await site.start()
     # wait for finish signal
-    await runner.cleanup()
+    await runner.serve_forever()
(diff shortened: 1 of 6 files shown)


docs/web_reference.rst
```diff
@@ -2554,6 +2554,12 @@ application on specific TCP or Unix socket, e.g.::
 
       Stop handling all registered sites and cleanup used resources.
 
+   .. method:: serve_forever()
+      :async:
+
+      Wait forever. Make sure to call :meth:`setup` prior calling this method.
+      Only one :meth:`serve_forever` task is allowed per one runner object.
```
Dreamsorcerer, line 2561:
```suggestion
      Only one :meth:`serve_forever` task is allowed per runner object.
```

Dreamsorcerer:
Sorry for taking so long to look at this. I have a few questions about the implementation and expected behaviour:

- The original attempt ran cleanup on cancellation. Should this still happen? Seems it would match the asyncio.Server behaviour which says "Cancellation of serve_forever task causes the server to be closed."
- The original also checks if there's any sites, should we include that check?
- Along those same lines, should we call `Site.start()` for each site, so the user doesn't need to call that themselves? This would match the asyncio.Server behaviour which call Server.start_serving() from within Server.serve_forever().
- Should an explicit call to Runner.cleanup() cancel the future in server_forever()?
- Should it be possible to call Runner.serve_forever() multiple times? You've added a test for this, but I'm not sure if that behaviour makes sense and doesn't seem to match Server.serve_forever().

tests/test_web_runner.py
```diff
@@ -265,3 +265,29 @@ async def shutdown():
...
+    runner = make_runner()
+    with pytest.raises(RuntimeError):
+        await runner.serve_forever()
+
+
+async def test_app_runner_serve_forever_concurrent_call(make_runner, loop) -> None:
+    runner = make_runner()
+    task = loop.create_task(runner.serve_forever())
+    await asyncio.sleep(0.01)
+    with pytest.raises(RuntimeError):
```
Dreamsorcerer, line 280: Can we use the match argument to match the text of the exception in these?
The codecov comment suggests this test isn't working correctly.

PR #7689, opened 2023-10-09
feature: allow link argument to be set to None/empty in HTTP 451 exception

hhromic:
## What do these changes do?

This PR allows for the link argument to be set to None/empty (so it can be optional) as per RFC in the HTTP 451 exception (`HTTPUnavailableForLegalReasons`). It also adds suitable unit tests for the `link` argument and aligns variable naming in the test class of this exception.

I also updated the documentation for this exception and adjusted the documented signatures of the other exceptions in the same section (the `body` argument does not exist anymore).

The exception constructor no longer requires the `link` argument nor accepts empty or invalid values.

## Checklist

### DIFF
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -420,20 +420,18 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
 
     def __init__(
         self,
-        link: Optional[StrOrURL] = None,
+        link: Optional[StrOrURL],
         *,
         headers: Optional[LooseHeaders] = None,
         reason: Optional[str] = None,
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
-        if link is not None and not link:
-            raise ValueError("link argument cannot be empty.")
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
         self._link = None
-        if self.link:
+        if link:
             self._link = URL(link)
             self.headers["Link"] = f'<{str(self._link)}>; rel="blocked-by"'
 
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -427,11 +427,16 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
+        if not link:
+            raise ValueError(
+                "HTTP unavailable for legal reasons needs a link to"
+                " a resource with information for the blocking reason."
+            )
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
-        self.headers["Link"] = f'<{str(link)}>; rel="blocked-by"'
         self._link = URL(link)
+        self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
 
     @property
     def link(self) -> URL:
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -436,7 +436,7 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
             self.headers["Link"] = f'<{str(self._link)}>; rel="blocked-by"'
 
     @property
-    def link(self) -> URL:
+    def link(self) -> Optional[URL]:
         return self._link
 
 
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -427,11 +427,11 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
+        if link is not None and not link:
+            raise ValueError("link argument cannot be empty.")
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
-        if link is not None and not link:
-            raise ValueError("link argument cannot be empty.")
         self._link = URL(link) if link else None
         if self._link:
             self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -420,23 +420,21 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
 
     def __init__(
         self,
-        link: StrOrURL,
+        link: Optional[StrOrURL] = None,
         *,
         headers: Optional[LooseHeaders] = None,
         reason: Optional[str] = None,
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
-        if not link:
-            raise ValueError(
-                "HTTP unavailable for legal reasons needs a link to"
-                " a resource with information for the blocking reason."
-            )
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
-        self._link = URL(link)
-        self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
+        if link is not None and not link:
+            raise ValueError("link argument cannot be empty.")
+        self._link = URL(link) if link else None
+        if self._link:
+            self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
 
     @property
     def link(self) -> URL:
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -432,9 +432,10 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
-        self._link = URL(link) if link else None
+        self._link = None
         if self.link:
-            self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
+            self._link = URL(link)
+            self.headers["Link"] = f'<{str(self._link)}>; rel="blocked-by"'
 
     @property
     def link(self) -> URL:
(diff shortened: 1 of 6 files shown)


hhromic:
Probably this change should be backported to supported versions.

aiohttp/web_exceptions.py
```diff
@@ -427,11 +427,16 @@ def __init__(
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
+        if not link:
+            raise ValueError(
+                "HTTP unavailable for legal reasons needs a link to"
+                " a resource with information for the blocking reason."
+            )
```
Dreamsorcerer: I think you've misread the spec:

> When an entity blocks access to a resource and returns status 451, it SHOULD include a "Link" HTTP header field [[RFC5988](https://datatracker.ietf.org/doc/html/rfc5988)] whose value is a URI reference [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)] identifying itself.  When used for this purpose, the "Link" header field MUST have a "rel" parameter whose value is "blocked-by".
> 
> The intent is that the header be used to identify the entity actually implementing blockage, not any other entity mandating it.  A human-readable response body, as discussed above, is the appropriate location for discussion of administrative and policy issues.
https://datatracker.ietf.org/doc/html/rfc7725#section-4

So, we should probably allow `None` and then exclude the header in that case. The link is also to identify who is implementing the blocking, not the reason for the blocking (which should be in the body).

Dreamsorcerer: So, I would suggest changing the parameter to allow `None`, and then only add the header under `if link:`. Plus, update the docs below to give the correct usecase for a link.

If you're happy to do that, I think it'll be good to merge.

hhromic: Oh, you are absolutely right.
I did read that a link is optional when reviewing the spec but then my brain just forgot it when implementing 🤦.
Regarding the link description, I just didn't want to change it too much from the existing description.
I will do all the suggested changes 👍.

hhromic: Implemented change in [8fde58b](https://github.com/aio-libs/aiohttp/pull/7689/commits/8fde58b3499cd3310f7b790e6a51b181e9bca75d).
Let me know if anything else is needed.

aiohttp/web_exceptions.py
```diff
@@ -420,18 +420,21 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
 
     def __init__(
         self,
-        link: StrOrURL,
+        link: Optional[StrOrURL] = None,
```
Dreamsorcerer: Given that the spec says that we _should_ include a URL, do you think we should remove the default? Thus requiring a user to explicitly opt for no link.

aiohttp/web_exceptions.py
```diff
@@ -420,18 +420,21 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
...
         *,
         headers: Optional[LooseHeaders] = None,
         reason: Optional[str] = None,
         text: Optional[str] = None,
         content_type: Optional[str] = None,
     ) -> None:
+        if link is not None and not link:
+            raise ValueError("link argument cannot be empty.")
```
Dreamsorcerer: Do you think this is worth checking? We could just treat an empty str the same as None.

aiohttp/web_exceptions.py
```diff
@@ -420,18 +420,21 @@ class HTTPUnavailableForLegalReasons(HTTPClientError):
...
         super().__init__(
             headers=headers, reason=reason, text=text, content_type=content_type
         )
-        self.headers["Link"] = f'<{str(link)}>; rel="blocked-by"'
-        self._link = URL(link)
+        self._link = URL(link) if link else None
+        if self.link:
+            self.headers["Link"] = f'<{str(self.link)}>; rel="blocked-by"'
```
Dreamsorcerer:
```suggestion
        self._link = None
        if self.link:
            self._link = URL(link)
            self.headers["Link"] = f'<{str(self._link)}>; rel="blocked-by"'
```
Maybe avoid the duplicate check?

docs/web_exceptions.rst
```diff
@@ -104,7 +104,14 @@ unsupported method and list of allowed methods::
 
     HTTPMethodNotAllowed(method, allowed_methods, *,
                          headers=None, reason=None,
-                         body=None, text=None, content_type=None)
+                         text=None, content_type=None)
+
+:exc:`HTTPUnavailableForLegalReasons` is constructed by optionally providing
+a link to the entity implementing the blockage (if any)::
```
Dreamsorcerer:
```suggestion
:exc:`HTTPUnavailableForLegalReasons` should be constructed with a ``link``
to yourself (as the entity implementing the blockage), and an explanation for
the block included in ``text``.::
```

docs/web_exceptions.rst
```diff
@@ -478,14 +485,14 @@ HTTP exceptions for status code in range 400-499, e.g. ``raise web.HTTPNotFound(
    An exception for *451 Unavailable For Legal Reasons*, a subclass of
    :exc:`HTTPClientError`.
 
-   :param link: A link to a resource with information for blocking reason,
+   :param link: A link to the entity implementing the blockage,
                 :class:`str` or :class:`~yarl.URL`
```
Dreamsorcerer:
```suggestion
   :param link: A link to yourself (as the entity implementing the blockage),
                :class:`str`, :class:`~yarl.URL` or ``None``.
```

docs/web_exceptions.rst
```diff
@@ -478,14 +485,14 @@ HTTP exceptions for status code in range 400-499, e.g. ``raise web.HTTPNotFound(
...
 
    For other parameters see :exc:`HTTPException` constructor.
```
Dreamsorcerer, line 492:
```suggestion
   For other parameters see :exc:`HTTPException` constructor.
   A reason for the block should be included in ``text``.
```

aiohttp/web_exceptions.py
hhromic: Umm, you mean the following cases?
```python
# valid construction
HTTPUnavailableForLegalReasons("http://example.com/legal")
HTTPUnavailableForLegalReasons(link="http://example.com/legal")
HTTPUnavailableForLegalReasons(None)
HTTPUnavailableForLegalReasons(link=None)

# invalid construction
HTTPUnavailableForLegalReasons()
```
In this case I would review other exception constructors to make sure this way of doing things is consistent.

EDIT: The API would look a bit weird imho, but maybe makes sense if a link should _really_ be provided and thus "non-intuitive" for a developer to avoid it. Let me know!

aiohttp/web_exceptions.py
hhromic: I implemented it that way to be consistent with other exceptions accepting URL arguments.
I can make it just `if link:` if you prefer. Let me know!

aiohttp/web_exceptions.py
hhromic: This depends on the previous comment's outcome, but can do of course. 👍

aiohttp/web_exceptions.py
Dreamsorcerer: Well, if there are similar RFC specs that say 'SHOULD', then yeah, it might be worth doing to others. But, those changes could only go into master, as it would break backwards compatibility. We can manage this one now, because it never had a default before.

aiohttp/web_exceptions.py
Dreamsorcerer: Which other ones are you referring to? The only similar one I see is location in HTTPMove, where the url is mandatory.

aiohttp/web_exceptions.py
hhromic: Sounds good to me then. I'll work the argument behaviour like suggested.
Yes, reviewing the others can be a later PR too indeed given the potential breakage.

aiohttp/web_exceptions.py
hhromic: Yes I meant the HTTPMove exception indeed, in particular this check:
```python
        if not location:
            raise ValueError("HTTP redirects need a location to redirect to.")
```
That only accepts `location` if not None _and_ not empty. Which is what I tried to mirror with `link`.

hhromic: I will rework the check 👍

hhromic: Implemented the changes in the recent version of this PR.

hhromic:
@Dreamsorcerer I implemented the changes as discussed and also updated the unit tests accordingly.
Let me know if anything else is missing/needed.

aiohttp/web_exceptions.py
```diff
@@ -430,8 +430,10 @@ def __init__(
...
         )
-        self.headers["Link"] = f'<{str(link)}>; rel="blocked-by"'
-        self._link = URL(link)
+        self._link = None
+        if link:
+            self._link = URL(link)
+            self.headers["Link"] = f'<{str(self._link)}>; rel="blocked-by"'
 
     @property
     def link(self) -> URL:
```
Dreamsorcerer:
```suggestion
    def link(self) -> Optional[URL]:
```

tests/test_web_exceptions.py
```diff
@@ -312,23 +312,51 @@ def test_pickle(self) -> None:
...
-        assert resp.headers == compare
-        assert resp.reason == "Zaprescheno"
-        assert resp.status == 451
+        assert exc.headers == compare
+        assert exc.reason == "Zaprescheno"
+        assert exc.status == 451
+
+    def test_no_link(self) -> None:
+        with pytest.raises(TypeError):
+            web.HTTPUnavailableForLegalReasons()
```
Dreamsorcerer:
```suggestion
            web.HTTPUnavailableForLegalReasons()  # type: ignore[call-arg]
```

Dreamsorcerer:
@hhromic Would be great if you can handle the backport, following the steps above.

hhromic:
@Dreamsorcerer thanks for merging! I'll check out the backport 👍.

PR #7698, opened 2023-10-13
Tightening the runtime type check for ssl

xiangyan99:
Currently, the valid types of ssl parameter are SSLContext, Literal[False], Fingerprint or None.

If user sets ssl = False, we disable ssl certificate validation which makes total sense. But if user set ssl = True by mistake, instead of enabling ssl certificate validation or raising errors, we silently disable the validation too which is a little subtle but weird.

In this PR, we added a check that if user sets ssl=True, we enable certificate validation by treating it as using Default SSL Context.


webknjaz (changes requested):
This would need motivation explanation, tests, change note. But overall, it just seems wrong.

aiohttp/client.py
```diff
@@ -784,9 +784,13 @@ async def _ws_connect(
             extstr = ws_ext_gen(compress=compress)
             real_headers[hdrs.SEC_WEBSOCKET_EXTENSIONS] = extstr
 
-        if not isinstance(ssl, SSL_ALLOWED_TYPES):
+        # We need this code to keep backward compatibility. If ssl is True, we transform it to None
+        if ssl is True:
+            ssl = None
+
+        if not isinstance(ssl, SSL_ALLOWED_TYPES) and ssl is not False:
```
Dreamsorcerer: Looking at it again, this check happens in 3 different places. So, we'd probably need to refactor the entire check into a function to keep the logic well contained. This is why it hasn't been done before.

xiangyan99: I reorganized the check method into helper.verify_ssl_type().

aiohttp/helpers.py
```diff
@@ -13,6 +13,7 @@
 import os
 import platform
 import re
+import ssl
```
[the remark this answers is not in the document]
webknjaz:
```suggestion
```

aiohttp/helpers.py
```diff
@@ -1063,3 +1071,13 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
             with suppress(ValueError):
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+def verify_ssl_type(
+    ssl: Optional[Union[SSLContext, Literal[False], Fingerprint]]
```
Dreamsorcerer: It's already a union.
```suggestion
    ssl: Union[SSLContext, Literal[False], Fingerprint, None]
```

aiohttp/helpers.py
```diff
@@ -1063,3 +1071,13 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
...
+) -> None:
+    if not isinstance(ssl, SSL_ALLOWED_TYPES) and ssl is not False:
+        raise TypeError(
+            "ssl should be SSLContext, Fingerprint, "
```
```suggestion
            "ssl should be SSLContext, Fingerprint, False "
```

aiohttp/client.py
```diff
@@ -784,11 +775,11 @@ async def _ws_connect(
...
             real_headers[hdrs.SEC_WEBSOCKET_EXTENSIONS] = extstr
 
-        if not isinstance(ssl, SSL_ALLOWED_TYPES):
-            raise TypeError(
-                "ssl should be SSLContext, bool, Fingerprint, "
-                "or None, got {!r} instead.".format(ssl)
-            )
+        # For the customers that mistakenly pass in True, convert it to None to use default ssl context
+        if ssl is True:
+            ssl = None
```
Dreamsorcerer: @webknjaz what do you think? Currently, nothing in the documentation mentions `True` as an allowed value. I'm wondering if it just makes more sense to replace `None` with `True` (in v4) and drop `None` as an allowed value entirely. In which case we can just keep the existing SSL type checks instead...

xiangyan99: At least from a user perspective, allowing `False` but not allowing `True` is a little weird. :)

xiangyan99: Ping. Do we have any luck to make progress? @Dreamsorcerer @webknjaz

Dreamsorcerer: If @webknjaz isn't around, maybe @bdraco can give an opinion on this.

bdraco: I expect most consumers of the client will have to handle the `ssl` flag, which makes the impact of a breaking change feel quite a bit more.

Even with a major version bump, I'd be inclined not to make a breaking change here or add new options if there is a substantial reason to do it.  

This PR needs more opening text and an explanation of the motivation for changing it, so it's hard to give good feedback here.

xiangyan99: @bdraco Thanks for the comments.

Currently, the valid types of ssl parameter are SSLContext, Literal[False], Fingerprint or None.

If user sets ssl = False, we disable ssl certificate validation which makes total sense. But if user set ssl = True by mistake, instead of enabling ssl certificate validation or raising errors, we silently disable the validation too which is a little subtle but weird.

As you said, we want to avoid breaking changes as much as possible.

Hence in this PR, we added a check that if user sets ssl=True, we enable certificate validation by treating it as using Default SSL Context.

Does this make sense? (I also added these into the PR description)

bdraco: That makes sense. `True` definitely doesn't do what one would expect. I'm sure there are cases where people have passed in `True` and hoped it would enable verification.

By changing the meaning of `True`, we introduce a breaking change since they now get verification, which may break their production code that may have relied on no verification (most likely mistakenly). It's a good change, but still a breaking one (even if they shouldn't be doing that).

I think that's OK, but we should warn them about the behavior change so they don't have to dig through the docs (and maybe the code) to determine what has changed. I imagine that would be a difficult change to find without some feedback from the code.

xiangyan99: Are you leaning towards reverting the type checking and only having the change

if ssl=True:
   ssl=None

?

I am open to it too if @Dreamsorcerer & @webknjaz are ok.

xiangyan99: After re-read your comments, are you saying that the change may break the case that user explicitly sets the ssl to True but relies on the behavior of no ssl verification?

I agree this is technically breaking. But I assume this is a positive breaking change. :)

bdraco: I think this PR is fine as-is if you add a warning when `True` is converted to `None` since `True` is not part of the api. 

The public type is `ssl: Optional[Union[SSLContext, Literal[False], Fingerprint]] = None,` which I think is good to stay as is.

bdraco:
```suggestion
            warnings.warn(
                "True is not a valid value for the ssl parameter; "
                "using default ssl context. Previously this "
                "would silently disable verification.",
            )
            ssl = None
```

bdraco: We should also get a test that setting `True` now enables verification

Dreamsorcerer: > By changing the meaning of `True`, we introduce a breaking change since they now get verification, which may break their production code that may have relied on no verification (most likely mistakenly). It's a good change, but still a breaking one (even if they shouldn't be doing that).

`True` has never had a meaning, it's an invalid input. So, I'd prefer to disallow `True`, or replace `None` with `True` in future releases (bearing in mind that `None` is the default, so may not be passed explicitly by many people).

Dreamsorcerer: If we want to change it to `True`, then we can always add a DeprecationWarning when using `None`, to give people plenty of time to switch any explicit arguments before v4.

bdraco: > `True` has never had a meaning, it's an invalid input. So, I'd prefer to disallow `True`, or replace `None` with `True` in future releases (bearing in mind that `None` is the default, so may not be passed explicitly by many people).

Raising when passing `True` seems fine to me as well. It makes it easy to see what went wrong.

> If we want to change it to `True`, then we can always add a DeprecationWarning when using `None`, to give people plenty of time to switch any explicit arguments before v4.

The downside to adding support for `True` is that anyone already passing it gets no warning that the behavior has changed because it becomes a valid option after an update, so we can't warn them of a behavior change.

Since previously it was an invalid option and a bug for them to pass it, assuming we are only talking about changing it out from under them if they explicitly pass a value (it should be secure by default with no passed value) to work more like one would intuitively expect (True/False) vs. (None/False) might be worth the cost of the few users who will get no warning.

Dreamsorcerer: > The downside to adding support for `True` is that anyone already passing it gets no warning that the behavior has changed because it becomes a valid option after an update, so we can't warn them of a behavior change.

This part doesn't concern me, as it is obvious their intention was to enable ssl, i.e. the default behaviour. The behaviour change will merely make it do what they were expecting.

I think the only users who need to update anything are those who get a deprecation warning from passing None (maybe because they wrap the client and have their own default arguments matching ours).

Dreamsorcerer: If you agree, then I'll leave that with you to oversee. I'll be back Wednesday.

bdraco: > This part doesn't concern me, as it is obvious their intention was to enable ssl, i.e. the default behaviour. The behaviour change will merely make it do what they were expecting.
> 
> I think the only users who need to update anything are those who get a deprecation warning from passing None (maybe because they wrap the client and have their own default arguments matching ours).

> If you agree, then I'll leave that with you to oversee. I'll be back Wednesday.

Sounds good. Let's change it to be `True`/`False` and deprecate explicit `None` with a `DeprecationWarning`  (passing nothing should default to `True`)

xiangyan99: To make sure I understand correctly, we want to change it to:
1. Default the value of ssl to True.
2. If user sets it to True explicitly, we will change the behavior from disabling ssl verification to enabling ssl verfication
3. If user explicitly sets it to None, we will keep current behavior which is to enable ssl verification and add a deprecation warning

Is this correct?

bdraco: That looks correct to me 👍

Dreamsorcerer: Slight tweak, as I forgot about backports.
So, this PR for master:
Change value to True. Remove None as an allowed value.

Then when that is merged, we can do the backports, I think in 3.9 we should allow None and True. Then in 3.10 backport we add a deprecation warning to None. That should cause the minimal amount of problems.

aiohttp/client.py
```diff
@@ -785,11 +775,15 @@ async def _ws_connect(
...
-        if not isinstance(ssl, SSL_ALLOWED_TYPES):
-            raise TypeError(
-                "ssl should be SSLContext, bool, Fingerprint, "
-                "or None, got {!r} instead.".format(ssl)
+        # For the sake of backward compatibility, if user passes in None, convert it to True
+        if ssl is None:
+            warnings.warn(
+                "None is a deprecated value for the ssl parameter; "
+                "using True for default ssl context.",
             )
```
webknjaz: Deprecation warnings should set the `DeprecationWarning` (or `FutureDeprecationWarning`) type + usually `stacklevel=2`.

aiohttp/connector.py
```diff
@@ -931,13 +934,7 @@ def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
...
             if isinstance(sslcontext, ssl.SSLContext):
                 return sslcontext
-            if sslcontext is not None:
-                # not verified or fingerprinted
-                return self._make_ssl_context(False)
-            sslcontext = self._ssl
-            if isinstance(sslcontext, ssl.SSLContext):
-                return sslcontext
-            if sslcontext is not None:
+            if sslcontext is False:
```
webknjaz: Can it be just
```suggestion
            if not sslcontext:
```
?

tests/test_connector.py
```diff
@@ -1348,6 +1348,15 @@ async def test___get_ssl_context6(loop: Any) -> None:
     assert conn._get_ssl_context(req) is conn._make_ssl_context(True)
 
 
+async def test___get_ssl_context7(loop: Any) -> None:
```
webknjaz: We should really start using meaningful names for the test functions...

alessio-locatelli: > meaningful names

Great point. "Meaningful" is a bit of a vague description, so I will add my two cents here to make it a bit clearer. It would be convenient if tests had the following naming pattern:

 `test_` + `<function_name>` + `(optional) <case or state or expected_behavior>`

aiohttp/client.py
Dreamsorcerer: This will be relevant for the 3.10 backport, yes.
But, this code shouldn't exist in master, where None should not be an allowed value anymore.

aiohttp/client.py
```diff
@@ -369,11 +363,7 @@ async def _request(
         if self.closed:
             raise RuntimeError("Session is closed")
 
-        if not isinstance(ssl, SSL_ALLOWED_TYPES):
-            raise TypeError(
-                "ssl should be SSLContext, bool, Fingerprint, "
-                "or None, got {!r} instead.".format(ssl)
-            )
```
Dreamsorcerer: This can probably just be reverted, as the check remains simple now.
(Just remember to keep the other code around for the backports.)

aiohttp/client_reqrep.py
Dreamsorcerer, line 1: None also needs to be removed from allowed types in here.

aiohttp/client.py
```diff
@@ -353,7 +347,7 @@ async def _request(
         proxy: Optional[StrOrURL] = None,
         proxy_auth: Optional[BasicAuth] = None,
         timeout: Union[ClientTimeout, _SENTINEL, None] = sentinel,
-        ssl: Optional[Union[SSLContext, Literal[False], Fingerprint]] = None,
+        ssl: Optional[Union[SSLContext, bool, Fingerprint]] = True,
```
```suggestion
        ssl: Union[SSLContext, bool, Fingerprint] = True,
```

aiohttp/connector.py
```diff
@@ -931,13 +934,7 @@ def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
             sslcontext = req.ssl
             if isinstance(sslcontext, ssl.SSLContext):
                 return sslcontext
-            if sslcontext is not None:
-                # not verified or fingerprinted
-                return self._make_ssl_context(False)
-            sslcontext = self._ssl
-            if isinstance(sslcontext, ssl.SSLContext):
-                return sslcontext
-            if sslcontext is not None:
```
Dreamsorcerer: None of this looks like it should be removed, only the `is None` condition changed.

aiohttp/connector.py
Dreamsorcerer: Shouldn't it actually be:
```suggestion
            if sslcontext is not True:
```

Dreamsorcerer: Yeah, I questioned myself there, but this condition should be hit (as the comment suggests) when it is either a non-validating context (i.e. False) or fingerprinted (i.e. Fingerprint()).
```suggestion
            if sslcontext is not True:
```

aiohttp/client.py
webknjaz: I'd argue that a PR deprecating it should go in separately, followed by a non-backported PR that actually does the removal.

aiohttp/client.py
webknjaz: Here's my opinion on the order of the changes: https://github.com/aio-libs/aiohttp/pull/7698#discussion_r1411152254.

aiohttp/client.py
Dreamsorcerer: Hmm, the deprecation is likely to create more changes though, which then need to be reverted in the removal. So, that seems like more work to me.

Dreamsorcerer: Actually, maybe not, I'm thinking of some changes that are no longer relevant. I don't mind which way it's done.

aiohttp/connector.py
xiangyan99: They are duplicated code. (exactly same as code above)

aiohttp/client.py
xiangyan99: This is moved into method verify_ssl_type()

aiohttp/connector.py
Dreamsorcerer: No it's not, it checks req.ssl and then self._ssl.

aiohttp/client.py
Dreamsorcerer: Yes, I'm just not sure we need a method anymore given it is a very simple check. It would make the review easier if we can minimise unrelated changes here.

aiohttp/client.py
```diff
@@ -504,7 +503,7 @@ async def _request(
                         proxy_auth=proxy_auth,
                         timer=timer,
                         session=self,
-                        ssl=ssl,
+                        ssl=ssl if ssl is not None else True,
```
Dreamsorcerer: This can't be None with the above changes.

aiohttp/client.py
```diff
@@ -711,7 +710,7 @@ def ws_connect(
                 headers=headers,
                 proxy=proxy,
                 proxy_auth=proxy_auth,
-                ssl=ssl,
+                ssl=ssl if ssl is not None else True,
```
Dreamsorcerer: Same

aiohttp/client.py
```diff
@@ -792,10 +791,14 @@ async def _ws_connect(
             extstr = ws_ext_gen(compress=compress)
             real_headers[hdrs.SEC_WEBSOCKET_EXTENSIONS] = extstr
 
+        # For the sake of backward compatibility, if user passes in None, convert it to True
```
Dreamsorcerer, line 794: Are we doing the backwards compatible approach here, and then removing it with a second PR, or just removing it in this PR? Currently, it seems to be mixing up both.

aiohttp/client_reqrep.py
```diff
@@ -147,7 +146,7 @@ def check(self, transport: asyncio.Transport) -> None:
 
 
 if ssl is not None:
-    SSL_ALLOWED_TYPES = (ssl.SSLContext, bool, Fingerprint, type(None))
```
Dreamsorcerer, line 150: I think if we're going for backwards compatible first, then maybe just this line needs reverting? (and we can keep the annotation changes to discourage None usage)

Probably need the DeprecationWarning somewhere as well.

aiohttp/client_reqrep.py
```diff
@@ -147,7 +146,7 @@ def check(self, transport: asyncio.Transport) -> None:
...
+    SSL_ALLOWED_TYPES = (ssl.SSLContext, bool, Fingerprint)
 else:  # pragma: no cover
     SSL_ALLOWED_TYPES = type(None)
```
Dreamsorcerer: Also, this presumably needs to support bool now.
```suggestion
    SSL_ALLOWED_TYPES = (bool, type(None))
```

aiohttp/client.py
xiangyan99: I am leaning towards doing the backwards compatible approach and keeping it. :)

xiangyan99: TBH, as long as we have a deterministic behavior for ssl=None, I am not a big fan of removing it from allowed type.

Dreamsorcerer: It will provide better consistency and v4 will likely still be some time away, so there's plenty of time to migrate.
I think we just need the DeprecationWarning in this PR then, and it should be good to go.

tests/test_connector.py
```diff
@@ -1309,7 +1309,7 @@ async def test_connect(loop: Any, key: Any) -> None:
     assert connection._protocol is proto
     assert connection.transport is proto.transport
     assert isinstance(connection, Connection)
-    connection.close()
+    await connection.close()
```
Dreamsorcerer:
```suggestion
    connection.close()
```

tests/test_connector.py
```diff
@@ -1591,10 +1591,19 @@ async def test___get_ssl_context6(loop: Any) -> None:
...
     assert conn._get_ssl_context(req) is conn._make_ssl_context(True)
 
 
+async def test__get_ssl_context_ssl_true(loop: Any) -> None:
+    ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+    conn = aiohttp.TCPConnector(ssl=True)
+    req = mock.Mock()
+    req.is_ssl.return_value = True
+    req.ssl = True
+    assert conn._get_ssl_context(req) is ctx
```
Dreamsorcerer:
```suggestion
    conn = aiohttp.TCPConnector(ssl=True)
    req = mock.Mock()
    req.is_ssl.return_value = True
    req.ssl = True
    assert conn._get_ssl_context(req) is True
```

tests/test_connector.py
```diff
@@ -1591,10 +1591,18 @@ async def test___get_ssl_context6(loop: Any) -> None:
...
 
 
+async def test__get_ssl_context_ssl_true(loop: Any) -> None:
+    conn = aiohttp.TCPConnector(ssl=True)
+    req = mock.Mock()
+    req.is_ssl.return_value = True
+    req.ssl = True
+    assert conn._get_ssl_context(req) is True
+
+
```
Dreamsorcerer:
```suggestion
```

PR #7718, opened 2023-10-17
Redesign shutdown process

Dreamsorcerer:
This contains a few changes to rework the shutdown process to be a bit clearer and better match user expectations.

To summarise, I'm proposing the shutdown process should look like this now:

1. Stop each site listening for new connections.
2. Close idle keep-alive connections (and set active ones to close upon completion).
3. Call on_shutdown signal (should be used to close websocket connections).
4. Wait for running tasks to complete (with timeout).
5. Close any remaining connections (with timeout again).
6. Call on_cleanup signal.
7. Cancel any remaining tasks and wait on them.

Docs demo preview: https://aiohttp--7718.org.readthedocs.build/en/7718/web_advanced.html#graceful-shutdown

### DIFF
--- a/CHANGES/7718.feature
+++ b/CHANGES/7718.feature
@@ -1,0 +1,3 @@
+Fixed keep-alive connections stopping a graceful shutdown.
+Added ``shutdown_timeout`` parameter to ``BaseRunner``, while
+removing ``shutdown_timeout`` parameter from ``BaseSite``. -- by :user:`Dreamsorcerer`
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -3,7 +3,7 @@ import signal
 import socket
 from abc import ABC, abstractmethod
 from contextlib import suppress
-from typing import Any, List, Optional, Set, Type
+from typing import Any, Awaitable, Callable, List, Optional, Set, Type
 
 from yarl import URL
 
@@ -235,7 +235,7 @@ class SockSite(BaseSite):
 
 class BaseRunner(ABC):
     __slots__ = (
-        "starting_tasks",
+        "pre_shutdown_callback",
         "_handle_signals",
         "_kwargs",
         "_server",
@@ -250,6 +250,7 @@ class BaseRunner(ABC):
         shutdown_timeout: float = 60.0,
         **kwargs: Any,
     ) -> None:
+        self.pre_shutdown_callback: Optional[Callable[[], Awaitable[None]]] = None
         self._handle_signals = handle_signals
         self._kwargs = kwargs
         self._server: Optional[Server] = None
@@ -288,11 +289,6 @@ class BaseRunner(ABC):
                 pass
 
         self._server = await self._make_server()
-        # On shutdown we want to avoid waiting on tasks which run forever.
-        # It's very likely that all tasks which run forever will have been created by
-        # the time we have completed the application startup (in self._make_server()),
-        # so we just record all running tasks here and exclude them later.
-        self.starting_tasks = asyncio.all_tasks()
 
     @abstractmethod
     async def pre_shutdown(self) -> None:
@@ -316,11 +312,8 @@ class BaseRunner(ABC):
         self._server.pre_shutdown()
         await self.pre_shutdown()
 
-        # Wait for pending tasks for a given time limit.
-        with suppress(asyncio.TimeoutError):
-            await asyncio.wait_for(
-                self._wait(asyncio.current_task()), timeout=self._shutdown_timeout
-            )
+        if self.pre_shutdown_callback:
+            await self.pre_shutdown_callback()
 
         await self.server.shutdown(self._shutdown_timeout)
         await self.shutdown()
@@ -356,11 +349,6 @@ class BaseRunner(ABC):
         if site not in self._sites:
             raise RuntimeError(f"Site {site} is not registered in runner {self}")
         self._sites.remove(site)
-
-    async def _wait(self, parent_task: Optional["asyncio.Task[object]"]) -> None:
-        exclude = self.starting_tasks | {asyncio.current_task(), parent_task}
-        while tasks := asyncio.all_tasks() - exclude:
-            await asyncio.wait(tasks)
 
 
 class ServerRunner(BaseRunner):
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -73,9 +73,6 @@ class BaseSite(ABC):
         self._runner._check_site(self)
         if self._server is not None:  # Maybe not started yet
             self._server.close()
-            # named pipes do not have wait_closed property
-            if hasattr(self._server, "wait_closed"):
-                await self._server.wait_closed()
 
         self._runner._unreg_site(self)
 
--- a/aiohttp/web_runner.py
+++ b/aiohttp/web_runner.py
@@ -313,7 +313,8 @@ class BaseRunner(ABC):
     async def cleanup(self) -> None:
         loop = asyncio.get_event_loop()
 
-        self._server.pre_shutdown()
+        if self._server is not None:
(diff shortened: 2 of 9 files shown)


Dreamsorcerer:
There are a couple of changes I'll also make for the backport, such as removing the `@abstractmethod` from pre_shutdown, to better manage backwards compatibility.

aiohttp/web_runner.py
```diff
@@ -45,20 +45,18 @@ def _raise_graceful_exit() -> None:
...
 
 class BaseSite(ABC):
-    __slots__ = ("_runner", "_shutdown_timeout", "_ssl_context", "_backlog", "_server")
+    __slots__ = ("_runner", "_ssl_context", "_backlog", "_server")
 
     def __init__(
         self,
         runner: "BaseRunner",
         *,
-        shutdown_timeout: float = 60.0,
```
Dreamsorcerer, line 54: For the 3.9 backport, I'll need to readd this for backwards-compatibility, with a deprecation.

aiohttp/web_runner.py
```diff
@@ -74,30 +71,11 @@ async def start(self) -> None:
 
     async def stop(self) -> None:
         self._runner._check_site(self)
-        if self._server is None:
-            self._runner._unreg_site(self)
-            return  # not started yet
-        self._server.close()
-        # named pipes do not have wait_closed property
-        if hasattr(self._server, "wait_closed"):
-            await self._server.wait_closed()
```
Dreamsorcerer, line 83: So, this method in Python 3.12 actually waits until all connections are closed. But, we haven't even started to shutdown our connections yet, so this is clearly wrong. I don't think we actually need to call it at all because the runner basically waits on the connections after cancelling them.

On older versions of Python, this method does literally nothing, so this code only worked in the past by accident.

docs/web_advanced.rst
```diff
@@ -927,25 +927,38 @@ Graceful shutdown
 Stopping *aiohttp web server* by just closing all connections is not
 always satisfactory.
 
-The first thing aiohttp will do is to stop listening on the sockets,
-so new connections will be rejected. It will then wait a few
-seconds to allow any pending tasks to complete before continuing
-with application shutdown. The timeout can be adjusted with
-``shutdown_timeout`` in :func:`run_app`.
+When aiohttp is run with :func:`run_app`, it will attempt a graceful shutdown
+by following these steps (if using a :ref:`runner<aiohttp-web-app-runners>`,
```
webknjaz:
```suggestion
by following these steps (if using a :ref:`runner <aiohttp-web-app-runners>`,
```

webknjaz:
> Cancel any remaining tasks and wait on them.

Does this pose a deadlock threat?

Dreamsorcerer:
> > Cancel any remaining tasks and wait on them.
> 
> Does this pose a deadlock threat?

I think there is the possibility that a task could stop the program exiting, but at that point the rest of shutdown has already happened, so it'll be as safe as it can be for someone to kill the program. This step is unchanged from the 3.8 release.

Dreamsorcerer:
> > > Cancel any remaining tasks and wait on them.
> > 
> > 
> > Does this pose a deadlock threat?
> 
> I think there is the possibility that a task could stop the program exiting, but at that point the rest of shutdown has already happened, so it'll be as safe as it can be for someone to kill the program. This step is unchanged from the 3.8 release.

In fact, I just remembered that this code is copied from asyncio.run(), so it'll be consistent with asyncio generally.

Dreamsorcerer:
I'm merging this, if there's any feedback, I can update it before the final release (maybe in a week or so).

bdraco:
> I'm merging this, if there's any feedback, I can update it before the final release (maybe in a week or so).

Sorry for the late feedback, I wasn't shut if this was ready to test or not.

I tested this with Home Assistant, and shutdown is delayed if we don't patch out `wait`.  

Home Assistant has long running tasks that are created after `aiohttp` starts up, and are expected to be torn down after `aiohttp` is shutdown. Unfortunately we can't change that design to shutdown `aiohttp` later.

Patching out `wait` it probably fine for Home Assistant. It might be nice to have a way to disable the shutdown logic a bit more cleanly though.

CHANGES/7718.feature
```diff
@@ -0,0 +1,3 @@
+Fixed keep-alive connections stopping a graceful shutdown.
```
webknjaz, line 1: @Dreamsorcerer “fixed” implies a “bugfix” changelog note type, but the filename says “feature”.  If it's both, it might be worth splitting into two fragments.

webknjaz:
> Home Assistant has long running tasks that are created after `aiohttp` starts up, and are expected to be torn down after `aiohttp` is shutdown. Unfortunately we can't change that design to shutdown `aiohttp` later.

Maybe, you could shield them from being cancelled, doing an unshielded cancellation later, manually?

webknjaz:
> Home Assistant has long running tasks that are created after `aiohttp` starts up, and are expected to be torn down after `aiohttp` is shutdown. Unfortunately we can't change that design to shutdown `aiohttp` later.

What if there was an interface to augment the new `starting_tasks` structure with your tasks? As I understand, you track what's yours, right?

webknjaz:
@Dreamsorcerer I'm not very familiar with the shutdown internals, so I can't contribute an involved feedback. Just want to say that I don't see anything obviously wrong, and it seems like a good idea.

webknjaz:
> Unfortunately we can't change that design to shutdown `aiohttp` later.

@bdraco could you explain why?

bdraco:
> > Unfortunately we can't change that design to shutdown `aiohttp` later.
> 
> 
> 
> @bdraco could you explain why?

Home Assistant has its own tasks that are created with our helper that we can track, but there are also ~1600 dependencies that can start their own tasks so we don't have a direct way to manage them. 

While most of them will listen for a stop event and cancel the tasks, there are many custom integrations (and libraries) that don't and we can't wait for them to shutdown because we would end up waiting forever in many cases.

To give some better context, there are likely quite a few integrations that have a task running a retry loop when a device is offline that keep trying to reconnect forever until Home Assistant shuts down the event loop from under them.

webknjaz:
@bdraco have you tried exploring having a separate event loop for those?

Dreamsorcerer:
@bdraco We already discussed this in the testing issue. My understanding is that HA is using the runners directly, so in this PR I've moved the wait logic into run_app(). This means it should have no impact on HA anymore, and there isn't anything there to patch out. For users using run_app(), this is the entry point of the event loop and we expect to control the lifecycle of the event loop.

Dreamsorcerer:
(This is also now highlighted in the docs section linked at the top of the PR)

bdraco:
> @bdraco We already discussed this in the testing issue. My understanding is that HA is using the runners directly, so in this PR I've moved the wait logic into run_app(). This means it should have no impact on HA anymore, and there isn't anything there to patch out. For users using run_app(), this is the entry point of the event loop and we expect to control the lifecycle of the event loop.

I agree that I shouldn't be seeing a problem anymore, but I can reproduce still.  I will double check that I haven't screwed something up with my testing.

bdraco:
TLDR: This PR solves the issue, there is an unrelated regression on my Home Assistant `integration` branch from trying to test too many changes at once that resulted in a bad test.

The problem goes away when I use Home Assistant stable + this PR.  I must have a regression my Home Assistant `integration` branch that is blocking shutdown.  I will bisect and find the bad commit but its certainly unrelated. 

I didn't notice since I was running 3.9 prior to this PR which was always shutting down slowly (I didn't want to forget to make sure this was sorted) since it wasn't patched and I assumed the slow shutdown was still from that.

Thanks for getting this sorted. Sorry about the bad test.

bdraco:
I found the bad commit, its definitely unrelated. Sorry again for the side trip. The good news is that 3.9 is working well in production 👍

PR #7719, opened 2023-10-17
Stop accepting `\x80-\xff` in header names; stop accepting `\n` as separating whitespace in status-lines

kenballus:
## What do these changes do?

This PR updates the python HTTP parser to stop accepting `\x80-\xff` in header names and stop accepting `\n` as separating whitespace in status-lines. Both of these are not allowed in the RFCs.

Only users of seriously misbehaving clients would notice a change in behavior for this patch. If a client sends non-ascii UTF-8 within header names, their messages will now 400. Note that (nearly) arbitrary values are still allowed within header values. Most HTTP servers (Apache, Nginx, IIS, Node) do not accept UTF-8 within header names because of the risk of control character injection.

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -66,7 +66,9 @@ ASCIISET: Final[Set[str]] = set(string.printable)
 METHRE: Final[Pattern[str]] = re.compile(r"[!#$%&'*+\-.^_`|~0-9A-Za-z]+")
 VERSRE: Final[Pattern[str]] = re.compile(r"HTTP/(\d).(\d)")
 STATUSLINESEP: Final[Pattern[str]] = re.compile(r"[ \t\v\f\r]+")
-HDRRE: Final[Pattern[bytes]] = re.compile(rb"[\x00-\x1F\x7F-\xFF()<>@,;:\[\]={} \t\"\\]")
+HDRRE: Final[Pattern[bytes]] = re.compile(
+    rb"[\x00-\x1F\x7F-\xFF()<>@,;:\[\]={} \t\"\\]"
+)
 HEXDIGIT = re.compile(rb"[0-9a-fA-F]+")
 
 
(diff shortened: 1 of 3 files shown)


aiohttp/http_parser.py
```diff
@@ -540,7 +543,7 @@ def parse_message(self, lines: List[bytes]) -> RawRequestMessage:
         # request line
         line = lines[0].decode("utf-8", "surrogateescape")
         try:
-            method, path, version = line.split(maxsplit=2)
+            method, path, version = re.split(STATUSLINESEP, line, maxsplit=2)
```
Dreamsorcerer: The spec actually says it should be a space specifically. Should we restrict it further, or do we anticipate clients sending other whitespace?

kenballus: The grammar in RFC 9112 lists `' '` only as the separator, but [section 3](https://www.rfc-editor.org/rfc/rfc9112.html#section-3) says this:
> Although the request-line grammar rule requires that each of the component elements be separated by a single SP octet, recipients MAY instead parse on whitespace-delimited word boundaries and, aside from the CRLF terminator, treat any form of whitespace as the SP separator while ignoring preceding or trailing whitespace; such whitespace includes one or more of the following octets: SP, HTAB, VT (%x0B), FF (%x0C), or bare CR. However, lenient parsing can result in request smuggling security vulnerabilities if there are multiple recipients of the message and each has its own unique interpretation of robustness (see [Section 11.2](https://www.rfc-editor.org/rfc/rfc9112.html#request.smuggling)).

Personally, I think splitting on only `' '` is the safest option, but I went with the conservative option. If you'd rather split on only `' '`, then I'm on board.

Dreamsorcerer: Hmm, I'm not sure, probably just match what llhttp does then. 99% of users should be using llhttp anyway, so this code shouldn't affect many people.

kenballus: llhttp allows only `' '`. I've changed this to match.

Dreamsorcerer:
If you can handle the backport, that'd be really helpful.

PR #7723, opened 2023-10-17
remove an unnecessary implicit string concatenation

booniepepper:
## What do these changes do?

Remove an unnecessary implicit string concatenation

Just a tiny thing I noticed while looking around here. I can see no reason to be concatenating two static strings here

No changes in behavior, except it may run infinitesimally faster if someone is throwing this error too frequently?

## Checklist

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -704,9 +704,7 @@ class TimerContext(BaseTimerContext):
         task = asyncio.current_task(loop=self._loop)
 
         if task is None:
-            raise RuntimeError(
-                "Timeout context manager should be used inside a task"
-            )
+            raise RuntimeError("Timeout context manager should be used inside a task")
 
         if self._cancelled:
             raise asyncio.TimeoutError from None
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
I mean, there's dozens more of these in the code... You can thank black (and someone not double checking the output after black ran).

webknjaz:
I think there's also a flake8 plugin for these cases. FWIW, it would be more useful to integrate that and fix the problem systemically globally, rather than pick one place which is counterproductive. The PR title implies that it fixes the issue while it actually only adjust one location…

Dreamsorcerer:
Yep, fixing individual cases seems like it'll just pollute the git history and make it more difficult to find the changes someone is looking for. So, I'll close this, but if you'd like to investigate a flake8 or similar fix and fix them all in one go, we'll happily merge that.

booniepepper:
Yeah this sounds reasonable

How do you feel if I added a dependency on this flake8 plugin?

https://pypi.org/project/flake8-no-implicit-concat/

PR #7755, opened 2023-10-27
Fix py http parser not treating 204/304/1xx as an empty body

bdraco:
## What do these changes do?

There was a disagreement on how to handle 204/304/1xx responses between the c parser and py parser.

204/304/1xx are now always treated as an empty body in the py parser to match the c parser and comply with
https://datatracker.ietf.org/doc/html/rfc9112#section-6.3

A empty chunked response body `0\r\n\r\n` will no longer be read for 204, 203, 1xx responses when using the py parser. This matches the behavior of the c parser.

## Checklist

TODO: 

- [x] timeline
- [x] mypy
- [x] more functional testing

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -527,7 +527,7 @@ class ClientSession:
                     assert conn.protocol is not None
                     conn.protocol.set_response_params(
                         timer=timer,
-                        skip_payload=method_must_be_empty_body(method.upper()),
+                        skip_payload=method_must_be_empty_body(method),
                         read_until_eof=read_until_eof,
                         auto_decompress=auto_decompress,
                         read_timeout=real_timeout.sock_read,
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -81,6 +81,7 @@ from .helpers import (
     TimeoutHandle,
     ceil_timeout,
     get_env_proxy_for_url,
+    method_must_be_empty_body,
     sentinel,
     strip_auth_from_url,
 )
@@ -526,8 +527,7 @@ class ClientSession:
                     assert conn.protocol is not None
                     conn.protocol.set_response_params(
                         timer=timer,
-                        skip_payload=method.upper()
-                        in (hdrs.METH_CONNECT, hdrs.METH_HEAD),
+                        skip_payload=method_must_be_empty_body(method.upper()),
                         read_until_eof=read_until_eof,
                         auto_decompress=auto_decompress,
                         read_timeout=real_timeout.sock_read,
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -526,7 +526,8 @@ class ClientSession:
                     assert conn.protocol is not None
                     conn.protocol.set_response_params(
                         timer=timer,
-                        skip_payload=method.upper() == "HEAD",
+                        skip_payload=method.upper()
+                        in (hdrs.METH_CONNECT, hdrs.METH_HEAD),
                         read_until_eof=read_until_eof,
                         auto_decompress=auto_decompress,
                         read_timeout=real_timeout.sock_read,
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -345,10 +345,8 @@ class HttpParser(abc.ABC, Generic[_MsgT]):
                         # calculate payload
                         # 204, 304, 1xx should not have a body per
                         # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
-                        code_indicates_empty_body = (
-                            code in (204, 304) or 100 <= code < 200
-                        )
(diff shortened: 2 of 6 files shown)


aiohttp/http_parser.py
```diff
@@ -338,12 +338,19 @@ def get_content_length() -> Optional[int]:
...
                         # calculate payload
+                        # 204, 304, 1xx should not have a body per
+                        # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+                        code_indicates_empty_body = (
+                            code in (204, 304) or 100 <= code < 200
+                        )
                         if (
                             (length is not None and length > 0)
-                            or msg.chunked
+                            or (msg.chunked and not code_indicates_empty_body)
```
Dreamsorcerer: Shouldn't this override it even if length>0? i.e. It shouldn't be bracketed.

tests/test_http_parser.py
```diff
@@ -1060,6 +1061,131 @@ def test_parse_no_length_payload(parser: Any) -> None:
     assert payload.is_eof()
 
 
+def test_parse_content_length_payload_multiple(response: Any) -> None:
+    text = b"HTTP/1.1 200 OK\r\n" b"content-length: 5\r\n\r\n" b"first"
```
```suggestion
    text = b"HTTP/1.1 200 OK\r\ncontent-length: 5\r\n\r\nfirst"
```

Dreamsorcerer: And same for all the other ones...

bdraco: Adjusted to collapse string

aiohttp/http_parser.py
bdraco: Refactored to be included with the length check as well

bdraco:
https://github.com/aio-libs/aiohttp/issues/6446 is a bit of a problem since its a special case

bdraco:
the `msg.upgrade` check looks like it might have a problem

bdraco:
not 100% sure if the msg.upgrade check it supposed to apply to chunk and content length or just chunked given https://github.com/aio-libs/aiohttp/commit/6301d6b75f86f31096fb0212c57238b4604f02a1

bdraco:
I'm not sure its right but I made it `or (msg.chunked and not msg.upgrade)` so the check is the same.

If its not right it should probably be fixed in another PR as I didn't want to break compatibility.

bdraco:
I think this one is ready for another look.

If it looks good I'll start doing production testing

I'm still working through regressions in the other one

aiohttp/http_parser.py
```diff
@@ -383,34 +390,33 @@ def get_content_length() -> Optional[int]:
                                 auto_decompress=self._auto_decompress,
                                 lax=self.lax,
                             )
+                        elif (
+                            not code_indicates_empty_body
```
Dreamsorcerer: I'm thinking in both PRs, maybe `must_be_empty_body` feels a little clearer to me when reading the if statements.

Dreamsorcerer (approved):

aiohttp/http_parser.py
bdraco: adjusted in 68c5bc2fbceb39a04c7b7abeaa4509bdac989eea

aiohttp/http_parser.py
```diff
@@ -338,10 +338,19 @@ def get_content_length() -> Optional[int]:
...
                         assert self.protocol is not None
                         # calculate payload
-                        if (
+                        # 204, 304, 1xx should not have a body per
+                        # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+                        must_be_empty_body = (
+                            method in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
+                            or code in (204, 304)
+                            or 100 <= code < 200
+                        )
```
bdraco: Maybe move this to helpers.py so its a bit more DRY

aiohttp/client.py
```diff
@@ -526,7 +527,7 @@ async def _request(
                     assert conn.protocol is not None
                     conn.protocol.set_response_params(
                         timer=timer,
-                        skip_payload=method.upper() == "HEAD",
+                        skip_payload=method_must_be_empty_body(method.upper()),
```
Dreamsorcerer: Probably a good idea to put the .upper() in the helper?

Dreamsorcerer:
This is still marked as a draft, so remember to update that if you think it's ready to merge.

aiohttp/client.py
bdraco: adjusted in d1fb2c103a9be5683a93e3106e06247bca0b3c5e

bdraco:
> This is still marked as a draft, so remember to update that if you think it's ready to merge.

Will do.  Still doing functional testing, and need to write the timeline message

PR #7756, opened 2023-10-27
Ensure empty body response for 1xx/204/304 per RFC 9112 sec 6.3

bdraco:
## What do these changes do?

`Transfer-Encoding` and `Content-Length` will be removed from the server response when sending a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status since no body is expected for these status codes

- `Content-Length` forbidden on 1xx/204 per https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2 and discouraged on 304 per https://datatracker.ietf.org/doc/html/rfc7232#section-4.1

We need to ensure `Content-Length` is kept for `HEAD` to ensure keep-alive works as expected and is allowed per https://datatracker.ietf.org/doc/html/rfc2616#section-14.13

- `Transfer-Encoding` removed per  https://datatracker.ietf.org/doc/html/rfc9112#section-6.1 
> any recipient on the response chain (including the origin server) can remove transfer codings when they are not needed.

`aiohttp` would incorrectly send an body of `0\r\n\r\n` when `chunked` was set for `Transfer-Encoding` with the above status code. 

This change ensures `aiohttp` does not send these invalid bodies.

The `Transfer-Encoding` and `Content-Length` headers will be removed when sending a 1xx (Informational), 204 (No Content), or 304 (Not Modified) except for `HEAD` responses since these status since no body is expected for these status codes.

An invalid body of `0\r\n\r\n` will no longer be sent with these status codes.

## Checklist

TODO: 

- [x] figure out why client closes connection on HEAD requests with content length
- [x] figure out why client functional tests fail with AIOHTTP_NO_EXTENSIONS=1 `test_keepalive_after_empty_body_status` and `test_keepalive_after_empty_body_status_stream_response` -- They do not fail anymore after the fixed in https://github.com/aio-libs/aiohttp/pull/7755
- [x] timeline
- [x] more functional testing

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1063,3 +1063,22 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
             with suppress(ValueError):
                 return datetime.datetime(*timetuple[:6], tzinfo=datetime.timezone.utc)
     return None
+
+
+def must_be_empty_body(method: str, code: int) -> bool:
+    """Check if a request must return an empty body."""
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+    return status_code_must_be_empty_body(code) or method_must_be_empty_body(method)
+
+
+def method_must_be_empty_body(method: str) -> bool:
+    """Check if a method must return an empty body."""
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+    return method in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
+
+
+def status_code_must_be_empty_body(code: int) -> bool:
+    """Check if a status code must return an empty body."""
+    # 204, 304, 1xx should not have a body per
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+    return code in (204, 304) or 100 <= code < 200
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1067,7 +1067,6 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
 
 def must_be_empty_body(method: str, code: int) -> bool:
     """Check if a request must return an empty body."""
-    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
     return status_code_must_be_empty_body(code) or method_must_be_empty_body(method)
 
 
@@ -1089,10 +1088,6 @@ def should_remove_content_length(method: str, code: int) -> bool:
 
     This should always be a subset of must_be_empty_body
     """
-    # forbidden on 1xx/204/CONNECT per:
-    # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
-    # discouraged on 304 per
-    # https://datatracker.ietf.org/doc/html/rfc7232#section-4.1
-    # HEAD requests should still include the content-length header
-    # per https://datatracker.ietf.org/doc/html/rfc2616#section-14.13
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-8
     return code in (204, 304) or 100 <= code < 200 or method == hdrs.METH_CONNECT
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1084,8 +1084,11 @@ def status_code_must_be_empty_body(code: int) -> bool:
     return code in (204, 304) or 100 <= code < 200
 
 
-def should_remove_content_length(code: int, method: str) -> bool:
-    """Check if a Content-Length header should be removed."""
+def should_remove_content_length(method: str, code: int) -> bool:
+    """Check if a Content-Length header should be removed.
+
+    This should always be a subset of must_be_empty_body
+    """
     # forbidden on 1xx/204/CONNECT per:
     # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
     # discouraged on 304 per
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1082,3 +1082,14 @@ def status_code_must_be_empty_body(code: int) -> bool:
     # 204, 304, 1xx should not have a body per
     # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
     return code in (204, 304) or 100 <= code < 200
+
+
+def should_remove_content_length(code: int, method: str) -> bool:
+    """Check if a Content-Length header should be removed."""
+    # forbidden on 1xx/204/CONNECT per:
+    # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
+    # discouraged on 304 per
+    # https://datatracker.ietf.org/doc/html/rfc7232#section-4.1
+    # HEAD requests should still include the content-length header
+    # per https://datatracker.ietf.org/doc/html/rfc2616#section-14.13
+    return code in (204, 304) or 100 <= code < 200 or method == hdrs.METH_CONNECT
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1070,7 +1070,7 @@ def must_be_empty_body(method: str, code: int) -> bool:
     return (
         status_code_must_be_empty_body(code)
         or method_must_be_empty_body(method)
-        or (code == 200 and method.upper() == hdrs.METH_CONNECT)
+        or (200 <= code < 300 and method.upper() == hdrs.METH_CONNECT)
     )
 
 
@@ -1097,5 +1097,5 @@ def should_remove_content_length(method: str, code: int) -> bool:
     return (
         code in {204, 304}
         or 100 <= code < 200
-        or (code == 200 and method.upper() == hdrs.METH_CONNECT)
+        or (200 <= code < 300 and method.upper() == hdrs.METH_CONNECT)
     )
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1092,8 +1092,8 @@ def should_remove_content_length(method: str, code: int) -> bool:
 
     This should always be a subset of must_be_empty_body
     """
-    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7
     # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-8
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-15.4.5-4
     return (
         code in {204, 304}
         or 100 <= code < 200
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1067,14 +1067,18 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
 
 def must_be_empty_body(method: str, code: int) -> bool:
     """Check if a request must return an empty body."""
-    return status_code_must_be_empty_body(code) or method_must_be_empty_body(method)
+    return (
+        status_code_must_be_empty_body(code)
+        or method_must_be_empty_body(method)
+        or (code == 200 and method.upper() == hdrs.METH_CONNECT)
+    )
 
 
 def method_must_be_empty_body(method: str) -> bool:
     """Check if a method must return an empty body."""
     # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3-2.1
     # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3-2.2
-    return method.upper() in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
+    return method.upper() == hdrs.METH_HEAD
 
 
 def status_code_must_be_empty_body(code: int) -> bool:
@@ -1090,4 +1094,8 @@ def should_remove_content_length(method: str, code: int) -> bool:
     """
     # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7
     # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-8
-    return code in (204, 304) or 100 <= code < 200 or method == hdrs.METH_CONNECT
+    return (
+        code in {204, 304}
+        or 100 <= code < 200
+        or (code == 200 and method.upper() == hdrs.METH_CONNECT)
+    )
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -1073,15 +1073,16 @@ def must_be_empty_body(method: str, code: int) -> bool:
 
 def method_must_be_empty_body(method: str) -> bool:
     """Check if a method must return an empty body."""
-    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
-    return method in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3-2.1
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3-2.2
+    return method.upper() in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
 
 
 def status_code_must_be_empty_body(code: int) -> bool:
     """Check if a status code must return an empty body."""
     # 204, 304, 1xx should not have a body per
-    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
-    return code in (204, 304) or 100 <= code < 200
+    # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3-2.1
+    return code in {204, 304} or 100 <= code < 200
 
 
 def should_remove_content_length(method: str, code: int) -> bool:
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -16,7 +16,7 @@ from typing import (
 
 from . import hdrs
 from .abc import AbstractStreamWriter
-from .helpers import ETAG_ANY, ETag
+from .helpers import ETAG_ANY, ETag, must_be_empty_body
 from .typedefs import LooseHeaders, PathLike
 from .web_exceptions import (
     HTTPNotModified,
@@ -266,12 +266,7 @@ class FileResponse(StreamResponse):
             )
 
         # If we are sending 0 bytes calling sendfile() will throw a ValueError
-        if (
-            count == 0
-            or request.method in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
-            or self.status in (204, 304)
-            or 100 <= self.status < 200
-        ):
+        if count == 0 or must_be_empty_body(request.method, self.status):
             return await super().prepare(request)
 
         fobj = await loop.run_in_executor(None, filepath.open, "rb")
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -268,7 +268,7 @@ class FileResponse(StreamResponse):
         # If we are sending 0 bytes calling sendfile() will throw a ValueError
         if (
             count == 0
-            or request.method in (hdrs.METH_CONNECT or hdrs.METH_HEAD)
+            or request.method in (hdrs.METH_CONNECT, hdrs.METH_HEAD)
             or self.status in (204, 304)
             or 100 <= self.status < 200
         ):
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -266,7 +266,12 @@ class FileResponse(StreamResponse):
             )
 
         # If we are sending 0 bytes calling sendfile() will throw a ValueError
-        if count == 0 or request.method == hdrs.METH_HEAD or self.status in [204, 304]:
+        if (
+            count == 0
+            or request.method in (hdrs.METH_CONNECT or hdrs.METH_HEAD)
+            or self.status in (204, 304)
+            or 100 <= self.status < 200
+        ):
             return await super().prepare(request)
 
         fobj = await loop.run_in_executor(None, filepath.open, "rb")
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -35,6 +35,7 @@ from .helpers import (
     populate_with_cookies,
     rfc822_formatted_time,
     sentinel,
+    should_remove_content_length,
     validate_etag_value,
 )
 from .http import SERVER_SOFTWARE, HttpVersion10, HttpVersion11
@@ -393,9 +394,9 @@ class StreamResponse(BaseClass, HeadersMixin, CookieMixin):
         if self._must_be_empty_body:
             # We need to remove the content-length for empty body
             # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
-            # HEAD requests should still include the content-length header
-            # per https://datatracker.ietf.org/doc/html/rfc2616#section-14.13
-            if hdrs.CONTENT_LENGTH in headers and request.method != hdrs.METH_HEAD:
+            if hdrs.CONTENT_LENGTH in headers and should_remove_content_length(
+                self.status, request.method
+            ):
                 del headers[hdrs.CONTENT_LENGTH]
             # remove transfer codings when they are not needed
             # per https://datatracker.ietf.org/doc/html/rfc9112#section-6.1
@@ -675,9 +676,9 @@ class Response(StreamResponse):
             await super().write_eof()
 
     async def _start(self, request: "BaseRequest") -> AbstractStreamWriter:
-        # HEAD requests should still include the content-length header
-        # per https://datatracker.ietf.org/doc/html/rfc2616#section-14.13
-        if self._must_be_empty_body and request.method != hdrs.METH_HEAD:
+        if self._must_be_empty_body and should_remove_content_length(
+            self.status, request.method
+        ):
             # We need to remove the content-length for empty body
             # https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
             if hdrs.CONTENT_LENGTH in self._headers:
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -378,27 +378,24 @@ class StreamResponse(BaseClass, HeadersMixin, CookieMixin):
             self.status in (204, 304) or 100 <= self.status < 200
         )
 
-        if status_code_indicates_empty_body:
-            if hdrs.TRANSFER_ENCODING in headers:
-                del headers[hdrs.TRANSFER_ENCODING]
-            if hdrs.CONTENT_LENGTH in headers:
-                del headers[hdrs.CONTENT_LENGTH]
-        elif self._chunked:
+        if self._chunked:
             if version != HttpVersion11:
                 raise RuntimeError(
                     "Using chunked encoding is forbidden "
                     "for HTTP/{0.major}.{0.minor}".format(request.version)
                 )
-            writer.enable_chunking()
-            headers[hdrs.TRANSFER_ENCODING] = "chunked"
+            if not status_code_indicates_empty_body:
+                writer.enable_chunking()
+                headers[hdrs.TRANSFER_ENCODING] = "chunked"
             if hdrs.CONTENT_LENGTH in headers:
                 del headers[hdrs.CONTENT_LENGTH]
         elif self._length_check:
             writer.length = self.content_length
             if writer.length is None:
-                if version >= HttpVersion11 and self.status != 204:
-                    writer.enable_chunking()
-                    headers[hdrs.TRANSFER_ENCODING] = "chunked"
+                if version >= HttpVersion11:
+                    if not status_code_indicates_empty_body:
+                        writer.enable_chunking()
+                        headers[hdrs.TRANSFER_ENCODING] = "chunked"
                     if hdrs.CONTENT_LENGTH in headers:
                         del headers[hdrs.CONTENT_LENGTH]
                 else:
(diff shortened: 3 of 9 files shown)


aiohttp/web_response.py
```diff
@@ -364,7 +364,25 @@ async def _prepare_headers(self) -> None:
...
+            # a 1xx (Informational), 204 (No Content), or 304
+            # (Not Modified) status code is always terminated by
+            # the first empty line after the header fields,
+            # regardless of the header fields present in the message,
+            # and thus cannot contain a message body or trailer section.
+            #
+            if hdrs.TRANSFER_ENCODING in headers:
+                del headers[hdrs.TRANSFER_ENCODING]
+            if hdrs.CONTENT_LENGTH in headers:
+                del headers[hdrs.CONTENT_LENGTH]
```
Dreamsorcerer: OK, running through the spec and what code we have here, I think we'll need to rework this a bit.

At line 407, it correctly removes the Content-Length header if needed already (though could do with checking the full range of 1xx codes).

At line 391, I think we just want to make enable_chunking() conditional on the status code. Then pretty much the same as line 407, delete the Transfer-Encoding header for the given status codes (maybe this part can be moved out a level and deduplicated).

This is also supposed to be done for a CONNECT request, regardless of status code. Not sure if we have any code for that (or if the method is available at this point in the code).

We should probably also make sure that any body provided by the user for these status codes results in an exception, if we don't do this already.

Dreamsorcerer (changes requested):

tests/test_web_response.py
```diff
@@ -624,6 +624,23 @@ async def write_headers(status_line, headers):
...
+@pytest.mark.parametrize("status", (100, 101, 204, 304))
+async def test_rm_transfer_encoding_rfc_9112_6_3_http_11(status: int) -> None:
+    """Remove transfer encoding for RFC 9112 sec 6.3 with HTTP/1.1."""
+    writer = mock.Mock()
+
+    async def write_headers(status_line, headers):
+        assert hdrs.CONTENT_LENGTH not in headers
+        assert hdrs.TRANSFER_ENCODING not in headers
+
+    writer.write_headers.side_effect = write_headers
```
Dreamsorcerer: I feel like we should be able to check this on the resp object normally?

But, either way, this check shouldn't be done for 304.

Dreamsorcerer:
Thanks for digging into this, I didn't realise from the original issue that it was a response from aiohttp that was causing the problem.

aiohttp/web_response.py
Dreamsorcerer: A quick test suggests we don't do the last 2 points currently.

The body is probably a particular issue. If we have a keep-alive connection and we send a body, a conforming client should stop before reading the body, and will then try to interpret the body as a new HTTP message.

bdraco:
Probably could use a test to check the raw body from a 304 with chunked pass through as well for the proxy case so it doesn't accidentally add the bad body

aiohttp/web_response.py
bdraco: TODO:

- [x] At line 407, it correctly removes the Content-Length header if needed already (though could do with checking the full range of 1xx codes).

Adjusted in dce3e25f8cc2cce3a98a423690c9b52a0d9ae28b

- [x] At line 391, I think we just want to make enable_chunking() conditional on the status code. Then pretty much the same as line 407, delete the Transfer-Encoding header for the given status codes (maybe this part can be moved out a level and deduplicated).

Adjusted in e3db2b88cff525fc20cc6399c79fd1cfd7827234

- [x] This is also supposed to be done for a CONNECT request, regardless of status code. Not sure if we have any code for that (or if the method is available at this point in the code).

I think that was supposed to be for `HEAD` as well??

Adjusted in 3fb3f7958aa3c0e48883f2e4c11acb096e4cacae

tests/test_web_response.py
bdraco: > I feel like we should be able to check this on the resp object normally?

Adjusted in 8f9f7527c455863e254fc7e3c7938f84dedee278 Kept the mock since it needs to be an async mock

> But, either way, this check shouldn't be done for 304.

https://github.com/nodejs/llhttp/pull/256#issuecomment-1782909852
> 
> > Transfer-Encoding MAY be sent in a response to a HEAD request or in a 304 (Not Modified) response ([Section 15.4.5](https://www.rfc-editor.org/rfc/rfc9110#section-15.4.5) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]) to a GET request, neither of which includes a message body, to indicate that the origin server would have applied a transfer coding to the message body if the request had been an unconditional GET. This indication is not required, however, because any recipient on the response chain (including the origin server) can remove transfer codings when they are not needed.

> It is allowed to send transfer-encoding header field in 304 response, but it must be ignored because the message body must be empty.

I think we should strip it out since its allowed to be ignored and because it will confuse the older pure python and pre llhttp 9.x clients if we don't

> This indication is not required, however, because any recipient on the response chain (including the origin server) can remove transfer codings when they are not needed.

https://datatracker.ietf.org/doc/html/rfc9112#section-6.1

aiohttp/web_response.py
bdraco: >  We should probably also make sure that any body provided by the user for these status codes results in an exception, if we don't do this already.

That seems like it might be better in another PR.

bdraco:
Need to figure out how I regressed

tests/test_client_functional.py::test_bad_payload_chunked_encoding[pyloop]

bdraco:
So it looks like if `writer.enable_chunking()` isn't called which is now the case for `HEAD` it won't keep the connection alive...

need to figure that out later though as out of time

bdraco:
Which also means we need keep-alive tests for 304/204/1xx

tests/test_web_response.py
Dreamsorcerer: Hmm, I guess it's a problem if we have clients that trip up on such a message. But, ideally, it could be used to give additional information about the response... I think the Content-Length header is probably more useful than the Transfer-Encoding header though.

aiohttp/web_response.py
Dreamsorcerer: > > We should probably also make sure that any body provided by the user for these status codes results in an exception, if we don't do this already.
> 
> That seems like it might be better in another PR.

Up to you, but the title of the PR says that it ensures there is an empty body...

aiohttp/web_response.py
```diff
@@ -364,32 +364,52 @@ async def _prepare_headers(self) -> None:
         if self._compression:
             await self._start_compression(request)
 
+        # The following should not have a body per
+        # https://datatracker.ietf.org/doc/html/rfc9112#section-6.3
+        # status code: 204, 304, 1xx
+        # method: HEAD
+        request_indicates_empty_body = (
```
Dreamsorcerer: Should we be doing this on a CONNECT 2xx response as well?

Dreamsorcerer: I think from my reading, Transfer-Encoding is forbidden, and Content-Length will be the same as 304 (so, if we're removing it for 304 we should remove it here too).

aiohttp/web_response.py
```diff
@@ -364,32 +364,52 @@ async def _prepare_headers(self) -> None:
...
                     keep_alive = False
             # HTTP 1.1: https://tools.ietf.org/html/rfc7230#section-3.3.2
             # HTTP 1.0: https://tools.ietf.org/html/rfc1945#section-10.4
-            elif version >= HttpVersion11 and self.status in (100, 101, 102, 103, 204):
+            elif version >= HttpVersion11 and request_indicates_empty_body:
                 del headers[hdrs.CONTENT_LENGTH]
+                # remove transfer codings when they are not needed
+                # per https://datatracker.ietf.org/doc/html/rfc9112#section-6.1
+                if hdrs.TRANSFER_ENCODING in headers:
+                    del headers[hdrs.TRANSFER_ENCODING]
```
Dreamsorcerer: Do you reckon we can just move this out a level and remove the duplicate check at lines 386-389?

```suggestion

        if version >= HttpVersion11 and request_indicates_empty_body:
            # remove transfer codings when they are not needed
            # per https://datatracker.ietf.org/doc/html/rfc9112#section-6.1
            if hdrs.CONTENT_LENGTH in headers:
                del headers[hdrs.CONTENT_LENGTH]
            if hdrs.TRANSFER_ENCODING in headers:
                del headers[hdrs.TRANSFER_ENCODING]
```

tests/test_web_response.py
Dreamsorcerer: > Adjusted in [8f9f752](https://github.com/aio-libs/aiohttp/commit/8f9f7527c455863e254fc7e3c7938f84dedee278) Kept the mock since it needs to be an async mock

Would be better if you can successfully create the mock with `mock.create_autospec(Writer, spec_set=True, instance=True)`

bdraco:
There is still another issue here because in Response we also force content length "0" when content length is missing and the body is empty. This is probably for clients that don't handle implicit empty body well.  We might need to do that in StreamResponse as well.  Also the rfc for http 1.0 says content length 0 is invalid ... but in practice it might be better to keep it as is as I expect many clients actually rely on that from experience 

Also there doesn't appear to be adequate coverage for validating how things behave with StreamResponse vs Response as well so might need to add that first and than back track and fix any discrepancies 

Will need to reconcile that before addressing review comments

aiohttp/web_response.py
bdraco: I think so.  Once I reconcile the other issue I mentioned below I'll do some testing with the suggested change

Dreamsorcerer:
> There is still another issue here because in Response we also force content length "0" when content length is missing and the body is empty. This is probably for clients that don't handle implicit empty body well.

I think that is needed for the client to know where the message ends. The only exception where it would be acceptable to have no content-length header is when the body must be empty (or chunked is being used):
https://www.rfc-editor.org/rfc/rfc9110#section-8.6-5

> Also the rfc for http 1.0 says content length 0 is invalid

It says "Any Content-Length greater than or **equal** to zero is a valid value."

bdraco:
> It says "Any Content-Length greater than or **equal** to zero is a valid value."

Right on, I mentally dropped the equal when I read that

aiohttp/web_response.py
bdraco: done in 5b557b983ed07c6ebec2c0dfbcd688b8d0ae6c8b

aiohttp/web_response.py
bdraco: I think that works but I don't have a good test case for this

04e6cfaf80c862ed7c805e61ffb60b88e6ad3d9f

tests/test_web_response.py
bdraco: b8d5991dea9f1754ae8f5d304692aecdd0e987c6

aiohttp/web_response.py
bdraco: It will be a bit more refactoring but I can do it here as well

bdraco: I'm not sure we can do that without a breaking change as these tests fail when I do

```
FAILED tests/test_client_functional.py::test_HTTP_304_WITH_BODY[pyloop] - aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
FAILED tests/test_client_functional.py::test_HTTP_200_OK_METHOD[pyloop] - aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
FAILED tests/test_client_functional.py::test_HTTP_200_OK_METHOD_connector[pyloop] - aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
FAILED tests/test_client_functional.py::test_HTTP_302_REDIRECT_HEAD[pyloop] - aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
FAILED tests/test_client_functional.py::test_shortcuts[pyloop] - aiohttp.client_exceptions.ServerDisconnectedError: Server disconnected
```

aiohttp/web_response.py
```diff
@@ -370,27 +380,36 @@ async def _prepare_headers(self) -> None:
...
 
-        if self.status not in (204, 304):
+        # HTTP 1.1: https://tools.ietf.org/html/rfc7230#section-3.3.2
+        # HTTP 1.0: https://tools.ietf.org/html/rfc1945#section-10.4
+        if self._must_be_empty_body:
+            # If the content-length is not set to "0" we will
+            # prematurely close a keep-alive connection that could
+            # have been reused. This matches the behavior of Response
+            # which would also set the content-length to "0" in _start
+            headers[hdrs.CONTENT_LENGTH] = "0"
```
bdraco: If I delete this head instead of setting it to zero the `test_keepalive_after_head_requests_success` test fails.

The test wasn't actually mocking `HEAD` before so it was actually returning an error.

We set "0" in `Response` for empty bodies already so I'm pretty sure this should be safe but it needs some 👀 to be sure.  I _think_ this behavior is safer, especially for clients that don't know about implicit empty body from 1xx/204/304 as the 0 should make them do the right thing

bdraco: https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2

> A server MAY send a Content-Length header field in a response to a
   HEAD request ([Section 4.3.2 of [RFC7231]](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2)); a server MUST NOT send
   Content-Length in such a response unless its field-value equals the
   decimal number of octets that would have been sent in the payload
   body of a response if the same request had used the GET method.


So it looks like we have to delete this

bdraco:
test_keepalive_after_head_requests_success is still failing but I think everything else will pass now

bdraco:
Its failing because the waiter is being cancelled in RequestHandler.start which breaks out of the loop

bdraco:
Its because connection_lost if called. I think the client is dropping the connection on the head response so there might be a bug reading head responses w/keep alive in the client

bdraco:
Out of time to dig for the client side issue so will have to pick this up later

bdraco:
The Content-Length entity-header field indicates the size of the entity-body, in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD method, the size of the entity-body that would have been sent had the request been a GET.



So for `HEAD` we should still send `Content-Length`

bdraco:
For 204/304 we need to ensure we don't send content-length.. looks like this has tripped up other projects as well https://github.com/heroku/cowboyku/issues/9

bdraco:
The tests are failing because of the differences between the c and py parsers which are fixed in https://github.com/aio-libs/aiohttp/pull/7755

bdraco:
That's all the time I have for this tonight.

Lots of functional testing tomorrow/monday left to do

bdraco:
I've looked it over and read everything related in the RFCs again so I think its good to start doing more functional testing with.

I can't think of any more tests to add at this point, but happy to add some more cases if there is something that stands out

aiohttp/web_response.py
Dreamsorcerer: Ah, OK, looking at those tests, it does ensure we end up with an empty body, even if a non-empty body is passed in. So, I guess we can leave it like that... Could maybe just tweak one of your tests (if it's not already done) to include a body as input, and assert there is no body as output (to ensure we cover all the status codes).

aiohttp/web_response.py
Dreamsorcerer: If you think it improves compatibility with keep-alive clients, it can stay there when it's not a HEAD request. It can't be done for a HEAD request though, as the client may be using the HEAD request to figure out how big the response will be.

Dreamsorcerer:
> So for `HEAD` we should still send `Content-Length`

If the user has set it, yes (or we have some other way of determining the size of the GET payload).

aiohttp/web_response.py
Dreamsorcerer: Ah, actually, 304 is the same. Should be left if present, but shouldn't be set to 0.

It appears that 1xx, 204 are actually forbidden to include the header at all:
https://datatracker.ietf.org/doc/html/rfc9110#section-8.6-8

So, yeah, that will need to go.

aiohttp/web_response.py
bdraco: I think its covered by the keep alive tests but I added an explicit one as well

f9e831a0bb1248537852be21b97894e3d21285ce

aiohttp/helpers.py
```diff
@@ -1076,3 +1081,13 @@ def status_code_must_be_empty_body(code: int) -> bool:
...
+
+
+def should_remove_content_length(method: str, code: int) -> bool:
+    """Check if a Content-Length header should be removed.
+
+    This should always be a subset of must_be_empty_body
+    """
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-8
+    return code in (204, 304) or 100 <= code < 200 or method == hdrs.METH_CONNECT
```
Dreamsorcerer: For CONNECT requests, it only applies if the response is 2xx, so I think there's a small mistake here.

Also, the wording for HEAD and 304 are basically identical, so I'm not 100% sure we should include 1 here and not the other. I think we really just need to ensure that any automatic Content-Length header is excluded in both cases.

aiohttp/web_fileresponse.py
```diff
@@ -266,7 +266,7 @@ async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter
             )
 
         # If we are sending 0 bytes calling sendfile() will throw a ValueError
-        if count == 0 or request.method == hdrs.METH_HEAD or self.status in [204, 304]:
+        if count == 0 or must_be_empty_body(request.method, self.status):
```
Dreamsorcerer, line 269: If the problem here is sending 0 bytes, then that'd include both 304 and HEAD, right?

Dreamsorcerer: Ignore that one, mixing up the methods.

aiohttp/helpers.py
Dreamsorcerer: Just realised the 2xx CONNECT is also an issue with the previously merged functions.

aiohttp/web_response.py
```diff
@@ -370,26 +374,33 @@ async def _prepare_headers(self) -> None:
...
+        # HTTP 1.1: https://tools.ietf.org/html/rfc7230#section-3.3.2
+        # HTTP 1.0: https://tools.ietf.org/html/rfc1945#section-10.4
+        if self._must_be_empty_body:
+            if hdrs.CONTENT_LENGTH in headers and should_remove_content_length(
+                request.method, self.status
+            ):
+                del headers[hdrs.CONTENT_LENGTH]
+            # remove transfer codings when they are not needed
+            # per https://datatracker.ietf.org/doc/html/rfc9112#section-6.1
+            if version >= HttpVersion11 and hdrs.TRANSFER_ENCODING in headers:
```
Dreamsorcerer:
```suggestion
            # https://datatracker.ietf.org/doc/html/rfc9112#section-6.1-10
            # https://datatracker.ietf.org/doc/html/rfc9112#section-6.1-13
            if hdrs.TRANSFER_ENCODING in headers:
```
The second link says not to send Transfer-Encoding at all for HTTP/1.0.

aiohttp/web_response.py
```diff
@@ -662,7 +673,12 @@ async def write_eof(self, data: bytes = b"") -> None:
             await super().write_eof()
 
     async def _start(self, request: "BaseRequest") -> AbstractStreamWriter:
-        if not self._chunked and hdrs.CONTENT_LENGTH not in self._headers:
+        if self._must_be_empty_body and should_remove_content_length(
+            request.method, self.status
+        ):
+            if hdrs.CONTENT_LENGTH in self._headers:
+                del self._headers[hdrs.CONTENT_LENGTH]
+        elif not self._chunked and hdrs.CONTENT_LENGTH not in self._headers:
```
Dreamsorcerer, line 679: So, this part can't be run if HEAD or 304, as the Content-Length header would be incorrect.

aiohttp/web_response.py
```diff
@@ -662,7 +673,12 @@ async def write_eof(self, data: bytes = b"") -> None:
             await super().write_eof()
 
     async def _start(self, request: "BaseRequest") -> AbstractStreamWriter:
-        if not self._chunked and hdrs.CONTENT_LENGTH not in self._headers:
+        if self._must_be_empty_body and should_remove_content_length(
```
Dreamsorcerer: Isn't the first check redundant, as a superset of the second?

tests/test_client_functional.py
```diff
@@ -82,6 +83,77 @@ async def on_reuseconn(session, ctx, params):
     await resp2.read()
 
     assert 1 == cnt_conn_reuse
+    await client.close()
```
Dreamsorcerer: This shouldn't be needed, our fixtures handle closing the client.

aiohttp/helpers.py
bdraco: I will prepare another PR to fix that

aiohttp/web_response.py
bdraco: agreed. removed in 8f7c0e2840ced7937c0b749db68dcdcf43599140

tests/test_client_functional.py
bdraco: Will validate and remove

bdraco: removed in a7dc41ce44c4c735c6a98234ab2c072c95e7d33a

aiohttp/helpers.py
bdraco: https://github.com/aio-libs/aiohttp/pull/7772

I will adjust the logic in this PR shortly

bdraco: logic adjusted in 5630ca593341df97d5551d3dd175163afcbc05d6

aiohttp/web_response.py
bdraco: I think its ok to run on HEAD

> A server MAY send a Content-Length header field in a response to a HEAD request ([Section 9.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#HEAD)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a response if the same request had used the GET method.

bdraco: I think its ok for 304 as well

> A server MAY send a Content-Length header field in a [304 (Not Modified)](https://www.rfc-editor.org/rfc/rfc9110.html#status.304) response to a conditional GET request ([Section 15.4.5](https://www.rfc-editor.org/rfc/rfc9110.html#status.304)); a server MUST NOT send Content-Length in such a response unless its field value equals the decimal number of octets that would have been sent in the content of a [200 (OK)](https://www.rfc-editor.org/rfc/rfc9110.html#status.200) response to the same request.

bdraco: https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7

aiohttp/helpers.py
```diff
@@ -1065,14 +1065,37 @@ def parse_http_date(date_str: Optional[str]) -> Optional[datetime.datetime]:
     return None
 
 
+def must_be_empty_body(method: str, code: int) -> bool:
+    """Check if a request must return an empty body."""
+    return (
+        status_code_must_be_empty_body(code)
+        or method_must_be_empty_body(method)
+        or (code == 200 and method.upper() == hdrs.METH_CONNECT)
```
Dreamsorcerer: I believe it was anything in the 2xx range, not just 200 itself.

aiohttp/helpers.py
bdraco: Tests adjusted in 79d55e28b248c48b90b53eef0d0239fdc1d17728

aiohttp/helpers.py
bdraco: Will hunt for it in the RFC and adjust as needed as well as the other PR

bdraco: > A server MUST NOT send a Content-Length header field in any response with a status code of [1xx (Informational)](https://www.rfc-editor.org/rfc/rfc9110.html#status.1xx) or [204 (No Content)](https://www.rfc-editor.org/rfc/rfc9110.html#status.204). A server MUST NOT send a Content-Length header field in any [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110.html#status.2xx) response to a CONNECT request ([Section 9.3.6](https://www.rfc-editor.org/rfc/rfc9110.html#CONNECT)).

Yes. 2xx 👍

bdraco: adjusted in 286a1084fd75344bda821116d0109e152dcaf6d5

aiohttp/web_response.py
Dreamsorcerer: Hmm, I think if the HEAD request was handled automatically by aiohttp, then it will be fine, as it just treats it as a GET request and will have included the entire body of the response (which will then be stripped later). We should ensure there is a test that verifies the HEAD request doesn't receive a 0 Content-Length for this case.

But, if a user manually handles a HEAD request, then this behaviour is incorrect, as they may return an empty body for the HEAD request, but not for a GET request. I don't know if we have an easy way to tell if this is from a user handler or an automatic aiohttp one...

bdraco: > Hmm, I think if the HEAD request was handled automatically by aiohttp, then it will be fine, as it just treats it as a GET request and will have included the entire body of the response (which will then be stripped later). We should ensure there is a test that verifies the HEAD request doesn't receive a 0 Content-Length for this case.

I think this is covered by `test_head_response_keeps_content_length_of_original_body`
> 
> But, if a user manually handles a HEAD request, then this behaviour is incorrect, as they may return an empty body for the HEAD request, but not for a GET request. I don't know if we have an easy way to tell if this is from a user handler or an automatic aiohttp one...

Need to think about how to handle this case

Dreamsorcerer: > I think this is covered by `test_head_response_keeps_content_length_of_original_body`

Yes, and I can see in test_web_functional.py we have a HEAD test there: test_head_returns_empty_body
Would be ideal to just add 1 more assert to that test for the Content-Length too.

Dreamsorcerer: > I think its ok for 304 as well

I think as per the above discussion, we can't do this for 304, as it will always be a user-supplied response. So, it should skip that step in all cases of 304, I believe.

Dreamsorcerer: The user can obviously set Content-Length themselves if they want to include that information.

bdraco: adjusted for head e851e0abe65720fe72f3533bca36f7dab9ef268f

Will adjust `test_head_returns_empty_body` in another commit

bdraco: adjusted `test_head_returns_empty_body` 02f22fa3626fd1c1e80003845902d1c38548837f

bdraco: > I think as per the above discussion, we can't do this for 304, as it will always be a user-supplied response. So, it should skip that step in all cases of 304, I believe.

I should probably do that in `_prepare_headers` instead... will adjust

bdraco: oh, its already done there

bdraco: `test_rm_content_length_1xx_204_304_responses` covers the 304

Dreamsorcerer: > oh, its already done there

That's because the helper function is still checking 304, right? The spec treats it the same as HEAD requests. So, basically don't delete the header if the user set it explicitly, but also don't add the header automatically if they didn't.

I guess the HEAD request case, we can go with the check of whether the body is set, as you've done (but, that does still leave a small chance for issues if the user passes a body by mistake, but that's probably the best we'll get for now).

bdraco: I made it a bit safer for 304 in cccbdb39466acd4a59903cda069d91adc1a47db1 so its only set if we actually know the length of the body

aiohttp/helpers.py
```diff
@@ -1076,3 +1085,17 @@ def status_code_must_be_empty_body(code: int) -> bool:
...
+
+def should_remove_content_length(method: str, code: int) -> bool:
+    """Check if a Content-Length header should be removed.
+
+    This should always be a subset of must_be_empty_body
+    """
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-7
+    # https://www.rfc-editor.org/rfc/rfc9110.html#section-8.6-8
+    return (
+        code in {204, 304}
```
Dreamsorcerer, line 1098: 304 should still be dropped from here, right?
If the user sets the header, we should keep it.

aiohttp/web_response.py
Dreamsorcerer: I had a quick look to see if we could get back to the handlers/urldispatcher (as we could then check if the same handler is registered for GET/HEAD), but looks like there's no path to it. So, will stick with what you've done.

aiohttp/helpers.py
bdraco: I'm leaning towards it should still be removed since `Content-Length` is discouraged (SHOULD NOT) for 304 since its not one of `Cache-Control`, `Content-Location`, `Date`, `ETag`, `Expires`, and `Vary`.

If you feel strongly otherwise, I can adjust the behavior to keep it in, adjust the tests, and add a test for keeping it.

https://datatracker.ietf.org/doc/html/rfc7232#section-4.1

> The server generating a 304 response MUST generate any of the
> following header fields that would have been sent in a 200 (OK)
> response to the same request: Cache-Control, Content-Location, Date,
> ETag, Expires, and Vary.

> Since the goal of a 304 response is to minimize information transfer
> when the recipient already has one or more cached representations, a
> sender SHOULD NOT generate representation metadata other than the
> above listed fields unless said metadata exists for the purpose of
> guiding cache updates (e.g., Last-Modified might be useful if the
> response does not have an ETag field).

Dreamsorcerer: I'll come back tomorrow to give this a last check, but that sounds reasonable. They should really have linked to that paragraph from the other section that makes it sound as if it should be treated the same as HEAD.

bdraco:
I'll get the backport done shortly

bdraco:
Thanks @Dreamsorcerer

PR #5238, opened 2020-11-14
[WIP] Complete chunked request on server eof

greshilov:
## What do these changes do?

Possible fix #5220  

Code is hacky and questionable, hope maintainers will help.

## Are there changes in behavior for the user?

No.

## Related issue number

#5220 

## Checklist

- [ ] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #7764, opened 2023-10-29
Fix missing eof when writer cancelled

### LINKED ISSUE #5220: Server returns 400: "invalid character in chunk size header" for valid request

Dreamsorcerer:
Fixes #5220.

I believe this is a better fix than #5238. That PR detects that we didn't finish sending a chunked response and then closes the connection. This PR ensures that we simply complete the chunked response by sending the EOF bytes, allowing the connection to remain open and be reused normally.


Dreamsorcerer:
@webknjaz Would be good if you can check this one over. The changes became more invasive than I originally imagined.

Basically, to fix the reported issue, I've changed the code to send the end bytes for a chunked request when the writer is cancelled. But, the method that does the cancellation is not async, so it can't wait for the writer to complete. To accommodate that, I've had to make all the exit/release code wait on the writer, to ensure it's finished writing before the connection can be reused.

I've had to make several small changes throughout the code in order to get the tests passing and maintain compatibility. The only behaviour change that can be seen in the tests, is that after a simple `await sess.get(...)` with a small response from the server, you'll find the response is not closed and connection not released immediately on the next line of code (as there's nothing to hook into to wait on the release). I don't believe anything in our docs has ever guaranteed that a connection would be released at this point, so I think the change is fine (the couple of affected tests I've just changed to use `async with`). We can still guarantee that the connection is released immediately after `async with` (our most recommended approach), after `.close()`, and after `.read()` (which I think our docs do say will release the connection).

Dreamsorcerer:
I'm going to merge this one too, would be great to get a review in before the final 3.9 release though.

PR #7773, opened 2023-10-31
Fix usage of proxy.py in test_proxy_functional

bdraco:
This test was running even if tls in tls was supported without the need to monkey patch because the failures were actually problems with usage of `proxy.py`.

The test would fail because of problems with usage of `proxy.py`, and not because tls in tls didn't work.  This one turned out to have a lot of side trips to discover why things failed because proxy.py doesn't expect to run inside pytest, and is not easy to get error reporting out of when run this way.

- On all systems:  https://github.com/abhinavsingh/proxy.py/blob/30574fd0414005dfa8792a6e797023e862bdcf43/proxy/common/utils.py#L226 would raise on warning causing the proxy to die off because of `DeprecationWarning: ssl.OP_NO_SSL*/ssl.OP_NO_TLS* options are deprecated`
- On windows `--threadless` does not work with proxy.py https://github.com/abhinavsingh/proxy.py/issues/492


Dreamsorcerer:
Huh, I thought it was meant to be supported on all platforms. @webknjaz Happy to just have this test skipped?

bdraco:
I think its only supported on all platforms after https://github.com/python/cpython/pull/31275

Dreamsorcerer:
> I think its only supported on all platforms after [python/cpython#31275](https://github.com/python/cpython/pull/31275)

That's a PR that appears to have been merged into 3.11+?

webknjaz:
Can it be that the warning mentioned @ https://github.com/aio-libs/aiohttp/discussions/6044#discussioncomment-3797009 is a result of a flawed logic there? I think I saw a problem in related code, but never got to fixing it.. Or is this about some other problem?

bdraco:
I think thats a different issue. AFAICT the proxy isn't ssl so its not tls in tls https://github.com/aio-libs/aiohttp/discussions/6044#discussioncomment-3797009 

The check might need to be `if runtime_has_start_tls and proxy_req.is_ssl():` instead?
https://github.com/aio-libs/aiohttp/blob/30b7a4e99677b4014dda2372504343bb05fc983e/aiohttp/connector.py#L1248

bdraco:
> > I think its only supported on all platforms after [python/cpython#31275](https://github.com/python/cpython/pull/31275)
> 
> That's a PR that appears to have been merged into 3.11+?

The way I read this is the test shouldn't run if that flag is set because tls in tls is supported on the version of python the test is running on, but before 3.11 monkey patching was needed to enable it

Dreamsorcerer:
> > > I think its only supported on all platforms after [python/cpython#31275](https://github.com/python/cpython/pull/31275)
> > 
> > 
> > That's a PR that appears to have been merged into 3.11+?
> 
> The way I read this is the test shouldn't run if that flag is set because tls in tls is supported on the version of python the test is running on, but before 3.11 monkey patching was needed to enable it

Right, I hadn't really looked at the test. Doesn't that mean the test should just be skipped on Python 3.11+? And if so, why does it pass on Ubuntu with 3.11+? I'm a bit confused why the behaviour is different on different platforms.

bdraco:
> > > > I think its only supported on all platforms after [python/cpython#31275](https://github.com/python/cpython/pull/31275)
> 
> > > 
> 
> > > 
> 
> > > That's a PR that appears to have been merged into 3.11+?
> 
> > 
> 
> > The way I read this is the test shouldn't run if that flag is set because tls in tls is supported on the version of python the test is running on, but before 3.11 monkey patching was needed to enable it
> 
> 
> 
> Right, I hadn't really looked at the test. Doesn't that mean the test should just be skipped on Python 3.11+? 

I think that's the case.

> And if so, why does it pass on Ubuntu with 3.11+? I'm a bit confused why the behaviour is different on different platforms.

I would have expected it to be failing on Ubuntu 3.11. Maybe it's OpenSSL version dependent as well. 

I think I'm going to have to build a Ubuntu setup to be sure

bdraco:
Marking as draft until I can put together a Ubuntu env for local testing

bdraco:
Unfortunately I'm getting different behavior on the ubuntu docker container.  It might be openssl version dependent or the fact I'm running in a docker host.

Instead I adjusted the other failing test to run if we think tls in tls is supported so we have:

`test_secure_https_proxy_absolute_path` running if we think tls in tls is supported
`test_https_proxy_unsupported_tls_in_tls` running if we think tls in tls is unsupported

bdraco:
If `test_secure_https_proxy_absolute_path` is failing on ubuntu w/3.11+ in the CI, I'll probably have to do some trial and error debugging here since it means there is something specific about the github runner that causes it to fail that I haven't identified yet

Dreamsorcerer:
> If `test_secure_https_proxy_absolute_path` is failing on ubuntu w/3.11+ in the CI

No problem. @webknjaz had a good tool for that, a github action that allows you to login to the server or something. I don't remember what it was called, but a quick search suggests it might be one of these 2: https://github.com/marketplace/actions/a-debugger-for-actions https://github.com/marketplace/actions/debug-via-ssh

bdraco:
I gave up trying to figure this out and ordered a mini pc with ubuntu on it. Hopefully it will have the issue

JCHacking:
Maybe I can help in some way, I work with linux so if I can run those tests on my computer and try to see why I get errors?

bdraco:
@JCHacking If you can figure out `test_secure_https_proxy_absolute_path` fails with py3.11/py3.12 on ubuntu that would be awesome

JCHacking:
Perfect, I will try to discover it and comment here with what I see.

I understand that the bug is in master? Or in 3.9.0rc0?

bdraco:
> Perfect, I will try to discover it and comment here with what I see.
> 
> I understand that the bug is in master? Or in 3.9.0rc0?

Please test against `master`, (it probably happens on `3.9` as well)

JCHacking:
> > Perfect, I will try to discover it and comment here with what I see.
> > I understand that the bug is in master? Or in 3.9.0rc0?
> 
> Please test against `master`, (it probably happens on `3.9` as well)

I have installed it on my linux system (ArchLinux specifically) and I see that when I run the test that you mentioned it returns this:
```
=============================================================================================== slowest 10 durations ===============================================================================================
0.81s teardown tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https]
0.79s teardown tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http]
0.03s call     tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https]
0.03s call     tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http]
0.02s setup    tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http]
0.02s setup    tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https]
============================================================================================= short test summary info ==============================================================================================
XFAIL tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-http] - The secure proxy fixture does not seem to work under Python 3.10 on Linux and any Python on Windows. See https://github.com/abhinavsingh/proxy.py/issues/622.
XFAIL tests/test_proxy_functional.py::test_secure_https_proxy_absolute_path[pyloop-https] - The secure proxy fixture does not seem to work under Python 3.10 on Linux and any Python on Windows. See https://github.com/abhinavsingh/proxy.py/issues/622.
======================================================================================== 20 deselected, 2 xfailed in 4.10s =========================================================================================
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7fe61ff91b60>
```

Did it fail you in Ubuntu before? Or is this error to be corrected?

bdraco:
I would expect the test to pass on ubuntu with python 3.11/3.12 (version with the new asyncio ssl implementation https://github.com/python/cpython/pull/31275) but likely fail with version before that

JCHacking:
Okay, so removing that warning the goal is to pass the test. 👍🏻

bdraco:
The mini pc I ordered show up. Ironically it was $60 cheaper to get the ubuntu version but it came with windows installed anyways 👎

bdraco:
got a different error testing on ubuntu on the mini pc

ERROR tests/test_proxy_functional.py - cryptography.utils.CryptographyDeprecationWarning: Blowfish has been deprecated

bdraco:
https://github.com/paramiko/paramiko/issues/2038

bdraco:
```
8675  write(10, "ERROR    proxy.core.work.fd.fd:fd.py:48 Exception occurred during initialization\nTraceback (most recent call last):\n  File \"/home/bdraco/.local/lib/python3.11/site-packages/proxy/core/work/fd/fd.py\", line 45, in work\n    self.works[fileno].initialize()\n  File \"/home/bdraco/.local/lib/python3.11/site-packages/proxy/http/handler.py\", line 63, in initialize\n    super().initialize()\n  File \"/home/bdraco/.local/lib/python3.11/site-packages/proxy/core/base/tcp_server.py\", line 129, in initialize\n    conn = self._optionally_wrap_socket(self.work.connection)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/home/bdraco/.local/lib/python3.11/site-packages/proxy/core/base/tcp_server.py\", line 240, in _optionally_wrap_socket\n    conn = wrap_socket(conn, self.flags.keyfile, self.flags.certfile)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/home/bdraco/.local/lib/python3.11/site-packages/proxy/common/utils.py\", line 226, in wrap_socket\n    ctx.options |= ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3 | ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1\n    ^^^^^^^^^^^\n  File \"/usr/lib/python3.11/ssl.py\", line 624, in options\n    super(SSLContext, SSLContext).options.__set__(self, value)\nDeprecationWarning: ssl.OP_NO_SSL*/ssl.OP_NO_TLS* options are deprecated\n", 1292) = 1292
8675  getpid()                          = 8671
8675  write(10, "DEBUG    proxy.http.handler:handler.py:87 Closing client connection 127.0.0.1:36374 has buffer False\n", 101) = 101
```

bdraco:
so the warning about the deprecated OP_NO_SSL makes the test fail.

`#ctx.options |= ssl.OP_NO_SSLv2 | ssl.OP_NO_SSLv3 | ssl.OP_NO_TLSv1 | ssl.OP_NO_TLSv1_1` and everything works as expected

tests/test_proxy_functional.py
```diff
@@ -113,32 +114,17 @@ async def handler(*args, **kwargs):
...
-
-
-@secure_proxy_xfail(raises=ClientProxyConnectionError)
+@pytest.mark.skipif(
+    not ASYNCIO_SUPPORTS_TLS_IN_TLS,
+    reason="asyncio on this python does not support TLS in TLS",
+)
 @pytest.mark.parametrize("web_server_endpoint_type", ("http", "https"))
-@pytest.mark.usefixtures("_pretend_asyncio_supports_tls_in_tls", "loop")
+@pytest.mark.filterwarnings(r"ignore:.*ssl.OP_NO_SSL*")
```
bdraco, line 109: The test was failing because ubuntu threw this warning which caused it to raise an exception and the proxy died

Dreamsorcerer: Interesting, did you find the cause of the warning? (wondering if we can fix it, or remove the filter in a later release)

bdraco: Yes. documented inline and above now

Dreamsorcerer: Ah, cool, so maybe we can get it fixed in proxy.py then.

bdraco:
> > > > I think its only supported on all platforms after [python/cpython#31275](https://github.com/python/cpython/pull/31275)
> > > 
> > > 
> > > That's a PR that appears to have been merged into 3.11+?
> > 
> > 
> > The way I read this is the test shouldn't run if that flag is set because tls in tls is supported on the version of python the test is running on, but before 3.11 monkey patching was needed to enable it
> 
> Right, I hadn't really looked at the test. Doesn't that mean the test should just be skipped on Python 3.11+? And if so, why does it pass on Ubuntu with 3.11+? I'm a bit confused why the behaviour is different on different platforms.

The failure on Ubuntu turned out to be the warning causing the proxy to die so it make it look like tls in tls did not work on ubuntu. Once the warning is filtered and it no longer causes the proxy to raise, the test can do tls in tls

bdraco:
I'd hazard to guess the windows failures are likely caused by the same problem but a different warning

bdraco:
I was only able to find the warning by stracing the process and seeing where it died because it doesn't appear to be logged to anywhere useful

bdraco:
It looks like proxy.py is kind enough to send the log output to /dev/null 

`8670  openat(AT_FDCWD, "/dev/null", O_WRONLY|O_CREAT|O_TRUNC|O_CLOEXEC, 0666) = 10`

bdraco:
I guess the other question is why does the warning cause an exception on ubuntu and not on macos

tests/test_proxy_functional.py
```diff
@@ -158,13 +149,19 @@ async def test_secure_https_proxy_absolute_path(
     await conn.close()
 
 
-@secure_proxy_xfail(raises=AssertionError)
 @pytest.mark.parametrize("web_server_endpoint_type", ("https",))
 @pytest.mark.usefixtures("loop")
+@pytest.mark.skipif(
+    ASYNCIO_SUPPORTS_TLS_IN_TLS, reason="asyncio on this python supports TLS in TLS"
+)
```
Dreamsorcerer, line 141: To ensure we don't miss any other mistakes like this, we should probably just skip based on Python version.

bdraco:
I haven't found a way to get the warnings to even appear yet other than by stracing the process since proxy.py? is attaching the log to /dev/null

bdraco:
It looks like there is a broad exception catch here https://github.com/abhinavsingh/proxy.py/blob/30574fd0414005dfa8792a6e797023e862bdcf43/proxy/core/work/fd/fd.py#L44 but it fails to log anything ..... I think its failing before the logging is setup

bdraco:
The proxy is likely dying for a different reason on windows.

Having the proxy run inside the same python process as the test harness is leading lots of unexpected side effects as it doesn't contemplate running in that type of enviorment.

Maybe it should be run an in isolated subprocess instead since we are effectively testing `proxy.py` and thats not what we want.

bdraco:
Unfortunately I've run out of weekend and need to get need to fly to Europe for a quick turn before the US holiday so I'll have to pick this up when I get more time.

bdraco:
The good news is this very much seems like a problem with the test and not a production issue

bdraco:
sadly it still doesn't work on windows.

last two iterations had all the tests passing on mac and linux

bdraco:
https://github.com/abhinavsingh/proxy.py/issues/492

so only `--threaded` works correctly on windows

bdraco:
saved `fix_test_https_proxy_unsupported_tls_in_tls_macos_subproc` and trying to go back to the in process but with `--threaded` to see if it behaves better

tests/test_proxy_functional.py
bdraco: f6881fd4f250cdde310fda22d175d7553f149765

bdraco:
Looks like this is good to go

The only one still failing reliably (and not related to this PR) is tests/test_web_server.py::test_unsupported_upgrade[pyloop] - TimeoutError on windows and mac only

Dreamsorcerer:
Great, one step closer.

bdraco:
I'll get the patch backs done after breakfast and coffee. Up early due to jet lag

webknjaz:
> > If `test_secure_https_proxy_absolute_path` is failing on ubuntu w/3.11+ in the CI
> 
> No problem. @webknjaz had a good tool for that, a github action that allows you to login to the server or something. I don't remember what it was called, but a quick search suggests it might be one of these 2: [github.com/marketplace/actions/a-debugger-for-actions](https://github.com/marketplace/actions/a-debugger-for-actions) [github.com/marketplace/actions/debug-via-ssh](https://github.com/marketplace/actions/debug-via-ssh)

@Dreamsorcerer it's this one https://github.com/marketplace/actions/debugging-with-tmate. I recommend using the options to restrict the user who can connect with it.

webknjaz:
@bdraco thanks for figuring this out! I still remember how exhausting it was to integrate this in the first place…

bdraco:
Cheers! This one was a bit of a rabbit hole. I can only imagine how it was to integrate in the first place 😉

webknjaz:
@bdraco so I just realized that I accidentally downgraded the proxy.py version at the beginning of August via these PRs:
- https://github.com/aio-libs/aiohttp/pull/7460/files#diff-7269b1b148c1b82edce5234e9cefabaaa80c1f2f6599db2d007856d9582b1e92L6
- https://github.com/aio-libs/aiohttp/pull/7460/files#diff-7269b1b148c1b82edce5234e9cefabaaa80c1f2f6599db2d007856d9582b1e92L6

We should probably recover allowing pre-releases there to get the latest fixes early, as that project doesn't release often, despite my extensive contributions to their automation.

bdraco:
I haven't used pip-tools before so I'm not 100% sure if this is the right way to go about it, but https://github.com/aio-libs/aiohttp/pull/7927 gives it a shot

webknjaz:
@bdraco that's good enough. I don't think we have the exact invocation documented + we don't normally run it manually — dependabot uses pip-tools to upgrade pairs of `.in`+`.txt` files. But I do recommend you get familiar with the tool — it's super useful. Eventually, I'll probably take it to the next level in this repo, lockfile-wise.

Issue #7583, opened 2023-09-06
CookieJar improvements


Dreamsorcerer:
While looking through #7577, we found a few details that could possibly be improved.

- [ ] It should really output duplicate cookies in the Cookie header (e.g. when cookies with different paths have the same name) to match the recommended behaviour in the spec. This may be fairly complex to refactor, as the cookies/morsels used from http.cookies don't seem to support this. Upon reviewing the documentation, I get the impression that http.cookies is intended for server-side, while http.cookiejar is intended for client-side. So, we may need to replace http.cookies to support this behaviour.
- [ ] Cookies are now ordered by path length, but it should also use creation timestamp as a tie-breaker (point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4). This is probably most relevant if cookies with the same name and path match (but different domains, like a root domain and subdomain). I don't think we currently store creation timestamp, so we'd probably need to add this.
- [x] (#7944) .filter_cookies() currently just iterates over all cookies and checks each one if they are valid. This could be made more efficient because they are stored like `self._cookies[(domain, path)][name]`, therefore we could do domain-matching and path-matching on the keys, instead of testing every single cookie. https://github.com/aio-libs/aiohttp/blob/4639b36852f3b4939e61311d7863f1d7450d500d/aiohttp/cookiejar.py#L141-L142

Rongronggg9:
![](https://github.com/aio-libs/aiohttp/assets/15956627/3c4454c8-78d6-49fe-a49b-7fa680212774)

I profiled my program with yappi and found that `.filter_cookies()` consumed 27.5% (23.1s/83.9s) of the total CPU time consumed by requests.

![](https://github.com/aio-libs/aiohttp/assets/15956627/ce70f5bd-8288-483c-9220-87d98018de77)

As we can see, the preparation before filtering is very expensive.
https://github.com/aio-libs/aiohttp/blob/7ed2dd3793955736def36ff67044d19a43bdf4d5/aiohttp/cookiejar.py#L237-L252

However, not all requests will have cookies in their jar, for example, the initial request, or, when the session is only used to request those URLs that never sent cookies (images, videos, files, etc).

So I have another suggestion: test if there are any cookies in the jar before really doing anything.

Dreamsorcerer:
Open PRs that probably resolve these performance issues: #7784 #7777 #7790

Rongronggg9:
[...]

PR #7777, opened 2023-11-02
Implement filter_cookies() with domain-matching and path-matching

xiangxli:
## What do these changes do?

Implement filter_cookies() with domain-matching and path-matching on the keys, instead of testing every single cookie.

no

## Related issue number

https://github.com/aio-libs/aiohttp/issues/7583

task 3

## Checklist

### DIFF
--- a/tests/test_cookiejar.py
+++ b/tests/test_cookiejar.py
@@ -4,21 +4,16 @@ import datetime
 import itertools
 import pathlib
 import pickle
-import sys
 import unittest
 from http.cookies import BaseCookie, Morsel, SimpleCookie
 from typing import Any
 from unittest import mock
 
 import pytest
+from freezegun import freeze_time
 from yarl import URL
 
 from aiohttp import CookieJar, DummyCookieJar
-
-try:
-    from time_machine import travel
-except ImportError:
-    travel = None  # type: ignore[assignment]
 
 
 def dump_cookiejar() -> bytes:  # pragma: no cover
@@ -427,10 +422,10 @@ class TestCookieJarSafe(TestCookieJarBase):
         elif isinstance(send_time, float):
             send_time = datetime.datetime.fromtimestamp(send_time)
 
-        with travel(update_time, tick=False):
+        with freeze_time(update_time):
             self.jar.update_cookies(self.cookies_to_send)
 
-        with travel(send_time, tick=False):
+        with freeze_time(send_time):
             cookies_sent = self.jar.filter_cookies(URL(url))
 
         self.jar.clear()
@@ -617,10 +612,6 @@ class TestCookieJarSafe(TestCookieJarBase):
         self.assertEqual(cookies_received["path-cookie"]["path"], "/somepath")
         self.assertEqual(cookies_received["wrong-path-cookie"]["path"], "/")
 
-    @unittest.skipIf(
-        sys.implementation.name != "cpython",
-        reason="time_machine leverages CPython specific pointers https://github.com/adamchainz/time-machine/issues/305",
-    )
     def test_expires(self) -> None:
         ts_before = datetime.datetime(
             1975, 1, 1, tzinfo=datetime.timezone.utc
@@ -642,10 +633,6 @@ class TestCookieJarSafe(TestCookieJarBase):
 
         self.assertEqual(set(cookies_sent.keys()), {"shared-cookie"})
 
-    @unittest.skipIf(
-        sys.implementation.name != "cpython",
-        reason="time_machine leverages CPython specific pointers https://github.com/adamchainz/time-machine/issues/305",
-    )
     def test_max_age(self) -> None:
         cookies_sent = self.timed_request("http://maxagetest.com/", 1000, 1000)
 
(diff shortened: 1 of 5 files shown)


aiohttp/cookiejar.py
```diff
@@ -256,8 +256,27 @@ def filter_cookies(
             and request_origin not in self._treat_as_secure_origin
         )
 
+        # d: domain
+        # p: path
+        d = hostname
+        p = request_url.path
+
+        cookies = list(self._cookies[(d, p)].values())
```
Dreamsorcerer: Thanks for looking into this. I think there's a couple of cases still missing here.

From my understanding, we need to handle subdomains. So, we probably want `d` and then `d.split(".", maxsplit=1)` and repeat until we've checked every domain level.

And then I believe the same again for paths.

So, maybe we end up with something like:
```
pairs = []
while d:
    p = request_url.path
    while p:
        pairs.append((d, p))
        p = p.rsplit("/", maxsplit=1)[0]
    d = d.split(".", maxsplit=1)[1]
cookies = itertools.chain.from_iterable(self._cookies[p].values() for p in pairs)
```

aiohttp/cookiejar.py
```diff
@@ -256,8 +256,27 @@ def filter_cookies(
...
+
+        # shared cookie, it should have max of 1 entry
+        shared_cookie = self._cookies.get(("", "/"))
```
Dreamsorcerer: Maybe I'm just forgetting. But, is there anything in the code, or something in the spec which mentions this case?

aiohttp/cookiejar.py
```diff
@@ -256,8 +256,27 @@ def filter_cookies(
...
+        if shared_cookie:
+            shared_cookie_morsel = list(shared_cookie.values())[0]
+            cookies.append(shared_cookie_morsel)
+
+        # no path cookie, it should have max of 1 entry
+        no_path_cookie = self._cookies.get((d, "/"))
```
Dreamsorcerer: This can be removed after doing the above change.

aiohttp/cookiejar.py
```diff
@@ -256,8 +256,27 @@ def filter_cookies(
...
+        if no_path_cookie:
+            no_path_cookie_morsel = list(no_path_cookie.values())[0]
+            cookies.append(no_path_cookie_morsel)
+
         # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
-        for cookie in sorted(self, key=lambda c: len(c["path"])):
+        for cookie in sorted(cookies, key=lambda c: len(c["path"])):
```
Dreamsorcerer: We should be able to remove the sorted() call at this point, because we'll have sorted the cookies with the loop we create above. I suspect there are other parts of the code we can remove too inside the loop, in order to make it more efficient (e.g. we shouldn't need to do the domain match, as we've already looked up the cookies by name).

tests/test_cookiejar.py
```diff
@@ -254,6 +254,36 @@ async def test_filter_cookies_str_deprecated(loop: Any) -> None:
...
+        " Max-Age=60; "
+        "invalid-max-age-cookie=fifteenth; Domain=invalid-values.com; "
+        " Max-Age=string; "
+        "invalid-expires-cookie=sixteenth; Domain=invalid-values.com; "
+        " Expires=string;"
+    )
+    jar.update_cookies(cookies)
+    cookies = jar.filter_cookies(URL("http://pathtest.com/"))
+
+    assert len(cookies) == 3
```
Dreamsorcerer: We should probably assert the actual values that should be in here.

We should also test with a call that includes a subdomain and a multilevel path.

Dreamsorcerer: I suggest making some failing tests here, before doing the above updates.

aiohttp/cookiejar.py
xiangxli: I am checking the cases, there are a few number of failed test cases.
Will commit the changes once done.

tests/test_cookiejar.py
xiangxli: Agreed, I will add some failing test cases.

Dreamsorcerer:
Still interested in finishing this PR?

xiangxli:
> Still interested in finishing this PR?

@Dreamsorcerer 
~~Sorry, been hold up on work. It does not seem I will have time before new year's eve. If somebody else could take it before that time it would be great.~~

I squeezed a few hours to finish the code, please review it when you got a change. @Dreamsorcerer

Dreamsorcerer:
Uhh, I'm not sure what you just pushed, but now there's 150 files changed, with lots of recent changes to master being reverted...

xiangxli:
> Uhh, I'm not sure what you just pushed, but now there's 150 files changed, with lots of recent changes to master being reverted...

Could be an lagged version from master, let me fix that first.

xiangxli:
PR closed by force push, continuing on the following new PR
https://github.com/aio-libs/aiohttp/pull/7944

PR #7782, opened 2023-11-03
Possibly fix some flaky tests

Dreamsorcerer:
Make some tests a little more reliable

### DIFF
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -1352,7 +1352,9 @@ async def test_POST_MultiDict(aiohttp_client: Any) -> None:
     app.router.add_post("/", handler)
     client = await aiohttp_client(app)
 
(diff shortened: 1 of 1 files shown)


webknjaz:
Urgh... I was sure I've tracked down all the unclosed resources in the past. Apparently, not.

@Dreamsorcerer ideally, the resource warning ignores should be removed from https://github.com/aio-libs/aiohttp/blob/96de6eb/setup.cfg#L137-L139 when this stuff is fully solved.

Dreamsorcerer:
Yep, I'll go through the ignores at some point. Just noticed an unclosed warning from one of these tests and figured explicit async closing would make it more reliable.

webknjaz:
That's exactly what I've been doing to many other tests in the past. But's it's a lot of manual monotonous work, so I probably got bored and didn't complete checking all the tests.

PR #7784, opened 2023-11-03
Fix duplicate cookie expiration calls in the CookieJar implementation

bdraco:
## What do these changes do?

`_do_expiration` was already being called `filter_cookies` because it iterates `self` which calls `__iter__` which also calls `_do_expiration`

Calling `len()` on the cookie jar has the side of expiring stale cookies since it also called `__iter__`

Calling `len()` on the cookie jar no longer has the side effect of expiring stale cookies.

## Checklist


bdraco:
Screenshot 2023-11-02 at 10 19 01 PM

bdraco:
looks like there is another call
Screenshot 2023-11-02 at 11 01 02 PM

bdraco:
Screenshot 2023-11-02 at 11 17 21 PM


Only called once now per filter_cookies

bdraco:
It also looks like every time a request is made without a hostname it has to reconstruct a netloc to get `origin` and make an ip address object
Screenshot 2023-11-02 at 11 19 03 PM

bdraco:
confirmed only happening once per filter_cookies now

filter_cookies_one_parent

aiohttp/cookiejar.py
```diff
@@ -234,7 +239,6 @@ def filter_cookies(
         self, request_url: URL = URL()
     ) -> Union["BaseCookie[str]", "SimpleCookie[str]"]:
         """Returns this jar's cookies filtered by their attributes."""
-        self._do_expiration()
```
Dreamsorcerer: I think this'll conflict with #7777, so I'd suggest leaving it in for now. Then that PR should end up removing the iter over self.

bdraco:
Marking as draft until #7777 merges as it will be much easier to rebase this PR vs that one

aiohttp/cookiejar.py
Dreamsorcerer: This PR should be unblocked now.

Dreamsorcerer (approved):

PR #7785, opened 2023-11-03
Fix coroutine wrapper to await after exception

### LINKED ISSUE #7117: RuntimeError: await wasn't used with future
### Describe the bug

This is a strange one, as it's not something that has ever happened to me, but one of my users reported as an reoccurring issue. There's no real repro steps. It seems like it's something wrong with my code, but I can't pin-point exactly what it is, and the error makes zero sense. I don't know aiohttp or asyncio internals, so I'm making this bug report in hopes of finding out if it's really just my code, or is there something wrong with aiohttp itself.

### To Reproduce

No steps as I said, but here are the two functions that are common to both tracebacks, right before it enters aiohttp internals:

```py
# from twitch.py
@asynccontextmanager
async def request(
    self, method: str, url: URL | str, *, invalidate_after: datetime | None = None, **kwargs
) -> abc.AsyncIterator[aiohttp.ClientResponse]:
    session = await self.get_session()
    method = method.upper()
    if self.settings.proxy and "proxy" not in kwargs:
        kwargs["proxy"] = self.settings.proxy
    logger.debug(f"Request: ({method=}, {url=}, {kwargs=})")
    session_timeout = timedelta(seconds=session.timeout.total or 0)
    backoff = ExponentialBackoff(maximum=3*60)
    for delay in backoff:
        if self.gui.close_requested:
            raise ExitRequest()
        elif (
            invalidate_after is not None
            # account for the expiration landing during the request
            and datetime.now(timezone.utc) >= (invalidate_after - session_timeout)
        ):
            raise RequestInvalid()
        try:
            response: aiohttp.ClientResponse | None = None
            response = await self.gui.coro_unless_closed(
                session.request(method, url, **kwargs)
            )
            assert response is not None
            logger.debug(f"Response: {response.status}: {response}")
            if response.status < 500:
                # pre-read the response to avoid getting errors outside of the context manager
                raw_response = await response.read()  # noqa
                yield response
                return
            self.print(_("error", "site_down").format(seconds=round(delay)))
        except (aiohttp.ClientConnectionError, asyncio.TimeoutError):
            # just so that quick retries that often happen, aren't shown
            if backoff.steps > 1:
                self.print(_("error", "no_connection").format(seconds=round(delay)))
        finally:
            if response is not None:
                response.release()
        with suppress(asyncio.TimeoutError):
            await asyncio.wait_for(self.gui.wait_until_closed(), timeout=delay)

# from gui.py
async def coro_unless_closed(self, coro: abc.Awaitable[_T]) -> _T:
        # In Python 3.11, we need to explicitly wrap awaitables
        tasks = [asyncio.ensure_future(coro), asyncio.ensure_future(self._close_requested.wait())]
        done: set[asyncio.Task[Any]]
        pending: set[asyncio.Task[Any]]
        done, pending = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
        for task in pending:
            task.cancel()
        if self._close_requested.is_set():
            raise ExitRequest()
        return await next(iter(done))  # L1929
```

### Expected behavior

Well, I expect no error.

### Logs/tracebacks

```python-traceback
  File "main.py", line 178, in <module>
  File "asyncio\base_events.py", line 646, in run_until_complete
  File "twitch.py", line 670, in run
  File "twitch.py", line 712, in _run
  File "twitch.py", line 1417, in fetch_inventory
  File "gui.py", line 1247, in add_campaign
  File "cache.py", line 97, in get
  File "contextlib.py", line 199, in __aenter__
  File "twitch.py", line 1258, in request
  File "gui.py", line 1861, in coro_unless_closed
  File "aiohttp\client.py", line 1125, in send
  File "aiohttp\client.py", line 536, in _request
  File "aiohttp\connector.py", line 540, in connect
  File "aiohttp\connector.py", line 901, in _create_connection
  File "aiohttp\connector.py", line 1175, in _create_direct_connection
  File "aiohttp\connector.py", line 980, in _wrap_create_connection
  File "asyncio\base_events.py", line 1049, in create_connection
  File "asyncio\base_events.py", line 960, in _connect_sock
  File "asyncio\proactor_events.py", line 705, in sock_connect
RuntimeError: await wasn't used with future


  File "utils.py", line 96, in wrapper
  File "twitch.py", line 928, in _watch_loop
  File "channel.py", line 343, in send_watch
  File "channel.py", line 230, in get_spade_url
  File "contextlib.py", line 199, in __aenter__
  File "twitch.py", line 1269, in request
  File "gui.py", line 1929, in coro_unless_closed
  File "aiohttp\client.py", line 1125, in send
  File "aiohttp\client.py", line 536, in _request
  File "aiohttp\connector.py", line 540, in connect
  File "aiohttp\connector.py", line 901, in _create_connection
  File "aiohttp\connector.py", line 1175, in _create_direct_connection
  File "aiohttp\connector.py", line 980, in _wrap_create_connection
  File "asyncio\base_events.py", line 1049, in create_connection
  File "asyncio\base_events.py", line 960, in _connect_sock
  File "asyncio\proactor_events.py", line 705, in sock_connect
RuntimeError: await wasn't used with future
```

### Python Version

```console
Theirs: 3.10.6 (where this happens)

Mine: 3.8.15 (where this doesn't happen)

In both cases, the code is frozen into an application via PyInstaller.
```

### aiohttp Version

```console
Version: 3.8.3 (both cases)
```

### multidict Version

```console
Version: 6.0.2 (both cases)
```

### yarl Version

```console
Version: 1.8.1 (both cases)
```

### OS

Windows 10 x64 (theirs)

Windows 7 x64 (mine)

### Related component

Client

### Additional context

The entirety of the source code, for reference, is available here: https://github.com/DevilXD/TwitchDropsMiner

Ref issue: https://github.com/DevilXD/TwitchDropsMiner/issues/80

### Code of Conduct

stalkerg:
This PR fixes #7117 and similar issues. Short explanation - our coroutine wrapper does not properly handle the exception, which breaks coroutine handling. As you can see, any task expects results from `throw` - https://github.com/python/cpython/blob/main/Lib/asyncio/tasks.py#L303 but it seems like in aiohttp it was acidently removed by this commit https://github.com/stalkerg/aiohttp/commit/f04ecc241119b44fa52658cc17c5eab4adcb1648#diff-f334e752b4894ef951105572ab8b195aeb8db90eb6e48b1dfbd9a01da4c854f5L842

This is repro a case without aiohttp:
```python
import ssl
import collections


class TestCoroWrapper(collections.abc.Coroutine):
    __slots__ = ("_coro", "_resp")
    def __init__(self, coro):
        self._coro = coro

    def __getattr__(self, attr):
        return getattr(self._coro, attr)

    def send(self, arg):
        return self._coro.send(arg)

    def throw(self, arg):
        self._coro.throw(arg)

    def close(self):
        return self._coro.close()

    def __await__(self):
        ret = self._coro.__await__()
        return ret

async def ssl_call(context):
    return await loop.create_connection(
        lambda: asyncio.Protocol(),
        '2404:6800:4004:824::2004',
        443,
        ssl=context,
        family=socket.AddressFamily.AF_INET6,
        proto=6,
        flags=socket.AddressInfo.AI_NUMERICHOST | socket.AddressInfo.AI_NUMERICSERV,
        server_hostname='www.google.com',
        local_addr=None
    )

async def prepare_call():
    context = ssl.create_default_context()
    try:
        connection = await ssl_call(context)
    except ssl.SSLError as e:
        print(f"Got exception1: {e}")
        raise e

    return connection

async def my_task():
    try:
        await prepare_call()
    except Exception as e:
        print(f"Got exception2: {e}")

    await asyncio.sleep(1)
    raise Exception("test")

async def main():
    my_coro = TestCoroWrapper(my_task())
    print(f"is coro? {asyncio.iscoroutine(my_coro)}")
    task = asyncio.create_task(my_coro)
    await task

asyncio.run(main())
```

The `TestCoroWrapper` here is equivalent of `_BaseRequestContextManager`. If you run such code like: `SSL_CERT_FILE=/dev/null SSL_CERT_DIR=/dev/null python test.py` you will get an error: `await wasn't used with future`. 
The main idea here is that you are trying to await the sleep function after getting and catching an exception from the native (SSL) module. 

Now I should explain why repro with aiohttp for some users return the same error:
```python
import asyncio
import aiohttp

async def main():
        try:
            response = await asyncio.ensure_future(session.get('https://www.google.com/'))
            print(await response.text())
        finally:
            response.release()

asyncio.run(main())
```

here it's happened because in `TCPConnector._create_direct_connection` we are getting all IPs for the given host and trying to connect one by one. If the first connection gets an error we will catch this error and try again for the next IP. If you have IPv6 you will have at least 2 IPs here (ipv6 and ipv4), and after the first error, you will try to connect to a second IP and get the same error. 

Why it's problem only for `asyncio.ensure_future`? Because `asyncio.ensure_future` creates a `task` such a task starts processing coroutines and directly communicates with our coroutine wrapper witch not return a result for `throw`. 

I will write changelog and etc after people validate this PR. But honestly, I don't think I can write a unit test for a such case. 

Regards,

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1108,8 +1108,8 @@ class _BaseRequestContextManager(Coroutine[Any, Any, _RetType], Generic[_RetType
     def send(self, arg: None) -> "asyncio.Future[Any]":
         return self._coro.send(arg)
 
-    def throw(self, arg: BaseException) -> None:  # type: ignore[override]
-        self._coro.throw(arg)  # type: ignore[unused-awaitable]
+    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":
+        return self._coro.throw(arg)
 
     def close(self) -> None:
         return self._coro.close()
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1108,8 +1108,8 @@ class _BaseRequestContextManager(Coroutine[Any, Any, _RetType], Generic[_RetType
     def send(self, arg: None) -> "asyncio.Future[Any]":
         return self._coro.send(arg)
 
-    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":  # type: ignore[override]
-        return self._coro.throw(arg)
+    def throw(self, *args: Any, **kwargs: Any) -> "asyncio.Future[Any]":
+        return self._coro.throw(*args, **kwargs)
 
     def close(self) -> None:
         return self._coro.close()
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1108,7 +1108,7 @@ class _BaseRequestContextManager(Coroutine[Any, Any, _RetType], Generic[_RetType
     def send(self, arg: None) -> "asyncio.Future[Any]":
         return self._coro.send(arg)
 
-    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":
+    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":  # type: ignore[override]
         return self._coro.throw(arg)
 
     def close(self) -> None:
--- a/CHANGES/7785.bugfix
+++ b/CHANGES/7785.bugfix
@@ -1,0 +1,1 @@
+Fixed a rare `RuntimeError: await wasn't used with future` exception -- by :user:`stalkerg`


Dreamsorcerer:
It seems rather awkward that we're subclassing Coroutine in the first place. Would be nice if we can refactor it out entirely. But, atleast the reported issue is fixed for now. Thanks for digging into this.

stalkerg:
Yeah, such a wrapper can have also a small performance penalty because usually coroutine is a C object.

aiohttp/client.py
```diff
@@ -1108,8 +1108,8 @@ def __init__(self, coro: Coroutine["asyncio.Future[Any]", None, _RetType]) -> No
     def send(self, arg: None) -> "asyncio.Future[Any]":
         return self._coro.send(arg)
 
-    def throw(self, arg: BaseException) -> None:  # type: ignore[override]
-        self._coro.throw(arg)  # type: ignore[unused-awaitable]
+    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":
```
```suggestion
    def throw(self, arg: BaseException) -> "asyncio.Future[Any]":  # type: ignore[override]
```

Dreamsorcerer:
> Yeah, such a wrapper can have also a small performance penalty because usually coroutine is a C object.

Well, if there's any chance you get time to look at that, that'd be great. I'll merge this one in for now.

stalkerg:
> Well, if there's any chance you get time to look at that, that'd be great. I'll merge this one in for now.

okey, should I also add to CONTRIBUTORS.txt?

Also, I found what this signature is not fully correct, Python support two signatures for `throw`  https://docs.python.org/3/reference/datamodel.html#coroutine.throw

Seems like function should be:
```python
    def throw(self, *args, **kwargs) -> Any:
        return self._coro.throw(*args, **kwargs)
```

Dreamsorcerer:
> okey, should I also add to CONTRIBUTORS.txt?

Oh, whatever you like, I've not even bothered to add myself. :P

stalkerg:
Thanks!

PR #7798, opened 2023-11-07
Fix infinite loop releasing the connection when the writer is not finished

bdraco:
## What do these changes do?

#7764 ensured the connection was released if the writer was not done yet, but it never cleared the writer so it would schedule callbacks in a loop

Screenshot 2023-11-07 at 7 53 35 AM

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -916,10 +916,12 @@ class ClientResponse(HeadersMixin):
             ):
                 return
 
+            self._cleanup_writer()
             self._release_connection()
+        else:
+            self._cleanup_writer()
 
         self._closed = True
-        self._cleanup_writer()
 
     @property
     def closed(self) -> bool:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -968,11 +968,12 @@ class ClientResponse(HeadersMixin):
                 headers=self.headers,
             )
 
-    def _cleanup_writer_and_release_connection(self, _: "asyncio.Future[None]") -> None:
+    def _cleanup_on_writer_done(self, _: "asyncio.Future[None]") -> None:
         """Cleanup writer and release connection.
 
-        This is called when the writer is not finished
-        and _release_connection is called.
+        This is called as a callback when the writer finishes
+        in the event _release_connection was called is called
+        before the writer is done.
         """
         self._writer = None
         self._release_connection()
@@ -983,9 +984,7 @@ class ClientResponse(HeadersMixin):
                 self._connection.release()
                 self._connection = None
             else:
-                self._writer.add_done_callback(
-                    self._cleanup_writer_and_release_connection
-                )
+                self._writer.add_done_callback(self._cleanup_on_writer_done)
 
     async def _wait_released(self) -> None:
         if self._writer is not None:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -968,7 +968,7 @@ class ClientResponse(HeadersMixin):
                 headers=self.headers,
             )
 
-    def _cleanup_writer_and_release_connection(self, _: asyncio.Future) -> None:
+    def _cleanup_writer_and_release_connection(self, _: asyncio.Future[None]) -> None:
         """Cleanup writer and release connection.
 
         This is called when the writer is not finished
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -974,10 +974,8 @@ class ClientResponse(HeadersMixin):
         This is called when the writer is not finished
         and _release_connection is called.
         """
-        if self._connection is not None:
-            self._cleanup_writer()
-            self._connection.release()
-            self._connection = None
+        self._writer = None
+        self._release_connection()
 
     def _release_connection(self) -> None:
         if self._connection is not None:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -968,7 +968,7 @@ class ClientResponse(HeadersMixin):
                 headers=self.headers,
             )
 
-    def _cleanup_writer_and_release_connection(self, _: asyncio.Future[None]) -> None:
+    def _cleanup_writer_and_release_connection(self, _: "asyncio.Future[None]") -> None:
         """Cleanup writer and release connection.
 
         This is called when the writer is not finished
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -968,13 +968,26 @@ class ClientResponse(HeadersMixin):
                 headers=self.headers,
             )
 
+    def _cleanup_writer_and_release_connection(self, _: asyncio.Future) -> None:
+        """Cleanup writer and release connection.
+
+        This is called when the writer is not finished
+        and _release_connection is called.
+        """
+        if self._connection is not None:
+            self._cleanup_writer()
+            self._connection.release()
+            self._connection = None
+
     def _release_connection(self) -> None:
         if self._connection is not None:
             if self._writer is None:
                 self._connection.release()
                 self._connection = None
             else:
-                self._writer.add_done_callback(lambda f: self._release_connection())
+                self._writer.add_done_callback(
+                    self._cleanup_writer_and_release_connection
+                )
 
     async def _wait_released(self) -> None:
         if self._writer is not None:
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -972,7 +972,7 @@ class ClientResponse(HeadersMixin):
         """Cleanup writer and release connection.
 
         This is called as a callback when the writer finishes
-        in the event _release_connection was called is called
+        in the event _release_connection was called
         before the writer is done.
         """
         self._writer = None
--- a/CHANGES/7798.bugfix
+++ b/CHANGES/7798.bugfix
@@ -1,0 +1,1 @@
+Fix infinite loop releasing the connection when the writer is not finished


aiohttp/client_reqrep.py
```diff
@@ -968,13 +968,26 @@ def raise_for_status(self) -> None:
...
     def _release_connection(self) -> None:
         if self._connection is not None:
             if self._writer is None:
                 self._connection.release()
                 self._connection = None
             else:
-                self._writer.add_done_callback(lambda f: self._release_connection())
+                self._writer.add_done_callback(
+                    self._cleanup_writer_and_release_connection
+                )
```
bdraco: I made it a separate function to ensure any future refactoring never generates a loop

Dreamsorcerer: Is there definitely an issue here? The writer task itself does `self._writer = None`, which is why I originally left this code like this. It might be worth making a change to make it clearer that the behaviour is correct, I'm just aware that we already have multiple blocks of code in this class doing almost identical things, so I was trying to avoid adding another one.

bdraco: The loop is 

`_release_connection` is called and `self._writer` is not None so it calls the `self._writer.add_done_callback`, the writer finishes and the callback fires, so `_release_connection` is called again, and since `self._writer` was never unset on `ClientResponse` it does the `self._writer.add_done_callback` again and since its already done, it fires via `call_soon`, and the loop repeats.

bdraco: Also it happened without intervention and I only found the issue because the container was using 100% cpu.

Dreamsorcerer: Hmm, but why does the callback happen when the writer has not been reset? The task itself resets the attribute before completing...

Dreamsorcerer: Ah, is it only when getting a 100 response? I missed setting it there:

https://github.com/aio-libs/aiohttp/blob/8e4b1c1e7cdca45ea48810c4f928acf07904bc12/aiohttp/client_reqrep.py#L531-L536

Dreamsorcerer: I'd be tempted to just refactor write_bytes a little. Move all of that code into a _write_bytes() and then just do:
```
async def write_bytes(...):
    try:
        await self._write_bytes(...)
    finally:
        self._writer = None
```

bdraco: AFAICT, only `_response_eof` will ever call `_release_connection` before `_cleanup_writer` but it always calls `_cleanup_writer` right after if it does call `_release_connection` so I'm not sure how it can even happen, but it definitely does

Dreamsorcerer: > AFAICT, only `_response_eof` will ever call `_release_connection` before `_cleanup_writer` but it always calls `_cleanup_writer` right after if it does call `_release_connection` so I'm not sure how it can even happen, but it definitely does

But, if the writer is in progress at that point, and in the above mentioned 100-continue code, then the cancellation won't reset the writer attribute. Hence my above suggestion to ensure that it is always reset upon completion.

bdraco: I'll add a `raise` before `self._writer.add_done_callback` to my production Home Assistant instance to see if I can sus out how its even getting there

Dreamsorcerer: I think try printing/raising before the above return statement. That's the only place I can see it going wrong currently. If it's reaching that return statement, then that is the problem.

bdraco:
```
2023-11-07 12:27:41.569 ERROR (MainThread) [aiohttp.internal] Exception in eof callback                     
  File "/usr/local/lib/python3.11/site-packages/aiohttp/streams.py", line 166, in on_eof                    
    callback()                                                                                                                                
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 978, in _response_eof
    self._release_connection()                                                                       
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 1047, in _release_connection
    raise RuntimeError("writer is not done")                                                                
RuntimeError: writer is not done   
```

bdraco: I guess it would make sense to refactor `_response_eof` to call `_cleanup_writer` first so avoid the callback?

bdraco:
```
Nov 07 18:32:36 homeassistant homeassistant[540]: 2023-11-07 12:32:36.220 ERROR (MainThread) [aiohttp.internal] Exception in eof callback
Nov 07 18:32:36 homeassistant homeassistant[540]: Traceback (most recent call last):
Nov 07 18:32:36 homeassistant homeassistant[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/streams.py", line 166, in on_eof
Nov 07 18:32:36 homeassistant homeassistant[540]:     callback()
Nov 07 18:32:36 homeassistant homeassistant[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 978, in _response_eof
Nov 07 18:32:36 homeassistant homeassistant[540]:     self._release_connection()
Nov 07 18:32:36 homeassistant homeassistant[540]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 1048, in _release_connection
Nov 07 18:32:36 homeassistant homeassistant[540]:     raise RuntimeError("writer is already done")
Nov 07 18:32:36 homeassistant homeassistant[540]: RuntimeError: writer is already done
```

bdraco: Thats with adding this guard
```
                if self._writer and self._writer.done():
                    raise RuntimeError("writer is already done")
```

bdraco:
```
    def _response_eof(self) -> None:
        if self._closed:
            return

        self._closed = True
        self._cleanup_writer()

        if self._connection is not None:
            # websocket, protocol could be None because
            # connection could be detached
            if (
                self._connection.protocol is not None
                and self._connection.protocol.upgraded
            ):
                return

            self._release_connection()
```

That seems a bit more efficient ?

bdraco: > I think try printing/raising before the above return statement. That's the only place I can see it going wrong currently. If it's reaching that return statement, then that is the problem.

Trying that now

Dreamsorcerer: > That seems a bit more efficient ?

Yes, but the behaviour changes if we meet the condition for that return statement. I deliberately didn't change this one, as I wasn't confident if that was correct or not. i.e. I'm not sure if there is a reason we don't want to run those 2 lines if we reach that return statement, or if that was just an accident when the code was originally written.

bdraco: makes sense

I guess
```python
    def _response_eof(self) -> None:
        if self._closed:
            return

        if self._connection is not None:
            # websocket, protocol could be None because
            # connection could be detached
            if (
                self._connection.protocol is not None
                and self._connection.protocol.upgraded
            ):
                return

            self._cleanup_writer()
            self._release_connection()
        else:
            self._cleanup_writer()

        self._closed = True
```
could work as well

bdraco: > > I think try printing/raising before the above return statement. That's the only place I can see it going wrong currently. If it's reaching that return statement, then that is the problem.
> 
> Trying that now

I haven't been able to get it to happen there ... yet

Dreamsorcerer: > > > I think try printing/raising before the above return statement. That's the only place I can see it going wrong currently. If it's reaching that return statement, then that is the problem.
> > 
> > 
> > Trying that now
> 
> I haven't been able to get it to happen there ... yet

Hmm, I suppose it'd still be possible for there to be an exception on conn.protocol or the following assertion maybe. I'd still be tempted to try: https://github.com/aio-libs/aiohttp/pull/7798#discussion_r1385338142

Or, maybe we could put a setter on _writer, so whenever it is set to non-None, we add a done_callback which sets it to None. I've also realised this class can be subclassed, so I suspect this is probably one of the safer options.

bdraco:
```python
        if self._continue is not None:
            try:
                await writer.drain()
                await self._continue
                _LOGGER.warning("100 Continue response was not received")
                raise RuntimeError("Writing data failed: request was cancelled")
```

Haven't got a hit on this yet, but it also possible I haven't waited long enough.  I'll report back in a few hours if it hits

bdraco: No hits on the `asyncio.CancelledError`. Will check it again in the morning

bdraco: No hits on that block over night

bdraco: Pushed what I've been running overnight with no issues

bdraco: Everything ran fine today with this PR.  I had another system hit the loop that I hadn't updated yet and I put this PR on it and so far so good with that one as well.

Dreamsorcerer: I'll come back to this shortly. Another possible cause I thought of, is that the task may get cancelled before it is started, in which case there would be no way for it to reset the attribute itself. I still think it could be more efficient by adding a done callback to the task immediately, to ensure that it is always reset as soon as the task is done (using a setter on _writer to add the callback whenever it is set to a non-None value).

Dreamsorcerer:
```
@setter
def _writer(self, writer: Task | None):
    if writer is not None:
        writer.add_done_callback(reset_to_none)
    self.__writer = writer
```
or similar.

Then we presumably wouldn't need to set it none anywhere else in the code (which I think is currently being done in several places).

bdraco: I gave that a shot but it seems like the writer doesn't get awaited because its set to None

Dreamsorcerer: This is what I had in mind, if you've got any time to look at it and give it a test:
https://github.com/aio-libs/aiohttp/pull/7815/files

It is causing all the tests to fail though, because they are currently passing a mock as the writer, which doesn't implement add/remove_done_callback() properly.

bdraco: I'll pull it in and test.  It also look like `ClientResponse` only sets `_writer` in `__init__` so it might be able to be simplified a bit?

Dreamsorcerer: Yeah, probably, I'm just concerned that it can be subclassed, so trying to be a little cautious and make it difficult to mess up.

bdraco:
```
Nov 10 18:26:44 homeassistant homeassistant[531]: 2023-11-10 12:26:44.705 ERROR (MainThread) [aiohttp.internal] Exception in eof callback
Nov 10 18:26:44 homeassistant homeassistant[531]: Traceback (most recent call last):
Nov 10 18:26:44 homeassistant homeassistant[531]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/streams.py", line 166, in on_eof
Nov 10 18:26:44 homeassistant homeassistant[531]:     callback()
Nov 10 18:26:44 homeassistant homeassistant[531]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 1014, in _response_eof
Nov 10 18:26:44 homeassistant homeassistant[531]:     self._cleanup_writer()
Nov 10 18:26:44 homeassistant homeassistant[531]:   File "/usr/local/lib/python3.11/site-packages/aiohttp/client_reqrep.py", line 1078, in _cleanup_writer
Nov 10 18:26:44 homeassistant homeassistant[531]:     assert not self._writer.done()
Nov 10 18:26:44 homeassistant homeassistant[531]: AssertionError

```

Looks like an assertion is failing

bdraco:
```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index 3f109fc6..5f04483a 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -1075,7 +1075,7 @@ class ClientResponse(HeadersMixin):
 
     def _cleanup_writer(self) -> None:
         if self._writer is not None:
-            assert not self._writer.done()
+            if not self._writer.done():
                 self._writer.cancel()
         self._session = None
 

```

Pushed to production with a small tweak

bdraco: the loop is back
```
top - 12:31:16 up 1 day, 13:38,  0 user,  load average: 1.06, 0.76, 0.39
Tasks:  10 total,   2 running,   8 sleeping,   0 stopped,   0 zombie
%Cpu(s): 12.3 us,  1.2 sy,  0.0 ni, 85.2 id,  0.0 wa,  0.0 hi,  1.2 si,  0.0 st 
MiB Mem :  15630.5 total,   6143.4 free,   1603.4 used,   8052.3 buff/cache     
MiB Swap:   5158.2 total,   5158.2 free,      0.0 used.  14027.0 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                                                                           
     66 root      20   0 2301904 665556  76948 R 100.0   4.2   1:19.32 python3    
```

bdraco: Screenshot 2023-11-10 at 12 33 22 PM
Screenshot 2023-11-10 at 12 34 46 PM

[callgrind.out.1699641101306196.zip](https://github.com/aio-libs/aiohttp/files/13322331/callgrind.out.1699641101306196.zip)
[loop.svg.zip](https://github.com/aio-libs/aiohttp/files/13322340/loop.svg.zip)

Dreamsorcerer: Yeah, if that assert fails then something is wrong with my logic. I thought that was happening in the tests just because it was a mock. If I don't think of anything by tomorrow, then we'll just go with this PR.

Dreamsorcerer: Wait, just realised, that assignment in the init needs to trigger the callback logic. Just pushed a commit to ensure that happens. Can you try again with that one?

bdraco: I'll give it a shot as soon as I get back home < 1h

bdraco: cherry-picked f45d9e48e29c12f2b839e121a20598c92ac359cf cleanly, results shortly

bdraco: Came up cleanly, loop did not happen right away as before. 

profile is clean, py-spy is clean

Will report back tomorrow as the original symptoms only happened after about 12 hours (not sure one which request)

bdraco: Everything ran fine overnight.
I rebooted a few switches and routers to generate some network chaos and everything recovered just fine

Closing this PR in favor of #7815

bdraco:
closing in favor of #7815

PR #7819, opened 2023-11-11
Skip filtering ``CookieJar`` when the jar is empty or all cookies have expired

Rongronggg9:
## What do these changes do?

The filtering itself and its preparation in `CookieJar.filter_cookies()` is expensive. Sometimes there are no cookies in the jar or all cookies have expired. Skip filtering and its preparation in this case.

Because the empty check is much cheaper than `_do_expiration()`, I think it deserves to be duplicated before and after calling `_do_expiration()`.

```console
$ python3.11 -m timeit -s 'from collections import defaultdict; d=defaultdict(foo="bar")' \
> 'if not d: pass'
50000000 loops, best of 5: 8.3 nsec per loop
$ python3.11 -m timeit -s 'from collections import defaultdict; d=defaultdict()' \
> 'if not d: pass'
50000000 loops, best of 5: 8.74 nsec per loop
$ python3.11 -m timeit -s 'from aiohttp import CookieJar; cj = CookieJar()' \
> 'cj._do_expiration()'
200000 loops, best of 5: 1.86 usec per loop
```

No.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/7583#issuecomment-1806328904

## Checklist

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -246,14 +246,11 @@ class CookieJar(AbstractCookieJar):
             SimpleCookie() if self._quote_cookie else BaseCookie()
         )
         if not self._cookies:
-            # Shortcut: empty jar
-            # We need the shortcut because the filtering itself and its
-            # preparation is expensive
+            # Skip do_expiration() if there are no cookies.
             return filtered
         self._do_expiration()
         if not self._cookies:
-            # Shortcut: all cookies expired
-            # Likewise.
+            # Skip rest of function if no non-expired cookies.
             return filtered
         hostname = request_url.raw_host or ""
         request_origin = URL()
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -234,7 +234,6 @@ class CookieJar(AbstractCookieJar):
         self, request_url: URL = URL()
     ) -> Union["BaseCookie[str]", "SimpleCookie[str]"]:
         """Returns this jar's cookies filtered by their attributes."""
-        self._do_expiration()
         if not isinstance(request_url, URL):
             warnings.warn(
                 "The method accepts yarl.URL instances only, got {}".format(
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
This looks like it's probably OK. But, we might want to look at seeing if we can replace datetime. In my quick tests with an empty cookiejar, the datetime operations are taking around 60% of the time in .clear() (with or without a predicate).

bdraco:
> This looks like it's probably OK. But, we might want to look at seeing if we can replace datetime. In my quick tests with an empty cookiejar, the datetime operations are taking around 60% of the time in .clear() (with or without a predicate).

Good idea. `time.time()` should be much faster here for comparisons

Rongronggg9:
> But, we might want to look at seeing if we can replace datetime.

Sure. Would you like me to update the PR to make it contain a bunch of optimizations? Or just open a new one?

As for what we have discussed in #7583 and here, plus those optimizations I thought of when I read the code, we can:
1. Skip filtering `CookieJar` when the jar is empty or all cookies have expired (this PR)
2. Skip filtering if `not CookieJar._unsafe and is_ip_address(hostname)` (https://github.com/aio-libs/aiohttp/issues/7583#issuecomment-1806861621)
3. Replace the usage of `datetime` with `time`
4. Only call `request_url.origin()` when we really need it (`request_url.scheme not in ("https", "wss")` and `self._treat_as_secure_origin` is not empty)

I think we can let 1,2,4 go to this PR since they are quite straightforward, and 3 go to another one?

Dreamsorcerer:
> Sure. Would you like me to update the PR to make it contain a bunch of optimizations? Or just open a new one?

Probably new ones, we can do them in steps.

aiohttp/cookiejar.py
```diff
@@ -246,6 +245,16 @@ def filter_cookies(
...
+        if not self._cookies:
+            # Shortcut: empty jar
+            # We need the shortcut because the filtering itself and its
+            # preparation is expensive
+            return filtered
+        self._do_expiration()
+        if not self._cookies:
+            # Shortcut: all cookies expired
+            # Likewise.
+            return filtered
```
```suggestion
            # Skip do_expiration() if there are no cookies.
            return filtered
        self._do_expiration()
        if not self._cookies:
            # Skip rest of function if no non-expired cookies.
            return filtered
```

Dreamsorcerer:
If you can handle the backport with the above instructions, that'd be really helpful.

Rongronggg9:
Sure. I will backport it later.

Issue #7790, opened 2023-11-05
Remove _next_expiration field from CookieJar


rgeronimi:
### Is your feature request related to a problem?

The _next_expiration field seems unused in CookieJar. It is only written to and never read in any part of aiohttp:

https://github.com/aio-libs/aiohttp/blob/a57dc3146839e2e1978b96db48e1de3af1a2bb50/aiohttp/cookiejar.py#L84

### Describe the solution you'd like

Remove the field if it's useless

### Describe alternatives you've considered

Maybe that it was intended for something that was never completely coded (e.g.,fast removal of expired cookies).

### Related component

Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

rgeronimi:
Indeed this _next_expiration field was used to perform fast removal of expired cookies, but a posterior commits deleted the line of code that was using it:
https://github.com/aio-libs/aiohttp/blob/d65f5cb4751651c7c43ac3744c92cbd4b23eec04/aiohttp/cookiejar.py#L87

Since that line of code has been deleted, _next_expiration is useless now (and the cleaning of cookies is slower as it doesn't use it to cut short the clear process).

The commit that deleted this line of code is
https://github.com/aio-libs/aiohttp/pull/5249 by @derlih

As an alternative to removing _next_expiration entirely, this line of code could be reintroduced. However this is more difficult now, as this commit made the _do_expiration code to be removed, and call instead the clear method, which now serves additional needs, like allowing custom selection of cookies to be cleared through a callback function. To revert this, the old _do_expiration code shall be reinstated, instead of delegating the work to the clear method.

More largely, frequent execution paths (_do_expiration is called very often across the entire API) shall have simple code benefiting from shortcuts (like the _next_expiration field to determine if the expiration process can be skipped) and rare paths (like a clear operation with a custom selection function) could be isolated into dedicated code.

Dreamsorcerer:
#7784 is also looking at removing duplicate calls to _do_expiration(). @bdraco Any opinion on whether this is worth reimplementing, or if it becomes a bit redundant after your changes?

bdraco:
Ideally the `_next_expiration` functionality would be restored.  

[...]

PR #7824, opened 2023-11-12
Use timestamp instead of datetime to achieve faster cookie expiration in CookieJar

Rongronggg9:
## What do these changes do?

Use timestamp instead of `datetime` to achieve faster cookie expiration in `CookieJar`

No.

## Related issue number

#7583
https://github.com/aio-libs/aiohttp/pull/7819#issuecomment-1806933086

## Checklist

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -8,6 +8,8 @@ import re
 import time
 import warnings
 from collections import defaultdict
+from math import ceil
+
 from http.cookies import BaseCookie, Morsel, SimpleCookie
 from typing import (  # noqa
     DefaultDict,
@@ -26,7 +28,7 @@ from typing import (  # noqa
 from yarl import URL
 
 from .abc import AbstractCookieJar, ClearCookiePredicate
-from .helpers import is_ip_address, next_whole_second
+from .helpers import is_ip_address
 from .typedefs import LooseCookies, PathLike, StrOrURL
 
 __all__ = ("CookieJar", "DummyCookieJar")
@@ -96,7 +98,7 @@ class CookieJar(AbstractCookieJar):
                 for url in treat_as_secure_origin
             ]
         self._treat_as_secure_origin = treat_as_secure_origin
-        self._next_expiration: Union[int, float] = next_whole_second()
+        self._next_expiration: Union[int, float] = ceil(time.time())
         self._expirations: Dict[Tuple[str, str, str], Union[int, float]] = {}
 
     def save(self, file_path: PathLike) -> None:
@@ -111,7 +113,7 @@ class CookieJar(AbstractCookieJar):
 
     def clear(self, predicate: Optional[ClearCookiePredicate] = None) -> None:
         if predicate is None:
-            self._next_expiration = next_whole_second()
+            self._next_expiration = ceil(time.time())
             self._cookies.clear()
             self._host_only_cookies.clear()
             self._expirations.clear()
@@ -134,8 +136,7 @@ class CookieJar(AbstractCookieJar):
                 del self._expirations[(domain, path, name)]
             self._cookies[(domain, path)].pop(name, None)
 
-        next_expiration = min(self._expirations.values(), default=self.MAX_TIME)
-        self._next_expiration = min(next_expiration + 1, self.MAX_TIME)
+        self._next_expiration = min(*self._expirations.values(), self.MAX_TIME - 1) + 1
 
     def clear_domain(self, domain: str) -> None:
         self.clear(lambda x: self._is_domain_match(domain, x["domain"]))
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -99,7 +99,7 @@ class CookieJar(AbstractCookieJar):
                 for url in treat_as_secure_origin
             ]
         self._treat_as_secure_origin = treat_as_secure_origin
-        self._next_expiration: Union[int, float] = ceil(time.time())
+        self._next_expiration: float = ceil(time.time())
         self._expirations: Dict[Tuple[str, str, str], Union[int, float]] = {}
 
     def save(self, file_path: PathLike) -> None:
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -56,9 +56,9 @@ class CookieJar(AbstractCookieJar):
 
     # calendar.timegm() fails for timestamps after datetime.datetime.max
     # Minus one as a loss of precision occurs when timestamp() is called.
-    MAX_TIME = int(
-        datetime.datetime.max.replace(tzinfo=datetime.timezone.utc).timestamp()
-    ) - 1
+    MAX_TIME = (
+        int(datetime.datetime.max.replace(tzinfo=datetime.timezone.utc).timestamp()) - 1
+    )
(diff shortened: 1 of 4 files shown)


Rongronggg9:
The test failure is Windows-specific. See also https://stackoverflow.com/questions/46133223/maximum-value-of-timestamp

aiohttp/cookiejar.py
```diff
@@ -125,13 +134,8 @@ def clear(self, predicate: Optional[ClearCookiePredicate] = None) -> None:
...
 
-        next_expiration = min(self._expirations.values(), default=self._max_time)
-        try:
-            self._next_expiration = next_expiration.replace(
-                microsecond=0
-            ) + datetime.timedelta(seconds=1)
-        except OverflowError:
-            self._next_expiration = self._max_time
+        next_expiration = min(self._expirations.values(), default=self.MAX_TIME)
+        self._next_expiration = min(next_expiration + 1, self.MAX_TIME)
```
Dreamsorcerer: This seems to be about twice as fast:
```suggestion
        self._next_expiration = min(*self._expirations.values(), self.MAX_TIME-1) + 1
```

aiohttp/helpers.py
```diff
@@ -530,11 +530,9 @@ def is_ip_address(host: Optional[Union[str, bytes, bytearray, memoryview]]) -> b
     return is_ipv4_address(host) or is_ipv6_address(host)
 
 
-def next_whole_second() -> datetime.datetime:
+def next_whole_second() -> int:
     """Return current time rounded up to the next whole second."""
-    return datetime.datetime.now(datetime.timezone.utc).replace(
-        microsecond=0
-    ) + datetime.timedelta(seconds=0)
+    return ceil(time.time())
```
Dreamsorcerer: I'm also thinking that we can just remove this function as it's been greatly simplified and is only used twice.

bdraco:
```python
import timeit
from http import cookies

from yarl import URL

from aiohttp import CookieJar


def filter_large_cookie_jar():
    """Filter out large cookies from the cookie jar."""
    jar = CookieJar()
    c = cookies.SimpleCookie()
    domain_url = URL("http://maxagetest.com/")
    other_url = URL("http://otherurl.com/")

    for i in range(5000):
        cookie_name = f"max-age-cookie{i}"
        c[cookie_name] = "any"
        c[cookie_name]["max-age"] = 60
        c[cookie_name]["domain"] = "maxagetest.com"
    jar.update_cookies(c, domain_url)
    assert len(jar) == 5000
    assert len(jar.filter_cookies(domain_url)) == 5000
    assert len(jar.filter_cookies(other_url)) == 0

    filter_domain = timeit.timeit(lambda: jar.filter_cookies(domain_url), number=1000)
    filter_other_domain = timeit.timeit(
        lambda: jar.filter_cookies(other_url), number=1000
    )
    print(f"filter_domain: {filter_domain}")
    print(f"filter_other_domain: {filter_other_domain}")


filter_large_cookie_jar()
```

Basic benchmark to test against (can probably be improved a bit), although I don't think this PR will affect it very much

aiohttp/cookiejar.py
Rongronggg9: It leads to an error if `self._expirations` is empty: `TypeError: 'int' object is not iterable`.
I read the code twice and found `self._next_expiration` never actually used anywhere, may we just remove it?

Rongronggg9: The usage of `self._next_expiration` has been removed since e65f1a9eba0ebde848bf0e50aa9be4753334b3c8. If it is intended to be a shortcut of `_do_expiration`, it should be added back.

bdraco: @Rongronggg9 please see the discussion in https://github.com/aio-libs/aiohttp/issues/7790

Rongronggg9: @bdraco Thanks for your information! I didn't notice that.
So for the time being I decided to just keep it :)

Dreamsorcerer: You can just unpack into a tuple, and it'll work with 0-length lists:
`self._next_expiration = min((*self._expirations.values(), self.SUB_MAX_TIME)) + 1`

Dreamsorcerer: The if statement seems to be a tiny bit faster though, so may as well stick with that.

aiohttp/cookiejar.py
```diff
@@ -52,9 +55,24 @@ class CookieJar(AbstractCookieJar):
...
+        int(datetime.datetime.max.replace(tzinfo=datetime.timezone.utc).timestamp()) - 1
+    )
+    try:
+        calendar.timegm(time.gmtime(MAX_TIME))
+    except OSError:
+        # Hit the maximum representable time on Windows
+        # https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64
+        MAX_TIME = calendar.timegm((3000, 12, 31, 23, 59, 59, -1, -1, -1))
+        # The below line should never raise an exception
+        time.gmtime(MAX_TIME)
```
Dreamsorcerer: Why is this line here? Should we just remove it?

aiohttp/cookiejar.py
```diff
@@ -81,14 +99,8 @@ def __init__(
...
         self._treat_as_secure_origin = treat_as_secure_origin
-        self._next_expiration = next_whole_second()
-        self._expirations: Dict[Tuple[str, str, str], datetime.datetime] = {}
-        # #4515: datetime.max may not be representable on 32-bit platforms
-        self._max_time = self.MAX_TIME
-        try:
-            self._max_time.timestamp()
-        except OverflowError:
-            self._max_time = self.MAX_32BIT_TIME
+        self._next_expiration: Union[int, float] = ceil(time.time())
```
```suggestion
        self._next_expiration: float = ceil(time.time())
```
int is compatible with float

aiohttp/cookiejar.py
```diff
@@ -81,14 +99,8 @@ def __init__(
...
+        self._expirations: Dict[Tuple[str, str, str], Union[int, float]] = {}
```
```suggestion
        self._expirations: Dict[Tuple[str, str, str], float] = {}
```

aiohttp/cookiejar.py
```diff
@@ -148,7 +158,7 @@ def _do_expiration(self) -> None:
         self.clear(lambda x: False)
 
     def _expire_cookie(
-        self, when: datetime.datetime, domain: str, path: str, name: str
+        self, when: Union[int, float], domain: str, path: str, name: str
```
```suggestion
        self, when: float, domain: str, path: str, name: str
```

Dreamsorcerer (approved):

aiohttp/cookiejar.py
Rongronggg9: We assume that OSError will only occur on Windows, without really verifying the current OS. The line is to ensure if there is any other (OS, arch) that raises OSError, the Windows-specific max time won't be too large (we should receive bug reports if it occurs).

Dreamsorcerer: But, if it does happen, then the whole library fails on import, even if the user would never have received a value anywhere near the MAX_TIME and wouldn't have experienced any issues at runtime.

Rongronggg9: OK then. I've removed the line

Dreamsorcerer:
#7837

PR #7829, opened 2023-11-12
Index resources in the UrlDispatcher to avoid linear search for most cases

### LINKED ISSUE #7828: UrlDispatcher improvements
### Describe the bug

The UrlDispatcher currently has to do a linear search to dispatch urls which has an average time complexity of `O(n)`. This means that urls that are registered first can be found faster than urls that are registered later.  

### To Reproduce

Results
```
% python3 url_dispatcher.py
resolve_time_first_url: 0.0017513339989818633
resolve_time_last_url: 2.258735582989175
```

This can be demonstrated with the following benchmark script
```python
import asyncio
import timeit
from unittest.mock import MagicMock, AsyncMock

from yarl import URL

from aiohttp import web
from aiohttp.web_urldispatcher import (
    UrlDispatcher,
)

URL_COUNT = 5000


async def url_dispatcher_performance():
    """Filter out large cookies from the cookie jar."""
    dispatcher = UrlDispatcher()
    for i in range(URL_COUNT):
        dispatcher.add_get(f"/static/sub/path{i}", AsyncMock())

    first_url = "/static/sub/path0"
    last_url = f"/static/sub/path{URL_COUNT-1}"

    request_first_url = web.Request(
        MagicMock(url=URL(first_url), method="GET"),
        MagicMock(),
        protocol=MagicMock(),
        host="example.com",
        task=MagicMock(),
        loop=asyncio.get_running_loop(),
        payload_writer=MagicMock(),
    )
    match_info = await dispatcher.resolve(request_first_url)
    assert match_info.route.resource.canonical == first_url

    request_last_url = web.Request(
        MagicMock(url=URL(last_url), method="GET"),
        MagicMock(),
        protocol=MagicMock(),
        host="example.com",
        task=MagicMock(),
        loop=asyncio.get_running_loop(),
        payload_writer=MagicMock(),
    )
    match_info = await dispatcher.resolve(request_last_url)
    assert match_info.route.resource.canonical == last_url

    async def resolve_times(request: web.Request, times: int) -> float:
        start = timeit.default_timer()
        for _ in range(times):
            await dispatcher.resolve(request)
        end = timeit.default_timer()
        return end - start

    run_count = 2000
    resolve_time_first_url = await resolve_times(request_first_url, run_count)
    resolve_time_last_url = await resolve_times(request_last_url, run_count)
    print(f"resolve_time_first_url: {resolve_time_first_url}")
    print(f"resolve_time_last_url: {resolve_time_last_url}")


asyncio.run(url_dispatcher_performance())

```

### Expected behavior

Ideally the time complexity is at most `O(url_parts)`

### Logs/tracebacks

```python-traceback
n/a
```

### Python Version

```console
n/a
```

### aiohttp Version

```console
master
```

### multidict Version

```console
n/a
```

### yarl Version

```console
n/a
```

### OS

n/a

### Related component

Server

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

The time complexity of the UrlDispatcher on a hit averaged `O(n)`

The UrlDispatcher now keeps an index of all resources and will attempt to avoid doing a linear search to resolve.  On a hit the performance should be nearly `O(url_parts)` on a miss the time complexity will be `O(domains) + O(url_parts)`

This works for prefixed subapps as well.

sub apps added by `add_domain` (aka `MatchedSubAppResource`) will fallback to linear searching as they are not indexable since they support wildcard/regexs

A fixed/static path will always be preferred over a dynamic path.

## Related issue number

fixes #7828

benchmark
```python
import asyncio
import timeit
from unittest.mock import MagicMock, AsyncMock

from yarl import URL

from aiohttp import web
from aiohttp.web_urldispatcher import (
    UrlDispatcher,
)

URL_COUNT = 5000


def make_mock_request(url: str) -> web.Request:
    return web.Request(
        MagicMock(url=URL(url), method="GET"),
        MagicMock(),
        protocol=MagicMock(),
        host="example.com",
        task=MagicMock(),
        loop=asyncio.get_running_loop(),
        payload_writer=MagicMock(),
    )


async def url_dispatcher_performance():
    """Filter out large cookies from the cookie jar."""
    dispatcher = UrlDispatcher()
    for i in range(URL_COUNT):
        dispatcher.add_get(f"/static/sub/path{i}", AsyncMock())

    first_url = "/static/sub/path0"
    last_url = f"/static/sub/path{URL_COUNT-1}"
    long_url = "/static/lots/of/sub/path/segments/0/1/2/3/4/5/6/7/8/9/10/11/12/13/14/15"
    mid_size_url = "/static/midsize/of/sub/path/segments"
    miss_url = "/is/a/miss"
    long_miss_url = "/is/a/miss/with/lots/of/sub/path/segments/0/1/2/3/4/5/6/7/8/9/10/11/12/13/14/15"

    dispatcher.add_get(long_url, AsyncMock())
    dispatcher.add_get(mid_size_url, AsyncMock())

    request_first_url = make_mock_request(first_url)
    match_info = await dispatcher.resolve(request_first_url)
    assert match_info.route.resource.canonical == first_url

    request_last_url = make_mock_request(last_url)
    match_info = await dispatcher.resolve(request_last_url)
    assert match_info.route.resource.canonical == last_url

    request_long_url = make_mock_request(long_url)
    match_info = await dispatcher.resolve(request_long_url)
    assert match_info.route.resource.canonical == long_url

    request_mid_size_url = make_mock_request(mid_size_url)
    match_info = await dispatcher.resolve(request_mid_size_url)
    assert match_info.route.resource.canonical == mid_size_url

    request_miss_url = make_mock_request(miss_url)
    match_info = await dispatcher.resolve(request_miss_url)
    assert match_info.route.status == 404

    request_long_miss_url = make_mock_request(long_miss_url)
    match_info = await dispatcher.resolve(request_long_miss_url)
    assert match_info.route.status == 404

    async def resolve_times(request: web.Request, times: int) -> float:
        start = timeit.default_timer()
        for _ in range(times):
            await dispatcher.resolve(request)
        end = timeit.default_timer()
        return end - start

    run_count = 2000
    resolve_time_first_url = await resolve_times(request_first_url, run_count)
    resolve_time_last_url = await resolve_times(request_last_url, run_count)
    resolve_time_mid_size_url = await resolve_times(request_mid_size_url, run_count)
    resolve_time_long_url = await resolve_times(request_long_url, run_count)
    resolve_time_miss = await resolve_times(request_miss_url, run_count)
    resolve_time_long_miss = await resolve_times(request_long_miss_url, run_count)
    print(f"resolve_time_first_url: {resolve_time_first_url}")
    print(f"resolve_time_last_url: {resolve_time_last_url}")
    print(f"resolve_time_mid_size_url: {resolve_time_mid_size_url}")
    print(f"resolve_time_long_url: {resolve_time_long_url}")
    print(f"resolve_time_miss: {resolve_time_miss}")
    print(f"resolve_time_long_miss: {resolve_time_long_miss}")


asyncio.run(url_dispatcher_performance())
```

before (5002 urls)
```
resolve_time_first_url: 0.0016546659753657877
resolve_time_last_url: 2.253484041953925
resolve_time_mid_size_url: 2.1529546250239946
resolve_time_long_url: 2.136738500033971
resolve_time_miss: 2.152159374963958
resolve_time_long_miss: 2.173428333015181
```
before (7 urls)
```
resolve_time_first_url: 0.0019221249967813492
resolve_time_last_url: 0.0038797500310465693
resolve_time_mid_size_url: 0.004664750013034791
resolve_time_long_url: 0.004066875029820949
resolve_time_miss: 0.00846966594690457
resolve_time_long_miss: 0.008330333046615124
```

after (5002 urls)
```
resolve_time_first_url: 0.0019938750192523003
resolve_time_last_url: 0.0020484999986365438
resolve_time_mid_size_url: 0.002027958049438894
resolve_time_long_url: 0.0019290419877506793
resolve_time_miss: 0.0035507080028764904
resolve_time_long_miss: 0.007968166959472
```
after (7 urls)
```
resolve_time_first_url: 0.0019937920151278377
resolve_time_last_url: 0.002007792005315423
resolve_time_mid_size_url: 0.0019652920309454203
resolve_time_long_url: 0.0019985410035587847
resolve_time_miss: 0.003473666962236166
resolve_time_long_miss: 0.007722083013504744
```

## Checklist

### DIFF
--- a/CHANGES/7829.misc
+++ b/CHANGES/7829.misc
@@ -1,1 +1,3 @@
-Index resources in the UrlDispatcher to avoid linear search for most cases
+Improved URL handler resolution time by indexing resources in the UrlDispatcher.
+For applications with a large number of handlers, this should increase performance significantly.
+-- by :user:`bdraco`
--- a/CHANGES/7829.misc
+++ b/CHANGES/7829.misc
@@ -1,1 +1,1 @@
-Index resources in the UrlDispatcher to avoid linear search on hit
+Index resources in the UrlDispatcher to avoid linear search for most cases
--- a/CHANGES/7829.misc
+++ b/CHANGES/7829.misc
@@ -1,0 +1,1 @@
+Index resources in the UrlDispatcher to avoid linear search on hit
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1067,7 +1067,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         self._resources.append(resource)
         canonical = resource.canonical
         if "{" in canonical:  # strip at the first { to allow for variables
-            canonical = canonical.partition("{")[0].rstrip("/")
+            canonical = canonical.partition("{")[0].rstrip("/") or "/"
         # There may be multiple resources for a canonical path
         # so we use a list to avoid falling back to a full linear search
         self._resource_index.setdefault(canonical, []).append(resource)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -979,8 +979,24 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         super().__init__()
         self._resources: List[AbstractResource] = []
         self._named_resources: Dict[str, AbstractResource] = {}
+        self._resource_index: dict[str, list[AbstractResource]] = {}
 
     async def resolve(self, request: Request) -> UrlMappingMatchInfo:
+        url_parts = request.rel_url.raw_parts
+        resource_index = self._resource_index
+
+        # Walk the url parts looking for candidates
+        for i in range(len(url_parts), 0, -1):
+            url_part = "/" + "/".join(url_parts[1:i])
+            if (resource_candidates := resource_index.get(url_part)) is not None:
+                for candidate in resource_candidates:
+                    if (
+                        match_dict := (await candidate.resolve(request))[0]
+                    ) is not None:
+                        return match_dict
+
+        # We didn't find any candidates, so will fallback to a linear search
+
         method = request.method
         allowed_methods: Set[str] = set()
 
@@ -1049,6 +1065,12 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                 )
             self._named_resources[name] = resource
         self._resources.append(resource)
+        canonical = resource.canonical
+        if "{" in canonical:  # strip at the first { to allow for variables
+            canonical = canonical.split("{")[0].rstrip("/")
+        # There may be multiple resources for a canonical path
+        # so we use a list to avoid falling back to a full linear search
+        self._resource_index.setdefault(canonical, []).append(resource)
 
     def add_resource(self, path: str, *, name: Optional[str] = None) -> Resource:
         if path and not path.startswith("/"):
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1024,7 +1024,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                 allowed_methods |= allowed
 
         if allowed_methods:
-            return MatchInfoError(HTTPMethodNotAllowed(method, allowed_methods))
+            return MatchInfoError(HTTPMethodNotAllowed(request.method, allowed_methods))
         else:
             return MatchInfoError(HTTPNotFound())
 
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1005,8 +1005,10 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                     else:
                         allowed_methods |= allowed
 
-        # We didn't find any candidates, so we'll try the sub-app resources
-        # which we have to walk in a linear fashion because they have match rules
+        # We didn't find any candidates, so we'll try the matched sub-app resources
+        # which we have to walk in a linear fashion because they have match rules.
+        # For most cases we do not expect there to be many of these since currently
+        # they are only added by add_domain
         method = request.method
         for resource in self._matched_sub_app_resources:
             match_dict, allowed = await resource.resolve(request)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1091,9 +1091,8 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
 
     def _get_resource_index_key(self, resource: AbstractResource) -> str:
         """Return a key to index the resource in the resource index."""
-        canonical = resource.canonical
-        if "{" in canonical:  # strip at the first { to allow for variables
-            canonical = canonical.partition("{")[0].rstrip("/")
+        # strip at the first { to allow for variables
+        canonical = resource.canonical.partition("{")[0].rstrip("/")
         return canonical or "/"
 
     def index_resource(self, resource: AbstractResource) -> None:
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -738,11 +738,6 @@ class PrefixedSubAppResource(PrefixResource):
         return {"app": self._app, "prefix": self._prefix}
 
     async def resolve(self, request: Request) -> _Resolve:
-        if (
-            not request.url.raw_path.startswith(self._prefix2)
-            and request.url.raw_path != self._prefix
-        ):
-            return None, set()
         match_info = await self._app.router.resolve(request)
         match_info.add_app(self._app)
         if isinstance(match_info.http_exception, HTTPMethodNotAllowed):
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1093,8 +1093,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
     def _get_resource_index_key(self, resource: AbstractResource) -> str:
         """Return a key to index the resource in the resource index."""
         # strip at the first { to allow for variables
-        canonical = resource.canonical.partition("{")[0].rstrip("/")
-        return canonical or "/"
+        return resource.canonical.partition("{")[0].rstrip("/") or "/"
 
     def index_resource(self, resource: AbstractResource) -> None:
         """Add a resource to the resource index."""
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -294,13 +294,6 @@ class MatchInfoError(UrlMappingMatchInfo):
     def http_exception(self) -> HTTPException:
         return self._exception
 
-    def freeze(self) -> None:
-        """Freeze the match info.
-
-        MatchInfoErrors are SystemRoutes and
-        shared across all apps so freeze is a noop.
-        """
-
     def __repr__(self) -> str:
         return "<MatchInfoError {}: {}>".format(
             self._exception.status, self._exception.reason
@@ -1226,6 +1219,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         super().freeze()
         for resource in self._resources:
             resource.freeze()
+        self._http_not_found_match.freeze()
 
     def add_routes(self, routes: Iterable[AbstractRouteDef]) -> List[AbstractRoute]:
         """Append routes to route table.
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -294,6 +294,13 @@ class MatchInfoError(UrlMappingMatchInfo):
     def http_exception(self) -> HTTPException:
         return self._exception
 
+    def freeze(self) -> None:
+        """Freeze the match info.
+
+        MatchInfoErrors are SystemRoutes and
+        shared across all apps so freeze is a noop.
+        """
+
     def __repr__(self) -> str:
         return "<MatchInfoError {}: {}>".format(
             self._exception.status, self._exception.reason
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -974,9 +974,6 @@ class RoutesView(Sized, Iterable[AbstractRoute], Container[AbstractRoute]):
         return route in self._routes
 
 
-NOT_FOUND_MATCH_ERROR = MatchInfoError(HTTPNotFound())
-
-
 class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
     NAME_SPLIT_RE = re.compile(r"[.:-]")
 
@@ -986,6 +983,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         self._named_resources: Dict[str, AbstractResource] = {}
         self._resource_index: dict[str, list[AbstractResource]] = {}
         self._matched_sub_app_resources: List[MatchedSubAppResource] = []
+        self._http_not_found_match = MatchInfoError(HTTPNotFound())
 
     async def resolve(self, request: Request) -> UrlMappingMatchInfo:
         resource_index = self._resource_index
@@ -1026,7 +1024,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         if allowed_methods:
             return MatchInfoError(HTTPMethodNotAllowed(request.method, allowed_methods))
 
-        return NOT_FOUND_MATCH_ERROR
+        return self._http_not_found_match
 
     def __iter__(self) -> Iterator[str]:
         return iter(self._named_resources)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -994,7 +994,11 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         resource_index = self._resource_index
         allowed_methods: Set[str] = set()
 
-        # Walk the url parts looking for candidates
+        # Walk the url parts looking for candidates. We walk the url backwards
+        # to ensure the most explicit match is found first. If there are multiple
+        # candidates for a given url part because there are multiple resources
+        # registered for the same canonical path, we resolve them in a linear
+        # fashion to ensure registration order is respected.
         for i in range(len(url_parts), 0, -1):
             url_part = "/" + "/".join(url_parts[1:i])
             if (resource_candidates := resource_index.get(url_part)) is not None:
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -995,7 +995,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                     ) is not None:
                         return match_dict
 
-        # We didn't find any candidates, so will fallback to a linear search
+        # We didn't find any candidates, so we fallback to a linear search
 
         method = request.method
         allowed_methods: Set[str] = set()
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -294,6 +294,13 @@ class MatchInfoError(UrlMappingMatchInfo):
     def http_exception(self) -> HTTPException:
         return self._exception
 
+    def freeze(self) -> None:
+        """Freeze the match info.
+
+        MatchInfoErrors are SystemRoutes and
+        shared across all apps so freeze is a noop.
+        """
+
     def __repr__(self) -> str:
         return "<MatchInfoError {}: {}>".format(
             self._exception.status, self._exception.reason
@@ -1219,7 +1226,6 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         super().freeze()
         for resource in self._resources:
             resource.freeze()
-        self._http_not_found_match.freeze()
 
     def add_routes(self, routes: Iterable[AbstractRouteDef]) -> List[AbstractRoute]:
         """Append routes to route table.
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -987,26 +987,28 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         self._resources: List[AbstractResource] = []
         self._named_resources: Dict[str, AbstractResource] = {}
         self._resource_index: dict[str, list[AbstractResource]] = {}
+        self._matched_sub_app_resources: List[MatchedSubAppResource] = []
 
     async def resolve(self, request: Request) -> UrlMappingMatchInfo:
         url_parts = request.rel_url.raw_parts
         resource_index = self._resource_index
+        allowed_methods: Set[str] = set()
+
         # Walk the url parts looking for candidates
         for i in range(len(url_parts), 0, -1):
             url_part = "/" + "/".join(url_parts[1:i])
             if (resource_candidates := resource_index.get(url_part)) is not None:
                 for candidate in resource_candidates:
-                    if (
-                        match_dict := (await candidate.resolve(request))[0]
-                    ) is not None:
+                    match_dict, allowed = await candidate.resolve(request)
+                    if match_dict is not None:
                         return match_dict
+                    else:
+                        allowed_methods |= allowed
 
-        # We didn't find any candidates, so we fallback to a linear search
-
+        # We didn't find any candidates, so we'll try the sub-app resources
+        # which we have to walk in a linear fashion because they have match rules
         method = request.method
-        allowed_methods: Set[str] = set()
-
-        for resource in self._resources:
+        for resource in self._matched_sub_app_resources:
             match_dict, allowed = await resource.resolve(request)
             if match_dict is not None:
                 return match_dict
@@ -1071,7 +1073,12 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
                 )
             self._named_resources[name] = resource
         self._resources.append(resource)
-        self.index_resource(resource)
+
+        if isinstance(resource, MatchedSubAppResource):
+            # We cannot index match sub-app resources because they have match rules
+            self._matched_sub_app_resources.append(resource)
+        else:
+            self.index_resource(resource)
 
     def _get_resource_index_key(self, resource: AbstractResource) -> str:
         """Return a key to index the resource in the resource index."""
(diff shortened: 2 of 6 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -979,8 +979,24 @@ def __init__(self) -> None:
...
+            url_part = "/" + "/".join(url_parts[1:i])
+            if (resource_candidates := resource_index.get(url_part)) is not None:
+                for candidate in resource_candidates:
+                    if (
+                        match_dict := (await candidate.resolve(request))[0]
+                    ) is not None:
+                        return match_dict
+
+        # We didn't find any candidates, so will fallback to a linear search
+
```
```suggestion
        # We didn't find any candidates, so we fallback to a linear search

```

aiohttp/web_urldispatcher.py
```diff
@@ -979,8 +986,23 @@ def __init__(self) -> None:
...
+        resource_index = self._resource_index
+        # Walk the url parts looking for candidates
+        for i in range(len(url_parts), 0, -1):
+            url_part = "/" + "/".join(url_parts[1:i])
+            if (resource_candidates := resource_index.get(url_part)) is not None:
+                for candidate in resource_candidates:
+                    if (
+                        match_dict := (await candidate.resolve(request))[0]
+                    ) is not None:
+                        return match_dict
```
Dreamsorcerer: I think this changes the resolution order, which could cause problems for a few users, if they were depending on the resolution order previously.

I wonder if we should define the resolution order differently in v4, such as using the most specific match first (which would be similar to several other web standards).

aiohttp/web_urldispatcher.py
```diff
@@ -979,8 +986,23 @@ def __init__(self) -> None:
...
+
+        # We didn't find any candidates, so we fallback to a linear search
```
Dreamsorcerer: Is there a situation where this is going to find a result if we didn't already find one?

I think if this is kept it's going to make performance worse, as any 404s are going to take even longer (and there's plenty of bots looking for non-existent pages).

aiohttp/web_urldispatcher.py
bdraco: It walks it backwards to ensure the most exact url is resolved first, and if they have multiple registered at the same path, it uses a list so they should still be resolved in order.

I guess its possible that someone could register

`/everything` and `/everything/hidden` and never want `/everything/hidden` to work but that seems like it would be a bug?

aiohttp/web_urldispatcher.py
bdraco: > Is there a situation where this is going to find a result if we didn't already find one?
> 
> I think if this is kept it's going to make performance worse, as any 404s are going to take even longer (and there's plenty of bots looking for non-existent pages).

It should only have to hit this for the `add_domain` case. The performance will be just a tiny bit be worse on a miss because it will have to do `O(n) + O(url_parts)` instead of `O(n)`.

aiohttp/web_urldispatcher.py
Dreamsorcerer: > It walks it backwards

Oh yes, I missed the -1.

We probably want to include some new tests to validate the ordering.

> /everything and /everything/hidden and never want /everything/hidden to work but that seems like it would be a bug?

You're probably right, I was thinking about variable routes, but I can't think of a situation where it'd cause a problem.

aiohttp/web_urldispatcher.py
bdraco: I adjusted it so only `MatchedSubAppResource` resources have to pay the price of the linear search

bdraco: 3d0b0f952bd5c2b144de7f6043051782c48e7ddf

bdraco: I will add before and after bench marks for miss as well

bdraco: miss is now fast as well

```
resolve_time_first_url: 0.002494540996849537
resolve_time_last_url: 0.0025858330773189664
resolve_time_miss: 0.007047250051982701
```

aiohttp/web_urldispatcher.py
bdraco: I added a comment to better explain whats going on.  I'll also add some more tests for shadowing

aiohttp/web_urldispatcher.py
bdraco: still need to fix system resources (working on that now)

aiohttp/web_urldispatcher.py
Dreamsorcerer: > You're probably right, I was thinking about variable routes, but I can't think of a situation where it'd cause a problem.

Probably doesn't matter in real usage, but it'd possible to have something like `/{x:.*/b}` vs `/x/{b}/` (where a request to /x/b/ would only be the same if the user had previously put the second route first).

bdraco: I pulled a few from https://docs.aiohttp.org/en/stable/web_quickstart.html#variable-resources

tests/test_web_urldispatcher.py
```diff
@@ -544,3 +544,74 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    r = await client.get("/second/frank/info")
+    assert r.status == 200
+    assert await r.text() == "/second/{user}/info"
+    await r.release()
+
+    r = await client.get("/second/bob/info")
+    assert r.status == 200
+    assert await r.text() == "/second/{user}/info"
+    await r.release()
```
bdraco: This test passes on `master` and fails on this PR.  I think this makes this PR DOA

aiohttp/web_urldispatcher.py
bdraco: Unfortunately the test fails https://github.com/aio-libs/aiohttp/pull/7829/files#r1393194544

While its probably fine for 99.9% of use cases, I think this PR is DOA since it changes behavior

Dreamsorcerer: That test shows the variable path shadowing a fixed path. As you said previously, that'd probably be a user mistake, why would they add an endpoint that can't be accessed?

Dreamsorcerer: The only concern is the edge case I showed above, which passes in your test, as you've ordered the /x/{b}/ route first, but would fail if you ordered it the other way on master. It's enough of an edge case that we can probably accept that risk (but, I'll probably push it back to the 3.10 release).

bdraco: The fixed paths are always going to win out over the dynamic paths. I guess thats OK as a breaking change since I don't think anyone will actually do that in production

bdraco: I adjusted the test to comment, and adjusted the opening text above to reflect the change in behavior.

Happy to wait for this to land in 4.0 as that seems fine for a behavior change.

bdraco: Also will need to adjust https://docs.aiohttp.org/en/stable/web_reference.html#resource

bdraco: adjusted that in https://github.com/aio-libs/aiohttp/pull/7829/commits/edf2f98c5eaa97547db5ff292940a1545cc54523

Will also add a note on fixed over dynamic

CHANGES/7829.misc
```diff
@@ -0,0 +1 @@
+Index resources in the UrlDispatcher to avoid linear search on hit
```
bdraco:
```suggestion
Index resources in the UrlDispatcher to avoid linear search for most cases
```

aiohttp/web_urldispatcher.py
Dreamsorcerer: > The fixed paths are always going to win out over the dynamic paths.

Another case I thought of last night, this doesn't hold with something like `/{x}/{y}/z` vs `/{x}/y/z`. In this case they'll be in the same list and end up back to the old behaviour.

I'm not sure if this is worth fixing, as it'll probably complicate the code further for a minimal edge case, but may affect how we document this...

docs/web_quickstart.rst
```diff
@@ -196,6 +196,12 @@ First one is *optimized*. You have got the idea.
 Variable Resources
 ^^^^^^^^^^^^^^^^^^
 
+.. versionchanged:: 4.0
```
Dreamsorcerer: If I read your benchmarks right, we are seeing 2+ seconds to resolve a URL on an application with 5000 resources (a lot, but not completely unreasonable). That seems completely unacceptable, so I'd be keen to release this in 3.10. I think the behaviour changes are low risk enough to introduce then.

```suggestion
.. versionchanged:: 3.10
```

bdraco: Yes, 2s on my m2 mac laptop.  Its ~6s on my production env actually 🙈

aiohttp/web_urldispatcher.py
bdraco: I'll adjust the docs later today

docs/web_quickstart.rst
Dreamsorcerer: Realistically, this is a DoS vector. I only need to send 1 request to a 404 page every 2 seconds to practically halt the server.

aiohttp/web_urldispatcher.py
bdraco: I was too tired to finish this yesterday, but I've just pushed a change to the docs

aiohttp/web_urldispatcher.py
```diff
@@ -979,12 +986,39 @@ def __init__(self) -> None:
...
         allowed_methods: Set[str] = set()
 
-        for resource in self._resources:
+        # Walk the url parts looking for candidates. We walk the url backwards
+        # to ensure the most explicit match is found first. If there are multiple
+        # candidates for a given url part because there are multiple resources
+        # registered for the same canonical path, we resolve them in a linear
+        # fashion to ensure registration order is respected.
+        for i in range(len(url_parts), 0, -1):
+            url_part = "/" + "/".join(url_parts[1:i])
```
Dreamsorcerer: Maybe I'm overthinking it, but I wonder if there's any optimisation here if we do something along the lines of:
```
url_part = "/" + "/".join(url_parts)
while url_part:
    url_part = url_part.rsplit("/", maxsplit=1)[0]
```

Dreamsorcerer: And, maybe too complicated, but a tree structure might allow bailing out quickly. Maybe it would look something like:
```
candidates = [tree := tree.get(p) for p in url_parts if tree]
for tree in candidates[::-1]:
    for c in tree[None]:  # str keys are used for the prefix matching, None contains candidates?
```

aiohttp/web_urldispatcher.py
```diff
@@ -979,12 +986,39 @@ def __init__(self) -> None:
...
+            if (resource_candidates := resource_index.get(url_part)) is not None:
+                for candidate in resource_candidates:
+                    match_dict, allowed = await candidate.resolve(request)
+                    if match_dict is not None:
+                        return match_dict
+                    else:
+                        allowed_methods |= allowed
```
Dreamsorcerer:
```suggestion
            for candidate in resource_index.get(url_part, ()):
                match_dict, allowed = await candidate.resolve(request)
                if match_dict is not None:
                    return match_dict
                else:
                    allowed_methods |= allowed
```

aiohttp/web_urldispatcher.py
```diff
@@ -979,12 +986,39 @@ def __init__(self) -> None:
...
+
+        #
+        # We didn't find any candidates, so we'll try the matched sub-app
+        # resources which we have to walk in a linear fashion because they
+        # have regex/wildcard match rules and we cannot index them.
+        #
+        # For most cases we do not expect there to be many of these since
+        # currently they are only added by `add_domain`
+        #
+        method = request.method
```
Dreamsorcerer: I don't think this is used?
```suggestion
```

aiohttp/web_urldispatcher.py
```diff
@@ -1050,6 +1084,32 @@ def register_resource(self, resource: AbstractResource) -> None:
...
+            # We cannot index match sub-app resources because they have match rules
+            self._matched_sub_app_resources.append(resource)
+        else:
+            self.index_resource(resource)
+
+    def _get_resource_index_key(self, resource: AbstractResource) -> str:
+        """Return a key to index the resource in the resource index."""
+        canonical = resource.canonical
+        if "{" in canonical:  # strip at the first { to allow for variables
+            canonical = canonical.partition("{")[0].rstrip("/")
```
Dreamsorcerer: This works without the if?
```suggestion
        # strip at the first { to allow for variables
        canonical = resource.canonical.partition("{")[0].rstrip("/")
```

docs/web_reference.rst
```diff
@@ -1865,20 +1865,38 @@ unique *name* and at least one :term:`route`.
...
-   finished.
-4. Otherwise router tries next resource from the *routing table*.
-5. If the end of *routing table* is reached and no *resource* /
-   *route* pair found the *router* returns special :class:`~aiohttp.abc.AbstractMatchInfo`
+1. The router splits the URL and checks the index from longest to shortest.
+   For example, '/one/two/three' will first check the index for
+   '/one/two/three', then '/one/two' and finally '/'.
+2. If the URL part is found in the index, the list of routes for
+   that URL part is iterated over. If a route matches to requested HTTP
+   method (or ``'*'`` wildcard) the route's handler is used as found
```
Dreamsorcerer:
```suggestion
   method (or ``'*'`` wildcard) the route's handler is used as the chosen
```

docs/web_reference.rst
```diff
@@ -1865,20 +1865,38 @@ unique *name* and at least one :term:`route`.
...
+   :term:`web-handler`. The lookup is finished.
+3. If the route is not found in the index, the router tries to find
+   the route in the list of :class:`~aiohttp.web.MatchedSubAppResource`,
+   (current only created from :meth:`~aiohttp.web.Application.add_domain`),
+   and will iterate over the list of
+   :class:`~aiohttp.web.MatchedSubAppResource` in a linear fashion
+   until a match is found.
+4. If no *resource* / *route* pair was found, the *router*
+   returns special :class:`~aiohttp.abc.AbstractMatchInfo`
```
Dreamsorcerer:
```suggestion
   returns the special :class:`~aiohttp.abc.AbstractMatchInfo`
```

docs/web_quickstart.rst
```diff
@@ -196,6 +196,20 @@ First one is *optimized*. You have got the idea.
...
+    if you have two routes ``/a/b`` and ``/a/{name}``, then the first
+    route will always be preferred over the second one.
+
+    If there are multiple dynamic paths with the same fixed prefix,
+    they will be resolved in order of registration.
+
+    For example, if you have two dynamic routes that are prefixed
+    with the fixed ``/users`` path such as ``/users/{x}/{y}/z`` and
+    ``/users/{x}/y/z``, the first one will be preferred over the
+    second one.
```
Dreamsorcerer: Given this is supposed to be a quickstart, and we've got the exact same description in reference, I'd suggest dropping this. (I think I'd like to take a look at dropping most of this quickstart at some point, and merging it with the tutorial).

tests/test_web_urldispatcher.py
```diff
@@ -544,3 +544,79 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    r = await client.get("/fifth/21")
+    assert r.status == 200
+    assert await r.text() == "/fifth/{name}"
+    await r.release()
+
+    r = await client.get("/fifth/42")
+    assert r.status == 200
+    assert await r.text() == "/fifth/42"
+    await r.release()
```
Dreamsorcerer, line 628: Strictly speaking, these aren't necessary. The connection is already released once .text() returns.

aiohttp/web_urldispatcher.py
Dreamsorcerer: I'm just throwing around ideas here, feel free to ignore them. But, might be interesting to see if there's some further performance gains to be had.

Dreamsorcerer (changes requested):

aiohttp/web_urldispatcher.py
bdraco: Yup, looks stale

docs/web_quickstart.rst
bdraco: removed in 2bac4583ee22a55a7ea2a7f3316fe9a8746952e1

tests/test_web_urldispatcher.py
bdraco: d126a828a821f828b41f10496d00502d1b51aee7

aiohttp/web_urldispatcher.py
bdraco: optimized in 5810f02750c88dfa6df7c3166eb5c0242f2c502c

bdraco: before
```
resolve_time_first_url: 0.002494540996849537
resolve_time_last_url: 0.0025858330773189664
resolve_time_miss: 0.007047250051982701
```

after
```
resolve_time_first_url: 0.0019859170133713633
resolve_time_last_url: 0.0020277079893276095
resolve_time_miss: 0.00666537499637343
```

bdraco: Looks like a little too much .. tests are failing now..

bdraco: Needed to use the `raw_path`

bdraco: and that seems to be a bit faster as well
```
resolve_time_first_url: 0.0018852080102078617
resolve_time_last_url: 0.0018728750001173466
resolve_time_miss: 0.005776290985522792
```

Dreamsorcerer: Nice, so looks like it may be performing better than the original code even in the first_url condition now.

aiohttp/web_urldispatcher.py
Dreamsorcerer, line 1: Checking the coverage report, it seems to suggest the if statement in resolve() is no longer being run:
https://github.com/bdraco/aiohttp/blob/fast_url_dispatcher/aiohttp/web_urldispatcher.py#L741-L745

Is this because the index ensures we never call it under these conditions, and therefore we can remove that code now?

bdraco: Need to check if https://github.com/bdraco/aiohttp/blob/c83d3a1c1e2911f1c941270a9ee134a761d920a5/aiohttp/web_urldispatcher.py#L619 is still reachable and if so which test hits it... doing that now

bdraco: `test_prefixed_subapp_overlap` and `test_prefixed_subapp_empty_route` are what hit it before

bdraco: Its unreachable now since we always try the longer url first

aiohttp/web_urldispatcher.py
bdraco: Thanks for pushing me to optimize it more. Almost nobody ever does that in reviews.

I'm working on down streaming the changes today so I can validate in my production

Dreamsorcerer: By the way, I only just realised while I was thinking about it overnight, but the benchmarks are biased.
Rather than first/last URL, it should really be best/worst time.
For the old code, the worst time will be the last URL. But, for the new code the worst time will be the one that is most nested and needs to go through multiple checks. The miss time is probably a reasonable estimate of the worst case, though it is really dependent on the layout of a given app.

bdraco: I updated the benchmarks to include a few more cases

bdraco: I added 7 url case and 5002 url case.

The new code still beats the old 7 url case (although it doesn't make much difference)

I guess we could probably optimize the 404 handler a bit more

bdraco: I sped up the 404 case quite a bit

bdraco: before (5002 urls)
```
resolve_time_first_url: 0.0016546659753657877
resolve_time_last_url: 2.253484041953925
resolve_time_mid_size_url: 2.1529546250239946
resolve_time_long_url: 2.136738500033971
resolve_time_miss: 2.152159374963958
resolve_time_long_miss: 2.173428333015181
```
before (7 urls)
```
resolve_time_first_url: 0.0019221249967813492
resolve_time_last_url: 0.0038797500310465693
resolve_time_mid_size_url: 0.004664750013034791
resolve_time_long_url: 0.004066875029820949
resolve_time_miss: 0.00846966594690457
resolve_time_long_miss: 0.008330333046615124
```

after (5002 urls)
```
resolve_time_first_url: 0.001852833025623113
resolve_time_last_url: 0.0018547910149209201
resolve_time_mid_size_url: 0.0019565830007195473
resolve_time_long_url: 0.001857625029515475
resolve_time_miss: 0.0009622920188121498
resolve_time_long_miss: 0.005029833992011845
```
after (7 urls)
```
resolve_time_first_url: 0.001815625000745058
resolve_time_last_url: 0.0018432079814374447
resolve_time_mid_size_url: 0.001861582975834608
resolve_time_long_url: 0.0017958339885808527
resolve_time_miss: 0.0009489589720033109
resolve_time_long_miss: 0.005105625023134053
```

Dreamsorcerer: Nice

bdraco: Looks like I'm going to have to revert it since the web.Application handler always adds the app every time and than throws away the handler and refactoring that would probably be outside of scope for this PR

bdraco: Was able to keep at least one object creation reduction so its still faster


after (5002 urls)
```
resolve_time_first_url: 0.0019938750192523003
resolve_time_last_url: 0.0020484999986365438
resolve_time_mid_size_url: 0.002027958049438894
resolve_time_long_url: 0.0019290419877506793
resolve_time_miss: 0.0035507080028764904
resolve_time_long_miss: 0.007968166959472
```
after (7 urls)
```
resolve_time_first_url: 0.0019937920151278377
resolve_time_last_url: 0.002007792005315423
resolve_time_mid_size_url: 0.0019652920309454203
resolve_time_long_url: 0.0019985410035587847
resolve_time_miss: 0.003473666962236166
resolve_time_long_miss: 0.007722083013504744
```

tests/test_web_urldispatcher.py
webknjaz: @Dreamsorcerer I've been sticking these around all over, at some point. The reason is that some things were leaking into the following tests, whenever responses or sessions weren't closed. Sometimes this led to flaky resource warnings that were hard to spot because they failed a different test than the one causing the leak.

Dreamsorcerer: Yeah, but we know they are closed if there is a .read() call completed. It's the other ones that are actually causing problems. Like, for example, this test failed because in some of the test cases it didn't call .read() and could result in the connection not being closed:
https://github.com/aio-libs/aiohttp/pull/7896/files#diff-e708432f3ba422740bfc04230f149eac4ff3787dde29204b6ee22b622a0e8915R77-R79

webknjaz: This must be what I'm remembering: https://github.com/aio-libs/aiohttp/pull/5494/files#diff-e708432f3ba422740bfc04230f149eac4ff3787dde29204b6ee22b622a0e8915R137

PR #7839, opened 2023-11-14
Add note about filtering warning

Dreamsorcerer:


docs/web_advanced.rst
```diff
@@ -394,7 +394,8 @@ and get it back in the :term:`web-handler`::
         data = request.app['my_private_key']
 
 Rather than using :class:`str` keys, we recommend using :class:`AppKey`.
-This is required for type safety (e.g. when checking with mypy)::
+This is required for type safety (e.g. when checking with mypy,
+or for autocomplete in an IDE)::
```
```suggestion
or for auto-complete in an IDE)::
```

docs/web_advanced.rst
```diff
@@ -403,6 +404,16 @@ This is required for type safety (e.g. when checking with mypy)::
         data = request.app[my_private_key]
         # reveal_type(data) -> str
 
+.. note::
+
+   If warnings are explicitly enabled (e.g. ``-We`` or ``-Wall``) then
+   using :class:`str` keys will trigger a warning. Like other warnings
+   this can be filtered out easily if you have no interest in this feature.
```
eoghanmurray, line 411: pet peeve! 'easy' should never be used in documentation; things are easy when you know how to do them, if you are reading documentation then this is prior to you knowing how to do them :)

webknjaz: That's a good point! I occasionally hear tech writers mentioning this, but it's so often overlooked. We need to do better!

docs/web_advanced.rst
```diff
@@ -403,6 +404,16 @@ This is required for type safety (e.g. when checking with mypy)::
...
+   e.g. Add ``-W 'ignore:It is recommended to use web.AppKey instances for keys.'``
```
webknjaz, line 412: I usually prefer to be as narrow as possible when making such exclusions:
```suggestion
   e.g. Add ``-W 'ignore:It is recommended to use web.AppKey instances for keys.:aiohttp.web_exceptions.NotAppKeyWarning:aiohttp.web_app'``
```
(although I'm not sure that the module part of the spec is right, because of the `stacklevel`, maybe drop it from my suggestion)

Can we advertise this? Besides, it's an “advanced” document.

Dreamsorcerer: I'm afraid it's not valid either way, otherwise I would have just used the type alone which is perfectly accurate: https://github.com/python/cpython/issues/66733

I'm tempted to not merge this in, as nobody else has reported any concerns. The original report was from someone using 3.9.0b0, which wasn't filtering out the warnings by default. Nobody who explicitly enabled warnings has expressed any concerns with the presence of these warnings.

docs/web_advanced.rst
Dreamsorcerer: I don't really mind either way, but I personally don't get it. It's easy in the context of the documentation, as you only need to copy a line of code or CLI argument. Plenty of things are still difficult after you know how to do them. The word doesn't add anything of value to the docs though, so it makes sense to remove it for that reason, if we do merge it in.

docs/web_advanced.rst
webknjaz: It's not? For some, I was convinced that I saw this work in cases when importing the exception wasn't causing it...

docs/web_advanced.rst
webknjaz: @Dreamsorcerer it appears we failed to convey the complete context of the sentiment.
The point here is not that it's _objectively_ easy or hard in the context, but how it affects the receiving side of the text (the reader).
When people hear (see) certain words, they have an effect on their perception. It's a subconscious thing. Hearing “just” often makes people _feel_ dumb. They get defensive and don't accept the information as easily. This is not the recommendation for docs only, it's applied to all the teaching contexts (tutoring/mentoring/training/coaching/public speaking etc.)

The solution that is typically suggested is avoiding the use of such trigger words (or phrases). The same goes for saying “easy” / “straightforward” / “obvious” / “intuitive”. All of these have a great potential to discourage and hurt self-confidence of the readers. The less experience one has, the more harmful it is. This is a straight way to diminish people's feeling of achievement.

Instead, of using these, it's recommended to just skip to the explanation, without showing how much the writer is smarter than a reader.

docs/web_advanced.rst
Dreamsorcerer: It basically only works when you manipulate the PYTHONPATH beforehand: https://github.com/python/cpython/issues/66733#issuecomment-1093665963
Otherwise, the warnings filters are resolved before the import paths are resolved. Requiring PYTHONPATH manipulation is not a robust enough experience to recommend in the docs.

Dreamsorcerer: I'll close this for now, as I've not seen any other user reports. Will reopen it if anybody reports any concerns over the warning.

docs/web_advanced.rst
eoghanmurray: @webknjaz I made a toot; i think we can stop annoying Dreamsorcerer any more :)
https://mastodon.green/@eoghanmurray/111521647225842098

PR #7850, opened 2023-11-18
Fix import under PyPy 3.8/3.9 on Windows

### LINKED ISSUE #7848: [3.9] Import fails on Windows PyPy
### Describe the bug

Importing aiohttp 3.9.0 on Windows in PyPy 3.8 fails with an error about `time.gmtime`.

### To Reproduce

* Run PyPy 3.8 on Windows
* import aiohttp

### Expected behavior

No crash

### Logs/tracebacks

```python-traceback
.tox\ci-pypy3\lib\site-packages\aiohttp\__init__.py:6: in <module>
    from .client import (
.tox\ci-pypy3\lib\site-packages\aiohttp\client.py:76: in <module>
    from .cookiejar import CookieJar
.tox\ci-pypy3\lib\site-packages\aiohttp\cookiejar.py:39: in <module>
    class CookieJar(AbstractCookieJar):
.tox\ci-pypy3\lib\site-packages\aiohttp\cookiejar.py:64: in CookieJar
    calendar.timegm(time.gmtime(MAX_TIME))
E   ValueError: Invalid argument
```

(Seen in [Black's CI](https://github.com/psf/black/actions/runs/6914280784/job/18811967454?pr=4010))
```


### Python Version

```console
Successfully set up PyPy 7.3.11 with Python (3.8.16)
```

### aiohttp Version

```console
Successfully installed aiohttp-3.9.0 aiosignal-1.3.1 async-timeout-4.0.3 attrs-23.1.0 black-0.1.dev1+g7e1ebdd click-8.1.7 frozenlist-1.4.0 idna-3.4 multidict-6.0.4 mypy-extensions-1.0.0 pathspec-0.11.2 typing-extensions-4.8.0 yarl-1.9.2
```

### multidict Version

```console
multidict-6.0.4
```

### yarl Version

```console
yarl-1.9.2
```

### OS

Windows

### Related component

Server

### Additional context

This seems related to #7824. It already catches OSError; I suspect PyPy just throws a different exception here.

### Code of Conduct

JelleZijlstra:
Fixes #7848.

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -65,7 +65,7 @@ class CookieJar(AbstractCookieJar):
     except (OSError, ValueError):
         # Hit the maximum representable time on Windows
         # https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64
-        # Throws ValueError on PyPy 3.8, OSError elsewhere
+        # Throws ValueError on PyPy 3.8 and 3.9, OSError elsewhere
         MAX_TIME = calendar.timegm((3000, 12, 31, 23, 59, 59, -1, -1, -1))
     except OverflowError:
         # #4515: datetime.max may not be representable on 32-bit platforms
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -62,9 +62,10 @@ class CookieJar(AbstractCookieJar):
     )
     try:
         calendar.timegm(time.gmtime(MAX_TIME))
-    except OSError:
+    except (OSError, ValueError):
         # Hit the maximum representable time on Windows
         # https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64
+        # Throws ValueError on PyPy 3.8, OSError elsewhere
         MAX_TIME = calendar.timegm((3000, 12, 31, 23, 59, 59, -1, -1, -1))
     except OverflowError:
         # #4515: datetime.max may not be representable on 32-bit platforms
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -153,7 +153,7 @@ jobs:
           - os: windows
             no-extensions: 'Y'
         include:
-          - pyver: pypy-3.9
+          - pyver: pypy-3.8
             no-extensions: 'Y'
             os: ubuntu
             experimental: false
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -155,7 +155,7 @@ jobs:
         include:
           - pyver: pypy-3.8
             no-extensions: 'Y'
-            os: ubuntu
+            os: windows
             experimental: false
           - os: macos
             pyver: "3.11"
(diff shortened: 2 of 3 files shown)


JelleZijlstra:
OK https://github.com/aio-libs/aiohttp/actions/runs/6914844384/job/18813153994?pr=7850 reproduces the issue. Now pushing a commit to try to fix it.

Dreamsorcerer:
Well, we can probably merge the fix in here. But, maybe need to consider another PR to add Pypy-Windows to the CI, looks like there are a few other failing tests which will need to be fixed or skipped.

aiohttp/cookiejar.py
```diff
@@ -62,9 +62,10 @@ class CookieJar(AbstractCookieJar):
     )
     try:
         calendar.timegm(time.gmtime(MAX_TIME))
-    except OSError:
+    except (OSError, ValueError):
         # Hit the maximum representable time on Windows
         # https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64
+        # Throws ValueError on PyPy 3.8, OSError elsewhere
```
AlexWaygood: Seems like it also throws the same error on Windows/PyPy-3.9. The latest commit in https://github.com/psf/black/pull/4035 bumps PyPy to 3.9 in black's CI, but there's still a `ValueError` being raised in the Windows/PyPy job: https://github.com/psf/black/actions/runs/6915180334/job/18813836767?pr=4035

JelleZijlstra:
@AlexWaygood pointed out that 3.8 is EOL so I switched CI back to pypy3.9. If we're lucky, that also makes some of the other errors go away.

I saw two categories of errors in the CI run with Windows PyPy 3.8:

1. Overlapped
```
  File "C:\hostedtoolcache\windows\PyPy\3.8.16\x86\Lib\asyncio\windows_events.py", line 612, in sendfile
    ov.TransmitFile(sock.fileno(),
AttributeError: 'Overlapped' object has no attribute 'TransmitFile'
```
That seems like a bug in pypy. Possibly can be worked around by not using sendfile() on Windows PyPy.

2. `NotImplementedError: symlink() is not implemented for PyPy on Windows`
Those tests presumably just need to be skipped.

There are also some other errors with less clear causes (e.g. `aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed`). Could be secondary effects of the above two, but could also be separate issues.

On 3.9 the symlink errors went away but the Overlapped one are still around.

I'll switch this PR back to running CI on Linux, and open an issue about Windows PyPy support.

JelleZijlstra:
OK this is ready. There are some other CI failures on 3.11 or 3.12 but they don't look related.

AlexWaygood:
The PR title should probably change ;)

webknjaz:
> There are also some other errors with less clear causes (e.g. `aiohttp.client_exceptions.ClientPayloadError: Response payload is not completed`). Could be secondary effects of the above two, but could also be separate issues.

FTR there are a number of issues with this exception reported but unclear causes happening under normal CPython (like #4581). So I ended up releasing v3.9.4rc0 that attaches those causes to the exceptions when they are raised (via #8049 and #8089).

PR #7862, opened 2023-11-21
Ensure separate ZLibCompressor instances per coroutine when using max_sync_chunk_size

### LINKED ISSUE #7859: ZLibCompressor issue leading to Websocket anomalies
### Describe the bug

After upgrading aiohttp to version 3.9.0, I encountered an intermittent error where the browser reports an 'Invalid frame header' during websocket access. 
On delving deeper, I found that the websocket module in aiohttp, when compressing the data, replaced zlib.compressobj with ZLibCompressor to support max_sync_chunk_size. The specific change can be traced to this commit: https://github.com/aio-libs/aiohttp/commit/586778f19b67a26a7d7ed9cd20add9b5b4dea96c. However, ZLibCompressor was only initialized once in the WebSocketWriter and shared across multiple coroutines. However, its logic does not support usage in multiple coroutines. When different coroutines simultaneously call the websocket to send messages, the compressed results get mixed up, leading to frame inconsistencies.
After replacing ZLibCompressor with an older version of zlib.compressobj(mostly remove await on compressobj.compress to avoid coroutine switching), everything returned to normal.

### To Reproduce

```
from aiohttp.compression_utils import ZLibCompressor
import zlib
import asyncio
import random
import string

compressobj = ZLibCompressor(
    level=zlib.Z_BEST_SPEED,
    wbits=-15,
    max_sync_chunk_size=5 * 1024,
)


def generate_random_string(length):
    letters = string.ascii_lowercase
    return ''.join(random.choice(letters) for i in range(length))


async def test(compressobj):
    # test code from aiohttp.http_websocket.py: _send_frame
    # https://github.com/aio-libs/aiohttp/commit/586778f19b67a26a7d7ed9cd20add9b5b4dea96c#diff-29db742fa68fd59f39e1ca70379c4d478101a8ed9af8895c8fc7604bbf1b4b11R625
    message = generate_random_string(1232323).encode()
    message = await compressobj.compress(message)  # await cause coroutine switching
    message += compressobj.flush(zlib.Z_SYNC_FLUSH)  # data maybe from other coroutine
    return message


# parallel running 10 coroutines
tasks = [test(compressobj) for i in range(10)]
messages = loop.run_until_complete(asyncio.gather(*tasks))

for message in messages:
    decompressor = zlib.decompressobj(wbits=-15)
    message = decompressor.decompress(message) # E   zlib.error: Error -3 while decompressing data: invalid stored block lengths
```

### Expected behavior

correct compression

### Logs/tracebacks

```python-traceback
Invalid frame header
```

### Python Version

```console
Python 3.11.1
```

### aiohttp Version

```console
Name: aiohttp
Version: 3.9.0
Author: 
Author-email: 
License: Apache 2
Location: /Users/xiaohuanshu/PycharmProjects/bazhen/venv/lib/python3.11/site-packages
Required-by: aiohttp-debugtoolbar, aiohttp-devtools, aiohttp-jinja2, aiohttp-pydantic, pytest-aiohttp
```

### multidict Version

```console
Name: multidict
Version: 6.0.4
Author: Andrew Svetlov
License: Apache 2
Location: /Users/xiaohuanshu/PycharmProjects/bazhen/venv/lib/python3.11/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
Name: yarl
Version: 1.9.2
Author: Andrew Svetlov
License: Apache-2.0
Location: /Users/xiaohuanshu/PycharmProjects/bazhen/venv/lib/python3.11/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

macOS

### Related component

Server

### Additional context

_No response_

### Code of Conduct

xiaohuanshu:
## What do these changes do?

fix #7859 

Ensure separate ZLibCompressor instances per coroutine when using max_sync_chunk_size

None

## Related issue number

Related issue :#7859 

## Checklist

### DIFF
--- a/tests/test_websocket_writer.py
+++ b/tests/test_websocket_writer.py
@@ -87,7 +87,7 @@ async def test_send_compress_text(protocol: Any, transport: Any) -> None:
     await writer.send(b"text")
     writer.transport.write.assert_called_with(b"\xc1\x06*I\xad(\x01\x00")
     await writer.send(b"text")
-    writer.transport.write.assert_called_with(b"\xc1\x05*\x01b\x00\x00")
+    writer.transport.write.assert_called_with(b"\xc1\x06*I\xad(\x01\x00")
 
 
 async def test_send_compress_text_notakeover(protocol: Any, transport: Any) -> None:
--- a/CHANGES/7859.bugfix
+++ b/CHANGES/7859.bugfix
@@ -1,0 +1,1 @@
+Ensure separate ZLibCompressor instances per coroutine when using max_sync_chunk_size
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -175,6 +175,7 @@ Jeongkyu Shin
 Jeroen van der Heijden
 Jesus Cea
 Jian Zeng
+Jing Yan
 Jinkyu Yi
 Joel Watts
 John Parton
--- a/aiohttp/compression_utils.py
+++ b/aiohttp/compression_utils.py
@@ -67,6 +67,8 @@ class ZLibCompressor(ZlibBaseHandler):
         return self._compressor.compress(data)
 
     async def compress(self, data: bytes) -> bytes:
+        # In a multi-coroutine environment where max_sync_chunk_size is in use,
+        # ensure that each coroutine has its own compressor instance.
         if (
             self._max_sync_chunk_size is not None
             and len(data) > self._max_sync_chunk_size
@@ -99,6 +101,8 @@ class ZLibDecompressor(ZlibBaseHandler):
         return self._decompressor.decompress(data, max_length)
 
     async def decompress(self, data: bytes, max_length: int = 0) -> bytes:
+        # In a multi-coroutine environment where max_sync_chunk_size is in use,
+        # ensure that each coroutine has its own decompressor instance.
         if (
             self._max_sync_chunk_size is not None
             and len(data) > self._max_sync_chunk_size
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -609,7 +609,6 @@ class WebSocketWriter:
         self._closing = False
         self._limit = limit
         self._output_size = 0
-        self._compressobj: Any = None  # actually compressobj
 
     async def _send_frame(
         self, message: bytes, opcode: int, compress: Optional[int] = None
@@ -632,13 +631,11 @@ class WebSocketWriter:
                     max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
                 )
             else:  # self.compress
-                if not self._compressobj:
-                    self._compressobj = ZLibCompressor(
-                        level=zlib.Z_BEST_SPEED,
-                        wbits=-self.compress,
-                        max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
-                    )
-                compressobj = self._compressobj
+                compressobj = ZLibCompressor(
+                    level=zlib.Z_BEST_SPEED,
+                    wbits=-self.compress,
+                    max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
+                )
 
             message = await compressobj.compress(message)
             message += compressobj.flush(


Dreamsorcerer:
@mykola-mokhnach

tests/test_websocket_writer.py
```diff
@@ -87,7 +87,7 @@ async def test_send_compress_text(protocol: Any, transport: Any) -> None:
     await writer.send(b"text")
     writer.transport.write.assert_called_with(b"\xc1\x06*I\xad(\x01\x00")
     await writer.send(b"text")
-    writer.transport.write.assert_called_with(b"\xc1\x05*\x01b\x00\x00")
+    writer.transport.write.assert_called_with(b"\xc1\x06*I\xad(\x01\x00")
```
Dreamsorcerer, line 90: Does this change signify that we are now losing some compression effectiveness, by starting a new compression each time? Whereas the old code was able to compress over time, and therefore send less bytes (I'm counting 1 more byte in this example).

Should we consider using a Queue or a Lock instead around:
https://github.com/aio-libs/aiohttp/blob/ccf74bbea6902a14c3a139ac836e3595c6b34c07/aiohttp/compression_utils.py#L74

Not sure exactly what the performance tradeoffs will be, so might need a little investigating.

This presumably needs something similar in the decompress one as well.

Dreamsorcerer: A quick test suggests that using a lock is less CPU-intensive than recreating an instance everytime:

```python
import asyncio
import zlib
from aiohttp.compression_utils import ZLibCompressor

self_compressobj = ZLibCompressor(level=zlib.Z_BEST_SPEED, wbits=-9, max_sync_chunk_size=1024)
lock = asyncio.Lock()

async def recreate():
    compressobj = ZLibCompressor(level=zlib.Z_BEST_SPEED, wbits=-9, max_sync_chunk_size=1024)

async def reuse():
    async with lock:
        if not self_compressobj:
            raise Exception()
        compressobj = self_compressobj
```

```bash
> python -m timeit -n 1000 -r 100 'import test, asyncio; loop = asyncio.get_event_loop()' 'loop.run_until_complete(test.recreate())'
1000 loops, best of 100: 58.7 usec per loop
> python -m timeit -n 1000 -r 100 'import test, asyncio; loop = asyncio.get_event_loop()' 'loop.run_until_complete(test.recreate())'
1000 loops, best of 100: 56.1 usec per loop
> python -m timeit -n 1000 -r 100 'import test, asyncio; loop = asyncio.get_event_loop()' 'loop.run_until_complete(test.reuse())'
1000 loops, best of 100: 53.1 usec per loop
> python -m timeit -n 1000 -r 100 'import test, asyncio; loop = asyncio.get_event_loop()' 'loop.run_until_complete(test.reuse())'
1000 loops, best of 100: 53.9 usec per loop
```

bdraco: I think it's possible to avoid the overhead of the lock for the common case where the packet is small by doing the compress, flush, and write without yielding to the event loop which would only make the packets that have to be compressed in the executor incur the overhead of the lock

bdraco: sample pseudo implementation

```diff
diff --git a/aiohttp/http_websocket.py b/aiohttp/http_websocket.py
index 3f124e17..ffd18647 100644
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -640,6 +640,13 @@ class WebSocketWriter:
                     )
                 compressobj = self._compressobj
 
+            if len(message) < compressobj._max_sync_chunk_size:
+                # Compress message if it is smaller than max sync chunk size
+                # without awaiting to ensure that the message written before
+                # the next message can be compressed.
+                message = compressobj.compress_sync(message)
+            else:
+                async with self._compress_lock:
                     message = await compressobj.compress(message)
             message += compressobj.flush(
                 zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
```

bdraco: As long as there are no await between when the message is compressed and written with `_write`, consistency should be ensured, we retain the original performance profile, and only the large message case has to pay for the lock

bdraco: It might be required to refactor it more since the compress, flush, and write likely need to be done under the lock for the await case, but when there is no await it can happen without the lock as before

xiaohuanshu: In your test code, the difference lies only in whether to recreate instances, so reuse is definitely faster than recreating. However, the test should cover the compression process, otherwise the meaning of asynchronous acceleration of ZlibCompressor will be lost.
So I think recreating an instance every time is better, the time taken to recreate is negligible compared to the overall compression process. In the case of concurrent multiple coroutines, it can better utilize CPU performance to accelerate. There are many cases where ZlibCompressor is recreated each time in other parts of aiohttp.

bdraco: so it ends up with something more complex to ensure the compress/flush/writes all either happen under the lock or without yielding to the event loop
```diff
diff --git a/aiohttp/http_websocket.py b/aiohttp/http_websocket.py
index 3f124e17..690969ba 100644
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -610,6 +610,7 @@ class WebSocketWriter:
         self._limit = limit
         self._output_size = 0
         self._compressobj: Any = None  # actually compressobj
+        self._compress_lock = asyncio.Lock()
 
     async def _send_frame(
         self, message: bytes, opcode: int, compress: Optional[int] = None
@@ -640,14 +641,37 @@ class WebSocketWriter:
                     )
                 compressobj = self._compressobj
 
+            if len(message) <= compressobj._max_sync_chunk_size:
+                # Compress message if it is smaller than max sync chunk size
+                # without awaiting to ensure that the message written before
+                # the next message can be compressed.
+                message = compressobj.compress_sync(message)
+                self._write_compressed_message(message, opcode, compressobj, rsv)
+            else:
+                async with self._compress_lock:
                     message = await compressobj.compress(message)
+                    self._write_compressed_message(message, opcode, compressobj, rsv)
+
+        else:
+            self._write_message(message, opcode, rsv)
+
+        if self._output_size > self._limit:
+            self._output_size = 0
+            await self.protocol._drain_helper()
+
+    def _write_compressed_message(
+        self, message: bytes, opcode: int, compressobj: ZLibCompressor, rsv: int
+    ) -> None:
         message += compressobj.flush(
             zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
         )
         if message.endswith(_WS_DEFLATE_TRAILING):
             message = message[:-4]
         rsv = rsv | 0x40
+        self._write_message(message, opcode, rsv)
 
+    def _write_message(self, message: bytes, opcode: int, rsv: int) -> None:
+        """Write a complete websocket frame to the transport."""
         msg_length = len(message)
 
         use_mask = self.use_mask
@@ -678,10 +702,6 @@ class WebSocketWriter:
 
             self._output_size += len(header) + len(message)
 
-        if self._output_size > self._limit:
-            self._output_size = 0
-            await self.protocol._drain_helper()
-
     def _write(self, data: bytes) -> None:
         if self.transport.is_closing():
             raise ConnectionResetError("Cannot write to closing transport")

```

xiaohuanshu: > so it ends up with something more complex to ensure the compress/flush/writes all either happen under the lock or without yielding to the event loop
> 
> ```diff
> diff --git a/aiohttp/http_websocket.py b/aiohttp/http_websocket.py
> index 3f124e17..85cbe826 100644
> --- a/aiohttp/http_websocket.py
> +++ b/aiohttp/http_websocket.py
> @@ -640,14 +640,35 @@ class WebSocketWriter:
>                      )
>                  compressobj = self._compressobj
>  
> +            if len(message) < compressobj._max_sync_chunk_size:
> +                # Compress message if it is smaller than max sync chunk size
> +                # without awaiting to ensure that the message written before
> +                # the next message can be compressed.
> +                message = compressobj.compress_sync(message)
> +                self._write_compressed_message(message, opcode, compressobj)
> +            else:
> +                async with self._compress_lock:
>                      message = await compressobj.compress(message)
> +                    self._write_compressed_message(message, opcode, compressobj)
> +
> +        else:
> +            self._write_message(message, opcode, rsv)
> +
> +        await self._maybe_drain()
> +
> +    def _write_compressed_message(
> +        self, message: bytes, opcode: int, compressobj: ZLibCompressor
> +    ) -> None:
>          message += compressobj.flush(
>              zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
>          )
>          if message.endswith(_WS_DEFLATE_TRAILING):
>              message = message[:-4]
>          rsv = rsv | 0x40
> +        self._write_message(message, opcode, rsv)
>  
> +    def _write_message(self, message: bytes, opcode: int, rsv: int) -> None:
> +        """Write a complete websocket frame to the transport."""
>          msg_length = len(message)
>  
>          use_mask = self.use_mask
> @@ -678,6 +699,7 @@ class WebSocketWriter:
>  
>              self._output_size += len(header) + len(message)
>  
> +    async def _maybe_drain(self) -> None:
>          if self._output_size > self._limit:
>              self._output_size = 0
>              await self.protocol._drain_helper()
> ```

I agree with your point. I prefer to operate without yielding to the event loop, as using a lock may lead to performance loss.

xiaohuanshu: The time to refactor the whole design will be long, but websocket of aiohttp 3.9.0 is completely unusable now. The urgent issue is how to temporarily solve it and release it as soon as possible. I think either we should roll back the https://github.com/aio-libs/aiohttp/commit/586778f19b67a26a7d7ed9cd20add9b5b4dea96c commit, or temporarily merge my PR first?

bdraco: Also recreating the compressor each time will likely break context takeover

bdraco: slightly cleaner version with comments about what is still TODO
```diff
diff --git a/aiohttp/http_websocket.py b/aiohttp/http_websocket.py
index 3f124e17..b18cbbbf 100644
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -610,6 +610,7 @@ class WebSocketWriter:
         self._limit = limit
         self._output_size = 0
         self._compressobj: Any = None  # actually compressobj
+        self._compress_lock = asyncio.Lock()
 
     async def _send_frame(
         self, message: bytes, opcode: int, compress: Optional[int] = None
@@ -640,14 +641,40 @@ class WebSocketWriter:
                     )
                 compressobj = self._compressobj
 
-            message = await compressobj.compress(message)
-            message += compressobj.flush(
-                zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
-            )
-            if message.endswith(_WS_DEFLATE_TRAILING):
-                message = message[:-4]
-            rsv = rsv | 0x40
+            # TODO: do not check ._max_sync_chunk_size twice since
+            # its already checked in .compress()
+            # TODO: do not access ._max_sync_chunk_size directly
+            if len(message) <= compressobj._max_sync_chunk_size:
+                # Compress message if it is smaller than max sync chunk size
+                # without awaiting to ensure that the message written before
+                # the next message can be compressed.
+                message = compressobj.compress_sync(message)
+                self._write_compressed_message(message, opcode, compressobj, rsv)
+            else:
+                async with self._compress_lock:
+                    message = await compressobj.compress(message)
+                    self._write_compressed_message(message, opcode, compressobj, rsv)
+
+        else:
+            self._write_message(message, opcode, rsv)
+
+        if self._output_size > self._limit:
+            self._output_size = 0
+            await self.protocol._drain_helper()
+
+    def _write_compressed_message(
+        self, message: bytes, opcode: int, compressobj: ZLibCompressor, rsv: int
+    ) -> None:
+        message += compressobj.flush(
+            zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
+        )
+        if message.endswith(_WS_DEFLATE_TRAILING):
+            message = message[:-4]
+        rsv = rsv | 0x40
+        self._write_message(message, opcode, rsv)
 
+    def _write_message(self, message: bytes, opcode: int, rsv: int) -> None:
+        """Write a complete websocket frame to the transport."""
         msg_length = len(message)
 
         use_mask = self.use_mask
@@ -662,6 +689,7 @@ class WebSocketWriter:
             header = PACK_LEN2(0x80 | rsv | opcode, 126 | mask_bit, msg_length)
         else:
             header = PACK_LEN3(0x80 | rsv | opcode, 127 | mask_bit, msg_length)
+
         if use_mask:
             mask = self.randrange(0, 0xFFFFFFFF)
             mask = mask.to_bytes(4, "big")
@@ -669,18 +697,15 @@ class WebSocketWriter:
             _websocket_mask(mask, message)
             self._write(header + mask + message)
             self._output_size += len(header) + len(mask) + len(message)
-        else:
-            if len(message) > MSG_SIZE:
-                self._write(header)
-                self._write(message)
-            else:
-                self._write(header + message)
+            return
 
-            self._output_size += len(header) + len(message)
+        if len(message) > MSG_SIZE:
+            self._write(header)
+            self._write(message)
+        else:
+            self._write(header + message)
 
-        if self._output_size > self._limit:
-            self._output_size = 0
-            await self.protocol._drain_helper()
+        self._output_size += len(header) + len(message)
 
     def _write(self, data: bytes) -> None:
         if self.transport.is_closing():
```

This passes the existing test suite, but we obviously need another test for the inter-leaved await case

bdraco: I think it would be good to first merge a PR with an xfail test so we have a test that demonstrates the problem so we know we have it fixed when we can remove the xfail

bdraco: attempt at implementing a concurrency safe fix without the lock for the common case: https://github.com/aio-libs/aiohttp/pull/7865

bdraco: seems to work ok but its never hitting this block:
https://github.com/aio-libs/aiohttp/pull/7865/files#diff-29db742fa68fd59f39e1ca70379c4d478101a8ed9af8895c8fc7604bbf1b4b11R643

which is where we have no test coverage, which is how we got here

bdraco: I need to get some sleep as I have a TATL flight back to the US in a few hours.  If you could get a PR going with a test case marked as xfail, that would be very helpful.

xiaohuanshu: > I need to get some sleep as I have a TATL flight back to the US in a few hours. If you could get a PR going with a test case marked as xfail, that would be very helpful.

PR with xfail just submit, https://github.com/aio-libs/aiohttp/pull/7866

bdraco: I've got a working test case that fails on master and passes in https://github.com/aio-libs/aiohttp/pull/7865

Dreamsorcerer: > In your test code, the difference lies only in whether to recreate instances, so reuse is definitely faster than recreating.

My assumption is that the compress call will be the same in both cases (or possibly better with reuse as it might be able to optimise across calls). I just wanted to see if using a lock would add more overhead than your existing solution. It seems to be better performance, so this makes a lot more sense to me. This should both use less CPU and less bandwidth.

> In the case of concurrent multiple coroutines, it can better utilize CPU performance to accelerate.

This seems like a rather rare edge case to prioritise for performance. There is a new instance for each writer, so this could only, possibly, provide improved performance in the rare case that someone has written a websocket handler which then tries to send large messages from multiple tasks. The tradeoff of increased bandwidth means that the end user of an application will likely still get the responses _slower_ than reusing an instance with a lock.

> As long as there are no await between when the message is compressed and written with `_write`, consistency should be ensured, we retain the original performance profile, and only the large message case has to pay for the lock

I think the problem is if you send a long message and then a short message. If the lock doesn't apply to the short message, then it will try to run at the some time as the long message is compressing. So, I think the lock will be needed across the whole thing.

bdraco: > > In your test code, the difference lies only in whether to recreate instances, so reuse is definitely faster than recreating.
> 
> My assumption is that the compress call will be the same in both cases (or possibly better with reuse as it might be able to optimise across calls). I just wanted to see if using a lock would add more overhead than your existing solution. It seems to be better performance, so this makes a lot more sense to me. This should both use less CPU and less bandwidth.
> 

The lock is very cheap (but not free) if there are no waiters which is likely to be the common case, its just flips a flag and flips it back in that case.

> > In the case of concurrent multiple coroutines, it can better utilize CPU performance to accelerate.
> 
> This seems like a rather rare edge case to prioritise for performance. There is a new instance for each writer, so this could only, possibly, provide improved performance in the rare case that someone has written a websocket handler which then tries to send large messages from multiple tasks. The tradeoff of increased bandwidth means that the end user of an application will likely still get the responses _slower_ than reusing an instance with a lock.
> 

If we overlook the fact it will break takeover, this could have a significant impact on systems do not have the cpu time to spare.  zlib is not that fast so I would expect creating a new instance will create a bottleneck.  

> > As long as there are no await between when the message is compressed and written with `_write`, consistency should be ensured, we retain the original performance profile, and only the large message case has to pay for the lock
> 
> I think the problem is if you send a long message and then a short message. If the lock doesn't apply to the short message, then it will try to run at the some time as the long message is compressing. So, I think the lock will be needed across the whole thing.

Right, the long message would be executing in the executor which would return control to the event loop, and than when the short message comes along, and than we don't know if the long message has finished or not so it would be possible for it to get out of order.  It  sadly means we always have to pay for the lock. Thankfully the lock is relatively cheap. 

I adjusted https://github.com/aio-libs/aiohttp/pull/7865

Dreamsorcerer:
Superseded by #7865

PR #7865, opened 2023-11-22
Restore async concurrency safety to websocket compressor

### LINKED ISSUE #7859: ZLibCompressor issue leading to Websocket anomalies

bdraco:
## What do these changes do?

Restore async concurrency safety to websocket compressor

## Related issue number

fixes #7859

## Checklist

### DIFF
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -610,7 +610,6 @@ class WebSocketWriter:
         self._limit = limit
         self._output_size = 0
         self._compressobj: Any = None  # actually compressobj
-        self._compress_lock: Optional[asyncio.Lock] = None
 
     async def _send_frame(
         self, message: bytes, opcode: int, compress: Optional[int] = None
@@ -628,47 +627,19 @@ class WebSocketWriter:
             if compress:
                 # Do not set self._compress if compressing is for this frame
                 compressobj = self._make_compress_obj(compress)
-                message = await compressobj.compress(message)
-                self._write_compressed_message(message, opcode, compressobj, rsv)
             else:  # self.compress
-                # If we are reusing the compress object, we need to hold a lock
-                # around the compress/flush/write to ensure that messages get
-                # compressed in order and takeover is not violated.
                 if not self._compressobj:
                     self._compressobj = self._make_compress_obj(self.compress)
-                    self._compress_lock = asyncio.Lock()
                 compressobj = self._compressobj
-                async with self._compress_lock:
-                    message = await compressobj.compress(message)
-                    self._write_compressed_message(message, opcode, compressobj, rsv)
 
-        else:
-            self._write_message(message, opcode, rsv)
+            message = await compressobj.compress(message)
+            message += compressobj.flush(
+                zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
+            )
+            if message.endswith(_WS_DEFLATE_TRAILING):
+                message = message[:-4]
+            rsv = rsv | 0x40
 
-        if self._output_size > self._limit:
-            self._output_size = 0
-            await self.protocol._drain_helper()
-
-    def _make_compress_obj(self, compress: int) -> ZLibCompressor:
-        return ZLibCompressor(
-            level=zlib.Z_BEST_SPEED,
-            wbits=-compress,
-            max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
-        )
-
-    def _write_compressed_message(
-        self, message: bytes, opcode: int, compressobj: ZLibCompressor, rsv: int
-    ) -> None:
-        message += compressobj.flush(
-            zlib.Z_FULL_FLUSH if self.notakeover else zlib.Z_SYNC_FLUSH
-        )
-        if message.endswith(_WS_DEFLATE_TRAILING):
-            message = message[:-4]
-        rsv = rsv | 0x40
-        self._write_message(message, opcode, rsv)
-
-    def _write_message(self, message: bytes, opcode: int, rsv: int) -> None:
-        """Write a complete websocket frame to the transport."""
         msg_length = len(message)
 
         use_mask = self.use_mask
@@ -683,7 +654,6 @@ class WebSocketWriter:
             header = PACK_LEN2(0x80 | rsv | opcode, 126 | mask_bit, msg_length)
         else:
             header = PACK_LEN3(0x80 | rsv | opcode, 127 | mask_bit, msg_length)
-
         if use_mask:
             mask = self.randrange(0, 0xFFFFFFFF)
             mask = mask.to_bytes(4, "big")
@@ -691,15 +661,25 @@ class WebSocketWriter:
             _websocket_mask(mask, message)
             self._write(header + mask + message)
             self._output_size += len(header) + len(mask) + len(message)
-            return
-
-        if len(message) > MSG_SIZE:
-            self._write(header)
-            self._write(message)
         else:
-            self._write(header + message)
+            if len(message) > MSG_SIZE:
+                self._write(header)
+                self._write(message)
+            else:
+                self._write(header + message)
 
-        self._output_size += len(header) + len(message)
+            self._output_size += len(header) + len(message)
+
+        if self._output_size > self._limit:
+            self._output_size = 0
+            await self.protocol._drain_helper()
+
+    def _make_compress_obj(self, compress: int) -> ZLibCompressor:
+        return ZLibCompressor(
+            level=zlib.Z_BEST_SPEED,
+            wbits=-compress,
+            max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
+        )
 
     def _write(self, data: bytes) -> None:
         if self.transport.is_closing():
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -628,27 +628,19 @@ class WebSocketWriter:
             if compress:
                 # Do not set self._compress if compressing is for this frame
                 compressobj = self._make_compress_obj(compress)
+                message = await compressobj.compress(message)
+                self._write_compressed_message(message, opcode, compressobj, rsv)
             else:  # self.compress
+                # If we are reusing the compress object, we need to hold a lock
+                # around the compress/flush/write to ensure that messages get
+                # compressed in order and takeover is not violated.
                 if not self._compressobj:
                     self._compressobj = self._make_compress_obj(self.compress)
                 compressobj = self._compressobj
-
-            if not compressobj.should_run_in_executor(message):
-                # Compress message if it is smaller than max sync chunk size
-                # without awaiting to ensure that the message written before
-                # the next message can be compressed.
-                message = compressobj.compress_sync(message)
-                self._write_compressed_message(message, opcode, compressobj, rsv)
-            else:
-                # Since we are compressing in an executor, and the await returns
-                # control to the event loop we need to hold a lock to ensure that
-                # the compressed message is written before the next message is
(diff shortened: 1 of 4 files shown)


bdraco:
For testing this I guess we gather a few writers to happen at the same time and than check the received messages are not corrupted

tests/test_websocket_writer.py
```diff
@@ -106,3 +108,60 @@ async def test_send_compress_text_per_message(protocol: Any, transport: Any) ->
...
+        call_bytes = call[0][0]
+        result, _ = reader.feed_data(call_bytes)
+        assert result is False
+        msg = await queue.read()
+        bytes_data: bytes = msg.data
+        assert len(bytes_data) == msg_length
+        assert bytes_data == bytes_data[0:1] * msg_length
+
+
+async def test_concurrent_messages_without_executor(
```
bdraco: I had this as a single test with params but I thought that would make it harder to understand what has gone wrong in the future so I split it into two tests since there are two different branches to test.  If there is a preference to have it a single test, I can put it back.

bdraco:
I think this will solve it.  I need to do more functional testing but my flight is taking off now

bdraco:
I think this will solve it.  I need to do more functional testing but my flight is taking off now

bdraco:
functional testing looks good

doing performance testing now

bdraco:
performance profile remains similar
Screenshot 2023-11-22 at 12 46 15 PM

tests/test_websocket_writer.py
Dreamsorcerer: If the only difference is the chunk size, then maybe just parametrize() it?

tests/test_websocket_writer.py
```diff
@@ -106,3 +108,92 @@ async def test_send_compress_text_per_message(protocol: Any, transport: Any) ->
...
+    are small enough to not be compressed in the executor
+    with messages that need to be compressed in the executor.
+    """
+    writer = WebSocketWriter(protocol, transport, compress=15)
+    queue: DataQueue[WSMessage] = DataQueue(asyncio.get_running_loop())
+    reader = WebSocketReader(queue, 50000)
+    writers = []
+    payloads = []
+    for count in range(1, 64 + 1):
+        payload = bytes((count,)) * count
```
Dreamsorcerer: To be a little more thorough, maybe these should be interleaved, something like:
```suggestion
    for count in range(1, 64 + 1):
        payload = bytes((count,)) * (64 if count % 2 else 8)
```
?

tests/test_websocket_writer.py
Dreamsorcerer: Alternatively, I reckon all 3 could be combined by just parametrizing msg_length (with a callable, so it works with the 3rd case).

bdraco: Yup, that works

bdraco: Yup, that works

aiohttp/http_websocket.py
```diff
@@ -626,28 +627,49 @@ async def _send_frame(
...
                 if not self._compressobj:
-                    self._compressobj = ZLibCompressor(
-                        level=zlib.Z_BEST_SPEED,
-                        wbits=-self.compress,
-                        max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
-                    )
+                    self._compressobj = self._make_compress_obj(self.compress)
                 compressobj = self._compressobj
+                if not self._compress_lock:
+                    self._compress_lock = asyncio.Lock()
```
Dreamsorcerer: Shall we just do this in the above if block along with `self._compressobj`, save us an extra check?

bdraco: Good idea, done

Dreamsorcerer:
@webknjaz I'd like to nominate @bdraco for membership to @aio-libs/aiohttp-dev. There's plenty of examples in the previous weeks of good quality PRs, with all the tests we'd expect: https://github.com/aio-libs/aiohttp/pulls?q=is%3Apr+author%3Abdraco+is%3Aclosed

Dreamsorcerer:
> I'd like to nominate @bdraco for membership

OK, done this properly now: #7880

aiohttp/http_websocket.py
```diff
@@ -626,28 +627,48 @@ async def _send_frame(
...
                 if not self._compressobj:
-                    self._compressobj = ZLibCompressor(
-                        level=zlib.Z_BEST_SPEED,
-                        wbits=-self.compress,
-                        max_sync_chunk_size=WEBSOCKET_MAX_SYNC_CHUNK_SIZE,
-                    )
+                    self._compressobj = self._make_compress_obj(self.compress)
+                    self._compress_lock = asyncio.Lock()
                 compressobj = self._compressobj
+                async with self._compress_lock:
```
Dreamsorcerer: I'm just thinking about this again. Would it be better to have the lock in the Compressor, so the logic is encapsulated and not going to cause issues if used elsewhere?

From my reading, the only yield point of this logic is when it's offloaded to a thread. If there is a lock around that call, then no other coro can enter it until it resumes. Once it resumes it will continue to call the remaining functions without yielding, so it should be safe?

bdraco: I _think_ that is safe since releasing the lock will only [wake up the next waiter](https://github.com/python/cpython/blob/dc0adb44d8d4a33121deaad398f24b5d8ae36d19/Lib/asyncio/locks.py#L137C18-L137C32) and the original should proceed on to flush and writer since [`set_result`](https://github.com/python/cpython/blob/dc0adb44d8d4a33121deaad398f24b5d8ae36d19/Lib/asyncio/locks.py#L154) will result in a [`call_soon`](https://github.com/python/cpython/blob/dc0adb44d8d4a33121deaad398f24b5d8ae36d19/Lib/asyncio/futures.py#L173) from [`__schedule_callbacks`](https://github.com/python/cpython/blob/dc0adb44d8d4a33121deaad398f24b5d8ae36d19/Lib/asyncio/futures.py#L263).  

That might be a implementation detail of the event loop implementation rather than a guarantee though.. I'm not really sure

bdraco: OTOH, I think the only place we need a lock for compression is the websocket and the other use cases are already fine

```
aiohttp/http_writer.py:            chunk = await self._compress.compress(chunk)
aiohttp/http_writer.py:                chunk = await self._compress.compress(chunk)
aiohttp/multipart.py:                chunk = await self._compress.compress(chunk)
```

Dreamsorcerer: I don't think the event loop implementation matters. As long as the lock's release doesn't yield to the event loop (which would be an odd thing to do, I think), then there is no opportunity for another coro to run before we complete the other code.

Dreamsorcerer: > OTOH, I think the only place we need a lock for compression is the websocket and the other use cases are already fine

Is it?

> ```
> aiohttp/http_writer.py:            chunk = await self._compress.compress(chunk)
> aiohttp/http_writer.py:                chunk = await self._compress.compress(chunk)
> aiohttp/multipart.py:                chunk = await self._compress.compress(chunk)
> ```

While I think it's slightly less likely you'd attempt to make multiple simultaneous StreamResponse.write() calls than on a websocket, it's still possible and looks to me like it'd have the same issue? Maybe the same for multipart..

Dreamsorcerer: e.g. I'm making some kind of custom output, where I'm just writing chunks out to the response, as those chunks become available asynchronously, without caring about the ordering of those chunks.

Dreamsorcerer: I think most of the decompress ones are fine, but only due to implementation details. i.e. They seem to be in feed_data() implementations, which I assume are called in a loop or similar as the parser goes through the data.

I think the one in multipart could have a tiny chance of causing issues, but it's been left as a sync method anyway, so will be fine.

bdraco: Assuming you are confident it's safe I can refactor this after the holiday to move the lock into the compressor class

Dreamsorcerer: I'm pretty sure it's safe today. The only question is how might the code be refactored in future. If someone adds an await between the calls, then it'll be unsafe. But, if we don't change, and then someone uses the compressor somewhere else without using a lock, then that'll be unsafe. So, there's a risk either way. To me, keeping it encapsulated seems slightly less worse. :P

bdraco: Haha. I think that's a yes on the refactor. I'll work on it after all the family events

bdraco: I refactored it and added some comments in the hope that it will help avoid problems from potential future refactoring you mention above.

PR #7879, opened 2023-11-23
Fix regression with connection upgrade

### LINKED ISSUE #7867: Response connection is None on 3.9.0
### Describe the bug

Sorry for the not easily reproducible bug, but since 3.9.0 connection upgrading doesn't work

### To Reproduce

I experienced the bug using aiodocker exec, so forgive me for indirect reproduce of the bug:
* get a running docker container in aiodocker
* `exec_ = await container.exec(command)` (the command doesn't matter, can be ls)
* `async with exec_.start(detach=False) as stream:`
* the above line will fail

### Expected behavior

I used aiodocker version 0.21.0, when using aiohttp version 3.8.6 it works well (but uninstallable for python3.12), on version 3.9.0 it fails with  Status code: 101 (https://github.com/aio-libs/aiodocker/issues/832)

### Logs/tracebacks

```python-traceback
Unfortunately, I only got aiodocker logs:

aiodocker.exceptions.DockerError: DockerError(500, "Cannot upgrade connection to vendored tcp protocol, the docker server has closed underlying socket. Status code: 101. Headers: <CIMultiDictProxy('Content-Type': 'application/vnd.docker.multiplexed-stream', 'Connection': 'Upgrade', 'Upgrade': 'tcp', 'Api-Version': '1.43', 'Docker-Experimental': 'false', 'Ostype': 'linux', 'Server': 'Docker/24.0.7 (linux)')>.")
```
```


### Python Version

```console
$ python --version
Tested on 3.12 and 3.11
```

### aiohttp Version

```console
$ python -m pip show aiohttp
works on 3.8.6, fails on 3.9.0
```

### multidict Version

```console
6.0.4
```

### yarl Version

```console
$ python -m pip show yarl
1.9.3
```

### OS

Ubuntu

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #7867.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -940,19 +940,9 @@ class ClientResponse(HeadersMixin):
         if self._closed:
             return
 
-        if self._connection is not None:
-            # websocket, protocol could be None because
-            # connection could be detached
-            if (
-                self._connection.protocol is not None
-                and self._connection.protocol.upgraded
-            ):
-                return
-
-            self._release_connection()
-
         self._closed = True
         self._cleanup_writer()
+        self._release_connection()
 
     @property
     def closed(self) -> bool:
@@ -978,7 +968,7 @@ class ClientResponse(HeadersMixin):
         self._closed = True
 
         self._cleanup_writer()
-        self._release_connection()
+        self._release_connection(skip_upgraded=False)
         return noop()
 
     @property
@@ -1003,8 +993,13 @@ class ClientResponse(HeadersMixin):
                 headers=self.headers,
             )
 
-    def _release_connection(self) -> None:
+    def _release_connection(self, skip_upgraded: bool = True) -> None:
         if self._connection is not None:
+            # protocol could be None because connection could be detached
+            protocol = self._connection.protocol
+            if skip_upgraded and protocol is not None and protocol.upgraded:
+                return
+
             if self._writer is None:
                 self._connection.release()
                 self._connection = None
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -941,7 +941,8 @@ class ClientResponse(HeadersMixin):
             return
 
         # protocol could be None because connection could be detached
-        protocol = self._connection and self._connection.protocol
+        # Mypy bug: https://github.com/python/mypy/issues/16565
+        protocol = self._connection and self._connection.protocol  # type: ignore[union-attr]
         if protocol is not None and protocol.upgraded:
             return
 
@@ -1043,7 +1044,8 @@ class ClientResponse(HeadersMixin):
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        protocol = self._connection and self._connection.protocol
+        # Mypy bug: https://github.com/python/mypy/issues/16565
+        protocol = self._connection and self._connection.protocol  # type: ignore[union-attr]
         if protocol is None or not protocol.upgraded:
             await self._wait_released()  # Underlying connection released
         return self._body
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -941,7 +941,7 @@ class ClientResponse(HeadersMixin):
             return
 
         # protocol could be None because connection could be detached
-        protocol = (self._connection or None) and self._connection.protocol
+        protocol = self._connection and self._connection.protocol
         if protocol is not None and protocol.upgraded:
             return
 
@@ -1043,7 +1043,7 @@ class ClientResponse(HeadersMixin):
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        protocol = (self._connection or None) and self._connection.protocol
+        protocol = self._connection and self._connection.protocol
         if protocol is None or not protocol.upgraded:
             await self._wait_released()  # Underlying connection released
         return self._body
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -940,11 +940,10 @@ class ClientResponse(HeadersMixin):
         if self._closed:
             return
 
-        if self._connection is not None:
-            # protocol could be None because connection could be detached
-            protocol = self._connection.protocol
-            if protocol is not None and protocol.upgraded:
-                return
+        # protocol could be None because connection could be detached
+        protocol = getattr(self._connection, "protocol", None)
+        if protocol is not None and protocol.upgraded:
+            return
 
         self._closed = True
         self._cleanup_writer()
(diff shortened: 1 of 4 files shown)


tests/test_client_functional.py
```diff
@@ -174,6 +174,22 @@ async def handler(request):
     assert 1 == len(client._session.connector._conns)
 
 
+async def test_upgrade_connection_not_released_after_read(aiohttp_client) -> None:
```
```suggestion
async def test_upgrade_connection_not_released_after_read(aiohttp_client: Any) -> None:
```

tests/test_client_functional.py
```diff
@@ -174,6 +174,22 @@ async def handler(request):
     assert 1 == len(client._session.connector._conns)
 
 
+async def test_upgrade_connection_not_released_after_read(aiohttp_client: Any) -> None:
+    async def handler(request):
```
```suggestion
    async def handler(request: web.Request) -> web.Response:
```

tests/test_client_functional.py
```diff
@@ -174,6 +174,22 @@ async def handler(request):
     assert 1 == len(client._session.connector._conns)
 
 
+async def test_upgrade_connection_not_released_after_read(aiohttp_client: Any) -> None:
+    async def handler(request: web.Request) -> web.Response:
+        body = await request.read()
+        assert b"" == body
+        return web.Response(status=101, headers={"Connection": "Upgrade"})
```
```suggestion
        return web.Response(status=101, headers={"Connection": "Upgrade", "Upgrade": "tcp"})
```

aiohttp/client_reqrep.py
```diff
@@ -940,19 +940,14 @@ def _response_eof(self) -> None:
...
-            # websocket, protocol could be None because
-            # connection could be detached
-            if (
-                self._connection.protocol is not None
-                and self._connection.protocol.upgraded
-            ):
-                return
-
-            self._release_connection()
+        # protocol could be None because connection could be detached
```
bdraco, line 943: Can the `getattr` be avoided?

```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index c0319074..2185aa85 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -941,7 +941,7 @@ class ClientResponse(HeadersMixin):
             return
 
         # protocol could be None because connection could be detached
-        protocol = getattr(self._connection, "protocol", None)
+        protocol = self._connection and self._connection.protocol
         if protocol is not None and protocol.upgraded:
             return
 
@@ -1043,7 +1043,7 @@ class ClientResponse(HeadersMixin):
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        protocol = getattr(self._connection, "protocol", None)
+        protocol = self._connection and self._connection.protocol
         if protocol is None or not protocol.upgraded:
             await self._wait_released()  # Underlying connection released
         return self._body
```

Dreamsorcerer: Yeah, I was just thinking about that and trying to decide which is preferred.

bdraco: Looks vscode can work it out, but it looks like mypy didn't like that.

bdraco: I guess you could write it as

```python
    @property
    def _upgraded(self) -> Optional[bool]:
        """Return True if connection was upgraded."""
        return (
            (connection := self._connection)
            and (protocol := connection.protocol)
            and protocol.upgraded
        )
```

Dreamsorcerer: It's a bit awkward, I think. We shouldn't write lower quality code because of a bug in mypy.

bdraco:
```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index 2185aa85..f3e5aeb2 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -758,6 +758,15 @@ class ClientResponse(HeadersMixin):
     def __reset_writer(self, _: object = None) -> None:
         self.__writer = None
 
+    @property
+    def _upgraded(self) -> Optional[bool]:
+        """Return True if connection was upgraded."""
+        return (
+            (connection := self._connection)
+            and (protocol := connection.protocol) is not None
+            and protocol.upgraded
+        )
+
     @property
     def _writer(self) -> Optional["asyncio.Task[None]"]:
         return self.__writer
@@ -937,14 +946,8 @@ class ClientResponse(HeadersMixin):
         return self
 
     def _response_eof(self) -> None:
-        if self._closed:
+        if self._closed or self._upgraded:
             return
-
-        # protocol could be None because connection could be detached
-        protocol = self._connection and self._connection.protocol
-        if protocol is not None and protocol.upgraded:
-            return
-
         self._closed = True
         self._cleanup_writer()
         self._release_connection()
@@ -1043,8 +1046,7 @@ class ClientResponse(HeadersMixin):
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        protocol = self._connection and self._connection.protocol
-        if protocol is None or not protocol.upgraded:
+        if not self._upgraded:
             await self._wait_released()  # Underlying connection released
         return self._body
 
```


Another option, but I think I like the ignore better

bdraco:
I'll throw this on production shortly

bdraco (approved):
LGTM

I put this on two production systems and everything is still working as expected.

bdraco:
I beat it up a bit with a lot of websocket connections and didn't hit any issues

aiohttp/client_reqrep.py
Dreamsorcerer: Hmm, the mypy developers have pointed out that if `Connection` is subclassed, then `__bool__()` could return False, in which case the mypy error is correct.

If it were made Final, then it'd be fine. But, I'm starting to think that the getattr() is the safer option.

bdraco: Somebody will subclass it so the getattr seems safer even if it's less desirable

bdraco: Also everything ran fine overnight. Logs look good. Nothing amiss

Dreamsorcerer: Or, they've suggested `(conn or None) and conn.protocol`.

Dreamsorcerer: The other suggestion is that we could add:
`def __bool__(self) -> Literal[True]: return True`

Which forces all subclasses to not be falsy.

Dreamsorcerer: I'm going to try the latter option, as it'd make code simpler throughout the project (and we probably already made this assumption elsewhere...).

bdraco (approved):

Dreamsorcerer:
I think that's the last regression I've seen. So, I'll try and push a 3.9.1 release today.

Dreamsorcerer:
By the way, if you have any ideas on an easy way to backport all the type annotations from the tests in master, that'd make the automatic backport a lot more reliable. I think doing it manually would be a lot of work, and I've not thought of an easy way to do it.

Currently, editing any test file basically fails the backport as it will conflict with the type annotation differences.

bdraco:
I really can't think of a good way to do it except manually. If I think of something, I'll be sure to bring it up

PR #7895, opened 2023-11-25
Message is not upgraded if Upgrade header is missing

Dreamsorcerer:
It is a required header for a 101 response. Behaviour is already correct in llhttp.

### DIFF
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -503,7 +503,7 @@ class HttpParser(abc.ABC, Generic[_MsgT]):
                 close_conn = True
             elif v == "keep-alive":
                 close_conn = False
-            elif v == "upgrade":
+            elif v == "upgrade" and headers.get(hdrs.UPGRADE):
                 upgrade = True
 
         # encoding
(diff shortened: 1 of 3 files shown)


tests/test_http_parser.py
```diff
@@ -473,6 +473,16 @@ def test_conn_upgrade(parser: Any) -> None:
     assert upgrade
 
 
+def test_bad_upgrade(parser: Any) -> None:
+    """Test not upgraded if missing Upgrade header."""
+
+    text = b"GET /test HTTP/1.1\r\n" b"connection: upgrade\r\n\r\n"
```
```suggestion
    text = b"GET /test HTTP/1.1\r\nconnection: upgrade\r\n\r\n"
```

aiohttp/http_parser.py
```diff
@@ -503,7 +503,7 @@ def parse_headers(
                 close_conn = True
             elif v == "keep-alive":
                 close_conn = False
-            elif v == "upgrade":
+            elif v == "upgrade" and headers.get(hdrs.UPGRADE):
```
Dreamsorcerer, line 507:
```suggestion
            # https://www.rfc-editor.org/rfc/rfc9110.html#name-101-switching-protocols
            elif v == "upgrade" and headers.get(hdrs.UPGRADE):
```

tests/test_http_parser.py
```diff
@@ -473,6 +473,16 @@ def test_conn_upgrade(parser: Any) -> None:
     assert upgrade
 
 
+def test_bad_upgrade(parser: Any) -> None:
+    """Test not upgraded if missing Upgrade header."""
+
```
```suggestion
```

PR #7916, opened 2023-11-27
Add more information to contributing page

Dreamsorcerer:
Mostly revamp coverage section with an example demonstrating how to analyse covecov reports.

I think #7829 works as a perfect example of how codecov is really useful to us.

Preview: https://aiohttp--7916.org.readthedocs.build/en/7916/contributing.html#tests-coverage


docs/contributing.rst
```diff
@@ -161,29 +161,61 @@ Any extra texts (print statements and so on) should be removed.
...
 
-   $ make cov-dev
+Here we see that the PR has introduced 1 miss and 2 partials. Now we
+click the link in the comment header to open the full report:
 
-to run test suite and collect coverage information. Once the command
-has finished check your coverage at the file that appears in the last
-line of the output:
-``open file:///.../aiohttp/htmlcov/index.html``
+.. image:: contributing-con-header.png
```
Dreamsorcerer:
```suggestion
.. image:: contributing-cov-header.png
```

docs/contributing.rst
```diff
@@ -161,29 +161,61 @@ Any extra texts (print statements and so on) should be removed.
...
 
-We are trying hard to have good test coverage; please don't make it worse.
+We use *codecov.io* as an indispensable tool for analysing our coverage
+results. Visit https://codecov.io/gh/aio-libs/aiohttp to see coverage
+reports for the master branch, history, pull requests etc.
 
-Use:
+We'll use an example from a real PR to demonstrate how we use this.
+Once the tests run in a PR, you'll see a comment posted by codecov.
+The most important thing to check here, is whether there are any new
```
bdraco:
```suggestion
The most important thing to check here is whether there are any new
```

docs/contributing.rst
```diff
@@ -161,29 +161,61 @@ Any extra texts (print statements and so on) should be removed.
...
+missed or partial lines in the report:
 
-.. code-block:: shell
+.. image:: contributing-cov-comment.png
 
-   $ make cov-dev
+Here we see that the PR has introduced 1 miss and 2 partials. Now we
```
bdraco:
```suggestion
Here, the PR has introduced 1 miss and 2 partials. Now we
```

bdraco (approved):

docs/contributing-cov-header.png
webknjaz, line 1: @Dreamsorcerer could you try replacing pictures with SVG, where possible? I suppose since this is taken from web, there must be a way to turn HTML into SVG...

docs/contributing.rst
```diff
@@ -161,29 +161,61 @@ Any extra texts (print statements and so on) should be removed.
...
+is also a useful tool for analyzing the coverage directly from *Files
+Changed* tab on the *GitHub Pull Request* review page.
+
+
+You can also produce coverage reports locally with ``make cov-dev``
+or just adding ``--cov-report=html`` to ``pytest``.
+
+This will run the test suite and collect coverage information. Once
+finished, coverage results can be view by opening:
+``open file:///.../aiohttp/htmlcov/index.html``
```
webknjaz: By the way, this is a macOS command which is `xdg-open` in Linux envs.

docs/contributing.rst
```diff
@@ -161,29 +161,61 @@ Any extra texts (print statements and so on) should be removed.
...
+
+Now, if we look through the diff under 'Files changed' we find one of
+our partials:
+
+.. image:: contributing-cov-partial.png
+   :alt: A while loop with partial coverage.
+
+In this case, the while loop is never skipped in our tests. This is
+probably not worth writing a test for (and may be a situation that is
+impossible to trigger anyway), so we leave this alone.
```
webknjaz, line 189: One way of dealing with such cases is using `# pragma: no cover`. This would allow setting higher required coverage level eventually, like this: https://adamj.eu/tech/2019/04/30/getting-a-django-application-to-100-percent-coverage/#the-report-section.

docs/spelling_wordlist.txt
```diff
@@ -7,6 +7,7 @@ aiohttpdemo
 aiohttp’s
 aiopg
 alives
+analysing
```
webknjaz: Would it work if you americanize the spelling?

docs/contributing.rst
```diff
@@ -249,7 +281,8 @@ The contents of this file are *reStructuredText* formatted text that
 will be used as the content of the news file entry. You do not need to
 reference the issue or PR numbers here as *towncrier* will automatically
 add a reference to all of the affected issues when rendering the news
-file.
+file. Please also see ``CHANGES/README.rst`` for the correct format of
```
webknjaz: This can be a link: https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme.

webknjaz: I also like including such documents so they become a part of the Sphinx tree and their sections are linkable. Here's examples of what I did in other places:
* https://yarl.aio-libs.org/en/latest/contributing/guidelines/#adding-change-notes-with-your-prs
* https://ansible-pylibssh.rtfd.io/en/latest/contributing/guidelines/#adding-change-notes-with-your-prs
* https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

docs/contributing-cov-header.png
Dreamsorcerer: That could work, I'll try and find something.

docs/contributing.rst
Dreamsorcerer: Yeah, that's from the old code. Maybe just reduce it to the filepath...

docs/contributing.rst
Dreamsorcerer: Yep, I'm aware of this, but have only been using such comments on things which are definitely never going to be covered in a test (like `if TYPE_CHECKING`). In this example, I'm not 100% sure that the condition can never happen, it just didn't seem worth even trying to write a test for it.

docs/contributing.rst
webknjaz: I kinda like having a command. I just remembered there's one cool stdlib module that will address this better:
````suggestion
```console
python -m webbrowser -n file://"$(pwd)"/htmlcov/index.html
```
````
We should probably make this change in the makefile too.

https://docs.python.org/3/library/webbrowser.html

File: docs/contributing.rst (hunk 5 above)
Comment (webknjaz) [reply docs/contributing.rst:189]: @Dreamsorcerer things like `if TYPE_CHECKING` should be added to the `.coveragerc`'s `exclude_also = ` setting. We should probably integrate https://pypi.org/project/covdefaults/ too, like many projects did over the past year or so.

As for this specific case, when the loop has at least one iteration, we expect the internal line to be covered. Which means that if we enforce coverage on that line, we also enforce coverage for the `True` branch of the `while` line. So adding a pragma would only suppress that "skipped loop" branch, and we won't stop covering that second branch if we enforce the internal line coverage.

Another point I forgot to make is that any “no cover” mark must be accompanied by a comment with the motivation and explanation of when it can be removed.

File: docs/contributing-cov-header.png
Comment (Dreamsorcerer) [reply docs/contributing-cov-header.png:1]: Nope, I don't think it's going to work. I tried a browser extension, I think because it inlines all the CSS etc. needed, it produced an SVG ~120x larger than the PNG.

Comment (webknjaz) [reply docs/contributing-cov-header.png:1]: Do you have examples of such files? I could try to take a look into what's happening...

Comment (Dreamsorcerer) [reply docs/contributing-cov-header.png:1]: Ah, I just deleted it and uninstalled the extension. If you want to try it yourself, extension was https://addons.mozilla.org/en-US/firefox/addon/svg-screenshots/
Screenshots are from #7915.

File: docs/contributing.rst (hunk 7 above)
Comment (Dreamsorcerer) [reply docs/contributing.rst]: Shall I just replace this whole section with the included file then?

Comment (webknjaz) [reply docs/contributing.rst]: Maybe, not replace blindly, but extend + adjust. I haven't looked into the similarities in the content. But I'd like if we adopted the same practices as in those examples.

File: docs/contributing-cov-header.png
Comment (Dreamsorcerer) [reply docs/contributing-cov-header.png:1]: @webknjaz Are you still planning to have a go at producing SVGs, or can I merge this now?

File: CHANGES/README.rst [hunk 8]
```diff
@@ -34,21 +31,22 @@ for the users to understand what it means.
...
 PR followed by a dot, then add a patch type, like ``feature``,
-``doc``, ``misc`` etc., and add ``.rst`` as a suffix. If you
-need to add more than one fragment, you may add an optional
-sequence number (delimited with another period) between the type
-and the suffix.
+``doc``, ``misc`` etc. If you need to add more than one fragment,
+you may add an optional sequence number (delimited with another period)
+between the type and the suffix.
 
-In general the name will follow ``<pr_number>.<category>.rst`` pattern,
```
webknjaz: Plz keep `.rst`. While it's not mandatory, it makes it easier for people to understand what syntax is inside.

CHANGES/README.rst
```diff
@@ -34,21 +31,22 @@ for the users to understand what it means.
 combined with others, it will be a part of the "news digest"
 telling the readers **what changed** in a specific version of
 the library *since the previous version*. You should also use
-reStructuredText syntax for highlighting code (inline or block),
-linking parts of the docs or external sites.
-If you wish to sign your change, feel free to add ``-- by
+*reStructuredText* syntax for highlighting code (inline or block),
+linking parts of the docs or external sites. However, you do not need to
+reference the issue or PR numbers here as *towncrier* will automatically
```
webknjaz: This will need to be copied over into other projects as I'm trying to have this in sync across many repos, including the ones that are not exactly mine.

CHANGES/README.rst
```diff
@@ -19,7 +16,7 @@ to the end-users most of the time. And so such details should be
 recorded in the Git history rather than a changelog.
 
 Alright! So how to add a news fragment?
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
webknjaz, line 22: plz keep this style since it's copied over multiple repos and would essentially mean problems when syncing with others

CHANGES/README.rst
```diff
@@ -1,15 +1,12 @@
 .. _Adding change notes with your PRs:
 
-Adding change notes with your PRs
```
webknjaz, line 3: Keep this title and drop the one in `contributing.rst` instead. Follow what I did in ansible/pylibssh.

CHANGES/README.rst
```diff
@@ -34,21 +31,22 @@ for the users to understand what it means.
...
+reference the issue or PR numbers here as *towncrier* will automatically
+add a reference to all of the affected issues when rendering the news
+file. If you wish to sign your change, feel free to add ``-- by
 :user:`github-username``` at the end (replace ``github-username``
 with your own!).
 
 Finally, name your file following the convention that Towncrier
 understands: it should start with the number of an issue or a
 PR followed by a dot, then add a patch type, like ``feature``,
-``doc``, ``misc`` etc., and add ``.rst`` as a suffix. If you
```
webknjaz: let's keep this

CHANGES/README.rst
```diff
@@ -64,7 +62,7 @@ necessary to make a separate documentation fragment for documentation
 changes accompanying the relevant code changes.
 
 Examples for adding changelog entries to your Pull Requests
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
webknjaz, line 80: same concern

CHANGES/7916.doc
```diff
@@ -0,0 +1 @@
+Updated contributing/Tests coverage section to show how we use ``codecov``. -- by :user:`Dreamsorcerer`
```
webknjaz: I didn't update examples in this repo but essentially, it's nicer to have a trailing period when there's a single sentence and the byline is on separated from the paragraph.
```suggestion
Updated contributing/Tests coverage section to show how we use ``codecov`` -- by :user:`Dreamsorcerer`.
```
Additionally, you can use RST syntax to actually reference those sections so that they are clickable on the changelog page on RTD.

docs/contributing-cov-header.png
webknjaz: I suppose I won't be able to anytime soon. Don't wait for it.

CHANGES/README.rst
```diff
@@ -34,21 +31,22 @@ for the users to understand what it means.
...
 telling the readers **what changed** in a specific version of
 the library *since the previous version*. You should also use
-reStructuredText syntax for highlighting code (inline or block),
-linking parts of the docs or external sites.
-If you wish to sign your change, feel free to add ``-- by
+*reStructuredText* syntax for highlighting code (inline or block),
+linking parts of the docs or external sites. However, you do not need to
+reference the issue or PR numbers here as *towncrier* will automatically
+add a reference to all of the affected issues when rendering the news
+file. If you wish to sign your change, feel free to add ``-- by
```
webknjaz: Also, no need to reflow existing lines with sentences starting at the beginning. It makes diffs cleaner and RST knows how to merge them into a paragraph anyway.
```suggestion
linking parts of the docs or external sites.
However, you do not need to
reference the issue or PR numbers here as towncrier_ will automatically
add a reference to all of the affected issues when rendering the news
file.
If you wish to sign your change, feel free to add ``-- by
```

docs/contributing-cov-header.png
webknjaz: @Dreamsorcerer So I managed to make that comment image smaller. Grab this file and replace the PNG with it:
pr-codecov-comment-hits-misses-partials

I did a "screenshot" with the dark mode, so you may want to edit it and drop the first `<g>` node with the background rectangle.

webknjaz: I used https://chromewebstore.google.com/detail/svg-screenshot/nfakpcpmhhilkdpphcjgnokknpbpdllg with Chrome. Then, I opened the SVG file in a text editor *and* in a browser tab. In the browser tab, I opened DevTools so that I could highlight XML nodes visually. I experimented with removing small parts in NeoVim and refreshing the page. Reverting the last change when necessary.

It's highly nested, so one of the last cleanup steps was to just remove like 20 layers of `<g>` nodes and make it more flat. I also edited the background rectangle dimmentions.

It's 6.6KB (smaller than your 17.9KB PNG). I didn't apply compression but there's tools to cut out spaces. I don't think it's worth it. Also, the uncompressed version is easier to edit if the need be.

Do you want to give it a try with the other one?

webknjaz: Oh, the first step was to remove the carriage returns:
```console
:%s/\r//g
```

webknjaz: Here's another one:
pr-codecov-comment-title

webknjaz: I also think that the images should be in the `docs/_static/img/` dir and not in the docs root.

CHANGES/README.rst
Dreamsorcerer: We don't have a single change file in CHANGES/ with .rst currently, and the other instructions didn't have this...

CHANGES/README.rst
Dreamsorcerer: Uhh, well,  that'll require redoing some of the other files, as this is not the header used in our other files and thus causes a Sphinx error when included.

CHANGES/README.rst
Dreamsorcerer: This looks more like an intro paragraph, and then everything else are subheadings. It looks a bit weird to me that these headings are on the same level (including when I look at https://ansible-pylibssh.readthedocs.io/en/latest/contributing/guidelines/)

Dreamsorcerer: Second level headings on that page are:

- Contributing docs
- Adding change notes with your PRs
- Alright! So how do I add a news fragment?
- Examples for changelog entries adding to your Pull Requests

The first heading appears to have nothing to do with the others, but all of the last 3 are related to the same thing.

Dreamsorcerer: I've readded the header, but at a different level. Let me know if that's OK.

docs/contributing.rst
Dreamsorcerer: I'll leave this comment for now, as we don't really follow this practice currently. Can return to it if we decide to start enforcing it.

CHANGES/README.rst
Dreamsorcerer: Couldn't get that to build correctly, so have tweaked it again. I suggest we use the standard described at https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#sections
With the first heading being a section (=) and the others being subsections (-). Any pages which want to include the file will need to adopt that standard.

docs/contributing-cov-header.png
Dreamsorcerer: I'd like to make a release tomorrow, and am out of time today. So, if there's any chance you could create those last 2 SVGs to complete the PR by tomorrow morning, that'd be really helpful.

docs/contributing.rst
```diff
@@ -159,34 +159,68 @@ Any extra texts (print statements and so on) should be removed.
      make test-3.10-no-extensions
 
 Tests coverage
```
webknjaz: @Dreamsorcerer it occurred to me that the wording of "Tests coverage" is rather weird. Can you think of an improvement here? I'm leaning towards "Code coverage" or "Covering code with tests". WDYT?

docs/contributing.rst
```diff
@@ -159,34 +159,68 @@ Any extra texts (print statements and so on) should be removed.
...
-The project uses *codecov.io* for storing coverage results. Visit
-https://codecov.io/gh/aio-libs/aiohttp for looking on coverage of
-master branch, history, pull requests etc.
+After reviewing the PR, we find that this code is no longer needed as
+the changes mean that this method will never be called under those
+conditions. Thanks to this report, we were able to remove some
+redundant code from a performance-critical part of our codebase (this
+check would have been run, probably multiple times, for every single
+incoming request).
+
```
webknjaz, line 204: Can we add a `.. tip::` block somewhere here mentioning that Codecov is sometimes drunk and the metrics reported may be incomplete at times. It might be a good transition paragraph to hint the contributor to keep reading and point them towards the importance of inspecting the coverage locally as well.

CHANGES/README.rst
webknjaz: Dunno. From the RST standpoint, it shouldn't matter. But it's nice to have some consistency. OTOH, I'd rather implement the header level changes in a separate PR, then here. I'd be fine with having the same-level headers for a while. Though, if we go for the Sphinx-suggested semantics, I'd extend that with documenting such a decision explicitly and maybe tried adding some linting for consistency. But then, it becomes a bigger standalone effort.

IIRC I picked up parts of that structure from https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request (but I also contributed back so I don't remember exactly where each specific sub-idea was coming from). It's one of the projects I wanted to keep in sync with the doc structure. There's also frozenlist/yarl/multidict, of course. And some others, in other orgs.

Currently, I'd like the document to diverge minimally, since it allows to copy parts over with less merge conflicts and cleaner diffs.

CHANGES/README.rst
webknjaz: True. But the common template I'm syncing over does. There's also `.rst` files in yarl/frozenlist/multidict. We just need to do a better job of consolidating those bits.

docs/contributing-cov-header.png
webknjaz: I expect those other two to be more tricky. Not sure how fast it'd be possible to complete them.

webknjaz: Here's one more:
contributing-cov-partial

webknjaz: And the most tricky one:
contributing-cov-miss

docs/contributing.rst
```diff
@@ -159,34 +159,68 @@ Any extra texts (print statements and so on) should be removed.
...
+Changed* tab on the *GitHub Pull Request* review page.
+
+
+You can also produce coverage reports locally with ``make cov-dev``
+or just adding ``--cov-report=html`` to ``pytest``.
+
+This will run the test suite and collect coverage information. Once
+finished, coverage results can be view by opening:
+```console
+python -m webbrowser -n file://"$(pwd)"/htmlcov/index.html
```
webknjaz: The `console` lexer expects a prompt line that it can highlight:
```suggestion
$ python -m webbrowser -n file://"$(pwd)"/htmlcov/index.html
```

CHANGES/7916.doc
```diff
@@ -0,0 +1 @@
+Updated contributing/Tests coverage section to show how we use ``codecov`` -- by :user:`Dreamsorcerer`.
```
webknjaz: Mind using `:ref:` to link those?

webknjaz:
@Dreamsorcerer I'm synching the changelog structure improvements I made in the subprojects so that it could be included in the new release. Unfortunately, merging https://github.com/aio-libs/aiohttp/pull/8066 will cause conflicts that will have to be resolved.

CHANGES/README.rst
Dreamsorcerer: > I picked up parts of that structure from https://setuptools.pypa.io/en/latest/development/developer-guide.html#making-a-pull-request

Right, but that one has a parent header, which has been removed here before inserting into a page with other content. We could readd that header with the 2 sentence intro (although seems a little redundant to me).

webknjaz: I kinda wanted the header in the docs folder. But maybe that's fine too.

docs/contributing.rst
```diff
@@ -158,35 +158,75 @@ Any extra texts (print statements and so on) should be removed.
...
-master branch, history, pull requests etc.
+After reviewing the PR, we find that this code is no longer needed as
+the changes mean that this method will never be called under those
+conditions. Thanks to this report, we were able to remove some
+redundant code from a performance-critical part of our codebase (this
+check would have been run, probably multiple times, for every single
+incoming request).
+
+.. tip::
+   Sometimes the diff on Codecov doesn't make sense. This is usually
```
Dreamsorcerer:
```suggestion
   Sometimes the diff on *codecov.io* doesn't make sense. This is usually
```

CHANGES/README.rst
Dreamsorcerer: In the interests of time, I'm going to merge this, but feel free to follow up with any other changes. It shouldn't be important if they miss this release.

webknjaz: Fair

docs/contributing.rst
```diff
@@ -158,35 +158,75 @@ Any extra texts (print statements and so on) should be removed.
...
+After reviewing the PR, we find that this code is no longer needed as
+the changes mean that this method will never be called under those
+conditions. Thanks to this report, we were able to remove some
+redundant code from a performance-critical part of our codebase (this
+check would have been run, probably multiple times, for every single
+incoming request).
+
+.. tip::
+   Sometimes the diff on *codecov.io* doesn't make sense. This is usually
+   caused by the branch being out of sync with master. Try merging
```
webknjaz, line 207: @Dreamsorcerer this isn't exactly the case I meant. What I'm talking about is GHA attempting to upload coverage and getting HTTP 500 from Codecov API in some or all of the jobs. Sometimes, restarting those jobs doesn't help and it just keeps erroring out for hours.
This means that part of the coverage uploads happened from some envs, but not all. And depending on which ones failed, the coverage may show up lower than it actually is.

To make things worse, by default, the GHA step doing the upload succeeds silently. And the only way to discover if some jobs didn't actually upload coverage is to click through each CI job, expand that step and read the log (it'll either have a URL of the upload for the current commit/PR or some error message at the end of the console output).

Dreamsorcerer: > What I'm talking about is GHA attempting to upload coverage and getting HTTP 500 from Codecov API in some or all of the jobs.

Oh, I think I enabled the option to fail CI if an error occurs in upload (unless it was a different repo), so we shouldn't get that happening anymore.

webknjaz: Different repo, probably. But still, I'm not sure if that option is what we want — it effectively marks entire jobs as failing. And whenever Codecov has platform issues, that means fully blocking merges for hours if not days.

Dreamsorcerer: Problem is that we've had several repos which have been failing coverage uploads for a long time, due to configuration issues. I've personally not seen codecov fail on upload as an issue on their end yet.

CHANGES/README.rst
```diff
@@ -1,15 +1,23 @@
-.. _Adding change notes with your PRs:
```
webknjaz, line 1: Changing this label is going to be problematic as other projects rely on it for referring the changelog guidelines from other places. Not all the PR recommendations set. It'll have to be reverted.

CHANGES/README.rst
```diff
@@ -1,15 +1,23 @@
...
+.. _Making a pull request:
+
+Making a pull request
```
webknjaz, line 3: @Dreamsorcerer this title was a mistake:
1) Since it's rendered on GitHub in a folder with change notes, on any branch, it shouldn't really mention the PR process that much — just the process for having the changelog.
2) The Sphinx site now has 2 subsequent titles with the same name but different capitalization — https://docs.aiohttp.org/en/stable/contributing.html#making-a-pull-request and https://docs.aiohttp.org/en/stable/contributing.html#id2.

CHANGES/README.rst
```diff
@@ -1,15 +1,23 @@
...
+=====================
+
+When making a pull request, please include a short summary of the changes
+and a reference to any issue tickets that the PR is intended to solve.
+All PRs with code changes should include tests. All changes should
```
webknjaz, line 8: Yeah, I missed that this now refers to tests that are not related to the https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme directory at all.

docs/contributing.rst
webknjaz: I've definitely seen some. Both in aio-libs and other places. It's sometimes flaky on their platform side. And restarts result in HTTP 500, then in several hours it *just works* when restarted. And I observed these outages affecting repos in different orgs in the same time frame, while their status pages happily reported "no platform issues".

And that is the primary reason https://hynek.me/articles/ditch-codecov-python/ is a thing.
I actually think, parts of that article should be implemented here too. Maybe, not in place of Codecov but additionally to it.

P.S. In the past I also blamed misconfiguration a lot of times. But that doesn't account for when its not the case.

docs/contributing.rst
```diff
@@ -220,47 +260,19 @@ To run spell checker on Linux box you should install it first:
    $ sudo apt-get install enchant
    $ pip install sphinxcontrib-spelling
 
-Changelog update
```
webknjaz, line 223: @Dreamsorcerer I think you had Sphinx problems because you didn't keep this title as in examples. I applied a partial revert and there's no errors.

PR #7918, opened 2023-11-27
Create new response on each HTTPException

bdraco:
## What do these changes do?

This is a limited backport of #5197 and #3462

## Checklist

### DIFF
--- a/aiohttp/web_exceptions.py
+++ b/aiohttp/web_exceptions.py
@@ -1,6 +1,7 @@
 import warnings
 from typing import Any, Dict, Iterable, List, Optional, Set  # noqa
 
+from multidict import CIMultiDict
 from yarl import URL
 
 from .typedefs import LooseHeaders, StrOrURL
@@ -116,6 +117,14 @@ class HTTPException(Response, Exception):
 
     def __bool__(self) -> bool:
         return True
+
+    @property
+    def status(self) -> int:
+        return self.status_code
+
+    @property
+    def headers(self) -> "CIMultiDict[str]":
+        return self._headers
 
 
 class HTTPError(HTTPException):
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -453,7 +453,10 @@ class RequestHandler(BaseProtocol):
             finally:
                 self._current_request = None
         except HTTPException as exc:
-            resp = exc
+            resp = Response(
+                status=exc.status, reason=exc.reason, text=exc.text, headers=exc.headers
+            )
+            resp._cookies = exc._cookies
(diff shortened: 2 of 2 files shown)


aiohttp/web_protocol.py
```diff
@@ -453,7 +453,10 @@ async def _handle_request(
             finally:
                 self._current_request = None
         except HTTPException as exc:
-            resp = exc
+            resp = Response(
+                status=exc.status, reason=exc.reason, text=exc.text, headers=exc.headers
+            )
+            resp._cookies = exc._cookies
```
Dreamsorcerer, line 459: I have a strong feeling this code is rather fiddly and probably shouldn't be touched..

bdraco: I pulled it directly from https://github.com/aio-libs/aiohttp/pull/5197/files#diff-2126b277e07e3fdd05e7a81da456accf24e5515a46c78c48a79db4eb166043e4R463

I'm also more than happy to close this and give up the optimization for 404s until 4.x

aiohttp/web_exceptions.py
```diff
@@ -117,6 +118,14 @@ def __init__(
     def __bool__(self) -> bool:
         return True
 
+    @property
+    def status(self) -> int:
+        return self.status_code
+
+    @property
+    def headers(self) -> "CIMultiDict[str]":
+        return self._headers
```
Dreamsorcerer, line 127: Also, this should already be present from Response. It was added in the linked PR due to it being separated from Response.

aiohttp/web_protocol.py
Dreamsorcerer: Yeah, that PR is making the exceptions compatible with cookies. As they are still Responses in 3.x we don't need those changes. I also recall needing to fix a problem in this code, and remember it being complex due to the Response/HTTPException thing, so I'd suggest just leaving this one.

bdraco: > so I'd suggest just leaving this one.



Let's close this than. We can always come back to it if the performance of the 404 turns out to be a problem but I don't think it will

Dreamsorcerer: To expand on why we need to keep this for backwards-compatibility, we have, for example, this error middleware example:
https://demos.aiohttp.org/en/latest/tutorial.html#middlewares

You'll see this catches HTTPException in order to customise error responses. Older examples would have checked the type of the returned response, as these weren't raised as exceptions previously.

So, in v4 they are exceptions only, but in 3.x they must still be a hybrid Response, otherwise those old middlewares etc. that users wrote will break badly.

PR #7930, opened 2023-12-02
Use .coveragerc for TYPE_CHECKING conditions

Dreamsorcerer:


.coveragerc
```diff
@@ -2,3 +2,7 @@
 branch = True
 source = aiohttp, tests
 omit = site-packages
+
+[report]
+exclude_also =
+    if TYPE_CHECKING
```
webknjaz, line 8: @Dreamsorcerer have you tried https://pypi.org/p/covdefaults/ ? It provides better regexes in general, including this one.

PR #7933, opened 2023-12-02
Support FileResponse from any pathlib.Path compatible object

### LINKED ISSUE #7928: pathlib.Path-alike no longer works in 3.9.0 for web.FileResponse
### Describe the bug

I use `web.FileResponse` to respond with a file that is not really on the disk. This is done with a `Path` instance that implements the `pathlib.Path` interface. Much like the `zipfile.Path` in the standard library. (`zipfile.Path` has no `stat` yet, so it is a no-go anyway.)

With upgrade to 3.9.0, this no longer works, but it fails with file not found exception.

I see the change came in #7111, the backport of #6876. For no apparent reason, the condition to only replace str with `pathlib.Path` was changed to create `pathlib.Path` for any argument.

```
@@ -55,10 +55,7 @@ class FileResponse(StreamResponse):
     ) -> None:
         super().__init__(status=status, reason=reason, headers=headers)

-        if isinstance(path, str):
-            path = pathlib.Path(path)
-
-        self._path = path
+        self._path = pathlib.Path(path)
         self._chunk_size = chunk_size
```

Python 3.12 is preparing for `pathlib.Path` to be subclassed more easily (see [whatsnew](https://docs.python.org/3/whatsnew/3.12.html#pathlib)). Whatever reasons there are to subclass, would also be broken by this kind of change.

Since this breaks something that used to work, I report this as a bug. As a new feature it would not require much. Only `Path` members `name`, `open`, `stat`, `with_name` and `is_file` for the value returned by `with_name`.

### To Reproduce

1. Subclass or implement own `pathlib.Path` like subclass.
2. Create instance.
3. Pass it to `web.FileResponse`.

### Expected behavior

Response created with the stat and file content of the `pathlib.Path`.

### Logs/tracebacks

```python-traceback
  File "***/.venv310/lib/python3.10/site-packages/aiohttp/web_protocol.py", line 531, in start
    resp, reset = await task
  File "***/.venv310/lib/python3.10/site-packages/aiohttp/web_protocol.py", line 477, in _handle_request
    reset = await self.finish_response(request, resp, start_time)
  File "***/.venv310/lib/python3.10/site-packages/aiohttp/web_protocol.py", line 630, in finish_response
    await prepare_meth(request)
  File "***/.venv310/lib/python3.10/site-packages/aiohttp/web_fileresponse.py", line 139, in prepare
    st: os.stat_result = await loop.run_in_executor(None, filepath.stat)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.10/pathlib.py", line 1097, in stat
    return self._accessor.stat(self, follow_symlinks=follow_symlinks)
FileNotFoundError: [Errno 2] No such file or directory: '***'
```

### Python Version

```console
$ python --version
Python 3.10.13
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.9.1
Author:
Author-email:
License: Apache 2
Location: /home/kohtala/andritz/.venv310/lib/python3.10/site-packages
Requires: aiosignal, async-timeout, attrs, frozenlist, multidict, yarl
Required-by:
```

### multidict Version

```console
Name: multidict
Version: 6.0.4
Author: Andrew Svetlov
License: Apache 2
Location: /home/kohtala/andritz/.venv310/lib/python3.10/site-packages
Requires:
Required-by: aiohttp, yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.9.3
Author: Andrew Svetlov
License: Apache-2.0
Location: /home/kohtala/andritz/.venv310/lib/python3.10/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

```
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

### Related component

Server

### Additional context

_No response_

### Code of Conduct

kohtala:
## What do these changes do?

Add support for passing to web.FileResponse any object with required attributes of pathlib.Path.

Responses can be sent from files not in the filesystem.

## Related issue number

Fixes #7928

## Checklist

### DIFF
--- a/aiohttp/typedefs.py
+++ b/aiohttp/typedefs.py
@@ -1,14 +1,17 @@
 import json
 import os
 from typing import (
+    IO,
     TYPE_CHECKING,
     Any,
     Awaitable,
     Callable,
     Iterable,
     Mapping,
+    Protocol,
     Tuple,
     Union,
+    runtime_checkable,
 )
 
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy, istr
@@ -52,3 +55,24 @@ Handler = Callable[["Request"], Awaitable["StreamResponse"]]
 Middleware = Callable[["Request", Handler], Awaitable["StreamResponse"]]
 
 PathLike = Union[str, "os.PathLike[str]"]
+
+
+class PathlibPathNamedLike(Protocol):
+    def is_file(self) -> bool:
+        ...
+
+
+@runtime_checkable
+class PathlibPathLike(Protocol):
+    """pathlib.Path interface used by aiohttp."""
+
+    name: str
+
+    def open(self, mode: str) -> IO[Any]:
+        ...
+
+    def stat(self, *, follow_symlinks=True) -> os.stat_result:
+        ...
+
+    def with_name(self, name: str) -> PathlibPathNamedLike:
+        ...
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -11,13 +11,14 @@ from typing import (
     Final,
     Optional,
     Tuple,
+    Union,
     cast,
 )
 
 from . import hdrs
 from .abc import AbstractStreamWriter
 from .helpers import ETAG_ANY, ETag, must_be_empty_body
-from .typedefs import LooseHeaders, PathLike
+from .typedefs import LooseHeaders, PathlibPathLike, PathLike
 from .web_exceptions import (
     HTTPNotModified,
     HTTPPartialContent,
@@ -43,7 +44,7 @@ class FileResponse(StreamResponse):
 
     def __init__(
         self,
-        path: PathLike,
+        path: Union[PathLike, PathlibPathLike],
         chunk_size: int = 256 * 1024,
         status: int = 200,
         reason: Optional[str] = None,
@@ -51,7 +52,12 @@ class FileResponse(StreamResponse):
     ) -> None:
         super().__init__(status=status, reason=reason, headers=headers)
 
-        self._path = pathlib.Path(path)
+        if isinstance(path, str) or (
+            not isinstance(path, PathlibPathLike) and isinstance(path, os.PathLike)
+        ):
+            path = pathlib.Path(path)
+
+        self._path = path
         self._chunk_size = chunk_size
 
     async def _sendfile_fallback(
--- a/tests/test_web_sendfile.py
+++ b/tests/test_web_sendfile.py
@@ -1,3 +1,5 @@
+from io import BytesIO
+from os import stat_result
 from pathlib import Path
 from typing import Any
 from unittest import mock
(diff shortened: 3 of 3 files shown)


Dreamsorcerer:
Why do we need the protocol? If you've subclassed Path, then shouldn't it be a subclass of `os.PathLike`?

kohtala:
> Why do we need the protocol? If you've subclassed Path, then shouldn't it be a subclass of `os.PathLike`?

If you do not have the file on filesystem, it would be best if it is not a subclass of `pathlib.Path`. `pathlib.Path` is a `os.PathLike`, i.e. it gives a path for the filesystem. The testcase in this PR adds a minimal class that provides the information needed for response without going to filesystem at all. For typing, you need a `Protocol` to check if the passed object implements the required interface.

Dreamsorcerer:
> If you do not have the file on filesystem, it would be best if it is not a subclass of `pathlib.Path`. `pathlib.Path` is a `os.PathLike`, i.e. it gives a path for the filesystem.

But, `os.PathLike` is an ABC, so you can just subclass that instead.

Though looking closer, `os.PathLike` only defines `__fspath__()`, so I'm not entirely sure why the type checking succeeded in the past...

Another complication with your custom classes, is that someone expects things like `open()` to work when passed a PathLike. It's probably a bit confusing if `open()` fails, but `path.open()` succeeds.

Dreamsorcerer:
> But, `os.PathLike` is an ABC, so you can just subclass that instead.

Hmm, I guess your point is you want something that looks like a `Path`, but doesn't support `__fspath__()`, and isn't permitted in `open()`?

kohtala:
> > But, `os.PathLike` is an ABC, so you can just subclass that instead.
> 
> Hmm, I guess your point is you want something that looks like a `Path`, but doesn't support `__fspath__()`, and isn't permitted in `open()`?

Yes. A `Path` that doesn't have working `__fspath__()` for `open()` since there is no file on the filesystem.

My application generates the file and has the content and timestamp in-memory. It does not change for every request, so much of the functionality in `web.FileResponse` is useful and I would not like to make a copy of it. I'd also rather not have some temporary file just to get around some silly minor detail in the existing implementation.

For now I assign the own Path like to `FileResponse._path` directly, but I'd rather have proper support for it.

Dreamsorcerer:
> My application generates the file and has the content and timestamp in-memory.

I guess my question is why use FileResponse? If you already have the file contents in memory, why not use a regular Response?

kohtala:
> > My application generates the file and has the content and timestamp in-memory.
> 
> I guess my question is why use FileResponse? If you already have the file contents in memory, why not use a regular Response?

`FileResponse` implements [RFC 7232](https://datatracker.ietf.org/doc/html/rfc7232) and [RFC 7233](https://datatracker.ietf.org/doc/html/rfc7233).

Dreamsorcerer:
OK, I'm getting the feeling it would make more sense to expand FileResponse, or add another subclass, which supports this more directly. e.g. Instead of a path object, it could accept a file-like object directly, then you can just pass io.StringIO.

Dreamsorcerer:
> That thought occurred to me also. It needs the modification time, size, maybe something for the content-type so any `typing.IO` is not enough.

Right, we could add attributes for these on the response object itself so they could be managed there. I was also thinking that modification time could be set automatically when assigning an attribute, but that'd only work if it was str/bytes rather than a file-like object (len() could also work for size in that instance).

PR #7942, opened 2023-12-04
added server_hostname param to ws_connect method

### LINKED ISSUE #7941: using server_host name param for ws_connect

if i understand correctly, SNI is currently supported using the server_hostname parameter.
the issue is i see it only exists in _request, and the call for ws_connect doesnt support this param.

simply add the server_hostname  parameter to ws_connect so it will be passed to _request method.

no alternatives, the solution is simple.

### Related component

Client

### Additional context

_No response_

### Code of Conduct

tamir1400:
added the server_hostname param to ws_connect method, so it will simply be passed to the _request method which already supports the server_hostname param.

Fixes #7941

### DIFF
--- a/CHANGES/7941.feature
+++ b/CHANGES/7941.feature
@@ -1,1 +1,1 @@
-added server_hostname param to ws_connect
+Added ``server_hostname`` parameter to ``ws_connect``.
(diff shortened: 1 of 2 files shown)


Dreamsorcerer:
I think that looks fine to me.

PR #7944, opened 2023-12-04
Implement filter_cookies() with domain-matching and path-matching

xiangxli:
## What do these changes do?

Implement filter_cookies() with domain-matching and path-matching on the keys, instead of testing every single cookie.

Fork from the below PR, force push caused it to be closed.
https://github.com/aio-libs/aiohttp/pull/7777

no

## Related issue number

https://github.com/aio-libs/aiohttp/issues/7583

## Checklist


xiangxli:
@Dreamsorcerer Hey buddy. The PR 7777 was closed by force push, creating this new PR to continue.

tests/test_cookiejar.py
```diff
@@ -253,6 +253,121 @@ async def test_filter_cookies_str_deprecated(loop: Any) -> None:
...
+        " Expires=Tue, 1 Jan 1980 12:00:00 GMT; "
+        "max-age-cookie=fourteenth; Domain=maxagetest.com; Path=/;"
+        " Max-Age=60; "
+        "invalid-max-age-cookie=fifteenth; Domain=invalid-values.com; "
+        " Max-Age=string; "
+        "invalid-expires-cookie=sixteenth; Domain=invalid-values.com; "
+        " Expires=string;"
+    )
+    jar.update_cookies(cookies)
+    cookies = jar.filter_cookies(URL("http://pathtest.com/one/two/"))
```
Dreamsorcerer: From previous PR, we probably want a couple more calls here to test different parts. This should include one with a subdomain (and probably one without the path). To verify we get the correct cookies for each kind of request.

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+            # handle last element for split
+            leftovers = d.split(".", maxsplit=1)
+            if len(leftovers) > 1:
+                d = leftovers[-1]
+            else:
+                d = None
+
+        # shared cookie, it should have max of 1 entry
+        pairs.add(("", "/"))
```
Dreamsorcerer: Question from previous PR:
Maybe I'm just forgetting. But, is there anything in the code, or something in the spec which mentions this case?

```suggestion
        # shared cookie, it should have max of 1 entry
        pairs.add(("", "/"))
```

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
                 request_origin = request_url.origin()
             is_not_secure = request_origin not in self._treat_as_secure_origin
 
+        # d: domain, d could be subdomain
+        # p: path
+        d = hostname
+
+        pairs = set()
```
Dreamsorcerer: Doesn't pairs need to be a list? i.e. It needs to handle longest paths first, so ordering is important.

Probably need another test case if this didn't get highlighted in the tests.

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+        while d:
+            p = request_url.path
+            while p:
+                pairs.add((d, p))
+                path_leftovers = p.rsplit("/", maxsplit=1)
+                # handle last element for rsplit
+                if len(path_leftovers) > 1:
+                    p = path_leftovers[0]
+                else:
+                    p = None
```
Dreamsorcerer: Not sure I understand why this is needed.

Isn't a path always starting with `/`, so we'll get to `/foo` which will be split to `["", "foo"]`. `""` will evaluate as False, so the while loop will break (before it ever reaches this else condition).

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+            # handle last element for split
+            leftovers = d.split(".", maxsplit=1)
+            if len(leftovers) > 1:
+                d = leftovers[-1]
+            else:
+                d = None
```
Dreamsorcerer: For this one, maybe would be more efficient as:
```suggestion
            try:
                d = d.split(".", maxsplit=1)[1]
            except IndexError:
                d = ""
```

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+        # shared cookie, it should have max of 1 entry
+        pairs.add(("", "/"))
+        print(f"pairs: {pairs}")
```
```suggestion
```

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+        import itertools
```
Dreamsorcerer: Move to top of file.

aiohttp/cookiejar.py
xiangxli: You are correct on path.

aiohttp/cookiejar.py
```diff
@@ -261,8 +261,43 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+        cookies = itertools.chain.from_iterable(
+            self._cookies[p].values() for p in pairs
+        )
+
+        # try push
+
         # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
-        for cookie in sorted(self, key=lambda c: len(c["path"])):
+        for cookie in list(cookies):
```
Dreamsorcerer: We made it an iterator for efficiency above, converting to a list defeats the purpose.
```suggestion
        for cookie in cookies:
```

tests/test_cookiejar.py
Dreamsorcerer: Oh, I didn't see the 3 tests above. We should rewrite it with pytest.parametrize to avoid duplication.

Dreamsorcerer: So, url and expected_cookies would be parameters.

aiohttp/cookiejar.py
xiangxli: Domain attribute without a value is undefined behavior, I added it based on the test cases data.
https://www.rfc-editor.org/rfc/rfc6265#section-5.2.3

Dreamsorcerer: Hmm, it does say:
> However, the user agent SHOULD ignore the cookie-av entirely.

But, I guess we'll just keep it compatible with the existing behaviour for now.

tests/test_cookiejar.py
```diff
@@ -253,6 +253,106 @@ async def test_filter_cookies_str_deprecated(loop: Any) -> None:
...
+        " Expires=Tue, 1 Jan 1980 12:00:00 GMT; "
+        "max-age-cookie=fourteenth; Domain=maxagetest.com; Path=/;"
+        " Max-Age=60; "
+        "invalid-max-age-cookie=fifteenth; Domain=invalid-values.com; "
+        " Max-Age=string; "
+        "invalid-expires-cookie=sixteenth; Domain=invalid-values.com; "
+        " Expires=string;"
+    )
+    jar.update_cookies(cookies)
+    cookies = jar.filter_cookies(URL("http://pathtest.com/"))
```
Dreamsorcerer: This one should also be in the parametrized set, right?

tests/test_cookiejar.py
```diff
@@ -253,6 +253,106 @@ async def test_filter_cookies_str_deprecated(loop: Any) -> None:
...
+@pytest.mark.parametrize(
+    ("url", "expected_cookies"),
+    [
+        (
+            "http://pathtest.com/one/two/",
+            [
+                "path2-cookie",
+                "shared-cookie",
+                "path3-cookie",
+                "path4-cookie",
```
Dreamsorcerer, line 245: Shouldn't this also have no-path-cookie and path1-cookie? Which is probably why there are 5 other tests failing.

aiohttp/cookiejar.py
```diff
@@ -261,8 +262,34 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+                d = d.split(".", maxsplit=1)[1]
+            except IndexError:
+                # handle last element for split
+                d = ""
+
+        # shared cookie, it should have max of 1 entry
+        pairs.append(("", "/"))
+
+        cookies = itertools.chain.from_iterable(
+            self._cookies[p].values() for p in pairs
```
Dreamsorcerer, line 288: Looking at the tests, this needs to be reversed, so longer paths are done later and therefore overwrite shorter paths.
```suggestion
            self._cookies[p].values() for p in reversed(pairs)
```

tests/test_cookiejar.py
xiangxli: Yes, it should be.

tests/test_cookiejar.py
xiangxli: I can't seem to comprehend the logic behind the no-path-cookie and path1-cookie, need more clarification on this.

Dreamsorcerer: They path-match don't they? Why would path1-cookie work and path2-cookie not?

aiohttp/cookiejar.py
Dreamsorcerer, line 1: So, I think the remaining fix needed for the failing tests is probably:
```diff
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -210,6 +210,7 @@ class CookieJar(AbstractCookieJar):
                     # Cut everything from the last slash to the end
                     path = "/" + path[1 : path.rfind("/")]
                 cookie["path"] = path
+            path = path.rstrip("/")
 
             max_age = cookie["max-age"]
             if max_age:
@@ -272,6 +273,7 @@ class CookieJar(AbstractCookieJar):
             while p:
                 pairs.append((d, p))
                 p = p.rsplit("/", maxsplit=1)[0]
+            pairs.append((d, ""))
 
             try:
                 d = d.split(".", maxsplit=1)[1]
@@ -280,7 +282,7 @@ class CookieJar(AbstractCookieJar):
                 d = ""
 
         # shared cookie, it should have max of 1 entry
-        pairs.append(("", "/"))
+        pairs.append(("", ""))
 
         cookies = itertools.chain.from_iterable(
             self._cookies[p].values() for p in reversed(pairs)
```

Don't worry about the pickle test at the moment.

xiangxli:
```diff
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -272,6 +272,9 @@ class CookieJar(AbstractCookieJar):
             while p:
                 pairs.append((d, p))
                 p = p.rsplit("/", maxsplit=1)[0]
+            # special handling for no-path and path1
+            pairs.append((d, "/"))
+            pairs.append((d, p))

```
This is what I was thinking, but trying yours at the moment.

Dreamsorcerer: My thought is that the rstrip() will deduplicate keys, which will make it more efficient and work correctly when a cookie had path '/foo/bar' and another '/foo/bar/' (where currently, the latter would be missed).

tests/test_cookiejar.py
```diff
@@ -253,6 +253,78 @@ async def test_filter_cookies_str_deprecated(loop: Any) -> None:
...
+                "no-path-cookie",
+                "path1-cookie",
+            ],
+        ),
+    ],
+    ids=(
+        "/one/two/ path",
+        "test1.example.com subdomain",
+        "pathtest.com",
+    ),
```
Dreamsorcerer: This adds a couple more test cases, which includes the error I mentioned in the other thread (with /one/two vs /one/two/).

```suggestion
    (
        (
            "http://pathtest.com/one/two/",
            (
                "no-path-cookie",
                "path1-cookie",
                "path2-cookie",
                "shared-cookie",
                "path3-cookie",
                "path4-cookie",
            ),
        ),
        (
            "http://pathtest.com/one/two",
            (
                "no-path-cookie",
                "path1-cookie",
                "path2-cookie",
                "shared-cookie",
                "path3-cookie",
            ),
        ),
        (
            "http://pathtest.com/one/two/three/",
            (
                "no-path-cookie",
                "path1-cookie",
                "path2-cookie",
                "shared-cookie",
                "path3-cookie",
                "path4-cookie",
            ),
        ),
        (
            "http://test1.example.com/",
            (
                "shared-cookie",
                "domain-cookie",
                "subdomain1-cookie",
                "dotted-domain-cookie",
            ),
        ),
        (
            "http://pathtest.com/",
            (
                "shared-cookie",
                "no-path-cookie",
                "path1-cookie",
            ),
        ),
    ),
```

Dreamsorcerer:
@bdraco As you have a PR depending on this, would you like to make a final review (and double check that my logic of stripping the last slash is correct)?

I'll sort out the pickle test once approved.

bdraco:
Will take a look shortly

aiohttp/cookiejar.py
```diff
@@ -261,8 +263,30 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
                 request_origin = request_url.origin()
             is_not_secure = request_origin not in self._treat_as_secure_origin
 
+        pairs = []
+        d = hostname
+        while d:
+            p = request_url.path
+            while p:
+                pairs.append((d, p))
+                p = p.rsplit("/", maxsplit=1)[0]
```
bdraco: `rpartition` is generally faster than `rsplit` assuming it works here

aiohttp/cookiejar.py
```diff
@@ -261,8 +263,30 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+            pairs.append((d, ""))
+
+            try:
+                d = d.split(".", maxsplit=1)[1]
```
bdraco: `partition` is generally faster than `split` assuming it works here

aiohttp/cookiejar.py
```diff
@@ -261,8 +263,30 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
                 request_origin = request_url.origin()
             is_not_secure = request_origin not in self._treat_as_secure_origin
 
+        pairs = []
+        d = hostname
+        while d:
+            p = request_url.path
```
bdraco: I wonder if `.parts` can be used instead as its cached

bdraco: Looks like thats slower

aiohttp/cookiejar.py
```diff
@@ -261,8 +263,30 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+            while p:
+                pairs.append((d, p))
+                p = p.rsplit("/", maxsplit=1)[0]
+            pairs.append((d, ""))
+
+            try:
+                d = d.split(".", maxsplit=1)[1]
+            except IndexError:
+                # handle last element for split
+                d = ""
```
```suggestion
        pairs: List[Tuple[str, str]] = []
        paths: List[str] = []
        path = request_url.path
        while path:
            paths.append(path)
            path = path.rpartition("/")[0]

        domain_part = hostname
        while domain_part:
            for path in paths:
                pairs.append((domain_part, path))
            pairs.append((domain_part, ""))
            domain_part = domain_part.partition(".")[2]
```

This was the fastest suggestion

bdraco:
```python
        pairs: List[Tuple[str, str]] = []
        paths: List[str] = []
        path = request_url.path
        while path:
            paths.append(path)
            path = path.rpartition("/")[0]

        domain_part = hostname
        while domain_part:
            for path in paths:
                pairs.append((domain_part, path))
            pairs.append((domain_part, ""))
            domain_part = domain_part.partition(".")[2]
```

This might be even better if there are lots of domain labels....

Dreamsorcerer: Haven't benchmarked it, but trying to keep it all within iterators so we don't do multiple splits etc., how about this?
```
paths = itertools.accumulate(request_url.path.split("/"), lambda x,y: f"{x}/{y}")
domains = itertools.accumulate(reversed(hostname.split(".")), lambda x,y: f"{y}.{x}")
pairs = itertools.chain((("", ""),), itertools.product(domains, paths))
```

I think this is the reverse order of what we already have, so the reversed() call below would need to be dropped (and I've merged the append into here too).

bdraco: Will put together a benchmark for all the options

bdraco: time_original: 0.04315150000911672
time_suggestion_with_partition: 0.026855791991692968
time_suggestion_with_partition_minimize_loops: 0.016633624996757135
time_suggestion_with_itertools: 0.020396792009705678

bdraco:
```python
import timeit

from yarl import URL
import itertools
from typing import Callable


run_count = 20000
hostname = "www.example.com"
request_url = URL(f"http://{hostname}/this/is/a/typical/url")
pairs = []


def original():
    d = hostname
    while d:
        p = request_url.path
        while p:
            pairs.append((d, p))
            p = p.rsplit("/", maxsplit=1)[0]
        pairs.append((d, ""))
        try:
            d = d.split(".", maxsplit=1)[1]
        except IndexError:
            # handle last element for split
            d = ""


def suggestion_with_partition():
    domain_part = hostname
    while domain_part:
        paths = request_url.path
        while paths:
            pairs.append((domain_part, paths))
            paths = paths.rpartition("/")[0]
        pairs.append((domain_part, ""))
        domain_part = domain_part.partition(".")[2]


def suggestion_with_partition_minimize_loops():
    paths = []
    path = request_url.path
    while path:
        paths.append(path)
        path = path.rpartition("/")[0]

    domain_part = hostname
    while domain_part:
        for path in paths:
            pairs.append((domain_part, path))
        pairs.append((domain_part, ""))
        domain_part = domain_part.partition(".")[2]


def suggestion_with_itertools():
    paths = itertools.accumulate(request_url.path.split("/"), lambda x, y: f"{x}/{y}")
    domains = itertools.accumulate(
        reversed(hostname.split(".")), lambda x, y: f"{y}.{x}"
    )
    pairs = itertools.chain((("", ""),), itertools.product(domains, paths))


def run_suggestion(suggestion: Callable[[], None], times: int) -> float:
    start = timeit.default_timer()
    for _ in range(times):
        pairs.clear()
        suggestion()
    end = timeit.default_timer()
    return end - start


time_original = run_suggestion(original, run_count)
time_suggestion_with_partition = run_suggestion(suggestion_with_partition, run_count)
time_suggestion_with_partition_minimize_loops = run_suggestion(
    suggestion_with_partition_minimize_loops, run_count
)
time_suggestion_with_itertools = run_suggestion(suggestion_with_itertools, run_count)

print(f"time_original: {time_original}")
print(f"time_suggestion_with_partition: {time_suggestion_with_partition}")
print(
    f"time_suggestion_with_partition_minimize_loops: {time_suggestion_with_partition_minimize_loops}"
)
print(f"time_suggestion_with_itertools: {time_suggestion_with_itertools}")
```

bdraco: I updated the suggestion above to `suggestion_with_partition_minimize_loops`

Dreamsorcerer: I added the following couple of lines of the codebase into the tests, and, atleast on my machine, itertools comes out marginally better:
```
> python test.py
time_original: 0.14913682299811626
time_suggestion_with_partition: 0.09907035499782069
time_suggestion_with_partition_minimize_loops: 0.07111960600013845
time_suggestion_with_itertools: 0.0670045239967294
> python test.py
time_original: 0.12457079599698773
time_suggestion_with_partition: 0.08952741000393871
time_suggestion_with_partition_minimize_loops: 0.0660165480003343
time_suggestion_with_itertools: 0.0644841980029014
> python test.py
time_original: 0.12488939300237689
time_suggestion_with_partition: 0.09736065800098004
time_suggestion_with_partition_minimize_loops: 0.06708528499439126
time_suggestion_with_itertools: 0.06588387100055115
```

Full script:
```
import timeit

from yarl import URL
import itertools
from typing import Callable


run_count = 20000
hostname = "www.example.com"
request_url = URL(f"http://{hostname}/this/is/a/typical/url")
pairs = []


def original():
    d = hostname
    while d:
        p = request_url.path
        while p:
            pairs.append((d, p))
            p = p.rsplit("/", maxsplit=1)[0]
        pairs.append((d, ""))
        try:
            d = d.split(".", maxsplit=1)[1]
        except IndexError:
            # handle last element for split
            d = ""
    pairs.append(("", ""))
    for p in reversed(pairs):
        ...


def suggestion_with_partition():
    domain_part = hostname
    while domain_part:
        paths = request_url.path
        while paths:
            pairs.append((domain_part, paths))
            paths = paths.rpartition("/")[0]
        pairs.append((domain_part, ""))
        domain_part = domain_part.partition(".")[2]
    pairs.append(("", ""))
    for p in reversed(pairs):
        ...


def suggestion_with_partition_minimize_loops():
    paths = []
    path = request_url.path
    while path:
        paths.append(path)
        path = path.rpartition("/")[0]

    domain_part = hostname
    while domain_part:
        for path in paths:
            pairs.append((domain_part, path))
        pairs.append((domain_part, ""))
        domain_part = domain_part.partition(".")[2]
    pairs.append(("", ""))
    for p in reversed(pairs):
        ...


def suggestion_with_itertools():
    paths = itertools.accumulate(request_url.path.split("/"), lambda x, y: f"{x}/{y}")
    domains = itertools.accumulate(
        reversed(hostname.split(".")), lambda x, y: f"{y}.{x}"
    )
    pairs = itertools.chain((("", ""),), itertools.product(domains, paths))
    for p in pairs:
        ...


def run_suggestion(suggestion: Callable[[], None], times: int) -> float:
    start = timeit.default_timer()
    for _ in range(times):
        pairs.clear()
        suggestion()
    end = timeit.default_timer()
    return end - start


time_original = run_suggestion(original, run_count)
time_suggestion_with_partition = run_suggestion(suggestion_with_partition, run_count)
time_suggestion_with_partition_minimize_loops = run_suggestion(
    suggestion_with_partition_minimize_loops, run_count
)
time_suggestion_with_itertools = run_suggestion(suggestion_with_itertools, run_count)

print(f"time_original: {time_original}")
print(f"time_suggestion_with_partition: {time_suggestion_with_partition}")
print(
    f"time_suggestion_with_partition_minimize_loops: {time_suggestion_with_partition_minimize_loops}"
)
print(f"time_suggestion_with_itertools: {time_suggestion_with_itertools}")
```

Dreamsorcerer: i.e. The original didn't include the extra append in the other tests, which was already included in itertools with chain().
It also didn't include the iteration which could obviously have been rather different when the itertools one does a lot of the work during iteration, rather than before.

bdraco:
```
% python3.12 cookies2.py
time_original: 0.044930542004294693
time_suggestion_with_partition: 0.02851945799193345
time_suggestion_with_partition_minimize_loops: 0.018505583007936366
time_suggestion_with_itertools: 0.026156499996432103
% python3.11 cookies2.py
time_original: 0.06036895800207276
time_suggestion_with_partition: 0.03111395899031777
time_suggestion_with_partition_minimize_loops: 0.01939904199389275
time_suggestion_with_itertools: 0.023035916994558647
% python3.10 cookies2.py
time_original: 0.07786262499575969
time_suggestion_with_partition: 0.037624666991177946
time_suggestion_with_partition_minimize_loops: 0.0272009159962181
time_suggestion_with_itertools: 0.02622854200308211
```

The results seem to be a bit different depending on the python version used

bdraco: Either of the last two are fine with me

Dreamsorcerer: Interesting, looks like newer versions of Python are becoming more performant with loops or lists maybe. Seems like itertools is fairly consistent.

Dreamsorcerer: Yeah, I think we should choose between those 2 based on readability/maintainability. Do you think the itertools is preferable? I like that the logic for each part is concise enough to be in a single line personally.

bdraco: I like the compactness of the partial but dislike having to use the lambdas so I'm on the fence and both solutions seem fine to me

Dreamsorcerer: Yeah, the lambda is probably where it loses some performance. If I take out the lambda it's definitely a bit faster (but, obviously gives the wrong result then).

aiohttp/cookiejar.py
```diff
@@ -261,8 +263,30 @@ def filter_cookies(self, request_url: URL = URL()) -> "BaseCookie[str]":
...
+
+        # shared cookie, it should have max of 1 entry
+        pairs.append(("", ""))
+
+        cookies = itertools.chain.from_iterable(
+            self._cookies[p].values() for p in reversed(pairs)
+        )
+
         # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
```
Dreamsorcerer: If we're happy with itertools, then I think this is the complete change:

```suggestion
        # Get all the path prefixes that might match a cookie (e.g. "", "/foo", "/foo/bar")
        paths = itertools.accumulate(request_url.path.split("/"), lambda x, y: f"{x}/{y}")
        # Get all the subdomains that might match a cookie (e.g. "foo.bar.com", "bar.com", "com")
        domains = itertools.accumulate(
            reversed(hostname.split(".")), lambda x, y: f"{y}.{x}"
        )
        # Create every combination of (domain, path) pairs, plus the shared cookie.
        pairs = itertools.chain((("", ""),), itertools.product(domains, paths))

        # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
        cookies = itertools.chain.from_iterable(
            self._cookies[p].values() for p in pairs
        )
```

aiohttp/cookiejar.py
Dreamsorcerer: Only way I can think to actually remove the lamba is using re.split() to keep the initial / in the split strings, but regex is a lot slower...

bdraco: I was hoping we could have used `"/".join` as the expression but we need multiple args

aiohttp/cookiejar.py
bdraco: Going to commit this change since xiangxli gave it a 👍

tests/test_cookiejar.py
```diff
@@ -837,7 +927,7 @@ def test_pickle_format(cookies_to_send) -> None:
...
...08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xd0h\x1b\x8c\nthirteenth\x94h\x1dh\xddubs\x8c\x0emaxagetest.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x0emax-age-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xde\x8c\x07max-age\x94\x8c\x0260\x94\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xe1h\x1b\x8c\nfourteenth\x94h\x1dh\xeeubs\x8c\x12invalid-values.com\x94h\x08\x86\x94h\x05)\x81\x94(\x8c\x16invalid-max-age-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xef\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xf2h\x1b\x8c\tfifteenth\x94h\x1dh\xfeub\x8c\x16invalid-expires-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94\x8c\x12invalid-values.com\x94\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xffh\x1b\x8c\tsixteenth\x94h\x1dj\x0c\x01\x00\x00ubuu."
```
Dreamsorcerer:
```suggestion
    pickled = b"\x80\x04\x95\xc8\x0b\x00\x00\x00\x00\x00\x00\x8c\x0bcollections\x94\x8c\x0bdefaultdict\x94\x93\x94\x8c\x0chttp.cookies\x94\x8c\x0cSimpleCookie\x94\x93\x94\x85\x94R\x94(\x8c\x00\x94h\x08\x86\x94h\x05)\x81\x94\x8c\rshared-cookie\x94h\x03\x8c\x06Morsel\x94\x93\x94)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94\x8c\x01/\x94\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\x08\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(\x8c\x03key\x94h\x0b\x8c\x05value\x94\x8c\x05first\x94\x8c\x0bcoded_value\x94h\x1cubs\x8c\x0bexample.com\x94h\x08\x86\x94h\x05)\x81\x94(\x8c\rdomain-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\x1e\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah!h\x1b\x8c\x06second\x94h\x1dh-ub\x8c\x14dotted-domain-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94\x8c\x0bexample.com\x94\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah.h\x1b\x8c\x05fifth\x94h\x1dh;ubu\x8c\x11test1.example.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x11subdomain1-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h<\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah?h\x1b\x8c\x05third\x94h\x1dhKubs\x8c\x11test2.example.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x11subdomain2-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94hL\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ahOh\x1b\x8c\x06fourth\x94h\x1dh[ubs\x8c\rdifferent.org\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x17different-domain-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\\\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah_h\x1b\x8c\x05sixth\x94h\x1dhkubs\x8c\nsecure.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\rsecure-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94hl\x8c\x07max-age\x94h\x08\x8c\x06secure\x94\x88\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ahoh\x1b\x8c\x07seventh\x94h\x1dh{ubs\x8c\x0cpathtest.com\x94h\x08\x86\x94h\x05)\x81\x94(\x8c\x0eno-path-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h|\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\x7fh\x1b\x8c\x06eighth\x94h\x1dh\x8bub\x8c\x0cpath1-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94\x8c\x0cpathtest.com\x94\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\x8ch\x1b\x8c\x05ninth\x94h\x1dh\x99ubu\x8c\x0cpathtest.com\x94\x8c\x04/one\x94\x86\x94h\x05)\x81\x94\x8c\x0cpath2-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x9b\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\x9a\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\x9eh\x1b\x8c\x05tenth\x94h\x1dh\xaaubs\x8c\x0cpathtest.com\x94\x8c\x08/one/two\x94\x86\x94h\x05)\x81\x94(\x8c\x0cpath3-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\xac\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xab\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xafh\x1b\x8c\x08eleventh\x94h\x1dh\xbbub\x8c\x0cpath4-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94\x8c\t/one/two/\x94\x8c\x07comment\x94h\x08\x8c\x06domain\x94\x8c\x0cpathtest.com\x94\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xbch\x1b\x8c\x07twelfth\x94h\x1dh\xcaubu\x8c\x0fexpirestest.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x0eexpires-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94\x8c\x1cTue, 1 Jan 2999 12:00:00 GMT\x94\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xcb\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xceh\x1b\x8c\nthirteenth\x94h\x1dh\xdbubs\x8c\x0emaxagetest.com\x94h\x08\x86\x94h\x05)\x81\x94\x8c\x0emax-age-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xdc\x8c\x07max-age\x94\x8c\x0260\x94\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xdfh\x1b\x8c\nfourteenth\x94h\x1dh\xecubs\x8c\x12invalid-values.com\x94h\x08\x86\x94h\x05)\x81\x94(\x8c\x16invalid-max-age-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94h\xed\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xf0h\x1b\x8c\tfifteenth\x94h\x1dh\xfcub\x8c\x16invalid-expires-cookie\x94h\r)\x81\x94(\x8c\x07expires\x94h\x08\x8c\x04path\x94h\x11\x8c\x07comment\x94h\x08\x8c\x06domain\x94\x8c\x12invalid-values.com\x94\x8c\x07max-age\x94h\x08\x8c\x06secure\x94h\x08\x8c\x08httponly\x94h\x08\x8c\x07version\x94h\x08\x8c\x08samesite\x94h\x08u}\x94(h\x1ah\xfdh\x1b\x8c\tsixteenth\x94h\x1dj\n\x01\x00\x00ubuu."
```

CHANGES/7583.feature
```diff
@@ -0,0 +1,14 @@
...
+This may break existing cookies that have been saved with `CookieJar.save()`. Cookies can be migrated with this script::
+
+    import pickle
+    with file_path.open("rb") as f:
+        cookies = pickle.load(f)
+
+    morsels = [(name, m) for c in cookies.values() for name, m in c.items()]
+    cookies.clear()
+    for name, m in morsels:
+        cookies[(m["domain"], "" if m["path"] == "/" else m["path"])][name] = m
```
Dreamsorcerer:
```suggestion
        cookies[(m["domain"], m["path"].rstrip("/"))][name] = m
```

Dreamsorcerer:
@bdraco Made a few more optimisations after looking at codecov. Would be good if you can double check them.

bdraco:
Starting my day soon, I'll be at my desk in about an hour run through everything

bdraco:
This looks good. I'll put it on production

bdraco:
Nothing obvious broke on production.

Will do a `py-spy` to look at the flow ...

bdraco:
Looks like `py-spy` won't work here as its too fast now to get any samples compared to the rest of the program.

Will do cProfile instead

bdraco:
I think this is good to go.

PR #7954, opened 2023-12-09
Implement happy eyeballs (RFC 8305)

### LINKED ISSUE #4451: Support Happy Eyeballs from Python 3.8

bdraco:
## What do these changes do?

Implements Happy Eyeballs support (https://datatracker.ietf.org/doc/html/rfc8305)

Dual stack connections where IPv6 is not working should now be successful.

## Related issue number

fixes #4451

## Checklist

--

TODO

- [x] docs (params `happy_eyeballs_delay` and `interleave` function the same as cpython but we should point that out, and how to disable `happy_eyeballs_delay` by setting it to `None`)
- [x] more production and functional testing
- [x] tests for ipv4 fallback
- [x] tests for interleave
- [x] tests for family
- [x] timeline

This PR is a first step as I need to make sure this library works for the other projects as well. Its likely to sit for a bit while I test the library with all the projects that need it so there aren't any need to do breaking changes before all the other projects start consuming it:
- https://github.com/esphome/aioesphomeapi/pull/789
- https://github.com/Jc2k/aiohomekit/pull/349

--


aiohttp/connector.py
```diff
@@ -956,16 +958,33 @@ def _get_fingerprint(self, req: ClientRequest) -> Optional["Fingerprint"]:
...
+        local_addrs_infos = None
+        if self._local_addr:
+            host, port = self._local_addr
+            is_ipv6 = helpers.is_ipv6_address(host)
+            family = socket.AF_INET6 if is_ipv6 else socket.AF_INET
+            if is_ipv6:
+                addr = (host, port, 0, 0)
+            else:
+                addr = (host, port)
+            local_addrs_infos = [(family, 0, 0, addr)]
```
bdraco: Should probably be a helper

aiohttp/connector.py
```diff
@@ -1160,6 +1170,17 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
...
+                    bad_peer = sock.getpeername()
+                    bad_addrs_infos: List[aiohappyeyeballs.AddrInfoType] = []
+                    for addr_info in addr_infos:
+                        if addr_info[-1][0] == bad_peer[0]:
+                            bad_addrs_infos.append(addr_info)
+                    if bad_addrs_infos:
+                        for bad_addr_info in bad_addrs_infos:
+                            addr_infos.remove(bad_addr_info)
+                    else:
+                        addr_infos.pop(0)
```
bdraco: This isn't great, would be better to make a helper here

bdraco:
functional tests seem to be ok

Anything that does this will fail since `aiohappyeyeballs.start_connection` isn't mocked
```
        self.loop.create_connection = make_mocked_coro((mock.Mock(), mock.Mock()))
```

aiohttp/connector.py
```diff
@@ -69,6 +72,76 @@
...
+    """Pop addr_info from the list of addr_infos by addr."""
+    bad_addrs_infos: List[aiohappyeyeballs.AddrInfoType] = []
+    for addr_info in addr_infos:
+        if addr_info[-1][0] == addr[0]:
+            bad_addrs_infos.append(addr_info)
+    if bad_addrs_infos:
+        for bad_addr_info in bad_addrs_infos:
+            addr_infos.remove(bad_addr_info)
+    else:
+        addr_infos.pop(0)
```
bdraco: This should probably raise instead of popping the first one since it means we can't find the bad peer for some reason.  Likely a problem comparing the addresses since the str may be formatted different. 

Need to add coverage for this

requirements/constraints.txt
```diff
@@ -1,11 +1,13 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.8
+# by the following command:
 #
-#    pip-compile --allow-unsafe --output-file=requirements/constraints.txt --resolver=backtracking --strip-extras requirements/constraints.in
+#    pip-compile --allow-unsafe --output-file=requirements/constraints.txt --strip-extras requirements/constraints.in
```
bdraco: TODO: revert header changes manually

requirements/dev.txt
```diff
@@ -1,11 +1,13 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
-# To update, run:
+# This file is autogenerated by pip-compile with Python 3.8
+# by the following command:
 #
-#    pip-compile --allow-unsafe --output-file=requirements/dev.txt --resolver=backtracking --strip-extras requirements/dev.in
+#    pip-compile --allow-unsafe --output-file=requirements/dev.txt --strip-extras requirements/dev.in
 #
```
bdraco: TODO: revert header changes manually

requirements/test.txt
```diff
@@ -1,11 +1,13 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
+# This file is autogenerated by pip-compile with Python 3.8
 # by the following command:
 #
-#    pip-compile --allow-unsafe --output-file=requirements/test.txt --resolver=backtracking --strip-extras requirements/test.in
+#    pip-compile --allow-unsafe --output-file=requirements/test.txt --strip-extras requirements/test.in
 #
```
bdraco: TODO: revert header changes manually

tests/conftest.py
```diff
@@ -208,3 +208,11 @@ def netrc_contents(
...
 
     return netrc_file_path
+
+
+@pytest.fixture
+def start_connection():
+    with mock.patch(
+        "aiohttp.connector.aiohappyeyeballs.start_connection"
+    ) as start_connection_mock:
+        yield start_connection_mock
```
bdraco:
```suggestion
```

aiohttp/connector.py
```diff
@@ -1160,6 +1199,10 @@ def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
                     if not self._cleanup_closed_disabled:
                         self._cleanup_closed_transports.append(transp)
                     last_exc = exc
+                    # Remove the bad peer from the list of addr_infos
+                    sock: socket.socket = transp.get_extra_info("socket")
+                    bad_peer = sock.getpeername()
+                    aiohappyeyeballs.remove_addr_infos(addr_infos, bad_peer[0])
```
bdraco: This should take the whole peer since it could mean we have different ports even if it's unlikely

bdraco: https://bugs.python.org/issue42620

requirements/test.txt
Dreamsorcerer: I wonder if it's reasonable to add `--no-header`, so this causes less issues in future...

setup.cfg
```diff
@@ -47,6 +47,7 @@ zip_safe = False
 include_package_data = True
 
 install_requires =
+  aiohappyeyeballs >= 2.0.0
```
Dreamsorcerer: Should we maybe put this under speedups and make it optional?

bdraco: I thought about this from the perspective that IPv6 support should be a base feature, and without Happy Eyeballs, IPV6 tends to get turned off because it cannot establish a connection before the timeout with many network setups.

At least from my experience with Home Assistant, we have IPv6 turned off in many places by setting the address family to AF_INET only because the lack of Happy Eyeballs makes using IPv6 untenable.

From my perspective, it is much more valuable if IPv6 works out of the box with the developer not having to tinker to get things working.

Dreamsorcerer: OK, seems reasonable then. Hopefully we can get it into cpython in the next couple of releases and then drop the dependency again.

Dreamsorcerer (approved):

bdraco:
After a brief detour to fix some mdns bugs in zeroconf, I've got the other libraries that need this working as expected.

I need to do more functional testing and update the connector tests

tests/test_connector.py
```diff
@@ -594,6 +601,7 @@ async def test_tcp_connector_multiple_hosts_errors(loop: Any) -> None:
     ip4 = "192.168.1.4"
```
bdraco, line 601: We need a similar test to make sure it falls back to the ipv4 host but should be quite light as to not reimplement testing the library

aiohttp/connector.py
```diff
@@ -69,6 +72,39 @@
...
+    local_addr: Optional[Tuple[str, int]]
+) -> Optional[List[aiohappyeyeballs.AddrInfoType]]:
+    """Convert a local_addr tuple to a list of addr_infos."""
+    if local_addr is None:
+        return None
+    host, port = local_addr
+    is_ipv6 = helpers.is_ipv6_address(host)
+    family = socket.AF_INET6 if is_ipv6 else socket.AF_INET
+    addr = (host, port, 0, 0) if is_ipv6 else (host, port)
+    return [(family, socket.SOCK_STREAM, socket.IPPROTO_TCP, "", addr)]
```
bdraco: This isn't aiohttp specific so it could be moved to aiohappyeyeballs

aiohttp/connector.py
```diff
@@ -69,6 +72,39 @@
...
+    addr_infos: List[aiohappyeyeballs.AddrInfoType] = []
+    for hinfo in hosts:
+        host = hinfo["host"]
+        is_ipv6 = helpers.is_ipv6_address(host)
+        family = socket.AF_INET6 if is_ipv6 else socket.AF_INET
+        addr = (host, hinfo["port"], 0, 0) if is_ipv6 else (host, hinfo["port"])
+        addr_infos.append(
+            (family, socket.SOCK_STREAM, socket.IPPROTO_TCP, hinfo["hostname"], addr)
+        )
+    return addr_infos
```
bdraco: We could store the addresses in the cache in python's `getaddrinfo` format to avoid having to convert.  It would be more efficient and more portable.... but its a bit less readable.  Something to consider for the future though

aiohttp/connector.py
```diff
@@ -69,6 +72,39 @@
...
+    """Converts the list of hosts to a list of addr_infos.
+
+    The list of hosts is the result of a DNS lookup. The list of
+    addr_infos is the result of a call to `socket.getaddrinfo()`.
+    """
+    addr_infos: List[aiohappyeyeballs.AddrInfoType] = []
+    for hinfo in hosts:
+        host = hinfo["host"]
+        is_ipv6 = helpers.is_ipv6_address(host)
+        family = socket.AF_INET6 if is_ipv6 else socket.AF_INET
```
bdraco: `is_ipv6_address` is likely overkill here as we don't need to validate it here, only determine the family

aiohttp/connector.py
bdraco: This one is a bit of a special case though since ipv6 is always a 4 arg tuple, and this only takes a 2 arg tuple

bdraco: Need to see how cpython handles this

bdraco: https://github.com/python/cpython/blob/9cdf05bc28c5cd8b000b9541a907028819b3d63e/Lib/asyncio/base_events.py#L1472

bdraco: cpython handles it the exact same way, it adds 0 0
https://github.com/python/cpython/blob/9cdf05bc28c5cd8b000b9541a907028819b3d63e/Lib/asyncio/base_events.py#L99C51-L99C72

So this is fine

aiohttp/connector.py
bdraco: We don't need validation, only to get the right family

9e36ed1d830d5d6f02cc79cfd4c2a1777f22fba5

aiohttp/connector.py
bdraco: Moved to the lib since its useful elsewhere and not aiohttp specific

bdraco:
At this point everything is working how I expect it to with functional testing.

The `aiohappyeyeballs` should be bug for bug compatible with cpython (hopefully there aren't any obvious ones) since its almost 100% the same code as `create_connection` with only added typing and unneeded code removed.

I've asked some of the Home Assistant devs to take a look at all the PRs to see if I missed anything that might require a breaking change in the `aiohappyeyeballs` library later

aiohttp/connector.py
```diff
@@ -1076,6 +1091,33 @@ async def _start_tls_connection(
...
+            family = socket.AF_INET6 if is_ipv6 else socket.AF_INET
+            if self._family and self._family != family:
+                continue
+            addr = (host, hinfo["port"], 0, 0) if is_ipv6 else (host, hinfo["port"])
+            addr_infos.append(
+                (
+                    family,
+                    socket.SOCK_STREAM,
+                    socket.IPPROTO_TCP,
+                    hinfo["hostname"],
```
bdraco: This value is completely ignored since it's only used as a result of dns resolution which is not done here.  We should remove it as it's useless here

```suggestion
                    "",
```

tests/conftest.py
```diff
@@ -208,3 +208,11 @@ def netrc_contents(
     monkeypatch.setenv("NETRC", str(netrc_file_path))
 
     return netrc_file_path
+
+
+@pytest.fixture
+def start_connection():
+    with mock.patch(
+        "aiohttp.connector.aiohappyeyeballs.start_connection"
```
Dreamsorcerer: If possible, it would be preferred to use `autospec=True, spec_set=True`, to minimise the possibility of misusing the mock.

tests/test_connector.py
```diff
@@ -623,11 +631,27 @@ async def _resolve_host(host, port, traces=None):
...
+        nonlocal os_error, certificate_error, ssl_error, fingerprint_error
+        nonlocal connected
+
+        addr_infos: List[AddrInfoType] = kwargs["addr_infos"]
+
+        first_addr_info = addr_infos[0]
+        first_addr_info_addr = first_addr_info[-1]
+        addrs_tried.append(first_addr_info_addr)
+
+        mock_socket = mock.Mock()
```
Dreamsorcerer: Similar with mocks like this. If we have a class to base it off, then we can hopefully do `create_autospec(Socket, spec_set=True, instance=True)`

bdraco: 349203019c80ce4af4e3514eff3e8976c4e485a5

tests/conftest.py
bdraco: 349203019c80ce4af4e3514eff3e8976c4e485a5

bdraco:
I'm still doing testing with this, and I think its ready to move to wider production testing.

I'm going to spend a few cycles on `test_unsupported_upgrade` in https://github.com/aio-libs/aiohttp/pull/7960 but its a bit difficult without a windows box

bdraco:
So far so good on production 3.9+backported.

Doing performance testing now

bdraco:
performance looks good. All the time is nearly `loop.sock_connect` so should have no significant change

bdraco:
Performance and production testing looks good.

I've got reviews on the other use cases so I think this is good to go for a final review.

I want to sleep on it and look it all over in the morning end to end again though before I mark it ready

bdraco:
Still testing out great this morning

requirements/test.txt
bdraco: That seems reasonable to do in another PR 👍

bdraco:
I think this is good to merge at this point, but I'm slow rolling it because ESPHome is currently in beta which also uses the `aiohappyeyeballs` lib so no rush in merging as I'm happy to wait for the beta feedback in case anything has to change with the lib, but it now seems unlikely at this point.

bdraco:
ESPHome is releasing today with the new lib, HA is doing beta next week with it

bdraco:
This change in esphome release seems to be going swimmingly.  Next week we get beta feedback on it in aiohomekit from Home Assistant

bdraco:
The changes in aiohomekit seem to be going well in the HA beta.

bdraco:
Ha beta has been without issues in this area.  The full release happens tomorrow.  As long as everything goes smoothly, I'll merge this shortly after

bdraco:
HA release shipped today.  No issues so far.  All looking good

bdraco:
Everything seems solid with the HA release related to this. 

That's the last hurdle.

PR #7956, opened 2023-12-11
Fixed Inappropriate Logical Expression

nydelzaf:
## What do these changes do?
* Removed a break statement inside the 'finally' block of a try-catch statement. Using 'break' or 'continue' inside a finally block can suppress the propagation of unhandled exception.

## Checklist

## Details
While triaging your project, our bug fixing tool generated the following message(s)-

> In file: [web_protocol.py](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_protocol.py#L606), method: start, a break statement is used inside a finally block. This will discard the temporarily saved unhandled exception which was raised inside the try, else or except block. iCR suggested that the break statement should not be used inside the finally block.

### How It Works
**Incorrect Use**
You can have a look at the example below-
```python
foo = [1, 2, 3, 4, 5]
bar = None
for _ in range(5):
    try:
        # IndexError
        bar = foo[10]
        print("Inside try block")
    except:
        print("Inside except block")
        # raising another exception
        bar = foo[100]
    else:
        print("Inside else block")
        # let's try another exc
        bar = foo[100]
    finally:
        break
```
```
Inside except block
```

**Correct Use**
```python
foo = [1, 2, 3, 4, 5]
bar = None
for _ in range(5):
    try:
        # IndexError
        bar = foo[10]
        print("Inside try block")
    except:
        print("Inside except block")
        # raising another exception
        bar = foo[100]
    else:
        print("Inside else block")
        # let's try another exc
        bar = foo[100]
    finally:
        pass
```
```
Inside except block
  File "try-catch-break.py", line 13, in <module>
    bar = foo[10]
IndexError: list index out of range

  File "try-catch-break.py", line 18, in <module>
    bar = foo[100]
IndexError: list index out of range
```
From these two examples, we can see that if `break` is used in the `finally` block, it ignores any exceptions that takes place in `except` or `else`. But if `pass` is used, or `finally` block isn't used, the exceptions are reported correctly. As a result, the correct use is not to use `break` or `continue` inside the finally block.

I've removed the `else and break` statement from the code since it'll lead to a better understanding of any exceptions that may take place during code execution.

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/CHANGES/7956.bugfix
+++ b/CHANGES/7956.bugfix
@@ -1,0 +1,1 @@
+Removed break statement inside the 'finally' block of a try-catch statement otherwise it can suppress unhandled exceptions.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -589,6 +589,9 @@ class RequestHandler(BaseProtocol):
             except Exception as exc:
                 self.log_exception("Unhandled exception", exc_info=exc)
                 self.force_close()
+            except BaseException:
+                self.force_close()
+                raise
             finally:
                 if self.transport is None and resp is not None:
                     self.log_debug("Ignored premature client disconnection.")
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -602,8 +602,6 @@ class RequestHandler(BaseProtocol):
                                 self._keepalive_handle = loop.call_at(
                                     now + keepalive_timeout, self._process_keepalive
                                 )
-                    else:
-                        break
 
         # remove handler, close transport if no handlers left
         if not self._force_close:
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -50,6 +50,7 @@ Arie Bovenberg
 Arseny Timoniq
 Artem Yushkovskiy
 Arthur Darcet
+Ataf Fazledin Ahamed
 Austin Scola
 Ben Bader
 Ben Greiner


aiohttp/web_protocol.py
```diff
@@ -602,8 +602,6 @@ async def start(self) -> None:
                                 self._keepalive_handle = loop.call_at(
                                     now + keepalive_timeout, self._process_keepalive
                                 )
-                    else:
-                        break
```
Dreamsorcerer, line 606: Good catch, I'm just trying to figure out if the transport.close() below should be run in the case of a BaseException that is uncaught..
@bdraco Do you think it's fine like this?

bdraco: I think this is a risk of the transport not getting closed on BaseException. I'm not sure what that exception would be though. 

maybe something like...
```diff
diff --git a/aiohttp/web_protocol.py b/aiohttp/web_protocol.py
index 3cefd5df..f2dcf5aa 100644
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -589,6 +589,9 @@ class RequestHandler(BaseProtocol):
             except Exception as exc:
                 self.log_exception("Unhandled exception", exc_info=exc)
                 self.force_close()
+            except BaseException:
+                self.force_close()
+                raise
             finally:
                 if self.transport is None and resp is not None:
                     self.log_debug("Ignored premature client disconnection.")
```

Dreamsorcerer: > I'm not sure what that exception would be though.

Well, in most cases, it'll probably be a KeyboardInterrupt, which may or may not lead to the program exiting..

> maybe something like...

Seems fine with me.

Dreamsorcerer: Currently, the tests don't even complete, so the current change doesn't seem correct. Maybe with the proposed change it will work correctly..

Dreamsorcerer: Tests are still not completing. @fazledyn-or Any interest in finishing this PR?

nydelzaf: > Tests are still not completing. @fazledyn-or Any interest in finishing this PR?

It's been long since I had worked on it, and yeah- I'll work toward finishing the PR. However, I don't understand the reason behind introducing an empty `raise`.

Dreamsorcerer: I think in the original code if a BaseException happened, it could trigger the `break`, which then leads to closing the transport outside the exception handler. We shouldn't be suppressing a BaseException though, so to be safe we want to force close on a BaseException and then reraise it.

Looking over that code again, it feels like that block of logic should maybe not be in the finally block...

mikezolo (approved):
One of the best based and explicitly explained pull requests I've read.

And the fix is indeed substantial.

CHANGES/7956.bugfix
```diff
@@ -0,0 +1 @@
+Removed break statement inside the 'finally' block of a try-catch statement otherwise it can suppress unhandled exceptions.
```
webknjaz, line 1: FWIW a changelog entry should explain the high-level effect for the end-users. Throwing internal implementation details at them doesn't seem useful..

webknjaz:
@fazledyn-or is this still on your radar? Unfortunately, we cannot merge this until it stops shuttering the CI...

Cycloctane:
superseded by #10434

PR #7960, opened 2023-12-11
DO NOT MERGE: debug test_unsupported_upgrade

bdraco:
## Checklist


bdraco:
its getting stuck at

bdraco:
```python
                        if not empty_body and (
                            (length is not None and length > 0)
                            or msg.chunked
                            and not msg.upgrade
                        ):
                            payload = StreamReader(
                                self.protocol,
                                timer=self.timer,
                                loop=loop,
                                limit=self._limit,
                            )
                            payload_parser = HttpPayloadParser(
```

`msg.upgrade` is set, so we don't ever read the payload so it times out

aiohttp/http_parser.py
```diff
@@ -341,17 +344,24 @@ def get_content_length() -> Optional[int]:
...
-                        self._upgraded = msg.upgrade
-
                         method = getattr(msg, "method", self.method)
                         # code is only present on responses
                         code = getattr(msg, "code", 0)
 
+                        import pprint
+
+                        pprint.pprint(["upgraded", msg.upgrade, msg])
+                        self._upgraded = code == 101 and msg.upgrade
```
bdraco: c parser seems to only set this if code is 101

https://github.com/nodejs/llhttp/blob/8498ef9d8b0e9539c8c331cf59213529287789e1/src/native/http.c#L21

aiohttp/http_parser.py
```diff
@@ -505,6 +517,9 @@ def parse_headers(
                 close_conn = False
             # https://www.rfc-editor.org/rfc/rfc9110.html#name-101-switching-protocols
             elif v == "upgrade" and headers.get(hdrs.UPGRADE):
```
steverep, line 545: Isn't this the source of the problem for #6446?  Just because the `Upgrade` header has content, doesn't mean there will be an upgrade.  The server has every right to ignore it if the scheme is unsupported.

tests/test_web_server.py
```diff
@@ -31,13 +37,29 @@ async def test_unsupported_upgrade(aiohttp_raw_server, aiohttp_client) -> None:
...
+                result = await request.read()
+            pprint.pprint(["handler read", result])
+            return web.Response(body=result)
+        except Exception as e:
+            import pprint
+
+            pprint.pprint(["handler except", e])
+            raise
 
     upgrade_headers = {"Connection": "Upgrade", "Upgrade": "unsupported_proto"}
```
steverep, line 54: I'm not sure what is parsing the `Upgrade` header, but the fact that it's not a valid token at all should also be considered ([IANA upgrade token registry](https://www.iana.org/assignments/http-upgrade-tokens/http-upgrade-tokens.xhtml).  In fact, it even contains an invalid character, "_", for URL schemes ([RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738#section-2.1)).

aiohttp/http_parser.py
Dreamsorcerer: If you think you can provide a fix, that'd be great. I believe there is already an xfail test from that issue, so as long as you can get that test passing..

tests/test_web_server.py
Dreamsorcerer: I would assume that would be handled in the parsers (i.e. llhttp or http_parser.py, depending on whether extensions are enabled).

steverep: Ah okay.  My point is, especially it it happens externally, the test should know and account for the behavior.

I would hope, no matter which parser is used, that "unsupported_proto" would simply be thrown away and equivalent to "".  Then the test should also test a valid, but unsupported, token.

aiohttp/http_parser.py
steverep: One problem at a time for this newbie... Brotli first 😉

bdraco:
Looks like its an actual bug and the test only passed because of a race

https://github.com/aio-libs/aiohttp/pull/8252

PR #7978, opened 2023-12-17
Fix websocket connection leak

### LINKED ISSUE #6325: Too many websocket client disconnect, memory not release
### Describe the bug

I have a aiohttp server, it supply a router for websoccket, when I use 10000 client to connect the server. and disconnect after 5 seconds, the server will not release the memory.

server handler
```python
async def websocket_handler(request):
    ws = web.WebSocketResponse()
    await ws.prepare(request)
    try:
        await ws.send_str('hi')
        async for msg in ws:
                if msg.data == 'close':
                    break
                await ws.send_str(msg)
    except (Exception, asyncio.CancelledError) as e:
        logging.warning(f'websocket connection closed: {str(e)}')
    finally:
        await ws.close()
    return ws
```

client script
```python
async def connect_ws(id):
    url = f'ws://localhost:8000/websocket'
    ws = create_connection(url)
    response = ws.recv()
    print(str(id) + response)
    await asyncio.sleep(5)
    ws.close()

async def run():
    for i in range(10000):
        asyncio.ensure_future(connect_ws(i))
    await asyncio.sleep(1000)

if __name__ == '__main__':
    loop.run_until_complete(run())
```

before run script:
image

after finish run script:
image

after I use gc.collectI():

### To Reproduce

1. start the server
2. start the client script
3. watching the memory used

### Expected behavior

memory will be released, 

### Logs/tracebacks

```python-traceback
not error info, but run script will reshow this problem
```

### Python Version

```console
$ python --version
Python 3.8.10
Python 3.7.8
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.8.1
Version: 3.6.2
```

### multidict Version

```console
Version: 4.7.6
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.6.3
```

### OS

Windows 10 21H1 19043.1348

### Related component

Server

### Additional context

_No response_

### Code of Conduct

bdraco:
fixes #6325

The existing implementation never closed the transport unless `heartbeat` was enabled (not default), the pong was not received, and `.close()` had not been called before the pong was not received since it cancelled the timer. 

We now always close the transport after we get the `WSMsgType.CLOSE` message or an abnormal closure occurs instead of hoping `asyncio` calls `connection_lost` which may not happen in a reasonable time frame (or ever?) for a variety of reasons deeper in the stack (network, bad client implementations, etc). 

Any time we are closing/setting `self._close_code` we must either call `self.close()` or close the transport otherwise its possible to leak the connection. 

To facilitate that, and reduce the chance of future refactoring re-introducing the problem, `_set_code_close_transport` is added which sets the code and closes the transport.  

All other places `self._close_code` is set have been audited to ensure they also call `self.close()` There is one exception if `self._autoclose` is set to `False` and the client closes the connection, than its the responsibility of the caller to close the connection as otherwise it can leak. This is now documented.

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -497,10 +497,10 @@ class WebSocketResponse(StreamResponse):
                 self._close_code = msg.data
                 # Could be closed while awaiting reader.
                 if not self._closed and self._autoclose:  # type: ignore[redundant-expr]
-                    # The client is going to close the connection
-                    # out from under us so we do not want to drain
-                    # any pending writes as it will likely result
-                    # writing to a broken pipe.
+                    # The client is likely going to close the
+                    # connection out from under us so we do not
+                    # want to drain any pending writes as it will
+                    # likely result writing to a broken pipe.
                     await self.close(drain=False)
             elif msg.type == WSMsgType.CLOSING:
                 self._closing = True
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -162,9 +162,8 @@ class WebSocketResponse(StreamResponse):
     def _pong_not_received(self) -> None:
         if self._req is not None and self._req.transport is not None:
             self._closed = True
-            self._close_code = WSCloseCode.ABNORMAL_CLOSURE
+            self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
             self._exception = asyncio.TimeoutError()
-            self._req.transport.close()
 
     async def prepare(self, request: BaseRequest) -> AbstractStreamWriter:
         # make pre-check to don't hide it by do_handshake() exceptions
@@ -444,7 +443,8 @@ class WebSocketResponse(StreamResponse):
     def _set_code_close_transport(self, code: WSCloseCode) -> None:
         """Set the close code and close the transport."""
         self._close_code = code
-        self._writer.transport.close()
+        if self._req is not None and self._req.transport is not None:
+            self._req.transport.close()
 
     async def receive(self, timeout: Optional[float] = None) -> WSMessage:
         if self._reader is None:
@@ -475,7 +475,7 @@ class WebSocketResponse(StreamResponse):
                     set_result(waiter, True)
                     self._waiting = None
             except (asyncio.CancelledError, asyncio.TimeoutError):
-                self._close_code = WSCloseCode.ABNORMAL_CLOSURE
+                self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
                 raise
             except EofStream:
                 self._close_code = WSCloseCode.OK
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -496,12 +496,15 @@ class WebSocketResponse(StreamResponse):
                 self._closing = True
                 self._close_code = msg.data
                 # Could be closed while awaiting reader.
-                if not self._closed and self._autoclose:  # type: ignore[redundant-expr]
+                if not self._closed:
                     # The client is going to close the connection
                     # out from under us so we do not want to drain
                     # any pending writes as it will likely result
                     # writing to a broken pipe.
-                    await self.close(drain=False)
+                    if self._autoclose:
+                        await self.close(drain=False)
+                    else:
+                        self._set_code_close_transport(msg.data)
             elif msg.type == WSMsgType.CLOSING:
                 self._closing = True
             elif msg.type == WSMsgType.PING and self._autoping:
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -496,15 +496,12 @@ class WebSocketResponse(StreamResponse):
                 self._closing = True
                 self._close_code = msg.data
                 # Could be closed while awaiting reader.
-                if not self._closed:
+                if not self._closed and self._autoclose:  # type: ignore[redundant-expr]
                     # The client is going to close the connection
                     # out from under us so we do not want to drain
                     # any pending writes as it will likely result
                     # writing to a broken pipe.
-                    if self._autoclose:
-                        await self.close(drain=False)
-                    else:
-                        self._set_code_close_transport(msg.data)
+                    await self.close(drain=False)
             elif msg.type == WSMsgType.CLOSING:
                 self._closing = True
             elif msg.type == WSMsgType.PING and self._autoping:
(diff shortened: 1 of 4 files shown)


aiohttp/web_ws.py
```diff
@@ -162,9 +162,8 @@ def _send_heartbeat(self) -> None:
     def _pong_not_received(self) -> None:
         if self._req is not None and self._req.transport is not None:
             self._closed = True
-            self._close_code = WSCloseCode.ABNORMAL_CLOSURE
+            self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
```
bdraco, line 165: Having trouble writing a test for this but that's probably my lack of sleep

We could leave this as is since it's not covered by existing tests.

Will keep trying though

bdraco:
pushed to production a new hours ago. all working well

Dreamsorcerer (approved):
I think if the leak in that demo is gone, then lets merge this.

bdraco:
Im not longer able reproduce it with the provided script after this change.

I will deploy the change more widely on my production later today and merge if no regressions are observed

bdraco:
deployed widely.  watching logs

bdraco:
Logs look good. Home Assistant's object logger / leak checker doesn't see the leak anymore.

johntdyer:
@Dreamsorcerer  any idea when 3.9.2 might be released....  presumably including this change?

Dreamsorcerer:
Will probably come round to it shortly, have been fully focused on a PR for aiohttp-admin for the last month.

spikefishjohn:
Howdy all and @bdraco! 

I'm working on a [bug report](https://github.com/GNS3/gns3-server/issues/2320) for GNS3 that I think this PR lines up with. The GNS3 issue is that when a web socket client sends a close message the server never closes the tcp socket. I tried the patch above but it doesn't address the issue. 

High level GNS3 is calling [ws.receive()](https://github.com/GNS3/gns3-server/blob/master/gns3server/handlers/api/controller/notification_handler.py#L33C14-L33C27) at which point aiohttp receive() the close message. 

I believe the problem is because ws.receive is setting [self._closing = True](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L474) when it reiceves a close message. This then causes self.close() to [return here](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) which prevents [if msg.type == WSMsgType.CLOSE:](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L413) from being reached.

I'm almost thinking [this should be removed](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) but i'm not sure what the intent of that is so i'm unsure if that is the proper fix. 

Should I open a new issue for this?

bdraco:
> Howdy all and @bdraco! 
> 
> 
> 
> I'm working on a [bug report](https://github.com/GNS3/gns3-server/issues/2320) for GNS3 that I think this PR lines up with. The GNS3 issue is that when a web socket client sends a close message the server never closes the tcp socket. I tried the patch above but it doesn't address the issue. 
> 
> 
> 
> High level GNS3 is calling [ws.receive()](https://github.com/GNS3/gns3-server/blob/master/gns3server/handlers/api/controller/notification_handler.py#L33C14-L33C27) at which point aiohttp receive() the close message. 
> 
> 
> 
> I believe the problem is because ws.receive is setting [self._closing = True](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L474) when it reiceves a close message. This then causes self.close() to [return here](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) which prevents [if msg.type == WSMsgType.CLOSE:](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L413) from being reached.
> 
> 
> 
> I'm almost thinking [this should be removed](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) but i'm not sure what the intent of that is so i'm unsure if that is the proper fix. 
> 
> 
> 
> Should I open a new issue for this?

Yes please.  Thank you

spikefishjohn:
Will do, thanks for the quick reply.

pxscott:
Hi,

With the changes for this bug, the behavior of asyncio.wait_for(ws.receive(), timeout=1) has changed. Before, this could be used to do a poll for new messages, but now hitting the timeout causes the websocket to be closed. This causes problems in our code, so we are pinning to 3.9.1.

- Scott McRae

bdraco:
> Hi,
> 
> With the changes for this bug, the behavior of asyncio.wait_for(ws.receive(), timeout=1) has changed. Before, this could be used to do a poll for new messages, but now hitting the timeout causes the websocket to be closed. This causes problems in our code, so we are pinning to 3.9.1.
> 
> * Scott McRae

Wrap the receive call in `asyncio.shield` to prevent cancellation

Dreamsorcerer:
> Use `asyncio.wait` instead or wrap the receive call in `asyncio.shield` to prevent cancellation

We should probably look at this more closely. The TimeoutError must come from our timeout() call, so that behaviour is probably correct. But, a CancelledError could come from anywhere and maybe shouldn't close the connection. It may be that we need to check if the cancellation came from outside the task (`asyncio.current_task().cancelling()`) and skip closing otherwise.

More information about this at:
https://superfastpython.com/asyncio-task-cancellation-best-practices/#Practice_1_Do_Not_Consume_CancelledError
And we recently implemented this in aiohttp-sse: https://github.com/aio-libs/aiohttp-sse/pull/459/files

bdraco:
https://github.com/aio-libs/aiohttp/pull/8251 will restore the behavior to keep the connection open if receive times out or is cancelled

Dreamsorcerer:
> asyncio.wait_for(ws.receive(), timeout=1)

Also, worth pointing out that receive has a `timeout` parameter, so you don't need `wait_for()` at all, in fact you're just doubling the number of timeouts being constructed.

pxscott:
Yeah, I noticed that when debugging this. This code was written awhile back, I think before that parameter existed. But changing to use the parameter didn't help with the close logic, so I just pinned the version for now. I'll try using the parameter when i do upgrade once this patch is released.

PR #7990, opened 2023-12-21
Updated Contributing.txt with more details

mdsahilnoob:
## Checklist

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -13,26 +13,26 @@ I hope everybody knows how to work with git and github nowadays :)
 Workflow is pretty straightforward:
 
 Clone the Repository:
-1.Open the provided GitHub link and click the "Fork" button on the upper-right of the web page to create your fork.
-2.Clone your forked repository using the --recurse-submodules argument to ensure submodules are also cloned.
+- 1.Open the provided GitHub link and click the "Fork" button on the upper-right of the web page to create your fork.
+- 2.Clone your forked repository using the --recurse-submodules argument to ensure submodules are also cloned.
 
 Set Up Development Environment:
-1.Configure your machine with the necessary development environment as specified in the project's documentation.
+- 1.Configure your machine with the necessary development environment as specified in the project's documentation.
 
 Make Changes:
-1.Implement the desired changes in your local clone.
+- 1.Implement the desired changes in your local clone.
 
 Run Tests:
-1.Ensure all tests pass before proceeding to the next step.
+- 1.Ensure all tests pass before proceeding to the next step.
 
 Update Changes File:
-1.Add a file to the CHANGES folder, naming it after the ticket or pull request number.
+- 1.Add a file to the CHANGES folder, naming it after the ticket or pull request number.
 
 Commit Changes:
-1.Commit your changes to your own aiohttp clone.
+- 1.Commit your changes to your own aiohttp clone.
 
 Create Pull Request:
-1.Make a pull request from the GitHub page of your clone against the master branch of the original repository.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
Dreamsorcerer, line 1: To me, this mostly feels more verbose without actually adding anything. @webknjaz Any thoughts?

CONTRIBUTING.rst
```diff
@@ -12,21 +12,30 @@ I hope everybody knows how to work with git and github nowadays :)
...
-  6. Commit changes to your own aiohttp clone
+Commit Changes:
+- 1.Commit your changes to your own aiohttp clone.
 
-  7. Make a pull request from the github page of your clone against the master branch
+Create Pull Request:
+- 1.Make a pull request from the GitHub page of your clone against the master branch of the original repository.
+Optional Backport:
 
-  8. Optionally make backport Pull Request(s) for landing a bug fix into released aiohttp versions.
```
Dreamsorcerer, line 29: This line should probably be changed though to better reflect current practices.

Maybe something like:
Once merged, backports may be created automatically by a bot. If the bot fails to do this, please follow the instructions from the bot to complete the backport PRs.

mdsahilnoob: Well, i will check that, thanks for the information!

webknjaz: That only works if somebody reviewing the PR puts the labels on. The contributors might still want to optionally offer manually-crafted backport PRs. (or semi-manually — we have a config for the CPython's `cherry-picker` tool!)

CONTRIBUTING.rst
```diff
@@ -12,21 +12,30 @@ I hope everybody knows how to work with git and github nowadays :)
 
 Workflow is pretty straightforward:
 
-  1. Clone the GitHub_ repo using the ``--recurse-submodules`` argument
+Clone the Repository:
+- 1.Open the provided GitHub link and click the "Fork" button on the upper-right of the web page to create your fork.
+- 2.Clone your forked repository using the --recurse-submodules argument to ensure submodules are also cloned.
```
webknjaz, line 17: This breaks the syntax highlighting.

CONTRIBUTING.rst
webknjaz: Yeah, agreed. We don't want an entire Git tutorial or things that are documented elsewhere to be copy-and-pasted there since this is an additional maintenance burden to keep copies in sync.

webknjaz: Also, there's a copy of this document in our Sphinx docs.

webknjaz: I typically keep something short like https://ansible-pylibssh.rtfd.io/en/latest/contributing/guidelines/, but in some cases, it makes sense to give more examples: https://github.com/python-attrs/attrs/blob/main/.github/CONTRIBUTING.md.

Still, I don't like trying to maintain the same thing somebody else does.

CONTRIBUTING.rst
Dreamsorcerer: Yeah, I'm just thinking it's usually safer to wait for the bot. A maintainer should add the label when they believe a backport is needed. Surely cherry picking won't work well if done before merging anyway, as they'd be picking commits from the branch, which won't end up on master (there will just be a squash commit instead).

webknjaz: Note that the bot experiences instability because of https://github.com/aio-libs/aiohttp/issues/4581, that I haven't really had time to tackle properly. I think it needs to be prioritized but there's a lot on my plate right now. A workaround is to re-apply the label (but for that, one must notice the failure first).

webknjaz:
This hasn't seen any progress or engagement in discussions for a while. Closing.

PR #8005, opened 2024-01-05
Implement happy eyeballs (RFC 8305) (#7954)

bdraco:
(cherry picked from commit c4ec3f130af3a53040b883aa2fab06fd215f9088)

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -1017,7 +1017,7 @@ class TCPConnector(BaseConnector):
                 raise
             raise client_error(req.connection_key, exc) from exc
 
-    async def _legacy_wrap_create_connection(
+    async def _wrap_existing_connection(
         self,
         *args: Any,
         req: ClientRequest,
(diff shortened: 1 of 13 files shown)


aiohttp/connector.py
```diff
@@ -1345,7 +1401,7 @@ async def _create_proxy_connection(
                 if not runtime_has_start_tls:
                     # HTTP proxy with support for upgrade to HTTPS
                     sslcontext = self._get_ssl_context(req)
-                    return await self._wrap_create_connection(
+                    return await self._wrap_existing_connection(
```
bdraco, line 1404: This path doesn't exist in 4.0 so I renamed the function to keep the code exactly as-is. I didn't want to risk changing it since there is no existing coverage for this path. Also we don't care about happyeyeballs in this case since the connection is already established so it likely doesn't matter anyways.

bdraco:
Will push this to production shortly to make sure I got all the conflicts done correctly again

bdraco:
Everything looks good after running for an hour. Is effectively the same code I've been running for weeks now though, just did the merge conflict again

Issue #8011, opened 2024-01-09
FileResponse + enable_compression + .gz will compress twice


messa:
### Describe the bug

When using `aiohttp.web.FileResponse` to return a file from disk, it is possible to use compression. As I understand from source code - I could not find documentation for FileResponse. There are two possible ways:

1. calling `response.enable_compression()` will make `FileResponse` automatically compress the file
2. when I call `FileResponse('some_file')` and `some_file.gz` exists, it will use the compressed version

Both ways work as expected.

But when both is appliend at the same time:

```python
response = FileResponse('some_file')
response.enable_compression()
return response
```

and _also_ `some_file.gz` exists, the response will be compressed twice - contents of `some_file.gz` is compressed via `deflate`.

### To Reproduce

```python
async def test_file_response_gz_with_enable_compression(aiohttp_client, tmp_path):

    async def file_handler(request):
        assert request.headers['Accept-Encoding'] == 'gzip, deflate'
        (tmp_path / 'file.gz').write_bytes(gzip.compress(b'Hello, compressed!'))
        response = FileResponse(tmp_path / 'file')
        response.enable_compression()
        return response

    app = Application()
    app.router.add_get('/file', file_handler)

    test_client = await aiohttp_client(app)
    resp = await test_client.get('/file')
    assert resp.status == 200
    assert resp.headers['Content-Type'] == 'application/octet-stream'
    assert resp.headers['Content-Encoding'] == 'deflate' # This should be gzip
    assert resp.headers['Transfer-Encoding'] == 'chunked'
    data = await resp.read()

    # This is how it should work:
    # assert data == b'Hello, compressed!'

    # But currently it works this way:
    assert gzip.decompress(data) == b'Hello, compressed!'
```

When using `enable_compression()` with non-compressed `file`, or using compressed `file.gz` without `enable_compression()`, it works as expected:

```python
async def test_file_response_gz_without_enable_compression(aiohttp_client, tmp_path):

    async def file_handler(request):
        assert request.headers['Accept-Encoding'] == 'gzip, deflate'
        (tmp_path / 'file.gz').write_bytes(gzip.compress(b'Hello, compressed!'))
        response = FileResponse(tmp_path / 'file')
        return response

    app = Application()
    app.router.add_get('/file', file_handler)

    test_client = await aiohttp_client(app)
    resp = await test_client.get('/file')
    assert resp.status == 200
    assert resp.headers['Content-Type'] == 'application/octet-stream'
[...]

PR #8012, opened 2024-01-09
Fix FileResponse doing blocking I/O in the event loop

bdraco:
## What do these changes do?

Fix `FileResponse` doing blocking I/O in the event loop

The call to `os.stat` was run in the executor but the call to `is_file()` was not. Additionally this resulted in two syscalls to `stat` since both `is_file` and `stat` will call `stat` under the hood. The code has been refactored to ensure there is only one `stat` syscall and it happens in the executor

No

## Related issue number

discovered while investigating #8011 but not does fix the issue reported

## Checklist

### DIFF
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -121,19 +121,31 @@ class FileResponse(StreamResponse):
         self.content_length = 0
         return await super().prepare(request)
 
-    async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
+    def _get_file_path_stat_and_gzip(
+        self, check_for_gzipped_file: bool
+    ) -> Tuple[pathlib.Path, os.stat_result, bool]:
+        """Return the file path, stat result, and gzip status.
+
+        This method should be called from a thread executor
+        since it calls os.stat which may block.
+        """
         filepath = self._path
-
-        gzip = False
-        if "gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, ""):
+        if check_for_gzipped_file:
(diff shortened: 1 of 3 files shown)


bdraco:
There is also a race here since nothing guarantees that the file can't change between the time we stat it and the time we open it

Fixing that is a lot harder though since we should `open` it , `fstat` it instead and keep it open.

bdraco:
local testing looks good

bdraco:
looks good on a second production system.

doing a `py-spy` to check the execution path

bdraco:
looks good

aiohttp/web_fileresponse.py
```diff
@@ -121,19 +121,31 @@ async def _precondition_failed(
...
+        This method should be called from a thread executor
+        since it calls os.stat which may block.
+        """
         filepath = self._path
-
-        gzip = False
-        if "gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, ""):
+        if check_for_gzipped_file:
             gzip_path = filepath.with_name(filepath.name + ".gz")
+            try:
```
Dreamsorcerer, line 135: Personally, I'd prefer a suppress() rather than having a pass statement.

Dreamsorcerer (approved):

aiohttp/web_fileresponse.py
bdraco: Usually I avoid `suppress` any time it might be performance sensitive since its much slower than a `try` pass

Dreamsorcerer: OK, never really considered there might be a performance difference..

PR #8022, opened 2024-01-11
Add runtime type check for `ClientSession` `timeout` param

### LINKED ISSUE #8021: ClientSession timeout AttributeError
### Describe the bug

When you create a client with `client.ClientSession(timeout=500)` does not raise or warn that timeout must be implemented as `ClientTimeout` but fails when sending a get request
Code lets you implement a client as `client.ClientSession(timeout='bad_arg')`

### To Reproduce

1. Implement a client with timeout of type `int` or `float` and send a get request 

```
import asyncio
import aiohttp

async def main():
    # No issues creating a client session
    aiohttp_client = aiohttp.ClientSession(timeout=100)

    try:
        # Fails sending GET request
        await aiohttp_client.get("https://example.com", )
    finally:
        # Close client session
        await aiohttp_client.close()

asyncio.run(main())
```

2. Implement a session with  timeout as `str` type  
```
aiohttp_client = aiohttp.ClientSession(timeout='bad_arg')
```

### Expected behavior

A `ValueError` has to be raised when using something other than `ClientTimeout` or `None`.

### Logs/tracebacks

```python-traceback
  File "/Users/voltaic/workspace/parsing_excersizes/async_requests/main.py", line 15, in <module>
    asyncio.run(main())
  File "/opt/homebrew/Cellar/python@3.12/3.12.1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 118, in run
  File "/opt/homebrew/Cellar/python@3.12/3.12.1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 684, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/voltaic/workspace/parsing_excersizes/async_requests/main.py", line 10, in main
    await aiohttp_client.get("https://example.com", )
  File "/Users/voltaic/Library/Application Support/hatch/env/virtual/async-requests/3bENKOFX/async-requests/lib/python3.12/site-packages/aiohttp/client.py", line 470, in _request
    self._loop, real_timeout.total, ceil_threshold=real_timeout.ceil_threshold
                ^^^^^^^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'total'
```

### Python Version

```console
Python 3.12.1
```

### aiohttp Version

```console
Name: aiohttp
Version: 3.9.1
Author:
Author-email:
License: Apache 2
Location: /Users/voltaic/Library/Application Support/hatch/env/virtual/async-requests/3bENKOFX/async-requests/lib/python3.12/site-packages
Required-by:
```

### multidict Version

```console
Name: multidict
Version: 6.0.4
Author: Andrew Svetlov
License: Apache 2
Location: /Users/voltaic/Library/Application Support/hatch/env/virtual/async-requests/3bENKOFX/async-requests/lib/python3.12/site-packages
Requires:
Required-by: aiohttp, yarl
```

### yarl Version

```console
Name: yarl
Version: 1.9.4
Author: Andrew Svetlov
License: Apache-2.0
Location: /Users/voltaic/Library/Application Support/hatch/env/virtual/async-requests/3bENKOFX/async-requests/lib/python3.12/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

macOS, Linux

### Related component

Client

### Additional context

_No response_

### Code of Conduct

igorvoltaic:
## What do these changes do?

Handle `ClientSession` `timeout` param type properly when creating a client.

A `ValueError` is raised when using something other than `ClientTimeout` in `timeout` param.

## Related issue number

Fixes #8021

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -269,15 +269,13 @@ class ClientSession:
         self._version = version
         self._json_serialize = json_serialize
         if timeout is sentinel or timeout is None:
-            self._timeout = DEFAULT_TIMEOUT
-        else:
-            if not isinstance(timeout, ClientTimeout):
-                raise ValueError(
-                    f"timeout parameter cannot be of {type(timeout)} type, "
-                    "please use 'timeout=ClientTimeout(...)'",
-                )
-            else:
-                self._timeout = timeout
+            timeout = DEFAULT_TIMEOUT
+        if not isinstance(timeout, ClientTimeout):
+            raise ValueError(
+                f"timeout parameter cannot be of {type(timeout)} type, "
+                "please use 'timeout=ClientTimeout(...)'",
+            )
+        self._timeout = timeout
         self._raise_for_status = raise_for_status
         self._auto_decompress = auto_decompress
         self._trust_env = trust_env
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -271,19 +271,10 @@ class ClientSession:
         if timeout is sentinel or timeout is None:
             self._timeout = DEFAULT_TIMEOUT
         else:
-            if not isinstance(timeout, (ClientTimeout, int, float)):
+            if not isinstance(timeout, ClientTimeout):
                 raise ValueError(
                     f"timeout parameter cannot be of {type(timeout)} type, "
                     "please use 'timeout=ClientTimeout(...)'",
-                )
-            if not isinstance(timeout, ClientTimeout):
-                self._timeout = ClientTimeout(total=timeout)
-                warnings.warn(
-                    "parameter 'timeout' of type 'float' or 'int' "
-                    "is deprecated, please use "
-                    "'timeout=ClientTimeout(total=...)'",
-                    DeprecationWarning,
-                    stacklevel=2,
                 )
             else:
                 self._timeout = timeout
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -271,7 +271,22 @@ class ClientSession:
         if timeout is sentinel or timeout is None:
             self._timeout = DEFAULT_TIMEOUT
         else:
-            self._timeout = timeout
+            if not isinstance(timeout, (ClientTimeout, int, float)):
+                raise ValueError(
+                    f"timeout parameter cannot be of {type(timeout)} type, "
+                    "please use 'timeout=ClientTimeout(...)'",
+                )
+            if not isinstance(timeout, ClientTimeout):
+                self._timeout = ClientTimeout(total=timeout)
+                warnings.warn(
+                    "parameter 'timeout' of type 'float' or 'int' "
+                    "is deprecated, please use "
+                    "'timeout=ClientTimeout(total=...)'",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
+            else:
+                self._timeout = timeout
         self._raise_for_status = raise_for_status
         self._auto_decompress = auto_decompress
         self._trust_env = trust_env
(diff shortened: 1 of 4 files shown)


aiohttp/client.py
```diff
@@ -271,7 +271,22 @@ def __init__(
...
+                )
+            if not isinstance(timeout, ClientTimeout):
+                self._timeout = ClientTimeout(total=timeout)
+                warnings.warn(
+                    "parameter 'timeout' of type 'float' or 'int' "
+                    "is deprecated, please use "
+                    "'timeout=ClientTimeout(total=...)'",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
Dreamsorcerer: I don't think we need a deprecation warning if it never worked before. You appear to be adding a feature and deprecating it at the same time.

tests/test_client_session.py
```diff
@@ -780,8 +780,9 @@ async def test_client_session_timeout_default_args(loop: Any) -> None:
 
 
 async def test_client_session_timeout_argument() -> None:
```
Dreamsorcerer: Although, this test shouldn't exist in that case...
Especially on master.

I think I'd just replace this test with one of your new ones.

aiohttp/client.py
igorvoltaic: Since this test is still here I'm afraid to break someone's code, yes it fails while executing `.get()` later, but might not be the case for other methods
https://github.com/aio-libs/aiohttp/blob/10ecad886fac40722f654b983a95d49e77f657ea/tests/test_client_session.py#L782-L785

tests/test_client_session.py
igorvoltaic: I do not think it is a good idea to introduce (semi-)breaking changes in a single PR, but agree docs suggest there should be only `ClientTimeout` or `None`. 

Pardon my hesitation, but have to ask again do I remove the possibility of using `float`/`int` entirely or leave it like this?

Dreamsorcerer: Everything goes through the request() method, so if it's not currently working, then it never worked.
There's backwards-compatibility with float already on 3.x, but appears to only be when passed in the request() method, rather than on the ClientSession...
But, it's clearly not working on the ClientSession, so there's no risk of breakage here. We don't want to introduce a deprecated feature.

igorvoltaic: removed deprecated feature

aiohttp/client.py
```diff
@@ -271,7 +271,13 @@ def __init__(
...
             self._timeout = DEFAULT_TIMEOUT
         else:
-            self._timeout = timeout
+            if not isinstance(timeout, ClientTimeout):
+                raise ValueError(
+                    f"timeout parameter cannot be of {type(timeout)} type, "
+                    "please use 'timeout=ClientTimeout(...)'",
+                )
+            else:
+                self._timeout = timeout
```
Dreamsorcerer: I feel like it might be simpler like:
```
        if timeout is sentinel or timeout is None:
            timeout = DEFAULT_TIMEOUT

        if not isinstance(timeout, ClientTimeout):
            raise ValueError(
                f"timeout parameter cannot be of {type(timeout)} type, "
                "please use 'timeout=ClientTimeout(...)'",
            )

        self._timeout = timeout
```

igorvoltaic: yep, you totally right ^_^

tests/test_client_session.py
Dreamsorcerer: Test needs to be removed though, as it will fail now. The new test replaces this one.

igorvoltaic: done

PR #8041, opened 2024-01-20
Close client connections when writer is unexpectedly cancelled

bdraco:
TODO: figure out how to test this

## What do these changes do?

If the writer is cancelled out while waiting for the connection to close or release, we now suppress the cancellation to prevent it from leaking upwards and force close the connection to ensure we do not reuse a connection which has an indeterminate state. 

https://github.com/aio-libs/aiohttp/discussions/8035#discussioncomment-8186816

## Checklist

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -709,6 +709,7 @@ class ClientResponse(HeadersMixin):
     _closed = True  # to allow __del__ for non-initialized properly response
     _released = False
     __writer = None
+    __writer_cancelled = False
 
     def __init__(
         self,
@@ -767,6 +768,7 @@ class ClientResponse(HeadersMixin):
         if self.__writer is not None:
             self.__writer.remove_done_callback(self.__reset_writer)
         self.__writer = writer
+        self.__writer_cancelled = False
         if writer is not None:
             writer.add_done_callback(self.__reset_writer)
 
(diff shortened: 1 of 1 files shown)


aiohttp/client_reqrep.py
Dreamsorcerer, line 1: One concern that occurred to me is what if this is part of a task that is being cancelled by the user? In that case we don't want to suppress.

Ideally, if we can detect if the cancellation is from our client timeout or similar, we'd suppress it, otherwise we'd reraise it. Or something along those lines...

Dreamsorcerer: Looks reasonable, but might also need something to handle ClientTimeout. Can't remember exactly how the logic works, but if we hit the timeout at this point, then we probably don't actually want to time out, as we already received the response before the timeout.

bdraco: I _think_ thats handled by https://github.com/aio-libs/aiohttp/blob/9e14ea19b5a48bb26797babc32202605066cb5f5/aiohttp/client_reqrep.py#L570

bdraco:
Tests are very light for this case so I'll have to work up something from scratch.  I'll do that once the issue is confirmed fixed by the op of the discussion

Dreamsorcerer:
I think this one is actually superseded by #9030.

bdraco:
closing in favor of #9030

PR #8049, opened 2024-01-21
Set cause for ClientPayloadError

Dreamsorcerer:
Should help improve some bug reports in future.

### DIFF
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -93,11 +93,11 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         if self._parser is not None:
             try:
                 uncompleted = self._parser.feed_eof()
-            except Exception:
+            except Exception as e:
                 if self._payload is not None:
-                    self._payload.set_exception(
-                        ClientPayloadError("Response payload is not completed")
-                    )
+                    exc = ClientPayloadError("Response payload is not completed")
+                    exc.__cause__ = e
+                    self._payload.set_exception(exc)
 
         if not self.is_eof():
             if isinstance(exc, OSError):


Dreamsorcerer:
We could really do with a way to also get the traceback added in StreamReader.set_exception(). Can't seem to figure it out at the moment, but ideally we'd grab the stack trace in that function and then add it the `exc.__cause__` somehow.

Dreamsorcerer:
Maybe at worst we can just add:
```
try:
    raise Exception()
except Exception as e:
    exc.__cause__ = e
```
Though feels like quite an awkward way to achieve it.

bdraco:
> Maybe at worst we can just add:
> 
> ```
> try:
>     raise Exception()
> except Exception as e:
>     exc.__cause__ = e
> ```
> 
> Though feels like quite an awkward way to achieve it.



> Maybe at worst we can just add:
> 
> ```
> try:
>     raise Exception()
> except Exception as e:
>     exc.__cause__ = e
> ```
> 
> Though feels like quite an awkward way to achieve it.

That seems awkward for sure.  `traceback.extract_tb` might work but its also really expensive.. `sys._getframe` is cheap but only works on cpython.

Setting `__cause__` is cheap and works everywhere.. its just a lot of copy pasta.

bdraco (approved):
This looks good

webknjaz:
I think, this one may need a change note.
Also, I seem to recall finding other places where it's unset, in that old issue. Have you looked into it?

Dreamsorcerer:
Not that I've seen, I just noticed a lack of information while debugging a couple of issues. The set_exception()s are the main other ones, but as described above it's a bit awkward to figure out how to put the traceback in. Best left to a separate PR if we do it.

webknjaz:
@Dreamsorcerer I was thinking about all the other places as mentioned @ https://github.com/aio-libs/aiohttp/issues/4581#issuecomment-1834283403. There's a `set_exception()` helper that isn't always used, it seems, and could be extended to be slightly more generic per my comment there.

Also, why didn't you reuse my patch @ https://github.com/aio-libs/aiohttp/issues/4581#issuecomment-1833977540? It's almost the same, but with better var names and the original exception repr backed into the message.

Dreamsorcerer:
> I was thinking about all the other places as mentioned @ [#4581 (comment)](https://github.com/aio-libs/aiohttp/issues/4581#issuecomment-1834283403).

Oh, I totally forgot about that conversation. I was just debugging something on my own yesterday and found that one missing. Doesn't look like you solved the .set_exception() yet though.

webknjaz:
> Doesn't look like you solved the .set_exception() yet though.

@Dreamsorcerer I ended up pushing #8089. I haven't tried it, though. And didn't add any tests. Plz check it for obvious mistakes.


> Setting `__cause__` is cheap and works everywhere.. its just a lot of copy pasta.

@bdraco my PR unifies that a bit. Interestingly, I've found the use of `__context__` in one place that I suspect may be problematic...

PR #8055, opened 2024-01-24
Make empty header names cause a 400 instead of an unhandled exception.

kenballus:
## What do these changes do?

This patch ensures that requests containing empty header names are rejected by the Python HTTP parser, instead of causing an unhandled exception that results in the closing of the connection without a response.

Shouldn't be!

## Checklist

- [X] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`names.

### DIFF
--- a/tests/test_http_parser.py
+++ b/tests/test_http_parser.py
@@ -251,6 +251,20 @@ def test_bad_chunked_c(loop: Any, protocol: Any) -> None:
         parser.feed_data(text)
 
 
+def test_empty_header_name_py(loop: Any, protocol: Any) -> None:
+    """Test that empty header names can't be used."""
+    parser = HttpRequestParserPy(
+        protocol,
+        loop,
+        2**16,
+        max_line_size=8190,
+        max_field_size=8190,
+    )
+    text = b"GET / HTTP/1.1\r\n:\r\n\r\n"
+    with pytest.raises(http_exceptions.InvalidHeader):
+        parser.feed_data(text)
+
+
 def test_whitespace_before_header(parser: Any) -> None:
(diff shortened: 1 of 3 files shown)


tests/test_http_parser.py
```diff
@@ -251,6 +251,20 @@ def test_bad_chunked_c(loop: Any, protocol: Any) -> None:
         parser.feed_data(text)
 
 
+def test_empty_header_name_py(loop: Any, protocol: Any) -> None:
+    """Test that empty header names can't be used."""
+    parser = HttpRequestParserPy(
```
webknjaz, line 256: Plz parametrize this to run against both parsers — C and Py should have feature parity.

kenballus: The C parser allows empty headers in the current release, but will begin disallowing them in the next release. See [this commit](https://github.com/nodejs/llhttp/commit/10ff94eb252e0e7cb792dcde6d40d0e46b794f8a). Until that commit makes it into a release, the two parsers will exhibit different behavior. I could add a C test and mark it xfail if you'd like?

webknjaz: Yes, `xfail` sounds reasonable. And make sure to add a `reason=` string with a link to the information regarding when it'll be okay to drop it. Could also make the xfail marker conditional on the llhttp version.

kenballus:
Closing this PR due to the existence of aiohttp-ghsa-8qpw-xqxj-h4r2 that also fixes this issue.

webknjaz:
@kenballus please, don't disclose the existence of private/embargoed advisories publicly.

PR #8063, opened 2024-01-26
Add server capability to check for Brotli compressed static files

### LINKED ISSUE #8062: Serve static Brotli compressed files

Currently only the _aiohttp_ client supports Brotli compression, and the _aiohttp_ server is missing such capability for even just serving Brotli files that are statically compressed by the user (i.e. a file with the `.br` extension exists).

For static requests, server should look for `.br` file extension in addition to `.gz`, and prefer Brotli over GZip if both exist.

I will submit a PR shortly for this.  I opened the issue as a formality.

None

### Related component

Server

### Additional context

#2518 is obviously a similar feature request, but this request focuses on just starting with precompressed Brotli capability, and then expanding to dynamic compression.

### Code of Conduct

steverep:
## What do these changes do?

Currently server only checks if static routes have a `.gz` extension and serves them with `gzip` encoding.  These changes do the same for `.br` files with `br` encoding.  Brotli is prioritized over gzip if both exist and are supported by the client, as it should almost always be a smaller content length.

I considered making a check for which is smaller if both exist, but figured it wouldn't be worth the extra file system call in the vast majority of cases (at least not for typical web formats).  Users should simply use gzip if it's smaller than Brotli for any file.

Only change would be if users are already including `.br` files, then obviously those would start being served, but IMO that's not "breaking" since the content should be the same.

## Related issue number

Fixes #8062 

## Checklist


aiohttp/web_response.py
```diff
@@ -53,7 +53,7 @@
     BaseClass = collections.abc.MutableMapping
 
 
-class ContentCoding(enum.Enum):
+class ContentCoding(enum.StrEnum):
```
bdraco: We still support older python, so I don't think this will work

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
...
+        # Fallback to the uncompressed file
+        return filepath, filepath.stat(), None
 
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
         loop = asyncio.get_event_loop()
-        check_for_gzipped_file = "gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, "")
-        filepath, st, gzip = await loop.run_in_executor(
-            None, self._get_file_path_stat_and_gzip, check_for_gzipped_file
+        accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
+        filepath, st, file_compression = await loop.run_in_executor(
```
bdraco: We should avoid generating an executor job if they didn't ask for compressed files.

edit: never mind we have to anyways to get the stat regardless.

tests/test_web_sendfile_functional.py
```diff
@@ -238,47 +243,60 @@ async def handler(request):
     app.router.add_get("/", handler)
     client = await aiohttp_client(app)
 
-    resp = await client.get("/")
+    resp = await client.get("/", headers={"Accept-Encoding": accept_encoding})
```
bdraco, line 276: We should probably keep this test as is and add a new one that sends the header so we cover both cases or params to do both here

bdraco:
I think this one should be safe for backport.

bdraco (dismissed):
Thanks for the PR. Please see comments above

aiohttp/web_fileresponse.py
steverep: That seems out of scope here.  This change doesn't add any new executor job - uses the one already being created.  

And from the method comment, it's created because of the file system call, which cannot be avoided.  You don't know if a compressed version exists or not unless you check.

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
...
-        """Return the file path, stat result, and gzip status.
+        self, accept_encoding: str
+    ) -> Tuple[pathlib.Path, os.stat_result, Optional[ContentCoding | Literal["br"]]]:
+        """Return the file path, stat result, and possible compression type.
 
         This method should be called from a thread executor
         since it calls os.stat which may block.
         """
         filepath = self._path
-        if check_for_gzipped_file:
```
bdraco, line 133: Let’s keep the guard here if the accept encoding isn’t set. It won’t make much difference but we could also avoid the lower on the header if it’s not set.

aiohttp/web_response.py
steverep: Okay not sure I need this anymore, but I would need to dumb down the dict type otherwise.  I'll check.

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
         return await super().prepare(request)
 
     def _get_file_path_stat_and_gzip(
-        self, check_for_gzipped_file: bool
-    ) -> Tuple[pathlib.Path, os.stat_result, bool]:
-        """Return the file path, stat result, and gzip status.
+        self, accept_encoding: str
```
bdraco, line 136:
```suggestion
        self, accept_encoding: Optional[str]
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
...
+
+        # Fallback to the uncompressed file
+        return filepath, filepath.stat(), None
 
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
         loop = asyncio.get_event_loop()
-        check_for_gzipped_file = "gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, "")
-        filepath, st, gzip = await loop.run_in_executor(
-            None, self._get_file_path_stat_and_gzip, check_for_gzipped_file
+        accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
```
bdraco, line 162:
```suggestion
        accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING)
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
...
-        if check_for_gzipped_file:
-            gzip_path = filepath.with_name(filepath.name + ".gz")
-            try:
-                return gzip_path, gzip_path.stat(), True
-            except OSError:
-                # Fall through and try the non-gzipped file
-                pass
-
-        return filepath, filepath.stat(), False
+        for encoding, extension in ENCODING_EXTENSION.items():
```
bdraco:
```suggestion
        if accept_encoding:
           lower_accept_encoding = accept_encoding.lower()
           for encoding, extension in ENCODING_EXTENSION.items():
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +130,31 @@ async def _precondition_failed(
...
+            if encoding in accept_encoding:
```
bdraco:
```suggestion
            if encoding in lower_accept_encoding:
```

aiohttp/web_fileresponse.py
bdraco: indent needs to be adjusted as well below

tests/test_web_sendfile_functional.py
steverep: Would adding `headers=None` or just `accept_encoding=None` as another test do the trick?

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +38,13 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+# File extensions for IANA encodings that will be checked in the order defined.
+# Brotli is specified as a literal until compression is supported.
+ENCODING_EXTENSION: Final[dict[ContentCoding | Literal["br"], str]] = {
+    "br": ".br",
+    ContentCoding.gzip: ".gz",
```
bdraco:
```suggestion
    "gzip": ".gz",
```

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +38,13 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+# File extensions for IANA encodings that will be checked in the order defined.
+# Brotli is specified as a literal until compression is supported.
+ENCODING_EXTENSION: Final[dict[ContentCoding | Literal["br"], str]] = {
```
bdraco:
```suggestion
ENCODING_EXTENSION: Final[Dict[str, str]] = {
```

We still need to use legacy typing for older python

tests/test_web_sendfile_functional.py
bdraco: `headers=None` should preserve the existing test behavior / coverage

aiohttp/web_response.py
bdraco:
```suggestion
class ContentCoding(enum.Enum):
```

Dreamsorcerer: If you think the code is better with a newer Python feature, then ideally add 2 versions with a `if sys.version_info >= (3, 10)`, or if too complicated, just leave a `# TODO (PY310): ...` comment. Then we'll drop the older code as we upgrade in future.

Dreamsorcerer:
@bdraco I'd suggest this should only go to 3.10 backport. We try to stick to bug fixes in the patch releases.

aiohttp/web_response.py
steverep: Personally, I'm a stickler for strong typing, so I would prefer to keep it instead of dumbing things down.  And I see spots in the code where it could be used more, especially if I try to follow this up with dynamic Brotli compression.

If you all are okay with it, I'd do an if-block just around the import for 3.11, and use either [`strenum`](https://pypi.org/project/StrEnum/#description) or [`backports.stremum`](https://pypi.org/project/backports.strenum/).

If so, a hint on how to easily add the dep in the right spot would be appreciated.  My python is a bit rusty.

bdraco: With the enum changes in python 3.11 I would prefer to keep this as is  since it introduces additional risk.

I have had hard to find problems with Enum related changes in cpython 3.11 and would prefer to be cautious here.

Once we drop support for python less than 3.11 we can come back and rework this.

steverep: Okay I removed the `StrEnum` while keeping the literal typing and use of the existing `Enum`.  Should pass tests now.

bdraco: ~~`Literal` is unfortunately not available on all python versions we currently support.~~

Never mind: its actually `dict` that is the problem here

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +38,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+# File extensions for IANA encodings that will be checked in the order defined.
+# TODO(py311): Replace Literal with ContentCoding as StrEnum and remove .value
+ENCODING_EXTENSION: Final[dict[Literal["br", "gzip"], str]] = {
```
bdraco:
```suggestion
ENCODING_EXTENSION: Final[Dict[str, str]] = {
```

I'm not sure we need `Literal` here as it would mean we would have every possible value that gets added to this dict in the type.

aiohttp/web_response.py
steverep: Yeah I was going to say I checked the 3.8 docs first... need to put my kids to bed then I'll finish this up.

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +37,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+# File extensions for IANA encodings that will be checked in the order defined.
+# TODO(py311): Use StrEnum conversion of ContentCoding for key type and remove .value
+ENCODING_EXTENSION: Final[dict[str, str]] = {
```
bdraco:
```suggestion
ENCODING_EXTENSION: Final[Dict[str, str]] = {
```

steverep: Yeah I just educated myself on the historical differences between typing.* and built in `dict`, `list`, etc.  Thanks.

aiohttp/web_response.py
steverep: Finally fixed - I think 🤞🏻

tests/test_web_sendfile_functional.py
steverep: Done

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +38,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+# File extensions for IANA encodings that will be checked in the order defined.
```
bdraco: Since we are offering a guarantee that they are checked in order, we should validate that in a test as well.

steverep: The tests already do it, otherwise they wouldn't pick up the `hello.txt.br` at all

aiohttp/web_fileresponse.py
steverep: I'm not following - there is no check for no accept-encoding header in the current code.  The logic here I believe is identical to the current code, just a loop over 2 ifs instead of one.

Regarding no `accept-encoding` header in general though:
1. No browser/client does that, and it's a bad idea as it indicates per spec that any coding is acceptable
2. The current code is (arguably but not really) not spec compliant in that sense - it's always sending uncompressed which is not best for the server if the client says anything goes (and it probably has never come up as an issue because of my previous point).  The same goes for dynamic compression here: https://github.com/steverep/aiohttp/blob/b0d58f7c6164bff5ca59fd3e97cfeb652eac3ef5/aiohttp/web_response.py#L330.

So, I would say a follow up could be to adjust that trivial compliance issue, and then you could save the `lower()` call if desired.  However, I would recommend not doing it as you suggested below and instead keeping it within the async code.

steverep:
I have no clue why the 3.8 test is not decompressing the Brotli file correctly.  Is there a way to force it to run the full matrix of python versions and OS to help narrow it down?

aiohttp/web_fileresponse.py
bdraco: >  there is no check for no accept-encoding header in the current code. 

`check_for_gzipped_file` previously meant `"gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, "")` so it would not enter this block if there was no `ACCEPT_ENCODING` header

bdraco: Here is the full suggestion https://github.com/aio-libs/aiohttp/commit/6409638c9e83c8bb7efeb86da97228919df471da

bdraco:
> I have no clue why the 3.8 test is not decompressing the Brotli file correctly. Is there a way to force it to run the full matrix of python versions and OS to help narrow it down?

You could open a PR to your repo with versions you don't want to run removed from https://github.com/aio-libs/aiohttp/blob/a379e6344432d5c033f78c2733fe69659e3cff50/.github/workflows/ci-cd.yml#L144

You could also remove `fail-fast: true`
https://github.com/aio-libs/aiohttp/blob/a379e6344432d5c033f78c2733fe69659e3cff50/.github/workflows/ci-cd.yml#L184

aiohttp/web_fileresponse.py
bdraco: > 2\. The current code is (arguably but not really) not spec compliant in that sense

I think you are referring to:
> https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3
...
> If no Accept-Encoding header field is in the request, any content coding is considered acceptable by the user agent.

While I agree the current code is not compliant with the RFC, I think if we changed the implementation to work this way, we would break clients that can't handle `br` encoding that fails to send an `Accept-Encoding` header.  

> No browser/client does that, and it's a bad idea as it indicates per spec that any coding is acceptable

I've seen out-of-spec clients not sending this header. Considering we are not in RFC compliance here, we can't safely make assumptions that clients will comply in this case.

> So, I would say a follow up could be to adjust that trivial compliance issue

While it's RFC correct to break them, it amounts to a breaking change, which we may get some fall-out from. We may want to avoid making that change as there is a practical risk we would need to revert it for a seemingly minor benefit. I'll leave the discussion to determine if it is worth the risk for a future PR.

tests/test_web_sendfile_functional.py
```diff
@@ -238,47 +246,64 @@ async def handler(request):
...
         return resp
 
     app = web.Application()
     app.router.add_get("/", handler)
     client = await aiohttp_client(app)
 
-    resp = await client.get("/")
+    resp = await client.get(
+        "/", headers={"Accept-Encoding": accept_encoding} if accept_encoding else None
+    )
```
bdraco: Please avoid branching in tests

Alternative suggestion https://github.com/bdraco/aiohttp/commit/f0fb424d8b151f715243247483e42a7e99d3cd44

bdraco: A test for an empty `Accept-Encoding` header and `*`  as well would be good since https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3 shows them as potential payloads

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +131,31 @@ async def _precondition_failed(
...
+
+        # Fallback to the uncompressed file
+        return filepath, filepath.stat(), None
 
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
         loop = asyncio.get_event_loop()
-        check_for_gzipped_file = "gzip" in request.headers.get(hdrs.ACCEPT_ENCODING, "")
-        filepath, st, gzip = await loop.run_in_executor(
-            None, self._get_file_path_stat_and_gzip, check_for_gzipped_file
+        accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
```
bdraco, line 162: https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1
> All content codings are case-insensitive and ought to be registered within the "HTTP Content Coding Registry", as described in [Section 16.6](https://www.rfc-editor.org/rfc/rfc9110#content.coding.extensibility)

It looks like this is a bugfix to make the check "case-insensitive" and comply with the above-quoted section of the RFC

For this case, we should do a separate bug fix PR to make this so we can backport the `lower()` change to 3.9.x, along with some tests to make sure various cases work.

aiohttp/web_fileresponse.py
bdraco: https://www.rfc-editor.org/rfc/rfc9110#name-content-codings

> All content codings are case-insensitive and ought to be registered within the "HTTP Content Coding Registry", as described in [Section 16.6](https://www.rfc-editor.org/rfc/rfc9110#content.coding.extensibility)

I missed that there is also a bug fix for RFC compliance to implement case-insensitivity here. For the typical case, we need to `lower()` it anyway, so this doesn't likely doesn't save anything, and it's fine how you have it.

Given the above, I've added a suggestion to where `.lower()` is implemented instead: https://github.com/aio-libs/aiohttp/pull/8063#discussion_r1468444690

aiohttp/web_fileresponse.py
bdraco: > All content codings are case-insensitive and ought to be registered within the "HTTP Content Coding Registry", as described in [Section 16.6](https://www.rfc-editor.org/rfc/rfc9110#content.coding.extensibility)

This looks like this should be a separate bug fix PR so we could backport the `lower()` change to 3.9.x

Dreamsorcerer: If we can include links to the relevant part of the spec where code is specifically following the spec (as I've done throughout http_parser.py, for example), that would help maintainability. You can get a link to a specific paragraph too.

steverep: > It looks like this is a bugfix to make the check "case-insensitive" and comply with the above-quoted section of the RFC
> 
> For this case, we should do a separate bug fix PR to make this so we can backport the `lower()` change to 3.9.x, along with some tests to make sure various cases work.

Yes, I essentially included that as a fix, and to be consistent with the equivalent check for dynamic compression.  I concur it should be a separate PR though so I'll remove it here.

bdraco: It would be good to merge the separate PR first since this one is likely to backport cleanly to 3.10 from `master` if we rebase it on top of that change

steverep:
The test failure was simply that Brotli was not added to the encodings map in `mimetypes` until 3.9, so it has to be added manually (https://github.com/python/cpython/issues/76202).

bdraco:
> The test failure was simply that Brotli was not added to the encodings map in `mimetypes` until 3.9, so it has to be added manually (https://github.com/python/cpython/issues/76202).

Thanks for referencing the cpython issue 👍

tests/hello.txt.br
webknjaz, line 1: Instead of adding more blobs to Git, please make use of the `tmp_path` fixture and generate the test file dynamically.

steverep: It's only 19 bytes.  The code to generate it on the fly would certainly add more plus more code to keep it parametrized with the current gzip tests.

webknjaz: Perhaps, but it'd keep the repo cleaner and the way it's generated would be documented. In any case, Git is optimized for text and adding/removing blobs has a price.

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +39,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+if sys.version_info < (3, 9):
```
webknjaz, line 43: Plz move this to a constant in `helpers.py`. Could call it `PY_OLDER_THAN_39`.

tests/test_web_sendfile_functional.py
```diff
@@ -224,9 +224,17 @@ async def handler(request):
...
+    [(None, "br"), ("gzip, deflate", "gzip")],
+)
 async def test_static_file_custom_content_type_compress(
-    aiohttp_client: Any, sender: Any
+    aiohttp_client: Any,
+    sender: Any,
+    accept_encoding: Optional[str],
+    expect_encoding: str,
 ):
+    """Test static compressed files are returned with expected content type and encoding"""
```
webknjaz:
```suggestion
    """Test static compressed files are returned with expected content type and encoding."""
```

tests/test_web_sendfile_functional.py
webknjaz: Wanted to say the same. Logic in tests makes them more fragile and harder to follow.

CHANGES/8062.feature
```diff
@@ -0,0 +1 @@
+Added server capability to check for static files with Brotli compression via a ``.br`` extension -- by :user:`steverep`
```
webknjaz: Let's use a full stop at the end of the sentence.
```suggestion
Added server capability to check for static files with Brotli compression via a ``.br`` extension -- by :user:`steverep`.
```
I recently updated the change note writing style guidelines to make them more clear: https://docs.aiohttp.org/en/stable/contributing.html#adding-change-notes-with-your-prs.

aiohttp/web_fileresponse.py
webknjaz: Alternatively, have an inverted name
```suggestion
if not PY_39_PLUS:
```

Dreamsorcerer: I'd prefer if we didn't. That often results in more work for me when dropping versions, and often doesn't play well with mypy (I suspect I may have removed a previous PY_39 constant or similar previously).

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
         return await super().prepare(request)
 
     def _get_file_path_stat_and_gzip(
-        self, check_for_gzipped_file: bool
-    ) -> Tuple[pathlib.Path, os.stat_result, bool]:
-        """Return the file path, stat result, and gzip status.
+        self, accept_encoding: str
+    ) -> Tuple[pathlib.Path, os.stat_result, Optional[str]]:
```
webknjaz, line 137: `Optional` is considered confusing, this name doesn't relay the semantics well. There's even a linter forbidding it (it's called `no-optional`). The alternative is to explicitly say that it's one of two types:
```suggestion
    ) -> Tuple[pathlib.Path, os.stat_result, Union[str, None]]:
```
(for py3.10+ it'd be `str | None`, of course)

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
...
+        """Return the file path, stat result, and possible compression type.
 
         This method should be called from a thread executor
         since it calls os.stat which may block.
         """
```
webknjaz: Let's make the title shorter and not imply that `None` is an impossible type. Instead, clarify that `None` is a sentinel, semantically referring to an uncompressed file:
```suggestion
        """Return the file path, stat result, and compression type.

        If no compressed copy of the file is detected, in that case,
        the compression is represented as :py:data:`None`.

        This method should be called from a thread executor
        since it calls os.stat which may block.
        """
```

aiohttp/web_fileresponse.py
Dreamsorcerer: I wouldn't worry about these, they'll get removed in a year or so with pyupgrade regardless.

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
...
-
-        return filepath, filepath.stat(), False
+        for extension, encoding in ENCODING_EXTENSIONS.items():
+            if encoding in accept_encoding:
+                compressed_path = filepath.with_name(filepath.name + extension)
+                try:
+                    return compressed_path, compressed_path.stat(), encoding
+                except OSError:
+                    # Try the next extension
+                    pass
```
webknjaz: Could you use `contextlib.suppress()` for clarity, instead?
```suggestion
                with suppress(OSError):
                    return compressed_path, compressed_path.stat(), encoding
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
...
-            try:
-                return gzip_path, gzip_path.stat(), True
-            except OSError:
-                # Fall through and try the non-gzipped file
-                pass
-
-        return filepath, filepath.stat(), False
+        for extension, encoding in ENCODING_EXTENSIONS.items():
+            if encoding in accept_encoding:
+                compressed_path = filepath.with_name(filepath.name + extension)
```
webknjaz: I think that using `with_suffix()` would relay the underlying intent better:
```suggestion
                compressed_path = filepath.with_suffix(filepath.suffix + extension)
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
...
-        if check_for_gzipped_file:
-            gzip_path = filepath.with_name(filepath.name + ".gz")
-            try:
-                return gzip_path, gzip_path.stat(), True
-            except OSError:
-                # Fall through and try the non-gzipped file
-                pass
-
-        return filepath, filepath.stat(), False
+        for extension, encoding in ENCODING_EXTENSIONS.items():
```
webknjaz: May I suggest a naming improvement?
```suggestion
        for file_extension, payload_encoding in ENCODING_EXTENSIONS.items():
```

aiohttp/web_fileresponse.py
```diff
@@ -122,29 +132,31 @@ async def _precondition_failed(
...
+            if encoding in accept_encoding:
```
webknjaz: Could you change this to a guard-style check? This will allow to dedent the rest of the loop instructions. Less nesting is easier to follow from the readability/maintainability perspective.
```suggestion
            if encoding not in accept_encoding:
                continue
```

tests/test_web_sendfile_functional.py
steverep: > Please avoid branching in tests
> 
> Alternative suggestion [bdraco@f0fb424](https://github.com/bdraco/aiohttp/commit/f0fb424d8b151f715243247483e42a7e99d3cd44)

I understand - I just didn't view it that way since `headers={"Accept-Encoding": None}` fails as it probably should, so `None` is just a signal to remove the header.

Anyway, thinking more about this, I would instead recommend to revert the [commit to test with the default client header](https://github.com/aio-libs/aiohttp/commit/b0d58f7c6164bff5ca59fd3e97cfeb652eac3ef5).  IMHO, your version reduces the readability and detracts from the point of the test, which is to test the **server** response to **known** `Accept-Encoding` headers.  It shouldn't care about the `aiohttp` client default.  For example:

- If the client changed its default to "gzip, deflate", the test would fail even though it shouldn't.
- If the client changed its default to "br, deflate", the test is weakened because the expectation of `.br` over `.gz` is not tested.

If it doesn't already exist somewhere in the suite, there should be a client test that when no headers are specified, assert that the expected default headers are transmitted to the server in the final request.


> A test for an empty `Accept-Encoding` header and `*` as well would be good since https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3 shows them as potential payloads

Completely agree, but I feel it's flirting with the boundary of scope creep here.  We've already established a few issues with the spec compliance of interpreting the `Accept-Encoding` header, so adding such tests should wait for a correct parsing and also alter the tests when `enable_compression` is used, which is purposefully left out here.  I just wrote up https://github.com/aio-libs/aiohttp/issues/8104 to cover this.

bdraco: > Anyway, thinking more about this, I would instead recommend to revert the [commit to test with the default client header](https://github.com/aio-libs/aiohttp/commit/b0d58f7c6164bff5ca59fd3e97cfeb652eac3ef5)

I think thats fine. The goal of the suggestion was to preserve what you had without branching in the test.

bdraco: > Completely agree, but I feel it's flirting with the boundary of scope creep here. We've already established a few issues with the spec compliance of interpreting the `Accept-Encoding` header, so adding such tests should wait for a correct parsing and also alter the tests when `enable_compression` is used, which is purposefully left out here. I just wrote up #8104 to cover this.

Its fine to do it in another PR. I'd like to solve it before merging this one to avoid conflicts since we can backport fixes to 3.9 but this PR will only go back to 3.10 since its a new feature.

aiohttp/web_fileresponse.py
steverep: Took your suggestion, but changed the wording as yours is not technically correct.

tests/test_web_sendfile_functional.py
steverep: > Its fine to do it in another PR. I'd like to solve it before merging this one to avoid conflicts since we can backport fixes to 3.9 but this PR will only go back to 3.10 since its a new feature.

Resolving the branching issue here.  Let's discuss backporting on #8104.

aiohttp/web_fileresponse.py
webknjaz: I don't see any technical difference. You're referring to an implementation detail (something happens if something is returned, internally) and I'm trying to assign a high-level to a sentinel constant. But alright, it's better than before.

webknjaz: Ah, I see. You're also replacing compression with encoding. That's fair.

aiohttp/web_fileresponse.py
```diff
@@ -121,30 +132,33 @@ async def _precondition_failed(
...
-    def _get_file_path_stat_and_gzip(
-        self, check_for_gzipped_file: bool
-    ) -> Tuple[pathlib.Path, os.stat_result, bool]:
-        """Return the file path, stat result, and gzip status.
+    def _get_file_path_stat_encoding(
+        self, accept_encoding: str
+    ) -> Tuple[pathlib.Path, os.stat_result, Optional[str]]:
+        """Return the file path, stat result, and encoding.
+
+        If an uncompressed file is returned, the encoding is set to ``None``.
```
webknjaz: Looks like you lost the role from my suggestion:
```suggestion
        If an uncompressed file is returned, the encoding
        is set to :py:data:`None`.
```
This way, it can be linked to CPython docs, in the Sphinx build.

tests/hello.txt.br
webknjaz, line 1: Looks like the file changed and the original request got lost. Re-linking it here for visibility: https://github.com/aio-libs/aiohttp/pull/8063#discussion_r1468971835.

tests/test_web_sendfile_functional.py
```diff
@@ -288,13 +308,12 @@ async def handler(request):
...
-    assert 200 == resp.status
-    body = await resp.read()
-    assert b"hello aiohttp\n" == body
-    ct = resp.headers["CONTENT-TYPE"]
-    assert "text/plain" == ct
-    encoding = resp.headers["CONTENT-ENCODING"]
-    assert "gzip" == encoding
+    assert resp.status == 200
+    assert (
+        resp.headers["Content-Type"],
```
webknjaz: Please, make this multiple assertions, each clearly making one simple check. Complex checks are hard to parse in the test output and in code.

tests/test_web_sendfile_functional.py
```diff
@@ -288,13 +308,12 @@ async def handler(request):
     client = await aiohttp_client(app)
 
     resp = await client.get("/")
-    assert 200 == resp.status
```
webknjaz: Why did you invert the operands? Please, revert. Unnecessary changes only create complications with backporting.

tests/test_web_sendfile_functional.py
```diff
@@ -238,47 +243,62 @@ async def handler(request):
...
+@pytest.mark.parametrize(
+    ("extension", "expect_encoding"), [(".gz", "gzip"), (".br", "br")]
+)
 async def test_static_file_with_content_encoding(
-    aiohttp_client: Any, sender: Any
+    aiohttp_client: Any, sender: Any, extension: str, expect_encoding: str
 ) -> None:
-    filepath = pathlib.Path(__file__).parent / "hello.txt.gz"
+    """Test requesting of static compressed files returns correct content type and encoding."""
+    filepath = pathlib.Path(__file__).parent / ("hello.txt" + extension)
```
webknjaz: This is probably better, than concat.
```suggestion
    filepath = pathlib.Path(__file__).parent / f"hello.txt{extension}"
```

tests/test_web_sendfile_functional.py
```diff
@@ -238,47 +243,62 @@ async def handler(request):
...
+    ("accept_encoding", "expect_encoding"),
+    [("gzip, deflate", "gzip"), ("gzip, deflate, br", "br")],
+)
+@pytest.mark.parametrize("forced_compression", [None, web.ContentCoding.gzip])
+async def test_static_file_with_encoding_and_enable_compression(
+    aiohttp_client: Any,
+    sender: Any,
+    accept_encoding: str,
+    expect_encoding: str,
+    forced_compression: Optional[web.ContentCoding],
```
webknjaz, line 297: Let's avoid `Optional`
```suggestion
    forced_compression: "web.ContentCoding | None",
```

webknjaz (changes requested):
Blockers so far:
* [ ] new binary blob in Git
* [ ] formatting changes in tests

Please, deal with this to making this mergeable.

aiohttp/web_fileresponse.py
steverep: You wrote:

> If no compressed copy of the file is detected, ...

Whether or not a compressed file exists is not the only condition; its encoding also has to be supported and acceptable by the client.

tests/hello.txt.br
steverep: The file did not change.  Your original comment is still unresolved in the conversation view of the PR.

tests/test_web_sendfile_functional.py
steverep: I took the advice of @Dreamsorcerer not to worry about these, and also from the rest of the code base.  There are 561 uses of `Optional[]`, and only 14 `Union[]` with `None`.

If this is that important, perhaps you could do a PR with a search and replace and enable the lint rule.

tests/test_web_sendfile_functional.py
steverep: Why is that better?

tests/test_web_sendfile_functional.py
steverep: Because different tests had them in a different order for identical assertions, which was confusing.  I altered one and copied to the others.  I was debugging with all 3 tests.

tests/test_web_sendfile_functional.py
steverep: Sorry, I disagree here, and think `pytest` does a good job of comparing sequences item by item.

I made that change specifically to aid me in debugging the issue with 3.8, and it made the issue very clear.  Asserting one at a time makes sense if a previous one failing guarantees the rest will fail (e.g. if the response status is not okay, then the body and headers won't exist).  But the other 3 were wrong and asserting one at a time meant I could only know one value at a time (the first one), which isn't very useful.  It was originally asserting the body first, which is probably the last one that would go wrong.  I shouldn't have to change a test multiple times to get a clear picture of the results in the log.

tests/test_web_sendfile_functional.py
webknjaz: String concatenation is generally discouraged. It's slower, for example. A non-nested f-string is a bit easier to parse for humans too.

tests/test_web_sendfile_functional.py
webknjaz: Okay, fair. Resolved.

tests/test_web_sendfile_functional.py
webknjaz: Alright. This still needs to be reverted so the diff is smaller and is less prone to merge conflicts on cherry-picks.

tests/test_web_sendfile_functional.py
webknjaz: That's understandable. But still, complex assertions of semantically different values are discouraged. Pytest can compare them but the test run output is a mess. Plus it's super ugly in code. A simple comparison mustn't be multiline, because it increases cognitive complexity for no good reason.

tests/hello.txt.br
webknjaz: It's marked as outdated, which happens when GitHub things that the file changed since the thread was started. That's the only indication GH gives in PRs, which is another reason not to have blobs — it's not possible to verify in the PR view in browser.

tests/test_web_sendfile_functional.py
steverep: > That's understandable. But still, complex assertions of semantically different values are discouraged.

Can you point out somewhere that is documented for my reading and edification?  I did some `pytest` researching before I made that change and found recommendations to organize assertions similarly to deal with the issue, so I'd like to read the opposing views.

> Pytest can compare them but the test run output is a mess.

Here's the output that made the problem clear to me: https://github.com/aio-libs/aiohttp/actions/runs/7683502126/job/20939089887

What do you feel is a mess there?

> Plus it's super ugly in code. A simple comparison mustn't be multiline, because it increases cognitive complexity for no good reason.

It seems not much different than any other conditional that gets formatted to multiple lines, for which I'm sure there are countless occurrences?  And are you saying the reason I just gave is not a good one?  Couldn't I just argue it's less complex with less assertion steps: (1) response is okay, (2) actual response = expected response?

I'm not trying to be difficult here - I'm just genuinely perplexed by your request.  I don't think I've ever been asked to make something objectively less useful in order to purely make the code subjectively prettier.

steverep: In any case, in the interest of avoiding an infinite review loop here, I made the change, but put them in an order that makes some sense: content encoding, then type, and finally body.  I hope that resolves this.

tests/hello.txt.br
steverep: > Perhaps, but it'd keep the repo cleaner and the way it's generated would be documented.

There's not much to document - Brotli has a well-defined spec just like gzip.  I just used the `brotli` binary on the command line. 🤷🏻‍♂️ 


> In any case, Git is optimized for text and adding/removing blobs has a price.

Even for a 19 B file 200 times smaller than a typical disk cluster?  I guess that might be fun to devise an experiment to prove... 🧪 

So again I was just following the existing test models for gzip when I created it.  But if you think it's worth holding up this long overdue feature, that's your call.  I've already spent way more time on this PR than I expected.  I was hoping to avoid digging into the 2 different Brotli packages used here, reading more pytest and python docs as I'm quite rusty, etc., at least until after this was merged and I decided to implement it for server dynamic compression.

Bottom line - waiting for me to do this might take a bit as I need to attend to other things.

aiohttp/web_fileresponse.py
steverep: Opened #8127 to resolve this.

bdraco: Thanks

tests/hello.txt.br
steverep: I looked at creating the Brotli file dynamically today.  It pretty much forces also doing the same for the gzip file without it resulting in an absolute mess.  Please merge #8136 first.

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +40,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+if sys.version_info < (3, 9):
+    mimetypes.encodings_map[".br"] = "br"
+
+# File extension to IANA encodings map that will be checked in the order defined.
+ENCODING_EXTENSIONS: Final[Dict[str, str]] = {
+    ext: mimetypes.encodings_map[ext] for ext in (".br", ".gz")
+}
```
Dreamsorcerer: We try to avoid mutable globals:
```suggestion
ENCODING_EXTENSIONS: Final[MappingProxyType[str, str]] = MappingProxyType({
    ext: mimetypes.encodings_map[ext] for ext in (".br", ".gz")
})
```

aiohttp/web_fileresponse.py
```diff
@@ -37,6 +40,14 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+if sys.version_info < (3, 9):
+    mimetypes.encodings_map[".br"] = "br"
+
+# File extension to IANA encodings map that will be checked in the order defined.
+ENCODING_EXTENSIONS: Final[Dict[str, str]] = {
+    ext: mimetypes.encodings_map[ext] for ext in (".br", ".gz")
```
Dreamsorcerer: Also, is there a reason we restrict it to .br and .gz? I wonder if it doesn't make sense to just open it up to more allowed formats?

Dreamsorcerer (approved):

aiohttp/web_fileresponse.py
steverep: > Also, is there a reason we restrict it to .br and .gz? I wonder if it doesn't make sense to just open it up to more allowed formats?

Each one will add an extra file system call, so I'd be cautious.  Others included in the map are `bzip2`, `xz`, and `compress`.  Of those, `compress` is the only one included in the specs, but it's long been overshadowed by the others.

bdraco: I think its hard to justify others given brotli's numbers and wider availability in most browsers https://github.com/powturbo/TurboBench/issues/43

Dreamsorcerer: Fair enough. I guess if we're pre-compressing the files, then it suggests that brotli is the only sensible choice. And I assume we don't deal with on-the-fly compression (which would presumably be handled by a reverse proxy).

bdraco (approved):
LGTM, https://github.com/aio-libs/aiohttp/pull/8136 should merge first so it can be backported

tests/test_web_sendfile_functional.py
webknjaz: > What do you feel is a mess there?

That one is short enough, but it would hurt consistency if we were to use different styles depending on each output example, which might even change over time.

webknjaz: > Can you point out somewhere that is documented for my reading and edification? I did some `pytest` researching before I made that change and found recommendations to organize assertions similarly to deal with the issue, so I'd like to read the opposing views.

Unfortunately, I don't have anything like that. Perhaps it's tribal knowledge in some sense. But I did see complex chunks of output at some point, that would be more complicated to understand. Also note, that people's perception may differ due to their experience and tooling used, and dev setup. And chasing inclusivity, I'm trying to use practices that would lean towards accessibility. For example, I use far larger fonts than many people — it helps to take some strain off of my eyes. This has a side effect of rendering weirdly and generating more line breaks, especially in various split views like diff pages. This informs a choice of shorter lines in the source code. But it is also something that sometimes has a similar effect in log outputs.

This is one of my main motivations to reduce cognitive complexity in text, wherever I see it. I hope this provides you with a bit more insight.

webknjaz: > > Plus it's super ugly in code. A simple comparison mustn't be multiline, because it increases cognitive complexity for no good reason.
> 
> It seems not much different than any other conditional that gets formatted to multiple lines, for which I'm sure there are countless occurrences? And are you saying the reason I just gave is not a good one? Couldn't I just argue it's less complex with less assertion steps: (1) response is okay, (2) actual response = expected response?
> 
> I'm not trying to be difficult here - I'm just genuinely perplexed by your request. I don't think I've ever been asked to make something objectively less useful in order to purely make the code subjectively prettier.

I appreciate your point. It's hard to judge if something is objectively useful or not outside of the context such metric is applied. Usefulness is an inherently subjective matter if you ask me — different people may value different aspects of the same things.

The complexity I'm talking about in this case, is expression/line-level. There are a few linters that I know help fight it:
* https://github.com/Miserlou/JonesComplexity#what-is-this
* https://github.com/best-doctor/flake8-expression-complexity#example

The thing is that the more stuff you stick into an expression, the more parts of it the reader has to hold in their mind while reading it. This is a pretty much universal statement.

With many assertion steps, the tests communicate the expectations more clearly by demonstrating that many properties together should be true. Those separate asserts communicate what those properties are. And the semantics is that property1, property2 and property3 are true. With a combined assertion, it turns into "this huge expression that compares various variables together is true" without assigning obvious semantics to the properties being relied on.

That said, I know of a recommendation to have a single assertion per test, and I agree with it. I just don't think that this recommendation is about the number of `assert` statements but rather that a combination of a tightly coupled properties counts as a single assertion regardless of the amount of sub-instructions.

To your point, it's indeed important to get all of the properties visible in the test output and having multiple instructions may stand in a way of that.

So I believe that this can be solved by assigning semantics to different parts of the expression through named variables and then constructing a simpler expression passed to a single `assert` statement. Pytest has `--showlocals` to reveal parts of expressions and the readability would be preserved.

It would look smth like this
```python
content_is_plain_text = "text/plain" == resp.headers["CONTENT-TYPE"]
encoding_is_expected = expect_encoding == resp.headers.get("Content-Encoding")
body_is_expected = b"hello aiohttp\n" == await resp.read()
assert content_is_plain_text and encoding_is_expected and body_is_expected
```

P.S. With that in mind, I think that such style should be consistent throughout all modules. Introducing refactoring in one place, in a functional PR, is kinda harmful. It'd be much better to just have refactoring applied everywhere and backported. Then, a new style would be used.

steverep: @webknjaz thanks for answering my questions in detail, and for the record...

> And chasing inclusivity, I'm trying to use practices that would lean towards accessibility.

Cheers to that as you're talking to a fellow a11y enthusiast (expert if I'm feeling boastful) and legally blind developer. 🥂 👨🏻‍🦯 

In any case, the changes here are now minimal with regard to assertions following #8136, so I'll tag this resolved.

webknjaz: Thank you for your insights too!

webknjaz (approved):

webknjaz:
@steverep looks like cherry-pick still has conflicts that need to be addressed manually. Please submit a backport PR per instructions above.

PR #8064, opened 2024-01-27
Add Request type generics to Server

javajawa:
## What do these changes do?

The addition of this Generic allows type checkers and linters to link the type returned by the RequestFactory used by a Server, and the Request type that will be passed to the RequestHandler. The generic type maintains the BaseRequest bound.

This does not change user behaviour, except for developers using custom Servers and type hinting, who may need to add 

Note: I'm honestly not sure if this change is worthwhile, and am happy for it to be closed as won't fix for the extra cognitive complexity it adds.

## Related issue number

N/A

## Checklist

(I'm considering unit tests / documentation as N/A)

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -148,7 +148,7 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         self,
         message: RawRequestMessage,
         payload: StreamReader,
-        protocol: "RequestHandler",
+        protocol: "RequestHandler[BaseRequest]",
         payload_writer: AbstractStreamWriter,
         task: "asyncio.Task[None]",
         loop: asyncio.AbstractEventLoop,
@@ -267,7 +267,7 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         return self._task
 
     @property
-    def protocol(self) -> "RequestHandler":
+    def protocol(self) -> "RequestHandler[BaseRequest]":
         return self._protocol
 
     @property
--- a/CHANGES/8064.feature
+++ b/CHANGES/8064.feature
@@ -1,0 +1,1 @@
+Add generics for request (factory) type in Server.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -55,6 +55,7 @@ Ben Bader
 Ben Greiner
 Ben Kallus
 Ben Timby
+Benedict Harcourt
 Benedikt Reinartz
 Bob Haddleton
 Boris Feld
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -162,7 +162,7 @@ class BaseTestServer(ABC):
         return self._closed
 
     @property
-    def handler(self) -> Server:
+    def handler(self) -> Server[Request]:
         # for backward compatibility
         # web.Server instance
         runner = self.runner
@@ -222,7 +222,7 @@ class TestServer(BaseTestServer):
 class RawTestServer(BaseTestServer):
     def __init__(
         self,
-        handler: _RequestHandler,
+        handler: _RequestHandler[Request],
         *,
         scheme: Union[str, _SENTINEL] = sentinel,
         host: str = "127.0.0.1",
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -13,10 +13,12 @@ from typing import (
     Awaitable,
     Callable,
     Deque,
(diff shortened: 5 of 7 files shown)


Dreamsorcerer:
I think we've considered something like this before, and probably not bothered due to the amount of complexity it adds for something very very few users will encounter. But, I can take a more in-depth look later in the week.

> This does not change user behaviour, except for developers using custom Servers and type hinting, who may need to add

This will impact many more users, as they may suddenly need to update a lot of annotations to include the generic type. How much they need to update depends on a) how well the default value can be inferred and b) what strictness settings they use in their mypy config.

But, this will definitely require a change from more users who don't use this feature than ones who do.

aiohttp/web_request.py
```diff
@@ -148,7 +148,7 @@ def __init__(
         self,
         message: RawRequestMessage,
         payload: StreamReader,
-        protocol: "RequestHandler",
+        protocol: "RequestHandler[BaseRequest]",
```
Dreamsorcerer, line 151: Without looking at the rest of the PR yet, this stands out to me as something where the complexity magnifies.

If we subclass BaseRequest, then we'll surely call super().__init__(), which is then requiring a protocol that can handle any BaseRequest. So, in what situation could you have a protocol that only handles a subclass, given that you can't pass it to a subclass?

javajawa: This type constraint is "really" `RequestHandler[type(self)]`, i.e. the protocol is able to handle the current type. As `type(self)` is logically constrained to be a subclass or `BaseRequest`, all protocols passed into this `__init__` will have to handle the properties of `BaseRequest`.

A subclass may change the annotation for clarity
```python
from __future__ import annotations

class Request(aiohttp.web.BaseRequest):
    def __init__(
        self,
        message: aiohttp.http.RawRequestMessage,
        payload: aiohttp.streams.StreamReader,
        protocol: aiohttp.web.RequestHandler[Request],
        payload_writer: aiohttp.abc.AbstractStreamWriter,
        task: asyncio.Task[None],
    ) -> None:
        super().__init__(
            message, payload, protocol, payload_writer, task, asyncio.get_running_loop()
        )
```

javajawa (commented):
As It didn't feel right in the PR main box, I'm going to take this opportunity to thank you and the other maintainers for this library -- it's the backbone of a lot of little projects I've written. :heart: 

> I think we've considered something like this before, and probably not bothered due to the amount of complexity it adds for something very very few users will encounter.

Yeah, I spent a while trying to decide if this was worth raising to fix one (1) type error in a tiny project. "closed, not worth benefit/" is probably the answer I'd give me in your shoes.

> > This does not change user behaviour, except for developers using custom Servers and type hinting, who may need to add
> 
> This will impact many more users, as they may suddenly need to update a lot of annotations to include the generic type.
> 
> But, this will definitely require a change from more users who don't use this feature than ones who do.

I'm less convinced on this front, as `Application`, `AppRunner`, and the `Site` classes do not gain generics. I think only people extracting server or protocol information out of them will be dealing with new type annotations.

I can't speak for all use cases, but I've only ever touched those features for projects where I need things like early request filtering or (as in the project which caused this change to get written) messing around with telemetry injection.

aiohttp/web_request.py
Dreamsorcerer: > A subclass may change the annotation for clarity

My point is that, that code will produce a type error. If protocol requires `[Request]` then it not compatible with a parameter than requires `[BaseRequest]`. Because if the protocol does something that specifically requires attributes that only exist on Request, then it would fail when it received a BaseRequest and therefore can't be passed through the super() call without a type error. In other words, the type annotation would need to be generic as well, rather than using BaseRequest, which is where it gets rather complicated.

Dreamsorcerer:
> I'm less convinced on this front, as `Application`, `AppRunner`, and the `Site` classes do not gain generics. I think only people extracting server or protocol information out of them will be dealing with new type annotations.

It's not common, but certainly a lot more frequent than using custom Server/Request subclasses. i.e. Getting the protocol from the request or for tracing purposes etc., people may have functions which have such an annotation.

webknjaz (commented):
Is this compatible all the way down to Python 3.8? If not, it'd be hard to backport it to aiohttp 3.9 and aiohttp 3.10.

Dreamsorcerer:
> Is this compatible all the way down to Python 3.8? If not, it'd be hard to backport it to aiohttp 3.9 and aiohttp 3.10.

I don't see any reason it wouldn't be.

PR #8074, opened 2024-01-28
Improve validation in HTTP parser

Dreamsorcerer:


tests/test_http_parser.py
```diff
@@ -689,6 +759,34 @@ def test_http_request_bad_status_line(parser: Any) -> None:
...
+def test_http_request_bad_status_line_number(
+    parser: Any, nonascii_digit: bytes
+) -> None:
+    text = b"GET /digit HTTP/1." + nonascii_digit + b"\r\n\r\n"
+    with pytest.raises(http_exceptions.BadStatusLine):
+        parser.feed_data(text)
+
+
+def test_http_request_bad_status_line_separator(parser: Any) -> None:
+    # single code point, old, multibyte NFKC, multibyte NFKD
```
webknjaz, line 783: Hey @pajod, could you expand what this comment refers to? Does the `utf8sep` variable contain a value that matches all the listed cases? I'm rather confused. Or did you mean to test different cases but added just one?

tests/test_http_parser.py
```diff
@@ -710,6 +808,31 @@ def test_http_request_upgrade(parser: Any) -> None:
     assert tail == b"some raw data"
 
 
+def test_http_request_parser_utf8_request_line(parser: Any) -> None:
+    if not isinstance(response, HttpResponseParserPy):
```
webknjaz, line 812: @pajod FYI it's much cleaner to use the `@pytest.mark.xfail` decorator since it allows pytest to make the decisions earlier in the process.

tests/test_http_parser.py
pajod: That utf8sep has those properties (and LTR), a popular choice for being multiple edge cases in one. None of which are strictly needed for.. *comparing to the literal ASCII dot* as we do here, but some of which I expect to regain relevance on future refactoring.

tests/test_http_parser.py
pajod: But that *only* explains why a llhttp-only code coverage report complains!? There is something much more wrong here. Like, *I completely broke that test* levels of wrong.

*Edit:* Sorry, I did. Four times. I meant to acknowledge (in [two cases, expected to be changed](https://github.com/nodejs/llhttp/commit/10ff94eb252e0e7cb792dcde6d40d0e46b794f8a)) behaviour differences of the C parser, while keeping my tests parametrized to keep running both parsers anyway. But each time I copied the `not isinstance(response, HttpResponseParserPy)` line where it should say `not isinstance(parser, HttpRequestParserPy)`. And mypy would have told me, if not for that `Any` and the duplicate use of the `response` identifier (global scope function but also function scope variable)

Dreamsorcerer: Yeah, I noticed there was no coverage on these tests for some reason. Something to look at later.

webknjaz: Yep. I started looking into the coverage drop and that's how I noticed this thing.

FWIW we _really_ should refactor how the tests with and without extensions are parametrized/generated, like I did in multidict recently. Essentially, there was an import loop in some places in tests that prevented C-extension tests from being executed (https://github.com/aio-libs/multidict/issues/837 / https://github.com/aio-libs/multidict/pull/915 / https://multidict.aio-libs.org/en/latest/changes/#contributor-facing-changes). So I fixed that by having an explicit global option for requiring one mode or the other, with a collection of fixtures reused everywhere and zero magic around import attempts and handling failures in weird ways.

Another thing I configured is the module classification in Codecov with different expected coverage thresholds — the goal should be that tests get 100% coverage in every CI run (from all jobs combined, of course). And the actual project code coverage should be measured as a separate metric. Currently, a global threshold value allows coverage to drop in tests if it's compensated by coverage in the project, meaning that we may be gaining more dead code (read: tests that are never executed), which results in a false sense of things being tested, when they aren't.

PR #8082, opened 2024-01-28
Release v3.9.2

Dreamsorcerer:


CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+---------
+
+- Fixed websocket connection leak.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`7978`.
+
+
+
```
Dreamsorcerer, line 26: @webknjaz If you could check the template code for this. These 3 lines include whitespace "  " in each section. I manually deleted them before commit.

webknjaz: I think that it's one of the artifacts of Jinja templating. Things like `{%- endwith -%}` drop spaces before and after the instruction, while the tag variants without those `-` keep them. It's rather tricky to get it right for all of the combinations of reference paragraphs.
I don't remember exactly, but I must've ended up with this compromise.

It's hard to debug and it was far more important to me that the rendered web page would look good, even at the cost of RST being less pretty.

Maybe, at some point I'll get to updating this, but right now it's not a priority for me.

If you're curious, you can take a look yourself, but that requires *manual* testing of variants with just commit hashes, just PR/issue numbers, just arbitrary references that are neither commit hashes nor numbers, and combinations of all or part of those together.

But I don't recommend doing this right now because it can be quite a lengthy trip down the rabbit hole.

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+  *Related issues and pull requests on GitHub:*
+  :issue:`8079`.
+
+
+
+
+Features
+--------
+
+- Added support for passing `True` to `ssl` while deprecating `None` -- by :user:`xiangyan99`.
```
webknjaz: The single backticks here should be double backticks or explicit roles. I like to make updates to the individual change notes before the release, but updating this file post-generation is also fine. It could look like this:
```suggestion
- Added support for passing :py:data:`True` to ``ssl`` while deprecating :py:data:`None` -- by :user:`xiangyan99`.
```

But then, this wording is still incomplete — it doesn't provide any context on where this is happening and whether it's in client or server mode etc. Can we include that?

Regarding the byline — it can be either at the end of the first sentence or paragraph, at the very end. If you choose the end position, don't add a trailing comma, just let it sit on the last line after one empty line-separator.

Here's some examples of what I mean: https://frozenlist.aio-libs.org/en/latest/changes/#contributor-facing-changes / https://ansible-pylibssh.readthedocs.io/en/latest/contributing/guidelines/#examples-for-changelog-entries-adding-to-your-pull-requests.

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+  (BACKWARD INCOMPATIBLE)
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8074`.
+
+
+
+- Improved validation of paths for static resources -- by :user:`bdraco`.
```
webknjaz: Server?

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+  *Related issues and pull requests on GitHub:*
+  :issue:`8066`.
+
+
+
+
+Miscellaneous internal changes
+------------------------------
+
+- Replace all tmpdir fixtures with tmp_path in test suite.
```
webknjaz:
```suggestion
- Replaced all ``tmpdir`` fixtures with ``tmp_path`` in test suite.
```
(could also be moved to a `contrib` changelog fragment, if you ask me..)

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+
+
+- Fixed an unhandled exception in the Python HTTP parser on header lines starting with a colon -- by :user:`pajod`.
+
+  Invalid request lines with anything but a dot between the HTTP major and minor version are now rejected.
+  Invalid header field names containing question mark or slash are now rejected.
+  Such requests are incompatible with :rfc:`9110#section-5.6.2` and are not known to be of any legitimate use.
+
+  (BACKWARD INCOMPATIBLE)
```
webknjaz: Instead of this line, I'd symlink the original change fragment to the new `breaking` category (before invoking towncrier) so it'd show up in both.

CHANGES.rst
Dreamsorcerer: Right, at a glance, I'd think that the change would need to be from:
`{% if something -%}` to `{%- if something %}`
At:
https://github.com/aio-libs/aiohttp/blob/master/CHANGES/.TEMPLATE.rst?plain=1#L48-L61

i.e. Keep the newline that's inside the block, but lose the newline with "  " before the block.

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
 
 .. towncrier release notes start
 
+3.9.2 (2024-01-28)
+==================
+
+Bug fixes
+---------
+
+- Fixed websocket connection leak.
```
webknjaz: Client or server?

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`7978`.
+
+
+
+- Fixed `web.FileResponse` doing blocking I/O in the event loop.
```
webknjaz: Client or server?

Also, it's linkable:

```suggestion
- Fixed :py:class:`web.FileResponse <aiohttp.web.FileResponse>` doing blocking I/O in the event loop.
```

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8012`.
+
+
+
+- Fixed double compress when compression enabled and compressed file exists.
```
webknjaz: Client or server?

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+
+Improved documentation
+----------------------
+
+- Updated :ref:`contributing/Tests coverage <aiohttp-contributing>` section to show how we use ``codecov`` -- by :user:`Dreamsorcerer`.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`7916`.
```
webknjaz: I'd move this fragment to the `contrib` section since it doesn't really concern the end-users, but contributors to the project.

CHANGES.rst
```diff
@@ -10,6 +10,188 @@
...
+
+
+
+- Fixed examples of ``fallback_charset_resolver`` function in client_advanced documentation. -- by :user:`henry0312`.
```
webknjaz:
```suggestion
- Fixed examples of ``fallback_charset_resolver`` function in the :doc:`client_advanced` document. -- by :user:`henry0312`.
```

CHANGES.rst
webknjaz: Yes, but I remember it breaking in some corner cases which irritated me enough to postpone figuring out how to make it work better.

webknjaz: I'd very much like to see this improved, but it's annoying to debug so I temporarily gave up. There's some important work to do in multidict that blocks it from being released, which I think takes priority.

CHANGES.rst
Dreamsorcerer: There's no reference documentation for this currently. So, can't link it.

CHANGES.rst
```diff
@@ -10,6 +10,201 @@
...
+  *Related issues and pull requests on GitHub:*
+  :issue:`8079`.
+
+
+
+
+Features
+--------
+
+- Added support for passing :py:data:`True` to ``ssl`` while deprecating :py:data:`None` -- by :user:`xiangyan99`.
```
webknjaz: Is `ssl` a function/method/class init param on server/client?

webknjaz (approved):

CHANGES.rst
webknjaz: I suppose it wasn't backported, because the `latest` RTD build does have this reference.

webknjaz:
FTR I just approved the PyPI upload in the GHA job. It should complete soon.

bdraco:
Retested the final version on production. Everything looks good. Thanks for getting this out.

CHANGES/8010.doc
```diff
@@ -1,2 +0,0 @@
-On the `CHANGES/README.rst <https://github.com/aio-libs/aiohttp/tree/master/CHANGES/README.rst>`_ page,
```
webknjaz, line 1: @Dreamsorcerer did you decide not to include this fragment into the resulting changelog? Why? Now, this change note only exists on `master` and shows up in the changelog draft for 4.0: https://docs.aiohttp.org/en/latest/changes.html.

webknjaz: Also, removal of this fragment didn't happen in `3.10` either: https://github.com/aio-libs/aiohttp/commit/0124dc58f4a06a7316925e4c10e00033cabe1d2d.

webknjaz: @Dreamsorcerer I checked the fragment list and this seems to be the only one that disappeared mysteriously.

Dreamsorcerer: Ah, not sure. Must have done something weird.

PR #8085, opened 2024-01-28
Fix xfail reason for unsupported upgrade test

steverep:
## What do these changes do?

I noticed this test is pretty finicky on Mac and Windows in #8063.  Again, my Python is quite rusty and still getting up to date, but the `reason` string given to `xfail` seems incorrectly spread over 2 lines.  The fact that it isn't caught by linting makes me think I'm wrong, but if I'm not it makes sense that Mac and Windows would be less forgiving and `pytest` might timeout.

In any case, I just shortened the string to one line since the URL is already within the test, and converted the comment to a doc string.  Hopefully that fixes the timeouts.

Nope

## Related issue number

n/a

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/8085.bugfix.rst
+++ b/CHANGES/8085.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed xfail reason for unsupported upgrade server test -- by :user:`steverep`
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -307,6 +307,7 @@ Stepan Pletnev
 Stephan Jaensch
 Stephen Cirelli
 Stephen Granade
+Steve Repsher
 Steven Seguin
 Sunghyun Hwang
 Sunit Deshpande
--- a/tests/test_web_server.py
+++ b/tests/test_web_server.py
@@ -24,12 +24,14 @@ async def test_simple_server(aiohttp_raw_server: Any, aiohttp_client: Any) -> No
 @pytest.mark.xfail(
     not helpers.NO_EXTENSIONS,
     raises=client.ServerDisconnectedError,
-    reason="The behavior of C-extensions differs from pure-Python: "
-    "https://github.com/aio-libs/aiohttp/issues/6446",
+    reason="The behavior of C-extensions differs from pure-Python (see #6446)",
 )
 async def test_unsupported_upgrade(aiohttp_raw_server, aiohttp_client) -> None:
-    # don't fail if a client probes for an unsupported protocol upgrade
-    # https://github.com/aio-libs/aiohttp/issues/6446#issuecomment-999032039
+    """don't fail if a client probes for an unsupported protocol upgrade
+
+    https://github.com/aio-libs/aiohttp/issues/6446#issuecomment-999032039
+    """
+
     async def handler(request: web.Request):
         return web.Response(body=await request.read())
 


Dreamsorcerer:
2 strings like that are simply concatenated together. It is also just a description of why we expect the test to fail. I'm unclear why you expect that to change the results of the test itself?

webknjaz:
Pretty sure this change does nothing. Why don't we.. not merge it?

steverep:
> 2 strings like that are simply concatenated together. It is also just a description of why we expect the test to fail. I'm unclear why you expect that to change the results of the test itself?

I had thought parentheses were required when the concatenation was over multiple lines.  Guess I was wrong. 🤷🏻‍♂️ 

The test just hangs and times out on Win/Mac 3.11/3.12, so clearly something is wrong...

bdraco:
I have attempted to debug it in https://github.com/aio-libs/aiohttp/pull/7960 but not having a system where I can replicate the hang on locally has been a challenge

webknjaz:
> I had thought parentheses were required when the concatenation was over multiple lines.

They are required. And in this case, those parents are the outer ones from the function call. Outside of this context, you'd need explicit parens.

steverep:
> I have attempted to debug it in #7960 but not having a system where I can replicate the hang on locally has been a challenge (along with time constraints)

From a contributor's view, especially first time to a repo, would be nice to have that context a priori.  I spent some time investigating whether or not it was my changes.

I'd suggest either disabling or restricting it until fixed - it's not preventing anything if it doesn't work.

webknjaz:
> From a contributor's view, especially first time to a repo, would be nice to have that context a priori. I spent some time investigating whether or not it was my changes.
> 
> I'd suggest either disabling or restricting it until fixed - it's not preventing anything if it doesn't work.

That's understandable. Though, in PRs, there's branch protection and a gate job is already marked as required — that's the one indicator that's important. It marks some jobs as non-mandatory, but due to GitHub's UI limitations, it's not always obvious that some jobs can be ignored safely.

We do want those tests to run. Disabling them would essentially mean removal since they won't be tracked anymore. We keep discovering tests that were added over 7 years ago but disabled (here's one example: #8096). This demonstrates that not running something just because it fails is not viable.

webknjaz:
@Dreamsorcerer are you talking about the job label in the UI?

Dreamsorcerer:
> @Dreamsorcerer are you talking about the job label in the UI?

Screenshot would help. :P
image

The one with true is experimental, the one with false is a required check. So, changing the true/false value could make it more readable.

Dreamsorcerer:
Likewise with NO_EXTESIONS actually...

webknjaz:
Fair. This is something to consider globally. I'll think about it as I'm making rounds of CI updates anyway.

steverep:
@Dreamsorcerer yes, it would really help to see something other than true or false in the job matrix. 👍🏻 

Two other simple suggestions:
1. For problematic tests, if you don't want to conditionally disable them, then make sure something useful to the contributor gets printed to the log telling me to ignore it.
2. If there's always a set of experimental tests I shouldn't worry about, put that in the contributing docs.

PR #6409, opened 2021-12-11
Support absolute-form and authority-form URLs by web server

asvetlov:
Fixes #6227

PR #8088, opened 2024-01-29
re-enable 4 tests, accidentally dropped

pajod:
## What do these changes do?

I copied the wrong line, earlier, so those new tests did not run.
It is the (request) parser param that should have informed the xfail() decision.

```diff
 def test_request_parser_foo(parser: Any) -> None:
-    if not isinstance(response, HttpResponseParserPy):
+    if not isinstance(parser, HttpRequestParserPy):
         pytest.xfail("duh")
```

Testing only.

However, I do expect behaviour and/or [documentation](https://docs.aiohttp.org/en/stable/web_advanced.html#unicode-support) changes are desirable.. after I figured out a way to rewrite these tests to more clearly pinpoint the problem.

## Commentary

* test_empty_header_name, which is a subset of test_invalid_header_spacing: [testing against llhttp 9.2](https://github.com/aio-libs/aiohttp/pull/8146) makes those pass

* test_parse_unusual_request_line: I expected applying #7651 would give me a descriptive exception text for the remaining differences, but I still get "BadStatusLine" so I stopped short of figuring out if they fail in any of the [actually relevant](https://github.com/aio-libs/aiohttp/pull/6409) branches, or need to be fine-tuned before even attempting at making them pass.

* test_http_request_parser_utf8_request_line: After reading up on more clever ways to abuse parser differences, maybe leaving the llhttp side as-is and making the python parser less liberal (as in rfc761#section-2.10) would be the better choice.

## Related issue number

Follow-up to #8074

## Checklist

- [x] I think the code looks ugly
 * Unfortunately, `pytest.xfail()` [does not appear to share](https://github.com/pytest-dev/pytest/issues/4216) the strict_xfail behaviour of `pytest.mark.xfail()` - but how else could I see the XPASS when testing against llhttp main branch?
- [x] No API changes to be documented
- [x] Incomplete, but safe to merge at this stage
 * While I expect no harm from merging the initial commit, going straight for a cleaner and more generic test refactoring would make more sense to me. The point of submitting is more about the verification CI run, and creating a place to leave additional comments on style and future for these tests without spamming the PR that introduced the problem

### DIFF
--- a/CHANGES/8088.contrib.rst
+++ b/CHANGES/8088.contrib.rst
@@ -1,0 +1,1 @@
+Enabled HTTP parser tests originally intended for 3.9.2 release -- by :user:`pajod`.
--- a/tests/test_http_parser.py
+++ b/tests/test_http_parser.py
@@ -282,9 +282,20 @@ def test_parse_headers_longline(parser: Any) -> None:
         parser.feed_data(text)
 
 
+@pytest.fixture
+def xfail_c_parser_status(request) -> None:
+    if isinstance(request.getfixturevalue("parser"), HttpRequestParserPy):
+        return
+    request.node.add_marker(
+        pytest.mark.xfail(
+            reason="Regression test for Py parser. May match C behaviour later.",
+            raises=http_exceptions.BadStatusLine,
+        )
+    )
+
+
+@pytest.mark.usefixtures("xfail_c_parser_status")
 def test_parse_unusual_request_line(parser: Any) -> None:
-    if not isinstance(response, HttpResponseParserPy):
-        pytest.xfail("Regression test for Py parser. May match C behaviour later.")
     text = b"#smol //a HTTP/1.3\r\n\r\n"
     messages, upgrade, tail = parser.feed_data(text)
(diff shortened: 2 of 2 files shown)


webknjaz:
> > _really_ should refactor how the tests with and without extensions are parametrized/generated

FWIW I didn't mean that you should be doing the refactoring. Just fixing these few places is a good starter. I wanted a more global refactoring with fixtures in `conftest.py`, injected in all of the relevant tests. This would be a major refactoring, a huge project. Before taking it on, it'd need to be talked through to synchronize the vision and stuff...

Dreamsorcerer:
@pajod Any chance you could look into the Pypy failure in order to get this PR finished?

Dreamsorcerer:
@pajod Thanks for that. If you've got time to complete the backport as above, that'd be appreciated.

webknjaz:
@pajod @Dreamsorcerer can this be backported to 3.9? All the test-only updates should be backported to the oldest active branch, provided they don't fail there. This is to reduce merge conflicts in future backports that might touch surrounding lines.

PR #8089, opened 2024-01-29
💅 Propagate error causes via asyncio protocols

webknjaz:
This is supposed to unify setting exceptions on the future objects, allowing to also attach their causes whenever available.

It's also supposed to help with tracking down what's happening with #4581.

## What do these changes do?

$sbj.

They should be able to see more detailed tracebacks.

## Related issue number

#4581

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -4,18 +4,21 @@ from typing import Any, Optional, Tuple
 
 from .base_protocol import BaseProtocol
 from .client_exceptions import (
+    ClientConnectionError,
     ClientOSError,
     ClientPayloadError,
     ServerDisconnectedError,
     SocketTimeoutError,
 )
 from .helpers import (
+    _EXC_SENTINEL,
     BaseTimerContext,
     set_exception,
     set_result,
     status_code_must_be_empty_body,
 )
 from .http import HttpResponseParser, RawResponseMessage, WebSocketReader
+from .http_exceptions import HttpProcessingError
 from .streams import EMPTY_PAYLOAD, DataQueue, StreamReader
 
 
@@ -80,33 +83,62 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
     def connection_lost(self, exc: Optional[BaseException]) -> None:
         self._drop_timeout()
 
-        if exc is not None:
-            set_exception(self.closed, exc)
-        else:
+        original_connection_error = exc
+        reraised_exc = original_connection_error
+
+        connection_closed_cleanly = original_connection_error is None
+
+        if connection_closed_cleanly:
             set_result(self.closed, None)
+        else:
+            assert original_connection_error is not None
+            set_exception(
+                self.closed,
+                ClientConnectionError(
+                    f"Connection lost: {original_connection_error !s}",
+                ),
+                original_connection_error,
+            )
 
         if self._payload_parser is not None:
-            with suppress(Exception):
+            with suppress(Exception):  # FIXME: log this somehow?
                 self._payload_parser.feed_eof()
 
         uncompleted = None
         if self._parser is not None:
             try:
                 uncompleted = self._parser.feed_eof()
-            except Exception as e:
+            except Exception as underlying_exc:
                 if self._payload is not None:
-                    exc = ClientPayloadError("Response payload is not completed")
-                    exc.__cause__ = e
-                    self._payload.set_exception(exc)
+                    client_payload_exc_msg = (
+                        f"Response payload is not completed: {underlying_exc !r}"
+                    )
+                    if not connection_closed_cleanly:
+                        client_payload_exc_msg = (
+                            f"{client_payload_exc_msg !s}. "
+                            f"{original_connection_error !r}"
+                        )
+                    set_exception(
+                        self._payload,
+                        ClientPayloadError(client_payload_exc_msg),
+                        underlying_exc,
+                    )
 
         if not self.is_eof():
-            if isinstance(exc, OSError):
-                exc = ClientOSError(*exc.args)
-            if exc is None:
-                exc = ServerDisconnectedError(uncompleted)
+            if isinstance(original_connection_error, OSError):
+                reraised_exc = ClientOSError(*original_connection_error.args)
+            if connection_closed_cleanly:
+                reraised_exc = ServerDisconnectedError(uncompleted)
             # assigns self._should_close to True as side effect,
             # we do it anyway below
-            self.set_exception(exc)
+            underlying_non_eof_exc = (
+                _EXC_SENTINEL
+                if connection_closed_cleanly
+                else original_connection_error
+            )
+            assert underlying_non_eof_exc is not None
+            assert reraised_exc is not None
+            self.set_exception(reraised_exc, underlying_non_eof_exc)
 
         self._should_close = True
         self._parser = None
@@ -114,7 +146,7 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         self._payload_parser = None
         self._reading_paused = False
 
-        super().connection_lost(exc)
+        super().connection_lost(reraised_exc)
 
     def eof_received(self) -> None:
         # should call parser.feed_eof() most likely
@@ -128,10 +160,14 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         super().resume_reading()
         self._reschedule_timeout()
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
         self._should_close = True
         self._drop_timeout()
-        super().set_exception(exc)
+        super().set_exception(exc, exc_cause)
 
     def set_parser(self, parser: Any, payload: Any) -> None:
         # TODO: actual types are:
@@ -208,7 +244,7 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         exc = SocketTimeoutError("Timeout on reading data from socket")
         self.set_exception(exc)
         if self._payload is not None:
-            self._payload.set_exception(exc)
+            set_exception(self._payload, exc)
 
     def data_received(self, data: bytes) -> None:
         self._reschedule_timeout()
@@ -234,14 +270,14 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
                 # parse http messages
                 try:
                     messages, upgraded, tail = self._parser.feed_data(data)
-                except BaseException as exc:
+                except BaseException as underlying_exc:
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
                     # should_close is True after the call
-                    self.set_exception(exc)
+                    self.set_exception(HttpProcessingError(), underlying_exc)
                     return
 
                 self._upgraded = upgraded
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -53,6 +53,7 @@ from .helpers import (
     noop,
     parse_mimetype,
     reify,
+    set_exception,
     set_result,
 )
 from .http import (
@@ -566,20 +567,29 @@ class ClientRequest:
 
                 for chunk in self.body:
                     await writer.write(chunk)  # type: ignore[arg-type]
-        except OSError as exc:
-            if exc.errno is None and isinstance(exc, asyncio.TimeoutError):
-                protocol.set_exception(exc)
-            else:
-                new_exc = ClientOSError(
-                    exc.errno, "Can not write request body for %s" % self.url
+        except OSError as underlying_exc:
+            reraised_exc = underlying_exc
+
+            exc_is_not_timeout = underlying_exc.errno is not None or not isinstance(
+                underlying_exc, asyncio.TimeoutError
+            )
+            if exc_is_not_timeout:
+                reraised_exc = ClientOSError(
+                    underlying_exc.errno,
+                    f"Can not write request body for {self.url !s}",
                 )
-                new_exc.__context__ = exc
-                new_exc.__cause__ = exc
-                protocol.set_exception(new_exc)
+
+            set_exception(protocol, reraised_exc, underlying_exc)
         except asyncio.CancelledError:
             await writer.write_eof()
-        except Exception as exc:
-            protocol.set_exception(exc)
+        except Exception as underlying_exc:
+            set_exception(
+                protocol,
+                ClientConnectionError(
+                    f"Failed to send bytes into the underlying connection {conn !s}",
+                ),
+                underlying_exc,
+            )
         else:
             await writer.write_eof()
             protocol.start_timeout()
@@ -1019,7 +1029,7 @@ class ClientResponse(HeadersMixin):
         content = self.content
         # content can be None here, but the types are cheated elsewhere.
         if content and content.exception() is None:  # type: ignore[truthy-bool]
-            content.set_exception(ClientConnectionError("Connection closed"))
+            set_exception(content, ClientConnectionError("Connection closed"))
         self._released = True
 
     async def wait_for_close(self) -> None:
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -797,9 +797,39 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
         fut.set_result(result)
 
 
-def set_exception(fut: "asyncio.Future[_T]", exc: BaseException) -> None:
-    if not fut.done():
-        fut.set_exception(exc)
+_EXC_SENTINEL = BaseException()
+
+
+class ErrorableProtocol(Protocol):
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ...,
+    ) -> None:
+        ...  # pragma: no cover
+
+
+def set_exception(
+    fut: "asyncio.Future[_T] | ErrorableProtocol",
+    exc: BaseException,
+    exc_cause: BaseException = _EXC_SENTINEL,
+) -> None:
+    """Set future exception.
+
+    If the future is marked as complete, this function is a no-op.
+
+    :param exc_cause: An exception that is a direct cause of ``exc``.
+                      Only set if provided.
+    """
+    if asyncio.isfuture(fut) and fut.done():
+        return
+
+    exc_is_sentinel = exc_cause is _EXC_SENTINEL
+    exc_causes_itself = exc is exc_cause
+    if not exc_is_sentinel and not exc_causes_itself:
+        exc.__cause__ = exc_cause
+
+    fut.set_exception(exc)
 
 
 @functools.total_ordering
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -28,10 +28,12 @@ from . import hdrs
 from .base_protocol import BaseProtocol
 from .compression_utils import HAS_BROTLI, BrotliDecompressor, ZLibDecompressor
 from .helpers import (
+    _EXC_SENTINEL,
     DEBUG,
     NO_EXTENSIONS,
     BaseTimerContext,
     method_must_be_empty_body,
+    set_exception,
     status_code_must_be_empty_body,
 )
 from .http_exceptions import (
@@ -439,13 +441,16 @@ class HttpParser(abc.ABC, Generic[_MsgT]):
                 assert self._payload_parser is not None
                 try:
                     eof, data = self._payload_parser.feed_data(data[start_pos:], SEP)
-                except BaseException as exc:
+                except BaseException as underlying_exc:
+                    reraised_exc = underlying_exc
                     if self.payload_exception is not None:
-                        self._payload_parser.payload.set_exception(
-                            self.payload_exception(str(exc))
-                        )
-                    else:
-                        self._payload_parser.payload.set_exception(exc)
+                        reraised_exc = self.payload_exception(str(underlying_exc))
+
+                    set_exception(
+                        self._payload_parser.payload,
+                        reraised_exc,
+                        underlying_exc,
+                    )
 
                     eof = True
                     data = b""
@@ -826,7 +831,7 @@ class HttpPayloadParser:
                             exc = TransferEncodingError(
                                 chunk[:pos].decode("ascii", "surrogateescape")
                             )
-                            self.payload.set_exception(exc)
+                            set_exception(self.payload, exc)
                             raise exc
                         size = int(bytes(size_b), 16)
 
@@ -929,8 +934,12 @@ class DeflateBuffer:
         else:
             self.decompressor = ZLibDecompressor(encoding=encoding)
 
-    def set_exception(self, exc: BaseException) -> None:
-        self.out.set_exception(exc)
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
+        set_exception(self.out, exc, exc_cause)
 
     def feed_data(self, chunk: bytes, size: int) -> None:
         if not size:
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -14,7 +14,13 @@ from typing import (
 )
 
 from .base_protocol import BaseProtocol
-from .helpers import BaseTimerContext, TimerNoop, set_exception, set_result
+from .helpers import (
+    _EXC_SENTINEL,
+    BaseTimerContext,
+    TimerNoop,
+    set_exception,
+    set_result,
+)
 from .log import internal_logger
 
 __all__ = (
@@ -146,19 +152,23 @@ class StreamReader(AsyncStreamReaderMixin):
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
         self._exception = exc
         self._eof_callbacks.clear()
 
         waiter = self._waiter
         if waiter is not None:
             self._waiter = None
-            set_exception(waiter, exc)
+            set_exception(waiter, exc, exc_cause)
 
         waiter = self._eof_waiter
         if waiter is not None:
             self._eof_waiter = None
-            set_exception(waiter, exc)
+            set_exception(waiter, exc, exc_cause)
 
     def on_eof(self, callback: Callable[[], None]) -> None:
         if self._eof:
@@ -499,7 +509,11 @@ class EmptyStreamReader(StreamReader):  # lgtm [py/missing-call-to-init]
     def exception(self) -> Optional[BaseException]:
         return None
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
         pass
 
     def on_eof(self, callback: Callable[[], None]) -> None:
@@ -574,14 +588,18 @@ class DataQueue(Generic[_T]):
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
         self._eof = True
         self._exception = exc
 
         waiter = self._waiter
         if waiter is not None:
             self._waiter = None
-            set_exception(waiter, exc)
+            set_exception(waiter, exc, exc_cause)
 
     def feed_data(self, data: _T, size: int = 0) -> None:
         self._size += size
--- a/CHANGES/8089.bugfix.rst
+++ b/CHANGES/8089.bugfix.rst
@@ -1,0 +1,3 @@
+The asynchronous internals now set the underlying causes
+when assigning exceptions to the future objects
+-- by :user:`webknjaz`.
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -19,7 +19,7 @@ from multidict import CIMultiDict as _CIMultiDict, CIMultiDictProxy as _CIMultiD
 from yarl import URL as _URL
 
 from aiohttp import hdrs
-from aiohttp.helpers import DEBUG
+from aiohttp.helpers import DEBUG, set_exception
 
 from .http_exceptions import (
     BadHttpMessage,
@@ -763,11 +763,13 @@ cdef int cb_on_body(cparser.llhttp_t* parser,
     cdef bytes body = at[:length]
     try:
         pyparser._payload.feed_data(body, length)
-    except BaseException as exc:
+    except BaseException as underlying_exc:
+        reraised_exc = underlying_exc
         if pyparser._payload_exception is not None:
-            pyparser._payload.set_exception(pyparser._payload_exception(str(exc)))
-        else:
-            pyparser._payload.set_exception(exc)
+            reraised_exc = pyparser._payload_exception(str(underlying_exc))
+
+        set_exception(pyparser._payload, reraised_exc, underlying_exc)
+
         pyparser._payload_error = 1
         return -1
     else:
--- a/aiohttp/base_protocol.py
+++ b/aiohttp/base_protocol.py
@@ -1,6 +1,7 @@
 import asyncio
 from typing import Optional, cast
 
+from .helpers import set_exception
 from .tcp_helpers import tcp_nodelay
 
 
@@ -76,7 +77,11 @@ class BaseProtocol(asyncio.Protocol):
         if exc is None:
             waiter.set_result(None)
         else:
-            waiter.set_exception(exc)
+            set_exception(
+                waiter,
+                ConnectionError("Connection lost"),
+                exc,
+            )
 
     async def _drain_helper(self) -> None:
         if not self.connected:
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -25,7 +25,7 @@ from typing import (
 
 from .base_protocol import BaseProtocol
 from .compression_utils import ZLibCompressor, ZLibDecompressor
-from .helpers import NO_EXTENSIONS
+from .helpers import NO_EXTENSIONS, set_exception
 from .streams import DataQueue
 
 __all__ = (
@@ -305,7 +305,7 @@ class WebSocketReader:
             return self._feed_data(data)
         except Exception as exc:
             self._exc = exc
-            self.queue.set_exception(exc)
+            set_exception(self.queue, exc)
             return True, b""
 
     def _feed_data(self, data: bytes) -> Tuple[bool, bytes]:
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -25,7 +25,7 @@ import yarl
 
 from .abc import AbstractAccessLogger, AbstractAsyncAccessLogger, AbstractStreamWriter
 from .base_protocol import BaseProtocol
-from .helpers import ceil_timeout
+from .helpers import ceil_timeout, set_exception
 from .http import (
     HttpProcessingError,
     HttpRequestParser,
@@ -577,7 +577,7 @@ class RequestHandler(BaseProtocol):
                         self.log_debug("Uncompleted request.")
                         self.close()
 
-                payload.set_exception(PayloadAccessError())
+                set_exception(payload, PayloadAccessError())
 
             except asyncio.CancelledError:
                 self.log_debug("Ignored premature client disconnection ")
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -42,6 +42,7 @@ from .helpers import (
     parse_http_date,
     reify,
     sentinel,
+    set_exception,
     set_result,
 )
 from .http_parser import RawRequestMessage
@@ -813,7 +814,7 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         return
 
     def _cancel(self, exc: BaseException) -> None:
-        self._payload.set_exception(exc)
+        set_exception(self._payload, exc)
         for fut in self._disconnection_waiters:
             set_result(fut, None)
 
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -11,7 +11,7 @@ from multidict import CIMultiDict
 
 from . import hdrs
 from .abc import AbstractStreamWriter
-from .helpers import call_later, set_result
+from .helpers import call_later, set_exception, set_result
 from .http import (
     WS_CLOSED_MESSAGE,
     WS_CLOSING_MESSAGE,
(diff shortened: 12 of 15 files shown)


webknjaz:
@bdraco @Dreamsorcerer please, consider this idea :pray: Hopefully, it's viable.

I expect that this needs polishing, of course. Especially since I didn't even bother touching the tests. Surprisingly, only 6 failed: https://github.com/aio-libs/aiohttp/actions/runs/7691292578/job/20956384838?pr=8089#step:11:3813.

aiohttp/client_proto.py
```diff
@@ -80,41 +82,71 @@ def is_connected(self) -> bool:
...
             try:
                 uncompleted = self._parser.feed_eof()
-            except Exception as e:
+            except Exception as underlying_exc:
                 if self._payload is not None:
-                    exc = ClientPayloadError("Response payload is not completed")
-                    exc.__cause__ = e
-                    self._payload.set_exception(exc)
+                    client_payload_exc_msg = (
+                        "Response payload is not completed: " f"{underlying_exc !r}"
```
webknjaz:
```suggestion
                        f"Response payload is not completed: {underlying_exc !r}"
```

aiohttp/client_proto.py
```diff
@@ -234,14 +270,19 @@ def data_received(self, data: bytes) -> None:
...
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
                     # should_close is True after the call
-                    self.set_exception(exc)
+                    self.set_exception(
+                        ClientPayloadError(
+                            "Unable to parse response payload: " f"{underlying_exc !r}"
```
webknjaz:
```suggestion
                            f"Unable to parse response payload: {underlying_exc !r}"
```

aiohttp/client_reqrep.py
```diff
@@ -566,20 +567,33 @@ async def write_bytes(
...
+            set_exception(protocol, reraised_exc, underlying_exc)
         except asyncio.CancelledError:
             await writer.write_eof()
-        except Exception as exc:
-            protocol.set_exception(exc)
+        except Exception as underlying_exc:
+            set_exception(
+                protocol,
+                ClientConnectionError(
+                    "Failed to send bytes into the " f"underlying connection {conn !s}",
```
webknjaz:
```suggestion
                    f"Failed to send bytes into the underlying connection {conn !s}",
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+    :param exc_cause: An exception that is a direct cause of ``exc``.
+                      Only set if provided.
+    """
+    if asyncio.isfuture(fut) and fut.done():
+        return
+
+    exc_is_sentinel = exc_cause is ErrorableMixin._EXC_SENTINEL
+    exc_causes_itself = exc is exc_cause
+    if not exc_is_sentinel and not exc_causes_itself:
+        exc.__cause__ = exc_cause
```
Dreamsorcerer, line 830: Maybe I misremember, but I don't think this includes a traceback, as the traceback is only present if we are catching a thrown exception.

If so, it'd be really useful to have a way to get the traceback at this point in the code and include it in the exception.

aiohttp/streams.py
```diff
@@ -84,7 +90,7 @@ def iter_chunks(self) -> ChunkTupleAsyncStreamIterator:
         return ChunkTupleAsyncStreamIterator(self)  # type: ignore[arg-type]
 
 
-class StreamReader(AsyncStreamReaderMixin):
+class StreamReader(AsyncStreamReaderMixin, ErrorableMixin):
```
Dreamsorcerer: I wonder if it makes more sense to use a Protocol, rather than needing to inherit from this mixin (which doesn't actually define any functionality).

webknjaz: Dunno. I never worked with the protocols much. Just know they exist and are callback/hook collections. Plz send examples of how you envision this. Currently, this was meant as a quick hack to please MyPy and get the PoC going.

aiohttp/helpers.py
webknjaz: You might be right, I don't know. This needs experimentation, then. Anyway, the current thing is definitely not harmful as it repeats the approach present in the code base already, just consolidating it.

aiohttp/client_reqrep.py
```diff
@@ -566,20 +567,33 @@ async def write_bytes(
...
+
+            exc_is_not_timeout = underlying_exc.errno is not None or not isinstance(
+                underlying_exc, asyncio.TimeoutError
+            )
+            if exc_is_not_timeout:
+                reraised_exc = ClientOSError(
+                    underlying_exc.errno,
+                    f"Can not write request body for {self.url !s}",
                 )
-                new_exc.__context__ = exc
```
webknjaz, line 576: @Dreamsorcerer @bdraco so this place also sets `__context__`, but doesn't `__cause__` have an effect of making `__context__` useless?

This https://stackoverflow.com/a/11235957/595220 suggests that it's to be used with the `__traceback__`, though. I wonder if something like [`with_traceback()`](https://docs.python.org/3/library/exceptions.html#BaseException.with_traceback) is needed.

https://docs.python.org/3/library/exceptions.html#exception-context also says
> Setting `__cause__` also implicitly sets the `__suppress_context__` attribute to `True`

and

> The default traceback display code shows these chained exceptions in addition to the traceback for the exception itself. **An explicitly chained exception in `__cause__` is always shown when present. An implicitly chained exception in `__context__` is shown only if `__cause__` is [None](https://docs.python.org/3/library/constants.html#None) and `__suppress_context__` is false.**

aiohttp/streams.py
Dreamsorcerer:
```
class Errorable(Protocol):
    def set_exception(
        self,
        exc: BaseException,
        exc_cause: BaseException = ...,
    ):
        ...
```

Then just use that for the parameter type, anything that has a compatible set_exception() method will work then, without needing to subclass anything.

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ):
+        raise NotImplementedError
+
+
+def set_exception(
+    fut: "asyncio.Future[_T] | ErrorableMixin",
```
webknjaz: @Dreamsorcerer so with https://github.com/aio-libs/aiohttp/pull/8089#discussion_r1469722650, do you mean that doing as follows would mean that I'd simply drop the inheritance from other modules even from `StreamReader`?

```suggestion
class ErrorableProtocol(asyncio.Protocol):
    _EXC_SENTINEL = BaseException()

    def set_exception(
        self,
        exc: BaseException,
        exc_cause: BaseException = _EXC_SENTINEL,
    ):
        raise NotImplementedError


def set_exception(
    fut: "asyncio.Future[_T] | ErrorableProtocol",
```

Also, did you mean `asyncio.Protocol` or `aiohttp.base_protocol.BaseProtocol`? Or maybe, `asyncio.BaseProtocol`?

Dreamsorcerer: Uhh, `typing.Protocol`. :P

Dreamsorcerer: You basically define a stub with Protocol, and then the type checker checks for compatibility with the interface of the stub (i.e. duck typing).

As it's a stub, I'd include the `...` as in my comment earlier.
`_EXC_SENTINEL` will need to be moved to a separate variable or something.

webknjaz: Ah, so just

```suggestion
class ErrorableProtocol(Protocol):
    _EXC_SENTINEL = BaseException()

    def set_exception(
        self,
        exc: BaseException,
        exc_cause: BaseException = _EXC_SENTINEL,
    ):
        raise NotImplementedError


def set_exception(
    fut: "asyncio.Future[_T] | ErrorableProtocol",
```

Is the existence of `_EXC_SENTINEL` on the class going to be a problem?

Dreamsorcerer: > Is the existence of `_EXC_SENTINEL` on the class going to be a problem?

Yep, it'll be part of the stub it's looking for. I don't think there's any need to have it as part of a class.

aiohttp/client_proto.py
```diff
@@ -4,13 +4,15 @@
...
 from .client_exceptions import (
+    ClientConnectionError,
     ClientOSError,
     ClientPayloadError,
     ServerDisconnectedError,
     SocketTimeoutError,
 )
 from .helpers import (
     BaseTimerContext,
+    ErrorableMixin,
```
webknjaz:
```suggestion
    ErrorableProtocol,
```

aiohttp/client_proto.py
```diff
@@ -128,10 +160,14 @@ def resume_reading(self) -> None:
         super().resume_reading()
         self._reschedule_timeout()
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+    exc: BaseException,
+    exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
```suggestion
    exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+
+    If the future is marked as complete, this function is a no-op.
+
+    :param exc_cause: An exception that is a direct cause of ``exc``.
+                      Only set if provided.
+    """
+    if asyncio.isfuture(fut) and fut.done():
+        return
+
+    exc_is_sentinel = exc_cause is ErrorableMixin._EXC_SENTINEL
```
```suggestion
    exc_is_sentinel = exc_cause is ErrorableProtocol._EXC_SENTINEL
```

aiohttp/http_parser.py
```diff
@@ -31,7 +31,9 @@
     DEBUG,
     NO_EXTENSIONS,
     BaseTimerContext,
+    ErrorableMixin,
```
webknjaz:
```suggestion
    ErrorableProtocol,
```

aiohttp/http_parser.py
```diff
@@ -929,8 +934,12 @@ def __init__(self, out: StreamReader, encoding: Optional[str]) -> None:
         else:
             self.decompressor = ZLibDecompressor(encoding=encoding)
 
-    def set_exception(self, exc: BaseException) -> None:
-        self.out.set_exception(exc)
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -14,7 +14,13 @@
 )
 
 from .base_protocol import BaseProtocol
-from .helpers import BaseTimerContext, TimerNoop, set_exception, set_result
+from .helpers import (
+    BaseTimerContext,
+    ErrorableMixin,
```
webknjaz:
```suggestion
    ErrorableProtocol,
```

aiohttp/streams.py
webknjaz:
```suggestion
class StreamReader(AsyncStreamReaderMixin):
```

aiohttp/streams.py
```diff
@@ -146,19 +152,23 @@ def get_read_buffer_limits(self) -> Tuple[int, int]:
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -499,7 +509,11 @@ def __repr__(self) -> str:
     def exception(self) -> Optional[BaseException]:
         return None
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -551,7 +565,7 @@ def read_nowait(self, n: int = -1) -> bytes:
 EMPTY_PAYLOAD: Final[StreamReader] = EmptyStreamReader()
 
 
-class DataQueue(Generic[_T]):
+class DataQueue(Generic[_T], ErrorableMixin):
```
webknjaz:
```suggestion
class DataQueue(Generic[_T]):
```

aiohttp/streams.py
```diff
@@ -574,14 +588,18 @@ def at_eof(self) -> bool:
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableMixin._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```

aiohttp/helpers.py
webknjaz: I guess, I'll need to move it to the module level, then.

Dreamsorcerer: Yeah, most likely the class won't need to referenced anywhere else, so you'll still only be importing one thing.

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
         fut.set_result(result)
 
 
-def set_exception(fut: "asyncio.Future[_T]", exc: BaseException) -> None:
-    if not fut.done():
-        fut.set_exception(exc)
+class ErrorableProtocol(Protocol):
+    _EXC_SENTINEL = BaseException()
```
```suggestion
_EXC_SENTINEL = BaseException()


class ErrorableProtocol(Protocol):
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ):
+        raise NotImplementedError
+
+
+def set_exception(
+    fut: "asyncio.Future[_T] | ErrorableProtocol",
+    exc: BaseException,
+    exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```
```suggestion
    exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,38 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+
+    If the future is marked as complete, this function is a no-op.
+
+    :param exc_cause: An exception that is a direct cause of ``exc``.
+                      Only set if provided.
+    """
+    if asyncio.isfuture(fut) and fut.done():
+        return
+
+    exc_is_sentinel = exc_cause is ErrorableProtocol._EXC_SENTINEL
```
```suggestion
    exc_is_sentinel = exc_cause is _EXC_SENTINEL
```

aiohttp/client_proto.py
```diff
@@ -4,13 +4,15 @@
...
 from .client_exceptions import (
+    ClientConnectionError,
     ClientOSError,
     ClientPayloadError,
     ServerDisconnectedError,
     SocketTimeoutError,
 )
 from .helpers import (
     BaseTimerContext,
+    ErrorableProtocol,
```
webknjaz:
```suggestion
    _EXC_SENTINEL,
    BaseTimerContext,
```

aiohttp/client_proto.py
```diff
@@ -128,10 +160,14 @@ def resume_reading(self) -> None:
         super().resume_reading()
         self._reschedule_timeout()
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/http_parser.py
```diff
@@ -31,7 +31,9 @@
     DEBUG,
     NO_EXTENSIONS,
     BaseTimerContext,
+    ErrorableProtocol,
```
webknjaz:
```suggestion
    _EXC_SENTINEL,
    BaseTimerContext,
```

aiohttp/http_parser.py
```diff
@@ -929,8 +934,12 @@ def __init__(self, out: StreamReader, encoding: Optional[str]) -> None:
         else:
             self.decompressor = ZLibDecompressor(encoding=encoding)
 
-    def set_exception(self, exc: BaseException) -> None:
-        self.out.set_exception(exc)
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -14,7 +14,13 @@
 )
 
 from .base_protocol import BaseProtocol
-from .helpers import BaseTimerContext, TimerNoop, set_exception, set_result
+from .helpers import (
+    BaseTimerContext,
+    ErrorableProtocol,
```
webknjaz:
```suggestion
    _EXC_SENTINEL,
    BaseTimerContext,
```

aiohttp/streams.py
```diff
@@ -146,19 +152,23 @@ def get_read_buffer_limits(self) -> Tuple[int, int]:
     def exception(self) -> Optional[BaseException]:
         return self._exception
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -574,14 +588,18 @@ def at_eof(self) -> bool:
...
```
webknjaz:
```suggestion
        exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/streams.py
```diff
@@ -499,7 +509,11 @@ def __repr__(self) -> str:
     def exception(self) -> Optional[BaseException]:
         return None
 
-    def set_exception(self, exc: BaseException) -> None:
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ErrorableProtocol._EXC_SENTINEL,
```
webknjaz:
```suggestion
        exc_cause: BaseException = _EXC_SENTINEL,
```

aiohttp/helpers.py
webknjaz: Yep. Seems to work.

aiohttp/client_proto.py
```diff
@@ -80,41 +82,71 @@ def is_connected(self) -> bool:
...
+        original_connection_error = exc
+        reraised_exc = original_connection_error
+
+        connection_closed_cleanly = original_connection_error is None
+        """Whether the connection got a clean EOF."""
+
+        if connection_closed_cleanly:
             set_result(self.closed, None)
+        else:
+            set_exception(
```
webknjaz, line 95:
```suggestion
            assert original_connection_error is not None
            set_exception(
```

aiohttp/client_proto.py
```diff
@@ -80,41 +82,72 @@ def is_connected(self) -> bool:
...
             # we do it anyway below
-            self.set_exception(exc)
+            set_exc_kwargs = (
+                {}
+                if connection_closed_cleanly
+                else {
+                    "exc_cause": original_connection_error,
+                }
+            )
+            assert reraised_exc is not None
```
webknjaz, line 140:
```suggestion
            assert original_connection_error is not None
            assert reraised_exc is not None
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,40 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+_EXC_SENTINEL = BaseException()
+
+
+class ErrorableProtocol(Protocol):
+
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ):
```
```suggestion
    ) -> None:
```

aiohttp/helpers.py
```diff
@@ -797,9 +797,40 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+
+
+class ErrorableProtocol(Protocol):
+
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = _EXC_SENTINEL,
+    ) -> None:
+        raise NotImplementedError
```
```suggestion
        exc_cause: BaseException = ...,
    ) -> None:
        ...
```

You probably missed my comment as you posted a second after, but this is a stub.
https://github.com/aio-libs/aiohttp/pull/8089#discussion_r1469763839

webknjaz: Ah, I wasn't sure that was mandatory.

Dreamsorcerer: Well, no, but it makes it clearer at a glance and avoids maintenance of things which are literally ignored and never run. The _EXC_SENTINEL is certainly incorrect, as no implementation of this method is defaulting to that value (but, as I say it's ignored by mypy as it's nothing to do with the type).

aiohttp/client_proto.py
```diff
@@ -80,41 +82,73 @@ def is_connected(self) -> bool:
...
+            set_exc_kwargs = (
+                {}
+                if connection_closed_cleanly
+                else {
+                    "exc_cause": original_connection_error,
+                }
+            )
+            assert original_connection_error is not None
+            assert reraised_exc is not None
+            self.set_exception(reraised_exc, **set_exc_kwargs)
```
webknjaz:
```suggestion
            underlying_exc = _EXC_SENTINEL if connection_closed_cleanly else original_connection_error
            assert underlying_exc is not None
            assert reraised_exc is not None
            self.set_exception(reraised_exc, underlying_exc)
```

aiohttp/helpers.py
webknjaz: > The _EXC_SENTINEL is certainly incorrect, as no implementation of this method is defaulting to that value (but, as I say it's ignored by mypy as it's nothing to do with the type).

That's not the case — I explicitly import that sentinel and use it as the default in all the signatures. This is the only use for that import outside this module.

aiohttp/client_proto.py
```diff
@@ -80,41 +82,67 @@ def is_connected(self) -> bool:
...
+                ),
+                original_connection_error,
+            )
 
         if self._payload_parser is not None:
-            with suppress(Exception):
+            with suppress(Exception):  # FIXME: log this somehow?
                 self._payload_parser.feed_eof()
 
         uncompleted = None
```
webknjaz, line 107:
```suggestion
        uncompleted = None
        underlying_exc: BaseException
```

aiohttp/helpers.py
Dreamsorcerer: Ah, yes, I didn't recheck the diff, thought the methods were left unchanged.

Still, it'd look a little odd to me for a stub. Up to you.

webknjaz: @Dreamsorcerer if we have the sentinel default value there in the stub, would MyPy enforce that the method implementations also have the same default value?

aiohttp/client_proto.py
```diff
@@ -80,41 +82,68 @@ def is_connected(self) -> bool:
...
+        underlying_exc: BaseException
```
webknjaz:
```suggestion
```

aiohttp/helpers.py
Dreamsorcerer: No, I'm pretty sure mypy completely ignores the values, as it's not typing information. The only benefit might be in IDEs showing the default value, which is why typeshed have started including simple defaults. But, I don't think that's relevant to a Protocol.

aiohttp/client_proto.py
```diff
@@ -80,41 +82,68 @@ def is_connected(self) -> bool:
...
-                exc = ServerDisconnectedError(uncompleted)
+            if isinstance(original_connection_error, OSError):
+                reraised_exc = ClientOSError(*original_connection_error.args)
+            if connection_closed_cleanly:
+                reraised_exc = ServerDisconnectedError(uncompleted)
             # assigns self._should_close to True as side effect,
             # we do it anyway below
-            self.set_exception(exc)
+            underlying_exc = _EXC_SENTINEL if connection_closed_cleanly else original_connection_error
+            assert underlying_exc is not None
```
webknjaz:
```suggestion
            underlying_non_eof_exc = _EXC_SENTINEL if connection_closed_cleanly else original_connection_error
            assert underlying_non_eof_exc is not None
```

aiohttp/client_proto.py
```diff
@@ -80,41 +82,68 @@ def is_connected(self) -> bool:
...
+            assert reraised_exc is not None
+            self.set_exception(reraised_exc, underlying_exc)
```
webknjaz:
```suggestion
            self.set_exception(reraised_exc, underlying_non_eof_exc)
```

aiohttp/client_proto.py
```diff
@@ -80,41 +82,67 @@ def is_connected(self) -> bool:
...
         self._drop_timeout()
 
-        if exc is not None:
-            set_exception(self.closed, exc)
-        else:
+        original_connection_error = exc
+        reraised_exc = original_connection_error
+
+        connection_closed_cleanly = original_connection_error is None
+        """Whether the connection got a clean EOF."""
```
Dreamsorcerer: Should this be a regular comment? Seems odd having a multiline str.

webknjaz: It's a variant of an attribute docstring for the variable above ([PEP 258](https://peps.python.org/pep-0258/#attribute-docstrings)). Sphinx recognizes these in some contexts. I like using these to explain the data, as in "_what_ a thing represents". The comments, OTOH, have different semantics — they are typically used to explain the unobvious motivation, as in "_why_" (although they are indeed often confused with docstrings).

webknjaz: But maybe a shorter comment would be fine too. I'll think about it.

webknjaz: I hope the variable name conveys the semantic meaning well enough. But it's unobvious that `None` is a sentinel for "the protocol got EOF as expected".

aiohttp/helpers.py
webknjaz: It'd be nice to enforce this, I suppose. But alright, I'll use the ellipsis.

aiohttp/client_proto.py
Dreamsorcerer: Right, although that PEP is rejected, and upon reading it, this isn't a valid attribute docstring as defined there either:

> The assignment must be in one of the following contexts:
> 
>     At the top level of a module (i.e., not nested inside a compound statement such as a loop or conditional): a module attribute.
>     At the top level of a class definition: a class attribute.
>     At the top level of the “__init__” method definition of a class: an instance attribute. Instance attributes assigned in other methods are assumed to be implementation details. (@@@ __new__ methods?)
>     A function attribute assignment at the top level of a module or class definition.

Dreamsorcerer: > I hope the variable name conveys the semantic meaning well enough

Yes, I'd say the variable name is pretty clear, probably don't need the comment at all.

Dreamsorcerer:
Would you like me to hold off from doing a 3.9.3 release for a bit? Would like to start on that in the next hour or so, to make sure people can upgrade shortly.

webknjaz:
@Dreamsorcerer don't wait up. I won't be merging this w/o proper reviews from you and @bdraco. It's a PoC that also changes a few exceptions along the way so we need to make sure it's well-understood first.

aiohttp/client_proto.py
webknjaz:
```suggestion
```

aiohttp/client_proto.py
```diff
@@ -234,14 +271,22 @@ def data_received(self, data: bytes) -> None:
...
                     # should_close is True after the call
-                    self.set_exception(exc)
+                    reraised_exc = ClientPayloadError(
+                        f"Unable to parse response payload: {underlying_exc !r}"
+                    )
+                    if isinstance(
+                        underlying_exc,
+                        (ClientError, HttpProcessingError),
+                    ):
+                        reraised_exc = underlying_exc
```
webknjaz: @Dreamsorcerer I did this to fix the tests. But what do you think about changing this to a different error unconditionally?

I'm a bit concerned about the changes in the PR influencing the end-users...

webknjaz:
@bdraco waiting for your opinion now

aiohttp/helpers.py
```diff
@@ -797,9 +797,39 @@ def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
...
+_EXC_SENTINEL = BaseException()
+
+
+class ErrorableProtocol(Protocol):
+    def set_exception(
+        self,
+        exc: BaseException,
+        exc_cause: BaseException = ...,
+    ) -> None:
+        ...
```
```suggestion
        ...  # pragma: no cover
```

aiohttp/client_proto.py
Dreamsorcerer: We've already been told by users to stop throwing exceptions that are not inhereting from ClientError. They expect to be able to catch all exceptions with that, so this is probably the desired behaviour.

webknjaz: Yes, but this hack makes the underlying error leak/bubble up directly to the end users, which means the traceback won't have this layer, right?

aiohttp/client_proto.py
```diff
@@ -234,14 +271,22 @@ def data_received(self, data: bytes) -> None:
...
                     # should_close is True after the call
-                    self.set_exception(exc)
+                    reraised_exc: BaseException = ClientPayloadError(
+                        f"Unable to parse response payload: {underlying_exc !r}"
+                    )
+                    if isinstance(
+                        underlying_exc,
+                        (ClientError, HttpProcessingError),
+                    ):
+                        reraised_exc = underlying_exc
```
bdraco:
```suggestion
                    reraised_exc = underlying_exc
                    if not isinstance(
                        underlying_exc,
                        (ClientError, HttpProcessingError),
                    ):
                        reraised_exc = ClientPayloadError(
                            f"Unable to parse response payload: {underlying_exc !r}"
                        )
```

trivial: I'd probably reverse this so it only reconstructs in the case where it doesn't match

aiohttp/client_reqrep.py
bdraco: We set `__context__` before `__cause__` before so changing the order might change the behavior here

Need to dig through https://github.com/python/cpython/blob/dc01b919c721f43ad024ba444a5d19541370e581/Objects/exceptions.c#L360

bdraco:
```
/* Steals a reference to cause */
void
PyException_SetCause(PyObject *self, PyObject *cause)
{
    PyBaseExceptionObject *base_self = _PyBaseExceptionObject_cast(self);
    base_self->suppress_context = 1;
    Py_XSETREF(base_self->cause, cause);
}
```

bdraco: Order doesn't matter, but setting `__context__` is likely useless here because it gets suppressed anyways

aiohttp/client_proto.py
```diff
@@ -80,41 +84,70 @@ def is_connected(self) -> bool:
...
-        else:
+        original_connection_error = exc
+        reraised_exc = original_connection_error
+
+        connection_closed_cleanly = original_connection_error is None
+
+        if connection_closed_cleanly:
             set_result(self.closed, None)
+        else:
+            assert original_connection_error is not None
```
bdraco, line 94: This looks like its only for typing. Usually I guard these with `TYPE_CHECKING` but thats just personal preference as I don't think anyone runs with `-O` here

aiohttp/helpers.py
bdraco: I think with `__cause__` set you should have everything that is needed to generate the full traceback at the point the exception is consumed

bdraco:
Looks good.  Only some minor comments.  I can throw it on production once its near ready to merge and look for obvious regressions

aiohttp/client_proto.py
Dreamsorcerer: I think it'd be preferable to just recommend `-O` in production environments so it has no impact. We have asserts like this elsewhere in the code too.

aiohttp/client_proto.py
Dreamsorcerer: I'm not sure, but I suspect it won't make a noticeable difference to users. Maybe just skip the 3.9 backport to minimise impact?

aiohttp/client_proto.py
webknjaz: Yep. I think that sticking `TYPE_CHECKING` everywhere is a burden we don't want to introduce all over. It's indeed easier to recommend doing `-O`.

aiohttp/client_proto.py
webknjaz: Fair. I tend to want less indentation so I optimize for readability. Though, your variant seems more efficient.

aiohttp/client_reqrep.py
webknjaz: So both `__context__` *and* `__cause__` have been first introduced almost a decade ago in https://github.com/aio-libs/aiohttp/commit/20ee192f6a92e73d6c96289ba5768839ded8a06d. The commit is not coming from a PR and is title `Support exception chaining for ClientRequestError classes`. It doesn't contain a longer description offering any explanation beyond that. The comments discuss the usefulness of the `raise ... from ...` syntax, which is not related to our context.

webknjaz: @bdraco @Dreamsorcerer do you think it's fine if I delete this assignment within this PR?

aiohttp/helpers.py
webknjaz: I'll keep it as is. @Dreamsorcerer please, mark the thread as resolved if you don't have any objections.

Dreamsorcerer: It's certainly not a blocker, but might be something we want to come back to, if it doesn't produce tracebacks.

aiohttp/client_proto.py
webknjaz: @Dreamsorcerer do you mean keep this the logic in the PR as it is now, backport it to 3.9 and 3.10, and only change the logic in a separate PR targeting 3.10?
I definitely want to use this patch to discover what's happening with my github bots with aiohttp 3.9 and not wait till 3.10.

I'll see if there's a meaningful way to make a patch that migrates to throwing something inheriting from `ClientError` while declaring it a bugfix.

@bdraco do you have any thoughts on this?

Dreamsorcerer: I don't mind either way. If you think it's likely to cause some minor breakage, it'd make sense to delay it to 3.10. But, I have no feeling for this, so I'm fine with it going into 3.9 if you are.

bdraco: I think it seems reasonable as a bugfix to make it always inherit from `ClientError`.  I expect consumers already expect that given https://github.com/aio-libs/aiohttp/pull/8089/files#r1471738716

aiohttp/client_proto.py
webknjaz: Marking as resolved.

aiohttp/client_reqrep.py
webknjaz: Deleted. Marking as resolved.

aiohttp/client_proto.py
webknjaz: So it occurred to me that the protocol is an internal layer, probably. And maybe, it should keep the internal exceptions as is with them being re-raised on the outer layer. Though, I'm yet to look deeper into it — this is the last thing keeping me from merging the patch.

aiohttp/client_proto.py
```diff
@@ -234,14 +271,14 @@ def data_received(self, data: bytes) -> None:
...
-                except BaseException as exc:
+                except BaseException as underlying_exc:
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
                     # should_close is True after the call
-                    self.set_exception(exc)
+                    self.set_exception(HttpProcessingError(), underlying_exc)
```
webknjaz, line 280: @Dreamsorcerer @bdraco I made this unconditional but turned it into a different error, because this turned out to be the wrong level of abstraction. It seems to be turned into what makes the tests pass in the end.

webknjaz:
@bdraco I'm triggering a release candidate via https://github.com/aio-libs/aiohttp/actions/runs/7926204466 — you'll be able to deploy this pre-release for testing before cutting a stable version.

webknjaz:
@bdraco I ended up having to re-trigger the release on another commit: https://github.com/aio-libs/aiohttp/actions/runs/7930478394/job/21655287776.

It's now available via https://pypi.org/project/aiohttp/3.9.4rc0/.

cc @Dreamsorcerer FYI

webknjaz:
I'm also deploying this into the Chronographer and Patchback bots now, that have Sentry hooked up. This should help with figuring out #4581 since Sentry alerts with that arrive every day and it's quite annoying.

bdraco:
I had to revert back right away when testing on production do to
```
2024-02-16 09:45:14.978 ERROR (MainThread) [pyunifiprotect.websocket] Websocket disconnect error: %s
  File "/usr/local/lib/python3.11/site-packages/pyunifiprotect/websocket.py", line 95, in _websocket_loop
    self._ws_connection = await session.ws_connect(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 884, in _ws_connect
    resp = await self.request(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/aiohttp/client.py", line 472, in _request
aiohttp.client_exceptions.NonHttpUrlClientError: wss://192.168.209.164/proxy/protect/ws/updates?lastUpdateId=80f54055-84de-4e15-95d7-80cac46aafbd
```

Dreamsorcerer:
> I had to revert back right away when testing on production due to:

I thought that PR was only merged into 3.10?

bdraco:
> > I had to revert back right away when testing on production due to:
> 
> I thought that PR was only merged into 3.10?

It did. Sorry for the lack of clarity. I pulled 3.10 on since that's what I've been running with as the staging production env is making use of 3.10 features. I'm going to pull it back down to 3.9 later today when I have free cycles

bdraco:
I've managed to get the smaller one downgraded to 3.9 (3.9.4rc0). So far so good.

I'm working on the larger one

webknjaz:
Great! I hope I'll get some tracebacks from the bots too, maybe.

Issue #6031, opened 2021-10-02
Recover non-running pretend tests in `tests/test_proxy_functional.py`


webknjaz:
### Describe the bug

While working on #6002 and #5992, I noticed several functions that are prefixed with `xtest_` and are marked with `@pytest.mark.xfail`. Clearly @asvetlov wanted them to run but not influence the test session outcome. But pytest only picks up functions marked as `test_` (by default) as tests so that never worked or was executed.

We need to drop that `x` and let them run. Also, we need to audit them to see if they are still relevant to the today's codebase. It sounds like they were added with these names originally back in 2017. And some of them were added earlier that year prefixed with `_test_` (which wouldn't work either). The `@pytest.mark.xfail` decorator was added later (also the same year) but the names were never corrected so it didn't have any effect. Further archeology revealed that it was @fafhrd91 who changed `test_` to `_test_` (in Feb 2017) and before that the tests had the proper names and the regular `xfail` mark. And the original tests (that existed at the time) were added in 2016 by @kserhii.

- https://github.com/aio-libs/aiohttp/commit/0958dde9c4aa34b057d50a2a9186baaf0e3c54d5#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R545
- https://github.com/aio-libs/aiohttp/commit/4f603349020d670d087e84ea2f85bf881313608f#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R261
- https://github.com/aio-libs/aiohttp/commit/adf00e73b6ee9c5cc4910e1e526a16d9f54f142c#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R280

There's no record of why those tests were disabled and forgotten so I'm posting these findings here for history.

### To Reproduce

Run `pytest -q --no-cov tests/test_proxy_functional.py --collect-only` and see that none of the functions that start with `xtest_` are listed there.

### Expected behavior

The functions in question are also marked with `@pytest.mark.xfail` meaning that they are perceived by the humans as tests but because of a small oversight they are never picked up and run by pytest.

### Logs/tracebacks

```python-traceback
N/A
```

### Python Version

```console
N/A
```

### aiohttp Version

```console
master branch
```

### multidict Version

```console
N/A
```

### yarl Version

```console
N/A
```

### OS

N/A

### Related component

Client

[...]

PR #1626, opened 2017-02-14
http parser implementation based on http-parser 

fafhrd91:
added c-based http parser, cython code is based on https://github.com/MagicStack/httptools

it gives ~15-25% increase

PR #1421, opened 2016-11-23
Proxy fix

dodge-ksv:
## What do these changes do?

 * Adds functional proxy tests (proxy_test_server created)
 * Fixes client proxy bugs: target path and proxy host header with port

## Are there changes in behavior for the user?

HTTPS proxy connection has been fixed.

## Related issue number

 * #1218 
 * #1413 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] Add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new entry to `CHANGES.rst`
  * Choose any open position to avoid merge conflicts with other PRs.
  * Add a link to the issue you are fixing (if any) using `#issue_number` format at the end of changelog message. Use Pull Request number if there are no issues for PR or PR covers the issue only partially.

PR #8096, opened 2024-01-29
Remove unused readall from Python parser

Dreamsorcerer:
Here's a test that hasn't been run, ever.


Dreamsorcerer:
Huh, seems we have a parameter which only exists on the Python parser...

webknjaz:
I think I saw a few other underscored tests somewhere around the proxy testing modules, IIRC. I don't think I ever got to re-enabling them..

webknjaz:
I think we should be backporting the test-only changes all the way to the oldest stable we released recently. This helps with future backports, reducing merge conflicts, while not having any effect on the SemVer considerations, because they aren't user-facing.
The other day, I spent hours resolving merge conflicts because of the test refactoring from 2019 that hasn't been backported back then but is super useful.

webknjaz:
@Dreamsorcerer here's some more: https://github.com/aio-libs/aiohttp/blob/28b574f/tests/test_proxy_functional.py#L461 — just search `def xtest` in that file to find all 10.

Interestingly, they were disabled in https://github.com/aio-libs/aiohttp/pull/1626#discussion_r101150685 because a maintainer didn't understand why a contributor added them in the first place (https://github.com/aio-libs/aiohttp/pull/1421/files#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R120). They were then renamed without re-enabling by another maintainer: https://github.com/aio-libs/aiohttp/pull/2256/files#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R279.

webknjaz:
@Dreamsorcerer :man_facepalming: I've actually found an issue where I documented this precisely: https://github.com/aio-libs/aiohttp/issues/6031.

Dreamsorcerer:
> @Dreamsorcerer 🤦‍♂️ I've actually found an issue where I documented this precisely: #6031.

Yeah, figured you'd looked at them before. I've no plans to look at the proxy stuff, but fixing the currently failing tests on Windows/Mac with Python 3.12 would be the first priority, before looking at these ones.

webknjaz:
Fair, I don't think I'll be looking into them anytime soon. But it's good that they are documented.

Dreamsorcerer:
OK, can someone check this, I think the parameter is totally useless and can be removed.
The only place the parameter is used is at https://github.com/aio-libs/aiohttp/blob/9736477ea0387fb9f1e626870894d6cc79bf47c2/aiohttp/http_parser.py#L377

After passing this check:
https://github.com/aio-libs/aiohttp/blob/9736477ea0387fb9f1e626870894d6cc79bf47c2/aiohttp/http_parser.py#L359-L363

The parameter passed in then used at the end here:
https://github.com/aio-libs/aiohttp/blob/9736477ea0387fb9f1e626870894d6cc79bf47c2/aiohttp/http_parser.py#L746-L762

But, if it passed the check above, then it must not be able to reach the code where the variable is used, if I'm reading that right.

Dreamsorcerer:
If correct, it must then be able to be removed from the latter code too, as the only other code that uses that parameter is hardcoded to `True`, so anytime that code is reached it must be `True`.

Dreamsorcerer:
> OK, can someone check this, I think the parameter is totally useless and can be removed.

@bdraco @webknjaz I think these changes are correct, all of that code seems useless and can never happen (except by interacting with HttpPayloadParser directly, as the deleted test was doing).

aiohttp/http_parser.py
```diff
@@ -757,16 +749,8 @@ def __init__(
...
-            if readall and code != 204:
-                self._type = ParseState.PARSE_UNTIL_EOF
-            elif method in ("PUT", "POST"):
-                internal_logger.warning(  # pragma: no cover
-                    "Content-Length or Transfer-Encoding header is required"
-                )
-                self._type = ParseState.PARSE_NONE
-                real_payload.feed_eof()
-                self.done = True
+        elif code != 204:
```
bdraco: I'm not sure we can get here if the code is 204

bdraco: 204 will always be empty body and the only place that can get in here is 

https://github.com/aio-libs/aiohttp/pull/8096/files#diff-197153342549adb4b25580f8b842876841eec42dc48ae12cb72ba2ae4f7ced10R380

AFAICT

Dreamsorcerer: I think if it's CONNECT and 204 it might reach here, which is why I wasn't sure about removing the check. Not sure if that's actually relevant or not...

Dreamsorcerer: Maybe we should stop CONNECT getting in here too? In which case, maybe we can refactor the code to avoid the duplicate `not empty_body` checks.

Dreamsorcerer: https://www.rfc-editor.org/rfc/rfc9110#section-9.3.6-7
> Any [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110#status.2xx) response indicates that the sender (and all inbound proxies) will switch to tunnel mode immediately after the response header section; data received after that header section is from the server identified by the request target. Any response other than a successful response indicates that the tunnel has not yet been formed.

> A CONNECT request message does not have content. The interpretation of data sent after the header section of the CONNECT request message is specific to the version of HTTP in use.


I think if I'm reading that correctly, then we should read the body regardless of whether it's 204 or not. So, I think we want to remove the check and always read?

bdraco: Anything after the response is going to be the proxied socket.  After the response headers I think we are only forwarding the stream.  I haven't dug through the paths to figure out how that works yet, but it seems like we shouldn't read the body and at that point its effectively forwarding read/writes of a stream?

Dreamsorcerer: So, maybe the body is supposed to be read when it's not a 2xx response? I'm not clear why this code seems to always try to read the response when the method is CONNECT..

webknjaz:
I'm not sure how I feel about the removal. On one hand, it's an unused private API. And on the other — `readall()` seem customary  for streaming APIs.
I suppose, it's fine to drop it and recover later as needed.
Though, maybe we'd just make use of it instead, implementing it for Cython too? Thoughts?

webknjaz:
Another thought: since it's used in some payload tests, they might rely on the underlying buffers being drained automatically or something like that. Could you check? I wouldn't be surprised if this thing exists as a test helper.

aiohttp/http_parser.py
bdraco: I was a bit worried that someone has a use case for doing CONNECT and than speaking some completely different protocol. AFAICT the only use case for CONNECT is for proxies. 

https://github.com/aio-libs/aiohttp/blob/0ec65c0f4dc08d027f659256b09ae9cff10ab404/aiohttp/connector.py#L1263

I think this change looks good as we are never going to get in to `HttpPayloadParser` on empty body responses so we only have to worry about it on `METH_CONNECT` and it seems like the logic remains unchanged.

Dreamsorcerer:
> `readall()` seem customary for streaming APIs.

We're removing a parameter called readall, not a method. I think this is also unrelated to streaming APIs, as the parameter is basically just saying "read until EOF when there is no length/chunked specified", which I'm pretty sure is the only sensible (i.e. spec-compliant) approach if the response is not 204 (or similar).

The only place where it ever gets set to False is when we've already determined that that the response has a body (i.e. not 204).

tests/test_http_parser.py
```diff
@@ -1205,8 +1205,8 @@ def test_parse_chunked_payload_chunk_extension(parser: Any) -> None:
     assert payload.is_eof()
 
 
-def _test_parse_no_length_or_te_on_post(loop, protocol, request_cls):
-    parser = request_cls(protocol, loop, readall=True)
+def test_parse_no_length_or_te_on_post(loop: Any, protocol: Any, request_cls: Any):
+    parser = request_cls(protocol, loop, 2**16)
```
webknjaz: Could you use a named kwarg so it's visible what this arbitrary constant stands for?

aiohttp/http_parser.py
webknjaz: Plot twist: I did some Git paleontology and it turned out that the patch implementing this bit of logic is the same commit that marked the exactly same proxy tests that I mentioned earlier in this PR (https://github.com/aio-libs/aiohttp/pull/8096#issuecomment-1914876703) with xfail. See https://github.com/aio-libs/aiohttp/commit/b26702449142733caa0e026b331381bf5bee4f01#diff-be15bec475f6c781d5d3e4413c19d0e24ccaf616b8a2495c6824fc3a16d139f5R270.

Does this mean that the original implementer wasn't sure the implementation was correct 7 years ago? Will those tests miraculously start passing if this is removed?

Dreamsorcerer: > Will those tests miraculously start passing if this is removed?

I don't think we're making any real change to behaviour, so no, I don't think they'll start passing. I'd note that someone renamed the tests to `xtest` though, so they are currently not running at all. The test you linked to still fails when I run locally with: `ClientConnectorError: Cannot connect to host www.google.com.ua:443 ssl:default [None]`

PR #8103, opened 2024-01-29
Do not merge

Dreamsorcerer:


Dreamsorcerer:
Now I seem to have a pypy failure for no apparent reason. No tests failed and it still exited with an error...

webknjaz:
> Now I seem to have a pypy failure for no apparent reason. No tests failed and it still exited with an error...

@Dreamsorcerer here's the failure: https://github.com/aio-libs/aiohttp/actions/runs/7702167031/job/20989877407#step:11:68.

Explanation: in the test jobs, we have 2 subsequent test invocations — a normal one and a "debug" one. The latter is very verbose and only runs if the former failed. Moreover, the re-run only restarts the failed tests.
In the log, when the first test run fails, the step is marked red. But if the re-run were to succeed, it'd look green, confusingly. This is why, the re-run has an `exit 1` injected, making sure that the step is marked as failing regardless of the second pytest run. GitHub also unfolds that step automatically, when you open the logs.

So the initial run crashed with `E   TypeError: expected some sort of stmt, got <ast.Name object at 0x00007f3a11981d60>` during the test collection/discovery stage, essentially leaving pytest with no tests marked as successful.
Because of that, the re-run invoked all the tests (which is the behavior for when there's no tests from the previous run marked as failed). That pytest run was a separate PyPy invocation in a separate process and wasn't affected by the flaky behavior that hit the first run. So all the tests succeeded this time. The output was green. But that `exit 1` kicked in to make sure the step is marked as red. This is quite intentional. Otherwise, the CI would hide flaky tests that succeed on re-run from us.

PR #8136, opened 2024-02-05
Create hello.txt.gz dynamically and improve related assertions

steverep:
## What do these changes do?

- Adds a fixture to create `hello.txt*` dynamically to help satisfy the edict from https://github.com/aio-libs/aiohttp/pull/8063#discussion_r1468952020
- Rearranges response assertions to be more useful if tests fail (e.g. if content is wrong, it's probably because encoding is wrong, so check that first)

This is separate so it can be fully backported.

No

No

## Related issue number

No

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


steverep:
What's with the random seg faults on PyPy?

Dreamsorcerer:
> What's with the random seg faults on PyPy?

No idea, been doing for about 3 months now and bumping to a newer version didn't help either...

CHANGES/8136.misc.rst
```diff
@@ -0,0 +1 @@
+Created ``hello.txt.gz`` dynamically and improved related assertions -- by :user:`steverep`.
```
webknjaz:
```suggestion
Started dynamically producing ``hello.txt.gz`` during testing and improved related assertions -- by :user:`steverep`.
```

CHANGES/8136.misc.rst
webknjaz, line 1: This could be of the `contrib` type. @Dreamsorcerer WDYT? Alternatively, we might want to just skip change notes in the test refactoring patches. Your call. I'm good with keeping it as is too.

tests/test_web_sendfile_functional.py
```diff
@@ -16,6 +17,19 @@
     ssl = None
 
 
+HELLO_AIOHTTP = b"Hello aiohttp! :-)\n"
+
+
+@pytest.fixture(scope="module")
+def hello_txt(tmp_path_factory) -> pathlib.Path:
```
webknjaz: How about getting the suffix thorough an indirect fixture request?
```suggestion
def hello_txt(request, tmp_path_factory) -> pathlib.Path:
```
It's accessible here via `request.param`. And in the test method, it'd be set through `@pytest.mark.parametrize('hello_txt', ('txt', ), indirect=('hello_txt', ))`

Alternatively, it could be provided through the fixture params in the decorator and there wouldn't be any need for the test function-level parametrization. I.e. `@pytest.fixture(params=('txt', ), scope="module")`.

It could then just return `hello_gzip` directly, not forcing the test to reconstruct the path.

Although, the fixture could be renamed to a `compressed_file_path` so it's more generic.

steverep: Used an indirect parameter with default to return the plain text path.  That way only the tests that use compressed paths need to be touched.

CHANGES/8136.misc.rst
steverep: What's the verdict here?  I imagine test only PRs aren't super uncommon, so perhaps revise the template to clarify after the gavel is sounded.

Dreamsorcerer: Personally, I prefer to skip changes that don't affect users. But, I'll happily merge it either way.

bdraco: I agree on skipping in this case. Anything that we can do to reduce the amount of lines users have to read is a better chance they won't miss an important chnge.

steverep: I know I'm the outsider here, but I'll throw in my two 🪙 for the other team.  I'm against leaving anything out of release notes for open source projects - that's for proprietary s/w.  

The whole point of creating well categorized notes is to create a better user and/or developer experience.  The important stuff goes first and stuff only devs might want to know goes at the bottom, perhaps collapsed by default.  Users will read features, maybe bug fixes, docs if they have time, then skip the rest.  Some do read on though and as a dev it's always a headache to go fishing for commits that **aren't** in the notes because nothing there looks like the problem.

steverep: I converted this to contribution type because it seems a precedent for PRs that only affect testing is already being established: #8088, #8139, and #8140.

Hopefully this is ready to go now?

webknjaz: @Dreamsorcerer @bdraco so I agree with skipping things that aren't user or contributor-facing. But I also agree with @steverep — the contributor-facing changes should be recorded. That's *why* I made the changelog fragment types more granular in the first place. Perhaps I didn't communicate the intent well enough, though.

I feel like we need to clarify what's a change note vs. what's a git commit and make it clearer what goes where.
TL;DR — if a change affects end-users — it's a breaking change, a bugfix, a doc or a feature most of the time. If it doesn't but does affect the contributors — it's a contrib fragment. Otherwise, if it affects downstreams — it's packaging. For things we don't know where to put but feel like they're worth an honorary mention, there's misc.
Everything else could just have motivation/explanation in Git commits but not the release notes. Refactoring a test doesn't seem worthy a change log entry. Changing the way of running tests, introducing stricter linting or PR acceptance prerequisites should definitely be advertised.
Change notes are about the high-level effect on things respective audiences do. Git commit messages are for the devs, which might intersect with contributors but also they go into the motivation and unobvious quirks more, then merely high-level effects.

As many people struggle with keeping PRs atomic, a change note may be required for a portion of changes in a patch, but not others. If a PR refactors tests by reformatting a few lines, a Git commit is more than enough. If a PR refactors a test by introducing a reusable pytest fixture, especially available globally — the change log could advertise that fixture, showing the contributors/other devs why its useful and where to apply it. For that reason, if a contributor is not sure whether a note is needed, they can ask and the maintainers would decide whether to apply the label making bot ignore missing change notes, on case-by-case basis.

I tried to explain this more @ https://github.com/aio-libs/aiohttp/pull/8139#discussion_r1485707484.

CHANGES/8136.contrib.rst
```diff
@@ -0,0 +1 @@
+Started dynamically producing ``hello.txt.gz`` during testing and improved related assertions -- by :user:`steverep`.
```
webknjaz: With https://github.com/aio-libs/aiohttp/pull/8136#discussion_r1485724640 in mind. Let's try this:
```suggestion
A pytest fixture ``hello_txt`` was introduced to aid
static file serving tests in
:file:`test_web_sendfile_functional.py`. It dynamically
provisions ``hello.txt`` file variants shared across the
tests in the module.

-- by :user:`steverep`
```

Whether the assertions were improved or not is an implementation detail of refactoring that's not worth mentioning. It doesn't add clarity with the current wording anyway. A change note should make it clear how the information in it is useful to the reader (a contributor in this case).

webknjaz: (though, I'd argue that the assertions aren't really as related to the fixture as the PR title leads us to believe)

CHANGES/8136.contrib.rst
webknjaz, line 1: _(preview: https://aiohttp--8136.org.readthedocs.build/en/8136/changes.html#contributor-facing-changes)_

tests/test_web_sendfile_functional.py
```diff
@@ -266,35 +285,34 @@ async def handler(request):
...
-    body = await resp.read()
-    assert b"hello aiohttp\n" == body
-    ct = resp.headers["CONTENT-TYPE"]
-    assert "text/plain" == ct
-    encoding = resp.headers["CONTENT-ENCODING"]
-    assert "gzip" == encoding
+    assert resp.status == 200
+    assert resp.headers.get("Content-Encoding") == expect_encoding
+    assert resp.headers["Content-Type"] == "text/plain"
+    assert await resp.read() == HELLO_AIOHTTP
```
webknjaz, line 315: If you don't mind, I'd like to keep the comparisons inverted:
```suggestion
    assert 200 == resp.status
    assert expect_encoding == resp.headers.get("Content-Encoding")
    assert "text/plain" == resp.headers["Content-Type"]
    assert HELLO_AIOHTTP == await resp.read()
```
This helps avoid awkward keyword combos like `assert await`.

@Dreamsorcerer @bdraco any objections to having this style? I wasn't the one to introduce it but it seems like a good idea to follow it.

tests/test_web_sendfile_functional.py
```diff
@@ -240,24 +257,26 @@ async def handler(request):
...
 
 
-async def test_static_file_with_gziped_counter_part_enable_compression(
-    aiohttp_client: Any, sender: Any
+@pytest.mark.parametrize("forced_compression", [None, web.ContentCoding.gzip])
+async def test_static_file_with_encoding_and_enable_compression(
+    hello_txt: pathlib.Path,
+    aiohttp_client: Any,
+    sender: Any,
+    forced_compression: Optional[web.ContentCoding],
```
webknjaz, line 273: `Optional` is considered confusing, so it'd be nice not to introduce it in favor of being explicit:
```suggestion
    forced_compression: "web.ContentCoding | None",
```

webknjaz (approved):
I think, this patch is good, and addressing the weird `assert await` places and other comments is optional. I'll let somebody else merge, though, in case there's extra opinions.

tests/test_web_sendfile_functional.py
Dreamsorcerer: I find it slightly awkward, as I'm rather accustomed to the expected value being on the right. I don't really care much though.

tests/test_web_sendfile_functional.py
Dreamsorcerer: We shouldn't use the new syntax, it may cause a SyntaxError if inspected at runtime on Python 3.8/9. I'd just leave these alone and let pyupgrade handle them later. FYI, as far as I'm aware the recommendation was always to use `Optional` until it became deprecated. Now we're just waiting till we can use the new syntax.

webknjaz: @Dreamsorcerer that's not entirely accurate. There's a quirk (or two). By default, `web.ContentCoding | None` is problematic, yes. But `"web.ContentCoding | None"` is not.

From the CPython parser's perspective, it's just a string so it does nothing in runtime and you can put anything you want in there, and it won't explode.

When it is being processed by MyPy, though, it allows using newer syntax in the annotations and it works regardless of the target CPython version for as long as MyPy isn't ancient.

Moreover, if we were to put `from __future__ import annotations` at the top, the quotes would be unnecessary and `web.ContentCoding | None` would just work due to the evaluation being postponed as it turns typing definitions in such modules to strings in runtime.

I've been using the future + no quotes for quite a while without any problems. As an example, there's instances of this syntax in yarl and I set up MyPy to test typing against Python 3.7 there — https://results.pre-commit.ci/run/github/64756820/1706701102.QzIu9jkORbOBrIsSqNFdRg.

Dreamsorcerer: > From the CPython parser's perspective, it's just a string so it does nothing in runtime and you can put anything you want in there, and it won't explode.

When the function is parsed, yes. But, if you inspect the type with get_type() or similar, I think it may evaluate it then, or if using something like Pydantic, then it may try to eval the string if not. It's an edge case, but I don't see any reason to break it. I'm also not sure if pyupgrade would remove the quotes when they are no longer needed, but it'll definitely remove the Optional.

webknjaz: I see your point. Though, I think we should rely on what's tested, not on something some people remember to do occasionally. But fine, I won't insist.l

bdraco:
Tagging for backport since we want to pull the other PR on top of this and it will conflict otherwise

tests/test_web_sendfile_functional.py
bdraco: I'm in favor of `from __future__ import annotations` as much as possible as the new syntax is so much nicer and more readable. Thats probably for another PR though.

bdraco (approved):

tests/test_web_sendfile_functional.py
```diff
@@ -213,24 +233,21 @@ async def handler(request):
...
     await resp.release()
     await client.close()
 
 
 async def test_static_file_custom_content_type_compress(
-    aiohttp_client: Any, sender: Any
+    hello_txt: pathlib.Path, aiohttp_client: Any, sender: Any
 ):
-    filepath = pathlib.Path(__file__).parent / "hello.txt"
+    """Test that custom type with encoding is returned for unencoded request."""
```
bdraco:
```suggestion
    """Test that custom type with encoding is returned for unencoded requests."""
```

tests/test_web_sendfile_functional.py
```diff
@@ -266,35 +285,34 @@ async def handler(request):
...
 
+@pytest.mark.parametrize(
+    ("hello_txt", "expect_encoding"), [["gzip"] * 2], indirect=["hello_txt"]
+)
 async def test_static_file_with_content_encoding(
-    aiohttp_client: Any, sender: Any
+    hello_txt: pathlib.Path, aiohttp_client: Any, sender: Any, expect_encoding: str
 ) -> None:
-    filepath = pathlib.Path(__file__).parent / "hello.txt.gz"
+    """Test requesting of static compressed files returns correct content type and encoding."""
```
bdraco:
```suggestion
    """Test requesting static compressed files returns the correct content type and encoding."""
```

bdraco:
I think this looks good. Some trivial nits in the docstrings above, but otherwise looks good to merge from my perspective.

bdraco (approved):

bdraco (approved):

bdraco:
Thanks @steverep

bdraco:
Looks like this one will need some manual backports.  I was hoping it was going to merge back cleanly.

steverep:
> Looks like this one will need some manual backports. I was hoping it was going to merge back cleanly.

I don't see anything that would actually conflict so it's probably trivial.  I'll take care of it.

PR #8139, opened 2024-02-07
rename a shadowed test and re-enable F811 to catch future cases

alexmac:
## What do these changes do?

Small fix to un-shadow a test that wasn't being run

no

Depends, was there a reason F811 was set to ignore?

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/8139.bugfix.rst
+++ b/CHANGES/8139.bugfix.rst
@@ -1,0 +1,1 @@
+renamed a shadowed test and re-enabled F811 to catch future cases -- by :user:`alexmac`.
--- a/CHANGES/8139.bugfix.rst
+++ b/CHANGES/8139.bugfix.rst
@@ -1,1 +1,0 @@
-renamed a shadowed test and re-enabled F811 to catch future cases -- by :user:`alexmac`.
--- a/setup.cfg
+++ b/setup.cfg
@@ -86,7 +86,7 @@ zip_ok = false
 [flake8]
 extend-select = B950
 # TODO: don't disable D*, fix up issues instead
-ignore = N801,N802,N803,E203,E226,E305,W504,E252,E301,E302,E501,E704,W503,W504,F811,D1,D4
+ignore = N801,N802,N803,E203,E226,E305,W504,E252,E301,E302,E501,E704,W503,W504,D1,D4
 max-line-length = 88
 per-file-ignores =
     # I900: Shouldn't appear in requirements for examples.
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -1146,14 +1146,16 @@ def test_subapp_iter(app: Any) -> None:
     assert list(resource) == [r1, r2]
 
 
-def test_invalid_route_name(router: Any) -> None:
+@pytest.mark.parametrize(
+    "route_name",
+    [
+        "invalid name",
+        "class",
+    ],
+)
+def test_invalid_route_name(router: Any, route_name: str) -> None:
     with pytest.raises(ValueError):
-        router.add_get("/", make_handler(), name="invalid name")
-
-
-def test_invalid_route_name_identifier(router) -> None:
-    with pytest.raises(ValueError):
-        router.add_get("/", make_handler(), name="class")  # identifier
+        router.add_get("/", make_handler(), name=route_name)
 
 
 def test_frozen_router(router: Any) -> None:
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -1151,7 +1151,7 @@ def test_invalid_route_name(router: Any) -> None:
         router.add_get("/", make_handler(), name="invalid name")
 
 
-def test_invalid_route_name(router) -> None:
+def test_invalid_route_name_identifier(router) -> None:
     with pytest.raises(ValueError):
         router.add_get("/", make_handler(), name="class")  # identifier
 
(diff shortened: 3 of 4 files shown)


CHANGES/8139.bugfix.rst
```diff
@@ -0,0 +1 @@
+renamed a shadowed test and re-enabled F811 to catch future cases -- by :user:`alexmac`.
```
webknjaz: This needs to be more specific and include details useful to the contributors. Make it a sentence with a capital letter at the beginning and a trailing period. Also, this is not a bugfix, because it's not user-facing. Use a contrib fragment type.

tests/test_urldispatch.py
```diff
@@ -1151,7 +1151,7 @@ def test_invalid_route_name(router: Any) -> None:
         router.add_get("/", make_handler(), name="invalid name")
 
 
-def test_invalid_route_name(router) -> None:
+def test_invalid_route_name_identifier(router) -> None:
```
webknjaz: Could you parametrize the previous test instead? It's basically the same with just one thing being different.

setup.cfg
```diff
@@ -86,7 +86,7 @@ zip_ok = false
 [flake8]
 extend-select = B950
 # TODO: don't disable D*, fix up issues instead
-ignore = N801,N802,N803,E203,E226,E305,W504,E252,E301,E302,E501,E704,W503,W504,F811,D1,D4
```
webknjaz, line 89: What's this rule? How is it related to changing a test name?

alexmac: This rule catches cases where there are multiple definitions of the same thing - cases like this are quite common in tests where someone copy/pastes a function to add a new variant, but then forgets to rename the test - which results in only one test being run.

If you run pyflakes/ruff with the F811 check enabled on master you'll see:

```
./tests/test_urldispatch.py:1154:1: redefinition of unused 'test_invalid_route_name' from line 1149
```

I've removed the ignore here so this doesn't happen again, and this is the only instance of it currently in Aiohttp

CHANGES/8139.bugfix.rst
alexmac: updated - let me know if that looks good, or is too verbose

setup.cfg
Dreamsorcerer: We also pay a lot more attention to codecov these days, which should also stop this happening again.

tests/test_urldispatch.py
```diff
@@ -1146,14 +1146,16 @@ def test_subapp_iter(app: Any) -> None:
...
-        router.add_get("/", make_handler(), name="invalid name")
-
-
-def test_invalid_route_name(router) -> None:
+@pytest.mark.parametrize(
+    "route_name",
+    [
+        "invalid name",
+        "class",
+    ],
```
```suggestion
    (
        "invalid name",
        "class",
    ),
```

Dreamsorcerer:
@alexmac If you could follow the instructions to create the backports, that'd be appreciated.

setup.cfg
webknjaz: @alexmac makes sense. We should be moving towards having less ignored rules and more enabled ones. I was mostly asking because there's no code comment explaining it. These days I preach only disabling linting rules accompanied by an explanation so it doesn't linger around for years with nobody knowing why.

CHANGES/8139.bugfix.rst
webknjaz: @alexmac I suppose it'd be useful to define the audience for the change log. I've been trying to present some hints in the PR template and https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme — have you read it? Did you find the way it's worded useful?

In general, release notes should describe what's changed since the previous release in terms of impact for the end-users. And for the developers, Git commits would have context with more details. They are to follow different writing styles. The former should use something like the past tense or present simple (non-imperative mood) with phrasing like "X is/does Y now" with a high-level effect. The former would use imperative mood for the commit title and normal prose explaining the motivation and unobvious decisions — these target developers/contributors.

But there's intersection between the two audiences — external (and core) contributors. It is useful for them to also have some change notes, but not about the end-user facing changes. But rather about the changes within the project, automation, contribution/testing processes etc. For those, I made a separate change note type — `contrib`. It may be challenging to distinguish what should be written where.
In the context of such changes, the contributors should be treated as end-users of the development processes and automation, not the project itself. If a way to configure dev env or run tests changed, this should be recorded in the changelog. But Git commits don't go anywhere — they also need some text. It would be different — it shouldn't just contain a copy of the change note, which is already a part of the patch. It should contain motivation and decision justification, unobvious quirks of the implementation internals.

Finally, there's cases when we can justify not including a change note. Like re-enabling the tests. This does not have a high-level impact on the end-users (unless there are companion changes in the runtime code, of course). And this does not have an impact on how the contributors should interact with the project. So I'd say that a change log is not needed for improving tests coverage.

That said, this PR has a change to the linting configuration. That the rule change affects the contributors in that it'll start popping up in their PRs and local linter runs. So this part could be useful to put in a `contrib` change note to let people know. One PR/issue may have multiple change notes, if necessary. Or symlink other notes so that the same text would show up in different categories, or reference several issues/PRs. Similarly, a PR's change note should only include things that affect non-devs, not refactoring internals.

One last thing — your last change note variant lacks specificity — it refers to an unspecified linting rule being enabled in an unspecified linter. It's unclear who's affected and how. Those are the things to mention in a good changelog frament.

P.S. We have a way to instruct the bot to ignore a missing change note via a PR label. If you feel like your PR is something that might not need one — ask the maintainers to have it applied and the check will turn from being a red-cross error into a neutral gray square, unblocking the PR.

PR #8140, opened 2024-02-07
Use NPM clean install and upgrade node to v18

### LINKED ISSUE #8116: Varying node versions & not using NPM clean install

For building `llhttp`, the docs, workflows, and `Makefile` all use `npm install`.  This is not a particularly safe approach as it does not freeze the lock file.  Instead, the [`npm ci` command](https://docs.npmjs.com/cli/v10/commands/npm-ci) should be used.

In addition, I noticed that the current `aiohttp` CI builds with node 14, while `llhttp` itself just uses the default in GH hosted runners.  Strongly recommend pinning this to the same version, and 14 is long out of maintenance.  Version 18 is the maintenance LTS and 20 is the current LTS.

- All `llhttp` installs should use `npm ci`.
- Consistently build `llhttp` with the same Node.js version (either current LTS or maintenance LTS).

n/a

### Related component

Server, Client

### Additional context

_No response_

### Code of Conduct

steverep:
## What do these changes do?

- Installs `llhttp` using `npm ci` for security and to preserve the lock file.  This command should always be used in CI.
- Upgrades Node.js from v14 (out of maintenance) to v18 (current LTS maintenance release and also what the `llhttp` Dockerfile is using)

No

No

## Related issue number

Fixes #8116 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -115,13 +115,13 @@ jobs:
       uses: actions/cache@v4.0.0
       id: cache
       with:
-        key: llhttp-${{ hashFiles('vendor/llhttp/package.json', 'vendor/llhttp/src/**/*') }}
+        key: llhttp-${{ hashFiles('vendor/llhttp/package*.json', 'vendor/llhttp/src/**/*') }}
         path:  vendor/llhttp/build
     - name: Setup NodeJS
       if: steps.cache.outputs.cache-hit != 'true'
       uses: actions/setup-node@v4
       with:
-        node-version: '14'
+        node-version: 18
     - name: Generate llhttp sources
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
--- a/vendor/README.rst
+++ b/vendor/README.rst
@@ -12,7 +12,7 @@ newer release, add ``--remote``)::
 Then build ``llhttp``::
 
     cd vendor/llhttp/
-    npm install
+    npm ci
     make
 
 Then build our parser::
--- a/CHANGES/8116.contrib.rst
+++ b/CHANGES/8116.contrib.rst
@@ -1,0 +1,1 @@
+Updated CI and documentation to use NPM clean install and upgrade node to version 18 -- by :user:`steverep`.
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ aiohttp/%.c: aiohttp/%.pyx $(call to-hash,$(CYS)) aiohttp/_find_header.c
 	cython -3 -o $@ $< -I aiohttp -Werror
 
 vendor/llhttp/node_modules: vendor/llhttp/package.json
-	cd vendor/llhttp; npm install
+	cd vendor/llhttp; npm ci
 
 .llhttp-gen: vendor/llhttp/node_modules
 	$(MAKE) -C vendor/llhttp generate


vendor/README.rst
```diff
@@ -12,7 +12,7 @@ newer release, add ``--remote``)::
 Then build ``llhttp``::
 
     cd vendor/llhttp/
-    npm install
+    npm ci
```
Dreamsorcerer, line 15: What is this change? I'm not familiar with this command, and the llhttp instructions appear to still tell users to run `npm install`.

Dreamsorcerer: Oh right, this is the clean install. I'll test it out shortly, as there is a new llhttp to upgrade to.

steverep: > ...and the llhttp instructions appear to still tell users to run `npm install`.

What can I say, they should fix it too, along with a few other node-related irregularities. They correctly use it in the Dockerfile, but not in their workflows.  I'll give them a PR. 😉 

If you're developing and okay with mutating the lock file, `npm install` and others are fine.  But for CI, releases, use as a submodule, etc, `npm ci` should be used to treat it as immutable.  That's why it exists.

.github/workflows/ci-cd.yml
```diff
@@ -115,13 +115,13 @@ jobs:
       uses: actions/cache@v4.0.0
       id: cache
       with:
-        key: llhttp-${{ hashFiles('vendor/llhttp/package.json', 'vendor/llhttp/src/**/*') }}
+        key: llhttp-${{ hashFiles('vendor/llhttp/package*.json', 'vendor/llhttp/src/**/*') }}
```
steverep, line 118: BTW, I updated this to include the lock file, but consider revamping this key to something safer.  There are other files that could affect the build, such as the `Makefile`, `bin/generator.ts`, and maybe others if I dug deeper.  Consider just getting the commit SHA from the checkout and using that as the key.

Feel free to push something like that to this branch if you want.

vendor/README.rst
steverep: https://github.com/nodejs/llhttp/pull/298

.github/workflows/ci-cd.yml
Dreamsorcerer: To be honest, this really only needs to know if the submodule has changed, right? So, if that information can be obtained, it could actually be simplified a lot.

vendor/README.rst
Dreamsorcerer: I tried this command for upgrading to 9.2 and everything seems to work fine, so I'm happy with this.

.github/workflows/ci-cd.yml
steverep: Yes, exactly.  I would just get the commit SHA of the submodule folder and use that as key.  IMO, there's some other streamlining of this job that could be done, so probably best for a different PR (e.g. I think only one checkout is needed, and no need to both cache and upload an artifact of the build).

Dreamsorcerer: I'll merge this one for now, feel free to add additional improvements.

PR #8148, opened 2024-02-08
rename a shadowed test and re-enable F811 to catch future cases (#8139)

alexmac:
(cherry picked from commit 3c0f1eb29d3512419ea65e7cdeb61ba3f3496f00)

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -154,11 +154,6 @@ async def test_threaded_negative_lookup() -> None:
         await resolver.resolve("doesnotexist.bla")
 
 
-async def test_close_for_threaded_resolver(loop) -> None:
-    resolver = ThreadedResolver(loop=loop)
-    await resolver.close()
-
-
 async def test_threaded_negative_lookup_with_unknown_result() -> None:
     loop = Mock()
 
@@ -182,6 +177,9 @@ async def test_threaded_negative_lookup_with_unknown_result() -> None:
         res = await resolver.resolve("www.python.org")
     assert len(res) == 0
 
+
+async def test_close_for_threaded_resolver(loop) -> None:
+    resolver = ThreadedResolver()
     await resolver.close()
 
 
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -154,30 +154,6 @@ async def test_threaded_negative_lookup() -> None:
         await resolver.resolve("doesnotexist.bla")
 
 
-async def test_threaded_negative_lookup_with_unknown_result() -> None:
-    loop = Mock()
-
-    # If compile CPython with `--disable-ipv6` option,
-    # we will get an (int, bytes) tuple, instead of a Exception.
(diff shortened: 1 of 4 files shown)


alexmac:
@Dreamsorcerer - here's one of the backports

tests/test_urldispatch.py
```diff
@@ -1130,14 +1130,16 @@ def test_subapp_iter(app) -> None:
...
-
-def test_invalid_route_name(router) -> None:
+@pytest.mark.parametrize(
+    "route_name",
+    (
+        "invalid name",
+        "class",
+    ),
+)
+def test_invalid_route_name(router: Any, route_name: str) -> None:
```
```suggestion
def test_invalid_route_name(router, route_name: str) -> None:
```

alexmac:
@Dreamsorcerer I've removed the other shadowed test (made the diff match master) - do you want the contrib file updated for this back port, and if so do I updated both the pr number and the text or just the text?

Dreamsorcerer:
Ah, I think it's fine.

tests/test_resolver.py
```diff
@@ -179,33 +179,7 @@ async def unknown_addrinfo(*args: Any, **kwargs: Any) -> List[Any]:
 
 
 async def test_close_for_threaded_resolver(loop) -> None:
-    resolver = ThreadedResolver(loop=loop)
```
webknjaz, line 182: @Dreamsorcerer did you miss that this removes an unrelated test in a test module that the original commit didn't even touch?

webknjaz: @alexmac it seems like the conflict resolution hasn't been completed correctly and this removed a legit test. Why?

webknjaz: @alexmac was it exactly identical to the other test? It'd be useful to have in a PR description. And perhaps, do such things in separate PRs. Backports must never include changes that aren't directly related to resolving conflicts with the original patch.

Dreamsorcerer: You can see the same test still exists, though the diff makes it a bit awkward to tell. The flake8 rule found this duplicate test that only existed on the 3.x branches. Could have been removed in another PR, but I don't think it makes any real difference.

webknjaz: Yeah, I managed to find that. It'd be much more obvious (== less mental overhead) if this was explained explicitly.

As for a separate PR — this is the way. Patchback would also work to port from 3.10 to 3.9. The thing is that the `git cherry-pick -x` is supposed to contain a copy of another change, it even generates a reference `(cherry picked from commit xxxxxx)`. By making arbitrary changes, that makes it hard to track, breaking transparency of the updates to Git. Atomic commits/PRs is a convention that is designed around certain expectations/promises. Breaking them hurts future git paleontology, making the introspection harder. This effect is often not visible/obvious to the commit authors, making it a hard sell, but is quite important for keeping sane history.

PR #8153, opened 2024-02-11
Ensure xtests are running

Dreamsorcerer:


Dreamsorcerer:
Seems to be unclosed transport warnings being triggered for a reason I can't tell, and I don't have the time to investigate.

PR #8154, opened 2024-02-13
Implement granular URL error hierarchy in the HTTP client #6722

setla:
What do these changes do?
Added new exception type to separate cases with invalid redirect url during request. Previously ValueError or InvalidURL was raised and screening out was hard (valid url that redirects to invalid one raised the same error as invalid url).

Slightly improved test coverage.

Docs changes previews:

https://aiohttp--6722.org.readthedocs.build/en/6722/client_reference.html#hierarchy-of-exceptions
https://aiohttp--6722.org.readthedocs.build/en/6722/changes.html#features
Are there changes in behavior for the user?
In case of invalid redirect url during request - new InvalidRedirectUrl will be raised instead of ValueError or InvalidURL, it derives from InvalidURL so it's backward compatibile

Related issue number
Resolves https://github.com/aio-libs/aiohttp/issues/2507
Resolves https://github.com/aio-libs/aiohttp/issues/2630
Resolves https://github.com/aio-libs/aiohttp/issues/3315

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```


webknjaz:
@setla did you follow the instructions of using `git cherry-pick` with the `-x` flag? It normally generates a new commit message based on the old one + adds a reference to the original commit. Could you reproduce that? The "merge" commit hash is visible in that PR.

Dreamsorcerer:
> @setla did you follow the instructions of using `git cherry-pick` with the `-x` flag?

I would note that the first couple of times I did this, I didn't realise that you need to use `git cherry-pick --continue` after resolving conflicts. Git is very fragile with these commands and it seems incredibly easy to cancel the cherry-pick without even realising. The same happens with merge commits, and I realised in some previous releases that I messed up the merge, because reverting a file in the merge actually cancels the merge and I just ended up doing a normal commit without the merge metadata (thus creating a load of duplicate conflicts when I did the next release).

webknjaz:
I'm pretty sure that doing `git commit` also continues the cherry-picking, since the commit message is pre-filled. But yes, it's easy to reset by accident. Also, it's a good insight that it's not obvious how to continue post conflict resolution. Maybe, it'd be useful to bake into the backporting instructions the bot posts.

As for cherry-picking true merge commits, there's `-m1` which is already suggested in those instructions. And it doesn't apply to this one since I merged it with the squash strategy, so the commit doesn't have more than one parent.

You're right that many people don't realize how stateful and low-level many Git procedures are. Thanks for reminding me that!

P.S. One tip I can give regarding the duplicate conflicts is to set [`rerere.enabled`](https://git-scm.com/book/en/v2/Git-Tools-Rerere) that records a number manual resolutions and is then able to replay them automatically.

PR #8163, opened 2024-02-18
Avoid creating a task to do DNS resolution if there is no throttle

bdraco:
## What do these changes do?

Avoid creating a task to do dns resolution if there is no throttle

When doing some debugging, I noticed that a large amount of tasks created on a Home Assistant instance were `TCPConnector._resolve_host` for cases where the host was an ip address or already in the cache.  On this instance, it reduced the number of `_resolve_host` tasks by 99%. Previously I observed ~500 per minute.

There is a small performance improvement on cache hit and when the host is an ip address.  Since requests no longer have to suspend on cache hit this can reduce latency of requests.  Scheduling and awaiting a task on the event loop can take 2 order of magnitude more wall clock time vs running a simple function.

no

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


tests/test_connector.py
```diff
@@ -1032,6 +1033,9 @@ async def test_tcp_connector_dns_throttle_requests_exception_spread(loop: Any) -
         r1 = loop.create_task(conn._resolve_host("localhost", 8080))
         r2 = loop.create_task(conn._resolve_host("localhost", 8080))
         await asyncio.sleep(0)
+        await asyncio.sleep(0)
```
bdraco, line 1036: These tests all rely on the event loop overhead introduced by the task. Without the task we need to let it run a few more iterations because the test gets to run sooner now

Dreamsorcerer: I'd probably just put a delay in (0.01 or something) if it's not guaranteed to be done in the next iteration.

Dreamsorcerer (approved):

aiohttp/connector.py
```diff
@@ -814,6 +814,7 @@ def clear_dns_cache(
     async def _resolve_host(
         self, host: str, port: int, traces: Optional[List["Trace"]] = None
     ) -> List[Dict[str, Any]]:
+        """Resolves host and returns list of addresses."""
```
webknjaz: PEP 257 requires a sentence in imperative mood, in docstring title lines: https://peps.python.org/pep-0257/#one-line-docstrings

aiohttp/connector.py
```diff
@@ -850,6 +850,39 @@ async def _resolve_host(
...
+        # for the first one to resolve and then use the result for all of them.
+        # We use a throttle event to ensure that we only resolve the host once
+        # and then use the result for all the waiters.
+        #
+        # In this case we need to create a task to ensure that we can shield
+        # the task from cancellation as cancelling this lookup should not cancel
+        # the underlying lookup or else the cancel event will get broadcast to
+        # all the waiters across all connections.
+        #
+        host_resolved = asyncio.create_task(
```
webknjaz: Maybe, call it a `resolved_host_task`?

aiohttp/connector.py
```diff
@@ -850,6 +850,39 @@ async def _resolve_host(
...
+            self._resolve_host_with_throttle(key, host, port, traces)
+        )
+        try:
+            return await asyncio.shield(host_resolved)
+        except asyncio.CancelledError:
+
+            def drop_exception(fut: "asyncio.Future[List[Dict[str, Any]]]") -> None:
```
webknjaz, line 871: No need to quote the annotation since there's no py310-only syntax here:
```suggestion
            def drop_exception(fut: asyncio.Future[List[Dict[str, Any]]]) -> None:
```

bdraco: I'm not sure asyncio.Future is sub-scriptable on older python.  Will drop it and see if the CI complains

aiohttp/connector.py
bdraco: I moved the code from below and didn't think about renaming it.  Sounds good to me.

aiohttp/connector.py
bdraco: 2367798d26f4d8928e618976f8d231e2160ce16b

bdraco:
How do we feel about backporting this to 3.9.  HA is having a lot of stability problems with too many tasks, and this would help a lot in reducing them.

tests/test_connector.py
bdraco: 8609c35605022650ca07d04fdedd76f708b7b372

I had to leave the two in test_tcp_connector_dns_throttle_requests_cancelled_when_close since it failed otherwise

aiohttp/connector.py
Dreamsorcerer: Just tried, it's not.

bdraco: restored quoting 38e0649023c32bcb2c986a0aa744f5ad17799c32

tests/test_connector.py
bdraco: I ended up reverting them since the test was only reliable as when it waited X iterations of the event loop instead of time

aiohttp/connector.py
webknjaz: Ah, I didn't realize that.

aiohttp/connector.py
```diff
@@ -814,6 +814,7 @@ def clear_dns_cache(
     async def _resolve_host(
         self, host: str, port: int, traces: Optional[List["Trace"]] = None
     ) -> List[Dict[str, Any]]:
+        """Resolve host and returns list of addresses."""
```
webknjaz:
```suggestion
        """Resolve host and return list of addresses."""
```

webknjaz:
> How do we feel about backporting this to 3.9. HA is having some stability problems with too many tasks, and this would help a lot in reducing them.

I think this can easily be declared a bugfix since (1) it's not a new feature (meaning no grounds for requiring a minor version bump) and (2) it fixes an undesired behavior.

aiohttp/connector.py
bdraco: thanks

CHANGES/8163.bugfix.rst
```diff
@@ -0,0 +1 @@
+Avoid creating a task to do DNS resolution if there is no throttle
```
webknjaz: Check out the recently improved guidelines: https://docs.aiohttp.org/en/latest/contributing.html#changelog-update / https://github.com/aio-libs/aiohttp/tree/master/CHANGES#readme.

Perhaps, something like this would be easier comprehend for the end-user audience:
```suggestion
Improved the not-throttled DNS resolution performance
-- by :user:`bdraco`.

This is achieverd by avoiding an :mod:`asyncio` task creation
in this case.
```

In case, that's hard to follow, here's somebody else's attempt to encode good release notes writing style: https://redhat-documentation.github.io/supplementary-style-guide/#release-notes (it doesn't fully correspond to our needs, though).

CHANGES/8163.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Improved the not-throttled DNS resolution performance
+-- by :user:`bdraco`.
+
+This is achieverd by avoiding an :mod:`asyncio` task creation
```
bdraco:
```suggestion
This is achieved by avoiding an :mod:`asyncio` task creation
```

CHANGES/8163.bugfix.rst
bdraco: Works for me. Thanks. It really helps to get another opinion on how to describe the change since its hard to guess how someone who didn't write the change will want to consume it.... I am not the user problem.

webknjaz: Yep, it's hard for everyone which is why I'm failing to come up with better guidelines :)

I just attempted rephrasing what you posted in the PR, that's all my context now.

Dreamsorcerer: "Performance: Avoided creating tasks unnecessarily during DNS resolution."

would cover it for me. I don't even know what throttled means in this context.

CHANGES/8163.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Improved the not-throttled DNS resolution performance
```
bdraco:
```suggestion
Improved the DNS resolution performance on cache hit
```

CHANGES/8163.bugfix.rst
bdraco: I adjusted it again to remove the word `throttled`

webknjaz: That's too low-level. Throttling refers to limiting the number of simultaneous connections to resolve the same hostname, it seems.

Dreamsorcerer: > Throttling refers to limiting the number of simultaneous connections to resolve the same hostname, it seems.

Right, to me that's just sensible caching (e.g. If you just added `@alru_cache` to the function it would also do that).

webknjaz: Brilliant! That's much better!

webknjaz (approved):

bdraco:
Thanks for the reviews.  I'll merge it tomorrow or later tonight after some sleep + recheck.

bdraco:
Still looks good, functional retest is good.

Issue #8184, opened 2024-02-26
web_ws - Can't close tcp socket when receiving a close message from client.


spikefishjohn:
### Describe the bug

I'm working on a https://github.com/GNS3/gns3-server/issues/2320 for GNS3 that I think [PR7978](https://github.com/aio-libs/aiohttp/pull/7978) lines up with. The GNS3 issue is that when a web socket client sends a close message the server never closes the tcp socket. I tried the patch above but it doesn't address the issue.

High level GNS3 is calling [ws.receive()](https://github.com/GNS3/gns3-server/blob/master/gns3server/handlers/api/controller/notification_handler.py#L33C14-L33C27) at which point aiohttp receive() the close message.

I believe the problem is because ws.receive is setting [self._closing = True](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L474) when it reiceves a close message. This then causes self.close() to [return here](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) which prevents [if msg.type == WSMsgType.CLOSE:](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L413) from being reached.

I'm almost thinking [this should be removed](https://github.com/aio-libs/aiohttp/blob/bf4edce5d6c967fa1d6a581b3aaab1bd2f5121cd/aiohttp/web_ws.py#L397) but i'm not sure what the intent of that is so i'm unsure if that is the proper fix.

### To Reproduce

I don't have a great way to reproduce this. I'm currently just using a fully installed GNS3 instance. I can work on making a reproduction now that I understand the issue.

### Expected behavior

aiohttp.web_ws should close the socket when it receives a close message. 

### Logs/tracebacks

```python-traceback
I'll add these later in the week. I wasn't prepared for this.
```

### Python Version

```console
$ python --version
Python 3.10.12
```

### aiohttp Version

```console
$ python -m pip show aiohttp
(gns3-server) root@ubuntu:~# source /usr/share/gns3/gns3-server/bin/activate
(gns3-server) root@ubuntu:~# python3 -m pip show aiohttp
Name: aiohttp
Version: 3.9.1
Summary: Async http client/server framework (asyncio)
[...]

PR #8201, opened 2024-03-02
Release v3.9.4

Dreamsorcerer:
Preparing for tomorrow, let me know if there's any reason to delay. @bdraco @webknjaz

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,93 @@
 
 .. towncrier release notes start
 
+3.9.4 (2024-03-03)
+==================
+
+Bug fixes
+---------
+
+- The asynchronous internals now set the underlying causes
+  when assigning exceptions to the future objects
+  -- by :user:`webknjaz`.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8089`.
+  
+  
+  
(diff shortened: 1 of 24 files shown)


webknjaz (approved):
LGTM

bdraco (approved):

bdraco:
https://github.com/aio-libs/aiohttp/issues/8184 might be ready tomorrow but its not a regression and its been a problem for a while so no need to wait for it

Dreamsorcerer:
> #8184 might be ready tomorrow but its not a regression and its been a problem for a while so no need to wait for it

Probably a good one to have in, not sure if we want to do any more 3.9 release after this. I'll do the release later in the week if you need time to finish it.

bdraco:
I was hoping to get a thumbs up from the op on if it fixed it for his user before moving it forward. I'll write the tests for it tomorrow regardless, since I'll be time poor at the end of the week as I have family in town.

webknjaz:
> I was hoping to get a thumbs up from the op on if it fixed it for his user before moving it forward.

@bdraco did you get that info?

Dreamsorcerer:
That one's merged now. We've had a couple of other last minute bug reports we've been fixing, but I think we're good for me to proceed with a release tomorrow.

webknjaz:
Sounds good!

PR #8251, opened 2024-03-26
Leave websocket transport open if receive times out or is cancelled

bdraco:
## What do these changes do?

Leave websocket transport open if receive times out or is cancelled.

This change restores the behavior before #7978

See https://github.com/aio-libs/aiohttp/pull/7978#issuecomment-2015714011

The connection is left open if the receive call times out or is cancelled

no

## Related issue number

https://github.com/aio-libs/aiohttp/pull/7978#issuecomment-2015714011

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/tests/test_web_websocket_functional.py
+++ b/tests/test_web_websocket_functional.py
@@ -879,8 +879,7 @@ async def test_receive_timeout_keeps_connection_open(
 
     ws = await client.ws_connect("/", autoping=False)
 
-    await asyncio.sleep(0)
-    await asyncio.sleep(sys.float_info.min)
+    await asyncio.sleep(0.1)
     await ws.ping("data")
 
     msg = await ws.receive()
--- a/tests/test_web_websocket_functional.py
+++ b/tests/test_web_websocket_functional.py
@@ -2,6 +2,8 @@
 # HTTP websocket server functional tests
 
 import asyncio
+import contextlib
+import sys
 from typing import Any
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/test_web_websocket_functional.py
```diff
@@ -799,3 +801,91 @@ async def ws_handler(request):
...
+        closed.set_result(None)
+        return ws
+
+    app = web.Application()
+    app.router.add_get("/", handler)
+    client = await aiohttp_client(app)
+
+    ws = await client.ws_connect("/", autoping=False)
+
+    await asyncio.sleep(0.1)
```
bdraco: Sleep probably doesn't make sense here, better to provide a future and resolve it once the server times out

bdraco:
Tested on production. No adverse effects observed. Old behavior restored without the leak

Dreamsorcerer:
And this still fixes the leak from before? I was thinking that we'd close the connection if the connection timed out (indicating that we may have lost the connection), but would keep the connection if it was cancelled externally.

Dreamsorcerer:
Also, setting the close code to abnormal closure still seems like we're attempting to close the connection to me. That still doesn't seem like the right behaviour when the task has been cancelled externally. I think no action should be taken in that situation (though, if someone is always putting a timeout on, then maybe there are issues with it never detecting the connection is timed out?).

bdraco:
> And this still fixes the leak from before? I was thinking that we'd close the connection if the connection timed out (indicating that we may have lost the connection), but would keep the connection if it was cancelled externally.

Yes, the original problem was in `close`, and the change to `receive` wasn't needed. I changed it to be consistent but that seems incorrect in hindsight.

bdraco:
> Also, setting the close code to abnormal closure still seems like we're attempting to close the connection to me. That still doesn't seem like the right behaviour when the task has been cancelled externally. I think no action should be taken in that situation (though, if someone is always putting a timeout on, then maybe there are issues with it never detecting the connection is timed out?).

I did a strait revert to the previous version, but I agree that seems wrong as well.  I think it should simply raise here instead

bdraco:
I made that change in the code, but I'm too tired to retest it so marked it back as draft until I've had more sleep

bdraco:
Functional testing looks good on production

Dreamsorcerer (approved):

bdraco:
Everything ran ok overnight. Nothing in the logs. All looks good

PR #8252, opened 2024-03-26
Fix handling of unsupported upgrades with the pure python http parser

bdraco:
## What do these changes do?

Fix handling of unsupported upgrades with the pure python http parser

The content is now read from an unsupported upgrade request

no

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/8252.bugfix.rst
+++ b/CHANGES/8252.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed content not being read when an upgrade request was not supported with the pure Python implementation.
+-- by :user:`bdraco`.
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -227,7 +227,7 @@ class HeadersParser:
 
 def _is_supported_upgrade(headers: CIMultiDictProxy[str]) -> bool:
     """Check if the upgrade header is supported."""
-    return headers.get(hdrs.UPGRADE, "").lower() in ("tcp", "websocket")
+    return headers.get(hdrs.UPGRADE, "").lower() in {"tcp", "websocket"}
 
 
 class HttpParser(abc.ABC, Generic[_MsgT]):
--- a/aiohttp/http_parser.py
+++ b/aiohttp/http_parser.py
@@ -227,7 +227,7 @@ class HeadersParser:
 
 def _is_supported_upgrade(headers: CIMultiDictProxy[str]) -> bool:
     """Check if the upgrade header is supported."""
-    return headers.get(hdrs.UPGRADE, "").lower() == "websocket"
+    return headers.get(hdrs.UPGRADE, "").lower() in ("tcp", "websocket")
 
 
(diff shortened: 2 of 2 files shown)


bdraco:
https://github.com/aio-libs/aiohttp/pull/7960 passed but I guess I need to scape more out of there.  Had to leave for the airport, so will have to pick it up later.

aiohttp/http_parser.py
```diff
@@ -225,6 +225,11 @@ def parse_headers(
         return (CIMultiDictProxy(headers), tuple(raw_headers))
 
 
+def _is_supported_upgrade(headers: CIMultiDictProxy[str]) -> bool:
+    """Check if the upgrade header is supported."""
+    return headers.get(hdrs.UPGRADE, "").lower() in ("tcp", "websocket")
```
bdraco: `tcp` is needed here or `test_upgrade_connection_not_released_after_read` failed. Are there any others?

```suggestion
    return headers.get(hdrs.UPGRADE, "").lower() in {"tcp", "websocket"}
```

I'd probably use a set for best practices (I think when I tested performance in the past, it's only ~10% slower than checking the first item of a tuple and then significantly faster if needing to check anything more than 1 item).

Dreamsorcerer (approved):

Dreamsorcerer:
I assume this affected all platforms then. Linux being fast enough that the race condition succeeded, Windows being slow enough that it always failed, and Mac being unreliable enough that it passes half the time?

bdraco:
> I assume this affected all platforms then. Linux being fast enough that the race condition succeeded, Windows being slow enough that it always failed, and Mac being unreliable enough that it passes half the time?

Yes that is the same conclusion I reached as well.

CHANGES/8252.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed content not being read when an upgrade request was not supported with the pure Python implementation.
```
webknjaz, line 1: @bdraco it's best not to put a trailing period at the end of the sentence if the byline is right after — RST will render it on the same line. If you want a separate line, add an extra empty separator line in between. Look how it's rendered now: https://docs.aiohttp.org/en/latest/changes.html#bug-fixes.

Also, I suggest not using any period after the byline if it's on a dedicated line as opposed to being at the end of a sentence.

Examples:
* https://yarl.aio-libs.org/en/latest/changes/#id5 — look at the last entry — here I have a byline at the end of the first sentence/paragraph. So the sentence itself doesn't have a period before the em dash but it ends with one right after the byline. It's followed by a separate paragraph of additional text.
* https://multidict.aio-libs.org/en/latest/changes/#id3 — most entries here have bylines on separate lines, I only missed to drop a trailing period from one, but others are nice.

PR #8257, opened 2024-03-28
Ensure websocket transport is closed when client does not close it (#8200)

bdraco:
(cherry picked from commit 6ec4747a02955dbf227fb66f4419d124a37fcc71)

### DIFF
--- a/CHANGES/8200.bugfix.rst
+++ b/CHANGES/8200.bugfix.rst
@@ -1,0 +1,6 @@
+Ensure websocket transport is closed when client does not close it
+-- by :user:`bdraco`.
+
+The transport could remain open if the client did not close it. This
+change ensures the transport is closed when the client does not close
+it.
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -395,6 +395,7 @@ class WebSocketResponse(StreamResponse):
             return True
 
         if self._closing:
+            self._close_transport()
             return True
 
         reader = self._reader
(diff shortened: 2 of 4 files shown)


bdraco:
That import time test running with coverage on is a bit useless since coverage is so much slower on 3.12

Dreamsorcerer:
> That import time test running with coverage on is a bit useless since coverage is so much slower on 3.12

Ah, so it's the coverage that is causing it to fail? Is there a way to run a test without coverage or something?

bdraco:
I was going to mark it `@pytest.mark.no_cover` but than I realized its just cpython 3.12 taking more time to import which is expected here. Maybe we should increase the time if its > 3.12 instead

Dreamsorcerer:
> nevermind, its running isolated. Its a regression in cpython 3.12 taking longer to import

Do we reckon it'll get improved in a future release?

Dreamsorcerer:
> Maybe we should increase the time if its > 3.12 instead

Can do, or we can skip it on 3.12 for now, if we think it's a temporary regression.

bdraco:
> > nevermind, its running isolated. Its a regression in cpython 3.12 taking longer to import
> 
> Do we reckon it'll get improved in a future release?

It might be better in 3.13, but from what I gather the timing changes are expected in 3.12 as there were trade-offs to make other things at run time faster

bdraco:
some more backstory https://github.com/python/cpython/issues/90081

Dreamsorcerer:
Well, if coverage is going to affect the performance, we should disable it. That test will have no impact on coverage anyway as it does nothing more than import the library. If that makes a notable difference, then we may want to lower the threshold to match the new timings.

If there is a regression in 3.12 we should probably just skip the test on 3.12. If there's still an issue on 3.13, then we can reevaluate.

PR #8259, opened 2024-03-28
Flag python 3.11/12 support

Dreamsorcerer:


bdraco (approved):

webknjaz (approved):
Nice, I didn't realize the tests are passing already!

Dreamsorcerer:
> Nice, I didn't realize the tests are passing already!

bdraco just fixed a race condition a few hours earlier which was the cause of the failures.

Issue #559, opened 2015-10-08
aiohttp client does not support ipv6 addresses


mx2323:
attempting to get a webpage from an ipv6 address doesnt appear to work in aiohttp 0.17.4

when i run the ipv6 address through curl it works. 

i also tried the ipv4 address through the aiohttp client and that worked as well. 

http://[::1]:9001/foo fails with the following exception
*\* 
socket.gaierror: [Errno -9] **Address family for hostname not supported**
**

```
 Cannot connect to host ::1:9001 ssl:False http://[::1]:9001/foo
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/aiohttp/connector.py", line 571, in _create_connection
    server_hostname=hinfo['hostname'] if sslcontext else None)
  File "/usr/lib/python3.4/asyncio/base_events.py", line 581, in create_connection
    infos = f1.result()
  File "/usr/lib/python3.4/asyncio/futures.py", line 275, in result
    raise self._exception
  File "/usr/lib/python3.4/concurrent/futures/thread.py", line 54, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.4/socket.py", line 533, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -9] **Address family for hostname not supported**

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/aiohttp/connector.py", line 292, in connect
    transport, proto = yield from self._create_connection(req)
  File "/usr/local/lib/python3.4/dist-packages/aiohttp/connector.py", line 588, in _create_connection
    (req.host, req.port)) from exc
aiohttp.errors.ClientOSError: Can not connect to ::1:9001
```

mx2323:
looks like through a client with a session, it has to be specified via the connector, which is not intuitive or called out clearly in the documentation imo.

asvetlov:
That makes sense, I see no objections for relaxing default _family_ param from IPv4 to both IPv4 and IPv6.

asvetlov:
Fixed

gwillem:
@asvetlov relaxing family breaks (cryptically) on systems with an IPv6 stack, `aiodns` installed but no IPv6 route. See https://github.com/saghul/aiodns/issues/23

[...]

PR #8270, opened 2024-03-30
Fix AsyncResolver to match ThreadedResolver behavior

bdraco:
## What do these changes do?

AsyncResolver was disabled by default because it did not implement all of functionality of ThreadedResolver because `aiodns` did not support `getaddrinfo` until https://github.com/saghul/aiodns/pull/118 which meant only IPv6 addresses were returned in most cases on systems that supported IPv6

see https://github.com/aio-libs/aiohttp/issues/559

`AsyncResolver` now behaves the same as the current default `ThreadedResolver` and will return both IPv6 and IPv4 addresses on systems that support it.

no

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

needs 
- [x] https://github.com/saghul/aiodns/pull/118
- [x] https://github.com/saghul/aiodns/pull/119
- [x] aiodns release (requested at https://github.com/saghul/aiodns/pull/119#issuecomment-2028487751)

### DIFF
--- a/CHANGES/8270.bugfix.rst
+++ b/CHANGES/8270.bugfix.rst
@@ -1,0 +1,9 @@
+Fix AsyncResolver to match ThreadedResolver behavior
+-- by :user:`bdraco`.
+
+On system with IPv6 support, the AsyncResolver would not fallback
+to providing A records when AAAA records were not available.
+Additionally, unlikely the ThreadedResolver, the AsyncResolver
+did not handle link-local addresses correctly.
+
+This change makes the behavior consistent with the ThreadedResolver.
--- a/CHANGES/8270.bugfix.rst
+++ b/CHANGES/8270.bugfix.rst
@@ -3,7 +3,7 @@ Fix AsyncResolver to match ThreadedResolver behavior
 
 On system with IPv6 support, the AsyncResolver would not fallback
 to providing A records when AAAA records were not available.
-Additionally, unlikely the ThreadedResolver, the AsyncResolver
+Additionally, unlike the ThreadedResolver, the AsyncResolver
 did not handle link-local addresses correctly.
 
 This change makes the behavior consistent with the ThreadedResolver.
--- a/CHANGES/8270.bugfix.rst
+++ b/CHANGES/8270.bugfix.rst
@@ -1,9 +1,9 @@
-Fix AsyncResolver to match ThreadedResolver behavior
+Fix ``AsyncResolver`` to match ``ThreadedResolver`` behavior
 -- by :user:`bdraco`.
 
-On system with IPv6 support, the AsyncResolver would not fallback
+On system with IPv6 support, the :py:class:`~aiohttp.resolver.AsyncResolver` would not fallback
 to providing A records when AAAA records were not available.
-Additionally, unlike the ThreadedResolver, the AsyncResolver
+Additionally, unlike the :py:class:`~aiohttp.resolver.ThreadedResolver`, the :py:class:`~aiohttp.resolver.AsyncResolver`
 did not handle link-local addresses correctly.
 
-This change makes the behavior consistent with the ThreadedResolver.
+This change makes the behavior consistent with the :py:class:`~aiohttp.resolver.ThreadedResolver`.
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -120,6 +120,18 @@ class AbstractView(ABC):
 
 
 class ResolveResult(TypedDict):
+    """Resolve result.
+
+    This is the result returned from an AbstractResolver's
+    resolve method.
+
+    :param hostname: The hostname that was provided.
+    :param host: The IP address that was resolved.
+    :param port: The port that was resolved.
+    :param family: The address family that was resolved.
+    :param proto: The protocol that was resolved.
+    :param flags: The flags that were resolved.
+    """
 
     hostname: str
     host: str
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -1,4 +1,5 @@
 import logging
+import socket
 from abc import ABC, abstractmethod
 from collections.abc import Sized
 from http.cookies import BaseCookie, Morsel
@@ -132,7 +133,9 @@ class AbstractResolver(ABC):
     """Abstract DNS resolver."""
 
     @abstractmethod
-    async def resolve(self, host: str, port: int, family: int) -> List[ResolveResult]:
+    async def resolve(
+        self, host: str, port: int = 0, family: int = socket.AF_INET
+    ) -> List[ResolveResult]:
         """Return IP address for given hostname"""
 
     @abstractmethod
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -13,6 +13,7 @@ from typing import (
     List,
     Optional,
     Tuple,
+    TypedDict,
 )
 
 from multidict import CIMultiDict
@@ -117,11 +118,21 @@ class AbstractView(ABC):
         """Execute the view handler."""
 
 
+class ResolveResult(TypedDict):
+
+    hostname: str
+    host: str
+    port: int
+    family: int
+    proto: int
+    flags: int
+
+
 class AbstractResolver(ABC):
     """Abstract DNS resolver."""
 
     @abstractmethod
-    async def resolve(self, host: str, port: int, family: int) -> List[Dict[str, Any]]:
+    async def resolve(self, host: str, port: int, family: int) -> List[ResolveResult]:
         """Return IP address for given hostname"""
 
     @abstractmethod
--- a/aiohttp/resolver.py
+++ b/aiohttp/resolver.py
@@ -113,7 +113,8 @@ class AsyncResolver(AbstractResolver):
                     # LL IPv6 is a VERY rare case. Strictly speaking, we should use
                     # getnameinfo() unconditionally, but performance makes sense.
                     result = await self._resolver.getnameinfo(
-                        address[0].decode("ascii"), *address[1:], _NUMERIC_SOCKET_FLAGS
+                        (address[0].decode("ascii"), *address[1:]),
+                        _NUMERIC_SOCKET_FLAGS,
                     )
                     resolved_host = result.node
                 else:
--- a/aiohttp/resolver.py
+++ b/aiohttp/resolver.py
@@ -27,10 +27,10 @@ class ThreadedResolver(AbstractResolver):
         self._loop = asyncio.get_running_loop()
 
     async def resolve(
-        self, hostname: str, port: int = 0, family: int = socket.AF_INET
+        self, host: str, port: int = 0, family: int = socket.AF_INET
     ) -> List[Dict[str, Any]]:
         infos = await self._loop.getaddrinfo(
-            hostname,
+            host,
             port,
             type=socket.SOCK_STREAM,
             family=family,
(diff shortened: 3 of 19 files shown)


aiohttp/resolver.py
```diff
@@ -85,19 +85,49 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+            if family == socket.AF_INET6:
+                if len(address) < 3:
+                    # IPv6 is not supported by Python build,
+                    # or IPv6 is not enabled in the host
+                    continue
+                if address[3]:
+                    # This is essential for link-local IPv6 addresses.
+                    # LL IPv6 is a VERY rare case. Strictly speaking, we should use
+                    # getnameinfo() unconditionally, but performance makes sense.
+                    host, _port = await self._loop.getnameinfo(
```
bdraco: pycares supports getnameinfo so we can avoid the executor here as well but I need to do another PR to aiodns

aiohttp/resolver.py
```diff
@@ -85,19 +85,49 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self._resolver = aiodns.DNSResolver(*args, loop=self._loop, **kwargs)
 
     async def resolve(
-        self, host: str, port: int = 0, family: int = socket.AF_INET
+        self, hostname: str, port: int = 0, family: int = socket.AF_INET
```
Dreamsorcerer: The parent class names this `host`, so should probably keep that name (not sure why mypy doesn't give an error about this...).

bdraco: I was matching it to ThreadedResolver. I guess we change both to host?

Dreamsorcerer: Probably makes more sense.

bdraco:
I've tested this on production and everything works as expected.  I'll do another turn and address review comments after the aiodns release

bdraco:
Functional testing looks good, but all the tests for the resolver are patching the old Cath so lots to update

aiohttp/resolver.py
```diff
@@ -48,19 +48,19 @@ async def resolve(
                     # This is essential for link-local IPv6 addresses.
                     # LL IPv6 is a VERY rare case. Strictly speaking, we should use
                     # getnameinfo() unconditionally, but performance makes sense.
-                    host, _port = socket.getnameinfo(
+                    resolved_host, _port = socket.getnameinfo(
```
bdraco: Looks like there is no coverage for this ?

bdraco: Also its using socket and not loop!

tests/test_resolver.py
```diff
@@ -49,21 +93,77 @@ async def fake(*args: Any, **kwargs: Any) -> List[Any]:
...
+        mock().getaddrinfo.assert_called_with(
+            "www.python.org",
+            family=socket.AF_INET,
+            flags=socket.AI_ADDRCONFIG,
+            port=0,
+            type=socket.SOCK_STREAM,
+        )
+
+
 @pytest.mark.skipif(not gethostbyname, reason="aiodns 1.1 required")
```
bdraco: This needs to be getaddrinfo as well as the other ones, but has to wait for a real release to change

aiohttp/resolver.py
```diff
@@ -48,24 +51,24 @@ async def resolve(
                     # This is essential for link-local IPv6 addresses.
                     # LL IPv6 is a VERY rare case. Strictly speaking, we should use
                     # getnameinfo() unconditionally, but performance makes sense.
-                    host, _port = socket.getnameinfo(
-                        address, socket.NI_NUMERICHOST | socket.NI_NUMERICSERV
+                    resolved_host, _port = await self._loop.getnameinfo(
```
bdraco, line 56: This has apparently never worked on python less than 3.9

bdraco: Yeah, it's the same problem we have in zeroconf where scope ID isn't supported on an IP address until 3.9.

Sisters never been in any test coverage here, the problem has gone unnoticed since it's very rare. This is even used anyway nobody seems to have complained

Dreamsorcerer: Huh, oh well, we'll drop support for 3.8 in a few months anyway.

bdraco: I'm going to mark the test as skip on 3.8 as I don't think there's any point in trying to fix it as it simply isn't going to work

I doubt anyone's using it on 3.8 who would even need this... I guess they really can't anyways

aiohttp/abc.py
```diff
@@ -117,11 +119,23 @@ def __await__(self) -> Generator[Any, None, StreamResponse]:
         """Execute the view handler."""
 
 
+class ResolveResult(TypedDict):
```
webknjaz, line 122: add a docstring?

webknjaz:
@bdraco does this need to include a minimal version bump in `setup.cfg` or is it API-compatible?

bdraco:
It's going to need aiodns bumped.  It was just released so I'll get that done after breakfast

bdraco:
I would like to backport this fix to 3.9 as well since AsyncResolver is currently unusable without it (currently if the system supports IPv6 and there are no AAAA records AsyncResolver doesn't fallback to A in many cases)

Ideally we can switch the default back to AsyncResolver now that the semantics should be the same as ThreadedResolver but I think that change should wait until 3.10

Dreamsorcerer:
> I would like to backport this fix to 3.9

Are we going to break anything if someone is using an older aiodns version? I'd prefer to avoid changing the dependency requirements in a patch version.

bdraco:
> > I would like to backport this fix to 3.9
> 
> 
> 
> Are we going to break anything if someone is using an older aiodns version? I'd prefer to avoid changing the dependency requirements in a patch version.

Yes it would. aiodns does not expose the needed API until the release done earlier today.

aiohttp/abc.py
bdraco: Good call. added

CHANGES/8270.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Fix AsyncResolver to match ThreadedResolver behavior
+-- by :user:`bdraco`.
+
+On system with IPv6 support, the AsyncResolver would not fallback
+to providing A records when AAAA records were not available.
+Additionally, unlikely the ThreadedResolver, the AsyncResolver
```
bdraco:
```suggestion
Additionally, unlike the ThreadedResolver, the AsyncResolver
```

webknjaz:
> > I would like to backport this fix to 3.9
> 
> Are we going to break anything if someone is using an older aiodns version? I'd prefer to avoid changing the dependency requirements in a patch version.

Technically, if *we* don't change our API, it's not a reason for us to include this in a minor or major version bump. SemVer-wise, that is. Fulfilling the dependencies is a job of a dependency resolver/installer. Similar to setting a specific `python_requires` value that wouldn't be an API change and wouldn't be breaking from this perspective.

CHANGES/8270.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Fix AsyncResolver to match ThreadedResolver behavior
+-- by :user:`bdraco`.
+
+On system with IPv6 support, the AsyncResolver would not fallback
+to providing A records when AAAA records were not available.
+Additionally, unlike the ThreadedResolver, the AsyncResolver
```
webknjaz: Could you double-backtick the identifiers?

aiohttp/abc.py
webknjaz: Perhaps, we should also link it from `abc.rst`?

CHANGES/8270.bugfix.rst
bdraco: 4f492651b3ae3e693cbbf33916fe05d73e2d046e

CHANGES/8270.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Fix ``AsyncResolver`` to match ``ThreadedResolver`` behavior
+-- by :user:`bdraco`.
+
+On system with IPv6 support, the :py:class:`~aiohttp.resolver.AsyncResolver` would not fallback
```
webknjaz, line 4: Is it linkable? Looking at https://webknjaz.github.io/intersphinx-untangled/docs.aiohttp.org/, it doesn't appear so. You'd need to add it to https://github.com/aio-libs/aiohttp/blob/1542348/docs/conf.py#L396 or document in RST. Otherwise, I expect Sphinx to choke: https://github.com/aio-libs/aiohttp/actions/runs/8499653747/job/23280716056?pr=8270#step:13:129.

bdraco: Thanks. Sadly it looks like they are undocumented so 3d5175f2ddeec003d01793a4c7bd7319c7abc25d

webknjaz: Yeah, that's why I initially suggested double backticks as this is the easiest way to highlight the identifiers without having to deal with Sphinx. Marking as resolved.

docs/conf.py
```diff
@@ -394,6 +394,8 @@
     ("py:class", "aiohttp.ClientRequest"),  # undocumented
     ("py:class", "aiohttp.payload.Payload"),  # undocumented
     ("py:class", "aiohttp.abc.AbstractResolver"),  # undocumented
+    ("py:class", "aiohttp.abc.AsyncResolver"),  # undocumented
+    ("py:class", "aiohttp.abc.ThreadedResolver"),  # undocumented
```
webknjaz:
```suggestion
    ("py:class", "aiohttp.resolver.AsyncResolver"),  # undocumented
    ("py:class", "aiohttp.resolver.ThreadedResolver"),  # undocumented
```

bdraco: Thanks 👍

aiohttp/abc.py
bdraco: It looks like the `AbstractResolver` isn't mentioned at all in `abc.rst`

bdraco: I added it to `abc.rst`

Dreamsorcerer:
> > > I would like to backport this fix to 3.9
> > 
> > 
> > Are we going to break anything if someone is using an older aiodns version? I'd prefer to avoid changing the dependency requirements in a patch version.
> 
> Technically, if _we_ don't change our API, it's not a reason for us to include this in a minor or major version bump. SemVer-wise, that is. Fulfilling the dependencies is a job of a dependency resolver/installer. Similar to setting a specific `python_requires` value that wouldn't be an API change and wouldn't be breaking from this perspective.

Sure, I just think it's preferable to minimise requirement changes in order to get security/bug fixes. I'll let you two decide, I don't mind too much.

bdraco:
The interface is flexible enough that anyone who needs to fix can copy the code until it's released so I'm fine with leaving it in the current state for 3.9. 

It looks like anyone using it already has a workaround in place of setting AF_INET instead of unspecified so they have been living with this issue for years anyways. That's not great for IPv6 adoption but they also have ThreadedResolver even if its performance is worse so there are viable workarounds here now.

docs/conf.py
```diff
@@ -394,6 +394,8 @@
     ("py:class", "aiohttp.ClientRequest"),  # undocumented
     ("py:class", "aiohttp.payload.Payload"),  # undocumented
     ("py:class", "aiohttp.abc.AbstractResolver"),  # undocumented
```
webknjaz: Since you're documenting it, the ignore shouldn't be needed anymore:
```suggestion
```

bdraco:
Got distracted yesterday with the HA release so I didn't have time to get back to this.

I'm going to push this to another production system to be sure everything is OK shortly before merging

bdraco:
So far so good on both production systems.

Will check it again after dinner before merging.

bdraco:
All good

Also good with IPv6 disabled in the kernel.

I can't think of any other way to beat this up. So I'm happy with it now

PR #8280, opened 2024-04-01
Fix handling of multipart/form-data

Dreamsorcerer:
https://datatracker.ietf.org/doc/html/rfc7578

### DIFF
--- a/CHANGES/8280.breaking.rst
+++ b/CHANGES/8280.breaking.rst
@@ -1,1 +1,2 @@
-Removed deprecated `content_transfer_encoding` parameter from `FormData.add_field()` -- by :user:`Dreamsorcerer`.
+Removed deprecated ``content_transfer_encoding`` parameter from :py:meth:`FormData.add_field()
+<aiohttp.FormData.add_field>` -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8280.breaking.rst
+++ b/CHANGES/8280.breaking.rst
@@ -1,2 +1,0 @@
-Removed deprecated ``content_transfer_encoding`` parameter from :py:meth:`FormData.add_field()
-<aiohttp.FormData.add_field>` -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8280.breaking.rst
+++ b/CHANGES/8280.breaking.rst
@@ -1,0 +1,1 @@
+Removed deprecated `content_transfer_encoding` parameter from `FormData.add_field()` -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8280.bugfix.rst
+++ b/CHANGES/8280.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed `multipart/form-data` compliance with RFC 7578 -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8280.bugfix.rst
+++ b/CHANGES/8280.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed `multipart/form-data` compliance with RFC 7578 -- by :user:`Dreamsorcerer`.
+Fixed ``multipart/form-data`` compliance with :rfc:`7578` -- by :user:`Dreamsorcerer`.
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -262,13 +262,18 @@ class BodyPartReader:
         content: StreamReader,
         *,
         _newline: bytes = b"\r\n",
+        subtype: str = "mixed",
+        default_charset: Optional[str] = None
     ) -> None:
         self.headers = headers
         self._boundary = boundary
         self._newline = _newline
         self._content = content
+        self._default_charset = default_charset
         self._at_eof = False
-        length = self.headers.get(CONTENT_LENGTH, None)
+        self._is_form_data = subtype == "form-data"
+        # https://datatracker.ietf.org/doc/html/rfc7578#section-4.8
+        length = None if self._is_form_data else self.headers.get(CONTENT_LENGTH, None)
         self._length = int(length) if length is not None else None
         self._read_bytes = 0
         self._unread: Deque[bytes] = deque()
@@ -346,6 +351,8 @@ class BodyPartReader:
 
         self._read_bytes += len(chunk)
         if self._read_bytes == self._length:
+            self._at_eof = True
+        if self._content.at_eof():
             self._at_eof = True
         if self._at_eof:
             newline = await self._content.readline()
@@ -486,7 +493,8 @@ class BodyPartReader:
         """
         if CONTENT_TRANSFER_ENCODING in self.headers:
             data = self._decode_content_transfer(data)
-        if CONTENT_ENCODING in self.headers:
+        # https://datatracker.ietf.org/doc/html/rfc7578#section-4.8
+        if not self._is_form_data and CONTENT_ENCODING in self.headers:
             return self._decode_content(data)
         return data
 
@@ -520,7 +528,7 @@ class BodyPartReader:
         """Returns charset parameter from Content-Type header or default."""
         ctype = self.headers.get(CONTENT_TYPE, "")
         mimetype = parse_mimetype(ctype)
-        return mimetype.parameters.get("charset", default)
+        return mimetype.parameters.get("charset", self._default_charset or default)
 
     @reify
     def name(self) -> Optional[str]:
@@ -581,10 +589,18 @@ class MultipartReader:
         *,
         _newline: bytes = b"\r\n",
     ) -> None:
+        self._mimetype = parse_mimetype(headers[CONTENT_TYPE])
+        assert self._mimetype.type == "multipart", "multipart/* content type expected"
+        if "boundary" not in self._mimetype.parameters:
+            raise ValueError(
+                "boundary missed for Content-Type: %s" % self.headers[CONTENT_TYPE]
+            )
+
         self.headers = headers
         self._boundary = ("--" + self._get_boundary()).encode()
         self._newline = _newline
         self._content = content
+        self._default_charset = None
         self._last_part: Optional[Union["MultipartReader", BodyPartReader]] = None
         self._at_eof = False
         self._at_bof = True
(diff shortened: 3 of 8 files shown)


tests/test_web_functional.py
```diff
@@ -523,7 +504,7 @@ async def handler(request):
         return web.Response()
 
     form = FormData()
-    form.add_field("name", b"123", content_transfer_encoding="base64")
+    form.add_field("name", b"123")
```
Dreamsorcerer: Not sure how to fix this. Currently, if the value is bytes and content_transfer_encoding is not sent, then the code assumes that it is a file. This seems kind of dumb. Not sure how best to deal with this, given that content_transfer_encoding should never be sent.

Ironically, the spec says:
> Currently, no deployed implementations that send such bodies have been discovered.
https://datatracker.ietf.org/doc/html/rfc7578#section-4.8

Well, I just discovered one...

bdraco:
I started digging into this one

Unrelated but it looks there is blocking I/O in the event loop here:
 
https://github.com/aio-libs/aiohttp/blob/742a7d94a239bc8804a670e17e31e8c0564eb8c4/aiohttp/web_request.py#L727
https://github.com/aio-libs/aiohttp/blob/742a7d94a239bc8804a670e17e31e8c0564eb8c4/aiohttp/web_request.py#L731
https://github.com/aio-libs/aiohttp/blob/742a7d94a239bc8804a670e17e31e8c0564eb8c4/aiohttp/web_request.py#L739

bdraco:
Its a bit unexpected that any formdata that is bytes is treated as a file

https://github.com/aio-libs/aiohttp/blob/742a7d94a239bc8804a670e17e31e8c0564eb8c4/aiohttp/formdata.py#L56

tests/test_web_functional.py
bdraco: On the client side, I think the only way to avoid a breaking change is to keep `content_transfer_encoding` in `add_field` for the client side but ignore it on the server side

Dreamsorcerer:
> Its a bit unexpected that any formdata that is bytes is treated as a file

That was my initial thought too. It might be a convenience if you avoid bytes for typical use, but a little confusing otherwise. Should we switch v4 to only create files with `filename` is passed or similar? We could use a sentinel so that `filename=None` would still create a file?

CHANGES/8280.breaking.rst
```diff
@@ -0,0 +1 @@
+Removed deprecated `content_transfer_encoding` parameter from `FormData.add_field()` -- by :user:`Dreamsorcerer`.
```
webknjaz: Remember that RST uses double backticks.
```suggestion
Removed deprecated ``content_transfer_encoding`` parameter from :py:meth:`FormData.add_field()
<aiohttp.FormData.add_field>` -- by :user:`Dreamsorcerer`.
```

CHANGES/8280.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed `multipart/form-data` compliance with RFC 7578 -- by :user:`Dreamsorcerer`.
```
webknjaz: RST (docutils) has a built-in role for RFCs (and PEPs for that matter):
```suggestion
Fixed ``multipart/form-data`` compliance with :rfc:`7578` -- by :user:`Dreamsorcerer`.
```

aiohttp/multipart.py
```diff
@@ -636,7 +652,24 @@ async def next(
...
+            self._last_part is None
+            and self._mimetype.subtype == "form-data"
+            and isinstance(part, BodyPartReader)
+        ):
+            _, params = parse_content_disposition(part.headers.get(CONTENT_DISPOSITION))
+            if params.get("name") == "_charset_":
+                # Longest encoding in https://encoding.spec.whatwg.org/encodings.json
+                # is 19 characters, so 32 should be more than enough for any valid encoding.
+                charset = await part.read_chunk(32)
+                if len(charset) >= 32:
```
webknjaz: Is `read_chunk(32)` able to return more than 32 bytes? I thought that argument is the max size...
Also, `>=` includes `==` so do you expect 31 bytes max?

aiohttp/multipart.py
```diff
@@ -636,7 +652,24 @@ async def next(
...
+                    raise RuntimeError("Invalid default charset")
```
webknjaz, line 658: Could you cover this branch in tests? It should be triggered with a charset of exactly 32 chars per my understanding.

aiohttp/multipart.py
```diff
@@ -870,32 +900,44 @@ def append(self, obj: Any, headers: Optional[Mapping[str, str]] = None) -> Paylo
...
+                encoding = None
+
+            # te encoding
+            te_encoding = payload.headers.get(
+                CONTENT_TRANSFER_ENCODING,
+                "",
+            ).lower()
+            if te_encoding not in ("", "base64", "quoted-printable", "binary"):
+                raise RuntimeError(
+                    "unknown content transfer encoding: {}" "".format(te_encoding)
```
webknjaz: weird str concat ^

aiohttp/multipart.py
Dreamsorcerer: > Is `read_chunk(32)` able to return more than 32 bytes? I thought that argument is the max size...

No, I just always write such comparisons defensively. Can be changed to `==` if preferred.

> Also, `>=` includes `==` so do you expect 31 bytes max?

As per the comment, I expect something closer to 19 bytes max, just picked the next power of 2 to give it some leeway. e.g. Maybe it includes the line endings, which would make it 21 bytes. Maybe somebody includes some whitespace characters that pushes it to 22-24 bytes or similar.

I just needed to pick a value to avoid any DoS issue with reading an entire chunk into memory. This should be more than enough for anything valid.

aiohttp/multipart.py
Dreamsorcerer: Yes, will do test coverage once everything's passing.

aiohttp/multipart.py
Dreamsorcerer: Ah, I just indented it. Good opportunity to fix it. There's still that open PR somewhere to resolve these in the linter.

aiohttp/multipart.py
webknjaz: It probably makes sense to put `32` into a constant and reuse it.

It's just confusing to read that you first request maximum 32 bytes and then check if it's not more even though the called function should be limiting that already. Should the check be `>`?

tests/test_web_functional.py
Dreamsorcerer: Are we in agreement to switch the behaviour in v4 to send files only when filename is passed (even if `filename=None`)?
This is probably the last blocker to getting 3.9.4 out.

bdraco: It looks like the example to upload a file doesn't pass filename so I expect it might be quite common

https://docs.aiohttp.org/en/stable/client_quickstart.html#post-a-multipart-encoded-file

Dreamsorcerer: > It looks like the example to upload a file doesn't pass filename so I expect it might be quite common
> 
> https://docs.aiohttp.org/en/stable/client_quickstart.html#post-a-multipart-encoded-file

The first example isn't using FormData, so shouldn't be affected by this change. The second example uses FormData in order to include a filename. Which suggests to me that most cases of using FormData won't be passing a file without a filename.

Dreamsorcerer: Both examples are also file objects, not bytes, further suggesting that passing bytes as a file is not a common thing in the first place.

Dreamsorcerer: Actually, that does appear to just end up calling the same method on FormData in the end, so could cause an issue when passed directly in post(). The example still uses file objects though, and other parts of the documentation utilise BytesIO, so it would still seem surprising to me to get a file field when passing bytes.

bdraco: > so it would still seem surprising to me to get a file field when passing bytes.

I agree that is unexpected. It seems reasonable to have a breaking change to require `filename` instead of `bytes/bytearray/memoryview` in 4.0

Dreamsorcerer: OK, reviewing again, I don't think `filename=None` would make sense, as it's the presence of a filename that makes it a FileField. The bytes version just sets filename to name, so I think we should just remove that bit of logic.

We can then add warnings, so if a user uses content_transfer_encoding it tells them to use a BytesPayload directly if they want forward compatibility, and if they pass bytes it tells them to use the filename parameter or use BytesIO.

Dreamsorcerer:
OK, I think this should be fine now. Will need a followup PR to do the actual removal in master.

bdraco:
I'll be back home in an hour and can throw this on production to make sure there's no unexpected side effects

bdraco:
The backport to 3.9 is messy because it look like https://github.com/aio-libs/aiohttp/pull/3492 isn't in 3.9

bdraco:
I squashed this, cherry-picked fcedc6603c18c96219bc3ac48484475d8e1cfc25, than cherry-picked this on to my 3.9 integration branch.

Testing that now

bdraco:
Started up OK, and file uploads work. Watching the logs now.

bdraco:
Nothing obvious showing up in the logs.  Everything I could think to test still works

Dreamsorcerer:
> The backport to 3.9 is messy because #3492 isn't in 3.9

Guess we just backport that first then? Clearly got missed from any backports.

bdraco:
> > The backport to 3.9 is messy because #3492 isn't in 3.9
> 
> Guess we just backport that first then? Clearly got missed from any backports.

While I'm not a big fan of #3492, given that PR is already accepted and merged, that seems like the most reasonable path forward.

Dreamsorcerer:
> > > The backport to 3.9 is messy because #3492 isn't in 3.9
> > 
> > 
> > Guess we just backport that first then? Clearly got missed from any backports.
> 
> While I'm not a big fan of the complexity added by #3492, given that PR is already accepted and merged, and this fix is developed on top of it, that seems like the most reasonable path forward.

Wait, is this multipart reader server-side or client-side? If it's server-side then the change will likely introduce another HTTP request smuggling vulnerability. If it's only server-side, then we'd probably be best reverting the commit. If it's both, then we probably want to backport it and add a followup PR to disable it when the parser is not in lax mode...

Dreamsorcerer:
> > > > The backport to 3.9 is messy because #3492 isn't in 3.9
> > > 
> > > 
> > > Guess we just backport that first then? Clearly got missed from any backports.
> > 
> > 
> > While I'm not a big fan of the complexity added by #3492, given that PR is already accepted and merged, and this fix is developed on top of it, that seems like the most reasonable path forward.
> 
> Wait, is this multipart reader server-side or client-side? If it's server-side then the change will likely introduce another HTTP request smuggling vulnerability. If it's only server-side, then we'd probably be best reverting the commit. If it's both, then we probably want to backport it and add a followup PR to disable it when the parser is not in lax mode...

I think this is server-side only, I think the best option right now is to revert that change as nobody seems to have missed the feature, given that it failed to make it into a release for 5 years.

Dreamsorcerer:
There are a couple of lines of coverage dropped in multipart.py:
https://app.codecov.io/gh/aio-libs/aiohttp/pull/8280/indirect-changes

I'm unclear how to get that cover that again, or whether the code should maybe be deleted now. If anyone figures it out by tomorrow, feel free to just edit the PR directly.

PR #8299, opened 2024-04-05
Create marker for internal tests

Dreamsorcerer:
To help avoid issues with downstream packagers etc. we should disable a few tests for them which are dependent on the CI environment.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -213,7 +213,7 @@ jobs:
         PIP_USER: 1
       run: >-
         PATH="${HOME}/Library/Python/3.11/bin:${HOME}/.local/bin:${PATH}"
-        pytest -m internal
+        pytest
       shell: bash
     - name: Re-run the failing tests with maximum verbosity
       if: failure()
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -213,7 +213,7 @@ jobs:
         PIP_USER: 1
       run: >-
         PATH="${HOME}/Library/Python/3.11/bin:${HOME}/.local/bin:${PATH}"
-        pytest
+        pytest -m internal
       shell: bash
     - name: Re-run the failing tests with maximum verbosity
       if: failure()
--- a/CHANGES/8299.packaging.rst
+++ b/CHANGES/8299.packaging.rst
@@ -1,2 +1,2 @@
-Added an 'internal' pytest marker for tests which should be skipped
+Added an ``internal`` pytest marker for tests which should be skipped
 by packagers and disabled them by default -- by :user:`Dreamsorcerer`.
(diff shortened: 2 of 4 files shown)


Dreamsorcerer:
Some of the tests in test_websocket_writer.py are a good candidate to add to this later, as the binary tests depend on the version of zlib installed. However, some new tests should also be added that test the cleartext as well, so there are still tests for packagers to run. I'll leave this as a future exercise.

CHANGES/8299.packaging.rst
```diff
@@ -0,0 +1,2 @@
+Added an 'internal' pytest marker for tests which should skipped
```
Dreamsorcerer:
```suggestion
Added an 'internal' pytest marker for tests which should be skipped
```

bdraco (approved):

Dreamsorcerer:
Well, that's not what I intended... :P

Dreamsorcerer:
I'm not sure if pytest has an option that simply cancels out the `-m 'not internal'`. The only option appears to be run only the tests marked internal...

.github/workflows/ci-cd.yml
```diff
@@ -213,7 +213,7 @@ jobs:
         PIP_USER: 1
       run: >-
         PATH="${HOME}/Library/Python/3.11/bin:${HOME}/.local/bin:${PATH}"
-        pytest
+        pytest -m internal
```
bdraco:
```suggestion
        pytest
```

I think this is not needed

Dreamsorcerer: It is while 'not internal' is in the default arguments. If nobody has a better idea, then we'll probably have to skip disabling it by default and just recommend packagers to explicitly disable it in their test runs.

bdraco: I don't see a way to override `not internal`

CHANGES/8299.packaging.rst
```diff
@@ -0,0 +1,2 @@
+Added an 'internal' pytest marker for tests which should be skipped
```
webknjaz: Let's highlight it:
```suggestion
Added an ``internal`` pytest marker for tests which should be skipped
```

webknjaz: It might make sense to either symlink a "contrib" change note or make a separate one, with a usage example and a code snippet.

.github/workflows/ci-cd.yml
webknjaz: I wouldn't add it to the default arguments — the downstreams can read changelog and add markers.

setup.cfg
```diff
@@ -134,6 +134,8 @@ addopts =
 
     # run tests that are not marked with dev_mode
     -m "not dev_mode"
+    # skip internal by default for other users (re-enabled explicitly in CI).
+    -m "not internal"
```
Dreamsorcerer:
```suggestion
```

PR #8304, opened 2024-04-07
Check for backports in CI

Dreamsorcerer:
Add a check for backport labels due to several PRs having missed backports by mistake.
We have a backport:skip label if we explicitly don't want to backport a PR.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -26,6 +26,15 @@ env:
 
 permissions: {}
 jobs:
+  backport:
+    runs-on: ubuntu-latest
+    name: Backport label added?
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -26,6 +26,19 @@ env:
...
+    if: ${{ github.base_ref == 'master' }}
+    steps:
+    - name: Check labels
+      run: |
+        for label in "${{ github.event.pull_request.labels }}"; do
+            case $label in backport*)
+                exit 0
+            esac
+        done
+        exit 1
```
Dreamsorcerer: @webknjaz Let me know if you've got any ideas how to do this correctly, I'm just guessing.

PR #8317, opened 2024-04-08
Escape filenames and paths in HTML when generating index pages

Dreamsorcerer:


Dreamsorcerer:
@bdraco We have a problem with Windows on this one.

tests/test_web_urldispatcher.py
```diff
@@ -47,14 +59,17 @@ async def test_access_root_of_static_handler(
...
     # Try to access the root of static file server, and make
     # sure that correct HTTP statuses are returned depending if we directory
     # index should be shown or not.
-    my_file = tmp_path / "my_file"
-    my_dir = tmp_path / "my_dir"
+    # Ensure that html in file names is escaped.
+    # Ensure that links are url quoted.
+    my_file = tmp_path / "<img src=0 onerror=alert(1)>.txt"
+    my_dir = tmp_path / "<img src=0 onerror=alert(1)>.dir"
     my_dir.mkdir()
```
Dreamsorcerer: Looks like these paths aren't valid, shall we change the test to Linux only, or create a new test?

tests/test_web_urldispatcher.py
```diff
@@ -47,12 +54,15 @@ async def test_access_root_of_static_handler(
...
     prefix: str,
+    request_path: str,
     data: Optional[bytes],
 ) -> None:
     # Tests the operation of static file server.
     # Try to access the root of static file server, and make
     # sure that correct HTTP statuses are returned depending if we directory
     # index should be shown or not.
+    # Ensure that html in file names is escaped.
+    # Ensure that links are url quoted.
```
Dreamsorcerer:
```suggestion
```

PR #8320, opened 2024-04-11
Fix Python parser to mark responses without length as closing

Dreamsorcerer:


aiohttp/http_parser.py
```diff
@@ -706,7 +706,16 @@ def parse_message(self, lines: List[bytes]) -> RawResponseMessage:
         ) = self.parse_headers(lines)
 
         if close is None:
-            close = version_o <= HttpVersion10
+            if version_o <= HttpVersion10:
+                close = True
+            # https://www.rfc-editor.org/rfc/rfc9112.html#name-message-body-length
+            elif 100 <= status_i < 200 or status_i in {204, 304}:
+                close = False
+            elif headers.keys() & {hdrs.CONTENT_LENGTH, hdrs.TRANSFER_ENCODING}:
```
Dreamsorcerer: Case-sensitivity doesn't extend to set operations.
```suggestion
            elif hdrs.CONTENT_LENGTH in headers or hdrs.TRANSFER_ENCODING in headers:
```

PR #8338, opened 2024-04-15
Add some coverage

Olegt0rr:
## What do these changes do?

Increases coverage

No

## Related issue number

No issues

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -3610,3 +3610,30 @@ async def test_rejected_upload(aiohttp_client: Any, tmp_path: Any) -> None:
         "/ok", timeout=aiohttp.ClientTimeout(total=0.01)
     ) as resp_ok:
         assert 200 == resp_ok.status
+
+
+async def test_request_with_wrong_ssl_type(aiohttp_client: AiohttpClient) -> None:
+    app = web.Application()
+    session = await aiohttp_client(app)
+
+    with pytest.raises(TypeError, match="ssl should be SSLContext, Fingerprint, .*"):
+        await session.get("/", ssl=42)  # type: ignore[arg-type]
+
+
+async def test_request_with_wrong_proxy(aiohttp_client: AiohttpClient) -> None:
+    app = web.Application()
+    session = await aiohttp_client(app)
+
+    with pytest.raises(TypeError):
+        await session.get("/", proxy=42)  # type: ignore[arg-type]
+
+
+async def test_raise_for_status_is_none(aiohttp_client: AiohttpClient) -> None:
+    async def handler(_: web.Request) -> web.Response:
+        return web.Response()
+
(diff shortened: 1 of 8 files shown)


tests/test_worker.py
```diff
@@ -19,7 +19,7 @@
 
 try:
     import uvloop
-except ImportError:
+except ImportError:  # pragma: no cover
```
Dreamsorcerer: I think this should be covered by pypy, so it shouldn't be ignored. We just need pypy to stop segfaulting half the time....

Olegt0rr: Sure, but it seems this case already covered by the code anyway

```
PARAMS = [AsyncioWorker]
if uvloop is not None:
    PARAMS.append(UvloopWorker)
```

And then test checks both workers.

Dreamsorcerer: I don't see that as a reason to skip coverage here?

Olegt0rr: Cause it looks redundant here - case with `uvloop is None` already covered anyway

But if you are against it, then I can remove it `¯\_(ツ)_/¯`

```suggestion
except ImportError:
```

tests/test_client_functional.py
```diff
@@ -3610,3 +3610,30 @@ async def not_ok_handler(request):
...
+
+    with pytest.raises(TypeError, match="ssl should be SSLContext, Fingerprint, .*"):
+        await session.get("/", ssl=42)  # type: ignore[arg-type]
+
+
+async def test_request_with_wrong_proxy(aiohttp_client: AiohttpClient) -> None:
+    app = web.Application()
+    session = await aiohttp_client(app)
+
+    with pytest.raises(TypeError):
```
Olegt0rr: Can't understand why my tests can't catch `InvalidURL` here :(

https://github.com/aio-libs/aiohttp/blob/5a6949da642d1db6cf414fd0d1f70e54c7b7be14/aiohttp/client.py#L431-L435

webknjaz: Probably because `TypeError` is not a subclass of `ValueError`.

Olegt0rr: Great sign for me to go to sleep)) thanks

tests/test_worker.py
Dreamsorcerer: Yep, I'm just unsure why redundancy would be a reason to ignore coverage (otherwise, we'd surely ignore 90% of the code...). The main reason to ignore code is that we know it'll never be run (primarily typing code). Another reason recently is some retry logic within a test used to make the test more stable, but the coverage could vary between test runs and confuse results.

PR #8340, opened 2024-04-16
Release v3.9.5

Dreamsorcerer:


CHANGES.rst
```diff
@@ -10,6 +10,39 @@
...
+
+- Fixed regression with adding Content-Disposition to form-data part after appending to writer -- by :user:`Dreamsorcerer`/:user:`Olegt0rr`.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8332`.
+
+
+
+- Added default Content-Disposition in multipart/form-data responses -- by :user:`Dreamsorcerer`.
```
webknjaz: Some highlighting wouldn't hurt:
```suggestion
- Added default ``Content-Disposition`` in ``multipart/form-data`` responses -- by :user:`Dreamsorcerer`.
```

CHANGES.rst
```diff
@@ -10,6 +10,39 @@
...
+
+- Fixed regression with adding Content-Disposition to form-data part after appending to writer -- by :user:`Dreamsorcerer`/:user:`Olegt0rr`.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8332`.
+
+
+
+- Added default ``Content-Disposition`` in ``multipart/form-data`` responses -- by :user:`Dreamsorcerer`.
```
webknjaz: Also, it's not clear what was broken from this entry, which would be useful since it's under the "bugfixes" section.

CHANGES.rst
```diff
@@ -10,6 +10,39 @@
...
+
+- Fixed "Unclosed client session" when initialization of ClientSession fails -- by :user:`NewGlad`.
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8253`.
+
+
+
+- Fixed regression with adding Content-Disposition to form-data part after appending to writer -- by :user:`Dreamsorcerer`/:user:`Olegt0rr`.
```
webknjaz:
```suggestion
- Fixed regression with adding ``Content-Disposition`` to the ``form-data`` part after appending to writer -- by :user:`Dreamsorcerer`/:user:`Olegt0rr`.
```

webknjaz: It might be useful to link the PR that caused the regression, using the `:pr:` role, bake it into the phrasing.

CHANGES.rst
```diff
@@ -10,6 +10,39 @@
 
 .. towncrier release notes start
 
+3.9.5 (2024-04-16)
+==================
+
+Bug fixes
+---------
+
+- Fixed "Unclosed client session" when initialization of ClientSession fails -- by :user:`NewGlad`.
```
webknjaz: Let's make it clickable:
```suggestion
- Fixed "Unclosed client session" when initialization of
:py:class:`~aiohttp.ClientSession` fails -- by :user:`NewGlad`.
```

webknjaz:
@Dreamsorcerer by the way, I figured out how to get rid of those trailing spaces in the template but haven't yet copied it over to other projects: https://github.com/cherrypy/cheroot/blob/86cb5ba/docs/changelog-fragments.d/.towncrier-template.rst.j2#L51-L71. The solution reads as a huge hack, though...

aiohttp/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "3.9.4.dev0"
```
webknjaz, line 1: Oh, does this mean that we forgot to bump it in the previous release? If did we bump it backwards? `3.9.4.dev0 < 3.9.4`

Dreamsorcerer: Nope, that's just my mistake. Let me update the instructions...

CHANGES.rst
```diff
@@ -10,6 +10,42 @@
 
 .. towncrier release notes start
```
webknjaz, line 12: FWIW, it seems like the previous changelog updates had two empty lines as a separator. Dunno if we want it that way, though.
```suggestion


```

webknjaz (approved):

CHANGES.rst
Dreamsorcerer: Once you've merged the updated template, I'll just update the existing changelog to follow whatever towncrier produces. But, this is what it produced when I ran it today.

Farhadsakhodi (approved):

PR #8353, opened 2024-04-19
Skip labels for dependabot

Dreamsorcerer:


.github/workflows/labels.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   backport:
     runs-on: ubuntu-latest
     name: Backport label added
+    if: ${{ github.actor != 'dependabot[bot]' }}
```
webknjaz, line 12: FYI you can also check for whether it's a GitHub App — the type would be `'Bot'`: https://api.github.com/users/dependabot[bot]

webknjaz (approved):

.github/workflows/labels.yml
Dreamsorcerer: I'm not clear on how to get to that as it doesn't appear to be on the github context (presumably it would require an API request which seems outside the scope of an if condition). As dependabot is the only thing that auto-merges, this is probably enough.

webknjaz: @Dreamsorcerer sure.

As for the github context, it should be retrievable through the event object like so `${{ github.event.pull_request.user.type }}` (the exact needs testing, I might've missed a level of nesting).

Checking the docs again, I also remembered why it's not always right to use `actor` for checks. An actor — is the one causing the action. So if you go into a pull request and add a label, then you'd be the actor while dependabot would remain the PR creator. So it seems to me that this check might fail if you were to edit a PR title of a Dependabot PR, or add/remove some label, or add a commit on top, or close+reopen it.

webknjaz: @Dreamsorcerer I confirmed that the flaw I described above indeed exists in how the user is checked: https://github.com/aio-libs/aiohttp/pull/8325#issuecomment-2067303116. We should check against the PR author here, not the actor.

Dreamsorcerer: Yep, if that works, then it should be better.

PR #8367, opened 2024-04-21
Delete size arg from StreamReader.feed_data #7265

Dreamsorcerer:
Finally completing the review for #7265.
Have added a small fix and now I think this is correct. Will double check it tomorrow before merging.


CHANGES/7265.breaking.rst
```diff
@@ -0,0 +1 @@
+Deleted `size` arg from `StreamReader.feed_data` -- by :user:`DavidRomanovizc`.
```
webknjaz: Don't forget it's rst, not md.

Dreamsorcerer:
@DavidRomanovizc Thanks for getting through that, turned out to be quite tricky in the end. If you're interested in picking up any other tasks, there should be plenty of easier ones that should be quick to review.

PR #8369, opened 2024-04-21
Update Matrix link

Dreamsorcerer:
Should probably update this now to the Matrix ID.


webknjaz:
Maybe update README as a part of this too?

PR #8384, opened 2024-04-25
Update .pre-commit-config.yaml

Dreamsorcerer:


aiohttp/web_response.py
```diff
@@ -176,7 +176,7 @@ def enable_compression(
     ) -> None:
         """Enables response compression encoding."""
         # Backwards compatibility for when force was a bool <0.17.
-        if type(force) == bool:
+        if isinstance(force, bool):
```
webknjaz, line 179: Are integers boolean? I remember there were some corner cases a around this. Or perhaps this was done with performance in mind.. FWIW this is not just a formatting change. It checks for subclasses too. Is this necessary?

tests/test_circular_imports.py
```diff
@@ -30,14 +31,16 @@ def _mark_aiohttp_worker_for_skipping(
     importables: List[str],
 ) -> List[Union[str, "ParameterSet"]]:
     return [
-        pytest.param(
-            importable,
-            marks=pytest.mark.skipif(
-                not hasattr(socket, "AF_UNIX"), reason="It's a UNIX-only module"
-            ),
+        (
```
webknjaz, line 34: Which linter wanted this? It seems like too much..

tests/test_proxy_functional.py
```diff
@@ -192,13 +192,16 @@ async def test_https_proxy_unsupported_tls_in_tls(
...
+    with pytest.warns(
+        RuntimeWarning,
+        match=expected_warning_text,
+    ), pytest.raises(
         ClientConnectionError,
         match=expected_exception_reason,
     ) as conn_err:
         await sess.get(url, proxy=secure_proxy_url, ssl=client_ssl_ctx)
 
-    assert type(conn_err.value.__cause__) == TypeError
```
webknjaz, line 201: I'm pretty sure the exact class is expected here..

tests/test_circular_imports.py
Dreamsorcerer: Isn't this one just black formatting? I didn't touch this one.

aiohttp/web_response.py
Dreamsorcerer: > Are integers boolean

No, other way round. bool is subclass of int, so this shoudn't make any difference. It's also ancient code that could be removed at this point anyway..

tests/test_proxy_functional.py
Dreamsorcerer: Didn't think it would make a difference. Feel free to revert and add a noqa.

PR #8405, opened 2024-05-09
Dependencies Update - Security Vulnerabilities fixed

GarryStraitYT:
## What do these changes do?

Updates dependencies to reduce security vulnerabilities

There is no changes in behavior that can be seen by the end user

I have not changed any core code but rather the version of the dependencies used by aiohttp to reflect the latest version.

## Related issue number

No related issues

## Checklist

No code was changed since last release. Dependencies were updated to prevent their security vulnerabilities from affecting the project.


Dreamsorcerer:
This is handled automatically by Dependabot (and has zero impact on users), but Pillow is failing due to needing to update or replace blockdiag: #8282 https://github.com/blockdiag/sphinxcontrib-blockdiag/issues/26

I think replacing it with the builtin graphviz extension might be a good approach if you'd like to contribute: https://www.sphinx-doc.org/en/master/usage/extensions/graphviz.html

PR #8432, opened 2024-06-01
Use cffi 1.17.0rc1 to test Python 3.13 support

edgarrmondragon:
## What do these changes do?

* Use a prerelease of `cffi` to support testing with Python 3.13

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


.github/workflows/ci-cd.yml
```diff
@@ -189,7 +189,7 @@ jobs:
             pip-ci-${{ runner.os }}-${{ matrix.pyver }}-${{ matrix.no-extensions }}-
     - name: Update pip, wheel, setuptools, build, twine
       run: |
-        python -m pip install -U pip wheel setuptools build twine
+        python -m pip install -U pip wheel setuptools build twine cffi==0.17.0rc1
```
webknjaz: This seems redundant as the following step would do the same.

edgarrmondragon: Yeah, but `pip` (I think) has a dependency on cffi via cryptography so this step was failing on 3.13

edgarrmondragon:
It seems [pycares](https://github.com/saghul/pycares) is also causing issues

.github/workflows/ci-cd.yml
webknjaz: Pip vendors all its deps so this would do nothing. Besides, all pip's deps are pure-python.

Dreamsorcerer:
Looks like uvloop is also failing to install currently. I suspect there's a good chance cffi will have made a stable release by the time that's sorted, so maybe not worth putting these changes in temporarily.

bdraco:
Looks like cffi 1.17.0 is now out

Dreamsorcerer:
I guess we close this then, as Dependabot will update it. Just waiting on uvloop now..

PR #8445, opened 2024-06-07
Fixes read_timeout on WS connection not respecting ws_connect's timeouts

### LINKED ISSUE #8444: Broken timeout system with ws_connect
### Describe the bug

# Context

We have very specific requirements with our HTTP connections: we expect extremely responsive API and reconnect rapidly if we don't receive responses because we consider it a potential failure.

To facilitate that we use `ClientTimeout(connect=1.5, sock_connect=1.0, sock_read=0.5)`, i.e. yes, if we attempt to read on a socket after sending the request and nothing arrives in 500 ms we consider it a failure and perform a retry. 

# Problem

This actually works great except when we use the `ws_connect` for websockets. Having nothing to read for 0.5s of `sock_read` results in a `ServerTimeoutError` requiring to retry with `ws_connect`.

Problem, however, is that `ws_connect`'s `timeout` and `receive_timeout` have no effect on the underlying timeout configuration because `_ws_connect`'s call to `self.request` does not alter the receive timeout for the connection by calling ` conn.protocol.set_response_params`. And after the WSS connection is established, disregarding the `ws_connect`'s `receive_timeout` the protocol inherits the parent's HTTP connection's `sock_read` of 0.5.

image

image

image

image

image

### To Reproduce

Above

### Expected behavior

`receive_timeout` passed to the `ws_connect` should be observed, by calling `conn.protocol.set_response_params` with `receive_timeout` as `read_timeout` 

### Logs/tracebacks

```python-traceback
N/A
```

### Python Version

```console
N/A
```

### aiohttp Version

```console
Latest stable
```

### multidict Version

```console
N/A
```

### yarl Version

```console
N/A
```

### OS

OS-independent

### Related component

Client

### Additional context

_No response_

### Code of Conduct

arcivanov:
Added `read_timeout` property to `ResponseHandler` to allow override

After WS(S) connection is established, adjust `conn.proto.read_timeout` to
be the largest of the `read_timeout` and the `ws_connect`'s
`timeout.ws_receive` with `None` treated as 'no timeout', i.e. the maximum.

fixes #8444

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -939,6 +939,17 @@ class ClientSession:
             assert conn is not None
             conn_proto = conn.protocol
             assert conn_proto is not None
+
+            # For WS connection the read_timeout must be either ws_timeout.ws_receive or greater
+            # None == no timeout, i.e. infinite timeout, so None is the max timeout possible
+            if ws_timeout.ws_receive is None:
+                # Reset regardless
+                conn_proto.read_timeout = None
+            elif conn_proto.read_timeout is not None:
+                conn_proto.read_timeout = max(
+                    ws_timeout.ws_receive, conn_proto.read_timeout
+                )
+
             transport = conn.transport
             assert transport is not None
             reader: FlowControlDataQueue[WSMessage] = FlowControlDataQueue(
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -240,6 +240,14 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
     def start_timeout(self) -> None:
         self._reschedule_timeout()
 
+    @property
+    def read_timeout(self) -> Optional[float]:
+        return self._read_timeout
+
+    @read_timeout.setter
+    def read_timeout(self, read_timeout: Optional[float]) -> None:
+        self._read_timeout = read_timeout
+
     def _on_read_timeout(self) -> None:
         exc = SocketTimeoutError("Timeout on reading data from socket")
         self.set_exception(exc)
--- a/tests/test_client_ws.py
+++ b/tests/test_client_ws.py
@@ -10,6 +10,7 @@ import pytest
 
 import aiohttp
 from aiohttp import client, hdrs
+from aiohttp.client_ws import ClientWSTimeout
 from aiohttp.http import WS_KEY
 from aiohttp.streams import EofStream
 from aiohttp.test_utils import make_mocked_coro
@@ -39,6 +40,7 @@ async def test_ws_connect(ws_key: Any, loop: Any, key_data: Any) -> None:
         hdrs.SEC_WEBSOCKET_ACCEPT: ws_key,
         hdrs.SEC_WEBSOCKET_PROTOCOL: "chat",
     }
+    resp.connection.protocol.read_timeout = None
     with mock.patch("aiohttp.client.os") as m_os:
         with mock.patch("aiohttp.client.ClientSession.request") as m_req:
             m_os.urandom.return_value = key_data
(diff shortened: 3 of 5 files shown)


aiohttp/client.py
```diff
@@ -939,6 +939,21 @@ async def _ws_connect(
...
             conn_proto = conn.protocol
             assert conn_proto is not None
+
+            # For WS connection the sock_read must be either receive_timeout
+            # or timeout (whichever is specified), unless read_timeout is greater
+            # None == no timeout, i.e. infinite timeout
+            if ws_timeout.ws_receive is None:
+                # Reset regardless
+                conn_proto.read_timeout = None
+            elif conn_proto.read_timeout is None:
```
bdraco: Is a bit unexpected to write this with a pass. It would be cleaner to reverse the condition and remove this block

aiohttp/client_proto.py
```diff
@@ -240,6 +240,14 @@ def _reschedule_timeout(self) -> None:
     def start_timeout(self) -> None:
         self._reschedule_timeout()
 
+    @property
```
bdraco, line 243: Ideally this would get set earlier so it doesn't need to be exposed as a property and setter.

arcivanov: It can't be. That's what differs HTTP connection and WS connection and that's why the bug exists.

When we upgrade from HTTP to WS the original HTTP request with the upgrade header needs to be done under the connection timeouts of HTTP connections. Once upgrade is established we need to *switch* connection timeout parameters to the WS mode. The crux is the upgrade.

aiohttp/client.py
arcivanov: I wrote this with a pass for the comment for logic clarity. I can of course reverse it if you'd like.

tests/test_client_ws.py
```diff
@@ -84,6 +174,7 @@ async def test_ws_connect_with_params(ws_key: Any, loop: Any, key_data: Any) ->
         hdrs.SEC_WEBSOCKET_ACCEPT: ws_key,
         hdrs.SEC_WEBSOCKET_PROTOCOL: "chat",
     }
+    resp.connection.protocol.read_timeout = None
```
bdraco, line 180: This seems like a breaking change if this now needs to be set in the below tests

arcivanov: I don't think so. This is because `Mock` is not comparable with `int` or `float` for the purposes of `max`.

arcivanov:
FYI I'm working on a backport for 3.9.

tests/test_client_ws.py
```diff
@@ -54,6 +56,94 @@ async def test_ws_connect(ws_key: Any, loop: Any, key_data: Any) -> None:
...
+    resp.status = 101
+    resp.headers = {
+        hdrs.UPGRADE: "websocket",
+        hdrs.CONNECTION: "upgrade",
+        hdrs.SEC_WEBSOCKET_ACCEPT: ws_key,
+        hdrs.SEC_WEBSOCKET_PROTOCOL: "chat",
+    }
+    resp.connection.protocol.read_timeout = 0.5
+    with mock.patch("aiohttp.client.os") as m_os:
+        with mock.patch("aiohttp.client.ClientSession.request") as m_req:
```
bdraco: I think you probably copied this from the above test, but please combine with statements

aiohttp/client.py
bdraco: Yes, please reverse

aiohttp/client_proto.py
bdraco: Ideally this would only be exposed for websocket connections as we need to take a look at all the places this is used and make sure it works as expected for everything since this exposes a new public setter.

tests/test_client_ws.py
arcivanov: I can do this, but I'm trying to keep uniformity between master and 3.9 patches. This is the state of affairs in the master. With your permission I'd like not to rework the entire `test_client_ws` suite - all tests there are like that.

bdraco: Its fine to leave the existing tests as-is. Ideally we don't create any new ones with that pattern though.

arcivanov: Sure, I'll fix this.

aiohttp/client_proto.py
arcivanov: Any suggestions? I unfortunately don't have time to dedicate to solving all the issues. 
We can make the property `_ws_read_timeout` to indicate specialization.
As-is this is absolutely harmless IMO.

Could I perhaps suggest this as an interim solution (since it works) and a followup ticket will "re-architect" this issue unless it's something obvious/trivial?

bdraco: I looked at all the usage, and changing it after construction doesn't seem like its a problem since we already allow changing it via `set_response_params` so I think its OK.  

The downside is that we have another method to maintain but at least its simple.

bdraco:
We will need to backport to 3.10 as well as I'm not sure if we plan on doing any more 3.9 releases before 3.10 starts rolling out.

arcivanov:
I want to leave this note here for whoever comes looking.

Between 3.9 and master ([f662958](https://github.com/aio-libs/aiohttp/commit/f662958b150a9d8d92fcbd0c9235e6bee1bedd67)) there was a change in the interpretation of WS timeouts.
`receive_timeout` and `timeout` in 3.9 became `ws_timeout.ws_receive` and `ws_timeout.ws_close` 
As such this solution only looks at `receive_timeout` in 3.9 and `ws_timeout.ws_receive` in master for the purposes of setting the `conn_proto.read_timeout`. 

Not considering `timeout` in 3.9, as I originally planned, will remove unexpected changes in behavior in 4.x when `timeout` becomes used for WS closing timeout only.

@bdraco

arcivanov:
> We will need to backport to 3.10 as well as I'm not sure if we plan on doing any more 3.9 releases before 3.10 starts rolling out.

Could I perhaps request a hotfix release with this for 3.9?  If you oblige, I'll backport to 3.10 right now :)

aiohttp/client.py
```diff
@@ -939,6 +939,17 @@ async def _ws_connect(
...
+            # For WS connection the read_timeout must be either ws_timeout.ws_receive or greater
+            # None == no timeout, i.e. infinite timeout, so None is the max timeout possible
+            if ws_timeout.ws_receive is None:
+                # Reset regardless
+                conn_proto.read_timeout = None
+            elif conn_proto.read_timeout is not None:
+                conn_proto.read_timeout = max(
+                    ws_timeout.ws_receive, conn_proto.read_timeout
+                )
+
```
Dreamsorcerer, line 984: Maybe I'm just tired and not understanding the issue, but can't we just pass the timeout above at:
https://github.com/aio-libs/aiohttp/blob/28f126cb673c258d8369a4fc2bcdf0ea4eef7a22/aiohttp/client.py#L851-L863

aiohttp/client_proto.py
Dreamsorcerer: If we want to make the attribute public, shouldn't we just rename it?

aiohttp/client.py
arcivanov: You can't do that. The above self.request is an HTTP request to upgrade to WS(S). That has to be done with the currently requested timeout settings for the HTTP connection. Once upgraded, the connection becomes WSS with WS-specific timeout settings.

That's the core of the bug and the crux of the issue. If you modify the connection here, you run the HTTP request to upgrade with (potentially infinite receive timeout) WS timeout settings.

arcivanov: WS is special that way that HTTP becomes WS. Nothing else in the library behaves that way, that's why it was missed - the connection morphs the semantics from HTTP behavior to WS behavior.

bdraco:
> > We will need to backport to 3.10 as well as I'm not sure if we plan on doing any more 3.9 releases before 3.10 starts rolling out.
> 
> @bdraco Could I perhaps request a hotfix release with this for 3.9? If you oblige, I'll backport to 3.10 right now 😄

That's one I can't answer as I've not done a release for this project before (only reviews + code) so I'll defer to other members to answer.

arcivanov:
@bdraco @Dreamsorcerer posted 3.10 backport as well. It's virtually identical to 3.9.

Dreamsorcerer:
> Between 3.9 and master ([f662958](https://github.com/aio-libs/aiohttp/commit/f662958b150a9d8d92fcbd0c9235e6bee1bedd67)) there was a change in the interpretation of WS timeouts.

I think that may be a missing backport. Seems sensible that the change should have been backported, same as the .request() method now uses ClientTimeout.

bdraco:
I'll push the 3.9 backport to a few production Home Assistant instances and run it overnight to test

arcivanov:
> > Between 3.9 and master ([f662958](https://github.com/aio-libs/aiohttp/commit/f662958b150a9d8d92fcbd0c9235e6bee1bedd67)) there was a change in the interpretation of WS timeouts.
> 
> I think that may be a missing backport. Seems sensible that the change should have been backported, same as the .request() method now uses ClientTimeout.

I don't thinks so. That's an API change and it was posted in 2019/0. It introduces new user-faces timeout data structures:

e51fb1ff1ebdde566b96af0090c5c63cf1a62b1b
c09c538e036619ae549280fb9051fd6084e8252c

image

Dreamsorcerer:
> That's one I can't answer as I've not done a release for this project before (only reviews + code) so I'll defer to other members to answer.

I'm not planning to do another 3.9 release unless there's a notable security issue (but, I think we can plan to do the 3.10 release soon if we want).

arcivanov:
> > That's one I can't answer as I've not done a release for this project before (only reviews + code) so I'll defer to other members to answer.
> 
> I'm not planning to do another 3.9 release unless there's a notable security issue (but, I think we can plan to do the 3.10 release soon if we want).

Pleeeeeeeease? :smile:  I really don't want to vendorize the entire aiohttp fork over this. Unless you're planning on releasing 3.10 this weekend?

Dreamsorcerer:
> I don't think so. That's an API change and it was posted in 2019/0. It introduces new user-facing timeout data structures in public API:

Yes, looks like the same change as ClientTimeout. The old parameters are deprecated, but still maintained for backwards compatibility in 3.x. I suspect the same should have been done for this. There are a lot of old PRs which missed backports by mistake (you'll see the PRs have no backport labels at all). It's why I added a CI check that backport labels are added to every PR, so we can be sure whether a PR was intended to be backported or not.

Dreamsorcerer:
> > > That's one I can't answer as I've not done a release for this project before (only reviews + code) so I'll defer to other members to answer.
> > 
> > 
> > I'm not planning to do another 3.9 release unless there's a notable security issue (but, I think we can plan to do the 3.10 release soon if we want).
> 
> Pleeeeeeeease? 😄 I really don't want to vendorize the entire aiohttp fork over this. Unless you're planning on releasing 3.10 this weekend?

I'm not doing any release this weekend. :P
The amount of work is probably less to release 3.10 than to do another 3.9 release.

arcivanov:
> > > > That's one I can't answer as I've not done a release for this project before (only reviews + code) so I'll defer to other members to answer.
> > > 
> > > 
> > > I'm not planning to do another 3.9 release unless there's a notable security issue (but, I think we can plan to do the 3.10 release soon if we want).
> > 
> > 
> > Pleeeeeeeease? 😄 I really don't want to vendorize the entire aiohttp fork over this. Unless you're planning on releasing 3.10 this weekend?
> 
> I'm not doing any release this weekend. :P The amount of work is probably less to release 3.10 than to do another 3.9 release.

Well, I had to try ;) 
Anyway, I've PRed 3.9 and 3.10 backports preserving the existing APIs. In the master the fix utilizes the new `ClientWSTimeout` APIs so you have both options.

Now I have to go patch the installed library at runtime to make sure this fix gets in.

Crying

bdraco:
Pushed to a few production Home Assistant instances. Nothing blew up.

Will report back in the morning before I'm traveling again.

bdraco:
All good overnight

bdraco (approved):

bdraco (approved):

bdraco:
@arcivanov We are getting closer to releasing 3.10.  Would you kindly take care of the backport?

arcivanov:
@bdraco I was under impression I already did. Is it conflicting? I'll confirm in a couple of hours.

bdraco:
You did, I forgot about them https://github.com/aio-libs/aiohttp/pull/8447

PR #8456, opened 2024-06-11
Add ClientConnectorDNSError for differentiating DNS errors from others

### LINKED ISSUE #8455: Separate exception for DNS errors

Currently the aiohttp client raises a `ClientConnectorError` when DNS resolution fails and there's no clean way to differentiate this `ClientConnectorError` from ones that happen at other points of the connection.

I'd like another Exception, maybe `ClientConnectorDNSError` that inherits from `ClientConnectorError` and is raised when DNS resolution fails.

I've considered inspecting the ClientConnectorError to determine whether it's a DNS error, but it's tough to do so without making my calling code aware of the implementation details of the underlying DNS resolver.

### Related component

Client

### Additional context

_No response_

### Code of Conduct

mstojcevich-cisco:
## What do these changes do?

Adds a ClientConnectorError that's specific to DNS resolution errors. See #8455

Existing usages should still keep working the same as before since this extends from ClientConnectorError.
Users can now catch ClientConnectorDNSError if they're looking for DNS errors specifically.

I can't see this being difficult to maintain since the Resolver abstraction already exists to determine when errors are related to DNS or some other part of the connection.

## Related issue number

Fixes #8455

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


tests/test_client_functional.py
```diff
@@ -2947,6 +2947,23 @@
...
+        async with aiohttp.request("GET", "http://wrong-dns-name.com"):
+            assert False, "never executed"  # pragma: no cover
+
+
+async def test_aiohttp_request_connector_error_non_dns() -> None:
+    # A non-dns counterpart to test_aiohttp_request_ctx_manager_not_found_exception_is_dns_specific
+    with pytest.raises(aiohttp.ClientConnectorError) as excinfo:
+        async with aiohttp.request("GET", "http://localhost:1/"):
+            assert False, "never executed"  # pragma: no cover
+    assert not isinstance(excinfo.value, aiohttp.ClientConnectorDNSError)
```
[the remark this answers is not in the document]
mstojcevich-cisco: Is there a comment I should add to ignore this, or is it fine to proceed with the warning?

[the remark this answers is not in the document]
webknjaz: There used to be comments back when the tool used to be called LGTM. I'm not sure anymore. But I can click the "dismiss alert" button since I can see that coverage is actually reported on this line.

OTOH, why don't you put the exception into `pytest.raises()`. That's what's supposed to be checked anyway.

I think CodeQL understands that `assert False` interrupts the control flow of a test function so it naturally thinks the outer line wouldn't be hit. It'd probably be better to call `pytest.fail()` there instead. I wonder if wrapping such a call into `assert_never()` would let us drop the coverage ignore comment as well...

CHANGES/8455.feature.rst
```diff
@@ -0,0 +1 @@
+Added `ClientConnectorDNSError` for differentiating DNS resolution errors from other connector errors.
```
webknjaz: Plz use RST, not MD.

webknjaz:
We need to think the DNS-related exception hierarchy through — perhaps it needs a base exception with subtypes. Similar to https://github.com/aio-libs/aiohttp/pull/6722#discussion_r1477103562.

Additionally, the documentation would need to be updated to list the new exceptions with explanations.

aiohttp/client_exceptions.py
```diff
@@ -166,6 +167,14 @@ def __str__(self) -> str:
     __reduce__ = BaseException.__reduce__
 
 
+class ClientConnectorDNSError(ClientConnectorError):
```
webknjaz, line 177: Not sure if this should be a connector error, though. Any reasoning?

CHANGES/8455.feature.rst
webknjaz: Here's a similar example: https://github.com/aio-libs/aiohttp/pull/6722/files#diff-06a2c33bab93f9984fab90f699c24c12c18cd0620eec8555deef04ae43b92925

aiohttp/client_exceptions.py
mstojcevich-cisco: It's currently a ConnectorError so it would allow existing code that tries to catch ConnectorErrors to keep working. I'm open to taking a different approach though.

webknjaz: Yeah, I understand. If you read the PR I linked, there were similar concerns, and I know that some bits of this might be breaking. In that instance, we ended up using multiple inheritance and intermediate exceptions in the hierarchy to maximize the compatibility while enabling flexibility and accuracy.

Dreamsorcerer: Is there much to discuss here? It looks to me like there's only 1 exception currently (there's only 1 place that raises this exception). It is also happening in the connector, so I don't really see any issue with it being a ConnectorError.

So, I think this change looks fine to me, and just needs the small bits of cleanup that has been asked for.

CHANGES/8455.feature.rst
bdraco:
```suggestion
Added :exc:`aiohttp.ClientConnectorDNSError` for differentiating DNS resolution errors from other connector errors.
```

tests/test_client_functional.py
```diff
@@ -3315,6 +3315,23 @@ async def test_aiohttp_request_ctx_manager_not_found() -> None:
...
```
bdraco: This test is redundant as there are already plenty of tests that raise for this case

tests/test_client_functional.py
```diff
@@ -3315,6 +3315,23 @@ async def test_aiohttp_request_ctx_manager_not_found() -> None:
...
 
 
+async def test_aiohttp_request_ctx_manager_not_found_exception_is_dns_specific() -> (
+    None
+):
+    # The error raised should be specific to DNS
+    with pytest.raises(aiohttp.ClientConnectorDNSError):
+        async with aiohttp.request("GET", "http://wrong-dns-name.com"):
+            assert False, "never executed"  # pragma: no cover
+
```
bdraco: This test should mock the resolver out so it does not do network I/O

tests/test_client_functional.py
bdraco: removed in 2089776ef6315521bc65b5fbeb9ea417214b9938

tests/test_client_functional.py
bdraco: mocked out the network I/O in 2089776ef6315521bc65b5fbeb9ea417214b9938

CHANGES/8455.feature.rst
bdraco:
```suggestion
Added :exc:`aiohttp.ClientConnectorDNSError` for differentiating DNS resolution errors from other connector errors.
```

CHANGES/8455.feature.rst
```diff
@@ -0,0 +1 @@
+Added :exc:`aiohttp.ClientConnectorDNSError` for differentiating DNS resolution errors from other connector errors.
```
bdraco:
```suggestion
Added :exc:`aiohttp.ClientConnectorDNSError` for differentiating DNS resolution errors from other connector errors -- by :user:`mstojcevich`.
```

bdraco (approved):
Thanks @mstojcevich-cisco @mstojcevich

PR #8457, opened 2024-06-12
Use async fixture in testing docs

Dreamsorcerer:
Instead of low-level asyncio APIs, which regular users shouldn't be using.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -111,11 +111,11 @@ app test client::
             body='value: {}'.format(request.app[value]).encode('utf-8'))
 
     @pytest.fixture
-    def cli(loop, aiohttp_client):
+    async def cli(aiohttp_client):
         app = web.Application()
         app.router.add_get('/', previous)
         app.router.add_post('/', previous)
-        return loop.run_until_complete(aiohttp_client(app))
+        return await aiohttp_client(app)
 
     async def test_set_value(cli):
         resp = await cli.post('/', data={'value': 'foo'})


webknjaz:
I think that injecting the `loop` fixture is still good for isolation. Even in our tests, we sometimes get one test influencing the other when the loop leaks.

P.S. I didn't know you could make async fixtures. Are they new?
P.P.S. I'll be going to the pytest development sprint next week and maybe will learn more new stuff as well..

Dreamsorcerer:
> I think that injecting the `loop` fixture is still good for isolation. Even in our tests, we sometimes get one test influencing the other when the loop leaks.

aiohttp_client depends on the loop fixture, so that's not something for the user to be concerned with.

> P.S. I didn't know you could make async fixtures. Are they new?

We've been using them in various places for atleast a couple of years. So, probably supported since 2-5 years ago.

webknjaz:
> > I think that injecting the `loop` fixture is still good for isolation. Even in our tests, we sometimes get one test influencing the other when the loop leaks.
> 
> aiohttp_client depends on the loop fixture, so that's not something for the user to be concerned with.

I wonder why our tests get hit by this, then..

Anyway, I don't have anything against the patch in general. Just wanted to clarify this bit.

webknjaz:
> > I think that injecting the `loop` fixture is still good for isolation. Even in our tests, we sometimes get one test influencing the other when the loop leaks.
> 
> aiohttp_client depends on the loop fixture, so that's not something for the user to be concerned with.
> 
> > P.S. I didn't know you could make async fixtures. Are they new?
> 
> We've been using them in various places for atleast a couple of years. So, probably supported since 2-5 years ago.

Could you check what's the oldest pytest where this works? Perhaps it'd be nice to document if this doesn't work on something ancient.

Dreamsorcerer:
> Could you check what's the oldest pytest where this works? Perhaps it'd be nice to document if this doesn't work on something ancient.

I think it's actually supported by pytest-asyncio (same as tests). Maybe our original code before using pytest-asyncio didn't support fixtures. It's not clear to me when exactly it became available (or if it even wasn't), but it seems unlikely people will encounter a problem.

webknjaz (approved):

PR #8466, opened 2024-06-19
add keyword argument encode_url for ClientSession._request

RomySaputraSihananda:
Hello Dev Team,

I have implemented a new feature enhancement in the ClientSession._fetch method. This enhancement introduces an optional parameter named encode_url, which allows users to control whether the URL should be encoded before being used in the HTTP request. Below are the technical details:

1. Changes: Added the optional parameter encode_url to the ClientSession._fetch method.
2. Implementation: I added conditional logic where the URL encoding is performed based on the value of encode_url.
3. Testing: I have included unit tests to ensure that this feature works as expected in various scenarios.
I believe that this feature will enhance the flexibility and usability of aiohttp in cases where direct control over URL encoding is necessary. I am open to addressing any feedback or changes required before merging.

Thank you for your time and consideration.

Best regards,

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -346,7 +346,7 @@ class ClientSession:
         return _RequestContextManager(self._request(method, url, **kwargs))
 
     def _build_url(self, str_or_url: StrOrURL, **kwargs: Any) -> URL:
-        url = URL(str_or_url, encoded=kwargs.get('encode_url'))
+        url = URL(str_or_url, encoded=kwargs.get("encode_url"))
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
I think this is just duplicating the URL object, so not sure it makes much difference to the user:

`.get(url, encode_url=False)` vs `.get(URL(url, encoded=True))`


We get enough bug reports from users that I'm thinking we should change the default in future though. @webknjaz What do you think about disabling the encoding option by default in v4?

Issue #8233, opened 2024-03-18
ASGI support


FeldrinH:
### Is your feature request related to a problem?

As far as I know most async Python web frameworks nowadays use ASGI with servers like Uvicorn. However, as far as I can tell aiohttp server does not support ASGI.

### Describe the solution you'd like

It would be nice if aiohttp could support ASGI, so that it can benefit from features provided by ASGI web servers (see for example https://github.com/aio-libs/aiohttp/issues/5999#issuecomment-930626734).

There may also be some performance benefits to ASGI over WSGI, but I don't know enough about the internals of how aiohttp interfaces with WSGI to make specific claims about this.

### Describe alternatives you've considered

Aiohttp isn't fundamentally broken without this feature so doing nothing is always an alternative.

### Related component

Server

### Additional context

There was pervious discussion of this in 2018 (https://github.com/aio-libs/aiohttp/issues/2902). The final decision there seemed to be to wait and see. Since then ASGI has become much more stable and much more widely used, so I think it is time to revisit this decision.

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
> There may also be some performance benefits to ASGI over WSGI, but I don't know enough about the internals of how aiohttp interfaces with WSGI to make specific claims about this.

We don't support either... (a quick search suggests it wouldn't even be possible to support WSGI as it's sync only). The discussion you linked to Andrew even says that performance may be lower with ASGI: https://github.com/aio-libs/aiohttp/issues/2902#issuecomment-377925008 https://github.com/aio-libs/aiohttp/issues/2902#issuecomment-419906320

I think, as per those previous discussions, I don't see any major advantages to implementing this, and so I doubt anyone here will want to invest time into it. However, if someone wants to give it a go, we'll consider merging it. If the integration is a simple as the attempt in 2017, then it looks reasonable: https://github.com/aio-libs/aiohttp/pull/2035/files

FeldrinH:
> We don't support either... (a quick search suggests it wouldn't even be possible to support WSGI as it's sync only).

I might have misunderstood something then. I noticed that aiohttp supports Gunicorn (https://docs.aiohttp.org/en/stable/deployment.html#nginx-gunicorn) and assumed that since Gunicorn is a WSGI server aiohttp somehow uses (or abuses) WSGI to communicate with Gunicorn.

Dreamsorcerer:
[...]

PR #8473, opened 2024-06-29
add: asgi library to third party list

thearchitector:
## What do these changes do?

Adding [aiohttp-asgi-connector](https://github.com/thearchitector/aiohttp-asgi-connector) to the list of third party libraries. It implements a custom connector to allow interacting directly with ASGI applications instead of going through a web server.

N/A

N/A

## Related issue number

N/A

## Checklist

N/A

### DIFF
--- a/docs/third_party.rst
+++ b/docs/third_party.rst
@@ -299,3 +299,6 @@ ask to raise the status.
 
 - `nacl_middleware <https://github.com/CosmicDNA/nacl_middleware>`_
   An aiohttp middleware library for asymmetric encryption of data transmitted via http and/or websocket connections.
+
+- `aiohttp-asgi-connector <https://github.com/thearchitector/aiohttp-asgi-connector>`_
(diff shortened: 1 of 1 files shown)


docs/third_party.rst
```diff
@@ -299,3 +299,6 @@ ask to raise the status.
 
 - `nacl_middleware <https://github.com/CosmicDNA/nacl_middleware>`_
   An aiohttp middleware library for asymmetric encryption of data transmitted via http and/or websocket connections.
+
+- `aiohttp-asgi-connector <https://github.com/thearchitector/aiohttp-asgi-connector>`_
+  An aiohttp connector for directly interfacing with ASGI applications.
```
Dreamsorcerer: I think the wording could make it clearer that this only affects the client.
i.e. It wouldn't resolve #8233, which is asking for full integration with the server-side (replacing aiohttp's internal server).

thearchitector: i clarified that it's for use with `ClientSession`

Issue #8182, opened 2024-02-23
404 returned if only compressed file exists


steverep:
### Is your feature request related to a problem?

Currently, the `StaticResource` handler immediately returns a 404 status if the requested file is not on disk: [`elif filepath.is_file():`](https://github.com/aio-libs/aiohttp/blob/e5e521e2aa1923223abd680bb431dc21c2c9c863/aiohttp/web_urldispatcher.py#L677-L681).  This is done before the response class checks for compressed versions of the file, making it impossible to store only compressed versions of static resources in order to conserve disk space.

Furthermore, this is an extra unnecessary call for a file system operation that should be eliminated to gain performance.  The file response class already gets a `stat` of the file in the executor to get things like the content length.

### Describe the solution you'd like

- Return the compressed file when permitted by the `Accept-Encoding` header, regardless of uncompressed existence.
- Return a 406 if the compression is not supported by the client, or permit dynamic decompression.
- Eliminate the extra `is_file()` check.

### Describe alternatives you've considered

Only solution is to make sure uncompressed files always exist.

### Related component

Server

### Additional context

Technically, I think serving a 404 when a representation with a supported content coding exists is a violation of RFC 9110.

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

steverep:
@Dreamsorcerer and @bdraco I'm looking at how to fix this but could use some guidance.  I basically see two possible routes:

1. Move all the `os.stat()` calls now in `FileResponse.prepare()` to `StaticResource._handler()` (which is currently making `is_dir` and `is_file` calls to raise any 404).  This would basically keep all exception handling the same but would need to alter `FileResponse.__init__` to accept the stat result.
2. Remove the call to `is_file` from `StaticResource._handler()`, and instead raise any 404 exception from `FileResponse.prepare()`, which AFAICT would then need to be handled in the `RequestHandler` class where `prepare` is called.

Both approaches would make the assumption that the `request` passed to `prepare` is the same as that passed to `_handler` (to use the `Accept-Encoding` header).  Is that at least correct?

After hours of reading code, my gut says the 1st is the right way, but everything going on in the middle of those methods is still very much a black box to me.  Thoughts?  Are there downstream, user, performance, etc. effects here I'm missing?

bdraco:
[...]

Issue #8104, opened 2024-01-29
Accept-Encoding header parsing and interpretation


steverep:
### Is your feature request related to a problem?

Forgive me - parts of this might be considered bug fix and parts feature request, but they are highly related and intertwined in the same code.  These are things I noticed while working on #8063.

- Dynamic compression via `enable_compression()` will compare `Accept-Encoding` case-insensitively, but retrieving static compressed files does not it should be case-insensitive per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-accept-encoding)).
- When the `force` parameter is used in `enable_compression()`, it does not even consider `Accept-Encoding`.  I fail to see any real use case for this and feel it should be treated like a preference instead (i.e. override the default preference to use `deflate` to something else).  If Brotli were implemented, it may very well be more performant than `deflate`, so preferring it might be desirable, but then the server no longer supports legacy browsers without Brotli support.
- AFAICT, there is no proper parsing of the `Accept-Encoding` field per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-accept-encoding).  There is simply an `in` operator test of the string.  For example:
  - ❌ If the header is not set, it indicates any encoding is okay, but the server currently will only send uncompressed.  Technically that's fine, but if a static compressed version exists but the uncompressed version does not (e.g. to save disk space), then the compressed version should be sent.
  - ❌ "*" in the field selects any encoding **not** explicitly listed.
  - ❌ Each coding is permitted to have a quality factor (e.g. "br;q=1, deflate;q=0, *;q=0).  Higher quality factors should be preferred over lower ones if supported by the server, and ";q=0" says not to send that encoding.
  - ✔️ If the header is empty ("") then no compression should be sent.

### Describe the solution you'd like

- [x] `Accept-Encoding` should be treated case-insensitively when returning static compressed files.
- [ ] The `force` parameter to `enable_compression()` should really act like a preference, not an override (i.e. it should not be forced if the encoding is not supported per the `Accept-Encoding` request header).
- [ ] Parse and interpret the header per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-accept-encoding) with quality factors, "*", etc.  Send a 415 response if merited.

### Describe alternatives you've considered

None for the first 2, but admittedly the parsing issue is more or less a moot point since most clients always send the header with an explicit list and no quality factors (i.e. "gzip, deflate, br").

### Related component

Server

### Additional context

n/a

[...]

PR #8485, opened 2024-07-08
Fix response headers for compressed file requests

### LINKED ISSUE #4462: aiohttp decompresses 'Content-Encoding: gzip' responses
## Long story (not so) short
aiohttp automatically decompresses responses with a `Content-Encoding` header. I believe this is incorrect – while decompression may of course be useful, it should not be on by default.

The `Content-Encoding` header is similar to the `Transfer-Encoding` one, and the two are often confused. The difference is that TE is about the data transfer, i.e. the encoding for a particular connection and client, whereas CE is about the data itself. CE exists to allow specifying this encoding without losing the MIME type of the actual data.

As an example, an ideal response for a `.tar.gz` file download would have a `Content-Type: application/x-tar` header and a `Content-Encoding: gzip` one (rather than e.g. `Content-Type: application/x-gzip` without `Content-Encoding`). When downloading such a file, one would still expect to get the `.tar.gz` file, not an uncompressed `.tar`. This is why I think that aiohttp should not automatically decompress on a `Content-Encoding` header.

## Steps to reproduce
Make a request for a file that has been gzipped, and read from the ClientResponse object (e.g. to write it to a file).

## Expected behaviour
Get the gzipped file.

## Actual behaviour
Returns the decompressed file.

## Your environment
All versions are affected (oldest checked: 2.3.10, but presumably much older than that).

## Additional notes
curl also used to have [this bug](https://github.com/curl/curl/issues/661) in their HTTP/2 implementation. I haven't searched in detail whether it existed in the HTTP/1.1 one as well, but current versions do not decompress such responses.

steverep:
## What do these changes do?

Currently the server responds to requests for compressed files with a `Content-Encoding` header based on the extension.  This is not correct as the encoding is supposed to be a transformation applied to the content type, and the type should be a representation of the request (i.e. if the request is for a compressed file, then the type should represent a compressed file).  The reasoning is further explained in https://github.com/aio-libs/aiohttp/issues/4462#issuecomment-2108230866.

Note this also addresses problems when non-standard or unsupported encodings are involved.  The server will currently respond to "/hello.txt.bz2" with encoding "bzip2" to an `aiohttp` client, despite the fact that it's not supported or sent in the `Accept-Encoding` header.  That header check is completely bypassed ATM.

Requests for static compressed files will no longer be decompressed by most clients (when the encoding is supported).  However this is not a server behavior change so much as an expected reaction to the server correction.

No

## Related issue number

Fixes #4462 and helps advance #8104 (by eliminating some ignorance of `Accept-Encoding`)

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -50,16 +50,22 @@ ENCODING_EXTENSIONS = MappingProxyType(
 
 FALLBACK_CONTENT_TYPE = "application/octet-stream"
 
+# Provide additional MIME type/extension pairs to be recognized.
+# IANA-registered types can be skipped (e.g. application/gzip).
 # https://en.wikipedia.org/wiki/List_of_archive_formats#Compression_only
-ENCODING_CONTENT_TYPES = MappingProxyType(
+ADDITIONAL_CONTENT_TYPES = MappingProxyType(
     {
-        "gzip": "application/gzip",
-        "br": "application/x-brotli",
-        "bzip2": "application/x-bzip2",
-        "compress": "application/x-compress",
-        "xz": "application/x-xz",
+        "application/x-brotli": ".br",
+        "application/x-bzip2": ".bz2",
+        "application/x-compress": ".Z",
+        "application/x-xz": ".xz",
     }
 )
+
+# Add custom pairs and clear the encodings map so guess_type ignores them.
+mimetypes.encodings_map.clear()
+for content_type, extension in ADDITIONAL_CONTENT_TYPES.items():
+    mimetypes.add_type(content_type, extension)
 
 
 class FileResponse(StreamResponse):
@@ -205,17 +211,6 @@ class FileResponse(StreamResponse):
         ):
             return await self._not_modified(request, etag_value, last_modified)
 
-        # If the Content-Type header is not already set, guess it based on the
-        # extension of the request path. If the request is for a compressed
-        # file, map the encoding back to the correct content type.
-        ct = None
-        if hdrs.CONTENT_TYPE not in self.headers:
-            ct, encoding = mimetypes.guess_type(str(self._path))
-            if encoding:
-                ct = ENCODING_CONTENT_TYPES.get(encoding, FALLBACK_CONTENT_TYPE)
-            elif not ct:
-                ct = FALLBACK_CONTENT_TYPE
-
         status = self._status
         file_size = st.st_size
         count = file_size
@@ -290,8 +285,14 @@ class FileResponse(StreamResponse):
                 # return a HTTP 206 for a Range request.
                 self.set_status(status)
 
-        if ct:
-            self.content_type = ct
+        # If the Content-Type header is not already set, guess it based on the
+        # extension of the request path. The encoding returned by guess_type
+        #  can be ignored since the map was cleared above.
+        if hdrs.CONTENT_TYPE not in self.headers:
+            self.content_type = (
+                mimetypes.guess_type(self._path)[0] or FALLBACK_CONTENT_TYPE
+            )
+
         if file_encoding:
             self.headers[hdrs.CONTENT_ENCODING] = file_encoding
             self.headers[hdrs.VARY] = hdrs.ACCEPT_ENCODING
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -40,14 +40,14 @@ _T_OnChunkSent = Optional[Callable[[bytes], Awaitable[None]]]
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
-content_types = MimeTypes()
+CONTENT_TYPES: Final[MimeTypes] = MimeTypes()
 
 if sys.version_info < (3, 9):
-    content_types.encodings_map[".br"] = "br"
+    CONTENT_TYPES.encodings_map[".br"] = "br"
 
 # File extension to IANA encodings map that will be checked in the order defined.
 ENCODING_EXTENSIONS = MappingProxyType(
-    {ext: content_types.encodings_map[ext] for ext in (".br", ".gz")}
+    {ext: CONTENT_TYPES.encodings_map[ext] for ext in (".br", ".gz")}
 )
 
 FALLBACK_CONTENT_TYPE = "application/octet-stream"
@@ -65,9 +65,9 @@ ADDITIONAL_CONTENT_TYPES = MappingProxyType(
 )
 
 # Add custom pairs and clear the encodings map so guess_type ignores them.
-content_types.encodings_map.clear()
+CONTENT_TYPES.encodings_map.clear()
 for content_type, extension in ADDITIONAL_CONTENT_TYPES.items():
-    content_types.add_type(content_type, extension)  # type: ignore[attr-defined]
+    CONTENT_TYPES.add_type(content_type, extension)  # type: ignore[attr-defined]
 
 
 class FileResponse(StreamResponse):
@@ -292,7 +292,7 @@ class FileResponse(StreamResponse):
         #  can be ignored since the map was cleared above.
         if hdrs.CONTENT_TYPE not in self.headers:
             self.content_type = (
-                content_types.guess_type(self._path)[0] or FALLBACK_CONTENT_TYPE
+                CONTENT_TYPES.guess_type(self._path)[0] or FALLBACK_CONTENT_TYPE
             )
 
         if file_encoding:
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -1,9 +1,9 @@
 import asyncio
-import mimetypes
 import os
 import pathlib
 import sys
 from contextlib import suppress
+from mimetypes import MimeTypes
 from types import MappingProxyType
 from typing import (
     IO,
@@ -40,12 +40,14 @@ _T_OnChunkSent = Optional[Callable[[bytes], Awaitable[None]]]
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+content_types = MimeTypes()
+
 if sys.version_info < (3, 9):
-    mimetypes.encodings_map[".br"] = "br"
+    content_types.encodings_map[".br"] = "br"
 
 # File extension to IANA encodings map that will be checked in the order defined.
 ENCODING_EXTENSIONS = MappingProxyType(
-    {ext: mimetypes.encodings_map[ext] for ext in (".br", ".gz")}
+    {ext: content_types.encodings_map[ext] for ext in (".br", ".gz")}
 )
 
 FALLBACK_CONTENT_TYPE = "application/octet-stream"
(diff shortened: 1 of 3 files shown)


tests/test_web_sendfile_functional.py
```diff
@@ -334,9 +343,9 @@ async def handler(request):
 
     resp = await client.get("/")
     assert resp.status == 200
-    assert resp.headers.get("Content-Encoding") == expect_encoding
-    assert resp.headers["Content-Type"] == "text/plain"
-    assert await resp.read() == HELLO_AIOHTTP
+    assert resp.headers.get("Content-Encoding") is None
```
Dreamsorcerer, line 346: I think this is more explicit as:
```suggestion
    assert "Content-Encoding" not in resp.headers
```

Dreamsorcerer (approved):
I believe this looks correct. @bdraco Think you could double check it?

tests/test_web_sendfile_functional.py
steverep: Actually, I considered that, but with that change, the log will only tell you it's true if the assertion fails. This way it tells you what the header is actually set to. It's done this way in another test above.

bdraco:
Will try to take a look soon. Hurricane Beryl is causing tons of problems in Houston right now so may not be today

steverep:
I added a couple extra sentences to the change notes just in case users need to adjust clients to the probable behavior of no longer decompressing the files.

Also, does this need to backport to 3.9 as well or are patches done for 3.9?

bdraco:
I don’t think we are doing any more 3.9a

bdraco:
The code looks good but I still haven’t been able to test it as all my equipment in Houston is on emergency power.  Grid power is expected to come back tomorrow

steverep:
Thanks and no rush.  The rain is currently headed my way for the next few days, albeit with much less intensity now, so hopefully I won't be saying the same.

I'm trying to work on #8182 in the meantime, so any advice there would be appreciated.

bdraco:
Going to give this a test now

bdraco:
backported to 3.10 for testing as https://github.com/bdraco/aiohttp/commit/db1eaec61aceb868d7d988e2e1b5320281431199

bdraco:
I tested this with a binary file by gziping it and than downloading it. All went ok

When I tested with a `.svg` (compressed to `.svg.gz`), it failed to download or render in both Chrome and Firefox
Screenshot 2024-07-10 at 8 31 22 PM

Screenshot 2024-07-10 at 8 34 21 PM

bdraco:
Downgrading back to 3.9.x and running the same test now

bdraco:
same problem happens on 3.9

Screenshot 2024-07-10 at 8 35 58 PM

bdraco:
Tried with a bzip2ed  jpeg on 3.9

The content-type that comes back is `image/jpeg`

Trying with this patch now

bdraco:
Tried with a bzip2ed  jpeg on 3.10 with this patch

The content-type that comes back is `image/jpeg` and the browser renders a broken image

The behavior is the same as 3.9 baseline

steverep:
If you request "file.svg.gz", then I would not expect it to render since the content type should be "application/gzip". That case is really no different than downloading "file.tar.gz". Images would only render if you request it without the compression extension.

steverep:
> Tried with a bzip2ed jpeg on 3.10 with this patch
> 
> The content-type that comes back is `image/jpeg` and the browser renders a broken image

I'm confused by this one.  Was the request with or without the `.bz2` extension?
- With: content type should be "application/x-bzip2"
- Without: `aiohttp` wouldn't serve it since it doesn't look for `.bz2`

Either way, browser wouldn't render it since they don't support `bzip2` encoding.

bdraco:
> I'm confused by this one. Was the request with or without the `.bz2` extension?

Request was with the bz2 extension

steverep:
> That was what I had expected as well but it was `image/jpeg`

Can you share that test for me to repro?

bdraco:
- Checkout https://github.com/bdraco/aiohttp/commit/db1eaec61aceb868d7d988e2e1b5320281431199
- Install it in a homeassistant container `docker exec -it homeassistant /bin/bash`
- Copy a jpeg file into `/config/www/something.jpeg`
- `cd /config/www ; bzip2 something.jpeg`
- navigate to http://xxxx:2083/local/something.jpeg.bz2

steverep:
Tests with HA won't work for compressed files because [HA incorrectly overrides the content type](https://github.com/home-assistant/core/blob/ab7c4244d2745b1915bc5ce5b6623672c21e91f4/homeassistant/components/http/static.py#L66-L68):

```py
                content_type = (mimetypes.guess_type(rel_url))[
                    0
                ] or "application/octet-stream"
```

bdraco:
I'll investigate the issue in HA this weekend (or first thing Monday) and see if I can come up with a solution so I can test this.

steverep:
I can make a quick fix PR for HA by just removing the code to cache and override the `Content-Type`.  It's just repeating the same code in `FileResponse.prepare()` that this PR is trying to fix.

But I wasn't going to bother with it because I wanted to fix #8182 too, which may end up rendering much more of that code unnecessary.

steverep:
HA change to allow this to take effect: https://github.com/home-assistant/core/pull/121831

aiohttp/web_fileresponse.py
```diff
@@ -48,6 +48,19 @@
...
+ENCODING_CONTENT_TYPES = MappingProxyType(
+    {
+        "gzip": "application/gzip",
+        "br": "application/x-brotli",
+        "bzip2": "application/x-bzip2",
+        "compress": "application/x-compress",
+        "xz": "application/x-xz",
+    }
+)
+
```
bdraco, line 66: Maybe it makes sense to let `mimetypes` handle this complexity

We could instead create an aiohttp specific [`mimetypes.MimeTypes`](https://docs.python.org/3/library/mimetypes.html#mimetypes.MimeTypes) object, clear the [`encodings_map`](https://github.com/python/cpython/blob/0ec761a96a80f4b84b98aad5e784e4ac8f385c3f/Lib/mimetypes.py#L417C5-L417C18) and add `ENCODING_CONTENT_TYPES` with `add_type` (need to verify `compress` is correct, as I'm not sure) ex: https://github.com/python/cpython/blob/0ec761a96a80f4b84b98aad5e784e4ac8f385c3f/Lib/mimetypes.py#L80

Than you have a `mimetypes_obj.guess_type(str(self._path))` which always returns the `ct` expected below and we don't need to do the `ENCODING_CONTENT_TYPES` lookup because `mimetypes` would handle the logic.

aiohttp/web_fileresponse.py
```diff
@@ -48,6 +48,19 @@
...
 
+FALLBACK_CONTENT_TYPE = "application/octet-stream"
+
+# https://en.wikipedia.org/wiki/List_of_archive_formats#Compression_only
+ENCODING_CONTENT_TYPES = MappingProxyType(
+    {
+        "gzip": "application/gzip",
+        "br": "application/x-brotli",
+        "bzip2": "application/x-bzip2",
+        "compress": "application/x-compress",
```
bdraco: `mime_types` uses `Z` instead of `compress` https://github.com/python/cpython/blob/0ec761a96a80f4b84b98aad5e784e4ac8f385c3f/Lib/mimetypes.py#L419

steverep: This maps the encoding to the type, not the extension.

aiohttp/web_fileresponse.py
steverep: Good idea, but might as well just call the methods directly.  No need to complicate with a subclass. 👍🏻

steverep:
In addition to @bdraco's suggestion that should improve performance, I also moved the lookup down (after any possible 416 responses) and avoided the string conversion (`mimetypes` supports `PathLike` after 3.8).

Another potential performance tweak would be to put the type lookup in parallel with opening the file.

aiohttp/web_fileresponse.py
```diff
@@ -48,6 +48,25 @@
...
+        "application/gzip": ".gz",
+        "application/x-brotli": ".br",
+        "application/x-bzip2": ".bz2",
+        "application/x-compress": ".Z",
+        "application/x-xz": ".xz",
+    }
+)
+
+# Add custom pairs and clear the encodings map so guess_type ignores them.
+mimetypes.encodings_map.clear()
```
bdraco: We need a https://docs.python.org/3/library/mimetypes.html#mimetypes.MimeTypes here to operate on so we don’t affect any other packages using mimetypes

steverep: We're already modifying it by adding Brotli for < 3.9.  Does that need to be part of the custom object too?

Dreamsorcerer: Well, 3.8 will be dropped shortly, so I wouldn't waste any time on it.

steverep:
Mypy is complaining `add_types()` isn't on the class, but it certainly is. 😕

aiohttp/web_fileresponse.py
```diff
@@ -40,14 +40,35 @@
 
 NOSENDFILE: Final[bool] = bool(os.environ.get("AIOHTTP_NOSENDFILE"))
 
+content_types = MimeTypes()
```
bdraco: maybe `CONTENT_TYPES` to indicate is constant at runtime.

bdraco:
> Mypy is complaining `add_types()` isn't on the class, but it certainly is. 😕

Its definitely defined and `mimetypes.add_type` is implemented as a wrapper around `MimeTypes.add_type` but it is missing from the pyi file.  I think its unlikely to change so an ignore is fine.

bdraco:
LGTM once https://github.com/aio-libs/aiohttp/pull/8485/files#r1676667798 is worked out 👍 

Thanks @steverep

bdraco (approved):

steverep:
FWIW... https://github.com/python/cpython/issues/121658

Dreamsorcerer:
> > Mypy is complaining `add_types()` isn't on the class, but it certainly is. 😕
> 
> Its definitely defined and `mimetypes.add_type` is implemented as a wrapper around `MimeTypes.add_type` but it is missing from the pyi file. I think its unlikely to change so an ignore is fine.

I don't know if either you had anything to do with it, but it was just added to typeshed:
https://github.com/python/typeshed/pull/12329

Not sure if the change will get synced into the 1.11 release or not.

Dreamsorcerer:
Ah, I see, it was result of the docs bug report.

steverep:
@bdraco FYI, the use of `MimeTypes()` here now guarantees that any `mime.types` files from the OS are read on module load.  Previously, that would not have happened until the first call to `mimetypes.guess_type()`, potentially significantly blocking that first response.

bdraco:
> @bdraco FYI, the use of `MimeTypes()` here now guarantees that any `mime.types` files from the OS are read on module load. Previously, that would not have happened until the first call to `mimetypes.guess_type()`, potentially significantly blocking that first response.

This is much nicer for async. In HA we preload mine types at bootstrap in the executor for this reason

steverep:
I mentioned above potentially another small improvement by putting that large table lookup concurrent with opening the file, e.g.:

```py
guess_type_fut = loop.run_in_executor(None, CONTENT_TYPES.guess_type, self._path)
# later on...
open_fut = loop.run_in_executor(None, file_path.open, "rb")
fobj,ct = await asyncio.gather([open_fut, guess_type_fut])
```

Worth it?  Is there a benchmark in place to test it?

PR #8495, opened 2024-07-13
Shutdown logic: Only wait on handlers

### LINKED ISSUE #8387: aiohttp ^C hangs when psycopg connection pool created
### Describe the bug

If I create a `psycopg_pool.AsyncConnectionPool` object, `aiohttp`'s Ctrl+C handler will hang while trying to handle the ^C.

### To Reproduce

The following minimal example reproduces the bug:

```python
import logging

from aiohttp import web
import psycopg_pool


async def hello(request):
    return web.Response(text="Hello, world")


def main():
    logging.basicConfig(level=logging.DEBUG)

    app = web.Application()
    app.add_routes([
        web.get('/', hello),
    ])
    app.cleanup_ctx.append(db_pool)

    web.run_app(app)


async def db_pool(app):
    kwargs = {
        'hostname': 'localhost',
        'port': 5432,
    }
    pool = psycopg_pool.AsyncConnectionPool(
        kwargs=kwargs,
        open=False,
        min_size=0,
        max_size=4,
    )
    print('Pool opening.')
    await pool.open()
    print('Pool open.')
    #app['db_driver'] = db_driver
    try:
        yield
    finally:
        print('Cleanup ctx done.')
    await pool.close()


if __name__ == '__main__':
    main()
```

Run it, hit ^C:
```
» venv2/bin/python3 aiohttp_bug.py
DEBUG:asyncio:Using selector: EpollSelector
Pool opening.
Pool open.
(Press CTRL+C to quit)
^C <hangs here for ~60s>
```

The hang is only after we create the pool. If we don't create the pool, no hang.

The reason we hang is this bit: https://github.com/aio-libs/aiohttp/blob/04b1212fe231a7e827959aa9a648d5c847e57c0a/aiohttp/web.py#L313-L318

Here, `aiohttp` attempts to wait on tasks to finish?

There's slightly more context here: https://github.com/aio-libs/aiohttp/blob/04b1212fe231a7e827959aa9a648d5c847e57c0a/aiohttp/web.py#L338-L343

> It's very likely that all tasks which run forever will have been created by the time we have completed the application startup

This just seems incorrect? In particular, `psycopg_pool` has a background task that is taking care of various async things that it appears to queue up for that task. That task has an event, which I presume is triggered when the queue has something new.

Its code looks like this:
```python
    async def run(self) -> None:
        """Execute the events scheduled."""
        q = self._queue
        while True:
            async with self._lock:
                now = monotonic()
                task = q[0] if q else None
                if task:
                    if task.time <= now:
                        heappop(q)
                    else:
                        delay = task.time - now
                        task = None
                else:
                    delay = self.EMPTY_QUEUE_TIMEOUT
                self._event.clear()

            if task:
                if not task.action:
                    break
                try:
                    await task.action()
                except Exception as e:
                    logger.warning(
                        "scheduled task run %s failed: %s: %s",
                        task.action,
                        e.__class__.__name__,
                        e,
                    )
            else:
                # Block for the expected timeout or until a new task scheduled
                await self._event.wait_timeout(delay)
```
That last `await self._event.wait_timeout(delay)` is what I am talking about here. `wait_timeout`, when called, calls `asyncio.wait_for`, which creates a task.

This is done every time we hit the lock, and it happens after the app has started, too, so this event wait's task ends up in the `tasks` set that `aiohttp` waits on.

Note that aiohttp's wait on all other running tasks to finish occurs _before_ we run the application cleanups, so _the pool is still alive_ — i.e., we're not first giving the cleanup context an opportunity to close the pool.

### Expected behavior

^C wraps up without hanging (for 60s).

### Logs/tracebacks

```python-traceback
(see above)
```

### Python Version

```console
$ python --version
Python 3.11.5
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.9.5
Author: 
Author-email: 
License: Apache 2
Location: /home/royiv/code/groceries/venv2/lib/python3.11/site-packages
Required-by:
```

### multidict Version

```console
Name: multidict
Version: 6.0.5
Author: Andrew Svetlov
License: Apache 2
Location: /home/royiv/code/groceries/venv2/lib/python3.11/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.9.4
Author: Andrew Svetlov
License: Apache-2.0
Location: /home/royiv/code/groceries/venv2/lib/python3.11/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

Arch Linux

### Related component

Server

### Additional context

```
Name: psycopg-pool
Version: 3.2.1
Summary: Connection Pool for Psycopg
Home-page: https://psycopg.org/psycopg3/
Author: Daniele Varrazzo
Author-email: daniele.varrazzo@gmail.com
License: GNU Lesser General Public License v3 (LGPLv3)
Location: /home/royiv/code/groceries/venv2/lib/python3.11/site-packages
Requires: typing-extensions
Required-by: 
```

### Code of Conduct

Dreamsorcerer:
In aiohttp 3.9 we introduced new shutdown logic in run_app() which waited for all pending tasks to complete. This has caused issues with libraries that spawn idle tasks at runtime to wait for connections.

This PR removes that logic and instead just waits on all active request handlers. Documentation will be updated to teach users how to manage tasks themselves.

@bdraco Please confirm this still works correctly with HA. The only difference should be that on application shutdown, active request handlers are now given time to complete instead of being cancelled immediately.

- [x] I think an edge case which needs another test is when handler_cancellation is enabled and the handler suppresses the cancellation.

Fixes #8387


aiohttp/web_server.py
```diff
@@ -69,6 +72,7 @@ def pre_shutdown(self) -> None:
     async def shutdown(self, timeout: Optional[float] = None) -> None:
         coros = (conn.shutdown(timeout) for conn in self._connections)
         await asyncio.gather(*coros)
+        print("LENGTH", len(self._connections))
```
Dreamsorcerer: I'd like to assert the number of connections at this point, but can't think of a good way to get this in the tests. Any ideas?

bdraco:
Will test later today.  Still catching up as I have a house full of family that doesn't have power from the hurricane in Houston

bdraco:
```
Jul 14 17:35:56 homeassistant homeassistant[569]: 2024-07-14 12:35:56.216 ERROR (MainThread) [homeassistant] Error doing job: Task exception was never retrieved (None)
Jul 14 17:35:56 homeassistant homeassistant[569]: Traceback (most recent call last):
Jul 14 17:35:56 homeassistant homeassistant[569]:   File "/usr/src/homeassistant/homeassistant/components/http/__init__.py", line 243, in stop_server
Jul 14 17:35:56 homeassistant homeassistant[569]:     await server.stop()
Jul 14 17:35:56 homeassistant homeassistant[569]:   File "/usr/src/homeassistant/homeassistant/components/http/__init__.py", line 624, in stop
Jul 14 17:35:56 homeassistant homeassistant[569]:     await self.runner.cleanup()
Jul 14 17:35:56 homeassistant homeassistant[569]:   File "/usr/local/lib/python3.12/site-packages/aiohttp/web_runner.py", line 307, in cleanup
Jul 14 17:35:56 homeassistant homeassistant[569]:     await self._server.shutdown(self._shutdown_timeout)
Jul 14 17:35:56 homeassistant homeassistant[569]:   File "/usr/local/lib/python3.12/site-packages/aiohttp/web_server.py", line 69, in shutdown
Jul 14 17:35:56 homeassistant homeassistant[569]:     await asyncio.gather(*coros)
Jul 14 17:35:56 homeassistant homeassistant[569]:   File "/usr/local/lib/python3.12/site-packages/aiohttp/web_protocol.py", line 269, in shutdown
Jul 14 17:35:56 homeassistant homeassistant[569]:     await self._current_request.wait_for_disconnection()
Jul 14 17:35:56 homeassistant homeassistant[569]:           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Jul 14 17:35:56 homeassistant homeassistant[569]: AttributeError: 'HomeAssistantRequest' object has no attribute 'wait_for_disconnection'
```

I get the following in HA when I backported this to 3.9 and updated

with the change backported as https://github.com/bdraco/aiohttp/commit/6aa52bdd5ffd45137e7eb06ff9c5b7189ce69ac7

Dreamsorcerer:
> I get the following in HA when I backported this to 3.9 and updated

Seem to be missing a backport of #4200. I assume it subclasses BaseRequest? In which case I expect it will work with that backported.

bdraco:
I'll make another integration branch with that backport to 3.9

I can't test with 3.10 yet until https://github.com/aio-libs/aiohttp/pull/8482 is backported or I make an integration branch for that.  But since HA is on 3.9 better to test with 3.9

bdraco:
https://github.com/aio-libs/aiohttp/pull/4200 has a whole lot of conflicts so I'll wait for the backport to happen instead of trying to patch it into the integration branch as there is a chance I'll screw it up and give you a bad test result.

Dreamsorcerer:
Should be sorted in #8504.

bdraco:
Tested with HA on 3.9+ this PR + https://github.com/aio-libs/aiohttp/pull/8504 cherry picked on top

Restarted a few times. Everything seems ok

aiohttp/web_server.py
bdraco: Maybe patch the `shutdown` function and see how many calls are made to it?

bdraco (approved):

bdraco:
Still needs a test and changelog, manual functional testing passed.

bdraco:
Adding a skip for now so we can get this into 3.10b0 as Sam is going to do another turn on this to add the test / changelog before the stable release

bdraco (approved):

aiohttp/web_server.py
Dreamsorcerer: I don't think that works. The number of shutdown() calls doesn't matter, it's the number of them which didn't result in the connections being removed from the set as part of the done callback.

Dreamsorcerer: I think if I can figure out how to get a reference to the Server object, then I could patch the .clear() call to check how many are still in the set when this is called.

PR #8498, opened 2024-07-14
Avoid creating a future on every websocket receive

bdraco:
## What do these changes do?

We should only create a future on close if we wait for the receive to finish. Instead of making the future every time `receive` is called, we should only create it when we are closing and let the `close` call await it only if `receive` is waiting. Since `close` is called far less frequently than `receive` in most cases, this is far more efficient.

There was also a nice decrease in `set_result` calls and a reduction of ~1000 future creations per minute on an HA install with 14 WebSocket connections open. Users with many WebSocket connections will see a much more substantial drop. WebSockets should scale better with this change, especially if they are highly active.

The future was added in https://github.com/aio-libs/aiohttp/commit/ae38c4ac7647d4b904a949088f44ed539553a26e , and the `test_concurrent_close` tests still pass with only creating the future on close

no

no

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -197,10 +197,10 @@ class ClientWebSocketResponse:
         # we need to break `receive()` cycle first,
         # `close()` may be called from different task
         if self._waiting and not self._closing:
-            assert self._loop is not None
             if not self._close_wait:
+                assert self._loop is not None
                 self._close_wait = self._loop.create_future()
-                self._closing = True
+            self._closing = True
             self._reader.feed_data(WS_CLOSING_MESSAGE)
             await self._close_wait
 
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -198,8 +198,9 @@ class ClientWebSocketResponse:
         # `close()` may be called from different task
         if self._waiting and not self._closing:
             assert self._loop is not None
-            self._close_wait = self._loop.create_future()
-            self._closing = True
+            if not self._close_wait:
+                self._close_wait = self._loop.create_future()
+                self._closing = True
             self._reader.feed_data(WS_CLOSING_MESSAGE)
             await self._close_wait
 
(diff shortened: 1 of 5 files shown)


bdraco:
I think this needs a test where concurrent tasks close the connection at the same time before receive is done in another task

aiohttp/client_ws.py
```diff
@@ -252,17 +255,17 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
                     ):
                         msg = await self._reader.read()
                     self._reset_heartbeat()
                 finally:
-                    waiter = self._waiting
-                    self._waiting = None
-                    set_result(waiter, True)
+                    self._waiting = False
+                    if close_wait := self._close_wait:
+                        set_result(close_wait, None)
```
Dreamsorcerer: I feel like it's more readable without the assignment.
```suggestion
                    if self._close_wait:
                        set_result(self._close_wait, None)
```

aiohttp/web_ws.py
```diff
@@ -479,15 +484,15 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
                     async with async_timeout.timeout(timeout or self._receive_timeout):
                         msg = await self._reader.read()
                     self._reset_heartbeat()
                 finally:
-                    waiter = self._waiting
-                    set_result(waiter, True)
-                    self._waiting = None
+                    self._waiting = False
+                    if close_wait := self._close_wait:
+                        set_result(close_wait, None)
```
```suggestion
                    if self._close_wait:
                        set_result(self._close_wait, None)
```

Dreamsorcerer (approved):

aiohttp/client_ws.py
bdraco: I usually use this pattern with futures to ensure we hold the future in case it gets unset inside the block if there is an await.  Its a learned pattern to avoid mistakes. It doesn't matter here though since execution is synchronous and there is no await. So let's not use the walrus here.

bdraco:
thanks

PR #8507, opened 2024-07-16
Remove blocking IO for static resources and refactor exception handling

steverep:
## What do these changes do?

Refactors the static resource handler to run all IO methods in an executor.  Also revamps the exception handling to be less confusing and throw client exceptions only for known scenarios.  The current handling clearly was designed prior to python 3.6, where `Path.resolve()` was called strictly - see https://github.com/aio-libs/aiohttp/issues/8182#issuecomment-2227680987 for details.

I doubt these would be noticed, but:
1. `PermissionError` will now always lead to a 403 response (previously there was a mix of 403 and 404 depending on the scenario).
2. "Unknown" exceptions are not caught and logged as before.  Instead they will raise up and be logged as server 500 errors as they should.

No

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
         rel_url = request.match_info["filename"]
+        filename = Path(rel_url)
+        if filename.anchor:
+            # rel_url is an absolute name like
+            # /static/\\machine_name\c$ or /static/D:\path
+            # where the static dir is totally different
+            raise HTTPForbidden()
+
+        unresolved_path = self._directory.joinpath(filename)
+        loop = asyncio.get_event_loop()
```
```suggestion
        loop = asyncio.get_running_loop()
```

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

> Because this function has rather complex behavior (especially when custom event loop policies are in use), using the [get_running_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) function is preferred to [get_event_loop()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) in coroutines and callbacks.

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
+                else:
                     raise HTTPForbidden()
-            else:
-                raise HTTPForbidden()
-        elif filepath.is_file():
-            return FileResponse(filepath, chunk_size=self._chunk_size)
-        else:
-            raise HTTPNotFound
+        except PermissionError as error:
+            raise HTTPForbidden() from error
```
bdraco, line 691: Missing coverage

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
-        # on opening a dir, load its contents if allowed
-        if filepath.is_dir():
-            if self._show_index:
-                try:
+        # if path is a directory, return the contents if permitted. Note the
+        # directory check will raise if a segment is not readable.
+        try:
+            if file_path.is_dir():
+                if self._show_index:
                     return Response(
```
bdraco, line 684: Response is being created in the executor now.  I think thats currently thread-safe based on digging though the code, but it might not be in the future.

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
+        except PermissionError as error:
+            raise HTTPForbidden() from error
+
+        # Not a regular file or does not exist.
+        if not file_path.is_file():
+            raise HTTPNotFound()
 
-    def _directory_as_html(self, filepath: Path) -> str:
-        # returns directory's index as html
+        return FileResponse(file_path, chunk_size=self._chunk_size)
```
bdraco, line 697: FileResponse is now being created in the executor as well. Need to verify this is thread-safe.

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
+        # root is permitted.
         try:
-            filename = Path(rel_url)
-            if filename.anchor:
-                # rel_url is an absolute name like
-                # /static/\\machine_name\c$ or /static/D:\path
-                # where the static dir is totally different
-                raise HTTPForbidden()
-            unresolved_path = self._directory.joinpath(filename)
             if self._follow_symlinks:
```
bdraco, line 668: NOTE TO SELF:

AUDIT: This block was the source of a vuln in the past. A full analysis needs to be performed here.

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
             raise HTTPNotFound() from error
 
-        # on opening a dir, load its contents if allowed
-        if filepath.is_dir():
-            if self._show_index:
-                try:
+        # if path is a directory, return the contents if permitted. Note the
+        # directory check will raise if a segment is not readable.
+        try:
+            if file_path.is_dir():
```
bdraco, line 682: Its unfortuate that this will stat the file again since `is_dir` and `is_file` are implemented like `return S_ISREG(self.stat().st_mode)`

It should be cached but still not free.

bdraco: I guess we could reverse this and do the `is_file` check first as it would be the common case and you only end up with an extra `stat` on the dir or not a normal file case.

CHANGES/8507.misc.rst
```diff
@@ -0,0 +1 @@
+Fixed blocking IO in the event loop for static resources and refactored exception handling -- by :user:`steverep`.
```
bdraco: I'd call this a bugfix

aiohttp/web_urldispatcher.py
steverep: I could move the response construction back to `_handle`, but then the code gets dirtier and harder to read.  I'd have to return either a `Path` or a string from the executor, then conditionally build the response based on the type.

aiohttp/web_urldispatcher.py
steverep: I agree perf could be improved here, but TBH I'd rather keep status quo for this PR.  

I'd rather think about this in a follow up to address [#8182](https://github.com/aio-libs/aiohttp/issues/8182).  Anything I change here would have to be rethought out, where there are weird cases to think about like someone has a directory `foo/` next to a `foo.gz`.

aiohttp/web_urldispatcher.py
bdraco: Confirmed nothing changed here. Only `filepath` -> `file_path`

aiohttp/web_urldispatcher.py
bdraco: I think the big potential risk we have to work about is something calling `asyncio.get_running_loop()` in `__init__`.  That doesn't happen now so this is probably ok.

aiohttp/web_urldispatcher.py
bdraco: Like the above, the big potential risk we have to work about is something calling `asyncio.get_running_loop()` in `__init__`.  I audited this path as well and it looks ok. We can always refactor it again later if we need to.

aiohttp/web_urldispatcher.py
mshandrovskiy: Anyone that cares about perf should not be using this in the first place. nginx or similar solution for serving static resources is virtually guaranteed to be the better choice.

bdraco: When we were fixing GHSA-5h86-8mv2-jq9f we found static usage much more common than expected. While more ideal, `nginix` isn't always something that can be deployed because of where the python application sits in the stack.  Millions+ of installs use the static handler every day because of this and it still makes sense to get what we can out of it.

mshandrovskiy: That's a fair point @bdraco I would argue that vast majority of those installs do have the option of a better solution, but don't know or care to implement it. I vaguely remember documentation steering people towards it.

aiohttp/web_urldispatcher.py
steverep: Take a look at the test I just modified, which of course fails on Windows.  I would prefer this to a mock and just skip Windows on the test.  What's your preference?  I don't know if it's even possible to test something being unauthorized on Windows.

That test covers `Path.iterdir` throwing.  I'd just add one for `is_dir` throwing too.

steverep: Went ahead and pushed my preference.  Let me know.

bdraco: I think its better to patch and mock out the actual file calls in this case since we will never be able to fully simulate the access failures in the test environments

steverep: It's easy to fully simulate in Linux and Mac using `chmod`.  Windows is the outlier where that only changes the read-only flag and you'd actually have to simulate a different user or group. 🤷🏻‍♂️

CHANGES/8507.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed blocking IO in the event loop for static resources and refactored exception handling -- by :user:`steverep`.
```
webknjaz: Could you update this with more context? When you say “Fixed blocking I/O”, this implies that it was broken somehow but avoids specifics. I think that the end-users reading the release notes deserve some background regarding how this affects them on the high level, without having to dig into the source code.
Would you be able to compose something for that audience?

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
+            raise HTTPForbidden()
+
+        unresolved_path = self._directory.joinpath(filename)
+        loop = asyncio.get_running_loop()
+        return await loop.run_in_executor(
+            None, self._resolve_path_to_response, unresolved_path
+        )
+
+    def _resolve_path_to_response(self, unresolved_path: Path) -> StreamResponse:
+        """Takes the unresolved path and queries the file system to form a response."""
```
webknjaz:
```suggestion
        """Take the unresolved path and query the file system to form a response."""
```

aiohttp/web_urldispatcher.py
```diff
@@ -639,59 +639,67 @@ def __iter__(self) -> Iterator[AbstractRoute]:
...
-            # relatively safe
-            raise HTTPNotFound() from error
-        except HTTPForbidden:
-            raise
-        except Exception as error:
-            # perm error or other kind!
-            request.app.logger.exception(error)
+                file_path = unresolved_path.resolve()
+                file_path.relative_to(self._directory)
+        # TODO(PY39): Remove OSError to appease Windows
```
webknjaz: From this comment, it's not very clear whether this needs to be handled when 3.9 is the lowest supported or when it's no longer supported...
Perhaps, this could be better expressed though code?

Something along the lines of
```python
known_file_lookup_errors = (
    ValueError, RuntimeError,
    *(() if IS_PY310_PLUS or not IS_WINDOWS else (OSError,)),
)

    ...
except known_file_lookup_errors as error:
    ...
```

steverep: It's when 3.9 is the lowest supported (which seems to be the syntax used elsewhere in the code?).  Seems silly to me to add a conditional for the Windows 3.8 bug just for a few months?

steverep: I updated the comment for clarity.

CHANGES/8507.bugfix.rst
steverep: I added context and changed "fixed" to "removed".

aiohttp/web_urldispatcher.py
webknjaz: We don't really use FIXME comments much. If somebody adds them, they just remain in the code base to rot. So I prefer actual code. Plus, this way, the CI would be testing both behavior branches, which is more future-proof. I'd rather keep a constant next to https://github.com/aio-libs/aiohttp/blob/3fbb093/aiohttp/helpers.py#L67 so that it'd be easier to track places that need to be cleaned up when the time comes.

CHANGES/8507.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Removed blocking I/O in the event loop for static resources and refactored
+exception handling -- by :user:`steverep`.
+
+File system calls when handling requests for static routs were moved to a
```
webknjaz:
```suggestion
File system calls when handling requests for static routes were moved to a
```

CHANGES/8507.bugfix.rst
webknjaz: Thanks! Marking this as resolved.

aiohttp/web_urldispatcher.py
steverep: Done.  I also misspoke and it affects 3.8 and 3.9 on Windows.

webknjaz: thanks!

webknjaz: It's good that you moved it into a conditional as code — I didn't realize it was one exception or the other. It seemed like one was just conditionally added to the rest.

aiohttp/web_urldispatcher.py
steverep: Kept the `chmod` test and added a monkey patched version.

aiohttp/web_urldispatcher.py
steverep: Windows throws the same error across them all: `WinError 1921`.  The difference is that 3.10+ `pathlib` catches it and correctly re-raises as `RuntimeError`.

CHANGES/8507.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Removed blocking I/O in the event loop for static resources and refactored
+exception handling -- by :user:`steverep`.
+
+File system calls when handling requests for static routes were moved to a
+separate thread in order to potentially improve performance. Exception handling
+was tightened in order to only return 403 Forbidden or 404 Not Found responses
+for expected scenarios; 500 Internal Server Error would be returned for any
+unknown errors.
```
bdraco, line 8:
```suggestion
File system calls when handling requests for static routes were moved to a separate thread to potentially improve performance. Exception handling was tightened to only return 403 Forbidden or 404 Not Found responses for expected scenarios; 500 Internal Server Error would be returned for any unknown errors.
```

CHANGES/8507.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Removed blocking I/O in the event loop for static resources and refactored
+exception handling -- by :user:`steverep`.
+
+File system calls when handling requests for static routes were moved to a
+separate thread in order to potentially improve performance. Exception handling
```
bdraco:
```suggestion
separate thread to potentially improve performance. Exception handling
```

tests/test_web_urldispatcher.py
```diff
@@ -388,31 +388,59 @@ async def async_handler(request: web.Request) -> web.Response:
...
+    my_dir = tmp_path / "my_dir"
+    my_dir.mkdir()
+
+    real_iterdir = pathlib.Path.iterdir
+    real_is_dir = pathlib.Path.is_dir
+
+    def mock_iterdir(self: pathlib.Path) -> Generator[pathlib.Path, None, None]:
+        if my_dir.samefile(self):
+            raise PermissionError()
+        return real_iterdir(self)
```
bdraco, line 428: This looks unreachable in the test

bdraco:
Looks like this is ready for manual testing

bdraco:
Testing this on production now

bdraco:
Manual testing looks good. As soon as the coverage issue is addressed, I think this is good to go.

tests/test_web_urldispatcher.py
steverep: I don't see why this matters.  The point of this line is just to make sure `Path.iterdir` behaves like normal unless it's called on the test directory.  It not being covered just means that `aiohttp` doesn't use it except in the one place we are trying to test.

I could remove the line altogether, but that just means the test will break if `iterdir` is used somewhere else.

bdraco: This project measures test coverage to ensure we have no unreachable lines.  If we aren't hitting the lines, we aren't testing the failure case in the test.  We should remove the unreachable code or move that path to another test to exercise this path.

bdraco (approved):
Thanks @steverep

aiohttp/web_urldispatcher.py
Dreamsorcerer: > We don't really use FIXME comments much. If somebody adds them, they just remain in the code base to rot.

This convention is probably from me. I check and follow all the steps each time we drop a Python release. So, as long as it contains `PY39` or whatever, it will get cleaned up.

bdraco:
Ran fine in production overnight.

PR #8510, opened 2024-07-17
Avoid scheduling the writer if it can finish synchronously

bdraco:
## What do these changes do?

python3.12 introduces eager_start for asyncio.Task, and in most cases writing the http request can be done synchronously because its a small amount of data and will never fill up the buffer causing the task to suspend. This means that the task never has to be scheduled on the event loop, and we avoid a call_soon to schedule the task, and we avoid a call_soon to cleanup the writer since its done right away.

requests will finish much faster in many cases

We have to maintain two branches until we only support python 3.12+ https://github.com/aio-libs/aiohttp/pull/8510/files#diff-720910fd022b9ba8b7ddf0cb446ce5e1afe443f4287ad26a1eb4dc4b5b5f4508R671. This should be minimal

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/8510.misc.rst
+++ b/CHANGES/8510.misc.rst
@@ -1,0 +1,1 @@
+Avoid scheduling the writer if it can finish synchronously -- by :user:`bdraco`.
--- a/CHANGES/8510.misc.rst
+++ b/CHANGES/8510.misc.rst
@@ -1,1 +1,1 @@
-Avoid scheduling the writer if it can finish synchronously -- by :user:`bdraco`.
+When using Python 3.12 or later, the writer is no longer scheduled on the event loop if it can finish synchronously. Avoiding event loop scheduling reduces latency and improves performance. -- by :user:`bdraco`.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -284,7 +284,11 @@ class ClientRequest:
             self.__writer.remove_done_callback(self.__reset_writer)
         self.__writer = writer
         if writer is not None:
-            writer.add_done_callback(self.__reset_writer)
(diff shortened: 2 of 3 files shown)


CHANGES/8510.misc.rst
```diff
@@ -0,0 +1 @@
+Avoid scheduling the writer if it can finish synchronously -- by :user:`bdraco`.
```
webknjaz: Could you change this from the imperative TODO-list instruction style to something that refers to what's changed since the previous release + what noticeable effect the end-users would observe?

aiohttp/client_reqrep.py
```diff
@@ -284,7 +284,11 @@ def _writer(self, writer: Optional["asyncio.Task[None]"]) -> None:
             self.__writer.remove_done_callback(self.__reset_writer)
         self.__writer = writer
         if writer is not None:
```
webknjaz: Could you reduce nesting by flipping this and bailing early + doing something similar to the block below?

CHANGES/8510.misc.rst
bdraco: Adjusted in 74718c1c

aiohttp/client_reqrep.py
bdraco: done in aee68303

Dreamsorcerer (approved):

bdraco:
This one has been running fine on my 3.12 production and 3.11 test setups.

bdraco:
thanks

PR #8512, opened 2024-07-17
[PR #5278/0f0c3759 backport][3.10] Drop Python 3.6 support

bdraco:
(cherry picked from commit 0f0c3759616e5369403cd711d8ff898c28f38548)

### DIFF
--- a/CHANGES/5278.misc
+++ b/CHANGES/5278.misc
@@ -1,0 +1,1 @@
+Drop Python 3.6 support -- by :user:`asvetlov`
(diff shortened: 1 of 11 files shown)


bdraco:
This is a missed backport

CHANGES/5278.misc
```diff
@@ -0,0 +1 @@
+Drop Python 3.6 support -- by :user:`asvetlov`
```
webknjaz: @bdraco there's 2 drop notes in the 3.9.0: https://docs.aiohttp.org/en/latest/changes.html#id24. Perhaps, this shouldn't be the third one?

bdraco: removed in e0b3542b2a4ecd027a0e45e422a1df48d5c93477

PR #8522, opened 2024-07-21
Restore AsyncResolver to be the default resolver

bdraco:
## What do these changes do?

Restore `AsyncResolver` to be the default resolver

Note that we do not install aiodns on windows because it uses protractor event loop by default, and the `DefaultResolver` will end up being `ThreadedResolver` on windows.

Reduction in executor jobs and thread switching. This can result in a significant performance boost.

no

## Related issue number

The `AsyncResolver` was disabled in 9fbb7d708375e0ba01d435c1e8cf41912381c0fc because of problems with IPv6 (see #559).  This is no longer an issue as the `AsyncResolver` uses the same underlying `getaddrinfo` call to match `ThreadedResolver`.  Home Assistant has been using `AsyncResolver` for months now, and nobody noticed the change over.

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -311,9 +311,11 @@ async def test_async_resolver_aiodns_not_present(loop: Any, monkeypatch: Any) ->
         AsyncResolver()
 
 
-def test_default_resolver() -> None:
-    # if getaddrinfo:
-    #     assert DefaultResolver is AsyncResolver
-    # else:
-    #     assert DefaultResolver is ThreadedResolver
+@pytest.mark.skipif(not getaddrinfo, reason="aiodns >=3.2.0 required")
+def test_aio_dns_is_default() -> None:
+    assert DefaultResolver is AsyncResolver
+
+
+@pytest.mark.skipif(getaddrinfo, reason="aiodns <3.2.0 required")
+def test_threaded_resolver_is_default() -> None:
     assert DefaultResolver is ThreadedResolver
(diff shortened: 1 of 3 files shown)


tests/test_resolver.py
```diff
@@ -311,9 +311,11 @@ async def test_async_resolver_aiodns_not_present(loop: Any, monkeypatch: Any) ->
...
-    #     assert DefaultResolver is AsyncResolver
-    # else:
-    #     assert DefaultResolver is ThreadedResolver
+@pytest.mark.skipif(not getaddrinfo, reason="aiodns >=3.2.0 required")
+def test_aio_dns_is_default() -> None:
+    assert DefaultResolver is AsyncResolver
+
+
+@pytest.mark.skipif(getaddrinfo, reason="aiodns <3.2.0 required")
+def test_threaded_resolver_is_default() -> None:
```
Dreamsorcerer, line 320: Surely only one of these will run in the test suite?

Dreamsorcerer (approved):

tests/test_resolver.py
bdraco: That's the hope. I'll drop `test_threaded_resolver_is_default` assuming codecov confirms

tests/test_resolver.py
```diff
@@ -311,9 +311,11 @@ async def test_async_resolver_aiodns_not_present(loop: Any, monkeypatch: Any) ->
...
     assert DefaultResolver is ThreadedResolver
```
bdraco, line 321:
```suggestion
```

tests/test_resolver.py
bdraco: Actually on windows we don't install aiodns because protractor is the default event loop, and aiodns needs selector event loop

https://github.com/saghul/aiodns#note-for-windows-users

PR #8535, opened 2024-07-25
Small speed up to cookiejar filter_cookies

bdraco:
## What do these changes do?
 
Small speed up to cookiejar

Using `str.format` is ~16% faster than the lambda

followup to https://github.com/aio-libs/aiohttp/pull/7944#discussion_r1430823536. I was hoping to use `join` there but later realized `str.format` will take `*args`

no

no

benchmark
```python
import timeit
import itertools

_FORMAT_PATH = "{0}/{1}".format

path = "lolonglonglonglonglonglongng/path/to/a/file"

print(
    timeit.timeit(
        'itertools.accumulate(path.split("/"), _FORMAT_PATH)', globals=globals()
    )
)


print(
    timeit.timeit(
        'itertools.accumulate(path.split("/"), lambda x, y: f"{x}/{y}")',
        globals=globals(),
    )
)
```

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -36,6 +36,9 @@ __all__ = ("CookieJar", "DummyCookieJar")
 
 CookieItem = Union[str, "Morsel[str]"]
 
+_FORMAT_PATH = "{}/{}".format
+_FORMAT_DOMAIN_REVERSED = "{1}.{0}".format
+
 
 class CookieJar(AbstractCookieJar):
     """Implements cookie storage adhering to RFC 6265."""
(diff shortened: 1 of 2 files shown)


aiohttp/cookiejar.py
```diff
@@ -36,6 +36,9 @@
 
 CookieItem = Union[str, "Morsel[str]"]
 
+_FORMAT_PATH = "{}/{}".format
```
Dreamsorcerer, line 40: As this is rather unconventional, I feel like there should be a comment.
```suggestion
# We cache these string methods here as their use is in performance critical code.
_FORMAT_PATH = "{}/{}".format
```

Dreamsorcerer (approved):

bdraco:
thanks

PR #8538, opened 2024-07-25
Handle 403 and 404 issues in FileResponse class

### LINKED ISSUE #8182: 404 returned if only compressed file exists

steverep:
## What do these changes do?

- Move the check in the static handler for `is_file()` (i.e. exists and regular) to the `FileResponse.prepare` method.  Revamp tests for non-regular files to be more realistic and robust.
- Add a catch for permission errors when opening the file (which are not handled anywhere currently).  Real and mock tests added here as well.
- Remove the plain text file from the `hello_txt` test fixture which automatically now ensures compressed can exist without uncompressed variants.

`FileResponse` class is now more robust.

No

## Related issue number

Closes #8182 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/8182.bugfix.rst
+++ b/CHANGES/8182.bugfix.rst
@@ -1,0 +1,9 @@
+Move file existence and type checks to ``FileResponse``,
+and fix it to handle permission errors -- by :user:`steverep`.
+
+The :py:class:`~aiohttp.web.FileResponse` class was modified to respond with
+ 403 Forbidden or 404 Not Found as appropriate.  Previously, it would cause a
+ server error if the path did not exist or could not be accessed.  Checks for
+ existence, non-regular files, and permissions were expected to be done in the
+ route handler.  For static routes, this now permits a compressed file to exist
+ without its uncompressed variant and still be served.
--- a/CHANGES/8182.bugfix.rst
+++ b/CHANGES/8182.bugfix.rst
@@ -1,9 +1,10 @@
-Move file existence and type checks to ``FileResponse``,
-and fix it to handle permission errors -- by :user:`steverep`.
+Adjusted ``FileResponse`` to check file existence and access when preparing the response -- by :user:`steverep`.
 
 The :py:class:`~aiohttp.web.FileResponse` class was modified to respond with
  403 Forbidden or 404 Not Found as appropriate.  Previously, it would cause a
  server error if the path did not exist or could not be accessed.  Checks for
  existence, non-regular files, and permissions were expected to be done in the
  route handler.  For static routes, this now permits a compressed file to exist
- without its uncompressed variant and still be served.
+ without its uncompressed variant and still be served.  In addition, this
+ changes the response status for files without read permission to 403, and for
+ non-regular files from 404 to 403 for consistency.
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -180,11 +180,10 @@ class FileResponse(StreamResponse):
         return file_path, file_path.stat(), None
 
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
+        loop = asyncio.get_running_loop()
         # Encoding comparisons should be case-insensitive
         # https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1
         accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
-
-        loop = asyncio.get_running_loop()
         try:
             file_path, st, file_encoding = await loop.run_in_executor(
                 None, self._get_file_path_stat_encoding, accept_encoding
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -4,6 +4,7 @@ import pathlib
 import sys
 from contextlib import suppress
 from mimetypes import MimeTypes
+from stat import S_ISREG
 from types import MappingProxyType
 from typing import (
     IO,
@@ -22,6 +23,8 @@ from .abc import AbstractStreamWriter
 from .helpers import ETAG_ANY, ETag, must_be_empty_body
 from .typedefs import LooseHeaders, PathLike
 from .web_exceptions import (
+    HTTPForbidden,
+    HTTPNotFound,
     HTTPNotModified,
     HTTPPartialContent,
     HTTPPreconditionFailed,
@@ -177,13 +180,23 @@ class FileResponse(StreamResponse):
         return file_path, file_path.stat(), None
 
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
-        loop = asyncio.get_event_loop()
         # Encoding comparisons should be case-insensitive
         # https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1
         accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
-        file_path, st, file_encoding = await loop.run_in_executor(
-            None, self._get_file_path_stat_encoding, accept_encoding
-        )
+
+        loop = asyncio.get_running_loop()
+        try:
+            file_path, st, file_encoding = await loop.run_in_executor(
+                None, self._get_file_path_stat_encoding, accept_encoding
+            )
+        except FileNotFoundError:
+            self.set_status(HTTPNotFound.status_code)
+            return await super().prepare(request)
+
+        # Forbid special files like sockets, pipes, devices, etc.
+        if not S_ISREG(st.st_mode):
+            self.set_status(HTTPForbidden.status_code)
+            return await super().prepare(request)
 
         etag_value = f"{st.st_mtime_ns:x}-{st.st_size:x}"
         last_modified = st.st_mtime
@@ -320,7 +333,12 @@ class FileResponse(StreamResponse):
         if count == 0 or must_be_empty_body(request.method, self.status):
             return await super().prepare(request)
 
-        fobj = await loop.run_in_executor(None, file_path.open, "rb")
+        try:
+            fobj = await loop.run_in_executor(None, file_path.open, "rb")
+        except PermissionError:
+            self.set_status(HTTPForbidden.status_code)
+            return await super().prepare(request)
+
         if start:  # be aware that start could be None or int=0 here.
             offset = start
         else:
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -690,10 +690,7 @@ class StaticResource(PrefixResource):
         except PermissionError as error:
             raise HTTPForbidden() from error
 
-        # Not a regular file or does not exist.
-        if not file_path.is_file():
-            raise HTTPNotFound()
-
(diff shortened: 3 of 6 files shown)


aiohttp/web_fileresponse.py
```diff
@@ -177,13 +180,23 @@ def _get_file_path_stat_encoding(
...
     async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter]:
-        loop = asyncio.get_event_loop()
         # Encoding comparisons should be case-insensitive
         # https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1
         accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
-        file_path, st, file_encoding = await loop.run_in_executor(
-            None, self._get_file_path_stat_encoding, accept_encoding
-        )
+
+        loop = asyncio.get_running_loop()
```
bdraco: Does this need to be moved? Any relocation we can avoid make backports easier.

aiohttp/web_fileresponse.py
```diff
@@ -177,13 +180,23 @@ def _get_file_path_stat_encoding(
...
+        try:
+            file_path, st, file_encoding = await loop.run_in_executor(
+                None, self._get_file_path_stat_encoding, accept_encoding
+            )
+        except FileNotFoundError:
+            self.set_status(HTTPNotFound.status_code)
+            return await super().prepare(request)
```
bdraco, line 193: Does this need to set the content_length to 0 like `_precondition_failed` does?

aiohttp/web_fileresponse.py
```diff
@@ -177,13 +180,23 @@ def _get_file_path_stat_encoding(
...
+
+        # Forbid special files like sockets, pipes, devices, etc.
+        if not S_ISREG(st.st_mode):
+            self.set_status(HTTPForbidden.status_code)
+            return await super().prepare(request)
```
bdraco, line 198: Does this need to set the content_length to 0 like `_precondition_failed` does?

aiohttp/web_fileresponse.py
```diff
@@ -320,7 +333,12 @@ async def prepare(self, request: "BaseRequest") -> Optional[AbstractStreamWriter
         if count == 0 or must_be_empty_body(request.method, self.status):
             return await super().prepare(request)
 
-        fobj = await loop.run_in_executor(None, file_path.open, "rb")
+        try:
+            fobj = await loop.run_in_executor(None, file_path.open, "rb")
+        except PermissionError:
+            self.set_status(HTTPForbidden.status_code)
+            return await super().prepare(request)
```
bdraco, line 339: Does this need to set the content_length to 0 like `_precondition_failed` does?

bdraco: doesn't seem like it does

aiohttp/web_fileresponse.py
bdraco: doesn't seem like it does

aiohttp/web_fileresponse.py
bdraco: doesn't seem like it does

CHANGES/8182.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Move file existence and type checks to ``FileResponse``,
+and fix it to handle permission errors -- by :user:`steverep`.
```
bdraco:
```suggestion
and handle permission errors -- by :user:`steverep`.
```

aiohttp/web_fileresponse.py
```diff
@@ -177,13 +180,23 @@ def _get_file_path_stat_encoding(
...
+            file_path, st, file_encoding = await loop.run_in_executor(
+                None, self._get_file_path_stat_encoding, accept_encoding
+            )
+        except FileNotFoundError:
+            self.set_status(HTTPNotFound.status_code)
+            return await super().prepare(request)
+
+        # Forbid special files like sockets, pipes, devices, etc.
+        if not S_ISREG(st.st_mode):
+            self.set_status(HTTPForbidden.status_code)
```
bdraco, line 197: This seems like a breaking change to not return HTTPNotFound since https://github.com/aio-libs/aiohttp/pull/8538/files#diff-9632858c8fdd45d817126b2735f763b9a2e24fb2800122eb2cb0c9a3b89c85c1L693 previously did

aiohttp/web_fileresponse.py
```diff
@@ -177,13 +180,23 @@ def _get_file_path_stat_encoding(
...
+        try:
+            file_path, st, file_encoding = await loop.run_in_executor(
+                None, self._get_file_path_stat_encoding, accept_encoding
+            )
+        except FileNotFoundError:
+            self.set_status(HTTPNotFound.status_code)
+            return await super().prepare(request)
+
+        # Forbid special files like sockets, pipes, devices, etc.
+        if not S_ISREG(st.st_mode):
```
bdraco, line 196: Not sure if `S_ISREG` will work on windows

aiohttp/web_fileresponse.py
steverep: Changed to `get_running_loop` instead of `get_event_loop`, so just moved closer to first occurrence.

Although I'm fine reverting it if you wish.  There are a ton of occurrences of `get_event_loop` that should probably be changed.

CHANGES/8182.bugfix.rst
steverep: I don't like my grammar here either, but I was trying to draw a clear distinction between what was just moved from the static handler to the prepare method, versus what wasn't handled at all.  Again from #8507 the legacy setup clearer used to expect the `resolve()` call to handle this.

aiohttp/web_fileresponse.py
steverep: I was going to comment myself here.  There are conflicting things written in the old test, which was not working properly.

403 seems like the correct response to me, so I just picked that expecting discussion.

bdraco: 403 does seems like a better response. I’m leaning towards keeping it not found though as 

- It’s revealing more information about the filesystem. Need to think about the security implications of that
- It’s technically a breaking change.


Maybe that’s ok at this point since we are going from 3.9 to 3.10. Need to sleep on this

bdraco: It's not escaping the sandbox so I do t think there is a security concern.  Let's keep it as you have it but explicitly call it out in the change message.

steverep: Spec does permit 404 for stuff that exists but otherwise inaccessible for the "hidden" effect.  However, if that's going to be the `aiohttp` mantra, then it should be consistent (i.e. all permission errors should also give 404).  That's not the case right now which is another reason I used 403 here.

aiohttp/web_fileresponse.py
bdraco: I think its a good change to switch to `get_running_loop`, but I'd leave it above and swap out `loop = asyncio.get_event_loop()` for `asyncio.get_running_loop()` to keep it easier to backport

aiohttp/web_fileresponse.py
steverep: It's what `Path.is_file()` uses under the hood.  Tests pass.

CHANGES/8182.bugfix.rst
steverep: Take another look.  I simplified the title part and added a sentence about the actual behavior changes.

aiohttp/web_fileresponse.py
Dreamsorcerer: It leaks information about something existing on the filesystem.

The only issue I can think of that leading to, is if the app supports uploading of files, using the same name for the upload could replace that file or result in the permissions changing. But, that seems like a bit of a stretch.

From the user perspective, if they are a general user of the app, then a 403 is not useful information and potentially confusing as it suggests they have the wrong login credentials or something. It would only be somewhat useful if the user is also the developer of the app (but, maybe a warning log could resolve this in a safer way).

steverep: For posterity, here's the short [RFC 9110 section on 403](https://www.rfc-editor.org/rfc/rfc9110#name-403-forbidden).

> It leaks information about something existing on the filesystem.

Right, but if your server becomes vulnerable by that info alone, you've got bigger problems, for which a 404 isn't the solution.

> From the user perspective, if they are a general user of the app, then a 403 is not useful information and potentially confusing as it suggests they have the wrong login credentials or something. 

That case is actually where it seems useful to me.  It says you've done everything right, but the admin didn't give your login the right permissions (as opposed to a 404 which says the app is broken for everyone).

My only strong opinion here would be consistency - either always send 404 instead of 403, or use 403 whenever it exists.  Don't mix and match.  If there's an outcry for security via obscurity, then it would not be difficult to introduce an app flag for a `hidden_forbidden` mode to swap the handful of places it's raised.

bdraco (approved):
Tested on production HA instances. All good

aiohttp/web_fileresponse.py
bdraco: I didn't merge this as I'm leaving this open for Sam to respond.

Dreamsorcerer: I don't mind. I'm on the fence at to which way to go.

bdraco:
Thanks @steverep

aiohttp/web_urldispatcher.py
```diff
@@ -690,10 +690,7 @@ def _resolve_path_to_response(self, unresolved_path: Path) -> StreamResponse:
         except PermissionError as error:
             raise HTTPForbidden() from error
 
-        # Not a regular file or does not exist.
-        if not file_path.is_file():
-            raise HTTPNotFound()
```
Dreamsorcerer, line 695: This breaks some code in aiohttp-devtools:
https://github.com/aio-libs/aiohttp-devtools/blob/270aa58741b9aad268e32cb7bd755fda4c562731/aiohttp_devtools/runserver/serve.py#L404

Any suggestions on how to proceed?

PR #8542, opened 2024-07-28
WSS: Fix heartbeat timeout logic

### LINKED ISSUE #8540: WSS: `_pong_not_received` works (mostly) correctly but outcome is erased by `receive`'s `except EofStream:`
### Describe the bug

I'm working on a patch.

### To Reproduce

### Expected behavior

`receive`:

### Logs/tracebacks

```python-traceback
Above
```

### Python Version

```console
$ python --version
```

### aiohttp Version

```console
$ python -m pip show aiohttp
3.9.5
```

### multidict Version

```console
Not relevant
```

### yarl Version

```console
$ python -m pip show yarl
Not relevant
```

### OS

Fedora 40, Linux, x86_64

### Related component

Client

### Additional context

_No response_

### Code of Conduct

arcivanov:

fixes #8540

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -4,7 +4,7 @@ import asyncio
 import sys
 from typing import Any, Optional, cast
 
-from .client_exceptions import ClientError
+from .client_exceptions import ClientError, ServerTimeoutError
 from .client_reqrep import ClientResponse
 from .helpers import call_later, set_result
 from .http import (
@@ -117,8 +117,12 @@ class ClientWebSocketResponse:
         if not self._closed:
             self._closed = True
             self._close_code = WSCloseCode.ABNORMAL_CLOSURE
-            self._exception = asyncio.TimeoutError()
+            self._exception = ServerTimeoutError()
             self._response.close()
+            if self._waiting is not None and not self._closing:
+                self._reader.feed_data(
+                    WSMessage(WSMsgType.ERROR, self._exception, None), 0
+                )
 
     @property
     def closed(self) -> bool:


bdraco:
Does this only happen on 3.9?

arcivanov:
@bdraco it's my next step. I caught this in production, production sits on 3.9.5 so I haven't explored yet 3.10 or master, but I will do so shortly.

arcivanov:
@bdraco 3.10.0b1 is definitely affected, same behavior - the code is identical too:

image
image

I just reproduced it with my app with unpatched 3.10.0b1 - app exits as if it was commanded to do so because OK EofStream has been received.

arcivanov:
The patch for 3.10 would be slightly different due to the `waiting` being changed just like here:

https://github.com/aio-libs/aiohttp/blob/df57b9f34ed4c696d3de3f3229dc84c2f9e6ac84/aiohttp/client_ws.py#L189-L194

bdraco:
3.10 should get an rc0 later today and hopefully a stable release early this week.  
3.9 is done at this point and there will be no more releases.

Please target master and a 3.10 backport if it’s different.  If it’s ready to go with tests I can try to get it into rc0 later this afternoon assuming everything else ready.  If not we can always do a 3.10.1 later next week.

Dreamsorcerer:
> Please target master and a 3.10 backport if it’s different

To clarify, we'd need a PR to master first. Then, ideally, follow the patchbot instructions after it is merged to cherry-pick into a 3.10 PR.

Dreamsorcerer:
We'd also really want a regression test that reproduces the behaviour.

arcivanov:
> We'd also really want a regression test that reproduces the behaviour.

Regression-wise, do you have an infrastructure to support? We need a WSS client/server setup where we can physically sever the connection between the two so that FIN-FIN/ACK-ACK don't propagate right away and heartbeat times out first.

bdraco:
You’ll need to patch some internals to simulate that condition.

arcivanov:
> You’ll need to patch some internals to simulate that condition.

Actually nope, I've added the functional test which simulates it here:

https://github.com/aio-libs/aiohttp/pull/8543/commits/56c0f1f4bc03d636f707a05a046f47caf0f69a50#

arcivanov:
I'm closing this one since 3.9.5 is dead.

PR #8543, opened 2024-07-28
[3.10] WSS: Fix heartbeat timeout logic

### LINKED ISSUE #8540: WSS: `_pong_not_received` works (mostly) correctly but outcome is erased by `receive`'s `except EofStream:`

arcivanov:

fixes #8540

## Checklist

### DIFF
--- a/CHANGES/8540.bugfix.rst
+++ b/CHANGES/8540.bugfix.rst
@@ -1,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
+
+The :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse._pong_not_received` did not
+ terminate a :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse.receive` operation.
+ This change causes `_pong_not_received` to feed the `reader` an error message, causing
+ pending `receive` to terminate and return the error message. The error message contains
+ the exception :py:class:`~aiohttp.client_exceptions.ServerTimeoutError`.
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -4,7 +4,7 @@ import asyncio
 import sys
 from typing import Any, Optional, cast
 
-from .client_exceptions import ClientError
+from .client_exceptions import ClientError, ServerTimeoutError
 from .client_reqrep import ClientResponse
 from .helpers import call_later, set_result
 from .http import (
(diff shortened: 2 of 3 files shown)


arcivanov:
@bdraco @Dreamsorcerer Please provide feedback based on this PR and if satisfactory I'll adopt it to master as well.

Please let me know if you think documentation changes are necessary.

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
+
+The :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse._pong_not_received` did not
+ terminate a :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse.receive` operation.
```
Dreamsorcerer, line 4: When I run the test, it seems like the .receive() terminates just fine. The only change is from receiving a WSCloseCode.OK to a WSCloseCode.ABNORMAL_CLOSURE.

arcivanov: Well, firstly that's a huge deal - it's indication of nominal closure vs abnormal termination. That was the crux for my application that mistook an abnormal disconnect for an orderly shutdown.
Secondly, I have a `receive` with a timeout, i.e. it will terminate naturally by itself on timeout, which is 1.0 s in the test.
Thirdly, it does terminate _in vivo_ as well during hard connection severance, just much later than the pong timeout is reached, whereas feeding the `reader` an error message terminates a pending receive immediately as should the intention be.

arcivanov: I have updated the test @Dreamsorcerer to simulate EOF not reaching either party (which is how hard-severed connection behaves) and the case is reproduced perfectly if you disable the `feed_data` stanza.

Dreamsorcerer: OK, I think this looks about right now, but @bdraco is more familiar with the expected behaviour of websockets, so I'll wait for him to confirm.
We do need the master PR to go in first though.

bdraco:
superseded by https://github.com/aio-libs/aiohttp/pull/8573

PR #8546, opened 2024-07-29
Fix WebSocket server heartbeat timeout logic

### LINKED ISSUE #8540: WSS: `_pong_not_received` works (mostly) correctly but outcome is erased by `receive`'s `except EofStream:`

arcivanov:

fixes #8540

## Checklist


tests/test_client_ws_functional.py
```diff
@@ -685,6 +685,36 @@ async def handler(request):
...
 
+async def test_heartbeat_no_pong_concurrent_receive(aiohttp_client: Any) -> None:
+    ping_received = False
+
+    async def handler(request):
+        nonlocal ping_received
+        ws = web.WebSocketResponse(autoping=False)
+        await ws.prepare(request)
+        msg = await ws.receive()
+        if msg.type == aiohttp.WSMsgType.PING:
```
```suggestion
        if msg.type is aiohttp.WSMsgType.PING:
```

tests/test_client_ws_functional.py
```diff
@@ -685,6 +685,36 @@ async def handler(request):
...
+            ping_received = True
+        ws._reader.feed_eof = lambda: None
+        await asyncio.sleep(10.0)
+        return ws
```
bdraco: Maybe we should raise here if its ever reached since that means something has gone wrong with the test and mark with no cover

webknjaz: Perhaps, mark with `typing.assert_never()` and stick that into `exclude_also` in `.coveragerc`?

tests/test_client_ws_functional.py
arcivanov: The `test_heartbeat_no_pong` uses an equals sign and I would keep this for consistency among the tests.

tests/test_client_ws_functional.py
arcivanov: If this is a question of coverage, we can exclude it. 
But I would draw your attention to the test this one is based on right above it in the code `test_heartbeat_no_pong` which I'm basing my code on, which has no assert as it's not a testable condition.

I would also kindly suggest that if there is a desire to refactor the tests overall, changing `==` or `is` and adding `asserts` then perhaps a separate issue/effort can be created for this. Refactoring one test leaving the other 20 _old style_ so to speak seems like trying to drain an ocean with a cup.

arcivanov: I have fixed the line 700 not being covered by deleting it:

https://app.codecov.io/gh/aio-libs/aiohttp/pull/8546?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aio-libs#68da58a6c1b4b9f2ac78950734ab927a-R699

tests/test_client_ws_functional.py
bdraco: Enums are singletons so an identity check should be used when both sides are an enum. The other test should be updated as well in a different PR. It’s also ok not to go refactor old code, but we shouldn’t propagate that pattern in new code.

arcivanov: fixed

tests/test_client_ws_functional.py
```diff
@@ -685,6 +685,35 @@ async def handler(request):
...
+async def test_heartbeat_no_pong_concurrent_receive(aiohttp_client: Any) -> None:
+    ping_received = False
+
+    async def handler(request):
+        nonlocal ping_received
+        ws = web.WebSocketResponse(autoping=False)
+        await ws.prepare(request)
+        msg = await ws.receive()
+        if msg.type is aiohttp.WSMsgType.PING:
+            ping_received = True
```
bdraco: It looks like we never get anything but a PING here so maybe make this an assertion instead?

arcivanov: The assertion is in the main test function. Can't really assert out of the request handler. The request will fail but it won't appear as a failed assertion.

bdraco: Do you have a suggestion for an alternate solution to solve the missing branch coverage?

arcivanov: You mean the partial missing branch coverage in a test source code? 

I don't use coverage to cover the tests and in 20+ years haven't seen other people doing it either. The goal is to have 100% coverage of the code **under test**, not the test itself. It's like taking the EKG of the patient AND the attending physician. You *can* do it, but the EKG of the physician provides no actionable information.

I would simply exclude the entire test suite from coverage reporting.

bdraco: I'll leave this PR for other members to comment on how the test coverage reporting was set up as its before my time, and I need to find out the history and motivation here before I can respond to your comment.

Dreamsorcerer: > I don't use coverage to cover the tests and in 20+ years haven't seen other people doing it either. The goal is to have 100% coverage of the code **under test**, not the test itself. It's like taking the EKG of the patient AND the attending physician. You _can_ do it, but the EKG of the physician provides no actionable information.

On the contrary, if test code isn't running (i.e. there isn't coverage) it shouldn't exist. When we added coverage for tests we found about a dozen tests that were not running at all, or not running the intended test code.

Dreamsorcerer: > The assertion is in the main test function. Can't really assert out of the request handler. The request will fail but it won't appear as a failed assertion.

I think an assert is fine. The test will fail due to it not running the rest of the code, and the logs should show the assertion failure with a traceback.

arcivanov: > On the contrary, if test code isn't running (i.e. there isn't coverage) it shouldn't exist. When we added coverage for tests we found about a dozen tests that were not running at all, or not running the intended test code.

The benefit of knowing whether the test code is running (it's the only arguable benefit, as you can't even correlate the coverage of the test code with coverage of the code under test) for me is absolutely not worth the effort required to achieve 100% line *and* branch coverage in the test. At worst you have some non-running code in your tests which is *functionally irrelevant* if you have 100%  line *and* branch coverage of the code under test.

arcivanov: > I think an assert is fine. The test will fail due to it not running the rest of the code, and the logs should show the assertion failure with a traceback.

No it really isn't fine: 

```python
async def test_heartbeat_no_pong_concurrent_receive(aiohttp_client: Any) -> None:
    ping_received = False

    async def handler(request):
        nonlocal ping_received
        ws = web.WebSocketResponse(autoping=False)
        await ws.prepare(request)
        msg = await ws.receive()
        assert False
        if msg.type is aiohttp.WSMsgType.PING:
            ping_received = True
        ws._reader.feed_eof = lambda: None
        await asyncio.sleep(10.0)

    app = web.Application()
    app.router.add_route("GET", "/", handler)

    resp = await client.ws_connect("/", heartbeat=0.1)
    resp._reader.feed_eof = lambda: None

    # Connection should be closed roughly after 1.5x heartbeat.
    msg = await resp.receive(5.0)
    # assert ping_received
    assert resp.close_code is WSCloseCode.ABNORMAL_CLOSURE
    assert msg
    assert msg.type is WSMsgType.ERROR
    assert isinstance(msg.data, ServerTimeoutError)
    
```

Results in a 100% passing test: 

```
Launching pytest with arguments test_client_ws_functional.py::test_heartbeat_no_pong_concurrent_receive --no-header --no-summary -q in /home/arcivanov/Documents/src/arcivanov/aiohttp/tests

============================= test session starts ==============================
collecting ... collected 1 item

test_client_ws_functional.py::test_heartbeat_no_pong_concurrent_receive[pyloop] /home/arcivanov/Documents/src/arcivanov/aiohttp/venv/.pyenv/versions/lib64/python3.12/site-packages/coverage/inorout.py:503: CoverageWarning: Module tests/ was never imported. (module-not-imported)
  self.warn(f"Module {pkg} was never imported.", slug="module-not-imported")


============================== 1 passed in 0.86s ===============================
PASSED [100%]
Process finished with exit code 0
```

Dreamsorcerer: > At worst you have some non-running code in your tests which is _functionally irrelevant_ if you have 100% line _and_ branch coverage of the code under test.

First, we don't have 100% coverage and achieving 100% coverage on the entire codebase would take a significant amount of work. Second, we have tests which test different expected behaviours of the code, even if a single test produces "full coverage" of that code (many of our regression tests when fixing bugs do not increase coverage of the code base). So, being aware that several of those tests are not even running is very useful. We've found numerous cases across projects where coverage has highlighted tests not running, broken tests or old code that should be removed or cleaned up (fixtures no longer used anywhere etc.).

> Results in a 100% passing test:

And an assert doesn't? I don't see how that makes it not fine to write an assert, the code makes sense either way, an assert just makes it clear that we should never see another type in this test.

arcivanov: > And an assert doesn't? I don't see how that makes it not fine to write an assert, the code makes sense either way, an assert just makes it clear that we should never see another type in this test.


I have added an `assert False` right above `if msg.type is aiohttp.WSMsgType.PING:`. The test passes, the assert in the handler is ignored.

Dreamsorcerer: Oh, I see what you're saying now. As I said, the test would fail due to the outer assert, and then I'd expect to see the inner assert log a traceback. To see the logs on a successful test, I think you need to run with `pytest -s`.

Dreamsorcerer: Yes, running that myself after putting the outer assert back in, the test fails and I see this in the output:
```
------------------------------------------------------------------------------------------------- Captured log call -------------------------------------------------------------------------------------------------
ERROR    aiohttp.server:web_protocol.py:442 Error handling request
  File "/home/ubuntu/hacking/aiohttp/aiohttp/web_protocol.py", line 472, in _handle_request
    resp = await request_handler(request)
  File "/home/ubuntu/hacking/aiohttp/aiohttp/web_app.py", line 388, in _handle
    resp = await handler(request)
  File "/home/ubuntu/hacking/aiohttp/tests/test_client_ws_functional.py", line 951, in handler
    assert False
AssertionError: assert False
```

That's what I meant.

arcivanov: So, do I understand you correctly: you're proposing replacing a functional assertion that works (works == fails the test when evaluated to False), with the one that doesn't work (i.e. will not fail the test) but will *indicate* a failure in the logs but only if the test harness is run in a verbose mode and only if manually parsed by human eye, all in pursuit of 100% test branch coverage which has no functional relevance for the code under test?

Dreamsorcerer: And again, if the test gets messed up in future and ends up something like what you posted, then coverage will tell us that a chunk of our test is not running and there is probably an issue with the test. So, I think you just demonstrated a perfect example of why we do coverage here (and I vaguely recall such tests like this where the handlers were failing and the tests were still passing).

Dreamsorcerer: > you're proposing replacing a functional assertion

No, I explicitly said that assertion needs to stay, not be replaced. I'm suggesting replacing an if statement which is unneeded (and suggests that could be a case where we get a different response) with an assert.

Dreamsorcerer: > with the one that doesn't work

So, all we'd be doing is _adding_ information. The test would still fail on the same assertion, but we'd _also_ see a log showing the assertion failure in the handler, making it even easier for a developer to understand what went wrong when the test fails.

Dreamsorcerer:
```suggestion
        assert msg.type is aiohttp.WSMsgType.PING
        ping_received = True
```

arcivanov: I have pushed a rewrite that eliminates the branch. Unfortunately I can't dedicate more time to this unless it's something functionally significant.

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
```
bdraco:
```suggestion
Fixed WebSocket server heartbeat timeout logic to terminate `receive` and return :py:class:`~aiohttp.ServerTimeoutError` -- by :user:`arcivanov`.
```

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
...
+
+The :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse._pong_not_received` did not
+ terminate a :py:meth:`~aiohttp.client_ws.ClientWebSocketResponse.receive` operation.
+ This change causes `_pong_not_received` to feed the `reader` an error message, causing
+ pending `receive` to terminate and return the error message. The error message contains
+ the exception :py:class:`~aiohttp.client_exceptions.ServerTimeoutError`.
```
bdraco:
```
/home/runner/work/aiohttp/aiohttp/docs/[towncrier-fragments]:209: WARNING: py:meth reference target not found: aiohttp.client_ws.ClientWebSocketResponse._pong_not_received
/home/runner/work/aiohttp/aiohttp/docs/[towncrier-fragments]:205: WARNING: py:meth reference target not found: aiohttp.client_ws.ClientWebSocketResponse.receive
/home/runner/work/aiohttp/aiohttp/docs/[towncrier-fragments]:205: WARNING: py:class reference target not found: aiohttp.client_exceptions.ServerTimeoutError
```

```suggestion
The :py:meth:`~aiohttp.ClientWebSocketResponse._pong_not_received` did not
 terminate a :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation.
 This change causes `_pong_not_received` to feed the `reader` an error message, causing
 pending `receive` to terminate and return the error message. The error message contains
 the exception :py:class:`~aiohttp.ServerTimeoutError`.
```

I'm not sure if this will fix the above, but let's give it a try.

CHANGES/8540.bugfix.rst
bdraco: Not sure thats exactly the best wording here so I left it as a suggestion.

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
+
+The :py:meth:`~aiohttp.ClientWebSocketResponse._pong_not_received` did not
+ terminate a :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation.
```
bdraco:
```suggestion
When a WebSocket pong message was not received, the
 :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation was not terminated.
```

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
+
+When a WebSocket pong message was not received, the
+ :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation was not terminated.
```
bdraco:
```suggestion
When a WebSocket pong message was not received, the
 :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation did not terminate.
```

CHANGES/8540.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+WSS: Fix heartbeat timeout logic -- by :user:`arcivanov`.
+
+When a WebSocket pong message was not received, the
+ :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation did not terminate.
+ This change causes `_pong_not_received` to feed the `reader` an error message, causing
+ pending `receive` to terminate and return the error message. The error message contains
+ the exception :py:class:`~aiohttp.ServerTimeoutError`.
```
bdraco, line 7: I adjusted this to get the linter to pass. I tried to keep it as close to the original intent as possible. Please feel free to make additional changes if I missed something here.

bdraco:
preview is here https://aiohttp--8546.org.readthedocs.build/en/8546/changes.html

CHANGES/8540.bugfix.rst
arcivanov: I'm OK with whatever wording you're ok with.

bdraco (approved):

bdraco:
I want to do some more manual testing with this before merging

bdraco:
testing seems ok

bdraco (approved):
Thanks @arcivanov

Issue #8529, opened 2024-07-22
** 3.10 testing **


bdraco:
We've introduced a few significant changes in 3.10, so it would be great to get some community testing before promoting it as a stable release. We've released 3.10.0rc0 as a prerelease to help with this.

Can be installed with: `pip install 'aiohttp<4' --pre --upgrade`

In particular, these are things we are interested on testing or gathering feedback on:

- Implement happy eyeballs https://github.com/aio-libs/aiohttp/pull/8005
- Improved URL handler resolution time by indexing resources in the UrlDispatcher. For applications with a large number of handlers, this should increase performance significantly. https://github.com/aio-libs/aiohttp/pull/7917
- Restore AsyncResolver to be the default resolver. https://github.com/aio-libs/aiohttp/pull/8522 https://github.com/aio-libs/aiohttp/pull/8270
- Changes to the shutdown logic Shutdown logic: Only wait on handlers (https://github.com/aio-libs/aiohttp/pull/8530 .. changelog details forthcoming)

Release checklist
- [x] docs/changelog/tests for https://github.com/aio-libs/aiohttp/pull/8495
- [x] https://github.com/aio-libs/aiojobs/pull/495
- [x] aiojobs https://github.com/aio-libs/aiojobs/releases/tag/v1.3.0

Releases
- https://github.com/aio-libs/aiohttp/pull/8531 https://github.com/aio-libs/aiohttp/actions/runs/10044635669
- https://github.com/aio-libs/aiohttp/pull/8533 https://github.com/aio-libs/aiohttp/actions/runs/10045374258
- https://github.com/aio-libs/aiohttp/pull/8545 https://github.com/aio-libs/aiohttp/actions/runs/10145560535
- https://github.com/aio-libs/aiohttp/pull/8550 https://github.com/aio-libs/aiohttp/actions/runs/10163532577 (cancelled)
- https://github.com/aio-libs/aiohttp/pull/8552 https://github.com/aio-libs/aiohttp/actions/runs/10168515264

bdraco:
Bumped both my Home Assistant large production and all good https://github.com/home-assistant/core/pull/122409

bdraco:
As the testing period has ended, and we are moving to release, please report any new issues by creating a new issue report https://github.com/aio-libs/aiohttp/issues/new/choose

webknjaz:
> Can be installed with: `pip install 'aiohttp<4' --pre --upgrade`

[...]

PR #8550, opened 2024-07-30
Release 3.10.0

bdraco:
release tracking issue closes https://github.com/aio-libs/aiohttp/issues/8529

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,7 @@
 
 .. towncrier release notes start
 
-3.10.0rc0 (2024-07-29)
+3.10.0 (2024-07-30)
 ========================
 
 Bug fixes
@@ -32,62 +32,6 @@ Bug fixes
   :issue:`8182`.
 
 
-
-
-Removals and backward incompatible breaking changes
----------------------------------------------------
-
-- The shutdown logic in 3.9 waited on all tasks, which caused issues with some libraries.
-  In 3.10 we've changed this logic to only wait on request handlers. This means that it's
-  important for developers to correctly handle the lifecycle of background tasks using a
-  library such as ``aiojobs``. If an application is using ``handler_cancellation=True`` then
-  it is also a good idea to ensure that any :func:`asyncio.shield` calls are replaced with
-  :func:`aiojobs.aiohttp.shield`.
-
-  Please read the updated documentation on these points:
-  https://docs.aiohttp.org/en/stable/web_advanced.html#graceful-shutdown
-  https://docs.aiohttp.org/en/stable/web_advanced.html#web-handler-cancellation
-
-  -- by :user:`Dreamsorcerer`
-
-
-  *Related issues and pull requests on GitHub:*
-  :issue:`8495`.
-
-
-
-
-Miscellaneous internal changes
-------------------------------
-
-- Improve performance of filtering cookies -- by :user:`bdraco`.
-
-  This change is a followup to the improvements in :issue:`7583`
-
-
-  *Related issues and pull requests on GitHub:*
-  :issue:`8535`.
-
-
-
-
-----
-
-
-3.10.0b1 (2024-07-22)
-========================
-
-No significant changes.
-
-
-----
-
-
-3.10.0b0 (2024-07-22)
-=========================
-
-Bug fixes
----------
 
 - Fixed server response headers for ``Content-Type`` and ``Content-Encoding`` for
   static compressed files -- by :user:`steverep`.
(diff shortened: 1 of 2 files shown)


bdraco:
There are no changes since rc0 so had to add the skip label

CHANGES.rst
```diff
@@ -268,9 +212,42 @@ Contributor-facing changes
...
+- The shutdown logic in 3.9 waited on all tasks, which caused issues with some libraries.
+  In 3.10 we've changed this logic to only wait on request handlers. This means that it's
+  important for developers to correctly handle the lifecycle of background tasks using a
+  library such as ``aiojobs``. If an application is using ``handler_cancellation=True`` then
+  it is also a good idea to ensure that any :func:`asyncio.shield` calls are replaced with
+  :func:`aiojobs.aiohttp.shield`.
+
+  Please read the updated documentation on these points:
+  https://docs.aiohttp.org/en/stable/web_advanced.html#graceful-shutdown
+  https://docs.aiohttp.org/en/stable/web_advanced.html#web-handler-cancellation
```
webknjaz, line 227: @bdraco FYI subsequent lines are merged into continuous paragraphs in RST. This is basically rendered as a single line: https://aiohttp--8550.org.readthedocs.build/en/8550/changes.html#removals-and-backward-incompatible-breaking-changes.
Did you expect this to be a list?

Perhaps, it's worth adding a changelog render preview step to the release checklist?

CHANGES.rst
```diff
@@ -268,9 +212,42 @@ Contributor-facing changes
 
 
 
+Removals and backward incompatible breaking changes
```
webknjaz, line 215: @bdraco I'd expect breaking changes to be closer to the beginning of the change log… Our current Towncrier config places it right after the deprecations: https://github.com/aio-libs/aiohttp/blob/a561fa9/pyproject.toml#L43-L48.

Any reason you wanted to relocate after the contrib section? In general, we should let towncrier order the sections (the order corresponds to the order in the config).
contrib/packaging/misc are placed at the very end since their audiences are small and most people won't need to read them.

CHANGES.rst
```diff
@@ -268,9 +212,42 @@ Contributor-facing changes
...
+
+
+
+
 Miscellaneous internal changes
 ------------------------------
 
+- Improve performance of filtering cookies -- by :user:`bdraco`.
+
+  This change is a followup to the improvements in :issue:`7583`
```
webknjaz, line 243: missing trailing period?

CHANGES.rst
```diff
@@ -268,9 +212,42 @@ Contributor-facing changes
...
+  *Related issues and pull requests on GitHub:*
+  :issue:`8495`.
+
+
+
+
 Miscellaneous internal changes
 ------------------------------
 
+- Improve performance of filtering cookies -- by :user:`bdraco`.
```
webknjaz, line 241: @bdraco I'm usually trying to edit the change log before releasing to make its writing style consistent. Since it's a summary of news for people to find out what's happened since the previous release, the writing style should be different from Git commit messages. Imperative mood (“Improve”) reads as if it's a TODO-list time, not a statement on something that has happened. I sometimes recommend using the past tense or non-imperative present tense, referencing the differences between the versions.

This entry says “Improve”, the following one is “Improved”, followed by “Add”. They should all be consistent between each other, really. Hence, the need to rephrase entries.

Personally, I like doing this before generating the change log with Towncrier, just by checking the draft in the docs. This way, I can edit separate change notes and let Towncrier compose the end result w/o editing this file directly.

With pre-releases, it'd be possible to just revert the previous commit that did the generation, apply changes to all the fragments and generate a new commit. In the past, though, I'd branch out the pre-release change log commit without merging that into stable branches: this way, I could cut another release from a stable branch later on, without concerning myself with reverting the commit and getting the fragments back.

CHANGES.rst
```diff
@@ -268,9 +212,42 @@ Contributor-facing changes
...
+
+  This change is a followup to the improvements in :issue:`7583`
+
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`8535`.
```
webknjaz, line 247: Another thing that caught my eye is that this says it's a follow-up to a change that first appeared in this release and has its own change log entry already. To the end-users, it doesn't really make sense to read about such series of changes within the same release cycle separately.
Instead, I'd merge this information into the pre-existing change note, updating it. You can credit more people there, if needed. You can also list multiple issues/PRs in the related section.

If you were to go the Towncrier route, you could make symlinks with different numbers to the same change log fragment, and it'd just enumerate them all in a comma-separated manner. Our configuration is capable of handling that. See: https://yarl.aio-libs.org/en/latest/changes/#packaging-updates-and-notes-for-downstreams.

CHANGES.rst
bdraco: I had manually merged the notes together.  I checked them all back out into https://github.com/aio-libs/aiohttp/pull/8552 and will let towncrier do it

CHANGES.rst
bdraco: I made 8535 a symlink to 7583

CHANGES.rst
bdraco: Deleted since 8535 is now a symlink to 7583

CHANGES.rst
bdraco: Deleted since 8535 is now a symlink to 7583

CHANGES.rst
bdraco: I made the links a list

bdraco: I'm not sure how to fix this one as it doesn't seem RST supports list inside of a list.  I've tried line blocks `|`, `|br|` and back slashes.

bdraco: Nevermind, it does look like the backslashes worked.

CHANGES.rst
bdraco: Adjusted tense on all the ones that needed to to be past tense.

PR #8552, opened 2024-07-30
Release 3.10.0 (attempt 2)

bdraco:
release tracking issue https://github.com/aio-libs/aiohttp/issues/8529

New release PR to address the feedback in https://github.com/aio-libs/aiohttp/pull/8550


bdraco:
Looks good now. Let's give this another shot.

CHANGES.rst
```diff
@@ -180,18 +180,41 @@ Features
...
+---------------------------------------------------
+
+- The shutdown logic in 3.9 waited on all tasks, which caused issues with some libraries.
+  In 3.10 we've changed this logic to only wait on request handlers. This means that it's
+  important for developers to correctly handle the lifecycle of background tasks using a
+  library such as ``aiojobs``. If an application is using ``handler_cancellation=True`` then
+  it is also a good idea to ensure that any :func:`asyncio.shield` calls are replaced with
+  :func:`aiojobs.aiohttp.shield`.
+
+  Please read the updated documentation on these points: \
```
webknjaz, line 193: @bdraco here's the preview: https://aiohttp--8552.org.readthedocs.build/en/8552/changes.html#removals-and-backward-incompatible-breaking-changes. This didn't change the render. The right fix would be to have a list with leading dashes or asterisks.

CHANGES.rst
```diff
@@ -314,7 +304,7 @@ Miscellaneous internal changes
 
 
 
-- Restore :py:class:`~aiohttp.resolver.AsyncResolver` to be the default resolver. -- by :user:`bdraco`.
+- Restored :py:class:`~aiohttp.resolver.AsyncResolver` to be the default resolver. -- by :user:`bdraco`.
```
webknjaz, line 307: That period in the middle could've been avoided.

CHANGES.rst
bdraco: I'm missing something as the render looks fine to me
Screenshot 2024-07-31 at 7 48 36 PM

Maybe its my screen size?

webknjaz: Yes, it's your screen. Inspect with DevTools and you'll see that it's a paragraph with no line breaks.
aiohttp-rm-changelog

FWIW, `\` in RST has a rather weird meaning. Sometimes, I use it to escape invisible spaces in between bits of syntax.

P.S. I'd rather have this generate `<li>`s.

webknjaz: (that's at 200% zoom in Firefox)

bdraco: I did try leading dashes but that didn't work either as its a list inside of a list

webknjaz: It would work, but might need an extra empty line after `:`. Feel free to ask me for RST advice next time — I have some experience with it and Sphinx internals+extensions.

webknjaz: Here's an example of a nested list that does work and is still a part of said changelog document: https://github.com/aio-libs/aiohttp/pull/8066/files#diff-bd97a3018f85ae60eb7b42978db8f4ae548d3090cc95bc3e8d64f1d825d4c5ee

bdraco: Thanks.  That helps.

CHANGES.rst
bdraco: Do you think it’s worth adjusting before 3.10.1, or should we leave it as is?

webknjaz: That's up to you. I know that it takes a lot of energy to make the change log nice. There's more things that could be fixed, like unlinked/unhighlited function mentions and perhaps other inconsistencies. I'd fix all of them if we were to work on improving it more systemically.

PR #8566, opened 2024-07-31
Fix url dispatcher index when variable is preceded by a fixed string after a slash

### LINKED ISSUE #8567: Matching tail pattern in add_route/add_* handling not working as expected in 3.10.0
### Describe the bug

The usage of https://docs.aiohttp.org/en/stable/faq.html#how-do-i-create-a-route-that-matches-urls-with-a-given-prefix seems to yield different results in 3.10.0 than in 3.9.5 for a certain condition. I've included a gist in 'reproduce' which mostly mimics what our code handles

While our code might not be the neatest, we use this for our testsuite (hence no functional user issue) here https://github.com/plugwise/python-plugwise/blob/672335ddf766799df5a8d0ba32d7d04332b8a5a2/tests/test_init.py#L114 with the constants defined around L29

### To Reproduce

https://gist.github.com/CoMPaTech/541eb68906431933d2b3ea48e886b92b

### Expected behavior

Parsing incoming URLs (for example) `PUT /core/locations;id=c50f167537524366a5af7aa3942feb1e/thermostat;id=711861a52ed84f37a2d854afba34d9e6` While we also frown upon the url setup, it's what the physical boxes have in their firmware so beyond our ability to change.

### Logs/tracebacks

```python-traceback
See gist I nto reproduce
```

### Python Version

```console
3.12
```

### aiohttp Version

```console
3.10.0 (failing)
3.9.5 and earlier (functioning as expected)
```

### multidict Version

```console
6.0.5
```

### yarl Version

```console
1.9.4
```

### OS

Able to reproduce in several containers, ubuntu and macOS (regardless of version)

### Related component

Server

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

Fix url dispatcher index when variable is preceded by a fixed string after a slash.  We need to strip off the fixed string when creating the index key so we can match against the path at that part of the url.  

Bugfix for a regression in 3.10.x. Regressed in #7829 fixes #8567

no

### DIFF
--- a/tests/test_web_urldispatcher.py
+++ b/tests/test_web_urldispatcher.py
@@ -919,7 +919,7 @@ async def test_url_with_many_slashes(aiohttp_client: AiohttpClient) -> None:
 
 
 async def test_route_with_regex(aiohttp_client: AiohttpClient) -> None:
-    """Test a route with a regex."""
+    """Test a route with a regex preceded by a fixed string."""
     app = web.Application()
 
     async def handler(request: web.Request) -> web.Response:
--- a/tests/test_web_urldispatcher.py
+++ b/tests/test_web_urldispatcher.py
@@ -916,3 +916,27 @@ async def test_url_with_many_slashes(aiohttp_client: AiohttpClient) -> None:
     r = await client.get("///a")
     assert r.status == 200
     await r.release()
+
+
+async def test_route_with_regex(aiohttp_client: AiohttpClient) -> None:
+    """Test a route with a regex."""
+    app = web.Application()
+
+    async def handler(request: web.Request) -> web.Response:
+        assert isinstance(request.match_info._route.resource, Resource)
+        return web.Response(text=request.match_info._route.resource.canonical)
+
+    app.router.add_get("/core/locations{tail:.*}", handler)
+    client = await aiohttp_client(app)
+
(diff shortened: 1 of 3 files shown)


tests/test_web_urldispatcher.py
```diff
@@ -916,3 +916,27 @@ async def get(self) -> web.Response:
     r = await client.get("///a")
     assert r.status == 200
     await r.release()
+
+
+async def test_route_with_regex(aiohttp_client: AiohttpClient) -> None:
+    """Test a route with a regex."""
```
bdraco:
```suggestion
    """Test a route with a regex preceded by a fixed string."""
```

bdraco:
This was reported to me out of band. An issue report is forthcoming tomorrow.

bdraco:
Will finish this up once I get confirmation from the reporter and an issue report

CoMPaTech:
As indicated can confirm that both test (via gist) and CI for module building (manually patching within respective `venv`s works)

CoMPaTech (approved):
LGTM (and tnx @bdraco)

bdraco:
Thanks for testing

Dreamsorcerer (approved):

PR #8587, opened 2024-08-02
Release 3.10.1

bdraco:
Screenshot 2024-08-02 at 8 05 26 AM


CHANGES.rst
```diff
@@ -10,6 +10,62 @@
 
 .. towncrier release notes start
 
+3.10.0.dev0 (2024-08-02)
```
bdraco:
```suggestion
3.10.1 (2024-08-02)
```

CHANGES.rst
```diff
@@ -10,6 +10,62 @@
...
+========================
+
+Bug fixes
+---------
+
+- Fixed WebSocket server heartbeat timeout logic to terminate `receive` and return :py:class:`~aiohttp.ServerTimeoutError` -- by :user:`arcivanov`.
+
+  When a WebSocket pong message was not received, the
+   :py:meth:`~aiohttp.ClientWebSocketResponse.receive` operation did not terminate.
+   This change causes `_pong_not_received` to feed the `reader` an error message, causing
```
webknjaz, line 23: @bdraco

FTR single backticks in RST don't have a defined role. At some point, we'll assign the default role of “any” for linting purposes, and it will attempt to link against arbitrary objects with those names (failing when they don't exist). We should catch these in reviews and turn into double backticks to remove the ambiguity. Currently, these are rendered as italics, not inline code.

Additionally, we have to decide on the extra indents here. We don't have a previously established practice of using them in the change log.
I don't know if you noticed, but these are rendered as HTML definition lists — you'll see `<dl>`, `<dt>`, `<dd>` in the source. It seems like this is a wrong semantic markup to be used — shouldn't it be just a normal paragraph `<p>`?

bdraco: I removed the extra indent in https://github.com/aio-libs/aiohttp/pull/8598

bdraco: Added the double back ticks as well for any undefined terms.

PR #8594, opened 2024-08-02
Close stale issues without info

Dreamsorcerer:
The goal here is to allow triagers to add a label to issues when they request more info from the reporter.
If nobody provides further info, then the issue will be closed automatically, instead of being left open indefinitely.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+name: 'Close stale issues'
+on:
+  schedule:
+    - cron: '50 5 * * *'
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: 30
+          any-of-labels: needs-info
```
Dreamsorcerer, line 16: Is this the most suitable name for the label, or anybody have other suggestions?

webknjaz: I think, this is good.

webknjaz (approved):

PR #8608, opened 2024-08-05
Fix timer handle churn in websocket heartbeat

bdraco:
## What do these changes do?

Each message would cancel and reset the timer handler which meant asyncio would have to rebuild the whole schedule heap frequently because so many timer handlers were being created and than cancelled.

When asyncio timer handle cancels reach the threadhold the whole heap has to be rebuilt
https://github.com/python/cpython/blob/1422500d020bd199b26357fc387f8b79b82226cd/Lib/asyncio/base_events.py#L1968 which is extremely inefficent.

To solve this, when a timer handle is already running, instead of cancelling it, we let it fire and reschedule if heartbeat would be sent to early.

The websocket ping/pong code was missing some coverage in this area so there are some new tests as well.

no

no


aiohttp/client_ws.py
```diff
@@ -85,48 +86,66 @@ def __init__(
...
+            # Heartbeat fired too early, reschedule
+            self._heartbeat_cb = loop.call_at(
+                self._heartbeat_when, self._send_heartbeat
             )
+            return
+
+        # fire-and-forget a task is not perfect but maybe ok for
+        # sending ping. Otherwise we need a long-living heartbeat
+        # task in the class.
+        self._loop.create_task(self._writer.ping())  # type: ignore[unused-awaitable]
```
bdraco: These are at risk of being garbage collected before they finish. 

This is an existing problem to be solved in a future PR

https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

aiohttp/web_ws.py
```diff
@@ -120,50 +122,66 @@ def __init__(
...
+            # Heartbeat fired too early, reschedule
+            self._heartbeat_cb = loop.call_at(
+                self._heartbeat_when, self._send_heartbeat
             )
+            return
+
+        # fire-and-forget a task is not perfect but maybe ok for
+        # sending ping. Otherwise we need a long-living heartbeat
+        # task in the class.
+        self._loop.create_task(self._writer.ping())  # type: ignore[unused-awaitable]
```
bdraco: These are at risk of being garbage collected before they finish. 

This is an existing problem to be solved in a future PR
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

aiohttp/client_ws.py
```diff
@@ -85,48 +86,62 @@ def __init__(
...
-                    self._conn._connector._timeout_ceil_threshold
-                    if self._conn is not None
-                    else 5
-                ),
+        self._heartbeat_cb = None
+        if self._heartbeat is None or self._closed:
+            return
+        loop = self._loop
+        now = loop.time()
+        if now < self._heartbeat_when:
```
Dreamsorcerer, line 124: So, basically, instead of cancelling we just check if we're running too early and then schedule a new one?
This feels like a bit of an anti-pattern and makes me wonder if asyncio can't be improved to make cancelling lightweight. It seems very counter-intuitive to think that it's faster to run a callback than to not run it...

Dreamsorcerer: Wouldn't using heappush() rather than append() then heapify() be faster? Though probably still not significant enough...

bdraco: > This feels like a bit of an anti-pattern and makes me wonder if asyncio can't be improved to make cancelling lightweight. It seems very counter-intuitive to think that it's faster to run a callback than to not run it...

asyncio is optimized to assume timer handlers will fire. I think this makes sense since thats going to be the common case. Timer handles don't actually get get removed from the heap when they are cancelled (unless too many build up, which is the problem we have here) and instead get marked as cancelled, and when they come up they are skipped as its too expensive otherwise. When they are being canceled/scheduled frequently its almost always better to let them fire and reschedule since there is no scheduling churn for the duration of the timer.

https://github.com/python/cpython/blob/1422500d020bd199b26357fc387f8b79b82226cd/Lib/asyncio/events.py#L158 https://github.com/python/cpython/blob/1422500d020bd199b26357fc387f8b79b82226cd/Lib/asyncio/events.py#L74

A few libraries have worked around this problem by extracting the heartbeat logic and than checking when it fires https://github.com/home-assistant-libs/aioshelly/blob/57c53a90115dac719a0f8befa724f713c5a4af83/aioshelly/rpc_device/wsrpc.py#L216 but it would be much nicer to not have to do that in lots of downstream libs.

Dreamsorcerer: Yep, thinking about it more, this is part of the low-level API that most users shouldn't interact with. So, I guess we should be dealing with this limitation.

Other places that might be improved:
client_proto.py:read_timeout: We cancel and reschedule everytime data is received.
client_ws.py: Do we need to deal with _pong_response_cb too? Cancelled and rescheduled on each heartbeat send.
web_protocol.py:.keep_alive(): Should we cancel if the val is already the same and a handle already exists?

Maybe some issues in connector.py if there are many connections that are getting closed.

Also, helpers.py:TimeoutHandle is only used once in client.py and seems like something that could be replaced by asyncio.timeout(). We'd need to use asyncio.Timeout.reschedule() in the resp.connection callback.

Dreamsorcerer: > Also, helpers.py:TimeoutHandle is only used once in client.py and seems like something that could be replaced by asyncio.timeout(). We'd need to use asyncio.Timeout.reschedule() in the resp.connection callback.

.reschedule() still cancels a TimerHandle though. As that's part of the high-level API of asyncio, I wonder if it should be changed in a similar way to reduce cancellations.

bdraco: > Yep, thinking about it more, this is part of the low-level API that most users shouldn't interact with. So, I guess we should be dealing with this limitation.
> 
> Other places that might be improved: 

> client_proto.py:read_timeout: We cancel and reschedule everytime data is received. 

I think it might be worth doing. It fires far less often than websockets though. Need to do some more profiling with this one.

> client_ws.py: Do we need to deal with _pong_response_cb too? Cancelled and rescheduled on each heartbeat send. 

This is a rare case where its almost never rescheduled since we expect it to respond right away or be closed.  Its probably not worth optimizing.

> web_protocol.py:.keep_alive(): Should we cancel if the val is already the same and a handle already exists?
> 

This one doesn't seem to get rescheduled frequently when I looked at the profile.

> Maybe some issues in connector.py if there are many connections that are getting closed.
> 

cleanup_closed could use some improvements.  Something to investigate for sure.

> Also, helpers.py:TimeoutHandle is only used once in client.py and seems like something that could be replaced by asyncio.timeout(). We'd need to use asyncio.Timeout.reschedule() in the resp.connection callback.

That looks like a good opportunity to cleanup there.

bdraco:
We don't have any tests for when the heartbeat would get rescheduled.   Need to write some.

bdraco:
95.2% reduction in websocket client timer handler churn.  That's even with the downstream libs that already have their own ping/pong implementation.  

I'm happy with this.  I'll let it bake overnight and than write some tests tomorrow if I don't have any more fires from my day job to put out.

bdraco:
95.1% reduction on the other instance I tested this on

CHANGES/8608.misc.rst
```diff
@@ -0,0 +1,3 @@
+Improved websocket performance when messages are sent or received frequently -- by :user:`bdraco`.
+
+The websocket keep-alive scheduling algorithm has been improved to reduce the ``asyncio`` scheduling overhead by reducing the number of ``asyncio.TimerHandle``s being created and cancelled.
```
bdraco:
```suggestion
The WebSocket heartbeat scheduling algorithm has been improved to reduce the ``asyncio`` scheduling overhead by reducing the number of ``asyncio.TimerHandle``s being created and cancelled.
```

CHANGES/8608.misc.rst
```diff
@@ -0,0 +1,3 @@
+Improved websocket performance when messages are sent or received frequently -- by :user:`bdraco`.
+
+The WebSocket heartbeat scheduling algorithm has been improved to reduce the ``asyncio`` scheduling overhead by reducing the number of ``asyncio.TimerHandle``s being created and cancelled.
```
bdraco:
```suggestion
The WebSocket heartbeat scheduling algorithm was improved to reduce the ``asyncio`` scheduling overhead by decreasing the number of ``asyncio.TimerHandle`` creations and cancellations.
```

aiohttp/client_ws.py
bdraco: https://github.com/aio-libs/aiohttp/issues/8613

aiohttp/web_ws.py
bdraco: https://github.com/aio-libs/aiohttp/issues/8614

aiohttp/client_ws.py
bdraco: https://github.com/aio-libs/aiohttp/issues/8614

bdraco:
Unfortunately, there are already some existing uncovered lines here

bdraco:
Unfortunately, there are already some existing uncovered lines here

tests/test_client_ws_functional.py
```diff
@@ -685,6 +686,37 @@ async def handler(request):
...
+    async def handler(request):
+        nonlocal ping_received
+        ws = web.WebSocketResponse(autoping=False)
+        await ws.prepare(request)
+        for _ in range(10):
+            await ws.send_str("test")
+        msg = await ws.receive()
+        ping_received = msg.type is aiohttp.WSMsgType.PING
+        await ws.receive()
+        return ws
```
```suggestion
 ```

File: aiohttp/client_ws.py [hunk 7]
```diff
@@ -85,48 +86,62 @@ def __init__(
...
-                self._pong_heartbeat,
-                self._loop,
-                timeout_ceil_threshold=(
-                    self._conn._connector._timeout_ceil_threshold
-                    if self._conn is not None
-                    else 5
-                ),
+        self._heartbeat_cb = None
+        if self._heartbeat is None or self._closed:
+            return
```
bdraco: I think this is unreachable unless we have a bug where we set _closed and don't cancel the timer handler

aiohttp/web_ws.py
```diff
@@ -120,50 +122,63 @@ def __init__(
...
-                self._pong_not_received,
-                self._pong_heartbeat,
-                self._loop,
-                timeout_ceil_threshold=(
-                    self._req._protocol._timeout_ceil_threshold
-                    if self._req is not None
-                    else 5
-                ),
+        self._heartbeat_cb = None
+        if self._heartbeat is None or self._closed:
```
bdraco: I think this is unreachable unless we have a bug where we set _closed and don't cancel the timer handler

bdraco: Well it looks like we don't always cancel the heartbeat when we close.  I'll make it a separate function so we don't ever forget to do that, and than remove this.

aiohttp/client_ws.py
bdraco: Well it looks like we don't always cancel the heartbeat when we close.  I'll make it a separate function so we don't ever forget to do that, and than remove this.

aiohttp/client_ws.py
```diff
@@ -278,7 +298,7 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
                 await self.close()
                 return WSMessage(WSMsgType.CLOSED, None, None)
             except ClientError:
```
bdraco, line 308: It looks like `ClientError` is unreachable since `self._reader.read()` will only throw `EofStream` or `WebSocketError`

bdraco: removed in e622cde79265bb51e5c0272a059f490fddcb1b43

bdraco: Maybe it can raise `ServerDisconnectedError` when the connection is lost

bdraco: Looks so. reverted, and added coverage.

aiohttp/helpers.py
```diff
@@ -598,12 +598,23 @@ def call_later(
...
+
+def calculate_timeout_when(
+    loop_time: float,
+    timeout: float,
+    timeout_ceiling_threshold: float,
+) -> float:
+    """Calculate when to execute a timeout."""
+    when = loop_time + timeout
+    if timeout > timeout_ceiling_threshold:
+        when = ceil(when)
```
Dreamsorcerer: I feel like that's an unneeded assignment.
```suggestion
        return ceil(when)
```

Dreamsorcerer (approved):
Looks good to me. I'm about to update these test files with type checking though, so would be great if you can update after those changes.

bdraco:
Thanks!

aiohttp/client_ws.py
bdraco: The cleanup closed one is likely worth prioritizing as it tends to create a lot of timer handlers, especially with the 2s timeout.

tests/test_client_ws_functional.py
```diff
@@ -686,6 +687,39 @@ async def handler(request: web.Request) -> NoReturn:
     assert resp.close_code is WSCloseCode.ABNORMAL_CLOSURE
 
 
+async def test_heartbeat_no_pong_after_many_messages(
```
bdraco: We should have one where we receive many messages in the handler as well

aiohttp/client_ws.py
```diff
@@ -85,52 +86,64 @@ def __init__(
...
-                    self._conn._connector._timeout_ceil_threshold
-                    if self._conn is not None
-                    else 5
-                ),
+        self._heartbeat_cb = None
+        loop = self._loop
+        now = loop.time()
+        if now < self._heartbeat_when:
+            # Heartbeat fired too early, reschedule
+            self._heartbeat_cb = loop.call_at(
```
bdraco, line 126: While its the same as web_ws so we know it works ok, should still try to cover this line.

PR #8609, opened 2024-08-05
Fix slowdown in tests waiting for server to close

bdraco:
## What do these changes do?

There were a few tests that had to wait 60s to teardown because we would wait for the server to close in places we were not testing the server close logic. This added a few minutes to the test runs (18 minutes vs 5 minutes). Set the timeout to the monotonic clock resolution to reduce test time. 

no

no

### DIFF
--- a/aiohttp/test_utils.py
+++ b/aiohttp/test_utils.py
@@ -113,11 +113,7 @@ class BaseTestServer(ABC):
         if self.runner:
             return
         self._ssl = kwargs.pop("ssl", None)
-        self.runner = await self._make_runner(
-            handler_cancellation=True,
-            shutdown_timeout=MONOTONIC_CLOCK_RESOLUTION,
-            **kwargs,
-        )
+        self.runner = await self._make_runner(handler_cancellation=True, **kwargs)
         await self.runner.setup()
         if not self.port:
             self.port = 0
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
I thought the handler_cancellation=True stopped our tests from hanging. I thought the only tests that take about 60s were the autobahn tests, which have never worked as they were not finished off. Maybe I missed another regression in my recent changes though..

bdraco:
> I thought the handler_cancellation=True stopped our tests from hanging. I also thought the only tests that take about 60s were the autobahn tests, which have never worked as they were not finished off. Maybe I missed another regression in my recent changes though..

I only noticed this because `test_client_ws_functional` was hanging on quite a few tests.

On master
```
% python3 -m pytest -vvvs tests/test_client_ws_functional.py::test_close_timeout_deprecated
================================================================================ test session starts =================================================================================
platform darwin -- Python 3.12.2, pytest-8.1.1, pluggy-1.4.0 -- /Users/bdraco/aiohttp/venv/bin/python3
cachedir: .pytest_cache
rootdir: /Users/bdraco/aiohttp
configfile: setup.cfg
plugins: cov-5.0.0, mock-3.14.0, xdist-3.6.1
collected 1 item                                                                                                                                                                     

tests/test_client_ws_functional.py::test_close_timeout_deprecated[pyloop] PASSED

...hang for 60s...
============================================================================ 1 passed in 61.67s (0:01:01) ============================================================================
```


This branch
```================================================================================= 1 passed in 1.51s ==================================================================================```

Dreamsorcerer:
It was a regression in #8495, let me see if anything obvious stands out. I expected this change to only wait on _less_ things, so unclear why it'd get slower.

Dreamsorcerer:
#8611

PR #8611, opened 2024-08-05
Fix handler waiting on shutdown

Dreamsorcerer:
There seems to be some kind of race condition in certain circumstance which can lead to .wait_for_disconnection() being called after the request has already been cancelled.

Additionally, I see no guarantee that when .wait_for_disconnection() completes, that the handler itself has completed (and not suppressed a cancellation etc.).

However, I couldn't figure out a shutdown test in test_run_app.py that triggers this edge case.

### DIFF
--- a/CHANGES/8611.bugfix.rst
+++ b/CHANGES/8611.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed an edge case where shutdown would wait for timeout when handler was already completed -- by :user:`Dreamsorcerer`.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -282,7 +282,9 @@ class RequestHandler(BaseProtocol):
 
         # Wait for graceful handler completion
         if self._handler_waiter is not None:
-            await self._handler_waiter
+            with suppress(asyncio.CancelledError, asyncio.TimeoutError):
+                async with ceil_timeout(timeout):
+                    await self._handler_waiter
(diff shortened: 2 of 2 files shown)


Dreamsorcerer:
So, I'm now not convinced that wait_for_disconnection() is useful (and possibly not correct). Maybe it should be removed again...?

bdraco (approved):

bdraco:
> So, I'm now not convinced that wait_for_disconnection() is useful (and possibly not correct). Maybe it should be removed again...?

I'm not sure its useful, but I've also never had a use case for it.  I'd open an issue to provide notice of intent to remove it, and than do so in 30 days if nobody objects.

Dreamsorcerer:
> I'm not sure its useful, but I've also never had a use case for it. I'd open an issue to provide notice of intent to remove it, and than do so in 30 days if nobody objects.

Well, it's only been in a release for a week or so. So, it's incredibly unlikely people are already using it, we can probably just strip in the next 3.10 release before most users have finished upgrading to 3.10.

bdraco:
> > I'm not sure its useful, but I've also never had a use case for it. I'd open an issue to provide notice of intent to remove it, and than do so in 30 days if nobody objects.
> 
> Well, it's only been in a release for a week or so. So, it's incredibly unlikely people are already using it, we can probably just strip in the next 3.10 release before most users have finished upgrading to 3.10.

I agree. No need to wait as its unlikely its being used.

PR #8620, opened 2024-08-06
Test typing round 2

Dreamsorcerer:
Another round of typing all the test files.

I'll check over the diff tomorrow.


aiohttp/typedefs.py
```diff
@@ -35,7 +35,13 @@
 Byteish = Union[bytes, bytearray, memoryview]
 JSONEncoder = Callable[[Any], str]
 JSONDecoder = Callable[[str], Any]
-LooseHeaders = Union[Mapping[Union[str, istr], str], _CIMultiDict, _CIMultiDictProxy]
+LooseHeaders = Union[
+    Mapping[str, str],
+    Mapping[istr, str],
```
bdraco, line 40: This is a nice fix. Might want to backport this one

bdraco (approved):

aiohttp/typedefs.py
Dreamsorcerer: Yeah, I'll backport the non-test changes.

PR #8632, opened 2024-08-07
Fix connecting to npipe://, tcp://, and unix:// urls

### LINKED ISSUE #8600: Error while trying to use unix socket, when `aiohttp`>=3.10
## Long story short

* Expected behaviour: Communicating via unix socket should work.
* Actual behaviour: Error. Its traceback:
```
  File "/tmp/aiodocker_fail/example.py", line 12, in <module>
    asyncio.run(main())
  File "/usr/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/asyncio/runners.py", line 118, in run
  File "/usr/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/example.py", line 8, in main
    await docker.version()
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 190, in version
    data = await self._query_json("version")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 319, in _query_json
    async with self._query(
  File "/usr/lib/python3.11/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 231, in _query
    yield await self._do_query(
          ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 257, in _do_query
    await self._check_version()
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 209, in _check_version
    ver = await self._query_json("version", versioned_api=False)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 319, in _query_json
    async with self._query(
  File "/usr/lib/python3.11/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 231, in _query
    yield await self._do_query(
          ^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiodocker/docker.py", line 267, in _do_query
    response = await self.session.request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/aiodocker_fail/venv/lib/python3.11/site-packages/aiohttp/client.py", line 510, in _request
aiohttp.client_exceptions.NonHttpUrlClientError: unix://localhost/version
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f5ede16bf10>
```
## How to reproduce

In fresh virtual environment run `pip install aiodocker`, the following script fails:

```python
#!/usr/bin/env python3
from aiodocker import Docker
import asyncio

async def main():
    docker = Docker()
    await docker.version()

if __name__ == "__main__":
    asyncio.run(main())
```
## Your environment

Just the latest version of aiodocker with combination of 3.10 aiohttp.

My complete `requirements.txt` of the environment:
```txt
aiodocker==0.22.2
aiohappyeyeballs==2.3.4
aiohttp==3.10.0
aiosignal==1.3.1
attrs==24.1.0
frozenlist==1.4.1
idna==3.7
multidict==6.0.5
yarl==1.9.4
```

Current fix for me is the usage of `aiohttp==3.9.0`

bdraco:
## What do these changes do?

Fixes a regression in 3.10.x connecting to `npipe://`, `tcp://`, and `unix://` urls

A future PR should remove `allowed_protocol_schema_set` in `BaseConnector` as its only there for backwards compat

no

no

## Related issue number
 fixes #8600

### DIFF
--- a/CHANGES/8632.bugfix.rst
+++ b/CHANGES/8632.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed connecting to ``unix://`` urls -- by :user:`bdraco`.
--- a/CHANGES/8632.bugfix.rst
+++ b/CHANGES/8632.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed connecting to ``unix://`` urls -- by :user:`bdraco`.
+Fixed connecting to ``npipe://`` and ``unix://`` urls -- by :user:`bdraco`.
--- a/CHANGES/8632.bugfix.rst
+++ b/CHANGES/8632.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed connecting to ``npipe://`` and ``unix://`` urls -- by :user:`bdraco`.
+Fixed connecting to ``npipe://``, ``tcp://``, and ``unix://`` urls -- by :user:`bdraco`.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -80,11 +80,16 @@ from .client_ws import (
     ClientWebSocketResponse,
     ClientWSTimeout,
 )
-from .connector import BaseConnector, NamedPipeConnector, TCPConnector, UnixConnector
+from .connector import (
+    HTTP_AND_EMPTY_SCHEMA_SET,
+    BaseConnector,
+    NamedPipeConnector,
+    TCPConnector,
+    UnixConnector,
+)
 from .cookiejar import CookieJar
 from .helpers import (
     _SENTINEL,
-    HTTP_SCHEMA_SET,
     BasicAuth,
     TimeoutHandle,
     ceil_timeout,
@@ -681,7 +686,7 @@ class ClientSession:
                             ) from e
 
                         scheme = parsed_redirect_url.scheme
-                        if scheme not in HTTP_SCHEMA_SET:
+                        if scheme not in HTTP_AND_EMPTY_SCHEMA_SET:
                             resp.close()
                             raise NonHttpUrlRedirectClientError(r_url)
                         elif not scheme:
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -84,6 +84,7 @@ from .connector import BaseConnector, NamedPipeConnector, TCPConnector, UnixConn
 from .cookiejar import CookieJar
 from .helpers import (
     _SENTINEL,
+    HTTP_SCHEMA_SET,
     BasicAuth,
     TimeoutHandle,
     ceil_timeout,
@@ -210,10 +211,7 @@ DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 
 # https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
-HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
-WS_SCHEMA_SET = frozenset({"ws", "wss"})
-PROTOCOL_SCHEMA_SET = frozenset({"npipe", "unix"})
-ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET | PROTOCOL_SCHEMA_SET
+
 
 _RetType = TypeVar("_RetType")
 _CharsetResolver = Callable[[ClientResponse, bytes], str]
@@ -456,7 +454,7 @@ class ClientSession:
         except ValueError as e:
             raise InvalidUrlClientError(str_or_url) from e
 
-        if url.scheme not in ALLOWED_PROTOCOL_SCHEMA_SET:
+        if url.scheme not in self._connector.allowed_protocol_schema_set:
             raise NonHttpUrlClientError(url)
 
         skip_headers = set(self._skip_auto_headers)
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -217,7 +217,6 @@ DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 # https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
 
-
 _RetType = TypeVar("_RetType")
 _CharsetResolver = Callable[[ClientResponse, bytes], str]
 
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -212,7 +212,7 @@ DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
 HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
 WS_SCHEMA_SET = frozenset({"ws", "wss"})
-PROTOCOL_SCHEMA_SET = frozenset({"unix"})
+PROTOCOL_SCHEMA_SET = frozenset({"npipe", "unix"})
 ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET | PROTOCOL_SCHEMA_SET
 
 _RetType = TypeVar("_RetType")
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -212,7 +212,8 @@ DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60)
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
 HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
 WS_SCHEMA_SET = frozenset({"ws", "wss"})
-ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET
+PROTOCOL_SCHEMA_SET = frozenset({"unix"})
+ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET | PROTOCOL_SCHEMA_SET
 
 _RetType = TypeVar("_RetType")
 _CharsetResolver = Callable[[ClientResponse, bytes], str]
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -459,6 +459,7 @@ class ClientSession:
         except ValueError as e:
             raise InvalidUrlClientError(str_or_url) from e
 
+        assert self._connector is not None
         if url.scheme not in self._connector.allowed_protocol_schema_set:
             raise NonHttpUrlClientError(url)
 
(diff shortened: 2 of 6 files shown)


aiohttp/client.py
```diff
@@ -212,7 +212,8 @@ class ClientTimeout:
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
 HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
 WS_SCHEMA_SET = frozenset({"ws", "wss"})
-ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET
+PROTOCOL_SCHEMA_SET = frozenset({"unix"})
```
bdraco: Ideally this would only be when using the `UnixConnector` but thats a breaking change, and the goal for the moment is to fix the regression.

https://github.com/aio-libs/aiohttp/issues/8633 will track moving this to be per connector

Dreamsorcerer: This will still leave NamedPipeConnector broken, right? I think we should delay a couple of days if needed to get this defined properly in the connectors.

bdraco:
Looks like we do need npipe as well https://github.com/aio-libs/aiodocker/blob/288d2168263364ad98ed7fd7bcf1a44d86cb98d9/aiodocker/docker.py#L96

CHANGES/8632.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed connecting to ``unix://`` urls -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed connecting to ``npipe://`` and ``unix://`` urls -- by :user:`bdraco`.
```

aiohttp/connector.py
```diff
@@ -243,6 +255,14 @@ def __init__(
         self._cleanup_closed_transports: List[Optional[asyncio.Transport]] = []
         self._cleanup_closed()
 
+    @cached_property
+    def allowed_protocol_schema_set(self) -> frozenset[str]:
+        """Return allowed protocol schema set.
+
+        By default we allow all protocols.
```
```suggestion
        By default we allow all protocols for backwards compatibility.
```

aiohttp/connector.py
```diff
@@ -1357,6 +1383,13 @@ def __init__(
         )
         self._path = path
 
+    @cached_property
+    def allowed_protocol_schema_set(self) -> frozenset[str]:
+        """Return allowed protocol schema set."""
+        return (
+            EMPTY_SCHEMA_SET | HTTP_SCHEMA_SET | WS_SCHEMA_SET | UNIX_PROTCOL_SCHEMA_SET
```
Dreamsorcerer: Shouldn't UnixConnector only work with unix:?

bdraco: There are specific tests to allow http etc over unix in connector

aiohttp/connector.py
```diff
@@ -243,6 +256,14 @@ def __init__(
         self._cleanup_closed_transports: List[Optional[asyncio.Transport]] = []
         self._cleanup_closed()
 
+    @cached_property
+    def allowed_protocol_schema_set(self) -> frozenset[str]:
+        """Return allowed protocol schema set.
+
+        By default we allow all protocols for backwards compatibility.
+        """
+        return ALLOWED_PROTOCOL_SCHEMA_SET
```
Dreamsorcerer: I was thinking about this last night, and maybe the BaseConnector should just have something like:
```
class AllProtocols:
    def __contains__(self, item: object) -> Literal[True]:
        return True
```

We also suggested making this abstract in future, but I think it'd be fine to leave this as allowing anything. Custom connectors can then choose to restrict the allowed set if they choose too.

aiohttp/connector.py
bdraco: https://github.com/aio-libs/aiohttp/blob/c99a1e27375285149ea82cbdcc2f2c40e57596dc/tests/test_connector.py#L2419

https://github.com/aio-libs/aiohttp/blob/c99a1e27375285149ea82cbdcc2f2c40e57596dc/tests/test_connector.py#L2265

aiohttp/connector.py
bdraco: Maybe this is fine as-is and we don't need to worry about removing it long term and can close https://github.com/aio-libs/aiohttp/issues/8633?

CHANGES/8632.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed connecting to ``npipe://`` and ``unix://`` urls -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed connecting to ``npipe://``, ``tcp://``, and ``unix://`` urls -- by :user:`bdraco`.
```

aiohttp/connector.py
Dreamsorcerer: As you mentioned, this could still be a breaking change if someone has a custom connector that handles other protocols. But, maybe it's fine to just default to http/ws...

Dreamsorcerer: But, defaulting to something that allows unix/npipe doesn't make sense to me, as there is unlikely to be any connector that supports all of these..

Dreamsorcerer: So, I'd vote for either:

1. Allow anything and then tighten in subclasses.
2. Allow http/ws and then expand in subclasses.

aiohttp/connector.py
```diff
@@ -63,6 +65,23 @@
     SSLContext = object  # type: ignore[misc,assignment]
 
 
+EMPTY_SCHEMA_SET = frozenset({""})
```
Dreamsorcerer, line 67: I'm not really following what the purpose of the empty set is?

bdraco: Its existing. The change it to break it out into its own constant.

I think its to allow connecting to "example.com/x/y/z"

aiohttp/connector.py
bdraco: I adjusted it to `BASE_PROTOCOL_SCHEMA_SET`

I think we should merge as-is for backport and than do another PR with a breaking change note to drop 
```
    | TCP_PROTOCOL_SCHEMA_SET
    | UNIX_PROTOCOL_SCHEMA_SET
    | NAMED_PIPE_PROTOCOL_SCHEMA_SET
```

from `BASE_PROTOCOL_SCHEMA_SET` that would only target 4.0

Dreamsorcerer: OK, I don't see what difference it makes personally. If a custom connector supports another scheme, then this is still a breaking change. If a custom connector supports one of these schemes, then they probably inherited from one of the subclasses, not BaseConnector, and are unaffected by this change regardless.

bdraco: Let's drop it now than

aiohttp/client.py
```diff
@@ -210,9 +216,7 @@ class ClientTimeout:
 
 # https://www.rfc-editor.org/rfc/rfc9110#section-9.2.2
 IDEMPOTENT_METHODS = frozenset({"GET", "HEAD", "OPTIONS", "TRACE", "PUT", "DELETE"})
-HTTP_SCHEMA_SET = frozenset({"http", "https", ""})
-WS_SCHEMA_SET = frozenset({"ws", "wss"})
-ALLOWED_PROTOCOL_SCHEMA_SET = HTTP_SCHEMA_SET | WS_SCHEMA_SET
+
```
Dreamsorcerer, line 219:
```suggestion
```

aiohttp/connector.py
```diff
@@ -244,6 +257,14 @@ def __init__(
         self._cleanup_closed_transports: List[Optional[asyncio.Transport]] = []
         self._cleanup_closed()
 
+    @cached_property
+    def allowed_protocol_schema_set(self) -> FrozenSet[str]:
+        """Return allowed protocol schema set.
+
+        By default we allow all protocols for backwards compatibility.
```
```suggestion
        By default we allow all high level protocols for backwards compatibility.
```

aiohttp/connector.py
```diff
@@ -791,6 +812,11 @@ def _close_immediately(self) -> List["asyncio.Future[None]"]:
             ev.cancel()
         return super()._close_immediately()
 
+    @cached_property
+    def allowed_protocol_schema_set(self) -> FrozenSet[str]:
+        """Return allowed protocol schema set."""
+        return super().allowed_protocol_schema_set | TCP_PROTOCOL_SCHEMA_SET
+
```
Dreamsorcerer: Personally, I'd have kept it to a simple:
```
allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET.union("tcp")
```

I'm not convinced on the single use globals or cached_property.

tests/test_client_session.py
```diff
@@ -536,15 +536,17 @@ async def test_ws_connect_allowed_protocols(
...
     }
-    resp.url = URL(f"{protocol}://example.com")
+    resp.url = URL(f"{protocol}://example")
     resp.cookies = SimpleCookie()
     resp.start = mock.AsyncMock()
 
     req = mock.create_autospec(aiohttp.ClientRequest, spec_set=True)
     req_factory = mock.Mock(return_value=req)
     req.send = mock.AsyncMock(return_value=resp)
+    # BaseConnector allows all protocols by default
```
bdraco:
```suggestion
    # BaseConnector allows all high level protocols by default
```

tests/test_client_session.py
```diff
@@ -536,15 +536,78 @@ async def test_ws_connect_allowed_protocols(
...
+        hdrs.SEC_WEBSOCKET_ACCEPT: ws_key,
+    }
+    resp.url = URL(f"{protocol}://example")
+    resp.cookies = SimpleCookie()
+    resp.start = mock.AsyncMock()
+
+    req = mock.create_autospec(aiohttp.ClientRequest, spec_set=True)
+    req_factory = mock.Mock(return_value=req)
+    req.send = mock.AsyncMock(return_value=resp)
+    # UnixConnector allows all protocols by default and unix sockets
```
bdraco:
```suggestion
    # UnixConnector allows all high level protocols by default and unix sockets
```

aiohttp/connector.py
bdraco: It did end up more complex than needed.

adjusted in 374821b2f847e250e7f2e69874401ad6e80c8c1a

Dreamsorcerer:
I think this is good.

I'd note that a frozenset() isn't actually needed as the second operator to result in a frozenset. The `.union("tcp")` syntax also allows to skip creating an extra set entirely. I don't really mind though, just thought I'd point it out.

Dreamsorcerer (approved):

bdraco:
> I think this is good.
> 
> I'd note that a frozenset() isn't actually needed as the second operator to result in a frozenset. The `.union("tcp")` syntax also allows to skip creating an extra set entirely. I don't really mind though, just thought I'd point it out.

I tried that first but it iterated each character in the string

bdraco:
Will fix the type when I get the next site. Have to drive now.

webknjaz:
@bdraco @Dreamsorcerer it would be worth also checking if connecting to abstract UNIX sockets works (it's a Linux-only feature).

bdraco:
The original PR only added validation for the schema part of the URL so I think we should be ok with the abstract Unix sockets

PR #8636, opened 2024-08-07
Remove Request.wait_for_disconnection() method

### LINKED ISSUE #8631: Remove wait_for_disconnection
This issue is for tracking purposes only so we don't forget

              > I'm not sure its useful, but I've also never had a use case for it. I'd open an issue to provide notice of intent to remove it, and than do so in 30 days if nobody objects.

Well, it's only been in a release for a week or so. So, it's incredibly unlikely people are already using it, we can probably just strip in the next 3.10 release before most users have finished upgrading to 3.10.

_Originally posted by @Dreamsorcerer in https://github.com/aio-libs/aiohttp/issues/8611#issuecomment-2273408467_

Dreamsorcerer:
Fixes #8631.

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -630,6 +630,7 @@ class RequestHandler(BaseProtocol):
         can get exception information. Returns True if the client disconnects
         prematurely.
         """
+        request._finish()
         if self._request_parser is not None:
             self._request_parser.set_upgraded(False)
             self._upgrade = False
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -630,7 +630,6 @@ class RequestHandler(BaseProtocol):
         can get exception information. Returns True if the client disconnects
         prematurely.
         """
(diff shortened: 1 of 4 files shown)


aiohttp/web_protocol.py
```diff
@@ -630,7 +630,6 @@ async def finish_response(
         can get exception information. Returns True if the client disconnects
         prematurely.
         """
-        request._finish()
```
Dreamsorcerer: @webknjaz Would be good if you can sanity check whether this method still makes sense. It was added only for this wait_for_disconnection() method, but then appears you added some multipart fix into the method. I'm wondering if that fix should live somewhere else, or we should keep the method.

bdraco (approved):

aiohttp/web_protocol.py
Dreamsorcerer: I assume your fix was never backported either: #5494.

bdraco (approved):

PR #8641, opened 2024-08-08
Fix WebSocket ping tasks being prematurely garbage collected

### LINKED ISSUE #8614: Risk of websocket ping tasks being garbage collected before they finish
These are at risk of being garbage collected before they finish. 

This is an existing problem to be solved in a future PR
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
> Important Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done. For reliable “fire-and-forget” background tasks, gather them in a collection:

_Originally posted by @bdraco in https://github.com/aio-libs/aiohttp/pull/8608#discussion_r1704039542_

bdraco:
## What do these changes do?

The event loop only keeps weak references to tasks, we need to hold a strong reference to ensure that the ping task is not prematurely garbage collected.
https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task

In almost all cases the ping can be done synchronously if the task is created eagerly which avoids scheduling the ping task on the event loop.

no

no
## Related issue number

fixes #8614

### DIFF
--- a/CHANGES/8641.bugfix.rst
+++ b/CHANGES/8641.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed WebSocket ping tasks being prematurely garbage collected -- by :user:`bdraco`.
+
+There was a small risk that WebSocket ping tasks would be prematurely garbage collected because the event loop only holds a weak reference to the task. The garbage collection risk has been fixed by holding a strong reference to the task. Additionally, the task is not scheduled eagerly with Python 3.12+ to increase the chance it can be completed immediately and avoid having to hold any references to the task.
--- a/CHANGES/8641.bugfix.rst
+++ b/CHANGES/8641.bugfix.rst
@@ -1,3 +1,3 @@
 Fixed WebSocket ping tasks being prematurely garbage collected -- by :user:`bdraco`.
 
-There was a small risk that WebSocket ping tasks would be prematurely garbage collected because the event loop only holds a weak reference to the task. The garbage collection risk has been fixed by holding a strong reference to the task. Additionally, the task is not scheduled eagerly with Python 3.12+ to increase the chance it can be completed immediately and avoid having to hold any references to the task.
+There was a small risk that WebSocket ping tasks would be prematurely garbage collected because the event loop only holds a weak reference to the task. The garbage collection risk has been fixed by holding a strong reference to the task. Additionally, the task is now scheduled eagerly with Python 3.12+ to increase the chance it can be completed immediately and avoid having to hold any references to the task.
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -146,7 +146,7 @@ class ClientWebSocketResponse:
         else:
             self._ping_task.add_done_callback(self._ping_task_done)
 
-    def _ping_task_done(self, task: asyncio.Task[None]) -> None:
+    def _ping_task_done(self, task: "asyncio.Task[None]") -> None:
         """Callback for when the ping task completes."""
         self._ping_task = None
 
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -146,7 +146,7 @@ class ClientWebSocketResponse:
             # the task on the event loop.
             ping_task = asyncio.Task(self._writer.ping(), loop=loop, eager_start=True)
         else:
-            ping_task = self.loop.create_task(self._writer.ping())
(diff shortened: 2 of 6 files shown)


aiohttp/helpers.py
```diff
@@ -592,6 +593,21 @@ def weakref_handle(
...
 
+def create_eager_task(
+    coro: Coroutine[Any, Any, None],
+    loop: asyncio.AbstractEventLoop,
+) -> "asyncio.Task[None]":
+    """Create a task that will be scheduled immediately if possible."""
+    if sys.version_info >= (3, 12):
+        # Optimization for Python 3.12, try to write
+        # bytes immediately to avoid having to schedule
+        # the task on the event loop.
```
bdraco:
```suggestion
        # Optimization for Python 3.12, try start eagerly
        # to avoid being scheduled on the event loop.
```

aiohttp/helpers.py
```diff
@@ -592,6 +593,20 @@ def weakref_handle(
     return None
 
 
+def create_eager_task(
+    coro: Coroutine[Any, Any, None],
+    loop: asyncio.AbstractEventLoop,
+) -> "asyncio.Task[None]":
+    """Create a task that will be scheduled immediately if possible."""
+    if sys.version_info >= (3, 12):
+        # Optimization for Python 3.12, try start eagerly
```
bdraco:
```suggestion
        # Optimization for Python 3.12+, try start eagerly
```

aiohttp/helpers.py
```diff
@@ -592,6 +593,20 @@ def weakref_handle(
     return None
 
 
+def create_eager_task(
+    coro: Coroutine[Any, Any, None],
+    loop: asyncio.AbstractEventLoop,
+) -> "asyncio.Task[None]":
+    """Create a task that will be scheduled immediately if possible."""
```
bdraco:
```suggestion
    """Create a task that will be run immediately if possible."""
```

CHANGES/8641.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed WebSocket ping tasks being prematurely garbage collected -- by :user:`bdraco`.
+
+There was a small risk that WebSocket ping tasks would be prematurely garbage collected because the event loop only holds a weak reference to the task. The garbage collection risk has been fixed by holding a strong reference to the task. Additionally, the task is not scheduled eagerly with Python 3.12+ to increase the chance it can be completed immediately and avoid having to hold any references to the task.
```
bdraco:
```suggestion
There was a small risk that WebSocket ping tasks would be prematurely garbage collected because the event loop only holds a weak reference to the task. The garbage collection risk has been fixed by holding a strong reference to the task. Additionally, the task is now scheduled eagerly with Python 3.12+ to increase the chance it can be completed immediately and avoid having to hold any references to the task.
```

aiohttp/client_ws.py
```diff
@@ -141,6 +140,15 @@ def _send_heartbeat(self) -> None:
         self._cancel_pong_response_cb()
         self._pong_response_cb = loop.call_at(when, self._pong_not_received)
 
+        ping_task = create_eager_task(self._writer.ping(), loop)
```
bdraco: ping task now happens at the end to better align with previous implementation since it previously would run one event loop iteration later on older python.

aiohttp/helpers.py
```diff
@@ -592,6 +593,20 @@ def weakref_handle(
     return None
 
 
+def create_eager_task(
```
Dreamsorcerer: Personally, I'd rather duplicate such code through the codebase, it makes cleaning it up in future easier.
e.g. If we just run pyupgrade or whatever on this, it's likely to result in a pointless-looking wrapper:
```
def create_eager_task(coro, loop):
    return asyncio.Task(coro, loop=loop, eager_start=True)
```

Whereas, running on duplicated code, we'd just end up with the expected code. For that same reason, I'd only add a comment explaining why we want eager_start. The availability of the feature on certain Python versions is obvious from the code itself.

Dreamsorcerer (approved):

aiohttp/helpers.py
bdraco: I'll undo the de-duplication.  I think we only have 4-5 potential places where this makes sense anyways.

PR #8655, opened 2024-08-08
Release 3.10.2

bdraco:
https://github.com/aio-libs/aiohttp/actions/runs/10306656747


bdraco:
Retested on two production systems just to be sure. All good

PR #8657, opened 2024-08-08
Fix multipart reading with split boundary

### LINKED ISSUE #8643: Multipart reader fails on Transfer-Encoding: chunked
### Describe the bug

The multipart reader will not parse if `Transfer-Encoding` is `chunked`.
The reader merges two different parts into one larger multipart.

This request-dump is also required (125 MB; github limits to 25 MB)
http://doppelbauer.name/request-data.txt

### To Reproduce

The test case is a bit clumsy, but reproduces the problem.

````python
import asyncio, base64, hashlib
from aiohttp.multipart import MultipartReader

headers = {
    'Content-Type': 'multipart/form-data; boundary=f9nvedV9b8roNRgHQEYH75IQSkw6UCOJsJSIzCXWhVAXCwPSMrs73NjPQTH5rKARGsWAFa',
}

# http://doppelbauer.name/request-data.txt
fp = open('request-data.txt')
class FakeReader:
    async def read(self, n=-1):
        kind, data = fp.readline().split('\t')
        assert kind == 'read'
        return base64.b64decode( data.encode() )

    async def readline(self, separator=b"\n"):
        kind, data = fp.readline().split('\t')
        assert kind == 'readuntil'
        return base64.b64decode( data.encode() )

    def at_eof(self):
        return False

    def unread_data(self, data):
        pass

async def test_multipart():
    multipart = MultipartReader( headers, FakeReader() )
    part_index = 0
    async for part in multipart:
        part_index += 1
        body = b''
        while True:
            chunk = await part.read_chunk()
            if not chunk:
                break
            body += chunk
        blake2 = hashlib.blake2b( body ).hexdigest()

        if int(part.headers['Content-Length']) == len(body) and part.headers['Content-BLAKE2'] == blake2:
            continue

        # Huh!? Boundary within body
        assert b'f9nvedV9b8roNRgHQEYH75IQSkw6UCOJsJSIzCXWhVAXCwPSMrs73NjPQTH5rKARGsWAFa' in body
        print( 'Multipart', '#' + str(part_index), part.filename )
        print( 'Size invalid:', len(body), 'instead of:', int(part.headers['Content-Length']) )
        print( 'Hash:', blake2, part.headers['Content-BLAKE2'],  )
        print()

asyncio.ensure_future(test_multipart())
loop.run_forever()
loop.close()
````

### Expected behavior

Correctly parse multipart

### Logs/tracebacks

```python-traceback
#
```

### Python Version

```console
python 3.12
```

### aiohttp Version

```console
Version: 3.10.1
```

### multidict Version

```console
$
```

### yarl Version

```console
$
```

### OS

Linux

### Related component

Server

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #8643.

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -266,6 +266,7 @@ class BodyPartReader:
     ) -> None:
         self.headers = headers
         self._boundary = boundary
+        self._boundary_len = len(boundary) + 2  # Boundary + \r\n
         self._content = content
         self._default_charset = default_charset
         self._at_eof = False
@@ -371,15 +372,25 @@ class BodyPartReader:
         # Reads content chunk of body part with unknown length.
         # The Content-Length header for body part is not necessary.
         assert (
-            size >= len(self._boundary) + 2
+            size >= self._boundary_len
         ), "Chunk size must be greater or equal than boundary length + 2"
         first_chunk = self._prev_chunk is None
         if first_chunk:
             self._prev_chunk = await self._content.read(size)
 
-        chunk = await self._content.read(size)
-        self._content_eof += int(self._content.at_eof())
-        assert self._content_eof < 3, "Reading after EOF"
+        chunk = b""
+        # content.read() may return less than size, so we need to loop to ensure
+        # we have enough data to detect the boundary.
+        while len(chunk) < self._boundary_len:
+            chunk += await self._content.read(size)
+            self._content_eof += int(self._content.at_eof())
+            assert self._content_eof < 3, "Reading after EOF"
+            if self._content_eof:
+                break
+        if len(chunk) > size:
+            self._content.unread_data(chunk[size:])
+            chunk = chunk[:size]
+
         assert self._prev_chunk is not None
         window = self._prev_chunk + chunk
         sub = b"\r\n" + self._boundary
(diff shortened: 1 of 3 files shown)


aiohttp/multipart.py
```diff
@@ -371,15 +372,25 @@ async def _read_chunk_from_stream(self, size: int) -> bytes:
         # Reads content chunk of body part with unknown length.
         # The Content-Length header for body part is not necessary.
         assert (
-            size >= len(self._boundary) + 2
+            size >= self._boundary_len
         ), "Chunk size must be greater or equal than boundary length + 2"
```
bdraco, line 376: stale assertion message

Dreamsorcerer: It's still accurate.

tests/test_multipart.py
```diff
@@ -657,6 +657,66 @@ async def test_invalid_boundary(self) -> None:
             with pytest.raises(ValueError):
                 await reader.next()
 
+    @pytest.mark.skipif(sys.version_info < (3, 10), reason="Needs anext()")
```
Dreamsorcerer, line 661: For keeping this test relatively simple, I'm just going to skip 3.8/9.

aiohttp/multipart.py
bdraco: Ahh my bad.. I missed the change above.

bdraco (approved):
LGTM

PR #8660, opened 2024-08-08
Improve performance of WebSockets when there is no timeout

bdraco:
## What do these changes do?

The common case for WebSockets is to call receive without a timeout, or use the WebSocketResponse as an async iterator. The read was always wrapped with an async timeout even if the timeout was not used which accounted for as much as 50% of the run time each loop. If there is no timeout, avoid wrapping the read call in the context manager which creates a new Timeout() object each time.

Performance improvement

related reading https://www.researchgate.net/publication/348993267_An_Analysis_of_the_Performance_of_Websockets_in_Various_Programming_Languages_and_Libraries

no

before
timeout

after
after

### DIFF
--- a/CHANGES/8660.misc.rst
+++ b/CHANGES/8660.misc.rst
@@ -1,3 +1,3 @@
-Improved performance of :py:meth:`~aiohttp.ClientWebSocketResponse.receive` and :py:meth:`~aiohttp.web.WebSocketResponse.receive` when there is timeout. -- by :user:`bdraco`.
+Improved performance of :py:meth:`~aiohttp.ClientWebSocketResponse.receive` and :py:meth:`~aiohttp.web.WebSocketResponse.receive` when there is no timeout. -- by :user:`bdraco`.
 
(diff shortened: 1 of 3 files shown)


bdraco:
This should significantly help `aiounifi`, `aioshelly`, and `uiprotect` which they are have quite a bit of incoming websocket traffic.

Dreamsorcerer:
I wonder if there's a way to improve the timeout version too...

Dreamsorcerer (approved):

bdraco:
> I wonder if there's a way to improve the timeout version too...

Nothing obvious stands out in https://github.com/python/cpython/blob/2f5c3b09e45798a18d60841d04a165fb062be666/Lib/asyncio/timeouts.py#L85 but I also haven't stared at it long enough.

bdraco:
Thanks!

bdraco:
> I wonder if there's a way to improve the timeout version too...

Found a way to make it faster https://github.com/python/cpython/pull/122882

PR #8661, opened 2024-08-08
Improve performance of starting request handlers with Python 3.12+

bdraco:
## What do these changes do?

#3406 wrapped _handle_request in a task to copy the contextvars which delays the start of the request by one iteration of the event loop and increases event loop overhead by forcing it to be scheduled which many of the handlers can finish synchronously.

In https://github.com/aio-libs/aiohttp/discussions/8170#discussioncomment-8802876 we did not have a way to improve with without removing the task, however since Python 3.12+ has eager start for tasks, we can now avoid the delay and scheduling overhead.

Performance improvement

no

### DIFF
--- a/CHANGES/8661.misc.rst
+++ b/CHANGES/8661.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of starting request handlers with Python 3.12+ -- by :user:`bdraco`.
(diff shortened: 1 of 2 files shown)


Dreamsorcerer (approved):

bdraco:
Thanks

Issue #8613, opened 2024-08-06
Analyze additional places where there is timer handle churn


bdraco:
> Yep, thinking about it more, this is part of the low-level API that most users shouldn't interact with. So, I guess we should be dealing with this limitation.
> 
> Other places that might be improved: 

> client_proto.py:read_timeout: We cancel and reschedule everytime data is received. 

I think it might be worth doing. It fires far less often than websockets though. Need to do some more profiling with this one.

> client_ws.py: Do we need to deal with _pong_response_cb too? Cancelled and rescheduled on each heartbeat send. 

This is a rare case where its almost never rescheduled since we expect it to respond right away or be closed.  Its probably not worth optimizing.

> web_protocol.py:.keep_alive(): Should we cancel if the val is already the same and a handle already exists?
> 

This one doesn't seem to get rescheduled frequently when I looked at the profile.

> Maybe some issues in connector.py if there are many connections that are getting closed.
> 

cleanup_closed could use some improvements.  Something to investigate for sure.

> Also, helpers.py:TimeoutHandle is only used once in client.py and seems like something that could be replaced by asyncio.timeout(). We'd need to use asyncio.Timeout.reschedule() in the resp.connection callback.

That looks like a good opportunity to cleanup there.

_Originally posted by @bdraco in https://github.com/aio-libs/aiohttp/pull/8608#discussion_r1704175784_

Dreamsorcerer:
> > web_protocol.py:.keep_alive(): Should we cancel if the val is already the same and a handle already exists?
> 
> This one doesn't seem to get rescheduled frequently when I looked at the profile.

For this one, I just figured it'd be a one-line fix to add an if statement in, which could reduce the impact even if it's not hit often.

bdraco:
> cleanup_closed could use some improvements. Something to investigate for sure.
> 
> >

This one is heavy. Should make it a priority.

bdraco:
The cleanup_closed case is because its every 2.0.  HA changed it to 60.0 but there are lot of places where libs don't use HA's subclassed connector so we still see it there.  2.0s might be too frequent, but I'm not sure how safe it is to increase for the generate use case.

bdraco:
> > > web_protocol.py:.keep_alive(): Should we cancel if the val is already the same and a handle already exists?
> > 
> > 
> > This one doesn't seem to get rescheduled frequently when I looked at the profile.
> 
[...]

PR #8662, opened 2024-08-08
Improve performance of keepalive rescheduling

bdraco:
## What do these changes do?

If not all handlers were done, the keep alive would get rescheduled every second until the current time was greater than self._keepalive_time + self._keepalive_timeout

Instead we will reschedule the timer for the expected keep alive close time if the timer handle fires too early.

no

no

related issue #8613

Before we would see 1 timer handle created every second for 60s per connection. After we get 1 timer handle total for when the keepalive timeout is due.

before
before

after
~~after~~

after_take_2

### DIFF
--- a/CHANGES/8662.misc.rst
+++ b/CHANGES/8662.misc.rst
@@ -1,0 +1,3 @@
+Reduce frequency of HTTP keep-alive checks -- by :user:`bdraco`.
+
+Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check now occurs when the keep-alive is expected to timeout or 1s later, whichever is later.
--- a/CHANGES/8662.misc.rst
+++ b/CHANGES/8662.misc.rst
@@ -1,3 +1,3 @@
 Improved performance of HTTP keep-alive checks -- by :user:`bdraco`.
 
-Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check now occurs when the keep-alive is expected to timeout or 1s later, whichever is later.
+Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check is now rescheduled if it fires too early instead.
--- a/CHANGES/8662.misc.rst
+++ b/CHANGES/8662.misc.rst
@@ -1,3 +1,3 @@
-Reduce frequency of HTTP keep-alive checks -- by :user:`bdraco`.
+Improved performance of HTTP keep-alive checks -- by :user:`bdraco`.
 
 Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check now occurs when the keep-alive is expected to timeout or 1s later, whichever is later.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -443,6 +443,7 @@ class RequestHandler(BaseProtocol):
         self.logger.exception(*args, **kw)
 
     def _process_keepalive(self) -> None:
+        self._keepalive_handle = None
         if self._force_close or not self._keepalive:
             return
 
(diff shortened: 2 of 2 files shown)


bdraco:
Before we would see 1 timer handle created every second for 60s per connection.  After we get 1 timer handle total.

bdraco:
I went to do some archeology before I mark this ready as the previous code didn't make much sense to me.

bdraco:
It looks like the 1s delay was added in https://github.com/aio-libs/aiohttp/commit/ad2366e31c0b05905082bd7b22340eb186128be0

Previously it was `next` but it looks like that caused a problem for websocket but thats since been removed.  It also looks like it caused a problem where it might fire too often, but thats not a problem with this implementation because we use `max(next, 1)`

CHANGES/8662.misc.rst
```diff
@@ -0,0 +1,3 @@
+Reduce frequency of HTTP keep-alive checks -- by :user:`bdraco`.
```
bdraco:
```suggestion
Improved performance of HTTP keep-alive checks -- by :user:`bdraco`.
```

aiohttp/web_protocol.py
```diff
@@ -449,17 +449,19 @@ def _process_keepalive(self) -> None:
...
-                self.force_close()
-                return
+        if self._waiter and now > next:
+            self.force_close()
+            return
 
         # not all request handlers are done,
-        # reschedule itself to next second
-        self._keepalive_handle = self._loop.call_later(
-            self.KEEPALIVE_RESCHEDULE_DELAY,
```
Dreamsorcerer, line 462: Ironically, this was introduced to fix a CPU spike: #2701.

I wonder if we can improve this further and avoid the 1 second reschedules entirely? From what I can tell, the 1 second rechecks will be happening while a request is being processed (i.e. we are not waiting for a new message, so there is no _waiter).

Can't we just reschedule once the request handler has finished processing? It appears to me the waiter is only set in start(), and there's no way _keepalive_time won't get updated just before creating another waiter:
https://github.com/aio-libs/aiohttp/blob/4d604eafefbee22ad35886c3bd12e4c3048cc45b/aiohttp/web_protocol.py#L612

Therefore, it seems to me that if we reach this point and `now > next`, then there's no point in rescheduling here, we can just exit and let it schedule again after updating the timeout.

bdraco: I'll adjust this a bit to reschedule if it fires too early and avoid the 1s reschedules.

aiohttp/web_protocol.py
```diff
@@ -445,23 +443,21 @@ def log_exception(self, *args: Any, **kw: Any) -> None:
...
+        self._keepalive_handle = None
         if self._force_close or not self._keepalive:
             return
 
-        next = self._keepalive_time + self._keepalive_timeout
+        loop = self._loop
+        now = loop.time()
+        close_time = self._next_keepalive_close_time
+        if now <= close_time:
+            # Keep alive fired close fired too early, reschedule
```
```suggestion
            # Keep alive close check fired too early, reschedule
```

CHANGES/8662.misc.rst
```diff
@@ -0,0 +1,3 @@
+Improved performance of HTTP keep-alive checks -- by :user:`bdraco`.
+
+Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check now occurs when the keep-alive is expected to timeout or 1s later, whichever is later.
```
bdraco:
```suggestion
Previously, when processing a request for a keep-alive connection, the keep-alive check would happen every second; the check is now rescheduled if it fires too early instead.
```

Dreamsorcerer (approved):

bdraco:
Thanks. I like the new version much better.

PR #8667, opened 2024-08-09
Improve performance of generating random WebSocket mask

bdraco:
## What do these changes do?

The mask was being generated with randint and than converted to bytes. We can skip a few steps and generate it with getrandbits() and pack it into bytes.

related RFC: https://datatracker.ietf.org/doc/html/rfc6455#section-5.3

No

no

I was looking at pings and noticed that most of the time is spent in randrange
Screenshot 2024-08-08 at 9 13 27 PM

### DIFF
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -8,6 +8,7 @@ import re
 import sys
 import zlib
 from enum import IntEnum
+from functools import partial
 from struct import Struct
 from typing import (
     Any,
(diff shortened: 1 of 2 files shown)


aiohttp/http_websocket.py
```diff
@@ -652,8 +654,7 @@ async def _send_frame(
         else:
             header = PACK_LEN3(0x80 | rsv | opcode, 127 | mask_bit, msg_length)
         if use_mask:
-            mask_int = self.randrange(0, 0xFFFFFFFF)
-            mask = mask_int.to_bytes(4, "big")
+            mask = PACK_RANDBITS(self.get_random_bits())
```
bdraco, line 657: pack is faster than `to_bytes`. Its already used above in a lot of places but this one wasn't for some reason.

aiohttp/http_websocket.py
```diff
@@ -596,7 +598,7 @@ def __init__(
         self.protocol = protocol
         self.transport = transport
         self.use_mask = use_mask
-        self.randrange = random.randrange
+        self.get_random_bits = partial(random.getrandbits, 32)
```
bdraco, line 601: Masking key is a 32-bit value https://datatracker.ietf.org/doc/html/rfc6455#section-5.3

Dreamsorcerer (approved):

bdraco:
thanks

Issue #3080, opened 2018-06-13
Frequent asyncio delays larger than 100ms


dimitar-petrov:
## Long story short
I am noticing asyncio event loop delays. And this is the stack trace of the handle when delay is detected:
```
[<FrameSummary file <string>, line 1 in <module>>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/multiprocessing/spawn.py, line 105 in spawn_main>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/multiprocessing/spawn.py, line 118 in _main>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/multiprocessing/process.py, line 258 in _bootstrap>,
 <FrameSummary file /foopath/3.6.3/envs/pyaio/lib/python3.6/site-packages/foo/bar.py, line 249 in run>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/base_events.py, line 454 in run_until_complete>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/base_events.py, line 421 in run_forever>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/base_events.py, line 1418 in _run_once>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/events.py, line 127 in _run>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/selector_events.py, line 731 in _read_ready>,
 <FrameSummary file /foopath/3.6.3/lib/python3.6/asyncio/sslproto.py, line 516 in data_received>,
 <FrameSummary file /foopath/3.6.3/envs/pyaio/lib/python3.6/site-packages/aiohttp/client_proto.py, line 211 in data_received>,
 <FrameSummary file /foopath/3.6.3/envs/pyaio/lib/python3.6/site-packages/aiohttp/streams.py, line 518 in feed_data>,
 <FrameSummary file /foopath/3.6.3/envs/pyaio/lib/python3.6/site-packages/aiohttp/helpers.py, line 745 in set_result>]
Executing <Handle <TaskWakeupMethWrapper object at 0x7f5edf78ca08>(<Future finis...events.py:275>) created at /foopath/3.6.3/envs/pyaio/lib/python3.6/site-packages/aiohttp/helpers.py:745> too
k 0.120 seconds

```

[...]

PR #8672, opened 2024-08-09
Fix TCPConnector doing blocking I/O in the event loop to create the SSLContext

bdraco:
## What do these changes do?

Create the SSLContext in the executor since it does blocking I/O to read the certificates from disk.

related issue #3080

no

no

### DIFF
--- a/CHANGES/8672.bugfix.rst
+++ b/CHANGES/8672.bugfix.rst
@@ -1,3 +1,3 @@
-Fix TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
+Fix :py:class:`aiohttp.TCPConnector` doing blocking I/O in the event loop to create the ``SSLContext`` -- by :user:`bdraco`.
 
-The blocking I/O would only happen once per verify mode. However, it could cause the event loop to block for a long time if the SSLContext creation is slow, which is more likely during startup when the disk cache is not yet present.
+The blocking I/O would only happen once per verify mode. However, it could cause the event loop to block for a long time if the ``SSLContext`` creation is slow, which is more likely during startup when the disk cache is not yet present.
--- a/CHANGES/8672.bugfix.rst
+++ b/CHANGES/8672.bugfix.rst
@@ -1,3 +1,3 @@
-Fix TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
+Fixed TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
 
 The blocking I/O would only happen once per verify mode. However, it could cause the event loop to block for a long time if the SSLContext creation is slow, which is more likely during startup when the disk cache is not yet present.
--- a/CHANGES/8672.bugfix.rst
+++ b/CHANGES/8672.bugfix.rst
@@ -1,0 +1,3 @@
+Fix TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
+
+The blocking I/O would only happen once per verify mode. However, it could cause the event loop to block for a long time if the SSLContext creation is slow, which is more likely during startup when the disk cache is not yet present.
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -752,7 +752,8 @@ class TCPConnector(BaseConnector):
     """
 
     allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET | frozenset({"tcp"})
-    _made_ssl_context: Dict[bool, asyncio.Lock] = {}
+    _made_ssl_context: Set[bool] = set()
+    _make_ssl_context_lock: Dict[bool, asyncio.Lock] = {}
 
     def __init__(
         self,
@@ -1012,10 +1013,11 @@ class TCPConnector(BaseConnector):
             return self._make_ssl_context(verified)
         # _make_ssl_context does blocking I/O to load certificates
         # from disk, so we run it in a separate thread.
-        async with self._made_ssl_context.setdefault(verified, asyncio.Lock()):
+        async with self._make_ssl_context_lock.setdefault(verified, asyncio.Lock()):
             context = await self._loop.run_in_executor(
                 None, self._make_ssl_context, verified
             )
+            self._made_ssl_context.add(verified)
         return context
 
     def _get_fingerprint(self, req: ClientRequest) -> Optional["Fingerprint"]:
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -752,6 +752,7 @@ class TCPConnector(BaseConnector):
     """
 
     allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET | frozenset({"tcp"})
+    _made_ssl_context: Set[bool] = set()
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


aiohttp/connector.py
```diff
@@ -982,25 +988,40 @@ def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
...
+            return self._make_ssl_context(verified)
+        # _make_ssl_context does blocking I/O to load certificates
+        # from disk, so we run it in a separate thread.
+        if not self._make_ssl_context_lock:
+            self._make_ssl_context_lock = asyncio.Lock()
+        async with self._make_ssl_context_lock:
+            context = await self._loop.run_in_executor(
+                None, self._make_ssl_context, verified
+            )
+            self._made_ssl_context.add(verified)
```
bdraco: There could be a thundering heard to create these at startup time. The executor job should be fast in most cases but without a lock there is a race here where connectors could create a flury of executor jobs.

aiohttp/connector.py
```diff
@@ -982,25 +987,37 @@ def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
...
+    async def _make_or_get_cached_ssl_context(self, verified: bool) -> SSLContext:
+        """Create or get cached SSL context."""
+        if verified in self._made_ssl_context:
+            return self._make_ssl_context(verified)
+        # _make_ssl_context does blocking I/O to load certificates
+        # from disk, so we run it in a separate thread.
+        async with self._made_ssl_context.setdefault(verified, asyncio.Lock()):
+            context = await self._loop.run_in_executor(
+                None, self._make_ssl_context, verified
+            )
```
bdraco: There is still a race here because the executor job could still be running.

aiohttp/connector.py
```diff
@@ -948,27 +950,31 @@ async def _create_connection(
...
+        sslcontext.verify_mode = ssl.CERT_NONE
+        try:
+            sslcontext.options |= ssl.OP_NO_COMPRESSION
+        except AttributeError as attr_err:
+            warnings.warn(
+                "{!s}: The Python interpreter is compiled "
+                "against OpenSSL < 1.0.0. Ref: "
+                "https://docs.python.org/3/library/ssl.html"
+                "#ssl.OP_NO_COMPRESSION".format(attr_err),
+            )
```
bdraco: https://docs.python.org/3/library/ssl.html
Changed in version 3.6: OpenSSL 0.9.8, 1.0.0 and 1.0.1 are deprecated and no longer supported. In the future the ssl module will require at least OpenSSL 1.0.2 or 1.1.0.

I think we can drop this

CHANGES/8672.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed TCPConnector doing blocking I/O in the event loop to create the SSLContext -- by :user:`bdraco`.
```

aiohttp/connector.py
```diff
@@ -982,25 +980,38 @@ def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
...
+        if verified in self._made_ssl_context:
+            return self._make_ssl_context(verified)
+        # _make_ssl_context does blocking I/O to load certificates
+        # from disk, so we run it in a separate thread.
+        async with self._make_ssl_context_lock.setdefault(verified, asyncio.Lock()):
+            context = await self._loop.run_in_executor(
+                None, self._make_ssl_context, verified
+            )
+            self._made_ssl_context.add(verified)
+        return context
```
Dreamsorcerer: Would it be a bit cleaner to do something along the lines of:

```suggestion
        try:
            future = self._make_ssl_context[verified]
        except KeyError:
            future = asyncio.Future()
            self._make_ssl_context[verified] = future
            try:
                result = await self._loop.run_in_executor(
                    None, self._make_ssl_context
                )
            except Exception as e:
                future.set_exception(e)
                self._make_ssl_context.pop(verified)
            else:
                future.set_result(result)
        return await future
```

Then the other function no longer needs to have a cache.

Dreamsorcerer: I think the original code could result in many executor calls just to run a cached function. It also just feels a little odd having a dict to check if the function is already cached, when you could instead store the result in there.

bdraco: I think we would need to worry about cancellation with this approach.... maybe we need to check for BaseException.  Need to think on this

bdraco:
Needs tests for

- [ ] `_make_ssl_context` raises due to `OSError` or similar
- [ ] `_make_or_get_ssl_context` handles cancellation

aiohttp/connector.py
```diff
@@ -752,6 +759,7 @@ class TCPConnector(BaseConnector):
     """
 
     allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET | frozenset({"tcp"})
+    _made_ssl_context: Dict[bool, asyncio.Lock] = {}
```
```suggestion
    _made_ssl_context: Dict[bool, SSLContext] = {}
```

aiohttp/connector.py
```diff
@@ -752,6 +759,7 @@ class TCPConnector(BaseConnector):
     """
 
     allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET | frozenset({"tcp"})
+    _made_ssl_context: Dict[bool, SSLContext] = {}
```
```suggestion
    _made_ssl_context: Dict[bool, asyncio.Future[SSLContext]] = {}
```

aiohttp/connector.py
```diff
@@ -752,6 +759,7 @@ class TCPConnector(BaseConnector):
     """
 
     allowed_protocol_schema_set = HIGH_LEVEL_SCHEMA_SET | frozenset({"tcp"})
+    _made_ssl_context: Dict[bool, asyncio.Future[SSLContext]] = {}
```
```suggestion
    _made_ssl_context: Dict[bool, "asyncio.Future[SSLContext]"] = {}
```

Dreamsorcerer (approved):

tests/test_connector.py
```diff
@@ -1709,67 +1709,109 @@ async def test_tcp_connector_clear_dns_cache_bad_args(
...
+
+@pytest.mark.parametrize("exception", [OSError, ssl.SSLError, asyncio.CancelledError])
+async def test_ssl_context_creation_raises(exception: BaseException) -> None:
+    """Test that we try again if SSLContext creation fails the first time."""
+    conn = aiohttp.TCPConnector()
+    conn._made_ssl_context.clear()
+
+    with mock.patch.object(
+        conn, "_make_ssl_context", side_effect=exception
+    ), pytest.raises(  # type: ignore[call-overload]
```
Dreamsorcerer, line 1809: Commit says this is missing BaseException on the overload?
Code appears to be correct: https://github.com/pytest-dev/pytest/blame/main/src/_pytest/python_api.py#L778

bdraco: It sure does look right there.

Maybe we have an old version?

I can't find the CI run so I started another one https://github.com/aio-libs/aiohttp/pull/8676

bdraco: https://github.com/aio-libs/aiohttp/actions/runs/10332753102/job/28604325639?pr=8676

bdraco: Just a silly mistake on my part. I forgot it needed to be wrapped in `type[]`

PR #8676, opened 2024-08-10
Fix type ignore in SSLContext creation connector test

bdraco:
Fixes the typing in `test_ssl_context_creation_raises`. This is not a runtime change.

https://github.com/aio-libs/aiohttp/pull/8672#discussion_r1712663038


Dreamsorcerer:
With a little tweak to the pytest parametrize API that could probably be caught as a type error in the function signature... 🤔️

PR #8680, opened 2024-08-11
Fix close race that prevented the close code from reaching the client

### LINKED ISSUE #3448: WebSocketResponse.close code/message don't reach the client
## Long story short

Following this example from the docs
https://docs.aiohttp.org/en/stable/web_advanced.html?highlight=going_away#graceful-shutdown

The code and message passed to `ws.close` don't reach the client side and instead, a default code and message is used.

## Expected behaviour

I would expect that the code and message should reach the client.
If I'm wrong, the documentation could explain why this is not the case.

## Actual behaviour

Default code and message are returned (`code=1000, message=b''`)

## Steps to reproduce

The commented-out asserts fail.

```python
import weakref

import aiohttp.http
import aiohttp.web


async def websocket_handler(request):
    websocket = aiohttp.web.WebSocketResponse()
    await websocket.prepare(request)
    request.app["websockets"].add(websocket)

    try:
        async for message in websocket:
            await websocket.send_json({"ok": True, "message": message.json()})
    finally:
        request.app["websockets"].discard(websocket)

    return websocket


async def on_shutdown(app):
    for websocket in set(app["websockets"]):
        await websocket.close(
            code=aiohttp.WSCloseCode.GOING_AWAY,
            message="Server shutdown",
        )


async def test_foo(aiohttp_client):
    url = "/ws"
    app["websockets"] = weakref.WeakSet()
    app.router.add_get(url, websocket_handler)
    app.on_shutdown.append(on_shutdown)

    websocket = await client.ws_connect(url)

    message = {"message": "hi"}
    await websocket.send_json(message)
    reply = await websocket.receive_json()
    assert reply == {"ok": True, "message": message}

    await app.shutdown()

    assert websocket.closed is False

    reply = await websocket.receive()

    assert reply.type == aiohttp.http.WSMsgType.CLOSE
    assert reply.data == aiohttp.WSCloseCode.OK
    assert reply.extra == ""
    # assert reply.data == aiohttp.WSCloseCode.GOING_AWAY
    # assert reply.extra == "Server shutdown"
    assert websocket.closed is True
```

## Your environment

Python 3.6.7

```
aiohttp==3.4.4
pytest==4.0.1
pytest-aiohttp==0.3.0
```

bdraco:
Fix close race that prevented the close code from reaching the client

fixes #3448

### DIFF
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -454,11 +454,9 @@ class WebSocketResponse(StreamResponse):
         # we need to break `receive()` cycle before we can call
         # `reader.read()` as `close()` may be called from different task
         if self._waiting:
-            assert (
-                self._close_wait is None
-            ), "Concurrent call to receive() is not allowed"
-            assert self._loop is not None
-            self._close_wait = self._loop.create_future()
+            if not self._close_wait:
+                assert self._loop is not None
+                self._close_wait = self._loop.create_future()
             reader.feed_data(WS_CLOSING_MESSAGE)
             await self._close_wait
 
--- a/aiohttp/web_ws.py
+++ b/aiohttp/web_ws.py
@@ -453,14 +453,13 @@ class WebSocketResponse(StreamResponse):
             self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
             return True
 
-        # we need to break `receive()` cycle before we
-        # can call `reader.read()` as
-        # `close()` may be called from different task
+        # we need to break `receive()` cycle before we can call
+        # `reader.read()` as `close()` may be called from different task
         if self._waiting:
             if not self._close_wait:
                 assert self._loop is not None
                 self._close_wait = self._loop.create_future()
-                reader.feed_data(WS_CLOSING_MESSAGE)
+            reader.feed_data(WS_CLOSING_MESSAGE)
             await self._close_wait
 
         if self._closing:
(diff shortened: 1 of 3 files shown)


Dreamsorcerer (approved):
Looks like that works.

bdraco:
need to do regression testing with Home Assistant and make sure everything is still covered

aiohttp/web_ws.py
```diff
@@ -462,6 +453,15 @@ async def close(
             self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
             return True
 
+        # we need to break `receive()` cycle before we can call
+        # `reader.read()` as `close()` may be called from different task
+        if self._waiting:
+            if not self._close_wait:
```
bdraco: now we have a branch miss on this line.

Need to think about how to cover it

bdraco: This one is going to be hard to cover since the race window is very small here

bdraco: Its is truely unreachable now since `self._closed = True` is set right away and there is no race there anymore

aiohttp/web_ws.py
```diff
@@ -431,23 +431,10 @@ async def close(
         if self._writer is None:
             raise RuntimeError("Call .prepare() first")
 
-        self._cancel_heartbeat()
```
Dreamsorcerer, line 434: Should this line have gone? Just double checking.

bdraco: Yes, `_set_closed()` handles it already https://github.com/aio-libs/aiohttp/blob/bbe71972b4452cd45ed94f5546722a31860a5dda/aiohttp/web_ws.py#L209

PR #8681, opened 2024-08-11
Improve performance of starting request handlers with Python 3.12+

bdraco:
## What do these changes do?

Improve performance of starting request handlers with Python 3.12+ by starting the task eagerly.

followup to #8661 as I missed one more place we create a task

no

no

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -510,7 +510,7 @@ class RequestHandler(BaseProtocol):
         keep_alive(True) specified.
         """
         loop = self._loop
-        handler = self._task_handler
+        handler = asyncio.current_task(loop)
         assert handler is not None
         manager = self._manager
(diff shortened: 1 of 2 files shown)


aiohttp/web_protocol.py
```diff
@@ -308,10 +308,16 @@ def connection_made(self, transport: asyncio.BaseTransport) -> None:
         if self._tcp_keepalive:
             tcp_keepalive(real_transport)
 
-        self._task_handler = self._loop.create_task(self.start())
         assert self._manager is not None
         self._manager.connection_made(self, real_transport)
 
+        loop = self._loop
```
bdraco, line 314: Moved to after `connection_made` since it will begin to execute immediately to keep the same semantics since previously may have expected the connection to always be there before the task began to execute.

PR #7238, opened 2023-03-22
consume ConnectionResetError in ping tasks

altendky:
## What do these changes do?

The purpose is to avoid uncaught exception tracebacks from ping tasks.

```python-traceback
Task exception was never retrieved
future: <Task finished name='Task-5559' coro=<WebSocketWriter.ping() done, defined at /home/ubuntu/chia-blockchain/venv/lib/python3.10/site-packages/aiohttp/http_websocket.py:672> exception=ConnectionResetError('Cannot write to closing transport')>
Traceback (most recent call last):
  File "/home/ubuntu/chia-blockchain/venv/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 676, in ping
    await self._send_frame(message, WSMsgType.PING)
  File "/home/ubuntu/chia-blockchain/venv/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 646, in _send_frame
    self._write(header + mask + message)
  File "/home/ubuntu/chia-blockchain/venv/lib/python3.10/site-packages/aiohttp/http_websocket.py", line 663, in _write
    raise ConnectionResetError("Cannot write to closing transport")
ConnectionResetError: Cannot write to closing transport
```

## Are there changes in behavior for the user?

The user both does not deal with the tracebacks and also gets the option to request consumption themselves.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/5182

## Checklist

- [ ] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names.
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #8685, opened 2024-08-12
Fix exceptions from WebSocket ping task not being consumed

### LINKED ISSUE #2309: WebSocketResponse does not throw/close when connection is abruptly cut

### LINKED ISSUE #4153: ClientWebSocket heartbeat should throw exceptions
## Long story short

A client websocket session with a heartbeat enabled will never indicate the connection is closed if a program is only receiving information.

This may be related to #2309 (which was server-side, but others appear to be commenting about client-side), where @asvetlov [stated](https://github.com/aio-libs/aiohttp/issues/2309#issuecomment-418641220):
> Workaround (the proper solution actually) is sending ping messages from a client.
> Unfortunately, browsers don't support WebSocket pings, you have to emulate them by small regular messages.

My understanding is the heartbeat is intended to create the ping message to detect the disconnection. Therefore, when someone is using the heartbeat, an exception should be raised when pong not received.
 
## Expected behaviour

The [Timeout exception](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_ws.py#L110) should be raised upon detecting that `_pong_not_received` for calling programs to catch.

## Actual behaviour

The connection closes silently. The only way to detect it would be to add additional logic to send a ping, which is the purpose of the heartbeat.

## Steps to reproduce

1. Create a receive only websocket connection to any server and enable a heartbeat
```python
async with session.ws_connect('http://example.org/ws', heartbeat=5) as ws:
    async for msg in ws:
        _LOGGER.debug("msg: %s", msg)
```
2. Simulate a disconnect by disconnecting internet.
3. Websocket messages will stop because the connection is closed. However, this is silent.

## Your environment

Client
OS isn't relevant

### LINKED ISSUE #5182: Unhandled exception for ping task
🐞 **Describe the bug**

Websocket handler with enabled `autoping` option does not handle sudden client disconnection right.  

When client calls `receive()` method, handler internally creates task with `ping()` call. If the client disconnects without sending close message, task will not be cancelled which leads us to the `ConnectionResetError` exception. `asyncio` logs it with `Task exception was never retrieved`, which is correct, but in this case annoying.

💡 **To Reproduce**

Consider following server/client example.
```python
# server.py
from aiohttp import web


async def handler(request):
    ws_response = web.WebSocketResponse(heartbeat=2)
    await ws_response.prepare(request)
    await ws_response.receive()
    return ws_response


app = web.Application()
app.add_routes([
    web.get('/ws', handler)
])

web.run_app(app, host='localhost')
```

```python
# client.py
import asyncio
import aiohttp


async def client_intruder():
        await session._ws_connect('http://localhost:8080/ws')
        # Look at me! I'm not closing my websocket connection >:P

asyncio.run(client_intruder(), debug=True)
```

💡 **Expected behavior**

Exception should be handled. We can't require `WSMsgType.CLOSE` message from client in all cases, because its disconnection may be caused by network failure.

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
future: <Task finished name='Task-6' coro=<WebSocketWriter.ping() done, defined at /home/s/python/aiohttp/aiohttp/http_websocket.py:660> exception=ConnectionResetError('Cannot write to closing transport') created at /home/s/python/aiohttp/aiohttp/web_ws.py:136>
source_traceback: Object created at (most recent call last):
  File "server.py", line 17, in <module>
    web.run_app(app, host='localhost', debug=True)
  File "/home/s/python/aiohttp/aiohttp/web.py", line 514, in run_app
    loop.run_until_complete(main_task)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 603, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 1851, in _run_once
    handle._run()
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/home/s/python/aiohttp/aiohttp/web_ws.py", line 136, in _send_heartbeat
    self._loop.create_task(self._writer.ping())  # type: ignore
  File "/home/s/python/aiohttp/aiohttp/http_websocket.py", line 664, in ping
    await self._send_frame(message, WSMsgType.PING)
  File "/home/s/python/aiohttp/aiohttp/http_websocket.py", line 641, in _send_frame
    self._write(header + message)
  File "/home/s/python/aiohttp/aiohttp/http_websocket.py", line 651, in _write
ConnectionResetError: Cannot write to closing transport
```

```console
$ python --version
Python 3.8.6
```

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 4.0.0a1
Author: Nikolay Kim
License: Apache 2
Location: /home/s/python/aiohttp
Requires: attrs, chardet, multidict, async-timeout, yarl, typing-extensions
Required-by:
```
```console
Name: multidict
Version: 5.0.0
Author: Andrew Svetlov
License: Apache 2
Location: /home/s/python/aiohttp/.direnv/python-3.8.6/lib/python3.8/site-packages
Requires: 
Required-by: yarl, aiohttp
```
```console
$ python -m pip show yarl
Name: yarl
Version: 1.6.1
Author: Andrew Svetlov
License: Apache 2
Location: /home/s/python/aiohttp/.direnv/python-3.8.6/lib/python3.8/site-packages
Requires: multidict, idna
Required-by: aiohttp
```

📋 **Additional context**

Looks like this problem affects `3.7+` versions of aiohttp.

bdraco:
## What do these changes do?

Fix exceptions from WebSocket ping task not being consumed

No more unexpected un-retrieved exceptions from ping tasks

no
## Related issue number
related issue https://github.com/home-assistant/core/issues/123653
replaces and closes #7238
fixes #5182
fixes #4153
fixes #2309

### DIFF
--- a/tests/test_client_ws_functional.py
+++ b/tests/test_client_ws_functional.py
@@ -679,6 +679,7 @@ async def test_heartbeat_connection_closed(aiohttp_client: AiohttpClient) -> Non
     # We patch write here to simulate a connection reset error
     # since if we closed the connection normally, the server would
     # would cancel the heartbeat task and we wouldn't get a ping
+    assert resp._conn is not None
     with mock.patch.object(
         resp._conn.transport, "write", side_effect=ConnectionResetError
     ), mock.patch.object(resp._writer, "ping", wraps=resp._writer.ping) as ping:
--- a/tests/test_client_ws_functional.py
+++ b/tests/test_client_ws_functional.py
@@ -661,6 +661,35 @@ async def test_heartbeat(aiohttp_client: AiohttpClient) -> None:
     assert ping_received
 
 
+async def test_heartbeat_connection_closed(aiohttp_client: AiohttpClient) -> None:
+    """Test that the connection is closed while ping is in progress."""
+
+    async def handler(request: web.Request) -> NoReturn:
+        ws = web.WebSocketResponse(autoping=False)
+        await ws.prepare(request)
+        await ws.receive()
+        assert False
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+
+    client = await aiohttp_client(app)
+    resp = await client.ws_connect("/", heartbeat=0.1)
+    ping_count = 0
+    # We patch write here to simulate a connection reset error
+    # since if we closed the connection normally, the server would
+    # would cancel the heartbeat task and we wouldn't get a ping
+    with mock.patch.object(
+        resp._conn.transport, "write", side_effect=ConnectionResetError
+    ), mock.patch.object(resp._writer, "ping", wraps=resp._writer.ping) as ping:
+        await resp.receive()
+        ping_count = ping.call_count
+    # Connection should be closed roughly after 1.5x heartbeat.
+    await asyncio.sleep(0.2)
+    assert ping_count == 1
+    assert resp.close_code is WSCloseCode.ABNORMAL_CLOSURE
+
+
 async def test_heartbeat_no_pong(aiohttp_client: AiohttpClient) -> None:
     """Test that the connection is closed if no pong is received without sending messages."""
     ping_received = False
--- a/tests/test_client_ws_functional.py
+++ b/tests/test_client_ws_functional.py
@@ -677,7 +677,7 @@ async def test_heartbeat_connection_closed(aiohttp_client: AiohttpClient) -> Non
     resp = await client.ws_connect("/", heartbeat=0.1)
     ping_count = 0
     # We patch write here to simulate a connection reset error
-    # since if we closed the connection normally, the server would
+    # since if we closed the connection normally, the client would
     # would cancel the heartbeat task and we wouldn't get a ping
     with mock.patch.object(
         resp._conn.transport, "write", side_effect=ConnectionResetError
--- a/CHANGES/8685.bugfix.rst
+++ b/CHANGES/8685.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed unconsumed exceptions raised by the WebSocket heartbeat -- by :user:`bdraco`.
+
+If the heartbeat ping raised an exception, it would not be consumed and would be logged as an warning.
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -157,19 +157,22 @@ class ClientWebSocketResponse:
     def _ping_task_done(self, task: "asyncio.Task[None]") -> None:
         """Callback for when the ping task completes."""
         if not task.cancelled() and (exc := task.exception()):
-            self._reader.feed_data(WSMessage(WSMsgType.ERROR, exc, None))
+            self._handle_ping_pong_exception(exc)
         self._ping_task = None
 
     def _pong_not_received(self) -> None:
-        if not self._closed:
-            self._set_closed()
-            self._close_code = WSCloseCode.ABNORMAL_CLOSURE
-            self._exception = ServerTimeoutError()
-            self._response.close()
-            if self._waiting and not self._closing:
-                self._reader.feed_data(
-                    WSMessage(WSMsgType.ERROR, self._exception, None)
-                )
+        self._handle_ping_pong_exception(ServerTimeoutError())
+
+    def _handle_ping_pong_exception(self, exc: BaseException) -> None:
+        """Handle exceptions raised during ping/pong processing."""
+        if self._closed:
+            return
+        self._set_closed()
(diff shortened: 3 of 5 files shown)


tests/test_client_ws_functional.py
```diff
@@ -661,6 +661,35 @@ async def handler(request: web.Request) -> NoReturn:
...
+        assert False
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+
+    client = await aiohttp_client(app)
+    resp = await client.ws_connect("/", heartbeat=0.1)
+    ping_count = 0
+    # We patch write here to simulate a connection reset error
+    # since if we closed the connection normally, the server would
```
```suggestion
    # since if we closed the connection normally, the client would
```

PR #3867, opened 2019-06-25
Fix decoding base64 chunk for BodyPartReader

hyzyla:
## What do these changes do?

Those changes add the additional reading of few bytes in `BodyPartReader.read_chunk` if the size of the chunk is not valid for base64 decoding.  

## Are there changes in behavior for the user?

Yes, because if someone will try to use `read_chunk` for the body that has transfer encoding equal to base64, can receive more than expected bytes.

## Related issue number

https://github.com/aio-libs/aiohttp/issues/3843

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [ ] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #8688, opened 2024-08-12
Fix decoding base64 chunk for BodyPartReader (#3867)

Dreamsorcerer:
(cherry picked from commit aeb01ce3bda7497d7cc6c434ac6806aca291e273)

### DIFF
--- a/aiohttp/multipart.py
+++ b/aiohttp/multipart.py
@@ -323,6 +323,31 @@ class BodyPartReader:
         else:
             chunk = await self._read_chunk_from_stream(size)
 
+        # For the case of base64 data, we must read a fragment of size with a
+        # remainder of 0 by dividing by 4 for string without symbols \n or \r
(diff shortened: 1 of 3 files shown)


aiohttp/multipart.py
```diff
@@ -323,6 +323,31 @@ async def read_chunk(self, size: int = chunk_size) -> bytes:
...
+
+                if self._prev_chunk:
+                    over_chunk = self._prev_chunk[:over_chunk_size]
+                    self._prev_chunk = self._prev_chunk[len(over_chunk) :]
+
+                if len(over_chunk) != over_chunk_size:
+                    over_chunk += await self._content.read(4 - len(over_chunk))
+
+                if not over_chunk:
+                    self._at_eof = True
```
Dreamsorcerer, line 345: @hyzyla If you get a chance, it'd great to have another test that covers this code. These last 2 if statements are never true in the tests, and the first one is never false.

PR #8690, opened 2024-08-12
Test cythonizing the websocket frame parser

bdraco:
DO NOT MERGE: for testing only


bdraco:
The tests pass locally but I'm not sure this is a good idea.

PR #8693, opened 2024-08-13
Fix missing logic in quickstart example

### LINKED ISSUE #3780: document bug in quickstart section session that always return None
https://docs.aiohttp.org/en/latest/web_quickstart.html#user-sessions
```
import asyncio
import time
import base64
from cryptography import fernet
from aiohttp import web
from aiohttp_session import setup, get_session, session_middleware
from aiohttp_session.cookie_storage import EncryptedCookieStorage

async def handler(request):
    session = await get_session(request)
    last_visit = session['last_visit'] if 'last_visit' in session else None
    text = 'Last visited: {}'.format(last_visit)
    return web.Response(text=text)

async def make_app():
    app = web.Application()
    # secret_key must be 32 url-safe base64-encoded bytes
    fernet_key = fernet.Fernet.generate_key()
    secret_key = base64.urlsafe_b64decode(fernet_key)
    setup(app, EncryptedCookieStorage(secret_key))
    app.add_routes([web.get('/', handler)])
    return app

web.run_app(make_app())
```

this'll always return `None`. problem is:
 `last_visit = session['last_visit'] if 'last_visit' in session else None`

considering fix? I'm able to PR

NewUserHa:
## What do these changes do?
an easy fix for https://docs.aiohttp.org/en/latest/web_quickstart.html#user-sessions

n/a

n/a

## Related issue number
Fixes https://github.com/aio-libs/aiohttp/issues/3780

## Checklist

### DIFF
--- a/docs/web_quickstart.rst
+++ b/docs/web_quickstart.rst
@@ -451,11 +451,8 @@ third-party library, :mod:`aiohttp_session`, that adds *session* support::
     async def handler(request):
         session = await get_session(request)
 
-        if "last_visit" in session:
-            last_visit = session["last_visit"]
-        else:
-            last_visit = None
-            session["last_visit"] = time.time()
+        last_visit = session.get("last_visit")
+        session["last_visit"] = time.time()
         text = "Last visited: {}".format(last_visit)
 
         return web.Response(text=text)
--- a/docs/web_quickstart.rst
+++ b/docs/web_quickstart.rst
@@ -450,8 +450,14 @@ third-party library, :mod:`aiohttp_session`, that adds *session* support::
 
     async def handler(request):
         session = await get_session(request)
-        last_visit = session['last_visit'] if 'last_visit' in session else None
-        text = 'Last visited: {}'.format(last_visit)
+
+        if "last_visit" in session:
+            last_visit = session["last_visit"]
+        else:
+            last_visit = None
+            session["last_visit"] = time.time()
+        text = "Last visited: {}".format(last_visit)
+
         return web.Response(text=text)
 
     async def make_app():


docs/web_quickstart.rst
```diff
@@ -450,8 +450,14 @@ third-party library, :mod:`aiohttp_session`, that adds *session* support::
...
     async def handler(request):
         session = await get_session(request)
-        last_visit = session['last_visit'] if 'last_visit' in session else None
-        text = 'Last visited: {}'.format(last_visit)
+
+        if "last_visit" in session:
+            last_visit = session["last_visit"]
+        else:
+            last_visit = None
+            session["last_visit"] = time.time()
```
Dreamsorcerer, line 458: Maybe this would be more concise?
```suggestion
        last_visit = session.get("last_visit")
        if last_visit is None:
            session["last_visit"] = time.time()
```

Dreamsorcerer: Wait, don't we want to update last_visit everytime, otherwise it wouldn't be the _last_ visit...

```suggestion
        last_visit = session.get("last_visit")
        session["last_visit"] = time.time()
```

NewUserHa: document for session should display the following visits are all in the same session,
therefore it was kept `last_visit` as the mark of the current session to be obviously seen.

Dreamsorcerer: I'm not sure I understand what you're saying.
This code will display `None` on first visit, and then the timestamp of the previous visit on all subsequent visits.

NewUserHa: I meant, the document/tutorial of `seesion` should indicate the session is the same after first request to show how it works, 
so the code only saved`last_visit` one time which is the first visit.

PR #8694, opened 2024-08-13
Add cleanup_ctx example for sharing resources across subapps

### LINKED ISSUE #3876: Sharing cleanup_ctx data with sub-apps
There is a documented approach of sharing resources from parent apps to nested apps in #2413. However that does not work for the common case of using CleanupContext managed resources (such as a database or redis pool).

Is there a recommended way of sharing resources provided by a root application's context signals?

_Originally posted in https://github.com/aio-libs/aiohttp/issues/2412#issuecomment-505492678_

Dreamsorcerer:
Fixes #3876.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -281,7 +281,7 @@ database object, do it explicitly::
    subapp[db_key] = mainapp[db_key]
    mainapp.add_subapp("/prefix", subapp)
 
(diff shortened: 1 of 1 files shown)


bdraco:
Screenshot 2024-08-13 at 8 47 49 PM

bdraco (approved):

PR #8698, opened 2024-08-14
Fix typing errors on content_disposition_header()

Dreamsorcerer:
The signature for the function doesn't type check well. If we unpack a `dict[str, str]`, we'll get a type error because quote_fields requires a bool and could be included in the unpacked arguments. All locations using the function seem to have a dict anyway...

This appears to be an internal function, not documented in our API, so risk should be low by changing the signature. Though will only do it on master anyway.

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -204,7 +204,7 @@ class Payload(ABC):
     ) -> None:
         """Sets ``Content-Disposition`` header."""
         self._headers[hdrs.CONTENT_DISPOSITION] = content_disposition_header(
-            disptype, quote_fields=quote_fields, _charset=_charset, params
+            disptype, quote_fields=quote_fields, _charset=_charset, params=params
         )
 
     @abstractmethod
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -200,11 +200,11 @@ class Payload(ABC):
         disptype: str,
         quote_fields: bool = True,
         _charset: str = "utf-8",
-        **params: Any,
+        **params: str,
     ) -> None:
         """Sets ``Content-Disposition`` header."""
         self._headers[hdrs.CONTENT_DISPOSITION] = content_disposition_header(
-            disptype, quote_fields=quote_fields, _charset=_charset, **params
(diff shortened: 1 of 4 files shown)


aiohttp/payload.py
```diff
@@ -200,11 +200,11 @@ def set_content_disposition(
         disptype: str,
         quote_fields: bool = True,
         _charset: str = "utf-8",
-        **params: Any,
+        **params: str,
     ) -> None:
         """Sets ``Content-Disposition`` header."""
         self._headers[hdrs.CONTENT_DISPOSITION] = content_disposition_header(
-            disptype, quote_fields=quote_fields, _charset=_charset, **params
+            disptype, quote_fields=quote_fields, _charset=_charset, params
```
Dreamsorcerer:
```suggestion
            disptype, quote_fields=quote_fields, _charset=_charset, params=params
```

tests/test_helpers.py
```diff
@@ -493,24 +493,29 @@ async def test_ceil_timeout_small_with_overriden_threshold(loop) -> None:
...
+            dict(filename='bär "\\'),
+            False,
+            "utf-8",
             'attachment; filename="bär \\"\\\\"',
         ),
     ],
 )
-def test_content_disposition(kwargs, result) -> None:
-    assert helpers.content_disposition_header("attachment", **kwargs) == result
+def test_content_disposition(params, quote_fields, _charset, result) -> None:
```
Dreamsorcerer:
```suggestion
def test_content_disposition(params, quote_fields, _charset, expected) -> None:
```

bdraco (approved):

bdraco:
Maybe this should get a changelog message as a breaking change

It does look like its used externally https://github.com/Starry-OvO/aiotieba/blob/1018eb96f26e6c17b7503f3457ad0accc2a9eb35/aiotieba/core/http.py#L135

Dreamsorcerer:
> some more on github code search https://github.com/search?q=content_disposition_header+aiohttp&type=code&p=2

Most of the results are just forks of aiohttp. I can see one which already passes params as a dict in the second argument, which wouldn't even work with the old code...

bdraco (approved):

PR #8597, opened 2024-08-03
Fix reading of body when ignoring an upgrade request

Dreamsorcerer:
Fixes #8414.

PR #8699, opened 2024-08-15
Reduce code indent in ResponseHandler.data_received

bdraco:
## What do these changes do?

Code cleanup only, no functional change

There were some else after returns that could be converted to guards to reduce code indent.

While looking at the full WebSockets path for https://github.com/aio-libs/aiohttp/discussions/8258 I noticed this could be improved a bit

Additionally, some missing test coverage is added and [unreachable code](https://github.com/aio-libs/aiohttp/pull/8699#issuecomment-2502632262) is removed.

no

no

### DIFF
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -258,7 +258,7 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
         if not data:
             return
 
-        # custom payload parser
+        # custom payload parser - currently always WebSocketReader
         if self._payload_parser is not None:
             payload_eof, payload_tail = self._payload_parser.feed_data(data)
             if payload_eof:
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -319,8 +319,5 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
             else:
                 self._drop_timeout()
 
-        if tail:
-            if upgraded:
-                self.data_received(tail)
-            else:
-                self._tail = tail
+        if upgraded and tail:
+            self.data_received(tail)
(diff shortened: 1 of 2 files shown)


bdraco:
Looks like we are missing coverage for `tail`

Dreamsorcerer:
I feel like that relates to https://github.com/aio-libs/aiohttp/pull/8597#issuecomment-2267101718 or one of my other parser changes. I'm not really sure if it's still possible to reach that code at all...

bdraco:
I'll dig through the code paths this weekend if flights cooperate.

bdraco:
Looks like websocket will never have a tail

bdraco:
I spent a few hours digging at it, and I can't find a case or make a case where `tail` exists and `upgraded` is `False`

Dreamsorcerer:
Then I reckon remove, but also skip the 3.10 backport as a precaution.

aiohttp/client_proto.py
```diff
@@ -270,51 +270,50 @@ def data_received(self, data: bytes) -> None:
...
-                if tail:
-                    if upgraded:
-                        self.data_received(tail)
-                    else:
-                        self._tail = tail
+        if tail:
+            if upgraded:
+                self.data_received(tail)
+            else:
+                self._tail = tail
```
bdraco:
```suggestion
            assert upgraded
            self.data_received(tail)
```

bdraco:
Still having trouble getting comfortable with removing that code, need to spend a few more hours pricing

bdraco:
I wonder if it is possible and we don't have coverage for it.

We didn't have test coverage for split WebSocket payloads until #9649

bdraco:
> Then I reckon remove it, but also skip the 3.10 backport as a precaution.

After discovering `feed_data` always return `b""` if its not upgraded, I'm confident its truely unreachable so I removed it
https://github.com/aio-libs/aiohttp/blob/2e369db2d9abcbef41031592b4f044adf5f89f59/aiohttp/_http_parser.pyx#L574

Issue #8028, opened 2024-01-12
Removal of 'loop' Parameter in web.run_app


libormatejek:
### Is your feature request related to a problem?

Yes, the absence of a detailed description for the 'loop' parameter in the web.run_app function documentation is causing confusion among developers. Users are unclear about the purpose and usage of this parameter, which can lead to improper implementations or avoidable errors.

### Describe the solution you'd like

I propose to add a comprehensive description of the 'loop' parameter including its purpose. This addition will provide clarity and assist developers in correctly utilizing the web.run_app function.

### Describe alternatives you've considered

An alternative would be to include links to relevant sections of the documentation where the 'loop' concept is explained in detail. However, directly adding the description within the web.run_app section ensures that the information is immediately accessible where it's most needed.

### Related component

Server

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
Hmm, the parameter should probably be removed. Like much of older asyncio APIs, loop parameters are generally not needed these days.

libormatejek:
Since version 3.8+ run_app evaluates asyncio.new_event_loop() instead of asyncio.get_event_loop() if no loop from kwargs is avaliable so there actually are use cases.

Dreamsorcerer:
There should not be, this is meant to work like asyncio.run() which does not take a loop parameter.
My only hesitation in removing it, is to first introduce a way that this might become compatible with asyncio.Runner.

DavidRomanovizc:
> There should not be, this is meant to work like asyncio.run() which does not take a loop parameter. My only hesitation in removing it, is to first introduce a way that this might become compatible with asyncio.Runner.

I can try this issue, but first, let me clarify smth: `web.run_app()` isn't a coroutine, and changing it to be compatible with `asyncio.Runner()` and save backward compatibility will be tricky. One possible solution is to use `web.AppRunner` for async launching, as it is specifically designed for this purpose. Another option could be using `web._run_app()` but, I think `web._run_app()` isn't meant for external use, so the solution I see is maybe creating another async func that calls `web._run_app()` and can be asynchronously run. Let me know what you think, and I'll be able to start

Dreamsorcerer:
Yes, my initial thoughts were more along the lines of introducing a new Runner, which would work like asyncio.Runner. Let me look again and I'll see if I can come up with a plan.

Dreamsorcerer:
[...]

PR #8723, opened 2024-08-16
Implement `web.Runner` context manager

dzhalaevd:
## What do these changes do?
These changes introduce the `web.Runner` context manager, based on `asyncio.Runner`. Added `web.Runner.run_app` method and refactored existing `web.run_app` to use `web.Runner.run_app` for running server. This feature preserves the backward compatibility

I think there aren't changes in behavior for users who will continue to use `web.run_app` as before; it remains fully compatible with existing code but users now have the option to use `web.Runner.run_app`

Supporting this change shouldn't be a substantial burden for maintainers because `web.Runner` builds on the `asyncio.Runner` with minor additions

Related to Issue: #8028 

## Checklist

- [ ] Documentation reflects the changes (Before updating docs, I want to make sure that I'm on the right track)


aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
 
+class _State(enum.Enum):
+    CREATED = "created"
+    INITIALIZED = "initialized"
+    CLOSED = "closed"
+
+
+@final
+class Runner:
+    """A context manager that controls event loop life cycle"""
```
Dreamsorcerer:
```suggestion
    """A copy of asyncio.Runner with the addition of .run_app() method."""
```

aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
+        handle_signals: bool = True,
+        reuse_address: Optional[bool] = None,
+        reuse_port: Optional[bool] = None,
+        handler_cancellation: bool = False,
+    ) -> None:
+        """Run an app locally"""
+        self._lazy_init()
+
+        self._loop.set_debug(self._debug)
+
```
Dreamsorcerer: This is already done in _laze_init().
```suggestion
```

aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
+                handle_signals=handle_signals,
+                reuse_address=reuse_address,
+                reuse_port=reuse_port,
+                handler_cancellation=handler_cancellation,
+            )
+        )
+
+        try:
+            if self._set_event_loop:
+                asyncio.set_event_loop(self._loop)
```
Dreamsorcerer: Also in _lazy_init()
```suggestion
```

aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
+            self._loop.run_until_complete(main_task)
+        except (GracefulExit, KeyboardInterrupt):  # pragma: no cover
+            pass
+        finally:
+            _cancel_tasks({main_task}, self._loop)
+            _cancel_tasks(asyncio.all_tasks(self._loop), self._loop)
+            self._loop.run_until_complete(self._loop.shutdown_asyncgens())
+            self.close()
+            asyncio.set_event_loop(None)
```
Dreamsorcerer: Already done in .close():
```suggestion
            _cancel_tasks({main_task}, self._loop)
            self.close()
```

aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
+            self._loop = events.new_event_loop()
+            if not self._set_event_loop:
+                # Call set_event_loop only once to avoid calling
+                # attach_loop multiple times on child watchers
+                events.set_event_loop(self._loop)
+                self._set_event_loop = True
+        else:
+            try:
+                self._loop = self._loop_factory()
+            except RuntimeError:
```
Dreamsorcerer: Why are we getting a RuntimeError here? This isn't in the upstream code.

aiohttp/web.py
```diff
@@ -277,6 +285,222 @@
...
+            sigint_handler = functools.partial(self._on_sigint, main_task=task)
+            try:
+                signal.signal(signal.SIGINT, sigint_handler)
+            except ValueError:
+                # `signal.signal` may throw if `threading.main_thread` does
+                # not support signals (e.g. embedded interpreter with signals
+                # not registered - see gh-91880)
+                sigint_handler = None
+        else:
+            sigint_handler = None
```
Dreamsorcerer: I wonder if this SIGINT code needs to be copied to run_app()...
Will have to do some more testing.

aiohttp/web.py
```diff
@@ -462,19 +686,15 @@ def run_app(
...
-
-    main_task = loop.create_task(
-        _run_app(
+    if loop is not None:
+
+        def loop_factory():
+            return loop
+
+    else:
+        loop_factory = events.get_running_loop
```
Dreamsorcerer: I think loop_factory should default to None (maybe this is the cause of the RuntimeError above?).
```suggestion
    loop_factory = None if loop is None else lambda: loop
```

tests/test_runner.py
```diff
@@ -0,0 +1,295 @@
...
+        policy = asyncio.get_event_loop_policy()
+        if policy.loop is not None:
+            self.assertTrue(policy.loop.is_closed())
+            self.assertTrue(policy.loop.shutdown_ag_run)
+
+        asyncio.set_event_loop_policy(None)
+        super().tearDown()
+
+
+class RunTests(BaseTest):
```
Dreamsorcerer: We don't need tests for asyncio.run(), only the Runner tests.

tests/test_runner.py
```diff
@@ -0,0 +1,295 @@
+import asyncio
```
Dreamsorcerer:
```suggestion
"""Tests from cpython for web.Runner()."""

import asyncio
```

tests/test_runner.py
Dreamsorcerer, line 1: We'll also want some tests in test_run_app.py to verify the .run_app() method separately.
e.g. Maybe test that running a simple task followed by .run_app() works (and happens in the same loop).

Dreamsorcerer (changes requested):
Looks like a good start. Still needs some fixes and documentation.

Dreamsorcerer:
Also, looking at the implementation, the .run_app() method only uses ._lazy_init() and ._loop from the class. So, I'm wondering if we should just ignore the comment about the class being final and subclass it anyway. Then just rely on decent tests to ensure we maintain compatibility.

bdraco:
Traveling today for the next 23 hours. Will take a look when I get to my destination

bdraco:
Looks like I'll be able to look today after all. Flight isn't happening due to engine issues so I'll be able to look once I get rebooked and back home.

bdraco:
> @bdraco Does this look reasonable overall? I think after merging this, I'll make a followup PR to move all of run_app/Runner code to a separate file, as this one is getting a bit too big.

It looks reasonable to me. I want to subclass it instead of copying all the code, as keeping it in sync imposes some maintenance burden. Unfortunately, upstream explicitly says not to do that. We should add some comments about periodically checking upstream for bug fixes if we don't want to ignore that and do it anyways since we will have an out-of-sync copy as soon as anything changes.

Dreamsorcerer:
Yeah, I think we should just subclass it, there's just too much code otherwise with test duplicated etc. We only depend on lazy_init and the loop attribute from the class, so I think risk of breakage is low.

@DavidRomanovizc Would you be alright changing it to a subclass and the removing the copied tests?
You'll need to define it with:
```
if sys.version_info >= (3, 11):
    class Runner(...
```
Then in run_app() also do a version check and use the new Runner code in 3.11+ or the old implementation otherwise (we'll then remove the old version when we drop support for 3.10).

aiohttp/web.py
Dreamsorcerer: Not relevant if we subclass.

aiohttp/web.py
Dreamsorcerer: Not relevant if we subclass.

tests/test_runner.py
Dreamsorcerer: File should be removed if we subclass.

tests/test_runner.py
Dreamsorcerer: File should be removed if we subclass.

DavidRomanovizc:
Sorry for the long inactivity. Decided to rename `Runner` to `WebRunner`, inherited `asyncio.Runner` and added check version. In `web.run_app` returned old implementation and use the new Runner code if py version 3.11+ or the old impl otherwise

aiohttp/web.py
```diff
@@ -277,6 +280,130 @@
...
+
+    class _State(enum.Enum):
+        CREATED = "created"
+        INITIALIZED = "initialized"
+        CLOSED = "closed"
+
+    class WebRunner(asyncio.Runner):  # type: ignore
+        """A context manager that controls event loop life cycle"""
+
+        def __init__(
```
Dreamsorcerer, line 293: Still a lot of copied code here. As far as I can tell, only the run_app() method needs to be defined here. init/close/_lazy_init can all be removed and just use the parent methods.

aiohttp/web.py
```diff
@@ -277,6 +280,130 @@
...
+                self._loop.get_debug()
+                and access_log
+                and access_log.name == "aiohttp.access"
+            ):
+                if access_log.level == logging.NOTSET:
+                    access_log.setLevel(logging.DEBUG)
+                if not access_log.hasHandlers():
+                    access_log.addHandler(logging.StreamHandler())
+
+            main_task = self._loop.create_task(
```
Dreamsorcerer, line 353: Should probably add the context parameter for compatibility:
https://github.com/python/cpython/blob/dc76a4ad3c46af3fb70361be10b8a791cd126931/Lib/asyncio/runners.py#L106-L109

aiohttp/web.py
```diff
@@ -463,54 +590,77 @@ def run_app(
...
-            print=print,
-            backlog=backlog,
-            access_log_class=access_log_class,
-            access_log_format=access_log_format,
-            access_log=access_log,
-            handle_signals=handle_signals,
-            reuse_address=reuse_address,
-            reuse_port=reuse_port,
-            handler_cancellation=handler_cancellation,
+    if sys.version_info < (3, 11):
```
Dreamsorcerer:
```suggestion
    if sys.version_info >= (3, 11):
```

aiohttp/web.py
```diff
@@ -277,6 +280,130 @@
...
 HostSequence = TypingIterable[str]
 
+if sys.version_info >= (3, 11):
+
+    class _State(enum.Enum):
+        CREATED = "created"
+        INITIALIZED = "initialized"
+        CLOSED = "closed"
+
+    class WebRunner(asyncio.Runner):  # type: ignore
```
Dreamsorcerer, line 290: We should probably get a new test that uses WebRunner directly too. Can probably just copy and modify one of the run_app() tests. The new test could also use WebRunner.run() to verify that the task and the app are run in the same loop.

DavidRomanovizc:
@Dreamsorcerer, I thought, why don't we replace inheritance with composition? We will still be able to use the functionality of `asyncio.Runner`. Also, we won't have to spend effort on maintenance because we ensure synchronization with the upstream

like that:

```python
class WebRunner:
    def __init__(
        self,
        *,
        debug: Optional[bool] = None,
        loop_factory: Optional[Callable[[], asyncio.AbstractEventLoop]] = None,
    ):
        self.debug = debug
        self.loop_factory = loop_factory
        self._runner = asyncio.Runner(debug=debug, loop_factory=loop_factory)
   ...
```

Dreamsorcerer:
> I thought, why don't we replace inheritance with composition?

We need to call _lazy_init(), right? So, we can't just rely on the public interface..

PR #8734, opened 2024-08-17
Test typing round 4

Dreamsorcerer:


aiohttp/web_protocol.py
```diff
@@ -530,14 +537,15 @@ async def start(self) -> None:
...
+                request_handler = self._request_handler
+            else:
                 # make request_factory work
                 request_handler = self._make_error_handler(message)
                 message = ERROR
-            else:
-                request_handler = self._request_handler
 
             request = self._request_factory(message, payload, self, writer, handler)
+            print(1, type(request))
```
Dreamsorcerer:
```suggestion
```

aiohttp/tracing.py
```diff
@@ -342,8 +350,8 @@ class Trace:
     def __init__(
         self,
         session: "ClientSession",
-        trace_config: TraceConfig,
-        trace_config_ctx: SimpleNamespace,
+        trace_config: TraceConfig[object],
+        trace_config_ctx: Any,
```
bdraco, line 354:
```suggestion
        trace_config: TraceConfig[_T],
        trace_config_ctx: _T,
```

Not sure if the TypeVar can be used here or not

Dreamsorcerer: It would require making Trace a Generic, which I think starts to add overhead to people's code, particularly as there's no way to pass the type through automatically, so every tracing handler would need manual annotations. Not sure if we want to go down that route...

Dreamsorcerer: This is essentially my cutoff point between type safety and keeping the code to allow anything the user wants.

PR #8735, opened 2024-08-18
Refactor http websocket reader parser frame handler

bdraco:
Testing with Home Assistant shows this is a ~22% speed up.

Will need to develop some benchmarks

related discussion https://github.com/aio-libs/aiohttp/discussions/8258

### DIFF
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -386,15 +386,6 @@ class WebSocketReader:
                 )
             return
 
-        # previous frame was non finished
-        # we should get continuation opcode
-        if self._partial and not is_continuation:
-            raise WebSocketError(
-                WSCloseCode.PROTOCOL_ERROR,
-                "The opcode in non-fin frame is expected "
-                "to be zero, got {!r}".format(opcode),
-            )
-
         if is_continuation:
             if self._opcode is None:
                 raise WebSocketError(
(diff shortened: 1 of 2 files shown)


aiohttp/http_websocket.py
```diff
@@ -304,134 +307,150 @@ def feed_data(self, data: bytes) -> Tuple[bool, bytes]:
...
+                "to be zero, got {!r}".format(opcode),
+            )
 
-        return False, b""
+        if is_continuation:
+            assert self._opcode is not None
+            opcode = self._opcode
+            self._opcode = None
+
+        self._partial.extend(payload)
```
bdraco: This is where most of the overhead is. When there are many small frames we end up making lots of copies that we do not need to make

bdraco: https://stackoverflow.com/questions/40004517/why-does-bytearray-obj-extendbytes-differ-from-bytearray-obj-bytes

bdraco:
Its only about 7% faster with continuations.

I think we can do a smaller change that optimizes for both cases

bdraco:
paired down version that should optimize well for both cases https://github.com/aio-libs/aiohttp/pull/8736

bdraco:
This got too big.

Going for a smaller iteration in https://github.com/aio-libs/aiohttp/pull/8736

PR #8736, opened 2024-08-19
Improve performance of WebSocketReader

bdraco:
Optimization

- Replace `bytearray.extend` with `+` and `+=` https://stackoverflow.com/questions/40004517/why-does-bytearray-obj-extendbytes-differ-from-bytearray-obj-bytes
- Avoid making copies of the incoming data when there is no continuation. `self._partial` is never needed for the non-continuation case and we can avoid having to `extend` and `clear` it for this case, which is the common case for many use cases.

This is a paired down version of #8735 to better optimize for both the non-continuation and continuation case

Testing `feed_data` on a live HA instance. Testing was done on an Odroid N2+ SBC. Performance numbers will vary based on CPU and memory performance.

~6% faster for the continuation case (real world use case is listening to a unifi instance)
~34% faster for the non-continuation case (real world use case is reading Bluetooth data from a shelly device, web based terminal shows a similar improvement)

Improvement percentages will be less impressive on systems with faster CPU/memory performance.

related discussion https://github.com/aio-libs/aiohttp/discussions/8258

benchmark for non-continuation case
```python
import random
import zlib
from typing import Any
import timeit
import asyncio
import aiohttp
from aiohttp.http import WSMsgType
from aiohttp.http_websocket import (
    _WS_DEFLATE_TRAILING,
    PACK_LEN1,
    PACK_LEN2,
    PACK_LEN3,
    WebSocketReader,
    _websocket_mask,
)


def build_frame(
    message: Any,
    opcode: Any,
    use_mask: bool = False,
    noheader: bool = False,
    is_fin: bool = True,
    compress: bool = False,
):
    # Send a frame over the websocket with message as its payload.
    if compress:
        compressobj = zlib.compressobj(wbits=-9)
        message = compressobj.compress(message)
        message = message + compressobj.flush(zlib.Z_SYNC_FLUSH)
        if message.endswith(_WS_DEFLATE_TRAILING):
            message = message[:-4]
    msg_length = len(message)
    if use_mask:  # pragma: no cover
        mask_bit = 0x80
    else:
        mask_bit = 0

    if is_fin:
        header_first_byte = 0x80 | opcode
    else:
        header_first_byte = opcode

    if compress:
        header_first_byte |= 0x40

    if msg_length < 126:
        header = PACK_LEN1(header_first_byte, msg_length | mask_bit)
    elif msg_length < (1 << 16):  # pragma: no cover
        header = PACK_LEN2(header_first_byte, 126 | mask_bit, msg_length)
    else:
        header = PACK_LEN3(header_first_byte, 127 | mask_bit, msg_length)

    if use_mask:  # pragma: no cover
        mask = random.randrange(0, 0xFFFFFFFF)
        mask = mask.to_bytes(4, "big")
        message = bytearray(message)
        _websocket_mask(mask, message)
        if noheader:
            return message
        else:
            return header + mask + message
    else:
        if noheader:
            return message
        else:
            return header + message


def out(loop):
    return aiohttp.DataQueue(loop)


def parser(out):
    return WebSocketReader(out, 4 * 1024 * 1024)


async def bench():
    data = build_frame(b"thisisthebinaryframe" * 256, WSMsgType.BINARY)
    out_queue = out(loop)
    parser_instance = parser(out_queue)
    print(
        timeit.timeit(
            "parser_instance.feed_data(data)", globals=locals(), number=1000000
        )
    )


asyncio.run(bench())
```

profiler
```python
import random
import zlib
from typing import Any
import asyncio
import aiohttp
from aiohttp.http import WSMsgType
import cProfile
from aiohttp.http_websocket import (
    _WS_DEFLATE_TRAILING,
    PACK_LEN1,
    PACK_LEN2,
    PACK_LEN3,
    WebSocketReader,
    _websocket_mask,
)


def build_frame(
    message: Any,
    opcode: Any,
    use_mask: bool = False,
    noheader: bool = False,
    is_fin: bool = True,
    compress: bool = False,
):
    # Send a frame over the websocket with message as its payload.
    if compress:
        compressobj = zlib.compressobj(wbits=-9)
        message = compressobj.compress(message)
        message = message + compressobj.flush(zlib.Z_SYNC_FLUSH)
        if message.endswith(_WS_DEFLATE_TRAILING):
            message = message[:-4]
    msg_length = len(message)
    if use_mask:  # pragma: no cover
        mask_bit = 0x80
    else:
        mask_bit = 0

    if is_fin:
        header_first_byte = 0x80 | opcode
    else:
        header_first_byte = opcode

    if compress:
        header_first_byte |= 0x40

    if msg_length < 126:
        header = PACK_LEN1(header_first_byte, msg_length | mask_bit)
    elif msg_length < (1 << 16):  # pragma: no cover
        header = PACK_LEN2(header_first_byte, 126 | mask_bit, msg_length)
    else:
        header = PACK_LEN3(header_first_byte, 127 | mask_bit, msg_length)

    if use_mask:  # pragma: no cover
        mask = random.randrange(0, 0xFFFFFFFF)
        mask = mask.to_bytes(4, "big")
        message = bytearray(message)
        _websocket_mask(mask, message)
        if noheader:
            return message
        else:
            return header + mask + message
    else:
        if noheader:
            return message
        else:
            return header + message


def out(loop):
    return aiohttp.DataQueue(loop)


def parser(out):
    return WebSocketReader(out, 4 * 1024 * 1024)


async def bench():
    data = build_frame(b"thisisthebinaryframe" * 256, WSMsgType.BINARY)
    out_queue = out(loop)
    parser_instance = parser(out_queue)
    pr = cProfile.Profile()
    pr.enable()
    for _ in range(1000000):
        parser_instance.feed_data(data)
    pr.disable()
    pr.create_stats()
    pr.dump_stats("feed.cprof")


asyncio.run(bench())
```

### DIFF
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -399,7 +399,6 @@ class WebSocketReader:
 
                 # Decompress process must to be done after all packets
                 # received.
-                payload_merged: Optional[bytes] = None
                 if compressed:
                     if not self._decompressobj:
                         self._decompressobj = ZLibDecompressor(
@@ -416,6 +415,8 @@ class WebSocketReader:
                                 self._max_msg_size + left, self._max_msg_size
                             ),
                         )
+                else:
+                    payload_merged = bytes(assembled_payload)
 
                 if opcode == WSMsgType.TEXT:
                     try:
@@ -428,8 +429,7 @@ class WebSocketReader:
                     self.queue.feed_data(WSMessage(WSMsgType.TEXT, text, ""))
                     continue
 
-                data = payload_merged or bytes(assembled_payload)
-                self.queue.feed_data(WSMessage(WSMsgType.BINARY, data, ""))
+                self.queue.feed_data(WSMessage(WSMsgType.BINARY, payload_merged, ""))
 
     def parse_frame(
         self, buf: bytes
--- a/aiohttp/http_websocket.py
+++ b/aiohttp/http_websocket.py
@@ -314,45 +314,7 @@ class WebSocketReader:
 
     def _feed_data(self, data: bytes) -> None:
         for fin, opcode, payload, compressed in self.parse_frame(data):
-            if opcode == WSMsgType.CLOSE:
-                if len(payload) >= 2:
-                    close_code = UNPACK_CLOSE_CODE(payload[:2])[0]
-                    if close_code < 3000 and close_code not in ALLOWED_CLOSE_CODES:
-                        raise WebSocketError(
-                            WSCloseCode.PROTOCOL_ERROR,
-                            f"Invalid close code: {close_code}",
-                        )
-                    try:
-                        close_message = payload[2:].decode("utf-8")
-                    except UnicodeDecodeError as exc:
-                        raise WebSocketError(
-                            WSCloseCode.INVALID_TEXT, "Invalid UTF-8 text message"
-                        ) from exc
-                    msg = WSMessage(WSMsgType.CLOSE, close_code, close_message)
-                elif payload:
-                    raise WebSocketError(
-                        WSCloseCode.PROTOCOL_ERROR,
-                        f"Invalid close frame: {fin} {opcode} {payload!r}",
-                    )
-                else:
-                    msg = WSMessage(WSMsgType.CLOSE, 0, "")
-
-                self.queue.feed_data(msg)
-
-            elif opcode == WSMsgType.PING:
-                self.queue.feed_data(WSMessage(WSMsgType.PING, payload, ""))
-
-            elif opcode == WSMsgType.PONG:
-                self.queue.feed_data(WSMessage(WSMsgType.PONG, payload, ""))
-
-            elif (
-                opcode not in (WSMsgType.TEXT, WSMsgType.BINARY)
-                and self._opcode is None
-            ):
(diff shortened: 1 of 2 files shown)


aiohttp/http_websocket.py
```diff
@@ -85,6 +85,12 @@ class WSMsgType(IntEnum):
     ERROR = 0x102
 
 
+MESSAGE_TYPES_WITH_CONTENT: Final = (
+    WSMsgType.BINARY,
+    WSMsgType.TEXT,
+    WSMsgType.CONTINUATION,
+)
```
Dreamsorcerer, line 92: Should be a frozenset if we're using `in` below?

bdraco: The performance won't make much difference until we get to more than a few items.

bdraco:
```
0.013342750000447268
0.018335040998863406
```
frozenset is tiny bit faster

bdraco: https://github.com/aio-libs/aiohttp/pull/8747

Dreamsorcerer: Yeah, tuples are only slightly faster when the first item matches and slower the rest.of the time, so good practice to just always use sets.

PR #8748, opened 2024-08-19
Add Python 3.13 classifier

bdraco:
## What do these changes do?

Add Python 3.13 classifier like https://github.com/aio-libs/aiohttp/pull/8259

#8740 will start shipping wheels for 3.13

We need to wait for 3.10.5 to be published with 3.13 wheels so uvloop can update and we can get a clean CI run, than we can do a 3.10.6 with this PR

no

no


bdraco:
I don't think the tests are going to pass though so will probably hold off on adding this into 3.10.5 and once uvloop does their release we can likely to 3.10.6 with this backported

webknjaz:
@bdraco I was also thinking that we need to start looking into 3.13t compat here and in the deps too..

bdraco:
I'm working on getting 3.10.5 with cibuildwheel updated to v2.20.0 so we get 3.13 wheels out and uvloop can update. Once they release than we will hopefully see green on the CI and can do 3.10.6 with the classifier.

Dreamsorcerer:
> @bdraco I was also thinking that we need to start looking into 3.13t compat here and in the deps too..

Yeah, so we're just waiting on uvloop, who are actually waiting on us. So, this'll get wheels out to unblock uvloop, then they'll unblock us.

bdraco:
3.13 wheels seem to be building ok https://github.com/aio-libs/aiohttp/actions/runs/10459267569/job/28963810550

bdraco:
We have 3.13 wheels now 🍾

webknjaz:
I think we usually only started uploading wheels after the final release of CPython. Should we yank them once it's out and republish? Or publish wheels into the same release with different build numbers?

@Dreamsorcerer why does uvloop wait for us to work on free-threading?

bdraco:
cibuildwheel starts building wheels by default once a python version reaches RC since thats the cut off for binary compatibility

https://github.com/pypa/cibuildwheel/releases/tag/v2.20.0

webknjaz:
Yeah, I know. Just not sure how safe it is.

bdraco:
> Yeah, I know. Just not sure how safe it is.

I've published wheels for lots of other projects when RC is out and never had an issue, but that's only my experience.

Dreamsorcerer:
We've always just let the release process build wheels as it wants. We just don't declare support until the CI passes (which for 3.11/12 was over a year...).

uvloop uses aiohttp for their tests. So, if we don't publish wheels until our CI passes, and they do the same, then neither of us will ever have a working release.

bdraco:
uv loop released 0.21.0beta1 https://github.com/MagicStack/uvloop/releases/tag/v0.21.0beta1

Dreamsorcerer:
Yeah, I'm hoping for a stable release next week, so we can just leave it to Dependabot. If doesn't happen next week, we'll have to bump ourselves.

Dreamsorcerer:
Probably, I never use those make commands. If nobody's noticed them missing, then seems like nobody else uses them either. So, maybe they can just be removed...

PR #8749, opened 2024-08-19
Bump cffi to 1.17.0

bdraco:
## What do these changes do?

cffi 1.17.0 is needed for python 3.13 support

no
no


bdraco:
I figured dependabot would have done this by now but since it hasn't it looks like we need to do it manually

Dreamsorcerer:
Hmm, it's seen them: https://github.com/aio-libs/aiohttp/network/updates/870981818
Not sure why it doesn't update it. Maybe it only updates the ones listed in the .in files?

Dreamsorcerer:
I think we want to add:
```
    allow:
      - dependency-type: "all"
```
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

PR #8763, opened 2024-08-19
DO NOT MERGE

bdraco:
back merge 3.10->3.11->master


bdraco:
might be a while before this can run since dependabot is eating all the CI time.

Hopefully it runs before I have to leave for dinner

PR #8797, opened 2024-08-20
Drop Python 3.8

Dreamsorcerer:
The 3.11 release won't happen for some time, so we can start dropping support from the codebase now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import sys
 
 from setuptools import Extension, setup
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 9):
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -4,7 +4,7 @@
 
 from setuptools import Extension, setup
 
-if sys.version_info < (3, 8):
+if sys.version_info < (3, 9):
     raise RuntimeError("aiohttp 4.x requires Python 3.8+")
```
bdraco:
```suggestion
    raise RuntimeError("aiohttp 4.x requires Python 3.9+")
```

bdraco (approved):

bdraco:
There is a resolver test to still cleanup as well

PR #8845, opened 2024-08-22
Return 500 error when handler has wrong return type

### LINKED ISSUE #4572: Server closes connection if wrong type returned by handler

Dreamsorcerer:
Fixes #4572.

Not entirely sure if this is the best solution. Will sort the test tomorrow.

### DIFF
--- a/CHANGES/8845.bugfix.rst
+++ b/CHANGES/8845.bugfix.rst
@@ -1,0 +1,1 @@
+Changed behaviour when returning an invalid response to send a 500 response -- by :user:`Dreamsorcerer`.
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -39,7 +39,7 @@ from .http import (
 from .log import access_logger, server_logger
 from .streams import EMPTY_PAYLOAD, StreamReader
 from .tcp_helpers import tcp_keepalive
-from .web_exceptions import HTTPException
+from .web_exceptions import HTTPException, HTTPInternalServerError
 from .web_log import AccessLogger
 from .web_request import BaseRequest
 from .web_response import Response, StreamResponse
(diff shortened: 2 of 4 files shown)


webknjaz:
Perhaps, instead of returning a tuple, an exception wrapping the response could be raised and unwrapped/handled on the  calling side?

Dreamsorcerer:
So, it's already covered by tests, they just asserted that the server would disconnect, which seems like an odd choice..

Dreamsorcerer:
> Perhaps, instead of returning a tuple, an exception wrapping the response could be raised and unwrapped/handled on the calling side?

I don't think that works without more substantial refactoring. The actual writing happens in finish_response() where I create the replacement Response, the only reason I then return the new response, is because the calling function passes the resp to something else which accesses an attribute on it one more time, so I need it to replace the invalid resp object it passed in with the new one.

Dreamsorcerer:
@webknjaz Backport's broken...

webknjaz:
I know, GitHub's API response started including a new field which broke the mapping. I'll try to fix it today.

Dreamsorcerer:
Oh, that's timeline as well. So, won't be able to merge anything new at the moment.

PR #8875, opened 2024-08-24
Fix unclosed transport warning

### LINKED ISSUE #5426: ResourceWarning: unclosed transport  when using SSL on server
🐞 **Describe the bug**
Server with ssl enabled do not closes `_SSLProtocolTransport` properly, which leads to ResourceWarning. This 

💡 **To Reproduce**

Run the following script (requires [trustme](https://pypi.org/project/trustme/) package).
```python
import asyncio
import ssl
import warnings

import aiohttp
import trustme
from aiohttp import web

# Enable all warnings
warnings.simplefilter('always')


async def run():
    # Prepare ssl
    certificate = trustme.CA().issue_server_cert("localhost", "127.0.0.1", "::1")
    ssl_context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
    certificate.configure_cert(ssl_context)

    # Prepare server
    app = web.Application()

    async def do_nothing(request: web.Request) -> web.Response:
        return web.Response(text="OK")

    app.router.add_routes([web.get("", do_nothing)])

    runner = web.AppRunner(app)
    await runner.setup()
    site = web.TCPSite(runner, "127.0.0.1", 8000, ssl_context=ssl_context)
    await site.start()

    # Do request
    connector = aiohttp.TCPConnector(ssl=False)
    async with aiohttp.ClientSession(connector=connector) as session:
        async with session.get("https://localhost:8000") as resp:
            assert await resp.text() == "OK"

    # Close server
    await runner.cleanup()


if __name__ == "__main__":
    asyncio.run(run())
```

💡 **Expected behavior**

No warnings are generated.

📋 **Logs/tracebacks**

```python-traceback (paste your traceback in the next line)
python3.9/asyncio/sslproto.py:320: ResourceWarning: unclosed transport <asyncio.sslproto._SSLProtocolTransport object at 0x107471e80>
  _warn(f"unclosed transport {self!r}", ResourceWarning, source=self)
```

```console
$ python --version
Python 3.9.1
```

```console
$ python -m pip show aiohttp
Version: 3.7.3
```
```console
Version: 5.1.0
```
```console
$ python -m pip show yarl
Version: 1.6.3
```

📋 **Additional context**

As I can see from source code, the reason is that `web_protocol.RequestHandler.connection_lost` is called when `transport` is still active because of keepalive. As a workaround, one can set `keepalive_timeout` to zero (it can be OK for tests):
 `runner = web.AppRunner(app, keepalive_timeout=0)`.

Dreamsorcerer:
Fixes #5426.

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -330,13 +330,12 @@ class RequestHandler(BaseProtocol, Generic[_Request]):
             return
         self._manager.connection_lost(self, exc)
 
-        super().connection_lost(exc)
-
         # Grab value before setting _manager to None.
         handler_cancellation = self._manager.handler_cancellation
 
+        self.force_close()
+        super().connection_lost(exc)
         self._manager = None
-        self._force_close = True
         self._request_factory = None
         self._request_handler = None
         self._request_parser = None
@@ -348,9 +347,6 @@ class RequestHandler(BaseProtocol, Generic[_Request]):
             if exc is None:
                 exc = ConnectionResetError("Connection lost")
             self._current_request._cancel(exc)
-
-        if self._waiter is not None:
-            self._waiter.cancel()
 
         if handler_cancellation and self._task_handler is not None:
             self._task_handler.cancel()
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -283,9 +283,6 @@ class RequestHandler(BaseProtocol, Generic[_Request]):
         if self._keepalive_handle is not None:
             self._keepalive_handle.cancel()
 
-        if self._waiter:
-            self._waiter.cancel()
-
         # Wait for graceful handler completion
         if self._handler_waiter is not None:
             with suppress(asyncio.CancelledError, asyncio.TimeoutError):
@@ -304,9 +301,7 @@ class RequestHandler(BaseProtocol, Generic[_Request]):
         if self._task_handler is not None:
             self._task_handler.cancel()
 
-        if self.transport is not None:
-            self.transport.close()
-            self.transport = None
+        self.force_close()
 
     def connection_made(self, transport: asyncio.BaseTransport) -> None:
         super().connection_made(transport)
--- a/setup.cfg
+++ b/setup.cfg
@@ -136,7 +136,6 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:unclosed transport <asyncio.sslproto._SSLProtocolTransport object.*:ResourceWarning
     ignore:unclosed transport <_ProactorSocketTransport closing fd=-1>:ResourceWarning
     ignore:Unclosed client session <aiohttp.client.ClientSession object at 0x:ResourceWarning
     ignore:The loop argument is deprecated:DeprecationWarning:asyncio
--- a/setup.cfg
+++ b/setup.cfg
@@ -136,7 +136,6 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:unclosed transport <_ProactorSocketTransport closing fd=-1>:ResourceWarning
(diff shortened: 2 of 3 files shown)


Dreamsorcerer:
Just checking that existing tests reliably cover the warning.

aiohttp/web_protocol.py
```diff
@@ -330,13 +330,12 @@ def connection_lost(self, exc: Optional[BaseException]) -> None:
             return
         self._manager.connection_lost(self, exc)
 
-        super().connection_lost(exc)
```
Dreamsorcerer, line 333: Basically, the parent class just sets transport to None without closing it. This function never even tried to close it before, so it looks like calling .force_close() is more sensible.

setup.cfg
```diff
@@ -136,7 +136,6 @@ addopts =
 filterwarnings =
     error
     ignore:module 'ssl' has no attribute 'OP_NO_COMPRESSION'. The Python interpreter is compiled against OpenSSL < 1.0.0. Ref. https.//docs.python.org/3/library/ssl.html#ssl.OP_NO_COMPRESSION:UserWarning
-    ignore:unclosed transport <asyncio.sslproto._SSLProtocolTransport object.*:ResourceWarning
     ignore:unclosed transport <_ProactorSocketTransport closing fd=-1>:ResourceWarning
```
Dreamsorcerer: Let's see if the Proactor warning is the same source.
```suggestion
```

aiohttp/web_protocol.py
bdraco: Shouldn't it already be closing if we get here?
https://github.com/python/cpython/blob/c4ee4e756a311f03633723445299bde90eb7b79c/Lib/asyncio/selector_events.py#L865

bdraco: I guess it doesn't matter since it will do nothing in that case

bdraco:
Looks good. Let me throw it on production before I start the day

bdraco:
Put it on production with a system that uses ssl. Forced disconnect and gracefully disconnected a few times. No errors observed

bdraco (approved):

aiohttp/web_protocol.py
Dreamsorcerer, line 1: It looks like line 286/287 is no longer needed? I can't seem to figure out why though..

Dreamsorcerer: I think the only place that method should be called in is via Server.shutdown(), which should be called after Server.pre_shutdown(). pre_shutdown() call .close() which cancels the waiter, so it should always be cancelled before .shutdown() is called.

Just not sure what my change has to do with this...

Dreamsorcerer: I don't know if I should actually remove those lines or not...

bdraco: Maybe we don't have a test that ever calls `shutdown` before `connection_lost`?

Dreamsorcerer: I'm not sure it's possible. I tried holding a handler open, ignoring cancellation and it still doesn't get there.

bdraco: I guess `self._waiter` always gets set to `None` in `start` since its `.cancel()` is always getting called and the cancellation gets delivered via `call_soon` on the next loop.

bdraco: I guess that happened before and the test was only hitting it on the connection_lost case which doesn't happen anymore because its already cancelled... but that doesn't make sense since the call_soon should not have fired yet because `connection_lost` isn't a coro that yields to the loop....hmm

bdraco: It seems the call_soon cancellation is now always able to get delivered before the shutdown code is run. That seems like any small change in the future could change the order again...

Dreamsorcerer: Maybe it should just be replaced with another .force_close() call?

bdraco: I was thinking that as well but unsure if `_task_handler` and `_handler_waiter` need to be awaited before calling `self.transport.close()`.

If not its even cleaner because than we can remove https://github.com/aio-libs/aiohttp/blob/48a5e07ad833bd1a8fcb2ce6f85a41ad0cef9dc6/aiohttp/web_protocol.py#L307 as well

Dreamsorcerer: I'm reasonable confident that they should be awaited beforehand. I'm not convinced that the waiter needs to be cancelled first though. I think the waiter should only be active if the handler is not yet running (it is awaited in .start() while waiting for an incoming message). So, putting force_close() after those awaits should be fine...

Dreamsorcerer: @bdraco Double check that last change, but I think that should be good. Tests are all passing still.

bdraco: I think this is fine. AFAICT, `_handler_waiter` should never be waiting for `_waiter` so there is no need to worry about it being canceled before waiting for `_handler_waiter`

bdraco (approved):

PR #8891, opened 2024-08-26
Return 500 error when handler has wrong return type (#8845)

Dreamsorcerer:
(cherry picked from commit 48a5e07ad833bd1a8fcb2ce6f85a41ad0cef9dc6)

### DIFF
--- a/aiohttp/web_protocol.py
+++ b/aiohttp/web_protocol.py
@@ -646,10 +646,10 @@ class RequestHandler(BaseProtocol):
             await prepare_meth(request)
             await resp.write_eof()
         except ConnectionError:
-            await self.log_access(request, resp, start_time)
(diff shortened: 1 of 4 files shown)


aiohttp/web_protocol.py
```diff
@@ -635,22 +631,26 @@ async def finish_response(
...
-            self.log_access(request, resp, start_time)
-            return True
-        else:
-            self.log_access(request, resp, start_time)
-            return False
+            await self.log_access(request, resp, start_time)
+            return resp, True
+
+        await self.log_access(request, resp, start_time)
+        return resp, False
```
Dreamsorcerer:
```suggestion
            self.log_access(request, resp, start_time)
            return resp, True

        self.log_access(request, resp, start_time)
        return resp, False
```

PR #8898, opened 2024-08-26
Fix router matching pre-encoded URLs

### LINKED ISSUE #5621: 3.7.0 regression: aiohttp route matching is now pre url decode instead of post decode

### LINKED ISSUE #6619: `custom regex` of router should support `\s` of the pattern
### Describe the bug

router like `/{fn:\w+ \d+}` or `/{fn:\w+\s\d+}` can't match path like `/abc 123`,
because of the [coroutine resolve(request)](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.UrlDispatcher.resolve) uses [raw_path](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.BaseRequest.raw_path), which is `/abc%20123` and needs pattern like `/{fn:\w+%20\d+}`.

Usually, users will write ` ` or `\s` for a pattern instead of `%20`

### Python Version

```console
$ python --version
Python 3.8.10
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.8.1
```

### Related component

Server

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #5621.
Fixes #6619.

### DIFF
--- a/CHANGES/8898.bugfix.rst
+++ b/CHANGES/8898.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed web router not matching pre-encoded URLs -- by :user:`Dreamsorcerer`.
+Fixed web router not matching pre-encoded URLs (requires yarl 1.9.6+) -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8898.bugfix.rst
+++ b/CHANGES/8898.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed web router not matching pre-encoded URLs -- by :user:`Dreamsorcerer`.
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -1140,7 +1140,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
             if resource.name == name and resource.raw_match(path):
                 return cast(Resource, resource)
         if not ("{" in path or "}" in path or ROUTE_RE.search(path)):
-            resource = PlainResource(_requote_path(path), name=name)
+            resource = PlainResource(path, name=name)
             self.register_resource(resource)
             return resource
         resource = DynamicResource(path, name=name)
@@ -1267,7 +1267,7 @@ def _quote_path(value: str) -> str:
 
 
 def _unquote_path(value: str) -> str:
-    return URL.build(path=value, encoded=True).path
+    return URL.build(path=value, encoded=True).path.replace("%2F", "/")
 
 
 def _requote_path(value: str) -> str:
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -358,7 +358,7 @@ class Resource(AbstractResource):
     async def resolve(self, request: Request) -> _Resolve:
         allowed_methods: Set[str] = set()
 
-        match_dict = self._match(request.rel_url.raw_path)
+        match_dict = self._match(request.rel_url.path)
         if match_dict is None:
             return None, allowed_methods
 
@@ -626,7 +626,7 @@ class StaticResource(PrefixResource):
         )
 
     async def resolve(self, request: Request) -> _Resolve:
-        path = request.rel_url.raw_path
+        path = request.rel_url.path
         method = request.method
         allowed_methods = set(self._routes)
         if not path.startswith(self._prefix2) and path != self._prefix:
@@ -1015,7 +1015,7 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
         # candidates for a given url part because there are multiple resources
         # registered for the same canonical path, we resolve them in a linear
         # fashion to ensure registration order is respected.
-        url_part = request.rel_url.raw_path
+        url_part = request.rel_url.path
         while url_part:
             for candidate in resource_index.get(url_part, ()):
                 match_dict, allowed = await candidate.resolve(request)
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -36,5 +36,5 @@ pycparser==2.22
     # via cffi
 uvloop==0.20.0 ; platform_system != "Windows" and implementation_name == "cpython"
     # via -r requirements/base.in
-yarl==1.9.5
+yarl==1.9.6
     # via -r requirements/runtime-deps.in
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -36,5 +36,5 @@ pycparser==2.22
     # via cffi
 uvloop==0.20.0 ; platform_system != "Windows" and implementation_name == "cpython"
     # via -r requirements/base.in
-yarl==1.9.4
+yarl==1.9.5
     # via -r requirements/runtime-deps.in
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -286,7 +286,7 @@ webcolors==24.8.0
     # via blockdiag
 wheel==0.44.0
     # via pip-tools
-yarl==1.9.5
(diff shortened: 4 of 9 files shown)


Dreamsorcerer:
I'm not too sure about this, but I think everything is working and passing if I make this change to yarl:
```diff
diff --git a/yarl/_quoting_py.py b/yarl/_quoting_py.py
index 585a1da..fc20b37 100644
--- a/yarl/_quoting_py.py
+++ b/yarl/_quoting_py.py
@@ -158,7 +158,7 @@ class _Unquoter:
                         if to_add is None:  # pragma: no cover
                             raise RuntimeError("Cannot quote None")
                         ret.append(to_add)
-                    elif unquoted in self._unsafe:
+                    elif unquoted in self._unsafe or unquoted == "/":
                         to_add = self._quoter(unquoted)
                         if to_add is None:  # pragma: no cover
                             raise RuntimeError("Cannot quote None")
```

Essentially, we shouldn't unquote `%2F` in the path, as this could lead to incorrectly treating it as a path separator.

bdraco:
> I'm not too sure about this, but I think everything is working and passing if I make this change to yarl:
> 
> ```diff
> diff --git a/yarl/_quoting_py.py b/yarl/_quoting_py.py
> index 585a1da..fc20b37 100644
> --- a/yarl/_quoting_py.py
> +++ b/yarl/_quoting_py.py
> @@ -158,7 +158,7 @@ class _Unquoter:
>                          if to_add is None:  # pragma: no cover
>                              raise RuntimeError("Cannot quote None")
>                          ret.append(to_add)
> -                    elif unquoted in self._unsafe:
> +                    elif unquoted in self._unsafe or unquoted == "/":
>                          to_add = self._quoter(unquoted)
>                          if to_add is None:  # pragma: no cover
>                              raise RuntimeError("Cannot quote None")
> ```
> 
> Essentially, we shouldn't unquote `%2F` in the path, as this could lead to incorrectly treating it as a path separator.

https://datatracker.ietf.org/doc/html/rfc3986#section-2.4

> When a URI is dereferenced, the components and subcomponents
   significant to the scheme-specific dereferencing process (if any)
   must be parsed and separated before the percent-encoded octets within
   those components can be safely decoded, as otherwise the data may be
   mistaken for component delimiters.

I read that as the URL needs to be separated by [delimiters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) before its decoded

Dreamsorcerer:
> I read that as the URL needs to be separated by [delimiters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.2) before its decoded

Hmm, I'm not sure how best to accommodate that...

bdraco:
https://www.w3.org/Addressing/URL/4_URI_Recommentations.html

> Example 2
> The URIs
> http://www.w3.org/albert/bertram/marie-claude
> 
> and
> http://www.w3.org/albert/bertram%2Fmarie-claude
>
> are NOT identical, as in the second case the encoded slash does not have hierarchical significance.

bdraco:
This gets messy quickly.. https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2007-0450

bdraco:
The safest course seems to never decode `%2F` to `/` when decoding a path

bdraco:
yarl treats `+` as unsafe in the unquoter but not `/`
https://github.com/aio-libs/yarl/blob/24073cec26aade02aa18f959e429a46e90cee001/yarl/_url.py#L165

but for the quoter `protected="/+"`
https://github.com/aio-libs/yarl/blob/24073cec26aade02aa18f959e429a46e90cee001/yarl/_url.py#L156

bdraco:
I think making `/` unsafe in `yarl` makes sense.... but its probably a breaking change

Dreamsorcerer:
> I think making `/` unsafe in `yarl` makes sense.... but its probably a breaking change

unsafe has a slightly different meaning, I tried that first.
`/my/path%2Ffoo` will become `%2Fmy%2Fpath%2Ffoo`

So, we'd probably want to add an ignore option or something separately, which would go where I hardcoded the value in the diff.

Dreamsorcerer:
Given the security implications, I suspect the change should be made regardless of backwards compatibility.

Dreamsorcerer:
https://github.com/aio-libs/yarl/pull/1057

bdraco:
Still haven't found a definitive answer on this one.....maybe we should be testing this against:
https://url.spec.whatwg.org/
https://github.com/web-platform-tests/wpt/tree/master/url

https://github.com/web-platform-tests/wpt/blob/master/url/resources/percent-encoding.json
https://github.com/web-platform-tests/wpt/blob/master/url/resources/urltestdata.json

Dreamsorcerer:
> Still haven't found a definitive answer on this one.....maybe we should be testing this against: https://url.spec.whatwg.org/ https://github.com/web-platform-tests/wpt/tree/master/url

Sounds like it could be a lot of work. Feel free to take that on, but I think this current solution is probably reasonable until that is done.

bdraco:
> > Still haven't found a definitive answer on this one.....maybe we should be testing this against: [url.spec.whatwg.org](https://url.spec.whatwg.org/) [web-platform-tests/wpt@`master`/url](https://github.com/web-platform-tests/wpt/tree/master/url?rgh-link-date=2024-08-28T12%3A00%3A26Z)
> 
> Sounds like it could be a lot of work. Feel free to take that on, but I think this current solution is probably reasonable until that is done.

I got a bit of a chuckle out of that. Its definitely a lot of work, and I wasn't intending for it to be done here.  While I think it would be great to test long term, I was thinking we could look this specific case is handled there as a reference point.

I plan on digging through it a bit more and provide some better analysis. Its Home Assistant beta week so I haven't had time to do that yet.

bdraco:
Will test this shortly

bdraco:
I'm hoping to find time to test this today

bdraco:
Everything looks like its working as expected.  Calling `path` is a bit heavier than `raw_path` but not much we can do about that.
Screenshot 2024-08-31 at 8 40 30 AM

The `cached_property` implementation in `yarl` is quite a bit slower than the `reify` implementation in `aiohttp`.  I noted that in https://github.com/aio-libs/yarl/issues/1065 .. maybe I need to do that sooner rather than later.

bdraco (approved):

PR #8908, opened 2024-08-26
Fix CancelledError stopping other cleanup contexts completing

### LINKED ISSUE #5672: await on the cancelled task will abort the application cleanup (py3.8+)
🐞 **Describe the bug**
If inside cleanup_ctx element you will use background tasks like this ([as in aiohttp docs](https://docs.aiohttp.org/en/stable/web_advanced.html#background-tasks))
```python
task = asyncio.create_task(...)
yield
task.cancel()
await task
```
Aiohttp will be aborted and no other contexts will be executed during the cleanup phase. This is because in python3.8 `asyncio.CancelledError` [changed](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) its ancestors: from `Exception` class to `BaseException`. 

💡 **To Reproduce**
```python
import asyncio
from aiohttp import web


async def some_coroutine():
    await asyncio.sleep(1000)


async def app_context2(app):
    print('Context #2: Starting...')
    some_task = asyncio.create_task(some_coroutine())
    yield
    some_task.cancel()
    await some_task  # will raise CancelledError  (expected)
    print('Context #2: Graceful shutdown')  # never prints (expected)


async def app_context1(app):
    print('Context #1: Starting...')
    yield
    print('Context #1: Graceful shutdown')  # prints in py3.7, but not in py3.8+ (unexpected)


def get_app():
    app = web.Application()
    app.cleanup_ctx.extend([app_context1, app_context2])
    return app


if __name__ == '__main__':
    web.run_app(get_app())
```

💡 **Expected behavior**
At the least, don't prevent processing from completing in other contexts.

📋 **Logs/tracebacks**
Python3.7.10
```python-traceback (paste your traceback in the next line)
Context #1: Starting...
Context #2: Starting...
(Press CTRL+C to quit)
^CContext #1: Graceful shutdown
```

Python3.8.4
```python-traceback (paste your traceback in the next line)
Context #1: Starting...
Context #2: Starting...
(Press CTRL+C to quit)
^C% 
```

```console
$ python --version
Python 3.7.10
```
```console
$ python --version
Python 3.8.4
```

Same for python 3.7.10 and 3.8.4
```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.7.4.post0
Author: Nikolay Kim
License: Apache 2
Location: /Users/<cropped>/lib/python3.8/site-packages
Requires: attrs, chardet, yarl, typing-extensions, multidict, async-timeout
Required-by:
```
```console
Name: multidict
Version: 5.1.0
Author: Andrew Svetlov
License: Apache 2
Location: /Users/<cropped>/venv3.8.4/lib/python3.8/site-packages
Requires: 
Required-by: yarl, aiohttp
```
```console
$ python -m pip show yarl
Name: yarl
Version: 1.6.3
Author: Andrew Svetlov
License: Apache 2
Location: /Users/<cropped>/venv3.8.4/lib/python3.8/site-packages
Requires: multidict, idna
Required-by: aiohttp
```

📋 **Additional context**

Relevant source [code](https://github.com/aio-libs/aiohttp/blob/4d58400477331704132cc9427d66699e5815ee25/aiohttp/web_app.py#L390) from master branch:
```python
    async def _on_cleanup(self, app: Application) -> None:
        errors = []
        for it in reversed(self._exits):
            try:
                await it.__anext__()
            except StopAsyncIteration:
                pass
            except Exception as exc:
                errors.append(exc)
            else:
                errors.append(RuntimeError(f"{it!r} has more than one 'yield'"))
        if errors:
            if len(errors) == 1:
                raise errors[0]
            else:
                raise CleanupError("Multiple errors on cleanup stage", errors)
```
----
PS. It's easy to revert the old behavior, just add the lines:
 ```python
try:
  ...
except asyncio.CancelledError as exc:
    errors.append(exc)
```
**But:**
Prior to python 3.8, with one such context nothing is output, but with two (or more) you will get `aiohttp.web_app.CleanupError: ('Multiple errors on cleanup stage', [CancelledError(), CancelledError()])`. It might be preferable to ignore the additional `CanceledError` in the cleanup phase?

### PR DESCRIPTION
**Pull request by Dreamsorcerer**
I wonder if we should even just catch BaseException here...?
Then a KeyboardInterrupt or similar could cancel each context individually.

Fixes #5672.

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -430,7 +430,7 @@ class CleanupContext(_CleanupContextBase):
                 await it.__anext__()
             except StopAsyncIteration:
                 pass
-            except Exception as exc:
+            except (Exception, asyncio.CancelledError) as exc:
                 errors.append(exc)
             else:
                 errors.append(RuntimeError(f"{it!r} has more than one 'yield'"))
--- a/CHANGES/8908.bugfix.rst
+++ b/CHANGES/8908.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed ``CancelledError`` from one cleanup context stopping other contexts from completng -- by :user:`Dreamsorcerer`.
--- a/CHANGES/8908.bugfix.rst
+++ b/CHANGES/8908.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed ``CancelledError`` from one cleanup context stopping other contexts from completng -- by :user:`Dreamsorcerer`.
+Fixed ``CancelledError`` from one cleanup context stopping other contexts from completing -- by :user:`Dreamsorcerer`.
--- a/docs/web_advanced.rst
+++ b/docs/web_advanced.rst
@@ -1068,13 +1068,10 @@ below::
       async with client.pubsub() as pubsub:
           await pubsub.subscribe(channel)
           while True:
-              try:
-                  msg = await pubsub.get_message(ignore_subscribe_messages=True)
-                  if msg is not None:
-                      for ws in app["websockets"]:
-                          await ws.send_str("{}: {}".format(channel, msg))
-              except asyncio.CancelledError:
-                  break
+              msg = await pubsub.get_message(ignore_subscribe_messages=True)
+              if msg is not None:
+                  for ws in app["websockets"]:
+                      await ws.send_str("{}: {}".format(channel, msg))
 
 
   async def background_tasks(app):
@@ -1083,7 +1080,8 @@ below::
       yield
 
       app[redis_listener].cancel()
-      await app[redis_listener]
+      with contextlib.suppress(asyncio.CancelledError):
+          await app[redis_listener]
 
 
   app = web.Application()
--- a/tests/test_web_app.py
+++ b/tests/test_web_app.py
@@ -334,9 +334,9 @@ async def test_cleanup_ctx_cleanup_after_exception() -> None:
     assert ctx_state == "CLEAN"
 
 
-@pytest.mark.parametrize("exc", (Exception, asyncio.CancelledError))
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

File: aiohttp/web_app.py [hunk 1]
```diff
@@ -430,7 +430,7 @@ async def _on_cleanup(self, app: Application) -> None:
                 await it.__anext__()
             except StopAsyncIteration:
                 pass
-            except Exception as exc:
+            except (Exception, asyncio.CancelledError) as exc:
```
bdraco, line 433: Is there a concern we would consume CancelledError if the current task was cancelling?

ie do we need?
```
if asyncio.current_task().cancelling() > 0:
   raise
```

Or is it ok to continue on ... maybe ok to consume it since its shutting down anyways and this isn't something we need to worry about?

Dreamsorcerer: I feel like if we've stopped the cleanup context from a cancellation, we'd still want to run the other contexts. If we were to reraise, I'd suggest doing after the loop. Not sure if that's really worth doing though.

My bigger question is should we catch BaseException, so for example, KeyboardInterrupt can interrupt each cleanup context one-at-a-time?

bdraco: Usually, when I catch keyboard interrupted, I log something about trying to continue to shut down gracefully, and let them press it again to shut down now.

Dreamsorcerer: Well, this is probably the second interrupt already. Question is whether it should require upto number of contexts + 1 to close, or stick with the current 2.

bdraco: I’d stick with the current two, as it’s confusing behavior to have to hit it many many times and the user might think something is broken, or hung

PR #8920, opened 2024-08-27
Re-enable keep-alive on proxies

### LINKED ISSUE #5765: Allow keep alive connection through the proxy
Once ago there was a merged pr - https://github.com/aio-libs/aiohttp/pull/3070 that disabled keep-alive connections through the proxy. There was a suggestion to make this functionality switchable in discussion.

Keep-alive with the proxy sometimes is need. E.g. to establish ntlm authorization the client must send 2 requests in one connection (a handshake). Current aiohttp logic restricts it, closing the connection on the first request.

I suggest to allow keep-alive connection when client explicitly sets keep-alive header.

Dreamsorcerer:
Fixes #5765.

This was disabled a long time ago in #3070. From what I've seen of various issues, the sentiment from users is that this causes problems and makes proxies too slow. asvetlov himself has suggested that modern proxies today may not have the issues he was experiencing when the change was made originally.

The original change had no test and I'm not familiar enough with the proxy code to write a new test. I suggest we put this into 3.11 and see if any users report regressions.

### DIFF
--- a/CHANGES/8920.misc.rst
+++ b/CHANGES/8920.misc.rst
@@ -1,0 +1,1 @@
+Enabled keep-alive support on proxies (which was originally disabled several years ago) -- by :user:`Dreamsorcerer`.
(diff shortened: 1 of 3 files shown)


bdraco:
`force_close` is being called externally

https://github.com/aio-libs/aiodocker/issues/918#issuecomment-2486278493

PR #8936, opened 2024-08-29
Test coverage of TextIOWrapper

Dreamsorcerer:


Dreamsorcerer:
Why does this fail on Windows...?

tests/test_formdata.py
```diff
@@ -48,6 +49,16 @@ def test_invalid_formdata_params2() -> None:
         FormData("as")  # 2-char str is not allowed
 
 
+async def test_formdata_textio_charset(buf: bytearray, writer: StreamWriter) -> None:
+    form = FormData()
+    body = io.TextIOWrapper(io.BytesIO(b"\xe6\x97\xa5\xe6\x9c\xac"))
```
Dreamsorcerer: Seem TextIOWrapper defaults to local encoding, just to be different from bytes.decode()...

```suggestion
    body = io.TextIOWrapper(io.BytesIO(b"\xe6\x97\xa5\xe6\x9c\xac"), encoding="utf-8")
```

Issue #6764, opened 2022-05-26
Client sends body after being redirected


luni3359:
### Describe the bug

aiohttp keeps sending the body every time it's redirected.

### To Reproduce

1. Send body data via a GET request
2. Site acknowledges data and redirects the client
3. After redirection, aiohttp sends the body again

### Expected behavior

aiohttp shouldn't send the body after being redirected. At the very least `requests` doesn't behave in this way and avoids sending the body a second time.

### Logs/tracebacks

Minimum reproducible code:
```python
import asyncio
import json
import aiohttp

headers = {'Content-Type': 'application/json'}
data = {
    "tags": "rating:general chibi",
    "random": True
}

async def main():
    async with aiohttp.ClientSession() as session:
        async with session.get("https://danbooru.donmai.us/posts.json", data=json.dumps(data), headers=headers) as response:
            response_payload = await response.read()
            print(response.status)
            print(response_payload)

asyncio.run(main())
```
```python-traceback
Traceback (most recent call last):
  File "/home/luni3359/.local/share/pyenv/versions/3.10.4/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/luni3359/.local/share/pyenv/versions/3.10.4/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/luni3359/.vscode/extensions/ms-python.python-2022.6.2/pythonFiles/lib/python/debugpy/__main__.py", line 45, in <module>
    cli.main()
  File "/home/luni3359/.vscode/extensions/ms-python.python-2022.6.2/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 444, in main
    run()
  File "/home/luni3359/.vscode/extensions/ms-python.python-2022.6.2/pythonFiles/lib/python/debugpy/../debugpy/server/cli.py", line 285, in run_file
    runpy.run_path(target_as_str, run_name=compat.force_str("__main__"))
  File "/home/luni3359/.local/share/pyenv/versions/3.10.4/lib/python3.10/runpy.py", line 269, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/home/luni3359/.local/share/pyenv/versions/3.10.4/lib/python3.10/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
[...]

PR #8966, opened 2024-09-01
Fix auth reset logic during redirects to different origin when _base_url set

MaximZemskov:
## What do these changes do?

Do not set ClientSession’s auth during redirects to a different origin when _base_url is set.

It might affect users who expect ClientSession to retain auth during redirects to different origins.

No.

## Related issue number

#6764 
[#8953 ](https://github.com/aio-libs/aiohttp/pull/8953#issuecomment-2322922080)

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


aiohttp/client.py
```diff
@@ -543,7 +544,13 @@ async def _request(
...
                     if auth is None:
                         auth = auth_from_url
-                    if auth is None:
+
+                    # If the session has a _base_url set and this is a redirect to a different origin
+                    # we should not set the auth
+                    is_auth_reset_needed = (
+                        is_redirect_to_different_origin and self._base_url is not None
+                    )
+                    if auth is None and not is_auth_reset_needed:
```
Dreamsorcerer: Rather than tracking this redirect variable, I'm wondering if we should just do something like `(not self._base_url or self._base_url.origin() == url.origin())`?

I was also thinking that maybe later we might change the base_url behaviour so that you can still make requests to absolute URLs. If we do that, then this check would ensure that we're not sending the auth when we request a different domain to what is defined in the base_url.

MaximZemskov: Yes, that makes sense. Fixed it.

aiohttp/client.py
```diff
@@ -543,7 +543,10 @@ async def _request(
 
                     if auth is None:
                         auth = auth_from_url
-                    if auth is None:
+
+                    if auth is None and (
+                        not self._base_url or self._base_url.origin() == url.origin()
```
bdraco, line 548: Calling `.origin()` is a bit expensive https://github.com/aio-libs/aiohttp/issues/7583#issuecomment-1806828707

We should probably cache `self._base_url.origin()` as self._base_url_origin` so we don't have to build it every time

bdraco: Probably could guard this with `self._default_auth` being set as well

bdraco: https://github.com/aio-libs/aiohttp/pull/9466

PR #8977, opened 2024-09-01
Add Generic for TestClient.app

### LINKED ISSUE #7226: TestClient.app can be None
A minor nuisance, but `TestClient.app` is `Application | None`.
When type checking in mypy this results in a bunch of errors unless extra assert calls are made.

Maybe we can make it generic or similar, so when passing an Application to the aiohttp_client fixture, for example, it will know that the attribute will not be None.

Dreamsorcerer:
This should avoid annoying `assert client.app is not None` lines in test code.

Fixes #7226.


CHANGES/8977.bugfix.rst
```diff
@@ -0,0 +1 @@
+Made ``TestClient.app`` a ``Generic`` so type checkers will know the correct type (avoid unneeded ``client.app is not None`` checks) -- by :user:`Dreamsorcerer`.
```
Dreamsorcerer:
```suggestion
Made ``TestClient.app`` a ``Generic`` so type checkers will know the correct type (avoiding unneeded ``client.app is not None`` checks) -- by :user:`Dreamsorcerer`.
```

bdraco (approved):

Dreamsorcerer:
#9245

PR #9027, opened 2024-09-05
Allow aiodns on Windows

### LINKED ISSUE #8121: aiodns can probably be enabled on windows, now that it ships with c-ares already built

The `aiodns` library from the `[speedup] `bundle is currently disabled on Windows per [this line ](https://github.com/aio-libs/aiohttp/blob/master/setup.cfg#L64).

However, `aiodns` currently ships with `pycares`, which bundles a binary of `c-ares` by default. I was just now able to install it on Win10 with pip.

It seems there's no need to exclude Windows from the aiodns speedup. Looks like changing https://github.com/aio-libs/aiohttp/blob/master/setup.cfg#L64 would be sufficient, but I'm by no means experienced, so I'd rather not mess with the repo myself.

There seems to be some requests to get rid of aiodns entirely in favour of pydns, but maybe till then we can use it like this?

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #8121.

### DIFF
--- a/CHANGES/9027.misc.rst
+++ b/CHANGES/9027.misc.rst
@@ -1,0 +1,1 @@
+Include aiodns in `[speedups]` for Windows -- by :user:`Dreamsorcerer`.
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --allow-unsafe --output-file=requirements/base.txt --strip-extras requirements/base.in
 #
-aiodns==3.2.0 ; sys_platform == "linux" or sys_platform == "darwin"
+aiodns==3.2.0
     # via -r requirements/runtime-deps.in
 aiohappyeyeballs==2.4.0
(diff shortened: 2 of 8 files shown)


bdraco:
Protractor does not work with aiodns Last time I checked

bdraco:
We probably need a guard to verify they’re using a selector event loop

Dreamsorcerer:
Proactor is default right? So, test should fail..

webknjaz:
@Dreamsorcerer @bdraco `aiodns` is now under our umbrella FYI: https://github.com/aio-libs/aiodns

CHANGES/9027.misc.rst
```diff
@@ -0,0 +1 @@
+Include aiodns in `[speedups]` for Windows -- by :user:`Dreamsorcerer`.
```
webknjaz, line 1: @Dreamsorcerer FYI RST uses double backticks for this

PR #9029, opened 2024-09-05
Fix SSLContext creation in the TCPConnector with multiple loops

### LINKED ISSUE #9020: Incorrect re-use future in _make_or_get_ssl_context
### Describe the bug

I meet following error:
```
RuntimeError: Task <Task pending name='Task-2' coro=<async_connect() running at <ipython-input-1-ba65b5596d34>:18>> got Future <Future pending> attached to a different loop
```
After going through the code, it seems the bug is caused by `aiohttp/connector.py::_make_or_get_ssl_context`. There is a global cache _made_ssl_context for asyncio.Future[SSLContext]. However, the cached future is bound to the first loop. When a new call with different loop to _make_or_get_ssl_context, the cached future is not expected because it has the old loop.

The bug exists since 3.10

### To Reproduce

Please see the description.

### Expected behavior

The calls to `client_session.ws_connect(url)` by multiple threads should not throw exception

### Logs/tracebacks

```python-traceback
RuntimeError: Task <Task pending name='Task-2' coro=<async_connect() running at <ipython-input-1-ba65b5596d34>:18>> got Future <Future pending> attached to a different loop
```

### Python Version

```console
$ python --version
3.8
```

### aiohttp Version

```console
$ python -m pip show aiohttp
3.10.5
```

### multidict Version

```console
n/a
```

### yarl Version

```console
$ python -m pip show yarl
n/a
```

### OS

Linux

### Related component

Client

### Additional context

N/A

### Code of Conduct

bdraco:
Creation of the SSLContexts did not account for multiple event loops in different threads.

Creation is now done at import time to ensure it does not block the event loop.

fixes #9020

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -984,7 +984,7 @@ class TCPConnector(BaseConnector):
 
         return proto
 
-    async def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
+    def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
         """Logic to get the correct SSL context
 
         0. if req.ssl is false, return None
@@ -1108,7 +1108,7 @@ class TCPConnector(BaseConnector):
         # `req.is_ssl()` evaluates to `False` which is never gonna happen
         # in this code path. Of course, it's rather fragile
         # maintainability-wise but this is to be solved separately.
-        sslcontext = cast(ssl.SSLContext, await self._get_ssl_context(req))
+        sslcontext = cast(ssl.SSLContext, self._get_ssl_context(req))
 
         try:
             async with ceil_timeout(
@@ -1186,7 +1186,7 @@ class TCPConnector(BaseConnector):
         *,
         client_error: Type[Exception] = ClientConnectorError,
     ) -> Tuple[asyncio.Transport, ResponseHandler]:
-        sslcontext = await self._get_ssl_context(req)
+        sslcontext = self._get_ssl_context(req)
         fingerprint = self._get_fingerprint(req)
 
         host = req.url.raw_host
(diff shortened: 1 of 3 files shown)


bdraco:
Probably easier to fix the issue and include a test then try to suppress all the asyncio errors that pytest really wants to raise

tests/test_connector.py
```diff
@@ -9,13 +9,15 @@
 import sys
 import uuid
 from collections import deque
+from concurrent.futures import ThreadPoolExecutor, as_completed
```
Dreamsorcerer: I'd suggest import `futures` given that as_completed() also exists in asyncio and could cause confusion.

Dreamsorcerer:
I'm not overly convinced we need to support multiple loops, but you didn't change the import_time test, so it can't be that slow to do this on import.

Dreamsorcerer (approved):

aiohttp/connector.py
```diff
@@ -963,24 +984,6 @@ async def _create_connection(
...
-        sslcontext = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-        sslcontext.options |= ssl.OP_NO_SSLv2
-        sslcontext.options |= ssl.OP_NO_SSLv3
-        sslcontext.check_hostname = False
-        sslcontext.verify_mode = ssl.CERT_NONE
-        sslcontext.options |= ssl.OP_NO_COMPRESSION
-        sslcontext.set_default_verify_paths()
-        return sslcontext
-
     async def _get_ssl_context(self, req: ClientRequest) -> Optional[SSLContext]:
```
bdraco: Nothing is being awaited here anymore. This doesn't need to be a Coro

bdraco:
Need to check the missing coverage

bdraco:
I'm not sure why I added skip on this one. I should have done a changelog message

bdraco:
I'll do another PR to add it

PR #9052, opened 2024-09-07
Fix lost details on HttpProcessingError

### LINKED ISSUE #8395: Error message not always propagated on 3.9.4
### Describe the bug

```python
from aiohttp import web
from aiohttp.client_exceptions import ClientResponseError

async def handler(_: web.Request) -> web.Response:
    return web.Response(headers={"Custom": "x" * 10000})

app = web.Application()
app.add_routes([web.get("/", handler)])

try:
    await client.get("/")
except ClientResponseError as e:
    assert e.status == 400
    assert "Got more than 8190 bytes" in e.message
else:
    raise AssertionError("Expected ClientResponseError")
```

In 3.9.3 this would work and the `ClientResponseError` thrown would look like `400, message='Got more than 8190 bytes (10000) when reading Header value is too long.', url=URL('http://127.0.0.1:51163/')`.
Starting in 3.9.4 however an empty `ClientResponseError` seems to be raised `0, message='', url=URL('http://127.0.0.1:51193/')`.

I can see that the expected error message is thrown up the stack (specifically in `aiohttp/client_proto.py`) but the actual information about it is lost when it reaches `aiohttp/client_reqrep.py`.

```python-traceback
    def data_received(self, data: bytes) -> None:
        self._reschedule_timeout()

        if not data:
            return

        # custom payload parser
            eof, tail = self._payload_parser.feed_data(data)
            if eof:
                self._payload = None
                self._payload_parser = None

                if tail:
                    self.data_received(tail)
            return
        else:
            if self._upgraded or self._parser is None:
                # i.e. websocket connection, websocket parser is not set yet
                self._tail += data
            else:
                # parse http messages
                try:
>                   messages, upgraded, tail = self._parser.feed_data(data)

../../.local/share/pyenv/versions/3.11.8/envs/mono-3.11.8/lib/python3.11/site-packages/aiohttp/client_proto.py:256:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
aiohttp/_http_parser.pyx:557: in aiohttp._http_parser.HttpParser.feed_data
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   aiohttp.http_exceptions.LineTooLong: 400, message:
E     Got more than 8190 bytes (10000) when reading Header value is too long.

aiohttp/_http_parser.pyx:732: LineTooLong

self = <ClientResponse(http://127.0.0.1:53889/) [None None]>
None

connection = Connection<ConnectionKey(host='127.0.0.1', port=53889, is_ssl=False, ssl=True, proxy=None, proxy_auth=None, proxy_headers_hash=None)>

    async def start(self, connection: "Connection") -> "ClientResponse":
        """Start response processing."""
        self._closed = False
        self._protocol = connection.protocol
        self._connection = connection

        with self._timer:
            while True:
                # read response
                try:
                    protocol = self._protocol
>                   message, payload = await protocol.read()  # type: ignore[union-attr]

../../.local/share/pyenv/versions/3.11.8/envs/mono-3.11.8/lib/python3.11/site-packages/aiohttp/client_reqrep.py:976:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <aiohttp.client_proto.ResponseHandler object at 0x104941240>

    async def read(self) -> _T:
        if not self._buffer and not self._eof:
            assert not self._waiter
            self._waiter = self._loop.create_future()
            try:
>               await self._waiter
E               aiohttp.http_exceptions.HttpProcessingError: 0, message:

../../.local/share/pyenv/versions/3.11.8/envs/mono-3.11.8/lib/python3.11/site-packages/aiohttp/streams.py:640: HttpProcessingError
```

### To Reproduce

See repro case above.

### Expected behavior

I expect `ClientResponseError` to have a 400 status rather than 0 and the message to be populated with the actual error.

### Logs/tracebacks

```python-traceback
See description above.
```

### Python Version

```console
3.11.8
```

### aiohttp Version

```console
Name: aiohttp
Version: 3.9.4
Author:
Author-email:
License: Apache 2
Location: /Users/damien/.local/share/pyenv/versions/mono-3.11.8/lib/python3.11/site-packages
```

### multidict Version

```console
Name: multidict
Version: 6.0.5
Author: Andrew Svetlov
License: Apache 2
Location: /Users/damien/.local/share/pyenv/versions/mono-3.11.8/lib/python3.11/site-packages
```

### yarl Version

```console
Name: yarl
Version: 1.9.4
Author: Andrew Svetlov
License: Apache-2.0
Location: /Users/damien/.local/share/pyenv/versions/mono-3.11.8/lib/python3.11/site-packages
```

### OS

macos sonoma 14.4.1

### Related component

Client

### Additional context

_No response_

### Code of Conduct

- [x] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
Fixes #8395.

### DIFF
--- a/CHANGES/9052.bugfix.rst
+++ b/CHANGES/9052.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed some exception information getting lost -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9052.bugfix.rst
+++ b/CHANGES/9052.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed exception information getting lost on :exc:`HttpProcessingError` -- by :user:`Dreamsorcerer`.
+Fixed exception information getting lost on ``HttpProcessingError`` -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9052.bugfix.rst
+++ b/CHANGES/9052.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed some exception information getting lost -- by :user:`Dreamsorcerer`.
+Fixed exception information getting lost on :exc:`HttpProcessingError` -- by :user:`Dreamsorcerer`.
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -275,18 +275,21 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
                 # parse http messages
                 try:
                     messages, upgraded, tail = self._parser.feed_data(data)
-                except HttpProcessingError as underlying_exc:
+                except BaseException as underlying_exc:
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
                     # should_close is True after the call
-                    exc = HttpProcessingError(
-                        code=underlying_exc.code,
-                        message=underlying_exc.message,
-                        headers=underlying_exc.headers,
-                    )
+                    if isinstance(underlying_exc, HttpProcessingError):
+                        exc = HttpProcessingError(
+                            code=underlying_exc.code,
+                            message=underlying_exc.message,
+                            headers=underlying_exc.headers,
+                        )
+                    else:
+                        exc = HttpProcessingError()
                     self.set_exception(exc, underlying_exc)
                     return
 
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -283,7 +283,11 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
                         self.transport.close()
                     # should_close is True after the call
                     if isinstance(underlying_exc, HttpProcessingError):
-                        exc = HttpProcessingError(code=underlying_exc.code, message=underlying_exc.message, headers=underlying_exc.headers)
+                        exc = HttpProcessingError(
+                            code=underlying_exc.code,
+                            message=underlying_exc.message,
+                            headers=underlying_exc.headers,
+                        )
                     else:
                         exc = HttpProcessingError()
                     self.set_exception(exc, underlying_exc)
--- a/aiohttp/client_proto.py
+++ b/aiohttp/client_proto.py
@@ -275,21 +275,18 @@ class ResponseHandler(BaseProtocol, DataQueue[Tuple[RawResponseMessage, StreamRe
                 # parse http messages
                 try:
                     messages, upgraded, tail = self._parser.feed_data(data)
-                except BaseException as underlying_exc:
+                except HttpProcessingError as underlying_exc:
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
                     # should_close is True after the call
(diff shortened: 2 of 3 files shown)


Dreamsorcerer:
Our codecov doesn't actually include any other types of exceptions...

Dreamsorcerer:
There doesn't appear to be any reason we should be catching other exceptions...

bdraco (approved):

aiohttp/client_proto.py
```diff
@@ -275,14 +275,19 @@ def data_received(self, data: bytes) -> None:
                 # parse http messages
                 try:
                     messages, upgraded, tail = self._parser.feed_data(data)
-                except BaseException as underlying_exc:
+                except HttpProcessingError as underlying_exc:
                     if self.transport is not None:
                         # connection.release() could be called BEFORE
                         # data_received(), the transport is already
                         # closed in this case
                         self.transport.close()
```
webknjaz, line 283: I feel like that `BaseException` was here to make sure to close transport in case of unexpected crashes, right? Is it possible for some non-matching exception to sneak in through this code path https://github.com/aio-libs/aiohttp/blob/bf022b3/aiohttp/_http_parser.pyx#L551C20-L551C36 ?

CHANGES/9052.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed some exception information getting lost -- by :user:`Dreamsorcerer`.
```
webknjaz: Can we get more details on what this looks like for the end-users?

aiohttp/client_proto.py
Dreamsorcerer: Hmm, it does catch BaseException in the parser, but I don't see any reason we'd have an exception there...

Dreamsorcerer: It feels a little awkward, but guess we'll keep it like that for now.

Dreamsorcerer: We're obviously missing coverage now, because there's no known situation where this could happen though..

PR #9053, opened 2024-09-07
Explain the importance of reproducers

Dreamsorcerer:


bdraco (approved):

bdraco:
Not sure if backport matters here as github only looks at master for the issue templates

webknjaz:
> Not sure if backport matters here as github only looks at master for the issue templates

Yes. But it may cause problems when backporting other PRs with functional changes that happen to also touch these files for whatever reason…

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -35,6 +35,8 @@ body:
       1. Implement the following server or a client '...'
       2. Then run '...'
       3. An error occurs.
+      The chances of someone looking at your issue are *vastly* improved if you provide
```
webknjaz, line 38: Personally, I'd put it into the description of the field rather than the placeholder that disappears as soon as somebody clicks on it... check the UX here https://github.com/aio-libs/aiohttp/issues/new?assignees=aio-libs%2Ftriagers&labels=bug&projects=&template=bug_report.yml

Dreamsorcerer: Feel free to adjust. I'm not sure how I'd format it into the description.

PR #9062, opened 2024-09-08
DNM: Test why update_query was not used

bdraco:
DO NOT MERGE

Exploring if we need an `extend_query`, we are missing test coverage, or `update_query` can be used here.

related issue: https://github.com/aio-libs/yarl/issues/1127


bdraco:
I expected the tests to fail

bdraco:
I will add some more coverage here so we can tell if this is an OK change or we need `extend_query` in yarl... I'd really like to get rid of the double query string build since its one of the most expensive parts of the lib

PR #9064, opened 2024-09-08
Fix sendfile test fixture

### LINKED ISSUE #8971: test_web_sendfile_functional isn't testing sendfile
We have this code in a fixture:
https://github.com/aio-libs/aiohttp/blob/8a525d98c512e624311b16e9891cac9050cc8b8c/tests/test_web_sendfile_functional.py#L75-L76

This looked rather dodgy as it appears, at first glance, to be changing the event loop in the middle of a handler, which you'd expect to break a lot of things.
However, in reality, the `loop_without_sendfile` fixture modifies the object from the `loop` fixture, which will be the default loop currently running. Which means that the loop has always been modified and this set_event_loop() call is a no-op.

That means the fixture is causing all the tests to be run twice under the exact same conditions. There is no test case where the sendfile has not been patched out. The fixtures need to be reworked to make this behave correctly, and a regression test for the fixture needs to be added.

Dreamsorcerer:
Not sure this is the best way to do it, but certainly better than before.

Fixes #8971.

### DIFF
--- a/tests/test_web_sendfile_functional.py
+++ b/tests/test_web_sendfile_functional.py
@@ -6,6 +6,7 @@ import pathlib
 import socket
 import zlib
 from typing import Any, Iterable, Optional
+from unittest import mock
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


bdraco (approved):

bdraco:
https://github.com/aio-libs/aiohttp/pull/9224

PR #9068, opened 2024-09-08
Use URL.extend_query to add params in `ClientRequest`

bdraco:
- [x] needs yarl bumped before this can merge so we test both paths https://github.com/aio-libs/aiohttp/pull/9072

## What do these changes do?

Use `URL.extend_query` to add `params` in `ClientRequest`

no

no

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -27,7 +27,7 @@ from typing import (
 )
 
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
-from yarl import URL
+from yarl import URL, __version__ as yarl_version
 
 from . import hdrs, helpers, http, multipart, payload
 from .abc import AbstractStreamWriter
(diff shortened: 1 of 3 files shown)


aiohttp/client_reqrep.py
```diff
@@ -90,6 +90,7 @@
 
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
+_YARL_SUPPORTS_EXTEND_QUERY = tuple(map(int, yarl_version.split(".")[:2])) > (1, 10)
```
```suggestion
_YARL_SUPPORTS_EXTEND_QUERY = tuple(map(int, yarl_version.split(".")[:2])) >= (1, 11)
```

Dreamsorcerer (approved):
I'd be tempted to just move this to 3.11 and increase the minimum yarl version (or do this in 3.10 and remove the old path in 3.11), as these kind of checks tend to be forgotten about for years after they're useful. We can do the same for using the Query type.

bdraco:
I'll backport this to 3.10 and 3.11 and do another PR to increment the yarl version for master/3.11 and cleanup the back compat

bdraco:
probably should merge https://github.com/aio-libs/aiohttp/pull/8564 first and than strip it out of 3.11 and master in another step once yarl requirement is bumped

bdraco:
Pretty significant drop on the profile 👍 
Screenshot 2024-09-08 at 9 11 58 PM

PR #9083, opened 2024-09-09
Remove unused backwards compatibility code for old yarl versions

bdraco:
## What do these changes do?

https://github.com/aio-libs/aiohttp/pull/9068#issuecomment-2336844811
Remove unused backwards compatibility code for old yarl versions
#9079 increased the yarl minimum version so we can remove these shims

no


bdraco:
We ended up having to bump yarl in 3.10 for path_safe anyways so we should backport this as well.

PR #9095, opened 2024-09-09
Reduce overhead to check if a host is an IP Address

bdraco:
## What do these changes do?

The IP Address regexes are a bit complex and take up ~35% of the time spent in `update_headers` https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client_reqrep.py#L361   

The checks for IP Addresses did not need to validate that the IP Address is valid, only that it was not a domain name https://github.com/aio-libs/aiohttp/pull/9095#issuecomment-2338631993 which means they can be much simpler.

Timings (more details below)
```
['ipv4 string - decode + isdigit', 0.05164358299225569]
['ipv4 string - regex', 0.24710870906710625]
['ipv4 binary - decode + isdigit', 0.07746362499892712]
['ipv4 binary - regex', 0.2532113748602569]
['ipv6 string - check for :', 0.00942283309996128]
['ipv6 string - regex', 1.4924992499873042]
['ipv6 binary - check for :', 0.11242458294145763]
['ipv6 binary - regex', 1.4484847909770906]
```

no

no

before
is_ip_address_before

after
is_ipv4_address_after
is_ip_address_after

### DIFF
--- a/CHANGES/9095.misc.rst
+++ b/CHANGES/9095.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of checking if a host is an IP Address -- by :user:`bdraco`.
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -483,56 +483,51 @@ try:
 except ImportError:
     pass
 
-_ipv4_pattern = (
-    r"^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}"
-    r"(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
-)
-_ipv6_pattern = (
-    r"^(?:(?:(?:[A-F0-9]{1,4}:){6}|(?=(?:[A-F0-9]{0,4}:){0,6}"
-    r"(?:[0-9]{1,3}\.){3}[0-9]{1,3}$)(([0-9A-F]{1,4}:){0,5}|:)"
-    r"((:[0-9A-F]{1,4}){1,5}:|:)|::(?:[A-F0-9]{1,4}:){5})"
-    r"(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}"
-    r"(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])|(?:[A-F0-9]{1,4}:){7}"
-    r"[A-F0-9]{1,4}|(?=(?:[A-F0-9]{0,4}:){0,7}[A-F0-9]{0,4}$)"
-    r"(([0-9A-F]{1,4}:){1,7}|:)((:[0-9A-F]{1,4}){1,7}|:)|(?:[A-F0-9]{1,4}:){7}"
-    r":|:(:[A-F0-9]{1,4}){7})$"
-)
-_ipv4_regex = re.compile(_ipv4_pattern)
-_ipv6_regex = re.compile(_ipv6_pattern, flags=re.IGNORECASE)
-_ipv4_regexb = re.compile(_ipv4_pattern.encode("ascii"))
-_ipv6_regexb = re.compile(_ipv6_pattern.encode("ascii"), flags=re.IGNORECASE)
-
 
 def is_ipv4_address(host: Optional[Union[str, bytes]]) -> bool:
-    """Check if host is an valid IPv4 address."""
+    """Check if host looks like an IPv4 address.
+
+    This function does not validate that the format is correct, only that
+    the host is a str or bytes, and its all numeric.
+
+    This check is only meant as a heuristic to ensure that
+    a host is not a domain name.
+    """
     if not host:
         return False
+    # For a host to be an ipv4 address, it must be all numeric.
     if isinstance(host, str):
-        # The last character of the host must be a digit to be an IPv4 address.
-        # We check this first to avoid the overhead of the regex.
-        return host[-1].isdigit() and bool(_ipv4_regex.match(host))
+        return host.replace(".", "").isdigit()
     if isinstance(host, (bytes, bytearray, memoryview)):
-        return bool(_ipv4_regexb.match(host))
+        return host.decode("ascii").replace(".", "").isdigit()
     raise TypeError(f"{host} [{type(host)}] is not a str or bytes")
 
 
 def is_ipv6_address(host: Optional[Union[str, bytes]]) -> bool:
-    """Check if host is an valid IPv6 address."""
+    """Check if host looks like an IPv6 address.
+
+    This function does not validate that the format is correct, only that
+    the host contains a colon and that it is a str or bytes.
+
+    This check is only meant as a heuristic to ensure that
+    a host is not a domain name.
+    """
     if not host:
         return False
+    # The host must contain a colon to be an IPv6 address.
     if isinstance(host, str):
-        # The host must contain a colon to be an IPv6 address.
-        # We check this first to avoid the overhead of the regex.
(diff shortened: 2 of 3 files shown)


Dreamsorcerer:
If this is a performance concern, wouldn't it make more sense to make this a much simpler heuristic? I believe we're only trying to decide if the user is attempting a request via an IP or a domain. There doesn't seem to be any real need to validate that the syntax is correct.

e.g. If we invert it to check if it looks like a domain, we could probably get away with something like `if ":" in host or host.rsplit(".", maxsplits=1)[-1].isdigit()`. Then it shouldn't be a domain according to the DNS syntax: https://www.rfc-editor.org/rfc/rfc1034#section-3.5

bdraco:
> If this is a performance concern, wouldn't it make more sense to make this a much simpler heuristic? I believe we're only trying to decide if the user is attempting a request via an IP or a domain. There doesn't seem to be any real need to validate that the syntax is correct.
> 
> e.g. If we invert it to check if it looks like a domain, we could probably get away with something like `if ":" in host or host.rsplit(".", maxsplit=1)[-1][0].isdigit()`. Then it shouldn't be a domain according to the DNS syntax: [rfc-editor.org/rfc/rfc1034#section-3.5](https://www.rfc-editor.org/rfc/rfc1034#section-3.5)

Yeah I think that makes a lot more sense.  I'll audit all usage to see if we actually care if its valid or not

bdraco:
is_ip_address is only used in cookies
https://github.com/aio-libs/aiohttp/blob/2b5a4b9d387d6a3f78d35c88ca68c38ff3725398/aiohttp/cookiejar.py#L180
https://github.com/aio-libs/aiohttp/blob/2b5a4b9d387d6a3f78d35c88ca68c38ff3725398/aiohttp/cookiejar.py#L339
https://github.com/aio-libs/aiohttp/blob/2b5a4b9d387d6a3f78d35c88ca68c38ff3725398/aiohttp/cookiejar.py#L282

is_ipv4_address is never called directly

is_ipv6_address is only used in client_reqrep to add `[]`
https://github.com/aio-libs/aiohttp/blob/2b5a4b9d387d6a3f78d35c88ca68c38ff3725398/aiohttp/client_reqrep.py#L361
https://github.com/aio-libs/aiohttp/blob/2b5a4b9d387d6a3f78d35c88ca68c38ff3725398/aiohttp/client_reqrep.py#L615

I don't think any of the use cases need to validate

bdraco:
Its the IPv6 regex that is heavy

```
['ipv4 string - decode + isdigit', 0.05164358299225569]
['ipv4 string - regex', 0.24710870906710625]
['ipv4 binary - decode + isdigit', 0.07746362499892712]
['ipv4 binary - regex', 0.2532113748602569]
['ipv6 string - check for :', 0.00942283309996128]
['ipv6 string - regex', 1.4924992499873042]
['ipv6 binary - check for :', 0.11242458294145763]
['ipv6 binary - regex', 1.4484847909770906]
```

```python
import re
import timeit

_ipv4_pattern = (
    r"^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}"
    r"(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$"
)

_ipv6_pattern = (
    r"^(?:(?:(?:[A-F0-9]{1,4}:){6}|(?=(?:[A-F0-9]{0,4}:){0,6}"
    r"(?:[0-9]{1,3}\.){3}[0-9]{1,3}$)(([0-9A-F]{1,4}:){0,5}|:)"
    r"((:[0-9A-F]{1,4}){1,5}:|:)|::(?:[A-F0-9]{1,4}:){5})"
    r"(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}"
    r"(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])|(?:[A-F0-9]{1,4}:){7}"
    r"[A-F0-9]{1,4}|(?=(?:[A-F0-9]{0,4}:){0,7}[A-F0-9]{0,4}$)"
    r"(([0-9A-F]{1,4}:){1,7}|:)((:[0-9A-F]{1,4}){1,7}|:)|(?:[A-F0-9]{1,4}:){7}"
    r":|:(:[A-F0-9]{1,4}){7})$"
)
_ipv4_regex = re.compile(_ipv4_pattern)
_ipv6_regex = re.compile(_ipv6_pattern, flags=re.IGNORECASE)
_ipv4_regexb = re.compile(_ipv4_pattern.encode("ascii"))
_ipv6_regexb = re.compile(_ipv6_pattern.encode("ascii"), flags=re.IGNORECASE)

host_str = "1.2.3.4"
print (['ipv4 string - decode + isdigit', timeit.timeit('host_str.replace(".","").isdigit()',globals=locals())])
print (['ipv4 string - regex', timeit.timeit('_ipv4_regex.search(host_str)',globals=locals())])
host_bin = b"1.2.3.4"
print(['ipv4 binary - decode + isdigit', timeit.timeit('host_bin.decode("ascii").replace(".","").isdigit()',globals=locals())])
print(['ipv4 binary - regex', timeit.timeit('_ipv4_regexb.search(host_bin)',globals=locals())])
host_str = "2001:db8::ff00:42:8329"
print(['ipv6 string - check for :', timeit.timeit('":" in host_str',globals=locals())])
print(['ipv6 string - regex', timeit.timeit('_ipv6_regex.search(host_str)',globals=locals())])
host_bin = b"2001:db8::ff00:42:8329"
print(['ipv6 binary - check for :', timeit.timeit('b":" in host_bin',globals=locals())])
print(['ipv6 binary - regex', timeit.timeit('_ipv6_regexb.search(host_bin)',globals=locals())])

```

Dreamsorcerer:
> ```python
> host_str.replace(".","").isdigit()
> ```

Strictly speaking, you could even do `host_str[0].isdigit()`, though I was initially thinking of checking the TLD specifically, to be on the safer side.

bdraco:
> > ```python
> > host_str.replace(".","").isdigit()
> > ```
> 
> Strictly speaking, you could even do `host_str[0].isdigit()`, though I was initially thinking of checking the TLD specifically, to be on the safer side.

I don't think we can do that since domains are allowed to start with a number `1password.com`

I'm not sure there any TLDs that end with a number so `host_str[-1].isdigit()` might work... but I wonder if that could change

Dreamsorcerer:
> I don't think we can do that since domains are allowed to start with a number `1password.com`

Oh, that actually violates the DNS syntax in the spec...

Dreamsorcerer:
> > I don't think we can do that since domains are allowed to start with a number `1password.com`
> 
> Oh, that actually violates the DNS syntax in the spec...

Nevermind, it was updated in https://www.rfc-editor.org/rfc/rfc1101 (section 3.1)

bdraco:
> > > > I don't think we can do that since domains are allowed to start with a number `1password.com`
> > > 
> > > 
> > > Oh, that actually violates the DNS syntax in the spec...
> > 
> > 
> > Nevermind, it was updated in [rfc-editor.org/rfc/rfc1101](https://www.rfc-editor.org/rfc/rfc1101) (section 3.1)
> 
> It says that `26.0.0.73.COM` is not a valid domain, but not clear how it reaches that conclusion.. But, I suspect what you've got will be good enough.

`host[-1].isdigit()` would be a tiny bit faster but considering we are going from 1.4924992499873042 to 0.00942283309996128 for the IPv6 check it probably doesn't matter so much as the IPv4 case wasn't so much of a problem

bdraco:
Tested on a few production Home Assistant instances.  There are lots of places that use cookies and ip addresses so this has some good real world use testing.

PR #9098, opened 2024-09-10
Remove extra ``MultiDictProxy`` wrapper from ``BaseRequest.query``

bdraco:
## What do these changes do?

Remove extra ``MultiDictProxy`` wrapper from ``BaseRequest.query``
query is already a MultiDictProxy even as far back as yarl 1.0.0 https://github.com/aio-libs/yarl/blob/829f6568916f9f3219a940720bee65103ca591c2/yarl/__init__.py#L472

no
no


CHANGES/9098.misc.rst
```diff
@@ -0,0 +1,3 @@
+Remove extra ``MultiDictProxy`` wrapper from ``BaseRequest.query`` -- by :user:`bdraco`.
```
Dreamsorcerer: This is surely not user-facing in any way? I'd skip any change note.

PR #9109, opened 2024-09-10
Remove None as a compress value

Dreamsorcerer:
Changed default value to ``compress`` from ``None`` to ``False`` (``None`` is no longer an expected value)

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -203,7 +203,7 @@ class ClientRequest:
         cookies: Optional[LooseCookies] = None,
         auth: Optional[BasicAuth] = None,
         version: http.HttpVersion = http.HttpVersion11,
-        compress: Union[str, bool, None] = None,
+        compress: Union[str, bool] = False,
         chunked: Optional[bool] = None,
         expect100: bool = False,
         loop: asyncio.AbstractEventLoop,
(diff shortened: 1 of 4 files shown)


aiohttp/client_reqrep.py
```diff
@@ -422,22 +421,20 @@ def update_cookies(self, cookies: Optional[LooseCookies]) -> None:
...
+    def update_content_encoding(self, data: Any, compress: Union[bool, str]) -> None:
         """Set request content encoding."""
+        self.compress = None
         if not data:
-            # Don't compress an empty body.
-            self.compress = None
             return
 
         enc = self.headers.get(hdrs.CONTENT_ENCODING, "").lower()
         if enc:
```
bdraco, line 431:
```suggestion
        if self.headers.get(hdrs.CONTENT_ENCODING):
```

I think there is no reason to call `.lower()` here if we only need to see if its set

bdraco: This is an existing issue though.  I'll do a new PR so we can backport

bdraco (approved):

aiohttp/client_reqrep.py
bdraco: https://github.com/aio-libs/aiohttp/pull/9234

PR #9136, opened 2024-09-12
Bump idna to 3.6

Dreamsorcerer:
Let's see which version is regressing.


Dreamsorcerer:
So, regression is clearly in 3.7.

PR #9137, opened 2024-09-13
Add and use ClientConnectionResetError

### LINKED ISSUE #8505: Doc Add ConnectionResetError exception to ClientWebSocketResponse documentation

In the documentation, for [send_json](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientWebSocketResponse.send_json), [send_str](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientWebSocketResponse.send_str), [send_bytes](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientWebSocketResponse.send_bytes), it should be noted that those methods will raise a `ConnectionResetError` if the transport is closing.

N/A

N/A

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Not sure if we should do it like this, or we should catch the exception in the client code and reraise this exception in order to be sure that a client exception doesn't get raised in the web side. Reraising would require updating a fair few places though in client_ws.py and client_reqrep.py and may make the code less readable with all the extra try/excepts.

Fixes #8505.

### DIFF
--- a/CHANGES/9137.bugfix.rst
+++ b/CHANGES/9137.bugfix.rst
@@ -1,0 +1,2 @@
+Added ``ClientConnectionResetError``. Client code that previously threw ``ConnectionResetError``
+will now throw this -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9137.bugfix.rst
+++ b/CHANGES/9137.bugfix.rst
@@ -1,2 +1,2 @@
-Added :exc:`ClientConnectionResetError`. Client code that previously threw :exc:`ConnectionResetError`
+Added :exc:`aiohttp.ClientConnectionResetError`. Client code that previously threw :exc:`ConnectionResetError`
 will now throw this -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9137.bugfix.rst
+++ b/CHANGES/9137.bugfix.rst
@@ -1,2 +1,2 @@
-Added ``ClientConnectionResetError``. Client code that previously threw ``ConnectionResetError``
+Added :exc:`ClientConnectionResetError`. Client code that previously threw :exc:`ConnectionResetError`
 will now throw this -- by :user:`Dreamsorcerer`.
--- a/aiohttp/client_exceptions.py
+++ b/aiohttp/client_exceptions.py
@@ -24,8 +24,8 @@ else:
 __all__ = (
     "ClientError",
     "ClientConnectionError",
-    "ClientOSError",
     "ClientConnectionResetError",
+    "ClientOSError",
     "ClientConnectorError",
     "ClientProxyConnectionError",
     "ClientSSLError",
--- a/aiohttp/client_exceptions.py
+++ b/aiohttp/client_exceptions.py
@@ -126,6 +126,10 @@ class ClientConnectionError(ClientError):
     """Base class for client socket errors."""
 
 
+class ClientConnectionResetError(ClientConnectionError, ConnectionResetError):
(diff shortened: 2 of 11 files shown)


CHANGES/9137.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Added ``ClientConnectionResetError``. Client code that previously threw ``ConnectionResetError``
```
webknjaz: This is how you link exceptions:
```suggestion
Added :exc:`ClientConnectionResetError`. Client code that previously threw :exc:`ConnectionResetError`
```
(provided they're documented in Sphinx)

webknjaz:
If this is being reused on server and client sides, it's probably going to be confusing UX-wise.

Dreamsorcerer:
> If this is being reused on server and client sides, it's probably going to be confusing UX-wise.

The thing is that I'm fairly confident this doesn't reach the user server-side. Any such exception is caught somewhere and handled.

webknjaz:
In that case, why would it need to derive from client exceptions?

Dreamsorcerer:
> In that case, why would it need to derive from client exceptions?

Because it propagates client-side and the user needs to catch it..

webknjaz:
Ah, I think I parsed your previous message wrong ;)

bdraco (approved):
This approach makes sense to me.

Would be good to add some coverage in these places that both exception types work
```
tests/test_base_protocol.py:    with pytest.raises(ConnectionResetError):
tests/test_client_ws.py:                with pytest.raises(ConnectionResetError):
tests/test_client_ws_functional.py:        resp._conn.transport, "write", side_effect=ConnectionResetError
tests/test_http_writer.py:    with pytest.raises(ConnectionResetError):
tests/test_http_writer.py:    """Test that writing to a closed transport raises ConnectionResetError.
tests/test_http_writer.py:    writing to the transport. If it is None, it raises ConnectionResetError.
tests/test_http_writer.py:    with pytest.raises(ConnectionResetError, match="Cannot write to closing transport"):
```

aiohttp/client_exceptions.py
```diff
@@ -18,13 +17,15 @@
...
+    from .http_parser import RawResponseMessage
 else:
-    RequestInfo = ClientResponse = ConnectionKey = None
+    RequestInfo = ClientResponse = ConnectionKey = RawResponseMessage = None
 
 __all__ = (
     "ClientError",
     "ClientConnectionError",
     "ClientOSError",
+    "ClientConnectionResetError",
```
Dreamsorcerer:
```suggestion
    "ClientConnectionError",
    "ClientConnectionResetError",
    "ClientOSError",
```

CHANGES/9137.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Added :exc:`ClientConnectionResetError`. Client code that previously threw :exc:`ConnectionResetError`
+will now throw this -- by :user:`Dreamsorcerer`.
```
bdraco, line 2: Should we tell them it inherits from `ConnectionResetError` so they know its backwards compat?

Dreamsorcerer: I figured they'd look at the exception in the docs, where this is shown. The main information I'm trying to convey here is that if they were catching ConnectionResetError as per the linked issue, then they can now remove that as it'll be caught as a regular ClientException.

PR #9157, opened 2024-09-16
[3.10] Cache construction of middleware handlers

bdraco:
Will do a PR to master once I'm happy with this, but since I'm traveling, I want to test live for now.  master will be much simpler since we don't have to deal with legacy middleware.
master version is https://github.com/aio-libs/aiohttp/pull/9158

## What do these changes do?

Cache construction of middleware.  Calling `update_wrapper` many times for each request with middleware is a more expensive than running all the middlewares 

No

no

Before
```
bdraco@MacBook-Pro-5 ~ % wrk -t 10 http://localhost:8123
Running 10s test @ http://localhost:8123
  10 threads and 10 connections
    Latency     6.83ms   33.14ms 348.84ms   96.25%
    Req/Sec     1.89k   799.64     2.41k    80.83%
  181206 requests in 10.01s, 0.95GB read
Requests/sec:  18100.19
Transfer/sec:     97.39MB
```

After
```
bdraco@MacBook-Pro-5 ~ % wrk -t 10 http://localhost:8123
Running 10s test @ http://localhost:8123
  10 threads and 10 connections
    Latency   386.34us  124.22us   5.11ms   98.73%
    Req/Sec     2.63k   156.95     2.77k    96.04%
  264316 requests in 10.10s, 1.39GB read
Requests/sec:  26169.50
Transfer/sec:    140.81MB
```

middleware

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -530,9 +530,7 @@ class Application(MutableMapping[Union[str, AppKey[Any]], Any]):
             handler = match_info.handler
 
             if self._run_middlewares:
-                handler = await self._apply_middlewares(
-                    self, handler, match_info.apps[::-1]
-                )
+                handler = await self._apply_middlewares(handler, match_info.apps[::-1])
 
             resp = await handler(request)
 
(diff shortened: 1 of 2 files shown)


aiohttp/web_app.py
```diff
@@ -525,19 +527,39 @@ async def _handle(self, request: Request) -> StreamResponse:
...
 
         return resp
 
+    async def _apply_middlewares(
+        self,
+        handler: Handler,
+        apps: Tuple["Application", ...],
+    ) -> Callable[[Request], Awaitable[StreamResponse]]:
+        """Apply middlewares to handler."""
+        cache_key = (handler, tuple(id(app) for app in apps))
```
Dreamsorcerer: Can't the `@cache` decorator do this automatically?

Dreamsorcerer:
Why would this be 3.10 only?

aiohttp/web_app.py
bdraco: Two problems

- `Application` is not Hashable because its `MutableMapping` and I was worried about the implications of making it hashable so it seemed safer to only allow the same `id` here.
- Currently this needs to be a coro for backwards compat (not a problem on master)

bdraco:
> Why would this be 3.10 only?

master version is https://github.com/aio-libs/aiohttp/pull/9158

I'm testing on 3.10 and since I'm traveling it was easier to do 3.10 first and work backwards

PR #9158, opened 2024-09-16
Cache construction of middleware handlers

bdraco:
## What do these changes do?

Cache construction of middleware.  Calling `update_wrapper` many times for each request with middleware is more expensive than running all the middlewares 

The `update_wrapper` call was introduced in https://github.com/aio-libs/aiohttp/pull/4195 which is what made middleware much slower.

No

no

Before
```
bdraco@MacBook-Pro-5 ~ % wrk -t 10 http://localhost:8123
Running 10s test @ http://localhost:8123
  10 threads and 10 connections
    Latency     6.83ms   33.14ms 348.84ms   96.25%
    Req/Sec     1.89k   799.64     2.41k    80.83%
  181206 requests in 10.01s, 0.95GB read
Requests/sec:  18100.19
Transfer/sec:     97.39MB
```

After
```
bdraco@MacBook-Pro-5 ~ % wrk -t 10 http://localhost:8123
Running 10s test @ http://localhost:8123
  10 threads and 10 connections
    Latency   386.34us  124.22us   5.11ms   98.73%
    Req/Sec     2.63k   156.95     2.77k    96.04%
  264316 requests in 10.10s, 1.39GB read
Requests/sec:  26169.50
Transfer/sec:    140.81MB
```

middleware

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -383,13 +383,13 @@ class Application(MutableMapping[Union[str, AppKey[Any]], Any]):
             handler = match_info.handler
 
             if self._run_middlewares:
-                handler = await self._apply_middlewares(handler, match_info.apps[::-1])
+                handler = self._apply_middlewares(handler, match_info.apps[::-1])
 
             resp = await handler(request)
 
         return resp
 
-    async def _apply_middlewares(
+    def _apply_middlewares(
         self,
         handler: Handler,
         apps: Tuple["Application", ...],
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -16,6 +16,7 @@ from typing import (
     MutableMapping,
     Optional,
     Sequence,
+    Tuple,
     Type,
     TypeVar,
     Union,
@@ -30,7 +31,7 @@ from frozenlist import FrozenList
 from . import hdrs
 from .helpers import AppKey
 from .log import web_logger
-from .typedefs import Middleware
+from .typedefs import Handler, Middleware
 from .web_exceptions import NotAppKeyWarning
 from .web_middlewares import _fix_request_current_app
 from .web_request import Request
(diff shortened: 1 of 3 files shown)


aiohttp/web_app.py
```diff
@@ -380,15 +383,31 @@ async def _handle(self, request: Request) -> StreamResponse:
...
 
         return resp
 
+    def _apply_middlewares(
+        self,
+        handler: Handler,
+        apps: Tuple["Application", ...],
+    ) -> Callable[[Request], Awaitable[StreamResponse]]:
+        """Apply middlewares to handler."""
+        cache_key = (handler, tuple(id(app) for app in apps))
```
Dreamsorcerer: I see no reason we couldn't just define `__hash__` in order to use `@cache`?

bdraco: I was a bit worried about that but as long as we hash on `id` I think its safe

Dreamsorcerer: I guess, strictly speaking, it's not safe as the ID could be reused for different objects. However, it seems incredibly unlikely to be an issue...

aiohttp/web_app.py
```diff
@@ -380,15 +384,25 @@ async def _handle(self, request: Request) -> StreamResponse:
...
-                    assert app.pre_frozen, "middleware handlers are not ready"
-                    for m in app._middlewares_handlers:
-                        handler = update_wrapper(partial(m, handler=handler), handler)
+                handler = Application._build_middlewares(handler, match_info.apps[::-1])
 
             resp = await handler(request)
 
         return resp
 
+    @staticmethod
```
Dreamsorcerer: Personally, staticmethod feels pointless to me and I'd probably just make a separate function.

Dreamsorcerer (approved):

elupus (commented):
Doesnt the cache need to be invalidated if an additional middleware is added at some late stage? (Know too little about internals to know if this can happen)

Dreamsorcerer:
> Doesnt the cache need to be invalidated if an additional middleware is added at some late stage? (Know too little about internals to know if this can happen)

They're frozen, they can't be modified once the app is started.

bdraco:
Need to look at the lru stats once I have stable internet

Issue #2779, opened 2018-02-27
Speedup aiohttp web server


asvetlov:
The typical web handler looks like:
```
async def handler(request):
    ...  # do business logic
    return web.Response(text=..., content_type=...)

Unfortunately construction of `web.Response` is relative slow, `resp.prepare()` is even slower.
The code has too many checks for overlapping `content_type` with `headers['Content-Type']`, response length with `headers['Content-Length']`, HTTP chunk processing etc. etc.

We can speed up the situation if we assume that `web.Response` is constructed with well knows single text or binary data.

The suggestion is:
- Introduce `web.text_response()` and `web.bin_response()` like already present `web.json_response()`. Functions will just make `web.Response` instances on this first step. The main point is teaching people to use the new API.
- `web.Response` should be kept as is for backward compatibility.
- After that we can refactor our response classes hierarchy by making a `BaseResponse` as parent for `StreamResponse` and deriving private classes for binary and text responses from `BaseResponse`. Obviously `BaseResponse.prepare()` should exist, not sure about other API attributes.
- These private classes can do as minimal job as possible, they don't need to pass redundant checks for conflicting HTTP headers.

We should forbid the following HTTP headers in `web.text_response()` / `web.bin_response()`:
-  Connection
-  Keep-Alive
-  Trailer
-  Transfer-Encoding
-  Content-Length
-  Upgrade

The check for headers is quite expensive, let's do it in debug mode only (`loop.get_debug() is True` or `PYTHONASYNCIODEBUG=1`.

That's it.

Discussion is welcome.
PR for the first step (adding `web.text_response()` and `web.bin_response()`) is even more welcome.

hubo1016:
I think it's OK to ignore any content-length and transfer-encoding settings for these responses, and always use the content-type specified by user if there is any. 

But I don't quite understand the speedup, I thought building the headers are always slower...

spcharc:
the check for headers is quite expensive  <- I don’t understand
It’s just some comparing work and it’s fast. Looking up in a multidict is also fast

asvetlov:
`response.prepare()` (and `response._start()`) take much longer time for simple responses in comparison as it should be.
I've created a gist for demonstration my steps to get profile results: https://gist.github.com/asvetlov/13295efdc1301bea241d7d35496e6f81
I hope you can reproduce snakeviz output by following the gist instructions (just snapshot in not informative because of interactive nature of the tool).
[...]

PR #9168, opened 2024-09-17
Avoid creating handler waiter until shutdown

bdraco:
## What do these changes do?

Every request created `_handler_waiter`. Instead only create it when we need to wait for graceful shutdown.  Avoids one future creation for every request if we are not shutting down.

no
no

related issue #2779

profile

before
handle_request_before

after
handle_request_after


aiohttp/web_protocol.py
```diff
@@ -494,7 +497,9 @@ async def _handle_request(
         else:
             resp, reset = await self.finish_response(request, resp, start_time)
         finally:
-            self._handler_waiter.set_result(None)
+            self._request_in_progress = False
+            if self._handler_waiter is not None and not self._handler_waiter.done():
```
Dreamsorcerer: We never needed the .done() check before? I don't think that should be possible..

Dreamsorcerer: Specifically, this method is called below in a `while not self._force_close` loop, and shutdown() sets that to True before making the future. So, there should be no way that it can call this twice while the future exists.

aiohttp/web_protocol.py
```diff
@@ -284,7 +286,8 @@ async def shutdown(self, timeout: Optional[float] = 15.0) -> None:
             self._keepalive_handle.cancel()
 
         # Wait for graceful handler completion
-        if self._handler_waiter is not None:
+        if self._request_in_progress:
+            self._handler_waiter = self._loop.create_future()
```
Dreamsorcerer, line 293: If this is genuinely a notable performance improvement, I think a comment should be added to explain this. Otherwise, it looks like code ready to be refactored to something simpler.

bdraco: Will do. I’ll add a profile for this one as well. Any time we can get rid of creating a future it’s always a nice performance bump.

aiohttp/web_protocol.py
bdraco: removed the done check

aiohttp/web_protocol.py
bdraco: Added comments, and profile above.

aiohttp/web_protocol.py
```diff
@@ -284,7 +286,12 @@ async def shutdown(self, timeout: Optional[float] = 15.0) -> None:
             self._keepalive_handle.cancel()
 
         # Wait for graceful handler completion
-        if self._handler_waiter is not None:
+        if self._request_in_progress:
+            # The future is only created when we are shutting
+            # down the connection while the handler is still
```
```suggestion
            # down while the handler is still
```

aiohttp/web_protocol.py
```diff
@@ -284,7 +286,12 @@ async def shutdown(self, timeout: Optional[float] = 15.0) -> None:
             self._keepalive_handle.cancel()
 
         # Wait for graceful handler completion
-        if self._handler_waiter is not None:
+        if self._request_in_progress:
+            # The future is only created when we are shutting
+            # down while the handler is still
+            # processing a request to avoid creating a future
+            # for every request.
```
```suggestion
            # down while the handler is still processing a request
            # to avoid creating a future for every request.
```

PR #9169, opened 2024-09-17
Speed up starting compression

bdraco:
## What do these changes do?

Enumerating a enum and accessing all the .value is not performant. Switching to a pre-built dict is significantly faster

no

no

related issue #2779

before
start_compression_before

after
start_compression_after


aiohttp/web_response.py
```diff
@@ -341,8 +343,8 @@ async def _start_compression(self, request: "BaseRequest") -> None:
             # Encoding comparisons should be case-insensitive
             # https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1
             accept_encoding = request.headers.get(hdrs.ACCEPT_ENCODING, "").lower()
-            for coding in ContentCoding:
-                if coding.value in accept_encoding:
+            for value, coding in CONTENT_CODINGS.items():
+                if value in accept_encoding:
```
Dreamsorcerer, line 351: I'm struggling with this one. It looks to me like all we've saved is one attribute access...

With the dict, I was expecting a reversal of logic or something here, like `for enc in accept_encoding: CONTENT_CODINGS.get(enc)`.

bdraco: Iterating enums in python is surprisingly expensive and good chunk of its the [`__get__`](https://github.com/python/cpython/blob/0c4884d8aa51df2dd013c3e78fcc3f2077d743c3/Lib/enum.py#L198). `_start_compression` was the most expensive call in `_prepare_headers` before:

We save:
- creating the iterator for ContentCoding enum (although it might be an even exchange to iterate `.items()`, but thats all going to be in native code for sure)
- creating a enum singleton for each one in the loop
- accessing the value which does some magic under the hood (its not a simple attribute)



... maybe we can save more here but I was thinking the `in` 3x was cheaper than a `dict` lookup

Zoomed in on `_prepare_headers` before the change:
prepare_headers_zoomed

Zoomed in on `_start_compression` before the change:
start_compression_zoomed

bdraco: a dict get might make sense if we parsed the string but parsing is probably more expensive `Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5`

PR #9171, opened 2024-09-18
Switch to using `yarl.URL.absolute` over `yarl.URL.is_absolute()`

bdraco:
## What do these changes do?

`.absolute` is directly cached and preferred

no

no

related issue #2779


bdraco:
not backported to 3.10 since the minimum yarl requirement is too old there

Dreamsorcerer (approved):

PR #9172, opened 2024-09-18
Improve performance of starting web requests when content length is not set

bdraco:
## What do these changes do?

Reorder the start logic to avoid the check to remove content length if its not present

no
no

related issue #2779

before
start_before

after
start_after


bdraco:
Looks like we are missing coverage for this case but it was not obvious before

Dreamsorcerer (approved):

tests/test_web_response.py
```diff
@@ -627,6 +627,22 @@ async def write_headers(status_line, headers):
     assert resp.content_length is None
 
 
+async def test_rm_content_length_if_for_204() -> None:
```
bdraco:
```suggestion
async def test_rm_content_length_if_204() -> None:
```

PR #9174, opened 2024-09-18
Add a cache to must_be_empty_body

bdraco:
## What do these changes do?

Nearly all the calls are going to be the same so we can cache the logic

I couldn't figure out how to avoid this call in most cases like https://github.com/aio-libs/aiohttp/pull/9172 so a cache made more sense.

no
no

related issue #2779

before
must_be_empty_body_before

after
must_be_empty_body_after

### DIFF
--- a/CHANGES/9174.misc.rst
+++ b/CHANGES/9174.misc.rst
@@ -1,1 +1,1 @@
-Improved performance of preparing web requests -- by :user:`bdraco`.
+Improved performance of web requests -- by :user:`bdraco`.
--- a/CHANGES/9174.misc.rst
+++ b/CHANGES/9174.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of preparing web requests -- by :user:`bdraco`.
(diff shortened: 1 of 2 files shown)


webknjaz (approved):
Looks good, though the change note could be made less broad..

Dreamsorcerer (approved):

Dreamsorcerer:
> Looks good, though the change note could be made less broad..

Personally, I don't think these are user-facing really, so all the recent PRs could probably be grouped to a single changenote about performance.

bdraco:
> > Looks good, though the change note could be made less broad..
> 
> Personally, I don't think these are user-facing really, so all the recent PRs could probably be grouped to a single changenote about performance.

Will symlink https://github.com/aio-libs/aiohttp/pull/9169 and https://github.com/aio-libs/aiohttp/pull/9172 to this one

CHANGES/9174.misc.rst
```diff
@@ -0,0 +1 @@
+Improved performance of preparing web requests -- by :user:`bdraco`.
```
bdraco:
```suggestion
Improved performance of web requests -- by :user:`bdraco`.
```

PR #9175, opened 2024-09-18
Speed up finding the reason if its unset

bdraco:
## What do these changes do?

Previously a new enum object had to be created and the reason phase looked up. Since there are a limited number of codes, we can pre-build a dict with all the phases which makes the runtime code much simpler

no
no

related issue #2779

init request

before
init_request_before

after
init_request_after


aiohttp/web_response.py
```diff
@@ -160,12 +163,7 @@ def set_status(
             not self.prepared
         ), "Cannot change the response status code after the headers have been sent"
         self._status = int(status)
-        if reason is None:
-            try:
-                reason = HTTPStatus(self._status).phrase
```
Dreamsorcerer, line 165: So, the problem is the actual object being instantiated?

aiohttp/web_response.py
```diff
@@ -160,12 +163,7 @@ def set_status(
...
-            except ValueError:
-                reason = ""
-        self._reason = reason
+        self._reason = reason or REASON_PHRASES.get(self._status, "")
```
Dreamsorcerer: Does the user want to be able to set an empty reason value explicitly? This change would break it.

aiohttp/web_response.py
```diff
@@ -122,7 +124,8 @@ def __init__(
         else:
             self._headers = CIMultiDict()
 
-        self.set_status(status, reason)
+        self._status = status
+        self._reason = reason or REASON_PHRASES.get(status, "")
```
Dreamsorcerer: I don't see a reason for the duplication here?

aiohttp/web_response.py
```diff
@@ -160,12 +163,7 @@ def set_status(
             not self.prepared
         ), "Cannot change the response status code after the headers have been sent"
         self._status = int(status)
```
Dreamsorcerer: Could probably also drop the int() call, atleast on master, given the type annotations.

aiohttp/web_response.py
bdraco: The idea was to avoid the `self.prepared` check since it was impossible to be an issue in `__init__`.  I'll break it out into an underscore function instead.

aiohttp/web_response.py
bdraco: dropped in 054e931248b5a8c34981be28131d56d9ddd2c72e, will leave it for backport

aiohttp/web_response.py
bdraco: Yes

aiohttp/web_response.py
bdraco: Fixed in 7ec583cbf2fd6f177bbcac9f8e7145fb6168d7c4

bdraco: coverage in a66573560b0f9d0bad9b3574fdb2ee6893f0b22d

PR #9200, opened 2024-09-19
Improve middleware performance

### LINKED ISSUE #9196: Middleware fix up is unexpectedly expensive
https://github.com/aio-libs/aiohttp/blob/f95bcaf4e0b2344d09df8dbb565150dcb4e73c0f/aiohttp/web_middlewares.py#L116

it’s likely the contextmanager decorator but need to dig in more

bdraco:
## What do these changes do?
The contextmanager in cff60ebb65e35a28b481969a8e91b72e95a9f649 [added quite a bit of overhead](https://github.com/aio-libs/aiohttp/issues/9196#issuecomment-2360799101) and was only used for fixing the current app when using middleware. I did a github code search and did not find usage of `set_current_app` outside of aiohttp so I think its safe to remove as its unlikely to be used outside of aiohttp.

closes #9196

The `set_current_app` function has been removed.

no

### DIFF
--- a/CHANGES/9200.breaking.rst
+++ b/CHANGES/9200.breaking.rst
@@ -1,0 +1,3 @@
+Improved middleware performance -- by :user:`bdraco`.
+
+The ``set_current_app`` method was removed from :class:`UrlMappingMatchInfo` because it is no longer used, and it was unlikely external caller would ever use it.
--- a/CHANGES/9200.breaking.rst
+++ b/CHANGES/9200.breaking.rst
@@ -1,3 +1,3 @@
 Improved middleware performance -- by :user:`bdraco`.
 
-The ``set_current_app`` method was removed from :class:`UrlMappingMatchInfo` because it is no longer used, and it was unlikely external caller would ever use it.
+The ``set_current_app`` method was removed from ``UrlMappingMatchInfo`` because it is no longer used, and it was unlikely external caller would ever use it.
--- a/aiohttp/web_middlewares.py
+++ b/aiohttp/web_middlewares.py
@@ -115,7 +115,12 @@ def normalize_path_middleware(
 
 def _fix_request_current_app(app: "Application") -> Middleware:
     async def impl(request: Request, handler: Handler) -> StreamResponse:
-        with request.match_info.set_current_app(app):
+        match_info = request.match_info
+        prev = match_info.current_app
+        match_info.current_app = app
+        try:
             return await handler(request)
+        finally:
+            match_info.current_app = prev
 
     return impl
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -8,7 +8,6 @@ import keyword
 import os
 import re
 import sys
-from contextlib import contextmanager
 from pathlib import Path
 from types import MappingProxyType
 from typing import (
@@ -271,8 +270,8 @@ class UrlMappingMatchInfo(BaseDict, AbstractMatchInfo):
         assert app is not None
         return app
 
-    @contextmanager
-    def set_current_app(self, app: "Application") -> Generator[None, None, None]:
+    @current_app.setter
+    def current_app(self, app: "Application") -> None:
         if DEBUG:  # pragma: no cover
             if app not in self._apps:
                 raise RuntimeError(
@@ -280,12 +279,7 @@ class UrlMappingMatchInfo(BaseDict, AbstractMatchInfo):
                         self._apps, app
                     )
                 )
-        prev = self._current_app
         self._current_app = app
-        try:
-            yield
-        finally:
-            self._current_app = prev
 
     def freeze(self) -> None:
         self._frozen = True


Dreamsorcerer:
Is it possibly caused by the exception handling or something?
https://github.com/python/cpython/blob/4420cf4dc9ef7bd3c1c9b5465fa9397304bf0110/Lib/contextlib.py#L149

bdraco:
> Is it possibly caused by the exception handling or something? [python/cpython@`4420cf4`/Lib/contextlib.py#L149](https://github.com/python/cpython/blob/4420cf4dc9ef7bd3c1c9b5465fa9397304bf0110/Lib/contextlib.py#L149)

Yeah it looks like the `__enter__` and mostly `__exit__` implementation is overkill for this use case.

Dreamsorcerer:
> > Is it possibly caused by the exception handling or something? [python/cpython@`4420cf4`/Lib/contextlib.py#L149](https://github.com/python/cpython/blob/4420cf4dc9ef7bd3c1c9b5465fa9397304bf0110/Lib/contextlib.py#L149)
> 
> Yeah it looks like the `__enter__` and mostly `__exit__` implementation is overkill for this use case.

I don't think enter is an issue. I believe `try` is now a 0-cost implementation in Python 3.13..

But, exit handles an exception in the expected case. I can't think of any way to make that more efficient though, other than not using the decorator...

bdraco:
The profile shows the most expensive part is `__init__` on line 104 with cpython 3.12.4, followed by `__exit__` so it looks like the context manager object has to be **created every time** from the `helper` function.

bdraco:
I recall we had a similar problem with rendering templates in Home Assistant (we can render many per second) and the solution was to create the context manager once and use it over and over... that worked because there was no await and everything finished synchronously, but we can't use that solution here.

Dreamsorcerer:
> The profile shows the most expensive part is `__init__` on line 104 with cpython 3.12.4, followed by `__exit__` so it looks like the context manager object has to be **created every time** from the `helper` function.
> 
> Line 108 on cpython `main` https://github.com/python/cpython/blob/4420cf4dc9ef7bd3c1c9b5465fa9397304bf0110/Lib/contextlib.py#L108

Looks to me like there's a ContextDecorator which uses _recreate_cm() to avoid redoing the init everytime. It seems to me like the decorators could be improved a little to do that same thing instead of instantiating a new class directly on every call. Not sure how much you'd get from that, but atleast would save fetching the docs attribute etc. in the init.

bdraco:
> > The profile shows the most expensive part is `__init__` on line 104 with cpython 3.12.4, followed by `__exit__` so it looks like the context manager object has to be **created every time** from the `helper` function.
> > Line 108 on cpython `main` [python/cpython@`4420cf4`/Lib/contextlib.py#L108](https://github.com/python/cpython/blob/4420cf4dc9ef7bd3c1c9b5465fa9397304bf0110/Lib/contextlib.py#L108)
> 
> Looks to me like there's a ContextDecorator which uses _recreate_cm() to avoid redoing the init everytime. It seems to me like the decorators could be improved a little to do that same thing instead of instantiating a new class directly on every call. Not sure how much you'd get from that, but atleast would save fetching the docs attribute etc. in the init.

Looks like thats already supposed to be happening but `__init__` is still running every time since its bound to an `UrlMappingMatchInfo` which gets recreated every time.

 https://github.com/python/cpython/blob/fcfe78664bf740a7fb059f9817e584a612e09f54/Lib/contextlib.py#L128 
https://github.com/python/cpython/blob/fcfe78664bf740a7fb059f9817e584a612e09f54/Lib/contextlib.py#L301

bdraco:
I played around with it for a bit and came to the conclusion that trying to make the context manager performant is likely not possible without writing it all out which got large quickly.  I think what we have now is probably the simplest solution.

Dreamsorcerer:
> > Looks to me like there's a ContextDecorator which uses _recreate_cm() to avoid redoing the init everytime. It seems to me like the decorators could be improved a little to do that same thing instead of instantiating a new class directly on every call. Not sure how much you'd get from that, but atleast would save fetching the docs attribute etc. in the init.
> 
> Looks like thats already supposed to be happening but `__init__` is still running every time since its bound to an `UrlMappingMatchInfo` which gets recreated every time.
> 
> https://github.com/python/cpython/blob/fcfe78664bf740a7fb059f9817e584a612e09f54/Lib/contextlib.py#L128 https://github.com/python/cpython/blob/fcfe78664bf740a7fb059f9817e584a612e09f54/Lib/contextlib.py#L301

Maybe I'm misreading that, but doesn't the decorator transform the original function into a new function which returns a _new_ _GeneratorContextManager on every call? Seems to me like the _GeneratorContextManager would need to be created in the outer function and then called in the inner function in order for _recreate_cm() to ever be called..

Dreamsorcerer:
> since its bound to an `UrlMappingMatchInfo` which gets recreated every time

The decorator would run on the function (unbound method), right? So, the decorator is not being run every time.

bdraco:
> The decorator would run on the function (unbound method), right? So, the decorator is not being run every time.

Decorator runs once, but context manager is created every time.

Example:

```
from contextlib import contextmanager
import cProfile

class X:

    @contextmanager
    def my_context(self):
        """My context."""
        yield


pr = cProfile.Profile()
pr.enable()
for _ in range(1000):
    with X().my_context():
        pass
pr.disable()
pr.create_stats()
pr.print_stats()
pr.dump_stats("cm.cprof")
```
```
         9001 function calls in 0.001 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1000    0.000    0.000    0.000    0.000 contextlib.py:104(__init__)
     1000    0.000    0.000    0.000    0.000 contextlib.py:132(__enter__)
     1000    0.000    0.000    0.000    0.000 contextlib.py:141(__exit__)
     1000    0.000    0.000    0.000    0.000 contextlib.py:299(helper)
     2000    0.000    0.000    0.000    0.000 test_context.py:7(my_context)
     1000    0.000    0.000    0.000    0.000 {built-in method builtins.getattr}
     2000    0.000    0.000    0.000    0.000 {built-in method builtins.next}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}


```

Dreamsorcerer:
> Decorator runs once, but context manager is created every time.

Well, yes, that's my point above..

Seems to me like the function should be written something like:
```
def contextmanager(func):
    return _GeneratorContextManager(func, args, kwds)(func)
```

At which point, each call would only be running:
```
            with self._recreate_cm():
                return func(*args, **kwds)
```

Dreamsorcerer:
Hacky proof-of-concept:
```
from contextlib import _GeneratorContextManager, wraps
import cProfile

class CM(_GeneratorContextManager):
    def __init__(self, func):
        self.func = func
        doc = getattr(func, "__doc__", None)
        if doc is None:
            doc = type(self).__doc__
        self.__doc__ = doc

    def _recreate_cm(self, args, kwargs):
        self.gen = self.func(*args, **kwargs)
        return self

    def __enter__(self):
        try:
            return next(self.gen)
        except StopIteration:
            raise RuntimeError("generator didn't yield") from None

def contextmanager(func):
    cm = CM(func)

    @wraps(func)
    def inner(*args, **kwds):
        return cm._recreate_cm(args, kwds)
    return inner

class X:

    @contextmanager
    def my_context(self):
        """My context."""
        yield


pr = cProfile.Profile()
pr.enable()
for _ in range(1000):
    with X().my_context():
        pass
pr.disable()
pr.create_stats()
pr.print_stats()
pr.dump_stats("cm.cprof")
```

```
         8001 function calls in 0.002 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1000    0.000    0.000    0.001    0.000 contextlib.py:117(__exit__)
     1000    0.000    0.000    0.000    0.000 test.py:12(_recreate_cm)
     1000    0.000    0.000    0.000    0.000 test.py:16(__enter__)
     1000    0.000    0.000    0.000    0.000 test.py:25(inner)
     2000    0.000    0.000    0.000    0.000 test.py:32(my_context)
     2000    0.000    0.000    0.000    0.000 {built-in method builtins.next}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

Note that here only the exit reaches 1ms cumtime, when I run the original I get 1ms

Dreamsorcerer:
```
from contextlib import _GeneratorContextManager, wraps, contextmanager
import cProfile

class CM(_GeneratorContextManager):
    def __init__(self, func):
        self.func = func
        doc = getattr(func, "__doc__", None)
        if doc is None:
            doc = type(self).__doc__
        self.__doc__ = doc

    def _recreate_cm(self, args, kwargs):
        self.gen = self.func(*args, **kwargs)
        return self

    def __enter__(self):
        try:
            return next(self.gen)
        except StopIteration:
            raise RuntimeError("generator didn't yield") from None

def contextmanager2(func):
    cm = CM(func)

    @wraps(func)
    def inner(*args, **kwds):
        return cm._recreate_cm(args, kwds)
    return inner

class X:

    @contextmanager2
    def my_context(self):
        """My context."""
        yield
```

Switching between contextmanager and contextmanager2 and timing with:
```
python3 -m timeit -s 'from test import X' 'with X().my_context(): pass'
```

I get around 1.4 - 2.3 µs with the original and 0.9 - 1.3 µs with the hack. So, original is about 50-80% slower. Though just setting self.gen is probably not correct (it won't be thread-safe for example), so would need to make some change for that..

Dreamsorcerer:
Taking the instantiation out:
```
python3 -m timeit -s 'from test import X; x = X()' 'with x.my_context(): pass'
```

I get as low as 0.8µs with the hack and no lower than 1.7 µs with the old one, so actually over 100% slower.

Dreamsorcerer:
Hmm, I actually can't see how that `__doc__` attribute is actually useful... I don't think it's actually relevant for contextmanager().

So, something like this should be safe and I can still get <1 µs from it.

PR #9203, opened 2024-09-19
Implement heapq for cookie expire times

### LINKED ISSUE #7790: Remove _next_expiration field from CookieJar

### LINKED ISSUE #8575: Expiring the cookie jar is unexpectedly expensive
### Describe the bug

clear_cookie_jar

### To Reproduce

I think the trace source if python-kasa since it clear the cookie jar on every connection

### Expected behavior

clear should be fast

### Logs/tracebacks

```python-traceback
n/a
```

### Python Version

```console
n/a
```

### aiohttp Version

```console
3.10.0
```

### multidict Version

```console
n/a
```

### yarl Version

```console
n/a
```

### OS

Linux

### Related component

Client

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

Implement heapq for cookie expire times

no

heapq is a bit of a complex structure to maintain

## Related issue number
fixes #8575
fixes #7790

benchmark script (we care about the `filter_other_domain` case for expires and that still includes the filtering time so its actually even faster): https://github.com/aio-libs/aiohttp/issues/7790#issuecomment-1807262618

Before
filter_other_domain: 0.451234458014369

After
filter_other_domain: 0.0018771658651530743

note that profile is not exactly the same number of iterations since its using a real world use case of 60s and its hard to get it exactly the same (1644 before) / (1413 after) but its such a significant difference, its more than enough to show the performance improvement of ~96-97% less run time (adjusted ratio for the iterations):

before
do_expiration_before

after
do_expiration_after

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -159,7 +159,10 @@ class CookieJar(AbstractCookieJar):
         return sum(len(cookie.values()) for cookie in self._cookies.values())
 
     def _cleanup_expirations(self) -> None:
-        """Remove expired entries from the expiration heap."""
+        """Remove expired entries from the expiration heap.
+
+        This method is called when the expiration heap grows too large.
+        """
         self._expire_heap = [
             entry
             for entry in self._expire_heap
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -107,9 +107,8 @@ class CookieJar(AbstractCookieJar):
                     for url in treat_as_secure_origin
                 }
             )
-        self._expirations: Dict[Tuple[str, str, str], float] = {}
-        # heap of cookies by expiration time
         self._expire_heap: List[Tuple[float, str, str, str]] = []
+        self._expirations: Dict[Tuple[str, str, str], float] = {}
 
     def save(self, file_path: PathLike) -> None:
         file_path = pathlib.Path(file_path)
@@ -123,10 +122,10 @@ class CookieJar(AbstractCookieJar):
 
     def clear(self, predicate: Optional[ClearCookiePredicate] = None) -> None:
         if predicate is None:
+            self._expire_heap.clear()
             self._cookies.clear()
             self._host_only_cookies.clear()
             self._expirations.clear()
-            self._expire_heap.clear()
             return
 
         now = time.time()
@@ -193,8 +192,8 @@ class CookieJar(AbstractCookieJar):
             self._expirations.pop((domain, path, name), None)
 
     def _expire_cookie(self, when: float, domain: str, path: str, name: str) -> None:
-        self._expirations[(domain, path, name)] = when
         heapq.heappush(self._expire_heap, (when, domain, path, name))
+        self._expirations[(domain, path, name)] = when
 
     def update_cookies(self, cookies: LooseCookies, response_url: URL = URL()) -> None:
         """Update cookies."""
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -179,6 +179,8 @@ class CookieJar(AbstractCookieJar):
         ):
             # Remove any expired entries from the expiration heap
             # that do not match the expiration time in the expirations
+            # as it means the cookie has been re-added to the heap
+            # with a different expiration time.
             self._expire_heap = [
                 entry
                 for entry in self._expire_heap
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -41,6 +41,11 @@ CookieItem = Union[str, "Morsel[str]"]
 _FORMAT_PATH = "{}/{}".format
 _FORMAT_DOMAIN_REVERSED = "{1}.{0}".format
 
+# The minimum number of scheduled cookie expirations before we start cleaning up
+# the expiration heap. This is a performance optimization to avoid cleaning up the
+# heap too often when there are only a few scheduled expirations.
+_MIN_SCHEDULED_COOKIE_EXPIRATION = 100
+
 
 class CookieJar(AbstractCookieJar):
     """Implements cookie storage adhering to RFC 6265."""
@@ -176,7 +181,10 @@ class CookieJar(AbstractCookieJar):
         if not expire_heap_len:
             return
 
-        if expire_heap_len > 100 and expire_heap_len > len(self._expirations) * 2:
+        if (
+            expire_heap_len > _MIN_SCHEDULED_COOKIE_EXPIRATION
+            and expire_heap_len > len(self._expirations) * 2
+        ):
             self._cleanup_expirations()
 
         now = time.time()
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -181,6 +181,11 @@ class CookieJar(AbstractCookieJar):
         if not expire_heap_len:
             return
 
+        # If the expiration heap grows larger than the number expirations
+        # times two, we clean it up to avoid keeping expired entries in
+        # the heap and consuming memory. We guard this with a minimum
+        # threshold to avoid cleaning up the heap too often when there are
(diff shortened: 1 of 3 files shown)


aiohttp/cookiejar.py
```diff
@@ -167,11 +166,49 @@ def __len__(self) -> int:
         return sum(len(cookie.values()) for cookie in self._cookies.values())
 
     def _do_expiration(self) -> None:
-        self.clear(lambda x: False)
+        # If the number of deleted items is greater than twice the number of
+        # expired heap, rebuild the expired heap without the deleted items
+        # and clear the deleted items.
+        deleted_len = len(self._deleted)
+        if deleted_len > 100 and deleted_len > len(self._expire_heap) * 2:
```
bdraco: probably should make 100 a constant `MIN_DELETED_TO_CLEANUP`

aiohttp/cookiejar.py
```diff
@@ -167,11 +160,41 @@ def __len__(self) -> int:
...
+        now = time.time()
+        to_del: List[Tuple[str, str, str]] = []
+        while self._expire_heap:
+            entry = self._expire_heap[0]
+            when = entry[0]
+            if when > now:
+                break
+            heapq.heappop(self._expire_heap)
+            cookie_key = entry[1:]
+            if self._expirations.get(cookie_key) == when:
```
bdraco, line 203: needs some tests to changing expire times

Going backwards, going forwards, and the same time seen twice

aiohttp/cookiejar.py
```diff
@@ -167,10 +159,40 @@ def __len__(self) -> int:
         return sum(len(cookie.values()) for cookie in self._cookies.values())
 
     def _do_expiration(self) -> None:
-        self.clear(lambda x: False)
+        expire_heap_len = len(self._expire_heap)
+        if not expire_heap_len:
+            return
+
+        if expire_heap_len > 100 and expire_heap_len > len(self._expirations) * 2:
```
bdraco: Need a test for the cleanup

aiohttp/cookiejar.py
```diff
@@ -121,34 +122,25 @@ def load(self, file_path: PathLike) -> None:
...
+            key
+            for (domain, path), cookie in self._cookies.items()
+            for name, morsel in cookie.items()
+            if (
+                (key := (domain, path, name)) in self._expirations
+                and self._expirations[key] <= now
+            )
+            or predicate(morsel)
+        ]
+        if to_del:
```
bdraco, line 147: Looks like we don't have coverage for clear matching nothing

aiohttp/cookiejar.py
```diff
@@ -121,34 +122,25 @@ def load(self, file_path: PathLike) -> None:
...
-            if self._expirations
-            else self.MAX_TIME
-        )
+        to_del = [
+            key
+            for (domain, path), cookie in self._cookies.items()
+            for name, morsel in cookie.items()
+            if (
+                (key := (domain, path, name)) in self._expirations
+                and self._expirations[key] <= now
```
Dreamsorcerer, line 143: Are we assuming that clear() is not called often? Otherwise, it'd probably make more sense to have _expirations sorted and then be able to break early here instead of needing to iterate every one.

Dreamsorcerer: The original issue does say that clear() is being used, but I don't see how these changes would make clear() much faster. The heapq isn't being utilised here.

bdraco: I think it only would help if predicate is callable that always returns False and that code has been removed.

AFAICT clear was only the issue because it was being called from _do_expiration

Clearing a domain will still be expensive but we would need another index to improve that.  I'm not sure how common it is to know if it's warranted... we could probably make that faster by eliminating the lambda and using a partial without an index being added... but something for another PR

Dreamsorcerer: Right, I was thinking the predicate was used when a cookie is expiring, but it's the other way around.

aiohttp/cookiejar.py
```diff
@@ -167,11 +159,40 @@ def __len__(self) -> int:
         return sum(len(cookie.values()) for cookie in self._cookies.values())
 
     def _do_expiration(self) -> None:
```
Dreamsorcerer, line 166: I'd put some comments in this function, it's a little dense to make sense of.

Dreamsorcerer (approved):

aiohttp/cookiejar.py
```diff
@@ -167,11 +159,40 @@ def __len__(self) -> int:
...
+        for domain, path, name in to_del:
+            self._host_only_cookies.discard((domain, name))
+            self._cookies[(domain, path)].pop(name, None)
+            self._expirations.pop((domain, path, name), None)
 
     def _expire_cookie(self, when: float, domain: str, path: str, name: str) -> None:
-        self._next_expiration = min(self._next_expiration, when)
-        self._expirations[(domain, path, name)] = when
+        cookie_key = (domain, path, name)
+        heapq.heappush(self._expire_heap, (when, cookie_key))
```
bdraco, line 220: Probably should have a guard here so if when hasn’t changed we don’t add the same entry in the heap.  Not sure if we can get here in that case but coverage will show us

aiohttp/cookiejar.py
```diff
@@ -167,11 +159,40 @@ def __len__(self) -> int:
         return sum(len(cookie.values()) for cookie in self._cookies.values())
 
     def _do_expiration(self) -> None:
-        self.clear(lambda x: False)
+        expire_heap_len = len(self._expire_heap)
+        if not expire_heap_len:
+            return
+
+        if expire_heap_len > 100 and expire_heap_len > len(self._expirations) * 2:
+            self._expire_heap = [
```
bdraco, line 184: Maybe break this out into a function to better show intent

aiohttp/cookiejar.py
bdraco: Definitely needs a comment as well

aiohttp/cookiejar.py
```diff
@@ -167,11 +159,40 @@ def __len__(self) -> int:
...
+            heapq.heapify(self._expire_heap)
+
+        now = time.time()
+        to_del: List[Tuple[str, str, str]] = []
+        while self._expire_heap:
+            when, cookie_key = self._expire_heap[0]
+            if when > now:
+                break
+            heapq.heappop(self._expire_heap)
+            if self._expirations.get(cookie_key) == when:
```
bdraco, line 203: Needs a comment explaining how it could still be in the heap but not reached and the expire time changes so we don’t want to delete it until we get far enough down in the heap that we hit the new expire time

tests/test_cookiejar.py
```diff
@@ -876,9 +876,95 @@ async def test_cookie_jar_clear_expired() -> None:
...
+        assert len(matched_cookies) == 1
+        assert "foo" in matched_cookies
+
+        freezer.move_to("1990-01-01 14:00:00+00:00")
+        matched_cookies = jar.filter_cookies(URL("/"))
+        assert len(matched_cookies) == 0
+
+
+async def test_cookie_jar_duplicates_with_expire_heap() -> None:
+    """Test that duplicate cookies to not grow the expires heap."""
```
Dreamsorcerer:
```suggestion
    """Test that duplicate cookies do not grow the expires heap."""
```

bdraco:
Running well on production overnight 👍

PR #9204, opened 2024-09-19
Significantly speed up filter_cookies

bdraco:
## What do these changes do?

Significantly speed up filter_cookies by adding a cache for morsel construction.  Previously every request had to create new Morsel objects.  Its a common pattern to set a cookie once and read it many times.  When testing with Home Assistant, the cache had nearly a 100% hit rate.

Benchmark https://github.com/aio-libs/aiohttp/issues/7790#issuecomment-1807262618

before 
filter_domain: 6.765125583391637

after
filter_domain: 1.3521433752030134

no
no


Dreamsorcerer:
This doesn't look like it should really make a difference.. I feel like the extra nesting might hurt readability too.

aiohttp/cookiejar.py
```diff
@@ -294,31 +294,28 @@ def filter_cookies(self, request_url: URL) -> "BaseCookie[str]":
         # Create every combination of (domain, path) pairs.
         pairs = itertools.product(domains, paths)
 
-        # Point 2: https://www.rfc-editor.org/rfc/rfc6265.html#section-5.4
-        cookies = itertools.chain.from_iterable(
-            self._cookies[p].values() for p in pairs
-        )
```
bdraco, line 343: Its probably the generator expression thats influencing the profile here.

bdraco:
Will do a profile later. If its not significant, I'll close this as it won't be churn if its only small.

bdraco:
https://github.com/aio-libs/aiohttp/issues/7790#issuecomment-1807262618

This benchmark should work

bdraco:
before

 % python3 filter_large_cookie_jar.py 
filter_domain: 6.765125583391637
filter_other_domain: 0.0018160408362746239

bdraco:
after

filter_domain: 6.75009624985978
filter_other_domain: 0.0017206249758601189

bdraco:
Based on that I'd say its not worth the churn

bdraco:
Still need to investigate if there is anything we can do to improve the filter domain case but that seems like it won't be a small change. I'm not sure if it's still a problem though

bdraco:
Did some more analysis and found out that most of the time is copying the cookie and making another Morsel object so I added a cache since most of the time we set once, ready many.


With adding a cache

filter_domain: 1.3521433752030134

bdraco:
Still in flight for a few more hours so will need to do some real world analysis to make sure the cache is effective as expected.

Right now we spend more time in cookie.py/cookiejar.py than http_writer.py in most cases so it would be a really nice speed up if we can make filter_cookies faster

bdraco:
The hit rate is near 100% at least on Home Assistant so it definitely seems worth caching

before
filter_cookies_before

after
filter_cookies_after


On another system
filter_cookies_after_other

aiohttp/cookiejar.py
```diff
@@ -285,7 +290,12 @@ def update_cookies(self, cookies: LooseCookies, response_url: URL = URL()) -> No
                 else:
                     cookie["expires"] = ""
 
-            self._cookies[(domain, path)][name] = cookie
+            key = (domain, path)
+            if self._cookies[key].get(name) != cookie:
+                # Don't blow away the cache is the same
```
```suggestion
                # Don't blow away the cache if the same
```

PR #9239, opened 2024-09-22
Stop raise_for_status() releasing when in a context

### LINKED ISSUE #8451: Allow reading already retrieved response body after connection is closed

The changes in https://github.com/aio-libs/aiohttp/pull/3365 have been discussed in other issues before and this is a related issue.

With `ClientReponse.text` I'm able to get the body of a response even if I did raise the exception. Assuming I `read()` it before.
https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/client_reqrep.py#L1083-L1091

If `_body` is already set it will just be decoded. 

However, I'm talking with a server that sets its headers incorrectly and therefore can't rely on `get_encoding()`. Calling `ClientReponse.read` actually breaks my code:
https://github.com/aio-libs/aiohttp/blob/f662958b150a9d8d92fcbd0c9235e6bee1bedd67/aiohttp/client_reqrep.py#L1040-L1058

Even if `._body` is already set I won't be able to read it again as there is now a check against `self._released`.

Would it be possible to expose a method that retrieves the body of a `ClientReponse` if it is already set without checking if the response is released yet?

Not really. 

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #8451.

### DIFF
--- a/CHANGES/9239.bugfix.rst
+++ b/CHANGES/9239.bugfix.rst
@@ -1,1 +1,1 @@
-Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an ``async with`` context -- by :user:`Dreamsorcerer`.
+Changed ``ClientResponse.raise_for_status()`` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9239.bugfix.rst
+++ b/CHANGES/9239.bugfix.rst
@@ -1,1 +1,1 @@
-Changed ``ClientResponse.raise_for_status()`` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
+Changed :py:meth:`ClientResponse.raise_for_status() <aiohttp.ClientResponse.raise_for_status>` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9239.bugfix.rst
+++ b/CHANGES/9239.bugfix.rst
@@ -1,0 +1,1 @@
+Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an `async with` context -- by :user:`Dreamsorcerer`.
--- a/CHANGES/9239.bugfix.rst
+++ b/CHANGES/9239.bugfix.rst
@@ -1,1 +1,1 @@
-Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an `async with` context -- by :user:`Dreamsorcerer`.
+Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an ``async with`` context -- by :user:`Dreamsorcerer`.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -737,6 +737,7 @@ class ClientResponse(HeadersMixin):
     # post-init stage allows to not change ctor signature
     _closed = True  # to allow __del__ for non-initialized properly response
     _released = False
+    _in_context = False
     __writer = None
 
     def __init__(
@@ -1022,7 +1023,12 @@ class ClientResponse(HeadersMixin):
         if not self.ok:
             # reason should always be not None for a started response
             assert self.reason is not None
-            self.release()
+
+            # If we're in a context we can rely on __aexit__() to release as the
+            # exception propagates.
+            if not self._in_context:
+                self.release()
+
             raise ClientResponseError(
                 self.request_info,
                 self.history,
@@ -1144,6 +1150,7 @@ class ClientResponse(HeadersMixin):
         return loads(self._body.decode(encoding))  # type: ignore[union-attr]
 
     async def __aenter__(self) -> "ClientResponse":
+        self._in_context = True
         return self
 
     async def __aexit__(
@@ -1152,6 +1159,7 @@ class ClientResponse(HeadersMixin):
         exc_val: Optional[BaseException],
         exc_tb: Optional[TracebackType],
     ) -> None:
+        self._in_context = False
         # similar to _RequestContextManager, we do not need to check
         # for exceptions, response object can close connection
         # if state is broken
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -3721,7 +3721,7 @@ async def test_read_after_raise_outside_context(aiohttp_client: AiohttpClient) -
         # No async with, so should release and therefore read() will fail.
         await resp.raise_for_status()
 
-    with pytest.raises(aiohttp.ClientConnectionError, match="Connection closed"):
+    with pytest.raises(aiohttp.ClientConnectionError, match=r"^Connection closed$"):
         await resp.read()
 
 
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -3722,7 +3722,7 @@ async def test_read_after_raise_outside_context(aiohttp_client: AiohttpClient) -
         await resp.raise_for_status()
 
     with pytest.raises(aiohttp.ClientConnectionError):
-        result = await resp.read()
+        await resp.read()
 
 
 async def test_read_from_closed_content(aiohttp_client: AiohttpClient) -> None:
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -3721,7 +3721,7 @@ async def test_read_after_raise_outside_context(aiohttp_client: AiohttpClient) -
         # No async with, so should release and therefore read() will fail.
         await resp.raise_for_status()
 
-    with pytest.raises(aiohttp.ClientConnectionError, matches="Connection closed"):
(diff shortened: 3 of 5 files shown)


tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError):
+        # No async with, so should release and therefore read() will fail.
+        await resp.raise_for_status()
+
+    with pytest.raises(aiohttp.ClientConnectionError):
+        result = await resp.read()
```
```suggestion
        await resp.read()
```

aiohttp/client_reqrep.py
```diff
@@ -1144,6 +1150,7 @@ async def json(
         return loads(self._body.decode(encoding))  # type: ignore[union-attr]
 
     async def __aenter__(self) -> "ClientResponse":
+        self._in_context = True
```
bdraco, line 1153: Is there any way they could enter/exit multiple times?

Dreamsorcerer: Once exited, the response is already released. So, doesn't really matter after that point whether we call .release() or not.

aiohttp/client_reqrep.py
```diff
@@ -1022,7 +1023,12 @@ def raise_for_status(self) -> None:
         if not self.ok:
             # reason should always be not None for a started response
             assert self.reason is not None
-            self.release()
+
+            # If we're in a context we can rely on __aexit__() to release as the
```
webknjaz, line 1027: Does this mean we *want to* support/tolerate cases where the callers don't use CMs? I'd say this encourages an antipattern. I'd rather delete the auto-release here in favor or telling people to always use a CM or release themselves…

CHANGES/9239.bugfix.rst
```diff
@@ -0,0 +1 @@
+Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an `async with` context -- by :user:`Dreamsorcerer`.
```
webknjaz:
```suggestion
Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an ``async with`` context -- by :user:`Dreamsorcerer`.
```

CHANGES/9239.bugfix.rst
```diff
@@ -0,0 +1 @@
+Stopped ``ClientResponse.raise_for_status()`` releasing the connection within an ``async with`` context -- by :user:`Dreamsorcerer`.
```
webknjaz:
```suggestion
Changed :py:meth:`ClientResponse.raise_for_status() <aiohttp.ClientResponse.raise_for_status>` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
```

aiohttp/client_reqrep.py
Dreamsorcerer: This has been like this for some time. My reasoning for why this is the expected behaviour is that if they have code to release it themselves, it may not get run because we're throwing an exception here that could cause it to bypass that code. So, instead we make this an explicit release for this exceptional case.

CHANGES/9239.bugfix.rst
```diff
@@ -0,0 +1 @@
+Changed ``ClientResponse.raise_for_status()`` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
```
webknjaz:
```suggestion
Changed :py:meth:`ClientResponse.raise_for_status() <aiohttp.ClientResponse.raise_for_status>` to only release the connection when invoked outside an ``async with`` context -- by :user:`Dreamsorcerer`.
```

aiohttp/client_reqrep.py
webknjaz: But if we expected that people were to do the right thing and always use the async context manager, that wouldn't be a problem, right?

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response(body=b"data", status=404)
+
+    app = web.Application()
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError):
```
webknjaz: Plz always use the `match=` regex with `pytest.raises()`, `pytest.warns()` and `pytest.deprecated()`. This improves the accuracy of such tests and reduces false-positives.

aiohttp/client_reqrep.py
Dreamsorcerer: We do explicitly allow `await` instead of the context manager in our code and documentation, even if we generally discourage it. We also have a substantial number of tests validating such behaviour (e.g. validating that .close() is not needed after a .read() call etc.).

If we wanted to remove the `await` usage, that seems like an entirely different discussion and could only go into v4.

webknjaz: Fair enough. Marking the thread as resolved.

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response(body=b"data", status=404)
+
+    app = web.Application()
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    async with client.get("/") as resp:
+        with pytest.raises(ClientResponseError):
```
```suggestion
        with pytest.raises(ClientResponseError, match="404"):
```

tests/test_client_functional.py
```suggestion
    with pytest.raises(ClientResponseError, match="404"):
```

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError):
+        # No async with, so should release and therefore read() will fail.
+        await resp.raise_for_status()
+
+    with pytest.raises(aiohttp.ClientConnectionError):
```
```suggestion
    with pytest.raises(aiohttp.ClientConnectionError, matches="Connection closed"):
```

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError, match="404"):
+        # No async with, so should release and therefore read() will fail.
+        await resp.raise_for_status()
+
+    with pytest.raises(aiohttp.ClientConnectionError, matches="Connection closed"):
```
```suggestion
    with pytest.raises(aiohttp.ClientConnectionError, match="Connection closed"):
```

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError, match="404"):
+        # No async with, so should release and therefore read() will fail.
+        await resp.raise_for_status()
+
+    with pytest.raises(aiohttp.ClientConnectionError, match="Connection closed"):
```
webknjaz: I recommend making it more explicit that it's a regex. Otherwise, you can get weird matches:
```suggestion
    with pytest.raises(aiohttp.ClientConnectionError, match=r"^Connection closed$"):
```

bdraco (approved):

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    app = web.Application()
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    async with client.get("/") as resp:
+        with pytest.raises(ClientResponseError, match="404"):
+            # Should not release response when in async with context.
+            await resp.raise_for_status()
```
```suggestion
            resp.raise_for_status()
```

tests/test_client_functional.py
```diff
@@ -3688,6 +3689,42 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    app = web.Application()
+    app.add_routes([web.get("/", handler)])
+
+    client = await aiohttp_client(app)
+
+    resp = await client.get("/")
+    with pytest.raises(ClientResponseError, match="404"):
+        # No async with, so should release and therefore read() will fail.
+        await resp.raise_for_status()
```
```suggestion
        resp.raise_for_status()
```

bdraco (approved):

PR #9240, opened 2024-09-22
Small cleanups to Application._handle

bdraco:
- Remove some unneeded intent
- Remove some intermediate variables

### DIFF
--- a/aiohttp/web_app.py
+++ b/aiohttp/web_app.py
@@ -386,22 +386,19 @@ class Application(MutableMapping[Union[str, AppKey[Any]], Any]):
         match_info.add_app(self)
         match_info.freeze()
 
-        resp = None
         request._match_info = match_info
-        expect = request.headers.get(hdrs.EXPECT)
-        if expect:
+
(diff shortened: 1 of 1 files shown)


Dreamsorcerer (approved):

aiohttp/web_app.py
```diff
@@ -386,22 +386,20 @@ async def _handle(self, request: Request) -> StreamResponse:
         match_info.add_app(self)
         match_info.freeze()
 
-        resp = None
         request._match_info = match_info
-        expect = request.headers.get(hdrs.EXPECT)
-        if expect:
+
+        if request.headers.get(hdrs.EXPECT):
             resp = await match_info.expect_handler(request)
```
bdraco, line 392: We might not have any coverage for `expect_handler` returning a response

bdraco: Looks like we do 👍

bdraco:
This cleanup is mostly for 3.1x as the code if larger there.

PR #9261, opened 2024-09-23
Cleanup changelog messages

bdraco:


CHANGES/8845.bugfix.rst
```diff
@@ -1 +1 @@
-Changed behaviour when returning an invalid response to send a 500 response -- by :user:`Dreamsorcerer`.
+Changed behavior when returning an invalid response to send a 500 response -- by :user:`Dreamsorcerer`.
```
bdraco, line 1: This showed up as a spelling, but not sure if we should be using American English or British English?

webknjaz: I tried asking about the spelling in other projects like pip and learned that they don't enforce a specific variant, accepting anything people write. And so I thought we could default to that here too.

I think Sam uses the British variant because of where he's based and that's okay.

On a related note, I've been meaning to explore integrating https://vale.sh to support style consistency efforts. I assume the tool would force us to select one way or the other, whenever we get to it. For now, I would only fix things that the linters enforce but wouldn't be too vocal about non-enforced personal preferences in reviews.

To summarize, I don't have a strong opinion here so long as others don't either.

CHANGES/9031.misc.rst
```diff
@@ -1 +1 @@
-Tracing overhead is avoided in the http writer when there are no active traces -- by user:`bdraco`.
```
webknjaz, line 1: I think that the  passive voice in present tense is fine in many places in the change log, if that fits well the surrounding style. Giving the guidance of using the past tense is the easiest to explain but that's definitely not the only way.

webknjaz (approved):
Reading this, I realize that I need to improve the change log guidance in general. I'll make a mental note for myself.

All the inline comments below are non-blocking.

PR #9266, opened 2024-09-23
Highlight fixing router matching pre-encoded URLs can be a breaking change

bdraco:
This regression was around for so long that its behavior is relied upon in downstream projects. Highlight the fix as breaking change.

https://github.com/aio-libs/aiohttp/pull/8898
https://github.com/aio-libs/aiohttp/issues/5621
https://github.com/aio-libs/aiohttp/issues/6619

details of how Home Assistant was using it https://github.com/home-assistant/core/pull/126468


bdraco:
superseded by https://github.com/aio-libs/aiohttp/pull/9267

PR #9267, opened 2024-09-23
Fix double unquoting in url dispatcher

bdraco:
#8898 now passes the unquoted path and we would unquote it again.  To fix this we need to bump the minimum yarl version to 1.12.0 and use the new `URL.path_safe` method.  This is done in a single PR since the yarl bump has a revert in it that is breaking (only for unreleased/non-stable versions of `aiohttp`) without this change.


bdraco:
If this passes the CI than we likely have a gap in test coverage

aiohttp/web_urldispatcher.py
```diff
@@ -1257,7 +1257,7 @@ def _quote_path(value: str) -> str:
 
 
 def _unquote_path(value: str) -> str:
-    return URL.build(path=value, encoded=True).path.replace("%2F", "/")
+    return value.replace("%2F", "/")
```
bdraco: This is still a bit of an issue because it could be %2f or %2F

aiohttp/web_urldispatcher.py
```diff
@@ -1257,7 +1257,23 @@ def _quote_path(value: str) -> str:
...
 def _unquote_path(value: str) -> str:
-    return URL.build(path=value, encoded=True).path.replace("%2F", "/")
+    """Unquote a URL path.
+
+    This method is used to unquote %2F to / in the URL path
+    for paths returned from yarl.URL.path. It is not intended
+    for use with yarl.URL.raw_path.
+    """
+    # Note that older yarl versions already unquote "%2F" to "/"
+    # so we only want to unquote if we see "%2F" or "%2f"
```
bdraco: This still won't work right because the path segment can be quoted and we would be double decoding by doing this

bdraco: While the test will pass its still wrong

aiohttp/web_urldispatcher.py
```diff
@@ -1257,7 +1257,21 @@ def _quote_path(value: str) -> str:
 
 
 def _unquote_path(value: str) -> str:
-    return URL.build(path=value, encoded=True).path.replace("%2F", "/")
```
Dreamsorcerer, line 1260: I thought when I looked at this originally, we were getting URL.path which is already normalised to uppercase encoded characters...

Dreamsorcerer: Oh, wait, we were creating a URL here, now we're not...
Ideally we'd want to utilise URL.path in some way though, which is already optimised.

bdraco: I'm not sure this is fixable because in this case `value` is already unquoted and when we run the regex we get the `%2F` in the match group but we don't know if its an encoded value or if its yarl skipping decoding `%2F` since its already decoded.

Dreamsorcerer: As far as I can see, both calls to this function are actually passing the value from `request.rel_url.path`, so it looks to me like it should already be safe?

bdraco: If the url has double quoted values in it, we don't know if we are converting the %2F that yarl skips unquoting or the unquoted value

```
>>> from urllib.parse import quote
>>> import yarl
>>> yarl.URL(quote("p/a/t/h=1", safe=''), encoded=True).path.replace('%2F','/')
'p/a/t/h=1'
>>> yarl.URL(quote(quote("p/a/t/h=1", safe=''), safe=''), encoded=True).path.replace('%2F','/')
'p/a/t/h%3D1'
>>> 
```

bdraco: I would expect the second one to return `'p%2Fa%2Ft%2Fh%3D1'` and not `'p/a/t/h%3D1'`

tests/test_urldispatch.py
```diff
@@ -774,7 +774,23 @@ async def test_dynamic_match_unquoted_path(router: web.UrlDispatcher) -> None:
     resource_id = "my%2Fpath%7Cwith%21some%25strange%24characters"
     req = make_mocked_request("GET", f"/path/{resource_id}")
     match_info = await router.resolve(req)
-    assert match_info == {"path": "path", "subpath": unquote(resource_id)}
```
Dreamsorcerer, line 777: This was a deliberate decision initially, so the user gets the unquoted path in the match info.
i.e. We only need to worry about path separators in the routing. Unless they change the default regex, it'd be impossible for a user to get an unquoted `/` in a variable (and I don't think we should worry about the difference even if they do change the regex).

aiohttp/web_urldispatcher.py
Dreamsorcerer: I see. Maybe we can add something to yarl...?

bdraco: I think if we revert the change in yarl everything will work as expected with this PR

Dreamsorcerer: Have you tried it? We made that change to fix the routing, I'm not sure I see any changes here to make that routing work..

bdraco: Yes https://github.com/aio-libs/yarl/pull/1149. I did have to change this test https://github.com/aio-libs/aiohttp/pull/9267/files#diff-ffe45e5e9ca19ae2918aa4ac4796a9871765331a2b862c9c5efc898e36ccc068R773

I looked at all the views in Home Assistant and found they were all made before the original regression so they have `:.+` in them

bdraco: I reverted the commits that require a new yarl and doing a CI run with yarl 1.9.4 to see if it passes still

bdraco: We have `app.router.add_route("GET", "/resource2/a/b%2Fc/", handler)` in the middleware tests

bdraco: https://github.com/aio-libs/aiohttp/pull/3051

bdraco: The middleware check expects paths to be raw
https://github.com/aio-libs/aiohttp/blob/203ca4273edcd6ea5372c630b054381ebb3bacdb/aiohttp/web_middlewares.py#L107

bdraco: The rabbit hole keeps getting deeper...

I think the safest course is to revert https://github.com/aio-libs/aiohttp/pull/8898 and https://github.com/aio-libs/yarl/pull/1057

tests/test_urldispatch.py
```diff
@@ -770,13 +770,24 @@ async def test_dynamic_match_two_part2(router: web.UrlDispatcher) -> None:
 
 async def test_dynamic_match_unquoted_path(router: web.UrlDispatcher) -> None:
     handler = make_handler()
-    router.add_route("GET", "/{path}/{subpath}", handler)
```
Dreamsorcerer: I think if this change is required, then the issue I fixed is now not fixed...

I don't think this is properly fixable without have separate attributes in yarl or something...

To distill the problem down to the minimal issue:
Problem with old yarl:
`URL("/", encoded=True).path == URL("%2F", encoded=True).path`
Problem with new yarl:
`URL("%2F", encoded=True).path == URL("%252F", encoded=True).path`

I don't see any way to fix that while having only one path attribute to handle it.

bdraco: I'm not sure that will take care of this case

The regex now runs with unquoted path at https://github.com/aio-libs/aiohttp/blob/203ca4273edcd6ea5372c630b054381ebb3bacdb/aiohttp/web_urldispatcher.py#L473

With the yarl change, %2F is never unquoted

So if when return the `match.groupdict()` its going to have `%2F` if there was a quoted `/`

We could replace `%2F` with `/` but that means that we would also convert `%252F` to `/`

bdraco: We don't have any way to convert the match back to the raw url parts and than unquote them after we get the match result

tests/test_urldispatch.py
```diff
@@ -770,11 +770,25 @@ async def test_dynamic_match_two_part2(router: web.UrlDispatcher) -> None:
...
+    router.add_route("GET", "/{path}/{subpath:.+}", handler)
```
bdraco:
```suggestion
    router.add_route("GET", "/{path}/{subpath}", handler)
```

tests/test_urldispatch.py
Dreamsorcerer: > We don't have any way to convert the match back to the raw url parts and than unquote them after we get the match result

Exactly, the problem exists with both new and old yarl though, it's just different situations. The only way I can think of, is to have 2 different attributes which are used here. e.g. Using the new-style .path for routing and the old-style .path for matchinfo.

Dreamsorcerer: I guess the matchinfo is part of the routing though, so that's not so easy...

bdraco: Yeah we have to return the match dict and once we decode it we can never know if the / was encoded in the original as `%2F` or `%252F`

bdraco:
```
> URL('%2F', encoded=True).path
'%2F'
> URL('%252F', encoded=True).path
'%2F'
```

Dreamsorcerer: Yeah, that's what I highlighted in the first comment. Likewise, we can't tell if the `/` was a path separator or not with the old yarl for the same reason.

Dreamsorcerer: In theory, I think we could do two matches, although that's might not be best for performance:
```
    def _match(self, path: str, unslashed_path: str) -> Optional[Dict[str, str]]:
        match = self._pattern.fullmatch(path)  # new-style path for routing
        if match is None:
            return None

        match = self._pattern.fullmatch(unslashed_path)  # old-style path for matchinfo
        return {key: _unquote_path(value) for key, value in match.groupdict().items()}
```

tests/test_urldispatch.py
```diff
@@ -774,7 +774,21 @@ async def test_dynamic_match_unquoted_path(router: web.UrlDispatcher) -> None:
     resource_id = "my%2Fpath%7Cwith%21some%25strange%24characters"
     req = make_mocked_request("GET", f"/path/{resource_id}")
     match_info = await router.resolve(req)
-    assert match_info == {"path": "path", "subpath": unquote(resource_id)}
+    assert match_info == {
+        "path": "path",
+        "subpath": "my%2Fpath|with!some%strange$characters",
```
bdraco: If we don't unquote `%2F` everything works downstream

It seems we already accepted that as OK...
https://github.com/aio-libs/aiohttp/blob/203ca4273edcd6ea5372c630b054381ebb3bacdb/tests/test_urldispatch.py#L489
https://github.com/aio-libs/aiohttp/blob/203ca4273edcd6ea5372c630b054381ebb3bacdb/tests/test_urldispatch.py#L667

...but should we be reverting this and the yarl change instead since this change is in a `.6` patch release?

tests/test_urldispatch.py
Dreamsorcerer: I suppose the pattern would need to be changed too though...

Dreamsorcerer: Actually, I don't it'd work...

Dreamsorcerer: What about not unquoting `%25` and then adding that to the .replace()? I think that could work, though it might make things more awkward for other users of yarl..

Dreamsorcerer: So, maybe we end up with `url.path_safe.replace("%2F", "/").replace("%25", "%")` or similar.

Dreamsorcerer: It subtly affects the regex matching, as it'd now be matching against `%25` instead of `%`, but I'm not sure of any better way.

tests/test_urldispatch.py
Dreamsorcerer: Those 2 tests are using url_for(), i.e. the URL needed to request those files. If it unquoted the %2F, then the URL would be wrong. I don't think that's related to what the matchinfo should be.

I guess the problem is that the handler for an endpoint with such a url may be written like:
`for parts in request.match_info["filepath"].split("/")`

If we don't decode the `%2F`, users will need to do the decoding now (or split on that), but all we've really done if offload the unsolvable issue onto them. Because, now the user won't be able to tell the difference between `%2F` and `%252F`.

tests/test_urldispatch.py
bdraco: pushed a change that seems to work ok with https://github.com/aio-libs/yarl/pull/1150

tests/test_urldispatch.py
bdraco: With the change I just pushed along with https://github.com/aio-libs/yarl/pull/1057 these tests go back to the original behavior before https://github.com/aio-libs/aiohttp/pull/8898 of decoding `%2F`

aiohttp/web_urldispatcher.py
```diff
@@ -1256,8 +1254,16 @@ def _quote_path(value: str) -> str:
     return URL.build(path=value, encoded=False).raw_path
 
 
-def _unquote_path(value: str) -> str:
-    return URL.build(path=value, encoded=True).path.replace("%2F", "/")
+def _unquote_path_safe(value: str) -> str:
+    if "%" not in value:
```
Dreamsorcerer, line 1258: As this value is coming from path_safe, it should still be normalised, right (guess code coverage would highlight it anyway..)?

I think it should still be fine to use:
`return value.replace("%2F", "/").replace("%25", "%")`

bdraco: These will be the original URL because `path_safe` skipped decoding them.

We don't know if the original url used `%2F` or `%2f`

tests/test_urldispatch.py
```diff
@@ -486,7 +486,7 @@ def test_add_static_quoting(router: web.UrlDispatcher) -> None:
     )
     assert router["static"] is resource
     url = resource.url_for(filename="/1 2/файл%2F.txt")
-    assert url.path == "/пре %2Fфикс/1 2/файл%2F.txt"
+    assert url.path == "/пре /фикс/1 2/файл%2F.txt"
```
Dreamsorcerer, line 489: I'm not clear why we want to change this (and the introduction of _requote_path())?

Feels to me like this might be an incorrect URL, given the add_statci() appears to be looking for a literaly `%2F`.

bdraco: It changes because I removed the ignore of `/` for `.path` in https://github.com/aio-libs/yarl/pull/1150/files#diff-657d7257a2bac78ab69f50e60aa9cef5620adc8cd7f902d4e4cbeb0c6e78a434R228

Do we want to keep the ignore of `/` for `.path` and `.path_safe` (there are some issues linked where its causing downstream problems with the `path` change ... https://github.com/conda-forge/aiohttp-feedstock/pull/94 https://github.com/NixOS/nixpkgs/pull/339112 https://github.com/mosquito/aiormq/pull/202 https://github.com/mosquito/aiormq/issues/203) ?

aiohttp/web_urldispatcher.py
Dreamsorcerer: The ignore list results in:
`to_add = self._quoter(unquoted)`

So, I think it should always be uppercase.

Dreamsorcerer: In fact, I'm fairly confident I put tests in...

Dreamsorcerer: OK, I put one test in the yarl one:
https://github.com/aio-libs/yarl/pull/1057/files#diff-6e407480bfb30fa51ddc2501641e06ed9ec6c0b351c523c59427341df198d572R325

tests/test_urldispatch.py
Dreamsorcerer: No, we should revert it. I'm just thinking about it wrong. The important thing is the raw URL that is given to the user, which is tested on the next line. The .path output here isn't so important.

aiohttp/web_urldispatcher.py
```diff
@@ -412,7 +411,7 @@ def get_info(self) -> _InfoDict:
         return {"path": self._path}
 
     def url_for(self) -> URL:  # type: ignore[override]
-        return URL.build(path=self._path, encoded=True)
+        return URL.build(path=_requote_path(self._path), encoded=True)
```
Dreamsorcerer: I'm still not entirely clear what this change relates to.

bdraco: probably not needed

```suggestion
        return URL.build(path=self._path, encoded=True)
```

bdraco: removed

tests/test_urldispatch.py
bdraco: Ok I will split that PR for yarl so its more obvious its being reverted

tests/test_urldispatch.py
```diff
@@ -774,7 +774,21 @@ async def test_dynamic_match_unquoted_path(router: web.UrlDispatcher) -> None:
     resource_id = "my%2Fpath%7Cwith%21some%25strange%24characters"
     req = make_mocked_request("GET", f"/path/{resource_id}")
     match_info = await router.resolve(req)
-    assert match_info == {"path": "path", "subpath": unquote(resource_id)}
+    assert match_info == {
+        "path": "path",
+        "subpath": "my/path|with!some%strange$characters",
+    }
```
bdraco:
```suggestion
    assert match_info == {"path": "path", "subpath": unquote(resource_id)}
```

tests/test_web_urldispatcher.py
```diff
@@ -864,6 +864,22 @@ async def handler(request: web.Request) -> web.Response:
         assert resp.status == expected_http_resp_status
 
 
+async def test_decoded_raw_match_regex(aiohttp_client: AiohttpClient) -> None:
+    """Verify that raw_match only matches decoded url."""
+    app = web.Application()
+
+    async def handler(request: web.Request) -> NoReturn:
+        return False
```
bdraco:
```suggestion
        return False  # pragma: no cover
```

tests/test_web_urldispatcher.py
```diff
@@ -864,6 +864,22 @@ async def handler(request: web.Request) -> web.Response:
         assert resp.status == expected_http_resp_status
 
 
+async def test_decoded_raw_match_regex(aiohttp_client: AiohttpClient) -> None:
+    """Verify that raw_match only matches decoded url."""
+    app = web.Application()
+
+    async def handler(request: web.Request) -> NoReturn:
+        return False  # pragma: no cover
```
bdraco:
```suggestion
        assert False
```

tests/test_urldispatch.py
```diff
@@ -4,7 +4,7 @@
 from collections.abc import Container, Iterable, Mapping, MutableMapping, Sized
 from functools import partial
 from typing import Awaitable, Callable, Dict, List, NoReturn, Optional, Type
-from urllib.parse import unquote
+from urllib.parse import quote
```
```suggestion
from urllib.parse import quote, unquote
```

bdraco:
https://github.com/home-assistant/core/pull/126468

all passing now.

I think we are good to do the stable release

PR #9268, opened 2024-09-23
Increase minimum yarl version to 1.12.0

bdraco:
needs https://github.com/aio-libs/yarl/actions/runs/11000860360

The minimum version must increase because we need ``URL.path_safe`` to be able to fix the issue in PR #9267 and the original PR #8898

### DIFF
--- a/CHANGES/9268.breaking.rst
+++ b/CHANGES/9268.breaking.rst
@@ -1,1 +1,1 @@
-Increase minimum yarl version to 1.12.0 -- by :user:`bdraco`.
+Increased minimum yarl version to 1.12.0 -- by :user:`bdraco`.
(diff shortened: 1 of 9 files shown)


CHANGES/9268.breaking.rst
```diff
@@ -0,0 +1 @@
+Increase minimum yarl version to 1.12.0 -- by :user:`bdraco`.
```
bdraco:
```suggestion
Increased minimum yarl version to 1.12.0 -- by :user:`bdraco`.
```

bdraco:
Looks like I'm going to have to merge this into https://github.com/aio-libs/aiohttp/pull/9267 to get it to be a single change without breaking anything

PR #9279, opened 2024-09-24
[3.10] Bump pydantic and deps for python 3.13 compat

bdraco:
## What do these changes do?

Bump pydantic to a version that works with python 3.13


Dreamsorcerer:
So annoying, you can't even cancel a workflow without a runner...

PR #9284, opened 2024-09-24
Bump aiohttp-theme

Dreamsorcerer:
Fix a rendering issue in the docs

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -11,14 +11,14 @@ submodules:
   recursive: true
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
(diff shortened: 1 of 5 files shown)


.readthedocs.yml
```diff
@@ -11,7 +11,7 @@ submodules:
   recursive: true
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
```
webknjaz, line 14: This seem to be a separate thing..

webknjaz (approved):

.readthedocs.yml
Dreamsorcerer: It's only purpose is to fix installing the new package.

Issue #9138, opened 2024-09-13
3.9.2 server keep-alive timeout can cause disconnect of active connections and hence client server disconnect errors


thehesiod:
### Describe the bug

as mentioned in https://github.com/aio-libs/aiohttp/issues/2682 we had an issue with keep-alive timeouts.  There appears to be a race condition between how often requests are made and the server keep alive timeout.  If they overlap often the server will cause disconnect errors on the client.

### To Reproduce

Running the following code in 3.9.2/3.10.5:
```python
import asyncio
import random
import sys
import logging
import time

import aiohttp
import aiohttp.web
import requests
import requests.adapters

logger = logging.getLogger()


async def get_foo(req: aiohttp.web.Request):
    response = aiohttp.web.StreamResponse(status=200, headers={'Content-Type': 'text/csv'})

    logger.info("sending response")
    await response.prepare(req)
    await response.write(b'abc' * 512)
    # await asyncio.sleep(1)
    await response.write_eof()
    logger.info("wrote eof")
    return response


port = int(sys.argv[1])

async def server():
    app = aiohttp.web.Application()
    app.add_routes([aiohttp.web.get('/foo', get_foo)])
    await aiohttp.web._run_app(app, port=port, keepalive_timeout=1.0)


async def aiohttp_client(session: aiohttp.ClientSession):
    while True:
        async with session.get(f'http://localhost:{port}/foo', verify_ssl=False) as r:
            result = await r.text()
            logger.info(f'{result[:15]}...')

        await asyncio.sleep(random.uniform(0.99, 1.01))


def requests_client(session):
    while True:
        res = session.get(f'http://localhost:{port}/foo', verify=False)
        result = res.text
        logger.info(f'{result[:15]}...')
        time.sleep(random.uniform(0.99, 1.01))


async def main():
    logging.basicConfig(level=logging.INFO, format='%(asctime)s %(levelname)-8s %(message)s')
    server_task = asyncio.create_task(server())

    num_clients = 4
    await asyncio.sleep(2)

    loop = asyncio.get_event_loop()

    session = requests.Session()
    session.mount("https://", requests.adapters.HTTPAdapter(pool_connections=num_clients, pool_maxsize=num_clients))
    session.mount("http://", requests.adapters.HTTPAdapter(pool_connections=num_clients, pool_maxsize=num_clients))
    await loop.run_in_executor(None, requests_client, session)

[...]

PR #9285, opened 2024-09-24
Increase default keepalive_timeout server-side.

Dreamsorcerer:
As we strongly recommend deploying apps behind a reverse proxy it makes sense to optimise the defaults for this use case.

When using a reverse proxy it's important to have a longer timeout than the proxy's timeout, otherwise we likely hit race conditions with the proxy trying to forward another request as we close the connection.

For proxies, it'd probably make the most sense to actually remove the timeout entirely, but for users without proxies the defaults should still be safe, therefore I'm proposing to just make it substantially longer by default.

I expect the vast majority of proxies to have a keepalive of 1 hour or less, so have chosen a value just over the 1 hour mark (for reference, nginx defaults to 75 seconds, ALB to 1 hour).

See #9138

### DIFF
--- a/CHANGES/9285.misc.rst
+++ b/CHANGES/9285.misc.rst
@@ -1,0 +1,1 @@
+Changed web ``keepalive_timeout`` default to around an hour in order to reduce race conditions on reverse proxies -- by :user:`Dreamsorcerer`.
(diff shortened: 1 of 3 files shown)


bdraco (commented):
This could result in significantly more memory being used on busy systems. Its probably fine though as its easy enough to override, but it may surprise some users.

bdraco (approved):

Dreamsorcerer:
> This could result in significantly more memory being used on busy systems.

Well, a busy system should definitely have a proxy, and this shouldn't have any impact in that situation.

PR #9292, opened 2024-09-24
Reject non-str values in form data

### LINKED ISSUE #8052: Form-encoded data with None as the value gets passed as string "None"
### Describe the bug

When i have form encoded data with `None`, by using requests the data is not sent to the server. But this is different for aiohttp, is sent `"None"` string. This happen also in httpx, take a look at this issue https://github.com/encode/httpx/issues/1500. Is this really the expected behavior?

### To Reproduce

Code

```python
import asyncio
from  aiohttp import ClientSession 

async def main():
    async with session.post("https://httpbin.org/post", data={"foo": None}) as resp:
      print(await resp.json())

asyncio.run(main())
```

Output

```
{'args': {}, 'data': '', 'files': {}, 'form': {'foo': 'None'}, 'headers': {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Content-Length': '8', 'Content-Type': 'application/x-www-form-urlencoded', 'Host': 'httpbin.org', 'User-Agent': 'Python/3.12 aiohttp/3.9.1', 'X-Amzn-Trace-Id': 'Root=1-65ae0f19-5b556c225f8d6df5174e435d'}, 'json': None, 'origin': '110.137.36.152', 'url': 'https://httpbin.org/post'}
```

Take a look at the `form`, its contain None `'form': {'foo': 'None'}`.

### Expected behavior

Code

```python
import requests

resp = requests.post("https://httpbin.org/post", data={"foo": None})
print(resp.json())
```

Output

```
{'args': {}, 'data': '', 'files': {}, 'form': {}, 'headers': {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Content-Type': 'application/x-www-form-urlencoded', 'Host': 'httpbin.org', 'Transfer-Encoding': 'chunked', 'User-Agent': 'python-requests/2.31.0', 'X-Amzn-Trace-Id': 'Root=1-65ae0f4c-569162ab7ff3089468c91391'}, 'json': None, 'origin': '110.137.36.152', 'url': 'https://httpbin.org/post'}
```

Take a look at the `form`, its empty `'form': {}`.

### Logs/tracebacks

```python-traceback
N/A
```

### Python Version

```console
Python 3.12.1
```

### aiohttp Version

```console
name         : aiohttp
version      : 3.9.1
description  : Async http client/server framework (asyncio)
```

### multidict Version

```console
name         : multidict
version      : 6.0.4
description  : multidict implementation
```

### yarl Version

```console
name         : yarl
version      : 1.9.4
description  : Yet another URL library
```

### OS

Windows 10

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Dreamsorcerer:
Fixes #8052.

### DIFF
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -209,7 +209,7 @@ async def test_merge_headers_with_list_of_tuples_duplicated_names(
 
 @pytest.mark.parametrize("obj", (object(), None))
 async def test_invalid_data(session: ClientSession, obj: object) -> None:
-    with pytest.raises(TypeError, matches="expected str"):
+    with pytest.raises(TypeError, match="expected str"):
         await session.post("http://example.test/", data={"some": obj})
 
 
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -209,7 +209,7 @@ async def test_merge_headers_with_list_of_tuples_duplicated_names(
 
 @pytest.mark.parametrize("obj", (object(), None))
 async def test_invalid_data(session: ClientSession, obj: object) -> None:
-    with pytest.raises(TypeError):
+    with pytest.raises(TypeError, matches="expected str"):
         await session.post("http://example.test/", data={"some": obj})
 
 
--- a/tests/test_client_session.py
+++ b/tests/test_client_session.py
@@ -207,6 +207,12 @@ async def test_merge_headers_with_list_of_tuples_duplicated_names(
     ]
 
 
+@pytest.mark.parametrize("obj", (object(), None))
+async def test_invalid_data(session: ClientSession, obj: object) -> None:
+    with pytest.raises(TypeError):
+        await session.post("http://example.test/", data={"some": obj})
+
+
 async def test_http_GET(session: ClientSession, params: _Params) -> None:
     with mock.patch(
         "aiohttp.client.ClientSession._request", autospec=True, spec_set=True
--- a/tests/test_formdata.py
+++ b/tests/test_formdata.py
@@ -36,14 +36,14 @@ def test_formdata_multipart(buf: bytearray) -> None:
 def test_invalid_formdata_payload_multipart(obj: object) -> None:
     form = FormData()
     form.add_field("test", obj, filename="test.txt")
-    with pytest.raises(TypeError):
+    with pytest.raises(TypeError, matches="expected str"):
         form()
 
 
 @pytest.mark.parametrize("obj", (object(), None))
 def test_invalid_formdata_payload_urlencoded(obj: object) -> None:
     form = FormData({"test": obj})
-    with pytest.raises(TypeError):
+    with pytest.raises(TypeError, matches="expected str"):
         form()
 
 
--- a/tests/test_formdata.py
+++ b/tests/test_formdata.py
@@ -36,14 +36,14 @@ def test_formdata_multipart(buf: bytearray) -> None:
 def test_invalid_formdata_payload_multipart(obj: object) -> None:
     form = FormData()
     form.add_field("test", obj, filename="test.txt")
-    with pytest.raises(TypeError, matches="expected str"):
+    with pytest.raises(TypeError, match="expected str"):
         form()
 
 
 @pytest.mark.parametrize("obj", (object(), None))
 def test_invalid_formdata_payload_urlencoded(obj: object) -> None:
     form = FormData({"test": obj})
-    with pytest.raises(TypeError, matches="expected str"):
+    with pytest.raises(TypeError, match="expected str"):
         form()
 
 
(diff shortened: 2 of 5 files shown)


Dreamsorcerer:
I made this branch 3 weeks ago, but didn't open a PR...
There might have been an issue with tests expecting an int, in which case this probably needs to be tweaked.

Dreamsorcerer:
Should we just restrict it to strings and only put it in v4? I'm not sure it really makes sense to allow non-str, given that urlencode() will just str() any arguments, and seems to lead to unexpected behaviour. There's probably a fair number of ints in the wild though, so probably don't want to push it in v3.

bdraco:
Need to think about this one. Yarl is very flexible about encoding params to string for query string and we know that’s heavily used…. But all the type checks add up quick.  Maybe that’s not even relevant for this case

Dreamsorcerer:
> Need to think about this one. Yarl is very flexible about encoding params to string for query string and we know that’s heavily used…. But all the type checks add up quick. Maybe that’s not even relevant for this case

Yeah, this is about FormData creating 'x-www-form-urlencoded' data. Yarl isn't involved here.

bdraco (approved):
I think it’s ok. I probably wouldn’t backport it though

tests/test_client_session.py
```diff
@@ -207,6 +207,12 @@ async def test_merge_headers_with_list_of_tuples_duplicated_names(
     ]
 
 
+@pytest.mark.parametrize("obj", (object(), None))
+async def test_invalid_data(session: ClientSession, obj: object) -> None:
+    with pytest.raises(TypeError):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, matches="expected str"):
```

tests/test_formdata.py
```diff
@@ -32,9 +32,17 @@ def test_formdata_multipart(buf: bytearray) -> None:
     assert form.is_multipart
 
 
-def test_invalid_formdata_payload() -> None:
+@pytest.mark.parametrize("obj", (object(), None))
+def test_invalid_formdata_payload_multipart(obj: object) -> None:
     form = FormData()
-    form.add_field("test", object(), filename="test.txt")
+    form.add_field("test", obj, filename="test.txt")
+    with pytest.raises(TypeError):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, matches="expected str"):
```

tests/test_formdata.py
```diff
@@ -32,9 +32,17 @@ def test_formdata_multipart(buf: bytearray) -> None:
...
+        form()
+
+
+@pytest.mark.parametrize("obj", (object(), None))
+def test_invalid_formdata_payload_urlencoded(obj: object) -> None:
+    form = FormData({"test": obj})
     with pytest.raises(TypeError):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, matches="expected str"):
```

tests/test_client_session.py
```diff
@@ -207,6 +207,12 @@ async def test_merge_headers_with_list_of_tuples_duplicated_names(
     ]
 
 
+@pytest.mark.parametrize("obj", (object(), None))
+async def test_invalid_data(session: ClientSession, obj: object) -> None:
+    with pytest.raises(TypeError, matches="expected str"):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, match="expected str"):
```

tests/test_formdata.py
```diff
@@ -32,10 +32,18 @@ def test_formdata_multipart(buf: bytearray) -> None:
...
-    with pytest.raises(TypeError):
+    form.add_field("test", obj, filename="test.txt")
+    with pytest.raises(TypeError, matches="expected str"):
+        form()
+
+
+@pytest.mark.parametrize("obj", (object(), None))
+def test_invalid_formdata_payload_urlencoded(obj: object) -> None:
+    form = FormData({"test": obj})
+    with pytest.raises(TypeError, matches="expected str"):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, match="expected str"):
        form()


@pytest.mark.parametrize("obj", (object(), None))
def test_invalid_formdata_payload_urlencoded(obj: object) -> None:
    form = FormData({"test": obj})
    with pytest.raises(TypeError, match="expected str"):
```

tests/test_formdata.py
```diff
@@ -32,10 +32,18 @@ def test_formdata_multipart(buf: bytearray) -> None:
...
 
 
-def test_invalid_formdata_payload() -> None:
+@pytest.mark.parametrize("obj", (object(), None))
+def test_invalid_formdata_payload_multipart(obj: object) -> None:
     form = FormData()
-    form.add_field("test", object(), filename="test.txt")
-    with pytest.raises(TypeError):
+    form.add_field("test", obj, filename="test.txt")
+    with pytest.raises(TypeError, match="expected str"):
```
Dreamsorcerer:
```suggestion
    with pytest.raises(TypeError, match="Can not serialize value"):
```

PR #9295, opened 2024-09-25
Avoid locking to compress when it can happen synchronously

bdraco:
## What do these changes do?

Avoid locking to compress when it can happen synchronously

If we are not going to await, there is no need to lock.  This is the most common case for WebSocket frames.

no

no

The locking was a non-trivial amount of the time
Screenshot 2024-09-25 at 2 01 54 PM

### DIFF
--- a/aiohttp/compression_utils.py
+++ b/aiohttp/compression_utils.py
@@ -69,25 +69,33 @@ class ZLibCompressor(ZlibBaseHandler):
     def compress_sync(self, data: bytes) -> bytes:
         return self._compressor.compress(data)
 
-    async def compress(self, data: bytes) -> bytes:
+    async def compress(self, data: bytes, flush_mode: Optional[int] = None) -> bytes:
         """Compress the data.
 
         If there is a max_sync_chunk_size, then the compression
         will happen in the executor.
         """
+        compressor = self._compressor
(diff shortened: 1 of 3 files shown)


bdraco:
github is broken Error: Name or service not known (internal-api.service.iad.github.net:443)

aiohttp/compression_utils.py
```diff
@@ -70,19 +70,24 @@ def compress_sync(self, data: bytes) -> bytes:
...
             # compress at a time.
-            if (
-                self._max_sync_chunk_size is not None
-                and len(data) > self._max_sync_chunk_size
-            ):
-                return await asyncio.get_event_loop().run_in_executor(
-                    self._executor, self.compress_sync, data
-                )
-            return self.compress_sync(data)
+            return await asyncio.get_event_loop().run_in_executor(
```
bdraco:
```suggestion
            return await asyncio.get_running_loop().run_in_executor(
```

bdraco:
I forgot why we did this... it won't work because we have mixed sizes and one will wait for the lock but not the other.

Dreamsorcerer:
^ Remember to delete the branch if it's not relevant anymore.

webknjaz:
> Remember to delete the branch

We should really migrate to keeping the branches in forks.. I'll look into enforcing that sometime.

PR #9301, opened 2024-09-27
Replace code that can now be handled by yarl

bdraco:
Needs yarl bumped to 1.13.0+

## What do these changes do?

- The manual join of the query string in ClientRequest.send can be replaced with url.raw_path_qs

- All the IPv6 bracket checks can be replaced with url.host_subcomponent with yarl 1.13.0+

no

no

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -93,7 +93,7 @@ if TYPE_CHECKING:
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
 _YARL_SUPPORTS_HOST_SUBCOMPONENT = tuple(map(int, yarl_version.split(".")[:2])) >= (
     1,
-    11,
+    13,
 )
 
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -27,7 +27,7 @@ from typing import (
 )
 
 from multidict import CIMultiDict, CIMultiDictProxy, MultiDict, MultiDictProxy
-from yarl import URL
+from yarl import URL, __version__ as yarl_version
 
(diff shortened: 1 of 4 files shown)


aiohttp/client_reqrep.py
```diff
@@ -91,6 +91,10 @@
 
 
 _CONTAINS_CONTROL_CHAR_RE = re.compile(r"[^-!#$%&'*+.^_`|~0-9a-zA-Z]")
+_YARL_SUPPORTS_HOST_SUBCOMPONENT = tuple(map(int, yarl_version.split(".")[:2])) >= (
+    1,
+    11,
```
```suggestion
    13,
```

bdraco:
I'm almost tempted to bump minimum yarl again so we don't have maintain the branching here

Dreamsorcerer:
> I'm almost tempted to bump minimum yarl again so we don't have maintain the branching here

Just bump it in 3.11 and the branching can sit in the 3.10 release for a short while.

aiohttp/client_reqrep.py
```diff
@@ -606,17 +614,19 @@ async def send(self, conn: "Connection") -> "ClientResponse":
...
-            if helpers.is_ipv6_address(connect_host):
-                connect_host = f"[{connect_host}]"
+            if _YARL_SUPPORTS_HOST_SUBCOMPONENT:
+                connect_host = self.url.host_subcomponent
+                assert connect_host is not None
+            else:
+                connect_host = self.url.raw_host
+                assert connect_host is not None
+                if helpers.is_ipv6_address(connect_host):
+                    connect_host = f"[{connect_host}]"
```
bdraco, line 624: https://app.codecov.io/gh/aio-libs/aiohttp/blob/master/aiohttp%2Fclient_reqrep.py#L611

This missing coverage is existing, but will soon disappear after https://github.com/aio-libs/aiohttp/pull/9305 and a followup to remove this branch

PR #9309, opened 2024-09-27
Fix building the URL in BaseRequest when the host contains a port or IPv6 address

### LINKED ISSUE #9307: BaseRequest.url passes the host and port to `URL.build` when the port is included in the host header (non-default port) -- fixed in 3.10.7
request.host can be `127.0.0.1:8123`

https://github.com/aio-libs/aiohttp/blob/0b8be7f6c8db0d59afe02109df9be874d0d637e6/aiohttp/web_request.py#L447

yarl 1.13.0+ caught this because host is now validated via https://github.com/aio-libs/yarl/pull/954

Pinning yarl to < 1.13.0 will also prevent the issue since the validation doesn't happen until that version.

bdraco:
## What do these changes do?

The port was incorrectly being passed in the `host` field to `URL.build` when the `Host` header included the `port` 

Calling `BaseRequest.url` will not fail with yarl 1.13.0+ when the `host` header has a port

no
## Related issue number
fixes #9307

### DIFF
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -449,12 +449,9 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
     @reify
     def url(self) -> URL:
         """The full URL of the request."""
-        host, has_port, port = self.host.partition(":")
-        if has_port:
-            url = URL.build(scheme=self.scheme, host=host, port=port)
-        else:
-            url = URL.build(scheme=self.scheme, host=host)
-        return url.join(self._rel_url)
+        # authority is used here because it may include the port number
+        # and we want yarl to parse it correctly
+        return URL.build(scheme=self.scheme, authority=self.host).join(self._rel_url)
 
     @reify
     def path(self) -> str:
--- a/aiohttp/web_request.py
+++ b/aiohttp/web_request.py
@@ -421,6 +421,10 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
         - overridden value by .clone(host=new_host) call.
         - HOST HTTP header
         - socket.getfqdn() value
+
+        For example, 'example.com' or 'localhost:8080'.
+
+        For historical reasons, the port number may be included.
         """
         host = self._message.headers.get(hdrs.HOST)
         if host is not None:
@@ -444,7 +448,12 @@ class BaseRequest(MutableMapping[str, Any], HeadersMixin):
 
     @reify
     def url(self) -> URL:
-        url = URL.build(scheme=self.scheme, host=self.host)
+        """The full URL of the request."""
+        host, has_port, port = self.host.partition(":")
+        if has_port:
+            url = URL.build(scheme=self.scheme, host=host, port=port)
+        else:
+            url = URL.build(scheme=self.scheme, host=host)
         return url.join(self._rel_url)
 
     @reify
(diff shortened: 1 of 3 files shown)


bdraco:
Will need a test as well .. but super late here so I'll have to pickup tomorrow

aiohttp/web_request.py
```diff
@@ -444,7 +448,12 @@ def remote(self) -> Optional[str]:
 
     @reify
     def url(self) -> URL:
-        url = URL.build(scheme=self.scheme, host=self.host)
+        """The full URL of the request."""
+        host, has_port, port = self.host.partition(":")
+        if has_port:
+            url = URL.build(scheme=self.scheme, host=host, port=port)
```
bdraco:
```suggestion
            url = URL.build(scheme=self.scheme, host=host, port=int(port))
```

aiohttp/web_request.py
```diff
@@ -444,7 +448,12 @@ def remote(self) -> Optional[str]:
 
     @reify
     def url(self) -> URL:
-        url = URL.build(scheme=self.scheme, host=self.host)
+        """The full URL of the request."""
+        host, has_port, port = self.host.partition(":")
```
bdraco: This isn't sufficient because it could be an IPv6 and we need to handle brackets

bdraco: https://github.com/aio-libs/yarl/blob/38ef0f95d25c240e17963971a41ed85b0b0c4e31/yarl/_url.py#L1089

bdraco:
It's even more messy because when it's cloned we use the host but host is the raw host in this case. Actually it's even the host subcomponent 

There aren't any tests for handling IPv6 hosts correctly either

We also probably want to let yarl parse the raw host instead of using URL.build

bdraco:
Maybe the simplest fix is to pass it as authority instead of host

bdraco:
> Maybe the simplest fix is to pass it as authority instead of host 

But that's really inefficient because it creates new split results and reparses it 4x

https://github.com/aio-libs/yarl/blob/38ef0f95d25c240e17963971a41ed85b0b0c4e31/yarl/_url.py#L371

Should fix that to use the same logic in __new__ but will need a lot of tests

aiohttp/web_request.py
```diff
@@ -444,8 +448,10 @@ def remote(self) -> Optional[str]:
 
     @reify
     def url(self) -> URL:
-        url = URL.build(scheme=self.scheme, host=self.host)
-        return url.join(self._rel_url)
+        """The full URL of the request."""
+        # authority is used here because it may include the port number
+        # and we want yarl to parse it correctly
+        return URL.build(scheme=self.scheme, authority=self.host).join(self._rel_url)
```
bdraco, line 454: Sadly this is ~63x slower when passing `authority` because it has to now parse it and than reparse is 4x because of how urllib handles `SplitResult`

bdraco:
New tests fail with the following on master before the fix

```
FAILED tests/test_web_request.py::test_url_non_default_port - ValueError: Host 'example.com:8123' cannot contain ':' (at position 11)
FAILED tests/test_web_request.py::test_url_ipv6 - ValueError: Host '[::1]:8123' cannot contain '[' (at position 0)
```

aiohttp/web_request.py
bdraco: We already have a fast helper in yarl for that so we can get it to be much faster https://github.com/aio-libs/yarl/pull/1163

PR #9312, opened 2024-09-27
replace aioredis with valkey in examples

### LINKED ISSUE #9283: Replace aioredis in example
We have aioredis in one of our examples:
https://github.com/aio-libs/aiohttp/blob/master/examples/background_tasks.py

This should be migrated to valkey (aioredis was merged into redis, then redis was forked to valkey to keep it open source).

Should be simple to update the example and then update the requirements files.

Cycloctane:
## What do these changes do?

- replace aioredis with valkey in background tasks examples.
- update the requirements files.

## Related issue number

Fixes #9283 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/examples/background_tasks.py
+++ b/examples/background_tasks.py
@@ -3,11 +3,11 @@
 import asyncio
 from typing import List
 
-import aioredis
+import valkey.asyncio as valkey
 
 from aiohttp import web
 
-redis_listener = web.AppKey("redis_listener", asyncio.Task[None])
+valkey_listener = web.AppKey("valkey_listener", asyncio.Task[None])
 websockets = web.AppKey("websockets", List[web.WebSocketResponse])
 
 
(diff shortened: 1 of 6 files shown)


examples/background_tasks.py
```diff
@@ -29,32 +29,35 @@ async def on_shutdown(app: web.Application) -> None:
...
-    app[redis_listener] = asyncio.create_task(listen_to_redis(app))
+    app[valkey_listener] = asyncio.create_task(listen_to_valkey(app))
 
 
 async def cleanup_background_tasks(app: web.Application) -> None:
     print("cleanup background tasks...")
-    app[redis_listener].cancel()
-    await app[redis_listener]
+    app[valkey_listener].cancel()
+    await app[valkey_listener]
```
Dreamsorcerer: While we're here, can we change this to a cleanup_ctx, as per:
https://docs.aiohttp.org/en/stable/web_advanced.html#complex-applications

Dreamsorcerer (approved):

examples/background_tasks.py
Dreamsorcerer: I think it'd also be slightly cleaner for us to suppress the CancelledError here (as per the example), rather than inside the listener function.

Dreamsorcerer:
Thanks for that. If you could handle the backports too, that'd be excellent.

PR #9333, opened 2024-09-28
Small cleanups to connector ConnectionKey usage

bdraco:
## What do these changes do?

- Add some missing typing
- Reduce some indent
- Reduce some dict lookups

no

no


bdraco:
Tested on production ok, but too tired so leaving it for tomorrow to look again after sleep

bdraco:
Still looks good after sleep

PR #9340, opened 2024-09-29
Use dunder writer internally in ClientResponse

bdraco:
## What do these changes do?

In #7815 the writer was wrapped with a setter to make sure it was always reset on completion and maintain compat for subclasses. This added a tiny bit of overhead since we access self._writer in a lot of places. We can access self.__writer instead since its all inside the class which has less overhead.

We still have a few spikes that would be nice to sort out, but they are getting to be more consistent
Screenshot 2024-09-29 at 8 43 17 AM

no

no

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -677,7 +677,7 @@ class ClientRequest:
         self.response = response_class(
             self.method,
             self.original_url,
-            writer=self._writer,
+            writer=self.__writer,
             continue100=self._continue,
             timer=self._timer,
             request_info=self.request_info,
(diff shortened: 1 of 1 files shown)


aiohttp/client_reqrep.py
```diff
@@ -761,7 +761,7 @@ def __init__(
         self._real_url = url
         self._url = url.with_fragment(None)
         self._body: Optional[bytes] = None
-        self._writer: Optional[asyncio.Task[None]] = writer
```
bdraco, line 764: typing was redundant as its a property now

PR #9344, opened 2024-09-29
Remove unused ip address helpers

bdraco:
## What do these changes do?

Remove `is_ipv6_address` and `is_ipv4_address` helpers as they are not used in the codebase anymore.  Not back-ported to 3.10 since there is still back-compat code that uses them.

Remove support for `bytes`/`bytesarray`/`memoryview` from `is_ip_address` as its never used.  Also passing a `memoryview` did not work so there was no chance it was being used.

The helpers have been removed. They are not expected to be called externally, however in case someone was using them, this was marked as a breaking change.

no


bdraco:
memoryview didn't actually work

also we never pass anything but string

PR #9348, opened 2024-09-29
Add new send_frame method to WebSockets

bdraco:
## What do these changes do?

Currently there is no way to send TEXT frames when the source data is a UTF-8 encoded bytes object without having decode the bytes to string before passing it to `send_str`, and than having the WebSocket internals encode it back to `bytes`. When sending at volume, the extra decode/encode cycle is a significant bottleneck.

no

no

### DIFF
--- a/docs/client_reference.rst
+++ b/docs/client_reference.rst
@@ -1600,13 +1600,14 @@ manually.
       Send a :const:`~aiohttp.WSMsgType` message *message* to peer.
 
       This method is low-level and should be used with caution as it
-      only accepts UTF-8 encoded bytes for *message*.
+      only accepts bytes which must conform to the correct message type
+      for *message*.
 
       It is recommended to use the :meth:`send_str`, :meth:`send_bytes`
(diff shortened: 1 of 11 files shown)


docs/client_reference.rst
```diff
@@ -1594,6 +1594,31 @@ manually.
          The method is converted into :term:`coroutine`,
          *compress* parameter added.
 
+   .. method:: send_frame(message, opcode, compress=None)
+      :async:
+
+      Send a :const:`~aiohttp.WSMsgType` message *message* to peer.
+
+      This method is low-level and should be used with caution as it
+      only accepts UTF-8 encoded bytes for *message*.
```
Dreamsorcerer: Isn't this untrue if opcode is BINARY? Maybe more accurate to say that the bytes must conform to the correct message type.

bdraco: Yes, thats better

Dreamsorcerer (approved):

bdraco:
Thanks for the review

PR #9365, opened 2024-10-01
Speed up the ConnectionKey

### LINKED ISSUE #9364: ConnectionKey hashing and equality has an unexpected impact on performance between master and 3.x
https://github.com/aio-libs/aiohttp/blob/e3b1011f2146ad0faa4c3d3c29f26b73e1400564/aiohttp/client_reqrep.py#L154

We have to hash the `ConnectionKey` quite a bit.

On 3.x its a attrs with the hash cached
attrs

On master its a dataclass no cache
dataclass

bdraco:
A large part of the connection management is constructing, hashing and comparing the connection key. I noticed this when I was profiling and looking at `py-spy` output to compare which code paths `aiohappyeyeballs` took on `master` and `3.10`

For additional details see #9364

To speed this up, `ConnectionKey` is now a `NamedTuple` which allows all the `__hash__` and `__eq__` operations to happen in native code.

https://github.com/python/cpython/blob/133e929a791d209b578b4822a7a07f4570b3803b/Objects/tupleobject.c#L319

The construction of the object is a bit faster by calling the `yarl.URL` methods directly.

I didn't find any third party code that was altering modifying `connection_key` https://github.com/search?q=req.connection_key&type=code so its likely this change is safe to backport to 3.10 but I was a bit worried about custom proxy Connectors (although that seems unlikely) so I marked it as breaking and only labeled it for 3.11.

## What do these changes do?

Speed up connections

The `ConnectionKey` is now a `NamedTuple`

no

fixes #9364


aiohttp/client_reqrep.py
```diff
@@ -150,8 +151,10 @@ def check(self, transport: asyncio.Transport) -> None:
     SSL_ALLOWED_TYPES = (bool,)
 
 
-@dataclasses.dataclass(frozen=True)
-class ConnectionKey:
+_SSL_SCHEMES = frozenset(("https", "wss"))
+
+
+class ConnectionKey(NamedTuple):
```
bdraco, line 160: Probably could use a comment here about why a NamedTuple is used

bdraco:
Here is what one of attrs ConnectionKey dict gets looks like (lots of `__hash__` and `__eq__`) 
Screenshot 2024-10-01 at 9 51 43 AM

bdraco:
Screenshot 2024-10-03 at 6 12 56 PM

Ever with the cached hash on attrs, the eq shows up in the profile often in 3.10 still.
Will be super nice to get this in 3.11

PR #9366, opened 2024-10-01
Small speed up to update_headers

bdraco:
## What do these changes do?

- Avoid checking to see if the port is default if no port is set: For most cases we can avoid the `is_default_port` check which has to work out the default port internally in yarl since if there is no explict port set, it must be the default port
- Avoid using strip for a single char check

Event if we have to call `is_default_port`, `explict_port` will already be cached
Screenshot 2024-10-01 at 4 33 00 AM

no

no

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -360,7 +360,7 @@ class ClientRequest:
         assert netloc is not None
         # See https://github.com/aio-libs/aiohttp/issues/3636.
         if netloc[-1] == ".":
-            netloc = netloc[:-1]
+            netloc = netloc.rstrip(".")
         explicit_port = self.url.explicit_port
         if explicit_port is not None and not self.url.is_default_port():
             netloc = f"{netloc}:{explicit_port}"
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -379,7 +379,7 @@ class ClientRequest:
 
         self.headers[hdrs.HOST] = host
 
-        if headers:
+        if not headers:
             return
 
         if isinstance(headers, (dict, MultiDictProxy, MultiDict)):
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -380,15 +380,17 @@ class ClientRequest:
         self.headers[hdrs.HOST] = host
 
         if headers:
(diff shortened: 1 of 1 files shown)


aiohttp/client_reqrep.py
```diff
@@ -359,9 +359,11 @@ def update_headers(self, headers: Optional[LooseHeaders]) -> None:
...
         assert netloc is not None
         # See https://github.com/aio-libs/aiohttp/issues/3636.
-        netloc = netloc.rstrip(".")
-        if self.url.port is not None and not self.url.is_default_port():
-            netloc += ":" + str(self.url.port)
+        if netloc[-1] == ".":
+            netloc = netloc.rstrip(".")
+        explicit_port = self.url.explicit_port
+        if explicit_port is not None and not self.url.is_default_port():
+            netloc = f"{netloc}:{explicit_port}"
```
Dreamsorcerer: Isn't the first check redundant?
```suggestion
        if not self.url.is_default_port():
            netloc = f"{netloc}:{self.url.explicit_port}"
```

aiohttp/client_reqrep.py
```diff
@@ -359,9 +359,11 @@ def update_headers(self, headers: Optional[LooseHeaders]) -> None:
         netloc = self.url.host_subcomponent
         assert netloc is not None
         # See https://github.com/aio-libs/aiohttp/issues/3636.
-        netloc = netloc.rstrip(".")
-        if self.url.port is not None and not self.url.is_default_port():
-            netloc += ":" + str(self.url.port)
+        if netloc[-1] == ".":
+            netloc = netloc.rstrip(".")
```
Dreamsorcerer: This seems like the kind of micro-optimisation that looks like redundant code and could easily be refactored back to the previous form by a future contributor...

aiohttp/client_reqrep.py
bdraco: `is_default_port` is never cached in `yarl` (and it could be optimized a bit more) so checking the cached property first is a guard to avoid calling `is_default_port`.  This needs a comment to explain why its first.

aiohttp/client_reqrep.py
bdraco: Definitely needs a comment here.  I'll also better explain why its done.

PR #9367, opened 2024-10-01
Speed up handling auth in urls

bdraco:
## What do these changes do?

Avoids calling the .user and .password methods if the raw versions are not set as auth in urls is going to be the rare case.

The cache of `raw_password` and `raw_user` is guaranteed to be set in yarl once the netloc is parsed

no
no

before

Screenshot 2024-10-01 at 4 50 34 AM

after
Screenshot 2024-10-01 at 4 49 06 AM

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -332,9 +332,8 @@ class ClientRequest:
             raise InvalidURL(url)
 
         # basic auth info
-        username, password = url.user, url.password
-        if username or password:
-            self.auth = helpers.BasicAuth(username or "", password or "")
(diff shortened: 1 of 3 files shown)


aiohttp/client_reqrep.py
```diff
@@ -332,9 +332,8 @@ def update_host(self, url: URL) -> None:
             raise InvalidURL(url)
 
         # basic auth info
-        username, password = url.user, url.password
-        if username or password:
-            self.auth = helpers.BasicAuth(username or "", password or "")
+        if url.raw_user is not None or url.raw_password is not None:
```
```suggestion
        if url.raw_user or url.raw_password:
```

bdraco:
needs comments and a test

PR #9368, opened 2024-10-01
Avoid using the proxy headers in the ConnectionKey if no proxy is in use

bdraco:
## What do these changes do?

Default proxy headers may be passed to the connector to be included in the event a proxy is being used. If no proxy is being used, we should not include the proxy auth and proxy headers in the `ConnectionKey` since it means we could end up creating a connection when one was already available for reuse

no

no


bdraco:
I was wondering why the proxy headers were always being considered in the connection key for Home Assistant even when no proxy is in use.   Its because default headers apply to proxy as well https://github.com/aio-libs/aiohttp/blob/e3b1011f2146ad0faa4c3d3c29f26b73e1400564/aiohttp/client.py#L474

bdraco:
needs a test

PR #9386, opened 2024-10-01
Small speed up to starting client requests

bdraco:
## What do these changes do?

Avoid some checks for uncommon cases

Avoid `lower()`ing default values that are rarely set.

no

no


bdraco:
master
aiohttp_master


3.9.5

Issue #4618, opened 2020-03-10
Memory leak for simple get call


dima-kov:
🐞 **Describe the bug**

I have faced with memory leak in my program: after loading `100k` URLs the process eats additional 300mb of RAM. I had extracted this piece of code, that reproduces memory issue (below) and ran it with `mprof` for 5 min for loading 4k urls:

![aiohttp-memory-leak](https://user-images.githubusercontent.com/8599032/76341241-6d6bf780-6305-11ea-8482-8de4f6ea784a.png)

💡 **To Reproduce**

Use env:
```
aiohttp==4.0.0a1
async-timeout==3.0.1
attrs==19.3.0
chardet==3.0.4
cycler==0.10.0
idna==2.9
kiwisolver==1.1.0
matplotlib==3.2.0
memory-profiler==0.57.0
multidict==4.7.5
numpy==1.18.1
psutil==5.7.0
pyparsing==2.4.6
python-dateutil==2.8.1
six==1.14.0
typing-extensions==3.7.4.1
uvloop==0.14.0
yarl==1.4.2
```
Tested with different versions of aiohttp (>3.5)

Code:
```
import asyncio
import uvloop
import aiohttp

url = 'https://docs.mapbox.com/mapbox-gl-js/assets/earthquakes.geojson'

load_times = 4000


def get_to_load():
    global load_times
    if load_times > 0:
        load_times -= 1
        return url
    return None


async def fetch(session, url):
    response = await session.get(url)
    return await response.text()


async def load(session, worker_id):
    to_load = get_to_load()
    while to_load is not None:
        r = await fetch(session, to_load)
        print('Done', worker_id)
        await asyncio.sleep(0.2)
        to_load = get_to_load()


async def main(workers_num=90):
    async with aiohttp.ClientSession() as session:
        await asyncio.gather(*[load(session, i) for i in range(workers_num)])


asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
loop = asyncio.get_event_loop()

asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
asyncio.run(asyncio.sleep(3))
asyncio.run(main())
asyncio.run(asyncio.sleep(5))
```
3. Run for 5 min with profiler.

💡 **Expected behavior**
Keep memory usage near straight during the whole program run.

📋 **Your version of the Python**
Python 3.7.5 (Also tested on 3.8)

📋 **Your version of the aiohttp/yarl/multidict distributions**

[...]

PR #9406, opened 2024-10-03
Add __slots__ to timer helpers

bdraco:
## What do these changes do?

We use `__slots__` almost everywhere else in the codebase, however `__slots__` was missing for these helpers

no

no

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,6 @@ skip = '.git,*.pdf,*.svg,Makefile,CONTRIBUTORS.txt,venvs,_build'
 ignore-words-list = 'te,assertIn'
 
 [tool.slotscheck]
-# Remove aiohttp.helpers once https://github.com/python/cpython/pull/106771
+# TODO(3.13): Remove aiohttp.helpers once https://github.com/python/cpython/pull/106771
 # is available in all supported cpython versions
 exclude-modules = "(^aiohttp\\.helpers)"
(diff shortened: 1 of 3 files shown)


bdraco:
https://github.com/python/cpython/pull/106771

pyproject.toml
```diff
@@ -82,3 +82,8 @@ skip = "pp*"
 [tool.codespell]
 skip = '.git,*.pdf,*.svg,Makefile,CONTRIBUTORS.txt,venvs,_build'
 ignore-words-list = 'te,assertIn'
+
+[tool.slotscheck]
+# Remove require-superclass = false once https://github.com/python/cpython/pull/106771
+# is available in all supported cpython versions
```
bdraco, line 88: I don't really like this. I think this check is valuable.  It's probably better to exclude helpers.py instead

pyproject.toml
```diff
@@ -82,3 +82,8 @@ skip = "pp*"
 [tool.codespell]
 skip = '.git,*.pdf,*.svg,Makefile,CONTRIBUTORS.txt,venvs,_build'
 ignore-words-list = 'te,assertIn'
+
+[tool.slotscheck]
+# Remove aiohttp.helkpers once https://github.com/python/cpython/pull/106771
```
bdraco:
```suggestion
# Remove aiohttp.helpers once https://github.com/python/cpython/pull/106771
```

pyproject.toml
```diff
@@ -82,3 +82,8 @@ skip = "pp*"
 [tool.codespell]
 skip = '.git,*.pdf,*.svg,Makefile,CONTRIBUTORS.txt,venvs,_build'
 ignore-words-list = 'te,assertIn'
+
+[tool.slotscheck]
+# Remove aiohttp.helpers once https://github.com/python/cpython/pull/106771
```
Dreamsorcerer: Appears to be 3.13 minimum.
```suggestion
# TODO(3.13): Remove aiohttp.helpers once https://github.com/python/cpython/pull/106771
```

PR #9407, opened 2024-10-03
Add __slots__ to stream classes

bdraco:
## What do these changes do?

We use `__slots__` almost everywhere else in the codebase, however `__slots__` were not implemented in stream

no

no

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -129,7 +129,6 @@ class StreamReader(AsyncStreamReaderMixin):
         "_exception",
         "_timer",
         "_eof_callbacks",
-        "total_bytes",
     )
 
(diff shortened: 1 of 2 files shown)


aiohttp/streams.py
```diff
@@ -134,6 +161,8 @@ def __init__(
         self._exception: Optional[Union[Type[BaseException], BaseException]] = None
         self._timer = TimerNoop() if timer is None else timer
         self._eof_callbacks: List[Callable[[], None]] = []
+        self._eof_counter = 0
```
bdraco, line 164: this was being defined outside of `__init__`

Dreamsorcerer (approved):

PR #9421, opened 2024-10-05
Fix all waiters being enumerated on connection release

bdraco:
## What do these changes do?

#2964 changed the waiter implementation to enumerate all the waiters on release. This may be to handle a case where waiters are not properly released, however I could not find one where we should be releasing waiters for other keys when we already have the key

no

no

### DIFF
--- a/CHANGES/9421.bugfix.rst
+++ b/CHANGES/9421.bugfix.rst
@@ -1,0 +1,1 @@
+Fix the connection limit so that the algorithm does not enumerate unrelated connections -- by user:`bdraco`.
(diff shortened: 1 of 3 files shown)


bdraco:
This is going to need some more test coverage for different keys

Dreamsorcerer (approved):

bdraco:
I've run through all the tests and this is already well covered. I can't think of a case where it would ever make sense to release the waiters for a different connection key.

CHANGES/9421.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix the connection limit so that the algorithm does not enumerate unrelated connections -- by user:`bdraco`.
```
bdraco:
```suggestion
Fix the connection limit so that the algorithm does not enumerate unrelated connection waiters -- by user:`bdraco`.
```

bdraco:
Digging in a bit more, this is actually correct because we could hit the overall limit but not the per host limit so we have to enumerate them all since any connection key being released could mean we bring it under the overall limit.

PR #9396, opened 2024-10-02
Disabling retry_persistent_connection in Tests

ShubhAgarwal-dev:
## What do these changes do?

It disables retry logic by default in tests.

## Are there changes in behavior for the user?

No

## Is it a substantial burden for the maintainers to support this?

No

## Related issue number

#9141 

## Checklist

- [X] I think the code is well written
- [ ] Unit tests for the changes exist
- [x] Add a new news fragment into the `CHANGES/` folder
- [X] Documentation reflects the changes
- [X] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

PR #9423, opened 2024-10-06
Patchback/backports/3.11/6d8562d8420b3e464877b941e04663fdac29a54c/pr 9396

ShubhAgarwal-dev:
Follow up to PR #9396.

I might have done a bad job of merging myself. Sorry for the inconvenience caused.


aiohttp/client.py
```diff
@@ -225,36 +225,34 @@ class ClientTimeout:
...
-            "_response_class",
-            "_ws_response_class",
-            "_trace_configs",
-            "_read_bufsize",
-            "_max_line_size",
-            "_max_field_size",
-            "_resolve_charset",
-            "_default_proxy",
-            "_default_proxy_auth",
-        ]
```
webknjaz: Yeah, this needs to be kept because it differs on this branch. The patch needs to be adapted for this branch to make an equivalent single-line change in diff.

tests/test_test_utils.py
```diff
@@ -334,9 +335,32 @@ def test_noop(self) -> None:
     result.stdout.fnmatch_lines(["*RuntimeError*"])
 
 
-async def test_server_context_manager(app, loop) -> None:
-    async with TestServer(app, loop=loop) as server:
-        async with aiohttp.ClientSession(loop=loop) as client:
```
webknjaz: Backports must not modify unrelated code. This should be reverted to its 3.11 state.

ShubhAgarwal-dev: Ok, I will do the same

aiohttp/client.py
ShubhAgarwal-dev: Ok

aiohttp/client.py
```diff
@@ -230,7 +230,6 @@ class ClientSession:
             "_base_url",
             "_source_traceback",
             "_connector",
-            "requote_redirect_url",
```
webknjaz, line 233: Keep this, everything else seems to correspond to the original patch.

webknjaz:
@ShubhAgarwal-dev pro tip: backport PRs must be open against corresponding stable branches. `3.11` in this case, not `master`. You can do this while creating the PR or edit later if you forget. I've changed it for you earlier. But perhaps next time you can do this yourself ;)

ShubhAgarwal-dev:
Can you explain what backport is..? I followed the instructions that were given by the patchbot. I will keep that in mind for the next time

Dreamsorcerer:
master represents a future v4 release. To get a feature released in a reasonable timeframe, it must also be merged into the relevant 3.x branch in order to make it into a 3.x release. Hence, backport.

Maybe the base branch for the PR needs to be included in the backport instructions, several people have also created PRs to master in the past, so this isn't the first time.

webknjaz:
@Dreamsorcerer I was also thinking that. I don't remember why I didn't include it. But it might be that this branch is unavailable in that context or something...

PR #9443, opened 2024-10-09
Fix handling of redirects with authentication

### LINKED ISSUE #9436: URLs that infinitely redirect to an authenticated Location will raise a ValueError instead of TooManyRedirects
### Describe the bug

I encountered a URL that seems to always respond with this, even for subsequent authenticated calls:

```http
HTTP/1.1 301 Moved Permanently
Server: nginx/1.18.0 (Ubuntu)
Date: Wed, 09 Oct 2024 07:12:26 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
X-redirected: 1
Location: http://bourospan@itheka.gr
```

This then leads to a `ValueError` once the redirect is handled, instead of a `TooManyRedirects` error (which is what both Chrome and `requests` throw for this).

### To Reproduce

```python
import aiohttp
import asyncio


async def main():
        await session.get('http://itheka.gr', allow_redirects=True)


if __name__ == "__main__":
    asyncio.run(main())
```

### Expected behavior

It should raise a `TooManyRedirectsError` instead. A `ValueError` gives the impression the user did something wrong, while this is out of their control.

### Logs/tracebacks

```python-traceback
>>> import aiohttp
>>> import asyncio
>>> 
>>> 
>>> async def main():
...     async with aiohttp.ClientSession() as session:
...         await session.get('http://itheka.gr', allow_redirects=True)
... 
>>> 
>>> if __name__ == "__main__":
...     asyncio.run(main())
... 
  File "<stdin>", line 2, in <module>
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/runners.py", line 118, in run
  File "/opt/homebrew/Cellar/python@3.12/3.12.4/Frameworks/Python.framework/Versions/3.12/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "<stdin>", line 3, in main
  File "/Users/peetersp/Documents/Swan/website-processors/venv/lib/python3.12/site-packages/aiohttp/client.py", line 591, in _request
    raise ValueError(
ValueError: Cannot combine AUTH argument with credentials encoded in URL
```

### Python Version

```console
$ python --version
Python 3.12.4
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.10.9
Author: 
Author-email: 
License: Apache 2
Location: /Users/peetersp/Documents/Swan/website-processors/venv/lib/python3.12/site-packages
Required-by:
```

### multidict Version

```console
Name: multidict
Version: 6.1.0
Author: Andrew Svetlov
License: Apache 2
Location: /Users/peetersp/Documents/Swan/website-processors/venv/lib/python3.12/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### propcache Version

```console
WARNING: Package(s) not found: propcache
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.12.1
Author: Andrew Svetlov
License: Apache-2.0
Location: /Users/peetersp/Documents/Swan/website-processors/venv/lib/python3.12/site-packages
Requires: idna, multidict
Required-by: aiohttp
```

### OS

Ubuntu, macOS

### Related component

Client

### Additional context

The root cause lies here:

https://github.com/aio-libs/aiohttp/blob/c4a3f276fab31b188c3fa27baa5aa1c5652ef516/aiohttp/client.py#L558-L562

I believe this could be fixed in one of two ways:

1. Checking whether `auth` and `auth_from_url` are different before raising
2. Only checking this for the initial URL and skipping the check for redirects

Option 2 is probably best, as the problem at hand is that URL-based auth can be out of the user's control, so it should only raise a `ValueError` if it is indeed the user that made a mistake, which can only really be the case for the initial URL in the redirect chain.

I'll happily PR this once I get the go-ahead for the method to use.

### Code of Conduct

PLPeeters:
## What do these changes do?

They make the client ignore auth clashes that are solely due to redirects, in addition to having redirect authentication take precedence over previously set authentication.

Users will no longer get a `ValueError` if a website suddenly includes multiple authenticated URLs in its redirect chain (see #9436 for an example).

While writing the test, I also wondered what should happen if we already have authentication set and get new authentication in a redirect URL. Mimicking what Chrome seems to be doing in this case, I opted to supersede the auth with that of the redirect. We can of course discuss this and tweak it if needed. I'm also on the fence on whether this warrants a separate PR, so I bundled it for now.

Probably not.

## Related issue number

Fixes #9436

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CHANGES/9436.bugfix.rst
+++ b/CHANGES/9436.bugfix.rst
@@ -1,0 +1,1 @@
+Authentication provided by a redirect now takes precedence over provided ``auth`` when making requests with the client -- by :user:`PLPeeters`.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -569,7 +569,9 @@ class ClientSession:
                             "credentials encoded in URL"
                         )
 
-                    if auth is None:
+                    # Override the auth with the one from the URL only if we
+                    # have no auth, or if we got an auth from a redirect URL
+                    if auth is None or (history and auth_from_url is not None):
                         auth = auth_from_url
 
                     if (
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -472,7 +472,7 @@ class ClientSession:
             data = payload.JsonPayload(json, dumps=self._json_serialize)
 
         redirects = 0
-        history = []
+        history: List[ClientResponse] = []
         version = self._version
         params = params or {}
 
@@ -560,7 +560,10 @@ class ClientSession:
                             else InvalidUrlClientError
                         )
                         raise err_exc_cls(url)
-                    if auth and auth_from_url:
+                    # If `auth` was passed for an already authenticated URL,
+                    # disallow only if this is the initial URL; this is to avoid issues
+                    # with sketchy redirects that are not the caller's responsibility
+                    if not history and (auth and auth_from_url):
                         raise ValueError(
                             "Cannot combine AUTH argument with "
                             "credentials encoded in URL"
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2930,6 +2930,7 @@ async def test_creds_in_auth_and_url() -> None:
 async def test_creds_in_auth_and_redirect_url(
     create_server_for_url_and_handler: Callable[[URL, Handler], Awaitable[TestServer]],
 ) -> None:
+    """Verify that credentials in redirect URLs can and do override any previous credentials."""
     url_from = URL("http://example.com")
     url_to = URL("http://user@example.com")
     redirected = False
@@ -2983,6 +2984,9 @@ async def test_creds_in_auth_and_redirect_url(
         assert len(resp.history) == 1
         assert str(resp.url) == "http://example.com"
         assert resp.status == 200
+        assert (
+            resp.request_info.headers.get("authorization") == "Basic dXNlcjo="
+        ), "Expected redirect credentials to take precedence over provided auth"
 
 
 @pytest.fixture
--- a/tests/test_client_functional.py
+++ b/tests/test_client_functional.py
@@ -2927,6 +2927,64 @@ async def test_creds_in_auth_and_url() -> None:
         await session.close()
 
 
+async def test_creds_in_auth_and_redirect_url(
+    create_server_for_url_and_handler: Callable[[URL, Handler], Awaitable[TestServer]],
+) -> None:
+    url_from = URL("http://example.com")
+    url_to = URL("http://user@example.com")
+    redirected = False
+
+    async def srv(request: web.Request) -> web.Response:
+        nonlocal redirected
+
+        assert request.host == url_from.host
+
+        if not redirected:
+            redirected = True
+            raise web.HTTPMovedPermanently(url_to)
+
+        return web.Response()
+
+    server = await create_server_for_url_and_handler(url_from, srv)
+
+    etc_hosts = {
+        (url_from.host, 80): server,
+    }
+
+    class FakeResolver(AbstractResolver):
+        async def resolve(
+            self,
+            host: str,
+            port: int = 0,
+            family: socket.AddressFamily = socket.AF_INET,
+        ) -> List[ResolveResult]:
+            server = etc_hosts[(host, port)]
+            assert server.port is not None
+
+            return [
+                {
+                    "hostname": host,
+                    "host": server.host,
+                    "port": server.port,
+                    "family": socket.AF_INET,
+                    "proto": 0,
+                    "flags": socket.AI_NUMERICHOST,
+                }
+            ]
+
+        async def close(self) -> None:
+            """Dummy"""
+
+    connector = aiohttp.TCPConnector(resolver=FakeResolver(), ssl=False)
+
+    async with aiohttp.ClientSession(connector=connector) as client, client.get(
+        url_from, auth=aiohttp.BasicAuth("user", "pass")
+    ) as resp:
+        assert len(resp.history) == 1
+        assert str(resp.url) == "http://example.com"
+        assert resp.status == 200
+
+
 @pytest.fixture
 def create_server_for_url_and_handler(
     aiohttp_server: AiohttpServer, tls_certificate_authority: trustme.CA
--- a/docs/client_advanced.rst
+++ b/docs/client_advanced.rst
@@ -67,6 +67,13 @@ argument. An instance of :class:`BasicAuth` can be passed in like this::
     async with ClientSession(auth=auth) as session:
         ...
 
+Note that if the request is redirected and the redirect URL contains
+credentials, those credentials will supersede any previously set credentials.
+In other words, if ``http://user@example.com`` redirects to
+``http://other_user@example.com``, the second request will be authenticated
+as ``other_user``. Providing both the ``auth`` parameter and authentication in
+the *initial* URL will result in a :exc:`ValueError`.
+
 For other authentication flows, the ``Authorization`` header can be set
 directly::
 


aiohttp/client.py
```diff
@@ -555,13 +555,18 @@ async def _request(
                             else InvalidUrlClientError
                         )
                         raise err_exc_cls(url)
-                    if auth and auth_from_url:
+                    # If `auth` was passed for an already authenticated URL,
+                    # disallow only if this is the initial URL; this is to avoid issues
+                    # with sketchy redirects that are not the caller's responsibility
+                    if not history and (auth and auth_from_url):
```
Dreamsorcerer, line 566: Doesn't it make sense to just move this before the loop? Then save doing this check on redirects/retries completely.

aiohttp/client.py
```diff
@@ -555,13 +555,18 @@ async def _request(
...
                         raise ValueError(
                             "Cannot combine AUTH argument with "
                             "credentials encoded in URL"
                         )
 
-                    if auth is None:
+                    # Override the auth with the one from the URL only if we
+                    # have no auth, or if we got an auth from a redirect URL
+                    if auth is None or (history and auth_from_url is not None):
```
Dreamsorcerer, line 574: I wonder about this part. This can get us into some inconsistent situations, if we do specify auth and the URL doesn't, then we're going to send the auth, which could then be rejected by the new server. I feel this could potentially be a security leak too if we end up sending the credentials to another server.

So, I'm wondering if we should throw away the auth on redirects completely...

aiohttp/client.py
PLPeeters: In principle it could, but `auth_from_url` is fetched in the loop which is why the check is there as well, otherwise we would end up with code duplication.

aiohttp/client.py
PLPeeters: I do agree, but isn't that what the current implementation is already doing? This change merely makes sure that any authentication in a redirect URL is properly taken into account, which cannot leak the input auth. It could leak the auth from a redirect further down the redirect chain however, but if I'm understanding this right that is a side-effect of the current auth handling and not this proposed change directly.

bdraco: If the server is compromised in some way and redirecting the auth to another server to capture credentials this is already a problem, and probably a bit moot, since if the server is compromised they could likely already steal the credentials and there isn't much point in redirecting them somewhere else.

CHANGES/9436.bugfix.rst
```diff
@@ -0,0 +1 @@
+Ignore client auth clashes caused by redirects and make redirect URL auth take precedence over input auth -- by :user:`PLPeeters`.
```
bdraco:
```suggestion
Authentication provided by a redirect now takes precedence over provided ``auth`` when making requests with the client -- by :user:`PLPeeters`.
```

aiohttp/client.py
bdraco, line 574: mypy is going to need to have `history` typed now since its used in Py_IsTrue

Dreamsorcerer: I guess I was thinking more that the server has some kind of blanket redirect, and maybe the user put auth credentials in the wrong request or something. If you don't think it's a problem, then we'll keep it like this.

webknjaz: > If the server is compromised in some way and redirecting the auth to another server to capture credentials this is already a problem, and probably a bit moot, since if the server is compromised they could likely already steal the credentials and there isn't much point in redirecting them somewhere else.

Yes, however, it might not be compromised fully. It might have some open redirect vulnerability unfixed. And in this case, the attackers wouldn't be able to steal the creds directly, but only by exploiting said redirect..

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,113 @@ async def test_creds_in_auth_and_url() -> None:
...
+        async def close(self) -> None:
+            """Dummy"""
+
+    connector = aiohttp.TCPConnector(resolver=FakeResolver(), ssl=False)
+
+    async with aiohttp.ClientSession(connector=connector) as client:
+        resp = await client.get(url_from, auth=aiohttp.BasicAuth("user", "pass"))
+        assert len(resp.history) == 1
+        assert str(resp.url) == "http://example.com"
+        assert resp.status == 200
```
webknjaz: Would this work outside the CM?
```suggestion

    assert len(resp.history) == 1
    assert str(resp.url) == "http://example.com"
    assert resp.status == 200
```

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,113 @@ async def test_creds_in_auth_and_url() -> None:
...
+            """Dummy"""
+
+    connector = aiohttp.TCPConnector(resolver=FakeResolver(), ssl=False)
+
+    async with aiohttp.ClientSession(connector=connector) as client:
+        resp = await client.get(url_from, auth=aiohttp.BasicAuth("user", "pass"))
+        assert len(resp.history) == 1
+        assert str(resp.url) == "http://example.com"
+        assert resp.request_info.headers.get("authorization") == "Basic dXNlcjo="
+        assert resp.status == 200
```
webknjaz: (same here)
```suggestion

    assert len(resp.history) == 1
    assert str(resp.url) == "http://example.com"
    assert resp.request_info.headers.get("authorization") == "Basic dXNlcjo="
    assert resp.status == 200
```

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,113 @@ async def test_creds_in_auth_and_url() -> None:
...
+                }
+            ]
+
+        async def close(self) -> None:
+            """Dummy"""
+
+    connector = aiohttp.TCPConnector(resolver=FakeResolver(), ssl=False)
+
+    async with aiohttp.ClientSession(connector=connector) as client:
+        resp = await client.get(url_from, auth=aiohttp.BasicAuth("user", "pass"))
```
webknjaz: Looking into this, I'd say that it should also be done through a CM or release the response object explicitly…

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,115 @@ async def test_creds_in_auth_and_url() -> None:
...
+
+    assert len(resp.history) == 1
+    assert str(resp.url) == "http://example.com"
+    assert resp.status == 200
+
+
+async def test_creds_redirect_url_creds_precedence(
```
webknjaz: Seeing that these two functions are almost identical, may I suggest using `@pytest.mark.parametrize` instead of keeping almost duplicates in the code?

tests/test_client_functional.py
PLPeeters: Totally fair! Note that a lot of tests are currently doing this, so it might be a good thing to do a test fix round at some point.

tests/test_client_functional.py
PLPeeters: Ah, that's my non-familiarity with PyTest showing up. I'll look into it!

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,73 @@ async def test_creds_in_auth_and_url() -> None:
...
+    url_to = URL("http://user@example.com")
+    redirected = False
+
+    async def srv(request: web.Request) -> web.Response:
+        nonlocal redirected
+
+        assert request.host == url_from.host
+
+        if not redirected:
+            redirected = True
```
PLPeeters, line 2944: @webknjaz Changed this as well when updating the test so the precedence test is able to fail separately if the precedence fix is undone, otherwise both tests just end up raising TooManyRequests due to the artificial infinite redirect, which is suboptimal.

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,73 @@ async def test_creds_in_auth_and_url() -> None:
         await session.close()
 
 
+@pytest.mark.parametrize(
+    ["check_redirect_auth_precedence"],
+    ([False], [True]),
+    ids=("redirect auth clash allowed", "redirect auth precedence"),
```
webknjaz: For a single param, you don't need to nest names or values. Additionally, it's best to use kebab-case in test IDs since these are often copied and pasted in terminal to run just one of the tests. And it's better to make test IDs short as they augment the test name, not replace it. It's not very clear what the variants are, I've taken a stub at trying to come up with new IDs:
```suggestion
    "check_redirect_auth_precedence",
    (False, True),
    ids=("clash-allowed", "precedence-enforced"),
```

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,73 @@ async def test_creds_in_auth_and_url() -> None:
...
+
+    async with aiohttp.ClientSession(connector=connector) as client, client.get(
+        url_from, auth=aiohttp.BasicAuth("user", "pass")
+    ) as resp:
+        assert len(resp.history) == 1
+        assert str(resp.url) == "http://example.com"
+        assert resp.status == 200
+
+        if check_redirect_auth_precedence:
+            assert resp.request_info.headers.get("authorization") == "Basic dXNlcjo="
```
webknjaz: This should probably have an `else` branch with the opposite assertion.

tests/test_client_functional.py
```diff
@@ -2927,6 +2927,73 @@ async def test_creds_in_auth_and_url() -> None:
...
 
 
+@pytest.mark.parametrize(
+    ["check_redirect_auth_precedence"],
+    ([False], [True]),
+    ids=("redirect auth clash allowed", "redirect auth precedence"),
+)
+async def test_creds_in_auth_and_redirect_url(
+    create_server_for_url_and_handler: Callable[[URL, Handler], Awaitable[TestServer]],
+    check_redirect_auth_precedence: bool,
```
webknjaz: I see that this argument is only used for optional assertion, so I assume what the behavior of both test variants is exactly the same. I'd expect some configuration difference between them..

PLPeeters: The (maybe wrong) rationale behind it is that you can break this test in 2 ways: you can either break the part that allows the clash on redirect, which will break both variants, or break the part that overrides the auth, which will only break the second variant. I figured this would make it easier to pinpoint the issue if someone breaks this somewhere down the line. Is there a better way to achieve that?

webknjaz: The tests should usually orchestrate for things they rely on to happen so that they check them. Are you saying it's somehow random?

PLPeeters: Not at all. I'm saying that if someone breaks this behaviour

https://github.com/aio-libs/aiohttp/blob/549279a36a815468a828a98c3375c1f7b2949285/aiohttp/client.py#L560-L563

the clash-allowed and precedence-enforced test variants will both fail. And if someone breaks this behaviour

https://github.com/aio-libs/aiohttp/blob/549279a36a815468a828a98c3375c1f7b2949285/aiohttp/client.py#L569-L571

only the precedence-enforced variant will fail. Nothing random about it :)

We _could_ test for both in one go and remove the parametrization altogether, but splitting the test allows for easier debugging if someone breaks one or both of these behaviours down the line. But again, maybe there's a better way to do this that I'm missing?

webknjaz: Ah, I see. I'm not sure if there's a better way, but such unobvious cases always require a code comment explaining these decisions. Otherwise, someone may just refactor this test to drop the parametrization and the if-block because it looks silly :man_shrugging:

PLPeeters: Very good point, I'll add that in tomorrow with rest of the suggested fixes!

tests/test_client_functional.py
PLPeeters: I suppose this `else` branch doesn't make sense anymore in light of our discussion below?

webknjaz: I'm not entirely sure. I'd like to hear what @Dreamsorcerer and @bdraco have to say.

tests/test_client_functional.py
PLPeeters: It appears you do need nesting even for a single param. If I don't nest I get

```
ERROR tests/test_client_functional.py - TypeError: object of type 'bool' has no len()
```

The rest is fixed as suggested :)

webknjaz: No, nesting is not necessary. You should've accepted the patch from the GH UI. I've suggested the correct change that updates several lines and your error implies that you only copied it partially. In general, it's recommended to commit code suggestions via GH and then pull them locally, if necessary.

PLPeeters: Ugh, my bad. Fixed.

tests/test_client_functional.py
Dreamsorcerer: Wouldn't this always be true? Seems to me like the parametrize is just deciding whether to check this or not. In which case, wouldn't it make more sense to just add a comment on this assert explaining the likely cause, rather than running the test twice with parametrize?

PLPeeters: That's a good point actually, I'll do that instead.

PLPeeters:
@webknjaz Is anything still required on my end to get this merged?

webknjaz (approved):
> Is anything still required on my end to get this merged?

All's good on my side. It'd be nice to have a docstring in the test function, but that's minor. I'll let somebody else merge, though.

bdraco:
That proxy test is flakey.  I restarted the CI.  I'm just about to walk out the door though so will check it when I get back home

bdraco:
If everything passes, I'll throw it on production and make sure there aren't any unexpected side effects as soon as I have some spare cycles

bdraco:
Testing this now

bdraco:
note for the future `docs/client_advanced.rst` has a conflict on backport.

bdraco (approved):
No unexpected side effects observed, however production has very limited use of auth, but enough that I feel comfortable it won't break most use cases.

bdraco:
Thanks @PLPeeters

bdraco:
https://github.com/aio-libs/aiohttp/pull/9570

bdraco:
https://github.com/aio-libs/aiohttp/pull/9571

PR #9448, opened 2024-10-10
Improve performance of fetching the content-length for web responses

bdraco:
## What do these changes do?

Small performance improvement for web responses ~1.8% speed up from avoiding all the multiple super calls

ref #2779 https://github.com/aio-libs/aiohttp/issues/2779#issuecomment-2400156629

no

no

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -747,9 +747,7 @@ class HeadersMixin:
     def content_length(self) -> Optional[int]:
         """The value of Content-Length HTTP header."""
         content_length = self._headers.get(hdrs.CONTENT_LENGTH)
-        if content_length is None:
-            return None
-        return int(content_length)
+        return None if content_length is None else int(content_length)
 
 
 def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -746,12 +746,10 @@ class HeadersMixin:
     @property
     def content_length(self) -> Optional[int]:
         """The value of Content-Length HTTP header."""
-        content_length = self._headers.get(hdrs.CONTENT_LENGTH)
-
+        content_length = self._headers[hdrs.CONTENT_LENGTH]
         if content_length is not None:
             return int(content_length)
-        else:
-            return None
+        return None
 
 
 def set_result(fut: "asyncio.Future[_T]", result: _T) -> None:
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -746,7 +746,7 @@ class HeadersMixin:
     @property
     def content_length(self) -> Optional[int]:
         """The value of Content-Length HTTP header."""
-        content_length = self._headers[hdrs.CONTENT_LENGTH]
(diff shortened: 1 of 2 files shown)


aiohttp/helpers.py
```diff
@@ -746,12 +746,10 @@ def charset(self) -> Optional[str]:
     @property
     def content_length(self) -> Optional[int]:
         """The value of Content-Length HTTP header."""
-        content_length = self._headers.get(hdrs.CONTENT_LENGTH)
-
-        if content_length is not None:
-            return int(content_length)
-        else:
+        content_length = self._headers[hdrs.CONTENT_LENGTH]
+        if content_length is None:
```
Dreamsorcerer: This won't be None if we don't use .get(), right?

bdraco: right in this case its typed to `str`.  Will adjust

bdraco: Little bit faster with that change as well 46931.13

bdraco: 1.2% speed up to 1.8% speed up

aiohttp/helpers.py
```diff
@@ -747,11 +747,9 @@ def charset(self) -> Optional[str]:
     def content_length(self) -> Optional[int]:
         """The value of Content-Length HTTP header."""
         content_length = self._headers.get(hdrs.CONTENT_LENGTH)
-
-        if content_length is not None:
-            return int(content_length)
-        else:
+        if content_length is None:
             return None
+        return int(content_length)
```
Dreamsorcerer: One-liner?
```suggestion
        return None if content_length is None else int(content_length)
```

aiohttp/helpers.py
bdraco: Also super interesting is that if we use an IP address its almost 7% faster

```
(venv) bdraco@MacBook-Pro-5 aiohttp % wrk -t 10 http://localhost:8080
Running 10s test @ http://localhost:8080
  10 threads and 10 connections
    Latency   210.76us   18.56us   1.30ms   82.25%
    Req/Sec     4.72k   106.61     5.42k    81.19%
  473971 requests in 10.10s, 71.87MB read
Requests/sec:  46931.13
Transfer/sec:      7.12MB
(venv) bdraco@MacBook-Pro-5 aiohttp % wrk -t 10 http://127.0.0.1:8080
Running 10s test @ http://127.0.0.1:8080
  10 threads and 10 connections
    Latency   195.56us   18.22us   1.04ms   82.81%
    Req/Sec     5.06k   101.73     5.40k    92.67%
  508676 requests in 10.10s, 77.13MB read
Requests/sec:  50365.54
Transfer/sec:      7.64MB
```

aiohttp/helpers.py
bdraco: as long as it doesn't wrap, sound good to me

aiohttp/helpers.py
bdraco: Something to investigate for a future PR to see if we have something going on with names vs ips

Dreamsorcerer: Not sure exactly what your test is running, but wouldn't that just be DNS resolution or something along those lines?

Issue #9447, opened 2024-10-10
DNS resolution errors are sometimes swallowed


bdraco:
### Describe the bug

There is definitely some weirdness going on in aiohttp or more likely in aiodns: When I drop packets to my DNS server, I get the following stack trace:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/aiohttp/resolver.py", line 105, in resolve
    resp = await self._resolver.getaddrinfo(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
aiodns.error.DNSError: (12, 'Timeout while contacting DNS servers')

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/aiohttp/connector.py", line 1294, in _create_direct_connection
    hosts = await self._resolve_host(host, port, traces=traces)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/aiohttp/connector.py", line 945, in _resolve_host
    return await asyncio.shield(resolved_host_task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/aiohttp/connector.py", line 982, in _resolve_host_with_throttle
    addrs = await self._resolver.resolve(host, port, family=self._family)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/aiohttp/resolver.py", line 114, in resolve
    raise OSError(msg) from exc
OSError: Timeout while contacting DNS servers

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/tmp/test3.py", line 16, in <module>
    asyncio.run(main())
  File "/usr/local/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #9451, opened 2024-10-10
Fix AsyncResolver swallowing the error message

bdraco:
## What do these changes do?

We would incorrectly create the `OSError` exception by passing the string as the first argument. The first argument of `OSError` is the errno. This meant that the message would end up in the wrong place.

https://docs.python.org/3/library/exceptions.html#OSError

proper errors

no

## Related issue number
#9447

### DIFF
--- a/CHANGES/9451.bugfix.rst
+++ b/CHANGES/9451.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed error messages from :py:class:`~aiohttp.resolver.AsyncResolver` being swallowed -- by :user:`bdraco`.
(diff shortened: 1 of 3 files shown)


bdraco:
much better
`aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host xcheckonline.home-assistant.io:443 ssl:default [DNS server returned answer with no data]`

instead of
`aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host xcheckonline.home-assistant.io:443 ssl:default [None]`

bdraco:
needs a test

bdraco:
We have fallback code for 3.11/3.10 that has the same problem.

bdraco:
I'll do another PR for 3.11 to fix the fallback code as well

bdraco:
Looks like this is worth doing a patch release for https://github.com/home-assistant/core/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+%22ClientConnectorError%22+%22%5BNone%5D%22

PR #9454, opened 2024-10-10
Simplify DNS throttle implementation

### LINKED ISSUE #3431: Timeout on one connection can cause CancelledError on other connections on DNS refresh
I ran into this on 2.3, but based on code it looks like this issue exists in current version

Conditions:
- DNS entry expires
- One task starts updating the DNS entry (and gets the lock)
- Another task sees the expired cached entry and waits for first one
- First task times out, causing async-timeout to cancel the task. This causes a CancelledError at this level. This is then passed on the lock and then raised to the second task

While the async-timeout for first task knows about the CancelledError and handles it (by throwing TimeoutError), the second task throws back CancelledError

DNS refresh code:

https://github.com/aio-libs/aiohttp/blob/56e39cd1e7cac90711805baaa89bb52842d0aaed/aiohttp/connector.py#L785-L816

Locks code:

https://github.com/aio-libs/aiohttp/blob/56e39cd1e7cac90711805baaa89bb52842d0aaed/aiohttp/locks.py#L29-L38

Example stack trace:

```
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/helpers.py", line 99, in __iter__
ret = yield from self._coro
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/client.py", line 267, in _request
conn = yield from self._connector.connect(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 402, in connect
proto = yield from self._create_connection(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 749, in _create_connection
_, proto = yield from self._create_direct_connection(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 813, in _create_direct_connection
hosts = yield from self._resolve_host(req.url.raw_host, req.port)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 718, in _resolve_host
yield from self._throttle_dns_events[key].wait()
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/locks.py", line 34, in wait
raise self._exc
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/client.py", line 267, in _request
conn = yield from self._connector.connect(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 402, in connect
proto = yield from self._create_connection(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 749, in _create_connection
_, proto = yield from self._create_direct_connection(req)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 813, in _create_direct_connection
hosts = yield from self._resolve_host(req.url.raw_host, req.port)
File "/Users/siddh/lib/python3.4/site-packages/aiohttp/connector.py", line 727, in _resolve_host
loop=self._loop)
File "/Users/siddh/python3.4/lib/python3.4/asyncio/futures.py", line 358, in __iter__
yield self # This tells Task to wait for completion./Users/siddh
File "/Users/siddh/python3.4/lib/python3.4/asyncio/tasks.py", line 297, in _wakeup
future.result()
File "/Users/siddh/python3.4/lib/python3.4/asyncio/futures.py", line 266, in result
raise CancelledError
```

bdraco:
## What do these changes do?

Simplify DNS throttle implementation

When testing #9447 I found there were so many race opportunities that it made sense to refactor the code to reduce the chance for races and make it quite a bit simpler.

This should also improve the performance a bit when we have a thundering herd backing up waiting for the resolver as they no longer need to create tasks and we only need a single task for the resolution that is happening to be able to shield it from cancellation.

no

no

likely fixes #3431


aiohttp/connector.py
```diff
@@ -820,16 +826,19 @@ def __init__(
...
         self._family = family
         self._local_addr_infos = aiohappyeyeballs.addr_to_addr_infos(local_addr)
         self._happy_eyeballs_delay = happy_eyeballs_delay
         self._interleave = interleave
         self._resolve_host_tasks: Set["asyncio.Task[List[ResolveResult]]"] = set()
 
     def _close_immediately(self) -> List[Awaitable[object]]:
-        for ev in self._throttle_dns_events.values():
-            ev.cancel()
+        for future_set in self._throttle_dns_futures.values():
```
Dreamsorcerer: Is it worth using itertools.chain() to flatten this?

bdraco: sounds good to me

f2f126a4a

aiohttp/connector.py
```diff
@@ -899,82 +907,93 @@ async def _resolve_host(
...
+    ) -> None:
+        futures: Set[asyncio.Future[None]]
+        #
+        # If multiple connectors are resolving the same host, we wait
+        # for the first one to resolve and then use the result for all of them.
+        # We use a throttle to ensure that we only resolve the host once
+        # and then use the result for all the waiters.
+        #
+        if key in self._throttle_dns_futures:
+            # get queue early, before any await (#4014)
```
```suggestion
            # get futures early, before any await (#4014)
```

aiohttp/connector.py
```diff
@@ -899,82 +907,93 @@ async def _resolve_host(
...
-                self._throttle_dns_events.pop(key)
 
-        return self._cached_hosts.next_addrs(key)
+            if traces:
+                for trace in traces:
+                    await trace.send_dns_resolvehost_end(host)
+
+            self._cached_hosts.add(key, addrs)  # type: ignore[possibly-undefined]
+            for fut in futures:
+                set_result(fut, None)
```
bdraco, line 989: We don't actually have any coverage for two (or more) requests in flight. This was hidden by the fact the previous event implementation had its own coverage.

bdraco: Considering how many different issues are in #9447 I think we should add a test for the following

- resolve while other resolve request in flight, in flight request is successful
- resolve while other resolve request in flight, in flight request is a failure
- resolve while other resolve request in flight, in flight request cancelled

bdraco:
The shielding is preventing the cancellation which prevents the timeout from working as well

aiohttp/connector.py
```diff
@@ -899,82 +907,93 @@ async def _resolve_host(
...
+        except BaseException as e:
+            # any DNS exception, independently of the implementation
+            # is set for the waiters to raise the same exception.
```
```suggestion
            # any DNS exception is set for the waiters to raise the same exception.
            # This coro is always run in task that is shielded from cancellation so
            # we should never be propagating cancellation here.
```

bdraco:
works fine in production.

Lot more coverage now as well

PR #9467, opened 2024-10-11
Small performance improvement to handling redirects

bdraco:
Reverse the conditional so the origin is only checked if not is_same_host_https_redirect

> Calling `.origin()` is a bit expensive https://github.com/aio-libs/aiohttp/issues/7583#issuecomment-1806828707


bdraco:
Looks like this is only in master anyways

PR #9473, opened 2024-10-11
Small speed up to checking connection in ClientResponse

bdraco:
## What do these changes do?

The check should have been `is not None` here to avoid the` __bool__` call which is always `True`

no

no


bdraco:
I guess its there for subclassed versions.

PR #9485, opened 2024-10-15
Avoid passing client writer to response when already finished

bdraco:
## What do these changes do?

With Python 3.12+ its likely that the writer can finish synchronously (on a production HA instance, it always did for all samples in a 60s period). In this case avoid passing it to the response since it will get unset right away needlessly

no

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -291,8 +291,8 @@ class ClientRequest:
         if writer is None:
             return
         if writer.done():
-            # The writer is already done, so we can reset it immediately.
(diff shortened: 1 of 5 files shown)


aiohttp/client_reqrep.py
```diff
@@ -709,7 +710,11 @@ async def send(self, conn: "Connection") -> "ClientResponse":
         else:
             task = self.loop.create_task(coro)
 
-        self._writer = task
+        if task.done():
+            task = None
+        else:
+            self.__writer = task
```
```suggestion
            self._writer = task
```

PR #9486, opened 2024-10-15
Add __slots__ to StreamWriter

bdraco:
## What do these changes do?

Add `__slots__` to StreamWriter

We use `__slots__` almost everywhere else in the codebase, however `__slots__` were not implemented in StreamWriter

no


bdraco:
This is too messy

PR #9503, opened 2024-10-18
Adjust URL type checks to checks to explictly check for URL

bdraco:
## What do these changes do?

We do not need to check for subclassed URL as subclassing is forbidden in yarl:
https://github.com/aio-libs/yarl/blob/03e634eaa7837209dc60065ece85bb3e54b8770b/yarl/_url.py#L438

The `type` check is a bit faster than an `isinstance` check

no


bdraco:
Its not called frequently enough that it makes any measurable difference on the profile

Issue #9506, opened 2024-10-18
WebSocketResponse.close() should wait for a close frame in its timeout window


lenard-mosys:
### Describe the bug

After calling `WebSocketResponse.close()` the server waits for a single message from the client, and if it's not a close message, then it indicates an abnormal closure:

https://github.com/aio-libs/aiohttp/blob/13dc0200d855bdfec1d4c4f3e1c9a1c66f88f1eb/aiohttp/web_ws.py#L489-L506

The problem with this is that a non-close client message might be in flight when the server sends its close message, and this results in an abnormal closure at the server's side, even if the client responds to the close message immediately. Here is the situation I'm describing:

![aiohttp_ws_close_issue_opt](https://github.com/user-attachments/assets/2e367a28-0761-48fe-a32c-95937fa763f1)

To my best understanding this is reported as an abnormal closure on the server side based on the implementation of `WebSocketResponse.close()`.

### To Reproduce

I didn't write a reproducer yet.

### Expected behavior

During the timeout wait for a close message, and discard messages that arrive before that. Only report an abnormal closure if that close message never arrives.

### Logs/tracebacks

```python-traceback
NA
```

### Python Version

```console
all
```

### aiohttp Version

```console
master, as of writing 13dc0200d855bdfec1d4c4f3e1c9a1c66f88f1eb
```

### multidict Version

```console
NA
```

### propcache Version

```console
NA
```

### yarl Version

```console
NA
```

### OS

NA

### Related component

Server

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
Yeah' that doesn't sound right. Feel free to make a PR, or maybe @bdraco might be interested in this one.

AchilleasTsiamoulis:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

bdraco:
fixed via https://github.com/aio-libs/aiohttp/pull/9508

PR #9508, opened 2024-10-18
Discard non-close messages in close timeout window

lenard-mosys:
## What do these changes do?

Waits for a close message in the close timeout window, not just the first message.

Previously spurious abnormally closed websocket connections could be reported to the server, even if the client responded to the close message with a close in a timely manner.

No

## Related issue number

https://github.com/aio-libs/aiohttp/issues/9506

## Checklist

- [ ] ~~Documentation reflects the changes~~
  - It is a bugfix.


lenard-mosys:
Breaks this test:

https://github.com/aio-libs/aiohttp/blob/9e975d275ec6421a0d7d1649dacfb0ebf6ff72bf/tests/test_web_websocket_functional.py#L1235-L1254

IMO the test is broken, or at least names and comments are misleading. I'm a bit perplexed why this test even passed before. Note that the test tests for the close code of the `ClientWebSocketResponse`, not for the server's `WebSocketResponse`. The close code for `ws` within the `handler` should indeed be abnormal closure, but that's never tested.

I'm not entirely sure what this test intends to test, or even if it intends to test server or client side behavior.

bdraco:
> I'm not entirely sure what this test intends to test, or even if it intends to test server or client side behavior.

The intent of the test is to verify the client gets an ABNORMAL_CLOSURE when sever initiated close while it was trying to send a `str`

lenard-mosys:
> The intent of the test is to verify the client gets an ABNORMAL_CLOSURE when sever initiated close while it was trying to send a `str`

I think this is questionable. The server closed the connection cleanly, even if the client tried to send a message. There could be connection schemes where the client sends fire-and-forget status updates, and the server could close the connection whenever. The client should get the close code sent from the server, not abnormal closure in this case.

Having said that I struggle to find normative wording in the RFC that would describe the conforming behavior in this situation. The best I can find is (under "Normal Closure of Connections"):

https://datatracker.ietf.org/doc/html/rfc6455#section-7.3

> Servers MAY close the WebSocket connection whenever desired.

And for 1006:

>It is designated for use in applications expecting a status code to indicate that the connection was closed abnormally, e.g., without sending or receiving a Close control frame.

The "e.g." doesn't help here, suggesting that there are other situations where this is applicable.

Anyway, I think there are some problems with the autoclose mechanism of the client, and I can make a similar test fail to have the expected result, even when the client doesn't try to send a message that the server doesn't expect.

bdraco:
Your changes seems reasonable to me. We likely need to adjust the test.  I agree the RFC is not crystal clear.

lenard-mosys:
> Your changes seems reasonable to me. We likely need to adjust the test. I agree the RFC is not crystal clear.

In any case, there is probably a missing test that checks for the server side behavior. This particular test can be made to pass with `ws = web.WebSocketResponse(timeout=0.)`, which restores (roughly) the server-side behavior the client can test against, but I don't know how palatable that is.

bdraco:
> > Your changes seems reasonable to me. We likely need to adjust the test. I agree the RFC is not crystal clear.
> 
> In any case, there is probably a missing test that checks for the server side behavior. This particular test can be made to pass with `ws = web.WebSocketResponse(timeout=0.)`, which restores (roughly) the server-side behavior the client can test against, but I don't know how palatable that is.

I think thats fine, but please add a comment to explain the 0 timeout.

bdraco:
@lenard-mosys It looks like a conflict has appeared

CHANGES/9506.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed ``WebSocketResponse.close()`` to discard non-close messages within its timeout window after sending close.
+-- by :user:`lenard-mosys`
```
bdraco:
```suggestion
Fixed ``WebSocketResponse.close()`` to discard non-close messages within its timeout window after sending close -- by :user:`lenard-mosys`.
```

aiohttp/web_ws.py
```diff
@@ -495,7 +495,11 @@ async def close(
 
         try:
             async with async_timeout.timeout(self._timeout):
-                msg = await reader.read()
+                while True:
+                    msg = await reader.read()
+                    if msg.type is WSMsgType.CLOSE:
```
bdraco, line 501: We might get WSMsgType.ERROR as well

tests/test_web_websocket_functional.py
```diff
@@ -1257,3 +1261,30 @@ async def handler(request: web.Request) -> web.WebSocketResponse:
...
+    close_code: Optional[WSCloseCode] = None
+
+    async def handler(request: web.Request) -> web.WebSocketResponse:
+        # Setting a short close timeout
+        ws = web.WebSocketResponse(timeout=0.1)
+        await ws.prepare(request)
+        await ws.close()
+
+        nonlocal close_code
+        if ws.close_code is not None:
```
bdraco: To fix the missing branch coverage, change this to an assert instead

aiohttp/web_ws.py
lenard-mosys: I reraise `msg.data` from here, so it's handled on the exception path.

bdraco: I think that should be fine. Will need a test for this case as well

lenard-mosys: Interestingly the `WSMsgType.ERROR` case seems to be already covered. The missing branch coverage is (probably) actually for the situation that this PR is meant to solve, which is when a regular message is received by the server within the close window. I added this test now.

lenard-mosys:
> @lenard-mosys We are likely doing another release soon. Not sure how keen you are to get this in the next one, but just in case you missed [my comment above](https://github.com/aio-libs/aiohttp/pull/9508#discussion_r1824682943) about the missing branch coverage for https://github.com/aio-libs/aiohttp/pull/9508/files#diff-37008174f47abcf385c9bd15d713b586a921dc6aad7d75acd1510aebc75d2b4fR503

I'm not in a hurry with this. I added a test that hopefully covers the previously missing branch.

bdraco:
Thanks. I'll take a look as soon as I can.

CI is a bit busy now as I'm cleaning up to prep another beta release

bdraco:
Looks like that branch is still coming up on the coverage report

Screenshot 2024-11-05 at 11 13 16 AM

bdraco:
https://app.codecov.io/gh/aio-libs/aiohttp/pull/9508/blob/aiohttp/web_ws.py

lenard-mosys:
> Looks like that branch is still coming up on the coverage report

Weird. I don't understand how the next line is covered then. My added test should cover the hidden `else` case here. I guess I could just add an `else: pass` and see what happens. Anyway, I don't have more time today for this, I will follow up tomorrow.

lenard-mosys:
I have some trouble covering the missing branch (that is producing a `WsMsgType.ERROR` message from `await reader.read()` within `WebSocketResponse.close()`). I tried sending an oversized message from the client side, with setting `WebSocketResponse(max_msg_size=8)` on the server size, but that produces an exception instead from `await reader.read()`.

It looks like that `WsMsgType.ERROR` is produced by exceptions, but those exceptions get reraised anyway from `await reader.read()`, therefore it's impossible to get this type of message at this level.

bdraco:
It looks like the only place we could get one is from
```
    def _handle_ping_pong_exception(self, exc: BaseException) -> None:
        """Handle exceptions raised during ping/pong processing."""
        if self._closed:
            return
        self._set_closed()
        self._set_code_close_transport(WSCloseCode.ABNORMAL_CLOSURE)
        self._exception = exc
        if self._waiting and not self._closing and self._reader is not None:
            self._reader.feed_data(WSMessageError(data=exc, extra=None))
```

bdraco:
For that to happen, close would have to be called right after `_handle_ping_pong_exception` happened

bdraco:
So I agree its truly unreachable.

aiohttp/web_ws.py
```diff
@@ -496,7 +496,13 @@ async def close(
...
+                        self._set_code_close_transport(msg.data)
+                        return True
+                    if msg.type is WSMsgType.ERROR:
+                        raise msg.data
```
```suggestion
```

Sorry for sending you down a wild goose chase on this one.

bdraco (approved):

CHANGES/9506.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ``WebSocketResponse.close()`` to discard non-close messages within its timeout window after sending close -- by :user:`lenard-mosys`.
```
bdraco:
```suggestion
Fixed :py:meth:`WebSocketResponse.close() <aiohttp.web.WebSocketResponse.close>` to discard non-close messages within its timeout window after sending close -- by :user:`lenard-mosys`.
```

bdraco (approved):

bdraco:
3.10 https://github.com/aio-libs/aiohttp/pull/9687

bdraco:
3.11 https://github.com/aio-libs/aiohttp/pull/9688

PR #9509, opened 2024-10-19
Avoid linear search to resolve resources

bdraco:
## What do these changes do?

Avoid linear search to resolve resources

`Resource` had another linear search of all methods to find the method for a resource.

no

no
## Related issue number
#2779

before 
after Requests/sec:  48394.40


bdraco:
Improvement is within the margin of error.

All the yarl improvements help so much more

PR #9511, opened 2024-10-20
Fix exceptions in websocket receive_* methods

### LINKED ISSUE #6800: EofStream and websocket.receive_* functions
aiohttp 3.8.1

`await websocket.receive_json()` Will die with `TypeError` on `WSMsgType.CLOSED`.

#2784 is fixed, yes:

```python
 async def __anext__(self) -> WSMessage:
        msg = await self.receive()
        if msg.type in (WSMsgType.CLOSE, WSMsgType.CLOSING, WSMsgType.CLOSED):
            raise StopAsyncIteration
        return msg
```

Yes, bug is fixed in iterator but NOT fixed for `websocket.receive_*` functions, like `receive_json()`

I would raise RuntimeError or so if corresponding message "type" (like bytes, str, or json) can not be received.

ara-25:
## What do these changes do?
- Add new `WSMessageTypeError` exception
- Raise `WSMessageTypeError` exception in `receive_*` methods in `WebSocketResponse` and `ClientWebSocketResponse`, when message without any content is received.

Websocket ``receive_str`` and ``receive_bytes`` methods will now raise ``WSMessageTypeError``, instead of ``TypeError``, when websocket messages without content are received.

No
## Related issue number

Fixes #6800 
## Checklist


aiohttp/client_ws.py
```diff
@@ -377,12 +377,20 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
 
     async def receive_str(self, *, timeout: Optional[float] = None) -> str:
         msg = await self.receive(timeout)
+        if msg.type not in MESSAGE_TYPES_WITH_CONTENT:
```
Dreamsorcerer: I don't think the issue is related to content, only a TEXT message will work here. The issue was just asking for the TypeError to be updated in the below check.

Dreamsorcerer:
But, should we use a connection error instead of RuntimeError as suggested by someone in that issue? That would make it easier for someone to catch the exception, as TypeError/RuntimeError is not really obvious. @bdraco Any opinions on the exception?

bdraco:
Changing it from `TypeError` to `RuntimeError` is a breaking change.  I think it would be safer to make a new exception that is derived from `TypeError`

aiohttp/client_ws.py
```diff
@@ -377,12 +377,20 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
+            raise RuntimeError(
+                f"Received message {msg.type}:{msg.data!r} has no content"
+            )
         if msg.type is not WSMsgType.TEXT:
             raise TypeError(f"Received message {msg.type}:{msg.data!r} is not str")
         return cast(str, msg.data)
 
     async def receive_bytes(self, *, timeout: Optional[float] = None) -> bytes:
         msg = await self.receive(timeout)
+        if msg.type not in MESSAGE_TYPES_WITH_CONTENT:
```
bdraco: This check can be moved inside the `if msg.type is not WSMsgType.BINARY:` block so we only check it if we know we are going to raise. Same with the others

aiohttp/client_ws.py
ara-25: in the original issue, according to my understanding, it was observed that these `receive_*` methods were raising `TypeError` when `EofStream` exception was caught in the `receive` method:
https://github.com/aio-libs/aiohttp/blob/d74cb5cd1221759c80cb84141ac042b7eb765a7f/aiohttp/client_ws.py#L342-L346

I understood that in the issue, it was proposed that some exception other than `TypeError` should be raised for these cases, as these seem to be different from type conversions. 
Please correct me if you find anything wrong with my understanding.

ara-25: Consequentially, I think that checking for `MESSAGE_TYPES_WITH_CONTENT` provides a more generic implementation for catching such cases.

ara-25:
> Changing it from `TypeError` to `RuntimeError` is a breaking change. I think it would be safer to make a new exception that is derived from `TypeError`

Should the new exception be included in the `client_exceptions.py` module

bdraco:
> > Changing it from `TypeError` to `RuntimeError` is a breaking change. I think it would be safer to make a new exception that is derived from `TypeError`
> 
> Should the new exception be included in the `client_exceptions.py` module?

I think one in `web_exceptions` for `web_ws`, and one in `client_exceptions` for `client_ws`

aiohttp/client_ws.py
Dreamsorcerer: Hmm, I'm not sure anymore. I was focusing more on:

> I would raise RuntimeError or so if corresponding message "type" (like bytes, str, or json) can not be received.

Which suggests changing the TypeError being thrown below.

And:

> The webscocket.receeive_* functions should throw ConnectionResetError when there are connection related errors instead of TypeError.

Which does make more sense to me than a RuntimeError, as people are likely already handling connection errors.

bdraco is more familiar with websockets, so I'll leave with him to decide what should be done.

ara-25:
@bdraco updated with the suggested changes. 

I have added the exception in just `client_exceptions.py` as `web_exceptions.py` seemed to contain http related exceptions.

ara-25:
@bdraco updated. Please have a look

tests/test_client_ws_functional.py
```diff
@@ -63,6 +70,24 @@ async def handler(request: web.Request) -> NoReturn:
...
+
+        await ws.close()
+        assert False
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+    client = await aiohttp_client(app)
+    resp = await client.ws_connect("/")
+
+    with pytest.raises(WSMessageTypeError):
```
bdraco: Please use match= with `pytest.raises`

ara-25: Updated

ara-25:
@bdraco I have completed tests for these changes

ara-25:
There might still be 2 uncovered lines showing up in the report, but they are not a part of the changes done in this PR and just include formatting changes. 

Let me know if you need me to add tests for those as well. Thanks

aiohttp/web_ws.py
```diff
@@ -605,6 +606,10 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
     async def receive_str(self, *, timeout: Optional[float] = None) -> str:
         msg = await self.receive(timeout)
         if msg.type is not WSMsgType.TEXT:
+            if msg.type not in [WSMsgType.CONTINUATION, WSMsgType.BINARY]:
```
bdraco: This case is missing branch coverage

aiohttp/web_ws.py
```diff
@@ -615,6 +620,10 @@ async def receive_str(self, *, timeout: Optional[float] = None) -> str:
     async def receive_bytes(self, *, timeout: Optional[float] = None) -> bytes:
         msg = await self.receive(timeout)
         if msg.type is not WSMsgType.BINARY:
+            if msg.type not in [WSMsgType.CONTINUATION, WSMsgType.TEXT]:
```
bdraco: This case is missing branch coverage

aiohttp/web_ws.py
```diff
@@ -615,6 +620,10 @@ async def receive_str(self, *, timeout: Optional[float] = None) -> str:
...
+                raise WSMessageTypeError(
+                    f"Received message {msg.type}:{msg.data!r} has no content"
+                )
             raise TypeError(f"Received message {msg.type}:{msg.data!r} is not bytes")
```
bdraco: Maybe it makes more sense to not add the additional check and change the existing exception?

```suggestion
            raise WSMessageTypeError(f"Received message {msg.type}:{msg.data!r} is not bytes")
```

ara-25: Now that you have pointed this out, this does make more sense.

ara-25: Updated

aiohttp/web_ws.py
```diff
@@ -602,7 +603,7 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
     async def receive_str(self, *, timeout: Optional[float] = None) -> str:
         msg = await self.receive(timeout)
         if msg.type is not WSMsgType.TEXT:
-            raise TypeError(
+            raise WSMessageTypeError(
                 "Received message {}:{!r} is not WSMsgType.TEXT".format(
                     msg.type, msg.data
                 )
```
bdraco: Please update this one to be an `f-string`

CHANGES/6800.bugfix.rst
```diff
@@ -0,0 +1 @@
+Modified websocket ``receive_str`` and ``receive_bytes`` methods to raise new ``WSMessageTypeError`` exception, instead of generic ``TypeError``, when websocket messages of incorrect types are received -- by :user:`ara-25`.
```
bdraco:
```suggestion
Modified websocket ``receive_str`` and ``receive_bytes`` methods to raise new :py:exc:`~aiohttp.WSMessageTypeError``` exception, instead of generic ``TypeError``, when websocket messages of incorrect types are received -- by :user:`ara-25`.
```

``TypeError``, ``receive_str`` and ``receive_bytes`` could also get references

ara-25: Updated

CHANGES/6800.bugfix.rst
```diff
@@ -0,0 +1 @@
+Modified websocket :py:exc:`~aiohttp.client_ws.ClientWebSocketResponse.receive_str`, :py:exc:`~aiohttp.client_ws.ClientWebSocketResponse.receive_bytes`, :py:exc:`~aiohttp.web_ws.WebSocketResponse.receive_str` & :py:exc:`~aiohttp.web_ws.WebSocketResponse.receive_bytes` methods to raise new :py:exc:`~aiohttp.WSMessageTypeError` exception, instead of generic :py:exc:`TypeError`, when websocket messages of incorrect types are received -- by :user:`ara-25`.
```
bdraco: The ones that are methods and not exceptions should use meth instead of exc

ara-25: Apologies for this. Corrected.

bdraco:
The object paths can be a bit tricky.  Check CHANGES.rst for previous usage to figure them out 

https://github.com/aio-libs/aiohttp/actions/runs/11586225078/job/32256466974#step:13:1

bdraco:
Once it's all passing check the read the docs preview as it will show how it will appear in the changelog

ara-25:
Fixed the changelog refs and also made some changes to the docs.

I had to add documentation for the `WSMessageTypeError` in `docs/client_reference.rst` for its reference to correctly show up in changelog.

Additionally, can you please guide me about how where this exception should be added in the `Hierarchy of exceptions` section:
https://github.com/aio-libs/aiohttp/blob/3f4f3adbc8f315b04e29e3bf3b4bc0ad27218a4a/docs/client_reference.rst#L2346-L2355

docs/client_reference.rst
```diff
@@ -2239,6 +2239,12 @@ Response errors
 
    Derived from :exc:`ClientResponseError`
 
+.. exception:: WSMessageTypeError
+
+   Received Web socket message of unexpected type
```
bdraco:
```suggestion
   Received WebSocket message of unexpected type
```

bdraco:
Screenshot 2024-11-02 at 1 46 52 PM

bdraco:
> Additionally, can you please guide me about how where this exception should be added in the `Hierarchy of exceptions` section:

I will take a look as soon as I can. Ran out of time for today.

asvetlov (commented):
Please don't merge until tomorrow. I need more time to review the PR. Thanks

asvetlov (approved):

asvetlov:
@bdraco please feel free to merge if you have no objections.

bdraco:
Will run through it later today. Thanks

ara-25:
Also @bdraco, I was thinking if it would be a good idea to add the websocket message-type and optionally data information in the exception class so that the end users can use those if required:
```python
raise WSMessageTypeError(
                f"Received message {msg.type}:{msg.data!r} is not WSMsgType.TEXT", received_msg_type=msg.type, received_msg_data=data, 
    
```

bdraco:
I'd worry a bit about storing a strong reference the whole payload in the Exception as Exceptions tend to get logged and collected by various systems which may have some undesirable memory impact

aiohttp/client_exceptions.py
```diff
@@ -377,3 +378,7 @@ def __str__(self) -> str:
             "[{0.certificate_error.__class__.__name__}: "
             "{0.certificate_error.args}]".format(self)
         )
+
+
+class WSMessageTypeError(TypeError):
+    """websocket message type is not valid."""
```
bdraco:
```suggestion
    """WebSocket message type is not valid."""
```

bdraco (approved):

bdraco:
Let's not expand the scope of this PR. We can consider other changes in a followup

bdraco:
backport https://github.com/aio-libs/aiohttp/pull/9676

PR #9523, opened 2024-10-22
Test typing final round

Dreamsorcerer:


webknjaz:
@Dreamsorcerer is it possible to backport at least some of these? This would help Patchback get more things done automatically...

Dreamsorcerer:
There's already so much diversion, it'd take just as much work to do the backports (which is weeks I've already spent on getting this finished). I'm hoping we can start prioritising v4 soon and then be able to essentially reset on the patchbacks by dropping v3.

webknjaz:
Yeah, if we want that, could you start a project board for tracking things that would have to go into that release stream?

Dreamsorcerer:
I've been tagging them into the milestone: https://github.com/aio-libs/aiohttp/milestone/20

PR #9530, opened 2024-10-23
Allow URLs with paths that end with `/` as base_url in ClientSession

### LINKED ISSUE #6647: Allow URLs with paths as base_url in ClientSession

Many sites have an API path in the URL. For example, Jira uses {jira_url}/rest/api/2/ as the base for its API requests.
The current implementation of the ClientSession only allows absolute URLs without a path. Thus, to make a number of Jira requests, you would do something like:
```python
async with aiohttp.ClientSession(base_url="https://jira.com") as client:
    baz = await client.get("/rest/api/2/issue/ISSUE-1")
    bom = await client.get("/rest/api/2/project/ISSUE")
```
instead of the simpler 
```python
async with aiohttp.ClientSession(base_url="https://jira.com/rest/api/2") as client:
    baz = await client.get("/issue/ISSUE-1")
    bom = await client.get("/project/ISSUE")
```

This somewhat defeats the purpose of having the base_url in the first place.

Allow using paths in the base_url, similar to how Requests and HTTPX do.

Just continue typing in the path or using variables with f-strings.

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Cycloctane:
## What do these changes do?

- Allow URLs with paths as base_url in ClientSession.
- Add tests for `_build_url` to test url joining when base_url includes path.

`base_url` in `ClientSession` now accepts URLs without paths or with paths that end with `/`

## Related issue number

Fixes #6647 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


aiohttp/client.py
```diff
@@ -415,8 +415,8 @@ def _build_url(self, str_or_url: StrOrURL) -> URL:
         if self._base_url is None:
             return url
         else:
-            assert not url.absolute and url.path.startswith("/")
-            return self._base_url.join(url)
+            assert not url.absolute
+            return self._base_url.join(URL(url.path_qs.lstrip("/")))
```
Dreamsorcerer: Hmm, I guess this is another complication. The spec suggests that `/foo` should replace the base path, while `foo` would get merged onto the base path.

@bdraco How do you think it should work here?

aiohttp/client.py
```diff
@@ -300,9 +300,9 @@ def __init__(
         else:
             self._base_url = URL(base_url)
             self._base_url_origin = self._base_url.origin()
-            assert (
-                self._base_url_origin == self._base_url
-            ), "Only absolute URLs without path part are supported"
+            assert self._base_url.absolute, "Only absolute URLs are supported"
+        if self._base_url is not None:
+            assert self._base_url.path.endswith("/")
```
```suggestion
        if self._base_url is not None and not self._base_url.path.endswith("/"):
            raise ValueError("base_url must have a trailing '/'")
```

aiohttp/client.py
bdraco: Reconstructing the URL again here is going to have an impact on every request.   It seems whats wanted it a new function in yarl to `append` urls instead of `join` them.


Side note .. URL processing used to be ~15-25% of the request cycle in some cases, and we have done a lot to improve the performance there recently.

Dreamsorcerer: > It seems whats wanted it a new function in yarl to `append` urls instead of `join` them.

Well, my first question is whether that should be the behaviour? Or should we stick with the RFC behaviour of merging (i.e. the same behaviour in browsers)? In which case this change can be reverted.

bdraco: It looks like the docs don't say how the URL will be joined to the `base_url`
https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession

It gets messy quickly once we start modifying the URL with string processing. For example, the current implementation in this PR is dropping the fragment because `path_qs` does not return it.

I think we shouldn't alter the url path and joining should be fully governed by the semantics of `URL.join`

Dreamsorcerer: > It looks like the docs don't say how the URL will be joined to the `base_url` https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession

We explicitly disallowed paths previously, likely due to these concerns. So, this is the point that we decide how to do it.

> I think we shouldn't alter the url path and joining should be fully governed by the semantics of `URL.join`

Right, so docs will need to explain this, and the most common usecase will be paths without a leading slash then.

CHANGES/9530.feature.rst
```diff
@@ -0,0 +1,2 @@
+Updated :py:class:`~aiohttp.ClientSession` to make ``base_url`` parameter allow url with path.
+``base_url`` must have a path that end with a ``/``  -- by :user:`Cycloctane`.
```
Dreamsorcerer:
```suggestion
Updated :py:class:`~aiohttp.ClientSession` to support paths in ``base_url`` parameter.
``base_url`` paths must end with a ``/``  -- by :user:`Cycloctane`.
```

docs/client_reference.rst
```diff
@@ -62,9 +62,11 @@ The client session supports the context manager protocol for self closing.
...
 
    :param base_url: Base part of the URL (optional)
-      If set, it allows to skip the base part (https://docs.aiohttp.org) in
-      request calls. It must not include a path (as in
-      https://docs.aiohttp.org/en/stable).
+
+      If set, it allows to prepend a base part to urls in request calls. It
+      must either have a path with a trailing ``/`` (as in
+      https://docs.aiohttp.org/en/stable/) or have no path at all (as in
+      https://docs.aiohttp.org).
```
```suggestion
      If set, allows to join a base part to relative URLs in request calls.
      If the URL has a path it must have a trailing ``/`` (as in
      https://docs.aiohttp.org/en/stable/).

      Note that URL joining follows :rfc:`3986`. This means, in the most
      common case the request URLs should have no leading slash:

      e.g.::

         session = ClientSession(base_url="http://example.com/foo/")

         await session.request("GET", "bar")
         # request for http://example.com/foo/bar

         await session.request("GET", "/bar")
         # request for http://example.com/bar
```

docs/client_reference.rst
```diff
@@ -372,6 +374,22 @@ The client session supports the context manager protocol for self closing.
...
+                  e.g.::
+
+                     session = ClientSession(base_url="http://example.com/foo/")
+
+                     await session.request("GET", "bar")
+                     # request for http://example.com/foo/bar
+
+                     await session.request("GET", "/bar")
+                     # request for http://example.com/bar
+
```
Dreamsorcerer: I'd probably leave this in the base_url documentation, so it's all together.
```suggestion
```

Dreamsorcerer (approved):
Minor suggestions on docs. Otherwise, I think this is ready to merge.

bdraco:
Formatting of the client reference block isn't great, but its an existing problem so not something that needs to be fixed here

bdraco (approved):

Issue #9520, opened 2024-10-21
update_cookies in client_reqrep.py has the same performance issue that was solved for the jar itself in #9204


bdraco:
The cookies get recreated every time.  I need to dig into the cookie code some more to see if we can avoid this

https://github.com/aio-libs/aiohttp/blob/d74e51b5cf93eb99a1fb6aa66950a863c7c056ee/aiohttp/client_reqrep.py#L510

https://github.com/aio-libs/aiohttp/pull/9204

<img width="1349" alt="Screenshot 2024-10-21 at 1 44 47 PM" src="https://github.com/user-attachments/assets/d076cd4d-00c6-4a22-8e3e-6f92cd2a1ea8">

bdraco:
<img width="1233" alt="Screenshot 2024-10-17 at 9 55 31 PM" src="https://github.com/user-attachments/assets/ca7ffb41-ad96-494f-bca3-24451cae36b8">

AchilleasTsiamoulis:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

bdraco:
I think we have to wait for https://github.com/python/cpython/pull/126157

bdraco:
If we don't find another solution by the time Python 3.14 is released, we can close this one

bdraco:
I think its unlikely we are going to be able to improve the performance without moving away from stdlib http.cookies.  As the change to improve Morsel creation time is already merged upstream, I think there is nothing left to do but wait so I'm going to close this issue out.

PR #9534, opened 2024-10-24
Add codspeed benchmarks

bdraco:
## What do these changes do?

Adds codspeed benchmarks

This will get the ball rolling so we know when we have solved #9520

no

no


tests/test_client_request_benchmarks.py
```diff
@@ -0,0 +1,22 @@
...
+
+from pytest_codspeed import BenchmarkFixture  # type: ignore[import-untyped]
+from yarl import URL
+
+from aiohttp.client_reqrep import ClientRequest
+
+
+def test_update_cookies(
+    loop: asyncio.AbstractEventLoop, benchmark: BenchmarkFixture
+) -> None:
```
bdraco: > This benchmark contains 10 system calls, totalling 93.9 µs of execution time.

That was unexpected

tests/test_benchmarks_client_request.py
```diff
@@ -0,0 +1,22 @@
+"""codspeed benchmarks for client requests."""
+
+import asyncio
+from http.cookies import Morsel
+
+from pytest_codspeed import BenchmarkFixture  # type: ignore[import-untyped]
```
Dreamsorcerer, line 6: https://github.com/CodSpeedHQ/pytest-codspeed/pull/53

.mypy.ini
```diff
@@ -34,3 +34,13 @@ ignore_missing_imports = True
...
+
+[mypy-test_benchmarks_client_request]
+disable_error_code =
+  no-any-unimported,
+  misc
+
+[mypy-test_benchmarks_cookiejar]
+disable_error_code =
+  no-any-unimported,
+  misc
```
Dreamsorcerer, line 50: It's not overly clear why we're disabling these here. Is it a case of being too noisy with pytest-codspeed untyped? In which case we can remove this once they've exposed type annotations.

bdraco: > Is it a case of being too noisy with pytest-codspeed untyped?

That’s exactly the reason.

tests/test_benchmarks_client_request.py
bdraco: Thanks!

.mypy.ini
bdraco: comments at in 19e6bd302fa43f617456f124c3a067e57b3a7d15

PR #9538, opened 2024-10-25
Add WebSocket read and send benchmarks for the http writer

bdraco:
## What do these changes do?

Add WebSocket read and send benchmarks for the http writer

no

no

### DIFF
--- a/tests/test_benchmarks_http_websocket.py
+++ b/tests/test_benchmarks_http_websocket.py
@@ -5,6 +5,7 @@ import asyncio
 from pytest_codspeed import BenchmarkFixture  # type: ignore[import-untyped]
 
 from aiohttp import DataQueue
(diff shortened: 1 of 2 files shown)


tests/test_benchmarks_http_websocket.py
```diff
@@ -0,0 +1,63 @@
...
+
+        async def _drain_helper(self) -> None:
+            """Swallow drain."""
+
+    writer = WebSocketWriter(MockProtocol(), MockTransport())
+    raw_message = b"Hello, World!" * 100
+
+    async def _send_one_hundred_websocket_text_messages() -> None:
+        for _ in range(100):
+            await writer.send_frame(raw_message, WSMsgType.TEXT, compress=False)
```
```suggestion
            await writer.send_frame(raw_message, WSMsgType.TEXT)
```

Let's not benchmark kwarg overhead

PR #9541, opened 2024-10-25
DNM: Explore performance of adding a Cython WebSocket frame parser

bdraco:
DNM: this is a rough version to check performance

It works out to ~30% faster.... I'm not sure if its worth it or not to have to maintain the parse_frame function twice

### DIFF
--- a/aiohttp/_websocket.pyx
+++ b/aiohttp/_websocket.pyx
@@ -7,8 +7,16 @@ cdef extern from "Python.h":
 
 from libc.stdint cimport uint32_t, uint64_t, uintmax_t
 
+from .websocket_models import (
+    UNPACK_LEN2,
+    UNPACK_LEN3,
+    WebSocketError,
+    WSCloseCode,
+    WSHandshakeError,
+)
 
-def _websocket_mask_cython(object mask, object data):
+
+cpdef _websocket_mask_cython(object mask, object data):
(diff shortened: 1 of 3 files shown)


bdraco:
The code is nearly identical. I wonder if a pxd file would make more sense. Then we only have to maintain that and keep it in sync with the python code.

bdraco:
Maybe move `WebsocketReader` into `http_websocket_reader.py` then we have `http_websocket_reader.pxd` which has the cython defs and we don't have two copies of the code and cython will add the types from the pxd

bdraco:
It looks like that would work.  Would have to move around some code first as `http_websocket.py` would need to import `WebsocketReader` from `http_websocket_reader.py` and since `WebsocketReader` needs ``WSMsgType` and `WSMessage` that would be a circular imports.

So maybe move them all to `http_websocket_models.py` first

bdraco:
I like that design better because we end up with ~100 lines of cython typing to maintain and the python code is the same and we don't have a copy

bdraco:
pxd is probably the way to do since the time in `_feed_data` and `parse_frame` are both high

bdraco:
PXD only approach seems to be more than good enough and has far less maint burden https://github.com/aio-libs/aiohttp/pull/9543

PR #9543, opened 2024-10-26
Add a PXD only Cython extension for the WebSocket reader to improve performance 

bdraco:
The idea is to be able to use a [PXD](https://github.com/aio-libs/aiohttp/pull/9543/files#diff-6b61b7dcd609475376fe433a38623b775dbb45aac3c0031c02a9cbd2ed73203d) file to augment the python code with types so cython can build an[ optimized version +~100%](https://github.com/aio-libs/aiohttp/pull/9543#issuecomment-2439728269) of it as an extension without having to maintain a separate PYX file.

The reader has been moved to `reader_py.py` to ensure `AIOHTTP_NO_EXTENSIONS` does not load `reader_c`. There was some missing coverage in `reader.py`. Tests have been added to close the gap.

I considered an [alternate approach of adding a PYX version](https://github.com/aio-libs/aiohttp/pull/9541) which would be a tiny bit faster, but the downside of having to maintain the same code twice and requiring significant Cython knowledge to work on the WebSocket reader lead me to a PXD only approach since only Cython types have to be maintained which has a much lower learning curve.

The cythonized version can be examined with the following:

1. `make cythonize`
2. Run `cythonize -X language_level=3 -a -i aiohttp/_websocket/reader_c.py`
3. Examine `aiohttp/_websocket/reader_c.html`

Yellow lines in the html hint at python interaction. See https://cython.readthedocs.io/en/latest/src/tutorial/cython_tutorial.html https://pythonprogramming.net/introduction-and-basics-cython-tutorial/ 

Note that if you make changes to the `.pxd` file, `cythonize` may not see them until you `touch aiohttp/_websocket/reader_c.py` and run the above again.


aiohttp/_websocket/reader_py.py
```diff
@@ -0,0 +1,351 @@
...
+
+ALLOWED_CLOSE_CODES: Final[Set[int]] = {int(i) for i in WSCloseCode}
+
+READ_HEADER = 1
+READ_PAYLOAD_LENGTH = 2
+READ_PAYLOAD_MASK = 3
+READ_PAYLOAD = 4
+
+# WSMsgType values unpacked so they can by
+# cythonized to unsigned int
```
```suggestion
# WSMsgType values unpacked so they can by cythonized to ints
```

aiohttp/_websocket/reader_c.py
```diff
@@ -0,0 +1 @@
+reader_py.py
```
bdraco, line 1: I was a little worried about adding a symlink, but I did a build with `python -m build` and it includes a copy of it so it works just fine

aiohttp/_websocket/reader_py.py
```diff
@@ -0,0 +1,352 @@
...
+        self._compressed: Optional[bool] = None
+        self._decompressobj: Optional[ZLibDecompressor] = None
+        self._compress = compress
+
+    def feed_eof(self) -> None:
+        self.queue.feed_eof()
+
+    def feed_data(self, data: bytes) -> Tuple[bool, bytes]:
+        if self._exc is not None:
+            return True, data
```
bdraco, line 87: missing coverage is existing and this line is unchanged.

aiohttp/_websocket/reader.py
```diff
@@ -1,350 +1,21 @@
...
-        self._payload_length = 0
-        self._payload_length_flag = 0
-        self._compressed: Optional[bool] = None
-        self._decompressobj: Optional[ZLibDecompressor] = None
-        self._compress = compress
-
-    def feed_eof(self) -> None:
-        self.queue.feed_eof()
-
-    def feed_data(self, data: bytes) -> Tuple[bool, bytes]:
```
bdraco, line 69: This has been typed wrong for a very long time since Protector uses bytearray and not bytes.

Technically it could be a memoryview as well but that seems unlikely

aiohttp/_websocket/reader_c.py
bdraco: I went ahead and added coverage here since we needed to fix the incoming type for `feed_data`

aiohttp/_websocket/reader_py.py
```diff
@@ -0,0 +1,360 @@
...
+        self._compressed: Optional[bool] = None
+        self._decompressobj: Optional[ZLibDecompressor] = None
+        self._compress = compress
+
+    def feed_eof(self) -> None:
+        self.queue.feed_eof()
+
+    # data can be bytearray on Windows because proactor event loop uses bytearray
+    # and asyncio types this to Union[bytes, bytearray, memoryview] so we need
+    # verify that data is bytes
```
```suggestion
    # coerce data to bytes if it is not
```

bdraco:
We could probably get another 30% out of it if we stopped shipping around the data in a tuple since it has to be converted back to pyint.  But thats probably not worth it

PR #9546, opened 2024-10-27
DNM: Websocket pxd create debug windows

bdraco:
https://github.com/aio-libs/aiohttp/pull/9543 is stalling the CI on windows with python 3.10 and 3.11 only with no obvious reason


bdraco:
So the type is wrong elsewhere as well

PR #9556, opened 2024-10-28
Fix Cython `WebSocketReader` byte signedness

bdraco:
The `first_byte` and `second_byte` are `unsigned`. This would occasionally cause the length to be wrong for the new `WebSocketReader` added in #9543 and #9554

We didn't have coverage for a longer multi-byte payload. Add coverage for this case.

### DIFF
--- a/CHANGES/9556.feature.rst
+++ b/CHANGES/9556.feature.rst
@@ -1,0 +1,1 @@
+9543.feature.rst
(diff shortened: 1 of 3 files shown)


bdraco:
For some reason I forget about the signedness of `char` but never `int`... but thats why I do downstream production re-testing after merge... to catch problems like this one

PR #9585, opened 2024-10-29
Add client request creation benchmarks for headers and cookies

bdraco:
Add client request creation benchmarks for headers and cookies.

We should watch this over time to make sure we don't regress performance here since this tends to be a place that creeps over time.

### DIFF
--- a/tests/test_benchmarks_client_request.py
+++ b/tests/test_benchmarks_client_request.py
@@ -7,6 +7,7 @@ from pytest_codspeed import BenchmarkFixture  # type: ignore[import-untyped]
 from yarl import URL
 
 from aiohttp.client_reqrep import ClientRequest
+from aiohttp.http_writer import HttpVersion11
 
 
 def test_client_request_update_cookies(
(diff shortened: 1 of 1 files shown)


tests/test_benchmarks_client_request.py
```diff
@@ -20,3 +21,47 @@ def test_client_request_update_cookies(
...
+        ClientRequest(
+            method="get",
+            url=url,
+            loop=loop,
+            headers=None,
+            data=None,
+            cookies={"cookie": "value"},
+            auth=None,
+            version=HttpVersion11,
+            compress=None,
```
bdraco:
```suggestion
            compress=False,
```

tests/test_benchmarks_client_request.py
```diff
@@ -20,3 +21,47 @@ def test_client_request_update_cookies(
...
+        ClientRequest(
+            method="get",
+            url=url,
+            loop=loop,
+            headers={"header": "value", "another": "header"},
+            data=None,
+            cookies=None,
+            auth=None,
+            version=HttpVersion11,
+            compress=None,
```
bdraco:
```suggestion
            compress=False,
```

bdraco:
Pretty clear all the time is spent in cookies

bdraco:
Screenshot 2024-10-29 at 10 18 34 AM
Screenshot 2024-10-29 at 10 18 37 AM

PR #9594, opened 2024-10-29
DNM: Experiment to speed up creating cookie morsels

bdraco:
This likely needs to be optimized upstream but we can get an idea if its worth it here

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -493,4 +493,4 @@ class FastMorsel(Morsel):
         # Set defaults
         self._key = self._value = self._coded_value = None
         # Set default attributes
-        super().__init__(self._empty_reserved)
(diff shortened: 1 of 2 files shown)


bdraco:
It's a really simple optimization. The loop can be replaced with a dict update upstream.

aiohttp/cookiejar.py
```diff
@@ -482,3 +483,16 @@ def update_cookies(self, cookies: LooseCookies, response_url: URL = URL()) -> No
...
+
+
+_empty_reserved = dict.fromkeys(Morsel._reserved, "")
+
+
+def fast_init(self):
+    # Set defaults
+    self._key = self._value = self._coded_value = None
+    # Set default attributes
+    self |= self._empty_reserved
```
```suggestion
    self |= _empty_reserved
```

PR #9595, opened 2024-10-30
Upgrade codspeed to 3.0.0

bdraco:
SSIA


bdraco:
Can't backport to 3.10 since 3.10 is still supporting python 3.8

PR #9597, opened 2024-10-30
DNM: Avoid constructing traces during client request if there are none

bdraco:
Most of the downstream objects already made them optional

Maybe long term it would be nicer to put all the traces in a manager object and let it do all the dispatches so we can check to see if the object is None and do nothing

TODO: need to write a benchmark that would cover this path since it turned out to be a bit larger than I expected, and it may not be worth it.  EDIT: rough profile shows 23% less time spent in `def _request` in `client.py`, but we should still get a benchmark written to validate this is worth it.

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -1287,7 +1287,7 @@ class ClientSession:
     @property
     def trace_configs(self) -> List[TraceConfig[Any]]:
         """A list of TraceConfig instances used for client tracing"""
-        return self._trace_configs
+        return self._trace_configs or []
 
     def detach(self) -> None:
         """Detach connector from session without closing the former.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -364,9 +364,10 @@ class ClientSession:
         self._response_class = response_class
         self._ws_response_class = ws_response_class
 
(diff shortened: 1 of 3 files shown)


Dreamsorcerer:
Seems surprising that it'd make that much difference. I wonder if cpython shouldn't do a similar check automatically if it makes that much difference.

My test of if/for suggests it's twice as fast, but it's still only a few nanoseconds, so not sure how that'd make a big difference to the whole function:
```
> python3 -m timeit -s 'x = []' 'for y in x: pass'
10000000 loops, best of 5: 33.3 nsec per loop
> python3 -m timeit -s 'x = []' 'if x: pass'
20000000 loops, best of 5: 16.5 nsec per loop
```

bdraco:
I definitely want to write benchmarks to see if its worth it but `_request` is rather large so it will be a bit more difficult to write a benchmark for that does no I/O

bdraco:
Even if the result is that its not worth it, having the benchmark to be able to watch for performance regressions will be valuable on its own

bdraco:
rough benchmark change was jitter.

This doesn't help in a meaningful way...less than 1%

PR #9600, opened 2024-10-31
Avoid starting connection timeout when a connection is already available

### LINKED ISSUE #9598: Explore moving the connect ceil_timeout in the client inside the Connector connect
https://github.com/aio-libs/aiohttp/blob/7d0f980af6d725970cda5cb011400f5459699301/aiohttp/client.py#L637

If the connection is reused, we create a TimerHandler and than cancel is right away.... in testing this seems to be the case 80%+ of the time so this seem ripe to optimize away.  Since we pass on the timeout object to `connect` anyways we could move the timeout into `connect` so it only does the timeout if its actually going to wait.  Since we expect connection reuse to be common with HTTP/1.1+ this takes up a significant amount of the request time.

There is a small risk someone has subclassed and replaced the `connect` implementation with a custom connector, however it `connect` is quite complex so it seems unlikely someone would reimplement it.  Either way it would need a breaking change note in the changelog.

### LINKED ISSUE #9670: deadlock with TCPConnector limit after timeout
### Describe the bug

When using a limit in TCPConnector, timeouts can lead to a condition where new requests are not actually sent, resulting in "sticky" timeouts.

After debugging, I believe the problem occurs in this line:https://github.com/aio-libs/aiohttp/blob/v3.10.10/aiohttp/connector.py#L541

When a `ValueError` is raised there, that's the result of:
- Awaiting the future failed, potentially due to a `CancelledError`: a timeout
- `_release_waiter` method was called, and it tried to wake up a coroutine waiting for this connection. But since this coroutine also suffered the timeout, it won't proceed. Since it doesn't proceed, it won't wake up other potentially waiting coroutines.

If the number of waiters never get to zero, the cycle never ends, and no coroutine wakes up to proceed. All coroutines wake up with a `CancelledError`.

Proposed fix:
Replace the `pass` in the mentioned line with `self._release_waiter()`. This seems to resolve the issue by awakening another coroutine, which was the original intent of the previous `self._release_waiter()` call.

Another way to fix it might be to use `asyncio.Semaphore` with a context manager.

### To Reproduce

1. Start a small server script that returns slowly only sometimes. For example, run this with `fastapi dev aio_bug_server.py`:
```python
from fastapi import FastAPI
import asyncio
from time import time

app = FastAPI()

sleep_time = 0.3
timestamp_last_change = time()
last_recv = time()

@app.get("/getme")
async def getme():
    global sleep_time, last_recv
    if time()-last_recv>8:
        # Go "up", forcing 0 sleep time if we don't know anymore about clients
        sleep_time = 0
    elif time()-timestamp_last_change>2:
        if sleep_time:
            sleep_time = 0
        else:
            sleep_time = 0.3
    print(f"recv {sleep_time=:.3f}")
    last_recv = time()

    await asyncio.sleep(sleep_time)
    print(f"fin {sleep_time=:.3f}")
    return {
        "asdf": "asdf",
    }

```
2. Start a script that uses aiohttp to make concurrent requests to the server, and check the number of timeouts/ok results. For example:
```python


import uvloop
uvloop.install()

import asyncio
import aiohttp
from time import time

timeouts = 0
errors = 0
oks = 0

async def do_request(session: aiohttp.ClientSession):
    global errors, timeouts, oks
    try:
        async with session.get(f"http://127.0.0.1:8000/getme") as resp:
            await resp.json()
        oks += 1
    except TimeoutError:
        timeouts += 1
    except BaseException as exc:
        print(f"UNEXPECTED EXCEPTION {exc=}")
        errors += 1



async def main():
    global timeouts, oks
                    connector=aiohttp.TCPConnector(limit=10),
                    timeout=aiohttp.ClientTimeout(total=0.2), 
                ) as session:
        tasks = set()
        done_reqs = 0
        i = 0
        t = time()
        last_ok = t
        while True:
            i+=1
            if i%100==0:
                dt = (time() - t)
                print(f"{done_reqs/dt:.1f}req/s {oks=} {timeouts=}")
                if oks:
                    last_ok = time()
                if time() - last_ok > 10:
                    # We hit the issue likely
                    # at this point the server should not have received anything in a while
                    print(f"ISSUE DETECTED")
                    t = time()
                    try:
                        async with session.get(f"http://127.0.0.1:8000/getme?q=a") as resp:
                            await resp.json()
                    except Exception as exc:
                        print(f"LAST REQUEST ERROR: EXC{exc=} {time()-t}")
                    for task in tasks:
                        await task
                    print(f"NEW REQUEST AFTER AWAITING ALL TASKS")
                    t = time()
                    try:
                        async with session.get(f"http://127.0.0.1:8000/getme?q=b") as resp:
                            await resp.json()
                    except Exception:
                        raise
                    finally:
                        print("FINAL DT:", time()-t)
                    return

                timeouts = 0
                oks=0

            if len(tasks) >= 200:
                done, tasks = await asyncio.wait(tasks, return_when=asyncio.FIRST_COMPLETED)
                done_reqs += len(done)
            tasks.add(asyncio.create_task(do_request(session)))

asyncio.run(main())
```

### Expected behavior

All requests should be sent, previous timeouts shouldn't change that.

### Logs/tracebacks

```python-traceback
no tracebacks
```

### Python Version

```console
$ python --version
Python 3.12.7
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.10.10
Author: 
Author-email: 
License: Apache 2
Location: /home/dv/code/search/services/local/venv/lib/python3.12/site-packages
Required-by:
```

### multidict Version

```console
Name: multidict
Version: 6.1.0
Author: Andrew Svetlov
License: Apache 2
Location: /home/dv/code/search/services/local/venv/lib/python3.12/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### propcache Version

```console
Name: propcache
Version: 0.2.0
Author: Andrew Svetlov
License: Apache-2.0
Location: /home/dv/code/search/services/local/venv/lib/python3.12/site-packages
Requires: 
Required-by: yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.17.1
Author: Andrew Svetlov
License: Apache-2.0
Location: /home/dv/code/search/services/local/venv/lib/python3.12/site-packages
Required-by: aiohttp
```

### OS

Linux

### Related component

Client

### Additional context

_No response_

### Code of Conduct

bdraco:
## What do these changes do?

Since connection re-use is common, we can avoid starting the `ceil_timeout` when we have a connection immediately available. Previously when a connection is already available for reuse, `ceil_timeout` had to create a `TimerHandle`, schedule it on the event loop, and then immediately cancel it, leaving the loop to have to remove it from the heap. See https://github.com/aio-libs/aiohttp/pull/8608#discussion_r1704175784 and https://github.com/aio-libs/aiohttp/issues/8613 for more history on the overhead of this set of operations.

If the server supports keep-alive, the hit rate on avoiding the timeout is over 80% in testing. https://github.com/aio-libs/aiohttp/pull/9601/commits/31b9637410d5cbf50e4ed1868b8203f3894807b7

closes #9598

If `BaseConnector.connect` was replaced with a custom connect function, the timeout must now be handled in this function. [The only abstract method the docs specify to be overwritten is `_create_connection`](https://docs.aiohttp.org/en/v3.8.4/client_reference.html#baseconnector). While it seems extremely unlikely that someone would write out a custom `connect` function since it would likely break all the protected calls back into the connector from `Connection`, I think this change is likely safe to backport. ~~To be on the safe side I only tagged it for 3.11 (tested in https://github.com/aio-libs/aiohttp/pull/9601) in case someone is doing that.~~ This was originally planned not to backport to 3.10, but its needed to fix https://github.com/aio-libs/aiohttp/pull/9670
```
aiohttp/connector.py:            self._connector._release(self._key, self._protocol, should_close=True)
aiohttp/connector.py:            self._connector._release(self._key, self._protocol, should_close=True)
aiohttp/connector.py:            self._connector._release(
```

no

Benchmarks
See script in https://github.com/aio-libs/aiohttp/issues/9598

aiohttp 3.10.0, yarl 1.9.5
aiohttp_3_10_0_yarl_1_9_5

aiohttp 3.11.0b0 yarl 1.17.0
aiohttp_3_11_0b0_yarl_1_17_0

aiohttp 3.11.0b0+ this change yarl 1.17.0
aiohttp 3 11 0b0+change yarl 1 17 0

aiohttp 3.11.0b0+ this change yarl 1.17.1 
aiohttp 3 11 0b0+change yarl 1 17 1

### DIFF
--- a/CHANGES/9600.breaking.rst
+++ b/CHANGES/9600.breaking.rst
@@ -1,3 +1,3 @@
 Improved performance of the connector when a connection can be reused -- by :user:`bdraco`.
 
-If ``BaseConnector.connect`` has sub-classed and replaced with custom logic, the ``ceil_timeout`` must be added.
+If ``BaseConnector.connect`` has been subclassed and replaced with custom logic, the ``ceil_timeout`` must be added.
--- a/CHANGES/9600.breaking.rst
+++ b/CHANGES/9600.breaking.rst
@@ -1,0 +1,3 @@
+Improved performance of the connector when a connection can be reused -- by :user:`bdraco`.
+
+If ``BaseConnector.connect`` has sub-classed and replaced with custom logic, the ``ceil_timeout`` must be added.
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -94,7 +94,6 @@ from .helpers import (
     _SENTINEL,
     BasicAuth,
     TimeoutHandle,
-    ceil_timeout,
     get_env_proxy_for_url,
     method_must_be_empty_body,
     sentinel,
@@ -634,13 +633,9 @@ class ClientSession:
 
                     # connection timeout
                     try:
-                        async with ceil_timeout(
-                            real_timeout.connect,
-                            ceil_threshold=real_timeout.ceil_threshold,
-                        ):
-                            conn = await self._connector.connect(
-                                req, traces=traces, timeout=real_timeout
-                            )
+                        conn = await self._connector.connect(
+                            req, traces=traces, timeout=real_timeout
+                        )
                     except asyncio.TimeoutError as exc:
                         raise ConnectionTimeoutError(
                             f"Connection timeout to host {url}"
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -497,9 +497,7 @@ class BaseConnector:
         if not wait_for_conn and (proto := self._get(key)) is not None:
             # If we do not have to wait and we can get a connection from the pool
             # we can avoid the timeout ceil logic and directly return the connection
-            if traces:
-                await self._send_connect_reuseconn(key, traces)
-            return self._acquired_connection(proto, key)
+            return await self._reused_connection(key, proto, traces)
 
         async with ceil_timeout(
             timeout.connect,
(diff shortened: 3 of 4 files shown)


CHANGES/9600.breaking.rst
```diff
@@ -0,0 +1,3 @@
+Improved performance of the connector when a connection can be reused -- by :user:`bdraco`.
+
+If ``BaseConnector.connect`` has sub-classed and replaced with custom logic, the ``ceil_timeout`` must be added.
```
bdraco:
```suggestion
If ``BaseConnector.connect`` has been subclassed and replaced with custom logic, the ``ceil_timeout`` must be added.
```

bdraco:
Tagging for backport to 3.10 as well since we need it to fix https://github.com/aio-libs/aiohttp/pull/9670

PR #9616, opened 2024-11-01
Add `__slots__` to `Connection`

bdraco:
SSIA

### DIFF
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -87,7 +87,16 @@ if TYPE_CHECKING:
 
 
 class Connection:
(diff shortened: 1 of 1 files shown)


asvetlov:
I definitely support this but please don't expect too much from this micro optimization. Most likely benchmarks will display no measurable difference at all.

Memory consumption is another beast. But again, a casual code has hundreds, or maybe, thousands of Connection instances.

So, the win could be very hardly visible. 

I'm happy to be wrong though.

bdraco:
It will make very little difference for sure, other than saving a tiny bit of RAM

PR #9623, opened 2024-11-01
Fix available connection calculation

bdraco:
## What do these changes do?

The available connection calculation is only ever used to check if its above 0 (it could be a `bool` return) and the actual available number of connections was not always correct if the total available was less than the available per host. However, this was unlikely to have any actual impact on production since all the consumers of the function only cared if it was > 0 and that case was correct. As such this is not labeled as a bugfix and is an internal cleanup.

I refactored this logic to be a bit simplier in #9610 but it still was not correct.

Add some tests to make sure it works as expected as it was not directly tested.

no

no

### DIFF
--- a/tests/test_connector.py
+++ b/tests/test_connector.py
@@ -3398,7 +3398,7 @@ def test_default_ssl_context_creation_without_ssl() -> None:
         assert connector_module._make_ssl_context(True) is None
 
 
-async def test_available_connections(
+async def test_available_connections_with_limit_per_host(
     key: ConnectionKey, other_host_key2: ConnectionKey
 ) -> None:
     """Verify expected values based on active connections."""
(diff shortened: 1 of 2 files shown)


tests/test_connector.py
```diff
@@ -3390,3 +3396,77 @@ def test_default_ssl_context_creation_without_ssl() -> None:
...
+    assert conn._available_connections(key) == 3
+    assert conn._available_connections(other_host_key2) == 3
+
+
+async def test_available_connections_no_limits(
+    key: ConnectionKey, other_host_key2: ConnectionKey
+) -> None:
+    """Verify expected values based on active connections with no limits."""
+    # No limits is a special case where available connections should always be 1.
+    conn = aiohttp.BaseConnector(limit=None, limit_per_host=None)
```
bdraco:
```suggestion
    conn = aiohttp.BaseConnector(limit=0, limit_per_host=0)
```

tests/test_connector.py
```diff
@@ -3390,3 +3396,77 @@ def test_default_ssl_context_creation_without_ssl() -> None:
...
+    other_proto1 = create_mocked_conn()
+    other_connection1 = conn._acquired_connection(other_proto1, other_host_key2)
+    assert conn._available_connections(key) == 2
+    assert conn._available_connections(other_host_key2) == 1
+    other_connection1.close()
+    assert conn._available_connections(key) == 2
+    assert conn._available_connections(other_host_key2) == 2
+
+
+@pytest.mark.parametrize("limit_per_host", [None, 10])
```
bdraco:
```suggestion
@pytest.mark.parametrize("limit_per_host", [0, 10])
```

tests/test_connector.py
```diff
@@ -3390,3 +3396,77 @@ def test_default_ssl_context_creation_without_ssl() -> None:
...
+async def test_available_connections_without_limit_per_host(
+    key: ConnectionKey, other_host_key2: ConnectionKey, limit_per_host: Optional[int]
```
bdraco:
```suggestion
    key: ConnectionKey, other_host_key2: ConnectionKey, limit_per_host: int
```

PR #9624, opened 2024-11-02
[PR #9623/898aa28 backport][3.10] Fix available connection calculation

bdraco:
(cherry picked from commit 898aa280f20c740ae7e3c020546d5be925bb6ddc)


bdraco:
Need to adjust the test since it wasn't back ported all the way to 3.10.

PR #9628, opened 2024-11-02
Add benchmark for client WebSocket round trip

bdraco:
SSIA


bdraco:
Looks like this won't work with a coro

PR #9635, opened 2024-11-03
Add benchmark for sending large WebSocket messages

bdraco:
SSIA

For validating if https://github.com/aio-libs/aiohttp/pull/9634 is a good idea or not

### DIFF
--- a/tests/test_benchmarks_http_websocket.py
+++ b/tests/test_benchmarks_http_websocket.py
@@ -64,7 +64,9 @@ def test_send_one_hundred_websocket_text_messages(
         for _ in range(100):
             await writer.send_frame(raw_message, WSMsgType.TEXT)
 
-    benchmark(loop.run_until_complete(_send_one_hundred_websocket_text_messages()))
+    @benchmark
+    def _run() -> None:
+        loop.run_until_complete(_send_one_hundred_websocket_text_messages())
 
 
 def test_send_one_hundred_large_websocket_text_messages(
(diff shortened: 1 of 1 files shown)


tests/test_benchmarks_http_websocket.py
```diff
@@ -41,7 +43,7 @@ def is_closing(self) -> bool:
         """Swallow is_closing."""
         return False
 
-    def write(self, data: bytes) -> None:
+    def write(self, data: Iterable[Union[bytes, bytearray, memoryview]]) -> None:
```
```suggestion
    def write(self, data: Union[bytes, bytearray, memoryview]) -> None:
```

tests/test_benchmarks_http_websocket.py
```diff
@@ -1,10 +1,12 @@
 """codspeed benchmarks for http websocket."""
 
 import asyncio
+from typing import Iterable, Union
```
```suggestion
from typing import Union
```

tests/test_benchmarks_http_websocket.py
```diff
@@ -67,6 +69,22 @@ def _run() -> None:
         loop.run_until_complete(_send_one_hundred_websocket_text_messages())
 
 
+def test_send_one_hundred_large_websocket_text_messages(
+    loop: asyncio.AbstractEventLoop, benchmark: BenchmarkFixture
+) -> None:
+    """Benchmark sending 100 WebSocket text messages."""
+    writer = WebSocketWriter(MockProtocol(loop=loop), MockTransport())
+    raw_message = b"x" * MSG_SIZE * 2
```
```suggestion
    raw_message = b"x" * MSG_SIZE * 4
```

PR #9640, opened 2024-11-03
Merge FlowControlDataQueue into DataQueue

bdraco:
## What do these changes do?

`DataQueue` was only used in tests and the `super()` calls for all the `read` and `feed_data` added overhead.

no

no, less code to maintain


bdraco:
nevermind client_proto uses it

PR #9643, opened 2024-11-03
DNM: Benchmark overhead of checking WebSocket str and bytes types

bdraco:
send_str has a bit more overhead than expected:
Screenshot 2024-11-03 at 1 25 37 AM

It seems like it has to be the `isinstance` check but the benchmark will tell for sure


bdraco:
Maybe its the encode or the kwargs...

PR #9644, opened 2024-11-03
DNM: Benchmark overhead of checking WebSocket msg type on receive

bdraco:
receive has a bit more overhead than expected:
Screenshot 2024-11-03 at 1 26 06 AM

It seem like it has to be the `property` call overhead of the `NamedTuple`, but now sure so running the benchmark


bdraco:
Its 4.99% of `receive` https://codspeed.io/aio-libs/aiohttp/branches/msg_type_lookup

not really worth changing.

PR #9645, opened 2024-11-03
Add benchmark for roundtrip of 100 large WebSocket messages

bdraco:
SSIA

For testing if https://github.com/aio-libs/aiohttp/pull/9634 is viable

Looks like almost all the overhead is `memcpy`


bdraco:
Will need to tweak it a bit for 3.10

PR #9646, opened 2024-11-03
DNM: Benchmark WebSocket sending without kwargs

bdraco:
`send_str` has a bit more overhead than expected:
Screenshot 2024-11-03 at 1 25 37 AM

Check if removing the kwargs makes a measurable difference... probably not, but worth checking


bdraco:
makes no difference

PR #9649, opened 2024-11-03
Avoid memory copy in the WebSocket reader for small payloads

bdraco:
Only convert to `bytearray` when we know we are going to append messages or unmask.  If the message comes in without being fragmented, we can avoid many conversions from `bytes` to `bytearray` which is the common case for small messages

### DIFF
--- a/CHANGES/9649.feature.rst
+++ b/CHANGES/9649.feature.rst
@@ -1,0 +1,1 @@
+9543.feature.rst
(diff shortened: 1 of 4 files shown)


bdraco:
websocket

bdraco:
Screenshot 2024-11-03 at 3 44 15 AM

Less time spent parsing now than the `receive` and `DataQueue` calls

PR #9654, opened 2024-11-03
DNM: Make get_extra_info calls lazy

bdraco:
Testing the performance of this change.  This may not work ok though since we might want to call them after the connection has been closed... but that probably doesn't make much sense


bdraco:
make no difference

bdraco:
Screenshot 2024-11-17 at 7 35 57 PM

Might be worth pursuing if it can be made to work

bdraco:
looks like we have to save it in `__init__` just in case we might use it, sadly which is rare

asvetlov:
> looks like we have to save it in `__init__` just in case we might use it, sadly which is rare

Exactly!
I didn't find a better way to save SSL info.
It is two dict lookups only internally though

bdraco:
Yeah its pretty small on the profile so I threw it away pretty quickly as there are still better choices to optimize

PR #9655, opened 2024-11-03
Bump yarl to 1.17.1

bdraco:
Doing this ahead of dependabot to see what the performance impact is


bdraco:
Screenshot 2024-11-03 at 11 40 01 AM

make_url is faster but it doesn't have much overall impact here

PR #9659, opened 2024-11-03
Refactor `FlowControlDataQueue` to improve performances

bdraco:
The `DataQueue` base class tracks `size` but it never uses it. Move all the `size` tracking into the `FlowControlDataQueue`

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -630,14 +630,12 @@ class DataQueue(Generic[_SizedT]):
         self._eof = True
         self._exception = exc
 
-        waiter = self._waiter
-        if waiter is not None:
+        if (waiter := self._waiter) is not None:
             self._waiter = None
             set_exception(waiter, exc, exc_cause)
 
     def _release_waiter(self) -> None:
-        waiter = self._waiter
-        if waiter is not None:
+        if (waiter := self._waiter) is not None:
             self._waiter = None
             set_result(waiter, None)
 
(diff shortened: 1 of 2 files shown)


bdraco:
why is that super call so expensive.....

aiohttp/streams.py
```diff
@@ -698,8 +696,23 @@ def feed_data(self, data: _SizedT) -> None:
             self._protocol.pause_reading()
 
     async def read(self) -> _SizedT:
-        try:
-            return await super().read()
```
bdraco, line 702: This super call is more expensive than I would expect.

... writing this out twice is not good though....

need to come up with something better

aiohttp/streams.py
Dreamsorcerer, line 1: As DataQueue doesn't have _size anymore, the Generic should not require Sized anymore. Also, probably want to have WSMessage queues move to that, as they don't need the _size calculations.

bdraco: `_SizedT` is in this case `WSMessage` .... I think could be changed to `_T`

`FlowControlDataQueue` is only used by the WebSocket?  Does the WebSocket not need to pause the protocol when the limit is hit?

bdraco: https://github.com/aio-libs/aiohttp/pull/9677

Dreamsorcerer: Oh yeah. That logic can simply be removed then. I thought I remembered it being used in the parser or something with bytes, where it tracked the size of the data stream..

PR #9660, opened 2024-11-04
Add __slots__ to DataQueue

bdraco:
## What do these changes do?

The `StreamReader` has `__slots__` but `DataQueue` does not.

We have one in memory for every connection so it will save a tiny bit of RAM.

no
no


bdraco:
Looks like we would have to mess with mro to make this work and that gets messy

PR #9663, opened 2024-11-04
[3.11] Refactor FlowControlDataQueue.read to improve performances

bdraco:
test backport of https://github.com/aio-libs/aiohttp/pull/9659


bdraco:
too messy, will need to do some missed backports

PR #9666, opened 2024-11-04
DNM: Attempt to improve WebSocket receive performance

bdraco:
- Collapse the nested `try` into a single `try`
- We have a lot of checks at the end of receive that are rarely hit. Return early if we do not need to do any additional processing.
- Remove dead code in the web side (loop is unused)

### DIFF
--- a/aiohttp/client_ws.py
+++ b/aiohttp/client_ws.py
@@ -311,6 +311,10 @@ class ClientWebSocketResponse:
         else:
             return False
 
+    async def _close_from_receive(self, code: int) -> None:
+        self._waiting = False
+        await self.close(code=code)
+
     async def receive(self, timeout: Optional[float] = None) -> WSMessage:
         receive_timeout = timeout or self._timeout.ws_receive
 
@@ -341,8 +345,7 @@ class ClientWebSocketResponse:
                 raise
             except EofStream:
                 self._close_code = WSCloseCode.OK
-                self._waiting = False
-                await self.close()
+                await self._close_from_receive(WSCloseCode.OK)
                 return WS_CLOSED_MESSAGE
             except ClientError:
                 # Likely ServerDisconnectedError when connection is lost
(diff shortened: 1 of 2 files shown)


bdraco:
Screenshot 2024-11-03 at 9 10 03 PM

19% speed up to `receive`

aiohttp/client_ws.py
```diff
@@ -351,14 +361,21 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
                 self._close_code = exc.code
-                await self.close(code=exc.code)
+                await self._close_from_receive(exc.code)
                 return WSMessageError(data=exc)
             except Exception as exc:
                 self._exception = exc
                 self._set_closing()
                 self._close_code = WSCloseCode.ABNORMAL_CLOSURE
-                await self.close()
+                await self._close_from_receive(WSCloseCode.OK)
```
bdraco, line 370: This is a bit unexpected, but this is how is currently works. Maybe something to reconsider later.

aiohttp/web_ws.py
```diff
@@ -549,36 +558,38 @@ async def receive(self, timeout: Optional[float] = None) -> WSMessage:
...
             except WebSocketError as exc:
                 self._close_code = exc.code
-                await self.close(code=exc.code)
+                await self._close_from_receive(exc.code)
                 return WSMessageError(data=exc)
             except Exception as exc:
                 self._exception = exc
                 self._set_closing(WSCloseCode.ABNORMAL_CLOSURE)
-                await self.close()
+                await self._close_from_receive(WSCloseCode.OK)
```
bdraco, line 589: This is a bit unexpected, but this is how is currently works. Maybe something to reconsider later.

aiohttp/web_ws.py
```diff
@@ -528,12 +533,21 @@ def _close_transport(self) -> None:
...
+        """
+        self._waiting = False
+        await self.close(code=code)
+
     async def receive(self, timeout: Optional[float] = None) -> WSMessage:
         if self._reader is None:
             raise RuntimeError("Call .prepare() first")
 
-        loop = self._loop
-        assert loop is not None
```
bdraco, line 536: loop is unused, this is dead code

bdraco:
seems like this could be split into 3 PRs

.. although I'm not sure the nested `try` is better..

bdraco:
Need to sleep on this one. Definitely should be split up

bdraco:
Extracted into https://github.com/aio-libs/aiohttp/pull/9678 and https://github.com/aio-libs/aiohttp/pull/9679

The `try` change doesn't produce enough performance improvement to justify the code churn

PR #9671, opened 2024-11-04
Refactor connection waiters to be cancellation safe

### LINKED ISSUE #9670: deadlock with TCPConnector limit after timeout

bdraco:
A deadlock that could occur while attempting to get a new connection slot after a timeout.

The connector was not cancellation-safe.

fixes #9670

will need to backport https://github.com/aio-libs/aiohttp/pull/9600 to 3.10 as well

### DIFF
--- a/CHANGES/9671.bugfix.rst
+++ b/CHANGES/9671.bugfix.rst
@@ -1,3 +1,3 @@
-Fix a deadlock that could occur while attempting to get a new connection slot after a timeout -- by :user:`bdraco`.
+Fixed a deadlock that could occur while attempting to get a new connection slot after a timeout -- by :user:`bdraco`.
 
 The connector was not cancellation-safe.
--- a/CHANGES/9671.bugfix.rst
+++ b/CHANGES/9671.bugfix.rst
@@ -1,0 +1,3 @@
+Fix a deadlock that could occur while attempting to get a new connection slot after a timeout -- by :user:`bdraco`.
+
+The connector was not cancellation-safe.
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -500,7 +500,9 @@ class BaseConnector:
             return conn
 
         async with ceil_timeout(timeout.connect, timeout.ceil_threshold):
-            await self._wait_for_available_connection(key, traces)
+            if self._available_connections(key) <= 0:
+                await self._wait_for_available_connection(key, traces)
+
             if (conn := await self._get(key, traces)) is not None:
                 return conn
 
@@ -550,7 +552,7 @@ class BaseConnector:
         # need to loop again to check if the connection
         # slot is still available.
         wait_count = 0
-        while self._available_connections(key) <= 0:
+        while True:
             wait_count += 1
             fut: asyncio.Future[None] = self._loop.create_future()
             keyed_waiters = self._waiters[key]
@@ -577,6 +579,9 @@ class BaseConnector:
                 keyed_waiters.pop(fut, None)
                 if not self._waiters.get(key, True):
                     del self._waiters[key]
+
+            if self._available_connections(key) > 0:
+                break
 
     async def _get(
         self, key: "ConnectionKey", traces: List["Trace"]
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -532,11 +532,9 @@ class BaseConnector:
                         for trace in traces:
                             await trace.send_connection_queued_end()
                 finally:
-                    # This is the only point that the waiters
-                    # are removed from the queue. This ensures that
-                    # the waiters are always cleaned up and there
-                    # is no race between other parts of the code
-                    del keyed_waiters[fut]
+                    # pop the waiter from the queue if its still
+                    # there and not already removed by _release_waiter
+                    keyed_waiters.pop(fut, None)
                     if not self._waiters.get(key):
                         del self._waiters[key]
 
@@ -640,11 +638,9 @@ class BaseConnector:
             if self._available_connections(key) < 1:
                 continue
 
-            # Waiters are not removed from the queue here,
-            # and only one waiter is released at a time.
-            # They are only removed in a finally block
-            # in the connect method where they are added.
-            for waiter in self._waiters[key]:
+            waiters = self._waiters[key]
+            while waiters:
+                waiter, _ = waiters.popitem(last=False)
                 if not waiter.done():
                     waiter.set_result(None)
                     return
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -497,10 +497,11 @@ class BaseConnector:
         if (proto := self._get(key)) is not None:
             # If we do not have to wait and we can get a connection from the pool
             # we can avoid the timeout ceil logic and directly return the connection
+            conn = Connection(self, key, proto, self._loop)
             if traces:
                 for trace in traces:
                     await trace.send_connection_reuseconn()
-            return Connection(self, key, proto, self._loop)
+            return conn
 
         async with ceil_timeout(timeout.connect, timeout.ceil_threshold):
             #
@@ -545,10 +546,11 @@ class BaseConnector:
                         await trace.send_connection_queued_end()
 
                 if (proto := self._get(key)) is not None:
+                    conn = Connection(self, key, proto, self._loop)
                     if traces:
                         for trace in traces:
                             await trace.send_connection_reuseconn()
-                    return Connection(self, key, proto, self._loop)
+                    return conn
 
             placeholder = cast(
                 ResponseHandler, _TransportPlaceholder(self._placeholder_future)
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -368,7 +368,13 @@ class BaseConnector:
                 timeout_ceil_threshold=self._timeout_ceil_threshold,
             )
 
+    def _mark_acquired(self, key: "ConnectionKey", proto: ResponseHandler) -> None:
+        """Mark connection acquired."""
+        self._acquired.add(proto)
+        self._acquired_per_host[key].add(proto)
+
     def _drop_acquired(self, key: "ConnectionKey", val: ResponseHandler) -> None:
+        """Drop acquired connection."""
         self._acquired.discard(val)
         if conns := self._acquired_per_host.get(key):
             conns.discard(val)
@@ -566,10 +572,6 @@ class BaseConnector:
                 await trace.send_connection_reuseconn()
         return Connection(self, key, proto, self._loop)
 
-    def _mark_acquired(self, key: "ConnectionKey", proto: ResponseHandler) -> None:
-        self._acquired.add(proto)
-        self._acquired_per_host[key].add(proto)
-
     def _get(self, key: "ConnectionKey") -> Optional[ResponseHandler]:
         """Get next reusable connection for the key or None.
 
@@ -628,6 +630,7 @@ class BaseConnector:
                     return
 
     def _release_acquired(self, key: "ConnectionKey", proto: ResponseHandler) -> None:
+        """Release acquired connection."""
         if self._closed:
             # acquired connection is already released on connector closing
             return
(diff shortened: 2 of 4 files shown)


aiohttp/connector.py
```diff
@@ -490,8 +495,18 @@ async def connect(
         """Get from pool or create new connection."""
         key = req.connection_key
         available = self._available_connections(key)
-        wait_for_conn = available <= 0 or key in self._waiters
-        if not wait_for_conn and (proto := self._get(key)) is not None:
+        fut: Optional[asyncio.Future[None]] = None
+        waiters: Optional[Deque[asyncio.Future[None]]] = None
+        if wait_for_conn := available <= 0 or key in self._waiters:
```
bdraco: I think `key in self._waiters` might still be a race risk because one of the waiters could be cancelled

aiohttp/connector.py
```diff
@@ -561,31 +579,22 @@ def _acquired_connection(
...
-                # remove a waiter even if it was cancelled, normally it's
-                #  removed when it's notified
-                with suppress(ValueError):
-                    # fut may no longer be in list
-                    self._waiters[key].remove(fut)
-
-            raise e
         finally:
-            if key in self._waiters and not self._waiters[key]:
+            waiters.remove(fut)
```
bdraco: Its still possible to get ValueError if its released by `_release_waiter`

bdraco:
https://github.com/aio-libs/aiohttp/issues/9670 No longer reproducible with these changes .. but now we can go over the limit.

Previously it was over-enforcing. Now its under-enforcing

bdraco:
The core problem is we have three sources of truth for limits and there is no synchronization between them

`self._acquired`, `self._acquired_per_host`, `self._waiters`

bdraco:
Any await after calling `self._available_connections` invalidates it because it could change after that

bdraco:
This is turning out to be really hard to fix with the current design

aiohttp/connector.py
```diff
@@ -446,6 +447,10 @@ def _close_immediately(self) -> List[Awaitable[object]]:
         finally:
             self._conns.clear()
             self._acquired.clear()
+            for keyed_waiters in self._waiters.values():
+                for keyed_waiter in keyed_waiters:
+                    if not keyed_waiter.done():
```
Dreamsorcerer: I think this call is redundant. .cancel() already does that, if I remember correctly.

bdraco: agree
```
        if self._state != _PENDING:
            return False
```

bdraco: removed in 5a5260c99

bdraco:
I couldn't come up with anything better than looping the available check to make sure it actually available after the `await`

The tests pass now and the reproducer in the issue is fixed....not sure how great that will be for performance

bdraco:
unrelated but `__del__` isn't thread safe since it can run in any thread.

It should be `self._loop.call_soon_threadsafe(functools.partial(self._connector._release, self._key, self._protocol, should_close=True))`

aiohttp/connector.py
```diff
@@ -125,7 +125,15 @@ def __del__(self, _warnings: Any = warnings) -> None:
...
-            self._connector._release(self._key, self._protocol, should_close=True)
+            # __del__ may be called in different thread. We have to use call_soon_threadsafe.
+            self._loop.call_soon_threadsafe(
+                functools.partial(
+                    self._connector._release,
+                    self._key,
+                    self._protocol,
+                    should_close=True,
+                )
+            )
```
```suggestion
            self._connector._release(self._key, self._protocol, should_close=True)
```

can be fixed in a different PR now

aiohttp/connector.py
```diff
@@ -252,9 +252,11 @@ def __init__(
         self._force_close = force_close
 
         # {host_key: FIFO list of waiters}
-        self._waiters: DefaultDict[ConnectionKey, deque[asyncio.Future[None]]] = (
-            defaultdict(deque)
-        )
+        # The FIFO is implemented with a dict with None keys because
```
```suggestion
        # The FIFO is implemented with an OrderedDict with None keys because
```

aiohttp/connector.py
```diff
@@ -489,111 +494,102 @@ async def connect(
...
             # we can avoid the timeout ceil logic and directly return the connection
-            return await self._reused_connection(key, proto, traces)
+            if traces:
+                for trace in traces:
+                    try:
+                        await trace.send_connection_reuseconn()
+                    except BaseException:
+                        self._release_acquired(key, proto)
+                        raise
+            return Connection(self, key, proto, self._loop)
```
bdraco: Probably move all this logic into _get so we don't have to have it twice here

bdraco:
It all works now... but I have no idea on the performance impact, and its going to need some new tests

bdraco:
tests are patch out the points where cancellation might happen, and raise `asyncio.CancelledError`, than verify `self._acquired` is cleaned up

aiohttp/connector.py
```diff
@@ -640,27 +640,23 @@ def _release_waiter(self) -> None:
...
 
-            waiters = self._waiters[key]
-            while waiters:
-                waiter = waiters.popleft()
+            # Waiters are not removed from the queue here,
+            # and only one waiter is released at a time.
+            # They are only removed in a finally block
+            # in the connect method where they are added.
+            for waiter in self._waiters[key]:
                 if not waiter.done():
```
bdraco: Performance is really bad here :(

bdraco: refactored to fix this

bdraco:
cancellation is more likely because we reuse the newest connection first instead of the oldest connection which means the second reuse is more likely to get a closing connection

bdraco:
> cancellation is more likely because we reuse the newest connection first instead of the oldest connection which means the second reuse is more likely to get a closing connection

fixed that in https://github.com/aio-libs/aiohttp/pull/9672

bdraco:
functional testing looks good

performance looks good now

bdraco:
Some tests tomorrow if I have time.  I'll try to wrap this before flying to LHR on Thursday

CHANGES/9671.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fix a deadlock that could occur while attempting to get a new connection slot after a timeout -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed a deadlock that could occur while attempting to get a new connection slot after a timeout -- by :user:`bdraco`.
```

aiohttp/connector.py
```diff
@@ -489,111 +494,102 @@ async def connect(
...
         async with ceil_timeout(timeout.connect, timeout.ceil_threshold):
-            # Wait if there are no available connections or if there are/were
-            # waiters (i.e. don't steal connection from a waiter about to wake up)
-            if wait_for_conn:
+            if self._available_connections(key) <= 0:
                 await self._wait_for_available_connection(key, traces)
-                if (proto := self._get(key)) is not None:
-                    return await self._reused_connection(key, proto, traces)
+
+            if (conn := await self._get(key, traces)) is not None:
```
bdraco: This should keep the indent so we don't check twice

bdraco:
Screenshot 2024-11-05 at 9 14 48 AM

Still a tiny perf regression in `_get`

bdraco:
Testing looks good, and it looks like we aren't regressing performance now

PR #9672, opened 2024-11-05
Reuse the oldest keep-alive connection first

bdraco:
Previously we would reuse the newest one which meant as more and more connections were cycled the first one in the list became more and more stale until it was more likely to drop

Change the order from LIFO to FIFO


Dreamsorcerer:
Is it possible that's a deliberate decision to avoid keeping too many connections open? e.g. A quick burst opens up 3 connections, then going forward only 1 request is needed at a time, so the other 2 connections will get closed. Whereas this will stop connections from getting closed.

If you have evidence this approach is better, I'm fine with it. Just checking it's thought through (and maybe a comment added to the code to explain the design choice).

bdraco:
> Is it possible that's a deliberate decision to avoid keeping too many connections open? e.g. A quick burst opens up 3 connections, then going forward only 1 request is needed at a time, so the other 2 connections will get closed. Whereas this will stop connections from getting closed.
> 
> If you have evidence this approach is better, I'm fine with it. Just checking it's thought through (and maybe a comment added to the code to explain the design choice).

My original thought was that it might make sense to reuse the newest one since its less likely to be closed, but it seems more connections get reused in production with the FIFO approach.

I still want to do more testing with this though before I move it forward.  Since it conflicts with https://github.com/aio-libs/aiohttp/pull/9671 I've put it on the back burner as that one seems much more important to solve.

asvetlov:
FIFO sounds like a good strategy for me.
I don't recall the reason for implementing LIFO. 
It sits in the code from the very beginning.
I suspect it was added by accident.

aiohttp/connector.py
```diff
@@ -241,7 +242,9 @@ def __init__(
         if loop.get_debug():
             self._source_traceback = traceback.extract_stack(sys._getframe(1))
 
-        self._conns: Dict[ConnectionKey, List[Tuple[ResponseHandler, float]]] = {}
+        self._conns: DefaultDict[
```
bdraco, line 248: TODO: document this is FIFO

bdraco:
78% connection reuse before
84% connection reuse after

bdraco:
On another system where there is more /10s polling

73% before
86% after

bdraco:
So its a little bit more likely to reuse a connection.  YMMV based on the server keep alive timeout.

CHANGES/9672.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed the keep-alive connection pool to by FIFO instead of LIFO -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed the keep-alive connection pool to be FIFO instead of LIFO -- by :user:`bdraco`.
```

Olegt0rr:
> On another system where there is more /10s polling
> 
> 73% before 86% after

Could you share your code snippet to reproduce this behaviour?
Wanna try it on my environment
Thanks :)

bdraco:
> > On another system where there is more /10s polling
> 
> > 
> 
> > 73% before 86% after
> 
> 
> 
> @bdraco, 
> 
> Could you share your code snippet to reproduce this measurement?
> 
> Wanna try it on my environment
> 
> Thanks :)

Sorry there is no snippet. I was profiling a few different Home Assistant instances to gauge the effect with a real world use case.

bdraco:
The change is available in the 3.11.0 betas if you would like to give it a test

bdraco:
If you do a callgrind, look at the number of `__aenter__` calls. That will tell you how many didn't get reused

PR #9675, opened 2024-11-05
[PR #9671/37d9fe6 backport][3.11] Refactor connection waiters to be cancellation safe

bdraco:
(cherry picked from commit 37d9fe68387e3672604ff5b183f48538bca34f16)


bdraco:
Ran the reproducer against this and its confirmed fixed

PR #9677, opened 2024-11-05
DNM: Remove reading flow control from WebSocket

bdraco:
edit: This works well enough on 3.11 and 3.10 because we send the size in https://github.com/aio-libs/aiohttp/blob/49f65e6ccbacc39127072bc0fa35c2efc965096e/aiohttp/streams.py#L700 so I think this needs to be restored to use that behavior instead.

alternative https://github.com/aio-libs/aiohttp/pull/9685

originally posted by Dreamsorcerer in https://github.com/aio-libs/aiohttp/pull/9659#discussion_r1828188535 :
> As DataQueue doesn't have _size anymore, the Generic should not require Sized anymore. Also, probably want to have WSMessage queues move to that, as they don't need the _size calculations.

It appears this has never worked correctly, and has always calculated the size of every WebSocket message as `3` which is the length of the `WSMessage` tuple, and since the limit was enforced at 2x, there would have to be `43690` queued WebSocket messages before reading would pause.

The downside to not pausing reads is we don't create back-pressure the other direction which means we keep filling memory by processing messages into the queue if for some reason the downstream application stops processing the data.  However since its always been `3` per message it seems unlikely this was needed.

This does not affect back-pressure for the writer as there is a drain when the outgoing builds up.

The alternative would be to rework it to be WebSocket specific and add the size of `msg.data`

codspeed shows ~4% speed up on WebSocket round trip.

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -30,11 +30,9 @@ __all__ = (
     "EofStream",
     "StreamReader",
     "DataQueue",
-    "FlowControlDataQueue",
 )
 
 _T = TypeVar("_T")
-_SizedT = TypeVar("_SizedT", bound=collections.abc.Sized)
 
 
 class EofStream(Exception):
@@ -600,7 +598,7 @@ class EmptyStreamReader(StreamReader):  # lgtm [py/missing-call-to-init]
 EMPTY_PAYLOAD: Final[StreamReader] = EmptyStreamReader()
 
 
-class DataQueue(Generic[_SizedT]):
+class DataQueue(Generic[_T]):
     """DataQueue is a general-purpose blocking queue with one reader."""
 
     def __init__(self, loop: asyncio.AbstractEventLoop) -> None:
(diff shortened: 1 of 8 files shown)


aiohttp/streams.py
```diff
@@ -663,42 +661,5 @@ async def read(self) -> _SizedT:
...
-        self._limit = limit * 2
-
-    def feed_data(self, data: _SizedT) -> None:
-        self._size += len(data)
-        self._buffer.append(data)
-        if (waiter := self._waiter) is not None:
-            self._waiter = None
-            set_result(waiter, None)
-        if self._size > self._limit and not self._protocol._reading_paused:
-            self._protocol.pause_reading()
```
Dreamsorcerer, line 691: Not sure if this is still relevant? If so, it could just use `len(self._buffer)` now. But, the old code would have tripled the size (i.e. size is incremented by 3 for every message).

Dreamsorcerer: I guess this is what you're referring to in the PR description. But, my point is that we don't need to track the _size attribute at all for this logic (I assumed this was for when the generic was `bytes`, but don't see anywhere in the code it's actually used like that currently).

Dreamsorcerer: Alternatively, maybe _size should have been tracking `len(msg.data)` instead, in order to limit the total amount of bytes?

Currently, it's `len(msg)` which is always 3, and clearly not useful.

bdraco: I think your previous comment is proving it's not needed since it doesn't actually work correctly and would be unlikely to pause in production, which supports removing it

PR #9679, opened 2024-11-05
Return early in WebSocket `receive` if there is no processing to do

bdraco:
`receive` is a hot path for many downstream applications. In most we can avoid the linear search of message types to see if they need additional processing

extracted from #9666

Screenshot 2024-11-03 at 9 10 03 PM

19% speed up to `receive` for common message types. closing/closing/ping/pong will be a tiny bit slower since it has to do the `PySequence_Contains` but for any WebSocket connection that has any sort of volume, `TEXT` or `BINARY` is going to significant exceed these message types.

### DIFF
--- a/CHANGES/9679.misc.rst
+++ b/CHANGES/9679.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of calling ``receive`` for WebSockets for the most common message types -- by :user:`bdraco`.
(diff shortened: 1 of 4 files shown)


bdraco:
Its ~20% improvement in of the performance of `receive`.. but we don't have a direct benchmark

PR #9685, opened 2024-11-05
Fix WebSocket reader flow control calculations        

bdraco:
Fix WebSocket reader flow control calculations        

It appears this does not work correctly on `master`, as the calculated the size of every WebSocket message is `3` which is the length of the `WSMessage` tuple, and since the limit was enforced at 2x, there would have to be `43690` queued WebSocket messages before reading would pause.

This works well enough on 3.11 and 3.10 because we send the size in https://github.com/aio-libs/aiohttp/blob/49f65e6ccbacc39127072bc0fa35c2efc965096e/aiohttp/streams.py#L700 

For master, I've moved the `size` into the `WSMessage`

originally posted by Dreamsorcerer in https://github.com/aio-libs/aiohttp/pull/9659#discussion_r1828188535 :
> As DataQueue doesn't have _size anymore, the Generic should not require Sized anymore. Also, probably want to have WSMessage queues move to that, as they don't need the _size calculations.

The queues have been split into `DataQueue` and `WebSocketDataQueue`. Sized has been removed from `DataQueue`. `WebSocketDataQueue` is own class now because I didn't want to cythonize `DataQueue`, and it doesn't use many of the methods on the former `DataQueue` base class anyways.

This wasn't broken in any public stable release so no bugfix changelog message.  

It likely wouldn't warrant a changelog message if it wasn't for `FlowControlDataQueue` was available for import in `aiohttp` top level namespace, although it would be a bit unexpected that someone would use it externally, its technically a breaking change.


aiohttp/_websocket/reader_py.py
```diff
@@ -45,9 +47,47 @@
...
+                waiter.set_result(None)
+        if self._size > self._limit and not self._protocol._reading_paused:
+            self._protocol.pause_reading()
+
+    async def read(self) -> WSMessage:
+        if not self._buffer and not self._eof:
+            await self._wait_for_data()
+        if self._buffer:
+            data = self._buffer.popleft()
+            self._size -= len(data)
```
```suggestion
            self._size -= len(data[0])
```

aiohttp/client.py
```diff
@@ -1035,7 +1035,7 @@ async def _ws_connect(
 
             transport = conn.transport
             assert transport is not None
-            reader: FlowControlDataQueue[WSMessage] = FlowControlDataQueue(
+            reader: WebSocketDataQueue[WSMessage] = WebSocketDataQueue(
```
bdraco:
```suggestion
            reader = WebSocketDataQueue(
```

aiohttp/client.py
```diff
@@ -1035,7 +1035,7 @@ async def _ws_connect(
 
             transport = conn.transport
             assert transport is not None
-            reader: FlowControlDataQueue[WSMessage] = FlowControlDataQueue(
+            reader = WebSocketDataQueue(
                 conn_proto, 2**16, loop=self._loop
             )
```
bdraco:
```suggestion
            reader = WebSocketDataQueue(conn_proto, 2**16, loop=self._loop)
```

aiohttp/client_ws.py
```diff
@@ -46,7 +47,7 @@ class ClientWSTimeout:
 class ClientWebSocketResponse:
     def __init__(
         self,
-        reader: "FlowControlDataQueue[WSMessage]",
+        reader: "WebSocketDataQueue[WSMessage]",
```
bdraco:
```suggestion
        reader: WebSocketDataQueue,
```

aiohttp/web_ws.py
```diff
@@ -105,7 +106,7 @@ def __init__(
         self._protocols = protocols
         self._ws_protocol: Optional[str] = None
         self._writer: Optional[WebSocketWriter] = None
-        self._reader: Optional[FlowControlDataQueue[WSMessage]] = None
+        self._reader: Optional[WebSocketDataQueue[WSMessage]] = None
```
```suggestion
        self._reader: Optional[WebSocketDataQueue] = None
```

bdraco:
This feels like it's at the wrong level.  We shouldn't be counting the data once it's parsed. We should be looking at it before it's parsed

bdraco:
This still feels wrong because we miss a large percentage of the actual payload since the header is decoded before its counted

aiohttp/_websocket/reader_py.py
```diff
@@ -45,9 +47,55 @@
...
+        super().__init__(loop=loop)
+        self._size = 0
+        self._protocol = protocol
+        self._limit = limit * 2
+
+    def feed_data(self, data: WSMessage) -> None:
+        payload = data[0]
+        payload_type = type(payload)
+        self._size += (
+            len(payload) if payload_type is str or payload_type is bytes else 0
```
Dreamsorcerer:
```suggestion
            len(payload) if payload_type in {str, bytes} else 0
```
?

aiohttp/_websocket/reader_py.py
```diff
@@ -45,9 +47,55 @@
...
+    def __init__(
+        self, protocol: BaseProtocol, limit: int, *, loop: asyncio.AbstractEventLoop
+    ) -> None:
+        super().__init__(loop=loop)
+        self._size = 0
+        self._protocol = protocol
+        self._limit = limit * 2
+
+    def feed_data(self, data: WSMessage) -> None:
+        payload = data[0]
```
Dreamsorcerer: I think this is more readable (and robust against refactorings) to use the attribute. Seems to be the same performance-wise too.

```suggestion
    def feed_data(self, msg: WSMessage) -> None:
        payload = msg.data
```

Dreamsorcerer:
> This still feels wrong because we miss a large percentage of the actual payload since the header is decoded before its counted

I feel like the headers are a separate thing. This is only concerned with the amount of websocket data queued up. What it is missing is the msg.extra, which could technically be of any length I believe (but, mostly only used on EXIT messages, if I remember correctly, so probably not an issue with well-behaving servers).

aiohttp/_websocket/reader_py.py
```diff
@@ -219,16 +302,15 @@ def _feed_data(self, data: bytes) -> None:
...
 
-                self._queue_feed_data(msg)
-
+                self.queue.feed_data(msg, payload_len)
             elif opcode == OP_CODE_PING:
-                msg = WSMessagePing(data=payload, extra="")
-                self._queue_feed_data(msg)
-
+                self.queue.feed_data(
+                    WSMessagePing(data=payload, extra=""), len(payload)
```
bdraco: Maybe move the `len()` inside the message

tests/test_websocket_parser.py
```diff
@@ -481,9 +482,9 @@ def test_continuation_with_close_empty(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=0, extra="")
+        assert res == (WSMessageClose(data=0, extra=""), 0)
         res = out._buffer[1]
-        assert res == WSMessageText(data="line1line2", extra="")
+        assert res == (WSMessageText(data="line1line2", extra=""), 10)
```
bdraco:
```suggestion
        assert res == WSMessageText(data="line1line2", size=10, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -481,9 +482,9 @@ def test_continuation_with_close_empty(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=0, extra="")
+        assert res == (WSMessageClose(data=0, extra=""), 0)
```
bdraco:
```suggestion
        assert res == WSMessageClose(data=0, size=0, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -411,9 +412,9 @@ def test_continuation_with_close(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=1002, extra="test")
+        assert res == (WSMessageClose(data=1002, extra="test"), 6)
         res = out._buffer[1]
-        assert res == WSMessageText(data="line1line2", extra="")
+        assert res == (WSMessageText(data="line1line2", extra=""), 10)
```
bdraco:
```suggestion
        assert res == WSMessageText(data="line1line2", size=10, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -411,9 +412,9 @@ def test_continuation_with_close(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=1002, extra="test")
+        assert res == (WSMessageClose(data=1002, extra="test"), 6)
```
bdraco:
```suggestion
        assert res == WSMessageClose(data=1002, size=6, extra="test")
```

tests/test_websocket_parser.py
```diff
@@ -376,9 +377,9 @@ def test_continuation_with_ping(
         parser._feed_data(data3)
 
         res = out._buffer[0]
-        assert res == WSMessagePing(data=b"", extra="")
+        assert res == (WSMessagePing(data=b"", extra=""), 0)
         res = out._buffer[1]
-        assert res == WSMessageText(data="line1line2", extra="")
+        assert res == (WSMessageText(data="line1line2", extra=""), 10)
```
bdraco:
```suggestion
        assert res == WSMessageText(data="line1line2", size=10, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -376,9 +377,9 @@ def test_continuation_with_ping(
         parser._feed_data(data3)
 
         res = out._buffer[0]
-        assert res == WSMessagePing(data=b"", extra="")
+        assert res == (WSMessagePing(data=b"", extra=""), 0)
```
bdraco:
```suggestion
        assert res == WSMessagePing(data=b"", size=0 extra="")
```

tests/test_websocket_parser.py
```diff
@@ -353,7 +354,7 @@ def test_continuation(
     parser._feed_data(data2)
 
     res = out._buffer[0]
-    assert res == WSMessageText(data="line1line2", extra="")
+    assert res == (WSMessageText(data="line1line2", extra=""), 10)
```
bdraco:
```suggestion
    assert res == WSMessageText(data="line1line2", size=10, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -340,7 +341,7 @@ def test_fragmentation_header(
     parser._feed_data(data[1:])
 
     res = out._buffer[0]
-    assert res == WSMessageText(data="a", extra="")
+    assert res == (WSMessageText(data="a", extra=""), 1)
```
bdraco:
```suggestion
    assert res == WSMessageText(data="a", size=1, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -329,7 +330,7 @@ def test_simple_binary(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageBinary(data=b"binary", extra="")
+        assert res == (WSMessageBinary(data=b"binary", extra=""), 6)
```
bdraco:
```suggestion
        assert res == WSMessageBinary(data=b"binary", size=6, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -309,7 +310,7 @@ def test_simple_text(
     data = build_frame(b"text", WSMsgType.TEXT)
     parser._feed_data(data)
     res = out._buffer[0]
-    assert res == WSMessageText(data="text", extra="")
+    assert res == (WSMessageText(data="text", extra=""), 4)
```
bdraco:
```suggestion
    assert res == WSMessageText(data="text", size=4, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -258,7 +259,7 @@ def test_close_frame_info(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=12337, extra="12345")
+        assert res == (WSMessageClose(data=12337, extra="12345"), 7)
```
bdraco:
```suggestion
        assert res == WSMessageClose(data=12337, size=7, extra="12345")
```

tests/test_websocket_parser.py
```diff
@@ -247,7 +248,7 @@ def test_close_frame(
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessageClose(data=0, extra="")
+        assert res == (WSMessageClose(data=0, extra=""), 0)
```
bdraco:
```suggestion
        assert res == WSMessageClose(data=0, size=0, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -236,7 +237,7 @@ def test_pong_frame(out: aiohttp.DataQueue[WSMessage], parser: WebSocketReader)
 
         parser.feed_data(b"")
         res = out._buffer[0]
-        assert res == WSMessagePong(data=b"data", extra="")
+        assert res == (WSMessagePong(data=b"data", extra=""), 4)
```
bdraco:
```suggestion
        assert res == WSMessagePong(data=b"data", size=4, extra="")
```

tests/test_websocket_parser.py
```diff
@@ -227,7 +228,7 @@ def test_ping_frame(
 
         parser.feed_data(data)
         res = out._buffer[0]
-        assert res == WSMessagePing(data=b"data", extra="")
+        assert res == (WSMessagePing(data=b"data", extra=""), 4)
```
bdraco:
```suggestion
        assert res == WSMessagePing(data=b"data", size=4, extra="")
```

tests/test_websocket_data_queue.py
```diff
@@ -0,0 +1,36 @@
...
+    loop: asyncio.AbstractEventLoop, protocol: BaseProtocol
+) -> WebSocketDataQueue:
+    return WebSocketDataQueue(protocol, limit=1, loop=loop)
+
+
+class TestWebSocketDataQueue:
+    def test_feed_pause(self, buffer: WebSocketDataQueue) -> None:
+        buffer._protocol._reading_paused = False
+        for _ in range(3):
+            buffer.feed_data(WSMessageBinary(b"x"), 1)
```
bdraco:
```suggestion
            buffer.feed_data(WSMessageBinary(b"x", size=1), 1)
```

tests/test_websocket_data_queue.py
```diff
@@ -0,0 +1,36 @@
...
+
+        assert buffer._protocol.pause_reading.called  # type: ignore[attr-defined]
+
+    async def test_resume_on_read(self, buffer: WebSocketDataQueue) -> None:
+        buffer.feed_data(WSMessageBinary(b"x"), 1)
```
bdraco:
```suggestion
        buffer.feed_data(WSMessageBinary(b"x", size=1), 1)
```

bdraco:
still -5% on test_read_one_hundred_websocket_text_messages

aiohttp/_websocket/models.py
```diff
@@ -70,36 +77,42 @@ def json(
...
 class WSMessageClose(NamedTuple):
     data: int
+    size: int
     extra: Optional[str] = None
     type: Literal[WSMsgType.CLOSE] = WSMsgType.CLOSE
 
 
 class WSMessageClosing(NamedTuple):
     data: None = None
+    size: int = 0
```
Dreamsorcerer, line 97: I'm assuming this size allows us to skip the checks for message types in order to get a little extra performance?

aiohttp/_websocket/reader_py.py
```diff
@@ -45,12 +50,88 @@
...
+        self._waiter = None
+        if not waiter.done():
+            waiter.set_result(None)
+
+    def feed_eof(self) -> None:
+        self._eof = True
+        self._release_waiter()
+
+    def feed_data(self, data: "WSMessage") -> None:
+        size = data[WS_MSG_SIZE]
```
Dreamsorcerer: `data.size` is equally as fast in my testing, and then we probably don't need these constants.

bdraco: Let me switch it back and compare the profile

aiohttp/_websocket/models.py
bdraco: Shouldn't affect performance very much (maybe one less dict key in the kwargs). I did it for convenience of not having to write it out for the synthetic messages since they are always going to be 0

aiohttp/_websocket/reader_py.py
bdraco: Its a little bit slower but no material difference so lets leave it as `.size`

aiohttp/_websocket/reader_py.py
```diff
@@ -45,12 +49,88 @@
...
+        self._waiter = None
+        if not waiter.done():
+            waiter.set_result(None)
+
+    def feed_eof(self) -> None:
+        self._eof = True
+        self._release_waiter()
+
+    def feed_data(self, data: "WSMessage") -> None:
+        size = data.size
```
bdraco, line 98: `size` is still assigned here so `cython` will convert it to an `unsigned int` and avoid all the py math for `self._size` and `self._limit`

aiohttp/_websocket/reader_py.py
```diff
@@ -45,12 +49,88 @@
...
+                await self._waiter
+            except (asyncio.CancelledError, asyncio.TimeoutError):
+                self._waiter = None
+                raise
+        return self._read_from_buffer()
+
+    def _read_from_buffer(self) -> WSMessage:
+        if self._buffer:
+            data = self._get_buffer()
+            size = data.size
```
bdraco, line 119: `size` is still assigned here so `cython` will convert it to an `unsigned int` and avoid all the py math for `self._size` and `self._limit`

aiohttp/_websocket/reader_py.py
```diff
@@ -189,15 +270,16 @@ def _feed_data(self, data: bytes) -> None:
                     # bottleneck, so we use tuple.__new__ to improve performance.
                     # This is not type safe, but many tests should fail in
                     # test_client_ws_functional.py if this is wrong.
-                    msg = TUPLE_NEW(WSMessageText, (text, "", WS_MSG_TYPE_TEXT))
+                    msg = TUPLE_NEW(WSMessageText, (text, size, "", WS_MSG_TYPE_TEXT))
```
bdraco, line 273:
```suggestion
                    msg = TUPLE_NEW(WSMessageText, (text, len(text), "", WS_MSG_TYPE_TEXT))
```

aiohttp/_websocket/reader_py.py
```diff
@@ -189,15 +270,16 @@ def _feed_data(self, data: bytes) -> None:
...
                 else:
                     msg = TUPLE_NEW(
-                        WSMessageBinary, (payload_merged, "", WS_MSG_TYPE_BINARY)
+                        WSMessageBinary, (payload_merged, size, "", WS_MSG_TYPE_BINARY)
```
bdraco, line 276:
```suggestion
                        WSMessageBinary, (payload_merged, len(payload_merged), "", WS_MSG_TYPE_BINARY)
```

aiohttp/_websocket/reader_py.py
```diff
@@ -177,6 +257,7 @@ def _feed_data(self, data: bytes) -> None:
                 else:
                     payload_merged = bytes(assembled_payload)
 
+                size = len(payload_merged)
```
bdraco, line 260:
```suggestion
```

aiohttp/_websocket/reader_py.py
```diff
@@ -189,15 +269,19 @@ def _feed_data(self, data: bytes) -> None:
                     # bottleneck, so we use tuple.__new__ to improve performance.
                     # This is not type safe, but many tests should fail in
                     # test_client_ws_functional.py if this is wrong.
-                    msg = TUPLE_NEW(WSMessageText, (text, "", WS_MSG_TYPE_TEXT))
+                    msg = TUPLE_NEW(
+                        WSMessageText, (text, len(text), "", WS_MSG_TYPE_TEXT)
```
bdraco: On second thought...

While 3.10 and 3.11 gets the length of the decoded text..... I think its actually better to get the len of the bytes since thats whats actually taking up memory

bdraco: I guess thats worth fixing on 3.11/3.10 as well since if its a lot of 4 byte chars, it would under calculate the limit quite a bit

bdraco: https://github.com/aio-libs/aiohttp/pull/9686 to fix 3.11/3.10

bdraco:
I'm pretty happy with this now. It does need some new tests to ensure the flow control calculation remains meaningful

bdraco:
Maybe it would be best to do the tests as xfail, backport them to 3.10/3.11, and remove the xfail there so we get coverage down the chain

CHANGES/9685.breaking.rst
```diff
@@ -0,0 +1 @@
+The ``FlowControlDataQueue`` has been replaced with the ``WebSocketDataQueue`` -- by :user:`bdraco`.
```
bdraco: This wasn't broken in any public stable release so no bugfix changelog message.  

It likely wouldn't warrant a changelog message if it wasn't for `FlowControlDataQueue` was available for import in `aiohttp` top level namespace, although it would be a bit unexpected that someone would use it externally, its technically a breaking change.

bdraco:
```suggestion
``FlowControlDataQueue`` has been replaced with the ``WebSocketDataQueue`` -- by :user:`bdraco`.
```

bdraco:
xfail tests unmarked xfail and all seems well here.

bdraco:
I did a github code search and could not find any use cases of FlowControlDataQueue outside of aiohttp. I think this is safe to partially backport, and it would make it a lot easier to not have to maintain different code in 3.11.

I'm going to tag, it backport it and tweak it to work with 3.11 so it still keeps the same format for the WebSocket messages and uses the tuple to pass the size like it does now

PR #9686, opened 2024-11-06
[3.11] Fix WebSocket reader flow control size calculation for multi-byte data

bdraco:
The decoded size was passed to the `FlowControlDataQueue` instead of the bytes size which meant multi-byte characters would be under counted.

Note that this code is completely different on 3.11/3.10 than `master` so this PR is directly to 3.11


bdraco:
https://www.merriam-webster.com/dictionary/undercount

Hmm.. guess its not in the spelling words

PR #9692, opened 2024-11-07
Change `RequestInfo` to be a `NamedTuple` to improve performances

### LINKED ISSUE #9504: RequestInfo is documented to be a namedtuple but its actually a dataclass or attrs
https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponse.request_info

It would be about 2x as fast to create if it were a namedtuple though, and although dataclasses are faster to access its rarely accessed so namedtuple probably is a better choice.  frozen dataclasses are also slower to create https://docs.python.org/3.9/library/dataclasses.html#frozen-instances.... and we create them for every request.
> There is a tiny performance penalty when using frozen=True: [__init__()](https://docs.python.org/3.9/reference/datamodel.html#object.__init__) cannot use simple assignment to initialize fields, and must use [object.__setattr__()](https://docs.python.org/3.9/reference/datamodel.html#object.__setattr__).

Additionally the attrs default isn't used on 3.11/3.10 at all

https://github.com/aio-libs/aiohttp/blob/13dc0200d855bdfec1d4c4f3e1c9a1c66f88f1eb/aiohttp/client_reqrep.py#L110

```diff
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index dad7fe78e..1eeaf7068 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -105,16 +105,12 @@ class ContentDisposition:
     filename: Optional[str]
 
 
-@attr.s(auto_attribs=True, frozen=True, slots=True)
-class RequestInfo:
+class RequestInfo(NamedTuple):
     url: URL
     method: str
     headers: "CIMultiDictProxy[str]"
-    real_url: URL = attr.ib()
+    real_url: URL
 
-    @real_url.default
-    def real_url_default(self) -> URL:
-        return self.url
 
 
 class Fingerprint:
@@ -401,7 +397,7 @@ class ClientRequest:
     @property
     def request_info(self) -> RequestInfo:
         headers: CIMultiDictProxy[str] = CIMultiDictProxy(self.headers)
-        return RequestInfo(self.url, self.method, headers, self.original_url)
+        return tuple.__new__(RequestInfo, (self.url, self.method, headers, self.original_url))
 
     def update_host(self, url: URL) -> None:
         """Update destination host, port and connection type (ssl)."""
```

bdraco:
closes #9504

2.7x faster creating the `RequestInfo`... 6.16% improvement in `send` performance

Screenshot 2024-11-06 at 9 51 27 PM

skipping backport to 3.10 like we did with #9365 (similar change)


bdraco:
skipping backport to 3.10 like we did with #9365 (similar change)

asvetlov (approved):

PR #9693, opened 2024-11-07
Avoid duplicate `content_length` fetch to prepare web response headers

bdraco:
issue #2779

Note: we don't have a benchmark for generating a web response yet so its only visible in the other side of client get request.
Screenshot 2024-11-06 at 10 19 30 PM

Screenshot 2024-11-06 at 10 18 11 PM

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -399,7 +399,7 @@ class StreamResponse(BaseClass, HeadersMixin, CookieMixin):
                 headers[hdrs.TRANSFER_ENCODING] = "chunked"
             if hdrs.CONTENT_LENGTH in headers:
                 del headers[hdrs.CONTENT_LENGTH]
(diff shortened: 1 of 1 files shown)


aiohttp/web_response.py
```diff
@@ -420,7 +420,7 @@ async def _prepare_headers(self) -> None:
             # https://datatracker.ietf.org/doc/html/rfc9112#section-6.1-13
             if hdrs.TRANSFER_ENCODING in headers:
                 del headers[hdrs.TRANSFER_ENCODING]
-        elif self.content_length != 0:
+        elif writer.length != 0:
```
bdraco: Would be safer to guard this to fallback to content length if length check is off.  Right now it can't happen though.  Maybe not worth the complexity since it already has a test to prevent regressions

PR #9695, opened 2024-11-07
Relax async-timeout version to >=4.0, <6.0

asvetlov:
`async-timeout` 5.0+ is backward-compatible with 4.0; so it works with aiohttp well.

While `aiohttp` itself uses the standard `asyncio.Timeout` on Python 3.11+,
dependent project could require the latest async-timeout version on Python 3.9-3.10.


Dreamsorcerer (approved):

bdraco:
I can take care of the backports. I'm going to try get a 3.11.0b4 out today before I leave for the airport.

bdraco:
haha.  Too slow :)

PR #9714, opened 2024-11-07
Enforce URL for `cookie_jar.filter_cookies()` call (#3753)

bdraco:
Note that this does not prevent `str` from being used, back-porting so users don't get a surprise when they update to 4.x and it will stop working.

* Enforce URL for cookie_jar.filter_cookies() call

* Add deprecation warning for strings

(cherry picked from commit 401c2560c4b8fc5d497ec6125953298135dc6f17)

### DIFF
--- a/aiohttp/cookiejar.py
+++ b/aiohttp/cookiejar.py
@@ -9,6 +9,7 @@ import pathlib
 import pickle
 import re
 import time
+import warnings
 from collections import defaultdict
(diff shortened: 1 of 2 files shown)


bdraco:
Actually lets wait until 3.11 to start warning. 3.10 is already a bit long in the tooth and its kinda odd to start telling them in a patch release

PR #9722, opened 2024-11-09
Simplify header and method lookups

bdraco:
Replace `method_must_be_empty_body` and `status_code_must_be_empty_body` with a simple set constant, and avoid many `.upper()` calls

~2% speed up
Screenshot 2024-11-09 at 9 18 18 AM


CHANGES/9722.misc.rst
```diff
@@ -0,0 +1 @@
+Replace internal helper methods ``method_must_be_empty_body`` and ``status_code_must_be_empty_body`` with simple set lookups -- by :user:`bdraco`.
```
bdraco:
```suggestion
Replace internal helper methods ``method_must_be_empty_body`` and ``status_code_must_be_empty_body`` with simple `set` lookups -- by :user:`bdraco`.
```

PR #9732, opened 2024-11-09
Add Client Middleware Support

### LINKED ISSUE #434: Implementing auth handlers with `aiohttp.client`

### LINKED ISSUE #6915: Support dynamically updating authentication headers

bdraco:
## Summary

This PR introduces middleware support for the aiohttp client, allowing users to intercept and modify requests and responses. This brings the client API closer to parity with the server API and enables powerful use cases like authentication, logging, retries, and request/response modification.

### Core Implementation
- Added `aiohttp.client_middlewares` module with:
  - `ClientMiddlewareType` type alias for middleware functions
  - `ClientHandlerType` type alias for request handlers  
  - `build_client_middlewares()` function to chain middlewares efficiently
  
- Modified `ClientSession` to accept `middlewares` parameter:
  - Supports both session-level and request-level middleware
  - Added `_middlewares` attribute to store session middlewares
  - Modified request handling to apply middleware chain

- Added `session` attribute to `ClientRequest`:
  - Allows middleware to access the parent session
  - Enables middleware to make additional requests (useful for OAuth token refresh, retries, etc.)

### Documentation
- Comprehensive documentation in `client_advanced.rst`:
  - Basic middleware structure and types
  - How to apply middleware at session and request level
  - Common middleware patterns (logging, authentication, retries)
  - Middleware execution order and chaining
  - Best practices and caveats

### Tests
- Extensive test coverage in `test_client_middleware.py` (~1100 lines):
  - Basic middleware functionality
  - Multiple middleware chaining
  - Order of execution
  - Request/response modification
  - Error handling
  - Retry patterns
  - Recursion prevention
  - Integration with existing client features

## Features Enabled

1. **Authentication Middleware**: Automatically add authentication headers or refresh tokens
2. **Logging Middleware**: Log all requests and responses
3. **Retry Middleware**: Implement custom retry logic with backoff
4. **Request/Response Modification**: Modify headers, URLs, or response data
5. **Circuit Breaker Pattern**: Implement circuit breakers for failing services
6. **Rate Limiting**: Implement client-side rate limiting

## Example Usage

```python
# Basic middleware
async def logging_middleware(request, handler):
    print(f"Request: {request.method} {request.url}")
    response = await handler(request)
    print(f"Response: {response.status}")
    return response

# Apply to session
async with ClientSession(middlewares=(logging_middleware,)) as session:
    resp = await session.get('http://example.com')

# Apply to specific request
    resp = await session.get('http://example.com', middlewares=(logging_middleware,))
```

## Technical Details

- Middleware execution is optimized for minimal overhead
- Single middleware case has a dedicated fast path
- Middleware chain is built without using `partial` or `update_wrapper` to reduce overhead
- No caching of middleware chains to avoid holding references to stateful middleware

## Breaking Changes

None. This is a backwards-compatible addition.

## Related Issues

- Fixes #6915 - Request for client middleware support
- Fixes #434 - Add middleware for client requests  
- Addresses #4939 - Client middleware patterns for authentication and retries


Dreamsorcerer:
Also #6915. There are some examples in there of different use cases which should work when this is complete.

aiohttp/client_middlewares.py
```diff
@@ -0,0 +1,58 @@
...
+    The middlewares are applied in reverse order, so the first middleware
+    in the list wraps all subsequent middlewares and the handler.
+
+    This implementation avoids using partial/update_wrapper to minimize overhead
+    and doesn't cache to avoid holding references to stateful middleware.
+    """
+    if not middlewares:
+        return handler
+
+    # Optimize for single middleware case
```
Dreamsorcerer, line 34: What's the difference in performance? Is it just due to how the variable is referenced?

aiohttp/client_middlewares.py
```diff
@@ -0,0 +1,58 @@
+"""Client middleware support."""
+
+from collections.abc import Awaitable, Callable
+
+from .client_reqrep import ClientRequest, ClientResponse
+
+__all__ = ("ClientMiddlewareType", "ClientHandlerType", "build_client_middlewares")
+
+# Type alias for client request handlers - functions that process requests and return responses
+ClientHandlerType = Callable[..., Awaitable[ClientResponse]]
```
Dreamsorcerer: Looks like this needs to be ClientRequest:
```suggestion
ClientHandlerType = Callable[[ClientRequest], Awaitable[ClientResponse]]
```

docs/client_advanced.rst
```diff
@@ -105,6 +105,141 @@ background.
...
+            if not hasattr(request, 'retry_count'):
+                request.retry_count = 0
+
+            response = await handler(request)
+
+            # Retry on 5xx errors
+            if response.status >= 500 and request.retry_count < self.max_retries:
+                request.retry_count += 1
+                _LOGGER.debug(f"Retrying request (attempt {request.retry_count})")
+                raise ClientMiddlewareRetry()
```
Dreamsorcerer: If I'm reading correctly, this is only needed to recreate the connection?
i.e. `conn = await self._connector.connect(`

What if we moved that creation into the handler (_send_request())? Would it then allow this middleware to work with a simple loop? Like:
```
for _ in range(retries):
    response = await handler(request)
    ...
```

Or is there another reason I'm missing?

aiohttp/client_middlewares.py
bdraco: There is no function call to make it for the single handler case

docs/client_advanced.rst
bdraco: The retrying on specific exception was making for cases like digest auth or something that needs to change the request and try again like https://github.com/aio-libs/aiohttp/pull/9732/files#diff-6a1c4584137f03afa592db25d959da016a36bcca1cc1473ce591e97f62f44bc4R148

docs/client_advanced.rst
```diff
@@ -105,6 +105,141 @@ background.
...
+
+      Callable[
+          [ClientRequest, ClientHandlerType],
+          Awaitable[ClientResponse]
+      ]
+
+:type:`ClientHandlerType`
+   Type alias for client request handler functions::
+
+      Callable[..., Awaitable[ClientResponse]]
```
bdraco:
```suggestion
      Callable[ClientRequest, Awaitable[ClientResponse]]
```

aiohttp/client.py
```diff
@@ -667,17 +675,40 @@ async def _request(
...
+                                await resp.start(_conn)
                             except BaseException:
                                 resp.close()
                                 raise
                         except BaseException:
-                            conn.close()
+                            _conn.close()
                             raise
+
+                        return resp
```
bdraco: This pattern is a problem because if the middleware raises we won't close the connection. We need to remove the outer try and instead catch it below

aiohttp/client.py
```diff
@@ -667,28 +675,49 @@ async def _request(
                         max_field_size=max_field_size,
                     )
 
-                    try:
+                    # Core request handler - use default parameter to avoid loop variable issues
+                    async def _send_request(
```
bdraco: I think this ends up running too late for the case we want to block the connection

bdraco: adjusted

docs/client_advanced.rst
bdraco: I improved the example and docs a bit

tests/test_connector.py
```diff
@@ -3307,10 +3307,7 @@ async def handler(request: web.Request) -> NoReturn:
 
     srv = await aiohttp_server(app, ssl=ssl_ctx)
 
-    port = unused_port()
-    conn = aiohttp.TCPConnector(local_addr=("127.0.0.1", port))
```
bdraco: This one appears to always have been failing but now the exception about the duplicate port use is obvious

docs/conf.py
```diff
@@ -407,6 +407,8 @@
     ("py:mod", "aiohttp.test_tools"),  # undocumented
     ("py:class", "list of pairs"),  # undocumented
     ("py:class", "aiohttp.protocol.HttpVersion"),  # undocumented
+    ("py:type", "ClientMiddlewareType"),  # type alias in client_middlewares
+    ("py:type", "ClientHandlerType"),  # type alias in client_middlewares
```
bdraco: note sure this makes sense.. maybe I'm just doing it wrong in the docs

tests/test_connector.py
bdraco: nope, we needed to trap connector errors and prevent retry. all fixed

aiohttp/client_middlewares.py
Dreamsorcerer: Oh, of course, I forgot we're rebuilding on every request.

examples/client_middleware_ssrf.py
```diff
@@ -0,0 +1,228 @@
...
+            except SSRFProtectionError as e:
+                if not should_succeed:
+                    _LOGGER.info("  ✓ Blocked: %s", e)
+                else:
+                    _LOGGER.error("  ✗ FAILED: Should have been allowed! %s", e)
+            except Exception as e:
+                _LOGGER.info("  ~ Error: %s - %s", type(e).__name__, str(e)[:50])
+
+
+async def test_redirect_protection(server_url: str) -> None:
```
Dreamsorcerer: I don't really see how this one is different from the one above?

I think to protect from redirects, we need a custom resolver which rejects anything that resolves to a blacklisted IP. So, full SSRF protection will consist of 1 middleware and 1 custom resolver.

docs/client_advanced.rst
Dreamsorcerer: So, the exception will result in:

- some error checks that probably don't need to be repeated.
- Reprocessing cookies that don't look like they'd need to be repeated (unless a rare situation sets a cookie in a response _and_ the middleware wants to retry while using that cookie).
- Rechecking proxy details that probably doesn't want to be repeated.
- Recreating a ClientRequest with (as far as I can tell) all the same details.
- Recreating all the middlewares.

Maybe I'm missing something, but it looks to me like the end result would be the same as the middleware running a loop (other than not rerunning any outer middlewares). I don't think I can see any code that changes the request in the digest middleware. Even if it did need to change it, it could always edit/clone the request object and pass that through to the handler.

I guess I'm still struggling to understand why this wouldn't work:
```python
while True:
    if self.last_challenge:
        ...
    response = await handler(request)
    if response.status != 401:
        return response
    ...
    continue
```

examples/client_middleware_ssrf.py
Dreamsorcerer: Oh wait, I see, the request is retried with the redirected domain. So, probably OK, but if a domain directly resolves to a restricted IP, it might need a custom resolver (I guess the security risk depends on whether that service will process a request with the wrong Host header or not).

I feel like these 2 tests should be merged though so they use the same middleware and blacklist.

Dreamsorcerer: Hmm, how is a CNAME processed? Will that produce the same result (i.e. resolves to the internal IP, but Host header is still set to the original malicious domain)?

docs/client_advanced.rst
bdraco: > I guess I'm still struggling to understand why this wouldn't work:
> 
> ```python
> while True:
>     if self.last_challenge:
>         ...
>     response = await handler(request)
>     if response.status != 401:
>         return response
>     ...
>     continue
> ```

Can you give a more in-depth example of where you are going with that? I'm having trouble parsing what you are suggesting without a bit more context.

Dreamsorcerer: I guess I'm just not clear why we need to raise to retry, instead of just repeating in a loop, without redoing all the setup work.

examples/client_middleware_ssrf.py
bdraco: A proper implementation would need a custom resolver. This example was intended to demonstrate the client middleware part of the equation only.  Both the examples are throw aways for demonstration purposes only and should be deleted before merging.

docs/client_advanced.rst
bdraco: In that case, how would the client middleware signal to aiohttp that the request should be retried, without raising?

Dreamsorcerer: Isn't calling `handler(request)` multiple times the same as retrying?

bdraco: What I'm not understanding is how the middleware should tell the loop to continue or finish?

Dreamsorcerer: My point is it's got it's own loop, so the loop you're referring to doesn't need to loop.

Dreamsorcerer: i.e. That loop is primarily for handling redirects (where the url etc. gets updated near the end of the loop). To retry with the same request, that loop is not really needed, it can just be done in a middleware with a tighter loop.

Dreamsorcerer: So, current version looks like:
```
while ...:
    # Do checks on url, cookies etc.
    # Create ClientRequest
    try:
        # Middleware modifies request
        response = await handler(request)
        if not response.ok:
            raise Retry()
    except Retry:
        continue
```

I'm proposing writing it like:
```
while ...:
    # Do checks on url, cookies etc.
    # Create ClientRequest
    while ...:
        # Middleware modifies request
        response = await handler(request)
        if not response.ok:
            continue
```

The inner loop (inside the middleware) avoids needing to redo a bunch of work to make the same request and doesn't need a custom exception to achieve. Unless there's something I'm misunderstanding...

bdraco: Thanks for the additional color. I'm a bit worried about using `response.ok` as it make the implementation a lot less flexible if they want to key on some other aspect about the response.

bdraco: With that pattern how would we distinguish between an expected value an one we wanted to retry?

ie how do I know that `not response.ok` should continue or should fail?

bdraco: I pushed a change to add the inner loop

Dreamsorcerer: > With that pattern how would we distinguish between an expected value an one we wanted to retry?
> 
> ie how do I know that `not response.ok` should continue or should fail?

I was just being too lazy to look up and write out the whole thing, that's not relevant.

> I pushed a change to add the inner loop

My suggestion was more for the retry middleware itself to have the loop, thus eliminating the exception.

bdraco: Thanks for sticking with me on this. I feel like I am missing a key part of your idea. How do you see the middleware signaling whether it should retry or not? I'm not quite seeing how a `continue` would work if the logic is user-provided and running in a separate function. I think a more detailed example might help clarify things since it seems like we're not quite on the same page yet.

Dreamsorcerer: Why does it need to signal anything, there is nothing outside the loop that needs to run?
Modifying my previous example, I've added some comments to make it clear where the middleware is here (I was just trying to flatten the example):
```
while ...:
    # Do checks on url, cookies etc.
    # Create ClientRequest
    ### START MIDDLEWARE
    while ...:
        # Middleware modifies request
        response = await handler(request)
        if not response.ok:
            continue
    ### END MIDDLEWARE
```

Dreamsorcerer: As far as I can see, what you've just done is:
```
while ...:
    # Do checks on url, cookies etc.
    # Create ClientRequest
    while ...:
        try:
            ### START MIDDLEWARE
            # Middleware modifies request
            response = await handler(request)
            if not response.ok:
                raise Retry()
            ### END MIDDLEWARE
        except Retry:
            continue
```

Which looks functionally the same to me, but with an exception instead of a continue.

bdraco: Sorry, I just not seeing it.

Let's take `test_client_middleware_multi_step_auth` as an example.

https://github.com/aio-libs/aiohttp/pull/9732/files#diff-2d48bab5c5d394f9efdae40fd7ce2355025d06d74b5cee1b8f196d5993ac61baR281

It only raises the `ClientMiddlewareRetry` exception on 401 and `X-Auth-Step` is in the headers.   How can we achieve that level of control when the handler returns a response object?  Are you suggesting a whole new format for the middleware other than whats in `client_middleware.py`? 

Should we unconditionally rerun the middleware until it gets into an `.ok` state?

Either way we probably need a max loops since in other case we would retry forever (if they keep raising `ClientMiddlewareRetry` or the response is not `.ok`

Dreamsorcerer: I hadn't looked at the multi-step one. But, wouldn't those steps normally need to be done against different URLs? In which case I'd expect the middleware to trigger completely new requests before proceeding with the current one (like in https://github.com/aio-libs/aiohttp/issues/6915#issuecomment-2466782302).


Still, I don't see a reason I can't just loop inside the middleware like:
```
    async def multi_step_auth_middleware(
        request: ClientRequest, handler: ClientHandlerType
    ) -> ClientResponse:
        request.headers["X-Client-ID"] = "test-client"
        middleware_state = {"step": 0}

        while True:
            # Apply auth based on current state
            if middleware_state["step"] == 1 and middleware_state["session"]:
                request.headers["Authorization"] = f"Bearer {middleware_state['session']}"
            elif middleware_state["step"] == 2 and middleware_state["challenge"]:
                request.headers["Authorization"] = (
                    f"Bearer {middleware_state['challenge']}-response"
                )

            response = await handler(request)

            # Handle multi-step auth flow
            if response.status == 401:
                auth_step = response.headers.get("X-Auth-Step")

                if auth_step == "1":
                    # First step: store session token
                    middleware_state["session"] = response.headers.get("X-Session")
                    middleware_state["step"] = 1
                    continue

                elif auth_step == "2":
                    # Second step: store challenge
                    middleware_state["challenge"] = response.headers.get("X-Challenge")
                    middleware_state["step"] = 2
                    continue

            return response
```

bdraco: I get what you are suggesting now. Let me give it a shot

bdraco: Looks like that will work with all the existing use cases

bdraco: Pushed changes. Let me know if thats was what you were thinking

aiohttp/client.py
```diff
@@ -642,32 +648,36 @@ async def _request(
...
-                    # connection timeout
-                    try:
-                        conn = await self._connector.connect(
-                            req, traces=traces, timeout=real_timeout
+                    # Core request handler - now includes connection logic
+                    async def _connect_and_send_request(
+                        req: ClientRequest,
+                        *,
+                        _url: URL = url,
+                        _method: str = method,
```
Dreamsorcerer: These are the first 2 arguments to ClientRequest, so we must be able to retrieve them from req. Once that is done, we don't need a closure anymore, right? In which case building the middlewares can be cached like the server-side ones, and the single middleware shortcut removed.

examples/client_middleware_auth.py
```diff
@@ -0,0 +1,525 @@
...
+                request.headers["Authorization"] = f"Bearer {self.access_token}"
+
+            # Make the request
+            response = await handler(request)
+
+            # Check if token expired
+            if response.status == 401 and not retry_with_refresh:
+                _LOGGER.info("Got 401, attempting to refresh token")
+                # Try to refresh the token
+                await self.refresh_access_token()
```
Dreamsorcerer: Can we verify this also works by just passing request.session rather than needing a reference to the session beforehand (we'll need to add a new property method)?

examples/client_middleware_auth.py
Dreamsorcerer, line 1: Just to sanity check, can we auth based on request body (https://github.com/aio-libs/aiohttp/discussions/6627)? Assuming the body is of a reparsable type (str, bytes...), I guess.

tests/test_client_middleware.py
Dreamsorcerer, line 1: I would also like to see one test that uses a simple middleware like https://github.com/aio-libs/aiohttp/issues/6915#issuecomment-2466782302
I can imagine a nested request potentially causing a problem, so would like to have some test coverage on that.

aiohttp/client.py
Dreamsorcerer: Hmm, maybe not quite, many of the conn.protocol.set_response_params() arguments are from the outer scope. Maybe we can pass them into the request object as well, if it avoids us rebuilding the middlewares every time. Could probably just make a TypedDict of the arguments and then unpack them directly into that call.

Dreamsorcerer: real_timeout appears to be needed too.

tests/test_client_middleware.py
bdraco: Are you worried about them capturing the session in the middleware somehow since `request.session` doesn't seem to exist?

`AttributeError: 'ClientRequest' object has no attribute 'session'`

bdraco: maybe something like this.... but seems kinda horrible

```diff
diff --git a/aiohttp/client.py b/aiohttp/client.py
index ff1610593..659029257 100644
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -10,6 +10,7 @@ import sys
 import traceback
 import warnings
 from contextlib import suppress
+from contextvars import ContextVar
 from types import TracebackType
 from typing import (
     TYPE_CHECKING,
@@ -219,6 +220,11 @@ class ClientTimeout:
     # to overwrite the defaults
 
 
+# Context variable to track middleware recursion depth per URL within a request
+_middleware_depth_context: ContextVar[dict[URL, int]] = ContextVar(
+    "_middleware_depth_context", default={}
+)
+
 # 5 Minute default read timeout
 DEFAULT_TIMEOUT: Final[ClientTimeout] = ClientTimeout(total=5 * 60, sock_connect=30)
 
@@ -702,8 +708,26 @@ class ClientSession:
                     else:
                         handler = _connect_and_send_request
 
+                    try:
+                        # Get current recursion tracking dict from context (isolated per request)
+                        current_depth = _middleware_depth_context.get().copy()
+
+                        # Check for middleware recursion
+                        if effective_middlewares and current_depth.get(req.url, 0) >= 10:
+                            raise RuntimeError(
+                                f"Recursive middleware detected for {req.url}. "
+                                "Middleware is calling itself with the same URL."
+                            )
+
+                        # Update depth tracking in a new context
+                        current_depth[req.url] = current_depth.get(req.url, 0) + 1
+                        token = _middleware_depth_context.set(current_depth)
+
                         try:
                             resp = await handler(req)
+                        finally:
+                            # Reset context
+                            _middleware_depth_context.reset(token)
                     # Client connector errors should not be retried
                     except (
```

aiohttp/client.py
bdraco: Feel free to push to this. Definitely would like to clean it up to the point we can cache the constructor for the session based ones and save it in the session 

Since they are stateful I wanted to avoid using an LRU

tests/test_client_middleware.py
Dreamsorcerer: My other comment said we'd need a new property. `request._session` currently exists.

examples/client_middleware_auth.py
Dreamsorcerer: https://github.com/aio-libs/aiohttp/pull/9732/commits/4417d9c8f8a4e5fdb07f6bb4bfdf7e4260d9e13c

aiohttp/client.py
Dreamsorcerer: Hmm, this would break subclasses of ClientRequest, so I think this would need to be a v4 optimisation. The change might look like:
```diff
diff --git a/aiohttp/client.py b/aiohttp/client.py
index 30a29c36c..b9b31ee6a 100644
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -622,9 +622,23 @@ class ClientSession:
                                 get_env_proxy_for_url, url
                             )
 
+                    response_params = {
+                        "timer": timer,
+                        "skip_payload": method in EMPTY_BODY_METHODS,
+                        "read_until_eof": read_until_eof,
+                        "auto_decompress": auto_decompress,
+                        "read_timeout": real_timeout.sock_read,
+                        "read_bufsize": read_bufsize,
+                        "timeout_ceil_threshold": self._connector._timeout_ceil_threshold,
+                        "max_line_size": max_line_size,
+                        "max_field_size": max_field_size,
+                    }
+
                     req = self._request_class(
                         method,
                         url,
+                        response_params=response_params,
+                        timeout=real_timeout,
                         params=params,
                         headers=headers,
                         skip_auto_headers=skip_headers,
@@ -656,7 +670,7 @@ class ClientSession:
                         assert self._connector is not None
                         try:
                             conn = await self._connector.connect(
-                                req, traces=traces, timeout=real_timeout
+                                req, traces=traces, timeout=req._timeout
                             )
                         except asyncio.TimeoutError as exc:
                             raise ConnectionTimeoutError(
@@ -664,17 +678,7 @@ class ClientSession:
                             ) from exc
 
                         assert conn.protocol is not None
-                        conn.protocol.set_response_params(
-                            timer=timer,
-                            skip_payload=req.method in EMPTY_BODY_METHODS,
-                            read_until_eof=read_until_eof,
-                            auto_decompress=auto_decompress,
-                            read_timeout=real_timeout.sock_read,
-                            read_bufsize=read_bufsize,
-                            timeout_ceil_threshold=self._connector._timeout_ceil_threshold,
-                            max_line_size=max_line_size,
-                            max_field_size=max_field_size,
-                        )
+                        conn.protocol.set_response_params(**req._response_params)
                         try:
                             resp = await req.send(conn)
                             try:
diff --git a/aiohttp/client_reqrep.py b/aiohttp/client_reqrep.py
index f0d4c0199..669185086 100644
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -22,6 +22,7 @@ from typing import (
     Optional,
     Tuple,
     Type,
+    TypedDict,
     Union,
 )
 
@@ -190,6 +191,18 @@ class ConnectionKey(NamedTuple):
     proxy_headers_hash: Optional[int]  # hash(CIMultiDict)
 
 
+class _ResponseParams(TypedDict):
+    timer: Optional[BaseTimerContext]
+    skip_payload: bool
+    read_until_eof: bool
+    auto_decompress: bool
+    read_timeout: Optional[float]
+    read_bufsize: int
+    timeout_ceil_threshold: float
+    max_line_size: int
+    max_field_size: int
+
+
 class ClientRequest:
     GET_METHODS = {
         hdrs.METH_GET,
@@ -225,6 +238,8 @@ class ClientRequest:
         method: str,
         url: URL,
         *,
+        response_params: _ResponseParams,
+        timeout: ClientTimeout,
         params: Query = None,
         headers: Optional[LooseHeaders] = None,
         skip_auto_headers: Optional[Iterable[str]] = None,
@@ -276,6 +291,8 @@ class ClientRequest:
         self.response_class: Type[ClientResponse] = real_response_class
         self._timer = timer if timer is not None else TimerNoop()
         self._ssl = ssl
+        self._response_params = response_params
+        self._timeout = timeout
         self.server_hostname = server_hostname
 
         if loop.get_debug():
```

aiohttp/client.py
```diff
@@ -642,32 +648,36 @@ async def _request(
...
+                            )
+                        except asyncio.TimeoutError as exc:
+                            raise ConnectionTimeoutError(
+                                f"Connection timeout to host {_url}"
+                            ) from exc
+
+                        assert conn.protocol is not None
+                        conn.protocol.set_response_params(
+                            timer=timer,
+                            skip_payload=_method in EMPTY_BODY_METHODS,
```
Dreamsorcerer: We can still simplify this part though:
```suggestion
                    async def _connect_and_send_request(
                        req: ClientRequest
                    ) -> ClientResponse:
                        # connection timeout
                        assert self._connector is not None
                        try:
                            conn = await self._connector.connect(
                                req, traces=traces, timeout=real_timeout
                            )
                        except asyncio.TimeoutError as exc:
                            raise ConnectionTimeoutError(
                                f"Connection timeout to host {req.url}"
                            ) from exc

                        assert conn.protocol is not None
                        conn.protocol.set_response_params(
                            timer=timer,
                            skip_payload=req.method in EMPTY_BODY_METHODS,
```

tests/test_client_middleware.py
Dreamsorcerer: Added in https://github.com/aio-libs/aiohttp/pull/9732/commits/4417d9c8f8a4e5fdb07f6bb4bfdf7e4260d9e13c
So that example should now work.

Dreamsorcerer: > maybe something like this.... but seems kinda horrible

I just meant a test to validate that using the session within a middleware continues to work (i.e. just as a regression test). I think we can just rely on normal Python stack limits etc. to error on accidental recursion.

Dreamsorcerer:
I'm generally happy with this now. Let me know when it's out of draft and I'll do a final review.

bdraco:
I tried to optimize this but the overhead of creating closures is minimal so caching doesn't make sense here because client_middleware is already designed to minimize the overhead.  Let me address the remaining outstanding review comments.

tests/test_client_middleware.py
bdraco: Pushed tests for this case.

Writing docs to add warnings about calling a session with client middleware inside the middleware

bdraco: added warnings to the docs

examples/client_middleware_auth.py
bdraco: Seems to work just fine..

bdraco: I added a test for this as well

bdraco:
Since we're generally happy with the design now (as discussed on Matrix), I'll go ahead and merge this to unblock the follow-up work. Any further adjustments can be handled in a separate PR.

aiohttp/client.py
```diff
@@ -642,32 +648,33 @@ async def _request(
                         trust_env=self.trust_env,
                     )
 
-                    # connection timeout
-                    try:
-                        conn = await self._connector.connect(
-                            req, traces=traces, timeout=real_timeout
+                    # Core request handler - now includes connection logic
```
Dreamsorcerer, line 651: Can probably be removed (second part sounds like it's comparing to an earlier commit).

aiohttp/client.py
```diff
@@ -678,6 +685,30 @@ async def _request(
...
+                    try:
+                        resp = await handler(req)
+                    # Client connector errors should not be retried
+                    except (
+                        ConnectionTimeoutError,
+                        ClientConnectorError,
+                        ClientConnectorCertificateError,
+                        ClientConnectorSSLError,
+                    ):
+                        raise
```
Dreamsorcerer, line 711: Is this related to middlewares? Are we sure we don't want to retry on these?

aiohttp/client.py
```diff
@@ -678,6 +685,30 @@ async def _request(
...
+                    effective_middlewares = (
+                        self._middlewares if middlewares is None else middlewares
+                    )
+
+                    if effective_middlewares:
+                        handler = build_client_middlewares(
+                            _connect_and_send_request, effective_middlewares
+                        )
+                    else:
+                        handler = _connect_and_send_request
```
Dreamsorcerer, line 700: This case appears to already be handled in build_client_middlewares(), so we can drop the condition here.

aiohttp/client_middlewares.py
```diff
@@ -0,0 +1,58 @@
...
+
+# Type for client middleware - similar to server but uses ClientRequest/ClientResponse
+ClientMiddlewareType = Callable[
+    [ClientRequest, ClientHandlerType], Awaitable[ClientResponse]
+]
+
+
+def build_client_middlewares(
+    handler: ClientHandlerType,
+    middlewares: tuple[ClientMiddlewareType, ...],
```
Dreamsorcerer, line 20: Strictly speaking, Sequence is plenty, there's no requirement for a tuple.

aiohttp/client_middlewares.py
```diff
@@ -0,0 +1,58 @@
...
+    Apply middlewares to request handler.
+
+    The middlewares are applied in reverse order, so the first middleware
+    in the list wraps all subsequent middlewares and the handler.
+
+    This implementation avoids using partial/update_wrapper to minimize overhead
+    and doesn't cache to avoid holding references to stateful middleware.
+    """
+    if not middlewares:
+        return handler
```
Dreamsorcerer, line 32: I think this check is actually irrelevant, unless we allow None. If there's no middlewares then the for loop below will not iterate and it'll immediately return the handler anyway (I guess there'd be a `reversed(())` call overhead though).

aiohttp/client.py
```diff
@@ -450,6 +455,7 @@ async def _request(
         auto_decompress: Optional[bool] = None,
         max_line_size: Optional[int] = None,
         max_field_size: Optional[int] = None,
+        middlewares: Optional[Tuple[ClientMiddlewareType, ...]] = None,
```
Dreamsorcerer, line 458: It might simplify the code checks and optimise slightly if we disallow None here and just default to `()`.

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
    Started keeping the ``Authorization`` header during HTTP → HTTPS
    redirects when the host remains the same.
 
+.. _aiohttp-client-middleware:
+
+Client Middleware
+-----------------
+
+aiohttp client supports middleware to intercept requests and responses. This can be
```
Dreamsorcerer, line 113: *The client

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
...
+        # Process request before sending
+        _LOGGER.debug(f"Request: {request.method} {request.url}")
+
+        # Call the next handler
+        response = await handler(request)
+
+        # Process response after receiving
+        _LOGGER.debug(f"Response: {response.status}")
+
+        return response
```
Dreamsorcerer, line 138: I feel like we should use a different example to start with. This might set up some bad practices, given that we already have an access logger. We should probably have a dedicated example/cookbook for a logging middleware.

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
...
+
+Here's a simple example showing request modification::
+
+    async def add_api_key_middleware(
+        request: ClientRequest,
+        handler: ClientHandlerType
+    ) -> ClientResponse:
+        # Add API key to all requests
+        request.headers['X-API-Key'] = 'my-secret-key'
+        return await handler(request)
```
Dreamsorcerer, line 161: I think maybe move this one to replace that initial one above and remove this section (although maybe this is another bad example, as this would be better done by setting the session headers parameter...).

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
...
+This pattern is particularly useful for:
+
+- Refreshing authentication tokens after a 401 response
+- Switching to fallback servers or authentication methods
+- Adding or modifying headers based on error responses
+- Implementing back-off strategies with increasing delays
+
+The middleware can maintain state between retries to track which strategies have been tried and modify the request accordingly for the next attempt.
+
+Example: Retrying requests with middleware
```
Dreamsorcerer, line 184: We probably want an example of refreshing an access token in response to a 401 as well. What do you think about creating a separate cookbook page or something? As we'll probably want to document several kinds of common middlewares (retry, couple of auth ones, logger one, ssrf etc.).

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
...
+    async def auth_middleware(
+        request: ClientRequest,
+        handler: ClientHandlerType
+    ) -> ClientResponse:
+        request.headers['Authorization'] = 'Bearer token123'
+        return await handler(request)
+
+    # Middlewares are applied in order: logging -> auth -> request
+    async with ClientSession(middlewares=(logging_middleware, auth_middleware)) as session:
+        resp = await session.get('http://example.com')
```
Dreamsorcerer, line 253: I've thought about explaining this before, and I wonder if it's worth putting something in that shows what the code looks like flattened, so people can see that it is simply nested.
e.g. (without me thinking of anything useful to put here):
```
async def middleware1():
    print(1)
    resp = await handler(request)
    print(2)
    return resp

async def middleware2():
    print(3)
    resp = await handler(request)
    print(4)
    return resp
```
is equivalent to:
```
# Start middleware 1
print(1)
# -> Start middleware 2
print(3)
resp = handler(request)
print(4)
# -> Return from middleware 2
print(2)
# Return from middleware 1
```

It's important that users understand that nesting when it comes to multiple middlewares handling exceptions etc. Do you think we should try to teach something like that here (though maybe it's more relevant to server middlewares than client...)?

docs/client_advanced.rst
```diff
@@ -105,6 +105,218 @@ background.
...
+      Callable[
+          [ClientRequest, ClientHandlerType],
+          Awaitable[ClientResponse]
+      ]
+
+.. type:: ClientHandlerType
+
+   Type alias for client request handler functions::
+
+      Callable[ClientRequest, Awaitable[ClientResponse]]
```
Dreamsorcerer, line 318: Should this be in reference instead?

Dreamsorcerer, line 318: *`[ClientRequest]`

docs/client_reference.rst
```diff
@@ -53,6 +53,7 @@ The client session supports the context manager protocol for self closing.
                          trust_env=False, \
                          requote_redirect_url=True, \
                          trace_configs=None, \
+                         middlewares=None, \
```
Dreamsorcerer, line 56: If we change the default, need to update here too: ()

docs/client_reference.rst
```diff
@@ -213,6 +214,13 @@ The client session supports the context manager protocol for self closing.
                          disabling.  See :ref:`aiohttp-client-tracing-reference` for
                          more information.
 
+   :param middlewares: A tuple of middleware instances to apply to all session requests.
+                      Each middleware must match the :type:`ClientMiddlewareType` signature.
+                      ``None`` (default) is used when no middleware is needed.
```
Dreamsorcerer, line 219: And this line removed.

Dreamsorcerer: Same 2 changes for request() below.

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+    async def retry_middleware(
+        request: ClientRequest, handler: ClientHandlerType
+    ) -> ClientResponse:
+        retry_count = 0
+        while True:
+            response = await handler(request)
+            if response.status == 503 and retry_count < 1:
+                retry_count += 1
+                continue
+            return response
```
Dreamsorcerer, line 83: Could probably do this cleaner without the while True:
```
for retry_count in range(2):
    response = await handler(request)
    if response.ok:
        return response
return response
```

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+
+        assert False, "Should not reach here - invalid auth scenario"
+
+    async def challenge_auth_middleware(
+        request: ClientRequest, handler: ClientHandlerType
+    ) -> ClientResponse:
+        challenge_data: Dict[str, Union[bool, str, None]] = {
+            "nonce": None,
+            "attempted": False,
+        }
```
Dreamsorcerer, line 250: I'd probably just make these 2 variables at this point.

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+                request.headers["Authorization"] = (
+                    f'Custom response="{challenge_data["nonce"]}-secret"'
+                )
+
+            response = await handler(request)
+
+            # If we get a 401 with challenge, store it and retry
+            if response.status == 401 and not challenge_data["attempted"]:
+                www_auth = response.headers.get("WWW-Authenticate")
+                if www_auth and "nonce=" in www_auth:  # pragma: no branch
```
Dreamsorcerer, line 264: Should we just assert, rather than using a pragma? i.e. If this does branch, it suggests something has gone wrong with the test, so an assert probably makes sense.

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+            if response.status == 401:
+                auth_step = response.headers.get("X-Auth-Step")
+
+                if auth_step == "1":
+                    # First step: store session token
+                    middleware_state["session"] = response.headers.get("X-Session")
+                    middleware_state["step"] = 1
+                    continue
+
+                elif auth_step == "2":  # pragma: no branch
```
Dreamsorcerer, line 350: Same here?

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+            request.headers["X-Auth-Token"] = str(token_state["token"])
+
+            response = await handler(request)
+
+            # Check if token needs refresh
+            if response.status == 401 and not token_state["refreshed"]:
+                data = await response.json()
+                if data.get("error") == "token_expired" and data.get(
+                    "refresh_required"
+                ):  # pragma: no branch
```
Dreamsorcerer, line 410: assert?

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+
+        if auth_token == "refreshed-token":
+            return web.json_response({"data": "success"})
+
+        assert False, "Should not reach here - invalid token refresh flow"
+
+    async def token_refresh_middleware(
+        request: ClientRequest, handler: ClientHandlerType
+    ) -> ClientResponse:
+        while True:
```
Dreamsorcerer, line 399: If this goes wrong, it's an infinite loop. Maybe safer to implement something with `for _ in range(2):` to keep it bounded?

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+
+async def test_client_middleware_stateful_retry(aiohttp_server: AiohttpServer) -> None:
+    """Test retry middleware using class with state management."""
+
+    class RetryMiddleware:
+        """Middleware that retries failed requests with backoff."""
+
+        def __init__(self, max_retries: int = 3) -> None:
+            self.max_retries = max_retries
+            self.retry_counts: Dict[int, int] = {}  # Track retries per request
```
Dreamsorcerer, line 493: This isn't used? I think in it's current state, this is a duplicate of a test above. I assume it was intended to use this though in some way (I guess this could track something over multiple requests).

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+        async with session.get(server.make_url("/")) as resp:
+            assert resp.status == 200
+            text = await resp.text()
+            assert text == "Auth: Bearer session-token"
+            assert session_middleware_called is True
+            assert request_middleware_called is False
+
+        # Reset flags
+        session_middleware_called = False
+        request_middleware_called = False
```
Dreamsorcerer, line 626: Maybe parametrize this test to avoid the need for these resets.

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+        with pytest.raises(BlockedByMiddleware) as exc_info:
+            await session.get("https://evil.com/path")
+
+        assert "Connection to evil.com is blocked" in str(exc_info.value)
+
+    # Verify that connections were attempted in the correct order
+    assert len(connection_attempts) == 3
+    assert allowed_url.host and allowed_url.host in connection_attempts[0]
+    assert "blocked.example.com" in connection_attempts[1]
+    assert "evil.com" in connection_attempts[2]
```
Dreamsorcerer, line 750: I think you could just do:
```
expected_attempts = (...)
assert connection_attempts == expected_attempts
```

tests/test_client_middleware.py
```diff
@@ -0,0 +1,1116 @@
...
+            self.secretkey = secretkey
+
+        def get_hash(self, request: ClientRequest) -> str:
+            if request.body:
+                data = request.body.decode("utf-8")
+            else:
+                data = "{}"
+
+            # Simulate authentication hash without using real crypto
+            signature = f"SIGNATURE-{self.secretkey}-{len(data)}-{data[:10]}"
```
Dreamsorcerer, line 1044: Just `return` without variable.

bdraco:
Thanks for the review. I'll clean this up between sprint work.

aiohttp/client.py
bdraco: The connection now happens inside the try block instead of before so we need to re-raise to preserve previous behavior

aiohttp/client.py
bdraco: 0a741b908 (in branch middleware_cleanup, PR not yet open)

aiohttp/client.py
bdraco: This is the no middleware case, we need to keep this

Dreamsorcerer: No, I mean if you call `build_client_middlewares(handler, ())`, the result is the same.

Dreamsorcerer: That function will return the handler unchanged.

aiohttp/client_middlewares.py
bdraco: 308a2400a (in branch middleware_cleanup, PR not yet open)

aiohttp/client.py
bdraco: I prefer the `None` check to creating an empty tuple

aiohttp/client_middlewares.py
bdraco: its already guarded in                     if effective_middlewares:
                        handler = build_client_middlewares(
                            _connect_and_send_request, effective_middlewares
                        )
So we can take it out

bdraco: 5c64626b6 (in branch middleware_cleanup, PR not yet open)

docs/client_advanced.rst
bdraco: 57597151a

docs/client_reference.rst
bdraco: Leave this as-is per my comment above

docs/client_reference.rst
bdraco: Leaving this as-is per my comment above.

aiohttp/client.py
Dreamsorcerer: > I prefer the `None` check to creating an empty tuple

My point further down is that it can allow us to have _no_ check if None is not allowed. i.e. we can just do `for m in middlewares:` without a check and know that the body won't execute if there are no middlewares.

tests/test_client_middleware.py
bdraco: We probably should make it really clear this is risk in the docs as well

docs/client_advanced.rst
bdraco: Removed logging example f9f8a0dd6

I had gone with something simple but I agree logging is not the best use case.

docs/client_advanced.rst
bdraco: removed in b8a5ecf63

docs/client_advanced.rst
bdraco: I think we will likely want to ship some common patterns. A cookbook is a good place to start to see what the use cases are. I expect there will be just a few very common ones

docs/client_advanced.rst
bdraco: moved in 1bffcd488

bdraco: fixed in 1bffcd488

tests/test_client_middleware.py
bdraco: changed in 91c8e8736

tests/test_client_middleware.py
bdraco: Its stale. Removed

aiohttp/client.py
bdraco: How should we handle the case were we want to override the middleware in the client request to disable it if the default is also a `()`?

tests/test_client_middleware.py
bdraco: I split it into two tests. It cleans it up a bit ce84abd53

tests/test_client_middleware.py
bdraco: oops, leftover cruft. fixed in d9ea1ca43

tests/test_client_middleware.py
bdraco: It used to be longer at one point. Yeah that makes sense to cleanup now

bdraco: cbe674710

tests/test_client_middleware.py
bdraco: b47aed9c8

tests/test_client_middleware.py
bdraco: 7bb620667

tests/test_client_middleware.py
bdraco: 3cad8280c

tests/test_client_middleware.py
bdraco: 3cad8280c

tests/test_client_middleware.py
bdraco: 3cad8280c

docs/client_advanced.rst
bdraco: attempt to explain in a12905883

docs/client_advanced.rst
bdraco: https://github.com/aio-libs/aiohttp/issues/10903

aiohttp/client.py
bdraco: captured in https://github.com/aio-libs/aiohttp/issues/10905 as GitHub is making it hard to find this with all the comments on this PR.

PR #9741, opened 2024-11-09
Fix race getting an unused port when there are multiple test runners

bdraco:
The `aiohttp_used_port` fixture finds a free port but it does not hold the socket so there is a race between when checking that the port is available, and when it actually gets used. When we run tests with xdist we would sometimes get a failure because multiple tests would try to use the same port

One step closer to unblocking https://github.com/aio-libs/aiohttp/pull/5431

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -212,7 +212,7 @@ jobs:
         PIP_USER: 1
       run: >-
         PATH="${HOME}/Library/Python/3.11/bin:${HOME}/.local/bin:${PATH}"
-        pytest --junitxml=junit.xml
+        pytest -vvvs --junitxml=junit.xml
       shell: bash
     - name: Re-run the failing tests with maximum verbosity
       if: failure()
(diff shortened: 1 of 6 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -212,7 +212,7 @@ jobs:
         PIP_USER: 1
       run: >-
         PATH="${HOME}/Library/Python/3.11/bin:${HOME}/.local/bin:${PATH}"
-        pytest --junitxml=junit.xml
+        pytest -vvvs --junitxml=junit.xml
```
bdraco: TODO: revert this before merging

tests/test_run_app.py
```diff
@@ -1030,14 +1048,14 @@ def __init__(self, *args: Any, **kwargs: Any):
...
-            await asyncio.sleep(0.5)
+            await asyncio.sleep(1)
             async with ClientSession() as sess:
                 for _ in range(5):  # pragma: no cover
                     try:
                         with pytest.raises(asyncio.TimeoutError):
                             async with sess.get(
                                 f"http://localhost:{port}/",
-                                timeout=ClientTimeout(total=0.1),
+                                timeout=ClientTimeout(total=0.2),
```
bdraco:
```suggestion
                                timeout=ClientTimeout(total=0.1),
```

PR #9755, opened 2024-11-09
DNM: Attempt to cache ConnectionKey hash

bdraco:


bdraco:
Ends up being slower because of the update_wrapper....

was worth a shot

PR #9757, opened 2024-11-09
Avoid creating body writer task when there is no body

bdraco:
The client would create a task to write the body even if the body was `b""`

Since `GET` requests tend to be some of the most common, it makes sense to optimize for the case where only headers are sent.

aiohttp 3.11.0b4 yarl 1.17.1
aiohttp_3_11_0b4_yarl_1_17_1

This PR:
b5

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -681,7 +681,7 @@ class ClientRequest:
             self.body
             or self._continue is not None
             or self.compress
-            or writer.chunked
+            or self.chunked
             or protocol.paused
         ):
             coro = self.write_bytes(writer, conn)
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -677,22 +677,29 @@ class ClientRequest:
         v = self.version
         status_line = f"{self.method} {path} HTTP/{v.major}.{v.minor}"
         await writer.write_headers(status_line, self.headers)
-        coro = self.write_bytes(writer, conn)
+        if self.body is not None or self._continue is not None or protocol.paused:
+            coro = self.write_bytes(writer, conn)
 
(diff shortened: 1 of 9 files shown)


tests/test_benchmarks_client_request.py
```diff
@@ -89,6 +89,10 @@ class MockProtocol(asyncio.BaseProtocol):
         def __init__(self) -> None:
             self.transport = MockTransport()
 
+        @property
+        def paused(self) -> bool:
```
bdraco:
```suggestion
        def writing_paused(self) -> bool:
```

aiohttp/client_reqrep.py
```diff
@@ -677,22 +677,30 @@ async def send(self, conn: "Connection") -> "ClientResponse":
...
+            if task.done():
+                task = None
+            else:
+                self._writer = task
         else:
-            self._writer = task
-
+            # We have nothing to write because
+            # - there is no body
+            # - the protocol is not paused
```
```suggestion
            # - the protocol does not have writing paused
```

tests/test_client_session.py
```diff
@@ -918,7 +918,7 @@ def to_url(path: str) -> URL:
             assert to_trace_urls(on_request_redirect) == []
             assert to_trace_urls(on_request_end) == [to_url("/?x=0")]
             assert to_trace_urls(on_request_exception) == []
-            assert to_trace_urls(on_request_chunk_sent) == [to_url("/?x=0")]
```
bdraco, line 921: This seems like a bug that we sent this trace when there was nothing actually sent so I didn't preserve this behavior

bdraco:
This might have introduced a regression on sending an empty body with chunked.. although we likely have a guard to not do chunked with empty body I'm not sure its tested

bdraco:
nope, seems ok but will add a test

PR #9762, opened 2024-11-10
Refactor payload registry to avoid linear searches for common types

### LINKED ISSUE #9419: Payload registry has to do a linear search to find payloads
https://github.com/aio-libs/aiohttp/blob/21f5f92a755dc5ac7225b5e76f561553cf86565e/aiohttp/payload.py#L97

There is a note that its inefficent.

We might be able to clean it up by giving each payload a key as part of the base class and then doing a dict lookup instead

bdraco:
closes #9419

Looks to be a ~2.4% speed up so probably not worth mentioning in a changelog message

Screenshot 2024-11-10 at 9 57 48 AM
Screenshot 2024-11-10 at 9 57 54 AM
Screenshot 2024-11-10 at 9 58 09 AM
Screenshot 2024-11-10 at 9 58 15 AM

### DIFF
--- a/aiohttp/payload.py
+++ b/aiohttp/payload.py
@@ -114,7 +114,7 @@ class PayloadRegistry:
             return data
         if self._first:
             for factory, type_ in self._first:
-                if isinstance(data, type):
+                if isinstance(data, type_):
                     return factory(data, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


bdraco:
Will add a changelog message if its substantial

PR #9771, opened 2024-11-10
Small cleanups to enabling compression in web_response

bdraco:
- Return early if identity to save some indent
- Use identity checks for enums since they are singletons
- Remove assertions that will always be True
- Add missing coverage

### DIFF
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -344,14 +344,14 @@ class StreamResponse(BaseClass, HeadersMixin, CookieMixin):
     async def _start_compression(self, request: "BaseRequest") -> None:
         if self._compression_force:
             await self._do_start_compression(self._compression_force)
-        else:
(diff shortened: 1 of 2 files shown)


aiohttp/web_response.py
```diff
@@ -705,30 +706,28 @@ async def _start(self, request: "BaseRequest") -> AbstractStreamWriter:
...
+        # Instead of using _payload_writer.enable_compression,
+        # compress the whole body
+        compressor = ZLibCompressor(
+            encoding=coding.value,
+            max_sync_chunk_size=self._zlib_executor_size,
+            executor=self._zlib_executor,
+        )
+        assert self._body is not None
+        if self._zlib_executor_size is None and len(self._body) > 1024 * 1024:
+            warnings.warn(
```
bdraco: Looks like we are missing coverage for this warning

PR #9780, opened 2024-11-10
Replace get_event_loop with get_running_loop in the compressor

bdraco:
## What do these changes do?

Replace `get_event_loop` with `get_running_loop` in the compressor.  This should never be called without a running event loop as its in an async function.

`asyncio.get_event_loop`:
> Deprecated since version 3.12: Deprecation warning is emitted if there is no current event loop. In some future Python release this will become an error

There was no need to call the sync wrapper function since we are already inside the object and can call the compressor/decompressor directly.

Adds missing test coverage

no

no

### DIFF
--- a/aiohttp/compression_utils.py
+++ b/aiohttp/compression_utils.py
@@ -79,8 +79,8 @@ class ZLibCompressor(ZlibBaseHandler):
                 self._max_sync_chunk_size is not None
                 and len(data) > self._max_sync_chunk_size
             ):
-                return await asyncio.get_event_loop().run_in_executor(
-                    self._executor, self.compress_sync, data
+                return await asyncio.get_running_loop().run_in_executor(
+                    self._executor, self._compressor.compress, data
                 )
             return self.compress_sync(data)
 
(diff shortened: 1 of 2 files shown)


bdraco:
Looks like we have some missing coverage here

aiohttp/compression_utils.py
```diff
@@ -111,8 +111,8 @@ async def decompress(self, data: bytes, max_length: int = 0) -> bytes:
             self._max_sync_chunk_size is not None
             and len(data) > self._max_sync_chunk_size
         ):
-            return await asyncio.get_event_loop().run_in_executor(
-                self._executor, self.decompress_sync, data, max_length
+            return await asyncio.get_running_loop().run_in_executor(
```
bdraco, line 128: This code looks like its unreachable because we never create ZLibDecompressor with max_sync_chunk_size

aiohttp/compression_utils.py
```diff
@@ -79,8 +87,8 @@ async def compress(self, data: bytes) -> bytes:
                 self._max_sync_chunk_size is not None
                 and len(data) > self._max_sync_chunk_size
             ):
-                return await asyncio.get_event_loop().run_in_executor(
-                    self._executor, self.compress_sync, data
+                return await asyncio.get_running_loop().run_in_executor(
```
Dreamsorcerer, line 90: Should we replace these low-level calls with asyncio.to_thread() now that we don't support 3.8?

PR #9783, opened 2024-11-10
Avoid pack overhead for large websocket messages

bdraco:
TODO: needs a test to unpack 2^63 as the length of the message


bdraco:
We have no benchmark for messages over 2**16 so we won't see any difference here until we add them

bdraco:
While it looks a bit faster it makes no overall difference because so much time is spent in memcpy for a message of this size

Screenshot 2024-11-10 at 10 10 32 PM

This isn't worth doing unless we can get rid of the memcpy first

bdraco:
Screenshot 2024-11-10 at 10 15 33 PM

bdraco:
Conclusion is we can't get rid of the memcpy so this doesn't matter

PR #9790, opened 2024-11-10
DNM: Attempt to reduce one copy in the WebSocket reader

bdraco:
The memcpy is 99% of the profile run time for large messages

### DIFF
--- a/aiohttp/_websocket/reader_py.py
+++ b/aiohttp/_websocket/reader_py.py
@@ -241,6 +241,8 @@ class WebSocketReader:
                         self._decompressobj = ZLibDecompressor(
                             suppress_deflate_header=True
                         )
+                    if type(assembled_payload) is memoryview:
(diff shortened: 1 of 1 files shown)


aiohttp/_websocket/reader_py.py
```diff
@@ -319,10 +321,14 @@ def _feed_data(self, data: bytes) -> None:
 
     def parse_frame(
         self, buf: bytes
-    ) -> List[Tuple[bool, Optional[int], Union[bytes, bytearray], Optional[bool]]]:
+    ) -> List[
+        Tuple[bool, Optional[int], Union[bytes, bytearray, memoryview], Optional[bool]]
```
Dreamsorcerer, line 325: Maybe worth making an alias of something here. The union atleast appears to be repeated in several places.

bdraco:
All it did was move the overhead out of that function and make it a bit slower for the small case

PR #9793, opened 2024-11-11
[PR #9685/5241897 backport][3.11] Fix WebSocket reader flow control calculations

bdraco:
Co-authored-by: pre-commit-ci[bot] <66853113+pre-commit-ci[bot]@users.noreply.github.com>
(cherry picked from commit 5241897f6459926028803caaa4782b8ba84f0562)

Note that the code that changed WebSocket messages to be a Union of different types has been stripped out, and the size argument/internal tuple to `feed_data` has been retained for compatibility with 3.11

### DIFF
--- a/aiohttp/__init__.py
+++ b/aiohttp/__init__.py
@@ -93,7 +93,6 @@ from .streams import (
     EMPTY_PAYLOAD as EMPTY_PAYLOAD,
     DataQueue as DataQueue,
     EofStream as EofStream,
-    FlowControlDataQueue as FlowControlDataQueue,
     StreamReader as StreamReader,
 )
 from .tracing import (
@@ -216,7 +215,6 @@ __all__: Tuple[str, ...] = (
     "DataQueue",
     "EMPTY_PAYLOAD",
     "EofStream",
-    "FlowControlDataQueue",
     "StreamReader",
     # tracing
     "TraceConfig",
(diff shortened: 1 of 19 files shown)


aiohttp/_websocket/reader_py.py
```diff
@@ -187,17 +264,18 @@ def _feed_data(self, data: bytes) -> None:
...
                 else:
-                    self._queue_feed_data(
+                    self.queue.feed_data(
                         TUPLE_NEW(WSMessage, (WS_MSG_TYPE_BINARY, payload_merged, "")),
                         len(payload_merged),
                     )
             elif opcode == OP_CODE_CLOSE:
-                if len(payload) >= 2:
+                payload_len = len(payload)
+                if payload_len >= 2:
```
```suggestion
                if len(payload) >= 2:
```

aiohttp/__init__.py
```diff
@@ -93,7 +93,6 @@
     EMPTY_PAYLOAD as EMPTY_PAYLOAD,
     DataQueue as DataQueue,
     EofStream as EofStream,
-    FlowControlDataQueue as FlowControlDataQueue,
```
bdraco, line 96: I couldn't find anything on GitHub code searching using this externally, and I don't think it would make too much sense to do so since it was only used for the WebSocket anyways.

bdraco:
Pushed to Home Assistant production systems. Everything tests out OK.  We should still do an rc1 so we can a full CI run though before moving to .0 release

PR #9795, opened 2024-11-11
Switch WebSocketWriter to use `removesuffix` to drop compression trailer

bdraco:
Now that Python 3.8 is no longer supported (`master` and `3.11` only) we can use `bytes.removesuffix`

This also takes care of one uncovered branch in the tests and brings the `WebSocketWriter` up to 100% line/branch coverage

Its small improvement, considering how much time the compression actually takes. It will be >10% if using `isal` for compression though since the compression overhead is far less when using alternative `zlib` see https://github.com/aio-libs/aiohttp/issues/9798
Screenshot 2024-11-11 at 8 31 43 AM

### DIFF
--- a/aiohttp/_websocket/writer.py
+++ b/aiohttp/_websocket/writer.py
@@ -92,16 +92,16 @@ class WebSocketWriter:
                     self._compressobj = self._make_compress_obj(self.compress)
                 compressobj = self._compressobj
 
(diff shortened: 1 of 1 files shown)


bdraco:
We have no benchmark for compression.

We need to add that before considering this change

PR #9796, opened 2024-11-11
Avoid calling the drain helper in `WebSocketWriter` if the protocol is not paused

bdraco:
We already know its connected or it would have raised for `is_closing` so we can guard calling the coro with a check to see if its paused

### DIFF
--- a/aiohttp/_websocket/writer.py
+++ b/aiohttp/_websocket/writer.py
@@ -155,7 +155,8 @@ class WebSocketWriter:
         # if the writer is not paused.
         if self._output_size > self._limit:
             self._output_size = 0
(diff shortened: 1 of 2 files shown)


aiohttp/_websocket/writer.py
```diff
@@ -155,7 +155,8 @@ async def send_frame(
         # if the writer is not paused.
         if self._output_size > self._limit:
             self._output_size = 0
-            await self.protocol._drain_helper()
+            if self.protocol.writing_paused:
```
bdraco:
```suggestion
            if self.protocol._paused:
```

PR #9800, opened 2024-11-11
Fix incorrect typing on `WebSocketResponse._handshake`

bdraco:
The compress type is an `int` not a `bool`. The protocol type can be `None`


aiohttp/web_ws.py
```diff
@@ -386,7 +386,7 @@ def ws_protocol(self) -> Optional[str]:
         return self._ws_protocol
 
     @property
-    def compress(self) -> bool:
+    def compress(self) -> Union[int, bool]:
```
bdraco, line 391: This is a bit strange but the type is correct as its `bool` until it connects and the handshake happens, than it changes to an `int`.

Would be nice to change that in 4.x but I didn't want to adjust it because it would be a breaking change to do so

PR #9803, opened 2024-11-11
Small performance improvement to drain

bdraco:
same as #9796 but for the http writer

Screenshot 2024-11-11 at 8 36 55 AM
Screenshot 2024-11-11 at 8 37 00 AM


bdraco:
not significant enough for a changelog message, but still a nice speed up for post requests

PR #9809, opened 2024-11-11
Add `__slots__` to `UrlMappingMatchInfo`

bdraco:
We churn these objects on every request, this will save a bit of RAM.

I was a bit surprised that the benchmark shows it makes a measurable difference in `__init__`
Screenshot 2024-11-11 at 9 33 53 AM
Screenshot 2024-11-11 at 9 33 49 AM

and `add_app`
Screenshot 2024-11-11 at 9 39 20 AM

### DIFF
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -59,6 +59,9 @@ class AbstractRouter(ABC):
 
 
 class AbstractMatchInfo(ABC):
+
+    __slots__ = ()
+
     @property  # pragma: no branch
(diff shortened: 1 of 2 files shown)


asvetlov:
Interesting!

At the time of adding `UrlMappingMatchInfo` (Python 3.4-3.5 IIRC) `__slots__` where slightly slower that `__dict__` based lookup.
For Python 3.9 it was still true, later I didn't measure.

PR #9810, opened 2024-11-11
Small performance improvement to Cython `HttpParser` `process_header`

bdraco:
- Avoid a `str` check on each header
- Avoid looking up the `add` method on every header


bdraco:
Need to look at the benchmark change to see if this is even worth changing... will take a peek after the run happens

bdraco:
makes no difference

PR #9812, opened 2024-11-11
Add `__slots__` to `PayloadRegistry`

bdraco:
I was a bit surprised the see the performance improvement in https://github.com/aio-libs/aiohttp/pull/9809 so I'm curious to see if we can observe any improvement here as well


bdraco:
In this case, it makes no difference because it's really the payloads themselves that could use slots

bdraco:
Still doesn't hurt to add this anyway

PR #9815, opened 2024-11-11
Remove unneeded assignment in BaseConnector._release

bdraco:
Not a functional change, only a small code reduction


asvetlov (commented):
I think the change is neutral, it saves nothing.

The next TODO removal would be more interesting thing, I suspect that all supported Python versions are safe already.

bdraco:
https://github.com/aio-libs/aiohttp/pull/9818

PR #9818, opened 2024-11-11
Remove connector close workaround for older Python versions

bdraco:
The TODO is fixed on all supported versions which was added in https://github.com/aio-libs/aiohttp/pull/6321

https://github.com/python/cpython/issues/84132

Note this code is not present in 3.x


bdraco:
I think we need to drop the `test_release_proto_closed_future` test as well

Dreamsorcerer (approved):

asvetlov (approved):

PR #9822, opened 2024-11-11
Improve performance of the access logger when its disabled

bdraco:
## What do these changes do?

Most of the time in the access logger is spent calling `clock_gettime`, move the check to see if its enabled to an `enabled` property so we do not have to fetch time and throw it away.

Screenshot 2024-11-11 at 2 09 02 PM

Screenshot 2024-11-11 at 2 18 35 PM

no

no

### DIFF
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -139,6 +139,15 @@ If your logging needs to perform IO you can instead inherit from
                                     f'"{request.method} {request.path} '
                                     f'done in {time}s: {response.status}')
 
+      @property
+      def enabled(self):
+          """Return True if logger is enabled.
+
+          Override this property if logging is disabled to avoid the
+          overhead of calculating details to feed the logger.
+          """
+          return True
+
 
 This also allows access to the results of coroutines on the ``request`` and
 ``response``, e.g. ``request.text()``.
--- a/docs/logging.rst
+++ b/docs/logging.rst
@@ -114,6 +114,16 @@ Example of a drop-in replacement for the default access logger::
                            f'"{request.method} {request.path} '
                            f'done in {time}s: {response.status}')
 
+      @property
+      def enabled(self):
+          """Return True if logger is enabled.
+
+          Override this property if logging is disabled to avoid the
+          overhead of calculating details to feed the logger.
+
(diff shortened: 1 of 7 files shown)


bdraco:
Probably should make it `_enabled` since we don't want to document this externally

bdraco:
OTOH logger tends to be expensive so an external logger might want to use this as well

docs/logging.rst
```diff
@@ -139,6 +139,15 @@ If your logging needs to perform IO you can instead inherit from
                                     f'"{request.method} {request.path} '
                                     f'done in {time}s: {response.status}')
 
+      @property
+      def enabled(self):
```
bdraco:
```suggestion
      def enabled(self) -> bool:
```

asvetlov (commented):
Maybe `None` access logger provides even less overhead?

bdraco:
> Maybe `None` access logger provides even less overhead?

Passing `None` to `access_log` in `RequestHandler` is definitely better but it lacks the ability to easily flip it on at run time.  Home Assistant turns on `aiohttp.access` at run time when debugging is on so its still super nice to be able to flip it on and only incur the overhead of the isEnabledFor on every request.

tests/test_web_log.py
```diff
@@ -275,12 +275,42 @@ def log(
...
+
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response()
+
+    class Logger(AbstractAccessLogger):
+
+        def log(
+            self, request: web.BaseRequest, response: web.StreamResponse, time: float
+        ) -> None:
+            self.logger.critical("This should not be logged")  # pragma: no cover
```
bdraco, line 304: I tried an `assert False` here but the test still passes so I did `no cover` instead since we want to know for sure the property works

bdraco:
Pushed `access_logger_perf_sq_311` as a 3.11 test backport for testing on production

bdraco:
Looks to be ~2.4% speed up in the overall request time.

[aiohttp_get_profiler.zip](https://github.com/user-attachments/files/17708431/aiohttp_get_profiler.zip)

asvetlov (approved):
Got you. LGTM.

PR #9823, opened 2024-11-12
DNM: Avoid bytes join and memory copy on single chunk

bdraco:
In many cases we end up reading a single chunk, especially from json apis

Note that we have no benchmarks that call `.read()` on the body of the response currently so this will not show any difference

Manual benchmarks for now.. Shows ~4.5% speed up, but we should write a proper codspeed benchmark to verify before considering moving this forward
[aiohttp_get_profiler.zip](https://github.com/user-attachments/files/17709767/aiohttp_get_profiler.zip)

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -396,12 +396,14 @@ class StreamReader(AsyncStreamReaderMixin):
             # deadlock if the subprocess sends more than self.limit
             # bytes.  So just call self.readany() until EOF.
             blocks = []
+            block_count = 0
             while True:
                 block = await self.readany()
                 if not block:
                     break
                 blocks.append(block)
-            return b"".join(blocks) if len(blocks) > 1 else blocks[0] if blocks else b""
+                block_count += 1
+            return b"".join(blocks) if block_count > 1 else blocks[0] if blocks else b""
 
         # TODO: should be `if` instead of `while`
         # because waiter maybe triggered on chunk end,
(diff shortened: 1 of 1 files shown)


bdraco:
Doesn't show as an improvement on codspeed. Need to take a look at why the profile looks better on the callgrind vs codspeed. 

This one needs more investigation.

bdraco:
Difference appears to be chunked vs content length

bdraco:
We also seem to spend time joining a url somewhere in the test server

bdraco:
There is also a lot of memcpy from llhttp on body callback

bdraco:
There is still something that doesn't line up.  Maybe its because I'm profiling locally on mac and the memory management is a bit different

bdraco:
https://github.com/python/cpython/blob/91f4908798074db6c41925b4417bee1f933aca93/Objects/stringlib/join.h#L8 bytes join is implemented here

bdraco:
There is already a fast path for a single element 

https://github.com/python/cpython/blob/91f4908798074db6c41925b4417bee1f933aca93/Objects/stringlib/join.h#L36

bdraco:
I can't replicate this in codspeed, and It looks like the improvement is local jitter since when I look at the bytes join implementation its clear its not making a copy for a single element

bdraco:


finishing a chunk has a lot of memcpy. We could probably switch it to a join there to save one copy

PR #9827, opened 2024-11-12
Adjust client GET read benchmarks to include chunked and content-length

bdraco:
chunked has a different performance profile than content-length (chunked is slower currently). We need benchmarks for both.

I'm still not sure if https://github.com/aio-libs/aiohttp/pull/9823 is a good change or not so I'm adding more benchmarks to validate it


bdraco:
chunked 1MiB is effectively spending ~all~ of its time in memcpy

Screenshot 2024-11-12 at 8 56 41 AM

PR #9828, opened 2024-11-12
[PR #9822/249df6e backport][3.11] Improve performance of the access logger when its disabled

bdraco:
(cherry picked from commit 249df6e1599db75a94dc95f63c498545358f6ef8)

### DIFF
--- a/CHANGES/9822.feature.rst
+++ b/CHANGES/9822.feature.rst
@@ -1,0 +1,1 @@
+Added an :attr:`~aiohttp.abc.AbstractAccessLogger.enabled` property to :class:`aiohttp.abc.AbstractAccessLogger` to dynamically check if logging is enabled -- by :user:`bdraco`.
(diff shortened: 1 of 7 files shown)


bdraco:
That import time test is still flakey

FAILED tests/test_imports.py::test_import_time - assert 257 < 250

Dreamsorcerer:
> That import time test is still flakey
> 
> FAILED tests/test_imports.py::test_import_time - assert 257 < 250

Still looks like 3.12+ as well. Do you think codspeed could handle an import time test? Would be more reliable if we can move it to there instead of CI workers.

bdraco:
> > That import time test is still flakey
> > FAILED tests/test_imports.py::test_import_time - assert 257 < 250
> 
> Still looks like 3.12+ as well. Do you think codspeed could handle an import time test? Would be more reliable if we can move it to there instead of CI workers.

It won't work with codspeed because we have to spawn a subprocess.  I suspect we have an actual problem to fix as a result of a change of some dep that has pushed it over the edge.

bdraco:
Screenshot 2024-11-12 at 9 24 55 AM

Profile by time spent in files...

bdraco:
I don't see any module that is specific or opportunities to optimize.  It looks like we are simply on the edge of the number and its time to bump it up

PR #9838, opened 2024-11-12
Reduce memcpy with chunked encoding

bdraco:

finishing a chunk has a lot of `memcpy`. We can switch it to a `b"".join()` since it has a more efficient implementation in https://github.com/python/cpython/blob/91f4908798074db6c41925b4417bee1f933aca93/Objects/stringlib/join.h#L36 vs constructing a new bytes string for every add operation

### DIFF
--- a/CHANGES/9838.misc.rst
+++ b/CHANGES/9838.misc.rst
@@ -1,0 +1,1 @@
+Improved performance of writing chunked payloads by reducing memory copy -- by :user:`bdraco`.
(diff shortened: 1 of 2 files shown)


bdraco:
Maybe we should use `writelines`

bdraco:
https://github.com/aio-libs/aiohttp/pull/9839 is likely going to be a better solution

bdraco:
replaced by https://github.com/aio-libs/aiohttp/pull/9839

PR #9839, opened 2024-11-12
Implement zero copy writes in `StreamWriter`

bdraco:
https://github.com/python/cpython/issues/91166 made `writelines` zero copy for Python 3.12+. 

finishing a chunk has a lot of `memcpy`. Even if zero copy isn't implement for old python it will use `b"".join()` since it has a more efficient implementation in https://github.com/python/cpython/blob/91f4908798074db6c41925b4417bee1f933aca93/Objects/stringlib/join.h#L36 vs constructing a new bytes string for every add operation https://github.com/aio-libs/aiohttp/pull/9838 shows the bytes join is also faster, just not as fast as the zero copy write

### DIFF
--- a/CHANGES/9839.misc.rst
+++ b/CHANGES/9839.misc.rst
@@ -1,1 +1,1 @@
-Implemented zerocopy writes for ``StreamWriter`` -- by :user:`bdraco`.
+Implemented zero copy writes for ``StreamWriter`` -- by :user:`bdraco`.
--- a/CHANGES/9839.misc.rst
+++ b/CHANGES/9839.misc.rst
@@ -1,0 +1,1 @@
+Implemented zerocopy writes for ``StreamWriter`` -- by :user:`bdraco`.
--- a/tests/test_client_request.py
+++ b/tests/test_client_request.py
@@ -81,7 +81,7 @@ def protocol(
 
 
 @pytest.fixture
-def transport(buf: bytearray) -> Any:
+def transport(buf: bytearray) -> mock.Mock:
     transport = mock.create_autospec(asyncio.Transport, spec_set=True, instance=True)
 
     def write(chunk: bytes) -> None:
@@ -94,7 +94,7 @@ def transport(buf: bytearray) -> Any:
     transport.write.side_effect = write
     transport.writelines.side_effect = writelines
     transport.is_closing.return_value = False
-    return transport
+    return transport  # type: ignore[no-any-return]
 
 
 @pytest.fixture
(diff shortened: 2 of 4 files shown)


bdraco:
Well that was much better than expected.

We don't have a benchmark for compressed payloads here. We should add that as well

tests/test_client_request.py
```diff
@@ -72,16 +81,20 @@ def protocol(
 
 
 @pytest.fixture
-def transport(buf: bytearray) -> mock.Mock:
-    transport = mock.create_autospec(asyncio.Transport, spec_set=True)
+def transport(buf: bytearray) -> Any:
```
bdraco:
```suggestion
def transport(buf: bytearray) -> mock.Mock:
```

tests/test_client_request.py
```diff
@@ -72,16 +81,20 @@ def protocol(
...
+    def writelines(chunks: Iterable[bytes]) -> None:
+        for chunk in chunks:
+            buf.extend(chunk)
+
     transport.write.side_effect = write
+    transport.writelines.side_effect = writelines
     transport.is_closing.return_value = False
-
-    return transport  # type: ignore[no-any-return]
+    return transport
```
bdraco:
```suggestion
    return transport  # type: ignore[no-any-return]
```

tests/test_client_request.py
```diff
@@ -73,14 +82,18 @@ def protocol(
...
 
+    def writelines(chunks: Iterable[bytes]) -> None:
+        for chunk in chunks:
+            buf.extend(chunk)
+
     transport.write.side_effect = write
+    transport.writelines.side_effect = writelines
     transport.is_closing.return_value = False
-
     return transport  # type: ignore[no-any-return]
```
bdraco, line 98:
```suggestion

    return transport  # type: ignore[no-any-return]
```

CHANGES/9839.misc.rst
```diff
@@ -0,0 +1 @@
+Implemented zerocopy writes for ``StreamWriter`` -- by :user:`bdraco`.
```
bdraco:
```suggestion
Implemented zero copy writes for ``StreamWriter`` -- by :user:`bdraco`.
```

asvetlov (approved):
Yeah, many years ago `transport.writelines(...)` was implemented as `transport.write(b''.join(lines))`.

Now it utilizes `sendmsg()` which is more efficient than `memcpy()` + `send()`

bdraco:
Still needs a test for a large compressed payload without chunked that is large enough to ensure compress and flush both output a payload

PR #9872, opened 2024-11-14
[3.11] Make creating `RequestInfo` backwards compatible with 3.10

### LINKED ISSUE #9866: `RequestInfo` change to a named tuple broke BC
### Describe the bug

Existing usage of aiohttp 3.10 no longer works with aiohttp 3.13, causing the following error: `TypeError: RequestInfo.__new__() missing 1 required positional argument: 'real_url'`

### To Reproduce

Run the following code:
```
"""
Reproduce a regression in aiohttp 3.11.0.
"""

from multidict import CIMultiDict, CIMultiDictProxy

from aiohttp.client_reqrep import RequestInfo
from yarl import URL

RequestInfo(URL(), "GET", CIMultiDictProxy(CIMultiDict()))
```
With aiohttp 3.10.11, this exits without an error. With aiohttp 3.11.0, this raises `TypeError: RequestInfo.__new__() missing 1 required positional argument: 'real_url'`

### Expected behavior

The expected behavior is for no error to be raised, by `real_url` not being a required argument.

### Logs/tracebacks

```python-traceback
.
```

### Python Version

```console
3.13
```

### aiohttp Version

```console
3.11.0
```

### multidict Version

```console
6.1.0
```

### propcache Version

```console
0.2.0
```

### yarl Version

```console
1.17.1
```

### OS

Ubuntu 24.04

### Related component

Client

### Additional context

_No response_

### Code of Conduct

bdraco:
It was unexpected that this object was being created directly
outside of aiohttp internals. While it looks like its only
used for mocking downstream, we can accomodate that by subclassing
the NamedTuple and providing a `__new__` while keeping the faster
`tuple.__new__` internally.

fixes #9866

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -42,6 +42,7 @@ from .client_exceptions import (
 from .compression_utils import HAS_BROTLI
 from .formdata import FormData
 from .helpers import (
+    _SENTINEL,
     BaseTimerContext,
     BasicAuth,
     HeadersMixin,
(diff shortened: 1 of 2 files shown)


bdraco:
Really not harm in doing this on `master` as well as we can always deprecate it later

PR #9876, opened 2024-11-14
DNM: Ensure manual builds use at least the same minimum Cython version used to build our wheels

bdraco:
https://github.com/aio-libs/aiohttp/discussions/9871#discussioncomment-11256618 showed some warnings/errors that are fixed in newer `Cython`.  We currently don't specify a minimum `Cython` version which means manual builds can result in broken builds if the `Cython` version being used for the build is too old.

This PR should not merge until we confirm this is the issue in the discussion

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,7 @@
 [build-system]
 requires = [
     "setuptools >= 46.4.0",
+    "Cython >= 3.0.11",
 ]
 build-backend = "setuptools.build_meta"
 


pyproject.toml
```diff
@@ -1,6 +1,7 @@
 [build-system]
 requires = [
     "setuptools >= 46.4.0",
+    "Cython >= 3.0.11",
```
webknjaz, line 4: This won't do anything since the current packaging implementation in aiohttp differs from the other libs. setuptools is configured to build from C-files. And those C-files have to exist. So the contributors have to call cythonize. And we call it when publishing. The end-users either get pre-built wheels or hit the sdist that contains those C-files.
So everything is dependent on whatever https://github.com/aio-libs/aiohttp/blob/9e3a328/requirements/cython.txt outputs at the time of release (this is also known to cause problems with older Cython-produced C-files not being compatible with bleeding-edge CPythons).
At some point, I'd like to change the build process but this is how things are right now.

Hence, we shouldn't add this here as it's never getting invoked in the current process.

asvetlov: Also I wonder how build will go on Python distro where Cython is not available?
python for java, .net, whatever...

webknjaz: Cython is not the problem. It can convert PYX into C with its pure-python code. I'd worry about the following step (calling GCC/CLANG).

PR #9885, opened 2024-11-14
Add benchmarks for web middleware

bdraco:
We didn't have any benchmarks that use Middleware, and since we have had some significant slow downs in the past using middleware, we should keep track of this.

### DIFF
--- a/tests/test_benchmarks_web_middleware.py
+++ b/tests/test_benchmarks_web_middleware.py
@@ -9,7 +9,7 @@ from aiohttp.pytest_plugin import AiohttpClient
 from aiohttp.typedefs import Handler
 
 
-def test_serialize_headers(
+def test_ten_web_middlewares(
     benchmark: BenchmarkFixture,
     loop: asyncio.AbstractEventLoop,
     aiohttp_client: AiohttpClient,
--- a/tests/test_benchmarks_web_middleware.py
+++ b/tests/test_benchmarks_web_middleware.py
@@ -1,0 +1,43 @@
+"""codspeed benchmarks for web middlewares."""
+
+import asyncio
+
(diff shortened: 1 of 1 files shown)


tests/test_benchmarks_web_middleware.py
```diff
@@ -0,0 +1,43 @@
...
+import asyncio
+
+from pytest_codspeed import BenchmarkFixture
+
+from aiohttp import web
+from aiohttp.pytest_plugin import AiohttpClient
+from aiohttp.typedefs import Handler
+
+
+def test_serialize_headers(
```
bdraco:
```suggestion
def test_ten_web_middlewares(
```

tests/test_benchmarks_web_middleware.py
```diff
@@ -0,0 +1,43 @@
...
+    """Benchmark 100 requests with 10 middlewares."""
+    message_count = 100
+
+    async def handler(request: web.Request) -> web.Response:
+        return web.Response()
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+
+    class MiddlewareClass:
```
Dreamsorcerer, line 26: The class seems a little odd here, I don't see it's purpose.

But, I wonder if it's also worth having a benchmark with something like 10 middlewares?

bdraco: The goal is to track the overhead of processing the middleware. I added the same "empty" middleware 10 times since I didn't want to benchmark the overhead of middleware itself, only the overhead of having/executing it.

Dreamsorcerer: Oh yes, I didn't look properly. Missed the append was in a loop.

PR #9895, opened 2024-11-15
Defer creation of SimpleCookie objects in the web server until needed

bdraco:
related issue #2779

We create these objects even if there are no cookies.  There are some small performance improvements and some memory savings on each request.

### DIFF
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -949,9 +949,6 @@ class CookieMixin:
         """
         if self._cookies is None:
             self._cookies = SimpleCookie()  # type: ignore[misc]
-        elif (old := self._cookies.get(name)) is not None and old.coded_value == "":
-            # deleted cookie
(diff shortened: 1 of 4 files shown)


bdraco:
Todo

- [ ] Focused benchmarks for creating web Responses and StreamResponses as we only have visibility on the more macro benchmark here
- [ ] Add test coverage for overriding an already set cookie

aiohttp/helpers.py
```diff
@@ -945,10 +947,8 @@ def set_cookie(
         Sets new cookie or updates existent with new value.
         Also updates only those params which are not None.
         """
-        old = self._cookies.get(name)
-        if old is not None and old.coded_value == "":
-            # deleted cookie
-            self._cookies.pop(name, None)
```
bdraco, line 951: This was unreachable since cookies are always set with a path

PR #9896, opened 2024-11-15
DNM: Attempt to speed up _prepare_headers

bdraco:
https://github.com/aio-libs/aiohttp/issues/2779#issuecomment-2477896415


bdraco:
no improvement

PR #9898, opened 2024-11-15
DNM: Benchmark overhead of setdefault calls in _prepare_headers

bdraco:
https://github.com/aio-libs/aiohttp/issues/2779#issuecomment-2477885530


bdraco:
no difference

PR #9899, opened 2024-11-15
Refactor Resource.resolve to avoid linear search of methods

bdraco:
The average time complexity of `Resource.resolve` is now `O(1)` instead of `O(number_of_methods)`

https://github.com/aio-libs/aiohttp/issues/2779#issuecomment-2477962741

Looks to be 
~14% faster with a single method
~2.5x faster for the multiple methods 

Screenshot 2024-11-15 at 8 30 25 AM

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -361,8 +361,7 @@ class Resource(AbstractResource):
             self._routes[route.method] = route
 
     async def resolve(self, request: Request) -> _Resolve:
-        match_dict = self._match(request.rel_url.path_safe)
-        if match_dict is None:
+        if (match_dict := self._match(request.rel_url.path_safe)) is None:
             return None, set()
         if route := self._routes.get(request.method) or self._any_route:
             return UrlMappingMatchInfo(match_dict, route), self._allowed_methods
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -339,11 +339,11 @@ class Resource(AbstractResource):
         *,
         expect_handler: Optional[_ExpectHandler] = None,
     ) -> "ResourceRoute":
-        if method in self._routes or self._any_route is not None:
+        if (route := self._routes.get(method)) or self._any_route is not None:
             raise RuntimeError(
                 "Added route will never be executed, "
                 "method {route.method} is already "
-                "registered".format(route=self._routes.get(method, self._any_route))
+                "registered".format(route=route or self._any_route)
             )
 
         route_obj = ResourceRoute(method, handler, self, expect_handler=expect_handler)
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -328,7 +328,9 @@ async def _default_expect_handler(request: Request) -> None:
 class Resource(AbstractResource):
     def __init__(self, *, name: Optional[str] = None) -> None:
         super().__init__(name=name)
-        self._routes: List[ResourceRoute] = []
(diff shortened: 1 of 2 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -337,13 +339,12 @@ def add_route(
...
         expect_handler: Optional[_ExpectHandler] = None,
     ) -> "ResourceRoute":
-        for route_obj in self._routes:
-            if route_obj.method == method or route_obj.method == hdrs.METH_ANY:
-                raise RuntimeError(
-                    "Added route will never be executed, "
-                    "method {route.method} is already "
-                    "registered".format(route=route_obj)
-                )
+        if route := self._routes.get(method) or self._any_route:
```
bdraco: Probably should be default instead of or

aiohttp/web_urldispatcher.py
```diff
@@ -353,23 +354,18 @@ def register_route(self, route: "ResourceRoute") -> None:
...
-            route_method = route_obj.method
-            allowed_methods.add(route_method)
-
-            if route_method == request.method or route_method == hdrs.METH_ANY:
-                return (UrlMappingMatchInfo(match_dict, route_obj), allowed_methods)
-        else:
-            return None, allowed_methods
+        if (match_dict := self._match(request.rel_url.path_safe)) is None:
+            return None, set()
+        if route := self._routes.get(request.method) or self._any_route:
```
bdraco: Probably use default instead of or

asvetlov (approved):
Looks good.
Regarding `or` vs getting with default -- up to you.
Getting with the default looks a little better, the speed on macro benchmark should be roughly the same I guess.

aiohttp/web_urldispatcher.py
```diff
@@ -353,23 +354,18 @@ def register_route(self, route: "ResourceRoute") -> None:
         assert isinstance(
             route, ResourceRoute
         ), f"Instance of Route class is required, got {route!r}"
-        self._routes.append(route)
+        if route.method == hdrs.METH_ANY:
+            self._any_route = route
+        else:
+            self._allowed_methods.add(route.method)
```
Dreamsorcerer, line 360: Why not use `self._routes.keys()`?

aiohttp/web_urldispatcher.py
Dreamsorcerer, line 1: Just double checking that this doesn't need updating (I don't think it changes anything for user..)?
https://docs.aiohttp.org/en/stable/web_quickstart.html#resources-and-routes

Dreamsorcerer (approved):

aiohttp/web_urldispatcher.py
bdraco: To avoid making a new set every time since the api currently wants a set returned. 

Maybe could be refactored more to change the type to allowed to be a Collection but I wanted to keep the change compatible as possible

aiohttp/web_urldispatcher.py
bdraco: Doesn't look like it. Should be an internals only change

aiohttp/web_urldispatcher.py
asvetlov: Technically `dict.keys()` provides `collections.abc.Set`, switching from `set[str]` to `collections.abc.Set[str]` is almost safe.

bdraco: `keys()` looks to be a tiny bit slower when benchmarking locally so I left it as-is.

asvetlov: up to you

PR #9907, opened 2024-11-15
POC for hyperscan usage in UrlDispatcher

asvetlov:
https://pypi.org/project/hyperscan/ could speed up url dispatcher if available.

Fallback to the current implementation is supported if hyperscan compilation fails or not available.


aiohttp/web_urldispatcher.py
```diff
@@ -1242,8 +1301,52 @@ def freeze(self) -> None:
         super().freeze()
         for resource in self._resources:
             resource.freeze()
+        if HAS_HYPERSCAN:
```
bdraco: Haven't tested this yet but I expect it will break downstream Home Assistant since there is code there to register new routes as new integrations get loaded after startup. It's pretty clear the intent here was to not allow that after freeze but that's currently the downstream use case and hard to change it.

asvetlov: Good point.
We can rebuild hypedscan db in `UrlDispatcher.register_resource()` if the dispatcher was frozen.
I'm curious how Home Assistant works with route table on the fly.
If it adds new routes to started web server it should remove them at some point, isn't it?

Before moving forward, I'd like to add some benchmarks first.

aiohttp/web_urldispatcher.py
```diff
@@ -997,21 +1009,56 @@ class UrlDispatcher(AbstractRouter, Mapping[str, AbstractResource]):
...
+        # N.B.: allowed_methods variable in modified in-place
+        found: list[int] = []
+
+        def on_match(
+            id_: int, from_: int, to: int, flags: int, context: Any | None
+        ) -> bool | None:
+            found.append(id_)
+            return None
+
+        self._hyperdb.scan(path.encode("utf8"), match_event_handler=on_match)
```
bdraco: This is really nice. Didn't know about hyperscan before and could have used it to solve other problems in the past.  Thanks for turning me on to it

aiohttp/web_urldispatcher.py
bdraco: About to go to the airport, but will dig up the Home Assistant code this weekend. It predates me though so will take a bit to figure it out.

We don't have any benchmarks for the url dispatcher.  I'll look at downstream use cases and come up with some after I get back from travel next week.

asvetlov: Have you safe flight!

bdraco: Once I figure out how the downstream code works for adding routes at run time. I'll see if I can make that into meaningful test cases here as well so we don't break anything.  AFAICT for a cursory review, routes are never removed when integrations are unloaded downstream.

aiohttp/web_urldispatcher.py
```diff
@@ -104,23 +116,23 @@ class _InfoDict(TypedDict, total=False):
...
 
-    rule: "AbstractRuleMatching"
+    rule: AbstractRuleMatching
 
     http_exception: HTTPException
 
 
 class AbstractResource(Sized, Iterable["AbstractRoute"]):
-    def __init__(self, *, name: Optional[str] = None) -> None:
+    def __init__(self, *, name: str | None = None) -> None:
```
Dreamsorcerer: We've been avoiding these typing changes. This is not valid on Python 3.9 (future import only delays the error, it will still happen if you inspect the types at runtime).

Pyupgrade will automatically update all of these across the entire project once 3.9 is dropped.

setup.cfg
```diff
@@ -70,6 +70,8 @@ speedups =
   aiodns >= 3.2.0; sys_platform=="linux" or sys_platform=="darwin"
   Brotli; platform_python_implementation == 'CPython'
   brotlicffi; platform_python_implementation != 'CPython'
+  Brotli; platform_python_implementation == 'CPython'
```
Dreamsorcerer: You duplicated this line.
```suggestion
```

aiohttp/web_urldispatcher.py
asvetlov: But aiohttp never inspects the types at runtime!
How can it can affect anybody?

Dreamsorcerer: A user could inspect them and (e.g.) pass the signature into pydantic or something.

bdraco:
Managed to finish the benchmarks before boarding the flight. They should run if you update with master.

Didn't have enough time to dig through downstream or write the tests yet though

asvetlov:
Thank you!

aiohttp/web_urldispatcher.py
asvetlov: Ok, I'll revert because typing changes not a subject of the PR.
I still think that type hints from aiohttp are used by type checkers only and never in runtime.

Dreamsorcerer: > I still think that type hints from aiohttp are used by type checkers only and never in runtime.

Well, that's down to the user. We've seen much weirder things. As I say, pyupgrade will handle all of this automatically anyway.

Dreamsorcerer:
Seems to be a lot slower in the benchmarks. I suspect that's due to our previous refactoring, so in the typical case we're doing a couple of dictionary lookups, which is probably a lot faster than hyperscan. Maybe if there's a hyperscan under each dictionary lookup, that might be an improvement?

Dreamsorcerer:
i.e. Instead of replacing the entire loop, maybe it could replace the `await candidate.resolve(request)` line in certain circumstances (ideally only when there are regex patterns involved).

bdraco:
Maybe if there are lot of DynamicResources in a list because they have the same anchor point it would be a lot faster for that case

Dreamsorcerer:
> Maybe if there are lot of DynamicResources in a list because they have the same anchor point it would be a lot faster for that case

Yeah, that's why I mean when I say it might be good if we can use it only when we get something from the dict, and it has (multiple) regexes.

asvetlov:
I think that regexes are pretty common in the real life, that's why https://github.com/aio-libs/aiohttp/pull/9925 was added.

I agree that for constant path dict lookup could be faster.
As we now prefer constant matches over regexes, maybe a `{path: resource}` dict for plain and prefixed resources + hyperscan for regexes could help.

Let's measure it first.

Dreamsorcerer:
> As we now prefer constant matches over regexes

Also, it's not just static paths, but paths that end with regexes will be faster.
If the path is `/foo/bar/{regex}`, then this will be 2 dict lookups and a single regex match, so is very likely to still beat hyperscan.
Where hyperscan will likely improve performance is URLs like `/org/{regex}/this/that/other`, in which case this might result in several regex matches with the current code.

Dreamsorcerer:
That's a massive improvement on the gitapi benchmark, so we'll definitely need to figure out some way to integrate this!

.mypy.ini
```diff
@@ -34,3 +34,6 @@ ignore_missing_imports = True
 
 [mypy-gunicorn.*]
 ignore_missing_imports = True
+
+[mypy-hyperscan]
+ignore_missing_imports = True
```
Dreamsorcerer, line 39: https://github.com/darvid/python-hyperscan/pull/166

aiohttp/web_urldispatcher.py
```diff
@@ -998,27 +1008,75 @@ def __init__(self) -> None:
...
+                allowed_methods |= allowed
+
+        if self._hyperdb is not None:
+            found: list[int] = []
+            resources = self._resources
+
+            self._hyperdb.scan(
+                path.encode("utf8"), match_event_handler=self._on_match, context=found
+            )
+            if found:
```
Dreamsorcerer: This if is probably not needed (and would clear a missing branch on the codecov).

setup.cfg
```diff
@@ -70,6 +70,7 @@ speedups =
   aiodns >= 3.2.0; sys_platform=="linux" or sys_platform=="darwin"
   Brotli; platform_python_implementation == 'CPython'
   brotlicffi; platform_python_implementation != 'CPython'
+  hyperscan >= 0.7.8; platform_python_implementation == 'CPython' and (sys_platform=="linux" or sys_platform=="darwin") and (python_version < "3.14")
```
Dreamsorcerer: Do we want the <3.14 here? Seems like something that'll be easy to forget.

aiohttp/web_urldispatcher.py
```diff
@@ -998,27 +1008,75 @@ def __init__(self) -> None:
...
+        path = request.rel_url.path_safe
+
+        if (plain_resource := self._plain_resources.get(path)) is not None:
+            match_dict, allowed = await plain_resource.resolve(request)
+            if match_dict is not None:
+                return match_dict
+            else:
+                allowed_methods |= allowed
+
+        for prefix, prefix_resource in self._prefix_resources:
```
Dreamsorcerer: I think this part can probably reuse the previous code with the `while url_part`, which would stop this growing in time as more prefixed resources are added.

aiohttp/web_urldispatcher.py
```diff
@@ -1239,6 +1297,55 @@ def freeze(self) -> None:
...
+        ids: list[int] = []
+        for id_, resource in enumerate(self._resources):
+            if isinstance(resource, PlainResource):
+                self._plain_resources[resource.get_info()["path"]] = resource
+                continue
+            elif isinstance(resource, DynamicResource):
+                pattern = resource.get_info()["pattern"].pattern
+            elif isinstance(resource, PrefixResource):
+                if isinstance(resource, MatchedSubAppResource):
+                    # wildcard resources doesn't fit hyperscan table
```
Dreamsorcerer, line 1340: This could probably work with a second hyperscan table. Though it's such a niche feature, it may not be worth the effort.

Dreamsorcerer (commented):
Looks good to me. Seems we need a couple more tests for missing coverage though.

.mypy.ini
asvetlov: https://github.com/darvid/python-hyperscan/pull/165 ! :)

aiohttp/web_urldispatcher.py
bdraco: https://github.com/home-assistant/core/blob/445690588c855498b56561805ed652a12f36a21d/homeassistant/components/http/__init__.py#L620

Home Assistant effectively prevents freeze from happening and relies on the current implementation detail of route registration being implemented synchronously.

That's obviously not so great but it is currently effective. There are also a LOT of routes being added during and after startup as integrations get loaded.

Need to think about the best solution for downstream. 

Maybe the simplest solution is to allow routes to be added after freeze as long as freeze is called again as that should give some control
Of when the hyperdb gets rebuilt to reduce churning it. While it's still a breaking change for downstream, given freeze is being prevented that's seems reasonable to say downstream needs to adapt as it wasn't guaranteed behavior in the first place.  Maybe there is a better way though.   Need to think on it

bdraco: Maybe a new feature is to have a way to add routes/views after startup but don't guarantee they will be efficient until calling some method after new routes are added, which would internally trigger the hyperdb rebuild or any other data structures needed to be rebuilt. I think that would avoid the downstream monkey-patching and still fill the existing use case.

Not sure if that should be a separate PR before or after this merges as I'm not sure how the final implementation of this PR will look yet.

bdraco:
Just realized Hyperscan likely doesn't work on armv7l so maybe we need benchmarks for with Hyperscan not available as well.  Maybe not though since armv7l is now slowly disappearing from the landscape now that the RPi shortage is over

aiohttp/web_urldispatcher.py
asvetlov: Done

setup.cfg
asvetlov: hyperscan doesn't exist for 3.14 yet. 
I'll try to make tests passed on 3.14. 
Now they fails on pydantic installation; it presents in our requirement files but never used by code. I guess the dependency could be removed in a separate PR.

aiohttp/web_urldispatcher.py
asvetlov: I'm experimented with alternative implementation for prefixed resources that should be faster than index dict, see updated PR.

aiohttp/web_urldispatcher.py
asvetlov: I'd prefer keeping this possible optimization out of the PR scope if you don't mind.

asvetlov:
> Just realized Hyperscan likely doesn't work on armv7l so maybe we need benchmarks for with Hyperscan not available as well. Maybe not though since armv7l is now slowly disappearing from the landscape now that the RPi shortage is over

Perhaps parametrized benchmarks will help to measure both hyperscan and domestic versions.
Also I'm planing to use parametrization for existing url_dispatcher tests (and maybe web_functional ones) for tests coverage and correctness checking.

bdraco:
From the flame graph on cod speed it looks  like test_resolve_gitapi_subapps is producing system routes (maybe 404s)

asvetlov:
Good finding, thanks!
I'll modify benchmarks to check the resolved route to prevent this situation from becoming possible.

aiohttp/web_urldispatcher.py
Dreamsorcerer: > Maybe the simplest solution is to allow routes to be added after freeze as long as freeze is called again as that should give some control

I feel like any change to this should probably go into v4 only. We've currently always assumed that once the app has started, the routes are frozen. Other libraries depend on this behaviour in some cases, for example aiohttp-apischema evaluates the endpoint schema on startup from the routes and doesn't expect them to change at runtime.

If we do change this, we should have an explicit mechanism for other libraries to know about these changes. e.g. By registering callbacks which can be triggered when the routes are updated.

setup.cfg
Dreamsorcerer: 3.14 test are unlikely to pass until we are close to release. I just put it in early in preparation. But, we'd need wheels for multidict/yarl etc. which will only happen when 3.14rc is available.

aiohttp/web_urldispatcher.py
Dreamsorcerer: That looks like it's doubled the improvement on gitapi_subapps, so looking good to me.

aiohttp/web_urldispatcher.py
asvetlov: BTW, is there a document for v4 roadmap?
What is the current status, blockers, things to be done?

bdraco: > > Maybe the simplest solution is to allow routes to be added after freeze as long as freeze is called again as that should give some control
> 
> I feel like any change to this should probably go into v4 only. We've currently always assumed that once the app has started, the routes are frozen. Other libraries depend on this behaviour in some cases, for example aiohttp-apischema evaluates the endpoint schema on startup from the routes and doesn't expect them to change at runtime.
> 
> If we do change this, we should have an explicit mechanism for other libraries to know about these changes. e.g. By registering callbacks which can be triggered when the routes are updated.

Maybe the safest way would be to require creating the app with `mutable_routes=True` and raise a warning if the user tries to modify routes after freeze without it telling them to carefully read the docs are there are performance considerations and a rebuild must be done when changing view/routes after freeze.  That should ensure we don’t break any current behavior and provide a path forward for anything that is currently needing to modify after freeze.

Dreamsorcerer: We literally just started thinking about this again, having triaged and sorted out a bunch of older issues.
Currently, it's just the v4 milestone, so we'll have to start getting through the tasks on there. I haven't checked whether the older things are still relevant, but have been adding things over time to the milestone that should be looked at before v4.

bdraco: I've been bucketing into the milestones with these conventions:

3.12: shorter term: https://github.com/aio-libs/aiohttp/milestone/35
4.0: old and longer term items: https://github.com/aio-libs/aiohttp/milestone/20

The caveat is that the 4.0 milestone curation is a bit stale.

asvetlov: > The caveat is that the 4.0 milestone curation is a bit stale.

Yes, this is my fault. Sorry, guys.

asvetlov: `mutable_routes=True` could be a good idea, I need sleep on it.

setup.cfg
asvetlov: OK, I see your point.

aiohttp/web_urldispatcher.py
Dreamsorcerer: My criteria is just anything that is likely to result in breaking changes or deprecations I've added to the v4 milestone.

Dreamsorcerer: > That should ensure we don’t break any current behavior and provide a path forward for anything that is currently needing to modify after freeze.

I feel like it's still breaking the libraries which aren't expected to advertise to users that they don't work when the flag is set.

Unless there's a clear tradeoff, I'd just put this into v4 without a flag. We just need to ensure that homeassistant has a way to continue hacking it in the mean time.

asvetlov: We can introduce a new API for rebuilding, e.g.
```
with rebuild(app):
    app.router.add_get(...)
    app.add_subapp(...)
    app.middlewares.append(...)
```
The idea is *unfreezing* on `__enter__` and freezing back on `__exit__`.
The first version could work with route table only, middlewares and other things could be added iteratively on demand.

@bdraco could the proposed API fit Home Assistance usage scenario?

Dreamsorcerer: > The idea is _unfreezing_ on `__enter__` and freezing back on `__exit__`.

That seems like a good way to avoid too much churn from recomputing everything after every call (if there are multiple calls making modifications).

aiohttp/web_urldispatcher.py
```diff
@@ -998,20 +1008,41 @@ def __init__(self) -> None:
...
+        url_part = path
+        prefix_resources = self._prefix_resources
 
         # Walk the url parts looking for candidates. We walk the url backwards
         # to ensure the most explicit match is found first. If there are multiple
         # candidates for a given url part because there are multiple resources
         # registered for the same canonical path, we resolve them in a linear
         # fashion to ensure registration order is respected.
-        url_part = request.rel_url.path_safe
         while url_part:
```
Dreamsorcerer: If we're back to this approach, I'm wondering if it worth merging plain and prefix resources again. Might be a tiny bit slower when a plain resource matches and a tiny bit faster in other cases. But, would also reduce the amount of code to maintain.

asvetlov: I'm still working on the PR, who knows how it will look at the end.
The existing index table is suboptimal for prefix resources.
Assume we have `/app` prefix subapp.
For handling path like `/app/path/inside/subapp` the code tries `/app/path/inside/subapp`, `/app/path/inside`, `/app/path`, and the only last `/app` test succeeds.
I would like to avoid both linear scan *and* walking up to the path for prefix resources.
The optimization target is good results for `test_resolve_gitapi_subapps` and `test_resolve_prefix_resources_many_prefix_many_plain` benchmark, the later was added by https://github.com/aio-libs/aiohttp/pull/9935
If you imagine other common cases for prefixed resources please share it. We can add another benchmarks and optimize the code to fit them as well.

aiohttp/web_urldispatcher.py
asvetlov: It also requires a locking probably to avoid concurrent route table modification from two web endpoints, for example.
Anyway, I want to finish optimizations first and maybe later return to freeze/unfreeze if we decide to use this approach.

aiohttp/web_urldispatcher.py
Dreamsorcerer: > If you imagine other common cases for prefixed resources please share it.

Subapps should probably be the priority for optimisation.

aiohttp/web_urldispatcher.py
bdraco: Freezing and unfreezing on enter and exit will likely be too many rebuilds.  ~297 routes get added at startup for Home Assistant. ~200 can be grouped together which would mean 200 rebuilds at startup

bdraco: Another method could be to keep adding to the fallback dict as new routes are added, clear the hyperscan db when a route is added after freeze, schedule a timer handle to rebuild the db a few (60... maybe less) seconds later if one is not already scheduled

Dreamsorcerer: > ~200 can be grouped together which would mean 200 rebuilds at startup

If they're grouped together, shouldn't that be 1 rebuild?

bdraco: Clarify: Very few can be grouped together. The result is 200 groups of various sizes, mostly 1

bdraco: Distribution of groupings looks something like 1, 9, 1, 7, 3, 3, 1, 1, 1, 1, 1 .....

However order is indeterminate as it relies on async task completion order

setup.cfg
asvetlov: Done

bdraco:
> > Just realized Hyperscan likely doesn't work on armv7l so maybe we need benchmarks for with Hyperscan not available as well. Maybe not though since armv7l is now slowly disappearing from the landscape now that the RPi shortage is over
> 
> Perhaps parametrized benchmarks will help to measure both hyperscan and domestic versions. Also I'm planing to use parametrization for existing url_dispatcher tests (and maybe web_functional ones) for tests coverage and correctness checking.

Opened https://github.com/darvid/python-hyperscan/issues/170

PR #9910, opened 2024-11-15
Add benchmarks for the URL dispatcher

bdraco:
For checking https://github.com/aio-libs/aiohttp/pull/9907 impact

### DIFF
--- a/tests/test_benchmarks_web_urldispatcher.py
+++ b/tests/test_benchmarks_web_urldispatcher.py
@@ -1,0 +1,133 @@
+"""codspeed benchmarks for the URL dispatcher."""
+
+import asyncio
+from typing import NoReturn
+from unittest import mock
+
+from multidict import CIMultiDict, CIMultiDictProxy
+from pytest_codspeed import BenchmarkFixture
+from yarl import URL
+
+from aiohttp import web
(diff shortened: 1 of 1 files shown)


tests/test_benchmarks_web_urldispatcher.py
```diff
@@ -0,0 +1,133 @@
...
+) -> None:
+    """Resolve different a DynamicResource when there are 250 PlainResources registered."""
+
+    async def handler(request: web.Request) -> NoReturn:
+        assert False
+
+    app = web.Application()
+    for count in range(250):
+        app.router.add_route("GET", f"/api/server/other/{count}/update", handler)
+    app.router.add_route("GET", "/api/server/dispatch/{customer}/update", handler)
```
bdraco, line 174: Probably need something that generates 1000+ routes and some longer ones since that better reflects downstream use cases

asvetlov: Sorry, I don't follow.

bdraco: Sorry, that was a note to self so I don't forget where I left off when I context switch back

PR #9911, opened 2024-11-15
Only construct the `allowed_methods` `set` once for a `StaticResource`

bdraco:
noticed in https://github.com/aio-libs/aiohttp/pull/9899#issuecomment-2479054068

Likely to have minimal performance improvement besides saving some memory churn so I didn't call it in out the changelog

### DIFF
--- a/aiohttp/web_urldispatcher.py
+++ b/aiohttp/web_urldispatcher.py
@@ -557,6 +557,7 @@ class StaticResource(PrefixResource):
                 "HEAD", self._handle, self, expect_handler=expect_handler
             ),
         }
+        self._allowed_methods = set(self._routes)
 
     def url_for(  # type: ignore[override]
         self,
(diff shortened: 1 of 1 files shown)


aiohttp/web_urldispatcher.py
```diff
@@ -620,15 +621,18 @@ def set_options_route(self, handler: Handler) -> None:
...
+        allowed_methods = self._allowed_methods
         if method not in allowed_methods:
             return None, allowed_methods
 
         match_dict = {"filename": _unquote_path_safe(path[len(self._prefix) + 1 :])}
-        return (UrlMappingMatchInfo(match_dict, self._routes[method]), allowed_methods)
+        return (
+            UrlMappingMatchInfo(match_dict, self._routes[method]),
+            self._allowed_methods,
+        )
```
```suggestion
        return (UrlMappingMatchInfo(match_dict, self._routes[method]), allowed_methods)
```

bdraco:
Going to have to revert this since `aiohttp-cors` accesses the protected `self._routes` 😢 

https://github.com/aio-libs/aiohttp-cors/blob/38c6c17bffc805e46baccd7be1b4fd8c69d95dc3/aiohttp_cors/urldispatcher_router_adapter.py#L187

PR #9925, opened 2024-11-16
Add benchmarks for github url resolving emulation

asvetlov:
I think github API is a representative example for possible route table.


bdraco:
That websocket test flaps a bit because of memory allocation inconsistency. We might want to add it to the ignores until it can be tuned a bit more.  Or simply ignore it for now since it's still a semi rare flap

tests/test_benchmarks_web_urldispatcher.py
```diff
@@ -92,11 +98,12 @@ async def handler(request: web.Request) -> NoReturn:
     for count in range(250):
         app.router.add_route("GET", f"/api/server/dispatch/{count}/update", handler)
     app.freeze()
+    router = app.router
```
Dreamsorcerer, line 101: Why all these changes? Seems like it's only accessed once, so assigning to a variable is surely just adding slightly more overhead and reducing conciseness.

Dreamsorcerer (approved):
Seems like a good idea to me.

tests/test_benchmarks_web_urldispatcher.py
asvetlov: No, it is accessed in a loop a few lines below,.
```
    async def run_url_dispatcher_benchmark() -> None:
        for _ in range(resolve_count):
            await router.resolve(request)
```
Benchmarks actually measure `run_url_dispatcher_benchmark()` execution, not the whole test.
Preparation time is not counted.

asvetlov:
> That websocket test flaps a bit because of memory allocation inconsistency. We might want to add it to the ignores until it can be tuned a bit more. Or simply ignore it for now since it's still a semi rare flap.
> 
I'm ok to do nothing with it until a champion fixes the flappines (or implementation).

PR #9939, opened 2024-11-17
Fix test_resolve_gitapi_subapps benchmark

asvetlov:
Also add checks that url dispatcher benchmarks find expected routes

### DIFF
--- a/tests/test_benchmarks_web_urldispatcher.py
+++ b/tests/test_benchmarks_web_urldispatcher.py
@@ -6,7 +6,7 @@ import pathlib
 import random
 import string
 from pathlib import Path
-from typing import NoReturn
+from typing import NoReturn, Optional
 from unittest import mock
 
 from multidict import CIMultiDict, CIMultiDictProxy
@@ -53,13 +53,15 @@ def test_resolve_root_route(
     router = app.router
     request = _mock_request(method="GET", path="/")
 
-    async def run_url_dispatcher_benchmark() -> None:
+    async def run_url_dispatcher_benchmark() -> Optional[web.UrlMappingMatchInfo]:
+        ret = None
         for _ in range(resolve_count):
             ret = await router.resolve(request)
 
         return ret
 
     ret = loop.run_until_complete(run_url_dispatcher_benchmark())
+    assert ret is not None
     assert ret.get_info()["path"] == "/", ret.get_info()
 
     @benchmark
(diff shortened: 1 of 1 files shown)


asvetlov:
I think the speed decrease is acceptable.
Checks add a constant execution time, it's ok.
Better to have checked results; it prevents silly errors like I made for `test_resolve_gitapi_subapps`

bdraco:
I'd probably write it like this to avoid
the assert inside the benchmark

```diff
diff --git a/tests/test_benchmarks_web_urldispatcher.py b/tests/test_benchmarks_web_urldispatcher.py
index bb6cf3aa3..ef1f2c4e8 100644
--- a/tests/test_benchmarks_web_urldispatcher.py
+++ b/tests/test_benchmarks_web_urldispatcher.py
@@ -53,11 +53,17 @@ def test_resolve_root_route(
     router = app.router
     request = _mock_request(method="GET", path="/")
 
-    async def run_url_dispatcher_benchmark() -> None:
+    async def run_url_dispatcher_assertions() -> None:
         for _ in range(resolve_count):
             ret = await router.resolve(request)
             assert ret.get_info()["path"] == "/", ret.get_info()
 
+    loop.run_until_complete(run_url_dispatcher_assertions())
+
+    async def run_url_dispatcher_benchmark() -> None:
+        for _ in range(resolve_count):
+            await router.resolve(request)
+
     @benchmark
     def _run() -> None:
         loop.run_until_complete(run_url_dispatcher_benchmark())
```

Sorry, no inline suggestion as I'm in flight and WiFi is like a 9600 baud modem

Dreamsorcerer:
> Checks add a constant execution time, it's ok.

Hopefully bdraco's suggestion resolves the issue. Otherwise, the drop in test_resolve_gitapi_subapps[pyloop] looks too big. If 97% of the time is spent on the new asserts, we'd be unable to detect any improvements to the benchmark.

asvetlov:
No, for GitHub subapps I'fixed the benchmark. 
Previously it returned 404 Not Found instead of 200 Ok, fail path was faster than lookup into the subapp

asvetlov:
@bdraco I've got your idea. 
I'll rewrite benchmarks to perform assertions outside of measured code.
Thanks for the idea!

asvetlov:
Now the only `test_resolve_gitapi_subapps` is about two times slower because I've fixed this incorrect benchmark.

bdraco (approved):

asvetlov:
@bdraco do you sleep sometimes? :)

bdraco:
Not so much at the end of the year as its always travel heavy for me and I'm flipping time zones every few days.

PR #9961, opened 2024-11-18
Disable Python 3.14 builds

asvetlov:
Now aiohttp cannot be *installed* on Python 3.14 because the lack of dependencies support.
It leads to non-required but annoying red cross label for every CI build.

Let's skip the Python 3.14 build matrix until 3.14rc will be published, aiohttp dependencies compile py-3.14 compatible distributions, and all will be ready for migration.

Right now I'm disappointing to see red cross for every CI run, I should open it to figure out is it a critical failure or py-3.14 problem that could be ignored.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -155,10 +155,10 @@ jobs:
             no-extensions: 'Y'
             os: ubuntu
             experimental: false
-          - os: ubuntu
-            pyver: "3.14"
-            experimental: true
-            no-extensions: 'Y'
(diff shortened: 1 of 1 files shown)


bdraco (approved):
I think it makes sense to turn this off for now as its going to be months before it will pass and in the mean time is using up another CI runner and can never cache.

Someone else might feel strongly about keeping it though so I'd give it a bit more time before merging.

asvetlov:
The waiting period is timed out.
If anybody objects, we can just uncomment 4 lines in the workflow yaml file.

PR #9962, opened 2024-11-18
Add target-version to black config

cdce8p:
## What do these changes do?
Add `target-version` option to black config in `pyproject.toml` and reformat code.
https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -46,7 +46,7 @@ class AiohttpClient(Protocol):
         __param: Application,
         *,
         server_kwargs: Optional[Dict[str, Any]] = None,
-        **kwargs: Any
+        **kwargs: Any,
     ) -> TestClient[Request, Application]: ...
     @overload
     async def __call__(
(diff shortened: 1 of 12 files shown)


bdraco (approved):

bdraco:
If we don't backport, it will increase the chance of conflicts for future backports

cdce8p:
> If we don't backport, it will increase the chance of conflicts for future backports 

Should I do the backports manually tomorrow. I doubt the automatic one would apply cleanly.

_Older versions also support older Python versions which in turn might effect the available formatting._

bdraco:
> > If we don't backport, it will increase the chance of conflicts for future backports
> 
> Should I do the backports manually tomorrow? I doubt the automatic one would apply cleanly.
> 
> _Older versions also support older Python versions which in turn might effect the available formatting._

Will think about the best way overnight.  Too much going on right not to work it out.

asvetlov (approved):

bdraco:
> > > If we don't backport, it will increase the chance of conflicts for future backports
> > 
> > 
> > Should I do the backports manually tomorrow? I doubt the automatic one would apply cleanly.
> > _Older versions also support older Python versions which in turn might effect the available formatting._
> 
> Will think about the best way overnight. Too much going on right now to work it out.

Probably good to do the manual cherry picks above, and adjust the target versions before pushing, and let the pre-commit bot clean it up.

bdraco:
3.10 is the only version that supports Python 3.8 still.

We will likely only publish another 3.10 if there is a severe regression. At this point it might be dead.

PR #9963, opened 2024-11-18
[3.11] Restore `FlowControlDataQueue` class

bdraco:
We no longer use `FlowControlDataQueue` internally as it was removed in #9793. It was assumed (incorrectly) it was not used externally. At least `aiodocker` uses it. There may be others so we will restore it for 3.11.

Please note that it will go away permanently in 4.x so its will probably be best to vendor the code in this PR if you will still need this class when 4.x is released.

fixes https://github.com/aio-libs/aiodocker/issues/918


bdraco:
I should have done this for 3.12 as well as the timeline for dropping it will be 4.x

bdraco:
Thankfully it forward ports cleanly

PR #9970, opened 2024-11-19
[PR #9895/d8ec1b4 backport][3.12] Defer creation of SimpleCookie objects in the web server until needed

bdraco:
(cherry picked from commit d8ec1b42a845460cc3966673d40130939fef2ca7)


bdraco:
backporting from this PR instead of `master` since its completely different on `master` as its been split into mixins and it will backport cleanly from here

PR #9972, opened 2024-11-19
Revert "Only construct the `allowed_methods` `set` once for a `StaticResource`"

bdraco:
Reverts aio-libs/aiohttp#9911

`aiohttp-cors` accesses the protected `self._routes` here and relies on being able to add new methods
https://github.com/aio-libs/aiohttp-cors/blob/38c6c17bffc805e46baccd7be1b4fd8c69d95dc3/aiohttp_cors/urldispatcher_router_adapter.py#L187

### DIFF
--- a/CHANGES/9972.bugfix.rst
+++ b/CHANGES/9972.bugfix.rst
@@ -1,0 +1,3 @@
+Reverted an optimization to avoid rebuilding the allowed methods for `StaticResource` on every request -- by :user:`bdraco`.
+
+``aiohttp-cors`` needs to be able to modify the allowed methods at run time via this internal.
--- a/CHANGES/9972.bugfix.rst
+++ b/CHANGES/9972.bugfix.rst
@@ -1,3 +1,3 @@
-Reverted an optimization to avoid rebuilding the allowed methods for `StaticResource` on every request -- by :user:`bdraco`.
+Reverted an optimization to avoid rebuilding the allowed methods for ``StaticResource`` on every request -- by :user:`bdraco`.
 
 ``aiohttp-cors`` needs to be able to modify the allowed methods at run time via this internal.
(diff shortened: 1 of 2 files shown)


CHANGES/9972.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Reverted an optimization to avoid rebuilding the allowed methods for `StaticResource` on every request -- by :user:`bdraco`.
```
bdraco:
```suggestion
Reverted an optimization to avoid rebuilding the allowed methods for ``StaticResource`` on every request -- by :user:`bdraco`.
```

aiohttp/web_urldispatcher.py
```diff
@@ -621,10 +620,10 @@ def set_options_route(self, handler: Handler) -> None:
     async def resolve(self, request: Request) -> _Resolve:
         path = request.rel_url.path_safe
         method = request.method
+        allowed_methods = set(self._routes)
         if not path.startswith(self._prefix2) and path != self._prefix:
             return None, set()
 
-        allowed_methods = self._allowed_methods
         if method not in allowed_methods:
```
bdraco, line 627:
```suggestion
        allowed_methods = set(self._routes)
        if method not in allowed_methods:
```

aiohttp/web_urldispatcher.py
```diff
@@ -621,10 +620,10 @@ def set_options_route(self, handler: Handler) -> None:
     async def resolve(self, request: Request) -> _Resolve:
         path = request.rel_url.path_safe
         method = request.method
+        allowed_methods = set(self._routes)
```
```suggestion
```

Dreamsorcerer:
> `aiohttp-cors` accesses the protected `self._routes` here and relies on being able to add new methods

We could just update that library?

bdraco:
Created an issue for that https://github.com/aio-libs/aiohttp-cors/issues/483

PR #9975, opened 2024-11-19
Add test coverage for accessing StaticResource._routes

bdraco:
#9911 optimized this to only build the set once, however ``aiohttp-cors`` relies on being able to access `StaticResouce._routes`
https://github.com/aio-libs/aiohttp-cors/blob/38c6c17bffc805e46baccd7be1b4fd8c69d95dc3/aiohttp_cors/urldispatcher_router_adapter.py#L187 so it was reverted in #9972

Since this requirement is unexpected, and therefore hard to discover, add a test for it.

### DIFF
--- a/tests/test_urldispatch.py
+++ b/tests/test_urldispatch.py
@@ -520,7 +520,7 @@ async def test_static_not_match(router: web.UrlDispatcher) -> None:
 async def test_add_static_access_resources(router: web.UrlDispatcher) -> None:
     """Test accessing resource._routes externally.
 
-    aiohttp-cors access the resource._routes, this test ensures that this
+    aiohttp-cors accesses the resource._routes, this test ensures that this
     continues to work.
     """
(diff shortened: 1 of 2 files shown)


tests/test_urldispatch.py
```diff
@@ -517,6 +517,41 @@ async def test_static_not_match(router: web.UrlDispatcher) -> None:
     assert (None, set()) == ret
 
 
+async def test_add_static_access_resources(router: web.UrlDispatcher) -> None:
+    """Test accessing resource._routes externally.
+
+    aiohttp-cors access the resource._routes, this test ensures that this
```
bdraco:
```suggestion
    aiohttp-cors accesses the resource._routes, this test ensures that this
```

bdraco:
Probably a better fix is to keep the change in #9911 and add OPTIONS on `set_options_route`, but lets add coverage first, and than do that in another PR

PR #9987, opened 2024-11-19
Fix ``ANY`` method not appearing in the ``UrlDispatcher`` routes

bdraco:
Regressed in #9899

The pseudo `ANY` method should still get added to `self._routes` and `self._allowed_methods` because external callers expect to inspect it.

Discovered in https://github.com/aio-libs/aiohttp-apischema/pull/44


CHANGES/9987.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed the ``ANY`` method not appearing in meth:`~aiohttp.web.UrlDispatcher.routes` -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed the ``ANY`` method not appearing in :meth:`~aiohttp.web.UrlDispatcher.routes` -- by :user:`bdraco`.
```

PR #10003, opened 2024-11-20
Fix client connection header not reflecting connector `force_close` value

bdraco:
The `Connection` header should be set based on the `force_close` value in the connector.

It appears that the code to set the connection header had no effect because it calls `keep_alive` which checks the existing connection header and uses the result to set the headers in a block that only ever called when the connection header is not set.

AFIACT its not worked as expected since https://github.com/aio-libs/aiohttp/commit/8e6723afb5513dea58b2f53e4a8d0071430e3e83 where the `test_http10_keep_alive_default` test was marked as `xfail` and `HTTP/1.0` connections never set the `keep-alive` header but still expected to reuse the connection. Additionally if `force_close` was set on the connector we never told the remote we were going to close the connection by setting the connection header to `close` for the `HTTP/1.1` case.

The code was only being exercised because the test was mocking out `keep_alive` to impossible values that disagreed with the block in `send` that was only entered when the connection header was not set.

https://github.com/aio-libs/aiohttp/blob/af33a8224203b996190c4782b65446849f2be445/aiohttp/client_reqrep.py#L575

It's likely not many people are using `force_close` or setting the http version to 1.0 so the impact is minimal. On the plus side it makes sending requests 8% faster since it's not having to do all the useless checks now

### DIFF
--- a/CHANGES/10003.bugfix.rst
+++ b/CHANGES/10003.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the HTTP client not considering the connector's ``force_close`` value when setting the ``Connection`` header -- by :user:`bdraco`.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -56,13 +56,7 @@ from .helpers import (
     set_exception,
     set_result,
 )
-from .http import (
-    SERVER_SOFTWARE,
-    HttpVersion,
-    HttpVersion10,
-    HttpVersion11,
-    StreamWriter,
-)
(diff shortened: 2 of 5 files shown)


bdraco:
Slept on it. Still think it can go. Need to do some git archeology to see why this was added and how we got here in case I'm missing something

bdraco:
https://github.com/aio-libs/aiohttp/commit/8e6723afb5513dea58b2f53e4a8d0071430e3e83

bdraco:
I think this was actually supposed to set the `keep-alive` header on http/1.0

PR #10004, opened 2024-11-20
Avoid building `skip_headers` in `ClientSession._request` if it will be thrown away

bdraco:
There is a very good chance that we will build `skip_headers` and than throw it away as this option is not commonly passed.

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -490,10 +490,12 @@ class ClientSession:
         if url.scheme not in self._connector.allowed_protocol_schema_set:
             raise NonHttpUrlClientError(url)
 
-        skip_headers: Optional[set[str]] = None
-        skip_headers = (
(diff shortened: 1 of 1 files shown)


aiohttp/client.py
```diff
@@ -490,10 +491,13 @@ async def _request(
...
             raise NonHttpUrlClientError(url)
 
-        skip_headers = set(self._skip_auto_headers)
+        skip_headers: Optional[Union[frozenset[istr], set[istr]]] = None
         if skip_auto_headers is not None:
-            for i in skip_auto_headers:
-                skip_headers.add(istr(i))
+            skip_headers = {istr(i) for i in skip_auto_headers}
+            if self._skip_auto_headers is not None:
+                skip_headers.update(self._skip_auto_headers)
```
bdraco: Looks like we don't have coverage for passing `skip_auto_headers` and also setting it on the session

PR #10018, opened 2024-11-21
Add url dispatcher benchmark for resolving root route for github simulated routes tree

asvetlov:
More benchmarks is better!

### DIFF
--- a/tests/test_benchmarks_web_urldispatcher.py
+++ b/tests/test_benchmarks_web_urldispatcher.py
@@ -6,7 +6,7 @@ import pathlib
 import random
 import string
 from pathlib import Path
-from typing import NoReturn, Optional
+from typing import NoReturn, Optional, cast
 from unittest import mock
 
 import pytest
@@ -30,7 +30,7 @@ def github_urls() -> list[str]:
     with (here / "github-urls.json").open() as f:
         urls = json.load(f)
 
-    return urls
+    return cast(list[str], urls)
 
 
 def _mock_request(method: str, path: str) -> web.Request:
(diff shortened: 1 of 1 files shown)


bdraco (approved):

bdraco:
there is a mypy nit
```
tests/test_benchmarks_web_urldispatcher.py:33:5: error: Returning Any from
function declared to return "list[str]"  [no-any-return]
        return urls
```

asvetlov:
Yep, I see. New commit should fix it.

bdraco:
Need to merge https://github.com/aio-libs/aiohttp/pull/10020 first.  Will take care of merging this one as soon as 10020 finishes

tests/test_benchmarks_web_urldispatcher.py
```diff
@@ -18,6 +19,20 @@
...
+    """GitHub api urls."""
+    # The fixture provides OpenAPI generated info for github.
+    # To update the local data file please run the following command:
+    # $ curl https://raw.githubusercontent.com/github/rest-api-description/refs/heads/main/descriptions/api.github.com/api.github.com.json | jq ".paths | keys" > github-urls.json
+
+    here = Path(__file__).parent
+    with (here / "github-urls.json").open() as f:
+        urls = json.load(f)
+
+    return cast(list[str], urls)
```
Dreamsorcerer, line 33: We shouldn't cast, it bypasses any type safety that may be present and could produce problems later on. Here we can just do:
```
        urls: list[str] = json.load(f)
```

PR #10030, opened 2024-11-24
Reduce initialization overhead in `ClientResponse`

bdraco:
Small speed up
Screenshot 2024-11-23 at 8 37 12 PM

When combined with https://github.com/aio-libs/aiohttp/pull/10029, its ~40% faster making `ClientResponse` objects

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -794,7 +794,8 @@ class ClientResponse(HeadersMixin):
         loop: asyncio.AbstractEventLoop,
         session: "ClientSession",
     ) -> None:
-        assert type(url) is URL, f"url is a {url!r}, not a yarl.URL"
+        if type(url) is not URL:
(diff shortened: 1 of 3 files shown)


aiohttp/client_reqrep.py
```diff
@@ -781,39 +790,34 @@ def __init__(
         continue100: Optional["asyncio.Future[bool]"],
         timer: Optional[BaseTimerContext],
         request_info: RequestInfo,
-        traces: List["Trace"],
+        traces: Optional[List["Trace"]],
```
bdraco: Should keep the signature the same in 3.11 since it's technically breaking to change it. Wait for 3.12 for this part and split to a new pr

bdraco: revert change in https://github.com/aio-libs/aiohttp/pull/10030/commits/91ca7bffda1c16c6e890afd718821abf12eec513

PR #10035, opened 2024-11-24
Small cleanup to the connection check in ``ClientResponse``

bdraco:
This should have been an explict is not None check as there is a __bool__ method on the object.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -1155,8 +1155,8 @@ class ClientResponse(HeadersMixin):
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        if (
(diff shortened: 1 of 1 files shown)


aiohttp/client_reqrep.py
```diff
@@ -1152,8 +1155,11 @@ async def read(self) -> bytes:
         elif self._released:  # Response explicitly released
             raise ClientConnectionError("Connection closed")
 
-        protocol = self._connection and self._connection.protocol
-        if protocol is None or not protocol.upgraded:
+        if (
+            self._connection is None
+            or (protocol := self._connection.protocol) is None
+            or not protocol.upgraded
```
```suggestion
            self._connection is not None and ((protocol := self._connection.protocol) is None or not protocol.upgraded)
```

PR #10036, opened 2024-11-24
DNM: Improve performance of starting client responses

bdraco:


bdraco:
Only makes a difference on the Py version. Cython version is already fast

PR #10037, opened 2024-11-24
Refactor requests and responses to use classvar defaults to avoid multiple `__init__`s

bdraco:
I was looking at the benchmarks and was wondering why 3.x was faster creating many objects. This PR re-aligns `master` to be similar to how these are implemented in 3.x as it was before https://github.com/aio-libs/aiohttp/pull/3942

Refactor mixins to avoid having to call `__init__` by using classvar defaults instead.

The downside is we have to drop `__slots__` but the trade-off seems worth it based on the benchmark.

### DIFF
--- a/CHANGES/10037.misc.rst
+++ b/CHANGES/10037.misc.rst
@@ -1,0 +1,1 @@
+Improved performances of creating objects during the HTTP request lifecycle -- by :user:`bdraco`.
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -795,7 +795,6 @@ class ClientResponse(HeadersMixin):
     ) -> None:
         # URL forbids subclasses, so a simple type check is enough.
         assert type(url) is URL
-        super().__init__()
 
         self.method = method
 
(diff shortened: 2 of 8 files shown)


bdraco:
Wasn't expecting that kind of improvement. I would say the trade off is worth it especially considering we can do it in a few more places

asvetlov:
I'm neutral to this change.
Such microoptimizations are not visible on macrobenchmarks.
Also, next Python version could change the picture.

bdraco:
Looks to be ~2.7% speed up in testing

before
Screenshot 2024-11-25 at 9 24 21 AM

after
Screenshot 2024-11-25 at 9 23 45 AM

asvetlov:
Wow! 2.5% is a high number, I didn't expect this.

bdraco:
Part of the issue with the benchmarks is that we have to use `loop.run_until_complete` which means we benchmark all the asyncio overhead and anything running in the loop (we get both the server and client side) so the benchmark only tends to pickup very significant changes as codspeed is also set to only report changes of 8%+ or more.  Which for a client side benchmark is really 8% of asyncio overhead, server request, and the actual client code.

https://github.com/aio-libs/aiohttp/blob/006fbc376bf8eac9cbe54a902b09e1975bd1ff41/tests/test_benchmarks_client.py#L33

PR #10038, opened 2024-11-24
Small speed up to `StreamWriter.__init__`

bdraco:
Use classvar defaults to avoid `__init__` overhead. `AbstractStreamWriter` already does this so we can drop a few duplicates as well.

Screenshot 2024-11-24 at 4 58 02 PM

### DIFF
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -38,6 +38,15 @@ _T_OnHeadersSent = Optional[Callable[["CIMultiDict[str]"], Awaitable[None]]]
 
 
 class StreamWriter(AbstractStreamWriter):
+
+    length: Optional[int] = None
+    chunked: bool = False
+    buffer_size: int = 0
+    output_size: int = 0
+    _eof: bool = False
+    _compress: Optional[ZLibCompressor] = None
+    _drain_waiter: Optional[asyncio.Future[None]] = None
+
     def __init__(
         self,
         protocol: BaseProtocol,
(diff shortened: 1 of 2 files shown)


aiohttp/http_writer.py
```diff
@@ -38,6 +38,13 @@ class HttpVersion(NamedTuple):
 
 
 class StreamWriter(AbstractStreamWriter):
+
+    length: Optional[int] = None
+    chunked: bool = False
+    _eof: bool = False
+    _compress: Optional[ZLibCompressor] = None
+    _drain_waiter: Optional[asyncio.Future[None]] = None
```
bdraco:
```suggestion
```
never used

aiohttp/http_writer.py
```diff
@@ -46,17 +52,7 @@ def __init__(
...
-
         self.loop = loop
-        self.length = None
-        self.chunked = False
-        self.buffer_size = 0
-        self.output_size = 0
-
-        self._eof = False
-        self._compress: Optional[ZLibCompressor] = None
-        self._drain_waiter = None
```
bdraco, line 58: This was never used.

aiohttp/http_writer.py
```diff
@@ -46,17 +52,7 @@ def __init__(
         on_headers_sent: _T_OnHeadersSent = None,
     ) -> None:
         self._protocol = protocol
-
         self.loop = loop
-        self.length = None
-        self.chunked = False
-        self.buffer_size = 0
-        self.output_size = 0
```
bdraco, line 54: Already set in `AbstractStreamWriter`

PR #10049, opened 2024-11-26
Improve client performance when there are no auto headers to skip

bdraco:
Skipping auto headers is a rarely used feature.

- Avoid making an empty `CIMultiDict` each request that we will throw away
- Avoid all the contains checks when there are no auto headers to skip
- Some unreachable branching has been removed from `update_body_from_data`

Screenshot 2024-11-26 at 4 20 04 PM

Screenshot 2024-11-26 at 4 10 07 PM

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -209,7 +209,7 @@ class ClientRequest:
     __writer: Optional["asyncio.Task[None]"] = None  # async task for streaming data
     _continue = None  # waiter future for '100 Continue' response
 
-    skip_auto_headers: Optional["CIMultiDict[None]"] = None
+    _skip_auto_headers: Optional["CIMultiDict[None]"] = None
 
     # N.B.
     # Adding __del__ method with self._writer closing doesn't make sense
@@ -294,6 +294,10 @@ class ClientRequest:
 
     def __reset_writer(self, _: object = None) -> None:
         self.__writer = None
+
+    @property
+    def skip_auto_headers(self) -> CIMultiDict[None]:
+        return self._skip_auto_headers or CIMultiDict()
 
     @property
     def _writer(self) -> Optional["asyncio.Task[None]"]:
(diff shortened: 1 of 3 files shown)


aiohttp/client_reqrep.py
```diff
@@ -522,19 +523,18 @@ def update_body_from_data(self, body: Any) -> None:
         self.body = body
 
         # enable chunked encoding if needed
-        if not self.chunked:
-            if hdrs.CONTENT_LENGTH not in self.headers:
-                size = body.size
-                if size is None:
-                    self.chunked = True
-                else:
-                    if hdrs.CONTENT_LENGTH not in self.headers:
```
bdraco, line 531: This check was done above so it would always be true, which allows this code block to be condensed

aiohttp/client_reqrep.py
```diff
@@ -412,12 +414,11 @@ def update_auto_headers(self, skip_auto_headers: Optional[Iterable[str]]) -> Non
         else:
             # Fast path when there are no headers to skip
             # which is the most common case.
-            self.skip_auto_headers = CIMultiDict()
             used_headers = self.headers
 
         for hdr, val in self.DEFAULT_HEADERS.items():
             if hdr not in used_headers:
-                self.headers.add(hdr, val)
+                self.headers[hdr] = val
```
bdraco, line 425: We are already checking if the header exists in `used_headers` which is the combination of `self.headers` and `skip_auto_headers` so we can use `[]` instead of `.add()` which is a bit faster

aiohttp/client_reqrep.py
```diff
@@ -293,6 +295,10 @@ def __init__(
     def __reset_writer(self, _: object = None) -> None:
         self.__writer = None
 
+    @property
+    def skip_auto_headers(self) -> CIMultiDict[None]:
+        return self._skip_auto_headers or CIMultiDict()
```
bdraco, line 300: This is documented as a public value

bdraco, line 300: It would be very unexpected that someone would want to get this value back after passing it in, but it was available so to avoid breaking the API it remains exposed.

PR #10050, opened 2024-11-26
DNM: Avoid checking for control chars in the method if its alpha numeric

bdraco:


bdraco:
regex isn't that expensive that it matters

PR #10055, opened 2024-11-27
Downgrade logging of invalid HTTP methods on first request to debug level

bdraco:
Its common when a client sends non-HTTP or encrypted traffic to an HTTP port for the server to encounter an error parsing the request. This is expected to happen when connected to the public internet (rather frequently) as browsers may try to speak SSL before plain-text. There are a variety of other reasons as well, such users mis-typing the port, bad redirects, etc which generates non HTTP traffic on the first request.

To avoid fill logs with noise, a new exception `BadHttpMethod`, which inherits from `BadStatusLine` is now raised internally to distinguish between status lines with invalid versions or status and ones one that start with garbage or an invalid method.

The new exception is logged at debug level on the first request only to avoid filling the log with noise when nothing is actually wrong as the vast majority of these are noise. Bad methods received on keep alive requests will continue to be logged at exception level.

Example of how common spurious errors are downstream: https://github.com/home-assistant/core/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aclosed+BadStatusLine https://github.com/home-assistant/core/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen+BadStatusLine

fixes #8065
fixes #8442
fixes #3287

PR #10067, opened 2024-11-27
Add new `NotHttpProtocol` exception to differentiate invalid methods and gibberish

bdraco:
Followup to https://github.com/aio-libs/aiohttp/pull/10055#issuecomment-2504524644 to better differentiate invalid methods and gibberish

### DIFF
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -834,7 +834,7 @@ cdef parser_error_from_errno(cparser.llhttp_t* parser, data, pointer):
                  cparser.HPE_INVALID_TRANSFER_ENCODING}:
         return BadHttpMessage(err_msg)
     elif errno == cparser.HPE_INVALID_METHOD:
-        if not data.partition(b" ")[0].isalpha():
+        if not data.partition(b" ")[0]:
             return NotHttpProtocol(err_msg)
         return BadHttpMethod(error=err_msg)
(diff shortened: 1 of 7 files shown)


bdraco:
I want to look back at all the HA issues and be sure to add some more tests for other types of gibberish

aiohttp/_http_parser.pyx
```diff
@@ -833,6 +834,8 @@ cdef parser_error_from_errno(cparser.llhttp_t* parser, data, pointer):
                  cparser.HPE_INVALID_TRANSFER_ENCODING}:
         return BadHttpMessage(err_msg)
     elif errno == cparser.HPE_INVALID_METHOD:
+        if b" " not in data:
+            return NotHttpProtocol(err_msg)
```
andersk, line 838: This seems unreliable since a binary protocol can easily include `b" "` by accident.

A TLS handshake always begins with `b"\x16"` ([RFC 8446 §5.1](https://datatracker.ietf.org/doc/html/rfc8446#section-5.1)); we could use that to reliably detect HTTPS at least.

bdraco: Yeah its a bit messy to guess here... maybe we should just leave it as-is.

bdraco: I think its better to leave it as-is and not try to differentiate as llhttp is always going to be able to do a better job than a heuristic as otherwise we end up with a whole list of possible exceptions for TLS or other garbage

PR #10072, opened 2024-11-29
DNM: Add a Cython implementation for ConnectionKey with hash caching

bdraco:
I noticed on the profiles that we spend a lot of time in `connection.py` in functions that hash and compare the `ConnectionKey`.  We moved to a `NamedTuple` to speed up the hashing but its still looks like its taking up quite
a bit of overhead because we use it as a key in so many operations.

Add a Cython implementation which caches the hash to see if there is a measurable improvement.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -43,6 +43,7 @@ from .formdata import FormData
 from .hdrs import CONTENT_TYPE
 from .helpers import (
     _SENTINEL,
+    NO_EXTENSIONS,
     BaseTimerContext,
     BasicAuth,
     HeadersMixin,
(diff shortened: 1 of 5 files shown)


aiohttp/client_reqrep.py
```diff
@@ -1239,3 +1240,16 @@ async def __aexit__(
         # if state is broken
         self.release()
         await self.wait_for_close()
+
+
+RawResponseMessagePy = ConnectionKey
```
```suggestion
ConnectionKeyPy = ConnectionKey
```

bdraco:
Doesn't make a measurable difference.

bdraco:
o wait. I don't think its actually worked

bdraco:
0.3% faster.... not worth it

PR #10073, opened 2024-11-29
Improve performance of parsing headers

bdraco:
After we have a complete header we always have to convert it to a new bytes object. For almost every case we get the header in a single read. In this case its much faster to convert to bytes once. If the header gets split across multiple reads it will be a bit slower but this is the uncommon case.

Screenshot 2024-12-01 at 3 13 50 PM

### DIFF
--- a/aiohttp/_http_parser.pyx
+++ b/aiohttp/_http_parser.pyx
@@ -97,11 +97,8 @@ cdef inline str http_method_str(int i):
     else:
         return "<unknown>"
 
-cdef inline object find_header(bytes raw_header):
-    cdef Py_ssize_t size
-    cdef char *buf
+cdef inline object find_header(char* buf, Py_ssize_t size, bytes raw_header):
     cdef int idx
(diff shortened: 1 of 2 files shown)


aiohttp/_http_parser.pyx
```diff
@@ -378,42 +379,35 @@ cdef class HttpParser:
...
-
-            name = find_header(raw_name)
-            value = raw_value.decode('utf-8', 'surrogateescape')
+        cdef str value
+        if self._raw_name is not EMPTY_BYTES:
+            name = find_header(self._raw_name)
+            value = self._raw_value.decode('utf-8', 'surrogateescape')
 
-            self._headers.add(name, value)
+            self._headers.append((name, value))
```
bdraco, line 388: Its a bit faster to build up a list of tuples and have `CIMultiDict` build from that instead of calling `.add()` over and over since we can stay in Cython without having to jump back to python code for every header since the only time we have to end up in Python is to pass the list to `CIMultiDict`

PR #10075, opened 2024-11-30
DNM: Migrate stream to use classvar defaults

bdraco:
same change as #10054 for streams


bdraco:
+0.04% .

not worth the change

Issue #3287, opened 2018-09-24
Add support for custom handling of requests that throw in `aiohttp.http_exceptions.*`


bazbar:
## Long story short
Not sure if this is a user error, or a bug.

On a bad request, the server raise `aiohttp.http_exceptions.BadStatusLine: invalid HTTP method`, and custom headers are ignored. I can't find any way to catch this exception.

My application works as it should, and custom headers are sent on both GET and POST, and also HEAD - using the  on_prepare signal.

I want to use custom headers and hide the `Server: Python/3.6 aiohttp/3.4.4`. But a simple request containing only "/" bypass all handlers, and display the default headers. And also raise a seemingly non-handable exception.

## Expected behaviour

I'd expect being able to handle all requests, no matter how erranous.

## Actual behaviour

In console, AioHttp drops
```
HTTP/1.0 400 Bad Request
Content-Type: text/html; charset=utf-8
Content-Length: 11
Date: Mon, 24 Sep 2018 14:13:37 GMT
Server: Python/3.6 aiohttp/3.4.4

Bad Request 
```
and in logs:
```
Traceback (most recent call last):
  File "site-packages\aiohttp\web_protocol.py", line 242, in data_received
  File "aiohttp\_http_parser.pyx", line 523, in aiohttp._http_parser.HttpParser.feed_data
aiohttp.http_exceptions.BadStatusLine: invalid HTTP method
```

## Steps to reproduce

1. Start a barebone server (like in Getting started)
2. telnet 127.0.0.1 8080, curl should work as well
3. type "/"

## Your environment
Windows
Python/3.6 aiohttp/3.4.4
aiohttp server

asvetlov:
[GitMate.io](https://gitmate.io) thinks possibly related issues are https://github.com/aio-libs/aiohttp/issues/2632 (aiohttp.http_exceptions.BadStatusLine: invalid HTTP method), https://github.com/aio-libs/aiohttp/issues/1619 (aiohttp.Client can't parse Trailer headers and throws exception), https://github.com/aio-libs/aiohttp/issues/1792 (aiohttp.http_exceptions.BadStatusLine caused by extra CRLF after POST data), https://github.com/aio-libs/aiohttp/issues/58 (aiohttp.HttpClient), and https://github.com/aio-libs/aiohttp/issues/1390 (Build chat server with aiohttp).

asvetlov:
The best we can do here is dropping connection without any output (but logging the fact).
[...]

PR #10076, opened 2024-11-30
Fix request count check for BadHttpMethod

bdraco:
The error handler was incorrectly checking for the first request of the server instead of the first request for the connection

https://github.com/aio-libs/aiohttp/pull/10055#discussion_r1864316780


aiohttp/web_protocol.py
```diff
@@ -212,6 +212,9 @@ def __init__(
     ):
         super().__init__(loop)
 
+        # _request_count is the number of request processed by the handler
+        # A new handler is created for each connection, so it is the number
+        # of requests processed by the connection.
```
Dreamsorcerer: I find this slightly confusing. I'd probably avoid the word handler here as I'm thinking of the endpoint handler (i.e. the function registered on a route) which is presumably not what you mean.

```suggestion
        # _request_count is the number of requests processed with the same connection.
```

bdraco: I simplified the comment. Hopefully it makes more sense now

Dreamsorcerer (commented):
I'm still not convinced this will behave as desired. Clients that don't use keep alive (whether unsupported, or a proxy doesn't allow it etc.) will never result in the issue getting logged.
While anyone using a reverse proxy like nginx will likely gets loads of logs as the "client" is the proxy server, not the connecting user.

bdraco:
> I'm still not convinced this will behave as desired. Clients that don't use keep alive (whether unsupported, or a proxy doesn't allow it etc.) will never result in the issue getting logged. While anyone using a reverse proxy like nginx will likely gets loads of logs as the "client" is the proxy server, not the connecting user.

If I remember correctly `nginx` has keep-alive disabled by default for reverse proxying..maybe that has changed https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

I'm not sure what you are suggesting for the path forward.

Dreamsorcerer:
> > I'm still not convinced this will behave as desired. Clients that don't use keep alive (whether unsupported, or a proxy doesn't allow it etc.) will never result in the issue getting logged. While anyone using a reverse proxy like nginx will likely gets loads of logs as the "client" is the proxy server, not the connecting user.
> 
> If I remember correctly `nginx` has keep-alive disabled by default for reverse proxying..maybe that has changed https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

I think it's still very common to enable keepalive. Regardless, when disabled we'll never log, when enabled we'll almost always log. Neither is the intended behaviour.

> I'm not sure what you are suggesting for the path forward.

Other than logging unconditionally as before, I don't have a better suggestion. Maybe when that other issue is fixed, users could customise this with a middleware?

Dreamsorcerer:
With a middleware they could choose to opt in to this behaviour, they could use the client IP from a trusted proxy header instead of per-connection etc.

bdraco:
> Other than logging unconditionally as before, I don't have a better suggestion. Maybe when that other issue is fixed, users could customise this with a middleware?

Yeah I think customization in https://github.com/aio-libs/aiohttp/issues/3287 probably the best way to handle this in the longer term. 

I think this PR is ok for the short term as its still logged at debug level so it should be discoverable if the user increases the logging level but considering this generates 100+ issues downstream when nothing is actually wrong I'd prefer to keep it at debug level as a log message isn't so valuable when the common case is noise.

Dreamsorcerer:
Yep, I think probably merge this for now as something reasonable.

I think #3287 will be a breaking change and won't get backported (which is why I put in on the v4 milestone). So, then we can drop the debug logs in v4 when that is ready.

PR #10088, opened 2024-12-02
Avoid calling len on the same data in the stream reader twice

bdraco:
This is trivial but it is noticeable in the profile.

before
before

after
after


bdraco:
I only found this because I was looking at function call counts..

I was trying to find why `_release_connection` is always called 4x per request

Screenshot 2024-12-01 at 6 57 37 PM

PR #10093, opened 2024-12-02
Use `quote_cookie` setting from ClientSession's cookiejar in `tmp_cookie_jar`

### LINKED ISSUE #9336: aiohttp ignoring cookie_jar quote_cookie parameter
### Describe the bug

```
sess = aiohttp.ClientSession(connector=conn, cookie_jar=aiohttp.CookieJar(quote_cookie=False)).request(...)
```
When using an client session object with a cookie jar with the cookie quoting routine disabled the requests sent still uses quoting routine when invoking the request:

[client_reqrep.py:468] ClientRequest().update_cookies():
```
def update_cookies(self, cookies: Optional[LooseCookies]) -> None:
        """Update request cookies header."""
        if not cookies:
            return

        c = SimpleCookie()
        .
        .
        .
        self.headers[hdrs.COOKIE] = c.output(header="", sep=";").strip()
```
The variable `c` used to output the cookie is a SimpleCookie object which defaults to using the quoting routine.
therefor when for example received a cookie with value such as `id=!IagQHEStKHLB3DFBAufSQIeK+Olw=` the header will output as `Cookie: id="!IagQHEStKHLB3DFBAufSQIeK+Olw="`.

A possible fix is to add a method to the CookieJar object like `is_quote_disabled()` and then in the update cookies use the `self.session.is_quoting_disabled()` to specifically handle the cookie in such case. i used:
```
self.headers[hdrs.COOKIE] = '; '.join([f"{name}={mrsl.value}" for name,mrsl in cookies.items()])
```

### To Reproduce
use the following script to reproduce:
```
import aiohttp
import asyncio

async def req():
    tcp_conn = aiohttp.TCPConnector(limit=10, force_close=True)
    c_jar = aiohttp.CookieJar(quote_cookie=False)
    sess = aiohttp.ClientSession(connector=tcp_conn, cookie_jar=c_jar)
    cookies = {"name": "val="}
    resp = await sess.request(url="http://localhost:8000/", method="GET", cookies=cookies)
    print(resp.request_info.headers.get("Cookie")) # name="val="
    await sess.close()

asyncio.run(req())
```

### Expected behavior

When specifying quote_cookie to False, cookie which by specifications are supposed to be quoted shouldn't be quoted because of the implicit flag.

### Logs/tracebacks

```python-traceback
-
```

### Python Version

```console
$ python --version
Python 3.9.6
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.10.5
```

### multidict Version

```console
Version: 6.1.0
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.11.1
```

### OS

Mac OS

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Cycloctane:
## What do these changes do?

Reuse `quote_cookie` setting from ClientSession's cookiejar when creating temporary CookieJar in request method.

This pr add a new property method to the abstract class of client's CookieJar.

## Related issue number

Fixes #9336 

## Checklist

    of this project.


aiohttp/client.py
```diff
@@ -601,7 +601,9 @@ async def _request(
                     all_cookies = self._cookie_jar.filter_cookies(url)
 
                     if cookies is not None:
-                        tmp_cookie_jar = CookieJar()
+                        tmp_cookie_jar = CookieJar(
+                            quote_cookie=self._cookie_jar.quote_cookie
```
Dreamsorcerer, line 603: Would it make sense to copy the other settings? unsafe and treat_as_secure_origin.

An alternative API might be to add a CookieJar.new() method or something, which simply returns a new CookieJar with the same settings.

Maybe someone else has an opinion on which approach to use...

Dreamsorcerer (approved):

aiohttp/cookiejar.py
```diff
@@ -471,6 +475,10 @@ def __iter__(self) -> "Iterator[Morsel[str]]":
     def __len__(self) -> int:
         return 0
 
+    @property
+    def quote_cookie(self) -> bool:
+        return True
```
bdraco, line 480: It looks like this line is missing coverage

Cycloctane: I'm not sure if `quote_cookie` in dummy cookie jar makes sense, for it does not handle any cookies. I let it return True simply because it returns `SimpleCookie` in `filter_cookies`.

bdraco (approved):

aiohttp/client.py
webknjaz: I think stdlib has a precedent for `.copy()`. Though, isn't there a `__copy__()`, too?

Dreamsorcerer: Yeah, it's not a copy though, in that we want an empty CookieJar with the same settings, not one with all the cookies copied over.

bdraco:
Doing a release shortly. Drafting while https://github.com/aio-libs/aiohttp/pull/10093#discussion_r1865912224 is being worked out.

aiohttp/client.py
Cycloctane: > I think stdlib has a precedent for `.copy()`. Though, isn't there a `__copy__()`, too?

CookieJar from ClienSession may be DummyCookieJar or user implemented CookieJar. If we use a copy of ClientSession's cookiejar to handle per-request cookie, cookies may not be processed as expected.

Dreamsorcerer: Right, sounds like we want to avoid copying the class, so probably stick with this approach then.

Dreamsorcerer:
Seems to be coverage of web_response.py:268 being lost by this change. Do we need to remove that check, or adjust a test..?

Dreamsorcerer:
Nevermind, probably a confusion with being out of sync with master, it's gone now. This is probably ready then.

bdraco (approved):

bdraco:
Screenshot 2024-12-17 at 8 18 22 AM

bdraco:
backport 3.12 https://github.com/aio-libs/aiohttp/pull/10176

Issue #8013, opened 2024-01-09
`FileResponse` has undefined behavior if the file is changed out from under it between the `stat` and `open` calls


bdraco:
This was discovered while fixing another issue in https://github.com/aio-libs/aiohttp/pull/8012#issuecomment-1883851654

This is a rare case so its not likely we need to prioritize fixing it but I'm sure this race has bitten someone with random unexplained failures who uses aiohttp to serve index or signature files that get changed out frequently.

### Describe the bug

We can fix this by `open`ing the file first, doing `fstat` to get the `stat`

### To Reproduce

replace the file on the FS between the `stat` and `open` call

### Expected behavior

The handler should still send the original file

### Logs/tracebacks

```python-traceback
n/a
```

### Python Version

```console
$ python --version
```

### aiohttp Version

```console
$ python -m pip show aiohttp
```

### multidict Version

```console
$ python -m pip show multidict
```

### yarl Version

```console
$ python -m pip show yarl
```

### OS

n/a

### Related component

Server

### Additional context

n/a

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

bdraco:
It looks like this is worth fixing as it will also give a nice performance boost by reducing one executor job per request https://github.com/aio-libs/aiohttp/pull/10095

steverep:
> It looks like this is worth fixing as it will also give a nice performance boost by reducing one executor job per request #10095

👍🏻 but I'd add a benchmark for a 304 response as well.  Those would end up with a penalty from any overhead opening the file and the extra executor job to close it (4xx responses would also be penalized but less a concern I guess).

bdraco:
> > It looks like this is worth fixing as it will also give a nice performance boost by reducing one executor job per request #10095
> 
> 👍🏻 but I'd add a benchmark for a 304 response as well. Those would end up with a penalty from any overhead opening the file and the extra executor job to close it (4xx responses would also be penalized but less a concern I guess).

I kept that in mind when I wrote the fix. The number of executor jobs remains the same for both a success and 304 response.  If you would like to add a benchmark for 304s as well in a fresh PR, that would be a nice addition to ensure we don't regress it in the future.

steverep:
[...]

PR #10095, opened 2024-12-03
Add a benchmark for web.FileResponse

bdraco:
We didn't have any benchmarks for file responses. From the benchmarks it turns out most of the time is creating and processing the executor jobs.

If we combine the stat into a job that returns the open fileobj it will likely be faster and solve https://github.com/aio-libs/aiohttp/issues/8013

### DIFF
--- a/tests/test_benchmarks_web_fileresponse.py
+++ b/tests/test_benchmarks_web_fileresponse.py
@@ -9,12 +9,12 @@ from aiohttp import web
 from aiohttp.pytest_plugin import AiohttpClient
 
 
-def test_simple_web_response(
+def test_simple_web_file_response(
     loop: asyncio.AbstractEventLoop,
     aiohttp_client: AiohttpClient,
     benchmark: BenchmarkFixture,
 ) -> None:
-    """Benchmark creating 100 simple web.Response."""
+    """Benchmark creating 100 simple web.FileResponse."""
     response_count = 100
     filepath = pathlib.Path(__file__).parent / "sample.txt"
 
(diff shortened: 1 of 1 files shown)


tests/test_benchmarks_web_fileresponse.py
```diff
@@ -0,0 +1,35 @@
+"""codspeed benchmarks for the web responses."""
```
bdraco:
```suggestion
"""codspeed benchmarks for the web file responses."""
```

tests/test_benchmarks_web_fileresponse.py
```diff
@@ -0,0 +1,35 @@
...
+import asyncio
+import pathlib
+
+from pytest_codspeed import BenchmarkFixture
+
+from aiohttp import web
+from aiohttp.pytest_plugin import AiohttpClient
+
+
+def test_simple_web_response(
```
bdraco:
```suggestion
def test_simple_web_file_response(
```

tests/test_benchmarks_web_fileresponse.py
```diff
@@ -0,0 +1,35 @@
...
+    loop: asyncio.AbstractEventLoop,
+    aiohttp_client: AiohttpClient,
+    benchmark: BenchmarkFixture,
+) -> None:
+    """Benchmark creating 100 simple web.Response."""
```
bdraco:
```suggestion
    """Benchmark creating 100 simple web.FileResponse."""
```

bdraco:
Probably need one for no sendfile to simulate over tls

bdraco:
On a side note, all the Python versions we support now implement sendfile fallback so its likely our sendfile fallback code is no longer needed.

PR #10099, opened 2024-12-04
Fix type of SSLContext for some static type checkers

AlanBogarin:
## What do these changes do?

Fix partially unknown bug in all references to SSLContext

No, it's only for type checkers

No, it only requires knowledge of run-time and type checking.

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


Dreamsorcerer:
We've been using this without issue for quite some time. You'll need to provide a reproducer demonstrating the bug you're claiming to fix.

webknjaz:
I don't even know if the try/except is needed these days. I don't know of a Python distribution being shipped without the `ssl` module by default post Python 2.7.9. Someone could disable it when compiling CPython but I don't know that we want to support such custom builds anyway.

Dreamsorcerer:
Yep, maybe it's time to remove it then.

asvetlov:
Python itself support the build without ssl.
Asyncio has such conditional imports also.
I think we should keep it until CPython decide to make ssl mandatory (as it happened for threading module).

asvetlov (approved):
OK, makes sense.
Looks like mypy has a different logic  than pyright, that's why we never encountered the error in out liter runs.

Dreamsorcerer:
It's a bit awkward, but looks like we'll have to go with it.

Please put the no cover comments back in first though, as you can see from the codecov comment above, this has caused a regression in coverage.

CHANGES/10099.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed unknown `SSLContext` symbol for type checkers
```
Dreamsorcerer:
```suggestion
Fixed type of ``SSLContext`` for some static type checkers (e.g. pyright).
```

bdraco:
```suggestion
Fixed unknown `SSLContext` symbol for type checkers -- by :user:`AlanBogarin`.
```

bdraco: It looks like this suggestion was marked as resolved without any context or commentary.

bdraco (changes requested):
Please see https://github.com/aio-libs/aiohttp/pull/10099#discussion_r1871966506

bdraco:
Usually we add credit to the end change note in the form. It's not clear if you are explicitly wanting to not be credited or its on oversight. 

```
 -- by :user:`AlanBogarin`.
```

aiohttp/client_exceptions.py
```diff
@@ -7,13 +7,17 @@
...
 
     SSLContext = ssl.SSLContext
-except ImportError:  # pragma: no cover
-    ssl = SSLContext = None  # type: ignore[assignment]
+else:
+    try:
+        import ssl
 
+        SSLContext = ssl.SSLContext
+    except ImportError:
```
Dreamsorcerer: Not sure if you missed my other comment, but still a coverage regression:
https://github.com/aio-libs/aiohttp/pull/10099#issuecomment-2520352679

AlanBogarin: I did (`# pragma: no cover`), Wouldn't it be better to remove the try/except block?

AlanBogarin:
What does `CI/Linter (pull_request)` do?

bdraco:
> What does `CI/Linter (pull_request)` do?

https://github.com/aio-libs/aiohttp/actions/runs/12192823875/job/34014087992?pr=10099

`[towncrier-fragments]:194: : Spell check: pyright:  for some static type checkers (e.g. pyright)..`

aiohttp/client_exceptions.py
Dreamsorcerer: Well, yes, that's what we suggested earlier. We don't need the TYPE_CHECKING either then. asvetlov is keen to keep it for now though (https://github.com/aio-libs/aiohttp/pull/10099#issuecomment-2517971410). Though I can't imagine anyone using a version of Python without SSL and wanting to use aiohttp on it..

Dreamsorcerer:
That was my suggestion, have adding to spelling list. Coverage is good now too.

Dreamsorcerer (approved):

bdraco (approved):

PR #10101, opened 2024-12-04
Fix race in `FileResponse` if file is replaced during `prepare`

### LINKED ISSUE #8013: `FileResponse` has undefined behavior if the file is changed out from under it between the `stat` and `open` calls

bdraco:
There was a race in ``FileResponse`` where the `stat` would be incorrect if the file was changed out between the `stat` and `open` syscalls. This would lead to various unexpected behaviors such as trying to read beyond the length of the file or sending a partial file. This problem is likely to occour when files are being renamed/linked into place.

An example of how this can happen with a system that provides weather data every 60s:

An external process writes `.weather.txt` at the top of each minute, and than renames it into place as `weather.txt`. In this case `FileResponse.prepare` may `stat` the old `weather.txt`, and than `open` the new `weather.txt`, and use the `os.stat_result` from the original file.

To fix this we now `fstat` the open file on operating systems where `fstat` is available.

This change looks much larger than it actually is because most of the function now needs to be wrapped in the `try`/`finally` to ensure that we close the file cleanly as we open it a lot sooner now.  Its best to compare without white space as well.

fixes #8013

### DIFF
--- a/CHANGES/10101.bugfix.rst
+++ b/CHANGES/10101.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed race condition in :class:`aiohttp.web.FileResponse` that could result in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during :meth:`aiohttp.web.FileResponse.prepare` -- by :user:`bdraco`.
--- a/CHANGES/10101.bugfix.rst
+++ b/CHANGES/10101.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could result in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
--- a/CHANGES/10101.bugfix.rst
+++ b/CHANGES/10101.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the file system during ``prepare`` -- by :user:`bdraco`.
--- a/CHANGES/10101.bugfix.rst
+++ b/CHANGES/10101.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during :meth:`aiohttp.web.FileResponse.prepare` -- by :user:`bdraco`.
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -74,21 +74,6 @@ for content_type, extension in ADDITIONAL_CONTENT_TYPES.items():
 _CLOSE_FUTURES: Set[asyncio.Future[None]] = set()
 
 
-def _stat_open_file(
-    file_path: pathlib.Path, fobj: io.BufferedReader, st: Optional[os.stat_result]
-) -> os.stat_result:
-    """Return the stat result of the file or the fallback stat result.
-
-    Ideally we can use fstat() to get the stat result of the file object,
-    to ensure we are returning the correct length of the open file,
-    but it is not possible to get the file descriptor from the file object
-    on some operating systems, so we have to use the stat result of the file path.
-    """
-    with suppress(OSError):  # May not work on Windows
-        return os.stat(fobj.fileno())
-    return st or file_path.stat()
-
-
 class FileResponse(StreamResponse):
     """A response object can be used to send files."""
 
@@ -199,7 +184,11 @@ class FileResponse(StreamResponse):
                 st = compressed_path.lstat()
                 if S_ISREG(st.st_mode):
                     fobj = compressed_path.open("rb")
-                    st = _stat_open_file(compressed_path, fobj, st)
+                    with suppress(OSError):
+                        # Once we open the file, we want the fstat() to ensure
+                        # the file has not changed between the first stat()
+                        # and the open().
+                        st = os.stat(fobj.fileno())
                     return fobj, st, file_encoding
 
         # Fallback to the uncompressed file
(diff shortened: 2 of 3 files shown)


CHANGES/10101.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could result in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during :meth:`aiohttp.web.FileResponse.prepare` -- by :user:`bdraco`.
```

CHANGES/10101.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during :meth:`aiohttp.web.FileResponse.prepare` -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
```

CHANGES/10101.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the filesystem during ``prepare`` -- by :user:`bdraco`.
```
bdraco:
```suggestion
Fixed race condition in :class:`aiohttp.web.FileResponse` that could have resulted in an incorrect response if the file was replaced on the file system during ``prepare`` -- by :user:`bdraco`.
```

bdraco:
close reopen because github flaked in the middle of the CI run.

https://www.githubstatus.com/

Hopefully I don't have to make a new PR

aiohttp/web_fileresponse.py
```diff
@@ -179,31 +184,72 @@ def _get_file_path_stat_encoding(
                 # Do not follow symlinks and ignore any non-regular files.
                 st = compressed_path.lstat()
```
steverep, line 185: Why not `open` before the `lstat` and then use `os.lstat` on the file object?

aiohttp/web_fileresponse.py
```diff
@@ -179,31 +184,72 @@ def _get_file_path_stat_encoding(
...
                 if S_ISREG(st.st_mode):
-                    return compressed_path, st, file_encoding
+                    fobj = compressed_path.open("rb")
+                    with suppress(OSError):
```
steverep, line 188: What's being suppressed here that isn't caught with the other `suppress`?

aiohttp/web_fileresponse.py
bdraco: It could be a socket, and each operating system will handle it slightly differently and raise different errno, some of which are not mapped in Cpython which would complicate the exception handling. ie on [MacOS its errno 102](https://stackoverflow.com/questions/36325912/socket-error-errno-102-operation-not-supported-on-socket) but does not have a named error.

aiohttp/web_fileresponse.py
bdraco: > fstat() may not be available on all platforms

Its broken in some cpython versions as well
https://github.com/python/cpython/issues/111856

steverep: Got it.  I thought this was under the previous `with`, which I now realize is not the case.

Dreamsorcerer:
@bdraco As far as I can see from codecov history, this PR is the one that loses coverage on web_response.py. Should we remove that check or adjust the tests to get coverage again?

https://app.codecov.io/gh/aio-libs/aiohttp/pull/10101/indirect-changes

bdraco:
After fighting with 

> @bdraco As far as I can see from codecov history, this PR is the one that loses coverage on web_response.py. Should we remove that check or adjust the tests to get coverage again?
> 
> [app.codecov.io/gh/aio-libs/aiohttp/pull/10101/indirect-changes](https://app.codecov.io/gh/aio-libs/aiohttp/pull/10101/indirect-changes)

That's a bit surprising.  I'm a bit curious what we were passing in there before that would have hit that branch

https://app.codecov.io/gh/aio-libs/aiohttp/pull/10095/blob/aiohttp/web_response.py#L268
https://app.codecov.io/gh/aio-libs/aiohttp/pull/10101/blob/aiohttp/web_response.py#L268

bdraco:
> That's a bit surprising. I'm a bit curious what we were passing in there before that would have hit that branch


@Dreamsorcerer 
Checked out fcce1bf6a5c339a3d63ab1678dcb6658ffc7d570

Made a small change

```diff
diff --git a/aiohttp/web_response.py b/aiohttp/web_response.py
index cb3e3717c..907f061f5 100644
--- a/aiohttp/web_response.py
+++ b/aiohttp/web_response.py
@@ -267,6 +267,10 @@ class StreamResponse(BaseClass, HeadersMixin, CookieMixin):
             )
         elif isinstance(value, str):
             self._headers[hdrs.LAST_MODIFIED] = value
+        else:
+            raise ValueError(
+                "Unsupported type for last_modified: %s" % type(value).__name__
+            )
 
     @property
     def etag(self) -> Optional[ETag]:
```

And... it looks like it was only covering the branch because it was passing in a `MagicMock` before
```
FAILED tests/test_web_sendfile.py::test_gzip_if_header_not_present_and_file_not_available[pyloop] - ValueError: Unsupported type for last_modified: MagicMock
FAILED tests/test_web_sendfile.py::test_using_gzip_if_header_present_and_file_available[pyloop] - ValueError: Unsupported type for last_modified: MagicMock
FAILED tests/test_web_sendfile.py::test_gzip_if_header_present_and_file_not_available[pyloop] - ValueError: Unsupported type for last_modified: MagicMock
FAILED tests/test_web_sendfile.py::test_status_controlled_by_user[pyloop] - ValueError: Unsupported type for last_modified: MagicMock
FAILED tests/test_web_sendfile.py::test_gzip_if_header_not_present_and_file_available[pyloop] - ValueError: Unsupported type for last_modified: MagicMock
```

bdraco:
Maybe raising ValueError when it's going to do nothing does make the most sense (along with a test for it).  However not sure if we should backport something like that to 3.x as the current behavior is to swallow unsupported types and raising an exception might be breaking change.

bdraco:
Captured as https://github.com/aio-libs/aiohttp/issues/10143

PR #10102, opened 2024-12-04
Fix deprecated calls to `guess_type` for paths with Python 3.13

bdraco:
## What do these changes do?

https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_type
> Deprecated since version 3.13: Passing a file path instead of URL is [soft deprecated](https://docs.python.org/3/glossary.html#term-soft-deprecated). Use [guess_file_type()](https://docs.python.org/3/library/mimetypes.html#mimetypes.guess_file_type) for this.

These should have been changed to `guess_file_type` on Python 3.13 as `guess_type` expects a URL and `guess_file_type` expects a path. Its a soft deprecation so they still work but that may change in the future.

no

no

### DIFF
--- a/CHANGES/10102.bugfix.rst
+++ b/CHANGES/10102.bugfix.rst
@@ -1,0 +1,1 @@
+Replaced deprecated call to :func:`mime_types.guess_type` with :func:`mime_types.guess_file_type` when using Python 3.13+ -- by :user:`bdraco`.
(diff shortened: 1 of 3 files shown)


CHANGES/10102.bugfix.rst
```diff
@@ -0,0 +1 @@
+Replaced deprecated call to :func:`mime_types.guess_type` with :func:`mime_types.guess_file_type` when using Python 3.13+ -- by :user:`bdraco`.
```
bdraco:
```suggestion
Replaced deprecated call to :func:`mimetypes.guess_type` with :func:`mimetypes.guess_file_type` when using Python 3.13+ -- by :user:`bdraco`.
```

PR #10113, opened 2024-12-05
Restore 304 performance after fixing `FileResponse` replace race

### LINKED ISSUE #10112: Restore performance of 304s after #10101
> The fix certainly handles the closings smartly, but I would be more worried about the additional overhead of the file openings.  If cache is controlled well, many implementations are likely serving 304s much more than 200s.  #10101 probably introduced a minor penalty there, but I don't see a better way to fix this issue. 🤷🏻‍♂️

_Originally posted by @steverep in https://github.com/aio-libs/aiohttp/issues/8013#issuecomment-2518555708_

bdraco:
Followup to #10101 to address the feedback in #10112

We can avoid opening the file is we are going to send a 304. 

fixes #10112


aiohttp/web_fileresponse.py
```diff
@@ -66,6 +67,16 @@
     }
 )
 
+
+class _FileResponseResult(Enum):
+    """The state of the file response."""
```
bdraco:
```suggestion
    """The result of the file response."""
```

aiohttp/web_fileresponse.py
```diff
@@ -163,17 +174,69 @@ async def _precondition_failed(
...
-    def _open_file_path_stat_encoding(
-        self, accept_encoding: str
-    ) -> Tuple[Optional[io.BufferedReader], os.stat_result, Optional[str]]:
-        """Return the io object, stat result, and encoding.
+    def _make_response(
+        self, request: "BaseRequest", accept_encoding: str
+    ) -> Tuple[
+        _FileResponseResult, Optional[io.BufferedReader], os.stat_result, Optional[str]
+    ]:
+        """Return the response state, io object, stat result, and encoding.
```
bdraco:
```suggestion
        """Return the response result, io object, stat result, and encoding.
```

steverep:
Nice!  I might have guessed maybe 5% penalty from the buffered open.  How big is the benchmark file (at least 4K I hope)?

bdraco:
The test uses this 11.1KB file https://github.com/aio-libs/aiohttp/blob/master/tests/sample.txt

PR #10121, opened 2024-12-05
Add host parameter to aiohttp_server fixture (#10120)

### LINKED ISSUE #10120: Add host parameter to aiohttp_server fixture.

I develop code that accesses 3rd party services available as docker images.

When testing my code using the aiohttp pytest plugin the following needs to happen:
* I create a server using the `aiohttp_server` fixture.
* I call the service running in the docker container and ...
* ... **pass the `aiohttp_server`'s URL as callback URL**.
* The docker service tries to send me some data.
* For the docker container to reach the host it needs to use either the host IP or `172.17.0.1`.

What happens is that aiohttp_server is unreachable.

This happens because:
* `aiohttp_server` calls `TestServer(app, port=port)` -- in `aiohttp/pytest_plugin.py` line 309
* `TestServer.__init__()` has a default value `host='127.0.0.1' in `aiohttp/test_utils.py` line 227

Change lines 306ff in `aiohttp.pytest_plugin.py` from 

````
    async def go(
        app: Application, *, port: Optional[int] = None, **kwargs: Any
    ) -> TestServer:
        server = TestServer(app, port=port)
````

to 

````
    async def go(
        app: Application, *, host: Optional[str] = '127.0.0.1', port: Optional[int] = None, **kwargs: Any
    ) -> TestServer:
        server = TestServer(app, host=host, port=port)
````

I created my own fixture, a copy of `aiohttp_server` with the fix applied.

### Related component

Server

### Additional context

The releated component selected above is misleading, as only the pytest plugin is affected.

### Code of Conduct

### PR DESCRIPTION
**Pull request by christianwbrock**
Fixes #10120 

It adds a host parameter to the aiohttp_server fixture to allow access
from non local clients, e.g. running in docker containers.

### DIFF
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -304,9 +304,9 @@ def aiohttp_server(loop: asyncio.AbstractEventLoop) -> Iterator[AiohttpServer]:
     servers = []
 
     async def go(
-        app: Application, *, port: Optional[int] = None, **kwargs: Any
+        app: Application, *, host: str = "127.0.0.1", port: Optional[int] = None, **kwargs: Any
     ) -> TestServer:
-        server = TestServer(app, port=port)
+        server = TestServer(app, host=host, port=port)
         await server.start_server(**kwargs)
         servers.append(server)
         return server
--- a/aiohttp/pytest_plugin.py
+++ b/aiohttp/pytest_plugin.py
@@ -304,7 +304,11 @@ def aiohttp_server(loop: asyncio.AbstractEventLoop) -> Iterator[AiohttpServer]:
     servers = []
 
     async def go(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: aiohttp/pytest_plugin.py [hunk 1]
```diff
@@ -304,9 +304,13 @@ def aiohttp_server(loop: asyncio.AbstractEventLoop) -> Iterator[AiohttpServer]:
     servers = []
 
     async def go(
-        app: Application, *, port: Optional[int] = None, **kwargs: Any
+        app: Application,
+        *,
+        host: Optional[str] = "127.0.0.1",
```
Dreamsorcerer: See the type error. host can't be None:
```suggestion
        host: str = "127.0.0.1",
```

Dreamsorcerer (approved):

PR #10122, opened 2024-12-05
Typing improvements for file responses

bdraco:
When I was working on this in #10113 and #10101 I found there was quite a bit of missing typing which made it a bit harder to validate changes.

Add some more typing here to ensure everything is working as expected

### DIFF
--- a/aiohttp/web_fileresponse.py
+++ b/aiohttp/web_fileresponse.py
@@ -18,7 +18,6 @@ from typing import (
     Optional,
     Set,
     Tuple,
-    cast,
 )
 
 from . import hdrs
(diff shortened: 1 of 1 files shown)


aiohttp/web_fileresponse.py
```diff
@@ -337,7 +334,7 @@ async def _prepare_open_file(
 
             # If a range request has been made, convert start, end slice
             # notation into file pointer offset and count
-            if start is not None or end is not None:
+            if start is not None:
```
bdraco, line 337: `end` is checked in the next if. It only makes sense to enter this block is `start is not None` since all the operations with `start` require it to be an `int`

PR #10128, opened 2024-12-05
Release 3.11.10

bdraco:
https://github.com/aio-libs/aiohttp/actions/runs/12189342218
Screenshot 2024-12-05 at 4 39 01 PM


bdraco:
The performance regression is expected, and there is nothing we can do about it for now until a new cpython is released

ref https://github.com/python/cpython/pull/127656

PR #10131, opened 2024-12-06
Bump Python version for benchmarks to 3.13

bdraco:
Python 3.13.1 has been released. As its rare to see performance changes in Python versions after .0, and even more rare to see them after .1 its time to start benchmarking against 3.13.1

https://github.com/python/cpython/pull/106555
https://github.com/python/cpython/pull/106879
https://github.com/python/cpython/pull/106528

Showing some modest gains from the above
Screenshot 2024-12-05 at 6 32 31 PM

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -266,11 +266,11 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: true
-    - name: Setup Python 3.12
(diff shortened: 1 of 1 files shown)


bdraco:
https://github.com/python/cpython/pull/106555
https://github.com/python/cpython/pull/106879
https://github.com/python/cpython/pull/106528

Showing some modest gains from the above

PR #10137, opened 2024-12-06
Restore zero copy writes on Python 3.12.9+/3.13.2+

bdraco:
TODO
- [x] Hold until 3.13.2 release day: Tuesday, 2025-02-04
- [x] Wait for GitHub to update 3.13 runners to 3.13.2
- [x] Re-run benchmarks
- [x] If benchmarks show the perf regression is fixed, this should be good to move forward and do a new release

## What do these changes do?

Restore zero copy support on Python 3.12.9+/3.13.2+ (unreleased at this time)
[CVE-2024-12254 aka GHSA-fw89-6wjj-8j95 is fixed on these Python versions](https://github.com/python/cpython/pull/127656)

The performance regression introduced by disabling zero copy writes in https://github.com/aio-libs/aiohttp/pull/10125 will be resolved

The relevant benchmarks (keep in mind we are benchmarking disabling zero copy writes), and this PR seeks to restore the performance:

`test_one_hundred_get_requests_with_1024_chunked_payload` +3% (`writelines` is a bit worse for tiny payloads)
`test_one_hundred_get_requests_with_512kib_chunked_payload` -21% (`writelines` is a lot better for large payloads)
`test_one_hundred_get_requests_with_30000_chunked_payload` -4% (`writelines` is a slightly better for medium size payloads)

Since we already have to branch here, I also added `MIN_PAYLOAD_FOR_WRITELINES` to 2048 to ensure we no longer take the 3% performance hit for `writelines` with small payloads


aiohttp/http_writer.py
```diff
@@ -24,6 +25,16 @@
 __all__ = ("StreamWriter", "HttpVersion", "HttpVersion10", "HttpVersion11")
 
 
+MIN_PAYLOAD_FOR_WRITELINES = 2048
+UNSAFE_WRITELINES = (3, 13, 0) <= sys.version_info < (3, 13, 2) or sys.version_info < (
+    3,
+    12,
+    9,
+)
```
bdraco: Annoying how black formats this. but 🤷

CHANGES/10137.misc.rst
```diff
@@ -0,0 +1,3 @@
+Restored support for zero copy writes when using Python 3.12 versions 3.12.9 and later or Python 3.13.2+ -- by :user:`bdraco`.
+
+Zero copy writes were previously disabled due to CVE-2024-12254 which is resolved in these Python versions.
```
webknjaz: Sphinx v8.1 (which I've just checked that our RTD setup uses already) added the `:cve:` role (it was previously only available through `sphinx-issues`). So we can now do this:
```suggestion
Zero copy writes were previously disabled due to :cve:`2024-12254` which is resolved in these Python versions.
```

bdraco: nice!

aiohttp/http_writer.py
```diff
@@ -24,6 +25,16 @@
 __all__ = ("StreamWriter", "HttpVersion", "HttpVersion10", "HttpVersion11")
 
 
+MIN_PAYLOAD_FOR_WRITELINES = 2048
+UNSAFE_WRITELINES = (3, 13, 0) <= sys.version_info < (3, 13, 2) or sys.version_info < (
```
webknjaz: The part after `or` will always evaluate to `False`. Because, for 3.12.0 the first part will be `True`. The check must be more granular if we want this to work.

aiohttp/http_writer.py
webknjaz: FWIW, you can fight formatting with putting these into separate named variables, which would improve readability greatly. I'd rather read `UNSAFE_WRITELINES = IS_PY313_BEFORE_313_2 or IS_PY312_BEFORE_312_9`.

aiohttp/http_writer.py
webknjaz: Oh... nevermind. I missed that it's a double-sided check with `<= x <`.

webknjaz:
@bdraco apparently, GHA is pinned to v7.1.2 and doesn't have this role (while it's not a problem with RTD):
```console
/home/runner/work/aiohttp/aiohttp/docs/changes.rst:815: ERROR: Unknown interpreted text role "cve".
```

CHANGES/10137.misc.rst
bdraco: /home/runner/work/aiohttp/aiohttp/docs/changes.rst:815: ERROR: Unknown interpreted text role "cve".

Looks like we might need something else?

webknjaz:
https://github.com/aio-libs/aiohttp/blob/f641aa4/requirements/doc.txt#L47C9-L47C14

Apparently, it was even downgraded in its last change: https://github.com/aio-libs/aiohttp/pull/8023/files#diff-9ba32e29b4b0a2ef5581fa10c8a8a79f01605f541f722957c098fc05a4addb22L55

aiohttp/http_writer.py
bdraco: Thats a good idea 👍

Dreamsorcerer: Though such variables often reduce maintainability when upgrading. i.e. they require someone to manually check and clean up code, whereas without variables pyupgrade can resolve it cleanly with no input.

bdraco:
almost released https://github.com/actions/python-versions/releases/tag/3.13.2-13149511920

bdraco:
as soon as https://github.com/actions/python-versions/pull/331 merges we can re-run

webknjaz:
It's merged!

bdraco:
Looks like we are still getting 3.13.1

  `Successfully set up CPython (3.13.1)`

bdraco:
There we go

Merging https://github.com/aio-libs/aiohttp/pull/10137 will improve performances by 18.02%

PR #10146, opened 2024-12-08
Raise `TypeError` when setting `StreamResponse.last_modified` to an unsupported type

### LINKED ISSUE #10143: Raise `TypeError` when `last_modified` gets an unexpected type
> Maybe raising ValueError when it's going to do nothing does make the most sense (along with a test for it).  However not sure if we should backport something like that to 3.x as the current behavior is to swallow unsupported types and raising an exception might be breaking change.

_Originally posted by @bdraco in https://github.com/aio-libs/aiohttp/issues/10101#issuecomment-2524873050_

bdraco:
## What do these changes do?

closes #10143

Currently this would silently fail and discard the new value.

Screenshot 2024-12-08 at 12 32 49 PM

### DIFF
--- a/CHANGES/10146.misc.rst
+++ b/CHANGES/10146.misc.rst
@@ -1,0 +1,1 @@
+Setting :attr:`aiohttp.web.StreamResponse.last_modified` to an unsupported type will now raise :exc:`ValueError` instead of silently failing -- by :user:`bdraco`.
--- a/CHANGES/10146.misc.rst
+++ b/CHANGES/10146.misc.rst
@@ -1,1 +1,1 @@
-Setting :attr:`aiohttp.web.StreamResponse.last_modified` to an unsupported type will now raise :exc:`ValueError` instead of silently failing -- by :user:`bdraco`.
+Setting :attr:`aiohttp.web.StreamResponse.last_modified` to an unsupported type will now raise :exc:`TypeError` instead of silently failing -- by :user:`bdraco`.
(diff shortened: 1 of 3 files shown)


bdraco:
Thought about it some more, and probably ok to backport to 3.12 but not 3.11

Dreamsorcerer (approved):

bdraco:
`TypeError` is probably better here actually

CHANGES/10146.misc.rst
```diff
@@ -0,0 +1 @@
+Setting :attr:`aiohttp.web.StreamResponse.last_modified` to an unsupported type will now raise :exc:`ValueError` instead of silently failing -- by :user:`bdraco`.
```
bdraco:
```suggestion
Setting :attr:`aiohttp.web.StreamResponse.last_modified` to an unsupported type will now raise :exc:`TypeError` instead of silently failing -- by :user:`bdraco`.
```

aiohttp/web_response.py
```diff
@@ -267,6 +267,9 @@ def last_modified(
             )
         elif isinstance(value, str):
             self._headers[hdrs.LAST_MODIFIED] = value
+        else:
+            msg = f"Unsupported type for last_modified: {type(value).__name__}"  # type: ignore[unreachable]
+            raise ValueError(msg)
```
bdraco:
```suggestion
            raise TypeError(msg)
```

tests/test_web_response.py
```diff
@@ -242,6 +242,13 @@ def test_last_modified_reset() -> None:
     assert resp.last_modified is None
 
 
+def test_last_modified_invalid_type() -> None:
+    resp = web.StreamResponse()
+
+    with pytest.raises(ValueError, match="Unsupported type for last_modified: object"):
```
bdraco:
```suggestion
    with pytest.raises(TypeError, match="Unsupported type for last_modified: object"):
```

asvetlov (approved):
lgtm

PR #10156, opened 2024-12-11
Add ALPN extension to client SSL Context

### LINKED ISSUE #10152: Add support for ALPN extension in client

aiohttp client doesn't send ALPN (TLS application_layer_protocol_negotiation extension). Some servers may refuse to do handshake with clients or block requests without this extension set in TLS layer. (As in #8478 , #10141 )

It has been implemented in cpython's urllib and curl. (https://bugs.python.org/issue40968) So I think enabling this in aiohttp is necessary.

aiohttp client only uses http/1.1 protocol for now. Simply setting it by `set_alpn_protocols['http/1.1']` after creating SSL default context should be fine.

none

### Related component

Client

### Additional context

_No response_

### Code of Conduct

Cycloctane:
## What do these changes do?

Add "http/1.1" ALPN extension to aiohttp client's SSL Context.

## Related issue number

Fixes #10152 

## Checklist

### DIFF
--- a/CHANGES/10156.feature.rst
+++ b/CHANGES/10156.feature.rst
@@ -1,1 +1,3 @@
-Add TLS ALPN extension to client SSL context -- by :user:`Cycloctane`.
+Enabled ALPN on default SSL contexts. This improves compatibility with some
+proxies which don't work without this extension.
+-- by :user:`Cycloctane`.
--- a/CHANGES/10156.feature.rst
+++ b/CHANGES/10156.feature.rst
@@ -1,0 +1,1 @@
+Add TLS ALPN extension to client SSL context -- by :user:`Cycloctane`.
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -772,14 +772,16 @@ def _make_ssl_context(verified: bool) -> SSLContext:
         # No ssl support
         return None  # type: ignore[unreachable]
     if verified:
-        return ssl.create_default_context()
-    sslcontext = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
-    sslcontext.options |= ssl.OP_NO_SSLv2
-    sslcontext.options |= ssl.OP_NO_SSLv3
-    sslcontext.check_hostname = False
-    sslcontext.verify_mode = ssl.CERT_NONE
-    sslcontext.options |= ssl.OP_NO_COMPRESSION
-    sslcontext.set_default_verify_paths()
+        sslcontext = ssl.create_default_context()
+    else:
+        sslcontext = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        sslcontext.options |= ssl.OP_NO_SSLv2
+        sslcontext.options |= ssl.OP_NO_SSLv3
+        sslcontext.check_hostname = False
+        sslcontext.verify_mode = ssl.CERT_NONE
+        sslcontext.options |= ssl.OP_NO_COMPRESSION
+        sslcontext.set_default_verify_paths()
+    sslcontext.set_alpn_protocols(["http/1.1"])
     return sslcontext
 
 
--- a/aiohttp/connector.py
+++ b/aiohttp/connector.py
@@ -781,7 +781,7 @@ def _make_ssl_context(verified: bool) -> SSLContext:
         sslcontext.verify_mode = ssl.CERT_NONE
         sslcontext.options |= ssl.OP_NO_COMPRESSION
         sslcontext.set_default_verify_paths()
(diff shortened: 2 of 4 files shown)


aiohttp/connector.py
```diff
@@ -772,14 +772,16 @@ def _make_ssl_context(verified: bool) -> SSLContext:
...
+        sslcontext = ssl.create_default_context()
+    else:
+        sslcontext = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        sslcontext.options |= ssl.OP_NO_SSLv2
+        sslcontext.options |= ssl.OP_NO_SSLv3
+        sslcontext.check_hostname = False
+        sslcontext.verify_mode = ssl.CERT_NONE
+        sslcontext.options |= ssl.OP_NO_COMPRESSION
+        sslcontext.set_default_verify_paths()
+    sslcontext.set_alpn_protocols(["http/1.1"])
```
Dreamsorcerer: Does this work with a tuple? I'd generally stick with immutable structures where possible.
```suggestion
    sslcontext.set_alpn_protocols(("http/1.1",))
```

tests/test_client_functional.py
```diff
@@ -632,6 +632,28 @@ async def handler(request: web.Request) -> web.Response:
...
+async def test_ssl_client_alpn(
+    aiohttp_server: AiohttpServer,
+    aiohttp_client: AiohttpClient,
+    ssl_ctx: ssl.SSLContext,
+) -> None:
+
+    async def handler(request: web.Request) -> web.Response:
+        sslobj = request.transport.get_extra_info("ssl_object")
+        assert sslobj.selected_alpn_protocol() == "http/1.1"
+        return web.Response(text="Test message")
```
Dreamsorcerer: Might make sense to return the protocol in the response and then assert it below?

tests/test_client_functional.py
```diff
@@ -632,6 +632,28 @@ async def handler(request: web.Request) -> web.Response:
...
+
+    app = web.Application()
+    app.router.add_route("GET", "/", handler)
+    ssl_ctx.set_alpn_protocols(["http/1.1"])
+    server = await aiohttp_server(app, ssl=ssl_ctx)
```
Dreamsorcerer: Do we need this on the server side? I thought this was defined entirely by the client..

CHANGES/10156.feature.rst
```diff
@@ -0,0 +1 @@
+Add TLS ALPN extension to client SSL context -- by :user:`Cycloctane`.
```
Dreamsorcerer: I'm just trying to think what this means for a user:
```suggestion
Enabled ALPN on default SSL contexts. This improves compatibility with some
proxies which don't work without this extension.
-- by :user:`Cycloctane`.
```

tests/test_client_functional.py
Cycloctane: alpn is also defined on server side. The protocol selected by alpn is based on alpn ext in both server's and client's hello. We need to set alpn on server side first and see if server can get selected protocol in tls handshake with client  to test alpn ext on client side.

CHANGES/10156.feature.rst
Cycloctane: I think enabling alpn on aiohttp client is only improving compatibility with some http/2 compatible servers that require alpn in tls. It is not actually reducing protocol negotiation round trips when there is only http/1.1 supported in client side.

tests/test_client_functional.py
Dreamsorcerer: Right, I think I misunderstood. If we don't define it, then I guess it just ignores the extension and continues with a normal handshake.

CHANGES/10156.feature.rst
Dreamsorcerer: Yeah, I think you're right. Updated suggestion.

tests/test_client_functional.py
```diff
@@ -632,6 +632,29 @@ async def handler(request: web.Request) -> web.Response:
...
 
 
+async def test_ssl_client_alpn(
+    aiohttp_server: AiohttpServer,
+    aiohttp_client: AiohttpClient,
+    ssl_ctx: ssl.SSLContext,
+) -> None:
+
+    async def handler(request: web.Request) -> web.Response:
+        sslobj = request.transport.get_extra_info("ssl_object")
```
Dreamsorcerer, line 643:
```suggestion
        assert request.transport is not None
        sslobj = request.transport.get_extra_info("ssl_object")
```

Dreamsorcerer (approved):

bdraco (approved):

PR #10159, opened 2024-12-11
Add distinction between error cases in TimerContext

### LINKED ISSUE #10153: "Timeout context manager should be used inside a task" slightly misleading
### Describe the bug

 This error is also raised when the timer context is used within a task, but that task is on a different event loop to the one the `ClientSession` was created with because it uses `asyncio.current_task(loop=self._loop)`. Calling `asyncio.current_task()` would show there is a task, but not on the correct event loop.

### To Reproduce

This stripped down example is a bit convoluted, but is the least amount of code I could come up with to reproduce what my application does. It has a long running context and spins up an event loop in another thread, which tries to use the session. Now that I have realised the problem have plans to change this entirely, but I thought it would be worth reporting as I think the error could be clearer.

```python
import asyncio
from threading import Thread

async def main():
    session = ClientSession()

    async def get():
        async with session.get("http://python.org") as response:
            print("Status:", response.status)
            print("Content-type:", response.headers["content-type"])

            html = await response.text()
            print("Body:", html[:15], "...")

    print("This loop")
    loop.call_soon_threadsafe(loop.create_task, get())

    await asyncio.sleep(1)

    loop = asyncio.new_event_loop()
    worker = Thread(target=loop.run_forever)
    worker.start()

    print("Wrong loop")
    loop.call_soon_threadsafe(loop.create_task, get())

    await asyncio.sleep(1)


asyncio.run(main())
```

### Expected behavior

The error could make it clear that this error could also occur if the task was created on the wrong event loop and not just that the timeout manager was not used inside a task. Or possibly it could distinguish the two cases and give different errors.

### Logs/tracebacks

```python-traceback
This loop
Status: 200
Content-type: text/html; charset=utf-8
Body: <!doctype html> ...
Wrong loop
future: <Task finished name='Task-7' coro=<main.<locals>.get() done, defined at /workspaces/fastcs-eiger/app.py:10> exception=RuntimeError('Timeout context manager should be used inside a task')>
  File "/workspaces/fastcs-eiger/app.py", line 11, in get
    async with session.get("http://python.org") as response:
  File "/venv/lib/python3.11/site-packages/aiohttp/client.py", line 1423, in __aenter__
    self._resp: _RetType = await self._coro
                           ^^^^^^^^^^^^^^^^
  File "/venv/lib/python3.11/site-packages/aiohttp/client.py", line 607, in _request
    with timer:
  File "/venv/lib/python3.11/site-packages/aiohttp/helpers.py", line 636, in __enter__
    raise RuntimeError("Timeout context manager should be used inside a task")
RuntimeError: Timeout context manager should be used inside a task
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7efc90fd9f50>
Unclosed connector
connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x7efc90e08910>, 115561.676192066)])']
connector: <aiohttp.connector.TCPConnector object at 0x7efc90e02110>
```

### Python Version

```console
$ python --version
Python 3.11.9
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Name: aiohttp
Version: 3.11.10
Author: 
Author-email: 
License: Apache-2.0
Location: /venv/lib/python3.11/site-packages
Requires: aiohappyeyeballs, aiosignal, attrs, frozenlist, multidict, propcache, yarl
Required-by: fastcs-eiger, tickit
```

### multidict Version

```console
Name: multidict
Version: 6.1.0
Author: Andrew Svetlov
License: Apache 2
Location: /venv/lib/python3.11/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### propcache Version

```console
Name: propcache
Version: 0.2.1
Author: Andrew Svetlov
License: Apache-2.0
Location: /venv/lib/python3.11/site-packages
Requires: 
Required-by: aiohttp, yarl
```

### yarl Version

```console
$ python -m pip show yarl
Name: yarl
Version: 1.18.3
Author: Andrew Svetlov
License: Apache-2.0
Location: /venv/lib/python3.11/site-packages
Required-by: aiohttp
```

### OS

RHEL8

### Related component

Client

### Additional context

_No response_

### Code of Conduct

GDYendell:
## What do these changes do?
This adds a distinct error message in the case that `TimerContext` is used within a task, but the task is on a different event loop.

The user will get a slightly different - and hopefully more useful - error message in the above case.

No.

## Related issue number
Fixes #10153 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -140,6 +140,7 @@ Frederik Peter Aalund
 Gabriel Tremblay
 Gang Ji
 Gary Wilson Jr.
+Gary Yendell
 Gennady Andreyev
 Georges Dubus
 Greg Holt
--- a/aiohttp/helpers.py
+++ b/aiohttp/helpers.py
@@ -658,7 +658,16 @@ class TimerContext(BaseTimerContext):
     def __enter__(self) -> BaseTimerContext:
         task = asyncio.current_task(loop=self._loop)
         if task is None:
-            raise RuntimeError("Timeout context manager should be used inside a task")
+            try:
+                asyncio.current_task()
+            except RuntimeError:
+                raise RuntimeError(
+                    "Timeout context manager should be used inside a task"
+                )
+            else:
+                raise RuntimeError(
+                    "Timeout context manager used in a task on a different event loop"
+                )
 
         if sys.version_info >= (3, 11):
             # Remember if the task was already cancelling
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -6,6 +6,7 @@ import sys
 import weakref
 from math import ceil, modf
 from pathlib import Path
+from threading import Thread
 from typing import Dict, Iterator, Optional, Union
 from unittest import mock
 from urllib.request import getproxies_environment
@@ -403,9 +404,32 @@ async def test_timer_context_timeout_does_swallow_cancellation() -> None:
 
 
 def test_timer_context_no_task(loop: asyncio.AbstractEventLoop) -> None:
-    with pytest.raises(RuntimeError):
+    with pytest.raises(
+        RuntimeError, match="Timeout context manager should be used inside a task"
+    ):
         with helpers.TimerContext(loop):
             pass
+
+
+async def test_timer_context_wrong_event_loop() -> None:
+    loop = asyncio.get_running_loop()
+
+    async def use_timer(loop: asyncio.AbstractEventLoop = loop) -> None:
+        with pytest.raises(
+            RuntimeError,
+            match="Timeout context manager used in a task on a different event loop",
+        ):
+            with helpers.TimerContext(loop):
+                pass
+
+    other_loop = asyncio.new_event_loop()
+    thread = Thread(target=other_loop.run_forever)
+    thread.start()
+
+    asyncio.run_coroutine_threadsafe(use_timer(), other_loop).result(timeout=5)
+
+    other_loop.stop()
+    thread.join()
 
 
 async def test_weakref_handle(loop: asyncio.AbstractEventLoop) -> None:


GDYendell:
Let me know if you would like a test that checks the specific error messages for the different cases. Is there any relevant documentation to update for this change? I don't think it is worth adding an entry to CHANGES/ but I can do.

Dreamsorcerer:
> Let me know if you would like a test that checks the specific error messages for the different cases.

Yeah, we'll want a test case for coverage.

> Is there any relevant documentation to update for this change? I don't think it is worth adding an entry to CHANGES/ but I can do.

Agreed, leave it.

Dreamsorcerer:
I'm a little confused what happened to the coverage, it says master has coverage of the existing exception, but this change doesn't have coverage of either of the exceptions...

Dreamsorcerer:
...unless asyncio.current_task() is managing to cause a RuntimeError itself and the tests aren't checking the exception text..

GDYendell:
> ...unless asyncio.current_task() is managing to cause a RuntimeError itself

Yep you're right. I have updated it to a try-except. Now the behaviour is correct. The original error case is still hit if used outside of an event loop entirely, as shown in the test. I get the new error in my application when used in the wrong event loop.

However, I am having trouble provoking the latter error case in a test, because the main event loop happens to have another task, so `asyncio.current_task(loop=self._loop)` does return a task, just not the one that the context manager is running in.

Do you have any suggestions on how this could be done? Do you think this is now adding too much complication for the sake of an error message?

Dreamsorcerer:
> > ...unless asyncio.current_task() is managing to cause a RuntimeError itself
> 
> Yep you're right. I have updated it to a try-except.

Odd, the docs don't seem to suggest such behaviour.

> Do you have any suggestions on how this could be done? Do you think this is now adding too much complication for the sake of an error message?

I'm not too sure how to make it work. Maybe it's not worth the effort, but I've never used multiple loops before. Up to you if you want to continue pursuing it.

PR #10171, opened 2024-12-17
Stream unpauses protocol before releasing connection

### LINKED ISSUE #10169: aiohttp hangs when uploading a file during streaming download
### Describe the bug

I've got some code that downloads a large archive file from S3 using aiobotocore, extracts it and uploads each file to another key in the same bucket, all using iterators to work on small chunks of data. We noticed that the process got stuck some times (maybe ~30% of the time) while doing so. No timeout or exceptions were raised but the process never finished.

Originally reported as an [aiobotocore issue](https://github.com/aio-libs/aiobotocore/issues/1235) but while debugging it, we managed to get to this minimum reproducing example using only aiohttp. Although in this example a timeout is being raised and the original it didn't (or it might have been swallowed by our code).

### To Reproduce

Run a localstack server as follows:
```
docker run \
  --rm -it \
  -p 127.0.0.1:4566:4566 \
  -p 127.0.0.1:4510-4559:4510-4559 \
  -v /var/run/docker.sock:/var/run/docker.sock \
  localstack/localstack

dd if=/dev/zero of=ae4bbfcdc47f450aa8557abefeba4a5ct bs=1M count=1
aws --endpoint-url=http://localhost:4566 s3api create-bucket --bucket bucket --region local
aws --endpoint-url=http://localhost:4566 s3 cp ae4bbfcdc47f450aa8557abefeba4a5ct s3://bucket
```

Then run a streaming download with concurrent uploads:

```python
import asyncio
import aiohttp

async def run():
    for _ in range(10):
            response = await session.get(
                "http://localhost:4566/bucket/ae4bbfcdc47f450aa8557abefeba4a5ct",
            )
            i = 0
            async for chunk in response.content.iter_chunked(1024):
                i += 1

                print(f"chunk {i}")
                # This process only fails near the end of the download, this condition is just here to speed up the testing
                # by skipping some uploads but the bug reproduces without it, it just takes more time. 
                if i >= 900:
                    print("Streamed, time to upload")
                    # It hangs awaiting this, no timeout is raised
                    await session.put(
                        "http://localhost:4566/bucket/output/some_file",
                        data=b""
                    )
                    print("Uploaded")

asyncio.run(run())
```

Sometimes, this will get stuck and eventually timeout after 5 minutes (or whatever the timeout is set to).

### Expected behavior

The script finishes successfully.

### Logs/tracebacks

```python-traceback
  File "venv/lib/python3.12/site-packages/aiohttp/client_reqrep.py", line 1055, in start
                       ^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/aiohttp/streams.py", line 668, in read
    await self._waiter
asyncio.exceptions.CancelledError

  File "replicate.py", line 24, in <module>
    asyncio.run(run())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "replicate.py", line 18, in run
    await session.put(
  File "venv/lib/python3.12/site-packages/aiohttp/client.py", line 728, in _request
    await resp.start(conn)
  File "venv/lib/python3.12/site-packages/aiohttp/client_reqrep.py", line 1050, in start
    with self._timer:
         ^^^^^^^^^^^
  File "venv/lib/python3.12/site-packages/aiohttp/helpers.py", line 671, in __exit__
    raise asyncio.TimeoutError from exc_val
TimeoutError
```

### Python Version

```console
3.12.7
```

### aiohttp Version

```console
3.11.10

Also tested 3.7.0, 3.8.0, 3.9.0, 3.10.0, 3.11.0 with same results.
```

### multidict Version

```console
4.7.6
```

### propcache Version

```console
0.2.1
```

### yarl Version

```console
1.18.3
```

### OS

Tested in Arch Linux and Ubuntu 24.04

### Related component

Client

### Additional context

A bit more context in the [original issue](https://github.com/aio-libs/aiobotocore/issues/1235). The important points are:

- It's a non-deterministic bug, it does not always happen. But with this script I managed to reproduce it ~30% of the tries.
- If it hangs, it always does so near the end of the download, at exactly the same place in the code. In the original code, downloading an archive of123 files, it either failed when trying to upload the 110th, or it didn't fail. Debugging it a bit, I think that's when the network download has completed but we still have buffered data pending to process.
- I attempted to reproduce against other services and I was unable to do so. I can reproduce against S3 (the cloud services) and the LocalStack emulator. I was not able to reproduce against minio (another S3 emulator).
- As a workaround, I'm using two separate clients, one to do the download an one to do the uploads. This works fine and I'm not experiencing any problems this way.
- Putting breakpoints through the code makes it much harder to reproduce. I'm assuming it's some kind of timing bug and that pausing in the debugger makes it go away.
- Looking at network dumps, it seems the issue manifest when the upload tries to reuse the download connection just after it finishes. At the network level, everything seems correct, the request is sent and the response received. But aiohttp never returns (times out with the trace above).

### Code of Conduct

javitonino:
## What do these changes do?

Fixes a hang when reusing a connection that was previously used for a streaming download and returned to the pool in paused state.

No, aside from the bugfix.

No. This adds bit of extra code to ensure correctness in this case. It's covered by tests to ensure the behaviour stays through code changes.

## Related issue number

Fixes https://github.com/aio-libs/aiohttp/issues/10169

## Checklist

### DIFF
--- a/aiohttp/streams.py
+++ b/aiohttp/streams.py
@@ -220,6 +220,9 @@ class StreamReader(AsyncStreamReaderMixin):
             self._eof_waiter = None
             set_result(waiter, None)
 
+        if self._protocol._reading_paused:
+            self._protocol.resume_reading()
+
         for cb in self._eof_callbacks:
             try:
                 cb()
--- a/CHANGES/10169.bugfix.rst
+++ b/CHANGES/10169.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed a hang where a connection previously used for a streaming
+download could be returned to the pool in a paused state.
+-- by :user:`javitonino`.
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -179,6 +179,7 @@ Jan Buchar
 Jan Gosmann
 Jarno Elonen
 Jashandeep Sohi
+Javier Torres
 Jean-Baptiste Estival
 Jens Steinhauser
 Jeonghun Lee
--- a/tests/test_flowcontrol_streams.py
+++ b/tests/test_flowcontrol_streams.py
@@ -107,3 +107,25 @@ class TestFlowControlStreamReader:
         res = stream.read_nowait(5)
         assert res == b""
         assert stream._protocol.resume_reading.call_count == 1  # type: ignore[attr-defined]
+
+    async def test_resumed_on_eof(self, stream: streams.StreamReader) -> None:
+        stream.feed_data(b"data")
+        assert stream._protocol.pause_reading.call_count == 1
+        assert stream._protocol.resume_reading.call_count == 0
+        stream._protocol._reading_paused = True
+
+        stream.feed_eof()
+        assert stream._protocol.resume_reading.call_count == 1
+
+
+async def test_stream_reader_eof_when_full() -> None:
+    loop = asyncio.get_event_loop()
+    protocol = BaseProtocol(loop=loop)
+    protocol.transport = asyncio.Transport()
+    stream = streams.StreamReader(protocol, 1024, loop=loop)
+
+    data_len = stream._high_water + 1
+    stream.feed_data(b"0" * data_len)
+    assert protocol._reading_paused
+    stream.feed_eof()
+    assert not protocol._reading_paused
--- a/tests/test_flowcontrol_streams.py
+++ b/tests/test_flowcontrol_streams.py
@@ -110,12 +110,12 @@ class TestFlowControlStreamReader:
 
     async def test_resumed_on_eof(self, stream: streams.StreamReader) -> None:
         stream.feed_data(b"data")
-        assert stream._protocol.pause_reading.call_count == 1
-        assert stream._protocol.resume_reading.call_count == 0
+        assert stream._protocol.pause_reading.call_count == 1  # type: ignore[attr-defined]
+        assert stream._protocol.resume_reading.call_count == 0  # type: ignore[attr-defined]
         stream._protocol._reading_paused = True
 
         stream.feed_eof()
-        assert stream._protocol.resume_reading.call_count == 1
+        assert stream._protocol.resume_reading.call_count == 1  # type: ignore[attr-defined]
 
 
 async def test_stream_reader_eof_when_full() -> None:


Dreamsorcerer:
I'm wondering if doing it in BaseConnector._release() may be the safest option, to ensure that it is always unpaused regardless of what any other components might do. Anybody else have any thoughts on it?

asvetlov:
Please give me some time for the review.

javitonino:
I think it makes sense to have this, at least, in the StreamReader, in the sense of having it clean up after itself.

However, I think that having a global catch-all at connection release is also a good idea, just in case we missed some other case that can end up in the same situation. I only did this change because I'm not sure of the consequences of doing this at a more general point might have (looks like it should be ok, but I'm just a newcomer here).

bdraco:
I'm assuming this happens on the last chunk of data for the payload which it just enough to go over the high water mark, and than the parser is complete and calls `feed_eof`.  I can't see how else it could happen.  I don't see an attack vector here that could be used to fill up memory since the next `feed_data` call would pause again anyways (not that we should get one after eof).

bdraco (approved):

aiohttp/streams.py
```diff
@@ -220,6 +220,9 @@ def feed_eof(self) -> None:
             self._eof_waiter = None
             set_result(waiter, None)
 
+        if self._protocol._reading_paused:
```
bdraco, line 223: Leaving an open conversation so this doesn't merge before https://github.com/aio-libs/aiohttp/pull/10171#issuecomment-2548475830

asvetlov (approved):
Looks good, let's merge and publish a bugfix release

bdraco:
If I don't run out of time, I'll do a release later today

bdraco:
3.11 backport https://github.com/aio-libs/aiohttp/pull/10179

bdraco:
3.12 backport https://github.com/aio-libs/aiohttp/pull/10180

bdraco:
Thanks @javitonino 

Working on release 3.11.11 now

https://github.com/aio-libs/aiohttp/pull/10181

bdraco:
3.11.11 is now available with this fix

PR #10178, opened 2024-12-18
Add py3.14 to CI test matrix

asvetlov:
It allows errors detection early.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -155,10 +155,10 @@ jobs:
             no-extensions: 'Y'
             os: ubuntu
             experimental: false
-          # - os: ubuntu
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -155,10 +155,10 @@ jobs:
             no-extensions: 'Y'
             os: ubuntu
             experimental: false
-          # - os: ubuntu
-          #   pyver: "3.14"
-          #   experimental: true
-          #   no-extensions: 'Y'
+          - os: ubuntu
+            pyver: "3.14.0-alpha.2"
```
Dreamsorcerer: Leave it as 3.14 as before and it will just take the latest release. I don't think we want to update this every month..

PR #10222, opened 2024-12-23
Revert incorrect typing changes

Dreamsorcerer:
From #9528. This should work again with mypy 1.14.

Add some new flags too.

Edit: Not actually fixed in mypy 1.14. Waiting for 1.15 for proper fix.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -49,7 +49,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.13
+        python-version: 3.11
     - name: Cache PyPI
       uses: actions/cache@v4.2.0
       with:
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -20,6 +20,7 @@ pretty = True
 show_column_numbers = True
 show_error_codes = True
 show_error_code_links = True
+strict_bytes = True
 strict_equality = True
 warn_incomplete_stub = True
 warn_redundant_casts = True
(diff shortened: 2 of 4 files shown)


Dreamsorcerer:
Pending dependabot PR, which should catch up over the next 2 days.

asvetlov (approved):

webknjaz:
> Pending dependabot PR, which should catch up over the next 2 days.

@Dreamsorcerer alternatively we could just bump all the deps manually, with a local pip-tools run. And avoid those annoying Dependabot notifications coming in every day..

asvetlov:
Sorry, guys, I've missed what are you talking about.
@webknjaz do you want to disable dependabot for the project?

Dreamsorcerer:
> > Pending dependabot PR, which should catch up over the next 2 days.
> 
> @Dreamsorcerer alternatively we could just bump all the deps manually, with a local pip-tools run. And avoid those annoying Dependabot notifications coming in every day..

I'm just being lazy. Could have also manually triggered Dependabot. I think using Dependabot is still preferable, but would be great to filter out those notifications...

webknjaz:
> do you want to disable dependabot for the project?

No, not right now anyway. The problem is that it was stuck for a while, and Sam fixed it just the other day. Now it's trying to catch up by creating a gazillion of PRs, one per each dependency. It calls `pip-compile` under the hood. A more efficient way could be running it manually to bump everything just this once, so dependabot would be less noisy.

The current setup is not very flexible and has a number of issues in the space of package management, so it's possible to fundamentally rework it, but it requires a certain amount of focused effort. This is why I'm not jumping right into it, even though I've done this research multiple times elsewhere and have really good practices that work.

PR #10281, opened 2024-12-31
Update workflow to use upload-artifact v4 and download-artifact v4

silamon:
## What do these changes do?

This updates the ci/cd workflow to use the upload-artifact v4 and download-artifact v4 github actions. The currently used upload-artifact and download-artifact will no longer work at the end of next month. 
The changes are needed since v4 no longer has mutable artifacts, which was used to collect wheels from different architectures.  

Fix #8588, Fix #8589, Fix #9009, Fix #10189, Fix #10191 

No, the ci/cd workflow has been tested although without publishing. 

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -421,6 +421,7 @@ jobs:
       uses: pypa/cibuildwheel@v2.22.0
       env:
         CIBW_ARCHS_MACOS: x86_64 arm64 universal2
+    - name: Upload wheels
       uses: actions/upload-artifact@v4
       with:
         name: >-  
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -424,11 +424,11 @@ jobs:
     - name: Upload wheels
       uses: actions/upload-artifact@v4
       with:
-        name: >-  
-          dist-${{ matrix.os }}-${{  
-            matrix.qemu  
-            && matrix.qemu  
-            || 'native'  
+        name: >-
+          dist-${{ matrix.os }}-${{
+            matrix.qemu
+            && matrix.qemu
+            || 'native'
           }}
         path: ./wheelhouse/*.whl
 
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1,6 +1,7 @@
 name: CI
 
 on:
+  merge_group:
   push:
     branches:
       - 'master'
@@ -420,15 +421,14 @@ jobs:
       uses: pypa/cibuildwheel@v2.22.0
       env:
         CIBW_ARCHS_MACOS: x86_64 arm64 universal2
-    - if: ${{ matrix.qemu }}
       uses: actions/upload-artifact@v4
       with:
-        name: dist-${{ matrix.os }}-${{ matrix.qemu }}
-        path: ./wheelhouse/*.whl
-    - if: ${{ !matrix.qemu }}
-      uses: actions/upload-artifact@v4
-      with:
-        name: dist-${{ matrix.os }}
+        name: >-  
+          dist-${{ matrix.os }}-${{  
+            matrix.qemu  
+            && matrix.qemu  
+            || 'native'  
+          }}
         path: ./wheelhouse/*.whl
 
   deploy:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -444,14 +449,19 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: true
+    - name: Merge distribution artifacts
+      uses: actions/upload-artifact/merge@v4
+      with:
+        name: dist
+        pattern: dist-*
```
silamon: This is optional, it's to ensure that the same outputs are outputted in case you use these artifacts for something else outside this workflow which I'm unsure about.

.github/workflows/ci-cd.yml
```diff
@@ -1,7 +1,6 @@
 name: CI
 
 on:
-  merge_group:
```
webknjaz: This breaks the merge queues support. Do not delete it.

.github/workflows/ci-cd.yml
```diff
@@ -421,9 +420,15 @@ jobs:
       uses: pypa/cibuildwheel@v2.22.0
       env:
         CIBW_ARCHS_MACOS: x86_64 arm64 universal2
-    - uses: actions/upload-artifact@v3
+    - if: ${{ matrix.qemu }}
```
webknjaz: No need for mustache in if-clauses. Also, every step must have a name.

.github/workflows/ci-cd.yml
```diff
@@ -421,9 +420,15 @@ jobs:
...
+      uses: actions/upload-artifact@v4
       with:
-        name: dist
+        name: dist-${{ matrix.os }}-${{ matrix.qemu }}
```
webknjaz: The last bit could just be made conditional instead of adding another step:
```suggestion
        name: >-
          dist-${{ matrix.os }}-${{
            matrix.qemu
            && matrix.qemu
            || 'native'
          }}
```

webknjaz: In other places, we don't really care about the conditional even: https://github.com/aio-libs/multidict/blob/fbf28ae/.github/workflows/reusable-build-wheel.yml#L79

.github/workflows/ci-cd.yml
webknjaz: Let's exclude this for now. They aren't used elsewhere. Maybe, submit it as a separate PR later. Also, if this is added, shouldn't the step below just access `dist` instead?

webknjaz (changes requested):
Thanks for looking into this! I have a few requests and problems mentioned in the inline comments. Please, address them.
Additionally, we probably need a contrib change note included.

.github/workflows/ci-cd.yml
Dreamsorcerer: It the duplicate in your suggestion a mistake @webknjaz? `matrix.qemu && matrix.qemu`

silamon: It works perfectly as suggested, without it is evaluated as a boolean expression.

.github/workflows/ci-cd.yml
silamon: Yeah, I removed it for now since it was only meant to be backwards compatible, which is unneeded.

.github/workflows/ci-cd.yml
webknjaz: @Dreamsorcerer no, it's not a mistake. It's a way to have ternary conditions, which is why I usually multiline such cases for clarity.
It's like when you do `cmd && cmd1 || cmd2`. The first occurrence is cast to a boolean.
A more generic way to demonstrate this. My typical idiom is `${{ bool && str || str }}`.

CHANGES/10281.contrib.rst
```diff
@@ -0,0 +1 @@
+The CI/CD workflow has been updated to use upload-artifact v4 and download-artifact v4 Github actions. -- by :user:`silamon`.
```
webknjaz: Let's drop this period right before the byline and this should be good to go.
```suggestion
The CI/CD workflow has been updated to use `upload-artifact` v4 and `download-artifact` v4 GitHub Actions -- by :user:`silamon`.
```

webknjaz (approved):
Thanks! Just one bit in the change note left to improve.

PR #10326, opened 2025-01-14
Add failing test for connection reuse for #10325

bdraco:


tests/test_client_functional.py
```diff
@@ -186,6 +186,61 @@ async def on_reuseconn(session: object, ctx: object, params: object) -> None:
     assert cnt_conn_reuse == 1
 
 
+pytest.mark.xfail(reason="Connection should be reused (#10325)")
+
+
```
```suggestion
@pytest.mark.xfail(reason="Connection should be reused (#10325)")
```

bdraco:
no longer needed since the issue will be fixed by #10915

Issue #10331, opened 2025-01-19
aiohttp._http_parser.HttpParser.feed_data aiohttp.http_exceptions.BadHttpMessage: 400, message:


skywalker0403:
### Describe the bug

I keep getting the above error message. This message accumulates to several hundred a day if I don't delete it in between. I have restored backups from the last 10 days so that I could perhaps rule out an update. Unfortunately, this was unsuccessful. The only update I could think of would be the first HA update this year. I don't know if there was another one. I don't have Duck DNS or anything similar to access my system from outside, I use VPN on my iPhone via the FritzBox instead. I don't know what to do anymore, especially since I can't do anything with this message because I don't know anything about Linux. I read about a similar problem here, but it was closed. There was no solution for me

### To Reproduce

As already described, I have 10 backup days and used the first backup of the 10, but the error was already there at that time.

### Expected behavior

By rewinding I hoped to find an update as the cause. Without success

### Logs/tracebacks

```python-traceback
Logger: aiohttp.server
Quelle: /usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py:451
Erstmals aufgetreten: 10:32:32 (5 Vorkommnisse)
Zuletzt protokolliert: 10:32:37

Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/aiohttp/web_protocol.py", line 380, in data_received
    messages, upgraded, tail = self._request_parser.feed_data(data)
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "aiohttp/_http_parser.pyx", line 558, in aiohttp._http_parser.HttpParser.feed_data
aiohttp.http_exceptions.BadHttpMessage: 400, message:
  Data after `Connection: close`:

    b'\x00'
          ^
```

### Python Version

```console
3.13.1
```

### aiohttp Version

```console
-
```

### multidict Version

```console
-
```

### propcache Version

```console
-
```

### yarl Version

```console
-
```

## System Information

version | core-2025.1.2
-- | --
installation_type | Home Assistant OS
dev | false
hassio | true
docker | true
user | root
virtualenv | false
python_version | 3.13.1
os_name | Linux
os_version | 6.6.66-haos
arch | x86_64
timezone | Europe/Berlin
config_dir | /config

<details><summary>Home Assistant Community Store</summary>

GitHub API | ok
-- | --
GitHub Content | ok
GitHub Web | ok
[...]

PR #10332, opened 2025-01-19
Log the remote that generates request errors

bdraco:
## What do these changes do?

Currently we only log that a request error occurred with no information about which remote the error
came from. This made it quite difficult to find
the offending remote

The remote address is now found in the logs

no

## Related issue number

#10331


Dreamsorcerer (approved):

asvetlov (approved):
Good addition, thanks!

Issue #10355, opened 2025-01-23
Parsing a missing CRLF at the end of a chunk in a malformed chunked encoding response


xdegaye:
### Describe the bug

A missing CRLF at the end of a chunk may happen for example when the parsed chunk size is wrong.

Parsing the end of a chunk is done by [HttpPayloadParser.feed_data(), see this part of the code](https://github.com/aio-libs/aiohttp/blob/d3dc087b8e9aa665d47045550e5d9f2eddf8f512/aiohttp/http_parser.py#L875).
When CRLF is missing:
- The HttpPayloadParser state remains `ChunkState.PARSE_CHUNKED_CHUNK_EOF` until there are no more chunks to parse
  since the next chunks processed by the following calls to feed_data() are prefixed with `self._chunk_tail` so that
  this code parses always the same start of sequence of bytes (which does not start with CRLF).
- All the next chunks are appended to `self._chunk_tail`.
- If the chunked body is large enough, such as an audio stream for example, the stream seems to hang.

### To Reproduce

In a well formed chunked encoded HTTP response, replace the CRLF at the end of a chunk with two bytes that are not CRLF.

### Expected behavior

A fix could be: 
- Change also the parser state to ChunkState.PARSE_CHUNKED_SIZE when the CRLF is missing.
- Alternatively raise an exception.

### Logs/tracebacks

```python-traceback
-
```

### Python Version

```console
-
```

### aiohttp Version

```console
commit d3dc087b8e9aa665d47045550e5d9f2eddf8f512 (HEAD -> master, origin/master, origin/HEAD)
Date:   Wed Jan 22 05:53:25 2025 -0800
```

### multidict Version

```console
-
```

### propcache Version

```console
-
```

### yarl Version

```console
-
```

### OS

-

### Related component

Server

### Additional context

This issue is the result of a code review. 

### Code of Conduct

- [x] I agree to follow the aio-libs Code of Conduct

Dreamsorcerer:
That fix may be OK. If you'd like to open a PR, please create a regression test first in test_http_parser.py (making sure to use the `response: HttpResponseParser` fixture).

xdegaye:
I will give it a try.

ekanshul:
[...]

PR #10359, opened 2025-01-25
Fixed missing CRLF at end of a chunk in malformed chunked encoding me…

xdegaye:
See issue #10355.

### DIFF
--- a/tests/test_http_parser.py
+++ b/tests/test_http_parser.py
@@ -1768,8 +1768,8 @@ class TestParsePayload:
         for lax, SEP in ((False, b"\r\n"), (True, b"\n")):
             out = aiohttp.StreamReader(protocol, 2**16, loop=loop)
             p = HttpPayloadParser(out, chunked=True, lax=lax)
-            p.feed_data(b"4\r\nasdf", SEP=SEP)  # type: ignore
-            p.feed_data(b"\r\n0\r\n\r\n", SEP=SEP)  # type: ignore
+            p.feed_data(b"4\r\nasdf", SEP=SEP)  # type: ignore[arg-type]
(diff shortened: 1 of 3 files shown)


xdegaye:
Do not know how to make the  `Labels / Backport label added` work.

tests/test_http_parser.py
```diff
@@ -1761,6 +1761,69 @@ async def test_parse_chunked_payload_split_end_trailers4(
         assert out.is_eof()
         assert b"asdf" == b"".join(out._buffer)
 
+    async def test_parse_chunked_payload_split_CRLF(
```
Dreamsorcerer, line 1764: Can we get a test that uses the `response: HttpResponseParser` fixture, rather than using HttpPayloadParser directly?
I think we'd want to ensure that both parsers work the same way, this only tests the Python parser (which is not used by most users).

xdegaye: Which is this second parser ?
Please bear with me, I have only been looking at aiohttp source code for few days.

The tests check the changes made in HttpPayloadParser, not in the second parser.

Dreamsorcerer: Yes, if you use the fixture as mentioned above, the test will be run against both parsers (i.e. run twice).

xdegaye: You did not answer my question. Please, what is the name of the second parser ?

Why should changes that affect HttpPayloadParser be tested against the second parser ?

What if the tests against the second parser fail ?
In that case:
- do you consider that these failures must be fixed within this PR ? 
- would'nt it be better to update the title of the PR and the issue to restrict their scope to HttpPayloadParser ?

Dreamsorcerer: There's a Python parser and a C parser (llhttp). Both should behave the same way, hence why we have a fixture to run the same tests against both parsers.

PR #10366, opened 2025-01-29
Fix sdist make

mgorny:
## What do these changes do?

Three fixes related to use of source distribution to build `aiohttp`:

1. Add missing `tools` directory to sdist, as it is required by the `Makefile` targets.
2. Fix `tools/gen.py` to work outside a git checkout, by looking for `pyproject.toml` rather than `.git`.
3. Add a `cythonize-nodeps` target that can be used by downstream packagers to perform cythonization without calling `pip`.

This fixes the ability to call `make` when working in an unpacked source distribution (currently it would fail due to missing `tools/gen.py`).

Don't think so. Worst case, the extra Makefile rule would go out of sync.

## Related issue number

n/a

## Checklist

### DIFF
--- a/CHANGES/10366.packaging
+++ b/CHANGES/10366.packaging
@@ -1,0 +1,2 @@
+Added missing files to the source distribution to fix ``Makefile`` targets.
+Added a ``cythonize-nodeps`` target to run Cython without invoking pip to install dependencies.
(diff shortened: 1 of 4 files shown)


mgorny:
Thanks a lot!

webknjaz:
@mgorny thanks! Eventually we'll migrate to something similar to what yarl and others do. But for now, this is good to have.

PR #10371, opened 2025-02-03
Added 'partitioned' parameter to set_cookie method.

### LINKED ISSUE #9870: Support for setting partitioned cookies

I need to use partitioned cookies to set cookies in contexts where third-party cookies are otherwise restricted.

A recent addition to the Set-Cookie header is the ability to mark cookies as partitioned (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#partitioned, https://developer.mozilla.org/en-US/docs/Web/Privacy/Privacy_sandbox/Partitioned_cookies).

`StreamResponse.set_cookie` should support setting partitioned cookies, e.g. using a `partitioned=True` keyword argument.

I can create a 'Set-Cookie' header manually, but would have to figure out edge cases with encoding and escaping special characters myself. This would be inconvenient and probably error-prone.

### Related component

Server

### Additional context

_No response_

### Code of Conduct

jpjacobpadilla:
## What do these changes do?

Fixes #9870 

## Checklist

-  [x] Add a new news fragment into the `CHANGES/` folder


docs/web_reference.rst
```diff
@@ -753,6 +754,11 @@ and :ref:`aiohttp-web-signals` handlers::
 
             .. versionadded:: 3.7
 
+      :param bool partitioned: ``True`` if you want the cookie to be
+         partitioned. (optional)
+
+            .. versionadded:: 3.14
+
```
Dreamsorcerer:
```suggestion
      :param bool partitioned: ``True`` to set a partitioned cookie.
         Available in Python 3.14+. (optional)

            .. versionadded:: 3.12

```

aiohttp/helpers.py
```diff
@@ -974,6 +975,9 @@ def set_cookie(
         if samesite is not None:
             c["samesite"] = samesite
 
+        if partitioned is not None:
+            c["partitioned"] = partitioned
```
Dreamsorcerer, line 979: Can we also get a quick test to cover this? Can add a `@pytest.mark.skipif(sys.version_info < (3, 14), reason="No partitioned support")` to run only on 3.14.

Dreamsorcerer: Another test around here should do the trick: https://github.com/aio-libs/aiohttp/blob/5bb6e9514ce6136d89932ebb9ee7558ec05e7a0b/tests/test_helpers.py#L911

Dreamsorcerer:
If you could follow the backport instructions, that'd be great.

jpjacobpadilla:
I should be able to do that! This is my first contribution to `aiohttp` so I'll have to read through those instructions tonight.

PR #10377, opened 2025-02-03
--disallow-any-decorated

Dreamsorcerer:
Most of these ignores will also disappear once we remove BasicAuth (in favour of middlewares).

### DIFF
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -12,6 +12,7 @@ from typing import (
     Dict,
     Generator,
     Iterable,
+    Iterator,
     List,
     Optional,
     Tuple,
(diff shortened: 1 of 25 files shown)


aiohttp/abc.py
```diff
@@ -12,6 +12,7 @@
     Dict,
     Generator,
     Iterable,
+    Iterator,
```
Dreamsorcerer:
```suggestion
```

aiohttp/abc.py
```diff
@@ -120,7 +120,7 @@ def request(self) -> Request:
         return self._request
 
     @abstractmethod
-    def __await__(self) -> Generator[Any, None, StreamResponse]:
+    def __await__(self) -> Generator[None, None, StreamResponse]:
```
asvetlov, line 123:
```suggestion
    def __await__(self) -> Generator[asyncio.Future[Any], None, StreamResponse]:
```
I guess, technically, asyncio.Task never sends non-None values into coroutines but respects to receive `asyncio.Future[Any]`

asvetlov (approved):
Looks good.
A part of ignores could be replaced with `Unpack[TypedDict]` for kwargs, but it could be done in a separate PR

aiohttp/abc.py
Dreamsorcerer: Typeshed at the moment doesn't seem to match this. We can update this later if typeshed gets updated.

bdraco (approved):

PR #10422, opened 2025-02-06
Add human readable error messages for WebSocket PING/PONG timeouts

bdraco:
## What do these changes do?

Currently the exception would stringify to an empty string which left the caller with no information about why the WebSocket disconnected in the log

Human readable errors

no

### DIFF
--- a/CHANGES/10422.misc.rst
+++ b/CHANGES/10422.misc.rst
@@ -1,3 +1,3 @@
-Added human readable error messages to the exceptions for WebSocket disconnects due to PONG not being received -- by :user:`bdraco`.
+Added human-readable error messages to the exceptions for WebSocket disconnects due to PONG not being received -- by :user:`bdraco`.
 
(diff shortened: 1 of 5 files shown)


CHANGES/10422.misc.rst
```diff
@@ -0,0 +1,3 @@
+Added human readable error messages to the exceptions for WebSocket disconnects due to PONG not being received -- by :user:`bdraco`.
```
bdraco:
```suggestion
Added human-readable error messages to the exceptions for WebSocket disconnects due to PONG not being received -- by :user:`bdraco`.
```

CHANGES/10422.misc.rst
```diff
@@ -0,0 +1,3 @@
...
+
+Previously the error messages were empty strings which made it hard to determine what went wrong.
```
bdraco:
```suggestion
Previously, the error messages were empty strings, which made it hard to determine what went wrong.
```

PR #10425, opened 2025-02-06
DNM: Attempt to avoid copy on read

bdraco:
This may already be optimized under the hood, but codspeed will tell us


bdraco:
Yeah its already optimized in the bytes join

PR #10433, opened 2025-02-06
Detect blocking calls in coroutines using BlockBuster

cbornet:
## What do these changes do?

This PR uses the [blockbuster](https://github.com/cbornet/blockbuster) library to detect blocking calls made in the asyncio event loop during unit tests.
Avoiding blocking calls is hard as these can be deeply buried in the code or made in 3rd party libraries.
Blockbuster makes it easier to detect them by raising an exception when a call is made to a known blocking function (eg: `time.sleep`).

No

I think no. In 5 years, it's just a few lines to remove if you don't want to use it anymore.

## Checklist

### DIFF
--- a/aiohttp/client.py
+++ b/aiohttp/client.py
@@ -607,11 +607,14 @@ class ClientSession:
                         if req_cookies:
                             all_cookies.load(req_cookies)
 
+                    proxy_: Optional[URL] = None
                     if proxy is not None:
-                        proxy = URL(proxy)
+                        proxy_ = URL(proxy)
                     elif self._trust_env:
                         with suppress(LookupError):
-                            proxy, proxy_auth = get_env_proxy_for_url(url)
+                            proxy_, proxy_auth = await asyncio.to_thread(
+                                get_env_proxy_for_url, url
+                            )
 
                     req = self._request_class(
                         method,
@@ -628,7 +631,7 @@ class ClientSession:
                         expect100=expect100,
                         loop=self._loop,
                         response_class=self._response_class,
-                        proxy=proxy,
+                        proxy=proxy_,
                         proxy_auth=proxy_auth,
                         timer=timer,
                         session=self,
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -35,7 +35,14 @@ IS_LINUX = sys.platform.startswith("linux")
 
 
 @pytest.fixture(autouse=True)
-def blockbuster() -> Iterator[None]:
+def blockbuster(request: pytest.FixtureRequest) -> Iterator[None]:
+    # No blockbuster for benchmark tests.
+    node = request.node.parent
+    while node:
+        if node.name.startswith("test_benchmarks"):
+            yield
+            return
+        node = node.parent
     with blockbuster_ctx(
         "aiohttp", excluded_modules=["aiohttp.pytest_plugin", "aiohttp.test_utils"]
     ) as bb:
@@ -43,7 +50,7 @@ def blockbuster() -> Iterator[None]:
         # https://github.com/aio-libs/aiohttp/issues/10435
         for func in ["io.TextIOWrapper.read", "os.stat"]:
             bb.functions[func].can_block_in("aiohttp/client_reqrep.py", "update_auth")
-        for func in ["os.readlink", "os.stat", "os.path.samestat"]:
+        for func in ["os.readlink", "os.stat", "os.path.abspath", "os.path.samestat"]:
             bb.functions[func].can_block_in(
                 "aiohttp/web_urldispatcher.py", "add_static"
             )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -50,7 +50,13 @@ def blockbuster(request: pytest.FixtureRequest) -> Iterator[None]:
         # https://github.com/aio-libs/aiohttp/issues/10435
         for func in ["io.TextIOWrapper.read", "os.stat"]:
             bb.functions[func].can_block_in("aiohttp/client_reqrep.py", "update_auth")
-        for func in ["os.getcwd", "os.readlink", "os.stat", "os.path.abspath", "os.path.samestat"]:
+        for func in [
+            "os.getcwd",
+            "os.readlink",
+            "os.stat",
+            "os.path.abspath",
+            "os.path.samestat",
+        ]:
             bb.functions[func].can_block_in(
                 "aiohttp/web_urldispatcher.py", "add_static"
             )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,10 @@ def blockbuster() -> Iterator[None]:
         # https://github.com/aio-libs/aiohttp/issues/10435
         for func in ["io.TextIOWrapper.read", "os.stat"]:
             bb.functions[func].can_block_in("aiohttp/client_reqrep.py", "update_auth")
+        for func in ["os.readlink", "os.stat", "os.path.samestat"]:
+            bb.functions[func].can_block_in(
+                "aiohttp/web_urldispatcher.py", "add_static"
+            )
         yield
 
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -36,20 +36,11 @@ IS_LINUX = sys.platform.startswith("linux")
 
 @pytest.fixture(autouse=True)
 def blockbuster() -> Iterator[None]:
-    with blockbuster_ctx("aiohttp") as bb:
+    with blockbuster_ctx("aiohttp", excluded_modules=["aiohttp.pytest_plugin", "aiohttp.test_utils"]) as bb:
         bb.functions["io.TextIOWrapper.read"].can_block_in(
             "aiohttp/client_reqrep.py", "update_auth"
         )
         bb.functions["os.stat"].can_block_in("aiohttp/client_reqrep.py", "update_auth")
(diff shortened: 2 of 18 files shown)


aiohttp/client.py
```diff
@@ -607,11 +607,14 @@ async def _request(
...
                             all_cookies.load(req_cookies)
 
+                    proxy_: Optional[URL] = None
                     if proxy is not None:
-                        proxy = URL(proxy)
+                        proxy_ = URL(proxy)
                     elif self._trust_env:
                         with suppress(LookupError):
-                            proxy, proxy_auth = get_env_proxy_for_url(url)
+                            proxy_, proxy_auth = await asyncio.to_thread(
```
cbornet, line 615: BlockBuster detected that `get_env_proxy_for_url` is blocking (through `netrc_from_env`)

tests/test_web_urldispatcher.py
```diff
@@ -813,19 +830,21 @@ async def test_static_absolute_url(
...
+    )
     assert resp.status == 403
 
 
 async def test_for_issue_5250(
     aiohttp_client: AiohttpClient, tmp_path: pathlib.Path
 ) -> None:
     app = web.Application()
-    app.router.add_static("/foo", tmp_path)
+    await asyncio.to_thread(app.router.add_static, "/foo", tmp_path)
```
bdraco: This seems like a bug in protection code that add_static does blocking IO.  I'm pretty sure Home Assistant has a workaround for this so it would be good to fix in the production code

bdraco: https://github.com/home-assistant/core/blob/634b754168c19a51b72079749cccdceec5551aea/homeassistant/components/http/__init__.py#L509

bdraco: https://github.com/home-assistant/core/blob/634b754168c19a51b72079749cccdceec5551aea/homeassistant/components/http/__init__.py#L477

tests/conftest.py
```diff
@@ -33,6 +34,25 @@
 IS_LINUX = sys.platform.startswith("linux")
 
 
+@pytest.fixture(autouse=True)
+def blockbuster() -> Iterator[None]:
+    with blockbuster_ctx("aiohttp") as bb:
+        bb.functions["io.TextIOWrapper.read"].can_block_in(
+            "aiohttp/client_reqrep.py", "update_auth"
```
cbornet: `update_auth` is blocking (through `netrc_from_env`) and called in `ClientRequest`'s constructor.
I'm not sure how to fix so putting an exclusion atm. Any idea ?

bdraco: Calling something like that in the constructor is very difficult to fix without a breaking change

bdraco:
I think this is a very nice change. I'm going to run out of time to review it today.

I'd add excludes for the known issues and add inline comments that they blocking I/O so they can be fixed when someone h as time to work out each one.

tests/conftest.py
cbornet: Caching the result of `netrc_from_env` could be a first step to minimize the impact and improve the perf.

aiohttp/client.py
```diff
@@ -607,11 +607,14 @@ async def _request(
                         if req_cookies:
                             all_cookies.load(req_cookies)
 
+                    proxy_: Optional[URL] = None
```
Dreamsorcerer, line 610: I'm not clear why the variable name change?

tests/test_web_urldispatcher.py
Dreamsorcerer: Given that this can't be done once the app is frozen (other than homeassistant hacking it), it's probably not been an issue so far.

tests/test_client_functional.py
Dreamsorcerer, line 1: Does it make sense to run this on tests/? Feels like it shouldn't make any real difference, and I feel it makes the code a little less readable (e.g. removing context managers).

aiohttp/client.py
cbornet: That’s to please mypy.
The original `proxy` may be a string.

Dreamsorcerer: Right, though I'm confused why the type information would have changed here, I'd expect the same types from this code as before. I'd probably name it proxy_url though.

tests/test_web_urldispatcher.py
cbornet: The blocking detected is through the path `add_static` -> `StaticResource.__init__` which does OS  calls to resolve the directory:
```python
        try:
            directory = Path(directory).expanduser().resolve(strict=True)
        except FileNotFoundError as error:
            raise ValueError(f"'{directory}' does not exist") from error
        if not directory.is_dir():
            raise ValueError(f"'{directory}' is not a directory")
```
I think in general `add_static` is called outside of an event loop. Also the `add_static` doc says it is not recommended for production code. So probably not a big issue if it blocks ?

tests/test_client_functional.py
cbornet: `blockbuster` has a mechanism to only trigger when a certain module is detected in the call stack.
It's configured to `aiohttp` in `conftest.py` and so you could expect that calls from `tests` that don't go through `aiohttp` would be allowed to block.
But the problem is that all tests first go through the `aiohttp.pytest_plugin` which is part of `aiohttp` so this mechanism does not work...
I'll see if I can add a feature to exclude certain modules from the scan.

aiohttp/client.py
cbornet: It seems the `asyncio.to_thread` makes mypy not as good as detecting the type narrowing.
Another solution would be to do `proxy=cast(proxy, Optional[URL]),`.

cbornet:
> I'd add excludes for the known issues and add inline comments that they blocking I/O so they can be fixed when someone h as time to work out each one.

done in https://github.com/aio-libs/aiohttp/pull/10433/commits/08e02f55fcc1cf7572b7af414907bbb8991a45e7

tests/test_client_functional.py
cbornet: done in https://github.com/aio-libs/aiohttp/pull/10433/commits/106d765cc5a578f5b74bf07a80b417954512bf20

aiohttp/client.py
Dreamsorcerer: cast would be a bad idea. Maybe I'll experiment and file a mypy bug later, but this should be fine.

tests/test_web_urldispatcher.py
Dreamsorcerer: If this is something we should fix (even if homeassistant were the only use case), then should we revert the to_thread() changes and whitelist the issue until fixed?

tests/test_client_functional.py
Dreamsorcerer: Awesome. Should we revert the file_contents read? I assume that doesn't need to be threaded anymore.

cbornet: If you prefer. But there are a lot of places where we read the content of the file. So is it not a good thing to factorize it in a fixture ?

Dreamsorcerer: Hmm, if the threading is not needed anymore then it doesn't seem like it saves much code (just a .read_bytes() call). But, mostly it just seems unrelated to the goal of this PR now.

tests/test_web_urldispatcher.py
cbornet: > If this is something we should fix (even if homeassistant were the only use case), then should we revert the to_thread() changes and whitelist the issue until fixed?

done in https://github.com/aio-libs/aiohttp/pull/10433/commits/b7cec65d06a678f88ee627e8b5dd07fe64f41fe7

tests/test_client_functional.py
cbornet: > Hmm, if the threading is not needed anymore then it doesn't seem like it saves much code (just a .read_bytes() call). But, mostly it just seems unrelated to the goal of this PR now.

Reverted in https://github.com/aio-libs/aiohttp/pull/10433/commits/b7cec65d06a678f88ee627e8b5dd07fe64f41fe7

bdraco:
It looks like the checks are running in the benchmarks which is benchmarking the time for the checks.  They should be excluded from the benchmark runs

cbornet:
> It looks like the checks are running in the benchmarks which is benchmarking the time for the checks. They should be excluded from the benchmark runs

done

Dreamsorcerer (approved):

bdraco (approved):

bdraco (approved):

bdraco:
Thanks @cbornet

bdraco:
@cbornet Looks like this couldn't be backported automatically, can you take a look?

tests/conftest.py
```diff
@@ -33,6 +34,36 @@
...
+            "os.getcwd",
+            "os.readlink",
+            "os.stat",
+            "os.path.abspath",
+            "os.path.samestat",
+        ]:
+            bb.functions[func].can_block_in(
+                "aiohttp/web_urldispatcher.py", "add_static"
+            )
+        bb.functions["os.getcwd"].can_block_in("coverage/control.py", "_should_trace")
```
cbornet, line 63: NOTE: not needed anymore with blockbuster v1.5.22+

bdraco: Would you please send another PR. We can merge on top and than backport once we get the backport PR merged.

cbornet: sure, no problem.

bdraco: Thanks!

PR #4890, opened 2020-07-30
Supports for multibytes and multidimensional memoryview

shagren:
## What do these changes do?

In some cases size for memoryview calculated incorrectly because _len()_ function used. I.e. if memoryview has more than one dimension.
In this PR _memoryview.nbytes_ used instead _len()_ for fixing issue.

## Are there changes in behavior for the user?

No

## Related issue number

None

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes
- [x] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`
  * The format is &lt;Name&gt; &lt;Surname&gt;.
  * Please keep alphabetical order, the file is sorted by names. 
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` for example (588.bugfix)
  * if you don't have an `issue_id` change it to the pr id after creating the pr
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: "Fix issue with non-ascii contents in doctest text files."

PR #10454, opened 2025-02-11
Fixes for strict-bytes

Dreamsorcerer:
This will be default behaviour in mypy 2, so best to fix it early.

It would be good if anybody else can check over these, the existing code seems a little confused with when it wants bytes only or will allow bytes/bytearray/memoryview.

### DIFF
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -202,7 +202,7 @@ class AbstractStreamWriter(ABC):
     length: Optional[int] = 0
 
     @abstractmethod
-    async def write(self, chunk: Union[bytes, bytearray, memoryview]) -> None:
+    async def write(self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]) -> None:
         """Write chunk into stream."""
 
     @abstractmethod
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -203,7 +203,7 @@ class AbstractStreamWriter(ABC):
 
     @abstractmethod
     async def write(
-        self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
+        self, chunk: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
     ) -> None:
         """Write chunk into stream."""
 
--- a/aiohttp/abc.py
+++ b/aiohttp/abc.py
@@ -202,7 +202,9 @@ class AbstractStreamWriter(ABC):
     length: Optional[int] = 0
 
     @abstractmethod
-    async def write(self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]) -> None:
+    async def write(
+        self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
+    ) -> None:
         """Write chunk into stream."""
 
     @abstractmethod
--- a/aiohttp/compression_utils.py
+++ b/aiohttp/compression_utils.py
@@ -1,7 +1,15 @@
 import asyncio
+import sys
 import zlib
 from concurrent.futures import Executor
 from typing import Optional, cast
+
+if sys.version_info >= (3, 12):
+    from collections.abc import Buffer
+else:
+    from typing import Union
+
+    Buffer = Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
 
 try:
     try:
@@ -66,10 +74,10 @@ class ZLibCompressor(ZlibBaseHandler):
             )
         self._compress_lock = asyncio.Lock()
 
-    def compress_sync(self, data: bytes) -> bytes:
+    def compress_sync(self, data: Buffer) -> bytes:
         return self._compressor.compress(data)
 
-    async def compress(self, data: bytes) -> bytes:
+    async def compress(self, data: Buffer) -> bytes:
         """Compress the data and returned the compressed bytes.
 
         Note that flush() must be called after the last call to compress()
@@ -111,10 +119,10 @@ class ZLibDecompressor(ZlibBaseHandler):
         )
         self._decompressor = zlib.decompressobj(wbits=self._mode)
 
-    def decompress_sync(self, data: bytes, max_length: int = 0) -> bytes:
+    def decompress_sync(self, data: Buffer, max_length: int = 0) -> bytes:
         return self._decompressor.decompress(data, max_length)
 
-    async def decompress(self, data: bytes, max_length: int = 0) -> bytes:
+    async def decompress(self, data: Buffer, max_length: int = 0) -> bytes:
         """Decompress the data and return the decompressed bytes.
 
         If the data size is large than the max_sync_chunk_size, the decompression
@@ -162,7 +170,7 @@ class BrotliDecompressor:
             )
         self._obj = brotli.Decompressor()
 
-    def decompress_sync(self, data: bytes) -> bytes:
+    def decompress_sync(self, data: Buffer) -> bytes:
         if hasattr(self._obj, "decompress"):
             return cast(bytes, self._obj.decompress(data))
         return cast(bytes, self._obj.process(data))
--- a/aiohttp/compression_utils.py
+++ b/aiohttp/compression_utils.py
@@ -9,7 +9,7 @@ if sys.version_info >= (3, 12):
 else:
     from typing import Union
 
-    Buffer = Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
+    Buffer = Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
 
 try:
     try:
--- a/aiohttp/http_writer.py
+++ b/aiohttp/http_writer.py
@@ -45,7 +45,7 @@ HttpVersion10 = HttpVersion(1, 0)
 HttpVersion11 = HttpVersion(1, 1)
 
 
-_T_OnChunkSent = Optional[Callable[[bytes], Awaitable[None]]]
+_T_OnChunkSent = Optional[Callable[[Union[bytes, bytearray, memoryview[int], memoryview[bytes]]], Awaitable[None]]]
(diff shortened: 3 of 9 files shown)


aiohttp/abc.py
```diff
@@ -202,7 +202,9 @@ class AbstractStreamWriter(ABC):
     length: Optional[int] = 0
 
     @abstractmethod
-    async def write(self, chunk: Union[bytes, bytearray, memoryview]) -> None:
+    async def write(
+        self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
```
Dreamsorcerer:
```suggestion
        self, chunk: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
```

aiohttp/compression_utils.py
```diff
@@ -1,8 +1,16 @@
...
 import zlib
 from concurrent.futures import Executor
 from typing import Optional, cast
 
+if sys.version_info >= (3, 12):
+    from collections.abc import Buffer
+else:
+    from typing import Union
+
+    Buffer = Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
```
Dreamsorcerer:
```suggestion
    Buffer = Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
```

aiohttp/http_writer.py
```diff
@@ -45,7 +45,11 @@ class HttpVersion(NamedTuple):
 HttpVersion11 = HttpVersion(1, 1)
 
 
-_T_OnChunkSent = Optional[Callable[[bytes], Awaitable[None]]]
+_T_OnChunkSent = Optional[
+    Callable[
+        [Union[bytes, bytearray, memoryview[int], memoryview[bytes]]], Awaitable[None]
```
Dreamsorcerer:
```suggestion
        [Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]], Awaitable[None]
```

aiohttp/http_writer.py
```diff
@@ -84,7 +88,9 @@ def enable_compression(
     ) -> None:
         self._compress = ZLibCompressor(encoding=encoding, strategy=strategy)
 
-    def _write(self, chunk: Union[bytes, bytearray, memoryview]) -> None:
+    def _write(
+        self, chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
```
Dreamsorcerer:
```suggestion
        self, chunk: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
```

aiohttp/http_writer.py
```diff
@@ -93,7 +99,10 @@ def _write(self, chunk: Union[bytes, bytearray, memoryview]) -> None:
             raise ClientConnectionResetError("Cannot write to closing transport")
         transport.write(chunk)
 
-    def _writelines(self, chunks: Iterable[bytes]) -> None:
+    def _writelines(
+        self,
+        chunks: Iterable[Union[bytes, bytearray, memoryview[int], memoryview[bytes]]],
```
Dreamsorcerer:
```suggestion
        chunks: Iterable[Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]],
```

aiohttp/http_writer.py
```diff
@@ -109,7 +118,7 @@ def _writelines(self, chunks: Iterable[bytes]) -> None:
 
     async def write(
         self,
-        chunk: Union[bytes, bytearray, memoryview],
+        chunk: Union[bytes, bytearray, memoryview[int], memoryview[bytes]],
```
Dreamsorcerer:
```suggestion
        chunk: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"],
```

aiohttp/web_response.py
```diff
@@ -435,7 +435,9 @@ async def _write_headers(self) -> None:
         status_line = f"HTTP/{version[0]}.{version[1]} {self._status} {self._reason}"
         await writer.write_headers(status_line, self._headers)
 
-    async def write(self, data: Union[bytes, bytearray, memoryview]) -> None:
+    async def write(
+        self, data: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
```
Dreamsorcerer:
```suggestion
        self, data: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
```

aiohttp/web_ws.py
```diff
@@ -606,7 +606,9 @@ async def receive_json(
         data = await self.receive_str(timeout=timeout)
         return loads(data)
 
-    async def write(self, data: bytes) -> None:
+    async def write(
+        self, data: Union[bytes, bytearray, memoryview[int], memoryview[bytes]]
```
Dreamsorcerer:
```suggestion
        self, data: Union[bytes, bytearray, "memoryview[int]", "memoryview[bytes]"]
```

Dreamsorcerer:
I'm unclear if the remaining errors are a bug with us or with typeshed (which says the method doesn't support `memoryview[bytes]`).

asvetlov (approved):

Dreamsorcerer:
> Would you backport the PR

It changes the (apparent) API, so I'd prefer to avoid it. Hopefully will be focusing on v4 after the typing work I'm doing.

asvetlov:
Up to you, any decision is good to me.

asvetlov:
Yes, focusing on V5 would be good.

Personally, I should review the diff and file a list of required changes related to my commits on master that were not backported.
I have quite a tight schedule this month; but the task is in my to-do list.

Dreamsorcerer:
> I'm unclear if the remaining errors are a bug with us or with typeshed (which says the method doesn't support `memoryview[bytes]`).

@bdraco I think you've played with that code fairly recently, any idea if memoryview[bytes] is actually supported?

bdraco:
asyncio Transport write is typed to `(method) def write(data: bytes | bytearray | memoryview[int]) -> Any`

so I think only `memoryview[int]` is needed.

Dreamsorcerer:
> asyncio Transport write is typed to `(method) def write(data: bytes | bytearray | memoryview[int]) -> Any`
> 
> so I think only `memoryview[int]` is needed.

We have tests that explicitly use `memoryview[bytes]` though. So, there's either a bug in our code, or in typeshed.

bdraco:
I couldn't find any tests where we have `memoryview[bytes]` .. AFAICT `memoryview(b"")` has a type of `memoryview[int]`....

```python
def memoryview_bytes(x: "memoryview[bytes]") -> None:
    pass


def memoryview_int(x: "memoryview[int]") -> None:
    pass


v = memoryview(b"")

memoryview_int(v)  # pass

memoryview_bytes(
    v
)  # error: Argument 1 to "memoryview_bytes" has incompatible type "memoryview[int]"; expected "memoryview[bytes]"  [arg-type]
```

Dreamsorcerer:
> I couldn't find any tests where we have `memoryview[bytes]` .. AFAICT `memoryview(b"")` has a type of `memoryview[int]`....

If you drop bytes from the union, then mypy will flag it up.

Dreamsorcerer:
```
tests/test_http_writer.py:454:21: error: Argument 1 to "write" of
"StreamWriter" has incompatible type "memoryview[bytes]"; expected
"bytes | bytearray | memoryview[int]"  [arg-type]
        await msg.write(payload)
                        ^~~~~~~
tests/test_http_writer.py:473:21: error: Argument 1 to "write" of
"StreamWriter" has incompatible type "memoryview[bytes]"; expected
"bytes | bytearray | memoryview[int]"  [arg-type]
        await msg.write(payload)
                        ^~~~~~~
```
https://github.com/aio-libs/aiohttp/blob/982ed47aa055bb117be0880f40c6ac4b8a2265a5/tests/test_http_writer.py#L442-L477

bdraco:
> payload = mv.cast("c", [3, 2])
That's what I was missing
```python
def memoryview_bytes(x: "memoryview[bytes]") -> None:
    pass


def memoryview_int(x: "memoryview[int]") -> None:
    pass


v = memoryview(b"")

memoryview_int(v)  # pass

payload = v.cast("c", [3, 2])
memoryview_bytes(payload)
```

passes now

bdraco:
> ```
> tests/test_http_writer.py:454:21: error: Argument 1 to "write" of
> "StreamWriter" has incompatible type "memoryview[bytes]"; expected
> "bytes | bytearray | memoryview[int]"  [arg-type]
>         await msg.write(payload)
>                         ^~~~~~~
> tests/test_http_writer.py:473:21: error: Argument 1 to "write" of
> "StreamWriter" has incompatible type "memoryview[bytes]"; expected
> "bytes | bytearray | memoryview[int]"  [arg-type]
>         await msg.write(payload)
>                         ^~~~~~~
> ```
> 
> https://github.com/aio-libs/aiohttp/blob/982ed47aa055bb117be0880f40c6ac4b8a2265a5/tests/test_http_writer.py#L442-L477

So it seems like we support it but `asyncio` is typed to only `memoryview[int]` since we end up calling `write` or `writelines` at https://github.com/aio-libs/aiohttp/blob/36aa738078ff8db20d418125a0da8cacd72735a7/aiohttp/http_writer.py#L106 https://github.com/aio-libs/aiohttp/blob/36aa738078ff8db20d418125a0da8cacd72735a7/aiohttp/http_writer.py#L108

`def writelines(list_of_data: Iterable[bytes | bytearray | memoryview[int]]) -> None`
`def write(data: bytes | bytearray | memoryview[int]) -> None`

I guess it works though since #4890 seems to use it

Dreamsorcerer:
> I guess it works though since #4890 seems to use it

Yeah, so typeshed bug then, I'll file a PR there.

bdraco:
It looks like it gets cast to `c` (char) https://github.com/aio-libs/aiohttp/blob/36aa738078ff8db20d418125a0da8cacd72735a7/aiohttp/http_writer.py#L129

bdraco (approved):

Dreamsorcerer:
Typeshed updated to `memoryview[object]`.

PR #10455, opened 2025-02-11
Cleanup questionable no cover comments

Dreamsorcerer:
I think all of these no cover comments shouldn't exist.

### DIFF
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -953,7 +953,8 @@ class ClientResponse(HeadersMixin):
 
         for val in re.split(r",(?=\s*<)", links_str):
             match = re.match(r"\s*<(.*)>(.*)", val)
-            assert match is not None
+            if match is None:  # Malformed link
+                continue
             url, params_str = match.groups()
             params = params_str.split(";")[1:]
 
--- a/aiohttp/client_reqrep.py
+++ b/aiohttp/client_reqrep.py
@@ -962,7 +962,8 @@ class ClientResponse(HeadersMixin):
 
             for param in params:
                 match = re.match(r"^\s*(\S*)\s*=\s*(['\"]?)(.*?)(\2)\s*$", param, re.M)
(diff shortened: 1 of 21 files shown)


asvetlov (commented):
Looks good in general except two replacements.

aiohttp/client_reqrep.py
```diff
@@ -953,19 +953,15 @@ def links(self) -> "MultiDictProxy[MultiDictProxy[Union[str, URL]]]":
 
         for val in re.split(r",(?=\s*<)", links_str):
             match = re.match(r"\s*<(.*)>(.*)", val)
-            if match is None:  # pragma: no cover
-                # the check exists to suppress mypy error
-                continue
+            assert match is not None
```
asvetlov: This line worries me a little. 
From my understanding, before we ignored malformed link headers.
After applying the change, malformed links raise AssertionError.

Please correct me if I'm wrong.

Dreamsorcerer: I was thinking it was related to paths originally. I think you're right, the comment is wrong, and there should be a test to cover this case.

Dreamsorcerer:
```suggestion
            if match is None:  # Malformed link
                continue
```

aiohttp/client_reqrep.py
```diff
@@ -963,9 +962,7 @@ def links(self) -> "MultiDictProxy[MultiDictProxy[Union[str, URL]]]":
 
             for param in params:
                 match = re.match(r"^\s*(\S*)\s*=\s*(['\"]?)(.*?)(\2)\s*$", param, re.M)
-                if match is None:  # pragma: no cover
-                    # the check exists to suppress mypy error
-                    continue
+                assert match is not None
```
Dreamsorcerer: This is probably the same.
```suggestion
                if match is None:  # Malformed param
                    continue
```

asvetlov (approved):

PR #10465, opened 2025-02-16
New IOResponse class for serving a response using any IO object

Derkades:
## What do these changes do?

FileResponse now accepts an opened file instead of only a file path. This allows sending a temporary file to the client by opening the file, deleting it, and then passing the open file to FileResponse. FileResponse handles closing the file when it is no longer used.

FileResponse is now a subclass of BaseIOResponse, I don't expect any compatibility issues for normal API usage

~~The FileResponse class is now more complicated. I am still thinking about if it makes sense to split FileResponse into two classes, one for handling IO and one for handling paths.~~ Now split into subclasses, I think it makes the code clearer than it was before.

It is a bit unusual for an API to take an open file and close it by itself. But it solves a problem. Other suggestions are welcome.

## Related issue number

https://github.com/aio-libs/aiohttp/discussions/10360

## Checklist

    number after creating the PR
      pre-agreed expectations.
      changes in behavior.
      process.
      runtime assumptions.
      environment.
      categories.
    for example:
    ```rst
    ```

    of this project.


aiohttp/web_fileresponse.py
```diff
@@ -377,11 +394,16 @@ async def _prepare_open_file(
...
+                    guesser = CONTENT_TYPES.guess_file_type
+                else:
+                    guesser = CONTENT_TYPES.guess_type
+                self.content_type = guesser(self._path)[0] or FALLBACK_CONTENT_TYPE
             else:
-                guesser = CONTENT_TYPES.guess_type
-            self.content_type = guesser(self._path)[0] or FALLBACK_CONTENT_TYPE
+                # content-type cannot be determined if self._io is used
+                # instead of self._path
+                self.content_type = FALLBACK_CONTENT_TYPE
```
Dreamsorcerer: Unless I'm missing something, it seems like FileResponse is missing Content-Disposition by default. If we were to add this in and recommend the user to specify a filename for the download, then we could use that filename here (or just use it if the header has been explicitly passed):
```
>>> mimetypes.guess_type("foo.html")
('text/html', None)
```

Dreamsorcerer (commented):
I think this looks reasonable. I suspect a subclass may be cleaner though (FileIOResponse maybe?).

Derkades:
Now rewritten with subclasses. It is much cleaner. I still need to bring back Accept-Encoding support.

Dreamsorcerer:
Looks promising. We'll need to update the docs too, and make sure we have a warning highlighting that the handler should not close the file object (including using a context manager).

Derkades:
I am having difficulty solving test failures with this error: `TypeError: Unsupported type for last_modified: MagicMock`

It happens when setting `self.last_modified` in the `_not_modified` function. I am unfortunately not very familiar with mock. Do you have an idea how to fix this?

Derkades:
CodSpeed reports a small regression, I think this is because the current implementation always opens the file even if it is unchanged. I can probably fix this.

Dreamsorcerer:
> It happens when setting `self.last_modified` in the `_not_modified` function. I am unfortunately not very familiar with mock. Do you have an idea how to fix this?

As far as I can see, last_modified comes from st_mtime, so we'll need to set it on the mock. e.g. Add a line after this one (presumably set to that value divided by 1000000 or whatever the appropriate order of magnitude is):
https://github.com/Derkades/aiohttp/blob/05e191d85cd65480b267cfccbe4d1f306a74ba8f/tests/test_web_sendfile.py#L152

Derkades:
With st_mtime added to the mock, the tests indeed pass. I am not sure why they didn't fail before, st_mtime was used previously.

`test_file_response_sends_headers_immediately` is failing because `send_headers()` is never called. That makes sense to me because `send_headers()` would be called in `_sendfile()`, but `_sendfile()` is replaced by a mock. Not sure how that worked before.

Dreamsorcerer:
Just noticing this PR must have been nearly ready. If it is ready, could you resolve the conflicts and mark it ready (not draft)?

Derkades:
Hey, thank you for your time once again. It is ready, but I was not able to resolve one test failure. 

Since submitting the PR I have been using this [vendored](https://codeberg.org/raphson/music-server/src/commit/69855c4480cd0a69bebc487e3619a26a9e79052f/raphson_mp/common/io_response.py) in production in my project. This is how I discovered [a bug](https://github.com/aio-libs/aiohttp/pull/10465/commits/afc0ef4f1158ffa67db7d7936d7127bb63a36465) in October 2025. Other than that I have not had any issues with it.

I have resolved the conflicts. I would greatly appreciate if you could take a look at the failing test.

> FAILED tests/test_web_sendfile.py::test_file_response_sends_headers_immediately - AssertionError: Expected 'send_headers' to have been called once. Called 0 times.

Derkades:
Is the type for BaseRequest.http_range correct? I think it should be `slice[int | None, int | None, int]`

Dreamsorcerer:
> Is the type for BaseRequest.http_range correct? I think it should be `slice[int | None, int | None, int]`

That's been confused by typeshed changing None -> Any, so no type errors occur. Feel free to update in a separate PR.
https://github.com/python/typeshed/blob/9a0dd0934a1293a8e91cc46dbbd182698f760f55/stdlib/builtins.pyi#L1103-L1126

Dreamsorcerer:
> I have resolved the conflicts. I would greatly appreciate if you could take a look at the failing test.

Think that test is fixed, but other tests are failing as well.

Dreamsorcerer:
> It is ready

Still marked as draft?

Derkades:
Ah OK, my type checker (basedpyright) does complain.

I have taken yet another look at the failing test (test_file_response_sends_headers_immediately). After staring at it for an hour I think there is a deeper issue.

This part is present in master but not in my branch for whatever reason:
```
with suppress(OSError):
            # fstat() may not be available on all platforms
            # Once we open the file, we want the fstat() to ensure
            # the file has not changed between the first stat()
            # and the open().
            st = os.stat(fobj.fileno())
```

This is the reason why the test succeeds in master. Path.stat is mocked but os.stat is not. The function then continues with whatever the size of fileno == 1 happens to be.

It then takes this path to call prepare()
```
        # If we are sending 0 bytes calling sendfile() will throw a ValueError
        if count == 0 or must_be_empty_body(request.method, status):
```

That is why prepare() is called in master but not in this branch. That must be an accident?

Derkades:
Ah yes, try this in the master branch:

* Add only `print('hello')` to test_file_response_sends_headers_immediately
* Observe that the test now fails because fileno 1 (stdout) is no longer zero in size!!

Derkades:
I have restored the second stat. There should probably a be a test for that, I will work on it :)

All tests pass locally for me.

Dreamsorcerer:
> Ah OK, my type checker (basedpyright) does complain.

I mean the mistake in http_range() does not produce an error because typeshed converts the inner type to Any.

Dreamsorcerer:
> All tests pass locally for me.

You'll need them passing in the CI though. Think some failures are only happening on certain platforms.

